Issue #3390, opened 2017-10-26
Bond Interface monitoring


ildarsv:
## Feature Request
Bond interface status monitoring
### Proposal:

There is no way to monitor bond interfaces status in the telegraf.
I'm starting develop the Bond input plugin. It will gather bond interfaces status from `/proc/net/bonding/*`.
Maybe somebody has an opinion regarding this approach?
### Current behavior:
There are no bond interface status metrics.
### Desired behavior:
To have some bond interface metrics.

danielnelson:
What bond modes are you interested in?  I have only ever used active-backup, in which case it could be nice to monitor which interfaces are up and how many times they have failed.

svetlovildar:
I am interested in active-backup and balance-rr. But anyway, I guess slave status and count of failures are good to monitor for every mode.

svetlovildar:
#3424

ildarsv:
Hello @danielnelson ! I've prepared pull request for this plugin, could you please take a look at this?

PR #3424, opened 2017-11-03
Add Bond Interface input plugin

ildarsv:
### Required for all PRs:


plugins/inputs/bond/bond.go
```diff
@@ -0,0 +1,172 @@
...
+	"strconv"
+	"strings"
+
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/plugins/inputs"
+)
+
+// default bond directory path
+const (
+	BOND_PATH = "/proc/net/bonding"
```
danielnelson: I would say call this `defaultBondPath` because Go naming guidelines, but to match existing style we should only specify the proc location.  We can add it to the config as `host_proc = "/proc"`, but we should also load it from the `HOST_PROC` environment variable.

plugins/inputs/bond/bond.go
```diff
@@ -0,0 +1,172 @@
...
+	// load path, get default value if config value and env variables are empty;
+	// list bond interfaces from bonding directory or gather all interfaces.
+	err := bond.listInterfaces()
+	if err != nil {
+		return err
+	}
+	for _, bondName := range bond.BondInterfaces {
+		file, err := ioutil.ReadFile(bond.BondPath + "/" + bondName)
+		if err != nil {
+			acc.AddError(fmt.Errorf("E! error due inspecting '%s' interface: %v", bondName, err))
```
danielnelson: The log level `E! ` will be added by the accumulator, so just `"error inspecting '%s' interface: %v"`, only add a log level if directly using the logger.  Check for this throughout pull request.

plugins/inputs/bond/bond.go
```diff
@@ -0,0 +1,172 @@
...
+	return nil
+}
+
+func (bond *Bond) gatherBondPart(bondName string, rawFile string, acc telegraf.Accumulator) error {
+	fields := make(map[string]interface{})
+	tags := map[string]string{
+		"bond": bondName,
+	}
+
+	lines := strings.Split(rawFile, "\n")
```
danielnelson: Use a `bufio.Scanner` to avoid EOL encoding issues.

plugins/inputs/bond/bond.go
```diff
@@ -0,0 +1,172 @@
...
+	for _, line := range lines {
+		stats := strings.Split(line, ":")
+		if len(stats) < 2 {
+			continue
+		}
+		name := strings.ToLower(strings.Replace(strings.TrimSpace(stats[0]), " ", "_", -1))
```
danielnelson: Why lowercase and replace spaces with underscores?

plugins/inputs/bond/bond.go
```diff
@@ -0,0 +1,172 @@
...
+		}
+	}
+	return fmt.Errorf("E! Couldn't find status info for '%s' ", bondName)
+}
+
+func (bond *Bond) gatherSlavePart(bondName string, rawFile string, acc telegraf.Accumulator) error {
+	var slave string
+	var status int
+
+	lines := strings.Split(rawFile, "\n")
```
danielnelson: `bufio.Scanner`

plugins/inputs/bond/bond.go
```diff
@@ -0,0 +1,172 @@
...
+		}
+	}
+	return nil
+}
+
+func (bond *Bond) listInterfaces() error {
+	if bond.BondPath == "" {
+		bond.BondPath = BOND_PATH
+	}
+	if len(bond.BondInterfaces) == 0 {
```
danielnelson: Don't cache the bond interfaces, or we won't be able to see new bonds without restarting Telegraf.  Return the paths instead of adding them to the struct.

plugins/inputs/bond/README.md
```diff
@@ -0,0 +1,81 @@
...
+```
+[[inputs.bond]]
+  ## Sets bonding directory path
+  ## If not specified, then default is:
+  bond_path = "/proc/net/bonding"
+
+  ## By default, telegraf gather stats for all bond interfaces
+  ## Setting interfaces will restrict the stats to the specified
+  ## bond interfaces.
+  bond_interfaces = ["bond0", "bond1"]
```
danielnelson: Comment this line out with a single `#`, so the default will be used.

plugins/inputs/bond/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+### Example output:
+
+Configuration:
+
+```
+[[inputs.bond]]
+  ## Sets bonding directory path
+  ## If not specified, then default is:
+  bond_path = "/proc/net/bonding"
```
danielnelson: Comment out since this is default.

plugins/inputs/bond/bond_test.go
```diff
@@ -0,0 +1,77 @@
...
+Bonding Mode: fault-tolerance (active-backup)
+Primary Slave: eth2 (primary_reselect always)
+Currently Active Slave: eth2
+MII Status: up
+MII Polling Interval (ms): 100
+Up Delay (ms): 0
+Down Delay (ms): 0
+
+Slave Interface: eth3
+MII Status: down
```
danielnelson: It's also possible for both of the interfaces to be up, right?  Maybe we should gather the active slave on the bond master?

plugins/inputs/bond/README.md
ildarsv: done

plugins/inputs/bond/README.md
ildarsv: done

plugins/inputs/bond/bond.go
ildarsv: done

plugins/inputs/bond/bond.go
ildarsv: done

plugins/inputs/bond/bond.go
ildarsv: done

plugins/inputs/bond/bond.go
ildarsv: done

plugins/inputs/bond/bond_test.go
ildarsv: Sure, I've added active_slave field. Thanks for the note.

plugins/inputs/bond/bond.go
ildarsv: done

plugins/inputs/bond/bond.go
ildarsv: It's too excessive, my bad. Removed.

ildarsv:
Hello @danielnelson ! Thank you for review. I've made some changes and improvements. Please take a look when you have time.

plugins/inputs/bond/bond.go
```diff
@@ -0,0 +1,218 @@
...
+type Bond struct {
+	HostProc       string   `toml:"host_proc"`
+	BondPath       string   `toml:"bond_path"`
+	BondInterfaces []string `toml:"bond_interfaces"`
+}
+
+var sampleConfig = `
+	## Sets 'proc' directory path
+	## If not specified, then default is /proc
+	# host_proc = "/proc"
```
danielnelson: Fix indention

plugins/inputs/bond/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+```toml
+[[inputs.bond]]
+  ## Sets 'proc' directory path
+  ## If not specified, then default is /proc
+  # host_proc = "/proc"
+
+  ## Sets 'bonding' directory relative path
+  ## If not specified, then default is /net/bonding
+  # bond_path = "/net/bonding"
```
danielnelson: Would this ever not be "/net/bonding"?

ildarsv: Seems no, would you like me to hardcode this parameter?

danielnelson: Yeah, will make the interface simpler which is always a good thing.

ildarsv: done

plugins/inputs/bond/bond.go
ildarsv: done

danielnelson:
Merged for 1.5, thanks!

ildarsv:
Thank you!
#3390 can be closed now

PR #3427, opened 2017-11-05
Add MessagePack output data format

dialogbox:
### Required for all PRs:

In addition to the current output data formats, I added a new binary output data format `msgpack`.

### Purporse:

To define msgpack messages and add output data format to support it.

### Current behavior:

Only text based output is supported.

### Desired behavior:

New format `msgpack` generates MessagePack encoded metrics as output.

### Use case:

Even the text based formats are easy to use, these are not the most efficient to parse and process. MessagePack can be used to define language-neutral, platform-neutral binary data encoding and the result can be easily encoded, decoded with many built-in safety checks. MessagePack is an efficient binary serialization format. It lets you exchange data among multiple languages like JSON. But it's faster and smaller.

https://msgpack.org/index.html

And it's already used by influxdb.

https://github.com/influxdata/influxdb/blob/1a920c5c420d2e7afcfb3e8781fb8dfbe557fcf9/services/httpd/response_writer.go#L196


Jaeyo:
any update?

Jaeyo:
@danielnelson is there any reason to holding? or is it enough to merge upstream?

sjwang90:
@Jaeyo Would you be able to look submitting this as an [external plugin](https://github.com/influxdata/telegraf/blob/master/docs/EXTERNAL_PLUGINS.md) and use it with that can be used with `processors.execd` to run seamlessly with Telegraf? 

Due to the level of interest around this output format this would be the better route. Telegraf users would be able to use the plugin sooner by having it as an external plugin since it wouldn't have to go through the typical review process.

dialogbox:
@sjwang90 I haven't touched telegraf quite a while but if my memory serves, the purpose of this PR was adding a serialization format. By doing this, users can send serialized msgpack data to any output plugins.
However, the external plugin system only supports inputs, processors and outputs type of plugin. Can we archive this with any of those external plugin types?

[processor](https://github.com/influxdata/telegraf/tree/master/plugins/processors/execd) can not be the option because it's output format must be the influx line protocol.

> The programs must accept influx line protocol on standard in (STDIN) and output metrics in influx line protocol to standard output (STDOUT).

I'm not sure telegraf supports chaining of output plugin. If it doesn't, there is no way to decouple serialization format and the output with external plugin.

One possible solution is creating a simple cli command which converts influxlp to msgpack(or any formats) and pipelining it with other output plugin command by creating a simple shell script. And using the script as an external output plugin. However, it looks like there is no way to do this with internal output plugins.

Once again, it's been quite a while since I created this PR. I've just skimmed the code again so I might be totally wrong. Please correct me if I'm wrong.

srebhan:
@dialogbox this is an act of art! If you are still interested in getting this PR in, please rebase and I will review.

dialogbox:
> @dialogbox this is an act of art! If you are still interested in getting this PR in, please rebase and I will review.

Thank you for the reply. I already deleted my fork so I had to open an another PR.

https://github.com/influxdata/telegraf/pull/8828

I did rebase and implemented some improvements which I missed in the original PR. Please review the new PR.

Issue #3408, opened 2017-10-30
Postgresql output plugin (for using with timescaledb)


leiserfg:

hlihhovac:
+1

svenklemm:
I've written an output plugin for postgresql. Let me know if you need any other features.

PR #3428

leiserfg:
It's, cool!. Can you add a more flexible way of creating the tables? For example: when creating the tables for using with timescaledb, after the table creation, but before the first insertion, you need to execute a command that create the table partition. By example:

```sql
CREATE TABLE conditions (
  time        TIMESTAMPTZ       NOT NULL,
  location    TEXT              NOT NULL,
  temperature DOUBLE PRECISION  NULL,
  humidity    DOUBLE PRECISION  NULL
);
SELECT create_hypertable('conditions', 'time');
```   
Another use case: using pg 10 and want to use a automatic partitioned table or citusdb with an sharded table.

I know that I can create the tables manually, but if I need a lot of metrics then it will imply a lot of repetitive sql writing.

leiserfg:
Maybe a template parameter for table creation can solve this.

svenklemm:
I guess a template is most flexible as a post_create_hook wouldnt help much with pg10 partitioning.

danielnelson:
I like the template idea, and we won't want to add an output which is not generally usable in many use cases.  If you design a template configuration please post it here for comments.

svenklemm:
I was thinking about having the following placeholders:
TABLE - tablename quoted as identifier
COLUMNS - comma separated list of columns with datatype
PK_COLUMNS - comma separated list of primary key columns (time+ all tags)
TABLENAME - tablename quoted as literal to use in function calls (required for timescale)

The template syntax is just a suggestion but curly braces shouldnt conflict with any SQL.

### Example Templates

```sql
-- Default Value
CREATE TABLE {TABLE}({COLUMNS},PRIMARY KEY({PK_COLUMNS}));

-- PostgreSQL 10 Partitioning by year and month
CREATE TABLE {TABLE}({COLUMNS}) PARTITION BY RANGE (EXTRACT(YEAR FROM time),EXTRACT(MONTH FROM time));

-- Timescale
CREATE TABLE {TABLE}({COLUMNS});
SELECT create_hypertable({TABLENAME}, 'time');
```

danielnelson:
Would a user need to design tables for every input?

leiserfg:
Awesome!

[...]

PR #3428, opened 2017-11-05
PostgreSQL output plugin

svenklemm:
This is an output plugin to store metrics in a postgresql database. It will create the tables required if they do not yet exist. Table names are derived from the metric name.

#3408 


cwadley:
@svenklemm Any update on the status of this PR? As per #3408 it seems like it might be ready to merge. Maybe add @danielnelson as a reviewer? I'd love to use this plugin.

svenklemm:
@cwadley I'm using it for a couple weeks now without problems but it could use more testing and proper integration tests.

cwadley:
@svenklemm I'll get to work :)

otherpirate:
Hi folks,

First things first, thanks @svenklemm I learned a lot with your code

I think this PR #3912 is a better/faster solution for large number of metrics.

saaros:
I wrote an experimental batching mode on top of this branch, see the commits in https://github.com/saaros/telegraf/tree/postgres_output_batching

This speeds up the processing of 10k metric batches 5-10x on my test environment when running locally and even more when telegraf & pg are running on separate hosts.

otherpirate:
Hi @saaros,

I did some pretty similar, but even faster using COPY command, you can check it in here #3912

saaros:
Thanks @otherpirate -- I had a look at your PR, but since it didn't have any DDL support I figured it's faster to add batching to this version than DDL to the other one.

Would be great to get something in good enough shape to be merged to master.

svenklemm:
I've merged the batch changes from @saaros into this PR. 
@danielnelson could you give any feedback whats missing to get this merged?

danielnelson:
@svenklemm Thanks for the heads up, it might take awhile yet but we will at this plugin as well as #3408, #3912, and #4205

svenklemm:
@danielnelson any update on this?

glinton:
oops, that review was for the most recent changes

plugins/outputs/postgresql/postgresql.go
```diff
@@ -0,0 +1,414 @@
...
+	query = strings.Replace(query, "{TABLELITERAL}", quoteLiteral(metric.Name()), -1)
+	query = strings.Replace(query, "{COLUMNS}", strings.Join(columns, ","), -1)
+	query = strings.Replace(query, "{KEY_COLUMNS}", strings.Join(pk, ","), -1)
+
+	sql = append(sql, query)
+	return strings.Join(sql, ";")
+}
+
+func (p *Postgresql) generateInsert(tablename string, columns []string) string {
+
```
glinton: nitpick: remove this line

plugins/outputs/postgresql/postgresql.go
```diff
@@ -0,0 +1,414 @@
...
+	var placeholder, quoted []string
+	for i, column := range columns {
+		placeholder = append(placeholder, fmt.Sprintf("$%d", i+1))
+		quoted = append(quoted, quoteIdent(column))
+	}
+
+	sql := fmt.Sprintf("INSERT INTO %s(%s) VALUES(%s)", quoteIdent(tablename), strings.Join(quoted, ","), strings.Join(placeholder, ","))
```
glinton: nitpick: just `return fmt.Sprintf(...`

plugins/outputs/postgresql/postgresql.go
```diff
@@ -0,0 +1,414 @@
...
+		if p.Tables[tablename] == false && p.tableExists(tablename) == false {
+			createStmt := p.generateCreateTable(metric)
+			_, err := p.db.Exec(createStmt)
+			if err != nil {
+				return err
+			}
+			p.Tables[tablename] = true
+		}
+
+		var columns []string
```
glinton: initialize `columns` with `"time"` as a value to skip line 231's `append`

plugins/outputs/postgresql/postgresql.go
```diff
@@ -0,0 +1,414 @@
...
+		var values []interface{}
```
glinton: initialize `values` with `metric.Time()` as a value to skip line 232's `append`

plugins/outputs/postgresql/postgresql.go
```diff
@@ -0,0 +1,414 @@
...
+
+				var where_parts []string
+				for i, column := range where_columns {
+					where_parts = append(where_parts, fmt.Sprintf("%s = $%d", quoteIdent(column), i+1))
+				}
+				query := fmt.Sprintf("SELECT tag_id FROM %s WHERE %s", quoteIdent(tablename+p.TagTableSuffix), strings.Join(where_parts, " AND "))
+
+				err := p.db.QueryRow(query, where_values...).Scan(&tag_id)
+				if err != nil {
+					// log.Printf("I! Foreign key reference not found %s: %v", tablename, err)
```
glinton: Remove commented code

plugins/outputs/postgresql/postgresql.go
```diff
@@ -0,0 +1,414 @@
...
+			i += len(params[table_and_cols]) * len(columns)
+			placeholder = append(placeholder, fmt.Sprintf("$%d", i+1))
+		}
+		params[table_and_cols] = append(params[table_and_cols], strings.Join(placeholder, ","))
+		colmap[table_and_cols] = columns
+		tabmap[table_and_cols] = tablename
+	}
+
+	for table_and_cols, values := range batches {
+		// log.Printf("Writing %d metrics into %s", len(params[table_and_cols]), table_and_cols)
```
glinton: Remove commented code

markusr:
@danielnelson @glinton 
Can you merge this? I would like to have this feature.

neographikal:
Nice work, would like to have this very much :)

blagojts:
Hey guys, we updated the code a bit. Added more unit tests, and created integration tests. 
Can you take a look again?

RobAtticus:
Hi @danielnelson @glinton, just wondering if there is any more we can do on our end to move this along?

danielnelson:
I'll try to look soon, one question comes to mind now though:  what does the plugin do if the same data is written twice?

blagojts:
> I'll try to look soon, one question comes to mind now though: what does the plugin do if the same data is written twice?

By default the table being created doesn't have any unique constraints. If the user modifies the `table_template` config property so that a unique constraint is also added, then the batch containing the duplicate data will fail, subsequent batches won't be affected

droidnxs:
Just something I noticed when using this plugin - I see multiple records for a single measurement per group, per-metric for a single timestamp. The input plugin is a shell script that outputs influx format as follows:
```
iostat,device=dm-1 readsPerSec=23.00
iostat,device=dm-1 writesPerSec=76.00
iostat,device=dm-1 avgRqSz=15.03
iostat,device=dm-1 avgQuSz=0.66
iostat,device=dm-1 percentUtil=61.20
```
However, in TimescaleDB / PostgreSQL I see 5 records for each metric:
image
Shouldn't these be on a single row? The current behaviour would end up significantly consuming more space. I guess due to this behaviour the inserts are broken down as well, and affects database performance too!

mcassaniti:
Just curious on the following:

- Are the timestamps in TimescaleDB different? I know from the screen shot you've sent that they appear the same, but do they differ.
- What happens when you send a measurement with multiple fields rather than one at a time?

svenklemm:
This is expected, if you send them in 1 line they will produce 1 row:
```
iostat,device=dm-1 readsPerSec=23.00,writesPerSec=76.00,avgRqSz=15.03,avgQuSz=0.66,percentUtil=61.20
```

kaleemullah360:
Why the plugin for `TimescaleDB`, can't use it for `InfluxDB` ? i.e send data stored in `InfluxDB` to `PostgreQSL`

droidnxs:
@mcassaniti @svenklemm 
Thanks for the help! I converted the script to output all metrics for each group on the same line and it worked! 

@kaleemullah360 
That's not a reliable solution. I can't have an influxdb just for the sake of these measurements, not to mention the additional maintenance and infrastructure. :)

1848:
Good to see timescaledb is coming.

I tested this code and got warnings about dropped metrics and missed flush intervals from telegraf when I had lots of input data.
From what I could see (I am not a golang guy so bear with me ;)) you are doing an INSERT (with auto commit) for each row, is that correct?
If so, did you consider batching up multiple rows into a transaction (to not use auto commit) or using the PostgreSQL *COPY* command?

blagojts:
@1848 thank you for trying out our plugin. We are using batches, but currently the batches consist of insert statements. We agree that we can improve performance by using COPY, and we'll try to issue an update soon. Can you tell us at what volume did you start getting the warnings?

1848:
Only did a short test to see how it performs against influxdb, cant remember to accurate numbers but I think I measured ~3000 commits/second in the database and roughly the same amount was dropped by telegraf.
Will test again when the update is released.
Thanks.

otherpirate:
Hi @1848 
PR #3912 implements Copy

1848:
@otherpirate thanks, will try to test on monday.

blagojts:
Hello people (@1848 and @otherpirate), inspired by your comments we updated our code and introduced a bunch of improvements
1. Batching is now implemented with COPY
2. Added tagID caching. When a users selects to keep the tags in a separate table we lost a lot of performance looking up the tagID to reference it in the main metrics table, so now we have a per-measurement configurable cache that keeps the LRU tag sets. 
3. Schema updates are done once per a batch of metrics. 
You can see the update here (https://github.com/svenklemm/telegraf/pull/5). Once the review is complete we'll update this PR with the changes also. 
 
Our initial tests show that we can now insert 10k telegraf.Metric(s) in ~90ms instead of the previous ~13 seconds. This is the speed we get when all the tags and fields are kept in one table. Converting them to JSON increases the time to ~130ms for 10k.

The time for inserting 10k metrics when the Tags are stored in a separate table is ~10s with a cardinality of 3000 (max possible tag combinations), but once the cache is populated subsequent writes are much faster (reaching <100ms again if the cache size is > cardinality)

e.g. A metric with 5 tags, each with 5 independent values results in a tagset cardinality 3125

1848:
@otherpirate could not test your solution because I need the tables to be created automatically (too many).

@blagojts very nice, works fine so far. I have only two little remarks (left a comment on the pull request).

blagojts:
The optimizations are now merged. Also a few other fixes went in. To summarize 
1. Batching is now implemented with COPY
2. Added tagID caching. When a users selects to keep the tags in a separate table we lost a lot of performance looking up the tagID to reference it in the main metrics table, so now we have a per-measurement configurable cache that keeps the LRU tag sets.
3. Schema updates are done once per a batch of metrics.
4. Reconnect to database if connection was lost

Our testing:
* Keep the tags in the same table as the fields = ~85ms for 10k telegraf.Metric(s)
* Keep the tags in a separate table with cache = ~13s for 10k telegraf.Metrics on the first insert (if all the metrics have different tags/no repeats), subsequent writes goes down to ~100ms
* Converting the tags or fields to JSON adds ~20ms on every 10k metrics

Testing done with 10K metrics, each had 10 fields, and 5 tags. Total possible different tags 5^5 = 3125.  
**Cache size is configurable**

1848:
@blagojts 
Tests worked fine, got good performance.
Now have again two feature requests (maybe for a future version?).

1) Add possibility to sanitize names. Had to escape some table names because the metric had upper chars in it. Maybe this could be done in a separate (processor?) plugin. Or is it already possible?

2) Had some trouble with the size of the database. I am already using a btrfs with compression (around 4x less space) but that is not good enough. So I was thinking to write the data as jsonb because there is another extension (zson) around which is able to compress jsonb. Did you test it already? Maybe its worth to include it as an option. zson needs to be retrained after some time which could be done automatically by your plugin...

russorat:
@svenklemm @blagojts could you take a look and address any test failures in this plugin?

blagojts:
@danielnelson Thank you for your extensive review. We have a few more commits with the following changes:
1. JSON support removed to simplify the plugin
2. Write first, if it fails try to recover (schema change)
3. Using slices of []telegraf.Metric instead of passing int indexes which metrics are needed
4. Templating removed, for TimescaleDB hypertables 2 other flags can be set
5. Mutex removed from connection checks

psura:
any update on this? I just came across this and it looks promising. Is someone going to resolve the merge request conflicts?

plugins/outputs/postgresql/tables/table_manager.go
```diff
@@ -0,0 +1,245 @@
+package tables
```
danielnelson: This file should be called `manager.go` to match the test file and because the containing package is already named tables.

I would reduce the number of packages personally, maybe just a single `schema` package in addition to the top package.

plugins/outputs/postgresql/tables/table_manager.go
```diff
@@ -0,0 +1,245 @@
...
+		"SELECT required.c as column_name, available.column_name IS NOT NULL as exists, available.data_type FROM required LEFT JOIN available ON required.c = available.column_name;"
+)
+
+type columnInDbDef struct {
+	dataType utils.PgDataType
+	exists   bool
+}
+
+func (c *columnInDbDef) String() string {
+	return fmt.Sprintf("{dType: %s, exists:%v}", c.dataType, c.exists)
```
danielnelson: Remove this function, for debugging you could always just print with `%+v`.

plugins/outputs/postgresql/README.md
```diff
@@ -0,0 +1,53 @@
+# PostgreSQL Output Plugin
+
+This output plugin writes all metrics to PostgreSQL. 
+The plugin manages the schema automatically updating missing columns, and checking if existing ones are of the proper type. 
+
```
danielnelson: Let's document the minimum version of Postgres and TimescaleDB the plugin supports.

plugins/outputs/postgresql/tables/table_manager.go
```diff
@@ -0,0 +1,245 @@
...
+		}
+	}
+
+	fullTableName := utils.FullTableName(t.schema, tableName).Sanitize()
+	return fmt.Sprintf(tagTableSQLTemplate, fullTableName, strings.Join(colDefs, ","))
+}
+
+// For a given table and an array of column names it checks the database if those columns exist,
+// and what's their data type.
+func (t *defTableManager) findColumnPresence(db db.Wrapper, tableName string, columns []string) ([]*columnInDbDef, error) {
```
danielnelson: I haven't checked if this is possible, but can we just attempt to alter all columns on the table?  This way we don't need to know what is in the database and it removes a race condition.

cameron-elliott:
@danielnelson thanks for working on this on a Friday night! (well, for some of us, anyway)
I can tell I am not the only one excited about this from the likes in @psura 's question.
So, a big thanks to the Influx team, and the PR team for the persistence and investment from both, we are so close, I can feel it! @blagojts @svenklemm

AtakanColak:
While trying to use this PostgreSQL plugin for my own SNMP Walker plugin, I got: 

` Error writing to outputs.postgresql: ERROR: invalid byte sequence for encoding "UTF8": 0x80 (SQLSTATE 22021)`

Web search implies there is a non-accepted character value in the values I'm sending, but when I did a brute force replace for 0x00 for all of the metrics, the result did not change.

rodo:
Thanks @svenklemm for this PR, I've started to look for the same feature, and won't write it as you did it. Hope to see this merged in telegraf V2.
My company is looking for a new monitoring system, as we have a we good postgresql DBA team, it will help a lot to not introduce a new technology in our data storage departement.

cameron-elliott:
> Thanks @svenklemm for this PR, I've started to look for the same feature, and won't write it as you did it. Hope to see this merged in telegraf V2.
> My company is looking for a new monitoring system, as we have a we good postgresql DBA team, it will help a lot to not introduce a new technology in our data storage departement.

@rodo I think you meant to say, "and because you wrote it, we won't need to write it, ourselves (thank goodness!)", but for some English speakers it can easily translate as "and I wouldn't write it the same way you would (I would do it different)", which can sound quite negative, but I don't think is what you meant to say.
That may be the reason for the downvote from negbie.
I hope this is helpful.

negbie:
@danielnelson any intension to push this forward and merge it?

mehdisadeghi:
I appreciate the work done by the contributors and the reviewers, but what is keeping this pull request from being accepted (apart from the obvious recent conflicts)?

negbie:
@mehdisadeghi I would say @danielnelson suggestions/requested changes from 23 Nov 2019

AtakanColak:
While I was using this plugin I came across a  ` duplicate key value violates unique constraint SQLSTATE 23505` error. Turns out when using with `tags_as_foreignkeys`, we must be really careful to not to reach the 63 byte table name limit with metric names. This is an issue that can be fixed with something like;

```
metricName := metric.Name()
if len(metricName) > 59 {
  metricName = metricName[:50] + "TLN" // TOO LONG NAME, truncate
}
metric.SetName(metricName)
```

I'm not sure where would be optimal for putting it, so wanted to share it instead of a pull request.

abhishekgetsb:
> While I was using this plugin I came across a ` duplicate key value violates unique constraint SQLSTATE 23505` error. Turns out when using with `tags_as_foreignkeys`, we must be really careful to not to reach the 63 byte table name limit with metric names. This is an issue that can be fixed with something like;
> 
> ```
> metricName := metric.Name()
> if len(metricName) > 59 {
>   metricName = metricName[:50] + "TLN" // TOO LONG NAME, truncate
> }
> metric.SetName(metricName)
> ```
> 
> I'm not sure where would be optimal for putting it, so wanted to share it instead of a pull request.

I'm kind of facing a similar issue since we recently downloaded the 1.13 unofficial release from TimescaleDB's website. Our logs somehow end up being duplicated by several orders magnitude in the database (sometimes up to **100 duplicates**), but the local files being tailed have only one instance per log event. 
Adding a constraint didn't mitigate the problem, because the postgresql output plugin uses COPY TO, and the database fails to insert the whole batch even if one log even had a constraint violation. We had to create an intermediate table where the duplicate data gets fed into, and added a trigger to insert the logs into the table with the constraint.
Any chance this is a regression caused by a recent change? I've haven't seen this occurring before 1.13.

droidnxs:
> While I was using this plugin I came across a ` duplicate key value violates unique constraint SQLSTATE 23505` error. Turns out when using with `tags_as_foreignkeys`, we must be really careful to not to reach the 63 byte table name limit with metric names. This is an issue that can be fixed with something like;
> 
> ```
> metricName := metric.Name()
> if len(metricName) > 59 {
>   metricName = metricName[:50] + "TLN" // TOO LONG NAME, truncate
> }
> metric.SetName(metricName)
> ```
> 
> I'm not sure where would be optimal for putting it, so wanted to share it instead of a pull request.

I'm kind of facing a similar issue since we recently downloaded the 1.13 unofficial release from TimescaleDB's website. Our logs somehow end up being duplicated by several orders magnitude in the database (sometimes up to **100 duplicates**), but the local files being tailed have only one instance per log event. 
Adding a constraint didn't mitigate the problem, because the postgresql output plugin uses COPY TO, and the database fails to insert the whole batch even if one log even had a constraint violation. We had to create an intermediate table where the duplicate data gets fed into, and added a trigger to insert the logs into the table with the constraint.
Any chance this is a regression caused by a recent change? I've haven't seen this occurring before 1.13.

plugins/outputs/postgresql/tags_cache.go
```diff
@@ -0,0 +1,164 @@
...
+	} else {
+		whereParts = make([]string, len(target.Names)-1)
+		whereValues = make([]interface{}, len(target.Names)-1)
+		whereIndex := 1
+		for columnIndex, tagName := range target.Names[1:] {
+			if val, ok := tags[tagName]; ok {
+				whereParts[columnIndex] = utils.QuoteIdent(tagName) + " = $" + strconv.Itoa(whereIndex)
+				whereValues[whereIndex-1] = val
+			} else {
+				whereParts[whereIndex-1] = tagName + " IS NULL"
```
Sarke, line 90:
```suggestion
				whereParts[whereIndex-1] = utils.QuoteIdent(tagName) + " IS NULL"
```

This case needs to be sanitized as well.  It failed on Docker input when trying on `AND org.opencontainers.image.description IS NULL`.

truijllo:
While I was using this plugin I came across an escaping issue, I have a json from kafka, one field is declared in the telegraf.con  as tag_key,  something like 

> {"name": "valle d\'aosta", "descr": "yadayada"}

now, putting "valle d\'aosta"  in the   *_tag table crashes for the unescaped single quote  ( for postgresql standard the proper escaping is a double single quote )

tk512:
Any news on merging this functionality? I see it has been a couple of years already. I think this functionality will be really useful to a lot of teams.

Sarke:
@tk512 I made an updated version for Telegraf 1.14 that you can try:

https://github.com/Sarke/telegraf/tree/1.14.0-with-pg

`docker pull sarkedev/telegraf-timescaledb:1.14.0-with-pg`

I've been running it for about a week now without issue.

brunokc:
Just wanted to add my support in getting this change merged. Learned about this work not long ago and just finished going through the history here, it's been in the making for a while. Thank you for the all the work so far. 

Is there anything still pending in review that needs to be addressed before a merge can happen?

AtakanColak:
As much as this plugin is quite useful, it uses up relatively high amount of CPU and is slow to send its data when compared to other output plugins. Perhaps some further work in performance optimisation would be desirable.

1848:
How about native timescaledb compression support?
Should not be too hard to implement imo and could be useful.

Does this plugin handle automatic deletion of data older than X time units?

tk512:
> Does this plugin handle automatic deletion of data older than X time units?

Correct me if I am wrong, but shouldn't a normal retention policy within TimescaleDB work out of the box?  Or do you mean a retention setting within the plugin?  That would be really cool.

1848:
> Correct me if I am wrong, but shouldn't a normal retention policy within TimescaleDB work out of the box? Or do you mean a retention setting within the plugin? That would be really cool.

Yes, correct and correct ;)

phemmer:
I'm using the version of this output plugin provided by TimescaleDB, (specifically `telegraf-1.13.0~with~pg-1.x86_64.rpm`), and it seems that it has some sort of out-of-control memory usage when connections fail. The issue leads to a panic when it runs out of memory.

The interesting thing of note is that this only seems to happen if the connection setup fails. If the connection setup is good, the memory usage is normal. In my case I was setting this output plugin up, had telegraf going through a proxy to get to postgres, and the proxy was terminating the connection immediately after establishing it.

```
fatal error: runtime: out of memory
runtime stack:
runtime.throw(0x27e69fd, 0x16)
        /usr/local/go/src/runtime/panic.go:774 +0x72
runtime.sysMap(0xc004000000, 0x58000000, 0x45881b8)
        /usr/local/go/src/runtime/mem_linux.go:169 +0xc5
runtime.(*mheap).sysAlloc(0x4564c00, 0x54546000, 0x0, 0x0)
        /usr/local/go/src/runtime/malloc.go:701 +0x1cd
runtime.(*mheap).grow(0x4564c00, 0x2a2a3, 0xffffffff)
        /usr/local/go/src/runtime/mheap.go:1252 +0x42
runtime.(*mheap).allocSpanLocked(0x4564c00, 0x2a2a3, 0x45881c8, 0x7fd8d0e98d50)
        /usr/local/go/src/runtime/mheap.go:1163 +0x291
runtime.(*mheap).alloc_m(0x4564c00, 0x2a2a3, 0x101, 0x3)
        /usr/local/go/src/runtime/mheap.go:1015 +0xc2
runtime.(*mheap).alloc.func1()
        /usr/local/go/src/runtime/mheap.go:1086 +0x4c
runtime.(*mheap).alloc(0x4564c00, 0x2a2a3, 0xc000010101, 0xc000000180)
        /usr/local/go/src/runtime/mheap.go:1085 +0x8a
runtime.largeAlloc(0x5454502b, 0x4550101, 0x7fd8db196301)
        /usr/local/go/src/runtime/malloc.go:1138 +0x97
runtime.mallocgc.func1()
        /usr/local/go/src/runtime/malloc.go:1033 +0x46
runtime.systemstack(0xc00013e480)
        /usr/local/go/src/runtime/asm_amd64.s:370 +0x66
runtime.mstart()
        /usr/local/go/src/runtime/proc.go:1146

goroutine 1 [running]:
runtime.systemstack_switch()
        /usr/local/go/src/runtime/asm_amd64.s:330 fp=0xc00051ecb0 sp=0xc00051eca8 pc=0x45f2a0
runtime.mallocgc(0x5454502b, 0x21613c0, 0x1, 0x4d5f61)
        /usr/local/go/src/runtime/malloc.go:1032 +0x895 fp=0xc00051ed50 sp=0xc00051ecb0 pc=0x40e605
runtime.makeslice(0x21613c0, 0x5454502b, 0x5454502b, 0x2000)
        /usr/local/go/src/runtime/slice.go:49 +0x6c fp=0xc00051ed80 sp=0xc00051ed50 pc=0x447c4c
github.com/influxdata/telegraf/vendor/github.com/jackc/pgx/chunkreader.(*ChunkReader).newBuf(...)
        /go/src/github.com/influxdata/telegraf/vendor/github.com/jackc/pgx/chunkreader/chunkreader.go:86
github.com/influxdata/telegraf/vendor/github.com/jackc/pgx/chunkreader.(*ChunkReader).Next(0xc000205d00, 0x5454502b, 0xc000210000, 0x5, 0x2000, 0x0, 0x0)
        /go/src/github.com/influxdata/telegraf/vendor/github.com/jackc/pgx/chunkreader/chunkreader.go:57 +0x31f fp=0xc00051ee08 sp=0xc00051ed80 pc=0x166f7af
github.com/influxdata/telegraf/vendor/github.com/jackc/pgx/pgproto3.(*Frontend).Receive(0xc0001f2c00, 0xc00027a001, 0x23, 0x40, 0x23)
        /go/src/github.com/influxdata/telegraf/vendor/github.com/jackc/pgx/pgproto3/frontend.go:122 +0x8e fp=0xc00051ee70 sp=0xc00051ee08 pc=0x16758de
github.com/influxdata/telegraf/vendor/github.com/jackc/pgx.(*Conn).rxMsg(0xc000114800, 0xc00027a000, 0x23, 0x40, 0x23)
        /go/src/github.com/influxdata/telegraf/vendor/github.com/jackc/pgx/conn.go:1788 +0x4f fp=0xc00051eed0 sp=0xc00051ee70 pc=0x170f11f
github.com/influxdata/telegraf/vendor/github.com/jackc/pgx.(*Conn).connect(0xc000114800, 0xc00030daa6, 0x9, 0x1538, 0xc00030dac1, 0x7, 0xc00030dace, 0x3, 0xc00030dadb, 0x18, ...)
        /go/src/github.com/influxdata/telegraf/vendor/github.com/jackc/pgx/conn.go:565 +0x721 fp=0xc00051f168 sp=0xc00051eed0 pc=0x1706241
github.com/influxdata/telegraf/vendor/github.com/jackc/pgx.connect(0xc00030daa6, 0x9, 0x1538, 0xc00030dac1, 0x7, 0xc00030dace, 0x3, 0xc00030dadb, 0x18, 0xc00013ec00, ...)
        /go/src/github.com/influxdata/telegraf/vendor/github.com/jackc/pgx/conn.go:418 +0xe7a fp=0xc00051f648 sp=0xc00051f168 pc=0x1704e9a
github.com/influxdata/telegraf/vendor/github.com/jackc/pgx.Connect(...)
        /go/src/github.com/influxdata/telegraf/vendor/github.com/jackc/pgx/conn.go:343
github.com/influxdata/telegraf/plugins/outputs/postgresql/db.NewWrapper(0xc00030daa1, 0x52, 0x8, 0xc00051f7a8, 0x504fca, 0x0)
        /go/src/github.com/influxdata/telegraf/plugins/outputs/postgresql/db/db_wrapper.go:32 +0xd9 fp=0xc00051f810 sp=0xc00051f648 pc=0x1e6c0a9
github.com/influxdata/telegraf/plugins/outputs/postgresql.(*Postgresql).Connect(0xc0001991e0, 0x28, 0xc00051f968)
        /go/src/github.com/influxdata/telegraf/plugins/outputs/postgresql/postgresql.go:57 +0x4c fp=0xc00051f8d0 sp=0xc00051f810 pc=0x1e6fa5c
github.com/influxdata/telegraf/agent.(*Agent).connectOutputs(0xc0000100b0, 0x2c98be0, 0xc000205040, 0x0, 0x0)
        /go/src/github.com/influxdata/telegraf/agent/agent.go:657 +0x1e3 fp=0xc00051f9a8 sp=0xc00051f8d0 pc=0x966803
github.com/influxdata/telegraf/agent.(*Agent).Run(0xc0000100b0, 0x2c98be0, 0xc000205040, 0x1, 0x1)
        /go/src/github.com/influxdata/telegraf/agent/agent.go:49 +0x251 fp=0xc00051faa0 sp=0xc00051f9a8 pc=0x9630a1
main.runAgent(0x2c98be0, 0xc000205040, 0x45856e0, 0x0, 0x0, 0x45856e0, 0x0, 0x0, 0x0, 0x0)
        /go/src/github.com/influxdata/telegraf/cmd/telegraf/telegraf.go:202 +0x8da fp=0xc00051fbc0 sp=0xc00051faa0 pc=0x1ed3a7a
main.reloadLoop(0xc000170000, 0x45856e0, 0x0, 0x0, 0x45856e0, 0x0, 0x0, 0xc00039bd28, 0x0, 0x0, ...)
        /go/src/github.com/influxdata/telegraf/cmd/telegraf/telegraf.go:110 +0x270 fp=0xc00051fcb0 sp=0xc00051fbc0 pc=0x1ed30b0
main.main()
        /go/src/github.com/influxdata/telegraf/cmd/telegraf/telegraf.go:427 +0x59f fp=0xc00051ff60 sp=0xc00051fcb0 pc=0x1ed496f
runtime.main()
        /usr/local/go/src/runtime/proc.go:203 +0x21e fp=0xc00051ffe0 sp=0xc00051ff60 pc=0x432b6e
runtime.goexit()
        /usr/local/go/src/runtime/asm_amd64.s:1357 +0x1 fp=0xc00051ffe8 sp=0xc00051ffe0 pc=0x461371

goroutine 34 [syscall]:
os/signal.signal_recv(0x0)
        /usr/local/go/src/runtime/sigqueue.go:147 +0x9c
os/signal.loop()
        /usr/local/go/src/os/signal/signal_unix.go:23 +0x22
created by os/signal.init.0
        /usr/local/go/src/os/signal/signal_unix.go:29 +0x41

goroutine 51 [select]:
github.com/influxdata/telegraf/vendor/go.opencensus.io/stats/view.(*worker).start(0xc000204240)
        /go/src/github.com/influxdata/telegraf/vendor/go.opencensus.io/stats/view/worker.go:150 +0xdd
created by github.com/influxdata/telegraf/vendor/go.opencensus.io/stats/view.init.0
        /go/src/github.com/influxdata/telegraf/vendor/go.opencensus.io/stats/view/worker.go:29 +0x57

goroutine 25 [select]:
main.reloadLoop.func1(0xc000170060, 0xc000180000, 0xc0002964d0, 0xc000170000)
        /go/src/github.com/influxdata/telegraf/cmd/telegraf/telegraf.go:97 +0xaf
created by main.reloadLoop
        /go/src/github.com/influxdata/telegraf/cmd/telegraf/telegraf.go:96 +0x207
```

phemmer:
Just ran across another issue.
Some inputs on telegraf take advantage of the fact that if you send multiple points with different fields to the same series with the same timestamp, InfluxDB will automatically merge them to the same point.
For example:
```
mymeasurement,mytag=x foo=1 1234567890
mymeasurement,mytag=x bar=2 1234567890
```
Will effectively be merged together as
```
mymeasurement,mytag=x foo=1,bar=2 1234567890
```

When the postgresql output sends these, they're getting inserted as separate records.
For example, one input which does this is the `cpu` input. Here's an example config:
```
[[inputs.cpu]]
 name_override = "cputest"
 collect_cpu_time = true
[[outputs.postgresql]]
 connection = "host=localhost port=5432 dbname=x user=telegraf password=y"
 table_template = "CREATE TABLE {TABLE}({COLUMNS}); SELECT create_hypertable({TABLELITERAL},'time');"
```
And if you query the data in postgres, it looks like this:
```
epitome=# select time, cpu, time_user, usage_user from cputest;
          time          |    cpu    | time_user |     usage_user      
------------------------+-----------+-----------+---------------------
 2020-05-19 04:24:40+00 | cpu-total |    954.28 |                    
 2020-05-19 04:24:40+00 | cpu1      |    450.14 |                    
 2020-05-19 04:24:40+00 | cpu0      |    504.14 |                    
 2020-05-19 04:24:50+00 | cpu-total |           | 0.40000000000020464
 2020-05-19 04:24:50+00 | cpu-total |    954.36 |                    
 2020-05-19 04:24:50+00 | cpu1      |           |   0.501002004007604
 2020-05-19 04:24:50+00 | cpu1      |    450.19 |                    
 2020-05-19 04:24:50+00 | cpu0      |           | 0.20000000000038654
 2020-05-19 04:24:50+00 | cpu0      |    504.16 |                    
 2020-05-19 04:25:00+00 | cpu-total |           | 0.40000000000020464
 2020-05-19 04:25:00+00 | cpu-total |    954.44 |                    
 2020-05-19 04:25:00+00 | cpu1      |           | 0.49950049950014835
 2020-05-19 04:25:00+00 | cpu1      |    450.24 |                    
 2020-05-19 04:25:00+00 | cpu0      |           |  0.4008016032056275
 2020-05-19 04:25:00+00 | cpu0      |     504.2 |                    
 2020-05-19 04:25:10+00 | cpu-total |           |  0.5505505505505631
 2020-05-19 04:25:10+00 | cpu-total |    954.55 |                    
 2020-05-19 04:25:10+00 | cpu1      |           |   0.800800800801388
 2020-05-19 04:25:10+00 | cpu1      |    450.32 |                    
 2020-05-19 04:25:10+00 | cpu0      |           | 0.29970029970031614
 2020-05-19 04:25:10+00 | cpu0      |    504.23 |                    
(21 rows)
```
Since it seems that TimescaleDB doesn't merge points like InfluxDB does (and vanilla postgres definitely doesn't), I think telegraf needs to be responsible for handling this and doing the merge.

Now I'm not sure that this needs to be handled by the postgres plugin itself though. This seems like it might be a good candidate for a processor plugin. I'm also not sure if processors receive points as a continuous stream, or batches like output plugins. If it is a continuous stream, I think a "merge" processor which adds a short time buffer (e.g. 10ms), such that when a point comes through, it's held back for that configured amount of time looking for other points to merge.

endersonmaia:
> When the postgresql output sends these, they're getting inserted as separate records.

if we have `time` field as a PK/Unique constraint and make an `UPSERT`, would solve ?

`INSERT INTO .... ON CONFLICT UPDATE...`

See: https://www.postgresql.org/docs/12/sql-insert.html#SQL-ON-CONFLICT

phemmer:
> if we have `time` field as a PK/Unique constraint and make an `UPSERT`, would solve ?
> 
> `INSERT INTO .... ON CONFLICT UPDATE...`

Well right now the plugin doesn't use `INSERT`, it uses `COPY`. While bulk inserts do exist, and would allow `ON CONFLICT` to work, I'm not sure of the implications of switching.

In addition, the creation of a unique constraint will impact performance, but I'm not sure by how much.

endersonmaia:
> Well right now the plugin doesn't use `INSERT`, it uses `COPY`. While bulk inserts do exist, and would allow `ON CONFLICT` to work, I'm not sure of the implications of switching.

🤔 

AFAIK, `COPY` doesn't support 'ON CONFLICT` nor something similar

we could use a `TEMPORARY TABLE` to `COPY` into and then make `INSERT... ON CONFLICT` from it, but it's a bit hacky to me

phemmer:
> Now I'm not sure that this needs to be handled by the postgres plugin itself though. This seems like it might be a good candidate for a processor plugin. I'm also not sure if processors receive points as a continuous stream, or batches like output plugins. If it is a continuous stream, I think it would be good to add a "merge" processor which adds a short time buffer (e.g. 10ms), such that when a point comes through, it's held back for that configured amount of time looking for other points to merge.

Well it seems that such a plugin already exists: https://github.com/influxdata/telegraf/tree/master/plugins/aggregators/merge
I swear I checked aggregators prior to my post, but it seems I'm blind.

The way the plugin is implemented greatest solution, as it operates in batches rather than as a stream, meaning that there's still the risk of a single point getting written twice if it's in 2 different batches. But it's better than nothing.

phemmer:
Just ran into another issue. I caused a situation (totally my fault) where a few records couldn't be written to the DB, and so errors were returned back to telegraf. I'm not sure how this plugin is handling transactioning, but other records were inserted fine. The problem is that the plugin kept trying to resend the bad records, along side the good records. This resulted in massive amounts of duplicate data in the database.

I think the error handling approach needs to be changed.
I'm not too familiar with error handling with `COPY`, but from some quick research, it seems like a good solution to this problem would be to have a config param that allows the user to select one of 3 options for error handling:
1. Use transactions, and fail the whole transaction, retrying it later. The whole thing succeeds, or the whole thing fails.
2. Ignore errors and continue. Just drop the bad records (with a log message telling the user), allowing the good records, and don't retry.
3. Use `COPY` error logging to insert the bad records into a temp table, and periodically retry those records. Perhaps also with an expiry time.

As it is, the only way I could stop the massive data duplication was to bounce telegraf so it'd discard the bad records and stop retrying them.

bithavoc:
Thanks for such great project, any ETA to merge this PR?

sjwang90:
@svenklemm or @blagojts, what are your thoughts around @phemmer's error handling concerns above?

Oleg26Dev:
Hi team, 
btw, still waiting  this plugin.
Thnx

phemmer:
Can we get some clarification on the state of this plugin & pull request.

After about 10 months without any update or response, it would appear that it has been abandoned. This is unfortunate, but it happens as priorities change. But if it needs a new maintainer, can we state that so people know it's open for someone to pick up and carry?

sjwang90:
@svenklemm, please chime in if you would like to continue to own this. Otherwise anyone who's interested in working on building on @svenklemm's code feel free to address necessary fixes.  

@phemmer Seems like you've provided a lot of great feedback if you'd like to take it over.

phemmer:
I don't know that I have the time to be able to take this on. But I'll keep it at near the top of my list of potential work when not pressed for other issues.

In the mean time I'll keep adding my experiences with it.  
So on that note, have another bit of feedback :-)

Currently the plugin's `table_template` value allows you to define some SQL that'll be run when the table needs to be created. I have one strong recommendation for a change regarding this, and a lesser request.

* Change it to accept a list of statements instead of a single statement.  
Ran into a limitation when I wanted to have the plugin create a timescaledb table, as well as enable compression on it. You cannot do `table_template = "CREATE TABLE ...; SELECT create_hypertable(...); ALTER TABLE ..."`. If you try you get `cannot insert multiple commands into a prepared statement (SQLSTATE 42601)`.  
So the solution is to allow passing multiple statements. All the statements should still be executed within the same transaction though.

* Add an option to create a view when using `tags_as_foreignkeys`. This view would transparently join the main table with the tag table. It might also be nice to throw the tables & the views in separate schemas, as with the view, you'll likely rarely use the underlying tables themselves, so putting them in a separate schema would get them out of the way.  
The construction of this view might be somewhat opinionated (what do you name the view, what do you name the underlying tables, etc). So an alternative might be to provide additional macros to the `table_template` so that the user could construct templates for these tasks.

AtakanColak:
As someone who used this plugin for months, I must warn everyone that this plugin does not scale well with many agents, many metrics and has many performance issues that quadruple when `tags_as_foreignkeys` are used. My recommendation for this plugin is a complete overhaul, remake, or even changing to Gorm v2 which has support for create from map. Would work on one if only had time.

phemmer:
> As someone who used this plugin for months, I must warn everyone that this plugin does not scale well with many agents, many metrics and has many performance issues that quadruple when `tags_as_foreignkeys` are used. My recommendation for this plugin is a complete overhaul, remake, or even changing to Gorm v2 which has support for create from map. Would work on one if only had time.

That's interesting. Can you elaborate any further? Given your statement that switching to an ORM will help performance, this implies that the performance issue is client side. I'm having a hard time imagining what that could be. I would expect the issue to be on the database side as the number of different series grows, in which it's not an ORM problem, but one of defining a proper segmentation key.

Also I'm not familiar with GORM, but ORMs tend to revolve around mapping defined types (e.g. `struct{}`s) to DB schemas. Given that you cannot create defined types for all the possible input metrics, how would that even work?

AtakanColak:
> > As someone who used this plugin for months, I must warn everyone that this plugin does not scale well with many agents, many metrics and has many performance issues that quadruple when `tags_as_foreignkeys` are used. My recommendation for this plugin is a complete overhaul, remake, or even changing to Gorm v2 which has support for create from map. Would work on one if only had time.
> 
> That's interesting. Can you elaborate any further? Given your statement that switching to an ORM will help performance, this implies that the performance issue is client side. I'm having a hard time imagining what that could be. I would expect the issue to be on the database side as the number of different series grows, in which it's not an ORM problem, but one of defining a proper segmentation key.
> 
> Also I'm not familiar with GORM, but ORMs tend to revolve around mapping defined types (e.g. `struct{}`s) to DB schemas. Given that you cannot create defined types for all the possible input metrics, how would that even work?

Performance issues exist both on DB side and client side. For example, `tags_as_foreignkeys` caused a block on the DB side with a high load of cpu and memory usage. On client (telegraf) side, this plugin parses the queries by hand to my knowledge whereas gorm already optimizes this, so I think that would be improvement over high CPU usage of this plugin which I measured to be about %60 of telegrafs total cpu usage. And regarding type part, I think it would need a workaround if gorm was to be used, with tables still created by hand, but metrics directly passed to gorm as maps.

I would like to reiterate my warning especially for commercial projects.

phemmer:
Just FYI, I've started looking into addressing some of the issues on this plugin.

@AtakanColak  I think I found the performance issue you're experiencing. The issue is that when using `tags_as_foreignkeys`, telegraf has to perform a query against the tag table to find the tag ID for a given series. Additionally it performes this task on a one-by-one basis, which is extremely slow. Now the plugin does cache the IDs it obtains, however it's only an LRU cache with a default size of 1000 (which isn't documented on timescaledb's docs), which is very easily exceeded. And once it is exceeded, the performance immediately tanks. There is no in-between. As soon as you hit 1001 items, the LRU is effectively useless. On top of that, when telegraf first starts up, the cache is cold, and in my case it took several minutes to populate it, during which time telegraf can't keep up and is dropping metrics.

I was able to solve this issue by changing the code to use a deterministic key. Meaning telegraf computes the tag ID and doesn't have to do a lookup.

With this change I was able to go from about 15k points per minute to about 50k.

Additionally with the need for using the `aggregators.merge` plugin as explained [here](https://github.com/influxdata/telegraf/pull/3428#issuecomment-630579777), this ended up being the next bottleneck. I opened PR #8391 to address that issue.

With that, and some other hackery involving multiple connections, I'm able to get telegraf up to about 1m points per minute, a massive improvement over 15k. But it's still telegraf bottlenecking, and not the DB.

I'm going to clean up the plugin a bit so I can remove some of the performance hacks I had to do, and will submit to a new branch.

phemmer:
I've put up a draft PR of the changes over at #8651. While it is a draft, it should be fully functional, and I would welcome feedback.

The plugin has been massively reworked. But it's a lot more flexible, fault tolerant, and performant.

sjwang90:
Closing this for #8651

devosalain:
I use freebsd13.0 Where can i download the sourcecode for this plugin to compile it ?

PR #3434, opened 2017-11-06
Add unbound input plugin

aromeyer:
This plugin gathers stats from Unbound - a validating, recursive, and caching DNS resolver (https://www.unbound.net/)

### Required for all PRs:

### DIFF
--- a/plugins/inputs/unbound/README.md
+++ b/plugins/inputs/unbound/README.md
@@ -1,4 +1,4 @@
-# Varnish Input Plugin
+# Unbound Input Plugin
 
 This plugin gathers stats from [Unbound - a validating, recursive, and caching DNS resolver](https://www.unbound.net/)
 
@@ -6,11 +6,11 @@ This plugin gathers stats from [Unbound - a validating, recursive, and caching D
 
 ```toml
  # A plugin to collect stats from Unbound - a validating, recursive, and caching DNS resolver
- [[inputs.varnish]]
+ [[inputs.unbound]]
    ## If running as a restricted user you can prepend sudo for additional access:
    #use_sudo = false
 
-   ## The default location of the varnishstat binary can be overridden with:
+   ## The default location of the unbound-control binary can be overridden with:
    binary = "/usr/sbin/unbound-control"
 
    ## By default, telegraf gathers stats for 3 metric points.
@@ -156,7 +156,7 @@ telegraf : telegraf
 $ usermod -a -G unbound telegraf
 
 $ groups telegraf
-telegraf : telegraf varnish
+telegraf : telegraf unbound
 ```
 
 **Sudo privileges**:
--- a/plugins/inputs/unbound/README.md
+++ b/plugins/inputs/unbound/README.md
@@ -13,135 +13,89 @@ This plugin gathers stats from [Unbound - a validating, recursive, and caching D
    ## The default location of the unbound-control binary can be overridden with:
    binary = "/usr/sbin/unbound-control"
 
-   ## By default, telegraf gathers stats for 3 metric points.
-   ## Setting stats will override the defaults shown below.
-   ## stats may also be set to ["all"], which will collect all stats
+   # The default timeout of 1000ms can be overriden with (in milliseconds):
+   #timeout = 1000
+
+  ## By default, telegraf gather stats for 4 metric points.
+  ## Setting stats will override the defaults shown below.
+  ## Glob matching can be used, ie, stats = ["total.*"]
+  ## stats may also be set to ["*"], which will collect all stats
+  ## except histogram.* statistics that will never be collected.
    stats = ["total.*", "num.*","time.up", "mem.*"]
 ```
 
 ### Measurements & Fields:
 
-This is the full list of stats provided by unbound. Stats will be grouped by their prefix (eg thread0,
-total, etc). In the output, the prefix will be used as a tag, and removed from field names. See
-https://www.unbound.net/documentation/unbound-control.html for details.
+This is the full list of stats provided by unbound-control and potentially collected by telegram
+depending of your unbound configuration. Histogram related statistics will never be collected,
+extended statistics can also be imported ("extended-statistics: yes" in unbound configuration).
+In the output, the dots in the unbound-control stat name are replaced by underscores(see
+https://www.unbound.net/documentation/unbound-control.html for details).
 
 - unbound
-    thread0.num.queries
-    thread0.num.cachehits
-    thread0.num.cachemiss
-    thread0.num.prefetch
-    thread0.num.recursivereplies
-    thread0.requestlist.avg
-    thread0.requestlist.max
-    thread0.requestlist.overwritten
-    thread0.requestlist.exceeded
-    thread0.requestlist.current.all
-    thread0.requestlist.current.user
-    thread0.recursion.time.avg
-    thread0.recursion.time.median
-    total.num.queries
-    total.num.cachehits
-    total.num.cachemiss
-    total.num.prefetch
-    total.num.recursivereplies
-    total.requestlist.avg
-    total.requestlist.max
-    total.requestlist.overwritten
-    total.requestlist.exceeded
-    total.requestlist.current.all
-    total.requestlist.current.user
-    total.recursion.time.avg
-    total.recursion.time.median
-    time.now
-    time.up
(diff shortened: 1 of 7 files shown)


plugins/inputs/unbound/unbound_windows.go
```diff
@@ -0,0 +1,3 @@
+// +build windows
```
danielnelson: Let's leave this file out, if somehow you have unbound-control on Windows then have at it.

plugins/inputs/unbound/README.md
```diff
@@ -0,0 +1,186 @@
...
+    num.answer.secure
+    num.answer.bogus
+    num.rrset.bogus
+    unwanted.queries
+    unwanted.replies
+
+### Tags:
+
+As indicated above, the  prefix of a unbound stat will be used as it's 'section' tag. So section tag may have one of
+the following values:
```
danielnelson: It seems like each section is very different, maybe we should call the measurement `unbound_<section>` and remove the tag.  If we did this we would probably want to do something special for the threadX section.

plugins/inputs/unbound/README.md
```diff
@@ -0,0 +1,186 @@
...
+```
+
+### Measurements & Fields:
+
+This is the full list of stats provided by unbound. Stats will be grouped by their prefix (eg thread0,
+total, etc). In the output, the prefix will be used as a tag, and removed from field names. See
+https://www.unbound.net/documentation/unbound-control.html for details.
+
+- unbound
+    thread0.num.queries
```
danielnelson: In this section list the fields how they will be emitted from Telegraf, so `thread0.num.queries` will be `num.queries`.

danielnelson: I think we should consider replace dots with underscores the field `num_queries` so the fields fit style wise with the other plugins.

plugins/inputs/unbound/README.md
```diff
@@ -0,0 +1,186 @@
...
+    total.recursion.time.median
+    time.now
+    time.up
+    time.elapsed
+    mem.total.sbrk
+    mem.cache.rrset
+    mem.cache.message
+    mem.mod.iterator
+    mem.mod.validator
+    histogram.000000.000000.to.000000.000001
```
danielnelson: How are these histogram metrics encoded?  Is the field named `000000.000000.to.000000.000001`?  Maybe we can do special processing of these, normally we use the format produced by the histogram aggregator.

plugins/inputs/unbound/unbound.go
```diff
@@ -0,0 +1,163 @@
...
+	cmd := exec.Command(cmdName, cmdArgs...)
+
+	if UseSudo {
+		cmdArgs = append([]string{cmdName}, cmdArgs...)
+		cmd = exec.Command("sudo", cmdArgs...)
+	}
+
+	var out bytes.Buffer
+	cmd.Stdout = &out
+	err := internal.RunTimeout(cmd, time.Millisecond*200)
```
danielnelson: This seems like an aggressive timeout, I would move it up to at least a second and consider making it configurable.

plugins/inputs/unbound/unbound.go
```diff
@@ -0,0 +1,163 @@
...
+// The prefix of each stat (eg MAIN, MEMPOOL, LCK, etc) will be used as a
+// 'section' tag and all stats that share that prefix will be reported as fields
+// with that tag
+func (s *Unbound) Gather(acc telegraf.Accumulator) error {
+	if s.filter == nil {
+		var err error
+		if len(s.Stats) == 0 {
+			s.filter, err = filter.Compile(defaultStats)
+		} else {
+			// legacy support, change "all" -> "*":
```
danielnelson: We can't have legacy support already since we are a new plugin :)

plugins/inputs/unbound/unbound.go
```diff
@@ -0,0 +1,163 @@
...
+		return &out, fmt.Errorf("error running unbound-control: %s", err)
+	}
+
+	return &out, nil
+}
+
+// Gather collects the configured stats from unbound_stat and adds them to the
+// Accumulator
+//
+// The prefix of each stat (eg MAIN, MEMPOOL, LCK, etc) will be used as a
```
danielnelson: These aren't actual sections, and there is no unbound_stat.  I guess this is holdovers from an earlier version of the code, can you update it?

plugins/inputs/unbound/unbound.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+
+	sectionMap := make(map[string]map[string]interface{})
+	scanner := bufio.NewScanner(out)
+	for scanner.Scan() {
+
+		cols := strings.Split(scanner.Text(), "=")
+
+		stat := cols[0]
+		value := cols[1]
```
danielnelson: This could panic if somehow there are not two fields, make sure to guard against this.

plugins/inputs/unbound/unbound.go
```diff
@@ -0,0 +1,163 @@
...
+
+		section := parts[0]
+		field := parts[1]
+
+		// Init the section if necessary
+		if _, ok := sectionMap[section]; !ok {
+			sectionMap[section] = make(map[string]interface{})
+		}
+
+		sectionMap[section][field], err = strconv.ParseUint(value, 10, 64)
```
danielnelson: Parse as int64, since this is what the Accumulator holds, if there is an error parsing don't add to the fields.

plugins/inputs/unbound/unbound.go
```diff
@@ -0,0 +1,163 @@
...
+		if err != nil {
+			sectionMap[section][field], err = strconv.ParseFloat(value, 64)
```
danielnelson: This fallback method could switch the type of the field, which will cause the metrics to be impossible to add to InfluxDB.  We may need to parse all fields as floats, or have a list of the types.

plugins/inputs/unbound/README.md
aromeyer: As there is not too much data (no really need to spread information in several measurements) and I do not control unbound-control future evolution, I will start to simply drop section/tag and push all fields as they are (just converting field name dots to underscores).  
I will just have a look on the "histograms" fields and see if I drop them or do something intelligible. I am not sure that these fields are really interesting for telegraf as we are collecting timeseries data... we don't really need partially pre-processed historical information.

plugins/inputs/unbound/README.md
aromeyer: Yes actually the field name will be 000000.000000.to.000000.000001 which is not really usefful. Anyway I don't think that this kind of information is relevant for telegraf. It is easier to construct the histogram from telegraf collected data rather than using these fields. I think I will skip them from collected metrics.

plugins/inputs/unbound/unbound.go
```diff
@@ -0,0 +1,161 @@
...
+	"github.com/influxdata/telegraf/plugins/inputs"
+)
+
+type runner func(cmdName string, Timeout int, UseSudo bool) (*bytes.Buffer, error)
+
+// Unbound is used to store configuration values
+type Unbound struct {
+	Stats   []string
+	Binary  string
+	Timeout int
```
danielnelson: Use `internal.Duration` for this, it will look like `timeout = "1s"` in toml.

plugins/inputs/unbound/unbound.go
```diff
@@ -0,0 +1,161 @@
+// +build !windows
```
danielnelson: Can you remove this build flag, and also in the test file.

plugins/inputs/unbound/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+   # The default timeout of 1s can be overriden with:
+   #timeout = "1s"
+
+  ## By default, telegraf gather stats for 4 metric points.
+  ## Setting stats will override the defaults shown below.
+  ## Glob matching can be used, ie, stats = ["total.*"]
+  ## stats may also be set to ["*"], which will collect all stats
+  ## except histogram.* statistics that will never be collected.
+  stats = ["total.*", "num.*","time.up", "mem.*"]
```
danielnelson: I think we should remove this filter, users can simply use the builtin fielddrop/fieldpass filters to get the same behavior.

PR #2990, opened 2017-07-06
Add result_type field to net_response input plugin

bobmshannon:
This PR adds a newly proposed field called `result_type` (inspired by the [http_response](https://github.com/influxdata/telegraf/pull/2814) input plugin) which indicates whether a collection was either successful or failed. This should be useful for alerting purposes, and also will fix #2794. In an effort to be consistent it also removes the `string_match` boolean field and replaces it with a `string_mismatch` result type (this is a breaking change, not sure if we should keep this for backward compatibility or not).

**Proposed result types**:
- **result_type (string)** # `success`, `timeout`, `connection_failed`, `read_failed`, `string_mismatch`
 
### Test results
```
* Plugin: inputs.net_response, Collection 1
> net_response,server=influxdata.com,port=8080,protocol=tcp,host=localhost result_type="timeout" 1499310361000000000
* Plugin: inputs.net_response, Collection 1
> net_response,server=influxdata.com,port=443,protocol=tcp,host=localhost result_type="success",response_time=0.088703864 1499310361000000000
* Plugin: inputs.net_response, Collection 1
> net_response,protocol=tcp,host=localhost,server=this.domain.does.not.exist,port=443 result_type="connection_failed" 1499310361000000000
* Plugin: inputs.net_response, Collection 1
> net_response,protocol=udp,host=localhost,server=influxdata.com,port=8080 result_type="read_failed" 1499310362000000000
* Plugin: inputs.net_response, Collection 1
> net_response,port=31338,protocol=udp,host=localhost,server=localhost result_type="string_mismatch",string_found=false,response_time=0.00242682 1499310362000000000
* Plugin: inputs.net_response, Collection 1
> net_response,protocol=udp,host=localhost,server=localhost,port=31338 response_time=0.001128598,result_type="success",string_found=true 1499310362000000000
* Plugin: inputs.net_response, Collection 1
> net_response,server=this.domain.does.not.exist,port=443,protocol=udp,host=localhost result_type="connection_failed" 1499310362000000000
```

### Required for all PRs:

- [x] Signed [CLA](https://influxdata.com/community/cla/).
- [x] Associated README.md updated.
- [x] Has appropriate unit tests.

PR #3450, opened 2017-11-09
Updating net_response as discussed in #2990

morfien101:
Updating the net_response plugin to send the field result_type as a int and moved the current result_type string value to a tag.

Fixing go linting issues.
Updating tests to reflect the changes.
Updating README.md to reflect the changes.

### Required for all PRs:

- [ x  ] Has appropriate unit tests.

### DIFF
--- a/plugins/inputs/net_response/README.md
+++ b/plugins/inputs/net_response/README.md
@@ -59,7 +59,7 @@ It can also check response text.
 
 - net_response
     - response_time (float, seconds)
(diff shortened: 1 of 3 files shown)


morfien101:
I will make a new PR as the changes here, although passed the tests caused issues with data in influxDB. Due to the tag and field having the same name one of them gets called result_type_1. I will change the name of the field to success.

PR #3451, opened 2017-11-09
move field result_type to be a tag

morfien101:
Updating the net_response plugin to move the result_type field to a tag. This will allow it so be used for filtering via the string values.
Replacing the field with value called success which has an int of 0 or 1 based on the result of the test.

Updating linting for go rules.
Updating README.md and tests to reflect the changes made.

### Required for all PRs:


morfien101:
rejected

PR #3471, opened 2017-11-14
 use MAX() instead of SUM() for latency measurements

erayaslan:
Using SUM() results in multi-second latency measurements when multiple files are used for a database - which is typical for tempdb.  Using MAX() gives a more meaningful measurement.   AVG() can be used as well but will hide problems when one of the db files encounters a problem.  Example:

File0: 30ms
File1: 40ms
File2: 50ms
File3: 60ms
File4: 70ms
File5: 80ms
File6: 90ms

SUM (currently what is reported) results in 420 ms latency, MAX (current proposal) in 90ms and AVG in 60ms


danielnelson:
@zensqlmonitor @m82labs Could you help review this change?

PR #3477, opened 2017-11-16
Add Particle Webhook Plugin

davidgs:
A Webhook Plugin for the Partricle.io platform

### Required for all PRs:

- [X ] Associated README.md updated.

### DIFF
--- a/plugins/inputs/webhooks/particle/particle_webhooks.go
+++ b/plugins/inputs/webhooks/particle/particle_webhooks.go
@@ -50,7 +50,7 @@ func (rb *ParticleWebhook) eventHandler(w http.ResponseWriter, r *http.Request) 
 	defer r.Body.Close()
 	e := newEvent()
 	if err := json.NewDecoder(r.Body).Decode(e); err != nil {
-		log.Println(err)
+		rb.acc.AddError(err)
 		w.WriteHeader(http.StatusBadRequest)
 		return
 	}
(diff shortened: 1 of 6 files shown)


plugins/inputs/webhooks/particle/particle_webhooks.go
```diff
@@ -0,0 +1,66 @@
...
+func (rb *ParticleWebhook) Register(router *mux.Router, acc telegraf.Accumulator) {
+	router.HandleFunc(rb.Path, rb.eventHandler).Methods("POST")
+	rb.acc = acc
+}
+
+func (rb *ParticleWebhook) eventHandler(w http.ResponseWriter, r *http.Request) {
+	defer r.Body.Close()
+	e := newEvent()
+	if err := json.NewDecoder(r.Body).Decode(e); err != nil {
+		log.Println(err)
```
danielnelson: Use `rb.acc.AddError(` here, which will log the message with a message about which plugin generated it and increment the error counter for the internal plugin.

plugins/inputs/webhooks/particle/particle_webhooks.go
```diff
@@ -0,0 +1,66 @@
...
+		w.WriteHeader(http.StatusBadRequest)
+		return
+	}
+
+	pTime, err := e.Time()
+	if err != nil {
+		pTime = time.Now()
+		log.Printf("error parsing particle event time: %s. Using telegraf host time instead: %s", e.PublishedAt, pTime)
```
danielnelson: I would probably use AddError here too, unless you think it is better to not increment the gather error count.

plugins/inputs/webhooks/particle/particle_webhooks.go
davidgs: Ok, sounds good.

plugins/inputs/webhooks/particle/particle_webhooks.go
davidgs: The error here can probably just be removed. It's not really an error, probably, so much as a "didn't work, but we can go on".

PR #3482, opened 2017-11-17
Prometheus plugin: support passing bearer token directly

### LINKED ISSUE #3481: Prometheus plugin: support passing bearer_token directly
## Feature Request

### Current behavior:

https://github.com/influxdata/telegraf/tree/master/plugins/inputs/prometheus

only support to pass bearer token as a path to file:

```
## Use bearer token for authorization
  # bearer_token = /path/to/bearer/token
```

### Desired behavior:

It should be supported to pass token directly and rename `bearer_token` to `bearer_token_file`. Then it will be consistent with https://prometheus.io/docs/prometheus/latest/configuration/configuration/

We are using it for [Grafana metrics](http://docs.grafana.org/installation/configuration/#metrics). After [generating the token](http://docs.grafana.org/http_api/auth/), no need to copy it to a file.

quantonganh:
Fix #3481 

### Required for all PRs:

### DIFF
--- a/plugins/inputs/prometheus/prometheus.go
+++ b/plugins/inputs/prometheus/prometheus.go
@@ -25,8 +25,11 @@ type Prometheus struct {
 	// An array of Kubernetes services to scrape metrics from.
 	KubernetesServices []string
 
-	// Bearer Token authorization file path
+	// Bearer Token authorization.
 	BearerToken string `toml:"bearer_token"`
+
+	// Bearer Token authorization file path.
+	BearerTokenFile string `toml:"bearer_token_file"`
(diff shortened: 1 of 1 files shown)


danielnelson:
This will break backwards compatibility in the config file which we are trying to avoid as much as possible.  How about instead we add support for general static headers similar to what we have in the httpjson input?

https://github.com/influxdata/telegraf/blob/8614445235a4fb5cd9a56b895221cbdac3d1a59a/plugins/inputs/httpjson/httpjson.go#L111-L113

russorat:

PR #3511, opened 2017-11-27
adding twitter input

russorat:
### Required for all PRs:

### DIFF
--- a/plugins/inputs/twitter/README.md
+++ b/plugins/inputs/twitter/README.md
@@ -1,0 +1,59 @@
+# Twitter Input Plugin
+
+This input plugin provides a way to ingest tweet metrics and events into InfluxDB.
+
+### Configuration:
+
+This section contains the default TOML to configure the plugin.  You can
+generate it using `telegraf --usage <plugin-name>`.
+
+```toml
+# Description
+[[inputs.example]]
+  example_option = "example_value"
+```
+
+### Metrics:
+
+Here you should add an optional description and links to where the user can
+get more information about the measurements.
+
+If the output is determined dynamically based on the input source, or there
+are more metrics than can reasonably be listed, describe how the input is
+mapped to the output.
+
+- measurement1
+  - tags:
+    - tag1 (optional description)
+    - tag2
+  - fields:
+    - field1 (type, unit)
+    - field2 (float, percent)
+
+- measurement2
+  - tags:
+    - tag3
+  - fields:
+    - field3 (integer, bytes)
+
+### Sample Queries:
+
+This section should contain some useful InfluxDB queries that can be used to
+get started with the plugin or to generate dashboards.  For each query listed,
+describe at a high level what data is returned.
+
+Get the max, mean, and min for the measurement in the last hour:
+```
+SELECT max(field1), mean(field1), min(field1) FROM measurement1 WHERE tag1=bar AND time > now() - 1h GROUP BY tag
+```
+
+### Example Output:
+
+This section shows example output in Line Protocol format.  You can often use
+`telegraf --input-filter <plugin-name> --test` or use the `file` output to get
+this information.
+
+```
+measurement1,tag1=foo,tag2=bar field1=1i,field2=2.1 1453831884664956455
+measurement2,tag1=foo,tag2=bar,tag3=baz field3=1i 1453831884664956455
+```
--- a/plugins/inputs/twitter/twitter.go
+++ b/plugins/inputs/twitter/twitter.go
@@ -1,0 +1,100 @@
+package twitter
+
+import (
+	"net/url"
+	"strings"
+
+	"github.com/ChimeraCoder/anaconda"
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/plugins/inputs"
+)
+
+const (
+	measurement = "tweets"
+)
+
+// Twitter - The object used to hold the configuration options from the user.
+type Twitter struct {
+	ConsumerKey       string
+	ConsumerSecret    string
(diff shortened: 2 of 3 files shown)


plugins/inputs/twitter/README.md
```diff
@@ -0,0 +1,59 @@
+# Twitter Input Plugin
+
+This input plugin provides a way to ingest tweet metrics and events into InfluxDB.
+
+### Configuration:
+
+This section contains the default TOML to configure the plugin.  You can
+generate it using `telegraf --usage <plugin-name>`.
```
danielnelson: Obviously the rest of this file would need updated

plugins/inputs/twitter/twitter.go
```diff
@@ -0,0 +1,100 @@
...
+
+// SampleConfig - returns a sample config block for the plugin
+func (s *Twitter) SampleConfig() string {
+	return `
+    ## These values can be obtained from apps.twitter.com
+    consumerKey = ""
+    consumerSecret = ""
+    accessToken = ""
+    accessTokenSecret = ""
+    keywordsToTrack = ""
```
danielnelson: Use snake_case in the config and 2 space indent to match other plugins.

plugins/inputs/twitter/twitter.go
```diff
@@ -0,0 +1,100 @@
...
+	anaconda.SetConsumerKey(s.ConsumerKey)
+	anaconda.SetConsumerSecret(s.ConsumerSecret)
+	api := anaconda.NewTwitterApi(s.AccessToken, s.AccessTokenSecret)
+	go s.fetchTweets(api, acc)
+	return nil
+}
+
+// Stop - Called once when stopping the plugin
+func (s *Twitter) Stop() {
+	return
```
danielnelson: Needs to stop goroutine started in Start, or things like reloading via SIGHUP will have problems.

plugins/inputs/twitter/twitter.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+func (s *Twitter) fetchTweets(api *anaconda.TwitterApi, acc telegraf.Accumulator) error {
+	// We will use this a little later for finding keywords in tweets
+	keywordsList := strings.Split(s.KeywordsToTrack, ",")
+	// Setting the keywords we want to track
+	v := url.Values{}
+	v.Set("track", s.KeywordsToTrack)
+	stream := api.PublicStreamFilter(v)
+	for {
```
danielnelson: Use `for item := range stream.C {`, this way you will stop if the channel is closed.

plugins/inputs/twitter/twitter.go
```diff
@@ -0,0 +1,100 @@
...
+			fields["user_verified"] = tweet.User.Verified
+			fields["raw"] = tweet.Text
+			time, _ := tweet.CreatedAtTime()
+			for i := 0; i < len(keywordsList); i++ {
+				if strings.Contains(strings.ToLower(tweet.Text), strings.ToLower(keywordsList[i])) {
+					tags["keyword"] = strings.ToLower(keywordsList[i])
+					acc.AddFields("tweets", fields, tags, time)
+				}
+			}
+		default:
```
danielnelson: Probably remove default label since it does nothing.

plugins/inputs/twitter/twitter.go
```diff
@@ -0,0 +1,100 @@
...
+			fields["retweet_count"] = tweet.RetweetCount
+			fields["tweet_id"] = tweet.IdStr
+			fields["followers_count"] = tweet.User.FollowersCount
+			fields["screen_name"] = tweet.User.ScreenName
+			fields["friends_count"] = tweet.User.FriendsCount
+			fields["favourites_count"] = tweet.User.FavouritesCount
+			fields["screen_name"] = tweet.User.ScreenName
+			fields["user_verified"] = tweet.User.Verified
+			fields["raw"] = tweet.Text
+			time, _ := tweet.CreatedAtTime()
```
danielnelson: Haven't looked at this function, but probably should do something on err.  Maybe just use `acc.AddError(err)` and continue.

plugins/inputs/twitter/twitter.go
```diff
@@ -0,0 +1,100 @@
...
+			for i := 0; i < len(keywordsList); i++ {
```
danielnelson: This might be the python developer in me, but indexes are the worst: `for _, keyword := range keywordList`

plugins/inputs/twitter/twitter.go
```diff
@@ -0,0 +1,100 @@
...
+				if strings.Contains(strings.ToLower(tweet.Text), strings.ToLower(keywordsList[i])) {
+					tags["keyword"] = strings.ToLower(keywordsList[i])
```
danielnelson: What happens if there are two matches?  Only the last will be assigned.  I don't have a real solution for this, it reminds me of a situation where a user wanted to assign multiple roles to a point without duplicating data, which is not something we can do.

danielnelson:
BTW, to make the tests pass you would need to add anaconda to Godeps, and the license should be added to `docs/LICENSE_OF_DEPENDENCIES.md`

russorat:
@danielnelson tried to address all the comments except the readme. can you double check to make sure i am using channels correctly?

danielnelson:
It looks correct, I would probably use a `context.Context` instead of a done channel normally these days, but in this case it looks like it is built into the library https://godoc.org/github.com/ChimeraCoder/anaconda#Stream.Stop

russorat:
closing since this won't be getting merged.

PR #354, opened 2015-11-08
Parse statsd lines with multiple metric bits

MerlinDMC:
This resembles the parsing which is done in etsy/statsd to add multiple datapoints for a measurement in a single line.
https://github.com/etsy/statsd/blob/243a1f2a166c2d573f7582dc0c42f50f257e4150/stats.js#L225-L277

It seems as if some client applications use those possibilities to pack more data into each UDP packet and simply omit the repeating measurement name where possible.

An example for this is the Kamon.io statsd reporter which produces lines like these:
```
kamon.host.akka-actor.kamon_user_kamon-system-metrics_sigar-metrics-recorder.processing-time:581632|ms:593920|ms:610304|ms:622592|ms:634880|ms:638976|ms:659456|ms:679936|ms:720896|ms:770048|ms
kamon.host.akka-actor.kamon_user_kamon-system-metrics_sigar-metrics-recorder.time-in-mailbox:5664|ms:6176|ms:6240|ms|@0.5:6304|ms:6336|ms:6656|ms:6720|ms:7072|ms:12288|ms
```

PR #3514, opened 2017-11-28
when tag includes ":", split won't behave as expected.

sagydr:
so "bits" would simply be the last split section, and bucketName is join of all-but-the-last sections

### Required for all PRs:

- [V] Signed [CLA](https://influxdata.com/community/cla/).

### DIFF
--- a/plugins/inputs/statsd/statsd.go
+++ b/plugins/inputs/statsd/statsd.go
@@ -488,7 +488,8 @@ func (s *Statsd) parseStatsdLine(line string) error {
 	}
 
 	// Extract bucket name from individual metric bits
-	bucketName, bits := bits[0], bits[1:]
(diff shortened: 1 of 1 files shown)


sagydr:
it's my first open-source contribution,
please contact me if i messed up something: sagysrael@gmail.com
thanks.

danielnelson:
I think this would break the "multi-bit" lines added in https://github.com/influxdata/telegraf/pull/354.

Looking closer into the statsd spec and implementations, I don't think there is any way to encode a colon before the value|type.  I'm still open to allowing it, but I think it will not be trivial and we need to ensure we don't slow the parser down if we do it.

danielnelson:
@sagydr I think it is probably best not to support this functionality due to the way the statsd spec and protocol is created, so I'm going to close this pull request.  Hope this does not discourage you from other contributions.

PR #3523, opened 2017-11-29
Valuecounter aggregator plugin

piotr1212:
A valuecounter aggregator plugin to count values in specified fields.
    Emits the aggregated count of the values.

A usecase for the valuecounter plugin is when you are processing a HTTP access
log (with the logparser input) and want to count the HTTP status codes.

The fields which will be counted must be configured with the `field_names`
configuration directive. When no `field_names` is provided the plugin will not
count any fields. The results are emitted in fields in the format:
`originalfieldname_fieldvalue = count`.

See also: https://community.influxdata.com/t/telegraf-count-field-values-processing-http-response-codes/3214/5

### Required for all PRs:


slogar:
This looks like a feature that we need. Can you please merge this one so we can deploy this?

plugins/aggregators/valuecounter/valuecounter.go
```diff
@@ -0,0 +1,94 @@
...
+
+type aggregate struct {
+	name       string
+	tags       map[string]string
+	fieldCount map[string]int
+}
+
+type ValueCounter struct {
+	cache       map[uint64]aggregate
+	Field_names []string
```
danielnelson: Call this `Fields`, which will match with some other aggregator work.

Strictly speaking, we could probably use the normal filtering options, since this is essentially fieldpass, but I think it is important to require this to be opt in.

plugins/aggregators/valuecounter/valuecounter.go
```diff
@@ -0,0 +1,94 @@
...
+
+var sampleConfig = `
+  ## General Aggregator Arguments:
+  ## The period on which to flush & clear the aggregator.
+  period = "30s"
+  ## If true, the original metric will be dropped by the
+  ## aggregator and will not get sent to the output plugins.
+  drop_original = false
+  ## The fields for which the values will be counted
+  field_names = ""
```
danielnelson: Should be a list: `fields = []`

plugins/aggregators/valuecounter/valuecounter.go
```diff
@@ -0,0 +1,94 @@
...
+			name:       in.Name(),
+			tags:       in.Tags(),
+			fieldCount: make(map[string]int),
+		}
+		vc.cache[id] = a
+	}
+
+	// Check if this metric has fields which we need to count, if so increment
+	// the count.
+	for fk, fv := range in.Fields() {
```
danielnelson: I wonder if we should skip floats?  One thing that would be bad is if there is a frequently changing float value, which could result in many new fields.

If we don't disallow this, maybe we should add a warning in the README.

plugins/aggregators/valuecounter/valuecounter.go
```diff
@@ -0,0 +1,94 @@
...
+		for _, cf := range vc.Field_names {
+			if fk == cf {
+				fn := fmt.Sprintf("%v_%v", fk, fv)
+				vc.cache[id].fieldCount[fn] += 1
+			}
+		}
+	}
+}
+
+// Emmit the counters
```
danielnelson: typo: Emit

plugins/aggregators/valuecounter/README.md
```diff
@@ -0,0 +1,70 @@
+# ValueCounter Aggregator Plugin
+
+The valuecounter plugin counts the occurance of values in fields and emmits the
+counter once every 'preriod' seconds.
```
danielnelson: typo: period

plugins/aggregators/valuecounter/valuecounter.go
piotr1212: good suggestion, will check if I can implement this

piotr1212:
Apart from the float check the suggestions are implemented. 

For some reason when running head I don't see events from inputs.logparser going through the aggregators at all. I do see events from other plugins passing by. I'm not sure if this is a bug or that I am doing something wrong here..

danielnelson:
Most likely the logs you are parsing are old, and only the metrics with a timestamp in the current period are added to the aggregators.  There is also a minor bug with service inputs adding metrics too early right at startup, sometimes I run into it when working with very small test sets.

piotr1212:
The last commit implements filtering on type as suggested. I've also updated the docs.

plugins/aggregators/valuecounter/valuecounter.go
```diff
@@ -0,0 +1,109 @@
...
+		for _, cf := range vc.Fields {
+			if fk == cf {
+				// Do not process float types to prevent memory from blowing up
+				switch fv.(type) {
+				default:
+					log.Printf("I! Valuecounter: Unsupported field type. " +
+						"Must be an int, string or bool. Ignoring.")
+					continue
+				case uint8, uint16, uint32, uint64, int8,
+					int16, int32, int64, string, bool:
```
danielnelson: Only need to support int64, uint64, bool, and string.

piotr1212: ok, thanks, updated.

piotr1212:
I think the failing 'release' test is unrelated

danielnelson:
I think it will clear up if you merge or rebase master.

piotr1212:
thanks

dataviruset:
It would be great if there would only be one field and the counted values would be tagged instead of having multiple fields. Makes it easier to do GROUP BY etc.

danielnelson:
@dataviruset You might want to use the regex and/or converter processors instead.  Open a new issue with more details about your use case or post on the [InfluxData Community](https://community.influxdata.com) site if you want to discuss further.

dataviruset:
@danielnelson Thanks for the reply. I created a post with more details on the community: https://community.influxdata.com/t/tail-and-count-log-file-with-tags/6484

plugins/aggregators/valuecounter/valuecounter.go
ramyam07: Hi Daniel 
My requirement is to filter the logs by response code. If the response code is 500 then only we have to insert in the influx db 

Please suggest me the approach

Issue #2647, opened 2017-04-11
Monitor available kernel entropy


phemmer:
### Proposal:
Telegraf should be able to monitor available kernel random entropy. On linux this is exposed through `/proc/sys/kernel/random/entropy_avail`. 

### Current behavior:
Telegraf cannot monitor available kernel random entropy.

### Desired behavior:
Monitor available kernel random entropy.

### Use case: [Why is this important (helps with prioritizing requests)]
For applications which utilize `/dev/random` a lot, the entropy pool can run out, and cause said applications to significantly slow down. Thus it is important to monitor for entropy pool exhaustion, and for an admin to intervene if the pool is frequently running low.

phemmer:
One thing that might be useful for this, and other requests such as #2643, and #2609, would be a generic "sysctl" or "file" input. Most of the time the sysctl files just contain a single number. We could create an input plugin which behaves something like:

```
[[inputs.file]]
  [[inputs.file.file]]
    path = "/proc/sys/kernel/random/entropy_avail"
    field = "entropy_avail"
    data_format = "value"
  [[inputs.file.file]]
    path = "foo"
    field = "bar"
    data_format = "value"
```

We can make configuration simpler by defaulting `data_format` to `"value"` and `field` to the filename component of `path`.

The only thing this might not handle is things such as the formats handled by the [linux_sysctl_fs](https://github.com/influxdata/telegraf/pull/2609) input, which can be multiple fields in one file.
For these we can either support an array of fields, such as `fields = ["field1","field2","field3"]`, or use a grok pattern. If we want this input to be able to handle stuff other than `/proc/sys`, a grok pattern is really the only option, as space delimitation is too inflexible.

danielnelson:
Yeah it is a good idea, could almost do it with tail, but not quite and also tail comes with problems.  I think grok should also probably become a data_format, though we do get a lot of support related questions when using it.

For this one in particular it's probably best to have it as a first class series.  What are you thinking for the measurement name?

phemmer:
I dunno. I personally dislike having tons of itty bitty measurements. I would be more inclined to put this in with something else, such as `kernel` or `system`.
But if it is going to be a separate measurement, "kernel_entropy".

danielnelson:
[...]

PR #3524, opened 2017-11-29
Monitor available entropy

gerardocorea92:
### Required for all PRs:

### DIFF
--- a/plugins/inputs/system/kernel.go
+++ b/plugins/inputs/system/kernel.go
@@ -41,7 +41,6 @@ func (k *Kernel) Gather(acc telegraf.Accumulator) error {
 		return err
 	}
 
-	//Assigning entropyData variable the output of ReadFile
 	entropyData, err := ioutil.ReadFile(k.entropyStatFile)
 	if err != nil {
 		return err
--- a/plugins/inputs/system/kernel.go
+++ b/plugins/inputs/system/kernel.go
@@ -54,7 +54,7 @@ func (k *Kernel) Gather(acc telegraf.Accumulator) error {
 
 	fields := make(map[string]interface{})
 
-	fields["entropy_available"] = int64(entropyValue)
+	fields["entropy_avail"] = int64(entropyValue)
 
 	dataFields := bytes.Fields(data)
 	for i, field := range dataFields {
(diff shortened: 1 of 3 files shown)


danielnelson:
closes #2647

plugins/inputs/system/kernel.go
```diff
@@ -33,13 +35,27 @@ func (k *Kernel) Description() string {
...
+
+	entropyString := string(entropyData)
+	entropyValue, err := strconv.ParseInt(strings.TrimSpace(entropyString), 10, 64)
+	if err != nil {
+		return err
+	}
+
 	fields := make(map[string]interface{})
 
+	fields["entropy_available"] = int64(entropyValue)
```
danielnelson: Let's just call it `entropy_avail` to match what is in proc.

kyzh:
Is there anything that I can do to help this PR to be merged and shipped ?
I could bake a cake, but I'm a long way away from SF to deliver it to your office.

hanej:
I have something like this to collect entropy in the mean time.

**telegraf.conf**
```
[[inputs.exec]]
commands = ["/etc/telegraf/collectors/collect_*.sh"]
data_format = "influx"
timeout = "10s"
```

**shell script**
```
#!/bin/bash

LOCATION=/etc/ansible/facts.d/location.fact

if [ ! -f $LOCATION ]; then
    exit 1
fi

DC=$($LOCATION | grep datacenter | awk '{print $2}' | sed -r 's/"(.*)".*/\1/')
ENV=$($LOCATION | grep '"environment"' | awk '{print $2}' | sed -r 's/"(.*)".*/\1/')
HOST=$(hostname -s)

if [ -z $DC ] || [ -z $ENV ]; then
    exit 1
fi

if [ ! -f /proc/sys/kernel/random/entropy_avail ]; then
    exit 1
fi

VALUE=$(cat /proc/sys/kernel/random/entropy_avail)

echo "system,dc=${DC},env=${ENV},host=${HOST} entropy=${VALUE}"
```

danielnelson:
@gerardocorea92 Do you think you could fix the unit test?

gerardocorea92:
@danielnelson pushed commit to fix CircleCI testing.

PR #3530, opened 2017-11-30
Support I (idle) process state on procfs+Linux

tzz:
`I` is a Linux process state: https://github.com/torvalds/linux/blob/master/fs/proc/array.c#L135

This is a fairly recent change but causes lots of Telegraf log noise with 4.x kernels.

I'll do the rest of the paperwork for the PR if the underlying idea is acceptable.

### Required for all PRs:

### DIFF
--- a/plugins/inputs/system/processes.go
+++ b/plugins/inputs/system/processes.go
@@ -85,6 +85,7 @@ func getEmptyFields() map[string]interface{} {
 		fields["dead"] = int64(0)
 		fields["paging"] = int64(0)
 		fields["total_threads"] = int64(0)
+		fields["idle"] = int64(0)
 	}
 	return fields
 }
@@ -174,6 +175,8 @@ func (p *Processes) gatherFromProc(fields map[string]interface{}) error {
 			fields["stopped"] = fields["stopped"].(int64) + int64(1)
 		case 'W':
 			fields["paging"] = fields["paging"].(int64) + int64(1)
+		case 'I':
+			fields["idle"] = fields["idle"].(int64) + int64(1)
 		default:
 			log.Printf("I! processes: Unknown state [ %s ] in file %s",
 				string(stats[0][0]), filename)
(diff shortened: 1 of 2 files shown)


danielnelson:
:+1:

phemmer:
I think it would be nice to not emit this field on systems which don't support it.

danielnelson:
How can we tell if its supported though, it looks like we are running `ps axo state` and counting the items.

danielnelson:
Oh whoops I was looking at the wrong code :/

danielnelson:
Now I'm looking at `/proc/1/stat`, it seems like its the same problem.  Is there a location we can get the known states from?  It would be great if we can do this, since 0 is not the same as unset.

phemmer:
Not sure. I was hoping for bright ideas, but the only thing I can think of is the kernel version. Which I'm not sure is justified to bother checking when it's just a field that's always 0.

plugins/inputs/system/processes.go
```diff
@@ -70,17 +70,15 @@ func getEmptyFields() map[string]interface{} {
 		"stopped":  int64(0),
 		"running":  int64(0),
 		"sleeping": int64(0),
+		"idle":     int64(0),
 		"total":    int64(0),
 		"unknown":  int64(0),
 	}
 	switch runtime.GOOS {
 	case "freebsd":
-		fields["idle"] = int64(0)
```
danielnelson: I think we should leave these as before, and add "idle" to the linux section.  In my opinion better to report 0 than unset.

tzz:
@danielnelson thanks for the review. I've updated the README and signed the CLA.

I changed it to do the minimum possible, which was to do `I` parsing and stats sending only on Linux, as you and others requested.

I'm not sure how to do the tests, since they will depend on the right version of Linux and you don't have anything in that direction currently in the processes tests. Would you like to add the tests yourself after merge?

Either way, I appreciate the help and attention.

danielnelson:
Thank you!

atc-cegodk:
I see that this got added to the 1.6.0 milestone.

Is there a plan to support I (idle) process state in version 1.4.x?

danielnelson:
I'm not planning to make another release on the 1.4 branch, but I will add this to the upcoming 1.5.0 release.

Issue #3388, opened 2017-10-26
support swing door data compression plugin


baoyicai:
Hi,
I'm wondering if telegraf can support more data compression plugin, the one algorithm I'm particularly interested is the swing door algorithm, it is implemented by OSIsoft PI system, in their interface node, which is very similar to telegraf ( the interface implements data exception algorithm) and the its core database (implements data compression algorithm)

the two algorithms can greatly reduce the data size without loosing too much of its details, and it's configurable so that users can choose how detailed they want to data to be. 

See link below
https://livelibrary.osisoft.com/LiveLibrary/content/en/server-v5/GUID-3BEE22F8-D525-44C1-85DF-41443553E233

danielnelson:
This is an interesting idea, something like this could be implemented as a processor.  One thing to note is that it seems that this could cause aggregation issues.

danielnelson:
To answer you question though, Telegraf currently has no support for this style of compression algorithm.

baoyicai:
Thanks for replying. I believe it will be a great feature for future development. It aims to reduce data size, which is the same as aggregation, but it keeps the details like the data spikes, which is important for analysis.

PR #2846, opened 2017-05-23
Add support for dropwizard format

atzoum:
### Required for all PRs:

- [x] Signed [CLA](https://influxdata.com/community/cla/).
- [x] Associated README.md updated.
- [x] Has appropriate unit tests.

Issue #1758, opened 2016-09-13
XML input


Ismael:
It would be helpful to have an XML input plugin that behaves like the httpjson plugin.
I have a service that only exposes its stats through an xml.

dipenshah90:
Just wanted to know if this is still being considered :)

danielnelson:
If we implement this I think we would define a new xml input data format and the service would be required to produce stats in this format.  Would this work with your use cases @Ismael @dipenshah90 ?

dipenshah90:
Hi @danielnelson by "service would be required to produce stats in this format" you mean the output would be in XML? For my use case, I am getting the data in XML and wanted to feed this data to Influxdb (Inlfuxdb- Grafana). I can get data in JSON, but, I need to add an extra header "Accept: application/json" which gives me a parsing error. I assume the option to add these extra headers are not available in httpjson?

Ismael:
I don't have control over the XML produced by the service, so it wouldn't work for me.

danielnelson:
@Ismael Can you show an example of your xml format?  I wonder if there is a good xml technology for parsing custom xml output to metrics.

danielnelson:
@dipenshah90 You can add arbitrary headers, check the plugin docs.

Ismael:
Here's an example

```
<WowzaStreamingEngine>
<ConnectionsCurrent>0</ConnectionsCurrent>
<ConnectionsTotal>0</ConnectionsTotal>
<ConnectionsTotalAccepted>0</ConnectionsTotalAccepted>
<ConnectionsTotalRejected>0</ConnectionsTotalRejected>
<MessagesInBytesRate>0.0</MessagesInBytesRate>
<MessagesOutBytesRate>0.0</MessagesOutBytesRate>
<VHost>
<Name>_defaultVHost_</Name>
<TimeRunning>64679.885</TimeRunning>
<ConnectionsLimit>0</ConnectionsLimit>
<ConnectionsCurrent>0</ConnectionsCurrent>
<ConnectionsTotal>0</ConnectionsTotal>
<ConnectionsTotalAccepted>0</ConnectionsTotalAccepted>
<ConnectionsTotalRejected>0</ConnectionsTotalRejected>
<MessagesInBytesRate>0.0</MessagesInBytesRate>
<MessagesOutBytesRate>0.0</MessagesOutBytesRate>
</VHost>
</WowzaStreamingEngine>
```

Ismael:
[...]

Issue #813, opened 2016-03-09
Add input to collect http data using any parser


davidbirdsong:
We have services that have implemented native influxdb data formats and expose them at scrapable endpoints. Instead of pushing these values directly to influxdb, we want telegraf to grab them which gives us all the flexibility that telegraf provides: prometheus scrape endpoint, kafka push, nsq, or push to influxdb.

Is there no way to scrape influx's native format?

OrangeTux:
I don't know whether I understand you correct, but you want Telegraf call HTTP endpoints that return data formatted according to InfluxDB's line protocol?

You could use the [exec](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/exec) input plugin.

```
[[inputs.exec]]
 commands = ["curl example.com"]
 data_format = "influx"
```

davidbirdsong:
Yes.

I'm sure it sounds odd at first blush, but I'll point out that telegraf is useful beyond collection and forwarding of metrics. It's the best so far, IMO, at providing a metrics router that allows external tools to intermingle data that most services or libraries don't want or need to be concerned with. For instance, we use ansible to drop data into the telegraf include dirs that merges inventory data like machines groups, tags, state, and other k/v data into metrics as they flow through--sub in any other config management or service discovery for similar approaches.

I was originally looking to heka for it's general purpose message router to mix in arbitrary data, but telegraf reads and writes to so many of the systems we use and in a metrics format that makes the most sense.

Given this approach, it makes sense for services to choose the influx line format regardless of whether they choose to push or pull.

something like
```toml
[[inputs.httpprobe]]
 data_format = "influx"
  servers = [
    "http://my.service.com/_stats"
  ]
```
 I'm happy to whip up a PR in the next week if it sounds like an acceptable idea.

OrangeTux:
I don't quite understand this line: 
>  it makes sense for services to choose the influx line format regardless of whether they choose to push or pull.

An input plugin fetches metrics from a source every once in a while. I think this is what you mean with 'pull'. The `exec` plug is an input plugin.   

A service input plugin is a special type of input plugin which runs in the background. This type of plugin can be used by externel tools to push metrics to. An example is the [TCP Listener](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/tcp_listener). This service input plugin listens on a port for metrics. Whould this plugin, or a new HTTP Listener plugin, solve your problem?

davidbirdsong:
[...]

PR #3531, opened 2017-12-01
add support for pulling metrics from a dropwizard application

grange74:
### Required for all PRs:

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,6 +144,7 @@ configuration options.
 * [dns query time](./plugins/inputs/dns_query)
 * [docker](./plugins/inputs/docker)
 * [dovecot](./plugins/inputs/dovecot)
+* [dropwizard](./plugins/inputs/dropwizard)
 * [elasticsearch](./plugins/inputs/elasticsearch)
 * [exec](./plugins/inputs/exec) (generic executable plugin, support JSON, influx, graphite and nagios)
 * [fail2ban](./plugins/inputs/fail2ban)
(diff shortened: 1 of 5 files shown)


danielnelson:
We have an open pull request for adding dropwizard as a parser, I think this is the best solution for dropwizard because it decouples the data format from the transport.  I encourage you to take a look at the pull request and comment on if it will work or any ideas you have:

https://github.com/influxdata/telegraf/pull/2846

I think the other half of this that we need is a http input plugin, similar to httpjson, but allowing any parser to be used:

https://github.com/influxdata/telegraf/issues/813
https://github.com/influxdata/telegraf/issues/1758#issuecomment-304418082

grange74:
Ok, separating the data format from the transport makes a lot of sense.

I guess in the long term a lot of the input plugins could be converted into parsers as many of the ones I looked at to model off this off just pulled json (or xml) from a HTTP endpoint (e.g. influxdb, kapacitor, tomcat). 
Having a simple http input plugin that supports custom parsers and various forms of authentication would be nice.
Please let me know if there is anything that I can help with.

One feature that was implemented in this PR, was to skip Idle metrics. That doesn't sound like something that should go into a parser. Would you see this as a generic feature and if so where would it fit in?

danielnelson:
For idle metrics, the best idea I have seen is the idea in #3388.

danielnelson:
I'm not aware of anyone working on the generic http input plugin, it would probably be fairly similar to this pull request without the dropwizard code but using the Parser interface.  Would you be interested in helping with this?

grange74:
Sure, i'd be happy to help with the "generic http input plugin". I had a quick look at some of the existing input plugins that implement the Parser interface (e.g. nsq_consumer) and they seem very straight forward. 

Given that there are already several http input plugins (i.e. http_listener, http_response, httpjson), what would you want this one called? just http or something else?

We can probably close this PR now and create a new issue to describe the new "generic http input plugin" to document the desired behaviour.

danielnelson:
We can use https://github.com/influxdata/telegraf/issues/813 for tracking the issue.  I think we should just name it `http`.

Issue #2506, opened 2017-03-07
Feature request: Elasticsearch extensible for aggregation queries 


lpic10:
## Feature Request

Similar to postgresql_extensible, allow to define ES aggregation queries in that the results (aggregated data) can feed the different outputs.

It is usual in Elasticsearch, when used as log management tool, to limit the number of days of logs to keep in order to limit storage consumption. But there is sometimes relevant information in these logs that may be necessary for a longer period, eg. if that information is used in a set of graphs or the information is used for anomaly detection and other monitoring needs. This plugin could be used to extract/calculate some useful metrics from events, logs and documents indexed in ES and store these metrics into a TSDB such as InfluxDB. 

The idea here is to get some early feedback about interest/feature set/etc. I have something similar to this for my own needs and I may submit a PR when it is more "ready" (code a bit more clear and documented).

### Proposal:
Each aggregation query could be defined by the following parameters:

* name of the aggregation: string used for the measurement name
* query: a lucene query to filter the data before aggregation
* metrics aggregation function: one of the supported [functions ](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-metrics.html) (it can be limited to few functions initially, as avg/sum/min/max/count)
* metric fields: list of the fields in a document to apply the metrics aggregation function
* tags: list of fields in a document to apply a "terms" aggregation (thus yielding results per tag, eg. per hostname and per FS path) 
* period: period of time of the aggregation (eg. "5m" will perform aggregation over last 5 minutes)
* datetime field: name of the datetime field in the ES document (eg. @timestamp)

### Use cases:

Calculate time-based metrics from events stored in ES, such as logs, events and metrics. 

* Count number of a particular log message per minute, aggregating per host or per instance
* Calculate the average response time and bytes from logs, per status code  
* Calculate the max value of a particular field seen in last 5 minutes, per use-case and per host

sureshsubramaniam:
Would be interested to see this feature in telegraf

sureshsubramaniam:
👍👍

brice86:
+! This would be awesome!

lpic10:
So, another unintended use-case of this plugin is to allow alarming over data stored in Elasticsearch (this is a paid feature of Elastic stack), we just figured this out using this internally for few use cases. 

[...]

PR #3536, opened 2017-12-03
Add Elasticsearch query input plugin

lpic10:
### Required for all PRs:

This implements #2506 

I've been using something similar to this code for some time already, I finally decided to clean it up a bit and add some tests and documentation.


lpic10:
Some to-do tasks in my list: 

- [ ] Support for Elasticsearch 6.x
- [ ] Support for percentiles aggregation

plugins/inputs/elasticsearch_query/elasticsearch_query.go
```diff
@@ -0,0 +1,287 @@
...
+  enable_sniffer = false
+  ## Set the interval to check if the Elasticsearch nodes are available
+  ## Setting to "0s" will disable the health check (not recommended in production)
+  health_check_interval = "10s"
+  ## HTTP basic authentication details (eg. when using x-pack)
+  # username = "telegraf"
+  # password = "mypassword"
+
+  ## Optional SSL Config
+  # ssl_ca = "/etc/telegraf/ca.pem"
```
glinton: Use `tls_[ca|cert|key]`

plugins/inputs/elasticsearch_query/elasticsearch_query.go
```diff
@@ -0,0 +1,287 @@
...
+type ElasticsearchQuery struct {
+	URLs                []string `toml:"urls"`
+	Username            string
+	Password            string
+	EnableSniffer       bool
+	Tracelog            bool
+	Timeout             internal.Duration
+	HealthCheckInterval internal.Duration
+	Aggregations        []Aggregation `toml:"aggregation"`
+	SSLCA               string        `toml:"ssl_ca"`   // Path to CA file
```
glinton: No need for these SSL/TLS attributes if you use `./internal/tls` and the shared tls config [like apache uses](https://github.com/influxdata/telegraf/blob/master/plugins/inputs/apache/apache.go#L96)

plugins/inputs/elasticsearch_query/elasticsearch_query.go
```diff
@@ -0,0 +1,287 @@
...
+
+	err := e.connectToES()
+	return err
+}
+
+func (e *ElasticsearchQuery) connectToES() error {
+
+	var clientOptions []elastic.ClientOptionFunc
+
+	tlsCfg, err := internal.GetTLSConfig(e.SSLCert, e.SSLKey, e.SSLCA, e.InsecureSkipVerify)
```
glinton: Use `internal/tls` and `TlsConfig()`

plugins/inputs/elasticsearch_query/elasticsearch_query.go
```diff
@@ -0,0 +1,287 @@
...
+	if err != nil {
+		return err
+	}
+	tr := &http.Transport{
```
glinton: Make the transport and http.Client a member of `ElasticsearchQuery`, again similar to [the apache plugin](https://github.com/influxdata/telegraf/blob/master/plugins/inputs/apache/apache.go#L69)

lpic10:
Thanks @glinton for the initial review, I've implemented the changes concerning TLS, the sample config and the README with examples. 

In addition I removed some commented code, and I changed the check for version to allow it to work on newer ES (6.x,) which I run the tests against and it seems to be working fine..

gl1ridae:
Hi! Is there any chanse this MR will be merged in the next version? Any additional help required on this?
@danielnelson

ddemin:
Really need this too =)

vishrantgupta:
Hi @lpic10 @glinton, any plans when it can be merged?

srebhan:
@lpic10 are you still interested in getting this merged?

hexinal:
@srebhan 
not op, but I maintain a "fork" of this in production so there is at least one person who would be interested in getting this in upstream :smile: . 
Let me know if I can help.

lpic10:
> @lpic10 are you still interested in getting this merged?

Yes, even if I'm not actively using this PR as my job attributions have changed since.
I will try to have a look at it during the weekend.

srebhan:
@hexinal please coordinate with @lpic10. You will be definitively required for testing... ;-)

plugins/inputs/elasticsearch_query/README.md
```diff
@@ -0,0 +1,137 @@
...
+  ## Fields to be used as tags
+  ## Must be non-analyzed fields, aggregations are performed per tag
+  tags = ["field.keyword", "field2.keyword"]
+  ## Set to true to not ignore documents when the tag(s) above are missing
+  include_missing_tag = true
+  ## String value of the tag when the tag does not exist
+  ## Used when include_missing_tag is true
+  missing_tag_value = "null"
+
+```
```
srebhan, line 63: Could you please make this a minimal useful version of the config. That is leave out as many options as possible and just keep what _has to_ be changed by the user. This way it is easier to for new users to start.

plugins/inputs/elasticsearch_query/README.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Examples
+
+### Search the average response time, per URI and per response status code
+
+```toml
+[[inputs.elasticsearch_query.aggregation]]
```
srebhan, line 70: Maybe you could add a not that the `[[inputs.elasticsearch_query]]` section is still required. :-)

plugins/inputs/elasticsearch_query/aggregation_parser.go
```diff
@@ -0,0 +1,117 @@
+package elasticsearch_query
+
+import elastic "gopkg.in/olivere/elastic.v5"
```
srebhan, line 3: Would it make sense to use the official client at https://github.com/elastic/go-elasticsearch?

plugins/inputs/elasticsearch_query/aggregation_parser.go
```diff
@@ -0,0 +1,117 @@
...
+	for _, aggregationQuery := range *aggregationQueryList {
+		m := measurements[aggregationQuery.measurement]
+		a := aggKey{function: aggregationQuery.function, field: aggregationQuery.field, name: aggregationQuery.name}
+
+		if m == nil {
+			m = []*aggKey{&a}
+		} else {
+			m = append(m, &a)
+		}
+		measurements[aggregationQuery.measurement] = m
```
srebhan, line 18: How about
```suggestion
		a := aggKey{function: aggregationQuery.function, field: aggregationQuery.field, name: aggregationQuery.name}

		if m := measurements[aggregationQuery.measurement]; m != nil {
			measurements[aggregationQuery.measurement] = append(m, &a)
		} else {
			measurements[aggregationQuery.measurement] = []*aggKey{&a}
		}
```

plugins/inputs/elasticsearch_query/aggregation_parser.go
```diff
@@ -0,0 +1,117 @@
...
+	return nil
+}
+
+func (e *ElasticsearchQuery) recurseResponse(aggKeys []*aggKey, aggs elastic.Aggregations, measurementName string, fields map[string]interface{}, tags map[string]string) error {
+
+	for _, aggKey := range aggKeys {
+		resp := e.getResponseAggregation(aggKey.function, aggKey.name, aggs)
+
+		if resp != nil {
+
```
srebhan, line 52: Please do
```suggestion

		if resp == nil {
			// Maybe do some logging here if this is an error...
			continue
		}

```

plugins/inputs/elasticsearch_query/aggregation_parser.go
```diff
@@ -0,0 +1,117 @@
...
+			switch resp := resp.(type) {
+			case *elastic.AggregationValueMetric:
+
+				// we've found a metric aggregation, add to field map
+				fields[aggKey.name] = float64(0)
+
+				if resp.Value != nil {
+					fields[aggKey.name] = *resp.Value
+				}
```
srebhan, line 61: Here you might use an else clause to not set the value twice.

plugins/inputs/elasticsearch_query/aggregation_parser.go
```diff
@@ -0,0 +1,117 @@
...
+
+			case *elastic.AggregationBucketKeyItems:
+				// we've found a terms aggregation, iterate over the buckets and try to retrieve the inner aggregation values
+				for item := range resp.Buckets {
```
srebhan, line 65: Don't you get a bucket when iterating over the buckets ([reference](https://github.com/olivere/elastic/blob/be977d75416e9a352a0853165c9d1786e3d43c51/search_aggs.go#L1251))?

plugins/inputs/elasticsearch_query/aggregation_parser.go
```diff
@@ -0,0 +1,117 @@
...
+					bucket := resp.Buckets[item]
+
+					fields["doc_count"] = bucket.DocCount
+					tags[aggKey.name] = bucket.Key.(string)
```
srebhan, line 69: Please check this type assertion as otherwise telegraf will panic when we encounter a non-string key.

plugins/inputs/elasticsearch_query/aggregation_parser.go
```diff
@@ -0,0 +1,117 @@
...
+		}
+		measurements[aggregationQuery.measurement] = m
+	}
+
+	// recurse over aggregation results
+	for measurementName, aggKey := range measurements {
+		fields := make(map[string]interface{})
+		tags := make(map[string]string)
+
+		err := e.recurseResponse(aggKey, searchResult.Aggregations, measurementName, fields, tags)
```
srebhan, line 26: Please put a comment here that you need to recursively resolve the fields and tags as otherwise it's strange to pass the data-structures down without using them afterwards.

plugins/inputs/elasticsearch_query/aggregation_parser.go
```diff
@@ -0,0 +1,117 @@
...
+
+					err := e.recurseResponse(aggKeys, bucket.Aggregations, measurementName, fields, tags)
+					if err != nil {
+						return err
+					}
+				}
+				return nil
+
+			case *elastic.AggregationPercentilesMetric:
+				// TODO
```
srebhan, line 79: :-) So please do or delete the case. :-)

plugins/inputs/elasticsearch_query/aggregation_parser.go
```diff
@@ -0,0 +1,117 @@
...
+				return nil
+			}
```
srebhan, line 81: You should add a `default` section and return an error for cases where you cannot handle the type. This catches problems where types are added to the client but we forgot to add them here.

plugins/inputs/elasticsearch_query/aggregation_parser.go
```diff
@@ -0,0 +1,117 @@
...
+		}
+	}
+
+	if len(fields) > 0 {
+		e.acc.AddFields(measurementName, fields, tags)
+	}
```
srebhan, line 87: I think this one has to go one level up. We want to collect the fields and values for the measurement/metric here and then add the metric when complete in `parseAggregationResult()`.

plugins/inputs/elasticsearch_query/aggregation_parser.go
```diff
@@ -0,0 +1,117 @@
...
+
+					fields["doc_count"] = bucket.DocCount
+					tags[aggKey.name] = bucket.Key.(string)
+
+					err := e.recurseResponse(aggKeys, bucket.Aggregations, measurementName, fields, tags)
+					if err != nil {
+						return err
+					}
+				}
+				return nil
```
srebhan, line 76: Is this one really correct? You will ignore all other responses after you've hit one of the `elastic.AggregationBucketKeyItems` type. That seems wrong.

plugins/inputs/elasticsearch_query/aggregation_parser.go
```diff
@@ -0,0 +1,117 @@
...
+		agg, found = aggs.Sum(aggName)
+	case "min":
+		agg, found = aggs.Min(aggName)
+	case "max":
+		agg, found = aggs.Max(aggName)
+	case "terms":
+		agg, found = aggs.Terms(aggName)
+		// TODO:
+		// case "percentile":
+		// 	agg, found = aggs.Percentiles(aggName)
```
srebhan, line 109: Please remove the TODOs.

plugins/inputs/elasticsearch_query/aggregation_query.go
```diff
@@ -0,0 +1,184 @@
...
+	"strings"
+	"time"
+
+	elastic "gopkg.in/olivere/elastic.v5"
+)
+
+func (e *ElasticsearchQuery) runAggregationQuery(ctx context.Context, aggregation Aggregation, aggregationQueryList []aggregationQueryData) (*elastic.SearchResult, error) {
+
+	now := time.Now().UTC()
+	from := now.Add(aggregation.QueryPeriod.Duration * -1)
```
srebhan, line 15:
```suggestion
	from := now.Sub(aggregation.QueryPeriod.Duration)
```

plugins/inputs/elasticsearch_query/aggregation_query.go
```diff
@@ -0,0 +1,184 @@
...
+	// add only parent elastic.Aggregations to the search request, all the rest are subaggregations of these
+	for _, v := range aggregationQueryList {
+		if v.isParent && v.aggregation != nil {
+			search.Aggregation(v.aggKey.name, v.aggregation)
+		}
+	}
+	searchResult, err := search.Do(ctx)
+
+	if err != nil {
+		if searchResult != nil {
```
srebhan, line 40: Please fuse those two ifs.

plugins/inputs/elasticsearch_query/aggregation_query.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+
+	query := elastic.NewBoolQuery()
+	query = query.Filter(elastic.NewQueryStringQuery(filterQuery))
+	query = query.Filter(elastic.NewRangeQuery(aggregation.DateField).From(from).To(now))
+
+	search := e.ESClient.Search().
+		Index(aggregation.Index).
+		Query(query).
+		Size(0)
```
srebhan, line 29: Put in one line please.

plugins/inputs/elasticsearch_query/aggregation_query.go
```diff
@@ -0,0 +1,184 @@
...
+
+		resp, err := e.ESClient.GetFieldMapping().Index(aggregation.Index).Field(metricField).Do(ctx)
+
+		if err != nil {
+			return mapMetricFields, err
+		}
+
+		for _, index := range resp {
+			for _, _type := range index.(map[string]interface{})["mappings"].(map[string]interface{}) {
+				for _, field := range _type.(map[string]interface{}) {
```
srebhan, line 72: Please check these type assertions!

plugins/inputs/elasticsearch_query/aggregation_query.go
```diff
@@ -0,0 +1,184 @@
...
+					fname := field.(map[string]interface{})["full_name"].(string)
+					for _, fieldType := range field.(map[string]interface{})["mapping"].(map[string]interface{}) {
+						ftype = fieldType.(map[string]interface{})["type"].(string)
```
srebhan, line 75: Same here.

plugins/inputs/elasticsearch_query/aggregation_query.go
```diff
@@ -0,0 +1,184 @@
...
+		agg = elastic.NewSumAggregation().Field(aggfield)
+	case "min":
+		agg = elastic.NewMinAggregation().Field(aggfield)
+	case "max":
+		agg = elastic.NewMaxAggregation().Field(aggfield)
+	// TODO:
+	// case "percentile":
+	// 	agg = elastic.NewPercentilesAggregation().Field(aggfield)
+	default:
+		return agg, fmt.Errorf("Aggregation function %s not supported", function)
```
srebhan, line 146: In case you error please return `nil` for the value if possible. This way you directly see that you cannot rely on the value and do not need to trace back the variable.

plugins/inputs/elasticsearch_query/elasticsearch_query.go
```diff
@@ -0,0 +1,277 @@
...
+
+	elastic "gopkg.in/olivere/elastic.v5"
+
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/internal"
+	"github.com/influxdata/telegraf/internal/tls"
+	"github.com/influxdata/telegraf/plugins/inputs"
+)
+
+const description = `Queries Elasticseach`
```
srebhan, line 22: Maybe "Derive metrics from aggregating Elasticsearch query results"?

plugins/inputs/elasticsearch_query/elasticsearch_query.go
```diff
@@ -0,0 +1,277 @@
...
+const sampleConfig = `
+  ## The full HTTP endpoint URL for your Elasticsearch instance
+  ## Multiple urls can be specified as part of the same cluster,
+  ## this means that only ONE of the urls will be written to each interval.
+  urls = [ "http://node1.es.example.com:9200" ] # required.
+  ## Elasticsearch client timeout, defaults to "5s" if not set.
```
srebhan, line 28: Please add a blank line between option groups.

plugins/inputs/elasticsearch_query/elasticsearch_query.go
```diff
@@ -0,0 +1,277 @@
...
+  timeout = "5s"
```
srebhan, line 29: Please comment default settings. Use the default setting as value. This way the user sees what is used but also knows that the option is optional.

plugins/inputs/elasticsearch_query/elasticsearch_query.go
```diff
@@ -0,0 +1,277 @@
...
+  ## Set to true to ask Elasticsearch a list of all cluster nodes,
+  ## thus it is not necessary to list all nodes in the urls config option
+  enable_sniffer = false
```
srebhan, line 32: Maybe `cluster_node_discovery`? We do not really sniff do we?

plugins/inputs/elasticsearch_query/elasticsearch_query.go
```diff
@@ -0,0 +1,277 @@
...
+  ## Valid values are: avg, sum, min, max, sum
+  metric_function = "avg"
+  ## Fields to be used as tags
+  ## Must be non-analyzed fields, aggregations are performed per tag
+  tags = ["field.keyword", "field2.keyword"]
+  ## Set to true to not ignore documents when the tag(s) above are missing
+  include_missing_tag = true
+  ## String value of the tag when the tag does not exist
+  ## Used when include_missing_tag is true
+  missing_tag_value = "null"
```
srebhan, line 70: Comment optional parameters in the form
```
## Some description
# parameter = "default value"
```

plugins/inputs/elasticsearch_query/elasticsearch_query.go
```diff
@@ -0,0 +1,277 @@
...
+`
+
+type ElasticsearchQuery struct {
+	URLs                []string `toml:"urls"`
+	Username            string
+	Password            string
+	EnableSniffer       bool
+	Tracelog            bool
+	Timeout             internal.Duration
+	HealthCheckInterval internal.Duration
```
srebhan, line 80: Please add toml annotations. This eases debugging in case you are searching for the use of an option.

plugins/inputs/elasticsearch_query/elasticsearch_query.go
```diff
@@ -0,0 +1,277 @@
...
+	Aggregations        []Aggregation `toml:"aggregation"`
+	tls.ClientConfig
+	httpclient *http.Client
+	ESClient   *elastic.Client
```
srebhan, line 84: Do we really need to export this?

plugins/inputs/elasticsearch_query/elasticsearch_query.go
```diff
@@ -0,0 +1,277 @@
...
+	acc        telegraf.Accumulator
```
srebhan, line 85: Usually the accumulator is passed down the functions and not stored in the structure. If you do not have a striking reason please follow this unwritten rule. :-)

plugins/inputs/elasticsearch_query/elasticsearch_query.go
```diff
@@ -0,0 +1,277 @@
...
+}
+
+type Aggregation struct {
```
srebhan, line 88: Can we make this struct private and if not can we use a more Elasticsearch-specific name!?

plugins/inputs/elasticsearch_query/elasticsearch_query.go
```diff
@@ -0,0 +1,277 @@
...
+	Index             string
+	MeasurementName   string
+	FilterQuery       string
+	QueryPeriod       internal.Duration
+	MetricFields      []string `toml:"metric_fields"`
+	DateField         string
+	Tags              []string `toml:"tags"`
+	IncludeMissingTag bool
+	MissingTagValue   string
+	MetricFunction    string
```
srebhan, line 98: Please also add toml tags.

plugins/inputs/elasticsearch_query/elasticsearch_query.go
```diff
@@ -0,0 +1,277 @@
...
+	name        string
+	function    string
+	field       string
+}
+
+type aggregationQueryData struct {
+	aggKey
+	isParent    bool
+	aggregation elastic.Aggregation
+}
```
srebhan, line 112: Please move these definitions to one of the `aggregation_...` files as you don't use them here.

plugins/inputs/elasticsearch_query/elasticsearch_query.go
```diff
@@ -0,0 +1,277 @@
...
+
+func (e *ElasticsearchQuery) SampleConfig() string {
+	return sampleConfig
+}
+
+func (e *ElasticsearchQuery) Description() string {
+	return description
+}
+
+func (e *ElasticsearchQuery) init() error {
```
srebhan, line 122: If you call this function `Init()` it will be automatically called by telegraf after the config parameters are parsed and set.

plugins/inputs/elasticsearch_query/elasticsearch_query.go
```diff
@@ -0,0 +1,277 @@
...
+	}
+
+	clientOptions = append(clientOptions,
+		elastic.SetHttpClient(e.httpclient),
+		elastic.SetSniff(e.EnableSniffer),
+		elastic.SetURL(e.URLs...),
+		elastic.SetHealthcheckInterval(e.HealthCheckInterval.Duration),
+	)
+
+	if e.Username != "" && e.Password != "" {
```
srebhan, line 150: This would prevent an empty password. If an empty password is valid use
```suggestion
	if e.Username != "" {
```
instead.

plugins/inputs/elasticsearch_query/elasticsearch_query.go
```diff
@@ -0,0 +1,277 @@
...
+		clientOptions = append(clientOptions,
+			elastic.SetBasicAuth(e.Username, e.Password),
```
srebhan, line 152: No line wrap here.

plugins/inputs/elasticsearch_query/elasticsearch_query.go
```diff
@@ -0,0 +1,277 @@
...
+	return description
+}
+
+func (e *ElasticsearchQuery) init() error {
+	if e.URLs == nil {
+		return fmt.Errorf("Elasticsearch urls is not defined")
+	}
+
+	err := e.connectToES()
+	return err
```
srebhan, line 128:
```suggestion
	return e.connectToES()
```

plugins/inputs/elasticsearch_query/elasticsearch_query.go
```diff
@@ -0,0 +1,277 @@
...
+		clientOptions = append(clientOptions,
+			elastic.SetHealthcheck(false),
+		)
+	}
+
+	if e.Tracelog {
+		clientOptions = append(clientOptions,
+			elastic.SetTraceLog(log.New(os.Stdout, "", log.LstdFlags)),
+		)
+	}
```
srebhan, line 166: Where is this option set? Usually logging to stdout is an absolute exception. Can we remove this? Or at least guard with a config option even though I prefer the removal.

plugins/inputs/elasticsearch_query/elasticsearch_query.go
```diff
@@ -0,0 +1,277 @@
...
+
+	// check for ES version on first node
+	esVersion, err := client.ElasticsearchVersion(e.URLs[0])
+
+	if err != nil {
+		return fmt.Errorf("Elasticsearch query version check failed: %s", err)
+	}
+
+	// quit if ES version is not supported
+	i, err := strconv.Atoi(strings.Split(esVersion, ".")[0])
```
srebhan, line 181: Please check if the split returned anything, otherwise telegraf will panic here.

plugins/inputs/elasticsearch_query/elasticsearch_query.go
```diff
@@ -0,0 +1,277 @@
...
+	if err != nil || i < 5 {
+		return fmt.Errorf("Elasticsearch query: ES version not supported: %s", esVersion)
```
srebhan, line 183: Errors have to start lower-case. Please also change the other instances. Furthermore it would be nice here to specify the supported version. Something like "ES versions < 5.x are not supported".

plugins/inputs/elasticsearch_query/elasticsearch_query.go
```diff
@@ -0,0 +1,277 @@
...
+	}
+
+	e.ESClient = client
+	return nil
+}
+
+func (e *ElasticsearchQuery) Gather(acc telegraf.Accumulator) error {
+	if err := e.init(); err != nil {
+		return err
+	}
```
srebhan, line 193: Please rename the function above to `Init()` and remove this code.

plugins/inputs/elasticsearch_query/elasticsearch_query.go
```diff
@@ -0,0 +1,277 @@
...
+
+	e.acc = acc
+
+	var wg sync.WaitGroup
+
+	for _, agg := range e.Aggregations {
+		wg.Add(1)
+		go func(agg Aggregation) {
+			defer wg.Done()
+			err := e.esAggregationQuery(agg)
```
srebhan, line 203: Please pass the accumulator down here instead of storing it in the structure.

plugins/inputs/elasticsearch_query/elasticsearch_query.go
```diff
@@ -0,0 +1,277 @@
...
+
+func (e *ElasticsearchQuery) esAggregationQuery(aggregation Aggregation) error {
+
+	ctx, cancel := context.WithTimeout(context.Background(), e.Timeout.Duration)
+	defer cancel()
+
+	if aggregation.DateField == "" {
+		return fmt.Errorf("Field 'date_field' not set")
+	}
+
```
srebhan, line 239: Maybe check this before constructing the context.

plugins/inputs/elasticsearch_query/elasticsearch_query.go
```diff
@@ -0,0 +1,277 @@
...
+	mapMetricFields, err := e.getMetricFields(ctx, aggregation)
+	if err != nil {
+		return err
+	}
+
+	aggregationQueryList, err := e.buildAggregationQuery(mapMetricFields, aggregation)
+	if err != nil {
+		return err
+	}
+
```
srebhan, line 249: Do you really need to do this in every gather cycle or is it sufficient to do it once at the beginning and only set the timeinterval here?

plugins/inputs/elasticsearch_query/elasticsearch_query.go
```diff
@@ -0,0 +1,277 @@
...
+		Transport: tr,
+		Timeout:   e.Timeout.Duration,
+	}
+
+	return httpclient, nil
+}
+
+func (e *ElasticsearchQuery) esAggregationQuery(aggregation Aggregation) error {
+
+	ctx, cancel := context.WithTimeout(context.Background(), e.Timeout.Duration)
```
srebhan, line 233: This way the timeout is for both operations that query the server (`getMetricFields()` and `runAggreagtionQuery()`). Is this intentional?

plugins/inputs/elasticsearch_query/elasticsearch_query.go
```diff
@@ -0,0 +1,277 @@
...
+			return err
+		}
+	} else {
+		err = e.parseSimpleResult(aggregation.MeasurementName, searchResult)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
```
srebhan, line 267: You can shorten to
```suggestion
	if searchResult.Aggregations != nil {
		err = e.parseAggregationResult(&aggregationQueryList, searchResult)
	} else {
		err = e.parseSimpleResult(aggregation.MeasurementName, searchResult)
	}

	return err
```
But that's a matter of taste...

srebhan (changes requested):
First of all thank you very much for submitting this @lpic10! Here comes a quick first round...

Please check the `recurseResponse()` function thoroughly as it looks fishy to me. That being said, it would be nice if you can come up with some test-cases that solely test the parsing part (`recurseResponse()` and friends) without the need of a server. My experience is that this are usually the tricky parts that break on changes.

Furthermore, make use of the `Init()` function that is automatically called by telegraf once after passing in the config variables. Use this function to prepare as much of your queries as possible and to validate user settings.

As an idea rather than a hard requirement I want to suggest to remember the last query time and dynamically adapt the aggregation window to the time between last run and now. This should be the default IMO and there should be the possibility to override this by the `query_period`.

Looking forward to the next round!

plugins/inputs/elasticsearch_query/aggregation_parser.go
lpic10: That would require a complete rewrite of the code I'm afraid.
Also this client is the same already in use on elasticsearch output plugin.

The fact that it is using a rather "old" (v5) version maybe is an issue or will become one soon.
One option would be to import multiple versions (v5, v6, v7...) and use the right client according to the version of the cluster detected.

plugins/inputs/elasticsearch_query/elasticsearch_query.go
lpic10: This setting is coming from the client: https://github.com/olivere/elastic/wiki/Sniffing
I'd be fine by removing or hiding this (and health check).

plugins/inputs/elasticsearch_query/elasticsearch_query.go
lpic10: I think I did that way because of the `recurseResponse` function. I thought it was better to have it "globally" than passing down each recursion. I will try to revisit that.

plugins/inputs/elasticsearch_query/aggregation_parser.go
lpic10: I think if we move up the metric may not be complete when adding to the Accumulator, I will re-check that

plugins/inputs/elasticsearch_query/elasticsearch_query.go
lpic10: That is for debugging the client, I agree it should not be in the release version as it pollutes the stdout and that's why I hid it from the config.

plugins/inputs/elasticsearch_query/aggregation_parser.go
srebhan: Fair enough. Let's start with this dependency then and maybe switch to the other client with higher versions...

plugins/inputs/elasticsearch_query/elasticsearch_query.go
srebhan: Yeah I'm only asking because the name is not very self-containing and it's barely expresses what it does. Furthermore, sniffing might have a negative co-notation... But if you want to keep it, it's up to you.

plugins/inputs/elasticsearch_query/elasticsearch_query.go
srebhan: Ok that is a point, but maybe you can change the recursive function in a way that it returns what should be added to `acc` and not doing it while traversing the tree...

plugins/inputs/elasticsearch_query/elasticsearch_query.go
srebhan: I see. However, I'd like this removed from the code that goes to telegraf. If possible and necessary add some debug-logging using the telegraf logger...

lpic10:
> First of all thank you very much for submitting this @lpic10! Here comes a quick first round...

Thanks for the review, very much needed!

I think I touched almost everything you mentioned. Except maybe implementing tests on the `recurseResponse` without the use of a server. 

This contribution was done few years ago, and because of that it is not supporting latest releases of Elasticsearch (>= 7.x)

Currently telegraf is using the elasticsearch version 7.2.0 for the integration tests, and this won't work with this plugin.
Maybe it is time already to upgrade the go elastic client to v7 for this (and the output plugin).

Another option would be to import v5, v6 and v7 client versions and use the appropriate one depending on the version of the cluster. This could be set via config parameter, or autodetected.

> As an idea rather than a hard requirement I want to suggest to remember the last query time and dynamically adapt the aggregation window to the time between last run and now. This should be the default IMO and there should be the possibility to override this by the `query_period`.

Is there any example of plugin using such mechanism?

> Looking forward to the next round!

srebhan:
> Currently telegraf is using the elasticsearch version 7.2.0 for the integration tests, and this won't work with this plugin.
Maybe it is time already to upgrade the go elastic client to v7 for this (and the output plugin).
Another option would be to import v5, v6 and v7 client versions and use the appropriate one depending on the version of the cluster. This could be set via config parameter, or autodetected.

This depends a bit on what you want. The approach using multiple client versions is preferable IMO as it allows to keep up with ElasticSearch releases in the future. But I'm also happy if you want to get this in first and "just" prepare later addition of v6 and v7.

Regarding the tests you could setup a `httptest` server (see `http/http_test.go` for an example) and return a prerecorded response for testing. This would eliminate the requirement for a real ES instance (keep this for long-tests :-)).

> Is there any example of plugin using such mechanism?
You might check `sysstat/sysstat.go`. It's just you remember the time of the last collection and use the time difference. However, please keep the user-option to override that mechanism... I guess I have to submit a PR to let plugins know the actually configured telegraf interval...

.vscode/launch.json
```diff
@@ -0,0 +1,19 @@
+{
```
srebhan, line 1: Please remove this file. :-)

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
srebhan, line 1: Same here. Please remove the `.vscode` subfolder and maybe add it to `gitignore`.

plugins/inputs/elasticsearch_query/README.md
```diff
@@ -0,0 +1,98 @@
...
+  # username = "telegraf"
+  # password = "mypassword"
+
+  ## Optional SSL Config
+  # ssl_ca = "/etc/telegraf/ca.pem"
+  # ssl_cert = "/etc/telegraf/cert.pem"
+  # ssl_key = "/etc/telegraf/key.pem"
+  ## Use SSL but skip chain & host verification
+  # insecure_skip_verify = false
+
```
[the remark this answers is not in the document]
srebhan: Please close this as resolved.

plugins/inputs/elasticsearch_query/README.md
```diff
@@ -0,0 +1,156 @@
...
+  # include_missing_tag = false
+
+  ## String value of the tag when the tag does not exist
+  ## Used when include_missing_tag is true
+  # missing_tag_value = "null"
+
+```
+
+## Examples
+
```
srebhan, line 85: Please add
```suggestion
Please note that the `[[inputs.elasticsearch_query]]` is still required for all of the examples below.
```

plugins/inputs/elasticsearch_query/aggregation_parser.go
```diff
@@ -0,0 +1,154 @@
+package elasticsearch_query
+
+import (
+	"fmt"
+
+	"github.com/influxdata/telegraf"
+	elastic "gopkg.in/olivere/elastic.v5"
```
srebhan, line 7: As per our discussion, please prepare for later versions by
```suggestion
	elastic5 "gopkg.in/olivere/elastic.v5"
```
or similar.

plugins/inputs/elasticsearch_query/aggregation_parser.go
```diff
@@ -0,0 +1,154 @@
...
+)
+
+type resultMetric struct {
+	name   string
+	fields map[string]interface{}
+	tags   map[string]string
+}
+
+func (e *ElasticsearchQuery) parseSimpleResult(measurement string, searchResult *elastic.SearchResult, acc telegraf.Accumulator) error {
+
```
srebhan, line 17: Please remove the newline.

plugins/inputs/elasticsearch_query/aggregation_parser.go
```diff
@@ -0,0 +1,154 @@
...
+	elastic "gopkg.in/olivere/elastic.v5"
+)
+
+type resultMetric struct {
+	name   string
+	fields map[string]interface{}
+	tags   map[string]string
+}
+
+func (e *ElasticsearchQuery) parseSimpleResult(measurement string, searchResult *elastic.SearchResult, acc telegraf.Accumulator) error {
```
srebhan, line 16: As this function cannot error please remove the return parameter to make this obvious to everybody.

srebhan: And can you please make `acc` the first parameter. It's not really a convention but nice to indicate "Hey we pass the acc down here.".

plugins/inputs/elasticsearch_query/aggregation_parser.go
```diff
@@ -0,0 +1,154 @@
...
+	acc.AddFields(measurement, fields, tags)
+	return nil
+}
+
+func (e *ElasticsearchQuery) parseAggregationResult(aggregationQueryList []aggregationQueryData, searchResult *elastic.SearchResult, acc telegraf.Accumulator) error {
+	var measurements = map[string]map[string]string{}
+
+	// organize the aggregation query data by measurement
+	for _, aggregationQuery := range aggregationQueryList {
+
```
srebhan, line 32: Remove newline. Please only use newlines to indicate new "logical code blocks" as this clutters the code otherwise IMO. :-)

plugins/inputs/elasticsearch_query/aggregation_parser.go
```diff
@@ -0,0 +1,154 @@
...
+	switch resp := resp.(type) {
+	case *elastic.AggregationBucketKeyItems:
+		// we've found a terms aggregation, iterate over the buckets and try to retrieve the inner aggregation values
+		for _, bucket := range resp.Buckets {
+			m.fields["doc_count"] = bucket.DocCount
+			if s, ok := bucket.Key.(string); ok {
+				m.tags[aggName] = s
+			} else {
+				return m, fmt.Errorf("bucket key is not a string")
```
srebhan, line 88: Maybe print the `bucket.Key` to help debugging the problem, but that's only cosmetics.

plugins/inputs/elasticsearch_query/aggregation_parser.go
```diff
@@ -0,0 +1,154 @@
...
+			if err != nil {
+				return m, err
+			}
+
+			// if there are fields present after finishing the bucket, it is a complete metric
+			// store it and clean the fields to start a new metric
+			if len(m.fields) > 0 {
+				acc.AddFields(m.name, m.fields, m.tags)
+				m.fields = make(map[string]interface{})
+			}
```
srebhan, line 102: I think you need a flag to denote that you are at the highest level again before adding the fields. Imagine you go down the tree, say 4 times, then you will add at least 3 times the same metric with a growing number of fields. Remember fields are non-zero in every level except maybe the lowest.
So either add a depth counter and only add the metric if you are at depth zero or split the recursion part from the top-level function...

plugins/inputs/elasticsearch_query/aggregation_parser.go
```diff
@@ -0,0 +1,154 @@
...
+
+	case *elastic.AggregationValueMetric:
+		if resp.Value != nil {
+			m.fields[aggName] = *resp.Value
+		} else {
+			m.fields[aggName] = float64(0)
+		}
+
+	default:
+		return m, fmt.Errorf("aggregation type returned not supported")
```
srebhan, line 116: Here you might also want to print the unknown aggregation type so we can start debugging from there.

plugins/inputs/elasticsearch_query/aggregation_query.go
srebhan: Huh? Where is this change?

plugins/inputs/elasticsearch_query/aggregation_query.go
```diff
@@ -0,0 +1,203 @@
...
+type aggregationQueryData struct {
+	aggKey
+	isParent    bool
+	aggregation elastic.Aggregation
+}
+
+func (e *ElasticsearchQuery) runAggregationQuery(ctx context.Context, aggregation esAggregation) (*elastic.SearchResult, error) {
+
+	now := time.Now().UTC()
+	from := now.Add(aggregation.QueryPeriod.Duration * -1)
```
srebhan, line 28: Still open...

plugins/inputs/elasticsearch_query/aggregation_query.go
```diff
@@ -0,0 +1,203 @@
...
+	filterQuery := aggregation.FilterQuery
+
```
srebhan, line 30: Remove newline.

plugins/inputs/elasticsearch_query/aggregation_query.go
```diff
@@ -0,0 +1,203 @@
...
+	search := e.esClient.Search().Index(aggregation.Index).Query(query).Size(0)
+
+	// add only parent elastic.Aggregations to the search request, all the rest are subaggregations of these
+	for _, v := range aggregation.aggregationQueryList {
+		if v.isParent && v.aggregation != nil {
+			search.Aggregation(v.aggKey.name, v.aggregation)
+		}
+	}
+	searchResult, err := search.Do(ctx)
+
```
srebhan, line 48: Remove newline.

plugins/inputs/elasticsearch_query/aggregation_query.go
```diff
@@ -0,0 +1,203 @@
...
+// GetMetricFields function returns a map of fields and field types on Elasticsearch that matches field.MetricFields
+func (e *ElasticsearchQuery) getMetricFields(ctx context.Context, aggregation esAggregation) (map[string]string, error) {
+	mapMetricFields := make(map[string]string)
+
+	if e.esClient == nil {
+		err := e.connectToES()
+		if err != nil {
+			return nil, err
+		}
+	}
```
srebhan, line 66: Do we really need this here? If so you should also care for this in `runAggregationQuery`.

plugins/inputs/elasticsearch_query/aggregation_query.go
```diff
@@ -0,0 +1,203 @@
...
+												}
+											}
+										}
+									}
+								}
+							}
+						}
+					}
+				}
+			}
```
srebhan, line 101: Please try to invert the logic to save indention levels! Furthermore, please check the type assertions, especially for `index` (first `if`).

plugins/inputs/elasticsearch_query/elasticsearch_query_test.go
```diff
@@ -0,0 +1,443 @@
...
+
+	err := e.connectToES()
+	require.NoError(t, err)
+
+	bulkRequest := e.esClient.Bulk()
+
+	// populate elasticsearch with nginx_logs test data file
+	file, err := os.Open("testdata/nginx_logs")
+	require.NoError(t, err)
+
```
srebhan, line 337: Remove newline.

srebhan (commented):
Almost. Please check the recursive function again as I think you add the metrics multiple times...

.vscode/launch.json
lpic10: ops, sorry, not sure how that ended up there :)

plugins/inputs/elasticsearch_query/aggregation_query.go
lpic10: I noticed I cannot just replace as suggested because `Sub` requires a `Time` as `Add` requires a `Duration`

plugins/inputs/elasticsearch_query/aggregation_query.go
lpic10: Yes, this does not look good... 
I tried reverting at first, but having to initialize each variable before assigning the value after the assertion check wasn't looking good either, unless I'm missing something.

plugins/inputs/elasticsearch_query/aggregation_parser.go
lpic10: The metric needs to be added after it reaches the highest level. 

Until it gets there, the tags are being added to the metric, and the fields too, except that the `doc_count` field is overwritten each time it goes up a level to another terms aggregation. Only when it reaches the `metric` aggregation (at the highest level of the recursion) is that it adds other field names like `field_avg` or `field_max`.

So I'm adding the metric when it goes down the first time. By resetting the `fields` and removing that bucket `tag` from the metric struct it means that it will not duplicate that metric when continuing to next bucket (new tag, new field) or going down the recursion. 

As the `metric` struct is returned each time from the higher recursion function down before checking if it needs to be added, normally it will not add again when going down the next time. 

By looking now I need to confirm if it is working correctly when more than one metric field is present.
It seems I will need to add the metric and reset fields at the end of the `AggregationValueMetric` switch case; and the current `addFields` would only be called when no metric field is present.

srebhan: Why don't you collect _all_ fields and _all_ tags during recursion and the call `acc.AddFields(m.name, m.fields, m.tags)` in the measurement loop of `parseAggregationResult()`? I think this will be much easier to understand... I'm still unsure if your description above matches what is done in the code after reading this function multiple times now. :-)

plugins/inputs/elasticsearch_query/aggregation_query.go
srebhan: Ok than at least make it
```suggestion
	from := now.Add(-aggregation.QueryPeriod.Duration)
```

plugins/inputs/elasticsearch_query/elasticsearch_query.go
```diff
@@ -0,0 +1,277 @@
...
+  ## Fields to be used as tags
+  ## Must be non-analyzed fields, aggregations are performed per tag
+  tags = ["field.keyword", "field2.keyword"]
+
+  ## Set to true to not ignore documents when the tag(s) above are missing
+  # include_missing_tag = false
+
+  ## String value of the tag when the tag does not exist
+  ## Used when include_missing_tag is true
+  # missing_tag_value = "null"
```
srebhan, line 81: Are all of the non-commented options mandatory and are different to their defaults? If not, please comment as many of them as possible to give the user a sense of what is the bare-minimum to configure.

plugins/inputs/elasticsearch_query/elasticsearch_query.go
```diff
@@ -0,0 +1,277 @@
...
+
+		aggregationQueryList, err := e.buildAggregationQuery(mapMetricFields, agg)
+		if err != nil {
+			return err
+		}
+		agg.mapMetricFields = mapMetricFields
+		agg.aggregationQueryList = aggregationQueryList
+		e.Aggregations[i] = agg
+	}
+	return err
```
srebhan, line 161: I guess this error can only be `nil`!? If so, please explicitly return `nil`.

srebhan (commented):
Some minor things remaining besides the big topic of recursion. :-) Can you please remove the commented plugins/lines from `plugins/inputs/elasticsearch_query/dev/telegraf.conf` in order to not add ~5,800 lines of rubbish. :-)

plugins/inputs/elasticsearch_query/aggregation_parser.go
lpic10: I could do that, so instead of calling `acc.AddFields` directly I could add the resulting metric to a map (or something else) and iterate over that in `parseAggregationResult`, but I think that would be more complex. 

Then I would need to see how to work with the tags slice inside the `metric` struct as I need to add/remove values from it during the recursive calls. As slices work by reference it would change the values in the resulting map as well (I figured that out because I tried that).

I want to add tests to this using JSON responses from these aggregations in Elasticsearch. Maybe by looking at these will help the understanding of this particular part of the code (and maybe to improve with a better solution).

lpic10:
> Some minor things remaining besides the big topic of recursion. :-) Can you please remove the commented plugins/lines from `plugins/inputs/elasticsearch_query/dev/telegraf.conf` in order to not add ~5,800 lines of rubbish. :-)

Well, that dev folder shouldn't be in the commit in the first place (I guess I added all files by mistake at some point). 
I'll remove that on next pull.

Thanks for the review again!

srebhan:
@lpic10 thank you for your effort! :+1:

plugins/inputs/elasticsearch_query/aggregation_query.go
srebhan: Your new solution is really nice!

plugins/inputs/elasticsearch_query/aggregation_parser.go
srebhan: To be honest it cannot be as hard to read as this. I _really_ tried to wrap my head around it and I'm still not sure I get all the possible paths, so splitting this into two parts might be a good idea. :-)
I still don't get the idea I think. As far as I get it you get a response that has a tree like structure. Is this correct? That's why you recurse into it and build aggregations on the way. However, you want to collect the whole subtree to form a metric, don't you?
Your code suggest that you call this recursions for _each_ **measurement**, so the first question is, do you expect multiple metrics for each measurement? If yes, do you have a clear signal of when a metric ends and another starts? Currently it looks like you want a metric for each `bucket` and each `metric`, but only at the highest level, all other things further down the tree should be collected/aggregated. Is this correct?

If all of the above is true, moving `acc.AddFields()` to `parseAggregationResult()` is a bad idea. In this case I would suggest to pass a `depth` counter down the recursive function and only call  `acc.AddFields()` if that counter is zero.

lpic10: > To be honest it cannot be as hard to read as this. I really tried to wrap my head around it and I'm still not sure I get all the possible paths, so splitting this into two parts might be a good idea. :-)

It was painful for me to go back to this code too :)

> I still don't get the idea I think. As far as I get it you get a response that has a tree like structure. Is this correct? 

Yes, exactly, and the objective is somehow to flat that tree.

> That's why you recurse into it and build aggregations on the way. However, you want to collect the whole subtree to form a metric, don't you?

Yes.

> Your code suggest that you call this recursions for _each_ **measurement**, so the first question is, do you expect multiple metrics for each measurement?

Yes, that's the case

> ... do you have a clear signal of when a metric ends and another starts? Currently it looks like you want a metric for each `bucket` and each `metric`, but only at the highest level, all other things further down the tree should be collected/aggregated. Is this correct?

Yes. The metric stops at the highest level as you mention. The highest level is a metric aggregation, when it exists. But a metric aggregation is optional, and the metric may end at the highest level for each bucket.

The metric aggregation (when it exists) is always at the edge of the (sub)tree. 

> If all of the above is true, moving `acc.AddFields()` to `parseAggregationResult()` is a bad idea. In this case I would suggest to pass a `depth` counter down the recursive function and only call `acc.AddFields()` if that counter is zero.

I will investigate this during the weekend if I can, but my feeling is that if the counter is zero it is too late to cal the `acc.AddFields()` as the tags are changing if there are more buckets. The part that I'm really in doubt is about calling it inside the metric aggregation though, I need to test the code using at least 3 terms aggregation and 2 metrics aggregation (the current test data only allowed me 1 metrics aggregation).

srebhan: Thanks for your effort but I think it's worth it... :-)

lpic10:
WIP

srebhan:
@lpic10 any estimated on when I'll be able to give it another round!?

plugins/inputs/elasticsearch_query/aggregation_query.go
```diff
@@ -0,0 +1,215 @@
...
+				function:    aggregation.MetricFunction,
+				field:       k,
+				name:        strings.Replace(k, ".", "_", -1) + "_" + aggregation.MetricFunction,
+			},
+			isParent:    true,
+			aggregation: agg,
+		}
+
+		aggregationQueryList = append(aggregationQueryList, aggregationQuery)
+
```
srebhan, line 154: Please change this to make the linter happy.

plugins/inputs/elasticsearch_query/elasticsearch_query.go
```diff
@@ -0,0 +1,304 @@
+package elasticsearch_query
+
+import (
+	"context"
+	"fmt"
+	"log"
```
srebhan, line 6: You should _not_ use `log` but rather define 
```golang
```
in your struct and the use the telegraf logging facility!

plugins/inputs/elasticsearch_query/elasticsearch_query.go
```diff
@@ -0,0 +1,304 @@
...
+  ## Fields to be used as tags
+  ## Must be text, non-analyzed fields. Metric aggregations are performed per tag
+  # tags = ["field.keyword", "field2.keyword"]
+
+  ## Set to true to not ignore documents when the tag(s) above are missing
+  # include_missing_tag = false
+
+  ## String value of the tag when the tag does not exist
+  ## Used when include_missing_tag is true
+  # missing_tag_value = "null"
```
srebhan, line 83: Please increase the indention of this subtable with two spaces.

plugins/inputs/elasticsearch_query/elasticsearch_query.go
```diff
@@ -0,0 +1,304 @@
...
+	}
+	return nil
+}
+
+func (e *ElasticsearchQuery) initAggregation(ctx context.Context, agg esAggregation, i int) {
+	// retrieve field mapping and build queries only once
+	mapMetricFields, err := e.getMetricFields(ctx, agg)
+	if err != nil {
+		log.Printf("E! %s", err.Error())
+		return
```
srebhan, line 158: Wouldn't it make sense to return this error and handle it in `Init()`? Can you do some sensible processing if this fails?

plugins/inputs/elasticsearch_query/elasticsearch_query_test.go
```diff
@@ -0,0 +1,682 @@
...
+	wantBuildQueryErr         bool
+	wantGetMetricFieldsErr    bool
+	wantQueryResErr           bool
+}
+
+type expectedMetric struct {
+	measurement string
+	fields      map[string]interface{}
+	tags        map[string]string
+}
```
srebhan, line 46: You might want to use `testutil.MustMetric(measurement, fields, tags, time.Unix(0,0)` instead. This way you can later directly test the metrics using `testutil.RequireMetricsEqual(t, expected, acc.GetTelegrafMetrics(), testutil.IgnoreTime())` where expected is `expected := []telegraf.Metric{ testutil.MustMetric(...), testutil.MustMetric(...),...}`

srebhan (changes requested):
There are still some points you need to address. The most important one is to _not_ use `log` but the telegraf logging factility. See my comments in the code.

srebhan:
@lpic10 any chance you work on this?

plugins/inputs/elasticsearch_query/elasticsearch_query.go
lpic10: The idea here is just to wait for the next plugin execution as the data for the query may not be available (yet) on ES, and the plugin init should not fail because of that.

plugins/inputs/elasticsearch_query/elasticsearch_query_test.go
srebhan: Can you please consider this comment.

plugins/inputs/elasticsearch_query/elasticsearch_query_test.go
```diff
@@ -0,0 +1,684 @@
...
+		Response     string    `json:"response"`
+		Size         float64   `json:"size"`
+		ResponseTime float64   `json:"response_time"`
+	}
+
+	err := e.connectToES()
+	if err != nil {
+		e.Log.Errorf("error setting up integration test: %s", err.Error())
+		return
+	}
```
srebhan, line 490: Please either simply return the error here and check it in the caller or pass the `testing.T` in here and use `require.NoError()` to fail hard. I prefer the former but I'm also OK with the latter.

srebhan (commented):
I have some comments for the tests. The other code looks good.

lpic10:
I've changed the tests as requested, let me know if anything else

plugins/inputs/elasticsearch_query/elasticsearch_query.go
srebhan: Fair enough.

srebhan:
@lpic10 thanks for keep this going! I think the only thing left is the following comment in the tests:

> You might want to use testutil.MustMetric(measurement, fields, tags, time.Unix(0,0) instead. This way you can later directly test the metrics using testutil.RequireMetricsEqual(t, expected, acc.GetTelegrafMetrics(), testutil.IgnoreTime()) where expected is expected := []telegraf.Metric{ testutil.MustMetric(...), testutil.MustMetric(...),...}

 It would be nice if you can care for this one. Everything else looks nice from my point of view.

plugins/inputs/elasticsearch_query/elasticsearch_query.go
```diff
@@ -0,0 +1,310 @@
...
+	esVersionSplit := strings.Split(esVersion, ".")
+
+	// quit if ES version is not supported
+	if len(esVersionSplit) == 0 {
+		return fmt.Errorf("elasticsearch version check failed")
+	}
+
+	i, err := strconv.Atoi(esVersionSplit[0])
+	if err != nil || i < 5 {
+		return fmt.Errorf("elasticsearch version %s not supported (currently supported versions are 5.x and 6.x)", esVersion)
```
akrantz01, line 232: Should this error on 7.x as well?

akrantz01 (commented):
Thanks for keeping this PR alive! Overall it looks good; I just left a couple of comments.

plugins/inputs/elasticsearch_query/elasticsearch_query.go
```diff
@@ -0,0 +1,310 @@
...
+func (e *ElasticsearchQuery) Init() error {
+	if e.URLs == nil {
+		return fmt.Errorf("elasticsearch urls is not defined")
+	}
+
+	err := e.connectToES()
+	if err != nil {
+		e.Log.Errorf("E! error connecting to elasticsearch: %s", err)
+		return nil
+	}
```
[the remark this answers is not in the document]
lpic10: That's the reason I don´t return the error here :)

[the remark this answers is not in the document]
lpic10: I don´t know about the Start(), I will see if it makes sense to move some logic there

plugins/inputs/elasticsearch_query/elasticsearch_query.go
lpic10: Yes, definitely, current code does not work on 7.x versions

plugins/inputs/elasticsearch_query/elasticsearch_query.go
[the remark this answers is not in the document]
akrantz01: Ok, that's what I figured. I think it's fine in the current state since part of the configuration validation depends on a connection to the ES instance.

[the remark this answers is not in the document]
lpic10: From what I've checked, Start() applies to service input plugins only. Is that right?

plugins/inputs/elasticsearch_query/elasticsearch_query_test.go
lpic10: Done :)

srebhan (approved):
Thanks @lpic10 for your epic effort in making this fly! Looks good to me.

sspaink (approved):
Looks great! Awesome work.

PR #3546, opened 2017-12-06
add generic http input plugin which supports setting an input data format

grange74:
Issue #813.

### Required for all PRs:

### DIFF
--- a/plugins/inputs/http/http.go
+++ b/plugins/inputs/http/http.go
@@ -1,0 +1,165 @@
+package http
+
+import (
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"sync"
+	"time"
+
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/internal"
+	"github.com/influxdata/telegraf/plugins/inputs"
+	"github.com/influxdata/telegraf/plugins/parsers"
+)
+
+type HTTP struct {
+	URLs []string `toml:"urls"`
+
+	// HTTP Basic Auth Credentials
+	Username string
+	Password string
+
+	// Path to CA file
+	SSLCA string `toml:"ssl_ca"`
+	// Path to host cert file
+	SSLCert string `toml:"ssl_cert"`
+	// Path to cert key file
+	SSLKey string `toml:"ssl_key"`
+	// Use SSL but skip chain & host verification
+	InsecureSkipVerify bool
+
+	Timeout internal.Duration
+
+	client *http.Client
+
+	// The parser will automatically be set by Telegraf core code because
+	// this plugin implements the ParserInput interface (i.e. the SetParser method)
+	parser parsers.Parser
+}
+
+var sampleConfig = `
+  ## One or more URLs from which to read formatted metrics
+  urls = [
+    "http://localhost:2015/simple.json"
+  ]
+
+  ## Optional HTTP Basic Auth Credentials
+  # username = "username"
+  # password = "pa$$word"
+
+  ## Optional SSL Config
+  # ssl_ca = "/etc/telegraf/ca.pem"
+  # ssl_cert = "/etc/telegraf/cert.pem"
+  # ssl_key = "/etc/telegraf/key.pem"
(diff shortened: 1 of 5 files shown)


plugins/inputs/http/http.go
```diff
@@ -0,0 +1,165 @@
...
+  ## Use SSL but skip chain & host verification
+  # insecure_skip_verify = false
+
+  ## http request & header timeout
+  ## defaults to 5s if not set
+  timeout = "10s"
+
+  ## Mandatory data_format
+  ## See available options at https://github.com/influxdata/telegraf/blob/master/docs/DATA_FORMATS_INPUT.md
+  data_format = "json"
```
danielnelson: This is actually optional, if unset it will default to `"influx"`.  Just use the default snippet here:
```
  ## Data format to consume.
  ## more about them here:
  # data_format = "influx"
```

plugins/inputs/http/http.go
```diff
@@ -0,0 +1,165 @@
...
+
+	// The parser will automatically be set by Telegraf core code because
+	// this plugin implements the ParserInput interface (i.e. the SetParser method)
+	parser parsers.Parser
+}
+
+var sampleConfig = `
+  ## One or more URLs from which to read formatted metrics
+  urls = [
+    "http://localhost:2015/simple.json"
```
danielnelson: Because of my comment about `data_format` we should set this to something more format neutral such as `http://localhost/metrics`

plugins/inputs/http/http.go
```diff
@@ -0,0 +1,165 @@
...
+func (h *HTTP) Gather(acc telegraf.Accumulator) error {
+	if h.client == nil {
+		tlsCfg, err := internal.GetTLSConfig(
+			h.SSLCert, h.SSLKey, h.SSLCA, h.InsecureSkipVerify)
+		if err != nil {
+			return err
+		}
+		h.client = &http.Client{
+			Transport: &http.Transport{
+				ResponseHeaderTimeout: h.Timeout.Duration,
```
danielnelson: Don't set a response header timeout since we have the client timeout.  I need to exercise these from the code so they stop getting copied around.

plugins/inputs/http/http.go
```diff
@@ -0,0 +1,165 @@
...
+  ## Optional SSL Config
+  # ssl_ca = "/etc/telegraf/ca.pem"
+  # ssl_cert = "/etc/telegraf/cert.pem"
+  # ssl_key = "/etc/telegraf/key.pem"
+  ## Use SSL but skip chain & host verification
+  # insecure_skip_verify = false
+
+  ## http request & header timeout
+  ## defaults to 5s if not set
+  timeout = "10s"
```
danielnelson: Update the comment based on removing header timeout (see below).  Use the default value and comment it out:
```
# timeout = "5s"
```

danielnelson:
I may have mentioned that I'm hoping to deprecate the `httpjson` plugin.  I looked through its feature set to see what we might still need.  (None of these are required to have this merged)

The `httpjson` plugin has support for setting general headers, and also using the `POST` or `GET` method.  I think these are important options we will need.

It also adds a tag `server` to every metric with the actual url.  I think this is also important but it seems that you should be able to configure if you want this tag added, since it could conflict with data provided at the url.  Perhaps we should have an option `tag_url = false`, and instead of calling this tag `server` we should call it `url`?

A float field is added to store the `response_time`.  This one I don't think we should add, it doesn't fit the purpose of the plugin in my opinion.  If there is strong demand for this then perhaps this could be a selfstat aka internal metric.

grange74:
I have updated based on your 4 comments. 
In the next few days, i will look to add support for setting generic headers, POST method and optional tagging with url.

danielnelson:
I added a few things after merging, notably method support and I decided to change the url tagging to method to automatically tag with url if it is not set in the remote host.  I decided to make this change because I felt like it was closer to the way the measurement name is set.

grange74:
@danielnelson Thanks for merging the PR and explaining your changes. The changes make perfect sense.

PR #3550, opened 2017-12-06
Use device name instead of abs path for devices tag

### LINKED ISSUE #3541: Remove "/dev/" prefix from smart device name tag
## Feature Request

### Proposal:
The new `smart` input plugin should not prefix device names with `/dev`. This is so that the device name is consistent with other measurements, such as the `diskio` measurement which outputs device names without the `/dev/` prefix.

### Current behavior:
A device such as `/dev/sda` is written as `device=/dev/sda`.

### Desired behavior:
A device such as `/dev/sda` is written as `device=sda`.

The consistency is important so that users who wish to use features such as Grafana's template variables can do so. If a user wishes to utilize both the `smart` and `diskio` measurements on a single dashboard which utilizes template variables to select the disk, it will be difficult because of the inconsistent naming.

danielnelson:
To be consistent with diskio and /proc entries

cc @rickard-von-essen

closes #3541

### Required for all PRs:

### DIFF
--- a/plugins/inputs/smart/README.md
+++ b/plugins/inputs/smart/README.md
@@ -129,7 +129,7 @@ the configuration to execute that.
 
 Example output from an _Apple SSD_:
 ```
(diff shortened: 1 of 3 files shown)


rickard-von-essen:
Looks good to me.

PR #3551, opened 2017-12-07
Add health status mapping from string to int.

arkady-emelyanov:
### Required for all PRs:

Add status_code field mapping when reporting health status.


danielnelson:
Thanks for the pull request, I think what may be simpler is if instead of the map we have a function with a switch over each health value.

arkady-emelyanov:
@danielnelson thanks for a quick review. you mean something like that?

danielnelson:
Perfect, I'll merge when the tests finish.

Issue #1896, opened 2016-10-13
MySQL Global Status values that should be strings aren't


skord:
## Feature Request

Many of the SHOW GLOBAL STATUS values from the MySQL plugin should be interpreted as strings, but aren't. A quick list of the keys from a MariaDB 10.1 instance:
- Compression
- Innodb_buffer_pool_dump_status
- Innodb_buffer_pool_load_status
- Innodb_have_atomic_builtins
- Innodb_have_lz4
- Innodb_have_lzo
- Innodb_have_lzma
- Innodb_have_bzip2
- Innodb_have_snappy
- Rpl_status
- Ssl_session_cache_mode
- wsrep_cluster_state_uuid
- wsrep_cluster_status
- wsrep_connected
- wsrep_evs_repl_latency
- wsrep_evs_state
- wsrep_gcomm_uuid
- wsrep_incoming_addresses
- wsrep_local_state_comment
- wsrep_local_state_uuid
- wsrep_provider_name
- wsrep_provider_vendor
- wsrep_provider_version
- wsrep_ready
### Proposal:

Ship the values as strings.
### Use case:

A few use cases. 
- Running a quick query against the data store to figure out if the replication cluster is in sync and which members are out of sync. 
- Being able to pinpoint when a configuration change led to a performance change.

h44z:
+1 would be usefull to have this values

sungiri88:
> select wsrep_incoming_addresses from mysql where time > now() - 9h order by time desc limit 10;
name: mysql
time                wsrep_incoming_addresses
----                ------------------------
1507875280000000000 0
1507875280000000000 0
1507875280000000000 0
1507875276000000000 0
1507875271000000000 0
1507875270000000000 0
1507875270000000000 0
1507875270000000000 0
1507875266000000000 0
1507875261000000000 0


Where as in cluter we get 

| wsrep_incoming_addresses                                     | 192.168.1.194:3306,192.168.1.197:3306,192.168.1.198:3306 |

DataType needs to be a string instead integer.

gabtastic:
+1

I have found that monitoring `wsrep_ready` is an efficient way to monitor node statuses in a Galera cluster.

Unfortunately, telegraf always report the value as `0`...

adrianlzt:
The problem is this line hardcoding all to integer
https://github.com/influxdata/telegraf/blame/7442b5645f3d8517f28ff4bf923e28ac40f0c093/plugins/inputs/mysql/mysql.go#L860
[...]

PR #3554, opened 2017-12-07
mysql fixes for #1896

skord:
### Required for all PRs:

This PR includes a number of fixes for MySQL's data types. As pointed out in the comments of #1896, many global variables and global statuses had incorrect reporting or were missing. Also many items that would be helpful to check against that were strings would never be reported at all. 

The Global Statuses were all being forced into an Integer type and this led to things like the wsrep_ready status always reporting a zero. 

Other fields that were being parsed by the parseValue function were being missed or were incorrectly typed as well. Some were being missed because the output of SHOW GLOBAL STATUS reports an uppercase "YES" or "NO" rather than "Yes" or "No". 

parseValue is used in a few places. The way it used to work it would try and check for MySQL booleans, set 0 or 1 accordingly, then attempt to convert to a float. There are very few floats actually in use (DOUBLE's) in MySQL, most numbers are integers. This PR adds a few more simple functions that check if a value is a MySQL integer, float, boolean, or string and returns the correct value.

In addition, the global status function was changed to remove the mappings structure. This is due to the fact that MySQL plugins can add new statuses and they could be missed if not in the mapping or assigned in the global status function. The mapping only really served to rename the "com_" prefix to "command" and lowercase key names. I just stuck with toLower on the key names as it seems like it makes more sense to me that the MySQL variable names line up with what you'd see if you ran the query on the server. 

As a warning, if this is introduced into an environment that is currently reporting MySQL metrics to an InfluxDB server, if there is a type change on the key, the values will not be reported. Seeing as though previous metrics were sometimes completely wrong, this could be an acceptable thing. 

Seeing as though I'm not a go wizard, I'd appreciate any review.

### DIFF
--- a/plugins/inputs/mysql/mysql.go
+++ b/plugins/inputs/mysql/mysql.go
@@ -541,7 +541,9 @@ func (m *Mysql) gatherGlobalVariables(db *sql.DB, serv string, acc telegraf.Accu
 			fields[key] = string(val)
 			tags[key] = string(val)
 		}
-		fields[key] = parseValue(val)
+		if value, ok := parseValue(val); ok {
+			fields[key] = value
+		}
 		// Send 20 fields at a time
 		if len(fields) >= 20 {
 			acc.AddFields("mysql_variables", fields, tags)
@@ -591,7 +593,9 @@ func (m *Mysql) gatherSlaveStatuses(db *sql.DB, serv string, acc telegraf.Accumu
 		// range over columns, and try to parse values
 		for i, col := range cols {
 			col = strings.ToLower(col)
-			fields["slave_"+col] = parseValue(*vals[i].(*sql.RawBytes))
+			if value, ok := parseValue(*vals[i].(*sql.RawBytes)); ok {
+				fields["slave_"+col] = value
+			}
 		}
 		acc.AddFields("mysql", fields, tags)
 	}
@@ -660,7 +664,9 @@ func (m *Mysql) gatherGlobalStatuses(db *sql.DB, serv string, acc telegraf.Accum
 
 		key = strings.ToLower(key)
 
-		fields[key] = parseValue(val)
+		if value, ok := parseValue(val); ok {
+			fields[key] = value
+		}
 
 		// Send 20 fields at a time
 		if len(fields) >= 20 {
@@ -1042,20 +1048,20 @@ func (m *Mysql) gatherInnoDBMetrics(db *sql.DB, serv string, acc telegraf.Accumu
 	}
 	defer rows.Close()
 
-	var key string
-	var val sql.RawBytes
-
 	// parse DSN and save server tag
 	servtag := getDSNTag(serv)
 	tags := map[string]string{"server": servtag}
 	fields := make(map[string]interface{})
 	for rows.Next() {
+		var key string
+		var val sql.RawBytes
 		if err := rows.Scan(&key, &val); err != nil {
 			return err
 		}
 		key = strings.ToLower(key)
-		fields[key] = parseValue(val)
-
+		if value, ok := parseValue(val); ok {
+			fields[key] = value
+		}
 		// Send 20 fields at a time
 		if len(fields) >= 20 {
 			acc.AddFields("mysql_innodb", fields, tags)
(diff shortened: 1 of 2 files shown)


skord:
I'm closing this because I'm seeing errors on my local copy between int and float evaluation. Will reopen when it's fixed.

skord:
This was user error, I didn't clear out local integration testing instance first.

plugins/inputs/mysql/mysql.go
```diff
@@ -820,98 +645,36 @@ func (m *Mysql) gatherBinaryLogs(db *sql.DB, serv string, acc telegraf.Accumulat
 // the mappings of actual names and names of each status to be exported
 // to output is provided on mappings variable
 func (m *Mysql) gatherGlobalStatuses(db *sql.DB, serv string, acc telegraf.Accumulator) error {
-	// If user forgot the '/', add it
-	if strings.HasSuffix(serv, ")") {
-		serv = serv + "/"
-	} else if serv == "localhost" {
-		serv = ""
-	}
-
```
danielnelson: Why is this removed?

plugins/inputs/mysql/mysql.go
```diff
@@ -820,98 +645,36 @@ func (m *Mysql) gatherBinaryLogs(db *sql.DB, serv string, acc telegraf.Accumulat
...
 	// run query
 	rows, err := db.Query(globalStatusQuery)
 	if err != nil {
 		return err
 	}
+	defer rows.Close()
+
+	var key string
+	var val sql.RawBytes
```
danielnelson: Move these declarations back into the loop, less error likely to cause errors if this code is modified in the future.

plugins/inputs/mysql/mysql.go
skord: The user can't forget the slash at the end, the driver will bubble up an error saying they forgot the slash at the end and the plugin will fail. It was inconsistent with the rest of the plugin and the default behavior instructs the user on how to fix it rather than continuing incorrectly.

skord:
Since this is already a feature breaking commit, was curious if prefixing the info_schema with mysql_ was a good idea to throw in this one. As it stands, all the namespaces for this plugin, and even some that are info schema related are prefixed with mysql_, except for the ones in the attached screenshot. 

selection_094

danielnelson:
What do you think about putting these into the existing mysql_info_schema like:
```
mysql_info_schema,<existing-tags> rows=42i data_free=42i data_length=42i index_length=42i
```

skord:
I'm playing with the addition to the mysql_info_schema, there's some strange behavior with it because the tags don't 100% line up, but it does seem like less noise. 

After checking with some people smarter than me, I'd made the assumption that MySQL will always return some sort of numeric value for numeric fields, but that assumption is incorrect. slave_seconds_behind master will return a blank value if it's not applicable, so the default behavior of defaulting to a string ensures once replication is setup, that the field in influx is a string and it gets dropped. I'll need to fix this behavior before this can be merged.

skord:
In the last two commits, the measurements were lined up to what their feature flags are regarding the results from the info schema. The existing mysql_info_schema was the auto-inc and processlist queries, which didn't make sense being lumped together with table schema as the tags don't line up at all it's vastly different data. 

The auto-inc and table schema queries do have their tags line up and they've been moved to mysql_table_schema. The data regarding the process list has been moved to the mysql_process_list measurement. 

I don't really have any other ambition with this PR, here's the run-down

- [x] Fixed case-sensitive parsing of MySQL boolean type configuration values. These are still output as integers rather than converting them to a boolean type.
- [x] Fixed incorrect conversion of global statuses that led to incorrect values being shipped.
- [x] Eliminated the mappings struct for reasons stated above.
- [x] parseValue now returns values that match MySQL's data types, including strings.
- [x] Data from queries about the table schema now lives in the **mysql_table_schema** measurement along with the auto increment query, as they describe similar values. Table schema was chosen as a name since it lines up with the configuration option.
- [x] Data from the process list query has been moved out of **mysql_info_schema** and put into **mysql_process_list**, lining it up with it's configuration value.
- [x] Table version data that also doesn't have tags that match other measurements has been moved to **mysql_table_schema_version**

danielnelson:
Looks good, is `mysql` the only measurement name that will potentially have type conflicts?

skord:
@danielnelson I did find one other problem in parse value, I expect to have that fixed in a few hours. It needs error checking and not to just return nil. Returning nil "fixes the glitch" but makes a lot of noise when shipping metrics. 

```mysql``` may have conflicts as well as ```mysql_variables``` and ```mysql_innodb``` may have conflicts. In the majority of cases, the existing data is invalid or incorrect, as the conversion was returning 0's in a lot of cases where it shouldn't.

skord:
@danielnelson I've checked for the noise in my kapacitor/influx logs for receiving null values and they've disappeared. Everything looks good from my end. Thanks so much.

danielnelson:
Can you take a look at the comment I added to the release notes in https://github.com/influxdata/telegraf/commit/de180d1e56dace1ddc557491ebe7ac1d7a13542a?

skord:
@danielnelson looks good.

PR #3559, opened 2017-12-08
procstats patterns without pgrep

### LINKED ISSUE #3558: Support for using procstat without pgrep
### Proposal:
I would like to submit a PR to change proc stats to use [go-ps](https://github.com/keybase/go-ps) instead of pgrep. This would allow pattern matching to work in a cross platform way. I am willing to do the work for this but I wanted to make sure that the change would be accepted.

### Current behavior:
proc stats can not do pattern matching on windows
### Desired behavior:
proc stats can do pattern matching on windows
This will allow me to leverage telegraf instead of rolling my own solutions to monitor java processes in windows.

vrecan:
I have pulled in a new dep for dealing with cross platform pattern matching for processes. I have validated everything on linux and I am going through some steps to validate it all on windows. Let me know if I did something incorrectly when I added the dependencies.

This fixes #3558 

This is my first PR for telegraf. I have signed the CLA. 

### Required for all PRs:

- [] Associated README.md updated.


vrecan:
this is failing because I am waiting for the change to land in gopsutil shirou/gopsutil/pull/468

vrecan:
I have reverted my change to gopsutil and just implemented the WMI queries here to make it work. WMI is the only way that I know to do this without using CGO. Let me know if there is another way you would like me to implement this or if there is anything you would like me to change!

Thanks!

danielnelson:
I think we shoud create another `PIDFinder`, we can call it `NativeFinder`.  We can implement the interface using regular gopsutil commands but for name/cmdline we can have it use the specialized WMI calls on Windows.  Then in the procstat config we can introduce an option: `finder = "pgrep"` or `finder = "native"`.  Then we can move the regex matching into the `NativeFinder`.

There are a few spots in the code where there are small unneeded changes, can you also clean these up so that we only modify lines related to the patchset.

vrecan:
OK so you want me to create a new finder that fully implements everything for all platforms using gopsutil only? And then in that version it will use the special WMI calls for windows? 

Just trying to make sure I understand what you want me to do before I implement it. I am open to implementing it however you want :)

vrecan:
There are a few things that are different for windows then unix. Windows has no pid files and it requires another WMI call to get processes by owner, it's a name though and not a UID.

danielnelson:
For the pid files, even though no one really uses them in Windows, maybe we can just use the same implementation.  Then someone could use it if they wanted for some reason and we don't have to explain that this option doesn't work.

danielnelson:
A little more clarification, a user should be able to use either the `native` or `pgrep` finder on any plaform.  We should have the default implementation use gopsutil, but we can have a windows specialization that uses the faster wmi calls as needed.

vrecan:
sounds good, thanks I will start working on that

vrecan:
@danielnelson do you mind looking at my current implementation. Does this look like what you want? If so I am going to add some more tests and update the readme

plugins/inputs/procstat/nativeFinder_unix.go
```diff
@@ -0,0 +1,59 @@
+// +build !windows
```
danielnelson: Call this file `native_finder_notwindows.go`

plugins/inputs/procstat/nativeFinder_windows.go
```diff
@@ -0,0 +1,109 @@
+// +build windows
```
danielnelson: Call this file `native_finder_windows.go`

plugins/inputs/procstat/nativeFinder_windows.go
```diff
@@ -0,0 +1,109 @@
...
+
+	"github.com/StackExchange/wmi"
+	"github.com/shirou/gopsutil/process"
+)
+
+//Timeout is the timeout used when making wmi calls
+var Timeout = 5 * time.Second
+
+func init() {
+	wmi.DefaultClient.AllowMissingFields = true
```
danielnelson: We should try to arrange for a new wmi.Client to be created and used once for each plugin.  This will avoid needing to modify the default client which might have unwanted effects on other pieces of code due to it being global.

plugins/inputs/procstat/nativeFinder_windows.go
```diff
@@ -0,0 +1,109 @@
...
+// 		return pids, err
+// 	}
+// 	for _, p := range procs {
+// 		pids = append(pids, PID(p.ProcessID))
+// 	}
+// 	return pids, nil
+// }
+
+//Pattern matches on the process name
+func (pg *NativeFinder) Pattern(pattern string) ([]PID, error) {
```
danielnelson: I think this looks the same as the notwindows version, lets put shared implementation in a `native_finder.go` file.

plugins/inputs/procstat/nativeFinder_windows.go
```diff
@@ -0,0 +1,109 @@
...
+		pids = append(pids, PID(p.ProcessID))
+	}
+	return pids, nil
+}
+
+//GetWin32ProcsByVariable allows you to query any variable with a like query
+func getWin32ProcsByVariable(variable string, qType queryType, value string, timeout time.Duration) ([]process.Win32_Process, error) {
+	var dst []process.Win32_Process
+	var query string
+	query = fmt.Sprint("WHERE ", variable, " ", qType, " \"", value, "\"")
```
danielnelson: Use fmt.Sprintf for readability.  I'm nervous about escaping here, `variable` and `value` need to be escaped more rigorously.

plugins/inputs/procstat/procstat.go
```diff
@@ -32,13 +33,19 @@ type Procstat struct {
...
 	procs           map[PID]Process
 	createProcess   func(PID) (Process, error)
 }
 
 var sampleConfig = `
+  ## pidFinder can be pgrep or native
+  ## pgrep tries to exec pgrep
+  ## native will work on all platforms, unix systems will use regexp. 
+  ## Windows will use WMI calls with like queries
+  pid_finder = native
```
danielnelson: Needs quotes: `pid_finder = "native"`

plugins/inputs/procstat/procstat.go
```diff
@@ -74,7 +81,17 @@ func (_ *Procstat) Description() string {
 
 func (p *Procstat) Gather(acc telegraf.Accumulator) error {
 	if p.createPIDFinder == nil {
-		p.createPIDFinder = defaultPIDFinder
+		fmt.Println("PIDFINDER: ", p.PidFinder)
+		switch p.PidFinder {
+		case "native":
+			fmt.Println("pid finder is now native")
```
danielnelson: Don't forget to cleanup debugging statements

plugins/inputs/procstat/procstat_test.go
```diff
@@ -1,3 +1,5 @@
+// +build !windows
```
danielnelson: Need to remove this build flag

plugins/inputs/procstat/README.md
```diff
@@ -27,8 +27,28 @@ Additionally the plugin will tag processes by their PID (pid_tag = true in the c
 * pid
 * process_name
 
+
+### Windows
```
danielnelson: We will have to remember to take a second pass at this documentation, so that it matches the current design

vrecan:
thanks for the comments, I am going to take a look at this today or tomorrow and update the PR

plugins/inputs/procstat/nativeFinder_windows.go
vrecan: should I have a seperate native finder struct for windows so that I can store it there?

plugins/inputs/procstat/nativeFinder_windows.go
vrecan: Agreed

plugins/inputs/procstat/procstat_test.go
vrecan: when I remove the build tag I get this error:
```
--- FAIL: TestGather_cgroupPIDs (0.00s)
        Error Trace:    procstat_test.go:363
        Error:          Received unexpected error:
                        open \sys\fs\cgroup\C:\Users\Vrecan\AppData\Local\Temp\565977507\cgroup.procs: The filename, directory name, or volume label syntax is incorrect.
```

plugins/inputs/procstat/nativeFinder_windows.go
vrecan: I was able to remove this and my queries still work. I took this from the gopsutil implementation and it looks like it's not really needed for our queries.

plugins/inputs/procstat/procstat_test.go
vrecan: I addressed this by skipping the test in windows

plugins/inputs/procstat/procstat.go
vrecan: done

plugins/inputs/procstat/procstat.go
vrecan: done

plugins/inputs/procstat/nativeFinder_windows.go
vrecan: I changed to sprintf, how would you like me to protect against escaping? based on [this](https://codereview.stackexchange.com/questions/160021/escaping-characters-in-a-wmi-query) I should be able to just escape backslash and single quote for the value. On the variable I can just fail anytime any of the non standard characters are in it. What do you think?

plugins/inputs/procstat/nativeFinder_windows.go
danielnelson: :+1:

plugins/inputs/procstat/nativeFinder_windows.go
danielnelson: In this case you can probably use the `%q` pattern, it will escape double quote and backslash using backslash, and it will add the outer quotation marks:
```
"WHERE %s %s %q"
```

plugins/inputs/procstat/procstat_test.go
danielnelson: Not sure if it will be enough, but we should use filepath.Join here: https://github.com/influxdata/telegraf/blob/master/plugins/inputs/procstat/procstat.go#L336

danielnelson:
Will need to rebase to fix the test error in cratedb

vrecan:
I pulled in master and I am still getting test failures. This time in leofs.

plugins/inputs/procstat/procstat_test.go
```diff
@@ -349,6 +350,10 @@ func TestGather_systemdUnitPIDs(t *testing.T) {
 }
 
 func TestGather_cgroupPIDs(t *testing.T) {
+	//no cgroups in windows
+	if runtime.GOOS == "windows" {
+		t.SkipNow()
```
danielnelson: You can put the comment in here like: `.Skip("no cgroups in windows")`

plugins/inputs/procstat/nativeFinder_windows.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+	return pids, nil
+}
+
+//GetWin32ProcsByVariable allows you to query any variable with a like query
+func getWin32ProcsByVariable(variable string, qType queryType, value string, timeout time.Duration) ([]process.Win32_Process, error) {
+	var dst []process.Win32_Process
+	var query string
+	// should look like "WHERE CommandLine LIKE "procstat"
+	query = fmt.Sprintf("WHERE %s %s '%s'", variable, qType, value)
```
danielnelson: Use %q here to get quoting with escaping of value. `"WHERE %s %s %q"`

plugins/inputs/procstat/nativeFinder.go
```diff
@@ -0,0 +1,66 @@
...
+	if err != nil {
+		return pids, err
+	}
+	pids = append(pids, PID(pid))
+	return pids, nil
+
+}
+
+//getPidsByUser ...
+func getPidsByUser(username string) ([]PID, error) {
```
danielnelson: Just merge this into `func (pg *NativeFinder) Uid(user string) ([]PID, error)` since it doesn't add any new abstraction.

vrecan:
@danielnelson I believe I have addressed all of your comments.

danielnelson:
@vrecan Merged in for 1.6, thanks.  I also did some updates to the procstat README so if you have a chance could you take a look and make sure it is correct?

vrecan:
Thanks for merging this, I will look at the docs now.

Issue #3394, opened 2017-10-26
Building telegraf manipulates other repos on $GOPATH


brianm:
## Bug report

Building telegraf via the included `Makefile` uses [gdm](https://github.com/sparrc/gdm) to manage dependencies. Unfortunately, `gdm` behaves poorly in regards to the common practice of using a singular `GOPATH`. A well behaved build will not manipulate things outside of the project being worked on.

### Relevant telegraf.conf:

n/a

### System info:

n/a

### Steps to reproduce:

1. `make`

### Expected behavior:

Elements outside the checked `telegraf/...` directory are not changed by the build.

### Actual behavior:

Many things outside the checkoued `telegraf/...` directory are manipulated.

### Additional info:

Go has introduced the `vendor/` directory behavior and dependency metadata (Gopkg.toml) to olve this exact problem, and the [dep](https://github.com/golang/dep) tool to manage these. 

### Proposal:

Switch to using `dep` to manage dependencies.

danielnelson:
We may do this in the future, but not until dep is a little more mature.  In the meantime you can use the `make telegraf` target to skip running gdm.

goller:
Hi @brianm I made a PR to convert to dep here:  https://github.com/influxdata/telegraf/pull/3564

Would you have time to look at it and give feedback?

james-lawrence:
as an aside, those of us who run gnome on linux get this wonderful error:
```
make
make deps
make[1]: Entering directory '/shared/james-lawrence/development/telegraf/src/github.com/influxdata/telegraf'
go get -u github.com/golang/lint/golint
go get github.com/sparrc/gdm
gdm restore
Only the root user can run GDM
```

PR #3564, opened 2017-12-11
Update vendoring to dep from gdm

goller:

Connect #3394

This PR changes Telegraf's vendoring from `gdm` to `dep`  Dep is getting closer and closer to be ready to go (see https://github.com/golang/dep#current-status) 

To generate the dep files (Gopkg.lock and .toml) I used deps automatic conversion from the `godep` vendoring solution.  

Gdm's Godep file is similar to godeps.  Here is a python program I wrote to convert gdm's Godep to godep's format:

```python
#!/usr/bin/env python
import json

with open('Godeps') as godeps:
    conversion = {
        "ImportPath": "github.com/tools/godep",
	"GoVersion": "go1.9",
	"GodepVersion": "v74",
	"Deps": []
    }
    deps = godeps.readlines()
    for dep in deps:
        package, sha = dep.split()
        conversion["Deps"].append({
            "ImportPath": package,
            "Rev": sha
        })

    print(json.dumps(conversion, indent=4))
```

... and the shell script ...

```sh
python convert.py > howdy
rm Godeps           # the godeps vendoring uses a Godeps directory
mkdir Godeps
mv howdy Godeps/Godeps.json
dep init -v         # initializes the vendoring directory using Godeps/Godeps.json
dep ensure          # ensures all files are in the vendor directory
dep status          # status of the vendoring... pretty interesting.
```

There are other details that this PR should have

- [ ] Probably update .gitignore to ignore vendor
- [ ] Possibly cache the vendoring for circle ( https://circleci.com/docs/1.0/how-cache-works/#what-is-cached )
- [ ] Understand why the Gopkg.toml has far far far fewer entries than the lock.


crazy-canux:
Looks good.
Gdm will never make succeed.
When can I use dep?

smyrman:
> Understand why the Gopkg.toml has far far far fewer entries than the lock.

This is normal. The Gopkg.toml gives _constrains_ to `dep ensure -u`, and nothing more. It doesn't need to contain a single entry if there is something that doesn't just work, e.g. something with breaking changes.

goller:
@smyrman Do you have some links you could share about that?  I'd love to read up to make sure I understand.

smyrman:
The statement above is based on us having used the tool for various projects for the last several months. Our `Gopkg.toml` files are always very short, and always hand-edited to contain the most basic constraints only. You can put more constraints if you wish, but it's not needed to generate a valid lockfile which is automatically kept up to date based on code changes. This separates it massively from other tools such as e.g. `glide`, and makes it a lot more convenient to use.

Please refer to the official documentation: https://golang.github.io/dep/docs/Gopkg.toml.html

Issue #3413, opened 2017-10-31
Add ability to limit stats in the basicstats aggregator


danielnelson:
## Feature Request

Option to limit the new values to a single calculation (stdv or median or all)

Requested in #3402 by @xkilian

### Proposal:

Add method to select which statistics are collected, perhaps:
```toml
[[aggregators.basicstats]]
  stats = ["stdev", "s2"]
```

### Current behavior:

Must generate all stats.

Due to the way measurement filtering works with processors/aggregators, you cannot use the filtering option to limit the output.  This is because filtering applies to the metrics that enter the plugin, but not to the output.  Unlike inputs/outputs, these plugins really need two sets of filters.

### Desired behavior:

A method for limiting the output.

### Use case: [Why is this important (helps with prioritizing requests)]

This will make the new aggregator much more flexible for hosts that generate lots of metrics.

JeffAshton:
I'd be interested in implementing this functionality.  I'd like to amend the proposal to allow configuration  per field.  For example, one field I might want Min/Max computed, and another field I might just want the mean computed.

A default configuration for non matched fields makes sense to me.

danielnelson:
Thanks for the offer @JeffAshton, the first step I would take is to design the configuration.

I'll also mention a workaround for this issue, you can use the standard [measurement filtering](https://github.com/influxdata/telegraf/blob/master/docs/CONFIGURATION.md#measurement-filtering) on the output plugins to remove stats you don't want.

JeffAshton:
@danielnelson ,  I've send you a PR for the first part.  I'll look at adding the field specific stuff as a separate PR.

danielnelson:
Maybe we can do something similar to how the histogram aggregator is configured, here is an example of how it looks:
```toml
[[aggregators.histogram]]
  [[aggregators.histogram.config]]
  #   ## The set of buckets.
  buckets = [0.0, 10.0, 20.0, 30.0, 40.0, 50.0, 60.0, 70.0, 80.0, 90.0, 100.0]
  #   ## The name of metric.
  measurement_name = "diskio"
  #   ## The concrete fields of metric
  fields = ["io_time", "read_time", "write_time"]
```

So fields is a list, and you can specify the measurement name (actually you must set this in the histogram aggregator) which might be useful for conflicts.

JeffAshton:
Yeah, I can get behind that

JeffAshton:
[...]

PR #3580, opened 2017-12-13
Configurable basicstats

JeffAshton:
Basic implementation of configuring which stats are pushed from basicstats aggregator (https://github.com/influxdata/telegraf/issues/3413)

### Required for all PRs:

### DIFF
--- a/plugins/aggregators/basicstats/basicstats.go
+++ b/plugins/aggregators/basicstats/basicstats.go
@@ -1,6 +1,7 @@
 package basicstats
 
 import (
+	"log"
 	"math"
 
 	"github.com/influxdata/telegraf"
@@ -8,10 +9,22 @@ import (
 )
 
 type BasicStats struct {
-	cache map[uint64]aggregate
+	Stats []string `toml:"stats"`
+
+	cache       map[uint64]aggregate
+	statsConfig *configuredStats
 }
 
-func NewBasicStats() telegraf.Aggregator {
+type configuredStats struct {
+	count    bool
+	min      bool
+	max      bool
+	mean     bool
+	variance bool
+	stdev    bool
+}
+
+func NewBasicStats() *BasicStats {
 	mm := &BasicStats{}
 	mm.Reset()
 	return mm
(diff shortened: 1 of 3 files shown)


plugins/aggregators/basicstats/basicstats.go
```diff
@@ -1,17 +1,30 @@
...
 
 type BasicStats struct {
-	cache map[uint64]aggregate
+	Stats []string `toml:"stats"`
+
+	cache       map[uint64]aggregate
+	statsConfig *configuredStats
+}
+
+type configuredStats struct {
```
JeffAshton: Could also do this as flags, but not sure it's a win

plugins/aggregators/basicstats/basicstats.go
```diff
@@ -114,23 +127,101 @@ func (m *BasicStats) Add(in telegraf.Metric) {
...
+
 	for _, aggregate := range m.cache {
 		fields := map[string]interface{}{}
 		for k, v := range aggregate.fields {
-			fields[k+"_count"] = v.count
-			fields[k+"_min"] = v.min
-			fields[k+"_max"] = v.max
-			fields[k+"_mean"] = v.mean
+
+			if config.count {
```
JeffAshton: For the first pass, only using the configuration to control which stats are pushed.  I'm not trying to optimize the computation of each stat as they are added.

The balance between checking the configuration per calculation seems like a micro optimization we can make later.

plugins/aggregators/basicstats/basicstats.go
```diff
@@ -114,23 +127,101 @@ func (m *BasicStats) Add(in telegraf.Metric) {
 }
 
 func (m *BasicStats) Push(acc telegraf.Accumulator) {
```
JeffAshton: For the first pass, only using the configuration to control which stats are pushed. I'm not trying to optimize the computation of each stat as they are added.

The balance between checking the configuration per calculation seems like a micro optimization we can make later.

plugins/aggregators/basicstats/basicstats.go
```diff
@@ -114,23 +127,101 @@ func (m *BasicStats) Add(in telegraf.Metric) {
...
+				}
+				if config.stdev {
+					fields[k+"_stdev"] = math.Sqrt(variance)
+				}
 			}
 			//if count == 1 StdDev = infinite => so I won't send data
 		}
-		acc.AddFields(aggregate.name, fields, aggregate.tags)
+
+		if len(fields) > 0 {
```
JeffAshton: If an empty array is specified in the configuration, then we push nothing.  I plan to use this with the field specific configuration, in other words, only produce aggregations for known fields.

plugins/aggregators/basicstats/basicstats.go
```diff
@@ -114,23 +127,101 @@ func (m *BasicStats) Add(in telegraf.Metric) {
...
+			parsed.max = true
+		case "mean":
+			parsed.mean = true
+		case "s2":
+			parsed.variance = true
+		case "stdev":
+			parsed.stdev = true
+
+		default:
+			log.Printf("W! Unrecognized basic stat '%s', ignoring", name)
```
JeffAshton: Sample output:

```
2017-12-13T23:09:19Z I! Loaded outputs: file
2017-12-13T23:09:19Z I! Loaded inputs: inputs.win_perf_counters
2017-12-13T23:09:19Z I! Tags enabled: host=DESKTOP-R9332IM
2017-12-13T23:09:19Z I! Agent Config: Interval:1s, Quiet:false, Hostname:"DESKTOP-R9332IM", Flush Interval:10s
2017-12-13T23:09:30Z W! Unrecognized basic stat '[wacky]', ignoring
```

plugins/aggregators/basicstats/basicstats.go
```diff
@@ -114,23 +127,101 @@ func (m *BasicStats) Add(in telegraf.Metric) {
...
+		}
 	}
+
+	return parsed
+}
+
+func defaultStats() *configuredStats {
```
JeffAshton: Mostly for backwards compatibility.  If not specified in the config, you get the old default of everything.

plugins/aggregators/basicstats/basicstats_test.go
```diff
@@ -149,3 +149,211 @@ func TestBasicStatsDifferentPeriods(t *testing.T) {
...
+		"d_stdev": math.Sqrt(8),
+	}
+	expectedTags := map[string]string{
+		"foo": "bar",
+	}
+	acc.AssertContainsTaggedFields(t, "m1", expectedFields, expectedTags)
+}
+
+// Test only aggregating minimum and maximum
+func TestBasicStatsWithMinAndMax(t *testing.T) {
```
JeffAshton: Example of 2 stats

plugins/aggregators/basicstats/basicstats.go
danielnelson: I don't think you can determine an empty array from a unset array with our current toml parsing, so you should always have at least one field.

Remember you can use the regular [measurement filtering](https://github.com/influxdata/telegraf/blob/master/docs/CONFIGURATION.md#measurement-filtering) options to control what fields are added in the first place.

JeffAshton: You are correct.  I still feel like this check isn't a bad thing.  It would still happen if I passed an invalid stat, for example:

```

PR #3586, opened 2017-12-14
Add total message rates for RabbitMQ input plugin

kerams:
### Required for all PRs:


danielnelson:
Can the rate values be calculated from the message counts at query time?  If so this will be more flexible since you will be able to calculate over any unit of time such as messages/minute, as well as using less storage on the database.

We can definitely add `messages_delivered_get`.

kerams:
No, I don't think so. The counts increase with every message and as far as I can tell they never reset (they certainly do not represent message counts for the sample period, which is 5s by default).

danielnelson:
Maybe you can use this query:
```
select non_negative_difference(last("messages_published")) as messages_published_rate from rabbitmq_overview where time > now() - 10m group by time(1m)
```

kerams:
Oh, yeah, that didn't occur to me.

danielnelson:
I'd rather not have them, because they can't be aggregated and they are duplicate information.  However, we should add the `messages_delivered_get` and we could add some documentation about this query.

kerams:
Alright, I'll open a new PR for that

PR #2560, opened 2017-03-21
Libvirt plugin

JanStrauss:
This adds support to gather basic domain metrics from libvirt VMs. 
This PR is based on https://github.com/influxdata/telegraf/pull/357 and originally I only rebased it to the current master, replaced the `libvirt-go` library used and adapted it to API changes. 

However afaik this would have added a hard runtime dependency on the libvirt C libraries since `libvirt-go` depends on them (?). 

To avoid this I've rewritten the plugin to use the `virsh` command that is part of `libvirt-bin` as source for the metrics. This isn't really a nice/clean solution and if there is a better way I'm happy to update this PR. 

Also my first time touching go, so if there is stuff to improve please let me know

### Required for all PRs:

- [ ] CHANGELOG.md updated (we recommend not updating this until the PR has been approved by a maintainer)
- [x] Sign [CLA](https://influxdata.com/community/cla/) (if not already signed)
- [x] README.md updated (if adding a new plugin)

PR #3592, opened 2017-12-15
libvirtd plugin with DO go package

dahendel:
This libvirt plugin uses digital oceans libvirt-go package to pull metrics. As of now it pulls VM metrics and disk metrics. Network metrics are being worked on.

### Required for all PRs:


plugins/inputs/libvirtd/libvirtd.go
```diff
@@ -0,0 +1,194 @@
...
+
+const (
+	sampleConfig = `
+  lib_virt_sock = "/var/run/libvirt/libvirt-sock"
+  network_devices = []
+`
+)
+
+// MockPlugin struct should be named the same as the Plugin
+type LibVirtD struct {
```
danielnelson: I would just call this `Libvirt`

plugins/inputs/libvirtd/libvirtd.go
```diff
@@ -0,0 +1,194 @@
...
+		if device.Tag == "devices" {
+			for _, disk := range device.SelectElements("disk") {
+				if disk.SelectElement("source") != nil {
+					diskDevices = append(diskDevices, disk.SelectElement("target").SelectAttr("dev").Value)
+				}
+			}
+		}
+	}
+
+	l.DiskDevices = diskDevices
```
danielnelson: Return values instead of stashing them on the struct

plugins/inputs/libvirtd/libvirtd.go
```diff
@@ -0,0 +1,194 @@
...
+	fields["state"] = state
+	fields["max_memory"] = maxMem
+	fields["total_memory"] = maxMem
+	fields["cpu_count"] = cpuTotal
+	fields["cpu_time"] = cpuTime
+
+	tags["name"] = dom.Name
+	tags["id"] = string(dom.ID)
+
+	acc.AddFields("libvirtd_vmdata", fields, tags)
```
danielnelson: I would call this `libvirt_domain`, we want to stick to the upstream terminology if possible.

plugins/inputs/libvirtd/README.md
```diff
@@ -0,0 +1,91 @@
...
+### Example Output:
+
+```
+> livbirtd_disk,device=hdc,name=r-2824-QA,ID=̡,host=njcloudhost.dev.ena.net wr_bytes="&{3 0}",rd_operations="&{3 68}",
+flush_operations="&{3 0}",wr_operations="&{3 0}",rd_bytes="&{3 450836}",wr_total_times="&{3 0}",
+rd_total_times="&{3 97202811}",flush_total_times="&{3 0}",allocation=78528512i,capacity=78526464i,physical=78526464i 
+1513349120000000000
+
+
+>libvirtd_vmdata,name=r-2828-QA,id=̢,host=njcloudhost.dev.ena.net branch_misses="&{6 0}",ref_cpu_cycles="&{6 0}",
```
danielnelson: Obviously some type conversions still to do...

plugins/inputs/libvirtd/libvirtd.go
```diff
@@ -0,0 +1,194 @@
...
+	return sampleConfig
+}
+
+func init() {
+	virt := &LibVirtD{}
+	inputs.Add("libvirtd", func() telegraf.Input { return virt })
+}
+
+// Find all the disks in the XML data
+func (l *LibVirtD) getDiskDevices(dom libvirt.Domain, acc telegraf.Accumulator) {
```
danielnelson: I wonder if instead of reading through the domain definition to pull out the disks we should just iterate through the  volumes.  This is more familiar to me as a libvirt user, since volumes are one of main parts of the virsh command, while disks are somewhat hidden inside the domain.  Maybe we can avoid parsing xml as a bonus.

dahendel: In our use case for example, we use ceph and none of the vol commands list our disks. I could not find another way around getting the data without looping through the XML. I am rather new to libvirt so there may be another way that I have not figured out yet. Side note, it seems like most virsh commands are built into the libvirtd-go lib.

tubemeister:
Hi. I've been playing with your plugin and it works, but it also seems to leak filehandles (the libvirt sock) at an alarming rate, growing by 6-700 open files per minute on my test server and crashing every few hours due to running out.

I've added a c.Close() before the final return nil, this seems to work. I still get metrics and the open file count stays at a reasonable level.

This was my first ever line of Go code, so by all means sanity check the hell out of it. ;-) 

Thanks!

tubemeister:
Oh, and you seem to have a typo in the volumes metric: livbirt_dom_volumes. :-)

dahendel:
@tubemeister thanks! I will get that changed.

dahendel:
I updated the Gather func with defer c.Close() and corrected the typo. However I am having trouble getting it to build because of the Microsoft lib.
```
# github.com/Microsoft/ApplicationInsights-Go/appinsights
../../Microsoft/ApplicationInsights-Go/appinsights/telemetry.go:382:27: multiple-value uuid.NewV4() in single-value context
../../Microsoft/ApplicationInsights-Go/appinsights/telemetry.go:412:23: multiple-value uuid.NewV4() in single-value context
../../Microsoft/ApplicationInsights-Go/appinsights/telemetrycontext.go:88:52: multiple-value uuid.NewV4() in single-value context
```

danielnelson:
I think it will be easiest to update the Godeps file by hand right now.  I merged the separate Windows and Linux Godeps into one but gdm doesn't like it.  I would reset it to the latest version on master and then add your new deps in by hand.

dahendel:
Closing in preference to #2560

PR #3594, opened 2017-12-16
Add information about how to set postfix permissions

### LINKED ISSUE #3591: Postfix plugin permissions
Hello,

I'm running 1.5.0 and wanted to benefit from the Postfix plugin.
How did you guys (@phemmer ?) did it with telegraf running with telegraf user which doesn't have access to files owned by postfix in /var/spool/postfix?

Example:
Error in plugin [inputs.postfix]: error scanning queue active: open /var/spool/postfix/active: permission denied

Thanks

danielnelson:
closes  #3591

### Required for all PRs:

- [ ] ~~Has appropriate unit tests.~~

### DIFF
--- a/plugins/inputs/postfix/README.md
+++ b/plugins/inputs/postfix/README.md
@@ -13,6 +13,23 @@ For each of the active, hold, incoming, maildrop, and deferred queues (http://ww
   # queue_directory = "/var/spool/postfix"
 ```
 
+#### Permissions:
+
+Telegraf will need read access to the files in the queue directory.  You may
+need to alter the permissions of this directory to provide access to the
+telegraf user.
+
+Unix permissions:
+```sh
+$ sudo chgrp -R telegraf /var/spool/postfix
+$ sudo chmod -R g+rX /var/spool/postfix
+```
+
+Posix ACL:
+```sh
+$ sudo setfacl -R -m u:telegraf:r /var/spool/postfix
+```
+
 ### Measurements & Fields:
 
 - postfix_queue
--- a/plugins/inputs/postfix/README.md
+++ b/plugins/inputs/postfix/README.md
@@ -22,12 +22,12 @@ telegraf user.
 Unix permissions:
 ```sh
 $ sudo chgrp -R telegraf /var/spool/postfix
-$ sudo chmod -R g+rX /var/spool/postfix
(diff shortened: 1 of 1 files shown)


danielnelson:
@dynek Can you review?

phemmer:
One caveat about these methods is that the directories in `deferred` are created on first use. So if they're created after these commands are run, then they'll need to be run again.

However in the case of ACL, you can set a default ACL, which should get inherited by the new directories (unless postfix is doing something silly like explicitly performing a `chown` instead of using a umask).

plugins/inputs/postfix/README.md
```diff
@@ -13,6 +13,23 @@ For each of the active, hold, incoming, maildrop, and deferred queues (http://ww
...
+
+Unix permissions:
+```sh
+$ sudo chgrp -R telegraf /var/spool/postfix
+$ sudo chmod -R g+rX /var/spool/postfix
+```
+
+Posix ACL:
+```sh
+$ sudo setfacl -R -m u:telegraf:r /var/spool/postfix
```
phemmer: It'll need execute as well to be able to read the dir. `setfacl` supports `X` (upper case) to only set execute bit on dirs, not files.

danielnelson:
I add the +X to setfacl, and the setgid bit to chmod

dynek:
Now getting:
Dec 16 08:02:52 server.host.tld postfix/postdrop[7766]: warning: mail_queue_enter: create file maildrop/6997.7766: Permission denied

Trying to fix it :-/

dynek:
OK fixed it with a previous dump or permissions (ls -lsaR) and `postfix check`.

Will check it a bit later.

Apparently we should not be overriding chgrping of postdrop for some dirs.

phemmer:
yeah, looks like the `maildrop` directory does utlize group permissions. So the `chgrp` should be targetted at just the `active`, `deferred`, `hold`, `incoming`, & `maildrop` dirs.

dynek:
Why not leave permissions as they are but play with ACL only?
```
sudo setfacl -Rdm u:telegraf:rX /var/spool/postfix/{active,hold,incoming,deferred,maildrop}
```

danielnelson:
Updated proposed docs, can you review again?

PR #3596, opened 2017-12-16
Add messages_delivered_get to rabbitmq_overview

kerams:
As discussed in #3586. `messages_delivered_get` represents the grand total of messages delivered since the node's startup.

### Required for all PRs:

### DIFF
--- a/plugins/inputs/rabbitmq/README.md
+++ b/plugins/inputs/rabbitmq/README.md
@@ -52,6 +52,7 @@ For additional details reference the [RabbitMQ Management HTTP Stats](https://cd
   - messages (int, messages)
   - messages_acked (int, messages)
   - messages_delivered (int, messages)
+  - messages_delivered_get (int, messages)
   - messages_published (int, messages)
   - messages_ready (int, messages)
   - messages_unacked (int, messages)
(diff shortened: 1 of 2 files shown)


plugins/inputs/rabbitmq/README.md
```diff
@@ -115,6 +116,12 @@ For additional details reference the [RabbitMQ Management HTTP Stats](https://cd
 
 ### Sample Queries:
 
+Message rates for the entire node can be calculated from total message counts. For instance, to get the rate of messages published per minute, use this query:
+
+```
+SELECT NON_NEGATIVE_DIFFERENCE(LAST("messages_published")) AS messages_published_rate
+FROM rabbitmq_overview WHERE time > now() - 10m GROUP BY time(1m)
```
phemmer: I might recommend `non_negative_derivative(..., 1m)` instead of `difference()`. That way the query always gives you the 1m rate, regardless of the group by interval. Both methods are perfectly valid, just the derivative tends to be more useful in graphing as your group by interval changes based on the time range.

kerams: Alright.

PR #3633, opened 2018-01-03
Fix deliver_get in RabbitMQ

kerams:
The json output has apparently never been parsed correctly.

### Required for all PRs:

### DIFF
--- a/plugins/inputs/rabbitmq/rabbitmq.go
+++ b/plugins/inputs/rabbitmq/rabbitmq.go
@@ -72,7 +72,7 @@ type MessageStats struct {
 	AckDetails        Details `json:"ack_details"`
 	Deliver           int64
 	DeliverDetails    Details `json:"deliver_details"`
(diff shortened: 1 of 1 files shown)


danielnelson:
I think this value is reported in rabbitmq_overview as `messages_delivered` and in rabbitmq_queue as `messages_deliver_get`.  Can you explain what this change does?

kerams:
These are 2 separate values (https://pulse.mozilla.org/doc/stats.html). In #3596 I exposed `messages_deliver_get`  into the `rabbitmq_overview` measurement, but it turns out it's always reported as 0 because of the bug that I am fixing in this PR.

kerams:
In other words, both `messages_delivered` and `messages_delivered_get` should be available in `rabbitmq_overview` as well as `rabbitmq_queue`, but the latter is currently always 0.

danielnelson:
I see, because of the underscore we need the struct tag.

PR #3210, opened 2017-09-08
CrateDB output plugin

felixge:
Hello,

this PR implements an output plugin for [CrateDB](http://crate.io/) using their [Postgres Wire Protocol](https://crate.io/docs/crate/reference/protocols/postgres.html).

Please let me know what kind of changes, if any, you'd like me to make in order to get this merged.

Many thanks in advance 😊
Felix

### Required for all PRs:

- [x] Signed [CLA](https://influxdata.com/community/cla/).
- [x] Associated README.md updated.
- [x] Has appropriate unit tests.

PR #3638, opened 2018-01-04
Convert datid to a string for backwards compatibility

### LINKED ISSUE #3622: PostgreSQL input error after upgrade to 1.5.0 - datid is integer instead of string
Since I upgraded Telegraf via apt to 1.5.0 I get an error with postgresql input:

```
Dec 27 10:39:17 vs-psql-03 telegraf[12274]: 2017-12-27T09:39:17Z I! Starting Telegraf v1.5.0
Dec 27 10:39:17 vs-psql-03 telegraf[12274]: 2017-12-27T09:39:17Z I! Loaded outputs: influxdb
Dec 27 10:39:17 vs-psql-03 telegraf[12274]: 2017-12-27T09:39:17Z I! Loaded inputs: inputs.system inputs.net inputs.nginx inputs.processes inputs.swap inputs.postgresql inputs.cpu inputs.disk inputs.diskio inputs.kernel inputs.mem inputs.redis
Dec 27 10:39:17 vs-psql-03 telegraf[12274]: 2017-12-27T09:39:17Z I! Tags enabled: environment=production host=vs-psql-03 rails_env=production role=cloud
Dec 27 10:39:17 vs-psql-03 telegraf[12274]: 2017-12-27T09:39:17Z I! Agent Config: Interval:10s, Quiet:false, Hostname:"vs-psql-03", Flush Interval:10s
Dec 27 10:39:30 vs-psql-03 telegraf[12274]: 2017-12-27T09:39:30Z E! Field type conflict, dropping conflicted points: Response Error: Status Code [400], expected [204], [field type conflict: input field "datid" on measurement "postgresql" is type integer, already exists as type string]
```

My current `telegraf.conf` postgresql input config:

```
  [[inputs.postgresql]]
    address = "host=127.0.0.1 user=postgres sslmode=disable dbname=cloud"
    databases = ["cloud"]
```

danielnelson:
closes #3622

### Required for all PRs:

### DIFF
--- a/plugins/inputs/postgresql/postgresql.go
+++ b/plugins/inputs/postgresql/postgresql.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"regexp"
 	"sort"
+	"strconv"
 	"strings"
 
 	// register in driver.
@@ -205,7 +206,15 @@ func (p *Postgresql) accRow(row scanner, acc telegraf.Accumulator) error {
 	for col, val := range columnMap {
 		_, ignore := ignoredColumns[col]
 		if !ignore {
-			fields[col] = *val
+			if col == "datid" {
+				// Convert column to a string for backwards compatibility with
+				// Telegraf <1.5
+				if val, ok := (*val).(int64); ok {
+					fields[col] = strconv.FormatInt(val, 10)
+				}
+			} else {
+				fields[col] = *val
+			}
 		}
 	}
 	acc.AddFields("postgresql", fields, tags)
--- a/plugins/inputs/postgresql/postgresql_test.go
+++ b/plugins/inputs/postgresql/postgresql_test.go
@@ -51,7 +51,6 @@ func TestPostgresqlGeneratesMetrics(t *testing.T) {
 		"checkpoints_req",
 		"checkpoints_timed",
 		"maxwritten_clean",
-		"datid",
 		"numbackends",
 	}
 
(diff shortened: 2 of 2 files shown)


danielnelson:
@james-lawrence I could use some advice here if you can spare the time.  Please ignore the code here so far, it won't work.

I introduced a regression into 1.5 causing oid types to be converted to ints instead of strings, so I'm trying to undo that.  Since this also affects postgresql_extensible I would like it to apply to all oids.

It seems like to fix this I will need to pass the `row.ColumnTypes()` into `accRow()` and then use that to setup the `columnMap` and `columnVars` with the correct types.  Do you think this is the best way to go about a fix?

james-lawrence:
@danielnelson do you know what commit caused the regression so I can take a look?

danielnelson:
It was part of #3210, when we updated pgx.  I believe it was added in https://github.com/jackc/pgx/commit/bec9bd261b6a0e11ecde620f2f7fd8a7ea07e529, quite some time back.

james-lawrence:
ah, so basically the driver added first class support for the type which strips the default output.

unfortunately atm there isn't a good way of doing this. once https://github.com/jackc/pgx/pull/365 is merged we should be able to configure the output for types.

danielnelson:
Too bad, I guess I'll just whip a release note for this item then.  Thanks for the help

PR #3647, opened 2018-01-08
Don't create AMQP exchange if not configured

### LINKED ISSUE #3646: AMQP_Consumer Altering Exchanges
### Relevant telegraf.conf:
exchange config

### System info:
Telegraf 1.5.0

### Steps to reproduce:

1. Create a RabbitMQ topic
2. Create another RabbitMQ topic or x-consistent-hash bound to the first topic (due to issue #2801)
3. Bind your queues to the second topic or x-consistent-hash

Your architecture should look something like this:
exch.telegraf (topic) -> exch.telegraf.consumer (x-consistent-hash) -> q.telegraf.1

4. Configure Telegraf amqp_consumer to use exchange exch.telegraf and one of the queues

On startup, the architecture now looks like:
exch.telegraf (topic) -> q.telegraf.1 AND exch.telegraf.consumer (x-consistent-hash) -> q.telegraf.1

### Expected behavior:
Exchanges are not reconfigured on startup for consuming messages

### Actual behavior:
This results in duplicate messages being sent to each queue.  It's much worse when you have multiple queues.  To get around this problem I have to provide a bogus exchange to just to make Telegraf happy when it starts up and leaves the main exchange configuration alone.

### Additional info:
When consuming messages, the only thing the client needs to provide the broker is the queue name and the prefetch count.  You should be able to delete lines 183 through 217.

hanej:
### Required for all PRs:

Removed AMQP publisher specific code and documentation from the consumer plugin.  This has been validated in our environment.  Fixes #3646

### DIFF
--- a/plugins/inputs/amqp_consumer/amqp_consumer.go
+++ b/plugins/inputs/amqp_consumer/amqp_consumer.go
@@ -65,6 +65,10 @@ func (a *AMQPConsumer) SampleConfig() string {
   url = "amqp://localhost:5672/influxdb"
   ## AMQP queue name
   queue = "telegraf"
+  ## Optional AMQP exchange
+  exchange = "telegraf"
+  ## Optional binding Key
+  binding_key = "#"
 
   ## Maximum number of messages server should give to the worker.
   prefetch_count = 50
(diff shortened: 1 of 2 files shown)


danielnelson:
If I understand this correctly, the user would now need to define the exchange and the queue, as well as bind them before using the plugin?

hanej:
That is correct

danielnelson:
We will need to find a way to provide what you need without breaking backwards compatibility, perhaps we could modify the code to skip the creation of the exchange if `exchange` is unset and then declare the exchange and queue only if they do not exist?

hanej:
That works.  I’ll give it a shot.

hanej:
The change has been committed.  It works with and without the exchange configuration option.

plugins/inputs/amqp_consumer/amqp_consumer.go
```diff
@@ -180,40 +180,42 @@ func (a *AMQPConsumer) connect(amqpConf *amqp.Config) (<-chan amqp.Delivery, err
...
-	err = ch.QueueBind(
-		q.Name,       // queue
-		a.BindingKey, // binding-key
-		a.Exchange,   // exchange
-		false,
-		nil,
-	)
-	if err != nil {
-		return nil, fmt.Errorf("Failed to bind a queue: %s", err)
+		err = ch.QueueBind(
```
danielnelson: I think we should always QueueBind, so that the binding key works even if no exchange/queue is created.

danielnelson:
Closing, assuming this is no longer required.  One related pull request is #5831 which adds passive queue declaration as well.

PR #3346, opened 2017-10-17
Ipset input plugin

sajoupa:
### Required for all PRs:

- [X] Signed [CLA](https://influxdata.com/community/cla/).
- [X] Associated README.md updated.
- [X] Has appropriate unit tests.

The ipset input plugin allows collecting metrics from Linux ipset counters.

PR #3649, opened 2018-01-09
feat: socketstat input plugin

sajoupa:
### Required for all PRs:

Hi,

The `ss` command (package iproute2) gives very useful metrics about sockets on Linux machines, this is a plugin to expose them with telegraf.

Thanks.


danielnelson:
@sajoupa Thanks for the pull request.  I think there may be two high level concerns with this pr:
- Since a new series is made for every state+src+dst, this output format will be very high cardinality, you will probably need to pair this with a short retention policy to avoid overloading your database.
- Shelling out to ss will be somewhat expensive, perhaps we can bypass this and directly retrieve the data from /proc or another method.

sajoupa:
Hi @danielnelson , you're totally right about data cardinality, I added a warning to the README.
I'm quite reluctant to parse /proc/net/<protocol> however. We would duplicate (and then maintain) logic which is already implemented by ss. I don't think that it would be worth it, given the small cost of shelling out to ss.

I have also updated the code to make the same changes which had been requested for the ipset input plugin (use a timeout, buffio scanner, ... ref https://github.com/influxdata/telegraf/pull/3346)

danielnelson:
I'm not sure if I will want to merge this in, in the meantime we can leave the pr open for anyone who would like to use it.

sajoupa:
@danielnelson which of the two reasons you first mentioned would block this ?
I can replace shelling out to ss by parsing /proc (I believe it's better to shell out, but both solutions work), however, there's no reducing data cardinality if we want to get ss data.
My intent is to have ss exported data sent to graylog. According to https://github.com/Graylog2/graylog-plugin-netflow and https://www.graylog.org/post/snmp-netflow-plug-ins-extend-graylog-network-infrastructure-support, graylog can handle the cardinality of network packets metadata.
I just find it a lot more convenient to get data from telegraf agents than network equipments. (also, activating netflow on network equipments won't help for cloud VMs).
Thanks.

danielnelson:
Shelling out to ss is the bigger issue, we can add warnings to the documentation about cardinality and how to deal with it.

sajoupa:
@danielnelson Hi, I looked further into it and think shelling out is the only option.
ss uses the tcp_diag kernel module to get its data much faster than parsing /proc/net (https://www.cyberciti.biz/files/ss.html). Also, /proc/net won't provide counters such as bytes_acked, received, etc.
And after looking at ss's 4900+ lines of code, I'm not sure we want to recreate them into telegraf ;)

sajoupa:
Hi @danielnelson it looks like the planning board is no longer.
Any chance we can get socket stats into telegraf at some point, and if so is there anything I can do to help it progress ?

Thanks !

danielnelson:
I'm not sure on this one yet, for now I want to keep it on hold primarily because of the concerns above.

plugins/inputs/socketstat/README.md
```diff
@@ -0,0 +1,66 @@
+# SocketStat plugin
+
+The socketstat plugin gathers indicators from established connections, using iproute2's `ss` command.
+
+The `ss` command does not require specific privileges.
+
+The output format will have a very high cardinality. It should either be stored by an engine which doesn't suffer from it, of with a short retention policy.
```
srebhan, line 7: Could you please make the warning **bold** as otherwise people will miss it.

plugins/inputs/socketstat/README.md
```diff
@@ -0,0 +1,66 @@
...
+        - segs_out (integer, count)
+        - segs_in (integer, count)
+        - data_segs_out (integer, count)
+        - data_segs_in (integer, count)
+
+### Tags:
+
+- All measurements have the following tags:
+    - proto
+    - state (tcp, dccp and sctp)
```
srebhan, line 36: `state` is already a field and in the light of exploding cardinality we might want to remove the tag!?

plugins/inputs/socketstat/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+#### recent ss version
+
+```
+$ ss -in --tcp
+State      Recv-Q Send-Q Local Address:Port               Peer Address:Port
+ESTAB      0      0      127.0.0.1:53692              127.0.0.1:7778
+	 cubic wscale:7,7 rto:224 rtt:21.55/14.181 ato:40 mss:25600 cwnd:10 bytes_acked:34525 bytes_received:2663883 segs_out:2331 segs_in:2136 data_segs_out:654 data_segs_in:1680 send 95.0Mbps lastsnd:15112 lastrcv:15084 lastack:15084 pacing_rate 190.1Mbps rcv_rtt:93.021 rcv_space:60281 minrtt:0.028
+
+./telegraf --config telegraf.conf --input-filter socketstat --test
```
srebhan, line 52: I think you can leave this out only showing the telegraf output. It would be nice tough to have an version of the `ss` you used for the user to get an idea if he is in the _recent_ or _old_ range.

plugins/inputs/socketstat/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+#### older ss version
+
+```
+$ ss -in --tcp
+State       Recv-Q Send-Q                           Local Address:Port                             Peer Address:Port
+ESTAB      0      0                                                  192.168.1.21:38776                                           1.2.3.4:18080
+	 cubic wscale:7,7 rto:284 rtt:84.5/8 ato:40 cwnd:5 ssthresh:3 send 685.4Kbps rcv_rtt:88 rcv_space:29200
+
+./telegraf --config telegraf.conf --input-filter socketstat --test
```
srebhan, line 64: Same here. Leave the lines out and add an example version for `ss`.

plugins/inputs/socketstat/socketstat.go
```diff
@@ -0,0 +1,188 @@
...
+// Socketstat is a telegraf plugin to gather indicators from established connections, using iproute2's  `ss` command.
+type Socketstat struct {
+	lister      socketLister
+	SocketProto []string
+	Timeout     internal.Duration
+}
+
+type socketLister func(proto string, Timeout internal.Duration) (*bytes.Buffer, error)
+
+const measurement = "socketstat"
```
srebhan, line 27: This is not required and should be set by telegraf anyway!

plugins/inputs/socketstat/socketstat.go
```diff
@@ -0,0 +1,188 @@
...
+
+var defaultTimeout = internal.Duration{Duration: time.Second}
```
srebhan, line 29: Maybe move this to an `Init()` function called at construction time of the plugin.

plugins/inputs/socketstat/socketstat.go
```diff
@@ -0,0 +1,188 @@
...
+  ## The default timeout of 1s for ss execution can be overridden here:
+  # timeout = "1s"
+`
+}
+
+// Gather gathers indicators from established connections
+func (ss *Socketstat) Gather(acc telegraf.Accumulator) error {
+	if len(ss.SocketProto) == 0 {
+		return nil
+	}
```
srebhan, line 51: You could check this in the `Init()` function mentioned above and fail early if no protocol is given. This reduces the amount of surprise... :-)

plugins/inputs/socketstat/socketstat.go
```diff
@@ -0,0 +1,188 @@
...
+	var out bytes.Buffer
+	cmd.Stdout = &out
+	err = internal.RunTimeout(cmd, Timeout.Duration)
+	if err != nil {
+		return &out, fmt.Errorf("error running ss -in --%s: ", err)
+	}
+	return &out, nil
+}
+
+var validFields = "(bytes_acked|bytes_received|segs_out|segs_in|data_segs_in|data_segs_out)"
```
srebhan, line 93: Can be a constant.

plugins/inputs/socketstat/socketstat.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+	return nil
+}
+
+func socketList(proto string, Timeout internal.Duration) (*bytes.Buffer, error) {
+	// Check that ss is installed
+	ssPath, err := exec.LookPath("ss")
+	if err != nil {
+		return new(bytes.Buffer), err
+	}
```
srebhan, line 74: Should be done in `Init()`.

plugins/inputs/socketstat/socketstat.go
```diff
@@ -0,0 +1,188 @@
...
+		e = ss.parseAndGather(out, proto, acc)
+		if e != nil {
+			acc.AddError(e)
+			continue
+		}
+	}
+	return nil
+}
+
+func socketList(proto string, Timeout internal.Duration) (*bytes.Buffer, error) {
```
srebhan, line 69: I think this is your `Init()` function...

plugins/inputs/socketstat/socketstat.go
```diff
@@ -0,0 +1,188 @@
...
+	if err != nil {
+		return new(bytes.Buffer), err
+	}
+
+	// Add needed args
+	cmdName := ssPath
+	var args []string
+	args = append(args, "-in")
+	args = append(args, "--"+proto)
+
```
srebhan, line 81: This should be done in the `Init()` function once.

plugins/inputs/socketstat/socketstat.go
```diff
@@ -0,0 +1,188 @@
...
+	err = internal.RunTimeout(cmd, Timeout.Duration)
+	if err != nil {
+		return &out, fmt.Errorf("error running ss -in --%s: ", err)
+	}
+	return &out, nil
+}
+
+var validFields = "(bytes_acked|bytes_received|segs_out|segs_in|data_segs_in|data_segs_out)"
+var validValues = regexp.MustCompile("^" + validFields + ":[0-9]+$")
+var beginsWithBlank = regexp.MustCompile("^\\s+.*$")
```
srebhan, line 95: Should be constructed in the `Init()` function.

plugins/inputs/socketstat/socketstat.go
```diff
@@ -0,0 +1,188 @@
...
+				flushData = false
+			}
+			// Delegate the real parsing to getTagsAndState, which manages various
+			// formats depending on the protocol
+			tags, fields = getTagsAndState(proto, words)
+			flushData = true
+		} else {
+			for _, word := range words {
+				if validValues.MatchString(word) {
+					kv := strings.Split(word, ":")
```
srebhan, line 131: Maybe add a comment that `kv` has to have two fields due to the matching regular expression. Otherwise people might wonder why you don't check the length of `kv`.

plugins/inputs/socketstat/socketstat.go
```diff
@@ -0,0 +1,188 @@
...
+					fields[kv[0]], err = strconv.ParseUint(kv[1], 10, 64)
+					if err != nil {
+						continue
```
srebhan, line 134: You might want to output the error somewhere as otherwise we might silently drop fields.

plugins/inputs/socketstat/socketstat.go
```diff
@@ -0,0 +1,188 @@
...
+	switch proto {
+	case "udp", "raw":
+		words = append([]string{"dummy"}, words...)
+	case "tcp", "dccp", "sctp":
+		fields["state"] = words[0]
+	}
+	switch proto {
+	case "tcp", "udp", "raw", "dccp", "sctp":
+		local := strings.Split(words[3], ":")
+		remote := strings.Split(words[4], ":")
```
srebhan, line 161: Maybe use `LastIndex()` instead of splitting and reassembling?

plugins/inputs/socketstat/socketstat.go
```diff
@@ -0,0 +1,188 @@
...
+		tags["remote_port"] = remote[len(remote)-1]
+	case "unix", "packet":
+		fields["netid"] = words[0]
+		tags["local_addr"] = words[4]
+		tags["local_port"] = words[5]
+		tags["remote_addr"] = words[6]
+		tags["remote_port"] = words[7]
+	}
+	var err error
+	fields["recv_q"], err = strconv.ParseUint(words[1], 10, 64)
```
srebhan, line 174: You should also check this error!

plugins/inputs/socketstat/socketstat.go
```diff
@@ -0,0 +1,188 @@
...
+	fields["send_q"], err = strconv.ParseUint(words[2], 10, 64)
+	if err != nil {
+	}
```
srebhan, line 177: What is the idea behind this? If you want to ignore the error, why getting it in the first place? I think you really need to handle invalid values here.

plugins/inputs/socketstat/socketstat.go
```diff
@@ -0,0 +1,188 @@
...
+	return tags, fields
+}
+
+func init() {
+	inputs.Add("socketstat", func() telegraf.Input {
+		return &Socketstat{
+			lister:  socketList,
+			Timeout: defaultTimeout,
```
srebhan, line 185: Default value handling can also be done in the `Init()` function.

srebhan (commented):
Hey @sajoupa, I think we can include this despite shelling out. If people complain we can fix it later anyway. However, I've several comments in the code. Furthermore, I want to encourage you to add a `Init()` function for all one-time setup code.
Last, but most important I want to understand your `flushData` idea. As far as I can see it is used to guard the `acc.AddFields()` call. Now in the loop `acc.AddFields()` is called exactly under the following conditions
1. The line starts with blanks. In this case all valid words are added and afterwards the `acc.AddFields()` is fired.
1. You first encounter a line starting non-blank and in a second run you hit the blank-line of 1.
1. You first encounter a line starting non-blank and in a second run you again git the non-blank starting line. Here you first flush and have the "first non-blank linestarting" state.

Now I realized that you **only** reset the tags and fields when entering the "non-blank-linestarting" case. If you encounter multiple lines starting with a blank, you might still carry data of the previous line if the current line does not have the same field. Is this what you want?

Furthermore, the only state where you could not directly call `acc.AddFields()` is case number 2 (first non-blank then blank). However in this case you might, in the second step, overwrite fields of the first step.

If this really is what you are intending, I think it's worth a comment as it is not obvious otherwise.

sajoupa:
Hi @srebhan, just a quick note to thank you for the review, I'll address the comments as soon as possible.

plugins/inputs/socketstat/README.md
sajoupa: This was  a typo, there is no state tag (only a field). Will fix.

sajoupa:
Comments addressed, thanks for the suggestions.
I'm not entirely convinced by the addition of an Init() function though, haven't done that yet.

plugins/inputs/socketstat/socketstat.go
srebhan: Minor: Typo in the comment.

plugins/inputs/socketstat/socketstat.go
srebhan: Declaration of global variables should be avoided. I think those two should go into the `Socketstat` struct.

plugins/inputs/socketstat/socketstat.go
srebhan: Not addressed yet.

srebhan:
@sajoupa some comments are not yet addressed, most severly the missing error check and the declaration of the regexp variables. Can you please go trough my comments and mark them resolved if you think they are (I cannot do this)!? This eases my review-life.
Regarding the `Init()` we can discuss, but I think the code is really more readable with such a function and you save some steps in the Gather function...

etc/telegraf.conf
```diff
@@ -6937,8 +6937,17 @@
 #   ## Content encoding for message payloads, can be set to "gzip" to or
 #   ## "identity" to apply no encoding.
 #   # content_encoding = "identity"
-
-
+#
+#
```
srebhan, line 6941: No hashes here, was correct before. Furthermore, this file needs no update as it is automatically updated at release time IIRC.

plugins/inputs/socketstat/socketstat.go
```diff
@@ -0,0 +1,194 @@
...
+
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/internal"
+	"github.com/influxdata/telegraf/plugins/inputs"
+)
+
+// Socketstat is a telegraf plugin to gather indicators from established connections, using iproute2's  `ss` command.
+type Socketstat struct {
+	lister      socketLister
+        Log         telegraf.Logger
```
srebhan, line 21: There seems to be a formatting error. Please rebase your code to trigger new CI runs.

plugins/inputs/socketstat/socketstat.go
srebhan: Can we at least move this to where the variable is used (in `init()`) so we don't work with global variables here!

plugins/inputs/socketstat/socketstat.go
srebhan: Ok if you are not willing to introduce an `Init()` function, then either fail here or remove the `if` section completely as the for loop is not executed if `ss.SocketProto` is empty.

plugins/inputs/socketstat/socketstat.go
```diff
@@ -0,0 +1,194 @@
...
+	ssPath, err := exec.LookPath("ss")
+	if err != nil {
+		return new(bytes.Buffer), err
+	}
+
+	// Add needed args
+	cmdName := ssPath
+	var args []string
+	args = append(args, "-in")
+	args = append(args, "--"+proto)
```
srebhan, line 81: If you don't add an `Init()` function can we please do
```suggestion
	args := []string {"-in", "--"+proto}
```

plugins/inputs/socketstat/socketstat.go
srebhan: Typo is still there.

plugins/inputs/socketstat/socketstat.go
srebhan: Please mark as resolved.

plugins/inputs/socketstat/socketstat.go
srebhan: Error check is still missing.

plugins/inputs/socketstat/socketstat.go
srebhan: Replace `defaultTimeout` by the definition above to get rid of the global variable if you don't want to add an `Init()` function.

srebhan:
Ok I get that you don't want to add an `Init()` function (even though I think your code would benefit). Can you please address the rest of the comments to get this done!?

srebhan:
@sajoupa any news here?

srebhan:
@sajoupa any news on this?

srebhan:

plugins/inputs/socketstat/README.md
sajoupa: done.

plugins/inputs/socketstat/README.md
sajoupa: now fixed.

plugins/inputs/socketstat/README.md
sajoupa: done.

plugins/inputs/socketstat/README.md
sajoupa: done.

plugins/inputs/socketstat/socketstat.go
sajoupa: Done (added an Init() and moved this variable initialization there).

plugins/inputs/socketstat/socketstat.go
sajoupa: done (moved to Init())

plugins/inputs/socketstat/socketstat.go
sajoupa: It's now in the SocketStat struct, as advised in another comment.

plugins/inputs/socketstat/socketstat.go
sajoupa: done.

plugins/inputs/socketstat/socketstat.go
sajoupa: Done.

plugins/inputs/socketstat/socketstat.go
sajoupa: Updated the comment.

plugins/inputs/socketstat/socketstat.go
sajoupa: It's now using LastIndex().

plugins/inputs/socketstat/socketstat.go
sajoupa: It's now caught, and logged.

plugins/inputs/socketstat/socketstat.go
sajoupa: It's now caught, and logged.

plugins/inputs/socketstat/socketstat.go
sajoupa: Now handled in the Init() function.

plugins/inputs/socketstat/socketstat.go
sajoupa: Fixed. I'll look into triggering a CI run when I'm sure that I've addressed all the comments.

plugins/inputs/socketstat/socketstat.go
sajoupa: I added an Init() function, and updated to do as suggested here.

etc/telegraf.conf
sajoupa: I reverted the etc/telegraf.conf changes

sajoupa:
> Hey @sajoupa, I think we can include this despite shelling out. If people complain we can fix it later anyway. However, I've several comments in the code. Furthermore, I want to encourage you to add a `Init()` function for all one-time setup code. Last, but most important I want to understand your `flushData` idea. As far as I can see it is used to guard the `acc.AddFields()` call. Now in the loop `acc.AddFields()` is called exactly under the following conditions
> 
>     1. The line starts with blanks. In this case all valid words are added and afterwards the `acc.AddFields()` is fired.
> 
>     2. You first encounter a line starting non-blank and in a second run you hit the blank-line of 1.
> 
>     3. You first encounter a line starting non-blank and in a second run you again git the non-blank starting line. Here you first flush and have the "first non-blank linestarting" state.
> 
> 
> Now I realized that you **only** reset the tags and fields when entering the "non-blank-linestarting" case. If you encounter multiple lines starting with a blank, you might still carry data of the previous line if the current line does not have the same field. Is this what you want?
> 
> Furthermore, the only state where you could not directly call `acc.AddFields()` is case number 2 (first non-blank then blank). However in this case you might, in the second step, overwrite fields of the first step.
> 
> If this really is what you are intending, I think it's worth a comment as it is not obvious otherwise.

It is indeed the intended behavior. I tried to improve the comments to explain it, without being too verbose.
It's quite subjective so suggestions are still welcome :)

sajoupa:
I @srebhan I finally got back to this PR, and implemented what you suggested. It is indeed much better with the Init() function (and complies with the telegraf development guidelines).
If you (or another reviewer) feel like reviewing it again, could you reopen the PR (I can't) ?
Thanks.

plugins/inputs/socketstat/socketstat.go
```diff
@@ -0,0 +1,219 @@
...
+// Socketstat is a telegraf plugin to gather indicators from established connections, using iproute2's  `ss` command.
+type Socketstat struct {
+	BeginsWithBlank *regexp.Regexp
+	cmdName         string
+	Lister          socketLister
+	Log             telegraf.Logger
+	Measurement     string
+	SocketProto     []string
+	Timeout         config.Duration
+	ValidValues     *regexp.Regexp
```
srebhan, line 28: As far as I can see, only two of those variables (`Timeout` and `SocketProto`) are actually config options, so please add toml-tags and make all possible values "unexported":
```suggestion
	SocketProto     []string `toml:"socket_proto"`
	Timeout         config.Duration `toml:"timeout"`
	Log             telegraf.Logger `toml:"-"`

	beginsWithBlank *regexp.Regexp
	validValues     *regexp.Regexp
	cmdName         string
	lister          socketLister
	measurement     string
```

plugins/inputs/socketstat/socketstat.go
```diff
@@ -0,0 +1,219 @@
...
+func (ss *Socketstat) Description() string {
+	return "Gather indicators from established connections, using iproute2's `ss` command."
+}
+
+// SampleConfig returns sample configuration options
+func (ss *Socketstat) SampleConfig() string {
+	return `
+  ## ss can display information about tcp, udp, raw, unix, packet, dccp and sctp sockets
+  ## Specify here the types you want to gather
+  # socket_proto = [ "tcp", "udp" ]
```
srebhan, line 43: Maybe
```suggestion
  ## List of protocol types to collect
  # protocols = [ "tcp", "udp" ]
```

plugins/inputs/socketstat/socketstat.go
```diff
@@ -0,0 +1,219 @@
...
+	ss.ValidValues = regexp.MustCompile("^" + validFields + ":[0-9]+$")
+	ss.BeginsWithBlank = regexp.MustCompile(`^\s+.*$`)
+
+	return nil
+}
+
+func init() {
+	inputs.Add("socketstat", func() telegraf.Input {
+		return &Socketstat{
+			Lister: socketList,
```
srebhan, line 216: Can we please also do this in `Init()`?

plugins/inputs/socketstat/socketstat.go
```diff
@@ -0,0 +1,219 @@
...
+
+func socketList(cmdName string, proto string, timeout config.Duration) (*bytes.Buffer, error) {
+	// Run ss for the given protocol, return the output as bytes.Buffer
+	args := []string{"-in", "--" + proto}
+	cmd := exec.Command(cmdName, args...)
+	var out bytes.Buffer
+	cmd.Stdout = &out
+	err := internal.RunTimeout(cmd, time.Duration(timeout))
+	if err != nil {
+		return &out, fmt.Errorf("error running ss -in --%s: ", err)
```
srebhan, line 76: This error message seems wrong. Did you mean
```suggestion
		return &out, fmt.Errorf("error running ss -in --%s: %v", proto, err)
```

plugins/inputs/socketstat/socketstat.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+	return &out, nil
+}
+
+func (ss *Socketstat) parseAndGather(data *bytes.Buffer, proto string, acc telegraf.Accumulator) error {
```
srebhan, line 81: Just a question here. AFAICS your parsing is that complicated due to the output format of `ss`. But you can influence this format. Did you try with `ss --no-header --oneline  ...` which provides the metrics with one-line per connection and without header?

plugins/inputs/socketstat/socketstat.go
```diff
@@ -0,0 +1,219 @@
...
+	flushData := false
+	for scanner.Scan() {
+		line := scanner.Text()
+		if line == "" {
+			continue
+		}
+		words := strings.Fields(line)
+
+		var err error
+		if !ss.BeginsWithBlank.MatchString(line) {
```
srebhan, line 105: Maybe rename this to `ss.isNewConnection` making it easier to read IMO.

plugins/inputs/socketstat/socketstat.go
```diff
@@ -0,0 +1,219 @@
...
+				acc.AddFields(ss.Measurement, fields, tags)
+			}
+
+			// Delegate the real parsing to getTagsAndState, which manages various
+			// formats depending on the protocol.
+			tags, fields = getTagsAndState(proto, words, ss.Log)
+
+			// This line containted metrics, so record that.
+			flushData = true
+		} else {
```
srebhan, line 118: Please add a `continue` in the `if` condition and unindent the `else` path.

plugins/inputs/socketstat/socketstat.go
```diff
@@ -0,0 +1,219 @@
...
+		tags["local_port"] = words[3][localIndex+1:]
+		tags["remote_addr"] = words[4][:remoteIndex]
+		tags["remote_port"] = words[4][remoteIndex+1:]
+	case "unix", "packet":
+		fields["netid"] = words[0]
+		tags["local_addr"] = words[4]
+		tags["local_port"] = words[5]
+		tags["remote_addr"] = words[6]
+		tags["remote_port"] = words[7]
+	}
```
srebhan, line 174: Please fuse the two.

plugins/inputs/socketstat/socketstat.go
```diff
@@ -0,0 +1,219 @@
...
+	var err error
+	fields["recv_q"], err = strconv.ParseUint(words[1], 10, 64)
+        if err != nil {
+                log.Infof("Couldn't read recv_q in: %s", words)
```
srebhan, line 178: Is this really an "info"? I think it's at least a warning or an error as the resulting metric has a missing field.

plugins/inputs/socketstat/socketstat.go
```diff
@@ -0,0 +1,219 @@
...
+        }
+	fields["send_q"], err = strconv.ParseUint(words[2], 10, 64)
+	if err != nil {
+		log.Infof("Couldn't read send_q in: %s", words)
```
srebhan, line 182: Same as above.

plugins/inputs/socketstat/socketstat_test.go
```diff
@@ -0,0 +1,145 @@
...
+ESTAB      0      0             192.168.1.21:6514                      192.168.1.21:443
+	cubic wscale:7,7 rto:204 rtt:0.057/0.033 ato:40 mss:22976 cwnd:10 bytes_acked:1126 bytes_received:532644751 segs_out:211249 segs_in:211254 data_segs_out:2 data_segs_in:211251 send 32247.0Mbps lastsnd:299082764 lastrcv:5248 lastack:5252 rcv_rtt:3.532 rcv_space:186557 minrtt:0.047
+ESTAB      0      0            192.168.122.1:55194                    192.168.122.1:6514
+	cubic wscale:7,7 rto:204 rtt:0.034/0.01 ato:40 mss:65483 cwnd:10 bytes_acked:790782896 bytes_received:1126 segs_out:333361 segs_in:333361 data_segs_out:333358 data_segs_in:2 send 154077.6Mbps lastsnd:5248 lastrcv:443892492 lastack:5248 rcv_rtt:250 rcv_space:43690 minrtt:0.009
+ESTAB      0      0                127.0.0.1:7778                         127.0.0.1:50378
+	cubic wscale:7,7 rto:220 rtt:16.009/21.064 ato:44 mss:65483 cwnd:10 bytes_acked:19983121 bytes_received:266383 segs_out:15431 segs_in:17633 data_segs_out:15119 data_segs_in:5098 send 327.2Mbps lastsnd:9792 lastrcv:9840 lastack:9748 pacing_rate 654.4Mbps retrans:0/1 rcv_rtt:129800 rcv_space:44057 minrtt:0.043`,
```
srebhan, line 42: Can you maybe put this data in files (in a `testdata` subdir) and load them in the tests?

srebhan (commented):
@sajoupa thanks for working on this plugin! It looks much nicer in my eyes. One more major comment is about using `ss --no-header --oneline --in ...` to have all data in one-line per connections. I guess this would ease the parsing quite a bit... Do you see any reason _not_ to use it?

plugins/inputs/socketstat/socketstat.go
sajoupa: Indeed, this is better. I pushed a fix (also fixing the struct to match)

plugins/inputs/socketstat/socketstat.go
sajoupa: Done.

plugins/inputs/socketstat/socketstat.go
sajoupa: done.

plugins/inputs/socketstat/socketstat.go
sajoupa: That's a good idea ! Unfortunately the --oneline option isn't present in 5.5.0 (e.g. Ubuntu 18.04 LTS), which was the version I was using when originally writing this plugin.
It's too early to deprecate such versions, so imho it's better to keep the code as is for now.

plugins/inputs/socketstat/socketstat.go
sajoupa: Much better. I renamed this way.

plugins/inputs/socketstat/socketstat.go
sajoupa: I believe that if we did that, we could miss the last measurement, by not running:
```
        if flushData {
                acc.AddFields(ss.measurement, fields, tags)
        }
```
at the end of this function.

plugins/inputs/socketstat/socketstat.go
sajoupa: This would create duplicates.

plugins/inputs/socketstat/socketstat.go
sajoupa: Fixed.

plugins/inputs/socketstat/socketstat.go
sajoupa: Fixed (now a warning).

plugins/inputs/socketstat/socketstat.go
srebhan: Oh I think my comment was misleading here. Please only move the `Lister = socketList` to `Init()`. We need the call to `inputs.Add` in `init()` (lowercase) as this function is called then on import and registers the plugin to be used... Sorry for the inconvinience!

plugins/inputs/socketstat/socketstat.go
srebhan: TBH I have mixed feelings here. I guess the code is so much simpler with the new option and if we now add the more complex code, we will never simplify it again. So what about make the version `>=5.5.0` a requirement and add support for older versions as an external plugin for the meantime?

plugins/inputs/socketstat/socketstat.go
srebhan: I don't think so, because this code is outside of the for-loop, isn't it? To me it seems that all code below if is in the else path... Did I miss something?

plugins/inputs/socketstat/socketstat.go
sajoupa: Right, I did that. I wasn't sure there'd be a way to assign a custom value to ss.lister for tests, but found it.

plugins/inputs/socketstat/socketstat.go
sajoupa: In my use case, which is managing a few thousand servers of various OS versions, more than half of the machines would be excluded. So I have a strong preference to keep it this way for now (and for a few more years, until versions without --oneline are EOL for most distributions, e.g. Ubuntu, Debian, RH, etc.).

plugins/inputs/socketstat/socketstat.go
sajoupa: My bad, you were right. I also reverted the test, as I found thenegative test (`if !<condition>`) less intuitive to read.

plugins/inputs/socketstat/socketstat_test.go
sajoupa: I moved it to a separate file. My understanding is that it must be in the same directory.

plugins/inputs/socketstat/README.md
```diff
@@ -0,0 +1,55 @@
+# SocketStat plugin
+
+The socketstat plugin gathers indicators from established connections, using iproute2's `ss` command.
+
+The `ss` command does not require specific privileges.
+
+**The output format will have a very high cardinality. It should either be stored by an engine which doesn't suffer from it, of with a short retention policy.**
```
srebhan, line 7:
```suggestion
**WARNING: The output format will produce series with very high cardinality.** You should either store those by an engine which doesn't suffer from it, use a short retention policy or do appropriate filtering.
```

plugins/inputs/socketstat/socketstat.go
```diff
@@ -0,0 +1,222 @@
...
+				if ss.validValues.MatchString(word) {
+					// kv will have 2 fields because it matched the regexp
+					kv := strings.Split(word, ":")
+					fields[kv[0]], err = strconv.ParseUint(kv[1], 10, 64)
+					if err != nil {
+						ss.Log.Infof("Couldn't parse metric: %s", word)
+						continue
+					}
+				}
+			}
```
srebhan, line 121: You can save one indention level here
```suggestion
			for _, word := range words {
				if !ss.validValues.MatchString(word) {
					continue
				}
				// kv will have 2 fields because it matched the regexp
				kv := strings.Split(word, ":")
				fields[kv[0]], err = strconv.ParseUint(kv[1], 10, 64)
				if err != nil {
					ss.Log.Infof("Couldn't parse metric: %s", word)
					continue
				}
			}
```

plugins/inputs/socketstat/socketstat.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+
+	// Initialize regexps to validate input data
+	validFields := "(bytes_acked|bytes_received|segs_out|segs_in|data_segs_in|data_segs_out)"
+	ss.validValues = regexp.MustCompile("^" + validFields + ":[0-9]+$")
+	ss.isNewConnection = regexp.MustCompile(`^\s+.*$`)
+
+	if ss.lister == nil {
+		ss.lister = socketList
+	}
```
srebhan, line 213: `Init` is only called once, so you can safely do
```suggestion
	ss.lister = socketList
```

plugins/inputs/socketstat/socketstat_test.go
```diff
@@ -0,0 +1,84 @@
...
+			}
+			acc := new(testutil.Accumulator)
+
+			err := ss.Init()
+			require.NoError(t, err)
+
+			err = acc.GatherError(ss.Gather)
+			if !reflect.DeepEqual(tt.err, err) {
+				t.Errorf("%d: expected error '%#v' got '%#v'", i, tt.err, err)
+			}
```
srebhan, line 32:
```suggestion
			require.ErrorIs(t, err, tt.err)
```

plugins/inputs/socketstat/socketstat_test.go
```diff
@@ -0,0 +1,84 @@
...
+					t.Errorf("%d: expected 0 fields if no protocol specified got %d", i, n)
+				}
+				return
+			}
+			if len(tt.tags) == 0 {
+				n := acc.NFields()
+				if n != 0 {
+					t.Errorf("%d: expected 0 values got %d", i, n)
+				}
+				return
```
srebhan, line 45:
```suggestion
				n := acc.NFields()
				require.Equalf(t, 0,  n, "%d: expected 0 values got %d", i, n)
				return
```

plugins/inputs/socketstat/socketstat_test.go
```diff
@@ -0,0 +1,84 @@
...
+			}
+			n := 0
+			for j, tags := range tt.tags {
+				for k, fields := range tt.fields[j] {
+					if len(acc.Metrics) < n+1 {
+						t.Errorf("%d: expected at least %d values got %d", i, n+1, len(acc.Metrics))
+						break
+					}
```
srebhan, line 53:
```suggestion
					require.LessOrEqual(t, len(acc.Metrics), n)
```

plugins/inputs/socketstat/socketstat_test.go
```diff
@@ -0,0 +1,84 @@
...
+					m := acc.Metrics[n]
+					if !reflect.DeepEqual(m.Measurement, ss.measurement) {
+						t.Errorf("%d %d %d: expected measurement '%#v' got '%#v'\n", i, j, k, ss.measurement, m.Measurement)
+					}
+					if !reflect.DeepEqual(m.Tags, tags) {
+						t.Errorf("%d %d %d: expected tags\n%#v got\n%#v\n", i, j, k, tags, m.Tags)
+					}
+					if !reflect.DeepEqual(m.Fields, fields) {
+						t.Errorf("%d %d %d: expected fields\n%#v got\n%#v\n", i, j, k, fields, m.Fields)
+					}
```
srebhan, line 63: Please use `require.Equal` here.

plugins/inputs/socketstat/socketstat_test.go
```diff
@@ -0,0 +1,84 @@
...
+		SocketProto: []string{"foobar"},
+		lister: func(cmdName string, proto string, timeout config.Duration) (*bytes.Buffer, error) {
+			return new(bytes.Buffer), errFoo
+		},
+	}
+	acc := new(testutil.Accumulator)
+	err := acc.GatherError(ss.Gather)
+	if !reflect.DeepEqual(err, errFoo) {
+		t.Errorf("Expected error %#v got\n%#v\n", errFoo, err)
+	}
```
srebhan, line 83: Please use `require.ErrorIs` here.

plugins/inputs/socketstat/testdata.go
```diff
@@ -0,0 +1,62 @@
+package socketstat
+
+var tests = []struct {
```
srebhan, line 3: What I meant here is to generate multiple _text_ files for the test cases that can be generated via `ss -... > some_test_file` and the define the expected metrics in the tests. Take a look at e.g. [this plugin](https://github.com/influxdata/telegraf/blob/master/plugins/inputs/nvidia_smi/nvidia_smi_test.go).

srebhan (commented):
Nice update @sajoupa! I agree with keeping the backward compatible parsing. Some more minor comments in the code. Please also use the `require` functions in your tests and checkout the mentioned test-example. Looking forward to the next round. We are almost there.

plugins/inputs/socketstat/socketstat.go
sajoupa: This will erase a custom lister for tests ?

plugins/inputs/socketstat/socketstat_test.go
sajoupa: Pushed a fix.

plugins/inputs/socketstat/socketstat_test.go
sajoupa: Pushed a fix.

plugins/inputs/socketstat/socketstat_test.go
sajoupa: Pushed a fix (we actually want it to be Greater).

plugins/inputs/socketstat/socketstat_test.go
sajoupa: Pushed a fix.

plugins/inputs/socketstat/socketstat_test.go
sajoupa: Pused a fix.

plugins/inputs/socketstat/testdata.go
sajoupa: fair enough ;) I pushed a fix to implement that.

plugins/inputs/socketstat/socketstat.go
srebhan: It will. However, you can modify that `lister` _after_ calling `Init()` in the tests. IMO the tests should not clutter the main code if this can be avoided...

plugins/inputs/socketstat/socketstat_test.go
srebhan: Looks like it didn't make it to the PR!?

srebhan (commented):
Very nice move forward @sajoupa! I have some more small comments.
Furthermore, please rebase your PR to latest master in order to get CI testing.

plugins/inputs/socketstat/socketstat.go
sajoupa: Gotcha. Pushed a fix.

plugins/inputs/socketstat/socketstat.go
```diff
@@ -0,0 +1,223 @@
...
+
+	// Check that ss is installed, get its path
+	ssPath, err := exec.LookPath("ss")
+	if err != nil {
+		return err
+	}
+	ss.cmdName = ssPath
+
+	ss.measurement = "socketstat"
```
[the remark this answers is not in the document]
sajoupa: Done.

plugins/inputs/socketstat/socketstat.go
```diff
@@ -0,0 +1,220 @@
...
+	var out bytes.Buffer
+	cmd.Stdout = &out
+	err := internal.RunTimeout(cmd, time.Duration(timeout))
+	if err != nil {
+		return &out, fmt.Errorf("error running ss -in --%s: %v", proto, err)
+	}
+	return &out, nil
+}
+
+func (ss *Socketstat) parseAndGather(data *bytes.Buffer, proto string, acc telegraf.Accumulator) error {
```
srebhan, line 83: Please make `acc` the first parameter. It's an unwritten convention... sorry!

plugins/inputs/socketstat/socketstat.go
```diff
@@ -0,0 +1,220 @@
...
+				}
+			}
+			if !flushData {
+				ss.Log.Warnf("Found orphaned metrics: %s", words)
+				ss.Log.Warnf("Added them to the last known connection.")
+			}
+			acc.AddFields(measurement, fields, tags)
+			flushData = false
+			continue
+		}
```
srebhan, line 131: Please move the error to the innermost for loop as this is the only location you use it or even better get rid of the declaration...
```suggestion
		if ss.isNewConnection.MatchString(line) {
			// A line with starting whitespace means metrics about the current connection.
			// We should never get 2 consecutive such lines. If we do, log a warning and in
			// a best effort, extend the metrics from the 1st line with the metrics of the 2nd
			// one, possibly overwriting.
			for _, word := range words {
				if !ss.validValues.MatchString(word) {
					continue
				}
				// kv will have 2 fields because it matched the regexp
				kv := strings.Split(word, ":")
				v, err := strconv.ParseUint(kv[1], 10, 64)
				if err != nil {
					ss.Log.Infof("Couldn't parse metric %q: %v", word, err)
					continue
				}
				fields[kv[0]] = v
			}
			if !flushData {
				ss.Log.Warnf("Found orphaned metrics: %s", words)
				ss.Log.Warnf("Added them to the last known connection.")
			}
			acc.AddFields(measurement, fields, tags)
			flushData = false
			continue
		}
```

plugins/inputs/socketstat/socketstat.go
```diff
@@ -0,0 +1,220 @@
...
+		fields["netid"] = words[0]
+		tags["local_addr"] = words[4]
+		tags["local_port"] = words[5]
+		tags["remote_addr"] = words[6]
+		tags["remote_port"] = words[7]
+	}
+	var err error
+	fields["recv_q"], err = strconv.ParseUint(words[1], 10, 64)
+	if err != nil {
+		log.Warnf("Couldn't read recv_q in: %s", words)
```
srebhan, line 181: Please show the error...
```suggestion
		log.Warnf("Couldn't read recv_q in %q: %v", words, err)
```

plugins/inputs/socketstat/socketstat.go
```diff
@@ -0,0 +1,220 @@
...
+	}
+	fields["send_q"], err = strconv.ParseUint(words[2], 10, 64)
+	if err != nil {
+		log.Warnf("Couldn't read send_q in: %s", words)
```
srebhan, line 185: See above.

plugins/inputs/socketstat/socketstat.go
```diff
@@ -0,0 +1,220 @@
...
+	ss.lister = socketList
+
+	return nil
+}
+
+func init() {
+	inputs.Add("socketstat", func() telegraf.Input {
+		return &Socketstat{}
+	})
+}
```
srebhan, line 220: Please allow for zero timeout (aka. no timeout) if the user wishes to...
```suggestion
	// Initialize regexps to validate input data
	validFields := "(bytes_acked|bytes_received|segs_out|segs_in|data_segs_in|data_segs_out)"
	ss.validValues = regexp.MustCompile("^" + validFields + ":[0-9]+$")
	ss.isNewConnection = regexp.MustCompile(`^\s+.*$`)

	ss.lister = socketList

	return nil
}

func init() {
	inputs.Add("socketstat", func() telegraf.Input {
		return &Socketstat{Timeout: time.Second}
	})
}
```

plugins/inputs/socketstat/socketstat_test.go
```diff
@@ -0,0 +1,123 @@
...
+			i++
+			ss := &Socketstat{
+				SocketProto: tt.proto,
+			}
+			acc := new(testutil.Accumulator)
+
+			err := ss.Init()
+			if err != nil && !strings.Contains(err.Error(), "executable file not found in $PATH") {
+				require.NoError(t, err)
+			}
```
srebhan, line 80: I think this is easier to understand
```suggestion
			if err != nil {
				require.EqualError(t, err, "executable file not found in $PATH")
			}
```

plugins/inputs/socketstat/socketstat_test.go
```diff
@@ -0,0 +1,123 @@
...
+	errFoo := errors.New("error foobar")
+	ss := &Socketstat{
+		SocketProto: []string{"foobar"},
+	}
+	ss.lister = func(cmdName string, proto string, timeout config.Duration) (*bytes.Buffer, error) {
+		return new(bytes.Buffer), errFoo
+	}
+	acc := new(testutil.Accumulator)
+	err := acc.GatherError(ss.Gather)
+	require.ErrorIs(t, errFoo, err)
```
srebhan, line 122: Please use `EqualError` here
```suggestion
	ss := &Socketstat{
		SocketProto: []string{"foobar"},
	}
	ss.lister = func(cmdName string, proto string, timeout config.Duration) (*bytes.Buffer, error) {
		return new(bytes.Buffer), errFoo
	}
	acc := new(testutil.Accumulator)
	err := acc.GatherError(ss.Gather)
	require.EqualErrorIs(t, err, "error foobar")
```

srebhan (commented):
@sajoupa this is almost perfect. Just a few small things left. I tried to make suggestions in the hope to make this the last iteration. I think we are almost there.

plugins/inputs/socketstat/socketstat_test.go
sajoupa: > @sajoupa this is almost perfect. Just a few small things left. I tried to make suggestions in the hope to make this the last iteration. I think we are almost there.

No worries at all, I'm not in a rush to get this plugin live, and it's never too late to make better code anyway. Thanks for all the reviews !

plugins/inputs/socketstat/socketstat.go
sajoupa: fixed.

plugins/inputs/socketstat/socketstat.go
sajoupa: fixed

plugins/inputs/socketstat/socketstat.go
sajoupa: fixed.

plugins/inputs/socketstat/socketstat.go
sajoupa: Fixed (with config.Duration(time.Second))

plugins/inputs/socketstat/socketstat_test.go
sajoupa: Pushed a fix (using a variable for "error foobar")

plugins/inputs/socketstat/socketstat.go
```diff
@@ -0,0 +1,216 @@
...
+				v, err := strconv.ParseUint(kv[1], 10, 64)
+				if err != nil {
+					ss.Log.Infof("Couldn't parse metric %q: %v", word, err)
+					continue
+				}
+				fields[kv[0]] = v
+			}
+			if !flushData {
+				ss.Log.Warnf("Found orphaned metrics: %s", words)
+				ss.Log.Warnf("Added them to the last known connection.")
```
srebhan, line 126: Maybe
```suggestion
				ss.Log.Warn("Added them to the last known connection.")
```
as we don't pass any variable for formatting.

plugins/inputs/socketstat/socketstat.go
```diff
@@ -0,0 +1,216 @@
...
+		out, e := ss.lister(ss.cmdName, proto, ss.Timeout)
+		if e != nil {
+			acc.AddError(e)
+			continue
+		}
+		e = ss.parseAndGather(acc, out, proto)
+		if e != nil {
+			acc.AddError(e)
+			continue
+		}
```
srebhan, line 65: Can you please name the error `err` as this is how we usually call it. This makes it easier for a reviewer / reader familiar with telegraf to work through the code...

plugins/inputs/socketstat/socketstat.go
```diff
@@ -0,0 +1,216 @@
...
+	var out bytes.Buffer
+	cmd.Stdout = &out
+	err := internal.RunTimeout(cmd, time.Duration(timeout))
+	if err != nil {
+		return &out, fmt.Errorf("error running ss -in --%s: %v", proto, err)
+	}
+	return &out, nil
+}
+
+func (ss *Socketstat) parseAndGather(acc telegraf.Accumulator, data *bytes.Buffer, proto string) error {
```
srebhan, line 83: This function does not return an error (only `nil`). So please change the signature to reflect this and make it visible to the caller.
```suggestion
func (ss *Socketstat) parseAndGather(acc telegraf.Accumulator, data *bytes.Buffer, proto string) {
```

plugins/inputs/socketstat/socketstat.go
```diff
@@ -0,0 +1,216 @@
...
+	if flushData {
+		acc.AddFields(measurement, fields, tags)
+	}
+	return nil
+}
+
+func getTagsAndState(proto string, words []string, log telegraf.Logger) (map[string]string, map[string]interface{}) {
+	tags := map[string]string{}
+	fields := make(map[string]interface{})
+	tags["proto"] = proto
```
srebhan, line 154:
```suggestion
	tags := map[string]string{
		"proto": proto,
	}
	fields := make(map[string]interface{})
```

plugins/inputs/socketstat/socketstat.go
```diff
@@ -0,0 +1,216 @@
...
+	}
+	var err error
+	fields["recv_q"], err = strconv.ParseUint(words[1], 10, 64)
+	if err != nil {
+		log.Warnf("Couldn't read recv_q in %q: %v", words, err)
+	}
+	fields["send_q"], err = strconv.ParseUint(words[2], 10, 64)
+	if err != nil {
+		log.Warnf("Couldn't read send_q in %q: %v", words, err)
+	}
```
srebhan, line 186: Maybe you should not set the fields on error. This way they will be `nil`/not existing and allow for downstream processing to detect the issue. Otherwise they will be zero without an indication of an invalid value.
```suggestion
	v, err := strconv.ParseUint(words[1], 10, 64)
	if err != nil {
		log.Warnf("Couldn't read recv_q in %q: %v", words, err)
	} else {
		fields["recv_q"] = v
	}
	v, err = strconv.ParseUint(words[2], 10, 64)
	if err != nil {
		log.Warnf("Couldn't read send_q in %q: %v", words, err)
	} else {
		fields["send_q"] = v
	}
```

srebhan (commented):
Thanks for your patience @sajoupa! There are a few additional improvements I'd like to see going in. Some are just a matter of taste, so please speak-up if you don't like them.

plugins/inputs/socketstat/socketstat.go
sajoupa: fixed.

plugins/inputs/socketstat/socketstat.go
sajoupa: fixed (and removed the `return nil` at the end of the function)

srebhan (approved):
Awesome! Thanks for driving this home @sajoupa! LGTM.

powersj (approved):

sajoupa:
> Awesome! Thanks for driving this home @sajoupa! LGTM.

Cheers, thanks for the patient reviews.

PR #3662, opened 2018-01-12
Parse string type values in Dropwizard parser

rickard-von-essen:
Handle parsing of a Dropwizard metric with string values e.g.

```json
{
  "version": "3.0.0",
  "gauges": {
    "measurement": {
      "value": "a string"
    }
  }
}
```

```
> measurement,metric_type=gauge value="a string" <timestamp>
```

### Required for all PRs:


rickard-von-essen:
FYI @atzoum

rickard-von-essen:
Follow up of #2846

atzoum:
nice catch! If you don't mind, I incorporated your changes in #3664 which also handles boolean fields and additionally properly escapes special characters from tags and fields

rickard-von-essen:
Great! I'll close this then.

PR #3664, opened 2018-01-12
Dropwizard format improvements

atzoum:
- Parse string and boolean fields 
- Handle special characters in parsed tags and fields

### Required for all PRs:

### DIFF
--- a/plugins/parsers/dropwizard/parser.go
+++ b/plugins/parsers/dropwizard/parser.go
@@ -14,6 +14,9 @@ import (
 	"github.com/tidwall/gjson"
 )
 
+var fieldEscaper = strings.NewReplacer("\\", "\\\\", "\"", "\\\"")
+var keyEscaper = strings.NewReplacer(" ", "\\ ", ",", "\\,", "=", "\\=")
+
 // Parser parses json inputs containing dropwizard metrics,
 // either top-level or embedded inside a json field.
 // This parser is using gjon for retrieving paths within the json file.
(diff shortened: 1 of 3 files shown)


atzoum:
Includes changes from @rickard-von-essen in #3662

plugins/parsers/dropwizard/parser.go
```diff
@@ -211,18 +214,22 @@ func (p *Parser) readDWMetrics(metricType string, dwms interface{}, metrics []te
 
 			measurementWithTags := measurementName
 			for tagName, tagValue := range tags {
-				tagKeyValue := fmt.Sprintf("%s=%s", tagName, tagValue)
+				tagKeyValue := fmt.Sprintf("%s=%s", keyEscaper.Replace(tagName), keyEscaper.Replace(tagValue))
```
danielnelson: We should use `metrics.New(` to make the metrics, then we won't need to worry about escaping here, and it should be faster as well.

atzoum: I wouldn't want to loose the default functionality. When you don't provide a custom template of your own, then each dropwizard measurement name is treated as line protocol keys `measurement<,tag_set>`.

danielnelson: Ah right, maybe in the future I'll add some functions that can handle this more elegantly... I'm making some changes to metrics this month so maybe I can include it.

atzoum: having a utility function to parse the keys part of a measurement sounds good! Don't hesitate to ping once there is something in place.

Issue #3158, opened 2017-08-23
http_listener_test hangs indefinitely ~50% of invocations 


rickard-von-essen:
master (8c68bd9ddbc2a7495796f44048b013c5098065f8)
Darwin Kernel Version 17.0.0: Thu Jul 20 17:24:26 PDT 2017; root:xnu-4556.0.0.1.2~1/RELEASE_X86_64 x86_64
go version go1.8.3 darwin/amd64

### Steps to reproduce:

```shell
$ go test -v ./plugins/inputs/http_listener
=== RUN   TestWriteHTTP
2017/08/23 09:07:03 I! Started HTTP listener service on :0
2017/08/23 09:07:03 E! http_listener received a single line longer than the maximum of 65536 bytes
2017/08/23 09:07:03 I! Stopped HTTP listener service on  :0
--- PASS: TestWriteHTTP (0.00s)
=== RUN   TestWriteHTTPNoNewline
2017/08/23 09:07:03 I! Started HTTP listener service on :0
2017/08/23 09:07:03 I! Stopped HTTP listener service on  :0
--- PASS: TestWriteHTTPNoNewline (0.00s)
=== RUN   TestWriteHTTPMaxLineSizeIncrease
2017/08/23 09:07:03 I! Started HTTP listener service on :0
2017/08/23 09:07:03 I! Stopped HTTP listener service on  :0
--- PASS: TestWriteHTTPMaxLineSizeIncrease (0.00s)
=== RUN   TestWriteHTTPVerySmallMaxBody
2017/08/23 09:07:03 I! Started HTTP listener service on :0
2017/08/23 09:07:03 I! Stopped HTTP listener service on  :0
--- PASS: TestWriteHTTPVerySmallMaxBody (0.00s)
=== RUN   TestWriteHTTPVerySmallMaxLineSize
2017/08/23 09:07:03 I! Started HTTP listener service on :0
2017/08/23 09:07:03 I! Stopped HTTP listener service on  :0
--- PASS: TestWriteHTTPVerySmallMaxLineSize (0.00s)
=== RUN   TestWriteHTTPLargeLinesSkipped
2017/08/23 09:07:03 I! Started HTTP listener service on :0
2017/08/23 09:07:03 E! http_listener received a single line longer than the maximum of 100 bytes
2017/08/23 09:07:03 I! Stopped HTTP listener service on  :0
--- PASS: TestWriteHTTPLargeLinesSkipped (0.00s)
[...]

PR #3668, opened 2018-01-12
Add notes about running tests on mac and the need to install dependencies

jmccann:
Some notes/steps that may help others when trying to run tests for first time.

### Required for all PRs:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -460,6 +460,21 @@ func init() {
 Before opening a pull request you should run the linter checks and
 the short tests.
 
+**Mac OSX Notes**
+
+On Mac OSX, tests can hang (https://github.com/influxdata/telegraf/issues/3158),
+so a workaround is to run tests from within a docker container:
+
+```bash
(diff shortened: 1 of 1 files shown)


danielnelson:
I'm torn on this, while this is good information it would be so much nicer if we could figure out why the tests are hanging.

jmccann:
I'll leave it to you.  ;)

My thoughts on this though are that having the info would have saved me personally several hours of time.  The issue has been around for 5-6months so not sure how timely it will get addressed.  

Perhaps a note to #3158 could be added to remove the section I've added once the issue is fixed?

danielnelson:
Sorry about the headache this caused, I think what we should do is add this to the test that hangs, is this enough to get tests working on Mac?:
```
if runtime.GOOS != "darwin" {
	t.Skip("Skipping due to hang on Darwin; issue #3158")
}
```

danielnelson:
We went through the tests this summer and got everything working on Darwin.  It is probably a matter of time before they break again, since we have no CI system setup for testing, which is unfortunate but I'm not sure how to solve.

I think instead of adding this documentation, we just need to fix any test issues as they come up.

PR #3186, opened 2017-08-29
NATS Monitoring Input Plugin

levex:
Hello,

I have a NATS bus in my setup and I'm looking to monitor its state via Grafana via InfluxDB, so I wrote a very simple Telegraf input plugin that would let me get information about the bus.

I feel the work is far from complete, but putting it here in case it gets more feedback!

Thanks!

### Required for all PRs:

- [x] Signed [CLA](https://influxdata.com/community/cla/).
- [x] Associated README.md updated.
- [x] Has appropriate unit tests.

PR #3674, opened 2018-01-16
NATS Monitoring Input Plugin

mjs:
This plugin gathers metrics a [NATS](https://nats.io/) server monitoring endpoint. Currently, only the `/varz` endpoint is used. See https://nats.io/documentation/server/gnatsd-monitoring/ for more information.

Note that this is a continuation of #3186. The original author isn't working on that PR any more.

### Required for all PRs:

### DIFF
--- a/plugins/inputs/nats/nats.go
+++ b/plugins/inputs/nats/nats.go
@@ -62,7 +62,7 @@ func (n *Nats) Gather(acc telegraf.Accumulator) error {
 		return err
 	}
 
-	acc.AddFields("nats",
+	acc.AddFields("nats_varz",
 		map[string]interface{}{
 			"in_msgs":           stats.InMsgs,
 			"out_msgs":          stats.OutMsgs,
--- a/plugins/inputs/nats/nats.go
+++ b/plugins/inputs/nats/nats.go
@@ -66,7 +66,7 @@ func (n *Nats) Gather(acc telegraf.Accumulator) error {
 		return err
 	}
 
-	acc.AddFields("nats_varz",
+	acc.AddFields("nats",
 		map[string]interface{}{
 			"in_msgs":           stats.InMsgs,
 			"out_msgs":          stats.OutMsgs,
--- a/plugins/inputs/nats/nats.go
+++ b/plugins/inputs/nats/nats.go
@@ -66,14 +66,18 @@ func (n *Nats) Gather(acc telegraf.Accumulator) error {
 		map[string]interface{}{
 			"in_msgs":           stats.InMsgs,
 			"out_msgs":          stats.OutMsgs,
-			"uptime":            time.Since(stats.Start).Seconds(),
+			"in_bytes":          stats.InBytes,
+			"out_bytes":         stats.OutBytes,
+			"uptime":            stats.Now.Sub(stats.Start).Nanoseconds(),
+			"cores":             stats.Cores,
+			"cpu":               stats.CPU,
+			"mem":               stats.Mem,
 			"connections":       stats.Connections,
 			"total_connections": stats.TotalConnections,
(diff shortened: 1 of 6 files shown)


levex:
Thanks @mjs for taking this over!

plugins/inputs/nats/nats.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+func (n *Nats) Gather(acc telegraf.Accumulator) error {
+	url, err := url.Parse(n.Server)
+	if err != nil {
+		return err
+	}
+	url.Path = path.Join(url.Path, "varz")
+
+	client := n.createHTTPClient()
```
danielnelson: Only create a new http client once and reuse each gather.

plugins/inputs/nats/nats.go
```diff
@@ -0,0 +1,104 @@
...
+			"cpu":               stats.CPU,
+			"mem":               stats.Mem,
+			"connections":       stats.Connections,
+			"total_connections": stats.TotalConnections,
+			"subscriptions":     stats.Subscriptions,
+			"slow_consumers":    stats.SlowConsumers,
+			"routes":            stats.Routes,
+			"remotes":           stats.Remotes,
+		},
+		map[string]string{"nats_server": n.Server},
```
danielnelson: Just call this server, this avoids stutter with measurement name.

plugins/inputs/nats/nats.go
```diff
@@ -0,0 +1,104 @@
...
+		return err
+	}
+
+	stats := new(gnatsd.Varz)
+	err = json.Unmarshal([]byte(bytes), &stats)
+	if err != nil {
+		return err
+	}
+
+	acc.AddFields("nats_varz",
```
danielnelson: What do you think about just calling this nats?  Adding varz could be useful if we anticipate other measurements but is otherwise unneeded IMO.

plugins/inputs/nats/nats.go
```diff
@@ -0,0 +1,104 @@
...
+
+	return nil
+}
+
+func (n *Nats) createHTTPClient() *http.Client {
+	timeout := n.ResponseTimeout.Duration
+	if timeout == time.Duration(0) {
+		timeout = 5 * time.Second
+	}
+	return &http.Client{
```
danielnelson: Create a new transport to avoid using the default one, all it needs set is `Proxy: http.ProxyFromEnvironment`

plugins/inputs/nats/nats.go
mjs: Ha! It was called `nats` originally but on #3186 you suggested using `nats_vars` :)

Some of the other measurements that NATS offers through other monitoring endpoints could be useful to expose at some stage (e.g detailed connection & subscription metrics) but right now it's just the high level metrics that are being exposed and there's no need to include "varz" in the name for those. 

I'm happy to change it back to just `nats`. We can add suffixes for the more detailed stuff.

plugins/inputs/nats/nats.go
mjs: Done

plugins/inputs/nats/nats.go
mjs: Done

plugins/inputs/nats/nats.go
mjs: Done

mjs:
@danielnelson All feedback addressed. PTAL.

plugins/inputs/nats/nats.go
danielnelson: :/ I guess this probably means it doesn't matter much, lets just go with what we have.

PR #3680, opened 2018-01-16
outputs.graphite: Retry sending metrics immediately after reconnect

piotr1212:
If writing to Graphite would fail the plugin would reconnect, but not retry to send the metrics until the next interval. In a situation when the connection would break before the next interval the metrics would never reach the Graphite server. Obviously this is a network issue which but Telegraf should handle these kind of situations better.

This patch retries to send the metrics one time immediately after reconnecting.

### Required for all PRs:

- [x] Associated README.md updated. <- not needed

### DIFF
--- a/plugins/outputs/graphite/graphite.go
+++ b/plugins/outputs/graphite/graphite.go
@@ -155,8 +155,22 @@ func (g *Graphite) Write(metrics []telegraf.Metric) error {
 		batch = append(batch, buf...)
 	}
 
+	err = g.send(batch)
+
+	// try to reconnect and retry to send
+	if err != nil {
+		log.Println("E! Graphite: Reconnecting and retrying: ")
+		g.Connect()
+		err = g.send(batch)
+	}
+
+	return err
+}
+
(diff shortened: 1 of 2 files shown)


danielnelson:
Can you check if your disconnects are detected by the code in `checkEOF`?  I think it might make more sense to reconnect before `Write`, so that we don't need to double send.

piotr1212:
Yes it does detect them, but I wasn't sure if I have to check all connections and then reconnect the broken ones, reconnect all connections or be happy with at least one connection. I suspect I would have to refactor connect() then as well, as it now reconnects all connections. Or add some way to mark a connection as broken so it won't try to send on it. I thought this would be safe as well, there might be situations which are not detected by checkEOF. I'll think a bit about it, if you have any suggestions let me know.

And this doesn't double send, the second send is only called when the first failed.

danielnelson:
Okay, I think we will want to refactor this code in the future so that we can reconnect individually to each server.  This way the logic can basically be:
```
for each server
  if not connected
    connect
  write
```

This is still an improvement, so I'm going to merge.

PR #3688, opened 2018-01-17
aggregate: Don't skip metrics which are in next aggretationperiod

piotr1212:
While running metrics from logparser through an aggregator I noticed some metrics were lost. Found out in some times the metrics timestamp was later than the current aggregation method. This PR queues metrics in the next aggregation method and processes them the next round. I'm not sure if this the correct way to fix this but it works for us. 

### Required for all PRs:

### DIFF
--- a/internal/models/running_aggregator.go
+++ b/internal/models/running_aggregator.go
@@ -1,8 +1,8 @@
 package models
 
 import (
-	"time"
 	"log"
+	"time"
 
 	"github.com/influxdata/telegraf"
 	"github.com/influxdata/telegraf/metric"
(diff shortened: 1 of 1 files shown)


danielnelson:
Do you know what caused the metrics to be too new?  I wonder if perhaps we are not setting the `periodEnd` correctly.

piotr1212:
I couldn't find the reason. Endperiod was looking fine.

danielnelson:
I don't want to add metric buffering to the aggregators, but I am interested in knowing why this was happening.  Perhaps the log had future timestamps due to clock skew?

piotr1212:
Found the issue. This PR makes no sense.

now is set here https://github.com/influxdata/telegraf/blob/master/agent/agent.go#L123 by the time the code reaches the if statement in in RunningAggregator.Run() on my system 600ms passed.

Issue #3701, opened 2018-01-19
Make rabbitmq_queue gathering be configurable


ildarsv:
## Feature Request

Some components can create a lot of temporary queues in RabbitMQ.
In RabbitMQ input plugin we have the `rabbitmq_queue` measurement with the `queue` tag. Gathering of the `rabbitmq_queue` metrics is not configurable in current implementation, we can only specify a list of queues to gather. In that case we might have a tons of tags for temporary queues.

### Proposal:

I propose to add `gather_queues` parameter into configuration, just to have possibility disable gathering for `rabbitmq_queue` measurement if needed.

### Current behavior:

`gather_queues` is gathered in any case.

### Desired behavior:

Disable `gather_queues` gathering.

### Use case:

I want to make my InfluxDB less stressed. For now we have 20+ new series for every temporary queue.

kerams:
I use `queues = [ "-" ]` as a workaround.

ildarsv:
Yes, I agree, but another side is `err := r.requestJSON("/api/queues", &queues)`. Telegraf makes request to RabbitMQ API every collection interval in any case.
But if you don't want gather `rabbitmq_queue` metrics, let's avoid these requests at all.

kerams:
Yeah, I've seen it timeouting on that call alone. We have almost a thousand active queues.

danielnelson:
Add for 1.6, thanks @ildarsv

PR #3702, opened 2018-01-19
Make rabbitmq_queue gathering be configurable

ildarsv:
### Required for all PRs:

PR for the #3701

### DIFF
--- a/plugins/inputs/rabbitmq/README.md
+++ b/plugins/inputs/rabbitmq/README.md
@@ -44,6 +44,11 @@ For additional details reference the [RabbitMQ Management HTTP Stats](https://cd
   ## A list of exchanges to gather as the rabbitmq_exchange measurement. If not
   ## specified, metrics for all exchanges are gathered.
   # exchanges = ["telegraf"]
+
+  ## Queues to include and exclude. Globs accepted.
+  ## Note that an empty array for both will include all queues
+  # queue_name_include = []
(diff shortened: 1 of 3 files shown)


danielnelson:
I'm somewhat adverse to adding new boolean options, as they don't scale up very well and become confusing especially if they become dependent on other options.  What I'd like to do is something similar to our other filter options and have an include and exclude list that supports globs:
```
queues_exclude = ["*"]
```

There is a good example of this in the docker input with `container_name_include` and `container_name_exclude`.

ildarsv:
@danielnelson 
Yes, It is not the best solution, but in case you have 1000+ queues (some of them can be temporary) and you don't want to gather `rabbitmq_queue` at all, you can apply `queues_exclude = ["*"]`. But the telegraf will make request to the RabbitMQ API to grab all these queues stats anyway. ~1Mb will be requested in every collecting interval, it causes timeouts from time to time and overloads network. Not only I have faced this issue, @kerams seems to have faced this issue as well (see #3701).

kerams:
@ildarsv Why not just check if `queues_exclude` is `*` and then decide whether it is necessary to gather them at all?

ildarsv:
Alright, I like this way too, let me do it

ildarsv:
@danielnelson I will appreciate if you make some comment/review for this PR when you have time.

PR #3707, opened 2018-01-21
ignore redis info when key name contain "replid"

hechen0:
when grab information from multiple redis nodes, some nodes' xx_replid
are sha1 string like '34055853fce268cfc81aadd5686e1dd79b21b5bf', but
some nodes' xx_replid are just
'0000000000000000000000000000000000000000',
when the all zero string get parsed in golang, it's a int 0, however
sha1 string is parsed as a string, so the type difference cause a type
conflict, which lead to drop of point when write to influxdb.
whatsmore the xx_replid section is useless, so drop this section is
harmless.

### Required for all PRs:

### DIFF
--- a/plugins/inputs/redis/redis.go
+++ b/plugins/inputs/redis/redis.go
@@ -207,6 +207,11 @@ func gatherInfoOutput(
 			}
 		}
 
+		// drop section when name contain "replid"
+		if strings.Contains(section, "replid"){
+			continue
+		}
+
 		if name == "mem_allocator" {
 			continue
 		}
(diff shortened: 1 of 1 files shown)


danielnelson:
@hechen0 Can you update the unittests or, if you prefer, send me the output of the `info` command from your server?

hechen0:
exp 1.
# Replication
role:master
master_replid:ceb8cecf390c08dbbf84c00ebc038eff1ffcd438
master_replid2:e65e4d3200b6534f4d738c62f681efb7b59d97b3

exp 2.
# Replication
role:master
master_replid:f58982804569417a42891e47ca5c8de99b82a3fa
master_replid2:0000000000000000000000000000000000000000

danielnelson:
I added unit tests and changed to skip only fields starting with `master_replid` in #3725

Issue #3710, opened 2018-01-22
Telegraf stops publishing metrics via wavefront output, all plugins take too long to collect - no meaningful error message


pberlowski:
## Bug report
After a certain amount of time (usually 5-10 minutes) under a load-test, the telegraf agent stops sending metrics onward via wavefront output. In addition to that, the log fills up with "took to long to collect" messages from all plugins.

No relevant error message is available in the logs.

This issue seems similar to #3629 however no aggregation plugin is in use.
There is a suspicion that the output may be too slow, even though the buffer doesn't seem to overfill and drop metrics. 

### Load test description
Measurements are submitted via http_listener from a load-generator host that replays a snapshot of data at selectable throughput. At 5000pps (fields), the error happens within minutes of starting the agent.

### Relevant telegraf.conf:

[http-proxy.conf.txt](https://github.com/influxdata/telegraf/files/1652365/http-proxy.conf.txt)
[telegraf.conf.txt](https://github.com/influxdata/telegraf/files/1652366/telegraf.conf.txt)

### System info:
Telegraf version: 1.4.5
OS: Centos 7.3

### Expected behavior:
Telegraf logs a meaningful error to describe a reason for overload and keeps buffering metrics.

### Actual behavior:
Telegraf seizes operations silently and then starts logging failures to collect on time.

### Additional info:
Stacktrace:
[telegraf.stacktrace.txt](https://github.com/influxdata/telegraf/files/1652360/telegraf.stacktrace.txt)

Hardware and software metrics screenshots for the failure window:
![telegraf failure - hardware](https://user-images.githubusercontent.com/7792100/35227906-48f1c030-ff87-11e7-9a3c-060871f9dc7a.png)
![telegraf failure - metrics](https://user-images.githubusercontent.com/7792100/35227908-4925dbae-ff87-11e7-9735-811a5ce22f73.png)
![telegraf failure - network](https://user-images.githubusercontent.com/7792100/35227910-494cf982-ff87-11e7-892f-b86454f8a6b8.png)

danielnelson:
This sounds like the wavefront plugin is waiting on a network response without the proper timeout.

@puckpuck Can you take a look at this one?  It is important that outputs return from the `Write()` function fairly quickly to avoid backing up the inputs, usually no more than 5s.

puckpuck:
@pberlowski 2 things

[...]

Issue #2919, opened 2017-06-13
Outputs block inputs when batch size is reached


rconn01:
## Bug report
If one of the output plugins become slow it can cause back pressure all the way up through the input plugins and start to block the input plugins from running. 

The reason for this is the channels between the input and output plugins only have a buffer of 100. This should be ok since most of the operations are in memory. The problem ends up being [here](https://github.com/influxdata/telegraf/blob/f93615672b02a41d9bc867bd92bf31c1d777989b/internal/models/running_output.go#L113) where the buffer gets written if the batch size is full. This ends up happening on the same thread that reads from the `outMetricC` channel in the [agent](https://github.com/influxdata/telegraf/blob/f93615672b02a41d9bc867bd92bf31c1d777989b/agent/agent.go#L281)
### Relevant telegraf.conf:
```
# Configuration for telegraf agent
[agent]
  interval = "5s"
  round_interval = true
  metric_batch_size = 1000
  metric_buffer_limit = 250000
  collection_jitter = "0s"
  flush_interval = "10s"
  flush_jitter = "5s"
  precision = ""
  debug = true
  quiet = false
  logfile = ""
  hostname = ""
  omit_hostname = false


###############################################################################
#                            OUTPUT PLUGINS                                   #
###############################################################################

## Configuration for influxdb server to send metrics to
[[outputs.influxdb]]
  urls = ["http://localhost:8086"] # required
  database = "telegraf" # required
  retention_policy = ""
  write_consistency = "any"
  timeout = "5s"


## Configuration for influxdb server to send metrics to
[[outputs.influxdb]]
  urls = ["http://localhost:8080"] # required
  database = "telegraf" # required
  retention_policy = ""
  write_consistency = "any"
  timeout = "10s"



###############################################################################
#                            INPUT PLUGINS                                    #
###############################################################################

# Read metrics about cpu usage
[[inputs.cpu]]
  percpu = true
  totalcpu = true
  collect_cpu_time = false

# Read metrics about disk usage by mount point
[[inputs.disk]]
  ignore_fs = ["tmpfs", "devtmpfs"]

# Read metrics about disk IO by device
[[inputs.diskio]]

# Get kernel statistics from /proc/stat
[[inputs.kernel]]

# Read metrics about memory usage
[[inputs.mem]]

# Read metrics about system load & uptime
[[inputs.system]]

[[inputs.internal]]


###############################################################################
#                            SERVICE INPUT PLUGINS                            #
###############################################################################

[...]

PR #3711, opened 2018-01-22
Update wavefront.go

puckpuck:
### Required for all PRs:

This fixes a potential issue where the output plugin could in some instances take more than 5 seconds causing all other input plugins to block.  

See #3710 and #2919 

### DIFF
--- a/plugins/outputs/wavefront/wavefront.go
+++ b/plugins/outputs/wavefront/wavefront.go
@@ -122,7 +122,7 @@ func (w *Wavefront) Write(metrics []telegraf.Metric) error {
 		return fmt.Errorf("Wavefront: TCP connect fail %s", err.Error())
 	}
 	defer connection.Close()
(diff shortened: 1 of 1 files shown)


danielnelson:
@puckpuck It would be even more robust if this output used a timeout capable replacement for Dial and ip resolution.  You might just want to merge the two functions into net.DialTimeout, you should be able to report any loopup error from that function.

puckpuck:
@danielnelson We have plans to do significant enhancements to how the the output plugin communications with the Wavefront proxy service.  We are adding HTTP in addition to the existing telnet/socket communication channel.  This would be planned in the near future, but we aren't quite ready yet.  When that happens I will also add more robustness around the entire communication channel including creation/testing of the network address.  Do you have a date already planned for the next Telegraf version post 1.6?

danielnelson:
1.7 will likely be in April

PR #3718, opened 2018-01-24
use telegraf.Accumulator

wi-cuckoo:
### Required for all PRs:
Maybe return telegraf.Accumulator should be more confortable


danielnelson:
Can you fix the unit tests?

wi-cuckoo:
@danielnelson the accumulator_test.go need to use the channel directly, so I'm not sure that modify the test file. Just I think should return an interface, not an unexported struct

danielnelson:
I added you changes and addressed the unittests in efb9d5b4cb5805ac21128fc975dd1bbef4fb1d3c

PR #3720, opened 2018-01-24
postgresql_extensible / extend README explaining the tagvalue Option

MaZderMind:
I had some issues understanding what the tagvalue-Option in the postgresql_extensible-Plugin does, especially as the README suggest that it is just a list of Tags added literally to the result-Row:
> the tagvalue field is used to define custom tags (separated by comas)

Neither does the README have an example on how the tagvalue can be used in conjunction with `GROUP BY` to add multiple result-rows to one measurement.

This PR extends the README wiith both, a more detailed explanation and such an example.

### Required for all PRs:


danielnelson:
Thanks!

PR #3722, opened 2018-01-25
openldap: Include more metrics, change up field names

acobaugh:
I noticed that this plugin wasn't including a bunch of metrics, specifically the monitoredObject entries. This PR includes those entries with numerical values for monitoredInfo.

It also changes up the field names so they sort more naturally. I know this was discussed on the original PR, and in retrospect and after using this plugin, reversing the names based on their distinguished names will make them easier to work with. This will be a breaking change for anyone already using this plugin, but I believe it will be worth it.

(please let me know if we'd rather make this two PRs so the additional metrics can be included in a release sooner on account of the breaking change)

### Required for all PRs:

### DIFF
--- a/plugins/inputs/openldap/README.md
+++ b/plugins/inputs/openldap/README.md
@@ -25,6 +25,10 @@ To use this plugin you must enable the [monitoring](https://www.openldap.org/dev
   # dn/password to bind with. If bind_dn is empty, an anonymous bind is performed.
   bind_dn = ""
   bind_password = ""
+  
+  # reverse metric names so they sort more naturally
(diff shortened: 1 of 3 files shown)


danielnelson:
We can add the new fields no problem, but on the reordered fields we need to preserve compatibility.  
What we can do if you like is add a new option that when set turns on the new ordering, it can be enabled when generating a new config but if not set it should default off.  In the comment we can recommend using the reverse order.

acobaugh:
Alright, added a `reverse_metric_names` option that defaults to false, but is set to true when generating a config.

PR #3726, opened 2018-01-27
ping plugin: Add cases to interface parameter

phlipse:
Fixes issue https://github.com/influxdata/telegraf/issues/3721.

Linux supports defining the source address or source interface with paramter **I**.
FreeBSD and Darwin do only support specifying the source address and also use a different parameter **S** to set it: see [manpage](https://www.freebsd.org/cgi/man.cgi?ping(8)). 

This commit adds cases to support pings from a defined source address on other operating systems than Linux. Source interface is linux only.

### Required for all PRs:


phlipse:
Maybe the variable should be called **Source** instead of **Interface**? I will let take this decision to an official maintainer.

PR #3738, opened 2018-01-31
Processors math

vi4eslav:
For partial preprocessing of a large volume of metrics, I added mathematical functions.

### Required for all PRs:


vi4eslav:
is there any news?

plugins/processors/math/README.md
```diff
@@ -0,0 +1,82 @@
+# Math Processor Plugin
+
+The math processor plugin applies the function to all or selected fields.
+
+### Configuration:
+
+```toml
+## Example config that processe all fields of the metric.
+# [[processor.math]]
```
ssoroka, line 9: should be `[[processors.math]]`, here and other places it's mentioned.

plugins/processors/math/math.go
```diff
@@ -0,0 +1,145 @@
...
+	return sampleConfig
+}
+
+func (p *Math) Description() string {
+	return "Math metrics that pass through this filter."
+}
+
+func (p *Math) Apply(in ...telegraf.Metric) []telegraf.Metric {
+
+	if _, ok := mathFunctions[p.Func]; ok {
```
ssoroka, line 100: invert this to do an error check and flatten the rest of the function. Save the mathFunc reference.

plugins/processors/math/math.go
```diff
@@ -0,0 +1,145 @@
...
+		for _, metric := range in {
+			if metric.Name() == p.Metric {
```
ssoroka, line 102: FYI you can invert the check to bring the nesting and complexity down in cases like this:
```
if metric.Name() != p.Metric {
  continue
}
// now you don't need the nesting.
```
However, you probably want to drop this completely in favour of the built-in `namepass` feature, as this is more flexible, and it's handled for you already. see [metric filtering](https://github.com/influxdata/telegraf/blob/master/docs/CONFIGURATION.md#metric-filtering)

plugins/processors/math/math.go
```diff
@@ -0,0 +1,145 @@
...
+
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/plugins/processors"
+)
+
+// Math function congfig
+type Math struct {
+	Func   string   `toml:"function"`
+	Metric string   `toml:"measurement_name"`
+	Fields []string `toml:"fields"`
```
ssoroka, line 16: might want to also take a float64 argument variable, which would let you support the entire _math_ package, as well as addition, subtraction, multiplication, and division

plugins/processors/math/math.go
```diff
@@ -0,0 +1,145 @@
...
+func (p *Math) Apply(in ...telegraf.Metric) []telegraf.Metric {
+
+	if _, ok := mathFunctions[p.Func]; ok {
+		for _, metric := range in {
+			if metric.Name() == p.Metric {
+				mathFunc := mathFunctions[p.Func]
+				if len(p.Fields) > 0 {
+					for _, name := range p.Fields {
+						if srcField, ok := metric.Fields()[name]; ok {
+							field, ok := srcField.(float64)
```
ssoroka, line 107: can this be other types?

plugins/processors/math/math.go
```diff
@@ -0,0 +1,145 @@
...
+
+func (p *Math) Apply(in ...telegraf.Metric) []telegraf.Metric {
+
+	if _, ok := mathFunctions[p.Func]; ok {
+		for _, metric := range in {
+			if metric.Name() == p.Metric {
+				mathFunc := mathFunctions[p.Func]
+				if len(p.Fields) > 0 {
+					for _, name := range p.Fields {
+						if srcField, ok := metric.Fields()[name]; ok {
```
ssoroka, line 106: add a function that returns true when the metric has the right name and fields, then simplify this code to remove all the "is this the right metric" checks.

plugins/processors/math/math.go
```diff
@@ -0,0 +1,145 @@
...
+							field, ok := srcField.(float64)
+							if ok {
+								metric.AddField(
+									fmt.Sprintf("%s_%s", name, p.Func),
+									mathFunc(field))
+							}
+						} else {
+							log.Printf(
+								"E! Metric %s doesn't have field %s",
```
ssoroka, line 115: remove this. not useful. Most metrics won't have the right fields, and we just ignore it and continue on.

plugins/processors/math/math.go
```diff
@@ -0,0 +1,145 @@
...
+								p.Func,
+								metric.Name())
+						}
+					}
+				} else {
+					for name, srcField := range metric.Fields() {
```
ssoroka, line 121: remove this. if no fields are specified the function should do nothing.

plugins/processors/math/math.go
```diff
@@ -0,0 +1,145 @@
...
+							metric.AddField(
+								fmt.Sprintf("%s_%s", name, p.Func),
+								mathFunc(field))
+						}
+					}
+				}
+			}
+		}
+	} else {
+		log.Printf(
```
ssoroka, line 133: this should be at the top and return. Avoid `else`, especially when buried many lines down.

ssoroka:
Discussed this with the team and we're going to close this in favor of the [Starlark processor plugin](https://github.com/influxdata/telegraf/tree/master/plugins/processors/starlark), which is already merged, and can handle more complex calculations. 
https://github.com/influxdata/telegraf/tree/master/plugins/processors/starlark

Issue #3739, opened 2018-02-01
Basic auth information are displayed in url tag with prometheus input


athoune:
When I query a prometheus probe with basic auth, login and password are displayed in the url tag.

Replacing password with '***' should be more secure.

danielnelson:
In 1.5.3 the userinfo part of the URL will be removed from the `url` tag.

PR #3743, opened 2018-02-02
prometheus input: Remove credentials from URL before setting url tag

phlipse:
Remove user data from URL before setting URL tag to prevent information disclosure.
See https://github.com/influxdata/telegraf/issues/3739.

Also run gofmt and turn variables with name *url* to uppercase to prevent overwrite of *net/url* package. Now it is more clear wheter the variable or package is used.

### Required for all PRs:

### DIFF
--- a/plugins/inputs/prometheus/README.md
+++ b/plugins/inputs/prometheus/README.md
@@ -67,7 +67,7 @@ Measurement names are based on the Metric Family and tags are created for each
 label.  The value is added to a field named based on the metric type.
 
 All metrics receive the `url` tag indicating the related URL specified in the
-Telegraf configuration. Existing username and password for basic authentication will be removed before. If using Kubernetes service discovery the `address`
+Telegraf configuration. If using Kubernetes service discovery the `address`
 tag is also added indicating the discovered ip address.
 
 ### Example Output:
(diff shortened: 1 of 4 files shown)


plugins/inputs/prometheus/README.md
```diff
@@ -67,7 +67,7 @@ Measurement names are based on the Metric Family and tags are created for each
 label.  The value is added to a field named based on the metric type.
 
 All metrics receive the `url` tag indicating the related URL specified in the
-Telegraf configuration.  If using Kubernetes service discovery the `address`
+Telegraf configuration. Existing username and password for basic authentication will be removed before. If using Kubernetes service discovery the `address`
```
danielnelson: Let's remove this, we shouldn't have to say this because doing it the other way would be bad.  BTW, using the userinfo section of the url is not [basic auth](https://tools.ietf.org/html/rfc7617).

phlipse: Oh sorry, you are right, this is not basic auth! I red it in the referenced issue and wasn't think about it when writing the commit message and Readme.

I'll remove the sentence in Readme.md with next commit.

phlipse:
I don't know what you have done but it fixed the failing test *TestTailBadLine*. :)

PR #3745, opened 2018-02-02
Makes the unbound plugin reset statistics on read

rhysrhaven:
Per the unbound documentation, the flag `statistics-cumulative: yes|no`
determines whether or not unbound clears its stat counters on a read
via the `stats` call to unbound-control. Since the original plugin is
hardcoded to use the `stats_noreset` command, it makes the configuration
option in the unbound.conf meaningless and provides no way to change it
in telegraf.

This changes to the plugin to calling the `stats` call to fix this problem.

### Required for all PRs:

### DIFF
--- a/plugins/inputs/unbound/README.md
+++ b/plugins/inputs/unbound/README.md
@@ -26,9 +26,11 @@ This is the full list of stats provided by unbound-control and potentially colle
 depending of your unbound configuration. Histogram related statistics will never be collected,
 extended statistics can also be imported ("extended-statistics: yes" in unbound configuration).
 In the output, the dots in the unbound-control stat name are replaced by underscores(see
-https://www.unbound.net/documentation/unbound-control.html for details).
+[https://www.unbound.net/documentation/unbound-control.html]() for details).
 
 - unbound
+
+```
     thread0_num_queries
     thread0_num_cachehits
     thread0_num_cachemiss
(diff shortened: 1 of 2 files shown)


danielnelson:
Usually cumulative stats are preferred, since they have better aggregation properties and handle errors and failure better.  You can use the non_negative_difference and non_negative_derivative functions to view the changes, here is an example using the cpu input:
```
select non_negative_derivative(last(time_idle), 1m) from cpu where time > now() - 5m group by time(1m)
```

rhysrhaven:
@danielnelson I'm aware of how to deal with cumulative stats, but fact remains that there exists specific behavior flags in unbound such that user decides this. setting `statistics-cumulative: yes` in unbound makes the `stats` call to operate exactly as the `stats_noreset` call. 

I'd like if it was up to the user, which this change allows. 

Separately do you have any docs/articles/explanations on how cumulative stats provide "better aggregation properties and handle errors and failure better" ? I've long held the opposite opinion, that a sampling style of getting stats was better (like are used as example metrics in the documentation of InfluxDB) but I've never had a discussion with anyone who knew. My team was all like "I heard that cumulative was better but ¯\_(ツ)_/¯". We're legit curious after having cumulative stats cause a fair few annoyances over the years.

danielnelson:
I prefer to avoid extra options if it's not required, especially since I would recommend using the cumulative version.  It is also possible that we won't always use the unbound-control utility to get the stats.

My aggregation comment was really more about multi-unit metrics such as messages/second vs a raw metrics such as message_count, so I don't think it really applies here.

The advantages of cumulative metrics I can think of are:

- If Telegraf is down or misses collections you can determine how much was missed.
- If the value is reset you need to know how long it has been since your last collection, while with absolute it doesn't matter.  This also means you don't need to worry about out of band calls to unbound-control or Telegraf being restarted/reloaded.
- Certain functions are less expensive, you can use first/last for min/max for instance, of course you don't need to mess with non_negative_derivative.

Downsides of cumulative are:

- Eventually the counter will reset
- Can capture larger changes due to int rollover

@desa Can you think of any other advantages/disadvantages I missed?

PR #3753, opened 2018-02-05
New processor Tagplugin

mortenhillbom:
This plugin makes it possible to add a new tag to metrics based on the value of an existing tag.

This can be useful for example in cases where nodes running telegraf are configured differently, and their cpus or network interfaces have different roles on each individual node.

Tagging them at this level can thus make it easier to group them by category or role later.

### Required for all PRs:

### DIFF
--- a/plugins/processors/all/all.go
+++ b/plugins/processors/all/all.go
@@ -2,4 +2,5 @@ package all
 
 import (
 	_ "github.com/influxdata/telegraf/plugins/processors/printer"
+	_ "github.com/influxdata/telegraf/plugins/processors/tagplugin"
 )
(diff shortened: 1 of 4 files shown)


danielnelson:
@mortenhillbom Sorry, I lost track of this PR.  I think you can do this now with the [override](https://github.com/influxdata/telegraf/tree/release-1.6/plugins/processors/override) plugin in 1.6.  You can do this by combining the [measurement filtering](https://github.com/influxdata/telegraf/blob/release-1.6/docs/CONFIGURATION.md#measurement-filtering) with the tags option:

This contrived example adds a tag to cpu2 of the cpu input:
```toml
[[processors.override]]
  namepass = "cpu"
  [processors.override.tagpass]
    cpu = ["cpu2"]

  [processors.override.tags]
    additional_tag = "tag_value"
```

PR #2785, opened 2017-05-11
Genericsqlplugin

lucadistefano:
The plugin executes simple queries or query scripts on multiple servers.
It permits to select the tags and the fields to export, if is needed fields can be forced to a choosen datatype. 
Supported drivers are  go-mssqldb (sqlserver) , oci8 ora.v4 (Oracle), mysql (MySQL), pq (Postgres) 

it is highly flexible and can do both type of counters gathering: horizontal and vertical .
horizontal: where each counter is on different columns of a row and the counter name is the name of the column
vertical : where each counter is on different row and the counter name is value of a specified column

### Required for all PRs:

- [ ] CHANGELOG.md updated (we recommend not updating this until the PR has been approved by a maintainer)
- [x] Sign [CLA](https://influxdata.com/community/cla/) (if not already signed)
- [x] README.md updated (if adding a new plugin)

PR #3758, opened 2018-02-06
New Plugin for Goldilocks Cluster DBMS 

ckh0618:
### Required for all PRs:


ckh0618:
It requires unixodbc library for linux enviroments.
 I worry that this change may break builds and need an extra dependency.

danielnelson:
I do want to eventually support loading C libraries, but there remains quite a bit of work in our build system to do it properly.  If it were possible for the driver to be loaded at runtime maybe it could be done like in #2785, this would be a faster path to inclusion.

ckh0618:
Well I couldn't find the way how to fix this problem.  I think it would be better not to include this changes to master because there is no way to eliminate external dependencies.  

If anyone who needs this feature, he/she has to build it by him/herself.  I wrote some instructions to  build telegraf with this feature manually. ( https://github.com/ckh0618/telegraf_goldilocks )

PR #3762, opened 2018-02-07
Add Derivative Aggregator Plugin

KarstenSchnitter:
Calculate metrics derivatives based on timestamp or other fields.

### Required for all PRs:


srebhan:
@KarstenSchnitter are you still interested in getting this PR merged? If so, please make sure you have a CLA signed and indicate this in the PR-message above.

KarstenSchnitter:
I would appreciate it very much if this PR would become part of telegraf. I have been running a fork of telegraf with this plugin in production ever since I created this PR. Signed CLA was provided for #3772 and #3773 as well, which were both merged in 2018.

plugins/aggregators/derivative/README.md
```diff
@@ -0,0 +1,72 @@
...
+### Configuration
+
+```toml
+[[aggregators.derivative]]
+  ## Specific Derivative Aggregator Arguments:
+  ## Configure a custom derivation variable. Timestamp is used if none is given.
+  # variable = "fieldname_of_derivation_vaiable"
+  ## Infix to separate field name from derivation variable.
+  infix = "_by_"  # default: _by_
+  ## Roll-Over last measurement to first measuremen of next period
```
srebhan, line 50:
```suggestion
  ## Roll-Over last measurement to first measurement of next period
```

plugins/aggregators/derivative/README.md
```diff
@@ -0,0 +1,72 @@
...
+cope with multiple "quiet" periods, where no new measurement is pushed to the
+aggregator. The roll-over will take place at most `max_roll_over`times.
+
+### Configuration
+
+```toml
+[[aggregators.derivative]]
+  ## Specific Derivative Aggregator Arguments:
+  ## Configure a custom derivation variable. Timestamp is used if none is given.
+  # variable = "fieldname_of_derivation_vaiable"
```
srebhan, line 47: This should be the default value
```suggestion
  # variable = "time"
```

plugins/aggregators/derivative/README.md
```diff
@@ -0,0 +1,72 @@
...
+  infix = "_by_"  # default: _by_
+  ## Roll-Over last measurement to first measuremen of next period
+  max_roll_over = 10 # default: 10
+
+  ## General Aggregator Arguments:
+  ## calculate derivative every 30 seconds
+  period = "30s"
+  ## fields for which the derivative should be calculated
+  ## Important: The derivation variable must be contained in that listed
+  fieldpass = ["field1", "field2", "variable"]
```
srebhan, line 58: This collides with the generic telegraf `fieldpass` option. Please drop this option here.

plugins/aggregators/derivative/README.md
```diff
@@ -0,0 +1,72 @@
...
+[[aggregators.derivative]]
+  ## Specific Derivative Aggregator Arguments:
+  ## Configure a custom derivation variable. Timestamp is used if none is given.
+  # variable = "fieldname_of_derivation_vaiable"
+  ## Infix to separate field name from derivation variable.
+  infix = "_by_"  # default: _by_
+  ## Roll-Over last measurement to first measuremen of next period
+  max_roll_over = 10 # default: 10
+
+  ## General Aggregator Arguments:
```
srebhan, line 53: Another general argument is `drop_original`. Please add this here.

plugins/aggregators/derivative/README.md
```diff
@@ -0,0 +1,72 @@
...
+  ## calculate derivative every 30 seconds
+  period = "30s"
+  ## fields for which the derivative should be calculated
+  ## Important: The derivation variable must be contained in that listed
+  fieldpass = ["field1", "field2", "variable"]
+```
+
+### Tags:
+No tags are applied by this aggregator.
```
srebhan, line 62: But I guess tags are passed through. Maybe you should mention that here.

plugins/aggregators/derivative/derivative.go
```diff
@@ -0,0 +1,209 @@
...
+	"time"
+
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/plugins/aggregators"
+)
+
+type Derivative struct {
+	Variable    string
+	Infix       string
+	MaxRollOver uint
```
srebhan, line 14: Please add the toml specifications here
```suggestion
	Variable    string   `toml:"variable"`
	Infix       string      `toml:"infix"`
	MaxRollOver uint `toml:"max_roll_over"`
```

plugins/aggregators/derivative/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+```toml
+[[aggregators.derivative]]
+  ## Specific Derivative Aggregator Arguments:
+  ## Configure a custom derivation variable. Timestamp is used if none is given.
+  # variable = "fieldname_of_derivation_vaiable"
+  ## Infix to separate field name from derivation variable.
+  infix = "_by_"  # default: _by_
+  ## Roll-Over last measurement to first measuremen of next period
+  max_roll_over = 10 # default: 10
```
srebhan, line 51: Comment this line if 10 is the default.

plugins/aggregators/derivative/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+### Configuration
+
+```toml
+[[aggregators.derivative]]
+  ## Specific Derivative Aggregator Arguments:
+  ## Configure a custom derivation variable. Timestamp is used if none is given.
+  # variable = "fieldname_of_derivation_vaiable"
+  ## Infix to separate field name from derivation variable.
+  infix = "_by_"  # default: _by_
```
srebhan, line 49: Do we really need this or can this be a `_derivative` suffix. or a `d` prefix or something?

plugins/aggregators/derivative/derivative.go
```diff
@@ -0,0 +1,209 @@
...
+  ## This Aggregator will estimate a derivative for each field, which is
+	## contained in both the first and last metric of the aggregation interval.
+  ## Without further configuration the derivative will be calculated with
+	## respect to the time difference between these two measurements in seconds.
+	## The formula applied is for every field:
+	##
+	##               value_last - value_first
+	## derivative = --------------------------
+  ##              time_difference_in_seconds
+  ##
```
srebhan, line 48: There seem to be formatting errors...

plugins/aggregators/derivative/derivative.go
```diff
@@ -0,0 +1,209 @@
...
+func (d *Derivative) Description() string {
+	return "Calculates a derivative for every field."
+}
+
+func (d *Derivative) Add(in telegraf.Metric) {
+	id := in.HashID()
+	if current, ok := d.cache[id]; !ok {
+		// hit an uncached metric, create caches for first time:
+		d.cache[id] = newAggregate(in)
+	} else {
```
srebhan, line 87: You might want to `return` here and omit the `else`.

plugins/aggregators/derivative/derivative.go
```diff
@@ -0,0 +1,209 @@
...
+		if current.first.time.After(in.Time()) {
+			current.first = newEvent(in)
+			current.rollOver = 0
+		}
+		if current.first.time.Equal(in.Time()) {
+			current.first.fields = upsertConvertedFields(in.Fields(), current.first.fields)
+			current.rollOver = 0
+		}
```
srebhan, line 95: I don't understand why you handle that differently... You always want to replace the field values or am I wrong here?

plugins/aggregators/derivative/derivative.go
```diff
@@ -0,0 +1,209 @@
...
+}
+
+func upsertConvertedFields(source map[string]interface{}, target map[string]float64) map[string]float64 {
+	for k, v := range source {
+		if value, ok := convert(v); ok {
+			target[k] = value
+		}
+	}
+	return target
+}
```
srebhan, line 138: I think it makes no sense to overwrite `target` and return it here as you probably touch every entry anyway. So either you just overwrite target without the return or you use a new map.

plugins/aggregators/derivative/derivative.go
```diff
@@ -0,0 +1,209 @@
...
+	case int64:
+		return float64(v), true
+	default:
+		return 0, false
+	}
+}
+
+func (d *Derivative) variableFieldName() string {
+	return strings.TrimSpace(d.Variable)
+}
```
srebhan, line 153: This should be done once in an `Init()` function. See `basicstats` for example.

plugins/aggregators/derivative/derivative.go
```diff
@@ -0,0 +1,209 @@
...
+func upsertConvertedFields(source map[string]interface{}, target map[string]float64) map[string]float64 {
+	for k, v := range source {
+		if value, ok := convert(v); ok {
+			target[k] = value
+		}
+	}
+	return target
+}
+
+func convert(in interface{}) (float64, bool) {
```
srebhan, line 140: Maybe also check `uint64`...

plugins/aggregators/derivative/derivative.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+}
+
+func (d *Derivative) variableFieldName() string {
+	return strings.TrimSpace(d.Variable)
+}
+
+func (d *Derivative) Push(acc telegraf.Accumulator) {
+	for _, aggregate := range d.cache {
+		if aggregate.first != aggregate.last {
```
srebhan, line 157: Save some indention levels...
```suggestion
		if aggregate.first == aggregate.last {
			continue
		}
```

plugins/aggregators/derivative/derivative.go
```diff
@@ -0,0 +1,209 @@
...
+			if denominator := d.calculateDenominator(aggregate); denominator != 0 {
```
srebhan, line 158: See above. Maybe issue a debug/info message here that you only got zero difference for the denominator!

plugins/aggregators/derivative/derivative.go
```diff
@@ -0,0 +1,209 @@
...
+				}
+				acc.AddFields(aggregate.name, derivatives, aggregate.tags)
+			}
+		}
+	}
+}
+
+func (d *Derivative) calculateDenominator(aggregate aggregate) float64 {
+	if variable, present := d.derivationVariableName(); present {
+		return aggregate.last.fields[variable] - aggregate.first.fields[variable]
```
srebhan, line 173: You need to make sure `variable` is actually contained in the fields!

plugins/aggregators/derivative/derivative.go
```diff
@@ -0,0 +1,209 @@
...
+	return strings.TrimSpace(d.Variable)
+}
+
+func (d *Derivative) Push(acc telegraf.Accumulator) {
+	for _, aggregate := range d.cache {
+		if aggregate.first != aggregate.last {
+			if denominator := d.calculateDenominator(aggregate); denominator != 0 {
+				derivatives := make(map[string]interface{})
+				for key, start := range aggregate.first.fields {
+					if end, ok := aggregate.last.fields[key]; key != d.variableFieldName() && ok {
```
srebhan, line 161: Split this in two ifs starting with skipping the denominator field.

plugins/aggregators/derivative/derivative.go
```diff
@@ -0,0 +1,209 @@
...
+func (d *Derivative) derivativeFieldName(field string) string {
+	if param := d.variableFieldName(); len(param) != 0 {
+		return field + d.trimmedInfix() + param
+	}
+	return field + d.trimmedInfix() + "seconds"
+}
+
+func (d *Derivative) trimmedInfix() string {
+	return strings.TrimSpace(d.Infix)
+}
```
srebhan, line 191: Do this in `Init()` once instead of calling this over and over in every push...

srebhan (commented):
There are only minor comments in the code. Please make sure that the information in README.md and `sampleConfig` match.
Furthermore, please rebase your code to retrigger CI.

KarstenSchnitter (commented):
Thanks for the review. I will work on the issues. Can you please give some guidance on the `metric.SeriesGrouper` comment, that I made?

plugins/aggregators/derivative/README.md
KarstenSchnitter: `time` is not the default variable. If it was the default, the aggregator would look for a field named `time`. But by default it uses the timestamp of the metric and not another field.

plugins/aggregators/derivative/README.md
KarstenSchnitter: These comments were included to point out the necessity to include the derivation variable in the `fieldpass` argument. If it was not contained, it would not be accessible to the aggregator. The calculation could not be done in that case. How should we go about that?

plugins/aggregators/derivative/README.md
KarstenSchnitter: I can do this for `infix` and `max_roll_over`.  What would I do for `period?

plugins/aggregators/derivative/README.md
KarstenSchnitter: The plugin can calculate the derivative of one field with respect to another field, e.g. disk io `read_time` by `read_count`. Using the infix allows the resulting field to be named using both field names by concatenating them with the infix. We can use another naming scheme. Would you prefer a simple suffix instead? It would mean more configuration effort for each use of the plugin, since the user needs to specify the result name more often.

plugins/aggregators/derivative/README.md
srebhan: Fair point. Maybe leave the string empty then. The documentation says the commented option should contain the default value.

plugins/aggregators/derivative/README.md
srebhan: Well then please add a big fat note somewhere, otherwise it is easy for the reader to confuse `fieldspass` with a plugin-option. IIRC, you also issue an error in the plugin if this is not the case.

So my solution would be to add a **bold** note to the readme stating that if you use custom divisor variables those **have to be present** in case of field filtering. If you also want to put it to the config add `The specified variable field has to be present even is case of field filtering (fieldpass/fielddrop) or the plugin will fail!` to the description of the `variable` option.

plugins/aggregators/derivative/README.md
srebhan: I think this is a general plugin option and has to be set... Not exactly sure though... As all other aggregators have this un-commented I would also stick with this.

plugins/aggregators/derivative/README.md
srebhan: To be honest I'm not sure what the best way is. I understood your intention and it is valid. IMO the most common use case is to compute the derivation by a time (even in your case you probably want `d_read_count / dread_time` I guess) and then `dread_count` or `read_count_derivative` is maybe easier to understand for the guy examining the database. But on the other hand this is probably a matter of taste...

So please go ahead with whatever is best in your opinion and mark this as resolved.

plugins/aggregators/derivative/derivative.go
srebhan: I see. Now imagine you have two such metrics leading to the same ID having the same timestamp but two field-sets, say the first on has `in` the second one `out`. Further assume a first `period` where those metric arrived you have a metric with `in, out` in `cache`, nice.
Now after the roll-over for some reason only the first metric arrived within `period` with a timestamp *earlier* than `first`. This time you will keep the replace `first with a metric having **only** the `in` field and you drop the `out` field... Is this really what you are intending?

srebhan: Furthermore, while writing this up I recognized another (potential) issue... Imaging the metrics that enter the plugin are strictly ordered by time (e.g. slow query intervals with only one plugin). In these cases the newly arriving metrics are **always later** than `first` even after a roll-over. In this scenario `first` stays the same until `MaxRollover` is reached!!! I guess you need to additionally check if the newly arrived metric is the *first* in the `period` and replace `first` in this case. You can use `current.rollOver > 0` to distinguish the two cases IMO.

srebhan:
> Thanks for the review. I will work on the issues. Can you please give some guidance on the `metric.SeriesGrouper` comment, that I made?

I would always only *update* first in this case and never replace the metric. But I'm not sure if this is ok in your use-case as you might keep very old fields this way and to circumvent that problem you would need another structure with a per-field roll-over counter.

So either you change to a per-metric-and-field `cache` **or** we stick with the current approach and describe the side-effects (see my in-code-comments to this) nicely in the readme... You decide! :-)

plugins/aggregators/derivative/README.md
KarstenSchnitter: How about introducing a new config parameter `fields` just like in histogram or value_counter? That would remove the side-effect on `fieldpass` and ease the documentation.

plugins/aggregators/derivative/derivative.go
KarstenSchnitter: To the question on dropping fields when replacing `first` event: There are two cases: derivatives by timestamp (default) and derivatives by variables (another field). In both cases receiving the earlier metric with only `in` can still be followed by a metric adding `out` for this earlier timestamp. I cannot know this at that time. I see the following implications:

**derivatives by timestamp**
In this case, the correctness of the calculation depends on the field values at the timestamp used in the denominator. If I replace the `first` event by an earlier event, I can only use the fields present for that timestamp and hope there will be an additional event adding the missing metrics. In you example, I would hope that there would be an update for `out` incoming. Otherwise I would rather not to emit a value for `out` than to emit a false one.

**derivatives by variables (other fields)**
Here it is essential, that the field for which the derivative is calculated and the variable in the denominator are correlated to the same event. Otherwise the calculation result would be wrong. Going with your example, calculus give us the following formula:

```
 d(in)    d(in) / dt
------ = -----------
d(out)   d(out) / dt
```

This relation is once again only true, if `in` and `out` are taken from the same timestamp. Once again I went for not emitting a value rather that a wrong value.

So yes, in both cases dropping the field was intentional. Did I miss something in this?

KarstenSchnitter: Your scenario with the the slow incoming metrics is a flaw in the `cache` clean-up. Elements are only removed from the `cache` when `MaxRollOver` is reached. Naively I would think, that it should also be removed, when a derivative is emitted in `Push`. I will look at the tests and perhaps add some examples to outline the desired behaviour. If there are changes required, I will make them in the code.

plugins/aggregators/derivative/README.md
srebhan: I don't like that approach of duplicating the `fieldpass` functionality. Please just add the documentation (which should be obvious for anybody anyway). The case of someone using a custom field (should be rare anyway) and then filtering that field out is clearly not a very common use-case either. Just make sure you catch that as an error and write a unit test for it.

plugins/aggregators/derivative/README.md
```diff
@@ -0,0 +1,81 @@
...
+variable in the denominator. This variable is assumed to be a monotonously
+increasing value. In this feature the following formula is used:
+```
+                 field_last - field_first
+derivative = --------------------------------
+              variable_last - variable_first
+```
+For each field the derivative then is emitted with a naming pattern
+`fieldname_by_variablename`.
+
```
srebhan, line 31: Put the discussed hint here.
```suggestion
**Make sure the specified variable is not filtered and exists in the metrics passed to this aggregator!**

```

plugins/aggregators/derivative/README.md
srebhan: I don't think this is resolved!?

plugins/aggregators/derivative/derivative.go
```diff
@@ -0,0 +1,211 @@
...
+	## feature is used by specifying a *variable*.
+	# variable = "parameter"
+	##
+	## When using a field as the derivation parameter the name of that field will
+	## be used for the resulting derivative, e.g. *fieldname_by_parameter*.
+	##
+	## As a genereal parameter the period needs to be given. It determines the
+	## period for which a derivative will be calculated
+	period = "30s"
+	##
```
srebhan, line 66: Can you please move this general option up and also add `drop_original`.

plugins/aggregators/derivative/derivative.go
```diff
@@ -0,0 +1,211 @@
...
+	## The resulting derivative will be named *fieldname_by_seconds*. The infix
+	## "_by_" can be configured by the *infix* parameter.
+	# infix = "_by_"
+	##
+	## As an abstraction the derivative can be calculated not only by the time
+	## difference but by the difference of a field, which is contained in the
+	## measurement. This field is assumed to be monotonously increasing. This
+	## feature is used by specifying a *variable*.
+	# variable = "parameter"
```
srebhan, line 58: Can you please sync this with the README and use
```suggestion
	# variable = ""
```

plugins/aggregators/derivative/README.md
srebhan: You can mark this as resolved.

plugins/aggregators/derivative/README.md
srebhan: Thanks, please also add in `sampleConfig`. Put this as a separate comment, so you can mark this one as resolved.

plugins/aggregators/derivative/derivative.go
srebhan: Resolved thanks!

plugins/aggregators/derivative/README.md
srebhan: Resolved thx.

plugins/aggregators/derivative/README.md
srebhan: Keep what you have as I cannot come-up with a better solution :-) and mark this as resolved.

plugins/aggregators/derivative/derivative.go
srebhan: Resolved thx.

plugins/aggregators/derivative/derivative.go
srebhan: > So yes, in both cases dropping the field was intentional. Did I miss something in this?

No I don't think you missed something, but the problem is that in the described cases you potentially drop fields and/or, if I interpret your code correctly, fail calculating the denominator in `calculateDenominator()`
```
￼	if len(d.Variable) != 0 {
￼		return aggregate.last.fields[d.Variable] - aggregate.first.fields[d.Variable]
￼	}
￼}
```
if for any reason `d.Variable` did not arrive. The lookup will then be `nil`.
Even if you go with the time denominator, the problem is that the user will end up with `null` field values in the database for the missing fields and has no indication of why. But I think to fix this, you would need to completely restructure the code, keeping track of the individual fields instead of metrics...
So to cut a long story short: Keep what you have, but please fix `calculateDenominator()` for cases where `d.Variable` did not arrive for any of the metrics.

srebhan: > Your scenario with the the slow incoming metrics is a flaw in the `cache` clean-up. Elements are only removed from the `cache` when `MaxRollOver` is reached. Naively I would think, that it should also be removed, when a derivative is emitted in `Push`. I will look at the tests and perhaps add some examples to outline the desired behaviour. If there are changes required, I will make them in the code.

I don't think it is about cleaning the `cache` but rather about the logic of replacing the values. I will outline 4 arriving samples with the form `timestamp: value` (imaging a metric with only one field). Say we get `0: 1.0; 1: 2.0; 2: 1.5; 3: 2.2` and compute the derivative by time. With your code you get
| incoming | cache first (start) | cache last (start)  | rollover after `Reset()` | cache first (end) | cache last (end) | outgoing | comment |
|---|---|---|---|---|---|---|---|
|0: 1.0| - | - | - | 0: 1.0 |  0: 1.0 | - | denominator zero |
|1: 2.0| 0: 1.0 | 0: 1.0 | 1 | 0: 1.0 |  1: 2.0 | 1.0 | so far so good, `rollover` is reset to zero |
|2: 1.5; 3: 2.2| 1: 2.0 | 1: 2.0 | 1 | 1: 2.0 |  3: 2.2 | 0.1 | `current.first.time` is neither after nor equal `in.Time()` for the first arriving metric|

Is this what you want? For the last line in the table I could expect
| incoming | cache first (start) | cache last (start)  | rollover after `Reset()` | cache first (end) | cache last (end) | outgoing | comment |
|---|---|---|---|---|---|---|---|
|2: 1.5; 3: 2.2| 1: 2.0 | 1: 2.0 | 1 | 2: 1.5 |  3: 2.2 | 0.7 | |

In case multiple metrics arrive in the `period` I would expect the first one in `period` replaces `current.first`...

plugins/aggregators/derivative/derivative.go
srebhan: Resolved thx.

plugins/aggregators/derivative/derivative.go
srebhan: Resolved thx.

plugins/aggregators/derivative/derivative.go
srebhan: Resolved thx.

plugins/aggregators/derivative/derivative.go
srebhan: Resolved thx.

plugins/aggregators/derivative/derivative.go
srebhan: This one is still open see the discussion above.

plugins/aggregators/derivative/derivative.go
```diff
@@ -0,0 +1,211 @@
...
+		denominator := d.calculateDenominator(aggregate)
+		if denominator == 0 {
+			d.Log.Infof("Got difference 0 in denominator for %q, skipping.", aggregate.name)
+			continue
+		}
+		derivatives := make(map[string]interface{})
+		for key, start := range aggregate.first.fields {
+			if end, ok := aggregate.last.fields[key]; key != d.Variable && ok {
+				derivatives[d.derivativeFieldName(key)] = (end - start) / denominator
+			}
```
srebhan, line 169: I think
```suggestion
			if key == d.Variable {
				// Skip the denominator variable
				continue
			}
			if end, ok := aggregate.last.fields[key];  ok {
				derivatives[d.derivativeFieldName(key)] = (end - start) / denominator
			}
```
is easier to understand.

plugins/aggregators/derivative/derivative.go
srebhan: See my new comment.

plugins/aggregators/derivative/derivative.go
srebhan: Resolved thx.

srebhan (commented):
This looks almost good now. :-) The biggest issue is the handling of `first` in the `periods` (see https://github.com/influxdata/telegraf/pull/3762#discussion_r551218841). Could you please comment on my questions and fix the remaining minor issues!?

plugins/aggregators/derivative/derivative.go
KarstenSchnitter: Actually 0.1 is exactly what I want. This way the derivative is continuously calculated, leaving no holes in the time series. In your example taking (2: 1.5) as first value of the period would omit the decline between (1: 2.0) and (2: 1.5). This would give the false impression, that the value was continuously increasing, when in fact, there was a small decrease. You can get your desired behaviour by configuring `max_roll_over` to 0 and choosing a period, that ensures multiple values are aggregated. Dealing with jumps between aggregation periods was one of the reasons, I implemented the roll-over.

plugins/aggregators/derivative/README.md
KarstenSchnitter: Can I get the `fieldpass` configuration injected into the derivative aggregator with something like this?
```go
type Derivative struct {
	FieldPass    string[]          `toml:"fieldpass"`
	Variable    string          `toml:"variable"`
        // ...
}
```

I would need this to check in `Init` whether `Variable` is contained as well.

srebhan: I think you shouldn't. Please simply check whether `Variable` is available in the metrics during `calculateDenominator()` or not and issue a log warn/error? if not.

plugins/aggregators/derivative/derivative.go
srebhan: Ok, I'm also fine with your interpretation, but can you please describe this a bit more in the README!? Feel free to use my example without the internal fields and show what happens for the different patterns (first arrival skipped, single arrival per period, no arrival per period, multiple arrivals per period....). Maybe also use a table or similar showing the different cases and describe what will be the output and give a short explanation of the "general rule".

plugins/aggregators/derivative/derivative.go
```diff
@@ -0,0 +1,234 @@
...
+	period = "30s"
+	##
+	## If true, the original metric will be dropped by the
+	## aggregator and will not get sent to the output plugins.
+	drop_original = false
+	##
+	## Fields for which the derivative should be calculated.
+	## Important: The derivation variable must be contained in that list, if used.
+	fieldpass = ["field1, field2, variable"]
+    ##
```
srebhan, line 50: Formatting.

srebhan, line 50: Please remove this section. See README comment.

plugins/aggregators/derivative/derivative.go
```diff
@@ -0,0 +1,234 @@
...
+	##
+	## The resulting derivative will be named *fieldname_by_seconds*. The infix
+	## "_by_" can be configured by the *infix* parameter.
+	# infix = "_by_"
+	##
+	## As an abstraction the derivative can be calculated not only by the time
+	## difference but by the difference of a field, which is contained in the
+	## measurement. This field is assumed to be monotonously increasing. This
+	## feature is used by specifying a *variable*.
+	# variable = ""
```
srebhan, line 69: Add a warning here
```suggestion
	## feature is used by specifying a *variable*.
	## Make sure the specified variable is not filtered and exists in the metrics passed to this aggregator!
	# variable = ""
```

plugins/aggregators/derivative/derivative.go
```diff
@@ -0,0 +1,234 @@
...
+	switch v := in.(type) {
+	case float64:
+		return v, true
+	case int64:
+		return float64(v), true
+	case uint64:
+		return float64(v), true
+	default:
+		return 0, false
+	}
```
srebhan, line 158: You **could** also write
```suggestion
	}
	return 0, false
```
but that's probably a matter of taste.

plugins/aggregators/derivative/derivative.go
```diff
@@ -0,0 +1,234 @@
...
+				continue
+			}
+			if last, found = aggregate.last.fields[d.Variable]; !found {
+				d.Log.Infof("Did not find %q in last event for %q.", d.Variable, aggregate.name)
+				continue
+			}
+			denominator = last - first
+		}
+		if denominator == 0 {
+			d.Log.Infof("Got difference 0 in denominator for %q, skipping.", aggregate.name)
```
srebhan, line 186: Not sure if `Info` is a sufficient logging level as we drop metrics. It is a balance between likelihood and severeness though. Please rethink if this shouldn't be `Warn`. But as I do not have any data, you should decide here.

plugins/aggregators/derivative/derivative.go
```diff
@@ -0,0 +1,234 @@
...
+			continue
+		}
+		derivatives := make(map[string]interface{})
+		for key, start := range aggregate.first.fields {
+			if key == d.Variable {
+				// Skip derivation variable
+				continue
+			}
+			if end, ok := aggregate.last.fields[key]; key != d.Variable && ok {
```
srebhan, line 195: You should remove the `key` check here
```suggestion
			if end, ok := aggregate.last.fields[key]; ok {
```

plugins/aggregators/derivative/derivative.go
```diff
@@ -0,0 +1,234 @@
...
+		}
+		acc.AddFields(aggregate.name, derivatives, aggregate.tags)
+	}
+}
+
+func (d *Derivative) derivativeFieldName(field string) string {
+	if len(d.Variable) != 0 {
+		return field + d.Infix + d.Variable
+	}
+	return field + d.Infix + "seconds"
```
srebhan, line 207: Maybe
```suggestion
	return field + d.Infix + "time"
```
would be better?
Btw: You could determine the `suffix` once in `Init()` adding it to the `Derivative` struct and skip the whole function.

srebhan:
Please also check the failing test.

plugins/aggregators/derivative/derivative.go
KarstenSchnitter: I will do this. It also noticed, that the behaviour is not straight-forward. But it is useful.

plugins/aggregators/derivative/derivative.go
KarstenSchnitter: This needs some more work. This code branch is also taken, when there was only one metric during the period. This needs to be changed, before changing the log level to warn. I will work on that.

plugins/aggregators/derivative/derivative.go
KarstenSchnitter: I will change it to a constant suffix determined in `Init()`.

KarstenSchnitter:
Fixing the tests is tricky. They run fine on my machine. Regardless wether I run all or just the plugin. Is there a special trick to reproduce the failing result?

plugins/aggregators/derivative/derivative.go
KarstenSchnitter: Actually, when there is only one metric in the period, calculation of the period is also skipped. Hence, a higher log level would be needed according to your suggestion. I will change all log levels to WARN and fix the initialisation of a new aggregate.

KarstenSchnitter:
I fixed a timestamp issue in the tests, that caused the failing checks. Now everything is green. Is there anything left to do?

plugins/aggregators/derivative/README.md
srebhan: Mark as resolved plz!

plugins/aggregators/derivative/README.md
srebhan: Could you please either remove the `fieldpass` option in the example or add a comment and link to the global telegraf option to clearly state that it is not specific to your aggregator!

plugins/aggregators/derivative/derivative.go
```diff
@@ -0,0 +1,229 @@
...
+		if current.last.time.Before(in.Time()) {
+			current.last = newEvent(in)
+			current.rollOver = 0
+		}
+		if current.last.time.Equal(in.Time()) {
+			upsertConvertedFields(in.Fields(), current.last.fields)
+			current.rollOver = 0
+		}
+
+		d.cache[id] = current
```
srebhan, line 114: Do we really need this?

plugins/aggregators/derivative/derivative.go
```diff
@@ -0,0 +1,229 @@
...
+	id := in.HashID()
+	if current, ok := d.cache[id]; !ok {
+		// hit an uncached metric, create caches for first time:
+		d.cache[id] = newAggregate(in)
+	} else {
+		if current.first.time.After(in.Time()) {
+			current.first = newEvent(in)
+			current.rollOver = 0
+		}
+		if current.first.time.Equal(in.Time()) {
```
srebhan, line 101: Convert those `if`s to `else if`s to save some comparisons.

srebhan (commented):
I think we are very close. Please check my comments. Regarding the testing it might be an issue with the timestamp of your test values. You are using `time.Now()` which might have different resolutions on your machine and the CI machines. Just use some fixed timestamps to avoid any non-deterministic behavior. Otherwise just run your tests in a loop (e.g. 100 times) and check if it fails. We sometimes have those problems when e.g. accessing maps where access is shuffled by golang...

plugins/aggregators/derivative/derivative.go
KarstenSchnitter: I will make some changes. Keep in mind it is possible to have `first.time` equal to `last.time` equal to `in.Time()`.

srebhan: Ok, but the two groups (two cases for `first` and two cases for `last`) can be `else if` right!?

plugins/aggregators/derivative/README.md
KarstenSchnitter: I removed `fieldpass` from the example.

plugins/aggregators/derivative/derivative.go
KarstenSchnitter: I changed the cache to use pointers to avoid this statement.

srebhan (approved):
Perfect! Thanks for all your effort!

plugins/aggregators/derivative/README.md
```diff
@@ -0,0 +1,166 @@
...
+derivative = --------------------------
+                  time_difference
+```
+For each field the derivative is emitted with a naming pattern
+`<fieldname>_by_time`.
+
+### Custom Derivation Variable
+
+The plugin supports to use a field of the aggregated measurements as derivation
+variable in the denominator. This variable is assumed to be a monotonously
```
ssoroka, line 22:
```suggestion
variable in the denominator. This variable is assumed to be a monotonically
```

plugins/aggregators/derivative/README.md
```diff
@@ -0,0 +1,166 @@
...
+the period. From these measurements the time difference in seconds is
+calculated. This time difference is than used to divide the difference of each
+field using the following formula:
+```
+              field_last - field_first
+derivative = --------------------------
+                  time_difference
+```
+For each field the derivative is emitted with a naming pattern
+`<fieldname>_by_time`.
```
ssoroka, line 17: how do you feel about _rate instead of _by_time?

plugins/aggregators/derivative/derivative.go
```diff
@@ -0,0 +1,225 @@
...
+func (d *Derivative) Description() string {
+	return "Calculates a derivative for every field."
+}
+
+func (d *Derivative) Add(in telegraf.Metric) {
+	id := in.HashID()
+	if current, ok := d.cache[id]; !ok {
+		// hit an uncached metric, create caches for first time:
+		d.cache[id] = newAggregate(in)
+	} else {
```
ssoroka, line 96: just return and flatten the else. makes it easier to read

plugins/aggregators/derivative/derivative.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+	return 0, false
+}
+
+func (d *Derivative) Push(acc telegraf.Accumulator) {
+	for _, aggregate := range d.cache {
+		if aggregate.first == aggregate.last {
+			d.Log.Warnf("Same first and last event for %q, skipping.", aggregate.name)
+			continue
+		}
```
ssoroka, line 163: wouldn't this be a 0 value derivative result?

plugins/aggregators/derivative/derivative.go
```diff
@@ -0,0 +1,225 @@
...
+		var denominator float64
+		if len(d.Variable) == 0 {
+			// If no derivation variable is known default to timestamp
+			denominator = aggregate.last.time.Sub(aggregate.first.time).Seconds()
+		} else {
+			var first float64
+			var last float64
+			var found bool
+			if first, found = aggregate.first.fields[d.Variable]; !found {
+				d.Log.Warnf("Did not find %q in first event for %q.", d.Variable, aggregate.name)
```
ssoroka, line 173: Might want to move these to debug log statements. I don't think they're that useful for normal operation.

plugins/aggregators/derivative/derivative.go
```diff
@@ -0,0 +1,225 @@
...
+func (d *Derivative) Push(acc telegraf.Accumulator) {
+	for _, aggregate := range d.cache {
+		if aggregate.first == aggregate.last {
+			d.Log.Warnf("Same first and last event for %q, skipping.", aggregate.name)
+			continue
+		}
+		var denominator float64
+		if len(d.Variable) == 0 {
+			// If no derivation variable is known default to timestamp
+			denominator = aggregate.last.time.Sub(aggregate.first.time).Seconds()
```
ssoroka, line 167: can you get rid of the else by making this the default value and flipping the condition?

ssoroka (changes requested):
looks really good. just some minor feedback.

plugins/aggregators/derivative/README.md
KarstenSchnitter: Actually the suffix is configurable. `_by_time` is currently the default value. I can change it to `_rate` if you prefer that.

plugins/aggregators/derivative/derivative.go
KarstenSchnitter: It is undefined, since the denominator would be 0.

KarstenSchnitter: Furthermore, notice, that the value can be "rolled-over" to the next interval, where a proper value may be calculated without emitting a 0 in between. That is actually one of the goals of the roll-over.

plugins/aggregators/derivative/derivative.go
KarstenSchnitter: I do not know if I understand your suggestion. Wouldn't this mean I would always calculate the time differences, even when the are not needed?

plugins/aggregators/derivative/derivative.go
KarstenSchnitter: It was a suggestion by @srebhan to log at warning level, whenever no value would be emitted. Logging the different reasons already proved valuable to me during debugging the plugin.

srebhan: @ssoroka my reasoning was to allow the user to find out why there was no metric emitted. We could change the log-level to debug if you fear too many log-messages... In general this should only trigger if you get a new _field_ key in an interval that was not seen before (something that should not happen IIRC).

plugins/aggregators/derivative/derivative.go
srebhan: The idea of @ssoroka  is to do
```golang
denominator = aggregate.last.time.Sub(aggregate.first.time).Seconds()
if len(d.Variable) > 0 {
```
instead of the if/else construct. However this means that you, in the non-default-case, do a superfluous computation of the time difference. Don't know if this matters...

KarstenSchnitter: This is how I understood @ssoroka. I just wanted to make sure, that I really should always calculate the time difference, even when it is not needed. I will make that change.

plugins/aggregators/derivative/derivative.go
KarstenSchnitter: Do I need to change something here?

srebhan: @ssoroka?

srebhan (approved):
Looks still good to me. @ssoroka any comments?

plugins/aggregators/derivative/derivative.go
ssoroka: yes, use debug please.

ssoroka:
```suggestion
				d.Log.Debugf("Did not find %q in first event for %q.", d.Variable, aggregate.name)
```

plugins/aggregators/derivative/derivative.go
```diff
@@ -0,0 +1,224 @@
...
+	case uint64:
+		return float64(v), true
+	}
+	return 0, false
+}
+
+func (d *Derivative) Push(acc telegraf.Accumulator) {
+	for _, aggregate := range d.cache {
+		if aggregate.first == aggregate.last {
+			d.Log.Warnf("Same first and last event for %q, skipping.", aggregate.name)
```
ssoroka, line 162:
```suggestion
			d.Log.Debugf("Same first and last event for %q, skipping.", aggregate.name)
```
probably this one too

plugins/aggregators/derivative/derivative.go
```diff
@@ -0,0 +1,224 @@
...
+		if len(d.Variable) > 0 {
+			var first float64
+			var last float64
+			var found bool
+			if first, found = aggregate.first.fields[d.Variable]; !found {
+				d.Log.Warnf("Did not find %q in first event for %q.", d.Variable, aggregate.name)
+				continue
+			}
+			if last, found = aggregate.last.fields[d.Variable]; !found {
+				d.Log.Warnf("Did not find %q in last event for %q.", d.Variable, aggregate.name)
```
ssoroka, line 176:
```suggestion
				d.Log.Debugf("Did not find %q in last event for %q.", d.Variable, aggregate.name)
```

plugins/aggregators/derivative/derivative.go
```diff
@@ -0,0 +1,224 @@
...
+				continue
+			}
+			denominator = last - first
+		}
+		if denominator == 0 {
+			d.Log.Warnf("Got difference 0 in denominator for %q, skipping.", aggregate.name)
```
ssoroka, line 182:
```suggestion
			d.Log.Debugf("Got difference 0 in denominator for %q, skipping.", aggregate.name)
```

ssoroka (changes requested):
just need to make logs ness noisy.

srebhan (approved):

ssoroka (approved):

KarstenSchnitter:
Wow, great to see the merge. Thanks for all the help and effort.

PR #3763, opened 2018-02-07
Add Drop Strings Processor

KarstenSchnitter:
This processor will drop all metrics with string values. This is helpful in
cases where the complete set of metrics is unknown or unsure but non-numerical
values should be avoided, e.g. for compliance reasons.

### Required for all PRs:

### DIFF
--- a/plugins/processors/all/all.go
+++ b/plugins/processors/all/all.go
@@ -1,5 +1,6 @@
 package all
 
 import (
+	_ "github.com/influxdata/telegraf/plugins/processors/drop_strings"
 	_ "github.com/influxdata/telegraf/plugins/processors/printer"
 )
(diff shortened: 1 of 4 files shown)


sjwang90:
Is this something that can now be addressed with [`processors.starlark`](https://github.com/influxdata/telegraf/tree/master/plugins/processors/starlark)?

KarstenSchnitter:
I agree, that with the right script, this issue can be addressed with the starlark processor plugin. The user has to develop and maintain that script though. Maybe this case can be included in the testdata?

ssoroka:
I think I'm going to close this in favor of https://github.com/influxdata/telegraf/pull/8438, which is more flexible. Otherwise we might end up with drop_x plugins for all available types.

PR #3350, opened 2017-10-18
Use labels in prometheus output for string fields

danielnelson:
closes: #2928 
closes: #2957

### Required for all PRs:

- [x] Signed [CLA](https://influxdata.com/community/cla/).
- [ ] Associated README.md updated.
- [x] Has appropriate unit tests.

PR #3765, opened 2018-02-07
Providing option to disable labels in prometheus output for string fields

shaheed121:
#3350 introduced a change where all string fields are turned to labels. However in certain cases where string value changes, it introduces a new metric due to the change in label value.

Hence this will provide an option to disable as well as enable only for certain string values. The default behavior doesn't change with this commit.

### DIFF
--- a/plugins/outputs/prometheus_client/README.md
+++ b/plugins/outputs/prometheus_client/README.md
@@ -23,4 +23,11 @@ This plugin starts a [Prometheus](https://prometheus.io/) Client, it exposes all
 
   # Expiration interval for each metric. 0 == no expiration
   expiration_interval = "60s"
+
+  # Disable labels in prometheus output for all string fields. (Default: false)
+  disable_string_to_label = false
+
+  # Enable labels in prometheus output for certain string fields.
+  # Won't work when disable_string_to_label is set to true.
+  string_to_label_names = []
 ```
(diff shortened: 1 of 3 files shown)


danielnelson:
What about using the [measurement filtering](https://github.com/influxdata/telegraf/blob/master/docs/CONFIGURATION.md#measurement-filtering) options to provide this using taginclude/tagexclude and fielddrop/fieldpass.

shaheed121:
@danielnelson Yes, I could do that. But I would have to know the list or patterns of tags beforehand. Instead if I disable converting string fields to labels globally, then I don't have to fiddle with all input plugins.

So what do you think about having atleast an option to disable this feature like how we had before 1.5.0 release?

danielnelson:
That makes sense, could you provide this in a similar way to how we we do it in the docker plugin with `docker_label_include` and `docker_label_exclude`?  Maybe we call it `string_label_include` and `string_label_exclude`.

shaheed121:
@danielnelson To keep it consistent across output plugins, I thought of doing something like `string_as_state` in Riemann output plugin. So have added `string_as_label` and set default as `true` to make sure that it exhibits the current behavior. 
The reason I didn't go ahead with `string_label_include` and `string_label_exclude` because it would be redundant since it can be managed using Measurement filtering like you said.

PR #3771, opened 2018-02-08
Flume input plugin

Icarus0xff:
### Required for all PRs:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ build
 /telegraf.gz
 *~
 *#
+cmd/telegraf/telegraf
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,8 +18,14 @@
 
 - The `sqlserver` input plugin has a new query and data model that can be enabled
   by setting `query_version = 2`.  It is encouraged to migrate to the new
-  model when possible as the old version is deprectated and will be removed in
+  model when possible as the old version is deprecated and will be removed in
   a future version.
+
+- An option has been added to the `openldap` input plugin that reverses metric
+  name to improve grouping.  This change is enabled when `reverse_metric_names = true`
+  is set.  It is encouraged to enable this option when possible as the old
+  ordering is deprecated.
+
 
 ### New Plugins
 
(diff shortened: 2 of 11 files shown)


danielnelson:
Thanks for the pull request @hikkikuma1991, could you modify the HTTP client to use a non-default Transport and add SSL/TLS support?  A good example of this is in the nginx plugin: https://github.com/influxdata/telegraf/blob/master/plugins/inputs/nginx/nginx.go#L63

Also, can you add some unit tests for the example JSON?  Again you could base it on the nginx tests: https://github.com/influxdata/telegraf/blob/master/plugins/inputs/nginx/nginx_test.go

AAbouZaid:
Well, I create [Flume plugin](https://github.com/AAbouZaid/telegraf/blob/flume_input_plugin/plugins/inputs/flume/flume.go) 2 weeks ago but didn't find time to push it :joy: 
But at least it's here now :tada: 

I will make some PR to @hikkikuma1991 repo to add some useful parts like:
- Multi urls/servers.
- Filtering.
- Numerical output (not string as they come from Flume metrics API).

@hikkikuma1991 I'd suggest default behavior to be collecting all metrics, and make option to filter them in case that needed, because in Flume the metrics could be extended with extra classes.

AAbouZaid:
@danielnelson I have a question.
I do see the HTTP client part again and again in many plugins, shouldn't this be a library or so to make it easier to use it?

danielnelson:
> I do see the HTTP client part again and again in many plugins, shouldn't this be a library or so to make it easier to use it?

Yes, likely.  I'll try to work something out soon.

russorat:
@hikkikuma1991 looks like according to the flume docs, monitoring is still a work in progress: https://flume.apache.org/FlumeUserGuide.html#monitoring

We'll probably hold off merging this until more people are using it or the flume monitoring stabilizes.

sjwang90:
Looks like monitoring for flume is still a work in progress. Considering it's been a couple years and still hasn't been updated may be better to have this as [external plugin](https://github.com/influxdata/telegraf/blob/master/docs/EXTERNAL_PLUGINS.md).

sjwang90:
Closing this issue due to inactivity and the low priority of this issue.

Issue #2791, opened 2017-05-12
Add ability to map string fields to numeric fields


bluebob:
### Proposal:
Introduce a numeric "cluster_health" or "cluster_status" value

### Current behavior:
Current implementation has a status field which is a text object, having values "green", "yellow" or "red".

### Desired behavior:
Add a numeric representation of this field, called cluster_health, cluster_status or just health. I suggest a simple mapping of 0, 1 and 2 for green,yellow and red.

### Use case: [Why is this important (helps with prioritizing requests)]
The "SignalFX" plugin for CollectD already has this behaviour. Having a numeric field permits Grafana presentation of state with time, so showing _when_ a cluster went into a degraded or failed state. This can be very useful in tracking down root cause. Current text-based representation has to be converted to a number in order to do this, which is cumbersome.
Alerting rules are also simplified, since != 0 means intervention is needed

phemmer:
Just for references:
Here's a grafana feature request which is to allow colorizing table / singlestat based on string values: https://github.com/grafana/grafana/issues/3601
Here's a influxdb feature request which would allow conditional substitution of values ("case when" stuff): https://github.com/influxdata/influxdb/issues/5383
Here's a telegraf feature request that would allow consitional substitution of values: https://github.com/influxdata/telegraf/issues/2667 (there's other issues with similar discussions about the end goal there, but can't find them at the moment)

And just for clarification, I'm not saying this request should be closed, just throwing out some related stuff.

bluebob:
Thanks. That info will be really handy as a "stop gap". 👍

danielnelson:
I think it is best to implement this with some enhancements to the #2667 proposal.  I'll write a follow up over there.

danielnelson:
Changing my mind on this, I guess it doesn't make sense to have a general purpose transform plugin so we would want a "enum" or "case" processor for this.

Here is my first idea for how to configure the processor.  I don't love it, so hopefully we can improve on the idea.
```toml
[[processors.enum]]
  namepass = ["elasticsearch"]
  [[processors.enum.fields]]
    key = "health"
    [processors.enum.table]
      green = 0
      yellow = 1
      red = 2
```

phemmer:
Kind of a crazy ass idea, but what about embedding kapacitor in telegraf? We wouldn't put the whole thing in, just the stream processing engine.
[...]

PR #3772, opened 2018-02-08
Add Enum Processor

KarstenSchnitter:
This processor allows substitution of field values of string or bool type with
numeric values. It can be used, e.g. to map status codes such as "red", "amber"
and "green" to values such as 0, 1, 2. Multiple fields can be configured, each
with its own mapping table.

### Required for all PRs:

### DIFF
--- a/plugins/processors/enum/README.md
+++ b/plugins/processors/enum/README.md
@@ -1,0 +1,27 @@
+# Enum Processor Plugin
+
+The Enum Processor allows the configuration of value mappings for metric fields.
+The main use-case for this is to rewrite status codes such as _red_, _amber_ and
+_green_ by numeric values such as 0, 1, 2. The plugin supports string and bool
+types for the field values. Multiple Fields can be configured with separate
+value mappings for each field.
+
+### Configuration
+Configuration using table syntax:
+`toml
+# Configure a status mapping for field 'status'
+[[processors.enum.fields]]
+  key = "status"
+  [processors.enum.fields.value_mappings]
+    green = 0
+    yellow = 1
+    red = 2
+`
+
+Configuration using inline syntax:
+`toml
+# Configure a status mapping for field 'status'
+[[processors.enum.fields]]
+  key = "status"
+  value_mappings = {green = 0, yellow = 1, red = 2 }
+`
--- a/plugins/processors/enum/README.md
+++ b/plugins/processors/enum/README.md
@@ -11,7 +11,7 @@ Configuration using table syntax:
 `toml
 # Configure a status mapping for field 'status'
 [[processors.enum.fields]]
-  key = "status"
+  source = "status"
   [processors.enum.fields.value_mappings]
     green = 0
     yellow = 1
@@ -22,6 +22,6 @@ Configuration using inline syntax:
 `toml
 # Configure a status mapping for field 'status'
 [[processors.enum.fields]]
-  key = "status"
+  source = "status"
   value_mappings = {green = 0, yellow = 1, red = 2 }
 `
--- a/plugins/processors/enum/README.md
+++ b/plugins/processors/enum/README.md
@@ -4,7 +4,8 @@ The Enum Processor allows the configuration of value mappings for metric fields.
 The main use-case for this is to rewrite status codes such as _red_, _amber_ and
 _green_ by numeric values such as 0, 1, 2. The plugin supports string and bool
 types for the field values. Multiple Fields can be configured with separate
-value mappings for each field.
+value mappings for each field. Default mapping values can be configured to be
+used for all values, which are not contained in the value_mappings.
 
 ### Configuration
 Configuration using table syntax:
@@ -12,6 +13,7 @@ Configuration using table syntax:
 # Configure a status mapping for field 'status'
 [[processors.enum.fields]]
   source = "status"
+  default = -1
   [processors.enum.fields.value_mappings]
     green = 0
     yellow = 1
@@ -23,5 +25,6 @@ Configuration using inline syntax:
 # Configure a status mapping for field 'status'
 [[processors.enum.fields]]
   source = "status"
+  default = -1
   value_mappings = {green = 0, yellow = 1, red = 2 }
 `
--- a/plugins/processors/enum/README.md
+++ b/plugins/processors/enum/README.md
@@ -5,7 +5,9 @@ The main use-case for this is to rewrite status codes such as _red_, _amber_ and
 _green_ by numeric values such as 0, 1, 2. The plugin supports string and bool
 types for the field values. Multiple Fields can be configured with separate
 value mappings for each field. Default mapping values can be configured to be
-used for all values, which are not contained in the value_mappings.
+used for all values, which are not contained in the value_mappings. The
+processor supports explicit configuration of a destination field. By default the
+source field is overwritten.
 
 ### Configuration
 Configuration using table syntax:
@@ -13,6 +15,7 @@ Configuration using table syntax:
 # Configure a status mapping for field 'status'
 [[processors.enum.fields]]
   source = "status"
+  destination = "code"
   default = -1
   [processors.enum.fields.value_mappings]
     green = 0
@@ -25,6 +28,7 @@ Configuration using inline syntax:
 # Configure a status mapping for field 'status'
 [[processors.enum.fields]]
   source = "status"
+  destination = "code"
   default = -1
   value_mappings = {green = 0, yellow = 1, red = 2 }
 `
--- a/plugins/processors/enum/enum.go
+++ b/plugins/processors/enum/enum.go
@@ -17,6 +17,10 @@ var sampleConfig = `
 # Name of the field source field to map
 #   source = "name"
 #
+# Optional destination field to be used for the mapped value. Source field is
+# used, when no explicit destination is configured.
+#   destination = "mapped"
+#
 # Optional default value to be used for all values not contained in the mapping
 # table. Only applied when configured.
 #   default = 0
@@ -36,6 +40,7 @@ type EnumMapper struct {
 
 type Mapping struct {
 	Source        string
+	Destination   string
 	Default       interface{}
 	ValueMappings map[string]interface{}
 }
@@ -60,8 +65,7 @@ func (mapper *EnumMapper) applyMappings(metric telegraf.Metric) telegraf.Metric 
 		if originalValue, isPresent := metric.GetField(mapping.Source); isPresent == true {
 			if adjustedValue, isString := adjustBoolValue(originalValue).(string); isString == true {
 				if mappedValue, isMappedValuePresent := mapping.mapValue(adjustedValue); isMappedValuePresent == true {
-					metric.RemoveField(mapping.Source)
-					metric.AddField(mapping.Source, mappedValue)
+					writeField(metric, mapping.getDestination(), mappedValue)
 				}
 			}
 		}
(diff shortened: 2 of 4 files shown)


plugins/processors/enum/README.md
```diff
@@ -0,0 +1,27 @@
...
+### Configuration
+Configuration using table syntax:
+`toml
+# Configure a status mapping for field 'status'
+[[processors.enum.fields]]
+  key = "status"
+  [processors.enum.fields.value_mappings]
+    green = 0
+    yellow = 1
+    red = 2
```
danielnelson: It might be nice if you could optionally modify the output fieldkey, so that you could transform like this:
```diff
-foo status="red"
+foo status="red",status_code=2i
```

Maybe something like this?:
```toml
[[processors.enum]]
  [[processors.enum.fields]]
    source = "status"
    destination = "status_code"
    [processors.enum.fields.value_mappings]
      green = 0
      yellow = 1
      red = 2
```

plugins/processors/enum/enum.go
```diff
@@ -0,0 +1,110 @@
...
+func (mapper *EnumMapper) SampleConfig() string {
+	return sampleConfig
+}
+
+func (mapper *EnumMapper) Description() string {
+	return "Map enum values according to given table."
+}
+
+func (mapper *EnumMapper) Apply(in ...telegraf.Metric) []telegraf.Metric {
+	out := make([]telegraf.Metric, 0, len(in))
```
danielnelson: We should try to make this as efficient as possible since it will be ran in over all metrics, perhaps we can modify `in` directly?

plugins/processors/enum/enum.go
```diff
@@ -0,0 +1,110 @@
...
+			out = append(out, target)
+		} else {
+			out = append(out, source)
+		}
+	}
+	return out
+}
+
+func (mapper *EnumMapper) applyMappings(source telegraf.Metric) (telegraf.Metric, error) {
+	if fields, changed := mapper.applyFieldMappings(source.Fields()); changed == true {
```
danielnelson: Instead of grabbing all the fields, it might be faster to use `HasField` to check for the fields we are interested in and modify the metric instead of creating a new one.

danielnelson:
closes #2791

plugins/processors/enum/README.md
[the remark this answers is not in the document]
KarstenSchnitter: I was using this plugin together with #3763, which drops any string values and thus removing unmatched values. For a more generic use-case I see the benefit in having a default behaviour to ensure a constant data type. I proposed to make dropping the field if it contains an unspecified value the default behaviour and optionally use a configurable default value:
```
[[processors.enum]]
  [[processors.enum.fields]]
    source = "status"
    default = -1
    [processors.enum.fields.value_mappings]
      green = 0
      yellow = 1
```
Would that be a reasonable solution?

KarstenSchnitter: That is a good idea. I will implement it.

plugins/processors/enum/enum.go
KarstenSchnitter: Checking with `HasField`first is a great suggestion I will implement that. As discussed above I have trouble when modifying the metric directly in the case where it has just the one metric we are interested in. Note, that the implementation will only create a new metric when a substitution was made and pass on the original object otherwise.

plugins/processors/enum/README.md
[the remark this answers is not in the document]
danielnelson: Sounds good

plugins/processors/enum/enum.go
danielnelson: I flipped through the other comments and couldn't find the discussion of this, can you tell me more about the problem modifying the metric directly or point me in the right direction?

KarstenSchnitter: I can provide a simple test to demonstrate my observations:

```go
import (
	"fmt"
	"testing"
	"time"

	"github.com/influxdata/telegraf/metric"
)

func TestDirectModification(t *testing.T) {
	metric, _ := metric.New("m",
		map[string]string{},
		map[string]interface{}{"test": "value"},
		time.Now(),
	)
	fmt.Println(metric.String()) // original state
	metric.RemoveField("test")
	fmt.Println(metric.String()) // field is not removed
	metric.AddField("test", 13)
	fmt.Println(metric.String()) // added new value
	metric.RemoveField("test")
	fmt.Println(metric.String()) // remove old value, but colon error
	metric.Fields()              // panic because of colon from above
}
```

This creates the following output:
```
$ go test
m test="value" 1519041597145903000

m test="value" 1519041597145903000

m test="value",test=13i 1519041597145903000

m ,test=13i 1519041597145903000

--- FAIL: TestDirectModification (0.00s)
panic: runtime error: index out of range [recovered]
	panic: runtime error: index out of range

goroutine 5 [running]:
testing.tRunner.func1(0xc4201080f0)
	/usr/local/Cellar/go/1.9.3/libexec/src/testing/testing.go:711 +0x2d2
panic(0x1292140, 0x1458a40)
	/usr/local/Cellar/go/1.9.3/libexec/src/runtime/panic.go:491 +0x283
github.com/influxdata/telegraf/metric.(*metric).Fields(0xc42009a200, 0x1)
	/Workspaces/go/src/github.com/influxdata/telegraf/metric/metric.go:345 +0xa22
github.com/influxdata/telegraf/plugins/processors/enum.TestDirectModification(0xc4201080f0)
	/Workspaces/go/src/github.com/influxdata/telegraf/plugins/processors/enum/demo_test.go:24 +0x4f9
testing.tRunner(0xc4201080f0, 0x12f5940)
	/usr/local/Cellar/go/1.9.3/libexec/src/testing/testing.go:746 +0xd0
created by testing.(*T).Run
	/usr/local/Cellar/go/1.9.3/libexec/src/testing/testing.go:789 +0x2de
exit status 2
FAIL	github.com/influxdata/telegraf/plugins/processors/enum	0.020s
```

So the following behaviour can be observed:

- `RemoveField(test)` on the last field does not change the metric, the field remains.
- `AddField(test,13)` creates a new field with a different value
- `RemoveField(test)` removes the old value but leads to a leading ',' which indicates a problem
- the problem from above manifests in a panic when the `Fields()` is called 

This problem is always present, if the leading field in the string representation is removed. Even when more fields are present. Therefore, I decided, that modifying a field value directly by this approach is not possible. The issue lies in the unreliability of `RemoveField(...)`. 

Is there another approach to a direct modification? I really want to keep the feature of replacing the original value not just adding the mapped value as a new field.

danielnelson: I'm rewriting the implementation of `telegraf.Metric` for 1.6 and I'll fix this, I'll let you know when its ready for testing.

ikovalyov:
What if we need to map few strings to the same numeric value?

KarstenSchnitter:
Using this processor you specify a map what string to map to what value. You can easily specify multiple strings mapping to the same value, e.g.:
```toml
[[processors.enum]]
  [processors.enum.fields.value_mappings]
    green = 0
    yellow = 1
    red = 1
```

KarstenSchnitter:
I committed changes addressing the questions by @danielnelson quite some time ago. Can we have another review on the current state of this PR?

KarstenSchnitter:
I reputed to fix the merge conflict.

gutschein:
I tried to map MAC-Addresses to hostname, but it doesn't work for me. Could it be possible that handling of more complex strings (containing : or _ ) and hwaddr is not yet covered? Or is therer a limitation for tag-fields? With this setting hostname is not filled with default value nor hostnames to MAC-Addresses.
```
  [[processors.enum]]
    [[processors.enum.mapping]]
      field = "caps_man_mac_address"
      dest =  "caps_man_hostname"
      default = "na"
      [processors.enum.mapping.value_mappings]
         "AA:11:BB:22:CC:33" = "hostname"
...
    [[inputs.snmp.table.field]]
      name = "caps_man_mac_address"
      oid = ".1.3.6.1.4.1.14988.1.1.1.5.1.1"
      conversion = "hwaddr"
      is_tag = true
    [[inputs.snmp.table.field]]     # doubled to get a taged field , replaced by value mapping
      name = "caps_man_hostname"
      oid = ".1.3.6.1.4.1.14988.1.1.1.5.1.1"
      is_tag = true
```

KarstenSchnitter:
Hi @gutschein,

thanks for trying out the enum processor plugin. Currently, the plugins supports only mapping for fields, but not for tags. If I understand your example correctly, the configuration `is_tag = true` adds `caps_man_mac_address` as a tag to the measurement. Therefore, the enum plugin is not able to process the mapping. 

You could try to first emit the mac address as a field using `is_tag = false`, then map the field value with the enum processor plugin using your configuration. If necessary you can use the [converter processor plugin](https://github.com/influxdata/telegraf/tree/master/plugins/processors/converter) to convert the mapped field into a tag. Note, that you may need to specify the `order` property of the processor plugins to ensure, that the measurements are first processed by the enum plugin and then by the converter plugin.

If your problem persists, please provide an example measurement generated by your input configuration and the desired output from the enum processor plugin. This makes it easier to understand the issue and provide a solution.

Best Regards,
Karsten

gutschein:
thanks for writing enum... :)  and thanks for your response.

Good idea to change field to tag later usinge the converter. I give it a try after the basic is running. You mentioned an option to order of the processing, too...interessting I wasn't aware of this option.

Back to the basics: I tried it with the destination field as a simple field (not tag) but it stilld doesn't work. Even the `default = "na"` doesn't work. It seems to me that delimiting at all doesn't work in my tests: I tried other strings (containing e.g. space) instead of the little bit complicated MAC-address as source field but my destination field was unchanged at all. 
And beside the target field: Is it allowed that the source field is a tag (in your example e.g status = "green") or do we have the same limitation there like for the destination field?

My target:
getting an MAC-address (e.g. "11:AA:22:BB:33:CC") from my access points in a tag i like to derive the device name (e.g. "Foos iphone") as a additional tag for using in grafana (known devices in WLAN). By the way: This may be a use case for many people if the source (here: my access point) can't resolve the names:
e.g. `"11:AA:22:BB:33:CC" = "Foos iphone"`

simplified target
I have a tag (pitfall 1?) with values containing space (pitfall 2?) as source to derive a string field (not a tag) (pitfall 3?)
e.g. `"VLAN 10" = "media"`

It seems to me that enum just accepts strings w/o space (an w/o `"`-Delimiter) as source and a (int) value as target by now. right?
e.g. `green = 1`

KarstenSchnitter:
I do not think, that a whitespace character is an issue. I think, that you suffer from pitfall 1: The enum processor plugin can only read fields, but no tags. In your example the plugin would look for a **field** `caps_man_mac_address` and replace any value of `AA:11:BB:22:CC:33` by `hostname` and any other value by `na`. However, your input plugin only provides a **tag** with that name but no field. Therefore the value remains unchanged. 

If you want support for tags in the enum processor plugin, I suggest to create an issue with the corresponding feature request.

Regarding the other pitfalls (2 and 3) it would be nice to have a short example that proves the misbehaviour. Can you provide that? So far I am not convinced, that there is an issue. But I will look into this in the next few days.

gutschein:
Ah, ok, thanks for explanation. So...I will try next weekend: Test with field instead of a tag containing a whitespace.

KarstenSchnitter:
@glinton, I missed out on that one. This is a great feature. I guess @gutschein just needs to change `field = "caps_man_mac_address"` to `tag = "caps_man_mac_address"` then.

gutschein:
I just tested it...indeed...my example from 7 days ago works perfect with `tag = "caps_man_mac_address"`. Great work! Thank you.

I think the online help (https://github.com/influxdata/telegraf/tree/master/plugins/processors/enum) should be adjusted to reflect this option:
* source can be field or tag (with choosing `field `or `tag`)
* destination can be field or tag (as it is defined, nothing to define here)
* source strings with whitespace or hwaddr or ip should be delimited
* destination strings sould be delimited.
Maybe mentioning this second use case "MAC-address to hostname" would be helpfull, too.

Issue #3467, opened 2017-11-13
Add "sum" to basicstats aggregator


puckpuck:
### Proposal:
Add `sum` metric to basicstats.

### Current behavior:
`sum` is not a computed aggregate.

### Desired behavior:
`sum` would be part of computed aggregates in conjunction with `count`, `min`, etc.

### Use case: 
Although this can be computed (mean * count) it would be easier and consistent with many other metrics platforms if this was delivered out of the box.  I would be happy to do this myself as a PR, but wanted to see if there was any reason why it wasn't done initially.

danielnelson:
I think it is probably something we can do, so long as we also do https://github.com/influxdata/telegraf/issues/3413.  I imagine there will be interest in other simple aggregations.

@toni-moreno Do you have any thoughts on this?

danielnelson:
Implemented in #3797 and will be in 1.6, thanks @cpacey

PR #3797, opened 2018-02-16
Add sum stat to basicstats aggregator

cpacey:
Implements #3467.  `sum` is not included by default to maintain backwards compatibility.

### Required for all PRs:

### DIFF
--- a/plugins/aggregators/basicstats/basicstats.go
+++ b/plugins/aggregators/basicstats/basicstats.go
@@ -41,6 +41,7 @@ type basicstats struct {
 	count float64
 	min   float64
 	max   float64
+	sum   float64
 	mean  float64
 	M2    float64 //intermedia value for variance/stdev
 }
@@ -78,6 +79,7 @@ func (m *BasicStats) Add(in telegraf.Metric) {
 					min:   fv,
 					max:   fv,
 					mean:  fv,
+					sum:   fv,
 					M2:    0.0,
 				}
 			}
@@ -93,6 +95,7 @@ func (m *BasicStats) Add(in telegraf.Metric) {
 						min:   fv,
 						max:   fv,
 						mean:  fv,
+						sum:   fv,
 						M2:    0.0,
 					}
 					continue
@@ -120,6 +123,8 @@ func (m *BasicStats) Add(in telegraf.Metric) {
 				} else if fv > tmp.max {
 					tmp.max = fv
 				}
+				//sum compute
+				tmp.sum += fv
 				//store final data
 				m.cache[id].fields[k] = tmp
 			}
@@ -148,7 +153,7 @@ func (m *BasicStats) Push(acc telegraf.Accumulator) {
 				fields[k+"_mean"] = v.mean
 			}
 			if config.sum {
-				fields[k+"_sum"] = (v.mean * v.count)
+				fields[k+"_sum"] = v.sum
 			}
 
 			//v.count always >=1
--- a/plugins/aggregators/basicstats/basicstats.go
+++ b/plugins/aggregators/basicstats/basicstats.go
@@ -22,6 +22,7 @@ type configuredStats struct {
 	mean     bool
 	variance bool
 	stdev    bool
+	sum      bool
 }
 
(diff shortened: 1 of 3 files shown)


plugins/aggregators/basicstats/basicstats.go
```diff
@@ -146,6 +147,9 @@ func (m *BasicStats) Push(acc telegraf.Accumulator) {
 			if config.mean {
 				fields[k+"_mean"] = v.mean
 			}
+			if config.sum {
+				fields[k+"_sum"] = (v.mean * v.count)
```
cpacey: This seemed the simplest way to implement it.

plugins/aggregators/basicstats/basicstats_test.go
```diff
@@ -357,3 +434,26 @@ func TestBasicStatsWithUnknownStat(t *testing.T) {
 
 	acc.AssertDoesNotContainMeasurement(t, "m1")
 }
+
+// Test that if Stats isn't supplied, then we only do count, min, max, mean,
+// stdev, and s2.  We purposely exclude sum for backwards compatability,
+// otherwise user's working systems will suddenly (and surprisingly) start
+// capturing sum without their input.
+func TestBasicStatsWithDefaultStats(t *testing.T) {
```
cpacey: This might be overkill, but I like the idea of verifying "default configuration" separately from verifying "aggregations are working".

plugins/aggregators/basicstats/basicstats.go
JeffAshton: And cheapest for backwards compatibility case

JeffAshton:
Hey @danielnelson , I would love to use this feature to help simplify my continuous queries.  Do you have time to take a look?

danielnelson:
One thing I'm worried about is that we might see some weird floating point artifacts that wouldn't appear if we kept a running sum:
```
cpu value=1 1519251291871635057
cpu value=1 1519251291871803032
cpu value=1 1519251291871857634
cpu value=1 1519251291871884218
cpu value=2 1519251291871907024
cpu value=1 1519251291871931726
cpu value_sum=6.999999999999999 1519251294000000000
```

cpacey:
@danielnelson I'll switch to tracking `sum` as we go.  Would you like to keep tracking `mean` as it's done now, or would you want to switch to calculating the mean based on `sum` and `count`?  The former simplifies the code change and reduces the likelihood of changing previous behaviour; the latter simplifies the code and reduces calculation cost.

danielnelson:
Let's switch to calculating the mean based on sum and count.

cpacey:
@danielnelson I've switched to keeping a running sum (and added a test for the floating-point problem you were concerned about).  However, when I tried to remove tracking the mean, the naive approach added another floating-point divide when calculating variance (example [here](https://github.com/cpacey/telegraf/pull/1)).  Do we care about a potential performance impact of this?  I haven't measured anything yet - I wanted to get a sense of whether you'd be concerned before investing time.

If you are concerned, options I see are to continue tracking the mean, or rearranging the variance equations to avoid the extra floating-point divide, but at the cost of an extra floating-point multiplication (example [here](https://github.com/cpacey/telegraf/pull/2)).

danielnelson:
In the first example, doesn't it equal out since we can remove this divide?
```
mean = mean + delta/n
```

cpacey:
Not without changing the `stdev` calculation, which I'm not excited to do.  The [online variance algorithm](https://en.m.wikipedia.org/wiki/Algorithms_for_calculating_variance#Online_algorithm) produces its estimate by using both the _current_ mean and the _previous_ mean.  In the current code, the previous mean is tracked, and the new mean is calculated with a single divide (`mean + delta/n`).  If we stop tracking the mean, then we need to compute _both_ the previous mean and the current mean, which (naively) requires 2 divides: either `oldSum / oldCount` and `newSum / newCount`; or `oldSum / oldCount` and `oldMean + delta/n`.

(Note that I'd be very happy to be wrong about this.)

cpacey:
@danielnelson, could we keep tracking `mean` and merge as-is?

PR #2673, opened 2017-04-14
Add Traefik input plugin

mlabouardy:
### Required for all PRs:

- [x] CHANGELOG.md updated (we recommend not updating this until the PR has been approved by a maintainer)
- [x] Sign [CLA](https://influxdata.com/community/cla/) (if not already signed)
- [x] README.md updated (if adding a new plugin)

PR #3798, opened 2018-02-16
Input plugin 2673 traefik

mkboudreau:
### Required for all PRs:

This submission is a continuance of https://github.com/influxdata/telegraf/pull/2673


plugins/inputs/traefik/traefik.go
```diff
@@ -0,0 +1,138 @@
...
+	TotalStatusCodeCount   HttpCodes `json:"total_status_code_count"`
+	TotalCount             int       `json:"total_count"`
+	TotalResponseTimeSec   float64   `json:"total_response_time_sec"`
+	AverageResponseTimeSec float64   `json:"average_response_time_sec"`
+}
+
+type HttpCodes map[string]int
+
+var sampleConfig = `
+# Required Traefik server address, host and port (default: "127.0.0.1")
```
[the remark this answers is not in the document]
mkboudreau: Great question. I took over this pull request to help drive it through completion. Adding credentials was not part of the original PR and not really needed to get through the initial completion of this PR. Obviously there are people out there who may want to secure this endpoint, so that functionality should probably be added. However, since the most common use case for this kind of metric data is probably unauthenticated, it makes sense to push this through as is and submit another PR with the added functionality. 

Would you want to work on adding the ability to add credentials?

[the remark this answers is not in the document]
danielnelson: Traefik supports several methods for authentication: https://docs.traefik.io/configuration/entrypoints/.  If we add support for this (and I agree it does not need to be in the initial PR) we should do it in a way that we could configure any of the methods even if we only do HTTP Basic Auth initially.

So, if you do work on this don't use the userinfo section of the URL, instead add options like https://github.com/influxdata/telegraf/blob/master/plugins/inputs/http/http.go#L58-L60

@mkboudreau  One thing I would like to add initially though is support for TLS.  Should be pretty easy we just need to mirror the code in the http input: https://github.com/influxdata/telegraf/blob/master/plugins/inputs/apache/apache.go#L100-L115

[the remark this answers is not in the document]
jkowalleck: @mkboudreau wrote
> Would you want to work on adding the ability to add credentials?
sorry, i dont know GO, so i think i would not be helpfull.
i tried to read the mongodb input plugin code to see how they strip credentials from the server... was not able to understand it at all ;-)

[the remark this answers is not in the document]
danielnelson: @jkowalleck Are you interested in HTTP Basic Auth then?  I could add this in once it's merged pretty easily.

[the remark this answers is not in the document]
jkowalleck: @danielnelson  Since traefik supports BASIC and DIGEST i suggest having both available.
see https://docs.traefik.io/configuration/entrypoints/

my current setup uses BASIC auth.
currently using `httpjson` input plugin as following:
```toml
[[inputs.httpjson]] ## traefik
#   ## NOTE This plugin only reads numerical measurements, strings and booleans
#   ## will be ignored.

#   ## Name for the service being polled.  Will be appended to the name of the
#   ## measurement e.g. httpjson_webserver_stats
#   ##
#   ## Deprecated (1.3.0): Use name_override, name_suffix, name_prefix instead.
#   name = "webserver_stats"
  name_override = "traefik"  # overwrite target - default target is "httpjson"

#   ## URL of each server in the service's cluster
#   ## !!!! DO NOT ADD auth here - it might(will) be written to the (influx) output plugin as a tag
  servers = ["http://traefik:57475/health"]

#   ## Set response_timeout (default 5 seconds)
#   response_timeout = "5s"

#   ## HTTP method to use: GET or POST (case-sensitive)
#   method = "GET"

#   ## HTTP Headers (all values must be strings)
  [inputs.httpjson.headers]
    authorization = "Basic c3RhdHM6c3RhdHM="
```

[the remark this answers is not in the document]
danielnelson: @jkowalleck I'll add support for Basic Auth after we merge, might take a bit longer to get digest.  It will probably look more like the `http` input (successor to httpjson in 1.6).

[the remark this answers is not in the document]
mkboudreau: > @mkboudreau One thing I would like to add initially though is support for TLS. Should be pretty easy we just need to mirror the code in the http input: https://github.com/influxdata/telegraf/blob/master/plugins/inputs/apache/apache.go#L100-L115

@danielnelson I should be able to update and add TLS support. Might have to wait until at least next week.

deric:
Any update on this PR?

mkboudreau:
I was planning on doing it weeks ago, but just have not had a moment to spare. Still hoping to add in the requested changes very soon unless someone beats me to it.

plugins/inputs/traefik/traefik.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+func copyFields(m map[string]interface{}) map[string]interface{} {
+	fields := make(map[string]interface{})
+	for k, v := range m {
+		fields[k] = v
+	}
+	return fields
+}
+func copyTags(m map[string]string) map[string]string {
```
glinton: nitpick: newline between functions

plugins/inputs/traefik/traefik.go
```diff
@@ -0,0 +1,167 @@
...
+	client := &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: tlsCfg,
+		},
+		Timeout: t.ResponseTimeout.Duration,
+	}
+
+	return client, nil
+}
+func (t *Traefik) Gather(acc telegraf.Accumulator) error {
```
glinton: newline between functions

plugins/inputs/traefik/traefik.go
```diff
@@ -0,0 +1,167 @@
...
+	TotalStatusCodeCount   HttpCodes `json:"total_status_code_count"`
+	TotalCount             int       `json:"total_count"`
+	TotalResponseTimeSec   float64   `json:"total_response_time_sec"`
+	AverageResponseTimeSec float64   `json:"average_response_time_sec"`
+}
+
+type HttpCodes map[string]int
+
+var sampleConfig = `
+    # Required Traefik server address, host and port (default: "127.0.0.1")
```
glinton: Should the default be `http://127.0.0.1:8080`? If not, maybe add the default port (`8080` from line 164) to the documentation here and in the readme.

plugins/inputs/traefik/traefik.go
```diff
@@ -0,0 +1,138 @@
...
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+
+	healthData := &HealthCheck{}
+	json.NewDecoder(resp.Body).Decode(&healthData)
+	t.lastRequestTiming = time.Since(start).Seconds()
+
+	tags := map[string]string{"server": t.Address}
```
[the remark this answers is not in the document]
glinton: I think we'll get this merged and it can be extended later to support basic authentication.

glinton:
@mkboudreau do you think you'll have time soon to address the latest feedback?

plugins/inputs/traefik/README.md
```diff
@@ -0,0 +1,87 @@
...
+    - **unixtime** (int) - unix timestamp of when this service was last restarted
+    - **uptime_sec** (int) - elapsed time (in seconds) since this service was restarted
+    - **health_response_time_sec** (float64) - round trip time to gather this set of metrics
+    - **count** (int) - count for this status code since the last time this service was restarted
+
+
+### Tags:
+
+- All measurements have the following tags:
+    - **server**
```
glinton: I believe we'll want this to be `source` and slightly modified, see #4413

plugins/inputs/traefik/traefik.go
```diff
@@ -0,0 +1,169 @@
...
```
glinton: Lets call the tag `source` (rather than `server` (4413)) and only use the parsed url's `.Hostname()` as the value. Also, save this as a variable on the struct so it doesn't have to be computed each call to `Gather`.

kerams:
Are stats split by frontend/backend something that can be added or does Traefik not expose those at all?

mkboudreau:
@danielnelson I haven't checked up on this in a while. I assumed it'd already be merged. Any idea what is preventing this PR from being merged?

plugins/inputs/traefik/traefik.go
```diff
@@ -0,0 +1,169 @@
...
+	start := time.Now()
+	resp, err := t.client.Do(req)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+
+	healthData := &HealthCheck{}
+	json.NewDecoder(resp.Body).Decode(&healthData)
+	t.lastRequestTiming = time.Since(start).Seconds()
```
danielnelson: Normally we add timing information like this as a selfstat (enabled by turning on the `internal` plugin).  You can see some examples of these in the influxdb_listener plugin.  It may not be worth adding though since I imagine it will be quite close to the built in gather_time_ns value.

plugins/inputs/traefik/README.md
```diff
@@ -0,0 +1,87 @@
...
+  # [inputs.traefik.tags]
+  #  instance = "prod"
+```
+
+### Measurements & Fields:
+
+- `traefik` measurement have the following fields:
+    - **total_count** (int) - total number of responses since the application was last restarted
+    - **average_response_time_sec** (float64) - average time (in seconds) of all responses
+    - **total_response_time_sec** (float64) - total time (in seconds) of all responses
```
danielnelson: For the duration types, we are trying to standardize on reporting as an integer in nanoseconds with the suffix `_ns`.

plugins/inputs/traefik/README.md
```diff
@@ -0,0 +1,87 @@
...
+    - **unixtime** (int) - unix timestamp of when this service was last restarted
```
danielnelson: Is this description accurate?  Traefik's docs say ` // current server date in seconds`.

At any rate, for times we are standardizing on saving them as integers in nanoseconds since the epoch.  I don't think we need to call it unixtime, something descriptive of the correct definition should work: `server_time` or ``startup_time`

plugins/inputs/traefik/README.md
```diff
@@ -0,0 +1,87 @@
...
+    - **uptime_sec** (int) - elapsed time (in seconds) since this service was restarted
```
danielnelson: `uptime_ns`

plugins/inputs/traefik/README.md
```diff
@@ -0,0 +1,87 @@
...
+Simple configuration, with one measurement per request capture:
+
+```toml
+# Description
+[[inputs.traefik]]
+  ## Required Traefik server address, host and port (default: "127.0.0.1")
+  # address = "http://127.0.0.1:8080"
+  #
+  # [inputs.traefik.tags]
+  #  instance = "prod"
```
danielnelson: Remove the `tags` usage example, we don't usually show it even though it can, of course, be set on all plugins.  Just showing the default config will be enough.

plugins/inputs/traefik/README.md
```diff
@@ -0,0 +1,87 @@
...
+    - **unixtime** (int) - unix timestamp of when this service was last restarted
+    - **uptime_sec** (int) - elapsed time (in seconds) since this service was restarted
+    - **health_response_time_sec** (float64) - round trip time to gather this set of metrics
+    - Http Codes - each status code count since this service was last restarted
+	- **status_code_200** (int)
+	- **status_code_400** (int)
+	- **status_code_500** (int)
+	- *status_code_*...
+
+- `traefik_status_codes` measurement have the following fields:
```
danielnelson: The cardinality of this measurement isn't going to be much of a problem, in almost all cases we are talking several dozen codes.  Even 1000 series wouldn't really be a problem, so I wouldn't make it configurable, keep it simple and always send the status code metrics.

So with that in mind, and given the shared data between the two measurements, I think a single measurement would work okay if we did it similar to:
```
traefik,source=localhost total_count=42i,average_response_time_ns=42i
traefik,source=localhost,status_code=400 status_code_count=42i
traefik,source=localhost,status_code=500 status_code_count=42i
```
If we do it like this, I don't mind having them all under the same measurement because there is no overlap between the field names.  However if you prefer, two measurements are still okay but we should keep the shared fields on the primary measurement only:

```
traefik,source=localhost total_count=42i,average_response_time_ns=42i
traefik_status_code,source=localhost,status_code=400 count=42i
traefik_status_code,source=localhost,status_code=500 count=42i
```

plugins/inputs/traefik/traefik.go
```diff
@@ -0,0 +1,169 @@
...
+		t.submitStatusCodeMeasurement(acc, healthData, tags, fields)
+	}
+
+	return nil
+}
+
+func init() {
+	inputs.Add("traefik", func() telegraf.Input {
+		return &Traefik{
+			Address: "127.0.0.1:8080",
```
danielnelson: I believe this address needs to be a URL: `http://127.0.0.1:8080`

ssoroka:
I took a look at what it would take to get this up to date and merged in to the project. Digging into this I discovered there's a better option. Traefik now supports exporting data to InfluxDB directly, with many more stats, and we can even pipe that through Telegraf if we want, using the influxdb_listener plugin. 

For Example:

Telegraf config:
```
[[inputs.influxdb_listener]]
  service_address = ":8186"
  read_timeout = "10s"
  write_timeout = "10s"
  max_body_size = "500MiB"
  max_line_size = "64KiB"
```

Traefik config:
```
traefik --metrics.influxdb=true \
  --metrics.influxdb.address=telegraf.server.location:8186 \
  --metrics.influxdb.protocol=http
```

Going to close this now as there's nothing else to do.

PR #3800, opened 2018-02-16
Add new ldap_response input plugin

mkboudreau:
### Required for all PRs:

New simple plugin for ldap servers to capture timing of a basic connect, bind, and simple search. Basic concept is similar to the http_response plugin, while some connection code is borrowed from the openldap plugin.


mkboudreau:
@danielnelson I haven't checked up on this in a while. I assumed it'd already be merged. Any idea what is preventing this PR from being merged?

glinton:
@mkboudreau I was trying to remember why you were assigned, I believe it was because the tests weren't passing. Update your branch and re-push and we'll give it a review.

mkboudreau:
@glinton Was trying to get this wrapped up for you tonight. I ran into an odd circleci failure that I don't understand what it is trying to do. The script `scripts/check-deps.sh` is reporting an error and spitting out the following to the console. Any help would be appreciated.
```
+gopkg.in/ldap.v2
```

p-zak:
@mkboudreau You need to add this dependency to `LICENSE_OF_DEPENDENCIES.md` (together with license type and link to licence)

sjwang90:
@mkboudreau - If you're still interested in this plugin, can you please address @p-zak's comment above?

sjwang90:
@mkboudreau please let us know if you are no longer interested in working in this plugin and am looking for someone else to take it over.

mkboudreau:
@sjwang90 I have addressed every "ask" in the 3 year history of this plugin. Can you tell me what the hold up is? If it is something on my end, I do not know what it is.

sjwang90:
I missed your updates from @p-zak's comments. I thought they were still missing. We'll get this on the backlog of reviewed items.

plugins/inputs/ldap_response/ldap_response.go
```diff
@@ -0,0 +1,209 @@
...
+	"time"
+
+	"gopkg.in/ldap.v2"
+
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/plugins/common/tls"
+	"github.com/influxdata/telegraf/plugins/inputs"
+)
+
+type Ldap struct {
```
srebhan, line 17: Can we use `LDAP` here? This style looks quite strange to me... :-) @p-zak will the linter shout at us if we use `LDAP`?

plugins/inputs/ldap_response/ldap_response.go
```diff
@@ -0,0 +1,209 @@
...
+	Host               string
+	Port               int
+	Ssl                string
+	InsecureSkipVerify bool
+	SslCa              string
+	BindDn             string
+	BindPassword       string
+	SearchBase         string
+	SearchFilter       string
+	SearchAttributes   []string
```
srebhan, line 27: Please add the `toml` tags to the user-configurable options!

plugins/inputs/ldap_response/ldap_response.go
```diff
@@ -0,0 +1,209 @@
...
+  port = 389
+
+  # ldaps, starttls, or no encryption. default is an empty string, disabling all encryption.
+  # note that port will likely need to be changed to 636 for ldaps
+  # valid options: "" | "starttls" | "ldaps"
+  ssl = ""
+
+  # skip peer certificate verification. Default is false.
+  insecure_skip_verify = false
+
```
srebhan, line 41: Please comment all options that are holding the default and don't need to be touched by the user by default. Furthermore, comments should start with a double-hash.
```suggestion
  # host = "localhost"
  # port = 389

  ## ldaps, starttls, or no encryption. default is an empty string, disabling all encryption.
  ## note that port will likely need to be changed to 636 for ldaps
  ## valid options: "" | "starttls" | "ldaps"
  # ssl = ""

  ## skip peer certificate verification. Default is false.
  ## insecure_skip_verify = false

```
Not sure about the `host` and `port`, but you should only leave options uncommented that _have_ to be set and do not have a sensible default.

Same for all other lines below.

plugins/inputs/ldap_response/ldap_response.go
```diff
@@ -0,0 +1,209 @@
...
+  # the attributes to return as fields. defaults to "objectclass" if unspecified.
+  search_attributes = [
+    "attribute1",
+    "attribute2",
+  ]
+`
+
+var DefaultSearchFilter = "(objectClass=*)"
+var DefaultSearchAttributes = []string{"objectclass"}
```
srebhan, line 63: Please do not export those and make at least `DefaultSearchFilter` a constant.

plugins/inputs/ldap_response/ldap_response.go
```diff
@@ -0,0 +1,209 @@
...
+	return "LDAP Response Input Plugin"
+}
+
+// return an initialized Ldap
+func NewLdap() *Ldap {
+	return &Ldap{
+		Host: "localhost",
+		Port: 389,
+	}
+}
```
srebhan, line 79: Please avoid this function and construct the struct instance directly at the caller's location.

plugins/inputs/ldap_response/ldap_response.go
```diff
@@ -0,0 +1,209 @@
...
+				return nil
+			}
+			err = server.StartTLS(tlsConfig)
+		} else {
+			acc.AddError(fmt.Errorf("Invalid setting for ssl: %s", l.Ssl))
+			return nil
+		}
+	} else {
+		server, err = ldap.Dial("tcp", fmt.Sprintf("%s:%d", l.Host, l.Port))
+	}
```
srebhan, line 117: This is initialization work and should be moved to an `Init() error` function called automatically by telegraf after the user config options are parsed and set in the plugin struct.

plugins/inputs/ldap_response/ldap_response.go
```diff
@@ -0,0 +1,209 @@
...
+	afterConnect := time.Now()
+
+	if err != nil {
+		acc.AddError(err)
+		return nil
+	}
```
srebhan, line 123: Please move this error checking to the correct spot!

plugins/inputs/ldap_response/ldap_response.go
```diff
@@ -0,0 +1,209 @@
...
+
+	fields := map[string]interface{}{
+		"connect_time_ms": float64(afterConnect.Sub(beforeConnect).Nanoseconds()) / 1000 / 1000,
+		"bind_time_ms":    float64(afterBind.Sub(beforeBind).Nanoseconds()) / 1000 / 1000,
+		"query_time_ms":   float64(afterSearch.Sub(beforeSearch).Nanoseconds()) / 1000 / 1000,
+		"total_time_ms":   float64(afterSearch.Sub(beforeConnect).Nanoseconds()) / 1000 / 1000,
+	}
+
+	gatherSearchResult(fields, sr, l, acc)
```
srebhan, line 172: You might move the function body here and rather move the individual measurement steps to functions...

plugins/inputs/ldap_response/ldap_response.go
```diff
@@ -0,0 +1,209 @@
...
+
+	return nil
+}
+
+func gatherSearchResult(fields map[string]interface{}, sr *ldap.SearchResult, l *Ldap, acc telegraf.Accumulator) {
```
srebhan, line 177: If you really _want_ to do this, `acc` has to be the first argument. Furthermore, please do not pass the plugin `Ldap` struct as parameter but rather make this function a member of the plugin.

plugins/inputs/ldap_response/ldap_response.go
```diff
@@ -0,0 +1,209 @@
...
+	tags := map[string]string{
+		"server": l.Host,
+		"port":   strconv.Itoa(l.Port),
+	}
+	for _, entry := range sr.Entries {
+		metricName := dnToMetric(entry.DN, l.SearchBase)
```
srebhan, line 183: The variable name is very confusing. From the code I guess it is a `fieldPrefix`?

plugins/inputs/ldap_response/ldap_response.go
```diff
@@ -0,0 +1,209 @@
...
+		for _, attr := range entry.Attributes {
+			if len(attr.Values[0]) >= 1 {
+				if v, err := strconv.ParseInt(attr.Values[0], 10, 64); err == nil {
+					fields[metricName+attr.Name] = v
+				}
+			}
```
srebhan, line 189:
```suggestion
			if len(attr.Values[0]) == 9 {
				continue
			}
			if v, err := strconv.ParseInt(attr.Values[0], 10, 64); err == nil {
				fields[metricName+attr.Name] = v
			}
```

plugins/inputs/ldap_response/ldap_response_test.go
```diff
@@ -0,0 +1,148 @@
+package ldap_response
+
+import (
+	"fmt"
+	"strconv"
+	"testing"
+
+	"github.com/influxdata/telegraf/testutil"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
```
srebhan, line 10: Please use `require` only if possible.

srebhan (commented):
Hey @mkboudreau,

thanks for the PR and sorry for this taking so long! :-( I have some comments in the code but my main question is why you don't fold this into the `openldap` plugin which is _very_ similar code-wise... As far as I can see, the main differences are the measured times and the different search parameters. Both features should be easy to add and we reduce the maintenance effort of two almost identical plugins and save one dependency to an older version of the ldap client.

Can you please outline what prevents you from adding the features to the openldap plugin!?

mkboudreau:
Hello @srebhan,

I must say this review caught me off-guard a bit after 3 years. Almost all the suggestions you have would also apply to existing plugins in the current codebase. This plugin was copied mostly from openldap, which is already in there. Some of your suggestions are good and I agree with them, others I think are a little bit of personal style (e.g. dont use`NewXXX()`) especially when there are already several instances in the codebase. However, regardless of my personal opinion, I just don't have the time or context anymore to do all these requested corrections. 

I will answer your question about merging with openldap. I did look to do that first (> 3 years ago at my former company); however, we were not using openldap. You'll notice the scope of that plugin does one thing, it gathers metrics from an internal openldap location. It is very simple. Nothing in that plugin was usable to us. We simply wanted a generic ldap heartbeat (i.e. connect, bind, search, disconnect) with measurements. I do not see what value would be added by overloading openldap with generic ldap stuff and then having users who do not use openldap "hope" their heartbeat functionality works with _their_ specific ldap server. I felt it was more clear to live on its own. 

@sjwang90 I guess you can remove this plugin or find someone else. I cannot speak for other contributors, but my experience here has been less than great.

sjwang90:
@mkboudreau Sorry for your less than great contribution experience. Standalone plugins that have limited or very specific use cases do fall back in our priority and we could have communicated that to your earlier. We have added [external plugins](https://github.com/influxdata/telegraf/blob/master/docs/EXTERNAL_PLUGINS.md) for these types of plugins to allow contributors to bypass the review process and use their plugins with [execd](https://github.com/influxdata/telegraf/tree/master/plugins/common/shim#telegraf-execd-go-shim).

Hopefully your experience will be better in the future. You can always reach out in the #telegraf channel in our [Community Slack](influxdata.com/slack) if you would like to chat with the developers or community directly.

srebhan:
Hey @mkboudreau,

first of all let me say that I'm sorry about your decision not to pursuing this further and about the long time it took to give this PR a review.

However, as you directly addressed your comment at me, let me answer. I know that it is bad to wait such a long time (I've experienced this myself) but I review telegraf PRs only since 6 month or so and I'm not employed by InfluxData. As I'm doing this completely in my spare-time, I think it's a bit unfair to complain to me about your experience.

> [...] Almost all the suggestions you have would also apply to existing plugins in the current codebase. This plugin was copied mostly from openldap, which is already in there. Some of your suggestions are good and I agree with them, others I think are a little bit of personal style (e.g. dont use`NewXXX()`) especially when there are already several instances in the codebase. 

Well, you submitted a _new_ plugin, so I review this from scratch not taking into account any other code-base. It might be that existing code is also weak in this respect, but "we've done this like this before" is a not an argument and we should strive for better code wherever we can.
Furthermore, if you remember my comments, having an almost identical code-base was my argument to fold this into the existing plugin. 

> [...] I will answer your question about merging with openldap. I did look to do that first (> 3 years ago at my former company); however, we were not using openldap. You'll notice the scope of that plugin does one thing, it gathers metrics from an internal openldap location. It is very simple. Nothing in that plugin was usable to us. We simply wanted a generic ldap heartbeat (i.e. connect, bind, search, disconnect) with measurements. I do not see what value would be added by overloading openldap with generic ldap stuff and then having users who do not use openldap "hope" their heartbeat functionality works with _their_ specific ldap server. I felt it was more clear to live on its own.

You also have to see the telegraf side, having > 200 plugins is a maintenance burden and adding almost identical ones instead of fusing them is only increasing this burden. And you yourself state above that the existing plugin has an almost identical code base. If you compare the code side-by-side (which I spent quite some time on during my review) you will see that your plugin differs by a) measuring the timings and b) by different search parameters! I think that's a pretty strong point for merging the two...

> However, regardless of my personal opinion, I just don't have the time or context anymore to do all these requested corrections.

Again, I can fully understand your frustration, but I do think that my review was thorough, fair and not personally attacking you. Furthermore, I still think that I have valid points and if you look at my other reviews I'm open to discuss stuff like keeping the `New...` functions etc. So I think addressing those comments to me is not fair in any way.

mkboudreau:
> Again, I can fully understand your frustration, but I do think that my review was thorough, fair and not personally attacking you.

@srebhan I do think your review was thorough and fair (so thank you), and I never felt personally attacked (please don't confuse frustration around the process with feeling personally attacked). 

I think if a thorough review was done 3+ years ago and not by the 6th person to come along, I'd be open to having more of a dialog on the kinds of things you brought up. At this point, I don't recall enough to answer all your questions, and I don't even work for the company who needed it anymore. Honestly, my expectation was that this was just about to be merged; hence, my frustration.

srebhan:
@mkboudreau no hard feelings from my side.
Again I understand your frustration and I agree that this is not what it is supposed to be, but I cannot change what was two or three years ago. Nonetheless, InfluxData tries to improve the process _now_ (or rather starting some month ago) involving more community members. So maybe, after the first frustration settled, you can submit the features with another PR. I really would like to see your use-case covered! Feel free to directly assign me then and I will answer in a timely manner! Promise! ;-)

PR #3809, opened 2018-02-19
Add Oracle input plugin

bobmshannon:
- Add a new input plugin that reads metrics from an Oracle database.

If this input plugin is to be considered, I will need to fix the build so the new integration tests run:
- [ ] address https://github.com/sparrc/gdm/issues/26
- [ ] update gcc compiler in build to 7.x. for ODPI-C
- [ ] install instant client for integration tests

### Required for all PRs:


danielnelson:
Thanks for opening this pull request.  As you probably know, the build issues around gcc and CGO are going to be sticking points for getting this included into an official release.

I don't have a detailed list of what we need to start including more C dependencies, but if we can solve these issues we would be close:
- No C compiler required (we should just expect odpi-c to already be compiled)
- No requirements for have specific C header files on the build system (optional is okay so long as no action required by default)
- Everything works with `CGO_ENABLED=0`

I should also mention that we may decide to push these types of changes over into a grpc based plugin system.

BTW there is also this approach to loading an Oracle driver: https://github.com/influxdata/telegraf/pull/2785/files#diff-c9862cb8a6c0bb799256ffe4cb790095R11

bobmshannon:
Thanks for taking a look. I'm going to close this as it sounds like a change like this will be a better fit once the RPC based plugin system is implemented. The pluggable driver looks really interesting, but   complicates deployability. Further, without modifying the build we won't be able to setup integration testing which is another reason why I think longer term this is a better fit once the RPC based plugin system is in place.

PR #3818, opened 2018-02-22
Added float datatype support for metrics and aerospike latency metrics.

shubham340:
### Required for all PRs:

Depenedency on RequestNodeLatency function in the aerospike go client.
https://github.com/aerospike/aerospike-client-go/pull/224


plugins/inputs/aerospike/aerospike.go
```diff
@@ -89,13 +94,22 @@ func (a *Aerospike) gatherServer(hostport string, acc telegraf.Accumulator) erro
 			}
 		}
 		acc.AddFields("aerospike_node", fields, tags, time.Now())
+		//Finding the latency metrics
+		infoLatency, err := as.RequestNodeInfo(n, "latency:")
```
danielnelson: I think it should be possible to get the latency at the same time as the namespaces in a single RequestNodeInfo call, can you look into this as it should be quite a bit more performant.

shubham340: Yaa, We can get the namespace names from the 'latency:' , but from the 'latency:' I got lesser namespace names as compared to 'namespaces'.
screen shot 2018-03-06 at 11 19 16 am
Black ones shows the namespaces obtained from 'latency:' whereas the green one is from 'namespaces'.

shubham340: Latency histogram initialization is lazy in nature. It won't exist if we have not done any operation in the namespace. Once we run read or write it will get created and will be shown then on. Hence the namespace names can't be fetched from 'latency:' appropriately.

shubham340:
Removed DS_Store and made new measurement for latency metric because it causes type conflict for *_pct fields for previously created measurements.

srebhan:
@shubham340 are you still interested in pushing this? If so, could you please rebase on current master and I will give it a review!?

srebhan:
Seems like @shubham340 is not interested anymore. If anyone is interested in pushing this forward, please step-up!

sjwang90:

PR #3829, opened 2018-02-25
Add new SSL input plugin

mgrabazey:
### Required for all PRs:

A new simple plug-in for checking the time to expiration of the ssl certificate and its correspondence to the domain

### DIFF
--- a/plugins/inputs/ssl/ssl.go
+++ b/plugins/inputs/ssl/ssl.go
@@ -1,0 +1,128 @@
+package ssl
+
+import (
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/plugins/inputs"
+	"crypto/x509"
+	"time"
+	"net"
+	"crypto/tls"
+	"github.com/pkg/errors"
+	"strings"
+)
+
+type Ssl struct {
+	Servers []Server
+}
+
+type Server struct {
+	Host string
+	Timeout int
+}
+
+var sampleConfig  = `
+  ## Server to check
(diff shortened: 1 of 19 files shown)


danielnelson:
@mgrabazey Can you take a look at this similar pull request and let me know if it would fulfill your use? #3768

mgrabazey:
This implementation does not check whether the certificate matches the domain on which the certificate is used

plugins/inputs/ssl/ssl.go
```diff
@@ -0,0 +1,128 @@
...
+
+func getServerCertsChain(d string, p string, t int) ([]*x509.Certificate, error) {
+	h := getServerAddress(d, p)
+	ipConn, err := net.DialTimeout("tcp", h, time.Duration(t)*time.Second)
+	if err != nil {
+		return nil, errors.New("[" + h + "] " + err.Error())
+	}
+	defer ipConn.Close()
+
+	tlsConn := tls.Client(ipConn, &tls.Config{ServerName: d, InsecureSkipVerify: true})
```
phemmer: If we're going to set `InsecureSkipVerify: true`, we should probably do a manual verification step and indicate the status. It's possible for an intermediate or root cert to be revoked (or expired in the case of a stupid CA), which would invalidate the cert.

danielnelson:
@mgrabazey We did merge the alternate plugin for certificate checking, and I think we should be able to integrate the certificate verification pieces into that plugin.  I opened a feature request for this addition #4877.

Issue #3670, opened 2018-01-13
Add general aggregations into Metric interface


yantang-msft:
I'm trying to write a output plugin to Azure Application Insights, which supports aggregated value natively in its datapoint.
The problem is for an aggregated metric, I won't know which aggregator emitted it. Thus the conversion logic can be funky. It also doesn't scale if a new aggregator is added.
So I'd like to have the Metric interface expose those aggregations (count, min, max, quantiles...) along with the IsAggregate() method.

danielnelson:
Why is it important to know which aggregator emitted the metric?

yantang-msft:
@danielnelson The App Insights has its own aggregated telemetry defined. https://github.com/Microsoft/ApplicationInsights-Go/blob/master/appinsights/telemetry.go#L194
I'll have to convert the aggregated metric from telegraf to App Insights' format. Otherwise, the backend of App Insights won't be able to understand the metric correctly.

Currently, to convert the aggregated metric from telegraf to App Insights, I'll need to parse each field based on the format of the aggregator plugins. But as an output plugin, I shouldn't need to know what aggregator is used, given count, min, max... are general aggregations.

danielnelson:
What if you just treat, for example, any field named `"*_min"` as a `AggregateMetricTelemetry` with a Min field without checking if it is `IsAggregate()`?  I can imagine there could be some false positives but I think that would be rare, and it would allow the output to work without knowledge of the source.

yantang-msft:
@danielnelson Yes, I can do that, but as you said, it's kind of unreliable.
Also, it requires all aggregator plugins' field follows the same pattern. And in this case, I'd rather have the aggregator plugins conform to the pattern explicitly by having these field strongly typed.

danielnelson:
Ultimately someone would need to guess in many cases as well, such as when the data originates upstream from a queuing system via a plugin like `kafka_consumer`.

We currently have something similar in the `Type()` function which is used by the prometheus output.  It has a few shortcomings, primarily that it applies to the entire Metric.  I am working on some changes to the Metric interface for 1.6, and I hope to rework this aspect, so I'll see if I can provide what you need as part of this.

danielnelson:
Adding this support on a per field basis while not taking a performance hit way is proving to be a bit more work and won't make it in for 1.6.

srebhan:
@yantang-msft are you still interested in this matter?

PR #3834, opened 2018-02-26
JSON serializer emits metric type information

amoghe:
JSON serializer now includes type information in the string it emits.

Also riding on this patch - ValueType now exposes a String()
method that returns the name of the ValueType. This is useful
when we print the metric type information (e.g. when serializing)
since the reader does not need to be aware of the enum types used
within telegraf.

### Required for all PRs:

- [ :heavy_check_mark: ] Signed [CLA](https://influxdata.com/community/cla/).
- [ :x: ] Associated README.md updated (<-- not needed)
- [ :heavy_check_mark: ] Has appropriate unit tests.

### DIFF
--- a/metric.go
+++ b/metric.go
@@ -17,6 +17,20 @@ const (
 	Histogram
 )
 
+var (
+	ValueTypeToName = map[ValueType]string{
+		Counter:   "counter",
+		Gauge:     "gauge",
+		Untyped:   "untyped",
+		Summary:   "summary",
+		Histogram: "histogram",
+	}
+)
+
+func (v ValueType) String() string {
+	return ValueTypeToName[v]
+}
+
 type Metric interface {
 	// Serialize serializes the metric into a line-protocol byte buffer,
 	// including a newline at the end.
--- a/plugins/serializers/json/json.go
+++ b/plugins/serializers/json/json.go
@@ -23,6 +23,7 @@ func (s *JsonSerializer) Serialize(metric telegraf.Metric) ([]byte, error) {
 	m["fields"] = metric.Fields()
 	m["name"] = metric.Name()
 	m["timestamp"] = metric.UnixNano() / units_nanoseconds
+	m["type"] = metric.Type().String()
 	serialized, err := ejson.Marshal(m)
 	if err != nil {
 		return []byte{}, err
--- a/plugins/serializers/json/json_test.go
+++ b/plugins/serializers/json/json_test.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/stretchr/testify/assert"
 
+	"github.com/influxdata/telegraf"
 	"github.com/influxdata/telegraf/metric"
 )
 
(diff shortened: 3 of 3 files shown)


danielnelson:
I think we shouldn't emit a type field if the metric is untyped.

amoghe:
> I think we shouldn't emit a type field if the metric is untyped.

I see your point. But don't you think that its better to have a json schema that is consistent (i.e. the reader can expect the json to have a consistent set of keys) rather than some keys missing sometimes?

If I were to write a parser to parse these messages, I'd want this behavior because otherwise I'd be writing a branch in my code to handle the `untyped` case when the key is missing.

Thoughts?

danielnelson:
It probably depends on what language/library you are using to parse if it helps or not.  I will say that most of the metrics in Telegraf are currently untyped because it is only used by the prometheus output.

danielnelson:
I've been thinking about this more and I don't think we should make this change, the main reason is that we want to make changes that will allow setting a type per field, instead of the current model of one type per Metric.  This would fix some issues with splitting metrics across lines that otherwise should logically remain together.

amoghe:
Could you point me to any documentation/writing that could explain this new scheme? (one-metric-many-values). I don't think I follow what you are suggesting.

danielnelson:
I'll try to explain here.  

Using line protocol, a metric can contain multiple values like:
```
cpu time_user=42 usage_user=20
```

However, there is no guarantee that all of the values are of the same type, in the above example `usage_user` is a gauge and `time_user` is a counter.  Currently we split these across metrics so that we can assign different types:
```
cpu time_user=42
cpu usage_user=20
```

In line protocol, which is our primary serialization format, there is actually no reason they can't be on the same line.  This is more efficient in terms of space and parsing speed, and also makes processing the data easier since the related values are all on one line.  A concrete example of these difficulties is reported in https://github.com/influxdata/kapacitor/issues/1217.

Because of this, in the future I'd like to be able to express this as a single metric/line, while still allowing outputs such as the prometheus output to have the required metadata.  There are also suggestions to add additional types of metadata that would be on a per field basis. (#3670)

I think what would be nice is if we had a new JSON output format that instead of having a hardcoded "shape", would allow the user to specify how they would like the data structured and what data, including metadata, would be included.  I'm not sure the exact details of how this would work, so that remains to be fleshed out.

danielnelson:
While I don't consider the Metric interface to be part of the versioned API, if we add the type to JSON output then we do need to support it until the next major version.  While this is not a disqualifier, I'm hesitant to add this when I know it will be changed.

That said, I had hoped to add per field types to 1.6, but it turns out doing this without using a lot more memory is a bit trickier than I anticipated and right now it is not a priority.  So I think it would be reasonable to move forward with this pull request.

metric.go
```diff
@@ -17,6 +17,20 @@ const (
...
+		Counter:   "counter",
+		Gauge:     "gauge",
+		Untyped:   "untyped",
+		Summary:   "summary",
+		Histogram: "histogram",
+	}
+)
+
+func (v ValueType) String() string {
+	return ValueTypeToName[v]
```
danielnelson: If ValueType is not in the map don't set a type, or optionally set "untyped".

danielnelson:
Closing due to lack of activity, let us know if you want to continue to work on this and we can reopen.

Issue #2667, opened 2017-04-13
Feature Request - "Transform" Processor plugin


PauloAugusto-Asos:
## Feature Request

Requesting a "Transform" processor plugin.

I am trying to import Web access logs into InfluxDB with Telegraf. However, some of the URL PATHs include identifiers _(session IDs, product IDs, etc)_. Ex:
`/products/cars/12345/view`
`/shoppingBasket/1234567890/view`

The URL PATH is being shipped as a Tag Value _(obviously)_. I need to to be able to replace those identifiers from the PATH Tag Value before shipping the data to Influx _(or whatever other DB)_ so that they become easily recognizable as the «same» URL PATH for searches and aggregations and to prevent an explosion of "series" in InfluxDB or Graphite.

### Proposal:
**`[[processors.transformer]]`**
`    tagpass = "ApacheLog"`
`    tagname = "path"`
`    matcher = "/products/cars/(\d+)/view/"`
`    matchertype = "regex" # "literal"`
`    replaceMatchedIndex = 1 # i0 being whole match. To replace *only* the ID`
`    replacement = "{CarID}"`
`    tagexclude = "ApacheLog"`
**`[[processors.transformer]]`**
`    tagpass = "ApacheLog"`
`    tagname = "path"`
`    matcher = "/shoppingBasket/(\\d+)/view"`
`    matchertype = "regex" # literal`
`    replaceMatchedIndex = 1`
`    replacement = "{SessionID}"`
`    tagexclude = "ApacheLog"`

### Simpler Proposal:
**`[[processors.transformer]]`**
`    tagpass = "ApacheLog"`
`    tagname = "path"`
`    matcher = "/products/cars/\\d+/view/"`
`    matchertype = "regex" # "literal"`
`    # replaceMatchedIndex = 1`
`    replacement = "/products/cars/{CarID}/view/"`
`    tagexclude = "ApacheLog"`

### SimplerSimpler Proposal:
**`[[processors.transformer]]`**
`    tagpass = "ApacheLog"`
`    tagname = "path"`
`    replaceDigits = 3 # replace all sequences of X+ digits`
`    replaceGuids = true`
`    replaceTrimmedGuids = true # guids stripped of dashes`
`    tagexclude = "ApacheLog"`

danielnelson:
I like it.  First version seems best, I think we need captures.  I suggest we change a few names:

- `matchertype` -> `transformation`
- `matcher` -> `regex_pattern`
- `literal` -> `replace`

[...]

PR #3839, opened 2018-02-26
Add regex processor plugin

44px:
Based on proposal from @danielnelson in https://github.com/influxdata/telegraf/issues/2667#issuecomment-301641261

Regex processor plugin allows to change tag and field values or create new tags/fields from existing ones.

My use-case for this plugin: it allows to use a simple predefined pattern such as `%{COMBINED_LOG_FORMAT}` in `logparser` plugin then do some processing to extract additional data from request field.

### Configuration:

```toml
[[processors.regex]]
  namepass = ["nginx_requests"]

  [[processors.regex.tags]]
    key = "resp_code"
    pattern = "^(\\d)\\d\\d$"
    replacement = "${1}xx"

  [[processors.regex.fields]]
    key = "request"
    pattern = "^/api(?P<method>/[\\w/]+)\\S*"
    replacement = "${method}"
    result_key = "method"

  [[processors.regex.fields]]
    key = "request"
    pattern = ".*category=(\\w+).*"
    replacement = "${1}"
    result_key = "search_category"
```

### Source Metric (from logparser):
```
nginx_requests,verb=GET,resp_code=200 request="/api/search/?category=plugins&q=regex&sort=asc",referrer="-",ident="-",http_version=1.1,agent="UserAgent",client_ip="127.0.0.1",auth="-",resp_bytes=270i 1519652321000000000
```

### Example Output:
```
nginx_requests,verb=GET,resp_code=2xx request="/api/search/?category=plugins&q=regex&sort=asc",method="/search/",search_category="plugins",referrer="-",ident="-",http_version=1.1,agent="UserAgent",client_ip="127.0.0.1",auth="-",resp_bytes=270i 1519652321000000000
```

### Required for all PRs:


syepes:
very nice

voiprodrigo:
Would be possible to use this to convert a field k/v pair to a tag, or extract tag name and value from partial field name (while renaming the field name too)?

44px:
Thanks for your interest!

As for now, this plugin operates only on values, so it can't convert tag to field or change its name.

I thought about some option which allows to convert between tag and field but later decided it would be better to have separate plugin for this. Such plugin can have more obvious name – "tag2field" instead of too general "regex" – so it would be easier to find it.

> extract tag name and value from partial field name (while renaming the field name too)

I'm not sure I properly understand your use case here. Can you give example of field name you have and what tag you want to create from it?

voiprodrigo:
> extract tag name and value from partial field name (while renaming the field name too)

>> I'm not sure I properly understand your use case here. Can you give example of field name you have and what tag you want to create from it?

Actually, what I was thinking would require splitting the metric point into multiple ones, which should probably be handled by the plugin itself. Or be a stand-alone processor maybe.

Let's say a metric has field names which denote a thread index in a multi-threaded process: 

`measurement,pname=whatever thread0_valueA=0,thread1_valueA=0,thread0_valueB=1,thread1_valueB=2`

This would translate to:

´measurement,pname=whatever,thread=0 valueA=0,valueB=1
measurement,pname=whatever,thread=1 valueA=0,valueB=2`

So this would be something more or less like:
- Match all fields named ^(thread)(\d+)_(.+)$
- Create new metric points with new tag \1=\2
- Matched fields renamed to \3

I implement something like this as an option to an existing input plugin (targeting specific field names created by the plugin), so I was wondering if your plugin could be used as a generic way to the same kind of thing.

danielnelson:
@voiprodrigo This seems like a fairly specialized operation, I'm not sure if it can be generalized nicely.  I think for this it would make sense if we had a processor that could run a user script to filter messages.  To avoid slowdown we would run the processor once and feed data in/out via stdin/stdout.

plugins/processors/regex/regex.go
```diff
@@ -0,0 +1,105 @@
...
+
+func getKey(c converter) string {
+	if c.ResultKey != "" {
+		return c.ResultKey
+	}
+	return c.Key
+}
+
+func getValue(c converter, value string) string {
+	regex := regexp.MustCompile(c.Pattern)
```
danielnelson: We should compile the regex pattern only once on startup, you could do this in Apply with a boolean field on the Regex type so it is only executed once.

plugins/processors/regex/regex.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+func (r *Regex) Description() string {
+	return "Transforms tag and field values with regex pattern"
+}
+
+func (r *Regex) Apply(in ...telegraf.Metric) []telegraf.Metric {
+	for _, metric := range in {
+		for _, converter := range r.Tags {
+			if value, ok := metric.Tags()[converter.Key]; ok {
```
danielnelson: Use `metric.GetTag` to avoid allocating.

plugins/processors/regex/regex.go
```diff
@@ -0,0 +1,105 @@
...
+				metric.AddTag(
+					getKey(converter),
+					getValue(converter, value),
+				)
+			}
+		}
+
+		for _, converter := range r.Fields {
+			if value, ok := metric.Fields()[converter.Key]; ok {
```
danielnelson: Use `metric.GetField` here as well.

44px:
Thanks for the tips! I applied `GetTag`/`GetField` and added cache for compiled regex patterns.

Here are results of running `go test -bench=. -benchmem` for this package:

Before:
```
pkg: github.com/influxdata/telegraf/plugins/processors/regex
BenchmarkConversions-4            100000             14945 ns/op           11192 B/op        147 allocs/op
PASS
ok      github.com/influxdata/telegraf/plugins/processors/regex 1.676s

```

After:
```
pkg: github.com/influxdata/telegraf/plugins/processors/regex
BenchmarkConversions-4            500000              2326 ns/op             632 B/op         25 allocs/op
PASS
ok      github.com/influxdata/telegraf/plugins/processors/regex 1.209s
```

PR #3650, opened 2018-01-09
Linux_Wireless Plugin

davidgs:
Gather basic stats on linux-based wireless interfaces.

### Required for all PRs:

- [X ] Signed [CLA](https://influxdata.com/community/cla/).
- [X ] Associated README.md updated.
- [X ] Has appropriate unit tests.

PR #3847, opened 2018-02-28
Add Wireless Input plugin

jamesmaidment:
### Required for all PRs:

The wireless plugin gathers metrics about wireless link quality by reading the `/proc/net/wireless` file. This plugin currently supports linux only.


plugins/inputs/wireless/wireless.go
```diff
@@ -0,0 +1,168 @@
...
+	// split the lines by newline
+	lines := bytes.Split(table, newLineByte)
+	// iterate over intefaces
+	for i := 2; i < len(lines); i = i + 1 {
+		if len(lines[i]) == 0 {
+			continue
+		}
+		fields := strings.Fields(string(lines[i]))
+		var values []int64
+		for i := 1; i < len(fields); i = i + 1 {
```
glinton: use another variable name like `k` instead of `i` for this internal loop

plugins/inputs/wireless/wireless.go
```diff
@@ -0,0 +1,168 @@
...
+			v, err := strconv.ParseInt(strings.Trim(fields[i], "."), 10, 64)
```
danielnelson: No need to strings.Trim because strings.Fields handles this.

plugins/inputs/wireless/wireless.go
```diff
@@ -0,0 +1,168 @@
...
+
+// Wireless is used to store configuration values.
+type Wireless struct {
+	ProcNetWireless string `toml:"proc_net_wireless"`
+}
+
+var sampleConfig = `
+  ## file paths for proc files. If empty default paths will be used:
+  ##    /proc/net/wireless
+  proc_net_wireless = "/proc/net/wireless"
```
danielnelson: Change this option to `# host_proc = "/proc"` and if set, use the environment variable `HOST_PROC`.  CHeck the bond input for an example.  This will make it work closer to the existing plugins including those that use gopsutil.

glinton:
@jamesmaidment do you think you can update so the tests pass then we can merge this

glinton:
Similar to #3650

jamesmaidment:
@glinton fixed now

plugins/inputs/wireless/wireless_linux.go
```diff
@@ -0,0 +1,157 @@
...
+func loadWirelessTable(table []byte) ([]*wirelessInterface, error) {
+	var w []*wirelessInterface
+	lines := bytes.Split(table, newLineByte)
+
+	// iterate over interfaces
+	for i := 2; i < len(lines); i = i + 1 {
+		if len(lines[i]) == 0 {
+			continue
+		}
+		values := []int64{}
```
danielnelson: Might as well initialize capacity to the expected length, since it is known.

plugins/inputs/wireless/wireless_linux.go
```diff
@@ -0,0 +1,157 @@
...
+		fields := strings.Fields(string(lines[i]))
+		for j := 1; j < len(fields); j = j + 1 {
+			v, err := strconv.ParseInt(strings.Trim(fields[j], "."), 10, 64)
+			if err != nil {
+				return nil, err
+			}
+			values = append(values, v)
+		}
+		w = append(w, &wirelessInterface{
```
danielnelson: Check expectations of the length of values, just in case the input data is not as expected.  If it's wrong, log and skip this line.

jamesmaidment: Resolved

plugins/inputs/wireless/wireless_linux.go
jamesmaidment: Resolved

PR #3861, opened 2018-03-06
graphite output module: "close_connect" option has added

ezhk:
### Required for all PRs:

### DIFF
--- a/plugins/outputs/graphite/README.md
+++ b/plugins/outputs/graphite/README.md
@@ -19,6 +19,8 @@ via raw TCP.
   template = "host.tags.measurement.field"
   ## timeout in seconds for the write connection to graphite
   timeout = 2
+  ## close connect every time, when data was sent
+  close_connect = false
 
   ## Optional SSL Config
   # ssl_ca = "/etc/telegraf/ca.pem"
@@ -30,10 +32,11 @@ via raw TCP.
 
 Parameters:
 
-    Servers  []string
-    Prefix   string
-    Timeout  int
-    Template string
+    Servers      []string
+    Prefix       string
+    Timeout      int
+    CloseConnect bool
+    Template     string
 
     // Path to CA file
     SSLCA string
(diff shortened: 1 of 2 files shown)


danielnelson:
Could you add a comment about why this feature was needed.

ezhk:
The first reason is when the metrics send with long intervals, minutes, so we do not need to keep the connection and its state. And also some daemons, that accept metrics, may wait for the FIN flag after receiving events; of course, this is the wrong behavior for these demons, but it's good if the module supports this case.

danielnelson:
Can you give some concrete example of daemons that require the connection to be closed?

ezhk:
I met it in self-written customer code, it wasn't open source.

danielnelson:
Can you tell me more about why you don't want to keep the connection open?  It seems to me that several connections is probably not enough resources to worry about especially if you will just need to reopen them after a few minutes.  Is it just the resource usage?

ezhk:
Several connections use not enough resources on client side, because, in main cases, clients are pushing data less than to 10 servers. But on server side it is not the same.

danielnelson:
That's a good point, thank you for helping me think this through.  Shouldn't it be the servers responsibility to close idle connections in this case?

ezhk:
Yep, it’s server repsonsibility. But look how does it works with current code.

Basic data for our example: client push events every 5 minutes, server has connection timeout 2 minutes.

Client open connection and push metrics — on this step all work correctly. After this step client does nothing. Server will be waiting for for 2 minutes and send FIN — as a trying to close connection correct. Client send ACK flag and that’s all! Client, by this code, doesn’t send FIN from their side, before next iteration to send data.

ezhk:
How else can I help?

danielnelson:
I think we need to redo the connection handling in this plugin to solve this issue more directly.  Perhaps we need to setup a goroutine on connection to wait for the remote side to close and call close on our side.  I'd also like to replace the checkEOF code because I don't think that it handles things correctly.

PR #3876, opened 2018-03-09
Add stackdriver output plugin

jamesmaidment:
### Required for all PRs:

Adds Google Stackdriver as an output for telegraf metric data, saved as a custom metric in Stackdriver


danielnelson:
Test still failed, very odd, when I pushed your branch to our repo it worked: https://circleci.com/gh/influxdata/telegraf/10024

jamesmaidment:
Might be a permissions issue? I'm contacting Circle CI to submit a bug report at the moment.

danielnelson:
I just opened an issue with CircleCI.

danielnelson:
CircleCI let me know they fixed an issue that should address the test failure.

plugins/outputs/stackdriver/stackdriver.go
```diff
@@ -0,0 +1,236 @@
...
+			},
+		}, nil
+	default:
+		return nil, fmt.Errorf("the value type \"%T\" is not supported for custom metrics", v)
+	}
+}
+
+// Close will terminate the session to the backend, returning error if an issue arises.
+func (s *GCPStackdriver) Close() error {
+	return nil
```
danielnelson: Do we need to close the client?

plugins/outputs/stackdriver/stackdriver.go
```diff
@@ -0,0 +1,236 @@
...
+							},
+						},
+						Points: []*monitoringpb.Point{
+							dataPoint,
+						},
+					},
+				}}
+
+			// Create the time series in Stackdriver.
+			err = s.client.CreateTimeSeries(ctx, timeSeries)
```
danielnelson: Curious how the performance is here, if this is indeed one rpc call per field it could be quite a few calls.  I also worry about the expensive of the code above at a per field basis.  What has your testing shown?

plugins/outputs/stackdriver/README.md
```diff
@@ -0,0 +1,18 @@
+# Stackdriver Output Plugin
+
+This plugin writes to the [Google Cloud Stackdriver API](https://cloud.google.com/monitoring/api/v3/)
+and requires [authentication](https://cloud.google.com/docs/authentication/getting-started) with Google Cloud using either a service account or user credentials.
+
+Requires `project` to specify where Stackdriver metrics will be delivered to.
+
+Metrics are grouped by the `namespace` variable and metric key - eg: `custom.googleapis.com/telegraf/system/load5`
+
```
danielnelson: Should we suggest increasing the agent `flush_interval`?


> Best practice: When adding multiple data points to the same time series, do not add them faster than once per minute. If you are adding data points to different time series, then there is no rate limitation.
>
> https://cloud.google.com/monitoring/custom-metrics/creating-metrics#writing-ts

danielnelson: We should probably mention the costs and link to https://cloud.google.com/stackdriver/pricing

plugins/outputs/stackdriver/stackdriver.go
```diff
@@ -0,0 +1,236 @@
...
+	// MaxInt is the max int64 value.
+	MaxInt = int(^uint(0) >> 1)
+)
+
+var sampleConfig = `
+  # GCP Project
+  project = "erudite-bloom-151019"
+
+  # The namespace for the metric descriptor
+  namespace = "telegraf"
```
danielnelson: I think we should add a timeout that would apply per call to CreateTimeSeries.

plugins/outputs/stackdriver/stackdriver.go
```diff
@@ -0,0 +1,236 @@
...
+							dataPoint,
+						},
+					},
+				}}
+
+			// Create the time series in Stackdriver.
+			err = s.client.CreateTimeSeries(ctx, timeSeries)
+			if err != nil {
+				log.Printf("E! Error writing to output [stackdriver]: %s", err)
+				continue
```
danielnelson: This will drop the metric, this may be reasonable because returning an error will fail the entire batch, and it would need restarted from the beginning.

As an aside, In the future we may add an immediate flush output type, it seems like this plugin would be a good candidate to use it.

Godeps
```diff
@@ -83,14 +84,18 @@ github.com/wvanbergen/kazoo-go 968957352185472eacb69215fa3dbfcfdbac1096
 github.com/yuin/gopher-lua 66c871e454fcf10251c61bf8eff02d0978cae75a
 github.com/zensqlmonitor/go-mssqldb ffe5510c6fa5e15e6d983210ab501c815b56b363
 golang.org/x/crypto dc137beb6cce2043eb6b5f223ab8bf51c32459f4
-golang.org/x/net f2499483f923065a842d38eb4c7f1927e6fc6e6d
+golang.org/x/net cbe0f9307d0156177f9dd5dc85da1a31abc5f2fb
```
danielnelson: Is the version change to `golang.org/x/net` required?

Godeps
```diff
@@ -17,18 +18,18 @@ github.com/eapache/go-resiliency b86b1ec0dd4209a588dc1285cdd471e73525c0b3
...
 github.com/gogo/protobuf 7b6c6391c4ff245962047fc1e2c6e08b1cdfa0e8
 github.com/golang/protobuf 8ee79997227bf9b34611aee7946ae64735e6fd93
 github.com/golang/snappy 7db9049039a047d955fe8c19b83c8ff5abd765c7
-github.com/go-ole/go-ole be49f7c07711fcb603cff39e1de7c67926dc0ba7
 github.com/google/go-cmp f94e52cad91c65a63acc1e75d4be223ea22e99bc
-github.com/gorilla/mux 53c1911da2b537f792e7cafcb446b05ffe33b996
-github.com/go-redis/redis 73b70592cdaa9e6abdfcfbf97b4a90d80728c836
-github.com/go-sql-driver/mysql 2e00b5cd70399450106cec6431c2e2ce3cae5034
+github.com/googleapis/gax-go 317e0006254c44a0ac427cc52a0e083ff0b9622f
+github.com/gorilla/mux 392c28fe23e1c45ddba891b0320b3b5df220beea
```
danielnelson: Is the version change to `github.com/gorilla/mux` required?

Godeps
```diff
@@ -83,14 +84,18 @@ github.com/wvanbergen/kazoo-go 968957352185472eacb69215fa3dbfcfdbac1096
...
+google.golang.org/grpc 583a6303969ea5075e9bd1dc4b75805dfe66989a
 gopkg.in/asn1-ber.v1 4e86f4367175e39f69d9358a5f17b4dda270378d
 gopkg.in/fatih/pool.v2 6e328e67893eb46323ad06f0e92cb9536babbabc
 gopkg.in/gorethink/gorethink.v3 7ab832f7b65573104a555d84a27992ae9ea1f659
 gopkg.in/ldap.v2 8168ee085ee43257585e50c6441aadf54ecb2c9f
 gopkg.in/mgo.v2 3f83fa5005286a7fe593b055f0d7771a7dce4655
 gopkg.in/olivere/elastic.v5 3113f9b9ad37509fe5f8a0e5e91c96fdc4435e26
 gopkg.in/tomb.v1 dd632973f1e7218eb1089048e0798ec9ae7dceb8
-gopkg.in/yaml.v2 4c78c975fe7c825c6d1466c42be594d1d6f3aba6
+gopkg.in/yaml.v2 4c78c975fe7c825c6d1466c42be594d1d6f3aba6
```
danielnelson: Add back the newline :)

plugins/outputs/stackdriver/stackdriver.go
```diff
@@ -0,0 +1,236 @@
...
+
+	for _, m := range metrics {
+		for k, v := range m.Fields() {
+			value, err := getStackdriverTypedValue(v)
+			if err != nil {
+				log.Printf("E! Error writing to output [stackdriver]: %s", err)
+				continue
+			}
+
+			metricKind, err := getStackdriverMetricKind(telegraf.Histogram)
```
danielnelson: Why do you pass telegraf.Histogram in here?  I was expecting metric.Type()?

Godeps
jamesmaidment: I'm unable to build the project using the existing `golang.org/x/net` version - seems the update is required by `google.golang.org/grpc`

Godeps
jamesmaidment: Not required, I have reverted this change

vishen:
Is this still being actively worked on?  I am just wondering, because, this feature would be super useful for me, and it hasn't seen much change for a while.

vishen:
@danielnelson I ran the build locally, and it seems to run fine. The only improvement I can see, is, if instead of sending one `dataPoint` per timeseries, a list of `dataPoint`s is built up and sent at the end, that would limit the number of rpc calls from `one-per-metric` to `one-per-metrictype` (or one per `telegraf.Metric.Name`), but I don't know if telegraf aggregates the metrics before it gets to the output plugins?

Otherwise seems reasonable enough to me.

danielnelson:
>  if telegraf aggregates the metrics before it gets to the output 

No aggregation is done unless an aggregator is used or the input aggregates (statsd input)

jamesmaidment:
I've been using this plugin in production since March, let me know if there's any other changes required so we can get this merged.

danielnelson:
@jamesmaidment We will need to update the pr to use dep instead of gdm, but in addition to that I see there are still a few comments that are unanswered from the previous review, can you take a look and respond to them?

jamesmaidment:
@danielnelson which comments are unanswered? Sorry I'm a bit slow 😄

danielnelson:
Can you check through the comments on this review and reply to all of them?  I assume some of them have already been addressed so just write "done" or such.  I think we may need to discuss performance a little bit https://github.com/influxdata/telegraf/pull/3876#discussion_r185380065

https://github.com/influxdata/telegraf/pull/3876#pullrequestreview-116755429

jamesmaidment:
Thanks @danielnelson will try and address the performance concerns and close out those final items

glinton:
@jamesmaidment Hows the progress coming on those performance concerns and whatnot?

plugins/outputs/stackdriver/stackdriver.go
glinton: I agree, build out a `[]*monitoringpb.TimeSeries` and submit it at the end of the loop.

plugins/outputs/stackdriver/stackdriver.go
```diff
@@ -0,0 +1,233 @@
...
+
+// Write the metrics to Google Cloud Stackdriver.
+func (s *Stackdriver) Write(metrics []telegraf.Metric) error {
+	ctx := context.Background()
+
+	for _, m := range metrics {
+		for k, v := range m.Fields() {
+			value, err := getStackdriverTypedValue(v)
+			if err != nil {
+				log.Printf("E! Error writing to output [stackdriver]: %s", err)
```
glinton: Personally, I would change these error strings to describe what failed. Eg: `E! [output.stackdriver] get type failed: %s`

jamesmaidment:
@glinton I should have the requested updates from @danielnelson and yourself completed over the next few days for review

plugins/outputs/stackdriver/stackdriver.go
jamesmaidment: Updated now

jamesmaidment:
@glinton @danielnelson I've resolved those requests -- let me know if there is anything else needed

plugins/outputs/stackdriver/stackdriver.go
```diff
@@ -0,0 +1,293 @@
...
+func getStackdriverMetricKind(vt telegraf.ValueType) (metricpb.MetricDescriptor_MetricKind, error) {
+	switch vt {
+	case telegraf.Untyped:
+		return metricpb.MetricDescriptor_GAUGE, nil
+	case telegraf.Gauge:
+		return metricpb.MetricDescriptor_GAUGE, nil
+	case telegraf.Counter:
+		return metricpb.MetricDescriptor_CUMULATIVE, nil
+	case telegraf.Histogram, telegraf.Summary:
+		fallthrough
```
danielnelson: Can we send these as GAUGE?

plugins/outputs/stackdriver/stackdriver.go
```diff
@@ -0,0 +1,293 @@
...
+		// Prepare time series request.
+		timeSeriesRequest := &monitoringpb.CreateTimeSeriesRequest{
+			Name:       monitoring.MetricProjectPath(s.Project),
+			TimeSeries: timeSeries,
+		}
+
+		// Create the time series in Stackdriver.
+		err := s.client.CreateTimeSeries(ctx, timeSeriesRequest)
+		if err != nil {
+			log.Printf("E! [output.stackdriver] create time series failed: %s", err)
```
danielnelson: I know we talked about this earlier, but I didn't really get a sense of why we don't return an error here.  If we return an error then Telegraf will treat all the data as not written and will try it again on the next `flush_interval`, without an error the data will be lost.

plugins/outputs/stackdriver/stackdriver.go
```diff
@@ -0,0 +1,293 @@
...
+func (s *Stackdriver) Write(metrics []telegraf.Metric) error {
+	ctx := context.Background()
+
+	for _, m := range metrics {
+		timeSeries := []*monitoringpb.TimeSeries{}
+
+		for k, v := range m.Fields() {
+			value, err := getStackdriverTypedValue(v)
+			if err != nil {
+				log.Printf("E! [output.stackdriver] get type failed: %s", err)
```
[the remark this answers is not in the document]
jamesmaidment: Added string value instead

plugins/outputs/stackdriver/stackdriver.go
danielnelson: @jamesmaidment What do you think about this item?

jamesmaidment: @danielnelson Stackdriver does support a Distribution type that could work for one or both of these https://cloud.google.com/monitoring/api/ref_v3/rest/v3/TimeSeries#distribution

I haven't had a lot of cycles to look into it and so far haven't had this use case. I'm not sure if I can get to this immediately and could add it as a patch later?

danielnelson: Yeah we can add proper support later.  Do you think we should send them as GAUGE values for now or do you think it is better to omit them as it is now?

jamesmaidment: I think perhaps we should omit them for now and add proper support for Histogram & Summary using the Distribution type in a future release just to not cause having to change the behavior in future if people come to rely on the Gauge metric types.

danielnelson:
@jamesmaidment Can you do one final merge/rebase to clear up the merge conflict?

jamesmaidment:
@danielnelson should be good now

PR #3877, opened 2018-03-09
Fix ping crash on Enimga2 linux distribution (mipsle)

marianob85:
### Required for all PRs:

Ping command crash on enimga2 stb ( linux mipsle ). 
- Fix problem with output parsing
- Fix problem with arguments


plugins/inputs/ping/ping.go
```diff
@@ -171,14 +171,24 @@ func (p *Ping) args(url string) []string {
 	// Build the ping command args based on toml config
 	args := []string{"-c", strconv.Itoa(p.Count), "-n", "-s", "16"}
 	if p.PingInterval > 0 {
-		args = append(args, "-i", strconv.FormatFloat(p.PingInterval, 'f', 1, 64))
```
danielnelson: I think we can just change this to use the -1 precision:
```
args = append(args, "-i", strconv.FormatFloat(p.PingInterval, 'f', -1, 64))
```

It still will require users to specify an even value, but I think it is better than switching on arch.

plugins/inputs/ping/ping.go
```diff
@@ -171,14 +171,24 @@ func (p *Ping) args(url string) []string {
...
+			args = append(args, "-i", strconv.FormatFloat(p.PingInterval, 'f', 1, 64))
+		}
+
 	}
 	if p.Timeout > 0 {
 		switch runtime.GOOS {
 		case "darwin":
 			args = append(args, "-W", strconv.FormatFloat(p.Timeout*1000, 'f', 1, 64))
 		case "linux":
-			args = append(args, "-W", strconv.FormatFloat(p.Timeout, 'f', 1, 64))
```
danielnelson: Can use -1 precision here too

plugins/inputs/ping/ping_test.go
```diff
@@ -41,6 +41,18 @@ PING www.google.com (216.58.218.164) 56(84) bytes of data.
 rtt min/avg/max/mdev = 35.225/43.628/51.806/5.325 ms
 `
 
+var enigma2PingOutput = `
```
danielnelson: Add a comment here with the output of `ping -V` (or the equivalent version option).

marianob85:
Honestly it is a problem because of BusyBox and I don't think this is a right way of fixing it. I make a condition based on mipsle compilation (GOARCH) but I think it should be done different. 
Maybe user should specify which arguments are supported and what is a prefix of them ? Maybe You have some idea.

marianob85:
This is how ping command look like on BusyBox

BusyBox v1.24.1 (2017-02-28 03:28:13 CET) multi-call binary.

Usage: ping [OPTIONS] HOST

Send ICMP ECHO_REQUEST packets to network hosts

        -4,-6           Force IP or IPv6 name resolution
        -c CNT          Send only CNT pings
        -s SIZE         Send SIZE data bytes in packets (default:56)
        -t TTL          Set TTL
        -I IFACE/IP     Use interface or IP address as source
        -W SEC          Seconds to wait for the first response (default:10)
                        (after all -c CNT packets are sent)
        -w SEC          Seconds until ping exits (default:infinite)
                        (can exit earlier with -c CNT)
        -q              Quiet, only display output at start
                        and when finished
        -p              Pattern to use for payload

marianob85:
Maybe add new parameter, something like 'target= busybox' ?

danielnelson:
I like the idea to introduce a variable where you can pick a dialect, we did something similar for procstat plugin recently to support Windows.  On ping, we also want to support a pure Go implementation in addition to the current method of exec'ing another command.

I think all of that would be a moderate amount of work though, as it's probably best to just start from scratch on the implementation.

plugins/inputs/ping/ping.go
```diff
@@ -171,17 +171,23 @@ func (p *Ping) args(url string) []string {
 	// Build the ping command args based on toml config
 	args := []string{"-c", strconv.Itoa(p.Count), "-n", "-s", "16"}
 	if p.PingInterval > 0 {
-		args = append(args, "-i", strconv.FormatFloat(p.PingInterval, 'f', 1, 64))
+		switch runtime.GOARCH {
+		case "mipsle":
+			// -i argument not supported in busybox
```
danielnelson: Let's not add anything for this then, if the ping_interval is set to "0s" it will not add this option.

danielnelson:
Can you take a look at the failed testcase?

marianob85:
Ping is already fixed but tests crash somewhere on statsd

plugins/inputs/ping/ping.go
```diff
@@ -268,7 +271,7 @@ func init() {
 	inputs.Add("ping", func() telegraf.Input {
 		return &Ping{
 			pingHost:     hostPinger,
-			PingInterval: 1.0,
+			PingInterval: 0.0,
```
danielnelson: Sorry about the delay merging, should/can we change this back to 1.0?

Issue #3879, opened 2018-03-13
Feature Request: add config flag to ignore network protocol metrics


datwiz:
### Proposal:
Add a config flag on inputs.net system plugin to skip collecting network protocol metrics.

### Current behavior:
On Linux, the system net plugin reports network protocol metrics in addition to interface specific metrics.

### Desired behavior:
Enable the collection of protocol metrics to be ignored or skipped.
The default behaviour maintains current functionality.  That is protocol metrics are reported.
By setting an `ignore_protocol_stats` flag to `true` will result in protocol metrics not being reported.  Only interface specific metrics are emitted.

### Use case: [Why is this important (helps with prioritizing requests)]
With metrics collection services that charge based on numbers of metrics collected per host, being able to filter out metrics that aren't critical to operations enables better management of metrics volume.

PR #3880, opened 2018-03-13
added config flag to skip collection of network protocol metrics

datwiz:
Added config flag to system net (inputs.net) to ignore collection of protocol stats.
Provides feature requested in #3879

### Required for all PRs:

### DIFF
--- a/plugins/inputs/system/NET_README.md
+++ b/plugins/inputs/system/NET_README.md
@@ -14,6 +14,11 @@ This plugin gathers metrics about network interface and protocol usage (Linux on
   ##
   # interfaces = ["eth*", "enp0s[0-1]", "lo"]
   ##
+  ## On linux systems telegraf also collects protocol stats.
+  ## Setting ignore_protocol_stats to true will skip reporting of protocol metrics.
+  ##
+  # ignore_protocol_stats = false
+  ##
 ```
 
 ### Measurements & Fields:
--- a/plugins/inputs/system/net.go
+++ b/plugins/inputs/system/net.go
@@ -14,8 +14,9 @@ type NetIOStats struct {
 	filter filter.Filter
 	ps     PS
 
-	skipChecks bool
-	Interfaces []string
+	skipChecks          bool
+	IgnoreProtocolStats bool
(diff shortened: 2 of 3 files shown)


danielnelson:
Would you be interested in chaining down a list of protocols to the underlying call to gopsutil?  Then we would have an option like `protocols = ["udp", "tcp"]` which eventually passes the list down to:
```
func ProtoCounters(protocols []string) ([]ProtoCountersStat, error) {
```

datwiz:
I can have a go at that.  2 questions:
1.  Would you see a `protocols` set option being used in place of an ignore_protocol_stats flag or in addition to a flag?
2.  If the set option is in addition to the ignore flag, would you like the work added into this PR or raising a new one?

danielnelson:
I guess we would need to have the boolean as well, otherwise we still can't disable them all and keep backwards compatibility.  I was kind of hoping to avoid adding a boolean option here, since they often don't compose well as the plugin grows.

Let's scratch the idea of the `protocols` array, users can just use the fielddrop filtering option to achieve the same results.

datwiz:
My apologies. I'm new to telegraf and hadn't fully appreciated the tag and filtering capabilities.
In my case, all the protocol stats are attributed to the `interface=all` tag.
I've been able to achieve the same filtering, without a new config flag using a tag drop section.
```
# # Read metrics about network interface usage
[[inputs.net]]
  ## By default, telegraf gathers stats from any up interface (excluding loopback)
  ## Setting interfaces will tell it to gather these explicit interfaces,
  ## regardless of status.
  ##
  interfaces = ["eth0"]
  [inputs.net.tagdrop]
    interface = [ "all" ]
```
At least for my use case, this config flag isn't required.

danielnelson:
It hadn't occurred to me either that it would be so easy to filter.  I think I will revert this change then since it is always better to have a smaller interface and we haven't had any other requests for this.

zabudskyi:
Would you be interested in adding option to gather stats from "interface to Internet" only? It is useful in case there are many hosts with different interface naming, numbering, etc.

PR #3892, opened 2018-03-16
Add disk input plugin ignore_mount_points property

Dark0096:
### disk input plugin
If we use tmpfs file system, metrics for unnecessary mount points are also collected. 

### Feature
I have added the ignore_mount_points property to ignore unnecessary mount point metric.

### Test Code
Wrote the test code.

### Required for all PRs:


Dark0096:
I would like to be able to detect the disk full by collecting metrics for a part of tmpfs through the merge request.

Dark0096:
Your comments is very helpful.
https://github.com/influxdata/telegraf/blob/master/docs/CONFIGURATION.md#input-config-tagpass-and-tagdrop

Thanks a lot.

Issue #272, opened 2015-10-16
Add integration to service discovery


pauldix:
If there's some standard service discovery to connect to like Consul, it would be cool to have Telegraf connect to that and automatically start collecting data for services that Telegraf supports.

So when a new MySQL server comes on, Telegraf will automatically start collecting data from it.

Just an idea. Users could also get this by just having Telegraf part of their deploys when they create new servers.

sparrc:
:+1:

rvrignaud:
Hello @pauldix,
What you suggest is, I think, what I tried to explain here: https://github.com/influxdb/telegraf/issues/193#issuecomment-140688471
Prometheus supports a wide range of discovery (consul included). I'm personally interested in kubernetes discovery.

titilambert:
@pauldix @rvrignaud see PR about etcd here : https://github.com/influxdata/telegraf/pull/651

chris-zen:
Hi @titilambert, your PR is really useful to update telegraf configuration dynamically, such as changing input and outputs configurations from time to time, but for service discovery in a system such as AWS, mesos or kubernetes where things scale dynamically, something like the service discovery features implemented in prometheus would be really great.

@rvrignaud explanation is [here](https://github.com/influxdata/telegraf/issues/193#issuecomment-140688471), and the [prometheus documentation](https://prometheus.io/docs/operating/configuration/#consul_sd_config) shows the different possibilities supported.

Having this feature would definitively make me move to influxdb, but keep using the prometheus instrumentation library.

titilambert:
@chris-zen that's very interesting !
I'm agree with you, I would love to see it, but this king of service discovery is more for scheduled (polling) monitoring systems (like Prometheus), isn't it ? I'm not sure that a pushing system like Telegraf is adapted to this...

What do you think about?

chris-zen:
Yes, agree that it is specially important for polling. But telegraf is already supporting polling inputs such as the one for prometheus. Right now the prometheus input only allows static config, but it would be very useful to support service discovery too. My understanding is that telegraf is quite versatile and allows both pull and push models, but the pull model without service discovery is worthless in such dynamic environments.

sparrc:
Just dropping this here for reference on what I think is a good service discovery model (from prometheus): https://prometheus.io/blog/2015/06/01/advanced-service-discovery/. Same as mentioned above but I think this blog post is a little more approachable than their documentation.

[...]

PR #3901, opened 2018-03-18
add scraping for Prometheus endpoint in Kubernetes

sorenmat:
This will allow users to add Prometheus annotations to pods
in Kubernetes, and then have telegraf scan for them, and add them
to the list of endpoints to collect metrics from.

### Required for all PRs:


sorenmat:
While reading #272 @abraithwaite @danielnelson might be interested in this.

plugins/inputs/prometheus/kubernetes.go
```diff
@@ -0,0 +1,101 @@
...
+func discoverAnnotations(s *v1.Service, p *Prometheus) {
+	url := scrapeURL(s)
+	if url != nil {
+		p.lock.Lock()
+		p.KubernetesServices = append(p.KubernetesServices, *url)
+		p.lock.Unlock()
+	}
+}
+
+func scrapeURL(s *v1.Service) *string {
```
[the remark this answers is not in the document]
sorenmat: So we can do a nil check on the value. if nil then no `prometheus.io/scrape` was present or set to false

sorenmat:
create a docker image with the change in sorenmat/telegraf:2 if you need to test it easily :)

lpic10:
@sorenmat I'm giving a try, but the scrape fails to resolve services running in different namespaces, as the discovery is not adding the namespace name to the url. 

In another subject, shouldn't the URL for the k8s API server be configurable ?

sorenmat:
I will address the first issue 😃
I think the second one should be done in another PR. There are many options/features you can add if you think about it, once the basic feature is merged people can add more features IMO

sorenmat:
I've done a bit of redesign of this:
Scraping pods instead of services, because that's what you want. it doesn't make sense to scrape services since they are load balanced, and you then only get a sample of each pod under the service.
Add some tags to the metrics so you can aggregate them together on the metric server.

Tried to add  scraping to 500 pods in our test system, didn't add much load to the telegraf process, so it seems like it scales quite well.

lpic10:
Yes, that makes sense, I thought that was the case already.

Do you have a docker image with these changes?

sorenmat:
@lpic10 yes sorenmat/telegraf:4

lpic10:
Seems something is not working, I don't see anything in logs related to k8s/prometheus and no collection is done at all.

sorenmat:
@lpic10 did you move the annotation from the service object to the pod object in your yaml files?

lpic10:
Ok, maybe what I was wanting for was for the k8s discovery to find the endpoints related to a service and retrieve metrics from all of them, and not directly discovering the pods. 

I think prometheus k8s service discovery can do both, maybe someone with experience on it could comment.

sorenmat:
@lpic10 Yeah, after googling kubernetes yaml files, it does seem like people are adding the annotation on the service, which I guess is making sense.
I will try to update it, to scan for services with the scrape annotation and the find the services pods and collect metrics form those.

sorenmat:
Looking more into this, and thinking about it.
I think adding the annotations to the pods make the most sense, it is the pods that you are scraping data from, and not the service. By adding it on the pod vs service you also get the benefit of scraping pods without a service, which could be useful in many cases.

abraithwaite:
I think that makes perfect sense as well.

plugins/inputs/prometheus/prometheus.go
```diff
@@ -244,8 +268,18 @@ func (p *Prometheus) gatherURL(u URLAndAddress, acc telegraf.Accumulator) error
...
 
+// Start will start the Kubernetes scraping if enabled in the configuration
+func (p *Prometheus) Start(a telegraf.Accumulator) error {
+	if p.KubernetesScraping {
+		return start(p)
+	}
+	return nil
+}
+
+func (p *Prometheus) Stop() {}
```
danielnelson: Need to stop the kubernetes pod monitor.  Easiest way to test this is by sending a SIGHUP, the config could be completely different after reload.

plugins/inputs/prometheus/prometheus.go
```diff
@@ -244,8 +268,18 @@ func (p *Prometheus) gatherURL(u URLAndAddress, acc telegraf.Accumulator) error
...
+
 func init() {
 	inputs.Add("prometheus", func() telegraf.Input {
-		return &Prometheus{ResponseTimeout: internal.Duration{Duration: time.Second * 3}}
+		return &Prometheus{ResponseTimeout: internal.Duration{Duration: time.Second * 3}, lock: &sync.Mutex{}}
```
danielnelson: Nitpick: can you wrap this line?

plugins/inputs/prometheus/prometheus.go
```diff
@@ -157,15 +187,6 @@ func (p *Prometheus) Gather(acc telegraf.Accumulator) error {
 	return nil
 }
 
-var tr = &http.Transport{
```
danielnelson: Thank you!

plugins/inputs/prometheus/prometheus.go
```diff
@@ -48,6 +57,12 @@ var sampleConfig = `
 
   ## An array of Kubernetes services to scrape metrics from.
   # kubernetes_services = ["http://my-service-dns.my-namespace:9100/metrics"]
+  
+  # Scrape Kubernetes pods for prometheus annotations.
+  # prometheus.io/scrape: Enable scraping for this pod
+  # prometheus.io/path: If the metrics path is not /metrics, define it with this annotation.
+  # prometheus.io/port: If port is not 9102 use this annotation
+  # kubernetes_scraping = true
```
danielnelson: `kubernetes_scraping` seems too ambiguous, maybe: `monitor_kubernetes_pods`?

plugins/inputs/prometheus/prometheus.go
```diff
@@ -109,11 +125,25 @@ func (p *Prometheus) GetAllURLs() ([]URLAndAddress, error) {
 
 		allURLs = append(allURLs, URLAndAddress{URL: URL, OriginalURL: URL})
 	}
+	p.lock.Lock()
+	defer p.lock.Unlock()
+	// loop through all pods scraped via the prometheus annotation on the pods
+	for _, pod := range p.KubernetesPods {
+		URL, err := url.Parse(pod.url)
+		if err != nil {
+			log.Printf("prometheus: Could not parse url %s, skipping it. Error: %s", pod.url, err)
```
danielnelson: I see that this is based on some existing log messages, but could you do me a favor and update them to look more like:
```
"E! [inputs.prometheus] could not parse URL %q: %v"`

File: plugins/inputs/prometheus/kubernetes.go [hunk 7]
```diff
@@ -0,0 +1,116 @@
...
+	)
+
+	go controller.Run(wait.NeverStop)
+	return nil
+}
+
+func registerPod(pod *v1.Pod, p *Prometheus) {
+	url := scrapeURL(pod)
+	if url != nil {
+		log.Printf("Will scrape metrics from %v\n", *url)
```
danielnelson: For these log at debug level: "`D! [inputs.prometheus] adding %q to scrape targets`"

plugins/inputs/prometheus/kubernetes.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+func unregisterPod(pod *v1.Pod, p *Prometheus) {
+	url := scrapeURL(pod)
+	if url != nil {
+		p.lock.Lock()
+		defer p.lock.Unlock()
+		log.Printf("Registred a delete request for %v in namespace '%v'\n", pod.Name, pod.Namespace)
+		var result []Target
+		for _, v := range p.KubernetesPods {
```
danielnelson: I think we should use a `map[string]bool` for the targets

plugins/inputs/prometheus/prometheus.go
```diff
@@ -109,11 +125,25 @@ func (p *Prometheus) GetAllURLs() ([]URLAndAddress, error) {
...
+			continue
+		}
+		podURL := p.AddressToURL(URL, URL.Hostname())
+		allURLs = append(allURLs, URLAndAddress{URL: podURL, Address: URL.Hostname(), OriginalURL: URL, Tags: pod.tags})
```
danielnelson: If we merge `URLAndAddress` with `Target` then we can do this logic in kubernetes.go and all we need to do here is add the items to the lists.

sorenmat:
I'm not 100% sure why the binary size would matter that much. I can see that it 'feels' bad, but if it doesn't eat more memory or CPU does it really matter that much?

I do think that re-implementing a client, that continuously watches the K8s API and knows about the auth features and can parse the various JSON formats will be a huge undertaking.
I also doubt it would result in a smaller binary to be honest, but it might :)
But we would have to play catch up with the k8s API all the time.
IMHO I don't think that's worth it.

sorenmat:
I just found this https://github.com/ericchiang/k8s
Seems like it might solve the problem, will try to re-implement this feature with that library to see if it helps with the size issue :)

danielnelson:
Thanks @sorenmat, I know for many the binary size is not an issue but for some embedded use cases it is a growing problem.

sorenmat:
@danielnelson This is ready for another review, when you have the time :)

plugins/inputs/prometheus/kubernetes.go
```diff
@@ -0,0 +1,161 @@
...
+		for k, v := range pod.GetMetadata().GetLabels() {
+			tags[k] = v
+		}
+		URL, err := url.Parse(*targetURL)
+		if err != nil {
+			log.Printf("E! [inputs.prometheus] could not parse URL %q: %v", targetURL, err)
+			return
+		}
+		podURL := p.AddressToURL(URL, URL.Hostname())
+		p.kubernetesPods = append(p.kubernetesPods, URLAndAddress{URL: podURL, Address: URL.Hostname(), OriginalURL: URL, Tags: tags})
```
danielnelson: Wrap this line and/or create URLAndAddress on separately.

plugins/inputs/prometheus/kubernetes.go
```diff
@@ -0,0 +1,161 @@
...
+		if err != nil {
+			log.Fatal(err)
+		}
+	}
+	type payload struct {
+		eventype string
+		pod      *corev1.Pod
+	}
+	in := make(chan payload)
+	go func() {
```
danielnelson: I think this goroutine may be leaked on close, how do we exit it?

plugins/inputs/prometheus/kubernetes.go
```diff
@@ -0,0 +1,161 @@
...
+		var pod corev1.Pod
+		watcher, err := client.Watch(context.Background(), "", &pod)
```
danielnelson: I think we need logic to reconnect the watch based on this: https://godoc.org/github.com/ericchiang/k8s#Client.Watch

danielnelson: I wasn't able to find this in the library documentation, does the watch send all pods initially?

plugins/inputs/prometheus/kubernetes.go
```diff
@@ -0,0 +1,161 @@
...
+	if err := yaml.Unmarshal(data, &config); err != nil {
+		return nil, fmt.Errorf("unmarshal kubeconfig: %v", err)
+	}
+	return k8s.NewClient(&config)
+}
+
+func start(p *Prometheus) error {
+	client, err := k8s.NewInClusterClient()
+	if err != nil {
+		client, err = loadClient(fmt.Sprintf("%v/.kube/config", os.Getenv("HOME")))
```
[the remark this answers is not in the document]
sorenmat: I do believe that the `k8s.NewInClusterClient()`does the correct thing to access the API.
Do you want me to add a kubeconfig file option to the TOML config file?

plugins/inputs/prometheus/kubernetes.go
sorenmat: Ran a test on our cluster and it started registering all pods. So yes :)

plugins/inputs/prometheus/kubernetes.go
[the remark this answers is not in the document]
danielnelson: Yes

plugins/inputs/prometheus/kubernetes.go
sorenmat: ok, will investigate. Thanks for the feedback

glinton:
@sorenmat Do you think you'll have time soon to address the most recent feedback items?

russorat:
fixing this via https://github.com/influxdata/telegraf/pull/4920

PR #3919, opened 2018-03-21
DHT22 temperature and humidity sensor plugin

gdunstone:
### Required for all PRs:

- [?] Has appropriate unit tests.

### DIFF
--- a/Godeps
+++ b/Godeps
@@ -17,7 +17,7 @@ github.com/eapache/go-resiliency b86b1ec0dd4209a588dc1285cdd471e73525c0b3
 github.com/eapache/go-xerial-snappy bb955e01b9346ac19dc29eb16586c90ded99a98c
 github.com/eapache/queue 44cc805cf13205b55f69e14bcb69867d1ae92f98
 github.com/eclipse/paho.mqtt.golang aff15770515e3c57fc6109da73d42b0d46f7f483
-github.com/gdunstone/go-dht d2c3f07368ef5cc78796d11563c5ba10bd1c4b70
+github.com/gdunstone/go-dht 9f2ed0c2c42357c4bde974d83239e50fbd6bacb4
 github.com/go-logfmt/logfmt 390ab7935ee28ec6b286364bba9b4dd6410cb3d5
 github.com/go-sql-driver/mysql 2e00b5cd70399450106cec6431c2e2ce3cae5034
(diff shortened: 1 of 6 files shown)


nerzhul:
excellent ! i have the same setup at home on a RPI with a python script i develop recently :D 
one unacceptable thing on your PR is to make sensor writeable for everybody instead of just setting the good rights (for example, can you try a setup with telegraf in root group, or try to do priviledge escalation) ?

gdunstone:
Originally that was what I was doing, create gpio group, add telegraf user to it and make /sys/class/gpio group writeable.
Im doing some more bughunting on it because there is some issue with the go-dht package that causes it to behave erratically when loaded as part of telegraf (I think it has to do with being run as a goroutine.
The original package used cgo to achieve fast enough reads.

I was also going to add a feature to support dht11 as well.

gdunstone:
it looks like this person ran into the same problem I am running into:
https://groups.google.com/forum/#!topic/golang-nuts/KFNYywZtL_4

gdunstone:
Due to issues with goroutines not being able to poll gpio at a sufficient rate, this PR (and my fork of go-dht) is fundamentally flawed.

It would be best to write a program with go-dht to output influx line protocol and use telegrafs exec plugin.

danielnelson:
Too bad about the polling issue.  I guess if we had a build system that could handle cross compiling CGO, like in your original pull request, it would work?  I'm not 100% sure about the syscall not working comment on the golang-nuts forum either, I don't know enough to say.

gdunstone:
I think its just a side effect the way that goroutines work. Its intended behaviour that other things are executed while waiting on io, but becomes a problem when you need something that is executed on a strict deadline.

danielnelson:
It seems like they should have the same accuracy as other native user space code, at least within a single function call.  I thought they only swap out the running goroutine when a function call happens (might be out of date on this though).

gdunstone:
there is a bunch of function calls within go-dht to read from the sensor so that might be it.

Ill revisit at a later date.

gdunstone:
@nerzhul I have created a PR for go-dht that adds an example that outputs the data in influx line format:
https://github.com/d2r2/go-dht/pull/6

@danielnelson if the build system could cross compile cgo that would be great, all it really needs is 
`arm-linux-gnueabihf-gcc`, I think. I'm just installing it on my build system now.

PR #3934, opened 2018-03-25
Add Riemann TLS output

mcorbin:
- Replace raidman by riemann-go-client (https://github.com/riemann/riemann-go-client) which support TLS.
- Add support for TLS

### Required for all PRs:

### DIFF
--- a/Godeps
+++ b/Godeps
@@ -63,6 +63,7 @@ github.com/prometheus/client_model fa8ad6fec33561be4280a8f0514318c79d7f6cb6
 github.com/prometheus/common dd2f054febf4a6c00f2343686efb775948a8bff4
 github.com/prometheus/procfs 1878d9fbb537119d24b21ca07effd591627cd160
 github.com/rcrowley/go-metrics 1f30fe9094a513ce4c700b9a54458bbb0c96996c
+github.com/riemann/riemann-go-client 125d29d80a3e783e9b18c5115158436edcb8510e
 github.com/samuel/go-zookeeper 1d7be4effb13d2d908342d349d71a284a7542693
 github.com/satori/go.uuid 5bf94b69c6b68ee1b541973bb8e1144db23a194b
 github.com/shirou/gopsutil 384a55110aa5ae052eb93ea94940548c1e305a99
--- a/etc/telegraf.conf
+++ b/etc/telegraf.conf
@@ -619,8 +619,17 @@
 
 # # Configuration for the Riemann server to send metrics to
 # [[outputs.riemann]]
-#   ## The full TCP or UDP URL of the Riemann server
+#   ## The full TCP, UDP or TLS URL of the Riemann server
 #   url = "tcp://localhost:5555"
+#
+#   ## Certificate path for TLS connection
+#   # cert_path = "cert.pem"
+#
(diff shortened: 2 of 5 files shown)


danielnelson:
At a high level I'm concerned about this libraries support for timeouts, I see they have it for connect but I don't see any ability to set timeouts on send.

mcorbin:
Thanks your for your review.

I have updated riemann-go-client to support timeout on send. I have also updated this pull request with the new version of riemann-go-client.

danielnelson:
Would you consider adding support for `context.Context` to your library?  Currently I have only basic support for context in Telegraf but I'm looking to expand this in the future.  If we could pass in a context to the Connect and Send functions that would be ideal, as we may want to apply a timeout to a group of commands, or simply want to cancel the connection.

Another question I have is about the channel you are using in the client, what is the purpose of this? https://github.com/riemann/riemann-go-client/blob/master/tls.go#L78

Any chance we can use an already create `tls.Config` for creating the TLSClient?  This will make it easier to hook into our standard TLS handling and allow us to add new options in the future.

Finally, is there a reason why you have separate clients for TLS vs non-TLS?

mcorbin:
> Would you consider adding support for context.Context to your library?

I will try ;)

> Another question I have is about the channel you are using in the client, what is the purpose of this? https://github.com/riemann/riemann-go-client/blob/master/tls.go#L78

For concurrency. We use a channel to be sure than only one message is write at a time. I'm new to Golang so maybe there is a more idiomatic way to do it ?

> Any chance we can use an already create tls.Config for creating the TLSClient? This will make it easier to hook into our standard TLS handling and allow us to add new options in the future.

I will try do to that.

> Finally, is there a reason why you have separate clients for TLS vs non-TLS?

Not really, i will merge the 2 clients.

Thanks you for your reviews ;)

danielnelson:
> For concurrency. We use a channel to be sure than only one message is write at a time. I'm new to Golang so maybe there is a more idiomatic way to do it ?

I see, the Send method is meant to be thread-safe?  I think what you have works, personally I would probably just use the channel as a semaphore and then call Write directly without the goroutine.   I did something like this in the dcos input to limit the number of concurrent HTTP requests made: https://github.com/influxdata/telegraf/blob/a4870e6a6d2f1503b2feea689207bd5093c6033a/plugins/inputs/dcos/client.go#L284-L299


>>Finally, is there a reason why you have separate clients for TLS vs non-TLS?
>
>Not really, i will merge the 2 clients.

I recently have been working on adding TLS support to our zookeeper input, the main difference between the two modes was the Dial function when working with timeouts, this isn't fully tested but maybe it will be helpful:
https://github.com/influxdata/telegraf/blob/c46c3a0b6e033668d7d0438c6d6a5ee28c112e74/plugins/inputs/zookeeper/zookeeper.go#L66-L77

mcorbin:
> I see, the Send method is meant to be thread-safe? 

Yes. And the format being protobuf, i need to do several write and read sequentially for a message.

I have started working on a refactoring (https://github.com/riemann/riemann-go-client/pull/21), i just need to add support for Context (i am reading doc about context and example code right now ;) ). 

I will update this PR once the refactoring merged.

Issue #3937, opened 2018-03-26
Passenger pid and process_group_id should not be tags


AnkurGel:
## Feature Request
Telegraf's Passenger plugin pushes `pid` and `process_group_id` as tags, which quickly becomes too big of a list for a _tag_.(they change on every passenger restart/deploy) I think, cardinality should minimized as much as possible and discussion on https://github.com/influxdata/influxdb/issues/3471#issuecomment-125278928 also seems to agree on that. Should we rather have them as `fields`?

### Proposal:
Move `tag` and `process_group_id` from _tag_ to _field_

### Current behavior:
`pid` and `process_group_id` are _tags_ in schema; which is polluting the tag space, impacting influx's granularity.

### Desired behavior:
Maybe moving `pid` and `process_group_id` to field may help. For querying, I am not sure in which case people will feel the need of grouping by them as tag.

danielnelson:
One thing to consider is if we will still have enough unique information, here is an example series key for passenger_process:
```
passenger_process,app_root=/var/app/current,code_revision=899ac7f,group_name=/var/app/current/public,life_status=ALIVE,pid=11553,process_group_id=13608,supergroup_name=/var/app/current/public
```

If we remove pid/process_group_id, you won't be able to track multiple processes whose remaining tags are the same.  Is it possible to have 2 processes at once with the same app_root, group_name, and supergroup_name?  You would have to consider the metric across multiple servers as well.

AnkurGel:
@danielnelson ah yes. You're right. We will miss the stats per passenger process with that. 
Though across multiple host, we will still get `host` as tag so we can differentiate there. But on a same host, across multiple passenger processes of same application, the absence of `pid` tag will remove the granularity (in _group by_) per process. 
Can `pid` and `process_group_id` rather be optional as tag? I'm curious how often are folks interested in seeing timeseries per process?

danielnelson:
Perhaps you could just remove the tags completely in your case using [measurement filtering](https://github.com/influxdata/telegraf/blob/master/docs/CONFIGURATION.md#measurement-filtering):
```
[[inputs.passenger]]
  # ... snip ...
  tagexclude = ["pid", "process_group_id"]
```

PR #3944, opened 2018-03-28
move pid tag to fields

arkpoah:
Having pid in tags creates lot of influxdb series on some architecture with severals passengers apps on severals front servers, then we get an influxdb error : max series per database exceeded.
Don't think pid is useful in tags, so putting it in fields can avoid to change max series per database value on influxdb side.

### Required for all PRs:


danielnelson:
@arkpoah Take a look at https://github.com/influxdata/telegraf/issues/3937, I think we may need pid to provide uniqueness to the series.

arkpoah:
Thx, I'll use tagexclude then.

PR #3946, opened 2018-03-28
uint64 support

ragzilla:
Didn't see an appropriate README to update. 

Pull adds uint64 support for InfluxDB line protocols (in and out), and to telegraf/metric.  telegraf metric uint64 support is gated behind a build flag until infludata/influxdb mainline uint64 is generally available to ensure behavior is consistent with how telegraf has behaved in the past.  

That will likely be a breaking change in the future unless the API for AddField and convertField are extended so the caller can specify whether or not uint64 storage and, by extension output, is desired.

### Required for all PRs:

### DIFF
--- a/plugins/parsers/influx/parser_test.go
+++ b/plugins/parsers/influx/parser_test.go
@@ -16,6 +16,12 @@ func Metric(v telegraf.Metric, err error) telegraf.Metric {
 	return v
 }
 
+const (
+	Uint64Overflow uint64 = 9223372036854775808
+	Uint64Max      uint64 = 18446744073709551615
+	Uint64Test     uint64 = 42
+)
+
 var DefaultTime = func() time.Time {
 	return time.Unix(42, 0)
 }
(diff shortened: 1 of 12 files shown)


ragzilla:
@danielnelson went with the influxdb approach of gating the feature behind a build flag.

plugins/parsers/influx/parser_test.go
```diff
@@ -16,6 +16,12 @@ func Metric(v telegraf.Metric, err error) telegraf.Metric {
 	return v
 }
 
+const (
+	Uint64Overflow uint64 = 9223372036854775808
```
danielnelson: Shouldn't this be larger than Uint64Max?  We need to test 18446744073709551616.

ragzilla: That was to verify Int64Max+1 properly typed to uint64.  If we attempt to define a const that’s Uint64Max+1 it’d error on compile.

danielnelson:
I did some follow-up work, if you can review: https://github.com/influxdata/telegraf/pull/3948

PR #3964, opened 2018-04-03
Add repl_oplog_window_s metric to mongodb input

subuk:
Hello!

Here is new metric named oplog_timediff.

References:
https://www.mongodb.com/blog/post/replica-set-health-is-more-than-just-replication
https://github.com/DataDog/integrations-core/blob/master/mongo/datadog_checks/mongo/mongo.py#L972

### Required for all PRs:

### DIFF
--- a/plugins/inputs/mongodb/README.md
+++ b/plugins/inputs/mongodb/README.md
@@ -56,6 +56,7 @@ and create a single measurement containing values e.g.
  * ttl_deletes_per_sec
  * ttl_passes_per_sec
  * repl_lag
+ * repl_oplog_window_s
  * jumbo_chunks (only if mongos or mongo config)
 
 If gather_db_stats is set to true, it will also collect per database stats exposed by db.stats()
--- a/plugins/inputs/mongodb/mongodb_server.go
+++ b/plugins/inputs/mongodb/mongodb_server.go
@@ -22,6 +22,41 @@ func (s *Server) getDefaultTags() map[string]string {
 	return tags
 }
 
+type oplogEntry struct {
+	Timestamp bson.MongoTimestamp `bson:"ts"`
+}
+
+func (s *Server) gatherOplogStats() *OplogStats {
+	stats := &OplogStats{}
+	localdb := s.Session.DB("local")
+
+	op_first := oplogEntry{}
+	op_last := oplogEntry{}
+	query := bson.M{"ts": bson.M{"$exists": true}}
+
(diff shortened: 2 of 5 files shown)


plugins/inputs/mongodb/README.md
```diff
@@ -57,6 +57,7 @@ and create a single measurement containing values e.g.
  * ttl_passes_per_sec
  * repl_lag
  * jumbo_chunks (only if mongos or mongo config)
+ * oplog_timediff
```
danielnelson: I think we should call the field `repl_oplog_window_s`.  This more closely matches the linked documentation from mongodb, while still using the same repl prefix as the other replication metrics and also includes the units.  I sort of dislike `_s` because its not immediately clear what it means, but it matches more in style with existing `_ns` readings.

plugins/inputs/mongodb/mongodb_server.go
```diff
@@ -22,6 +22,56 @@ func (s *Server) getDefaultTags() map[string]string {
...
+type OplogStats struct {
+	TimeDiff int64
+}
+
+func (s *Server) gatherOplogStats() *OplogStats {
+	stats := &OplogStats{}
+	localdb := s.Session.DB("local")
+	oplog_collection_name := ""
+
+	collection_names, err := localdb.CollectionNames()
```
danielnelson: It seems like instead of checking for the existence of this items we should just try to get the data and if it's not there then we don't report anything.  This will result in less network calls and checking doesn't guarantee the data will be there later anyway due to possible changes.

How about we just try to get the entries for `oplog.rs` and if that fails we can try `oplog.$main`?

plugins/inputs/mongodb/README.md
subuk: Okay, I will rename oplog_timediff to repl_oplog_window_s and move it closer to other repl_* metrics in code.

plugins/inputs/mongodb/mongodb_server.go
subuk: Yeah, it makes sense, will be fixed.

subuk:
Fixed. Metric renamed to repl_oplog_window_s, collection list query removed.

danielnelson:
Thank you, this will be in version 1.7.0

danielnelson:
@subuk I renamed this `repl_oplog_window_sec` as I noticed that this matches the style of the other fields in this plugin.  1890efbb70448642bcebfddda7f77cccfda0c8aa

PR #677, opened 2016-02-10
Added zabbix output support

UnderGreen:
Have a lot of problems on rebase in #470. Started new MR.

PR #470, opened 2015-12-30
Add zabbix output plugin

UnderGreen:
Added output plugin, which sends metrics to Zabbix Server.

It's my first Golang code, please review it in depth. :smile:

PR #3966, opened 2018-04-03
feat: output and aggregator for Zabbix

adrianlzt:
Output plugin to send metrics to Zabbix, using item parameters to map
Telegraf args.
Also added an aggregator to call Zabbix Low Level Discovery rules to
generate new keys needed for metrics with tags.

### Required for all PRs:


adrianlzt:
Other PRs related: https://github.com/influxdata/telegraf/pull/677 https://github.com/influxdata/telegraf/pull/470

adrianlzt:
If a global tag is added to the telegraf config, it will be used by the output plugin and the aggregator as if it were a parameter from the input plugin, changing how the Zabbix key is generated and, probably, not reaching the appropiate item in Zabbix.

adrianlzt:
Please, run the tests again, was a error in golint: https://github.com/golang/lint/issues/397

rockaut:
Is there any plan when to include this PR? Would love to see that in 1.9 or so? *prettyplease*

rockaut:
Question to @adrianlzt : as i see the last changes it seams that mostly LLD is some work to do but default zabbix output works?
What if you remove the LLD completely for now so that we can get zabbix output before telegrafs 1.11.0 release? LLD is changing significantly with Zabbix 4.2 release so maybe it will be easyer to get along then?

adrianlzt:
Thanks for the info @rockaut. I didn't know about those changes.
If some at InfluxData agree to merge this PR without LLD I'm ok removing the LLD part. In fact, right now we are modifying it to generate only LLDs when needed to avoid having so many.

rockaut:
maybe  @russorat who changed the tag last time can say something?

danielnelson:
I won't have time to look at it before 1.10, but will try for 1.11.  Keeping the plugins together is fine for now, we can split them later on if they need to go in seperately.

adrianlzt:
I have some improvements to add before merge.

glinton:
Thanks @adrianlzt. Don't forget to update your branch with master and ensure the tests pass.

plugins/aggregators/zabbix_lld/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+This metrics are expected to be sent to Zabbix using the Zabbix output plugin.
+
+**[Zabbix 4.2](https://www.zabbix.com/documentation/4.2/manual/discovery/low_level_discovery)** has changed the LLD format and it is not currently supported.
+
+#### Idea
+
+Zabbix needs an `item` created before receiving any metric. In some cases we do not know in advance what are we going to send, for example, the name of a container to send its cpu and memory consumption.
+
+For this case Zabbix provides [low level discovery](https://www.zabbix.com/documentation/4.0/manual/discovery/low_level_discovery) that allow to create dinamically new items based on the parameters sent by the trap.
```
v-zhuravlev: dynamically?

plugins/outputs/zabbix/README.md
```diff
@@ -0,0 +1,120 @@
...
+```
+measurement,host=hostname valueA=0,valueB=1
+->
+measurement.valueA
+measurement.valueB
+```
+
+If the metric has tags, they will be added, in alphabetical order, like:
+```
+measurement,host=hostname,tagA=keyA,tagB,keyB valueA=0,valueB=1
```
v-zhuravlev: tagB=keyB?

v-zhuravlev:
@adrianlzt , is it possible to test your implementation already? By building telegraf from your branch?
thanks

adrianlzt:
Yes. We are using it in production since one year. Just be careful with LLDs, too many impact badly in zabbix

alvarolmedo:
+1

russorat:
@adrianlzt looks like your tests are failing. could you take a look?

v-zhuravlev:
Tested you plugin with some real-life cases. Worked like a charm. Would be great to see this merged.

rpmcdougall:
Is there any plans to get this into an upcoming release? I have a use case for this currently that I would like to try it out for.

sjwang90:
This output plugin actually looks like a great candidate for an [external plugin](https://github.com/influxdata/telegraf/blob/master/docs/EXTERNAL_PLUGINS.md). Would you be willing to submit this plugin as an [external plugin](https://github.com/influxdata/telegraf/blob/master/EXTERNAL_PLUGINS.md) that can be used with `execd` to run seamlessly with Telegraf? Telegraf users would be able to use the plugin sooner by having it as an external plugin since it wouldn't have to go through the typical review process. 

Feel free to respond with any questions you may have.

adrianlzt:
This is a small output plugin, without licensed software, neither large dependencies, so I don't see why it will fit as an output external plugin.
Also, having to run a second binary breaks this clean "one binary deploy" that we love in golang, also it imposes more monitoring to check that second proc is running correctly.

frankgerhardt:
I just wrote this in #470: 

I'm looking at this and it's years old. My question is: is there Zabbix support by now and if yes where?

adrianlzt:
We have been using this plugin in production for ~3 years.

You can create your own build of Telegraf adding this PR

michalg91:
Any change to see this PR merged? It looks solid.

srebhan:
@adrianlzt can you please rebase this on the current master and switch to go modules? I promise to give it a review then...

srebhan:
@adrianlzt looking at the code, I wonder why you cannot generate the traps in the output plugin instead of using an intermediate aggregator. The output can scan all received metrics and generate the traps just like the aggregator...

The benefit when doing it in the output is that you do not run into timing issues where e.g. the output is flushed before the new aggregated traps arrive etc. Furthermore, it **only** makes send to use the aggregator with the provided output... So why do you split those?

adrianlzt:
My idea was to reduce as much as possible the send of LLD packets to Zabbix. Those packets are costly to be processed by Zabbix.

Also, those packets should contain all the possible tags for a particular measurement, for example, if we have configured different inputs.procstat with different intervals, LLD packet should contain all those possible values.

Using an aggregator felt natural to this uses cases (waiting and aggregating) but I see your point that's werid to have to configure an aggregator to make an output plugin works.

I guess the other option is to move all the logic from the aggregator to the output plugin. That will impose more knowledge of the state in the output.

Or do you mean using ``AggregatingOutput``? (I have just discovered it, it didn't exist when I started this PR)

srebhan:
TBH I never came across an `AggregatingOutput` and I think this is semantically a different concept, but not sure.

Regarding _Aggregator_ vs. _Output_: My understanding is that you need to send those LLD packets (sorry I'm by no means an expert in Zabbix) whenever you encounter a new tag. In best case you can accumulate all possible tags that are sent through the output to reduce the cost on the Zabbix side. Is this correct?
If so, my idea is the following, add a list of known tags (best as `map[string]bool`) to the output. Then, iterate over the `metrics` in `Write()` first and check for "unknown" tags. Send a LLD if you found one. Then send the metrics. So yes you will need a kind of state in the output, but that would be a natural thing IMO. If you do so, please put the LLD logic in an own file so it is clear for the reader what you are after...

Does that make sense to you?

adrianlzt:
The idea of how Zabbix works is simple, yet different from all the new wave of monitoring systems.
To be able to send data to Zabbix first you need to create the associated _item_, that creation implies some writes in the database to register it.

Some data is _static_, meaning that its always the same, mem_pct and mem_swap for example. Those _items_ are created when you register your node (yes, you always have to declarate things before sending anything).

The problems comes with _dynamic_ data, for example disks or network interfaces, that you don't know, a priori, how many of them have each server.
That is when LLD comes into play, declaring an special item that receives, for example, the names of the disks and create the _items_ (registering them in the database) so Zabbix could accept those metrics.

So yes, I guess I could move the logic from the aggregator to the output.

srebhan:
@adrianlzt any news here?

natenho:
@adrianlzt 🆙 👍🏻

srebhan:
@natenho feel free to take control of the PR or submit a new one... :-)

adrianlzt:
I don't have much time to do the refactor right now, but I can guide if needed

powersj:
@adrianlzt - is there history in this PR that you need? Otherwise, I might suggest opening a new PR if you have a new approach? Thanks!

adrianlzt:
Ok. I will open a new PR once I have changed the code to have a simple Output with the LLD integrated.

PR #3971, opened 2018-04-04
added details about why not all logstash patterns are supported, reso…

sanderson:
…lves influxdata/docs.influxdata.com#991

### DIFF
--- a/plugins/inputs/logparser/README.md
+++ b/plugins/inputs/logparser/README.md
@@ -89,8 +89,8 @@ current time.
   - drop     (drops the field completely)
 - Timestamp modifiers:
   - ts               (This will auto-learn the timestamp format)
-  - ts-ansic         ("Mon Jan _2 15:04:05 2006")
-  - ts-unix          ("Mon Jan _2 15:04:05 MST 2006")
+  - ts-ansic         ("Mon Jan \_2 15:04:05 2006")
+  - ts-unix          ("Mon Jan \_2 15:04:05 MST 2006")
   - ts-ruby          ("Mon Jan 02 15:04:05 -0700 2006")
   - ts-rfc822        ("02 Jan 06 15:04 MST")
   - ts-rfc822z       ("02 Jan 06 15:04 -0700")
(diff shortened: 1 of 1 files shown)


plugins/inputs/logparser/README.md
```diff
@@ -89,8 +89,8 @@ current time.
   - drop     (drops the field completely)
 - Timestamp modifiers:
   - ts               (This will auto-learn the timestamp format)
-  - ts-ansic         ("Mon Jan _2 15:04:05 2006")
-  - ts-unix          ("Mon Jan _2 15:04:05 MST 2006")
+  - ts-ansic         ("Mon Jan \_2 15:04:05 2006")
+  - ts-unix          ("Mon Jan \_2 15:04:05 MST 2006")
```
danielnelson: I know underscore is technically supposed to be escaped, but I prefer to leave them out for typeability and readability.  Github's markdown parser does handle it as desired without the escapes.  If you like we can use backtick's on these strings instead.

plugins/inputs/logparser/README.md
```diff
@@ -108,10 +108,11 @@ CUSTOM time layouts must be within quotes and be the representation of the
...
 See https://golang.org/pkg/time/#Parse for more details.
 
-Telegraf has many of its own
-[built-in patterns](./grok/patterns/influx-patterns),
-as well as supporting
+Telegraf has many of its own [built-in patterns](./grok/patterns/influx-patterns),
+as well as support for most of
 [logstash's builtin patterns](https://github.com/logstash-plugins/logstash-patterns-core/blob/master/patterns/grok-patterns).
+_Golang regular expressions do not support lookahead or lookbehind.
+logstash patterns that depend on these are not supported._
```
danielnelson: Maybe we can have a list of the patterns linked above that are not in Telegraf?  We should probably also pin the list to a specific version.

plugins/inputs/logparser/README.md
sanderson: Fair enough. I'll revert this change.

plugins/inputs/logparser/README.md
sanderson: Is there an easy way to get the list of logstash patterns not supported by Telegraf?

danielnelson: I guess not really, we could write a script that compares the pattern names
between the logstash patterns[1] with the vjeantet/grok libraries patterns[2],
but it would probably go out of date, so lets not do that.

[1] https://github.com/logstash-plugins/logstash-patterns-core/blob/master/patterns/grok-patterns
[2] https://github.com/vjeantet/grok/tree/d73e972b60935c7fec0b4ffbc904ed39ecaf7efe/patterns

PR #3974, opened 2018-04-05
Microsoft Application Insights output plugin

karolz-ms:
This PR adds an output plugin that sends data to Microsoft Application Insights https://docs.microsoft.com/azure/application-insights/

Hopefully this addition aligns with telegraf roadmap, but if there are concerns, we (here at Microsoft) are open to other suggestions--thanks!


karolz-ms:
@brahmnes @yantang-msft @jjjordanmsft Could you please take a peek? TIA!

plugins/outputs/application_insights/application_insights.go
```diff
@@ -0,0 +1,410 @@
...
+		a.diagMsgListener.Remove()
+	}
+
+	return nil
+}
+
+func init() {
+	outputs.Add("application_insights", func() telegraf.Output {
+		return &ApplicationInsights{
+			DiagMsgSubscriber: appinsightsDiagnosticsMessageSubscriber{},
```
yantang-msft: Is there a way to disable the diagMsgListener?
I can't find the doc about configuration, but I assume user can only specify simple data type in the configuration, thus the diagnostics listener is always created?

plugins/outputs/application_insights/application_insights.go
```diff
@@ -0,0 +1,410 @@
...
+func (a *ApplicationInsights) Connect() error {
+	if a.Timeout.Duration == 0 {
+		a.Timeout = internal.Duration{Duration: time.Second * 5}
+	}
+
+	if a.Transmitter == nil {
+		a.useDefaultTransmitter()
+	}
+
+	if a.Transmitter != nil && a.DiagMsgSubscriber != nil {
```
yantang-msft: I feel like a boolean configuration like "developer_mode" is simpler than this extra interface and implementation.

plugins/outputs/application_insights/application_insights.go
```diff
@@ -0,0 +1,410 @@
...
+			a.Transmitter.Track(telemetry)
+		}
+	}
+
+	return nil
+}
+
+func (a *ApplicationInsights) Close() error {
+	if a.Transmitter == nil {
+		return nil
```
yantang-msft: nit: The current logic won't create diagMsgListener if Transmitter is null, but still feel safer to always check diagMsgListener before return.
How about
```
if a.Transmitter != nil {
    select {....}
}
// check and remove diaMsgListener
```

plugins/outputs/application_insights/application_insights.go
```diff
@@ -0,0 +1,410 @@
...
+		}
+
+		return int(v), nil
+	}
+
+	return 0.0, fmt.Errorf("[%s] cannot be converted to an int value", value)
+}
+
+func logOutputMsg(format string, v ...interface{}) {
+	log.Printf("Output [application_insights] "+format, v...)
```
yantang-msft: nit: space between '+'

plugins/outputs/application_insights/application_insights.go
```diff
@@ -0,0 +1,410 @@
...
+				return err
+			}
+
+			*telemetryProperty = metricValue
+
+			if usedFields != nil {
+				*usedFields = append(*usedFields, fieldName)
+			}
+
+			return nil
```
yantang-msft: Is this return statement necessary?

plugins/outputs/application_insights/application_insights.go
```diff
@@ -0,0 +1,410 @@
...
+
+func setFloat64TelemetryProperty(
+	telemetryProperty *float64,
+	candidateFields []string,
+	metric telegraf.Metric,
+	usedFields *[]string) error {
+
+	for _, fieldName := range candidateFields {
+		if metric.HasField(fieldName) {
+			metricValue, err := toFloat64(metric.Fields()[fieldName])
```
yantang-msft: What if the fieldName afterwards can be converted? Do we still want to return err?

plugins/outputs/application_insights/application_insights.go
```diff
@@ -0,0 +1,410 @@
...
+			}
+			suffix++
+		}
+		properties[name+"_"+strconv.Itoa(suffix)] = value
+	}
+}
+
+func toFloat64(value interface{}) (float64, error) {
+	switch v := value.(type) {
+	case int:
```
yantang-msft: Can use list in the case statement
case int, int8, int16...:

plugins/outputs/application_insights/application_insights.go
```diff
@@ -0,0 +1,410 @@
...
+		return v, nil
+	}
+
+	return 0.0, fmt.Errorf("[%s] cannot be converted to a float64 value", value)
+}
+
+func toInt(value interface{}) (int, error) {
+	if !is32BitChecked {
+		is32BitChecked = true
+		i := int(-1)
```
yantang-msft: What do you think of using sizeof to check 32bit? That is more straightforward to me.

plugins/outputs/application_insights/application_insights.go
```diff
@@ -0,0 +1,410 @@
...
+	telemetry.Properties = metric.Tags()
+	telemetry.Timestamp = metric.Time()
+
+	err := setFloat64TelemetryProperty(&(telemetry.Value), []string{"value"}, metric, nil)
+	if err == nil {
+		addRemainingMeasurements(telemetry, metric, []string{"value"})
+	} else {
+		// Either the metric does not have a "value" field, or it was not convertable to float64
+		// Heuristics: if the metric has just one field that is convertible to float64, take it as value
+		// Otherwise give up
```
yantang-msft: I think the "value" field is just a special case. And the general case is metric with multiple dimensions (fields), and we shouldn't give up in this case.
For example, I can have a metric like
{
    metricName: "cpu",
    fields: [
        "Percent_Idle_Time": 0.3,
        "Percent_User_Time": 0.5
    ]
}
and we can generate multiple AI telemetries from this metric, with metric names cpu_Percent_Idle_Time, cpu_Percent_User_Time

plugins/outputs/application_insights/README.md
```diff
@@ -0,0 +1,13 @@
+# Application Insights Output Plugin
+
+This plugin writes telegraf metrics to Auzre Application Insights
```
jwilder: Azure

plugins/outputs/application_insights/application_insights.go
jwilder: @yantang-msft That's correct.  A measurement (in telegraf/influx) could have multiple fields.  

The logic here would would only pick up a measurement that happens to have a `value` field and is convertible to a `float64`... dropping all other fields which is unlikely what is intended.

plugins/outputs/application_insights/application_insights.go
```diff
@@ -0,0 +1,410 @@
...
+
+func addRemainingMeasurements(telemetry appinsights.Telemetry, metric telegraf.Metric, usedFields []string) {
+	properties := telemetry.GetProperties()
+
+	for fieldName, fieldValue := range metric.Fields() {
+		if contains(usedFields, fieldName) {
+			continue
+		}
+
+		float64Value, err := toFloat64(fieldValue)
```
yantang-msft: Since 'properties' is a string to string map, is it still necessary to require the fieldValue to be numeric? Although I believe the value is always a numeric. :)

plugins/outputs/application_insights/application_insights.go
```diff
@@ -0,0 +1,410 @@
...
+	return false
+}
+
+func addUniquelyNamedProperty(properties map[string]string, name string, value string) {
+	if _, found := properties[name]; !found {
+		properties[name] = value
+	} else {
+		suffix := 1
+		for {
+			_, found = properties[name+"_"+strconv.Itoa(suffix)]
```
yantang-msft: Looks like you intended not to having space with string concatenation.

yantang-msft: Is it possible to use the same if statement like above to simplify this? i.e.,
if _, found = properties[...]; !found {
    break;
}

plugins/outputs/application_insights/application_insights.go
jjjordanmsft: It's a function callback, so you could disable it in your implementation.  The SDK provides only Add/Remove.

plugins/outputs/application_insights/application_insights.go
```diff
@@ -0,0 +1,410 @@
...
+	err = setIntTelemetryProperty(&(telemetry.Count), []string{"count", "samples"}, metric, &usedFields)
+	if err != nil {
+		return nil
+	}
+
+	// We attempt to set Min, Max, StdDev and Variance fields but do not really care if they are not present
+	setFloat64TelemetryProperty(&(telemetry.Min), []string{"min"}, metric, &usedFields)
+	setFloat64TelemetryProperty(&(telemetry.Max), []string{"max"}, metric, &usedFields)
+	setFloat64TelemetryProperty(&(telemetry.StdDev), []string{"stddev"}, metric, &usedFields)
+	setFloat64TelemetryProperty(&(telemetry.Variance), []string{"variance"}, metric, &usedFields)
```
jjjordanmsft: Is there a reason to set both?  The SDK will only use one or the other, with stddev taking precedence.

plugins/outputs/application_insights/application_insights.go
```diff
@@ -0,0 +1,410 @@
...
+		float64Value, err := toFloat64(fieldValue)
+		if err != nil {
+			continue
+		}
+
+		// We would like to add the field to telemetry.Measurements here
+		// But because of issue https://github.com/Microsoft/ApplicationInsights-Go/issues/12
+		// this is currently impossible. So we are just going to add the measurement as a property instead.
+		// This is what we want to be able to do, eventually:
+		//     telemetry.GetMeasurements()[fieldName] = float64Value
```
jjjordanmsft: I assume that `telemetry` is always a metric (whether single or multiple value)?

This is unfortunate.  We lose many possible ways to analyze these measurements by stringifying them into properties (which, btw, I'm not sure you can even see except in the AA portal)

Is it really not feasible to log these as separate metrics?

plugins/outputs/application_insights/application_insights.go
jjjordanmsft: @yantang-msft FYI `go fmt` would have output the string concatenation that way, and it's the accepted convention.

plugins/outputs/application_insights/application_insights.go
jjjordanmsft: See [type switches](https://golang.org/doc/effective_go.html#type_switch)

plugins/outputs/application_insights/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+### Configuration
```
jjjordanmsft: I guess I don't know a whole lot about this system, but there may be some value in specifying things like RoleName, RoleInstance, and custom properties -- all things that would just end up in the TelemetryClient.Context() object.  These help isolate regions, resource groups, and instances in our queries.  Perhaps the default (hostname in RoleInstance) is enough, but I find myself always filling in these fields for services that I monitor.

karolz-ms:
Hey, thanks everyone for lots of good comments! The main takeaway seems that, facing with a metric with multiple fields, it is better to submit multiple metric telemetry records to AppInsights rather than squishing the fields into telemetry properties. I completely agree with this, but it will require some rework and new tests, so I am going to close this PR and submit another one that incorporates this change.

@brahmnes FYI

plugins/outputs/application_insights/application_insights.go
karolz-ms: @yantang-msft Agreed, DiagMsgSubscriber was exposed for testability, but it is probably too much of a good thing. Will rework

plugins/outputs/application_insights/application_insights.go
karolz-ms: Agreed

plugins/outputs/application_insights/application_insights.go
karolz-ms: This is how standard go code formatting formats this expression

plugins/outputs/application_insights/application_insights.go
```diff
@@ -0,0 +1,410 @@
...
+	outputs.Add("application_insights", func() telegraf.Output {
+		return &ApplicationInsights{
+			DiagMsgSubscriber: appinsightsDiagnosticsMessageSubscriber{},
+		}
+	})
+}
+
+func (a *ApplicationInsights) useDefaultTransmitter() {
+	// Env variable, if present, overrides what is in the configuration
+	iKey := os.Getenv("APPINSIGHTS_INSTRUMENTATIONKEY")
```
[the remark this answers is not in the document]
karolz-ms: Makes sense, will change this

plugins/outputs/application_insights/application_insights.go
karolz-ms: @yantang-msft Not sure I understand the question. The error is about inability to convert to float64 and has nothing to do with the field name.

plugins/outputs/application_insights/application_insights.go
karolz-ms: Meaning, you can't

plugins/outputs/application_insights/application_insights.go
karolz-ms: Sure, we can to that

plugins/outputs/application_insights/application_insights.go
karolz-ms: Agreed, will submit new PR to fix this

plugins/outputs/application_insights/README.md
karolz-ms: thx

plugins/outputs/application_insights/application_insights.go
karolz-ms: @jjjordanmsft The logic here is that we look for both fields, and use what is available--the call will be a no-op if . I will make sure that we test for stddev after testing for variance to align with the AppInsights SDK

plugins/outputs/application_insights/application_insights.go
karolz-ms: It is, will change that

plugins/outputs/application_insights/README.md
karolz-ms: @jjjordanmsft This is good point and we are having some discussions about how to set "special" properties on AppInsights telemetry. I will add you to the email thread

plugins/outputs/application_insights/application_insights.go
```diff
@@ -0,0 +1,410 @@
...
+	}
+
+	if a.InstrumentationKey != "" {
+		a.Transmitter = NewAppinsightsTransmitter(a.InstrumentationKey)
+	}
+}
+
+func canBeTreatedAsAggregate(metric telegraf.Metric) bool {
+	// If we can identify the Sum and Count properties, then we are good
+	hasSum := metric.HasField("value") || metric.HasField("sum")
```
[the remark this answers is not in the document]
yantang-msft: We hardcoded the rule of how an metric can be treated as aggregated, which would be hard to follow. What if the field name is "sample" or "sampling" instead of "samples", or "total" instead of "sum"?
Someone will have to decide whether a metric is aggreated, but I hope it's done by the input plugin, e.g., AppMetric input plugin will know the property name that means aggregated plugin, or some override plugin with more flexibility.
As an output plugin, we can just rely on this property Metric.IsAggregate.

plugins/outputs/application_insights/application_insights.go
yantang-msft: I mean we have a list of candidateFields. If the first candidate field failed to convert, but the second candidate field can be converted, do we want to return as soon as there is an error or wait after all fields are checked.

karolz-ms: Ah, good point--we should continue. Thanks

plugins/outputs/application_insights/application_insights.go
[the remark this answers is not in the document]
karolz-ms: telegraf.Metric.IsAggregate() is just a flag on the (Telegraf) metric type. From Application Insights perspective, to create an AppInsights aggregate metric, we have to know at least the sum of all values and the number of samples, so telegraf.Metric.IsAggregate() alone does not help us with that task. 

Application Insights aggregate metric is a concept specific to AppInsighst, so it *has to be the AppInsights output plugin* that makes the decision whether a Telegraf metric can be interpreted as an AppInsights aggregate metric or not. We might eventually need more flexibility here in terms of how this interpretation is made, but it quickly gets complicated and I do not want to try to solve this in the first iteration.

Issue #3975, opened 2018-04-05
Influxdb Output no longer works with HTTPS urls


JeffAshton:
Introduced in the change 43bd23e5556ba947758cd23af0364ce808add059, the urls for HTTPS scheme are no longer formatted correctly.

```
	switch u.Scheme {
	case "unix":
		u.Scheme = "http"
		u.Host = "127.0.0.1"
		u.Path = "/write"
	case "http":
		u.Path = path.Join(u.Path, "write")
	}
```

```
	switch u.Scheme {
	case "unix":
		u.Scheme = "http"
		u.Host = "127.0.0.1"
		u.Path = "/query"
	case "http":
		u.Path = path.Join(u.Path, "query")
	}
```

Both of these switch statements are missing ```https``` case.

Added a log line to ```makeQueryRequest``` to see what url was 404ing:

```
	log.Printf("D! [outputs.influxdb] Posting query to %s", c.QueryURL)
```

And produced the following debug output:

```
./telegraf.exe --config telegraf.conf
2018-04-05T17:10:20Z D! Attempting connection to output: influxdb
2018-04-05T17:10:20Z D! [outputs.influxdb] Posting query to https://influxdb.example.com:8086
2018-04-05T17:10:20Z W! [outputs.influxdb] when writing to [https://influxdb.example.com:8086]: database "telegraf" creation failed: 404 Not Found
2018-04-05T17:10:20Z D! Successfully connected to output: influxdb
2018-04-05T17:10:20Z I! Starting Telegraf v1.6.0-rc2
```

After fixing the switch statements or using http:

```
./telegraf.exe --config telegraf.conf
2018-04-05T17:10:54Z D! Attempting connection to output: influxdb
2018-04-05T17:10:54Z D! [outputs.influxdb] Posting query to http://influxdb.example.com:8086/query
2018-04-05T17:10:54Z D! Successfully connected to output: influxdb
2018-04-05T17:10:54Z I! Starting Telegraf v1.6.0-rc2
```

You'll notice the query url is correct in the 2nd case.

PR #3976, opened 2018-04-05
Fixing https urls in Influxdb Output Plugin

JeffAshton:
Fixing https://github.com/influxdata/telegraf/issues/3975, where the influxdb output plugin now longer formats valid write/query urls for https scheme.

### Required for all PRs:

### DIFF
--- a/plugins/outputs/influxdb/http.go
+++ b/plugins/outputs/influxdb/http.go
@@ -418,6 +418,7 @@ func makeWriteURL(loc *url.URL, db, rp, consistency string) string {
 		u.Host = "127.0.0.1"
 		u.Path = "/write"
 	case "http":
+	case "https":
 		u.Path = path.Join(u.Path, "write")
 	}
 	u.RawQuery = params.Encode()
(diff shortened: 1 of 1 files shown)


plugins/outputs/influxdb/http.go
```diff
@@ -432,6 +433,7 @@ func makeQueryURL(loc *url.URL) string {
 		u.Host = "127.0.0.1"
 		u.Path = "/query"
 	case "http":
+	case "https":
 		u.Path = path.Join(u.Path, "query")
 	}
```
JeffAshton: I really hate that there is no default case that logs an error, like here:

https://github.com/influxdata/telegraf/blob/master/plugins/outputs/influxdb/influxdb.go#L163

danielnelson: That's a good point, I added some handling of the default case: df78133bf3d20edaf20fb565559a0b4109b09779

danielnelson:
I'll release a rc3 today with this.

Issue #3239, opened 2017-09-15
socket listener: needs a way to set socket permissions for unix/dgram sockets.


james-lawrence:
### System info:
Telegraf v1.4.0 (git: release-1.4 34b7a4c3611d1ede908ef275401544c34a4a3ba3)

## Feature Request

being able to set socket permissions for socket_listener, will allow administrators to have various applications on the system write to the telegraf socket.

```
# Configuration for telegraf agent
[agent]
  interval = "20s"
  round_interval = true
  metric_batch_size = 1000
  metric_buffer_limit = 10000
  collection_jitter = "5s"
  flush_interval = "60s"
  flush_jitter = "15s"
  debug = false
  quiet = false
  omit_hostname = false

[[inputs.socket_listener]]
  service_address = "unix:///run/telegraf-ops/telegraf.sock"
```

telegraf -conf=test.conf

danielnelson:
See implementation discussion starting here: https://github.com/influxdata/telegraf/pull/3996#issuecomment-394164349

jgitlin-bt:
@danielnelson I have a concern with what was proposed on #3996, namely that if it becomes mandatory in a future version (I.E. there's no way to disable it) then it will require that processes sending to the socket have the same group id as telegraf... this will be an issue for my usage, where processes send InfluxDB line protocol stats to the socket but these process need to be in a separate group from the telegraf process.

Do you know if the checks for SCM_CREDENTIALS will be able to be disabled via a config option? That would be ideal for me, as I could then rely on directory permissions for access control...

danielnelson:
> Do you know if the checks for SCM_CREDENTIALS will be able to be disabled via a config option? That would be ideal for me, as I could then rely on directory permissions for access control...

We can do this, I think we would only use SCM_CREDENTIALS if it is configured, and use directory permissions by default.

jegj:
is there a way i can solve the permission problem while this feature is ready for a new release?  `chmod` + `chown` works great but i want to keep the permissions after a service restart... the socket always has `rwxr-xr-x telegraf telegraf`  as permissions after a restart... so a external service will lost the write access

danielnelson:
Not sure if it will work for you, but you could try setting setgid on the directory containing the socket and adjusting the umask to 002 for Telegraf so the file is group writable.

jegj:
[...]

PR #3996, opened 2018-04-10
Provide a way to set socket permissions for unix/dgram sockets

gentstr:
#3239

* Defaults to world-writable by default (777)
* New configuration option file_mode allows for custom permissions

### Required for all PRs:

### DIFF
--- a/plugins/inputs/socket_listener/README.md
+++ b/plugins/inputs/socket_listener/README.md
@@ -25,6 +25,9 @@ This is a sample configuration for the plugin.
   # service_address = "unix:///tmp/telegraf.sock"
   # service_address = "unixgram:///tmp/telegraf.sock"
 
+  ## File mode for unix sockets
+  # file_mode = "777"
+
   ## Maximum number of concurrent connections.
   ## Only applies to stream sockets (e.g. TCP).
   ## 0 (default) is unlimited.
--- a/plugins/inputs/socket_listener/socket_listener.go
+++ b/plugins/inputs/socket_listener/socket_listener.go
@@ -7,6 +7,7 @@ import (
 	"log"
 	"net"
 	"os"
+	"strconv"
 	"strings"
 	"sync"
 
@@ -164,6 +165,7 @@ type SocketListener struct {
 	ReadBufferSize  int
 	ReadTimeout     *internal.Duration
 	KeepAlivePeriod *internal.Duration
+	FileMode        string
 
 	parsers.Parser
 	telegraf.Accumulator
@@ -187,6 +189,9 @@ func (sl *SocketListener) SampleConfig() string {
   # service_address = "udp6://:8094"
   # service_address = "unix:///tmp/telegraf.sock"
   # service_address = "unixgram:///tmp/telegraf.sock"
+
+  ## File mode for unix sockets
+  # file_mode = "777"
 
   ## Maximum number of concurrent connections.
   ## Only applies to stream sockets (e.g. TCP).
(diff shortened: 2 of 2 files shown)


danielnelson:
Thanks for the PR.  Can you make it so the default is not to chmod, so that the umask settings take effect, and also see if you can move the chmod code after the switch statement so you can share that code (alternately use a function).  Finally don't forget to gofmt.

phemmer:
Also relying on permissions of the socket file for security is not a good practice. It works on linux, but it might be a good idea to toss in a warning in the docs:

http://man7.org/linux/man-pages/man7/unix.7.html
>       On Linux, connecting to a stream socket object requires write permis‐
>       sion on that socket; sending a datagram to a datagram socket likewise
>       requires write permission on that socket.  POSIX does not make any
>       statement about the effect of the permissions on a socket file, and
>       on some systems (e.g., older BSDs), the socket permissions are
>       ignored.  Portable programs should not rely on this feature for secu‐
>       rity.

Using `chmod` for security is also racy. Because the socket is created with different permissions at first, a client can connect before they have been adjusted.

The best practice for permissions on unix sockets is to either set the permissions on the directory, or to use [`SCM_CREDENTIALS`](http://man7.org/linux/man-pages/man7/unix.7.html)

danielnelson:
> either set the permissions on the directory...

This seems like something that should be done outside of Telegraf.  Before starting the plugin the user would create a directory with posix acls or setgid or another method to ensure the socket created inside has the right permissions.  I imagine this would work today with no changes.

> to use SCM_CREDENTIALS

I'm new to this, but it looks like this sends us the uid, gid of the connecting user.  At that point we would  verify that it matches our selected user/group in Telegraf application code?

phemmer:
> This seems like something that should be done outside of Telegraf

Yes, sorry. I wasn't proposing that telegraf do this :-)

> I'm new to this, but it looks like this sends us the uid, gid of the connecting user.

Something along those lines. Note that it only sends their primary GID, not their supplementary groups. So you'd have to do a name service lookup on the UID to get all the other groups the user is in.

danielnelson:
Okay, this makes sense to me.  I'm going to close this pull request, we should do the SCM_CREDENTIALS instead.

danielnelson:
It is possible to use this feature to either restrict or loosen permissions to a socket.  Creating a parent directory with tighter permissions and `SCM_CREDENTIALS` are good ways to tighten the permissions.  However, I suspect most actually want the opposite; to allow processes not in the `telegraf` group to write to the socket.

It seems that before `SCM_CREDENTIALS` can be used, you need filesystem access to be able to connect to the socket permissions, so they aren't helpful for loosening permissions.  Instead, it seems that the only methods are by using chmod after socket creation or changing the process umask.

If we wanted to change the umask, we would need to lock any file creation operations to avoid creating files with the wrong permissions.  This isn't undoable but would be a bit of work and a possible source of bugs.

That leaves us with chmod.  Considering only the case where we make the socket more permissive the race between creations and chmod will result in a moment where the the socket exists but connection will result in permission denied.  It does not appear that this can be avoided, but seems like something a client could take in consideration and since the permissions are being loosened it doesn't seem like a security issue.

In the chmod solution where the permissions are being tightened it is a security issue, but as pointed out the socket should already be put into a directory with restrictive permissions instead of relying on the socket perms.

This makes me think we should resurrect this pull request, and add some additional guidance to the sample config and README.

danielnelson:
Closed in #5760

Issue #3849, opened 2018-03-01
Unable to override some settings in /etc/init.d/telegraf


bmailhe:
I override vars `config` and `configdir` in file `/etc/default/telegraf` but they're not taken into account when I restart the service.

### System info:
**version**: telegraf-1.5.2-1.x86_64
**OS** : CentOS release 6.9 

### Steps to reproduce:
Set a new value for var `configdir` in `/etc/default/telegraf` and restart, it is not used.

### Expected behavior:
Be able to override "config" and "configdir" in  `/etc/default/telegraf` 

### Actual behavior:
Only possible to override `TELEGRAF_OPTS`, `USER` and `GROUP` in  `/etc/default/telegraf` 

### Proposal:
Move the section sourcing  `/etc/default/telegraf`  after the variables declaration.

### Use case: 
I want to be able to move the configuration file and directory to another location without modifying the file `/etc/init.d/telegraf`

danielnelson:
Can you add more details about why you want to change the configuration file and directory?

danielnelson:
Closing due to lack of details.

PR #3997, opened 2018-04-10
Override some settings in service startup scripts

gentstr:
#3849

* SysV: CONFIG, CONFDIR, and PIDFILE are now env vars and can be overridden
* Systemd: CONFIG and CONFDIR are now env vars and can be overridden
* Note: Systemd does not allow setting the User from env vars

### Required for all PRs:

No README or unit tests for these changes. Would you like a new README to cover overrides?

### DIFF
--- a/scripts/init.sh
+++ b/scripts/init.sh
@@ -22,10 +22,16 @@
 
 # Command-line options that can be set in /etc/default/telegraf.  These will override
 # any config file values.
+CONFIG=/etc/telegraf/telegraf.conf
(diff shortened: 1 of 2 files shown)


danielnelson:
Can you tell me more about why you need to modify these settings?

danielnelson:
I'm inclined not to make these changes in the official initscripts.  I don't know of any reason the defaults would not be acceptable on the platforms we ship the scripts with.

PR #4000, opened 2018-04-10
Provide aggregated metrics for ntpq input plugin

vvanholl:
### Required for all PRs:

### DIFF
--- a/plugins/inputs/ntpq/README.md
+++ b/plugins/inputs/ntpq/README.md
@@ -35,7 +35,7 @@ server (RMS of difference of multiple time samples, milliseconds);
 
 ### Measurements & Fields:
 
-- ntpq
+- ntpq (Detail about all NTP peers)
     - delay (float, milliseconds)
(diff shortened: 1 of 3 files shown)


vvanholl:
**Why provide aggregated values on NTP from Telegraf ?**
Because we noticed that the actual implementation generates a huge number of series in InfluxDB.

For instance, we have about 800 instances running Telegraf with ntpq input plugin enabled.
This generates up to 50k series.
In most cases having detailed values ( a time serie per remote peer) is not well interesting and aggregated value are just fine.

danielnelson:
Before we discuss aggregation, can you help me understand the cardinality issue a bit better?  I would expect that you would have only a few remote hosts that don't change frequently.  Also, does it help if you use `dns_lookup = true` so that you get hostname?

danielnelson:
Closing due to no response, I think this could probably be done with the basicstats aggregator as well.  Please feel free to comment and we can discuss further.

PR #4001, opened 2018-04-10
New input plugin: Cosmic

borisroman:
Hi!

This PR contains a new input provider. It will collect metrics from the Cosmic infrastructure orchestration platform.

### Required for all PRs:

### DIFF
--- a/plugins/inputs/cosmic/README.md
+++ b/plugins/inputs/cosmic/README.md
@@ -1,0 +1,142 @@
+# Cosmic Input Plugin
+
+This plugin will collect statistics about virtualmachines, volumes and public ip adresses available in Cosmic.
+It uses [Cosmic API](https://apidoc.cosmiccloud.io/) to query the data.
+
+See [Cosmic](https://github.com/MissionCriticalCloud/cosmic/) for more information on the infrastructure orchestrator.
+
+**Series Cardinality Warning**
+
+Depending on the work load of your DC/OS cluster, this plugin can quickly
+create a high number of series which, when unchecked, can cause high load on
+your database.
+
+- Use the
+  [measurement filtering](https://docs.influxdata.com/telegraf/latest/administration/configuration/#measurement-filtering)
+  options to exclude unneeded tags.
+- Write to a database with an appropriate
+  [retention policy](https://docs.influxdata.com/influxdb/latest/guides/downsampling_and_retention/).
+- Limit series cardinality in your database using the
+  [`max-series-per-database`](https://docs.influxdata.com/influxdb/latest/administration/config/#max-series-per-database-1000000) and
+  [`max-values-per-tag`](https://docs.influxdata.com/influxdb/latest/administration/config/#max-values-per-tag-100000) settings.
+- Consider using the
+  [Time Series Index](https://docs.influxdata.com/influxdb/latest/concepts/time-series-index/).
+- Monitor your databases
+  [series cardinality](https://docs.influxdata.com/influxdb/latest/query_language/spec/#show-cardinality).
+
+
+### Configuration:
+
+```toml
+# Gather metrics from virtualmachines available in Cosmic
+[[inputs.cosmic]]
+  ##
+  ## Connection parameters
+  ##
+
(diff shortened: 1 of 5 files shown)


plugins/inputs/cosmic/cosmic.go
```diff
@@ -0,0 +1,216 @@
...
+	return "Gather metrics from resources available in Cosmic"
+}
+
+func (c *Cosmic) SampleConfig() string {
+	return sampleConfig
+}
+
+func (c *Cosmic) Gather(acc telegraf.Accumulator) error {
+	if c.client == nil {
+		c.client = cosmic.NewClient(c.Url, c.Apikey, c.Secretkey, c.VerifySSL)
```
danielnelson: Another feature which would be nice is the ability to configure a timeout.

danielnelson: This is where I'm sticking all my go-cosmic questions :).  What about the possibility of canceling a call to this library, this can be useful during shutdown/restart/etc.  Telegraf doesn't currently have much support for `context.Context`, but this is probably something we will pass in to the `Gather` function in the future.

plugins/inputs/cosmic/README.md
```diff
@@ -0,0 +1,115 @@
...
+
+```toml
+# Gather metrics from virtualmachines available in Cosmic
+[[inputs.cosmic]]
+  ##
+  ## Connection parameters
+  ##
+
+  ## Cosmic API endpoint
+  url = "localhost"
```
danielnelson: Could you write this a full URL with like: `https://localhost`

plugins/inputs/cosmic/README.md
```diff
@@ -0,0 +1,115 @@
+# Cosmic Input Plugin
+
+This plugin will collect statistics about virtualmachines, volumes and public ip adresses available in Cosmic.
+It uses [Cosmic API](https://apidoc.cosmiccloud.io/) to query the data.
+
+See [Cosmic](https://github.com/MissionCriticalCloud/cosmic/) for more information on the infrastructure orchestrator.
+
```
danielnelson: This plugin looks it probably has the potential to create lots of series, similar to many of container orchestration plugins that use UUIDs.  If there are any opportunities to remove UUIDs in favor of names, without it causing problems due to lack of uniqueness, it might allow users to be able to store data for longer.  This is especially useful if the UUID changes frequently, and less so on things that rarely change like user UUIDs.

Ultimately there is probably limited changes you can make, so perhaps we should add the "Series Cardinality Warning" like the [DC/OS](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/dcos#dcos-input-plugin) has?

borisroman:
@danielnelson Thanks for your comments! I'll incorporate and answer them next week.

borisroman:
Hi @danielnelson 

I've incorporated all comments in the PR. The only thing I still need to do is to add the option to cancel a call to the library.

How would this actually look like from a Telegraf perspective?

Cheers

danielnelson:
In the future we will add context.Context to interfaces such as Input:
```diff
type Input interface {
-    Gather(Accumulator) error
+    Gather(context.Context, Accumulator) error
}
```
The agent could then cancel the Context using [WithCancel](https://golang.org/pkg/context/#WithCancel); the plugin itself would probably still be handle timeouts using [WithTimout](https://golang.org/pkg/context/#WithTimeout).

It would be the plugins (and the libraries it uses) job to react and return when the Context is cancelled or the timeout is reached.  Adding timeouts are pretty straigtforward from what I have seen, cancel is more complicated.

Some network operations, such as HTTP, are not a problem to cancel. Others, such as TCP connections, are [not so simple to cancel](https://medium.com/@zombiezen/canceling-i-o-in-go-capn-proto-5ae8c09c5b29) in Go, to the point where it is probably is not worth attempting to support it.  Hopefully there will be improvements in the future.

The change to the Input interface is probably a ways off, but we can make it smoother if plugins create their own context and make use of it internally.

I hope this answered your question.

PR #4010, opened 2018-04-12
Microsoft Application Insights output plugin

karolz-ms:
This PR adds an output plugin that sends data to Microsoft Application Insights https://docs.microsoft.com/azure/application-insights/

(this is a revised and improved version of https://github.com/influxdata/telegraf/pull/3974)

Hopefully this addition aligns with telegraf roadmap, but if there are concerns, we (here at Microsoft) are open to other suggestions--thanks!

### Required for all PRs:


karolz-ms:
@brahmnes @yantang-msft FYI

plugins/outputs/application_insights/application_insights.go
```diff
@@ -0,0 +1,427 @@
...
+func (a *ApplicationInsights) SampleConfig() string {
+	return sampleConfig
+}
+
+func (a *ApplicationInsights) Description() string {
+	return "Send telegraf metrics to Azure Application Insights"
+}
+
+func (a *ApplicationInsights) Connect() error {
+	if a.Timeout.Duration == 0 {
```
yantang-msft: So it's never possible that timeout is 0s? The doc said it can be set to 0s meaning no timeout.

plugins/outputs/application_insights/application_insights.go
```diff
@@ -0,0 +1,427 @@
...
+		a.Timeout = internal.Duration{Duration: time.Second * 5}
+	}
+
+	if a.transmitter == nil && a.InstrumentationKey != "" {
+		a.transmitter = NewAppinsightsTransmitter(a.InstrumentationKey)
+	}
+
+	if a.EnableDiagnosticLogging && a.diagMsgSubscriber != nil {
```
yantang-msft: I feel like this subscriber can be dropped.

plugins/outputs/application_insights/application_insights.go
```diff
@@ -0,0 +1,427 @@
...
+	hasCount := metric.HasField("count") || metric.HasField("samples")
+	return hasSum && hasCount
+}
+
+func (a *ApplicationInsights) createTelemetry(metric telegraf.Metric) []appinsights.Telemetry {
+	if canBeTreatedAsAggregate(metric) {
+		aggregateTelemetry, usedFields := a.createAggregateMetricTelemetry(metric)
+		if aggregateTelemetry != nil {
+			remainingTelemetry := a.createTelemetryForUnusedFields(metric, usedFields)
+			retval := make([]appinsights.Telemetry, 1, 1+len(remainingTelemetry))
```
yantang-msft: nit: I just checked the go tutorial, and they do add space between +.
I didn't find any example for string format, i.e., space between "+" for "Output [application_insights] "+format

plugins/outputs/application_insights/application_insights.go
```diff
@@ -0,0 +1,427 @@
...
+		}
+
+		return nil
+	}
+
+	return fmt.Errorf("No field from the candidate list was found in the metric")
+}
+
+func contains(set []string, val string) bool {
+	for _, elem := range set {
```
yantang-msft: It will throw if set is nil correct?

plugins/outputs/application_insights/application_insights.go
```diff
@@ -0,0 +1,427 @@
...
+		telemetry := a.createSimpleMetricTelemetry(metric, metric.FieldList()[0].Key, false)
+		if telemetry != nil {
+			return []appinsights.Telemetry{telemetry}
+		} else {
+			return nil
+		}
+	} else {
+		// AppInsights does not support multi-dimensional metrics at the moment, so we need to disambiguate resulting telemetry
+		// by adding field name as the telemetry name suffix
+		retval := a.createTelemetryForUnusedFields(metric, nil)
```
yantang-msft: Looks like it will throw in the contains function since the usedFields is nil?

plugins/outputs/application_insights/application_insights.go
```diff
@@ -0,0 +1,427 @@
...
+			retval = append(retval, telemetry)
+		}
+	}
+
+	return retval
+}
+
+func (a *ApplicationInsights) addContextTags(metric telegraf.Metric, telemetry appinsights.Telemetry) {
+	for tagSourceName, contextTagName := range a.ContextTagSources {
+		contextTagValue, found := metric.GetTag(tagSourceName)
```
yantang-msft: I saw you use style like below before, which looks pretty concise
if contextTagValue, found := metric.GetTag(tagSourceName); found {
      telemetry.ContextTags()[contextTagName] = contextTagValue 
}

plugins/outputs/application_insights/application_insights.go
```diff
@@ -0,0 +1,427 @@
...
+## Instrumentation key of the Application Insights resource.
+instrumentationKey = "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxx"
+
+## Timeout on close. If not provided, will default to 5s. 0s means no timeout (not recommended).
+# timeout = "5s"
+
+## Determines whether diagnostic logging (for Application Insights endpoint traffic) is enabled. Default is false.
+# enableDiagnosticLogging = "true"
+
+## ContextTagSources dictionary instructs the Application Insights plugin to set Application Insights context tags
```
yantang-msft: Add link to contextTagKeys
https://github.com/Microsoft/ApplicationInsights-dotnet/blob/develop/Schema/PublicSchema/ContextTagKeys.bond

plugins/outputs/application_insights/application_insights.go
karolz-ms: Agreed, this is not the right place to do apply the default. Will fix

plugins/outputs/application_insights/application_insights.go
karolz-ms: It is used by tests

plugins/outputs/application_insights/application_insights.go
karolz-ms: Go fmt formats the code without spaces around the + operator

plugins/outputs/application_insights/application_insights.go
karolz-ms: Nope. This is one of the nice things about Golang: nil values are generally useful/harmless. In this case, ranging over nil map is a no-op.

plugins/outputs/application_insights/application_insights.go
karolz-ms: True, but in this particular case this idiom cannot be used because I need the contextTagValue two lines below

plugins/outputs/application_insights/application_insights.go
karolz-ms: OK

plugins/outputs/application_insights/application_insights.go
yantang-msft: Just tested if it's typed then it's fine, but throws if it's untyped nil.

plugins/outputs/application_insights/application_insights.go
yantang-msft: The contextTagValue is used inside the scope?

plugins/outputs/application_insights/application_insights.go
karolz-ms: I see. I do not think this case applies to the plugin code

plugins/outputs/application_insights/application_insights.go
karolz-ms: Oh, I see what you mean. Will fix, thx

plugins/outputs/application_insights/appinsightsDiagnosticMsg.go
```diff
@@ -0,0 +1,12 @@
+package application_insights
```
danielnelson: Rename the source files using snake_case, I would probably call this file `subscriber.go`

plugins/outputs/application_insights/appinsightsTransmitter.go
```diff
@@ -0,0 +1,19 @@
+package application_insights
```
danielnelson: Rename file `transmitter.go`

plugins/outputs/application_insights/application_insights.go
```diff
@@ -0,0 +1,422 @@
...
+func (a *ApplicationInsights) SampleConfig() string {
+	return sampleConfig
+}
+
+func (a *ApplicationInsights) Description() string {
+	return "Send telegraf metrics to Azure Application Insights"
+}
+
+func (a *ApplicationInsights) Connect() error {
+	if a.transmitter == nil && a.InstrumentationKey != "" {
```
danielnelson: If the instrumentation key is unset, should we return an error?

plugins/outputs/application_insights/application_insights.go
```diff
@@ -0,0 +1,422 @@
...
+		}
+
+		return int(v), nil
+	}
+
+	return 0.0, fmt.Errorf("[%s] cannot be converted to an int value", value)
+}
+
+func logOutputMsg(format string, v ...interface{}) {
+	log.Printf("Output [application_insights] "+format, v...)
```
danielnelson: Use a log level and format like: `"E! [outputs.application_insights] "`.

plugins/outputs/application_insights/application_insights.go
```diff
@@ -0,0 +1,422 @@
...
+			remainingTelemetry := a.createTelemetryForUnusedFields(metric, usedFields)
+			retval := make([]appinsights.Telemetry, 1, 1+len(remainingTelemetry))
+			retval[0] = aggregateTelemetry
+			retval = append(retval, remainingTelemetry...)
+			return retval
+		}
+	}
+
+	fields := metric.Fields()
+	if len(fields) == 1 {
```
danielnelson: I think we should try to use the same naming strategy regardless of the number of fields, this will be simpler to understand.

plugins/outputs/application_insights/application_insights.go
```diff
@@ -0,0 +1,422 @@
...
+			telemetry.ContextTags()[contextTagName] = contextTagValue
+		}
+	}
+}
+
+func setFloat64TelemetryProperty(
+	telemetryProperty *float64,
+	candidateFields []string,
+	metric telegraf.Metric,
+	usedFields *[]string) error {
```
danielnelson: I would have this function return `(float64, error)` and return the telemetryPropery.  No need to pass the usedFields by pointer it will work the same.

plugins/outputs/application_insights/application_insights.go
```diff
@@ -0,0 +1,422 @@
...
+	for _, elem := range set {
+		if elem == val {
+			return true
+		}
+	}
+
+	return false
+}
+
+func toFloat64(value interface{}) (float64, error) {
```
danielnelson: You can shorten this list, fields will be one of int64, uint64, float64, bool, string.  Keep the error message but in theory it should never trigger.

plugins/outputs/application_insights/application_insights.go
```diff
@@ -0,0 +1,422 @@
...
+	case float32:
+		return float64(v), nil
+	case float64:
+		return v, nil
+	}
+
+	return 0.0, fmt.Errorf("[%s] cannot be converted to a float64 value", value)
+}
+
+func toInt(value interface{}) (int, error) {
```
danielnelson: Consider returning an int64 and eliminating the 32bit cases?  Same comment as on the toFloat function, this will either be an int64, uint64, float64, bool or string.

plugins/outputs/application_insights/application_insights.go
```diff
@@ -0,0 +1,422 @@
...
+}
+
+func (a *ApplicationInsights) createTelemetry(metric telegraf.Metric) []appinsights.Telemetry {
+	if canBeTreatedAsAggregate(metric) {
+		aggregateTelemetry, usedFields := a.createAggregateMetricTelemetry(metric)
+		if aggregateTelemetry != nil {
+			remainingTelemetry := a.createTelemetryForUnusedFields(metric, usedFields)
+			retval := make([]appinsights.Telemetry, 1, 1+len(remainingTelemetry))
+			retval[0] = aggregateTelemetry
+			retval = append(retval, remainingTelemetry...)
```
danielnelson: Does the aggregate telemetry need to be first?  If not just append to remainingTelemetry and name variable `telemetry`.

plugins/outputs/application_insights/application_insights.go
```diff
@@ -0,0 +1,422 @@
...
+			// It is very common to set Cloud.RoleName and Cloud.RoleInstance context properties, hence initial capacity of two
+			ContextTagSources: make(map[string]string, 2),
+		}
+	})
+}
+
+func canBeTreatedAsAggregate(metric telegraf.Metric) bool {
+	// If we can identify the Sum and Count properties, then we are good
+	hasSum := metric.HasField("value") || metric.HasField("sum")
+	hasCount := metric.HasField("count") || metric.HasField("samples")
```
danielnelson: These methods are just as expensive as getting the field value, it may be faster to just try to create aggregates instead of testing up front.

plugins/outputs/application_insights/application_insights.go
```diff
@@ -0,0 +1,422 @@
...
+			logOutputMsg("%s", msg)
+			return nil
+		})
+	}
+
+	return nil
+}
+
+func (a *ApplicationInsights) Write(metrics []telegraf.Metric) error {
+	if len(metrics) == 0 || a.transmitter == nil || a.InstrumentationKey == "" {
```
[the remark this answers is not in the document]
karolz-ms: Will remove the metric length test.

Regarding missing instrumentation key, the standard procedure for Application Insights is to no-op if the key is missing, but otherwise continue with the work. I will emit a warning, but I am pretty sure I do not want to stop Telegraf just because AppInsights instrumentation key is missing.

plugins/outputs/application_insights/application_insights.go
karolz-ms: I talked about this with the owners of Application Insights service--this naming strategy was the solution they suggested. We have a proof-of-concept app that uses single- and multi-dimensional metrics and it really isn't that difficult to reconcile which is which in the Application Insights portal as long as the strategy is followed consistently--which it is in this implementation.

plugins/outputs/application_insights/application_insights.go
karolz-ms: I am not sure if applying this suggestion is an improvement

If the `usedFields` slice is not passed as a pointer, the modifications made to it will not be visible outside of `setFloat64TelemetryProperty` (which is required)

As to returning `(float64, error)` this will complicate consumer code. The point of the method is that the modification succeeds _sometimes, but not always_. If I return a value and error tuple, I will need an extra variable to grab the returned value, check the error, and if the error is nil, set the telemetry property to the actual value. This is more code than just calling `setFloat64TelemetryProperty` passing the address of the telemetry property to set. 

There is a few places in the code where the property set is a "best effort" and `setFloat64TelemetryProperty` is especially useful in this case because one can just call it and ignore the returned error.

karolz-ms:
@danielnelson thank you for a thorough review, much appreciated. I believe I have addressed all your comments, but please let me know if I missed something or if you disagree with what I have done.

Regarding https://github.com/influxdata/telegraf/pull/4089 this is, honestly, work in progress. The Azure Monitor folks are coming from the infrastructure monitoring angle; my contribution is targeting Application Insights, which is an application monitoring solution, with a different set of experiences. We are in the middle of working out unified story for metrics on Azure, where you will be able to visualize, query and alert on metrics for everything, infra and applications, but it will take some time to get us there. In the meantime we want to give users of Application Insights more flexibility in terms of  feeding metrics into Application Insights and Telegraf is a perfect fit for that. This is the best answer I can give you at the moment.

plugins/outputs/application_insights/application_insights.go
danielnelson: It's not guaranteed that metrics will be delivered together, so you could have fields changing names.  Imagine that an input plugin is running for multiple hosts but on one of them a value is not available for a period:
```
foo,host=a first=42,second=43 1525292885000000000
foo,host=b first=42 1525292885000000000

foo,host=a first=42,second=43 1525293034000000000
foo,host=b first=42,second=43 1525293034000000000
```

This would result in the `first` field being renamed during one write, which seems undesired.

plugins/outputs/application_insights/application_insights.go
danielnelson: Slices work like a struct containing a pointer to the underlying data, so no copy will be made and the changes will be visible.  It will work either way but passing a pointer to a slice is not Go convention, and is never used in the Go standard library.

On returning `(float64, error)`, again it is just Go convention.  Keep in mind that float64 types are already set to a zero value and so you would only have to conditionally assign if you have set a default value that you want to preserve.

plugins/outputs/application_insights/application_insights.go
karolz-ms: I see your point. What about a metric which always comes as 

```
foo,host=a value=23 1525293034000000000
```

Do you think it makes sense to special-case single-field metric with a "value" field? Or should the output send a "foo.value" telemetry to AppInsights regardless?

danielnelson: I think this could be okay, we do something similar in the prometheus output.  Strictly speaking though it does introduce some corner cases.  Let's make sure to document the conversions in the README.

plugins/outputs/application_insights/application_insights.go
karolz-ms: Maybe I am misunderstanding what you are saying regarding the slice point, but run this on Golang play--there is a difference:

```
package main

import (
	"fmt"
)

func main() {
  var sl []string
  changeSlice(sl, "Foo")
  fmt.Printf("%v\n", sl)
  changeSlicePointer(&sl, "Bar")
  fmt.Printf("%v\n", sl)
}

func changeSlice(sl []string, new string) {
  sl = append(sl, new)
}

func changeSlicePointer(sl *[]string, new string) {
  *sl = append(*sl, new)
}
```

Good point regarding zero value, I will change that

danielnelson: I see, because we are appending.  I guess the Go way would probably be to return the new slice (just as append does) but I'll let you decide.

plugins/outputs/application_insights/application_insights.go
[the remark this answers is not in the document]
karolz-ms: Two reasons:

1. There might be other outputs that are functional and sending data somewhere
1. If Telegraf is used as a basis for metrics sidecar container in a Kubernetes environment, the failure of metrics container will cause the whole pod to be considered Failed and subject to relocation & restarts, affecting the primary pod functionality. IMO it is better to just warn about invalid configuration, but let the main container do its job, because that is the most important thing. Lack of metrics from such pod will be, hopefully, noticed fairly quickly, but the primary functionality will not be affected.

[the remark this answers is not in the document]
danielnelson: I don't think we should support having the plugin both enabled and non-functional, it will just cause confusion and can be resolved before runtime, it should be easy enough to just not enable the plugin if no output is desired.  Also, this is contrary to how all the other outputs work so I don't want to introduce a different behavior.

It might be helpful to use the `--config-directory` to more easily enable/disable plugins.

plugins/outputs/application_insights/application_insights.go
karolz-ms: OK, I have changed the return value but kept the slice update via pointer argument

plugins/outputs/application_insights/application_insights.go
karolz-ms: Done--updated README (which was also missing some config-related information) and added a test verifying the new naming logic.

PR #3191, opened 2017-08-31
TLS and MTLS enhancements to HTTPListener input plugin

allingeek:
This enhancement allows a user to enable TLS for the endpoint by providing certificate and private key files. They can use private PKI by specifying a list of certificate authority file names. Mutual TLS authentication can be enabled by providing a list of allowed client CA certificate files.

Signed-off-by: Jeff Nickoloff <jeff@allingeek.com>

### Required for all PRs:

- [x] Signed [CLA](https://influxdata.com/community/cla/).
- [x] Associated README.md updated.
- [x] Has appropriate unit tests.

PR #4021, opened 2018-04-14
Add TLS support to socket_writer and socket_listener plugins

### LINKED ISSUE #2866: Add TLS/SSL encryption support to socket_writer and socket_listener
### Proposal:
Add SSL encryption support to the socket_writer and socket_listener plugins with client certificate support.

### Current behavior:
socket_writer allows to send data over tcp or udp. socket_listener allows to receive data over tcp or udp. This data is transferred in cleartext over the connection.

### Desired behavior:
It should be possible to encrypt and authenticate the data transferred preferably using tls/ssl and authenticate the writer and listener using certificates.

### Use case:
I want to authenticate the data sent by the clients by using the cn field of a client certificate specific for each client. This would ensure that a compromised client cannot send measurements for other clients.

bobmshannon:
Fixes #2866.
 
### Required for all PRs:


internal/internal.go
```diff
@@ -112,12 +112,21 @@ func RandomString(n int) string {
...
 	InsecureSkipVerify bool,
+) (*tls.Config, error) {
+	return GetClientTLSConfig(SSLCert, SSLKey, SSLCA, InsecureSkipVerify)
+}
+
+// GetClientTLSConfig gets a tls.Config object from the given certs, key, and CA files
+// for use with a client.
+// The full path to each file must be provided.
+// Returns a nil pointer if all files are blank and InsecureSkipVerify=false.
+func GetClientTLSConfig(
```
danielnelson: Just leave this as GetTLSConfig for now.

plugins/inputs/socket_listener/README.md
```diff
@@ -35,6 +35,15 @@ This is a sample configuration for the plugin.
...
   # read_timeout = "30s"
 
+  ## Optional SSL configuration.
+  ## Only applies to stream sockets (e.g. TCP).
+  # ssl_cert = "/etc/telegraf/cert.pem"
+  # ssl_key = "/etc/telegraf/key.pem"
+  ## Enable and require client certificate authentication.
+  # ssl_client_auth = false
+  ## CAs used to verify client certificates.
+  # ssl_ca = ["/etc/telegraf/ca.pem"]
```
danielnelson: We have a little bit of prior art on the server side: #3191, I would prefer if we use the same variable names, but let me know if you think that is a problem.  The biggest difference is that client_auth is enabled automatically if there are cacerts specified.

internal/internal.go
bobmshannon: Done.

plugins/inputs/socket_listener/README.md
bobmshannon: Makes sense -- definitely want to stay consistent. The config now matches those variable names.

bobmshannon:
I've addressed the comments, however, I still need to investigate why the tests are failing.

bobmshannon:
OK, tests should be sorted now.

danielnelson:
Thanks!

plugins/inputs/socket_listener/socket_listener.go
```diff
@@ -313,6 +338,14 @@ func (uc unixCloser) Close() error {
 	return err
 }
 
+func (uc unixCloser) Accept() (net.Conn, error) {
+	return uc.closer.(net.Listener).Accept()
```
phemmer: if we're assuming that `uc.closer` is always going to be a `net.Listener`, you can avoid the type assertion by redefining `uc.closer` to be a `net.Listener`.
This is good practice anyway as then you get compile time verification that you never assign anything other than a `net.Listener` to `uc.closer`, as if you do otherwise this code will panic.

plugins/outputs/socket_writer/socket_writer.go
```diff
@@ -58,12 +70,26 @@ func (sw *SocketWriter) SetSerializer(s serializers.Serializer) {
 }
 
 func (sw *SocketWriter) Connect() error {
+	var (
+		c   net.Conn
+		err error
+	)
```
phemmer: nit: it's idiomatic go to declare variables right before they're used, not at the top of the function.
Also `err` doesn't need to be declared at all. It'll get declared on line 83.

danielnelson:
Thanks, I'll get these.

plugins/inputs/socket_listener/socket_listener.go
danielnelson: I don't see why we need these functions at all...

phemmer: if you embed `net.Listener` directly within `unixCloser` so that `unixCloser` picks up `net.Listener`'s methods, then yeah, you shouldn't.

danielnelson:
1486ae25c025706f915abe7638367453c878165a

plugins/inputs/socket_listener/socket_listener.go
phemmer: Oh, you mean they aren't needed because they're not even called. yeah... :-)

Issue #3791, opened 2018-02-14
Add an input plugin for jenkins "Global Build Stats"


fchiorascu:
New Feature:

Kindly is possible to have:
1\. Input plugins: [[inputs.jenkins]] ?
2\. This oprion it seems that already exists. InfluxDB plugin: https://plugins.jenkins.io/influxdb

The second option I think is more reliable, but for 1\. will be developed something in future?

Kind Regards,

danielnelson:
This would be nice, we also have a pull request open for a plugin to monitor Jenkins itself: https://github.com/influxdata/telegraf/pull/3292

lpic10:
I have something working with that already internally. It makes use of [gojenkins](https://github.com/bndr/gojenkins) that uses only Jenkins API (so no need to install a jenkins plugin) to retrieve jobs data (job name, id, duration, result) and some monitoring info from the nodes like memory, swap and disk space. 

I'm my internal fork I call this input plugin "jenkins", but I see there are #3292 that uses this name as well.

If you are interested I can do some cleaning and open a PR, but I'd need to rename it.

danielnelson:
Searching for metrics at https://plugins.jenkins.io/, it seems there are quite a few options.  It should already be possible to use the influxdb plugin sending to http_listener, the prometheus exporter, the graphite plugin, etc.  It should also be possible to replicate #3292 now in master by using the http input with the dropwizard parser.

Perhaps it's not worth creating more code for this considering there are so many possible integration points?

I do see a big advantage to your design @lpic10, since you wouldn't need to add any plugins.  Can you post the output of `telegraf --input-filter jenkins --test`?

fchiorascu:
I think will be great to have:
- [[inputs.jenkins]]
- and also to monitor jenkins itself.

Kind Regards,

lpic10:
@danielnelson , see below from an initial readme I wrote. 

There are 2 important details:

The jenkins API is not very efficient, so to gather all this data a lot of http calls are needed (to get a list of jobs, then to get a list of builds for a particular job, then for each build). This can escalate rather quickly depending on the number of jobs and builds in a jenkins instance. That's the reason I implemented two basic filters, one by job name and another one to ignore old builds (as telegraf will be collecting data regularly I don't need to get same data again and again). 

The timestamp of the job build is used instead of the collection time. This works fine with influxDB as it does not duplicate the point, I don't know about other output plugins.

[...]

PR #4022, opened 2018-04-16
Add jenkins input plugin based on gojenkins

lpic10:
### Required for all PRs:

As discussed on #3791 this telegraf input plugin does not require a plugin on jenkins side.
See https://github.com/influxdata/telegraf/issues/3791#issuecomment-366799524


danielnelson:
Thanks @lpic10, I will try to review this soon.  On the CI failure, it probably makes sense to modify the Godeps by hand for now since I caused some problems with `gdm save` when I combined the Windows and Linux Godeps into one file.

lpic10:
Ok, I noticed there were too many changes to Godeps... I will resend changing it by hand later today

plugins/inputs/jenkins/jenkins.go
```diff
@@ -0,0 +1,265 @@
...
+func (j *Jenkins) GetNodesData(nodes []*gojenkins.Node, acc telegraf.Accumulator) error {
+
+	measurement := "jenkins_node"
+
+	tags := map[string]string{}
+	fields := make(map[string]interface{})
+
+	// get node data
+	for _, node := range nodes {
+		info, _ := node.Info()
```
danielnelson: If Info fails, I think we should `continue`.

plugins/inputs/jenkins/jenkins.go
```diff
@@ -0,0 +1,265 @@
...
+		tags["node_name"] = node.GetName()
+
+		// skip nodes without information
+		if info.MonitorData.Hudson_NodeMonitors_ArchitectureMonitor == nil {
+			continue
+		}
+
+		tags["arch"] = info.MonitorData.Hudson_NodeMonitors_ArchitectureMonitor.(string)
+		isOnline, _ := node.IsOnline()
```
danielnelson: Add a comment about why we are ignoring the error

plugins/inputs/jenkins/jenkins.go
```diff
@@ -0,0 +1,265 @@
...
+
+	LastbuildFilterInterval internal.Duration `toml:"lastbuild_interval"`
+	JobFilterName           []string          `toml:"job_exclude"`
+}
+
+type byBuildNumber []gojenkins.JobBuild
+
+func (a byBuildNumber) Len() int           { return len(a) }
+func (a byBuildNumber) Swap(i, j int)      { a[i], a[j] = a[j], a[i] }
+func (a byBuildNumber) Less(i, j int) bool { return a[i].Number > a[j].Number }
```
danielnelson: I suggest using `sort.Slice` instead of defining these funcs

plugins/inputs/jenkins/jenkins.go
```diff
@@ -0,0 +1,265 @@
...
+			buildSecAgo := time.Now().Sub(jobLastBuild.GetTimestamp()).Seconds()
+
+			if buildSecAgo > j.LastbuildFilterInterval.Duration.Seconds() {
+				log.Printf("D! Last job too old, last %s build was %v seconds ago", jobName, buildSecAgo)
+				continue
+			}
+		}
+
+		buildIds, err := job.GetAllBuildIds()
+		sort.Sort(byBuildNumber(buildIds))
```
danielnelson: sort.Slice, move under the err check

plugins/inputs/jenkins/jenkins.go
```diff
@@ -0,0 +1,265 @@
...
+
+		if err != nil {
+			log.Printf("E! Error retrieving all builds from job \"%s\": %s", jobName, err.Error())
+			return err
+		}
+
+		for _, buildId := range buildIds {
+			build, err := job.GetBuild(buildId.Number)
+			log.Printf("D! Reading data from job %s build %v", jobName, buildId.Number)
```
danielnelson: I would remove this log message, if the build does not result in a metric we will get another log message so I think we are able to debug in either case.

plugins/inputs/jenkins/jenkins.go
```diff
@@ -0,0 +1,265 @@
...
+					break
+				}
+			}
+
+			tags := map[string]string{"job_name": jobName, "result": build.GetResult()}
+			fields := make(map[string]interface{})
+			fields["duration"] = build.GetDuration()
+			fields["result_code"] = mapResultCode(build.GetResult())
+
+			acc.AddFields(measurement, fields, tags, build.GetTimestamp())
```
danielnelson: Will a completed build ever change?  I wonder if we should make any effort to ignore builds we have already collected?

plugins/inputs/jenkins/jenkins.go
```diff
@@ -0,0 +1,265 @@
...
+
+				if time.Now().Sub(build.GetTimestamp()).Seconds() > j.LastbuildFilterInterval.Duration.Seconds() {
+					log.Printf("D! Job %s build %v too old (%v seconds ago), skipping to next job", jobName, buildId.Number, buildSecAgo)
+					break
+				}
+			}
+
+			tags := map[string]string{"job_name": jobName, "result": build.GetResult()}
+			fields := make(map[string]interface{})
+			fields["duration"] = build.GetDuration()
```
danielnelson: What are the units?  Let's name it with the units, ex: `duration_ms`

plugins/inputs/jenkins/jenkins_test.go
```diff
@@ -0,0 +1,352 @@
...
+		case "/api/json":
+			w.WriteHeader(http.StatusOK)
+			resp = response
+		default:
+			w.WriteHeader(http.StatusNotFound)
+		}
+		w.Write([]byte(resp))
+	}))
+
+	l, _ := net.Listen("tcp", "127.0.0.1:8080")
```
danielnelson: I would probably just use httptest.NewServer, but if you listen separately then use port 0 to get a auto assigned port.

plugins/inputs/jenkins/jenkins.go
```diff
@@ -0,0 +1,265 @@
...
+
+  ## Optional SSL Config
+  # ssl_ca = /path/to/cafile
+  # ssl_cert = /path/to/certfile
+  # ssl_key = /path/to/keyfile
+  ## Use SSL but skip chain & host verification
+  # insecure_skip_verify = false
+
+  ## Job & build filter
+  lastbuild_interval = "1h"
```
danielnelson: I wasn't sure what this did until I read the code, maybe `max_build_age` would be a better name?

plugins/inputs/jenkins/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+- jenkins_job
+    - job_name
+    - result
+
+### Example Output:
+
+```
+$ ./telegraf --config telegraf.conf --input-filter jenkins --test
+jenkins_node,arch=Linux\ (amd64),disk_path=/var/jenkins_home,temp_path=/tmp,host=myhost,node_name=master swap_total=4294963200,memory_available=586711040,memory_total=6089498624,online=true,response_time=1000i,disk_available=152392036352,temp_available=152392036352,swap_available=3503263744 1516031535000000000
```
danielnelson: I think we should make `online` a tag.

plugins/inputs/jenkins/jenkins.go
```diff
@@ -0,0 +1,265 @@
...
+type byBuildNumber []gojenkins.JobBuild
+
+func (a byBuildNumber) Len() int           { return len(a) }
+func (a byBuildNumber) Swap(i, j int)      { a[i], a[j] = a[j], a[i] }
+func (a byBuildNumber) Less(i, j int) bool { return a[i].Number > a[j].Number }
+
+const sampleConfig = `
+  url = "http://my-jenkins-instance:8080"
+  user = "admin"
+  passwd = "admin"
```
danielnelson: Super nit-pick, but can we call this `username` and `password` only because it matches more with existing plugins.

lpic10:
Thanks for the review @danielnelson. I'm unavailable until end of next week, I will send the updates after.

plugins/inputs/jenkins/jenkins.go
lpic10: AFAIK a completed build should not change. How do you see this could be implemented? Maybe keeping a list of last build ID & timestamp collected per jenkins instance?

plugins/inputs/jenkins/jenkins.go
lpic10: Yes, it is milliseconds. That is a breaking change for me, I will need to see how I can change my data in InfluxDB :)

plugins/inputs/jenkins/jenkins.go
lpic10: Ok, I agree, I'm definitely not good at naming stuff

plugins/inputs/jenkins/jenkins_test.go
lpic10: Ok, I will have a look at that

plugins/inputs/jenkins/jenkins.go
danielnelson: It might be helpful to look at this [migration guide for the mysql plugin](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/mysql#metric-version), you could change to the TICKScript to (untested):
```
batch
    |query('''
        SELECT duration as duration_ms FROM "telegraf"."autogen"."jenkins"
    ''')
        .period(5m)
        .every(5m)
	|influxDBOut()
		.database('telegraf')
		.retentionPolicy('autogen')
```

plugins/inputs/jenkins/jenkins.go
danielnelson: Perhaps we could store the last reported build per job in a `map[string]int64`? If there is a gap we could record the last build before the gap for simplicity:

lastCollectedBuild = 1
```
build status
1     success
2     running
3     success
```

We would probably need to clear the previous map each Gather to remove Jobs that no longer exist.

plugins/inputs/jenkins/jenkins.go
```diff
@@ -0,0 +1,265 @@
...
+	instance, err := gojenkins.CreateJenkins(httpclient, j.URL, j.User, j.Passwd).Init()
+	if err != nil {
+		return fmt.Errorf("E! It was not possible to connect to the Jenkins instance\n")
+	}
+
+	j.Instance = instance
+
+	nodes, err := instance.GetAllNodes()
+	if err != nil {
+		return fmt.Errorf("E! Something went wrong retrieving nodes information from Jenkins\n")
```
danielnelson: Make sure to show the error, I would use a format like this for errors.
```
return fmt.Errorf("error retrieving nodes[%s]: %v", j.URL, err)
```

Telegraf will add the prefix `E! Error in plugin [jenkins]: ` when and log it.

danielnelson:
@lpic10 Do you mind if we hand this plugin over to one of my colleagues to complete?

lpic10:
> @lpic10 Do you mind if we hand this plugin over to one of my colleagues to complete?

No, not at all. I had implemented most of the changes you requested, just pushed now so you can review it.  Only part missing in my view is the improvement to ignore builds already collected.

kelwang:
Hi @lpic10 , do you mind we only support Jenkins version >= 2.35? Since 2.35, Jenkins API support 2 fields "fullName" and "fullDisplayName", which will include the parents name of a sub job.

PR #4035, opened 2018-04-17
Add docker input server version

### LINKED ISSUE #4029: FeatureRequest: add docker engine version to inputs.docker
Having `docker_version` as a tag in `inputs.docker` would be great.

danielnelson:
closes #4029

### Required for all PRs:

### DIFF
--- a/plugins/inputs/docker/README.md
+++ b/plugins/inputs/docker/README.md
@@ -181,19 +181,24 @@ based on the availability of per-cpu stats on your system.
 - docker (memory_total)
     - unit=bytes
     - engine_host
+    - server_version
 - docker (pool_blocksize)
     - unit=bytes
     - engine_host
+    - server_version
 - docker_data
     - unit=bytes
     - engine_host
+    - server_version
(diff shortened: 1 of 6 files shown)


danielnelson:
@rdxmb Can you review?

rdxmb:
Oh, I am not a Go-developer, so I cannot make a "real review". But, I can have a look at it, of course.

What I think about is if we need the tag really with all measurements. Having it with `docker` could be enough, I guess.

danielnelson:
Hmm, maybe it should be a string field then, what sort of queries are you planning to use?

rdxmb:
As mostly the strings I use are stored as tags, I did not have the idea to use a string field for that.

However, for anybody else a use-case could also be to get what engine-version is running **under a container**. For that, your PR would be the best.

My personal use case is to just check if the docker-engines are up-to-date. So for now I am very flexible about the query.

danielnelson:
I think having it as a tag to all of the docker_* measurements will end up be the least confusing.

PR #4050, opened 2018-04-20
Deprecate the cassandra input plugin

### LINKED ISSUE #4049: Deprecate cassandra input
### Proposal:

I believe we should deprecate the `cassandra` input.  This plugin uses the jolokia API to collect metrics, but is not as fast, configurable, or bug free as the jolokia2 input.  If we deprecate it we will have less code to maintain when it is removed.

### Current behavior:

cassandra input is not deprecated

### Desired behavior:

- cassandra input is deprecated
- sample cassandra configuration exists for use with the jolokis2 input

danielnelson:
closes #4049

### Required for all PRs:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 ## v1.7 [unreleased]
 
+### Release Notes
+
+- The `cassandra` input plugin has been deprecated in favor of the `jolokia2`
+  input plugin which is much more configurable and more performant.  There is
+  an [example configuration](./plugins/inputs/jolokia2/examples) to help you
+  get started.
+
 ### New Inputs
 
 - [fibaro](./plugins/inputs/fibaro/README.md) - Contributed by @dynek
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@ and writing metrics.
 
 Design goals are to have a minimal memory footprint with a plugin system so
 that developers in the community can easily add support for collecting metrics
-from local or remote services.
+.  For an example configuration referencet from local or remote services.
 
 Telegraf is plugin-driven and has the concept of 4 distinct plugins:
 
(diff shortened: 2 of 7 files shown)


danielnelson:
cc @dylanmei @subhachandrachandra @albundy83

The Cassandra plugin had a good run, but the jolokia2 plugin is just more flexible and much faster due to the bulk requests.

Can you all take a look at the sample cassandra.conf I added and give me any feedback, it's my first jolokia2 config and I don't use Cassandra in any meaningful way, so I'm sure it can be improved.

albundy83:
You first try is a good try :)
However, to try to be closer with current cassandra plugin, it could be good to prefix all metrics with "cassandra" and java metrics with "java".
That's what is done here : https://github.com/albundy83/telegraf/blob/cassandra-jolokia2-examples/plugins/inputs/cassandra/cassandra.go
```go
if classAndPairs[0] == "org.apache.cassandra.metrics" {
		tokens["class"] = "cassandra"
	} else if classAndPairs[0] == "java.lang" {
		tokens["class"] = "java"
	} else {
		return tokens
	}
```
Here another proposal for a sample of cassandra.conf file:
https://github.com/albundy83/telegraf/blob/cassandra-jolokia2-examples/plugins/inputs/jolokia2/examples/cassandra.conf

danielnelson:
@albundy83 I updated the example config to use your config, with the exception of keeping the snake case names `cassandra_file_cache`.  The reason for this is I'm nervous that we might conflict with existing metrics from the original cassandra input if we keep the style exactly the same, and also it matches style with the other example configs and plugins better.

I'm sure you can do a better job than I at selecting the right metrics, since you know much more about Cassandra and I'd love to keep this example up to date with whatever changes make in the future.

albundy83:
I understand and you are right to not put metrics with different format in the same place.
But you should preserve the original cassandra jmx name. 
I mean don't add extra underscores in the names of the metrics. 
You should even keep the case name. 
Adding an underscore between cassandra and metrics name should be enough. 
Same for Java metrics, just an underscore should be enough.
ex : cassandra_ClientRequestMetrics

It will be easier to generate configuration or even to look for details about metrics on internet.

albundy83:
I have written a new version with name_prefix = "java_" and name_prefix = "cassandra_"
It's a cleaner way I think:
https://github.com/albundy83/telegraf/commit/bcaaceda0e363317cdb1589de466f42535011379

danielnelson:
I added your modifications @albundy83, thanks!

albundy83:
Links to conf files (cassandra, kafka and java) are broken in the file:
https://github.com/influxdata/telegraf/blob/master/plugins/inputs/jolokia2/README.md

And here also, there is a broken link to cassandra.conf file:
https://github.com/influxdata/telegraf/blob/master/plugins/inputs/cassandra/README.md

danielnelson:
Good catch, I believe the paths are fixed now.

Issue #3840, opened 2018-02-27
inputs.sysstat took longer to collect


zhou-mfk:
Hi,
I encountered the following problem when using telegarf.

![image](https://user-images.githubusercontent.com/15190461/36714108-bfa2414a-1bcb-11e8-81ce-584a25d94e99.png)

sometimes 11 s ,
/usr/lib64/sa/sadc -S DISK 11 2

I think this command caused the timeout.

E! Error in plugin [inputs.sysstat]: took longer to collect than collection interval (10s)

### Relevant telegraf.conf:
[global_tags]
[agent]
  interval = "10s"
  round_interval = true
  metric_batch_size = 1000
  metric_buffer_limit = 10000
  collection_jitter = "3s"
  flush_interval = "10s"
  flush_jitter = "0s"
  precision = ""
  debug = true
  quiet = false
  logfile = ""
  hostname =""
  omit_hostname = false
[[inputs.cpu]]
  percpu = true
  totalcpu = true
  collect_cpu_time = false
[[inputs.disk]]
  ignore_fs = ["tmpfs", "devtmpfs", "devfs"]
[[inputs.diskio]]
[[inputs.kernel]]
[[inputs.mem]]
[[inputs.processes]]
[[inputs.swap]]
[[inputs.system]]
[[inputs.kernel_vmstat]]
[[inputs.linux_sysctl_fs]]
[[inputs.net]]
[[inputs.sysstat]]
  sadc_path = "/usr/lib64/sa/sadc"
  [inputs.sysstat.options]
    -d = "disk"

### System info:

Telegraf v1.3.4 (git: release-1.3 7bbd3daa98d5931c6e9025a98105092b0722de0d)

Red Hat Enterprise Linux Server release 6.5 (Santiago)

### Log info:
/var/log/telegraf/telegraf.log
2018-02-27T06:09:47Z E! Error in plugin [inputs.sysstat]: took longer to collect than collection interval (10s)
2018-02-27T06:10:00Z E! Error in plugin [inputs.sysstat]: took longer to collect than collection interval (10s)
2018-02-27T06:10:12Z E! Error in plugin [inputs.sysstat]: took longer to collect than collection interval (10s)
2018-02-27T06:10:25Z E! Error in plugin [inputs.sysstat]: took longer to collect than collection interval (10s)
2018-02-27T06:10:37Z E! Error in plugin [inputs.sysstat]: took longer to collect than collection interval (10s)
[...]

PR #4068, opened 2018-04-24
New parameter to force the interval of gather for sysstat

adrianlzt:
Added new config parameter for sysstat plugin, sadc_interval, to force
the gather interval of the sadc command.

This is useful when we have configured jitter in the agent, because
jitter breaks the measurement of the interval used in sadc.

Example:
 - agent with interval=10, jitter=5
 - sysstat is started in t=0 and runs sadc with interval=1
 - next execution of sysstat is in t=14 (interval + some jitter)
 - sysstat plugin set the value of "s.interval" to 14s
 - sadc is run with interval=13 (s.interval - parseInterval)
 - sysstat plugin is killed because is taking longer that collection
 interval

### Required for all PRs:

### DIFF
--- a/plugins/inputs/sysstat/sysstat.go
+++ b/plugins/inputs/sysstat/sysstat.go
@@ -33,6 +33,9 @@ type Sysstat struct {
 	// Sadc represents the path to the sadc collector utility.
 	Sadc string `toml:"sadc_path"`
 
+	// Force the execution time of sadc
+	SadcInterval internal.Duration `toml:"sadc_interval"`
+
 	// Sadf represents the path to the sadf cmd.
 	Sadf string `toml:"sadf_path"`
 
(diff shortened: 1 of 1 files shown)


danielnelson:
Thanks @adrianlzt, I've been wondering if we should always require a hard timeout, and completely remove the automatic computation of the interval size.  We would of course set a default timeout so it would still remain optional, but the timeout would not be dependent on the first collection.  Do you think this is a good idea?

danielnelson:
closes #3840

adrianlzt:
The computation of the interval size is a little bit hacky and I guess it will cause problems, because of the jitter or because sadc or sadf being slow.

I guess a fixed interval could be useful, but you are going to miss some data between runs.

Maybe allowing to use this plugin in two different forms could be better.
First one with a fixed interval (maybe with a default of 8s and warning if you lower the agent interval is going to fail).
And for the second one you will need to have sar running in your system and the plugin just issuing commands with ``sar -s HH:MM:SS -e HH:MM:SS`` storing the last ``end`` value.

danielnelson:
I confess I'm not up to speed on how sysstat works.  This is a step in the right direction but I will probably need more advice at some point in the future.

Issue #2468, opened 2017-02-23
win_perf_counter plugin does not work on 386


PierreF:
## Bug report

Telegraf i386 crash on Windows:
```
2017-02-23T15:41:04Z I! Starting Telegraf (version 1.2.1)
2017-02-23T15:41:04Z I! Loaded outputs: file
2017-02-23T15:41:04Z I! Loaded inputs: inputs.win_perf_counters
2017-02-23T15:41:04Z I! Tags enabled: host=MSEDGEWIN10
2017-02-23T15:41:04Z I! Agent Config: Interval:10s, Quiet:false, Hostname:"MSEDGEWIN10", Flush Interval:10s
unexpected fault address 0x5566b687
fatal error: fault
[signal 0xc0000005 code=0x0 addr=0x5566b687 pc=0x48a345]

goroutine 17 [running]:
runtime.throw(0xf13bde, 0x5)
        /usr/local/go/src/runtime/panic.go:566 +0x7f fp=0x12302e60 sp=0x12302e54
runtime.sigpanic()
        /usr/local/go/src/runtime/signal_windows.go:164 +0x116 fp=0x12302e78 sp=0x12302e60
syscall.UTF16ToString(0x5566b687, 0x20000000, 0x20000000, 0x0, 0x0)
        /usr/local/go/src/syscall/syscall_windows.go:51 +0x35 fp=0x12302ea0 sp=0x12302e78
github.com/influxdata/telegraf/plugins/inputs/win_perf_counters.UTF16PtrToString(0x5566b687, 0x0, 0x0)
        /home/ubuntu/telegraf-build/src/github.com/influxdata/telegraf/plugins/inputs/win_perf_counters/pdh.go:418 +0x62 fp=0x12302ec4 sp=0x12302ea0
github.com/influxdata/telegraf/plugins/inputs/win_perf_counters.(*Win_PerfCounters).Gather(0x1265a8e0, 0x14046a0, 0x1265af60, 0x0, 0x0)
        /home/ubuntu/telegraf-build/src/github.com/influxdata/telegraf/plugins/inputs/win_perf_counters/win_perf_counters.go:277 +0x32d fp=0x12302f98 sp=0x12302ec4
github.com/influxdata/telegraf/agent.gatherWithTimeout.func1(0x122a8a00, 0x1265aaa0, 0x1265af60)
[...]

Issue #1899, opened 2016-10-14
win_perf_counters and remote Windows host


skonet:
Hello,

In fact this is maybe a feature request but documentation does not seem to detail this point.

Telegraf win_perf_counters input plugin does not seem to allow to collect Windows counters from a remote machine (ie limited to local machine). This functionality is however present in Windows standard counters features (eg in Perfmon using metrics like \hostname\category\object).

I tried to force in conf file a remote hostname through several tricks, with no success.
Is this feature included? How can I activate it?

Configuration : Telegraf 1.0.1 on Win7 Pro SP1 FR.

Best regards.

maklaut:
:+1:

sjwang90:
For those interested in this feature, please test out the artifacts on PR https://github.com/influxdata/telegraf/pull/6590#issuecomment-1099684738. Please provide any comments or issues on PR #6590. Your testing and feedback will assist in getting this feature out sooner.

sspaink:
Closing due to lack of activity, if you are still interested in this feature request please re-open or leave a comment. We can look into reviving https://github.com/influxdata/telegraf/pull/11090.

PR #4076, opened 2018-04-25
386fix

srclosson:
### Required for all PRs:

- [*] Signed [CLA](https://influxdata.com/community/cla/).
- [*] Associated README.md updated.
- [*] Has appropriate unit tests.

### DIFF
--- a/plugins/inputs/win_perf_counters/README.md
+++ b/plugins/inputs/win_perf_counters/README.md
@@ -38,6 +38,12 @@ It is recommended NOT to use this on OSes starting with Vista and newer because 
 Example for Windows Server 2003, this would be set to true:
 `PreVistaSupport=true`
 
+#### UseWinTimestamps
+
+Bool, if set to `true` will request a timestamp along with the PerfCounter data. 
+
+Note: This call may not be supported on all versions of windows. If you receive an error, set this value to false to disable the feature, and searching for this call entirely.
+
 ### Object
 
 See Entry below.
--- a/plugins/inputs/win_perf_counters/README.md
+++ b/plugins/inputs/win_perf_counters/README.md
@@ -144,7 +144,7 @@ if any of the combinations of ObjectName/Instances/Counters are invalid.
 
   [[inputs.win_perf_counters.object]]
     # Processor usage, alternative to native, reports on a per core.
-    NodeName = "localhost"
+    Computer = "localhost"
     ObjectName = "Processor"
     Instances = ["*"]
     Counters = ["% Idle Time", "% Interrupt Time", "% Privileged Time", "% User Time", "% Processor Time"]
@@ -153,7 +153,7 @@ if any of the combinations of ObjectName/Instances/Counters are invalid.
 
   [[inputs.win_perf_counters.object]]
     # Disk times and queues
-    NodeName = "localhost"
+    Computer = "localhost"
     ObjectName = "LogicalDisk"
     Instances = ["*"]
     Counters = ["% Idle Time", "% Disk Time","% Disk Read Time", "% Disk Write Time", "% User Time", "Current Disk Queue Length"]
@@ -161,7 +161,7 @@ if any of the combinations of ObjectName/Instances/Counters are invalid.
     #IncludeTotal=false #Set to true to include _Total instance when querying for all (*).
 
   [[inputs.win_perf_counters.object]]
-    NodeName = "localhost"
+    Computer = "localhost"
     ObjectName = "System"
     Counters = ["Context Switches/sec","System Calls/sec", "Processor Queue Length"]
     Instances = ["------"]
@@ -170,7 +170,7 @@ if any of the combinations of ObjectName/Instances/Counters are invalid.
 
   [[inputs.win_perf_counters.object]]
     # Example query where the Instance portion must be removed to get data back, such as from the Memory object.
-    NodeName = "localhost"
+    Computer = "localhost"
     ObjectName = "Memory"
     Counters = ["Available Bytes","Cache Faults/sec","Demand Zero Faults/sec","Page Faults/sec","Pages/sec","Transition Faults/sec","Pool Nonpaged Bytes","Pool Paged Bytes"]
     Instances = ["------"] # Use 6 x - to remove the Instance bit from the query.
@@ -180,7 +180,7 @@ if any of the combinations of ObjectName/Instances/Counters are invalid.
   [[inputs.win_perf_counters.object]]
     # more counters for the Network Interface Object can be found at
     # https://msdn.microsoft.com/en-us/library/ms803962.aspx
-    NodeName = "localhost"
+    Computer = "localhost"
     ObjectName = "Network Interface"
     Counters = ["Bytes Received/sec","Bytes Sent/sec","Packets Received/sec","Packets Sent/sec"]
     Instances = ["*"] # Use 6 x - to remove the Instance bit from the query.
@@ -191,7 +191,7 @@ if any of the combinations of ObjectName/Instances/Counters are invalid.
 ### Active Directory Domain Controller
 ```
   [[inputs.win_perf_counters.object]]
-    NodeName = "localhost"
+    Computer = "localhost"
     ObjectName = "DirectoryServices"
     Instances = ["*"]
     Counters = ["Base Searches/sec","Database adds/sec","Database deletes/sec","Database modifys/sec","Database recycles/sec","LDAP Client Sessions","LDAP Searches/sec","LDAP Writes/sec"]
@@ -200,7 +200,7 @@ if any of the combinations of ObjectName/Instances/Counters are invalid.
     #IncludeTotal=false #Set to true to include _Total instance when querying for all (*).
 
   [[inputs.win_perf_counters.object]]
-    NodeName = "localhost"
+    Computer = "localhost"
     ObjectName = "Security System-Wide Statistics"
     Instances = ["*"]
     Counters = ["NTLM Authentications","Kerberos Authentications","Digest Authentications"]
@@ -208,7 +208,7 @@ if any of the combinations of ObjectName/Instances/Counters are invalid.
     #IncludeTotal=false #Set to true to include _Total instance when querying for all (*).
 
   [[inputs.win_perf_counters.object]]
-    NodeName = "localhost"
+    Computer = "localhost"
     ObjectName = "Database"
     Instances = ["*"]
     Counters = ["Database Cache % Hit","Database Cache Page Fault Stalls/sec","Database Cache Page Faults/sec","Database Cache Size"]
@@ -220,7 +220,7 @@ if any of the combinations of ObjectName/Instances/Counters are invalid.
 ```
   [[inputs.win_perf_counters.object]]
     # AD, DFS N, Useful if the server hosts a DFS Namespace or is a Domain Controller
-    NodeName = "localhost"
+    Computer = "localhost"
     ObjectName = "DFS Namespace Service Referrals"
     Instances = ["*"]
     Counters = ["Requests Processed","Requests Failed","Avg. Response Time"]
@@ -234,7 +234,7 @@ if any of the combinations of ObjectName/Instances/Counters are invalid.
 ```
   [[inputs.win_perf_counters.object]]
     # AD, DFS R, Useful if the server hosts a DFS Replication folder or is a Domain Controller
-    NodeName = "localhost"
+    Computer = "localhost"
     ObjectName = "DFS Replication Service Volumes"
     Instances = ["*"]
     Counters = ["Data Lookups","Database Commits"]
@@ -247,7 +247,7 @@ if any of the combinations of ObjectName/Instances/Counters are invalid.
 ### DNS Server + Domain Controllers
 ```
   [[inputs.win_perf_counters.object]]
-    NodeName = "localhost"
+    Computer = "localhost"
     ObjectName = "DNS"
     Counters = ["Dynamic Update Received","Dynamic Update Rejected","Recursive Queries","Recursive Queries Failure","Secure Update Failure","Secure Update Received","TCP Query Received","TCP Response Sent","UDP Query Received","UDP Response Sent","Total Query Received","Total Response Sent"]
     Instances = ["------"]
@@ -259,7 +259,7 @@ if any of the combinations of ObjectName/Instances/Counters are invalid.
 ```
   [[inputs.win_perf_counters.object]]
     # HTTP Service request queues in the Kernel before being handed over to User Mode.
-    NodeName = "localhost"
+    Computer = "localhost"
     ObjectName = "HTTP Service Request Queues"
     Instances = ["*"]
     Counters = ["CurrentQueueSize","RejectedRequests"]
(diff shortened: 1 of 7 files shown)


plugins/inputs/win_perf_counters/pdh_386.go
```diff
@@ -0,0 +1,74 @@
...
+// Kevin Pors <krpors@gmail.com>
+
+// +build windows
+
+package win_perf_counters
+
+// Union specialization for double values
+type PDH_FMT_COUNTERVALUE_DOUBLE struct {
+	CStatus     uint32
+	padding     uint32 // TODO: could well be broken on amd64
```
danielnelson: Change the type of padding to `[4]byte`.  Remove these comments, make sure there is a testcase to check each type.

plugins/inputs/win_perf_counters/pdh_386.go
```diff
@@ -0,0 +1,74 @@
...
+	padding    uint32 // TODO: could well be broken on amd64
+	LargeValue int64
+}
+
+// Union specialization for long values
+type PDH_FMT_COUNTERVALUE_LONG struct {
+	CStatus   uint32
+	LongValue int32
+	padding   [4]byte
+}
```
danielnelson: Shouldn't these all be the same size?  This one is smaller than the rest by 32bits.

plugins/inputs/win_perf_counters/README.md
```diff
@@ -38,6 +38,12 @@ It is recommended NOT to use this on OSes starting with Vista and newer because
 Example for Windows Server 2003, this would be set to true:
 `PreVistaSupport=true`
 
+#### UseWinTimestamps
```
danielnelson: Maybe a more descriptive name would be `UsePerfCounterTime`?

plugins/inputs/win_perf_counters/pdh.go
```diff
@@ -156,76 +167,89 @@ const (
...
 	pdh_ValidatePathW = libpdhDll.MustFindProc("PdhValidatePathW")
+
+	krn_FileTimeToSystemTime = libkrnDll.MustFindProc("FileTimeToSystemTime")
+	krn_FileTimeToLocalFileTime = libkrnDll.MustFindProc("FileTimeToLocalFileTime")
+	krn_LocalFileTimeToFileTime = libkrnDll.MustFindProc("LocalFileTimeToFileTime")
+	krn_WideCharToMultiByte = libkrnDll.MustFindProc("WideCharToMultiByte")
+}
+
+func PdhUseWinTimestamps() {
+	pdh_CollectQueryDataWithTime = libpdhDll.MustFindProc("PdhCollectQueryDataWithTime")
```
danielnelson: Let's do like above on `AddEnglishCounterW`?

plugins/inputs/win_perf_counters/pdh.go
```diff
@@ -156,76 +167,89 @@ const (
...
+}
+
+func PdhConnectMachine(szMachineName string) uint32 {
```
danielnelson: I can't seem to find where this is called from, is it used?

plugins/inputs/win_perf_counters/pdh.go
```diff
@@ -414,11 +469,64 @@ func PdhValidatePath(path string) uint32 {
...
+//   }
+//
+//	 fmt.Println("Length bytes: ", cc)
+//   n, err := WideCharToMultiByte(65001, 0, s, 1<<29, &outStr[0], 1<<29)
+//   if err != nil {
+// 	   fmt.Println("CONVERSION ERROR: ", err)
+//   }
+//   fmt.Println("Converted bytes: ", n)
+//
+func WideCharToMultiByte(codePage uint32, dwFlags uint32, wchar *uint16, nwchar int32, str *byte, nstr int32) (nwrite int32, err error) {
```
danielnelson: Where is this called from?

plugins/inputs/win_perf_counters/pdh.go
```diff
@@ -156,76 +167,89 @@ const (
...
-	FmtValue PDH_FMT_COUNTERVALUE_LONG
+type FILETIME struct {
+	dwLowDateTime  uint32
+	dwHighDateTime uint32
 }
 
 var (
 	// Library
 	libpdhDll *syscall.DLL
+	libkrnDll *syscall.DLL
```
danielnelson: Can you move all the kernel32 code to a new file: kernel32.go

plugins/inputs/win_perf_counters/README.md
srclosson: Yup. I like it. Changed

plugins/inputs/win_perf_counters/pdh.go
srclosson: Good idea. Done

plugins/inputs/win_perf_counters/pdh.go
srclosson: I learn something new everyday. Much cleaner. Thank-you

plugins/inputs/win_perf_counters/pdh.go
srclosson: No, and I neglected to  write a comment header stating this. In experimenting I had added the call, but ended up not requiring it (it was actually orders of magnitude slower doing the connect call and then querying for the counters. I've noted this in a comment header.

plugins/inputs/win_perf_counters/pdh.go
srclosson: It's not. It was one of those things I used to try and fix the win32 bug without padding the structures. The call itself works good and I thought it might be useful in the future. I've noted this in a comment, but I can remove it as well.

plugins/inputs/win_perf_counters/pdh_386.go
srclosson: Done

plugins/inputs/win_perf_counters/pdh_386.go
srclosson: I didn't actually change this one. It was already written this way, so  my question is yours. It just looks like I changed it because I moved it to a new file, but amd64 is the same. I don't understand why it was done this way either but there is no need to access data in the struct after the padding so I was inclined not to worry about it...

plugins/inputs/win_perf_counters/win_perf_counters.go
```diff
@@ -90,6 +97,7 @@ type item struct {
 	counter       string
 	instance      string
 	measurement   string
+	nodename      string
```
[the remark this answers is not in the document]
srclosson: Sure. Good idea.

plugins/inputs/win_perf_counters/win_perf_counters.go
```diff
@@ -135,20 +143,25 @@ func (m *Win_PerfCounters) SampleConfig() string {
...
 	if len(m.Object) > 0 {
 		for _, PerfObject := range m.Object {
 			for _, counter := range PerfObject.Counters {
 				for _, instance := range PerfObject.Instances {
 					objectname := PerfObject.ObjectName
+					nodename := PerfObject.NodeName
 
 					if instance == "------" {
-						query = "\\" + objectname + "\\" + counter
+						query = "\\\\" + nodename + "\\" + objectname + "\\" + counter
```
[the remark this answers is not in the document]
srclosson: I can tell you now. The node name is required to be specified. I personally like this, because the measurement gets tagged with the node name, which is now different from the collecting host, but it's not backwards compatible.

Let me check a different idea in which would be backwards compatible.

plugins/inputs/win_perf_counters/win_perf_counters.go
```diff
@@ -195,17 +208,24 @@ func (m *Win_PerfCounters) Gather(acc telegraf.Accumulator) error {
...
+			ret, timestamp = PdhCollectQueryDataWithTime(metric.handle)
+		} else {
+			timestamp = time.Now()
+			ret = PdhCollectQueryData(metric.handle)
+		}
+
 		if ret == ERROR_SUCCESS {
-			ret = PdhGetFormattedCounterArrayDouble(metric.counterHandle, &bufSize,
-				&bufCount, &emptyBuf[0]) // uses null ptr here according to MSDN.
+			ret = PdhGetFormattedCounterArrayDouble(metric.counterHandle, &bufSize, &bufCount, &emptyBuf[0]) // uses null ptr here according to MSDN.
```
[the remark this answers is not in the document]
srclosson: Roger doger

plugins/inputs/win_perf_counters/pdh.go
danielnelson: Can you instead remove all the calls that are not used?

plugins/inputs/win_perf_counters/README.md
```diff
@@ -50,6 +56,16 @@ beneath the main win_perf_counters entry, `[[inputs.win_perf_counters]]`.
 
 Following this is 3 required key/value pairs and the three optional parameters and their usage.
 
+#### Computer
+**Required**
```
danielnelson: Mark this as optional

srclosson:
Can this be re-reviewed?

danielnelson:
Can you take another pass and remove **all** code that is no actively being used right now.  If we need to bring back in this code we can grab it either from your tree or from https://github.com/lxn/win (who authored most of the Windows code).

Also, it looks like https://github.com/influxdata/telegraf/pull/4036 conflicted with the changes here so we will need to sort that out.

danielnelson:
closes #1899

danielnelson:
closes #2468

danielnelson:
@srclosson I believe the 32-bit path is fixed by #4189, but I believe you will need to base the remote computer and counter times fixes on the new code.  I'm going to close this pull request since I think it will be easier to open a new one.

vlastahajek:
@danielnelson, #4189 didn't contain fix for 386 issue. It doesn't work in 32bit build, I tested it. 
 I deliberately skipped 32bit support, to later merge it with this PR. 
As #4189 is merged now, I can:
-  either help @srclosson  with merging of current master to  386fix branch
- or create separated fix for 32bit support  in different PR. 
Maybe it would be more convenient to keep 386 support separated from other features (collecting from remote computers and adding timestamp to metrics)

danielnelson:
I'm not getting the crash anymore when I compile with GOARCH=386 and run on a 64-bit system, without #4189 it crashes immediately.  That said I haven't tested anything more than running `--test` on the default Windows config.

@srclosson Depends on how much time you have available, I'm hoping to release 1.7.0-rc1 at the end of next week, and it would be great if we could have the 32-bit fix in.  If your schedule is full until then it might be best for @vlastahajek to adapt your work so we can get it in for 1.7.

Ideally we would have separate pull requests for 32-bit fix and the new features.

vlastahajek:
@danielnelson: Yes, it's not crashing, at least for several tens minutes, but it writes invalid values for all the stuff, measurement names, fieldnames, fieldvalues.

As discussed in the original issue, there is problem with aligning data within structures with different field types between Go and Wind 32 API.

Crash before occurred fast because there were, IMHO, wrong allocation of the buffer for PDH_FMT_COUNTERVALUE_ITEM_DOUBLE array.

I've prepared separated fix for 386problem,  cause I primarily needed to fix my code for 32bit support.
It took me some time to figure out where to place padding in PDH_COUNTER_INFO structure.
It's now ready and tested and all I need to do is create a PR.

@srclosson: Now it depends on you. Did you have a time to look at changes? I did quite a large refactor.

vlastahajek:
Ok, #4206

vlastahajek:
I merged @srclosson's branch with master. The question is how to push it. 

@srclosson, As those features are your work, I would prefer if you could add me as a contributor to your forked telegraf repo and will commit it there. We can then continue with this PR (rename it) and with comments.

My merges is purely melding my refactor and fixes with @srclosson's enhacements, except one change, that I removed the `Computer` param from example config, cause default value _localhost_ doesn't work on Windows 10. You have to specify local computer name to have it working.

I would also propose a few improvements:
- Rename the `Computer` parameter to `RemoteHost`, or so. 
- Allow to monitor multiples hosts - if we change monitoring to remote host, we loose ability to monitor host where telegraf  is running easily. And if we have to monitor more such _nodes_? Mutiple instances of telegraf doesn't seem to be nice solution.

Both should be easy to implement. Multiple remote host can be easily added as stacking performance counters for multiple hosts,  although solution using more workers would be prettier.

danielnelson:
@vlastahajek In the meantime if you have time to prepare a branch with your changes that would be nice.

I think the RemoteHost change is a good one, but for multiple hosts doesn't it work with a mix of local and remote hosts?
```
  RemoteHost = "Foo"
  ObjectName = "Processor"
  Instances = ["*"]
  Counters = ["% Idle Time"]
  Measurement = "win_cpu"

  RemoteHost = "Bar"
  ObjectName = "Processor"
  Instances = ["*"]
  Counters = ["% Idle Time"]
  Measurement = "win_cpu"

  ObjectName = "LogicalDisk"
  Instances = ["*"]
  Counters = ["% Idle Time"]
  Measurement = "win_disk"
```

DmitryMaletin:
Hi, 
I play with forked version of telegraf 

My config looks like 

  Servers = ["$COMPUTERNAME", "RemoteHost","RemoteHost"]
  ObjectName = "Processor"
  Instances = ["*"]
  Counters = ["% Idle Time"]
  Measurement = "win_cpu"

So it's work with mixed set of hosts

vlastahajek:
I already have it in branch in my fork: https://github.com/
bonitoo-io/telegraf/tree/386fix. It;s merged imported branch.

Yes, this should work as a workaround.

Dekuji,

S pozdravem,
Vlastimil Hajek

DmitryMaletin:
@vlastahajek 
If I understood correctly you propose to have a counter per host config. 
In my case it's not very convenient cuz I can  have a dozen of host to get exactly the same counters.   

My current fork: https://github.com/DmitryMaletin/telegraf/blob/dma/plugins/inputs/win_perf_counters/win_perf_counters.go

vlastahajek:
Host per counter object is as it is now. 

I would vote  the same as you, having global remote host(s) for all counters

danielnelson:
Would plugin level be reasonable?  Remember you can always have multiple plugins defined.  Obviously not as fine grained as per object but in many cases this will be easier since you don't have to continually redefine the same remote hosts:
```
[[inputs.win_perf_counters]]
  # If empty, collect only from localhost
  Servers = ["$COMPUTERNAME", "RemoteHost","RemoteHost"]

    ObjectName = "Processor"
    Instances = ["*"]
    Counters = ["% Idle Time"]
    Measurement = "win_cpu"
```

DmitryMaletin:
yes. for me it sounds reasonable.

vlastahajek:
So, how to proceed further? 

@srclosson, Are going to look at it?

srclosson:
@danielnelson/@vlastahajek, I can certianly look at it. @danielnelson, could we get on the phone for 30 minutes? I've got a lot of stuff to contribute, and I think it would make sense to bring you up to speed on how to best go about merging some of it. There is a lot... Everything from a GPS Nmea plugin, to generic alarm parsing, windows event logs, webdav/sftp data transfer etc... I just don't want to hold up the line and would like to just show you through the ones I've developed, get your feedback, and some ideas on how I could/should go about merging them...

srclosson:
Another idea I had for the remote nodes is to do something like, (pseudocode)

for _, entry := range PerfCounterItems {
   wg.Add(1)
   go GetAndProcessData(entry)
}
wg.Wait()

This way, all of the requests get sent off at the same time. Probably more performant than doing each one individually...

danielnelson:
Yes, this would be a good thing, ideally network operations are done concurrently with a connection limit and with timeouts.  Support for cancellation using a Context is also a nice touch but Telegraf doesn't have a way to do this yet.

PR #4077, opened 2018-04-25
Add input plugin for McRouter

cthayer:
McRouter is a smart proxy for Memcached

https://github.com/facebook/mcrouter

Code is based on the memcached input plugin

### Required for all PRs:

### DIFF
--- a/plugins/inputs/mcrouter/README.md
+++ b/plugins/inputs/mcrouter/README.md
@@ -1,0 +1,103 @@
+# Mcrouter Input Plugin
+
+This plugin gathers statistics data from a Mcrouter server.
+
+### Configuration:
+
+```toml
+# Read metrics from one or many mcrouter servers.
+[[inputs.mcrouter]]
+  ## An array of address to gather stats about. Specify an ip or hostname
+  ## with port. ie tcp://localhost:11211, tcp://10.0.0.1:11211, etc.
+  servers = ["tcp://localhost:11211", "unix:///var/run/mcrouter.sock"]
+
+  ## Timeout for metric collections from all servers.  Minimum timeout is "1s".
+  # timeout = "5s"
+```
+
+### Measurements & Fields:
+
+The fields from this plugin are gathered in the *mcrouter* measurement.
+
+Description of gathered fields can be found [here](https://github.com/facebook/mcrouter/wiki/Stats-list).
+
+Fields:
+
+* uptime
+* num_servers
+* num_servers_new
+* num_servers_up
+* num_servers_down
+* num_servers_closed
+* num_clients
+* num_suspect_servers
+* destination_batches_sum
+* destination_requests_sum
+* outstanding_route_get_reqs_queued
+* outstanding_route_update_reqs_queued
+* outstanding_route_get_avg_queue_size
+* outstanding_route_update_avg_queue_size
+* outstanding_route_get_avg_wait_time_sec
+* outstanding_route_update_avg_wait_time_sec
+* retrans_closed_connections
+* destination_pending_reqs
+* destination_inflight_reqs
+* destination_batch_size
+* asynclog_requests
+* proxy_reqs_processing
+* proxy_reqs_waiting
+* client_queue_notify_period
+* rusage_system
+* rusage_user
+* ps_num_minor_faults
+* ps_num_major_faults
+* ps_user_time_sec
+* ps_system_time_sec
+* ps_vsize
+* ps_rss
+* fibers_allocated
+* fibers_pool_size
+* fibers_stack_high_watermark
+* successful_client_connections
+* duration_us
+* destination_max_pending_reqs
+* destination_max_inflight_reqs
+* retrans_per_kbyte_max
+* cmd_get_count
+* cmd_delete_out
+* cmd_lease_get
+* cmd_set
+* cmd_get_out_all
+* cmd_get_out
+* cmd_lease_set_count
+* cmd_other_out_all
+* cmd_lease_get_out
+* cmd_set_count
+* cmd_lease_set_out
+* cmd_delete_count
+* cmd_other
+* cmd_delete
+* cmd_get
+* cmd_lease_set
+* cmd_set_out
+* cmd_lease_get_count
+* cmd_other_out
+* cmd_lease_get_out_all
+* cmd_set_out_all
+* cmd_other_count
+* cmd_delete_out_all
+* cmd_lease_set_out_all
+
+### Tags:
+
+* Mcrouter measurements have the following tags:
+    - server (the host name from which metrics are gathered)
+
+
+
+### Example Output:
+
+```
+$ ./telegraf --config telegraf.conf --input-filter mcrouter --test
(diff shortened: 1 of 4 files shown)


plugins/inputs/mcrouter/README.md
```diff
@@ -0,0 +1,102 @@
...
+### Configuration:
+
+```toml
+# Read metrics from one or many mcrouter servers.
+[[inputs.mcrouter]]
+  # An array of address to gather stats about. Specify an ip on hostname
+  # with optional port. ie localhost, 10.0.0.1:11211, etc.
+  servers = ["localhost:11211"]
+  # An array of unix mcrouter sockets to gather stats about.
+  # unix_sockets = ["/var/run/mcrouter.sock"]
```
danielnelson: Let's use a single list of this style:
```
servers = ["tcp://localhost:11211", "unix:///var/run/mcrouter.sock", "udp://localhost:11211"]
```

We don't need to add support for UDP at first, assuming it is even available with mcrouter, but this will give us a nice way to add it later.

plugins/inputs/mcrouter/mcrouter.go
```diff
@@ -0,0 +1,224 @@
...
+  ## An array of address to gather stats about. Specify an ip on hostname
+  ## with optional port. ie localhost, 10.0.0.1:11211, etc.
+  servers = ["localhost:11211"]
+  # unix_sockets = ["/var/run/mcrouter.sock"]
+`
+
+var defaultTimeout = 5 * time.Second
+
+// The list of metrics that should be sent
+var sendMetrics = []string{
```
danielnelson: I would love to have these locked to specific types, since it would be a problem if a field changed values.  Maybe we can turn this into a `map[string]statType` where statType is an enum being int64, float64 or string?  Then when we parse we try the expected value and log an error if it doesn't match.

plugins/inputs/mcrouter/mcrouter.go
```diff
@@ -0,0 +1,224 @@
...
+}
+
+var sampleConfig = `
+  ## An array of address to gather stats about. Specify an ip on hostname
+  ## with optional port. ie localhost, 10.0.0.1:11211, etc.
+  servers = ["localhost:11211"]
+  # unix_sockets = ["/var/run/mcrouter.sock"]
+`
+
+var defaultTimeout = 5 * time.Second
```
danielnelson: Let's make this configurable and have it apply to all servers.

plugins/inputs/mcrouter/mcrouter.go
```diff
@@ -0,0 +1,224 @@
...
+			return err
+		}
+		defer conn.Close()
+	} else {
+		_, _, err = net.SplitHostPort(address)
+		if err != nil {
+			address = address + ":11211"
+		}
+
+		conn, err = net.DialTimeout("tcp", address, defaultTimeout)
```
danielnelson: Would be a nice touch to create a context.Context at the beginning of Gather, and use DialContext, I recently updated the zookeeper input to do something like this:

https://github.com/influxdata/telegraf/blob/master/plugins/inputs/zookeeper/zookeeper.go#L81

plugins/inputs/mcrouter/mcrouter.go
```diff
@@ -0,0 +1,224 @@
...
+
+	if conn == nil {
+		return fmt.Errorf("Failed to create net connection")
+	}
+
+	// Extend connection
+	conn.SetDeadline(time.Now().Add(defaultTimeout))
+
+	// Read and write buffer
+	rw := bufio.NewReadWriter(bufio.NewReader(conn), bufio.NewWriter(conn))
```
danielnelson: I don't think this is necessary, just use `conn.Write(` and you won't need to flush.

plugins/inputs/mcrouter/mcrouter.go
```diff
@@ -0,0 +1,224 @@
...
+			} else {
+				fields[key] = value
+			}
+		}
+	}
+	acc.AddFields("mcrouter", fields, tags)
+	return nil
+}
+
+func parseResponse(r *bufio.Reader) (map[string]string, error) {
```
danielnelson: We should use a `bufio.Scanner` to read lines, the ReadLine() function can return partial lines.

plugins/inputs/mcrouter/mcrouter.go
```diff
@@ -0,0 +1,224 @@
...
+			return values, errRead
+		}
+		// Done
+		if bytes.Equal(line, []byte("END")) {
+			break
+		}
+		// Read values
+		s := bytes.SplitN(line, []byte(" "), 3)
+		if len(s) != 3 || !bytes.Equal(s[0], []byte("STAT")) {
+			return values, fmt.Errorf("unexpected line in stats response: %q", line)
```
danielnelson: Return nil as the map.

plugins/inputs/mcrouter/mcrouter.go
cthayer: switched to map with types and only attempt a convert for the expected type

plugins/inputs/mcrouter/README.md
cthayer: I've updated to use a single array of servers and pull the protocol from the connection string.

(mcrouter does not support UDP connections)

plugins/inputs/mcrouter/mcrouter.go
cthayer: I followed the zookeeper example here

plugins/inputs/mcrouter/mcrouter.go
cthayer: I think I followed the zookeeper example here

plugins/inputs/mcrouter/mcrouter.go
cthayer: I followed the zookeeper example and just do `fmt.Fprint(conn`

plugins/inputs/mcrouter/mcrouter.go
cthayer: Using `bufio.Scanner` now

plugins/inputs/mcrouter/mcrouter.go
cthayer: Made the change.  I assume that the standard is to return no stats on error rather than partial stats?

cthayer:
I think I could update the memcached plugin afterwards.

It seems like a breaking change for the memcached plugin to remove the `UnixSockets` member from the `struct`?

plugins/inputs/mcrouter/mcrouter.go
```diff
@@ -0,0 +1,255 @@
...
+	return nil
+}
+
+func (m *Mcrouter) gatherServer(ctx context.Context, address string, acc telegraf.Accumulator) error {
+	var conn net.Conn
+	var err error
+	var protocol string
+	var dialer net.Dialer
+
+	s := strings.SplitN(address, "://", 2)
```
danielnelson: Use `url.Parse` here, I think it will help simplify things and prevent errors.

plugins/inputs/mcrouter/mcrouter.go
danielnelson: Yes, returning an error is probably best here.

danielnelson:
Yeah, for memcached we will need to keep backwards compatibility.  I usually try to translate the old format to the new one on the first call to Gather.

plugins/inputs/mcrouter/mcrouter.go
cthayer: I updated the code to use `url.Parse`

I feel like it makes it a bit stricter on the server connection string format, but I think that's ok.

I make some attempt to tease partial connection strings into a proper format, but ultimately default to `tcp://localhost:11211` if the server connection string does not parse properly.

plugins/inputs/mcrouter/mcrouter.go
```diff
@@ -0,0 +1,273 @@
...
+	var dialer net.Dialer
+
+	u, parseError := url.Parse(address)
+
+	if parseError != nil {
+		return fmt.Errorf("Invalid server address")
+	}
+
+	if u.Scheme != "tcp" && u.Scheme != "unix" {
+		protocol = defaultServerURL.Scheme
```
danielnelson: return error

plugins/inputs/mcrouter/mcrouter.go
```diff
@@ -0,0 +1,273 @@
...
+
+	if protocol == "unix" {
+		if u.Path == "" {
+			return fmt.Errorf("Invalid unix socket path")
+		}
+
+		address = u.Path
+	} else {
+		if u.Host == "" {
+			u.Host = defaultServerURL.Host
```
danielnelson: return error

plugins/inputs/mcrouter/mcrouter.go
```diff
@@ -0,0 +1,273 @@
...
+		}
+
+		host, port, err = net.SplitHostPort(u.Host)
+
+		if err != nil {
```
danielnelson: I did a few quick tests and unfortunately I think we will need to handle the case when either host or port are empty strings:
```
func Check(input string) {
	host, port, err := net.SplitHostPort(input)
	fmt.Printf("%q %q %q: %v\n", input, host, port, err)
}

func TestSplitHostPort(t *testing.T) {
	Check("localhost:8086")
	Check("localhost") # we should accept this and use default port
	Check("localhost:") # use default port
	Check(":8086") # use default hostname
	Check(":") # default host:port or reject
	Check("") # reject, but we already do that above
}
```
```
"localhost:8086" "localhost" "8086": <nil>
"localhost" "" "": address localhost: missing port in address
"localhost:" "localhost" "": <nil>
":8086" "" "8086": <nil>
":" "" "": <nil>
"" "" "": missing port in address
```

plugins/inputs/mcrouter/mcrouter.go
cthayer: returning error

plugins/inputs/mcrouter/mcrouter.go
cthayer: returning error

plugins/inputs/mcrouter/mcrouter.go
cthayer: I added a test for parsing addresses and separated the address parsing into its own function.

I'm just using the `url.URL` functions to get the hostname and port parts and not using `net.SplitHostPort` any more.

danielnelson:
Great job, merged for 1.7.0

cthayer:
Thanks!

PR #4078, opened 2018-04-26
agent/accumulator: expose GetTime to give input plugins a chance to r…

zerodeux:
…euse a timestamp for several Add* calls + use this on plugins/system to fix telegraf-sample/influxdb-point mismatch pb

I stumbled upon curious single system measures spanning over two distinct timestamps, on a second boundary (this bug happening a few times in a day over 300 servers collecting system measures every 10s) : 
```
> SELECT * FROM system WHERE time > now() - 5m and host='xxxxxxxxx-xxxxxxx'
name: system
time                host              load1 load15 load5 n_cpus n_users uptime  uptime_format
----                ----              ----- ------ ----- ------ ------- ------  -------------
...
1524726770000000000 xxxxxxxxx-xxxxxxx 0.79  0.64   0.6   4      0       6685039 77 days,  8:57 
1524726780000000000 xxxxxxxxx-xxxxxxx 1.14  0.67   0.68  4      0     
1524726781000000000 xxxxxxxxx-xxxxxxx                                   6685049 77 days,  8:57 
1524726790000000000 xxxxxxxxx-xxxxxxx 1.41  0.69   0.76  4      0       6685059 77 days,  8:57 
...
```

I realized that the input/system plugins call three accumulator.Add* methods in a row, each fetching a distinct timestamp. This is cumbersome since its much more useful to gather all those fields as a single InfluxDB point and hence with the same timestamp.

In order to to this, I had to expose the accumulator.getTime() function. This should not break any plugin assumption and behaviour.

### DIFF
--- a/accumulator.go
+++ b/accumulator.go
@@ -41,7 +41,6 @@ type Accumulator interface {
 		t ...time.Time)
 
 	SetPrecision(precision, interval time.Duration)
-	GetTime(t []time.Time) time.Time
 
 	AddError(err error)
 }
--- a/accumulator.go
+++ b/accumulator.go
@@ -41,6 +41,7 @@ type Accumulator interface {
 		t ...time.Time)
 
 	SetPrecision(precision, interval time.Duration)
+	GetTime(t []time.Time) time.Time
 
 	AddError(err error)
 }
--- a/agent/accumulator.go
+++ b/agent/accumulator.go
@@ -49,7 +49,7 @@ func (ac *accumulator) AddFields(
 	tags map[string]string,
 	t ...time.Time,
 ) {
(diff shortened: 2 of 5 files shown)


danielnelson:
This is a good fix but I think we should just create one time and use it in both calls, no need to add the GetTime function.
```
now := time.Now()
acc.AddGauge(..., now)
acc.AddCounter(..., now)
```

zerodeux:
OK, and I missed the point that other plugins are alreay doing this. I thought that Telegraf was somewhat in control of the time notion (especially wrt. precision) and didn't dare to go to time.Now() directly. I'll push a simpler fix for plugins/inputs/system ASAP.

danielnelson:
Thanks!  I include this in the next patch release (1.6.2)

zerodeux:
Thanks for your quick review and integration !

PR #4080, opened 2018-04-27
[jolokia2] [cassandra] Metrics values should have same names than old cassandra plugin

albundy83:
### Required for all PRs:

The modifications are important as after working with the old version, it's better to add the prefix of the metrics name.
Here an example with the output of the deprecated cassandra plugin:

```
cassandraThreadPools,cassandra_host=localhost,dc=dc1,env=prod,host=my_hostname.com,mname=MaxPoolSize,path=transport,rack=rack1,scope=Native-Transport-Requests MaxPoolSize_Value=256 1524670394000000000
```

Here the same output with the current version of the cassandra.conf of jolokia2:
```
cassandra_ThreadPools,dc=dc1,env=prod,host=my_hostname.com,jolokia_agent_url=http://localhost:8778/jolokia,name=MaxPoolSize,path=transport,rack=rack1,scope=Native-Transport-Requests Value=256 1524670394000000000
```
As you can see the value name `MaxPoolSize_Value` has been renamed to `Value` only

With the version of this pull request, the output will be like this:
```
cassandra_ThreadPools,dc=dc1,env=prod,host=my_hostname.com,jolokia_agent_url=http://localhost:8778/jolokia,name=MaxPoolSize,path=transport,rack=rack1,scope=Native-Transport-Requests MaxPoolSize_Value=256 1524670394000000000
```
The value name `MaxPoolSize_Value` is now identical.

It will just remain the tag `mname` in deprecated cassandra which is now called `name`.

The "price" to migrate to jolokia2 (and migrate boards!!) from deprecated cassandra plugin will be much lower.

### DIFF
--- a/plugins/inputs/jolokia2/examples/cassandra.conf
+++ b/plugins/inputs/jolokia2/examples/cassandra.conf
@@ -10,6 +10,7 @@
     name  = "GarbageCollector"
     mbean = "java.lang:name=*,type=GarbageCollector"
     tag_keys = ["name"]
+    field_prefix = "$1_"
 
 [[inputs.jolokia2_agent]]
   urls = ["http://localhost:8778/jolokia"]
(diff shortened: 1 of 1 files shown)


albundy83:
I have also removed `HintedHandoffManager` as it's not an usable metric

albundy83:
This follow the request https://github.com/influxdata/telegraf/pull/4050

danielnelson:
Much nicer!

PR #4089, opened 2018-04-30
Add Azure Monitor output plugin

gunnaraasen:
### Required for all PRs:

This is a new output plugin for Azure Monitor. I will be adding more unit tests soon.


plugins/outputs/azuremonitor/azuremetadata.go
```diff
@@ -0,0 +1,213 @@
...
+
+	msiEndpoint.RawQuery = msiParameters.Encode()
+	req, err := http.NewRequest("GET", msiEndpoint.String(), nil)
+	if err != nil {
+		return nil, err
+	}
+	req.Header.Add("Metadata", "true")
+
+	// Create the HTTP client and call the token service
+	client := http.Client{
```
danielnelson: Use http.Client and transport shared between requests.

plugins/outputs/azuremonitor/azuremonitor.go
```diff
@@ -0,0 +1,490 @@
...
+	Region              string            `toml:"region"`
+	Timeout             internal.Duration `toml:"Timeout"`
+	AzureSubscriptionID string            `toml:"azureSubscription"`
+	AzureTenantID       string            `toml:"azureTenant"`
+	AzureClientID       string            `toml:"azureClientId"`
+	AzureClientSecret   string            `toml:"azureClientSecret"`
+
+	useMsi           bool
+	metadataService  *AzureInstanceMetadata
+	instanceMetadata *VirtualMachineMetadata
```
danielnelson: I think this is unused.

plugins/outputs/azuremonitor/azuremonitor.go
```diff
@@ -0,0 +1,490 @@
...
+	msiToken         *msiToken
+	msiResource      string
```
danielnelson: Make this a package const

plugins/outputs/azuremonitor/azuremonitor.go
```diff
@@ -0,0 +1,490 @@
...
+	bearerToken      string
```
danielnelson: Use directly from msiToken

plugins/outputs/azuremonitor/azuremonitor.go
```diff
@@ -0,0 +1,490 @@
...
+	expiryWatermark  time.Duration
```
danielnelson: I see this being used but where is it set?

plugins/outputs/azuremonitor/azuremonitor.go
```diff
@@ -0,0 +1,490 @@
...
+
+	oauthConfig *adal.OAuthConfig
+	adalToken   adal.OAuthTokenProvider
```
danielnelson: This I see being set but not used.

plugins/outputs/azuremonitor/azuremonitor.go
```diff
@@ -0,0 +1,490 @@
...
+
+	client *http.Client
+
+	cache       map[string]*azureMonitorMetric
+	period      time.Duration
+	delay       time.Duration
+	periodStart time.Time
+	periodEnd   time.Time
```
danielnelson: I think these two times can just be locals

plugins/outputs/azuremonitor/azuremonitor.go
```diff
@@ -0,0 +1,490 @@
...
+	}
+	return
+}
+
+func (a *AzureMonitor) postData(msg *[]byte) (*http.Request, error) {
+	if err := a.validateCredentials(); err != nil {
+		return nil, fmt.Errorf("Error authenticating: %v", err)
+	}
+
+	metricsEndpoint := fmt.Sprintf("https://%s.monitoring.azure.com%s/metrics",
```
asheniam: One thing to be careful here -- As custom metrics is in preview on the Azure Monitor side, we don't support all regions of public Azure.  We are only going to be available for a few regions as part of the reviews so not all these endpoints will exist.

plugins/outputs/azuremonitor/azuremonitor.go
```diff
@@ -0,0 +1,490 @@
...
+	if err != nil {
+		log.Printf("Error creating HTTP request")
+		return nil, err
+	}
+
+	req.Header.Set("Authorization", "Bearer "+a.bearerToken)
+	req.Header.Set("Content-Type", "application/x-ndjson")
+
+	tr := &http.Transport{
+		TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
```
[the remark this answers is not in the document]
asheniam: +1.  Metric endpoint supports TLS1.2

plugins/outputs/azuremonitor/azuremonitor.go
asheniam: For any of the endpoints that the output plugin is communicating with (ex. https://<region>.monitoring.azure.com, we should ideally move all these endpoints into single place in the code, not sprinkle them across all files. This will help future proof the plugin to more easily support other Azure clouds down the road (Azure Germany, Azure China, Azure US Government) which will all have their own endpoints.

gunnaraasen:
@danielnelson I've updated this branch to use the running output aggregator pattern you suggested. I moved the original PR code to the `ga-azure-monitor-original` branch. I will add some tests if you think the architecture makes sense now.

danielnelson:
Looks good, like how you dealt with not having the filter for old metrics.  Gives me some ideas for improving this with the normal aggregators.

plugins/outputs/azuremonitor/azuremonitor.go
```diff
@@ -0,0 +1,445 @@
...
+
+var sampleConfig = `
+  ## The resource ID against which metric will be logged.  If not
+  ## specified, the plugin will attempt to retrieve the resource ID
+  ## of the VM via the instance metadata service (optional if running 
+  ## on an Azure VM with MSI)
+  #resource_id = "/subscriptions/<subscription_id>/resourceGroups/<resource_group>/providers/Microsoft.Compute/virtualMachines/<vm_name>"
+  ## Azure region to publish metrics against.  Defaults to eastus.
+  ## Leave blank to automatically query the region via MSI.
+  #region = "useast"
```
asheniam: Use "eastus" as the default, not "useast".  "useast" will not work -- it won't resolve to any monitoring endpoint.

plugins/outputs/azuremonitor/azuremonitor.go
```diff
@@ -0,0 +1,445 @@
...
+	cache map[time.Time]map[uint64]*aggregate
+}
+
+type aggregate struct {
+	telegraf.Metric
+	updated bool
+}
+
+const (
+	defaultRegion          string = "eastus"
```
asheniam: We shouldn't have a default region constant.  This value either needs to come from instance metadata or come from user configuration.  The region must match the region of the Azure resource ID and can't be guessed.

plugins/outputs/azuremonitor/azuremonitor.go
```diff
@@ -0,0 +1,445 @@
...
+	if err := a.validateCredentials(); err != nil {
+		return fmt.Errorf("E! Unable to fetch authentication credentials: %v", err)
+	}
+
+	req, err := http.NewRequest("POST", a.url, bytes.NewBuffer(body))
+	if err != nil {
+		return err
+	}
+
+	req.Header.Set("Authorization", "Bearer "+a.msiToken.AccessToken)
```
asheniam: If we are not using MSI, where are we setting a.adalToken?

plugins/outputs/azuremonitor/azuremonitor.go
```diff
@@ -0,0 +1,445 @@
...
+	req.Header.Set("Content-Type", "application/x-ndjson")
+
+	resp, err := a.client.Do(req)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode >= 300 || resp.StatusCode < 200 {
```
asheniam: Why are we only doing fmt.Errorf for status codes in the [300, 200) range.  We should also follow this pattern for any errors encountered in the 4xx or 5xx range

plugins/outputs/azuremonitor/azuremonitor.go
```diff
@@ -0,0 +1,445 @@
...
+	min, _ := m.GetField("min")
+	max, _ := m.GetField("max")
+	sum, _ := m.GetField("sum")
+	count, _ := m.GetField("count")
+	return &azureMonitorMetric{
+		Time: m.Time(),
+		Data: &azureMonitorData{
+			BaseData: &azureMonitorBaseData{
+				Metric:         m.Name(),
+				Namespace:      "default",
```
asheniam: Any reason we choose to set the metric namespace to be "default"?  This might be good for users to override via config but as a default, it might be better to go with a value which has "Telegraf" in the namespace.

plugins/outputs/azuremonitor/azuremonitor.go
```diff
@@ -0,0 +1,445 @@
...
+	tbucket := time.Date(t.Year(), t.Month(), t.Day(), t.Hour(), t.Minute(), 0, 0, t.Location())
+	if tbucket.Before(time.Now().Add(-time.Minute * 30)) {
+		log.Printf("W! attempted to aggregate metric over 30 minutes old: %v, %v", t, tbucket)
+		return
+	}
+
+	// Azure Monitor doesn't have a string value type, so convert string
+	// fields to dimensions (a.k.a. tags) if enabled.
+	if a.StringAsDimension {
+		for fk, fv := range m.Fields() {
```
asheniam: How many fields is typical?  There can only be max of 10 tags submitted

plugins/outputs/azuremonitor/azuremonitor.go
```diff
@@ -0,0 +1,445 @@
...
+				translate(m).Data.BaseData.Series...,
+			)
+		}
+	}
+
+	var body []byte
+	for _, m := range azmetrics {
+		// Azure Monitor accepts new batches of points in new-line delimited
+		// JSON, following RFC 4288 (see https://github.com/ndjson/ndjson-spec).
+		jsonBytes, err := json.Marshal(&m)
```
asheniam: How large can azmetrics be?  I believe the Azure Monitor metric API has a max request body size of 4MB.  If we exceed this limit, we should issue multiple POST requests

plugins/outputs/azuremonitor/azuremonitor.go
```diff
@@ -0,0 +1,445 @@
...
+  ## Azure region to publish metrics against.  Defaults to eastus.
+  ## Leave blank to automatically query the region via MSI.
+  #region = "useast"
+
+  ## Write HTTP timeout, formatted as a string.  If not provided, will default
+  ## to 5s. 0s means no timeout (not recommended).
+  # timeout = "5s"
+
+  ## Whether or not to use managed service identity.
+  #use_managed_service_identity = true
```
asheniam: Nit: In the sample configuration, we should make it clear that when use_managed_service_identify is false, it's required that the user supply resource_id, region, azure_subscription, azure_tenant, azure_client_id, and azure_client_secret.  These become mandatory parameters.

plugins/outputs/azuremonitor/azuremonitor.go
```diff
@@ -0,0 +1,445 @@
...
+
+var _ telegraf.AggregatingOutput = (*AzureMonitor)(nil)
+var _ telegraf.Output = (*AzureMonitor)(nil)
+
+// AzureMonitor allows publishing of metrics to the Azure Monitor custom metrics service
+type AzureMonitor struct {
+	useMsi              bool              `toml:"use_managed_service_identity"`
+	ResourceID          string            `toml:"resource_id"`
+	Region              string            `toml:"region"`
+	Timeout             internal.Duration `toml:"Timeout"`
```
asheniam: Nit: Should this be lower case "timeout" instead of "Timeout"?

plugins/outputs/azuremonitor/azuremonitor.go
```diff
@@ -0,0 +1,445 @@
...
+	AzureSubscriptionID string            `toml:"azure_subscription"`
+	AzureTenantID       string            `toml:"azure_tenant"`
+	AzureClientID       string            `toml:"azure_client_id"`
+	AzureClientSecret   string            `toml:"azure_client_secret"`
+	StringAsDimension   bool              `toml:"string_as_dimension"`
```
asheniam: What is StringAsDimension?  There isn't such example in the sample config?

plugins/outputs/azuremonitor/azuremonitor.go
```diff
@@ -0,0 +1,445 @@
...
+		}
+	}
+}
+
+func init() {
+	outputs.Add("azuremonitor", func() telegraf.Output {
+		return &AzureMonitor{
+			StringAsDimension: false,
+			Timeout:           internal.Duration{Duration: time.Second * 5},
+			Region:            defaultRegion,
```
asheniam: As mentioned earlier, we shouldn't treat region special with a default region.  This should be treated as the rest of the configuration -- either it comes from instance metadata or from user supplied config.

asheniam:
@danielnelson - I saw the milestone changed from 1.7 to 1.8.  What is the timeline for 1.8?

danielnelson:
I believe 1.8 will be finished around the end of August?  Does that sounds right @russorat?

plugins/outputs/azuremonitor/azuremonitor.go
```diff
@@ -0,0 +1,445 @@
...
+// Reset clears the cache of aggregate metrics
+func (a *AzureMonitor) Reset() {
+	for tbucket := range a.cache {
+		// Remove aggregates older than 30 minutes
+		if tbucket.Before(time.Now().Add(-time.Minute * 30)) {
+			delete(a.cache, tbucket)
+			continue
+		}
+		for id := range a.cache[tbucket] {
+			a.cache[tbucket][id].updated = false
```
danielnelson: If the interval is lower than the 1m aggregation period, this can get set to false before the metric has had a chance to be returned in `Push()` since you may have multiple calls to `Reset()` before the metric is returned.  This causes the plugin not to write any metrics unless you set the set the flush_interval to at least `1m`.

plugins/outputs/azuremonitor/azuremonitor.go
```diff
@@ -0,0 +1,445 @@
...
+		oauthConfig, err := adal.NewOAuthConfig(azure.PublicCloud.ActiveDirectoryEndpoint, a.AzureTenantID)
+		if err != nil {
+			return fmt.Errorf("E! Could not initialize AD client: %s", err)
+		}
+		a.oauthConfig = oauthConfig
+	}
+
+	// Pull region and resource identifier
+	err := a.GetInstanceMetadata()
+	if err != nil && a.ResourceID == "" && a.Region == "" {
```
danielnelson: I think this should be `||` instead of `&&`, we should always return `if err != nil`, so that we are sure to see all errors and I think we need both of these set.

Also, if Region and ResourceID are set beforehand, maybe we can skip this function completely?

plugins/outputs/azuremonitor/azuremetadata.go
```diff
@@ -0,0 +1,186 @@
...
+	if err := json.Unmarshal(reply, &token); err != nil {
+		return nil, err
+	}
+	token.parseTimes()
+	token.raw = string(reply)
+	return &token, nil
+}
+
+// GetInstanceMetadata retrieves metadata about the current Azure VM
+func (a *AzureMonitor) GetInstanceMetadata() error {
```
danielnelson: I think ideally this function would return `region, resource, error`.  The calling function would combine these with the plugin settings and create the url.  Also, consider passing in the client and making this and the function above a free function.

plugins/outputs/azuremonitor/azuremonitor.go
```diff
@@ -0,0 +1,445 @@
...
+		return fmt.Errorf("E! No resource id specified, and Azure Instance metadata service not available.  If not running on an Azure VM, provide a value for resource_id")
+	}
+
```
danielnelson: Once we have the final discovered URL, I think we should add a log message with the final version at debug level.

plugins/outputs/azure_monitor/azure_monitor_test.go
```diff
@@ -0,0 +1,275 @@
+package azure_monitor
```
danielnelson: Let us know when the tests are ready.

plugins/outputs/azure_monitor/azure_monitor.go
```diff
@@ -0,0 +1,491 @@
...
+				nf["min"] = aggfields["min"]
+			}
+			if fv < aggfields["max"].(float64) {
+				nf["max"] = aggfields["max"]
+			}
+			nf["sum"] = fv + aggfields["sum"].(float64)
+			nf["count"] = aggfields["count"].(int64) + 1
+		}
+
+		na, _ := metric.New(name, m.Tags(), nf, tbucket)
```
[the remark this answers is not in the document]
danielnelson: https://github.com/influxdata/telegraf/pull/4441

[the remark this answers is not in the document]
gunnaraasen: `telegraf.Metric` was required to separate the aggregation step and the write step to utilize Telegraf's internal buffering, via https://github.com/influxdata/telegraf/commit/19e435fc618b25a66b0cb935c37199e452aca1a8. I don't see a way to use a new type and still use the internal buffering. Let me know if you have any ideas. This could also be left as an optimization for a later time.

plugins/outputs/azure_monitor/azure_monitor.go
```diff
@@ -0,0 +1,491 @@
...
+}
+
+// Add will append a metric to the output aggregate
+func (a *AzureMonitor) Add(m telegraf.Metric) {
+	// Azure Monitor only supports aggregates 30 minutes into the past and 4
+	// minutes into the future. Future metrics are dropped when pushed.
+	t := m.Time()
+	tbucket := time.Date(t.Year(), t.Month(), t.Day(), t.Hour(), t.Minute(), 0, 0, t.Location())
+	if tbucket.Before(time.Now().Add(-time.Minute * 30)) {
+		log.Printf("W! attempted to aggregate metric over 30 minutes old: %v, %v", t, tbucket)
```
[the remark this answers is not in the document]
gunnaraasen: I switched this to a TODO comment.

plugins/outputs/azure_monitor/azure_monitor.go
```diff
@@ -0,0 +1,491 @@
...
+		if ok {
+			aggfields := agg.Fields()
+			if fv > aggfields["min"].(float64) {
+				nf["min"] = aggfields["min"]
+			}
+			if fv < aggfields["max"].(float64) {
+				nf["max"] = aggfields["max"]
+			}
+			nf["sum"] = fv + aggfields["sum"].(float64)
+			nf["count"] = aggfields["count"].(int64) + 1
```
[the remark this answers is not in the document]
gunnaraasen: I haven't had a chance to look into this or test on i386 aside. It does not panic on x86_64 builds but I haven't had a chance to test on i386.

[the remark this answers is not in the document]
danielnelson: You won't need to run this on a 32-bit system to test it, just compile with GOARCH=386 and try running it.  Looking at it again I guess it will work because we always convert int to int64 when adding them to a metric, but I do feel like this is brittle and a slight change to metric will introduce a panic here.

plugins/outputs/azure_monitor/azure_monitor.go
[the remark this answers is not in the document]
danielnelson: The metric would come in via Add as a telegraf.Metric, you would convert it to a custom type for the cache, and then it would be turned back into a telegraf.Metric in the Push step.  It seems like a lot of conversions but you are actually already making a new cache Metric for each telegraf.Metric added.  However, you don't really want or need the loose field typing of a telegraf.Metric or any of the other functionality.  This would allow you to remove all the conversions above.

plugins/outputs/azure_monitor/azure_monitor_test.go
```diff
@@ -0,0 +1,239 @@
...
+		auth                autorest.Authorizer
+		client              *http.Client
+		cache               map[time.Time]map[uint64]*aggregate
+	}
+	tests := []struct {
+		name    string
+		fields  fields
+		wantErr bool
+	}{
+		// TODO: Add test cases.
```
glinton: Add tests

PR #4094, opened 2018-05-03
Emqtt batch output

jvrahav:
### Required for all PRs:


danielnelson:
There will still be multiple `telegraf.Metric` going to a single topic though, if we send them together in a batch message we will need to use `SerializeBatch` in order to have the JSON data encode correctly.  For influx and graphite output it will be the same, but for JSON the data shouldn't be line delimited because there are unicode characters that are newlines.  Also, you can imagine serialization formats that cannot be delimited with newlines such as yaml/toml that we might want to support in the future.

jvrahav:
@danielnelson can you review the request. I wasnt able to remove the diff in telegraf conf file. I dont see any difference in my file locally.

danielnelson:
Thanks, this will be in 1.7.0.

Issue #4111, opened 2018-05-06
Arista LANZ (Latency analyzer) input plugin


timhughes:
### Proposal:
I am proposing a new input plugin for monitoring Arista switches. They provide a stream in protobuf and I have written most of the code but I am having some difficulty with the testing which is probably my misunderstanding of how the accumulator works. Code is available at https://github.com/timhughes/telegraf/tree/c4c91ca220109d2f3cba09d5e207e51664ab23bf/plugins/inputs/lanz_consumer

In the test I am creating two endpoints and passing the accumulator to both endpoints. When I input a message `testProtoBufCongestionRecord1` to the first endpoint I get the correct data in the accumulator but when I input the second message `testProtoBufCongestionRecord2` I still only have the first message's  data.

https://www.arista.com/en/solutions/telemetry-analytics

If this question is better off in the community site please let me know and I will move the question there. 

### Use case: [Why is this important (helps with prioritizing requests)]
New plugin allowing more monitoring

timhughes:
I think I have worked out the testing side of things and now have a working multi endpoint consumer service. Currently untested in the real world so  I will put it through it's paces during the week. If someone is able to do a bit of a code review and let me know what needs improving before making a MR it would be appreciated https://github.com/timhughes/telegraf/tree/lanz_consumer_multi_server/plugins/inputs/lanz_consumer

littlespace:
@timhughes , just wanted to see if you have got the plugin working with Arista? also is it gRCP/gNMI or do you need LANZ server to get the plugin work?

timhughes:
We have been using it for a year without any issues. it something i keep meaning to get back and write tests for so that i can get it accepted into the mainline. We have 1 telegraph instance per data centre collecting from the gRPC stream. The code is here https://github.com/timhughes/telegraf/tree/lanz_consumer_multi_server/plugins/inputs/lanz_consumer

glinton:
When you get time to update your pr branch to resolve merge conflicts and add a readme, we can give it another review, seems useful.

timhughes:
I have fixed up the code, rebased and added a readme. PR is #4112 if someone wants to review the code again

612wong:
@timhughes 

Thank you so much for the lanz_consumer plugins. I found it working very well.
I'm using Grafana to graph the LANZ statistics but end users found they need to lookup what devices are connecting to the interesting interface. As a result, I used snmp plugin to get the ifAlias.

[...]

PR #4112, opened 2018-05-06
Arista LANZ consumer multi server. Production tested. 

timhughes:
Code for #4111 

This has been running for over 6 months in production. Handling multiple lanz endpoints publishing data to it. 

. 

### Required for all PRs:

- [ * ] Signed [CLA](https://influxdata.com/community/cla/).
- [ * ] Associated README.md updated.
- [ * ] Has appropriate unit tests.


plugins/inputs/lanz_consumer/lanz_consumer.go
```diff
@@ -0,0 +1,160 @@
...
+	c.done = make(chan bool)
+	go c.receiver()
+	return nil
+}
+
+func (c *LanzClient) receiver() {
+
+	u, err := url.Parse(c.Server)
+	if err != nil {
+		log.Fatal(err)
```
danielnelson: Move this check into Start, pass the *url.URL in as an argument.

plugins/inputs/lanz_consumer/lanz_consumer.go
```diff
@@ -0,0 +1,160 @@
...
+		select {
+		case <-c.done:
+			return
+		case msg := <-c.in:
+			cr := msg.GetCongestionRecord()
+			if cr != nil {
+				vals := map[string]interface{}{
+					"timestamp":     int64(cr.GetTimestamp()),
+					"queueSize":     int64(cr.GetQueueSize()),
+					"timeOfMaxQLen": int64(cr.GetTimeOfMaxQLen()),
```
danielnelson: Use snake_case instead of camelCase for fields/measurement.

plugins/inputs/lanz_consumer/lanz_consumer.go
```diff
@@ -0,0 +1,160 @@
...
+	return &LanzClient{}
+}
+
+func (c *LanzClient) Start(acc telegraf.Accumulator) error {
+	c.Lock()
+	defer c.Unlock()
+	c.acc = acc
+	u, err := url.Parse(c.Server)
+	if err != nil {
+		log.Fatal(err)
```
danielnelson: Return the error here

plugins/inputs/lanz_consumer/lanz_consumer.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+	client := lanz.New(lanz.WithAddr(u.Host), lanz.WithBackoff(1*time.Second), lanz.WithTimeout(10*time.Second))
+	c.client = &client
+	go func() {
+		client.Run(c.in)
+		c.done <- true
+	}()
+
+	c.in = make(chan *pb.LanzRecord)
+	c.done = make(chan bool)
```
danielnelson: Make the channel before starting the goroutine above, or it has a chance of writing to nil channel.

plugins/inputs/lanz_consumer/lanz_consumer.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+func NewLanzClient() *LanzClient {
+	return &LanzClient{}
+}
+
+func (c *LanzClient) Start(acc telegraf.Accumulator) error {
+	c.Lock()
+	defer c.Unlock()
+	c.acc = acc
```
danielnelson: Add shared items to the struct in the LanzConsumer, items that need to be different could be added to the LanzClient here.

I don't think you will need locks, Telegraf will serialize calls to the interface functions.

timhughes:
Thanks @danielnelson I will get on those.

plugins/inputs/lanz_consumer/lanz_consumer.go
timhughes: done

plugins/inputs/lanz_consumer/lanz_consumer.go
timhughes: done

plugins/inputs/lanz_consumer/lanz_consumer.go
timhughes: done

plugins/inputs/lanz_consumer/lanz_consumer.go
timhughes: done

lukebigum:
I've had this custom build of Telegraf running a LANZ input in one of our production sites for a few days now. It's been connected to ~20 switches and writing to InfluxDB.  A sample of the data can be seen here, along with a Grafana dashboard displaying the data:

https://gist.github.com/lukebigum/725e46c18f06f6b4053409ff3d21e43f

An example Telegraf config is attached above as well.

plugins/inputs/lanz_consumer/lanz_consumer.go
timhughes: done

plugins/inputs/lanz_consumer/lanz_consumer.go
timhughes: done

plugins/inputs/lanz_consumer/lanz_consumer.go
timhughes: Removed the locks

timhughes: Not 100% sure what you mean by `Add shared items to the struct in the LanzConsumer, items that need to be different could be added to the LanzClient here.` but other than that i think i have cleaned up all the issues

timhughes:
@danielnelson @glinton request another review please when you have some time.

timhughes:
@danielnelson @glinton request another review please when you have some time.

plugins/inputs/lanz_consumer/lanz_consumer.go
```diff
@@ -0,0 +1,154 @@
...
+	"github.com/influxdata/telegraf/plugins/inputs"
+)
+
+var sampleConfig = `
+  	## URL to Arista LANZ endpoint
+  	# servers = "tcp://localhost:50001"
+  	servers = [
+		"tcp://switch02.int.example.com:50001",
+		"tcp://switch02.int.example.com:50001"
+	]
```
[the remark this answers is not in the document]
timhughes: the security is based on acls configured within the switch. From the docs 

> To ensure client access to LANZ data, add a rule to any relevant ACL permitting traffic destined for the LANZ port (50001) before initiating a client connection for streaming from a remote host. A static rule (sequence number 130) in the default control plane ACL permits LANZ traffic, but a similar rule must be added to any user-created ACL.

plugins/inputs/lanz_consumer/README.md
```diff
@@ -0,0 +1,54 @@
+# Arista LANZ Consumer Input Plugin
```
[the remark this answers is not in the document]
timhughes: I think this is done to the correct way

Gopkg.toml
```diff
@@ -52,7 +52,7 @@
 
 [[constraint]]
   name = "github.com/golang/protobuf"
-  version = "1.1.0"
+  version = "1.3.1"
 
 [[constraint]]
   name = "github.com/google/go-cmp"
```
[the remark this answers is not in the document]
timhughes: I have set the version to the one i previously tested. setting it to latest caused some errors

plugins/inputs/lanz_consumer/lanz_consumer.go
```diff
@@ -0,0 +1,154 @@
...
+	for {
+
+		select {
+		case <-c.done:
+			return
+		case msg := <-c.in:
+			cr := msg.GetCongestionRecord()
+			if cr != nil {
+				vals := map[string]interface{}{
+					"timestamp":        int64(cr.GetTimestamp()),
```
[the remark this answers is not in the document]
timhughes: It isn't the receive time. it is the time reported internally by the switch. we use it when diagnosing issues occasionally to compare it with logs generated to syslog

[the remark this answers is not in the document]
danielnelson: Yes, right now we have a integer field with the switches timestamp and the metric timestamp is the time we receive it in Telegraf.  It is often useful to have these both of these timestamps, one of the advantages is that you can easily calculate the differences between the switch timestamps.

However, for selecting by time range or grouping by time for aggregation it is better supported for the metric timestamp, so another way it could make sense to handle this is to use the switches timestamp as the metric timestamp.

Both of these layouts are valid and have their own advantages/disadvantages, so I just wanted to point out that there is another option to consider.

plugins/inputs/lanz_consumer/lanz_consumer.go
```diff
@@ -0,0 +1,154 @@
...
+				c.acc.AddFields("global_buffer_usage_record", vals, tags)
+			}
+		}
+	}
+}
+
+func (c *LanzClient) Stop() {
+	c.Lock()
+	defer c.Unlock()
+	close(c.done)
```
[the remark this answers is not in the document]
timhughes: Thanks for this, it makes it much simpler. The tests i was a little unsure. In particular I don't know that running `receive` in a goroutine is a sensible way of doing these tests but it appears to do what I wanted. https://github.com/timhughes/telegraf/commit/6c51f716c351511e60b344094299e00f063d35af#r37639022

timhughes:
Would you like this all rebased and squashed ?

danielnelson:
> Would you like this all rebased and squashed?

No need, we always squash commits when we merge pull requests.

timhughes:
> Probably not, should be able to write to the input channel and then close it. After that call receive and it should exit after it processes all the data in the channel.

Extracted the business logic to and new func to make it easier to test.

timhughes:
@danielnelson This still say that it is waiting on a change from me and i don't know how to clear it down. I think i have done all the changes you requested.

Cheers

Tim

danielnelson:
Looks great, merged for 1.14

Issue #4110, opened 2018-05-05
CrateDB output plugin fails with memory input plugin


nslater:
### Relevant telegraf.conf:

```
# Read metrics about memory usage
[[inputs.mem]]
  # no configuration
```

### System info:

CrateDB 2.3

tested failing with two versions of Telegraf

installed via Homebrew on macOS 10.13.4:

```
$ telegraf --version
Telegraf v1.6.1 (git: unknown unknown)
```

installed via `go get` on macOS 10.13.4:

```
$ ./telegraf --version
Telegraf v1.7.0~d96bcac3 (git: master d96bcac3)
```

### Steps to reproduce:

1. `telegraf --input-filter mem --output-filter cratedb config > telegraf.conf`
2. configure the CrateDB connection string
3. `telegraf --config telegraf.conf`

### Expected behavior:

Telegraf runs with no errors, and the CrateDB table is created and populated

### Actual behavior:

repeated messages like this:

```
2018-05-05T20:51:42Z E! Error writing to output [cratedb]: unexpected type: uint64: 0x16618f000
```

the table is created, but no records are written to CrateDB

### Additional info:

this does not happen with the cpu input plugin

I have not tested any other plugins

PR #4117, opened 2018-05-07
Add uint support to cratedb output

danielnelson:
Since 1.6, fields can be int64, uint64, float64, bool, string.

For uint64 values too large to store in CrateDB, we save as the maximum int64 value.

@felixge Can you review?

closes: #4110

### Required for all PRs:

### DIFF
--- a/plugins/outputs/cratedb/cratedb.go
+++ b/plugins/outputs/cratedb/cratedb.go
@@ -121,6 +121,9 @@ func escapeValue(val interface{}) (string, error) {
 	case int64, float64:
 		return fmt.Sprint(t), nil
 	case uint64:
+		// The long type is the largest integer type in CrateDB and is the
+		// size of a signed int64.  If our value is too large send the largest
+		// possible value.
 		if t <= uint64(MaxInt64) {
 			return strconv.FormatInt(int64(t), 10), nil
 		} else {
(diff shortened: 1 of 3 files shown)


plugins/outputs/cratedb/cratedb.go
```diff
@@ -115,11 +118,19 @@ func escapeValue(val interface{}) (string, error) {
...
-	// produce those types, so it's hopefully ok.
-	case int, int32, int64, float32, float64:
+	case int64, float64:
 		return fmt.Sprint(t), nil
+	case uint64:
+		if t <= uint64(MaxInt64) {
+			return strconv.FormatInt(int64(t), 10), nil
+		} else {
+			return strconv.FormatInt(MaxInt64, 10), nil
+		}
```
felixge: So the idea here is to silently truncate large numbers because CrateDB can't store them? I generally prefer to return an error in these situations, but I think there should be at least a comment explaining the reason for this code.

danielnelson: Yes, this is closely based on the behavior of the InfluxDB output but I can definitely leave a comment.

We could log an error here however in most cases the value would probably be oversized on every interval, so we would want to log only once.  We can't return an error from `Write()` because then Telegraf would retry the metric.

felixge: OK

PR #4118, opened 2018-05-08
Don't report 0ms on timeout in dns_query

### LINKED ISSUE #2548: dns_query should not report 0ms query time on error
### Relevant telegraf.conf:
```
[[inputs.dns_query]]
  servers = ["8.8.8.8"]
  record_type = "A"
  domains = ["google.com"]
  timeout = 10
```

### System info:

Telegraf 1.2.1

### Steps to reproduce:

1. Run telegraf
2. Interrupt resolution to cause an error, such as a timeout.

### Expected behavior:
One of:
1. `query_time_ms=10000`
2. null value

### Actual behavior:
`query_time_ms=0`

### Additional info:
This is highly problematic as the value indicates that the response came back immediately. However this is not the case. The response didn't come back at all. The result should somehow indicate this by either setting the value to the maximum (the timeout value), or not setting it at all, and providing some other field indicating an error occurred.

danielnelson:
Adds `result` tag and `result_code` field for results: success, timeout, error.  If the result is not success, the `query_time_ms` field will be omitted.

closes #2548

### Required for all PRs:

### DIFF
--- a/plugins/inputs/dns_query/README.md
+++ b/plugins/inputs/dns_query/README.md
@@ -3,9 +3,8 @@
 The DNS plugin gathers dns query times in miliseconds - like [Dig](https://en.wikipedia.org/wiki/Dig_\(command\))
 
 ### Configuration:
-
-```
-# Sample Config:
+```toml
+# Query given DNS server and gives statistics
 [[inputs.dns_query]]
   ## servers to query
   servers = ["8.8.8.8"]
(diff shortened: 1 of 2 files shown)


discoduck2x:
@danielnelson exited to test this - is there a way to get hold of nightly builds for telegraf?

danielnelson:
Yes, these packages are updated by the CircleCI build:

https://github.com/influxdata/telegraf#nightly-builds

randallt:
Why wouldn't you report the timeout value for query_time_ms? I'm trying to use this right now to help debug an issue and if a request times out, I would expect to see a high response time recorded.

danielnelson:
It has been some time, but I believe it stems back to the changes on http_response where we had an existing field `response_time` and we didn't want to set it to the timeout as there was no response.  The changes made here were based on the earlier change although you could argue that with `query_time_ms` the field name is valid even if the query times out.

One potential issue with mixing success and timeout together is that when you aggregate the timeout values can skew the results.  This is probably not a big problem because you could use a tag to exclude the timeout times.

randallt:
The timeouts are exactly the thing I'm looking for. At first I didn't even notice them until I saw the log entries. Seems a backward way of doing it.

danielnelson:
If the `query_time_ms` is used, one might still miss them as it could appear that they are successfully completing after timeout seconds.  I think gaps in the data are probably more noticeable but it depends on how you are viewing the data, and for alerts you really need to watch the `result` tag.

PR #4142, opened 2018-05-12
Fix dropwizard parsing error for metrics that need escaped

danielnelson:
If the dropwizard parser cannot convert the metric name into a valid line protocol series then we will accept the name as is.

This adds a way to parse only the `<measurement>,<tagset>` portion of line protocol and uses it to create new metrics instead of serializing the entire dropwizard metric to line protocol.  Anything that would require escaping to be valid cannot be converted, because escaping it fully would prevent tags from being extracted.

### Required for all PRs:

### DIFF
--- a/metric.go
+++ b/metric.go
@@ -54,6 +54,8 @@ type Metric interface {
 	AddField(key string, value interface{})
 	RemoveField(key string)
 
+	SetTime(t time.Time)
+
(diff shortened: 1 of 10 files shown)


danielnelson:
@atzoum Can you review/test?

atzoum:
LGTM :)

PR #4150, opened 2018-05-15
Added async producer mode for kafka

apollo-3:
### Required for all PRs:

### DIFF
--- a/plugins/outputs/kafka/README.md
+++ b/plugins/outputs/kafka/README.md
@@ -10,6 +10,10 @@ This plugin writes to a [Kafka Broker](http://kafka.apache.org/07/quickstart.htm
   ## Kafka topic for producer messages
   topic = "telegraf"
 
+  ## Producer mode, either sync or async
+  ## With async producer required_acks setting doesn't matter
+  # mode = "async"
+
   ## Optional topic suffix configuration.
   ## If the section is omitted, no suffix is used.
   ## Following topic suffix methods are supported:
--- a/plugins/outputs/kafka/kafka.go
+++ b/plugins/outputs/kafka/kafka.go
@@ -25,6 +25,8 @@ type (
 		Brokers []string
 		// Kafka topic
 		Topic string
+		// Producer mode
+		Mode string
 		// Kafka topic suffix option
 		TopicSuffix TopicSuffix `toml:"topic_suffix"`
 		// Routing Key Tag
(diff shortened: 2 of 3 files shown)


danielnelson:
Can you tell me a little bit about why the async producer is helpful?  When would a user want to enable async?

apollo-3:
Hello,

We have a Telegraf fleet of 4 nodes that receive metrics from ~ 10K nodes or even more. Write rate to Kafka with sync producer is low in comparison with influx... eventually we start seeing errors like “internal agent didn’t manage to gather metrics during 10s, skipping flush interval”. And cpu and memory usage on the fleet is totally normal as well as network usage... we were desperate until I rewrote the code.... now all errors are gone and tick stacks works through Kafka. Telegraf consumer fleet reads metrics from Kafka and writes into InfluxDB.
I also saw other posts in the Internet where people stumble over Kafka output plugin write rate limit ~ 10KB per second if I’m not mistaken...
Could you please help us with my original request to merge it?
Thank you

Thank you



Sent from Mail.Ru app for iOS


Tuesday, May 15, 2018, 8:59 PM +0300 from notifications@github.com  <notifications@github.com>:

danielnelson:
I think the main difference between this method and the current one is that you are not waiting for success or errors.  I'd like to stop using the sync producer (which just wraps the async producer) completely, but then we need to have a configurable number of messages in flight and we still need to monitor the errors channel.

Another potential performance improvement would be to use SerializeBatch to place the full batch into a single Kafka message.

apollo-3:
Hello,

So does it mean that you going to issue the fix?

Sent from Mail.Ru app for iOS


Wednesday, May 16, 2018, 1:47 AM +0300 from notifications@github.com  <notifications@github.com>:

apollo-3:
Hello Daniel,

We ran into another quite annoying issue with Kafka Plugin.
I hoped you could tell me where to look for a problem.

When Kafka is not available telegraf can't write to Kafka of source. But when it is back up and running Telegraf Kafka Plugin doesn't try to reconnect.
I believe this logic is outside of plugin itself... Could you please give me a hint, ideally a file name where reconnection is handled.
Or maybe you could share what needs to be added, modified in order to force output plugin reconnect on failures.

Thank you

danielnelson:
@apollo-3 In 1.8.0 we switched to using `SendMessages` instead of `SendMessage`, and it improved performance substantially.

https://github.com/influxdata/telegraf/pull/4491

With this change I don't believe we need to switch to the AsyncProducer directly.  Can you test out the new release?

Sorry about missing your last comment about not reconnecting, it should be automatic.  If you are still having trouble in 1.8.0 can you open a new issue?

PR #4155, opened 2018-05-16
Add support for additional Consul checks' tags and splitting them by a delimiter

lcharkiewicz:
This PR introduces 2 new options:
1) Gathering all Consul check's tags, not only node, service_name and check_id.
If a check's tag is a single word then Telegraf will send this tag's value as a key and a value. This probably isn't the best option, but that's why there's another one below.
2) Splitting check's tag by a specified delimiter.
It's useful if Consul checks have tags formatted like: "env:production". After setting tag_delimiter = ":" these tags will be sent as proper Telegraf tags.

Example:
`consul_health_checks,cluster=bigcluster,host=11.22.33.44,release=2612cb161526038358,service=testservice,service_name=http-testservice-bigcluster,task_type=http check_name="service_tcp_port_check",critical=0i,passing=1i,service_id="ewpft7gojkilm2exq2mldlhpzrabzfkl",status="passing",warning=0i 1526477851000000000`

### Required for all PRs:

### DIFF
--- a/Godeps
+++ b/Godeps
@@ -30,7 +30,7 @@ github.com/gorilla/mux 53c1911da2b537f792e7cafcb446b05ffe33b996
 github.com/go-redis/redis 73b70592cdaa9e6abdfcfbf97b4a90d80728c836
 github.com/go-sql-driver/mysql 2e00b5cd70399450106cec6431c2e2ce3cae5034
 github.com/hailocab/go-hostpool e80d13ce29ede4452c43dea11e79b9bc8a15b478
-github.com/hashicorp/consul 63d2fc68239b996096a1c55a0d4b400ea4c2583f
+github.com/hashicorp/consul cf57a23bb1698dde5365aafeffd8bcf88d95d646
 github.com/influxdata/tail c43482518d410361b6c383d7aebce33d0471d7bc
 github.com/influxdata/toml 5d1d907f22ead1cd47adde17ceec5bda9cacaf8f
 github.com/influxdata/wlog 7c63b0a71ef8300adc255344d275e10e5c3a71ec
--- a/plugins/inputs/consul/consul.go
+++ b/plugins/inputs/consul/consul.go
@@ -2,6 +2,8 @@ package consul
 
 import (
 	"net/http"
+	"regexp"
+	"strings"
 
 	"github.com/hashicorp/consul/api"
 	"github.com/influxdata/telegraf"
@@ -17,6 +19,8 @@ type Consul struct {
 	Password   string
 	Datacentre string
 	tls.ClientConfig
+	GatherAllTags bool
+	TagDelimiter  string
 
 	// client used to connect to Consul agnet
 	client *api.Client
@@ -45,6 +49,14 @@ var sampleConfig = `
   # tls_key = "/etc/telegraf/key.pem"
   ## Use TLS but skip chain & host verification
   # insecure_skip_verify = true
+
+  ## Gather all Consul checks' tags
+  # gather_all_tags = false
+
+  ## Consul checks' tag splitting
+  # When tags are formatted like "key:value" with ":" as a delimiter then
+  # they will be splitted and reported as proper key:value in Telegraf
+  # tag_delimiter = ":"
 `
 
 func (c *Consul) Description() string {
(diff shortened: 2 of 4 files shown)


plugins/inputs/consul/consul.go
```diff
@@ -111,6 +123,22 @@ func (c *Consul) GatherHealthCheck(acc telegraf.Accumulator, checks []*api.Healt
 		tags["service_name"] = check.ServiceName
 		tags["check_id"] = check.CheckID
 
+		if c.GatherAllTags {
+			for _, checkTag := range check.ServiceTags {
+				if c.TagDelimiter != "" && len(c.TagDelimiter) == 1 {
+					r, _ := regexp.Compile("\\w+" + c.TagDelimiter + "\\w+")
```
danielnelson: I don't think we need a regex, just `SplitN(tag, delimiter, 2)`, and then assign the correct tag value based on the len of the returned value.

plugins/inputs/consul/consul.go
```diff
@@ -45,6 +49,14 @@ var sampleConfig = `
   # tls_key = "/etc/telegraf/key.pem"
   ## Use TLS but skip chain & host verification
   # insecure_skip_verify = true
+
+  ## Gather all Consul checks' tags
+  # gather_all_tags = false
```
danielnelson: I think we should remove this option and always gather all tags.  Plugin users can remove unwanted tags using [measurement filtering](https://github.com/influxdata/telegraf/blob/master/docs/CONFIGURATION.md#measurement-filtering).

Godeps
```diff
@@ -30,7 +30,7 @@ github.com/gorilla/mux 53c1911da2b537f792e7cafcb446b05ffe33b996
 github.com/go-redis/redis 73b70592cdaa9e6abdfcfbf97b4a90d80728c836
 github.com/go-sql-driver/mysql 2e00b5cd70399450106cec6431c2e2ce3cae5034
 github.com/hailocab/go-hostpool e80d13ce29ede4452c43dea11e79b9bc8a15b478
-github.com/hashicorp/consul 63d2fc68239b996096a1c55a0d4b400ea4c2583f
+github.com/hashicorp/consul cf57a23bb1698dde5365aafeffd8bcf88d95d646
```
danielnelson: Set to the rev of the last release, currently `5174058f0d2bda63fa5198ab96c33d9a909c58ed`.

danielnelson:
Thanks!

plugins/inputs/consul/consul.go
freeseacher: That is acctually not true. My app registers in consul and pass as tag uniq id. That is how it looks like now with telegraf 1.7.1
```
consul_health_checks_passing{="",
018ecd7a_4ca0_109d_cc67_45f4ba438280="",
03b93403_3960_49bb_81fc_8ebe0ff4c79a="",
03e9dd4a_b8de_48bf_ba98_050823dbe816="",
05cb54a7_1748_405b_9671_ccd630256364="",
06c7b6ac_9fd9_4494_ae3c_ac5d1daa9580="",
07b1163b_16ca_497b_afcc_2c9e6662f4ab="",
0b4c9bc8_78cf_407d_8a93_7d0d022cbab1="",
0b720e27_dc8a_e821_2638_08890c361783="",
0eaf2333_5929_4fc6_a304_c2452a05146a="",
10a7f37d_ff1b_4b90_b613_d3d500991355="",
135df3c5_822a_478c_a5e1_4fcdb9db23af="",
14e02d39_0afb_4e33_9a77_5d17a62b613b="",
```
So tags are random and i really don't know the way i exclude them from tags.
BTW there are also tag '' that makes my prometheus sad with 
`prom_1          | level=warn ts=2019-01-04T12:03:29.464871059Z caller=scrape.go:835 component="scrape manager" scrape_pool=stage target=http://10.143.114.19:9273/metrics msg="append failed" err="expected label name, got \"EQUAL\""
`
Of cause i can setup taginclude but that have to be split responsibility between devs and ops.

danielnelson: Can you open a new issue for this?

danielnelson: #5522

PR #4164, opened 2018-05-18
Add new output plugin httpjson

opan:
### Required for all PRs:

- [v] Signed [CLA](https://influxdata.com/community/cla/).
- [v] Associated README.md updated.
- [v] Has appropriate unit tests.

### DIFF
--- a/plugins/outputs/all/all.go
+++ b/plugins/outputs/all/all.go
@@ -11,7 +11,7 @@ import (
 	_ "github.com/influxdata/telegraf/plugins/outputs/file"
 	_ "github.com/influxdata/telegraf/plugins/outputs/graphite"
 	_ "github.com/influxdata/telegraf/plugins/outputs/graylog"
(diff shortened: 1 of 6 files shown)


danielnelson:
We recently added the [http output](https://github.com/influxdata/telegraf/tree/master/plugins/outputs/http) which will be included in 1.7.0.  I believe it covers the same ground as this pull request, can you take a look at it to see if it fulfills you needs?

opan:
Thank you for the response! Okay, it looks like the one I created. But only it is missing one feature, I need to able to send an additional data.

My version have one more options, I called it `data`, we can add as many params we need to be sent to the targeted API.

I think I will take a look at the source code and see if I am able to made a PR and adding this feature.

opan:
And btw, when version 1.7.0 will be released? Because I need to use it soon. At least for monday.

danielnelson:
I'm going to close this pull request, but let me know if using HTTP headers will work for your use case.

Issue #3830, opened 2018-02-25
Add Graphite Tag support


eMerzh:
## Feature Request

Since latest versions (1.1) graphite support tags , it would be cool if telegraf support them as they might be a good way to structure your metrics and it might be closer to what influx and friends are doing.

http://graphite.readthedocs.io/en/latest/tags.html

6d6178:
I would also love to see this implemented, especially because telegraf already has a tagging system in place.

R4scal:
I was opened PR https://github.com/influxdata/telegraf/pull/4165. This is a concept that working for me. Any feedback is welcomed

danielnelson:
Thanks @R4scal, support will be in 1.7.

gena01:
Am i reading this wrong that this only adds support for the "output" side of Telegraf. Anyway to parse these on the input side?

danielnelson:
@gena01 That's a good point, could you open a new feature request specifically for the graphite input data format?

PR #4165, opened 2018-05-18
Add support for Graphite 1.1.x protocol

R4scal:
Hi
This PR will add support for Graphite 1.1.x protocol (with tags support) http://graphite.readthedocs.io/en/latest/tags.html#graphite-tag-support 

This is a concept that working for me. Any feedback is welcomed


danielnelson:
Thanks for the PR, would it be possible for you to add the new functionality into the old graphite serializer and add an option `graphite_tag_support`?  I think this will cut down on duplicate logic.

R4scal:
@danielnelson Done. I was add `graphite_tag_support` option to original graphite oputput

R4scal:
`name` as tag name is reserved for graphite metric and will be replaced with `_name` in last commit,  because graphite server silently drop this tag from metrics

danielnelson:
closes #3830

PR #4181, opened 2018-05-22
Syslog input plugin

leodido:
This PR introduces a syslog input service plugin (acting as a syslog receiver) supporting:

- TLS (or TCP) as mandated by RFC5425
- UDP as mandated by RFC5426

Thanks a lot to @goller for his feedback, help, and support.

### Required for all PRs:


plugins/inputs/syslog/syslog_test.go
```diff
@@ -0,0 +1,67 @@
...
+			i--
+		}
+		cache >>= letterIdxBits
+		remain--
+	}
+
+	return string(b)
+}
+
+func init() {
```
danielnelson: No reason for init, just initialize them where they are declared.

plugins/inputs/syslog/syslog.go
```diff
@@ -0,0 +1,378 @@
...
+	return ts
+}
+
+func fields(msg *rfc5424.SyslogMessage) map[string]interface{} {
+	flds := map[string]interface{}{
+		"version": msg.Version(),
+	}
+
+	if msg.Timestamp() != nil {
+		flds["timestamp"] = *msg.Timestamp()
```
danielnelson: Should we use the timestamp as the metric timestamp?  What have we decided about dealing with conflicting timestamps with the same series?

plugins/inputs/syslog/README.md
```diff
@@ -0,0 +1,110 @@
...
+  protocol = "udp"
+  address = ":6514"
+```
+
+#### Other configs
+
+Other available configurations are:
+
+- `keep_alive_period`, `max_connections` for stream sockets
+- `best_effort` to tell the parser to work until it is able to do and extract partial but valid info
```
danielnelson: Please add more details on how this works or add link to go-syslog repo.

plugins/inputs/syslog/README.md
```diff
@@ -0,0 +1,110 @@
...
+    - **facility** (`string`)
+    - **facility_message** (`string`)
+    - hostname (`string`)
+    - appname (`string`)
+
+The name of fields in _italic_ corresponds to their runtime value.
+
+The fields/tags which name is in **bold** will always be present when a valid Syslog message has been received.
+
+### Syslog transport sender
```
danielnelson: Title this section something like `RSYSLOG Integration`

plugins/inputs/syslog/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+### Configuration:
+
+#### TCP
+
+The minimal configuration is the following:
+
+```toml
+[[inputs.syslog]]
+  address = ":6514"
```
danielnelson: Just have a single configuration, no need to show it in multiple potential configurations.

plugins/inputs/syslog/syslog_test.go
```diff
@@ -0,0 +1,67 @@
...
+)
+
+func TestListenError(t *testing.T) {
+	receiver := &Syslog{
+		Address: "wrong address",
+	}
+	require.Error(t, receiver.Start(&testutil.Accumulator{}))
+}
+
+func getRandomString(n int) string {
```
danielnelson: It would be preferable to have non random strings in unittests, to increase determinism.  Programmatic creation is fine of course.

plugins/inputs/syslog/rfc5426_test.go
```diff
@@ -0,0 +1,276 @@
...
+				want = tc.wantStrict
+			}
+			if !cmp.Equal(want, got) {
+				t.Fatalf("Got (+) / Want (-)\n %s", cmp.Diff(want, got))
+			}
+		})
+	}
+}
+
+func TestUDPInBestEffortMode(t *testing.T) {
```
danielnelson: The way these tests are structured make it more likely that they will fail due to a previous test case, because you are reusing the unit under test.  I would create all tested code new for each test, unless you are specifically testing multiple connections.

plugins/inputs/syslog/syslog.go
```diff
@@ -0,0 +1,378 @@
...
+    ## If no port is specified, 6514 is used (RFC5425#section-4.1).
+	server = ":6514"
+	
+	## Protocol (default = tcp)
+	## Should be one of the following values:
+	## tcp, tcp4, tcp6, unix, unixpacket, udp, udp4, udp6, ip, ip4, ip6, unixgram.
+	## Otherwise forced to the default.
+	# protocol = "tcp"
+
+    ## TLS Config
```
goller: This needs to be indented to the same level as the rest of the config

plugins/inputs/syslog/syslog.go
```diff
@@ -0,0 +1,378 @@
...
+func (s *Syslog) listenStream(acc telegraf.Accumulator) {
+	defer s.wg.Done()
+
+	s.connections = map[string]net.Conn{}
+
+	for {
+		conn, err := s.tcpListener.Accept()
+		if err != nil {
+			if !strings.HasSuffix(err.Error(), ": use of closed network connection") {
+				log.Println(err)
```
goller: Remove this log statement

plugins/inputs/syslog/syslog.go
```diff
@@ -0,0 +1,378 @@
...
+			}
+			break
+		}
+
+		if s.ReadTimeout != nil && s.ReadTimeout.Duration > 0 {
+			s.udpListener.SetReadDeadline(time.Now().Add(s.ReadTimeout.Duration))
+		}
+
+		p := rfc5424.NewParser()
+		mex, err := p.Parse(b[:n], &s.BestEffort)
```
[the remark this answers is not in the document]
leodido: Changed to `message`.

plugins/inputs/syslog/syslog.go
```diff
@@ -0,0 +1,378 @@
...
+}
+
+func (s *Syslog) listenPacket(acc telegraf.Accumulator) {
+	defer s.wg.Done()
+	b := make([]byte, ipMaxPacketSize)
+	for {
+		n, _, err := s.udpListener.ReadFrom(b)
+		if err != nil {
+			if !strings.HasSuffix(err.Error(), ": use of closed network connection") {
+				log.Println(err)
```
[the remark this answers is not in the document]
leodido: Done.

plugins/inputs/syslog/syslog.go
leodido: Done.

plugins/inputs/syslog/syslog.go
leodido: Done.

plugins/inputs/syslog/syslog.go
```diff
@@ -0,0 +1,378 @@
...
+	tcpListener   net.Listener
+	tlsConfig     *tls.Config
+	connections   map[string]net.Conn
+	connectionsMu sync.Mutex
+
+	udpListener net.PacketConn
+}
+
+var sampleConfig = `
+    ## Specify an ip or hostname with port - eg., localhost:6514, 10.0.0.1:6514
```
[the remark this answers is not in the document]
leodido: Ok.

plugins/inputs/syslog/syslog.go
```diff
@@ -0,0 +1,378 @@
...
+	for {
+		conn, err := s.tcpListener.Accept()
+		if err != nil {
+			if !strings.HasSuffix(err.Error(), ": use of closed network connection") {
+				log.Println(err)
+				acc.AddError(err)
+			}
+			break
+		}
+		var tcpConn, _ = conn.(*net.TCPConn)
```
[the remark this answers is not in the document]
leodido: Yes, to set keep alive since `net.Conn` interface does not have `SetKeepAlive` method while `net.TCPConn` has it.

plugins/inputs/syslog/syslog.go
```diff
@@ -0,0 +1,378 @@
...
+`
+
+// SampleConfig returns sample configuration message
+func (s *Syslog) SampleConfig() string {
+	return sampleConfig
+}
+
+// Description returns the plugin description
+func (s *Syslog) Description() string {
+	return "Influx syslog receiver as per RFC5425"
```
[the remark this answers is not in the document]
leodido: Done.

plugins/inputs/syslog/syslog.go
```diff
@@ -0,0 +1,378 @@
...
+
+func (s *Syslog) removeConnection(c net.Conn) {
+	s.connectionsMu.Lock()
+	delete(s.connections, c.RemoteAddr().String())
+	s.connectionsMu.Unlock()
+}
+
+func (s *Syslog) handle(conn net.Conn, acc telegraf.Accumulator) {
+	defer s.removeConnection(conn)
+	defer conn.Close()
```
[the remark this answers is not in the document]
leodido: Right, done.

plugins/inputs/syslog/syslog.go
```diff
@@ -0,0 +1,378 @@
...
+	if msg.StructuredData() != nil {
+		for sdid, sdparams := range *msg.StructuredData() {
+			if len(sdparams) == 0 {
+				// When SD-ID does not have params we indicate its presence with a bool
+				flds[sdid] = true
+				continue
+			}
+			for name, value := range sdparams {
+				// Using whitespace as separator since it is not allowed by the grammar within SDID
+				flds[sdid+" "+name] = value
```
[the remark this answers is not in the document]
leodido: I've chosen to use the whitespace as a separator since this character cannot be present by specification (and so by go-syslog library) within a `SDID` or a within a `SDPARAMNAME`.
Look [here](https://github.com/influxdata/go-syslog/blob/develop/rfc5424/rfc5424.rl#L59).

If we choose to use `_` we'll not longer be able to distinguish a `SDID` without parameters, from a `SDPARAM` looking only at the field name.

In such setting, a field `abc_xzy` could represent at the same time a `SDID` without parameters -  `.. [abc_xyz] ..` - or a `SDID` with a parameter with name `xyz` - `.. [abc xyz="value"] ..`.
Thus, a user, to distinguish a posteriori, should look at the kind of the field's value - ie.,`bool` vs `string`.

Let's talk about it together pls :)

plugins/inputs/syslog/syslog.go
```diff
@@ -0,0 +1,378 @@
...
+func (s *Syslog) Stop() {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	if s.Closer != nil {
+		s.Close()
+	}
+	s.wg.Wait()
+
+	log.Printf("I! Stopped syslog receiver at %s\n", s.Address)
```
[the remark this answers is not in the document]
leodido: Ok.

plugins/inputs/syslog/syslog_test.go
leodido: Right, done.

plugins/inputs/syslog/syslog.go
leodido: In a previous version of this plugin we were using it as a metric timestamp.

The reason we have chosen to store it as a fields depends on the fact that the syslog timestamp represents when the event that the message represents has occurred, not when telegraf received it.

plugins/inputs/syslog/README.md
leodido: Ok, what configuration should we show? The minimal one?

plugins/inputs/syslog/README.md
leodido: Ok.

plugins/inputs/syslog/README.md
leodido: Ok.

plugins/inputs/syslog/syslog.go
```diff
@@ -0,0 +1,378 @@
...
+	}
+
+	if s.isTCP {
+		l, err := net.Listen(s.Protocol, s.Address)
+		if err != nil {
+			return err
+		}
+		s.Closer = l
+		s.tcpListener = l
+		if tlsConfig, _ := s.TLSConfig(); tlsConfig != nil {
```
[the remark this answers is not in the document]
leodido: Done.

plugins/inputs/syslog/syslog_test.go
leodido: There's a pseudo-random source within this method using now as seed.
If you prefer we can stuck to a preconfigured seed value (so not now) or we can completely remove this and simply generate a string with length 7681.

plugins/inputs/syslog/README.md
leodido: Done showing the config for TLS.

plugins/inputs/syslog/syslog.go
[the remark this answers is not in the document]
leodido: Talking about it with @goller he suggested to create an option - eg., `sd_separator` - that defaults to `_`.

plugins/inputs/syslog/README.md
danielnelson: Run `telegraf -usage syslog` and use the output.

plugins/inputs/syslog/syslog_test.go
danielnelson: I would probably just generate a string with the desired length.

plugins/inputs/syslog/syslog.go
```diff
@@ -0,0 +1,378 @@
...
+				acc.AddError(err)
+			}
+			break
+		}
+
+		if s.ReadTimeout != nil && s.ReadTimeout.Duration > 0 {
+			s.udpListener.SetReadDeadline(time.Now().Add(s.ReadTimeout.Duration))
+		}
+
+		p := rfc5424.NewParser()
```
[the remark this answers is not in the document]
leodido: Done.

plugins/inputs/syslog/rfc5425_test.go
```diff
@@ -0,0 +1,657 @@
...
+	"time"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/influxdata/telegraf/internal"
+	"github.com/influxdata/telegraf/testutil"
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	serviceRootPEM = `-----BEGIN CERTIFICATE-----
```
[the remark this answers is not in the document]
leodido: Thanks, done.

plugins/inputs/syslog/syslog_test.go
leodido: Done.

plugins/inputs/syslog/rfc5426_test.go
leodido: Should be better now.

plugins/inputs/syslog/syslog.go
```diff
@@ -0,0 +1,378 @@
...
+	if res.Message != nil {
+		acc.AddFields("syslog", fields(res.Message), tags(res.Message), s.now())
+	}
+}
+
+func tags(msg *rfc5424.SyslogMessage) map[string]string {
+	ts := map[string]string{}
+	if lvl := msg.SeverityLevel(); lvl != nil {
+		ts["severity"] = strconv.Itoa(int(*msg.Severity()))
+		ts["severity_level"] = *lvl
```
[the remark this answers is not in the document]
leodido: Now stores `severity_code` (int) as a field and `severity` (string) as a tag (short version).
Same for facility.

Godeps
```diff
@@ -32,6 +32,7 @@ github.com/go-redis/redis 73b70592cdaa9e6abdfcfbf97b4a90d80728c836
 github.com/go-sql-driver/mysql 2e00b5cd70399450106cec6431c2e2ce3cae5034
 github.com/hailocab/go-hostpool e80d13ce29ede4452c43dea11e79b9bc8a15b478
 github.com/hashicorp/consul 5174058f0d2bda63fa5198ab96c33d9a909c58ed
+github.com/influxdata/go-syslog dcd9920f1eea047ffa10928fd2b7fbad6c7abe83
```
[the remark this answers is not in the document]
leodido: Library `go-syslog` has a `v1.0.0` tag.
Updated `Godeps` to contain the tag commit hash.

plugins/inputs/syslog/README.md
leodido: Done.

plugins/inputs/syslog/syslog.go
```diff
@@ -0,0 +1,378 @@
...
+
+	udpListener net.PacketConn
+}
+
+var sampleConfig = `
+    ## Specify an ip or hostname with port - eg., localhost:6514, 10.0.0.1:6514
+    ## Address and port to host the syslog receiver.
+    ## If no server is specified, then localhost is used as the host.
+    ## If no port is specified, 6514 is used (RFC5425#section-4.1).
+	server = ":6514"
```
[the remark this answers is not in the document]
leodido: Done.

plugins/inputs/syslog/syslog.go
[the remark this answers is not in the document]
leodido: Done.

plugins/inputs/syslog/syslog.go
leodido: So at the end as @goller suggested me, assuming system clock is monotonic, we are using its timestamps as metric. Incrementing a time by 1 nanoseconds in the case it is equal to the previous done.

For future reference, in cases where this assumption can not be made or external timestamps are used as metrics, a solution based on bloom filter (ie., definitely not in set versus false-positives) could fit well our needs. I have one in stash.

Anyway should be ok now.

plugins/inputs/syslog/syslog.go
```diff
@@ -0,0 +1,419 @@
...
+func fields(msg rfc5424.SyslogMessage, s *Syslog) map[string]interface{} {
+	// Not checking assuming a minimally valid message
+	flds := map[string]interface{}{
+		"version": msg.Version(),
+	}
+	flds["severity_code"] = int(*msg.Severity())
+	flds["facility_code"] = int(*msg.Facility())
+
+	if msg.Timestamp() != nil {
+		flds["timestamp"] = *msg.Timestamp()
```
danielnelson: I think this doesn't work because this is a time.Time and a field has to be one of these types.  If this returns nil then the field is silently dropped!

https://github.com/influxdata/telegraf/blob/ce3b367dacd415d5b9fbe3be104e671fe51546af/metric/metric.go#L251-L286

danielnelson: Probably should encode as an integer in nanosecond precision

leodido: :+1:

PR #4184, opened 2018-05-22
Write standard output to /var/log/telegraf with systemd

### LINKED ISSUE #4174: Log to /var/log/telegraf/telegraf.log on systems using systemd
### Proposal:

On systems using sysvinit, logs are set to write to `/var/log/telegraf/telegraf.log`.  For systems using systemd the logs instead end up in `/var/log/messages`, the `/var/log/telegraf` directory remains empty.

### Current behavior:

- systemd systems log to `/var/log/messages`

### Desired behavior:

- systemd systems log to `/var/log/telegraf/telegraf.log`

danielnelson:
closes #4174 

### Required for all PRs:

### DIFF
--- a/scripts/telegraf.service
+++ b/scripts/telegraf.service
@@ -11,6 +11,7 @@ ExecReload=/bin/kill -HUP $MAINPID
 Restart=on-failure
 RestartForceExitStatus=SIGPIPE
 KillMode=control-group
+StandardError=/var/log/telegraf/telegraf.log
 
 [Install]
 WantedBy=multi-user.target


danielnelson:
@phemmer Can you review?

phemmer:
This is essentially a half-reversion of #1732

I personally don't agree with the change as it prevents the logs from going to the systemd journal. This means if people have systemd configured to do log shipping to an external system, the telegraf error logs won't be picked up.
Systemd based systems also do not inherently have log file rotation, meaning this log file will go unchecked and could fill the drive. *Some* systems utilizing systemd do, but those systems generally have a log daemon (such as rsyslog or syslog-ng) read from the journal and write to the files. As such it is expected to have to clean up the files those daemons generate (and the OS generally comes preconfigured to do so).

danielnelson:
That makes sense, and of course I'd like Telegraf to act as well behaved as possible, while being fairly consistent across init systems.  That said, my instinct is that an admin would rather have the log file split out due to how talkative Telegraf can be.

Since journald often is setup to forward to syslog, perhaps we should add rules for syslog handling to rsyslog?

phemmer:
> my instinct is that an admin would rather have the log file split out due to how talkative Telegraf can be

`journalctl` has a `-p`/`--priority` option you can use to only see a certain range.

I also don't think telegraf is that noisy. It spits out a little noise sure, but it's not overly verbose (unless misconfigured).


> Since journald often is setup to forward to syslog, perhaps we should add rules for syslog handling to rsyslog?

You could, but what about systems which don't have a syslog daemon? Or systems which use syslog-ng vs rsyslog? And then having to configure log rotation to clean these up? You could provide all of the above, using the `.d` directories (if the locations for these `.d` directories are standard across distros, and I have no idea if this is the case), so that they only take effect if that daemon is present on the system. This feels a little icky, but I suppose it's not terribly evil.

&nbsp;

However I should note that my objection is only on principal. I override the telegraf unit file entirely. So whatever is provided by the package doesn't affect me.

danielnelson:
Closing, this is a change we want to make.  Further discussion on #4174

PR #4185, opened 2018-05-22
Initial Commit for Splunk Output plugin

tmartin14:
### Required for all PRs:

Initial commit of an output plugin to send telegraf metrics directly to a Splunk metrics index via HEC.

### DIFF
--- a/plugins/outputs/splunk/README.md
+++ b/plugins/outputs/splunk/README.md
@@ -1,0 +1,76 @@
+# Splunk Output Plugin
+
+This plugin writes metrics to a [Splunk Metric Index](https://docs.splunk.com/Documentation/Splunk/latest/Indexer/Setupmultipleindexes#Create_metrics_indexes) via a [Splunk HTTP Event Collector (HEC)](http://docs.splunk.com/Documentation/Splunk/latest/Data/UsetheHTTPEventCollector).  This requires an `Authorization Token` which will be created when you setup your Splunk HEC Token.
+
+Notes:
+1. If any point value cannot be converted to a float64, that metric will be skipped.
+2. Metrics are grouped by converting any `_` characters to `.` in the Metric Name.
+
+
+
+### Configuration:
+
+```toml
+# Configuration for Splunk server to send metrics to
+[[outputs.splunk]]
+  ## REQUIRED
+  ## URL of the Splunk Enterprise HEC endpoint (i.e.: http://localhost:8088/services/collector)
+  SplunkUrl = "http://localhost:8088/services/collector"
+
+  ## REQUIRED
+  ## Splunk Authorization Token for sending data to a Splunk HTTPEventCollector (HEC).
+  ##   Note:  This Token should map to a 'metrics' index in Splunk.  
+  AuthString = "XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX"
+
+  ## OPTIONAL:  prefix for metrics keys
+  #prefix = "my.specific.prefix."
+
+  ## OPTIONAL:  whether to use "value" for name of simple fields
+  #simple_fields = false
+
+  ## OPTIONAL:  character to use between metric and field name.  defaults to . (dot)
+  #metric_separator = "."
+
+  ## OPTIONAL:  Convert metric name paths to use metricSeperator character
+  ## When true (default) will convert all _ (underscore) chartacters in final metric name
+  #convert_paths = true
+
+  ## OPTIONAL:  Use Regex to sanitize metric and tag names from invalid characters
+  ## Regex is more thorough, but significantly slower
+  #use_regex = false
+
+  ## OPTIONAL:  whether to convert boolean values to numeric values, with false -> 0.0 and true -> 1.0.  default true
+  #convert_bool = true
+```
+
+
+### Convert Path & Metric Separator
+If the `convert_path` option is true any `_` in metric and field names will be converted to the `metric_separator` value. 
+By default, the `convert_path` option is true, and `metric_separator` is `.` (dot). 
+
+
+### Use Regex
+Most illegal characters in the metric name are automatically converted to `-`.  
+The `use_regex` setting can be used to ensure all illegal characters are properly handled, but can lead to performance degradation.
+
+
+
+### Splunk Metric Data format
+The expected input for a Splunk Metric is:
+```
+{  "time":<timestamp>,
+   "event":"metric",
+   "source":"",
+   "host":"<host>",
+   "fields": {
+      "metric_name":"<your.metric.name>,
+      "_value":<your_metric_value>,
+      "<dimension1_name>":"<dimension1_value>",
+      "<dimension2_name>":"<dimension2_value>",
+      ...,
+      "<dimensionN_name>":"<dimensionN_value>"
+   }
+}
+```
+More information about the Splunk Metric data format is available [here](https://docs.splunk.com/Documentation/Splunk/latest/Metrics/Overview)
+
--- a/plugins/outputs/splunk/README.md
+++ b/plugins/outputs/splunk/README.md
@@ -26,14 +26,19 @@ Notes:
   #prefix = "my.specific.prefix."
 
   ## OPTIONAL:  whether to use "value" for name of simple fields
+  ##  Default is false which will result in using only the measurement name as the metric name, not "value"
   #simple_fields = false
 
   ## OPTIONAL:  character to use between metric and field name.  defaults to . (dot)
   #metric_separator = "."
 
   ## OPTIONAL:  Convert metric name paths to use metricSeperator character
-  ## When true (default) will convert all _ (underscore) chartacters in final metric name
-  #convert_paths = true
+  ## When true will convert all _ (underscore) chartacters in final metric name
+  #convert_paths = false
+
+  ## OPTIONAL:  Replace special characters in metric names with "-".  
+  ## This can be useful if metric names contain special characters  
+  #replace_special_chars = false
 
   ## OPTIONAL:  Use Regex to sanitize metric and tag names from invalid characters
   ## Regex is more thorough, but significantly slower
@@ -46,12 +51,12 @@ Notes:
 
 ### Convert Path & Metric Separator
 If the `convert_path` option is true any `_` in metric and field names will be converted to the `metric_separator` value. 
-By default, the `convert_path` option is true, and `metric_separator` is `.` (dot). 
+By default, the `convert_path` option is false and `metric_separator` is `.` (dot). 
 
 
-### Use Regex
-Most illegal characters in the metric name are automatically converted to `-`.  
-The `use_regex` setting can be used to ensure all illegal characters are properly handled, but can lead to performance degradation.
+### Replace Special Characters & Use Regex
+Special characters in the metric name can be automatically converted to `-` by setting replace_special_chars to true.  This is false by default.  
+The `use_regex` setting can be used to ensure all special characters are properly handled, but can lead to performance degradation.
 
 
 
--- a/plugins/outputs/splunk/splunk.go
+++ b/plugins/outputs/splunk/splunk.go
@@ -33,12 +33,13 @@ type Splunk struct {
 	StringToNumber  map[string][]map[string]float64
 	serializer      serializers.Serializer
 }
+
 // Descriptions of the flags in Splunk struct
 /*
 SimpleFields    - boolean to determine whether or not to use xxx.value as the metric name (true) or ommit the .value from the metric name (false)
 MetricSeparator - character to use between metric and field name.  defaults to . (dot)
 ConvertPaths    - boolean to convert all _ (underscore) chartacters in metric name to MetricSeparator
-ConvertBool     - boolean to convert all true/false values to 1/0 
+ConvertBool     - boolean to convert all true/false values to 1/0
 ReplaceSpecials - boolean to sanitize special characters in metric names with "-"
 UseRegex        - boolean to use Regex to sanitize metric and tag names from invalid characters
 StringToNumber  - map used internally to convert string values to numerics
--- a/plugins/outputs/splunk/splunk.go
+++ b/plugins/outputs/splunk/splunk.go
@@ -95,6 +95,7 @@ AuthString = "XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX"
 
 ## OPTIONAL:  whether to convert boolean values to numeric values, with false -> 0.0 and true -> 1.0.  default true
 #convert_bool = true
+
 `
 
 type SplunkMetric struct {
--- a/plugins/outputs/splunk/splunk.go
+++ b/plugins/outputs/splunk/splunk.go
@@ -84,7 +84,6 @@ AuthString = "XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX"
 
 ## OPTIONAL:  whether to convert boolean values to numeric values, with false -> 0.0 and true -> 1.0.  default true
 #convert_bool = true
-
 `
 
 type SplunkMetric struct {
--- a/plugins/outputs/splunk/splunk.go
+++ b/plugins/outputs/splunk/splunk.go
@@ -28,6 +28,7 @@ type Splunk struct {
 	MetricSeparator string
 	ConvertPaths    bool
 	ConvertBool     bool
+	ReplaceSpecials bool
 	UseRegex        bool
 	StringToNumber  map[string][]map[string]float64
 	serializer      serializers.Serializer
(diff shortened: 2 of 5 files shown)


plugins/outputs/splunk/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+  ## REQUIRED
+  ## Splunk Authorization Token for sending data to a Splunk HTTPEventCollector (HEC).
+  ##   Note:  This Token should map to a 'metrics' index in Splunk.  
+  AuthString = "XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX"
+
+  ## OPTIONAL:  prefix for metrics keys
+  #prefix = "my.specific.prefix."
+
+  ## OPTIONAL:  whether to use "value" for name of simple fields
```
phemmer: What is a "simple field"?

Looking at the code, it seems that what this really does is that if not enabled, and the field name is "value", that the metric name in splunk is just the name of the measurement.
I think this could be made more clear as it took me a while to figure out what the feature was really doing.

plugins/outputs/splunk/splunk.go
```diff
@@ -0,0 +1,277 @@
...
+	AuthString string
+	Timeout    internal.Duration
+	client     *http.Client
+
+	SimpleFields    bool
+	MetricSeparator string
+	ConvertPaths    bool
+	ConvertBool     bool
+	UseRegex        bool
+	StringToNumber  map[string][]map[string]float64
```
phemmer: There's no documentation on this parameter, or what it's supposed to be doing.

plugins/outputs/splunk/splunk.go
```diff
@@ -0,0 +1,277 @@
...
+				}
+			}
+		}
+		return 0, fmt.Errorf("unexpected type: %T, with value: %v, for: %s", v, v, name)
+	default:
+		return 0, fmt.Errorf("unexpected type: %T, with value: %v, for: %s", v, v, name)
+	}
+
+	return 0, fmt.Errorf("unexpected type: %T, with value: %v, for: %s", v, v, name)
+}
```
phemmer: There's lots of other numeric types being missed: `int`, `uint`, `uint8`, `uint16`, `uint32`, `int8`, `int16`, `int32`, `float32`.

plugins/outputs/splunk/splunk.go
```diff
@@ -0,0 +1,277 @@
...
+	case bool:
+		if s.ConvertBool {
+			if p {
+				return 1, nil
+			} else {
+				return 0, nil
+			}
+		}
+	case int64:
+		return float64(v.(int64)), nil
```
phemmer: You don't need the type assertion if you use `p`.

E.g.:
```
return float64(p), nil
```

Ditto for the `uint64` and `float64` cases below.

plugins/outputs/splunk/splunk.go
```diff
@@ -0,0 +1,277 @@
...
+		name = fmt.Sprintf("%s%s", s.Prefix, measureName)
+	} else {
+		name = fmt.Sprintf("%s%s%s%s", s.Prefix, measureName, s.MetricSeparator, fieldName)
+	}
+
+	if s.UseRegex {
+		name = sanitizedRegex.ReplaceAllLiteralString(name, "-")
+	} else {
+		name = sanitizedChars.Replace(name)
+	}
```
phemmer: I don't think we should be forcing the user to have scrubbed field names. Defaulting to enabled might be fine, but it should be possible to turn it off entirely.
Splunk doesn't require that the name be within a certain character set. Having some characters might make it harder to work with, but it's not impossible. And people might not want their names being messed with.

plugins/outputs/splunk/README.md
```diff
@@ -0,0 +1,76 @@
...
+  #simple_fields = false
+
+  ## OPTIONAL:  character to use between metric and field name.  defaults to . (dot)
+  #metric_separator = "."
+
+  ## OPTIONAL:  Convert metric name paths to use metricSeperator character
+  ## When true (default) will convert all _ (underscore) chartacters in final metric name
+  #convert_paths = true
```
phemmer: This doesn't seem like a good idea, at least not as a default. In splunk, the . is supposed to be for namespacing. Lots of telegraf metrics use underscore in field names, and not as namespace boundaries.

plugins/outputs/splunk/splunk.go
```diff
@@ -0,0 +1,277 @@
...
+			}
+			fields["_value"] = metricValue
+
+			// -----------------------------------------------------------------------------------------------------------------
+			//  Create a []byte array to send via an HTTP POST
+			// -----------------------------------------------------------------------------------------------------------------
+			var payload []byte
+			var err error
+
+			splunkMetric.Time = m.Time().UnixNano() / 1000000000 // convert metric nanoseconds to unix time
```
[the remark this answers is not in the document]
tmartin14: updated to use the default precision of UnixNano().

plugins/outputs/splunk/splunk.go
```diff
@@ -0,0 +1,277 @@
...
+			if err != nil {
+				return fmt.Errorf("unable to create http.Request \n    URL:%s\n\n", s.SplunkUrl)
+			}
+
+			req.Header.Add("Content-Type", "application/json")
+			req.Header.Add("Authorization", "Splunk "+s.AuthString)
+
+			//// Check for existence of s && req to prevent race ('panic: runtime error: invalid memory address or nil pointer dereference')
+			if s != nil && req != nil {
+				resp, err := s.client.Do(req)
```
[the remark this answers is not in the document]
tmartin14: implemented batch sending of metrics.

plugins/outputs/splunk/splunk.go
```diff
@@ -0,0 +1,277 @@
...
+				if err != nil {
+					return fmt.Errorf("error POST-ing metrics to Splunk[%s]  Sending Data:%s\n", err, payload)
+				}
+
+				defer resp.Body.Close()
```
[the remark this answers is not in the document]
tmartin14: implemented batch sending of metrics.

plugins/outputs/splunk/splunk.go
```diff
@@ -0,0 +1,277 @@
...
+				splunkMetric.Host = value
+			} else {
+				newTags[tagName] = value
+			}
+		}
+
+		// -----------------------------------------------------------------------------------------------------------------------
+		//  Loop through each metric and create a map of all tags + the name and value of the metric  (Splunk Metric format)
+		// -----------------------------------------------------------------------------------------------------------------------
+		fields := newTags
```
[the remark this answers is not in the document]
tmartin14: fixed.

plugins/outputs/splunk/splunk.go
```diff
@@ -0,0 +1,277 @@
...
+			req, err := http.NewRequest("POST", s.SplunkUrl, bytes.NewBuffer(payload))
+			if err != nil {
+				return fmt.Errorf("unable to create http.Request \n    URL:%s\n\n", s.SplunkUrl)
+			}
+
+			req.Header.Add("Content-Type", "application/json")
+			req.Header.Add("Authorization", "Splunk "+s.AuthString)
+
+			//// Check for existence of s && req to prevent race ('panic: runtime error: invalid memory address or nil pointer dereference')
+			if s != nil && req != nil {
```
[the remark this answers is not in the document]
tmartin14: corrected.

plugins/outputs/splunk/splunk.go
```diff
@@ -0,0 +1,277 @@
...
+			}
+		}
+	}
+	return nil
+}
+
+// -----------------------------------------------------------------------------------------------------------------------
+//  Check for metric name ending with .value and remove it, sanitize characters and add separators
+// -----------------------------------------------------------------------------------------------------------------------
+func processFieldName(measureName string, fieldName string, s *Splunk) string {
```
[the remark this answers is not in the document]
tmartin14: no particular reason ;).   Left as is.

plugins/outputs/splunk/splunk.go
```diff
@@ -0,0 +1,277 @@
...
+	var name string
+	if !s.SimpleFields && fieldName == "value" {
+		name = fmt.Sprintf("%s%s", s.Prefix, measureName)
+	} else {
+		name = fmt.Sprintf("%s%s%s%s", s.Prefix, measureName, s.MetricSeparator, fieldName)
```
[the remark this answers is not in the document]
tmartin14: corrected.   Thanks for the tip!

plugins/outputs/splunk/README.md
tmartin14: Added comments in the code as well as the readme.

plugins/outputs/splunk/splunk.go
tmartin14: added comments throughout.

plugins/outputs/splunk/splunk.go
tmartin14: added the additional types.

plugins/outputs/splunk/splunk.go
tmartin14: left the type assertion, but created new cases for multiple types.

plugins/outputs/splunk/splunk.go
tmartin14: modified the default to NOT change any names and added a new setting to let users choose if they want to replace special characters.

plugins/outputs/splunk/README.md
tmartin14: Updated the default to false, but the option is there if a user wants to convert "_" to "." (or any other metric separator).

tmartin14:
circleci tests fail with code formatting error and recommends running 'make fmt' to resolve the issue.   

I ran make fmt, but there's nothing changed in the resulting source file (splunk.go).   

Any suggestions?

tmartin14:
@phemmer - I'm getting failures on the code formatting, but 'make fmt' does not make any changes to the code.     Can you help?

danielnelson:
Maybe try running it directly:
```
gofmt -w plugins/outputs/splunk/splunk.go
```

plugins/outputs/splunk/splunk.go
```diff
@@ -0,0 +1,298 @@
...
+	case bool:
+		if s.ConvertBool {
+			if p {
+				return 1, nil
+			} else {
+				return 0, nil
+			}
+		}
+	case int, int8, int16, int32, int64:
+    	return float64(v.(int64)), nil
```
phemmer: Unfortunately this will not work. As an example, in the case where `v` is an `int8`, the type assertion `v.(int64)` will panic.
Type switches have to be repetitive. One `case` for each type.

```
switch p := v.(type) {
case bool:
  ...
case int:
  return float64(p), nil
case int8:
  return float64(p), nil
case int16:
  return float64(p), nil
case int32:
  return float64(p), nil
...
}
```
Even though each `case` statement contains the same code, they have to be separate because `p` is a different type, and thus when compiled, they aren't actually the same.

plugins/outputs/splunk/splunk.go
```diff
@@ -0,0 +1,298 @@
...
+	Timeout    internal.Duration
+	client     *http.Client
+
+	SimpleFields    bool
+	MetricSeparator string
+	ConvertPaths    bool
+	ConvertBool     bool
+	ReplaceSpecials bool
+	UseRegex        bool
+	StringToNumber  map[string][]map[string]float64
```
phemmer: The response to the 'lack of documentation' comment on the previous review said documentation had been added on this parameter. But I'm still not seeing it. Maybe it was accidentally left out of the update?

plugins/outputs/splunk/README.md
```diff
@@ -0,0 +1,81 @@
...
+  ## When true will convert all _ (underscore) chartacters in final metric name
+  #convert_paths = false
+
+  ## OPTIONAL:  Replace special characters in metric names with "-".  
+  ## This can be useful if metric names contain special characters  
+  #replace_special_chars = false
+
+  ## OPTIONAL:  Use Regex to sanitize metric and tag names from invalid characters
+  ## Regex is more thorough, but significantly slower
+  #use_regex = false
```
phemmer: `replace_special_chars` and `use_regex` combined cover all the possible cases, but maybe a single parameter would be better.
E.G.
```
name_sanatize = "none"
name_sanatize = "regex"
name_sanatize = "transliterate"
```
Note, this is not a requirement, or even a suggestion on the name, just throwing out ideas. @danielnelson might have a stronger preference.

plugins/outputs/splunk/splunk.go
```diff
@@ -0,0 +1,298 @@
...
+
+
+	// -----------------------------------------------------------------------------------------------------------------
+	//  Now we can send this set of metrics to Splunk
+	//
+	//  Create a []byte array to send via an HTTP POST
+	// -----------------------------------------------------------------------------------------------------------------
+	var payload []byte
+	var err error
+	payload, err = json.Marshal(allMetrics)
```
phemmer: Does this work? I haven't tested with an array, but the splunk [http event collector doc](http://dev.splunk.com/view/event-collector/SP-CAAAE6P) says it wants JSON streaming format:

> Here is an example of batched data. The batch protocol for HTTP Event Collector is simply event objects stacked one after the other as shown here, and not in a JSON array. Note that these events, though they only contain the "event" and "time" keys, are still valid:
```
{
  "event":"event 1", 
  "time": 1447828325
}

{
  "event":"event 2", 
  "time": 1447828326
}
```

----

To get this format in GO, you need to use a [`json.Encoder`](https://golang.org/pkg/encoding/json/#Encoder). E.G.:

```
buf := bytes.NewBuffer(nil)
enc := json.NewEncoder(buf)
for _, m := range measures {
  ...
  enc.Encode(splunkMetric)
}
http.NewRequest("POST", s.SplunkUrl, buf)
```

tmartin14: This data will be placed into a Splunk metrics index as opposed to an event index.  The format of a metric is still JSON, but different from event data.    The byte array does indeed work ;-).

plugins/outputs/splunk/splunk.go
tmartin14: fixed.

plugins/outputs/splunk/splunk.go
tmartin14: The documentation was added to the readme and the notes in the config.   I've add comments to the code as well.

plugins/outputs/splunk/README.md
tmartin14: by keeping them separate the user will have the option to use both regex and the predefined character set.  I'm going to leave this one as is.

plugins/outputs/splunk/splunk.go
danielnelson: Fields can only be int64, uint64, float64, string, and bool.  You can only support these types, but also include a default case that logs an error.

tmartin14: I've converted all numerics to float64 in the latest commit.   String and bool are also handled.  

Is this acceptable?

phemmer: Hrm. That seems like a restriction that should be removed. The type is a limitation of the storage engine. I don't think it makes sense to impose InfluxDB's limitations on the other outputs.

danielnelson: I'm open to allowing additional types, but I don't want fields to be anything and everything.

tmartin14: I'm confused.   This is an output plugin that's doing the conversion.    If I'm missing something I apologize, but if there really is an issue please let me know how best to resolve it in this code.

danielnelson: @tmartin14 On this output, only worry about fields being one of int64, uint64, float64, string, and bool.  Anything else just log an error and ignore it.

Regarding field types in general, I don't want each output plugin to have to anticipate a huge list of possible types, and I want the types that a field can be to be well defined.  So, I don't really want the field to be an `interface{}`, I want a variant type.  I would also like at some point to return an error if the input adds other types.  I think these steps will reduce the number of bugs around type conversion, and simplify output writing.  We can expand this list of types if needed (I've been considering adding a time type for instance).

tmartin14: Thanks @danielnelson.   I'm updating this back to the original types and will re-commit.

danielnelson:
@tmartin14 I think we are going to go with the serializer implementation in #4339 instead of this output, if you haven't already could you double check that the serializer meets your requirements?

I would still be open to a Splunk output plugin, I think it would simplify using the serializer and we might be able to do it without writing any new code by creating a specially configured http output in the init function.

tmartin14:
Thanks for the update @danielnelson.  I am in favor of closing this PR in favor of #4339 as the serializer approach adds support for both Splunk metric and event index types without the need for a separate output.

PR #4190, opened 2018-05-23
Add special syslog timestamp parser that uses current year

### LINKED ISSUE #3379: Logparser : Epoch convertion impossible with SYSLOGTIMESTAMP Grok pattern
### Relevant telegraf.conf:

[[inputs.logparser]]
files = ["/root/samples/*mail.log"]
from_beginning = true

[inputs.logparser.grok]
patterns = ["%{SYSLOGTIMESTAMP:timestamp:ts-“Mon Jan 2 15:04:05”} %{SYSLOGHOST:hostname} mta-in/milter\[%{MILTER_PROCESSID}\]: %{POSTFIX_QUEUEID:postfix_queue_id}: from=<%{DATA:postfix_from}>, firstto=<%{DATA:postfix_firstto}>, nrcpt=%{DATA:postfix_nrcpt}, size=%{DATA:postfix_msg_size}, score=%{DATA:score}, state=%{DATA:state:int}, status=%{DATA:status:tag}, level=%{DATA:level}, actions=%{DATA:actions}, subject="%{DATA:subject}", cause="%{DATA:cause}""]
## Name of the outputted measurement name.
measurement = “test”
…
custom_patterns = ‘’‘
POSTFIX_QUEUEID ([0-9A-F]{6,}|[0-9a-zA-Z]{15,})
MILTER_PROCESSID ([0-9]{5,})
’’’

### System info:

- Telegraf v1.4.2 (git: release-1.4 0cc5fc0ce40b09467012fffddcea6985b2763a97)
- InfluxDB 1.3.6-1
- Docker official image

### Steps to reproduce:

1. Deploy Telegraf using docker with the 
2. Inject an existing log in SYSLOGTIMESTAMP format
For ex. Sep 25 08:27:29 mailhost postfix/cleanup[17659]: ABFF15C..
3. The time is not correctly calculated.

### Expected behavior:

The syslog_timestamp should be converted into time in Influxdb

### Actual behavior:

As the year is missing, As Telegraf can't find the year he set it to 0 which generate a wrong time in influxdb

### Additional info:

## Feature Request

### Proposal:

Like the timezone option we should be able to set manually the year.
Assuming automatically the year could cause some trouble when we change of it.
if a person has more than 2-3 years of logs to inject into influxdb he should pass this option in each logparser.

### Current behavior:

Telegraf assume automatically 0 has year if it's not provided

### Desired behavior:

By setting manually the date, Telegraf will be able to build the correct time in influxdb.

This is a standard output for the syslog, this pattern should be recognize per default as many solution use it.

### Forum Discussion
https://community.influxdata.com/t/custom-timestamp-for-syslogtimestamp-and-telegraf/2798/

danielnelson:
Previously it was impossible to parse syslog timestamps without the date being reported as year 0, due to the year not being specified.

This pull request adds the special timestamp format: `ts-syslog` that will set the year after parsing to the current year.  `%{SYSLOGTIMESTAMP:timestamp:ts-syslog}`

closes  #3379

### Required for all PRs:

### DIFF
--- a/plugins/inputs/logparser/README.md
+++ b/plugins/inputs/logparser/README.md
@@ -104,6 +104,7 @@ You must capture at least one field per line.
   - ts-httpd         ("02/Jan/2006:15:04:05 -0700")
   - ts-epoch         (seconds since unix epoch, may contain decimal)
   - ts-epochnano     (nanoseconds since unix epoch)
+  - ts-syslog        ("Jan 02 15:04:05", parsed time is set to the current year)
   - ts-"CUSTOM"
 
 CUSTOM time layouts must be within quotes and be the representation of the
--- a/plugins/inputs/logparser/grok/grok.go
+++ b/plugins/inputs/logparser/grok/grok.go
@@ -32,6 +32,7 @@ var timeLayouts = map[string]string{
 	// will get handled in the ParseLine function.
 	"ts-epoch":     "EPOCH",
 	"ts-epochnano": "EPOCH_NANO",
+	"ts-syslog":    "SYSLOG_TIMESTAMP",
 	"ts":           "GENERIC_TIMESTAMP", // try parsing all known timestamp layouts.
 }
 
@@ -44,6 +45,7 @@ const (
 	DROP              = "drop"
 	EPOCH             = "EPOCH"
 	EPOCH_NANO        = "EPOCH_NANO"
+	SYSLOG_TIMESTAMP  = "SYSLOG_TIMESTAMP"
 	GENERIC_TIMESTAMP = "GENERIC_TIMESTAMP"
 )
 
@@ -112,6 +114,7 @@ type Parser struct {
 	// layouts.
 	foundTsLayouts []string
 
+	timeFunc func() time.Time
 	g        *grok.Grok
 	tsModder *tsModder
 }
@@ -172,6 +175,10 @@ func (p *Parser) Compile() error {
 	if err != nil {
 		log.Printf("W! improper timezone supplied (%s), setting loc to UTC", p.Timezone)
 		p.loc, _ = time.LoadLocation("UTC")
+	}
+
+	if p.timeFunc == nil {
+		p.timeFunc = time.Now
 	}
 
 	return p.compileCustomPatterns()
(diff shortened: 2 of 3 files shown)


plugins/inputs/logparser/README.md
```diff
@@ -104,6 +104,7 @@ You must capture at least one field per line.
   - ts-httpd         ("02/Jan/2006:15:04:05 -0700")
   - ts-epoch         (seconds since unix epoch, may contain decimal)
   - ts-epochnano     (nanoseconds since unix epoch)
+  - ts-syslog        ("Jan 02 15:04:05", parsed time is set to the current year)
```
leodido: @danielnelson this is the BSD syslog timestamp format (RFC3164) not the syslog RFC5424 timestamp (basically a RFC3339 micro timestamp).
So probably you want to call it `ts-bsd-syslog`, and maybe also add another format for `ts-syslog`.
Just my 2 cents.

leodido: The timestamp depends on the format (so RFC) chosen.
For example you can configure `RSYSLOG` to use `RSYSLOG_SyslogProtocol23Format` format, which means RFC5424. With this setting `/var/log/*` will contains `RFC3339MICRO` timestamps (as per RFC5424). Without this setting it will default to the old (RFC3164) BSD syslog (and timestamp) format.

leodido: Regarding localization. RFC5425 with `RFC3339MICRO` timestamps is not affected.
Rather, timestamps of syslog messages following the old RFC3164 are not localized (must be always in english) ([ref](https://tools.ietf.org/html/rfc3164#section-4.1.2)).

danielnelson: I think I'll leave this as `ts-syslog` since it pairs with the built in grok pattern `SYSLOGTIMESTAMP` and there isn't any reason to add a special pattern for rfc5424 style timestamps since we already have `ts-rfc3339` and `ts-rfc3339nano`.  We only need a special type here to inject the year.

plugins/inputs/logparser/grok/grok.go
```diff
@@ -285,6 +292,16 @@ func (p *Parser) ParseLine(line string) (telegraf.Metric, error) {
 			} else {
 				timestamp = time.Unix(0, iv)
 			}
+		case SYSLOG_TIMESTAMP:
+			ts, err := time.ParseInLocation("Jan 02 15:04:05", v, p.loc)
```
mamafun: Hi @danielnelson 
Should we use "Jan 2 15:04:05" instead of "Jan 02 15:04:05" ?

danielnelson: I think maybe it should be `Jan _2 15:04:05`, but I can't seem to create a testcase that fails.  Do you have a example that is failing?

mamafun: If we try to parse "Sep  2 09:01:55 value=42" in test function TestSyslogTimestampParser(),
the test will show the following:
`=== RUN   TestSyslogTimestampParser
2018/06/22 09:36:41 E! Error parsing Sep 2 09:01:55 to time layout [SYSLOG_TIMESTAMP]: parsing time "Sep 2 09:01:55" as "Jan 02 15:04:05": cannot parse "2 09:01:55" as "02"`

Does this the example count?
Thanks.

danielnelson: Ahh, the test was "passing" even though the parsing failed.  Here is the fix: https://github.com/influxdata/telegraf/pull/4334

mamafun: That fix looks nice. Thank you.

PR #4201, opened 2018-05-26
Add detect_leader config option to docker input plugin

palp:
This option causes Swarm data (gather_services) to only be collected if the current node is the leader. By doing so I can use the same config on all nodes in the Swarm.

It might make sense to make this default to true, but I didn't want to introduce any changes to existing behavior on my own.

### Required for all PRs:


palp:
Pushed the wrong thing here by mistake, will reopen.

PR #4202, opened 2018-05-26
Add detect_leader config option to docker input plugin

palp:
This option causes Swarm data (gather_services) to only be collected if the current node is the leader. By doing so I can use the same config on all nodes in the Swarm.

It might make sense to make this default to true, but I didn't want to introduce any changes to existing behavior on my own.

### Required for all PRs:

### DIFF
--- a/plugins/inputs/docker/docker.go
+++ b/plugins/inputs/docker/docker.go
@@ -58,7 +58,7 @@ type Docker struct {
 	labelFilter     filter.Filter
 	containerFilter filter.Filter
 	stateFilter     filter.Filter
-	isLeader	bool
+	isLeader        bool
 }
 
 // KB, MB, GB, TB, PB...human friendly
@@ -182,7 +182,7 @@ func (d *Docker) Gather(acc telegraf.Accumulator) error {
 			err := d.gatherSwarmInfo(acc)
 			if err != nil {
 				acc.AddError(err)
-			}				
+			}
 		}
 	}
 
--- a/plugins/inputs/docker/docker.go
+++ b/plugins/inputs/docker/docker.go
@@ -240,14 +240,19 @@ func (d *Docker) gatherSwarmInfo(acc telegraf.Accumulator) error {
 		return err
 	}
 
+	isLeader := false
 	activeNodes := make(map[string]struct{})
 	for _, n := range nodes {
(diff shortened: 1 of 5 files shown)


plugins/inputs/docker/docker.go
```diff
@@ -82,7 +84,12 @@ var sampleConfig = `
   endpoint = "unix:///var/run/docker.sock"
 
   ## Set to true to collect Swarm metrics(desired_replicas, running_replicas)
+  ## Note: configure this in one of the manager nodes in a Swarm cluster.
+  ## configuring in multiple Swarm managers results in duplication of metrics.
   gather_services = false
+  ## If gather_services is true, only collect if this node is the leader.
+  ## This allows the same configuration to be used on all nodes in a Swarm.
+  detect_leader = false
```
danielnelson: I think we should turn these bool pairs into a tri-bool type setting:
```
  ## Service mode sets the condition for collecting Swarm data.  Can be one of
  ## "leader", "always", or "never".
  swarm_reporting = "leader"
```

Leave the old option as is, but if the new option is not set translate it to "always" or "never".  When the new option is set ignore the old value.

plugins/inputs/docker/docker.go
```diff
@@ -303,6 +324,10 @@ func (d *Docker) gatherInfo(acc telegraf.Accumulator) error {
 	d.engine_host = info.Name
 	d.serverVersion = info.ServerVersion
 
+	if d.SwarmReporting == "leader" {
+		d.nodeID = info.Swarm.NodeID
```
danielnelson: I think we should return this ID, and then we can pass the `leaderID` back into `gatherSwarmInfo`.  This can be done regardless of the setting but double check that it works if there is no swarm leader, I worry that info.Swarm may be unset in this case.  This will help ensure that these operations are done in the correct order.

palp: Good idea - I also figured it wouldn't be populated, which is why I put the check in there, but it's probably better to know what to expect and handle it appropriately.

sjwang90:
Hey @palp, thanks for your contribution. Checking in as it's been awhile. If you'd like to continue working on this, I think you still haven't committed updates from the last comment.

palp:
Although ironically I had my modification still running until recently, I don't have an environment set up to test this right now. But I'll see if I can get one going soon and finish it up.

powersj:
Hi,

Thanks!

PR #4204, opened 2018-05-27
[unbound] add option to reset stats each time they are collected

mcfedr:
Unbound stats normally collected cumulatively by the plugin, but for some setups its more useful to collect stats for the time period. This change adds option to change the command used.

### Required for all PRs:

### DIFF
--- a/plugins/inputs/unbound/README.md
+++ b/plugins/inputs/unbound/README.md
@@ -27,6 +27,9 @@ a validating, recursive, and caching DNS resolver.
   ## true in a future version.  It is recommended to set to true on new
   ## deployments.
   thread_as_tag = false
+
+  ## When set to true metrics are reset each time they are collected
+  reset_stats = false
 ```
 
 #### Permissions:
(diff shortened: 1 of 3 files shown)


mcfedr:
Not sure I follow the reasoning, most plugins work where each point collected is independent. This plugin is currently unique in that it collects the cumulative stats from unbound. This makes it next to useless, each restart of unbound resets all the stats, and I cannot see any long term trends.

danielnelson:
There are quite a few plugins that are cumulative, such as cpu and net.  You can use the non_negative_derivative function to view the differences.  There is a bit of prior discussion on https://github.com/influxdata/telegraf/pull/3745.

mcfedr:
hmm, thats really interesting, didnt know about that function. Thanks!

voiprodrigo:
non_negative_difference is also useful if you don’t need a rate. I use both all the time with cumulative counters. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #4205, opened 2018-05-27
Added Generic SQL output plugin

illuusio:
*First small word of warning. This is SQL output plugin is far from perfect. It tries to works automatically with some plugins but with some it fails miserably.*

I took generic SQL input plugin and Postgres output plugin PR#3428. Then formulated them into generic SQL plugin. As I know Telegraf is made by InfluxData which makes first grade InfluxDB but there is corner cases when SQL based database is needed to store metrics. It's not fast it's more convenient.
Plugin is at very basic level as I though if it can be get in it should be extended some features in future which are missing (like transactions) and _again it's not fastest way to store metrics currently_!
I have tested it with Postgres and MariaDB but it should be so generic it could be used with most SQL databases.
As said above this is far from perfect and I really appreciate comments which are not about speed of SQL (I know there is several pull request for Postgres Output plugin which probably do faster) but what is missing and is this even needed output plugin.  

### Required for all PRs:

### DIFF
--- a/README.md
+++ b/README.md
@@ -296,6 +296,7 @@ For documentation on the latest development code see the [documentation index][d
 * [solr](./plugins/inputs/solr)
 * [sql server](./plugins/inputs/sqlserver) (microsoft)
 * [stackdriver](./plugins/inputs/stackdriver) (Google Cloud Monitoring)
+* [sql](./plugins/outputs/sql) (SQL generic output)
 * [statsd](./plugins/inputs/statsd)
 * [suricata](./plugins/inputs/suricata)
 * [swap](./plugins/inputs/swap)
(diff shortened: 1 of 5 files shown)


fgenoese:
Are you still working on this @illuusio ?

illuusio:
Yes I would if there is interest. I kind of settled to Fluentd but it's harder to maintain but supports SQL with plugin.

fgenoese:
Would it be feasible to write an [External Plugin](https://github.com/influxdata/telegraf/blob/master/EXTERNAL_PLUGINS.md)?

illuusio:
Maybe I have to investigate that.

illuusio:
If I understood correctly It works only for Input plugins with execd. It could be done with output Exec but needs again more thinking.

sjwang90:
`outputs.sql` released in Telegraf 1.19 - merged in #9280

Issue #4161, opened 2018-05-17
Kafka output plugin should support batching at the producer level


JimHagan:
## Support Batching at the Producer Level

This will make it possible to scale to higher degree, by making fewer calls and may allow our compression of output objects to be more efficient.

### Proposal:

Currently the `Write` method uses an individual call to sendMessage per metric.  This could be refactored to allow for a batch paramter passed to write (and injected down from the Kafka output plugin).  We could use `sendMessages` instead.

Suggestion...

```
func (k *Kafka, batch) Write(metrics []telegraf.Metric) error {
    if len(metrics) == 0 {
        return nil
    }

   counter:=0
   m_list:= []

    for _, metric := range metrics {
        buf, err := k.serializer.Serialize(metric)
        if err != nil {
            return err
        }

       counter += 1
       m_list.append(m)

        topicName := k.GetTopicName(metric)

        m := &sarama.ProducerMessage{
            Topic: topicName,
            Value: sarama.ByteEncoder(buf),
        }
        if h, ok := metric.Tags()[k.RoutingTag]; ok {
            m.Key = sarama.StringEncoder(h)
        }

        if counter%batch == 0 {
        _, _, err = k.producer.SendMessages(m_list)
         counter = 0;
         m_list = []
       }

        if err != nil {
            return fmt.Errorf("FAILED to send kafka message: %s\n", err)
        }
    }

    if counter > 0  {
        _, _, err = k.producer.SendMessages(m_list)
    }


    return nil
}
```

### Use case: [Why is this important (helps with prioritizing requests)]

We have a very large multi-data center Kafka pipeline and we need batching to lower the number of calls as well as to make Kafka compression possibly more efficient.

russorat:
see also https://github.com/influxdata/telegraf/issues/2825

JimHagan:
@russorat Is there code for this in 1.8 that we could give a look at? I was just talking with @edbernier

danielnelson:
@JimHagan I have two potential patches that I could use your help performance testing, if you are up for it.

I have them in separate branches: `kafka-perf-send-messages` and `kafka-perf-batch`.

The send-messages branch works by simply using the `SendMessages` method on `sarama.SyncProducer`, this sends all messages out in a `Write` batch and waits for them all to complete, instead of the previous method of send/recv, send/recv, etc.

[...]

PR #4210, opened 2018-05-30
Fix crush on non-English Windows (win_perf_counters input plugin ) + Extend Kafka output plugin

korservick:
Fix non English name of metrics on Windows (win_perf_counters input plugin) + extend kafka output plugin (add LZ4 commpression + batch sending)

### Required for all PRs:

### DIFF
--- a/plugins/inputs/win_perf_counters/performance_query.go
+++ b/plugins/inputs/win_perf_counters/performance_query.go
@@ -181,7 +181,7 @@ func UTF16ToStringArray(buf []uint16) []string {
 	stringLine := UTF16PtrToString(&buf[0])
 	for stringLine != "" {
 		strings = append(strings, stringLine)
-		nextLineStart += len(stringLine) + 1
+		nextLineStart += len([]rune(stringLine)) + 1
 		remainingBuf := buf[nextLineStart:]
 		stringLine = UTF16PtrToString(&remainingBuf[0])
 	}
--- a/plugins/outputs/kafka/kafka.go
+++ b/plugins/outputs/kafka/kafka.go
@@ -21,6 +21,8 @@ var ValidTopicSuffixMethods = []string{
 
 type (
 	Kafka struct {
+		// Kafka version
+		Version string
 		// Kafka brokers to send metrics to
 		Brokers []string
 		// Kafka topic
@@ -45,6 +47,8 @@ type (
 		CA string
 
 		tlsint.ClientConfig
+		// Batch messages (if output format support it)
+		BatchMessage bool `toml:"batch"`
 
 		// SASL Username
 		SASLUsername string `toml:"sasl_username"`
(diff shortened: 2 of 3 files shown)


korservick:
@danielnelson Hi! 
With batch sending (I use JSON output format with batching + LZ4 compression) this patch significally decrease CPU usage (2-5%->0-1%) + some memory usage reduction (40MB->30MB)

plugins/outputs/kafka/kafka.go
```diff
@@ -186,6 +197,11 @@ func (k *Kafka) Connect() error {
 	}
 	config := sarama.NewConfig()
 
+	kafkaVersion, err := sarama.ParseKafkaVersion(k.Version)
+	if err != nil {
+		return err
+	}
+	config.Version = kafkaVersion
```
danielnelson: Is this required for this patch to work, what happens if it is removed?  The documentation seems to indicate that this is used to control some features, what is improved by adding it?

korservick: If not set this parameter sarama lib return error if we switch to LZ4 compression. 
https://github.com/Shopify/sarama/blob/70f6a705d4a17af059acbc6946fb2bd30762acd7/config_test.go#L199
https://github.com/Shopify/sarama/blob/44e7121d3b5189096ae4ef90c442f5f806c10fc9/config.go#L416

In some places of sarama the 'Version' parameter used to switch to "new" Kafka API that supposed to be more perfomance than "old" one.

plugins/inputs/win_perf_counters/performance_query.go
```diff
@@ -181,7 +181,7 @@ func UTF16ToStringArray(buf []uint16) []string {
 	stringLine := UTF16PtrToString(&buf[0])
 	for stringLine != "" {
 		strings = append(strings, stringLine)
-		nextLineStart += len(stringLine) + 1
+		nextLineStart += len([]rune(stringLine)) + 1
```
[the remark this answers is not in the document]
korservick: Ok, no problem. Little bit later

plugins/outputs/kafka/kafka.go
```diff
@@ -64,85 +68,92 @@ type (
...
+
+## Optional SASL Config
+# sasl_username = "kafka"
+# sasl_password = "secret"
+
+## Data format to output.
+## Each data format has its own unique set of configuration options, read
+## more about them here:
+## https://github.com/influxdata/telegraf/blob/master/docs/DATA_FORMATS_OUTPUT.md
+# data_format = "influx"
```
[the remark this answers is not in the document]
korservick: Fixed

korservick:
Done, but CI fail

danielnelson:
closes #4161

danielnelson:
Looks like a bug in the syslog tests, I'll rerun the tests.

plugins/outputs/kafka/kafka.go
```diff
@@ -186,6 +202,11 @@ func (k *Kafka) Connect() error {
 	}
 	config := sarama.NewConfig()
 
+	kafkaVersion, err := sarama.ParseKafkaVersion(k.Version)
```
danielnelson: If version is set to an empty string I believe this will panic.

plugins/outputs/kafka/kafka.go
```diff
@@ -235,30 +256,69 @@ func (k *Kafka) Description() string {
...
+
 	if len(metrics) == 0 {
 		return nil
 	}
 
 	for _, metric := range metrics {
-		buf, err := k.serializer.Serialize(metric)
-		if err != nil {
-			return err
+		routingTag = metric.Tags()[k.RoutingTag]
```
danielnelson: So long as you are in this code, can you change this to use `metric.GetTag(k.RoutingTag)`, which doesn't allocate.

plugins/outputs/kafka/kafka.go
```diff
@@ -235,30 +256,69 @@ func (k *Kafka) Description() string {
...
+		if _, found := routingTags[routingTag]; !found {
+			metricsmap[routingTag] = make(map[string][]telegraf.Metric)
 		}
+		routingTags[routingTag] = true
```
danielnelson: I think you can remove the `routingTags` map, and just base the creation of the submap on if `_, found := routingTags[routingTag]; !found` is true.

plugins/outputs/kafka/kafka.go
```diff
@@ -235,30 +256,69 @@ func (k *Kafka) Description() string {
 }
 
 func (k *Kafka) Write(metrics []telegraf.Metric) error {
+	var (
+		metricsmap  = map[string]map[string][]telegraf.Metric{}
+		routingTags = map[string]bool{}
+		routingTag  string
```
danielnelson: Don't declare `routingTag` here, just create it on 270.

danielnelson:
@korservick Would you be able to sign the CLA too?

JimHagan:
Guys, one thing about #4161 is that it allows a configurable batch size.  While allowing the producer to flush the entire buffer per flush will be fine on most cases, there are certain extremely high volume use cases for which we may want to have a separate batch size parameter associated with this feature.

danielnelson:
There is a several flush Config options in sarama that seem useful, such as `Producer.Flush.Bytes` and `Producer.Flush.MaxMessages`.  We could expose these, but they won't mesh well if we are doing the batching ourselves instead of leaving it to the Producer.  Has anyone tried using these?

JimHagan:
@danielnelson Each kafka topic has a match message bytes config.  Let's say it's set at 2MEG.  one thing we'll want to make sure is that we have the granularity  of control to ensure our system isn't bumping up agains tthat with batching.  Is Sarama coded to respect that parameter?  Just curious.

danielnelson:
This option defaults to 1MB and currently we don't expose it as a configurable setting.  With the current method in master where each message currently contains a single metric, it is unlikely we would bump up against this, but we could allow it to be configured.  If we do end up adding batching then we definitely need to expose this setting.

PR #4254, opened 2018-06-08
Copy Varz type instead of importing from gnatsd

themartorana:
For discussion, mostly - wanted to use telegraf to monitor FreeBSD gnatsd server cluster, and found it was disabled due to C extensions that have to build with gnatsd.

The gnatsd package is imported for the sole purpose of using the `Varz` type to deserialize metric data, and because of C extension requirements in gnatsd, the collector has been disabled for FreeBSD. See commit: https://github.com/influxdata/telegraf/commit/1b0a4e49cdf30cd0ae4984253549c443520c11ca

This PR simply removes the import requirement by copying the Varz type definition into the plugin. I don't know if that violates the Apache 2 license (or places requirements on telegraf) and I'm not informed enough to make that judgement, but I thought I'd propose this solution for my own benefit.

### Required for all PRs:

### DIFF
--- a/plugins/inputs/nats/nats.go
+++ b/plugins/inputs/nats/nats.go
@@ -1,5 +1,3 @@
-// +build !freebsd
-
 package nats
 
 import (
@@ -14,8 +12,6 @@ import (
 	"github.com/influxdata/telegraf"
 	"github.com/influxdata/telegraf/internal"
 	"github.com/influxdata/telegraf/plugins/inputs"
-
-	gnatsd "github.com/nats-io/gnatsd/server"
 )
 
 type Nats struct {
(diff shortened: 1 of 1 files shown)


danielnelson:
It seems to me that the [problematic file in gnatsd](https://github.com/nats-io/gnatsd/blob/master/server/pse/pse_freebsd.go) could be replaced using non-CGO calls using `golang.org/x/sys/unix`, it would be similar [to this gopsutil code](https://github.com/shirou/gopsutil/blob/master/mem/mem_freebsd.go#L17).

sjwang90:
Closing due to inactivity. Feel free to re-open if you'd like to continue discussion.

PR #4259, opened 2018-06-09
Telegraph Docker Input Plugin : per docker running status support

prashanthjbabu:
Currently the docker input plugin does not provide the running status of the docker . It does provide health status of the docker but that is only retrieved if the docker application supports it . This enhancement extends the plugin to give out the running status of the docker which is taken from the Docker Container Inspect API.
### Required for all PRs:

### DIFF
--- a/plugins/inputs/docker/docker.go
+++ b/plugins/inputs/docker/docker.go
@@ -437,16 +437,16 @@ func (d *Docker) gatherContainer(
 	}
 	if info.State != nil {
 		statefields := map[string]interface{}{
-			"status":  info.State.Status,
-			"running": info.State.Running,
-			"paused": info.State.Paused,
+			"status":     info.State.Status,
+			"running":    info.State.Running,
+			"paused":     info.State.Paused,
 			"restarting": info.State.Restarting,
-			"oomkilled": info.State.OOMKilled,
-			"dead": info.State.Dead,
-			"pid": info.State.Pid,
-			"exitcode": info.State.ExitCode,
-			"error": info.State.Error,
-			"startedat": info.State.StartedAt,
+			"oomkilled":  info.State.OOMKilled,
+			"dead":       info.State.Dead,
+			"pid":        info.State.Pid,
+			"exitcode":   info.State.ExitCode,
+			"error":      info.State.Error,
+			"startedat":  info.State.StartedAt,
 			"finishedat": info.State.FinishedAt,
 		}
 		acc.AddFields("docker_container_status", statefields, tags, time.Now())
--- a/plugins/inputs/docker/docker.go
+++ b/plugins/inputs/docker/docker.go
@@ -438,14 +438,9 @@ func (d *Docker) gatherContainer(
 	if info.State != nil {
 		statefields := map[string]interface{}{
 			"status":     info.State.Status,
-			"running":    info.State.Running,
-			"paused":     info.State.Paused,
-			"restarting": info.State.Restarting,
 			"oomkilled":  info.State.OOMKilled,
-			"dead":       info.State.Dead,
 			"pid":        info.State.Pid,
 			"exitcode":   info.State.ExitCode,
-			"error":      info.State.Error,
 			"startedat":  info.State.StartedAt,
 			"finishedat": info.State.FinishedAt,
 		}
--- a/plugins/inputs/docker/docker.go
+++ b/plugins/inputs/docker/docker.go
@@ -436,13 +436,13 @@ func (d *Docker) gatherContainer(
 		}
 	}
 	if info.State != nil {
+		tags["container_status"] = info.State.Status
 		statefields := map[string]interface{}{
-			"status":     info.State.Status,
-			"oomkilled":  info.State.OOMKilled,
-			"pid":        info.State.Pid,
-			"exitcode":   info.State.ExitCode,
-			"startedat":  info.State.StartedAt,
-			"finishedat": info.State.FinishedAt,
+			"oomkilled":   info.State.OOMKilled,
+			"pid":         info.State.Pid,
+			"exitcode":    info.State.ExitCode,
+			"started_at":  info.State.StartedAt,
+			"finished_at": info.State.FinishedAt,
 		}
 		acc.AddFields("docker_container_status", statefields, tags, time.Now())
 	}
--- a/plugins/inputs/docker/docker.go
+++ b/plugins/inputs/docker/docker.go
@@ -438,11 +438,17 @@ func (d *Docker) gatherContainer(
 	if info.State != nil {
 		tags["container_status"] = info.State.Status
 		statefields := map[string]interface{}{
-			"oomkilled":   info.State.OOMKilled,
-			"pid":         info.State.Pid,
-			"exitcode":    info.State.ExitCode,
-			"started_at":  info.State.StartedAt,
-			"finished_at": info.State.FinishedAt,
+			"oomkilled": info.State.OOMKilled,
+			"pid":       info.State.Pid,
+			"exitcode":  info.State.ExitCode,
+		}
+		container_time, err := time.Parse(time.RFC3339, info.State.StartedAt)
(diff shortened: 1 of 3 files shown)


plugins/inputs/docker/docker.go
```diff
@@ -435,6 +435,22 @@ func (d *Docker) gatherContainer(
 			}
 		}
 	}
+	if info.State != nil {
+		statefields := map[string]interface{}{
+			"status":     info.State.Status,
```
danielnelson: I think we should have this value as a tag on all `docker_container_*` measurements.

plugins/inputs/docker/docker.go
```diff
@@ -435,6 +435,22 @@ func (d *Docker) gatherContainer(
...
+			"running":    info.State.Running,
+			"paused":     info.State.Paused,
+			"restarting": info.State.Restarting,
```
danielnelson: I don't think we need to save the running, paused, restarting, dead values, since this is all encompassed by the status field.

plugins/inputs/docker/docker.go
```diff
@@ -435,6 +435,22 @@ func (d *Docker) gatherContainer(
...
+			"oomkilled":  info.State.OOMKilled,
+			"dead":       info.State.Dead,
+			"pid":        info.State.Pid,
+			"exitcode":   info.State.ExitCode,
+			"error":      info.State.Error,
+			"startedat":  info.State.StartedAt,
+			"finishedat": info.State.FinishedAt,
```
danielnelson: I'm unsure if error, startedat, finishedat are interesting enough to include, I feel like many people will want to exclude them.

plugins/inputs/docker/docker.go
prashanthjbabu: Wouldn't this be too much of duplication since the same information is across multiple measurements?

plugins/inputs/docker/docker.go
prashanthjbabu: Fair point . I will remove them....

plugins/inputs/docker/docker.go
prashanthjbabu: I feel startedat,finishedat is important since it gives useful information on the uptime of the container or even how long ago the container died(based on "status") . If you still think its not needed , we can take it off!

prashanthjbabu: Regarding error . I will remove it..

plugins/inputs/docker/docker.go
danielnelson: It is a little bit of duplication, but it seems useful to be able to "group by" status on any of the container measurements.

plugins/inputs/docker/docker.go
danielnelson: Okay, lets just rename them `started_at` and `finished_at` and convert them to unixnano format (unix epoch timestamp in nanosecond precision).  This is the format we have been using recently for encoding timestamps into a field.

plugins/inputs/docker/docker.go
prashanthjbabu: I went through the code to add status for each container . It can get tricky for cases like CPU(where there is a per-cpu option) and Network(where there is a per interface option). Also since these collectors are all over the place , we would have to make a separate "docker inspect" API call at each of these places . It seems a bit inefficient considering how it currently is designed . Would you still like me to do it that way?
   To handle the "group by" feature you requested , we could plug in "container_id" into this which allows you to match with it.

danielnelson: No, not if we need to make extra API calls.  Let's just add it the new measurement as a tag and we can try to fish it through later on if needed.

prashanthjbabu: I've fixed this one ! I've added it as a tag.. So now it comes not only for docker_container_status , but also for cpu,mem etc

Attached some logs :

docker_container_status,container_image=ubuntu,container_name=vibrant_benz,container_status=running,container_version=unknown,deviceid=mydeviceid,engine_host=raspberrypi,server_version=18.05.0-ce started_at="2018-06-14T05:48:53.266176036Z",finished_at="0001-01-01T00:00:00Z",oomkilled=false,pid=5278i,exitcode=0i 1528956922000000000
docker_container_mem,container_image=ubuntu,container_name=vibrant_benz,container_status=running,container_version=unknown,deviceid=mydeviceid,engine_host=raspberrypi,server_version=18.05.0-ce limit=0i,usage=0i,max_usage=0i,usage_percent=0,container_id="f85ef98b08bde3ad6a93fadcacfa33a57a1f833119cce0309cadb15c45df5cf4" 1528956921000000000
docker_container_cpu,container_image=ubuntu,container_name=vibrant_benz,container_status=running,container_version=unknown,cpu=cpu-total,deviceid=mydeviceid,engine_host=raspberrypi,server_version=18.05.0-ce usage_total=102424740i,usage_in_usermode=60000000i,usage_system=10927430000000i,throttling_periods=0i,throttling_throttled_time=0i,usage_in_kernelmode=20000000i,throttling_throttled_periods=0i,container_id="f85ef98b08bde3ad6a93fadcacfa33a57a1f833119cce0309cadb15c45df5cf4",usage_percent=0 1528956921000000000
docker_container_cpu,container_image=ubuntu,container_name=vibrant_benz,container_status=running,container_version=unknown,cpu=cpu0,deviceid=mydeviceid,engine_host=raspberrypi,server_version=18.05.0-ce usage_total=7975104i,container_id="f85ef98b08bde3ad6a93fadcacfa33a57a1f833119cce0309cadb15c45df5cf4" 1528956921000000000

plugins/inputs/docker/docker.go
prashanthjbabu: I've renamed this to started_at and finished_at . 
I have a query regarding converting the time from RFC3339 format (Example : 2018-06-14T05:48:53.266176036Z ) to Unix Nano format . Golang has the option to convert using the following code : 
``
t1, err := time.Parse(time.RFC3339,"0001-01-01T00:00:00Z")
``
and then use 
``
t1.UnixNano()
``
Here's some sample output of the timestamps : 

> docker_container_status,container_image=ubuntu,container_name=vibrant_benz,container_status=running,container_version=unknown,deviceid=mydeviceid,engine_host=raspberrypi,server_version=18.05.0-ce started_at="2018-06-14T05:48:53.266176036Z",finished_at="0001-01-01T00:00:00Z",oomkilled=false,pid=5278i,exitcode=0i 1528956922000000000
> 
In the above case the docker is still running so it has a valid start time , although its endtime is invalid which is "0001-01-01T00:00:00Z" . If i use the above go code to convert this to unix nano , I get "-6795364578871345152" . As you can see it returns a long negative invalid value . Would this be okay? How else would you like me to handle this ?If this is fine then I will go ahead and commit this

plugins/inputs/docker/docker.go
prashanthjbabu: The unit tests seem to be failing after adding "container_status" as a tag . I'm not quite sure how to deal with this. Any ideas?

plugins/inputs/docker/docker.go
danielnelson: I think you should be able to use time.IsZero(), and if true skip adding the field.

prashanthjbabu: Thanks ! I've done that.

plugins/inputs/docker/docker.go
prashanthjbabu: I've fixed this . All tests pass now.

danielnelson:
I updated the README in https://github.com/influxdata/telegraf/commit/c98b58dacc9748740b6dfaa0d56c958ef5985cf2

danielnelson:
@Puneeth-n This feature will be included in Telegraf 1.8.0, but if you are okay with a slightly higher chance of encountering bugs you can try using the [nightly development builds](https://github.com/influxdata/telegraf#nightly-builds) to demo this feature now.

PR #4261, opened 2018-06-10
Fixed typos in nvidia_smi plugin doc

marcv81:
### Required for all PRs:


danielnelson:
@jackzampolin Can you verify these units?

marcv81:
I assumed this was a typo, but on second thought this change might be wrong. While I can guarantee that on my system these values are in MiB, nvidia-smi does not specify what unit is returned. Instead it returns both the value and an explicit unit. For all we know the unit might depend on the card or even the driver.

Could we try to get several users to report the unit on their card/driver? We should then be able decide how to follow up.

danielnelson:
For now let's just keep an eye out for any systems where it is not in MiB.

PR #4267, opened 2018-06-11
Use counter time in win perf counters

### LINKED ISSUE #4250: 1.7.0-rc1 PerfMon % Processor Time discrepancy
### Relevant telegraf.conf:

```
[global_tags]

[agent]
  interval = "60s"
  round_interval = true
  metric_batch_size = 1000
  metric_buffer_limit = 10000
  collection_jitter = "5s"
  flush_jitter = "0s"
  precision = ""
  debug = false
  quiet = false
  logfile = ""
  hostname = "***"
  omit_hostname = false

[[outputs.influxdb]]
  urls = ["https://***:8086"]
  database = "telegraf"
  retention_policy = ""
  write_consistency = "any"
  timeout = "5s"
  username = "***"
  password = "***"

[[inputs.win_perf_counters]]
    ObjectName = "Processor"
    Instances = ["*"]
    Counters = ["% Idle Time", "% Interrupt Time", "% Privileged Time", "% User Time", "% Processor Time"]
    Measurement = "win_cpu"
    IncludeTotal=true
```

### System info:

Telegraf: 1.7.0-rc1
Windows: Server 2012 R2

### Steps to reproduce:

Compare % Process Time in Grafana (or the graphing tool of your choice) and perfmon

### Expected behavior:

The collected values are identical to the actual values on the system

### Actual behavior:

During the time span where CPU Load was at 100% in the following screenshot perfmon reported a steady load of 82-88% of the same exact counter, so 12-15% less than the collected value.
grafik

vlastahajek:
Trying to fix #4250: added possibility to use timestamp from performance counters

### DIFF
--- a/plugins/inputs/win_perf_counters/win_perf_counters.go
+++ b/plugins/inputs/win_perf_counters/win_perf_counters.go
@@ -353,6 +353,11 @@ func (m *Win_PerfCounters) Gather(acc telegraf.Accumulator) error {
 						collectFields[instance][sanitizedChars.Replace(metric.counter)] = float32(cValue.Value)
 					}
 				}
+			} else {
+				//ignore invalid data from as some counters from process instances returns this sometimes
+				if phderr, ok := err.(*PdhError); ok && phderr.ErrorCode != PDH_INVALID_DATA && phderr.ErrorCode != PDH_CALC_NEGATIVE_VALUE {
+					return fmt.Errorf("error while getting value for counter %s: %v", metric.counterPath, err)
+				}
 			}
 		}
 	}
--- a/plugins/inputs/win_perf_counters/win_perf_counters.go
+++ b/plugins/inputs/win_perf_counters/win_perf_counters.go
@@ -254,18 +254,15 @@ func (m *Win_PerfCounters) Gather(acc telegraf.Accumulator) error {
 			m.counters = m.counters[:0]
 		}
 
-		err = m.query.Open()
-		if err != nil {
+		if err = m.query.Open(); err != nil {
 			return err
 		}
 
-		err = m.ParseConfig()
-		if err != nil {
+		if err = m.ParseConfig(); err != nil {
 			return err
 		}
 		//some counters need two data samples before computing a value
-		err = m.query.CollectData()
-		if err != nil {
+		if err = m.query.CollectData(); err != nil {
 			return err
 		}
 		m.lastRefreshed = time.Now()
@@ -289,8 +286,7 @@ func (m *Win_PerfCounters) Gather(acc telegraf.Accumulator) error {
 		}
 	} else {
 		timestamp = time.Now()
-		err = m.query.CollectData()
-		if err != nil {
+		if err = m.query.CollectData(); err != nil {
 			return err
 		}
 	}
--- a/plugins/inputs/win_perf_counters/win_perf_counters.go
+++ b/plugins/inputs/win_perf_counters/win_perf_counters.go
@@ -307,7 +307,7 @@ func (m *Win_PerfCounters) Gather(acc telegraf.Accumulator) error {
 				collectFields[instance][sanitizedChars.Replace(metric.counter)] = float32(value)
 			} else {
 				//ignore invalid data from as some counters from process instances returns this sometimes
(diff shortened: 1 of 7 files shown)


matthenning:
@danielnelson I'm not sure this can be classified as an enhancement.
If it indeed fixes #4250 this is a bugfix.

plugins/inputs/win_perf_counters/win_perf_counters.go
```diff
@@ -349,12 +363,20 @@ func (m *Win_PerfCounters) Gather(acc telegraf.Accumulator) error {
...
-		acc.AddFields(instance.name, fields, tags)
+		acc.AddFields(instance.name, fields, tags, timestamp)
 	}
 
 	return nil
 }
 
+//returns true if err is an error we count with
+func isKnowError(err error) bool {
+	if phderr, ok := err.(*PdhError); ok && (phderr.ErrorCode == PDH_INVALID_DATA || phderr.ErrorCode == PDH_CALC_NEGATIVE_VALUE || phderr.ErrorCode == PDH_CSTATUS_INVALID_DATA) {
```
danielnelson: Could you line wrap this around 78 chars?

plugins/inputs/win_perf_counters/win_perf_counters.go
```diff
@@ -349,12 +363,20 @@ func (m *Win_PerfCounters) Gather(acc telegraf.Accumulator) error {
...
 		}
-		acc.AddFields(instance.name, fields, tags)
+		acc.AddFields(instance.name, fields, tags, timestamp)
 	}
 
 	return nil
 }
 
+//returns true if err is an error we count with
+func isKnowError(err error) bool {
```
danielnelson: Can you rename this function and redo the comment?  I think based on your comment above maybe you could name it `isCounterDataError`.  You can probably remove the comment above if this name is more precise.

plugins/inputs/win_perf_counters/win_perf_counters_integration_test.go
```diff
@@ -98,6 +103,10 @@ func TestWinPerformanceQueryImpl(t *testing.T) {
 	require.NoError(t, err)
 
 	arr, err := query.GetFormattedCounterArrayDouble(hCounter)
+	if phderr, ok := err.(*PdhError); ok && phderr.ErrorCode != PDH_INVALID_DATA && phderr.ErrorCode != PDH_CALC_NEGATIVE_VALUE {
+		time.Sleep(time.Second)
```
danielnelson: This is a red flag for me because it indicates the test may be timing dependent, why do we need to sleep?

vlastahajek: Invalid data error happens randomly, but can be seen several times in an hour. It generally means that some counter provided invalid value or it was unable to compute value from samples gathered during provided period. If we give it more time (which happens automatically during next gather cycle in normal Telegraf run), it solves the problem.

When using wildcards expansion, each counter is queried for value separately is it is filtered in my code on that level.
When using __PdhGetFormattedCounterArray_, it queries multiple counters at once and it seems that if any counter returns status about invalid data, the function fails completely.

The code historically didn't bothered about errors, they were totally ignored. Maybe it is the way.

It makes sense for me to tell about unexpected errors.

danielnelson: If this can happen in the actual code then we should probably ignore or ignore and log.  In the tests though this will actually be more problematic, we don't want to miss failures but we also don't want slow tests or intermittent failures.

The best fix would be to mock the calls to the library, in the meantime mark this test to skip if `-short` is set during testing, which I just added in ee6e4b0afd5107d7ce8bcd9a9123a3f6129aae5f.

vlastahajek: The test is already skipped in the short mode. 

Do you mean to log such error as warning ? Not as error as it is now..

Regarding commit ee6e4b0. Are all tests (not the short mode) run some check during a pull request checking?

danielnelson: Okay I think we are good then, obviously timing based stuff is problematic but since it is skipped on short it is okay.  We always only run `go test -short` for CI unit tests, which means tests like this are almost always skipped unless a developer runs them by hand.

Issue #4260, opened 2018-06-10
Cannot select nested json fields using tag_keys with json parser


Freakqency:
"total_devices": 5,
"total_threads": 10,
"shares": {
        "total": 5,
        "accepted": 5,
        "rejected": 0,
        "avg_find_time": 4
    },
I am able to pass "total_devices"  ,"total_threads" as my tag_keys and they get stored in influxdb and i able to visualize them.But i am not able to pass the subfields of shares like shares.total,shares.accepeted etc .to telegraft
this is my current telegraf configuration 

 [[inputs.httpjson]]
   servers = [
     "http://192.168.137.1:21555/"
   ]
   response_timeout = "5s"
   method = "GET"
   tag_keys = [ "total_devices","total_threads","shares_total"]

PR #4284, opened 2018-06-13
Allow for Tags on non base level JSON format

maxunt:
### Required for all PRs:

### DIFF
--- a/plugins/parsers/json/parser.go
+++ b/plugins/parsers/json/parser.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"log"
 	"strconv"
 	"strings"
 	"time"
@@ -44,18 +45,6 @@ func (p *JSONParser) parseObject(metrics []telegraf.Metric, jsonOut map[string]i
 		tags[k] = v
 	}
 
-	for _, tag := range p.TagKeys {
-		switch v := jsonOut[tag].(type) {
-		case string:
-			tags[tag] = v
-		case bool:
-			tags[tag] = strconv.FormatBool(v)
-		case float64:
-			tags[tag] = strconv.FormatFloat(v, 'f', -1, 64)
-		}
-		delete(jsonOut, tag)
-	}
-
 	f := JSONFlattener{}
 	err := f.FullFlattenJSON("", jsonOut, true, true)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


plugins/parsers/json/parser.go
```diff
@@ -57,19 +57,59 @@ func (p *JSONParser) parseObject(metrics []telegraf.Metric, jsonOut map[string]i
...
+		if fields[name] != nil {
+			switch fields[name].(type) {
+			//case of field is string
+			case string:
+				tags[name] = fields[name].(string)
+				delete(fields, name)
+			case bool:
+				tags[name] = strconv.FormatBool(fields[name].(bool))
+				delete(fields, name)
+			case int:
```
danielnelson: I think it is not possible to have ints in parsed JSON, so we can remove this case.  Let's also add a warning though if we hit the default case:
```
log.Printf("E! [parsers.json] Unrecognized type %T", value)
```

plugins/parsers/json/parser.go
```diff
@@ -57,19 +57,59 @@ func (p *JSONParser) parseObject(metrics []telegraf.Metric, jsonOut map[string]i
...
 
+//will take in field map with strings and bools,
+//search for TagKeys that match fieldnames and add them to tags
+//will delete any strings/bools that shouldn't be fields
+//assumes that any non-numeric values in TagKeys should be displayed as tags
+func (p *JSONParser) switchFieldToTag(tags map[string]string, fields map[string]interface{}) (map[string]string, map[string]interface{}) {
+	for _, name := range p.TagKeys {
+		//switch any fields in tagkeys into tags
+		if fields[name] != nil {
+			switch fields[name].(type) {
```
danielnelson: To avoid needing to type assert in each case, you can do this:
```
switch value := fields[name].(type) {
```

Then in each case the value will already be in the type of the case label.

plugins/parsers/json/parser.go
```diff
@@ -57,19 +57,59 @@ func (p *JSONParser) parseObject(metrics []telegraf.Metric, jsonOut map[string]i
...
 }
 
+//will take in field map with strings and bools,
+//search for TagKeys that match fieldnames and add them to tags
+//will delete any strings/bools that shouldn't be fields
+//assumes that any non-numeric values in TagKeys should be displayed as tags
+func (p *JSONParser) switchFieldToTag(tags map[string]string, fields map[string]interface{}) (map[string]string, map[string]interface{}) {
+	for _, name := range p.TagKeys {
+		//switch any fields in tagkeys into tags
+		if fields[name] != nil {
```
danielnelson: Nitpick, instead of nesting the switch statement in this check, I would check and continue:
```
if fields[name] == nil {
    continue
}

switch ....
```

This help with readability because the scope of the conditional is reduced.

plugins/parsers/json/parser.go
```diff
@@ -119,6 +159,9 @@ func (f *JSONFlattener) FlattenJSON(
 	if f.Fields == nil {
 		f.Fields = make(map[string]interface{})
 	}
+
+	//bool values to be changed to allow strings, bools as fields
+	//changed to true true, func switchFieldToTag will remove values
```
danielnelson: I would remove this comment

plugins/parsers/json/parser_test.go
```diff
@@ -440,3 +441,34 @@ func TestHttpJsonBOM(t *testing.T) {
...
+	"tester": "work",
+	"tester2": "don't want this",
+	"tester3": 7.93
+	}
+	}`
+
+	parser := JSONParser{
+		MetricName: "json_test",
+		TagKeys:    []string{"total_devices", "total_threads", "shares_tester", "shares_tester3"},
+		//TagKeys: []string{"mytag", "a", "b_c", "b_d"},
```
danielnelson: Remove any commented out code

plugins/parsers/json/parser_test.go
```diff
@@ -440,3 +441,34 @@ func TestHttpJsonBOM(t *testing.T) {
...
+	}
+
+	metrics, err := parser.Parse([]byte(testString))
+	t.Logf("error: %v", err)
+	t.Logf("num of metrics: %v", len(metrics))
+	t.Logf("metric tags: %v", metrics[0].Tags())
+	t.Logf("metric fields: %v", metrics[0].Fields())
```
danielnelson: Instead of logging the values, use require to assert that the values are as expected.

plugins/parsers/json/parser.go
```diff
@@ -57,19 +57,59 @@ func (p *JSONParser) parseObject(metrics []telegraf.Metric, jsonOut map[string]i
...
 
 	f := JSONFlattener{}
-	err := f.FlattenJSON("", jsonOut)
+	err := f.FullFlattenJSON("", jsonOut, true, true)
 	if err != nil {
 		return nil, err
 	}
 
-	metric, err := metric.New(p.MetricName, tags, f.Fields, time.Now().UTC())
+	tags, nFields := p.switchFieldToTag(tags, f.Fields)
```
danielnelson: Now that we are adjusting the values after flattening, is it still required to keep the loop above on line 47?

danielnelson:
closes #4260

PR #4285, opened 2018-06-14
Doc how to parse telegraf logs

adrianlzt:
### Required for all PRs:


plugins/inputs/logparser/README.md
```diff
@@ -213,6 +213,48 @@ A multi-line literal string allows us to encode the pattern:
     custom_patterns = 'UNICODE_ESCAPE (?:\\u[0-9A-F]{4})+'
 ```
 
+#### Parsing Telegraf log file
+We can use logparser to convert the log lines generated by Telegraf in metrics.
+
+To do this we need to configure Telegraf to write logs to a file (if we have systemd, by
+default, it will be written to journald):
```
danielnelson: This isn't always the case, on most systems I have tested by default journald is setup to forward to rsyslog, when then writes the logs to `/var/log/syslog`.  See #4174

What's important is that the admin sets up the telegraf logs to go into a dedicated file, and this will do that, but maybe we should also mention that on some systems you can also do this from your syslog handler.

plugins/inputs/logparser/README.md
```diff
@@ -213,6 +213,48 @@ A multi-line literal string allows us to encode the pattern:
...
+```
+
+Logparser configuration:
+```toml
+[[inputs.logparser]]
+  files = ["/var/log/telegraf/telegraf.log"]
+
+  [inputs.logparser.grok]
+    measurement = "telegraf_log"
+    patterns = ['\A%{TIMESTAMP_ISO8601:timestamp:ts-rfc3339} %{TELEGRAF_LOG_LEVEL:level:tag}! %{GREEDYDATA:msg}']
```
danielnelson: `\A`?

Issue #3883, opened 2018-03-13
Input to read and parse a file every interval


mjf:
Hello.

I would like to propose new input plugin to **read and parse** (*Grok*) a file every interval, preferably called `[[inputs.parser]]` (or `[[inputs.reader]]` in case #3479 is planned to be done one day).

-----
- Also see #3878 for reference and some random thoughts and
- it may have connection to #3479 too.

dirkdevriendt:
We're also stuck on a seemingly simple use case where AFAICT we bump into a (sometimes surprising) shortcoming in the existing tail & logparser input inputs.

We have logfiles that have the date in their name and data in influx formatted lines.
`[[inputs.tail]]` with `files = ["/data/*.log"]` does not work because it does not pick up newly created files
`[[inputs.logparser]]` does not work because it does not support `data_format = "influx"` (and we do not know up front which metrics will be in there)

A solution would be if:
  - `tail` had a setting that would re-evaluate the list of files at every interval
  - `logparser` could read influx formatted files (assuming it is able to load new files in flight)

In addition to that, a setting to not only recreate the file list but also parse every file from the beginning at very interval would also fulfill the OP's request.

Note that this may also prevent some of the confusion around the issues #1829 #2141 #2847 #3492, where I suspect that the cause could for instance be that inotify keeps references to files that get truncated, but "loses" the ones to files that get deleted and recreated.
An option to refresh the file list would make debugging easier.

danielnelson:
@dirkdevriendt Keep an eye on #3479, once we do this the next step will be to merge logparser and tail.

mjf:
The `[[inputs.reader]]` should support at least these options:

- **parser type** (i.e. `"none"`, `"grok"`, `"influx"`, etc.);
- **change watch method**:
  -  `"inotify"`, `"mtime"` or `"poll"` for *files*;
  - `"list"`, `"mtime"`, `"dnotify"` for *directories*;
- **action on event**:
  - `"parse"` to run the **parser**;
  - `"exec"` (for "preprocessing" etc.);
- **continuous reading** (a.k.a. *"tail"*);
- **reuse last known position** in a file;
- **handle list of objects changes** (i.e. updates *on reload*, *every n-th interval* etc.)

and mabe few other options...

I would also suggest to let users form *sets* (groups) sharing similar combination of these options for list of path resources (directory file from *glob patterns*) etc. It should also support the *Golang* glob patterns etc.

[...]

Issue #3479, opened 2017-11-16
Add Grok as top level parser to be used with any plugin


russorat:
### Proposal:
Add GROK as a top level parser so that all plugins can take advantage of it. Currently it is only available for the logparser plugin. The specific ask was regarding the Kafka input. 

### Current behavior:
Grok parsing is only available for the logparser plugin

### Desired behavior:
Grok parsing is available for all plugins.

### Use case: [Why is this important (helps with prioritizing requests)]
Some users are pushing log messages from many machines into Kafka, and would like to add their grok patterns so that fields and tags are decoded automatically.

PR #4292, opened 2018-06-15
Added filename tag to tail input plugin.

wingsof:
### Required for all PRs:

- [v] Signed [CLA](https://influxdata.com/community/cla/).
- [v] Associated README.md updated.
- [v] Has appropriate unit tests.

Added filename tag for tail input plugin.
This information is needed when the case telegraf captures log from multiple files in the directory structure and there's a need to identify source.

### DIFF
--- a/plugins/inputs/tail/tail.go
+++ b/plugins/inputs/tail/tail.go
@@ -146,7 +146,11 @@ func (t *Tail) receiver(tailer *tail.Tail) {
 
 		m, err = t.parser.ParseLine(text)
 		if err == nil {
-			t.acc.AddFields(m.Name(), m.Fields(), m.Tags(), m.Time())
+			if m != nil {
+				tags := m.Tags()
+				tags["path"] = tailer.Filename
+				t.acc.AddFields(m.Name(), m.Fields(), tags, m.Time())
+			}
 		} else {
 			t.acc.AddError(fmt.Errorf("E! Malformed log line in %s: [%s], Error: %s\n",
 				tailer.Filename, line.Text, err))


plugins/inputs/tail/tail.go
```diff
@@ -146,7 +146,11 @@ func (t *Tail) receiver(tailer *tail.Tail) {
 
 		m, err = t.parser.ParseLine(text)
 		if err == nil {
-			t.acc.AddFields(m.Name(), m.Fields(), m.Tags(), m.Time())
+			if m != nil {
+				tags := m.Tags()
+				tags["filename"] = tailer.Filename
```
danielnelson: Can you change the name of this tag to `path`?  This will just be more consistent with the logparser input, and I hope to combine these plugins at some point.

wingsof:
Hello @danielnelson , it seems that circleci deps check failed due to operational failure. Would you please re-run check or should I make additional commit to rerun this?

danielnelson:
@wingsof If you can merge/rebase against our master to get the new CircleCI build image it should resolve this issue.

wingsof:
tag name changed to 'path' =-)

and would you share your plan about merging tail and logparser plugin? I was planning to implement capture log from newly created file while telegraf running, but if tail plugin will be merged eventually, maybe I can check logparser plugin first.

danielnelson:
What we would like to do is make the grok parser in the logparser plugin available as a first class parser available to all plugins that support the `data_format` option: https://github.com/influxdata/telegraf/issues/3479

Once this is done then tail will be able to do everything that logparser can do, and there will no longer be a need to have both plugins.

mjf:
> Once this is done then tail will be able to do everything that logparser can do, and there will no longer be a need to have both plugins.

Does that mean we will be finaly able to use *Grok* parser for gathering metrices from the `/proc` *procfs* filesystem as well (if not using the *tail* plugin, at least with *exec* plugin with `cat ...` or something alike)?

danielnelson:
@mjf Yes, once #3479 is done you could do this with an exec script.  You will also be able to use it along with a solution for #3883 soon.

mjf:
@danielnelson That's great! Thanks!

dirkdevriendt:
Ouch, just noticed this change and now our data is being "polluted" with an extra tag that has no value for us...

I understand this can be a very useful feature (so, thanks! :-)), but considering how much of a pain it can be to change tags in influx, it might be better to have it as an optional / opt-in /configurable feature?

glinton:
@dirkdevriendt That's not a bad idea, in the meantime you can use [`tagexclude`](https://github.com/influxdata/telegraf/blob/1.8.2/docs/CONFIGURATION.md#measurement-filtering) to remove that tag from your metrics.

dirkdevriendt:
Thanks @glinton!
That's one of those golden nuggets I must have actually seen without connecting the dots. From now on I'll be using `taginclude` for these use cases to make sure I'm only getting the tags I opt-in to!

joriws:
I'd like to see same path-added to input.file too! Because if I glob there I want to be able set measurement name by filename-prefix..

glinton:
@joriws that functionality should be present in telegraf versions newer than 1.8.0. File a new issue or feature request if the behavior is unsatisfactory [here](https://github.com/influxdata/telegraf/issues/new/choose).

Issue #4293, opened 2018-06-15
procstat memory fields missing in aggregation


adrianlzt:
### Relevant telegraf.conf:
```
[[outputs.file]]
  files = ["stdout"]

[agent]
  interval = "3s"
  flush_interval = "3s"

[[inputs.procstat]]
  pattern = "nginx: worker"
  fieldpass = ["pid","memory_rss"]

[[aggregators.minmax]]
  period = "3s"
```

### System info:
Telegraf v1.7.0 (git: release-1.7 f4d22dd4)

Arch Linux

### Steps to reproduce:

1. ``telegraf -config telegraf.conf``

### Expected behavior:
All fields from procstat being processed by the aggregator

### Actual behavior:
Only ``pid`` field gets computed by the aggregator.
```
procstat,host=archer,pattern=nginx:\ worker,process_name=nginx pid=29094i,memory_rss=1949696i 1529058174000000000                                                                                                  
procstat,host=archer,pattern=nginx:\ worker,process_name=nginx pid=29080i,memory_rss=1949696i 1529058174000000000                                                                                                  
procstat,host=archer,pattern=nginx:\ worker,process_name=nginx pid=29084i,memory_rss=1949696i 1529058174000000000                                                                                                  
procstat,host=archer,pattern=nginx:\ worker,process_name=nginx memory_rss=1949696i,pid=29094i 1529058177000000000                                                                                                  
procstat,host=archer,pattern=nginx:\ worker,process_name=nginx pid=29080i,memory_rss=1949696i 1529058177000000000                                                                                                  
procstat,host=archer,pattern=nginx:\ worker,process_name=nginx memory_rss=1949696i,pid=29084i 1529058177000000000                                                                                                  
procstat,host=archer,pattern=nginx:\ worker,process_name=nginx pid_min=29080,pid_max=29094 1529058177000000000
```

### Additional info:
With the procstat input, only the ``memory_*`` fields are not processed.

adrianlzt:
[``convert``](https://github.com/influxdata/telegraf/blob/master/plugins/aggregators/minmax/minmax.go#L104) function fails if the number is not float64 or int64, but memory fields, comming from psutil lib, [are uint64](https://github.com/shirou/gopsutil/blob/master/process/process.go#L43)

danielnelson:
Closed by #4294

PR #4294, opened 2018-06-15
Fix minmax and basicstats aggregators to use all number types

adrianlzt:
These aggregators were not working with other numbers type not being
float64 or int64.
Telegraf uses externals libs that could use other types.

### Required for all PRs:


adrianlzt:
Fix #4293

danielnelson:
Fields types can now (since 1.6) only be int64, uint64, float64, string, and bool \[1\].  Let's just add support for uint64 so we don't need to use the reflect package.

There are a variety of unittests that use other units, but the unittests just need to be fixed.

\[1\]: https://github.com/influxdata/telegraf/blob/master/metric/metric.go#L251

PR #3382, opened 2017-10-24
Adapt the MySQL input plugin to handle specific MariaDB implementation of user statistics tables

etiennecarriere:
Linked to https://github.com/influxdata/telegraf/issues/2910
### Required for all PRs:

- [ ] Signed [CLA](https://influxdata.com/community/cla/).
- [x] Associated README.md updated.
- [x] Has appropriate unit tests.

Issue #1645, opened 2016-08-17
Percona input plugin


xert:
### Proposal:

I wish to add Percona user statistics https://www.percona.com/doc/percona-server/5.6/diagnostics/user_stats.html into the mysql input plugin.

I have a working patch for this. Should I add PR?
### Current behavior:

Percona user statistics currently not implemented in the mysql plugin
### Desired behavior:

Add Percona user statistics to mysql input plugin
### Use case: [Why is this important (helps with prioritizing requests)]

Allows to monitor user, client, table and index usage (see the Percona documentation linked above)

sparrc:
@xert rather than overloading the mysql plugin, I would prefer if we had a separate `percona` input plugin, unless there is a significant efficiency to be gained from integrating it directly into the mysql plugin.

xert:
@sparrc That was my initial perspective too and it's still a possible option, of course.

I see this possible solutions:

1. Make separate percona plugin, collecting just userstat metrics (disadvantage - the DSN must be configured twice, telegraf will make unnecessary additional connections to the server)
2. Copy mysql as percona and add userstats to it (DSN needs to be configured just once, but this solution it's not good for maintenance)
3. Extend mysql plugin capabilities (better then 2. for sure, but adds complexity to the plugin)
4. Separate common functionality needed by both plugins to a shared piece of code (generic sql plugin?) and make plugins mysql and percona (maybe mariadb too - I am not sure if mariadb has some differences in performance / information schema)

I think that the best approach is 3. and then work on 4. Currently I see there are opportunities for improvement in the mysql plugin. For example the metric queries are defined as constants at the top of the file and are used later in functions. If You change the SQL it's not obvious which parts of code are affected by this change. I think it lead to missing column in #1647.

What do You think?

sparrc:
Personally I still prefer (1), because when we're talking about a MySQL database I don't think adding 1 more connection is much overhead. Although percona is obviously closely related to MySQL, it's definitely a separate product and I think it makes sense to collect percona-specific metrics only.

xert:
Ok, I'll try to make a new plugin collecting just percona specifics.

joelsdc:
Hi @xert, can you share your current working patch please? I'd love to give it a try... :D

danielnelson:
Closed in #4306

PR #4306, opened 2018-06-15
Handle variations in the user_statistics collecting

### LINKED ISSUE #2910: MariaDB user statistics are different from Percona version
### System info:

Telegraf 1.3.1, MariaDB 10.1.24

### Steps to reproduce:

1. Configure Telegraf with MySQL input plugin
2. Turn on gather_user_statistics in Telegraf and userstat in MariaDB

### Expected behavior:
Information should be collected.

### Actual behavior:
Error in logs:
Jun 12 14:43:10 db1 telegraf[14838]: 2017-06-12T13:43:10Z E! MySQL Error gathering user stats: Error 1054: Unknown column 'rows_fetched' in 'field list'
Jun 12 14:43:10 db1 telegraf[14838]: 2017-06-12T13:43:10Z E! Error in plugin [inputs.mysql]: sql: expected 26 destination arguments in Scan, not 23

### Additional info:
                USER: stats
          TOTAL_CONNECTIONS: 83
     CONCURRENT_CONNECTIONS: 0
             CONNECTED_TIME: 55
                  BUSY_TIME: 1.2070639999999992
                   CPU_TIME: 1.1825710000000003
             BYTES_RECEIVED: 82791
                 BYTES_SENT: 1655069
       BINLOG_BYTES_WRITTEN: 0
                  ROWS_READ: 0
                  ROWS_SENT: 2264
               ROWS_DELETED: 0
              ROWS_INSERTED: 0
               ROWS_UPDATED: 0
            SELECT_COMMANDS: 443
            UPDATE_COMMANDS: 0
             OTHER_COMMANDS: 0
        COMMIT_TRANSACTIONS: 0
      ROLLBACK_TRANSACTIONS: 0
         DENIED_CONNECTIONS: 27
           LOST_CONNECTIONS: 0
              ACCESS_DENIED: 6
              EMPTY_QUERIES: 117
      TOTAL_SSL_CONNECTIONS: 0
       MAX_STATEMENT_TIME_EXCEEDED: 0

glinton:
There are differences in the `user_statistics` table (if it exists) between MariaDB, Percona, and MySQL. This should allow the stats to be collected properly.

MariaDB has two different table definitions:
 - https://mariadb.com/kb/en/library/information-schema-user_statistics-table/

Percona has it's own:
 - https://www.percona.com/doc/percona-server/LATEST/diagnostics/user_stats.html#USER_STATISTICS

MySQL doesn't have the table, but there may be a patch:
 - https://www.byte.nl/blog/enhancing-mysql-with-user-statistics

Resolves #2910
Closes #3382

### Required for all PRs:


plugins/inputs/mysql/mysql.go
```diff
@@ -920,74 +861,210 @@ func (m *Mysql) GatherUserStatisticsStatuses(db *sql.DB, serv string, acc telegr
...
+				return fmt.Errorf("Unknown column type - %T", v)
+			}
 		}
 		acc.AddFields("mysql_user_stats", fields, tags)
 	}
 	return nil
 }
 
+// convertColumns converts selected column names to lowercase.
+func convertColumns(s []string, e error) ([]string, error) {
```
danielnelson: Maybe rename `columnsToLower` or `lowerColumns`?

plugins/inputs/mysql/mysql.go
```diff
@@ -920,74 +861,210 @@ func (m *Mysql) GatherUserStatisticsStatuses(db *sql.DB, serv string, acc telegr
...
-		rollback_transactions  int64
-		denied_connections     int64
-		lost_connections       int64
-		access_denied          int64
-		empty_queries          int64
-		total_ssl_connections  int64
-		count                  uint32
-	)
+
+	cols, err := convertColumns(rows.Columns())
```
danielnelson: This is just an observation.  I never knew you could pass multiple arguments like this, interesting.  I'm not sure I like how it becomes the outer functions job to deal with an upstream error, but I like how it can compress the error handling needed in this function.

plugins/inputs/mysql/mysql.go
```diff
@@ -920,74 +861,210 @@ func (m *Mysql) GatherUserStatisticsStatuses(db *sql.DB, serv string, acc telegr
...
-			"commit_transactions":    commit_transactions,
-			"rollback_transactions":  rollback_transactions,
-			"denied_connections":     denied_connections,
-			"lost_connections":       lost_connections,
-			"access_denied":          access_denied,
-			"empty_queries":          empty_queries,
-			"total_ssl_connections":  total_ssl_connections,
+		for i := range cols {
+			if i == 0 {
+				continue // skip "user"
```
danielnelson: Percona index 0 comment applies here.

plugins/inputs/mysql/mysql.go
```diff
@@ -920,74 +861,210 @@ func (m *Mysql) GatherUserStatisticsStatuses(db *sql.DB, serv string, acc telegr
...
-			"update_commands":        update_commands,
-			"other_commands":         other_commands,
-			"commit_transactions":    commit_transactions,
-			"rollback_transactions":  rollback_transactions,
-			"denied_connections":     denied_connections,
-			"lost_connections":       lost_connections,
-			"access_denied":          access_denied,
-			"empty_queries":          empty_queries,
-			"total_ssl_connections":  total_ssl_connections,
+		for i := range cols {
```
danielnelson: I would use `for _, column := range cols {` and avoid using the index directly.  I know in some situations it can be slightly slower to iterate in this manner, but I think it almost always is more readable and less error prone.

plugins/inputs/mysql/mysql.go
```diff
@@ -920,74 +861,210 @@ func (m *Mysql) GatherUserStatisticsStatuses(db *sql.DB, serv string, acc telegr
...
+			if i == 0 {
+				continue // skip "user"
+			}
+			switch v := read[i].(type) {
+			case *bool:
```
danielnelson: Only include expected types and the default case can handle anything else.

plugins/inputs/mysql/mysql.go
```diff
@@ -920,74 +861,210 @@ func (m *Mysql) GatherUserStatisticsStatuses(db *sql.DB, serv string, acc telegr
...
-			&empty_queries, &total_ssl_connections, &count,
-		)
+		err = rows.Scan(read...)
 		if err != nil {
 			return err
 		}
 
-		tags := map[string]string{"server": servtag, "user": user}
-		fields := map[string]interface{}{
+		tags := map[string]string{"server": servtag, "user": *read[0].(*string)}
```
danielnelson: User is not always index 0 (percona)

plugins/inputs/mysql/mysql.go
```diff
@@ -920,74 +861,210 @@ func (m *Mysql) GatherUserStatisticsStatuses(db *sql.DB, serv string, acc telegr
...
+	if err != nil {
+		return err
+	}
+
+	read, err := getColSlice(len(cols))
+	if err != nil {
+		return err
+	}
+
+	if len(read) != len(cols) {
```
danielnelson: I would probably remove this check, since in theory it is impossible unless there is a programming error and it should. be picked up by the rows.Scan(

plugins/inputs/mysql/mysql.go
glinton: Good call. Is it stylistically ok to leave out the default case since it is immediately followed by the function's return?

plugins/inputs/mysql/mysql.go
glinton: :+1:

plugins/inputs/mysql/mysql.go
glinton: I'll make it more robust, but in all the [docs](https://www.percona.com/doc/percona-server/LATEST/diagnostics/user_stats.html#USER_STATISTICS) I checked it would be index 0.

plugins/inputs/mysql/mysql.go
glinton: In this case, I think leaving it is best. I'll still need the iterator for the `read` slice and it's pretty clear where `cols[i]` is only used in the switch block.

plugins/inputs/mysql/mysql.go
danielnelson: I usually leave the default in to catch any programming mistakes.

plugins/inputs/mysql/mysql.go
danielnelson: What about in this list of columns?

https://github.com/influxdata/telegraf/pull/4306/files#diff-d267f290fd16dce7d39997e69d775747R1041

plugins/inputs/mysql/mysql.go
```diff
@@ -790,72 +790,13 @@ func (m *Mysql) gatherGlobalStatuses(db *sql.DB, serv string, acc telegraf.Accum
 
 	// gather connection metrics from user_statistics for each user
 	if m.GatherUserStatistics {
-		conn_rows, err := db.Query("select user, total_connections, concurrent_connections, connected_time, busy_time, cpu_time, bytes_received, bytes_sent, binlog_bytes_written, rows_fetched, rows_updated, table_rows_read, select_commands, update_commands, other_commands, commit_transactions, rollback_transactions, denied_connections, lost_connections, access_denied, empty_queries, total_ssl_connections FROM INFORMATION_SCHEMA.USER_STATISTICS GROUP BY user")
+		err = m.GatherUserStatisticsStatuses(db, serv, acc)
```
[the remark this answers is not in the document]
danielnelson: What about L455 though?

plugins/inputs/mysql/mysql.go
```diff
@@ -790,72 +790,13 @@ func (m *Mysql) gatherGlobalStatuses(db *sql.DB, serv string, acc telegraf.Accum
...
-					"denied_connections":     denied_connections,
-					"lost_connections":       lost_connections,
-					"access_denied":          access_denied,
-					"empty_queries":          empty_queries,
-					"total_ssl_connections":  total_ssl_connections,
-				}
-
-				acc.AddFields("mysql_user_stats", fields, tags)
+			// disable collecting if table is not found (suppresses repeat errors)
+			if strings.Contains(err.Error(), "nknown table 'user_statistics'") {
```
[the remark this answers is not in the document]
danielnelson: Okay, personally I would probably just ToLower it, but ideally this type of error matching will be uncommon.

I did find an interesting link off of http://go-database-sql.org/errors.html#identifying-specific-database-errors to a [package of error codes for mysql](https://github.com/VividCortex/mysqlerr), this might be helpful, though I'm unsure if these work on all mysql flavors.

plugins/inputs/mysql/mysql.go
glinton: alias for user iirc

glinton: AH! I was referencing `client_statistics` table instead of `user_statistics` table

glinton:
Closes #1645

PR #4307, opened 2018-06-15
Bugfix/procstat adds new procstat metric about pgrep search

maxunt:
### Required for all PRs:

### DIFF
--- a/plugins/inputs/procstat/procstat_test.go
+++ b/plugins/inputs/procstat/procstat_test.go
@@ -11,6 +11,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/influxdata/telegraf/internal"
 	"github.com/influxdata/telegraf/testutil"
 	"github.com/shirou/gopsutil/cpu"
 	"github.com/shirou/gopsutil/process"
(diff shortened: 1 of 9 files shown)


plugins/inputs/procstat/procstat_test.go
```diff
@@ -369,3 +370,39 @@ func TestGather_cgroupPIDs(t *testing.T) {
 	assert.Equal(t, []PID{1234, 5678}, pids)
 	assert.Equal(t, td, tags["cgroup"])
 }
+
+func TestExitStatus(t *testing.T) {
+	out, err := exec.Command("pgrep", "-G", "sys").Output()
```
danielnelson: This command could return different results on different systems, so it could create random failures, perhaps you can make the needed error by hand.  Also, it tests a function in `internal/internal.go` so it should be placed in `internal/internal_test.go`

plugins/inputs/procstat/procstat_test.go
```diff
@@ -369,3 +370,39 @@ func TestGather_cgroupPIDs(t *testing.T) {
...
+	t.Logf("cmd out: %v", out)
+	exitNum, _ := internal.ExitStatus(err)
+	t.Logf("error status: %v", exitNum)
+	require.Equal(t, 1, exitNum)
+}
+
+func TestNoPgrepResults(t *testing.T) {
+	p := Procstat{
+		PidFinder: "pgrep",
```
danielnelson: Same comment as above for this and the next test. You should be able to use testPgrep in this same file to fake the results.

plugins/inputs/procstat/procstat_test.go
```diff
@@ -369,3 +370,39 @@ func TestGather_cgroupPIDs(t *testing.T) {
 	assert.Equal(t, []PID{1234, 5678}, pids)
 	assert.Equal(t, td, tags["cgroup"])
 }
+
+func TestExitStatus(t *testing.T) {
+	out, err := exec.Command("pgrep", "-G", "sys").Output()
+	t.Logf("cmd out: %v", out)
```
danielnelson: Remove any logging in the test cases, use assertions to document results.

PR #4300, opened 2018-06-15
output plugin for Splunk HEC API

shrug42:
### Required for all PRs:

- [x] Signed [CLA](https://influxdata.com/community/cla/).
- [x] Associated README.md updated.
- [x] Has appropriate unit tests.

PR #4339, opened 2018-06-22
Splunk Metrics serializer

ronnocol:
This serializer properly formats the metrics data according to the Splunk metrics JSON specification, it can be used with any output that supports serializers, including file or HTTP. I decided that just formatting the data according to the Splunk spec was a better investment than managing the whole output pipeline. Now you can use telegraf without requiring a HEC (by using a file output), or you can use the well tested and maintained HTTP output and just set the extra headers required by the HEC.

Please see the docs in README.md

### Required for all PRs:

### DIFF
--- a/docs/DATA_FORMATS_OUTPUT.md
+++ b/docs/DATA_FORMATS_OUTPUT.md
@@ -7,7 +7,7 @@ plugins.
 1. [InfluxDB Line Protocol](#influx)
 1. [JSON](#json)
 1. [Graphite](#graphite)
-1. [SplunkMetric](#splunkmetric)
+1. [SplunkMetric](../plugins/serializers/splunkmetric/README.md)
 
 You will be able to identify the plugins with support by the presence of a
 `data_format` config option, for example, in the `file` output plugin:
@@ -210,67 +210,3 @@ reference the documentation for the specific plugin.
   ## the power of 10 less than the specified units.
   json_timestamp_units = "1s"
 ```
-
-## SplunkMetric
-
-The SplunkMetric format translates the telegraf metrics into a JSON format compatible
-with a Splunk Metrics index. Normally you would use this to send the metrics to a
-HTTP Event Collector (HEC), it follows the format specified in the document [here.](http://dev.splunk.com/view/SP-CAAAFDN#json)
-
-```toml
-[[outputs.http]]
-   ## URL is the address to send metrics to
-   url = "https://localhost:8088/services/collector"
-
-   ## Timeout for HTTP message
-   # timeout = "5s"
-
-   ## HTTP method, one of: "POST" or "PUT"
-   # method = "POST"
-
-   ## HTTP Basic Auth credentials
-   # username = "username"
-   # password = "pa$$word"
-
-   ## Optional TLS Config
-   # tls_ca = "/etc/telegraf/ca.pem"
-   # tls_cert = "/etc/telegraf/cert.pem"
-   # tls_key = "/etc/telegraf/key.pem"
-   ## Use TLS but skip chain & host verification
-   # insecure_skip_verify = false
-
-   ## Data format to output.
-   ## Each data format has it's own unique set of configuration options, read
-   ## more about them here:
-   ## https://github.com/influxdata/telegraf/blob/master/docs/DATA_FORMATS_OUTPUT.md
-   data_format = "splunkmetric"
-   # Enable hec_routing 
-   hec_routing = true
-
-   ## Additional HTTP headers
-   [outputs.http.headers]
-   # Should be set manually to "application/json" for json data_format
-      Content-Type = "application/json"
-      Authorization = "Splunk xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
-      X-Splunk-Request-Channel = "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
-```
-
-It also has the option to output a more basic JSON structure that can be used by a
-Splunk Universal Forwarder or modular input.
-
-```toml
-[[outputs.file]]
-  ## Files to write to, "stdout" is a specially handled file.
-  files = ["stdout", "/tmp/metrics.out"]
-
-  ## Data format to output.
-  ## Each data format has its own unique set of configuration options, read
-  ## more about them here:
-  ## https://github.com/influxdata/telegraf/blob/master/docs/DATA_FORMATS_OUTPUT.md
-  data_format = "json"
-  ## hec_routing defaults to false
-  # hec_routing = false
-```
-
-Please see the [Splunk Metrics README](../plugins/serializers/splunkmetric/README.md) for additional
-information.
--- a/docs/DATA_FORMATS_OUTPUT.md
+++ b/docs/DATA_FORMATS_OUTPUT.md
@@ -7,6 +7,7 @@ plugins.
 1. [InfluxDB Line Protocol](#influx)
 1. [JSON](#json)
 1. [Graphite](#graphite)
+1. [SplunkMetric](#splunkmetric)
 
 You will be able to identify the plugins with support by the presence of a
 `data_format` config option, for example, in the `file` output plugin:
@@ -209,3 +210,67 @@ reference the documentation for the specific plugin.
   ## the power of 10 less than the specified units.
   json_timestamp_units = "1s"
 ```
+
+## SplunkMetric
+
+The SplunkMetric format translates the telegraf metrics into a JSON format compatible
+with a Splunk Metrics index. Normally you would use this to send the metrics to a
+HTTP Event Collector (HEC), it follows the format specified in the document [here.](http://dev.splunk.com/view/SP-CAAAFDN#json)
+
+```toml
+[[outputs.http]]
+#   ## URL is the address to send metrics to
+   url = "https://localhost:8088/services/collector"
+#
+#   ## Timeout for HTTP message
+#   # timeout = "5s"
+#
+#   ## HTTP method, one of: "POST" or "PUT"
+#   # method = "POST"
+#
+#   ## HTTP Basic Auth credentials
+#   # username = "username"
+#   # password = "pa$$word"
+#
+#   ## Optional TLS Config
+#   # tls_ca = "/etc/telegraf/ca.pem"
+#   # tls_cert = "/etc/telegraf/cert.pem"
+#   # tls_key = "/etc/telegraf/key.pem"
+#   ## Use TLS but skip chain & host verification
+#   # insecure_skip_verify = false
+#
+#   ## Data format to output.
+#   ## Each data format has it's own unique set of configuration options, read
+#   ## more about them here:
+#   ## https://github.com/influxdata/telegraf/blob/master/docs/DATA_FORMATS_OUTPUT.md
+   data_format = "splunkmetric"
+   # Enable hec_routing 
+   hec_routing = true
+#
+#   ## Additional HTTP headers
+    [outputs.http.headers]
+#   # Should be set manually to "application/json" for json data_format
+      Content-Type = "application/json"
+      Authorization = "Splunk xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
+      X-Splunk-Request-Channel = "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
+```
+
+It also has the option to output a more basic JSON structure that can be used by a
+Splunk Universal Forwarder or modular input.
+
+```toml
+[[outputs.file]]
+  ## Files to write to, "stdout" is a specially handled file.
+  files = ["stdout", "/tmp/metrics.out"]
+
+  ## Data format to output.
+  ## Each data format has its own unique set of configuration options, read
+  ## more about them here:
+  ## https://github.com/influxdata/telegraf/blob/master/docs/DATA_FORMATS_OUTPUT.md
+  data_format = "json"
+  ## hec_routing defaults to false
+  # hec_routing = false
+```
+
+Please see the [Splunk Metrics README](../plugins/serializers/splunkmetric/README.md) for additional
+information.
--- a/docs/DATA_FORMATS_OUTPUT.md
+++ b/docs/DATA_FORMATS_OUTPUT.md
@@ -219,37 +219,37 @@ HTTP Event Collector (HEC), it follows the format specified in the document [her
 
 ```toml
 [[outputs.http]]
-#   ## URL is the address to send metrics to
+   ## URL is the address to send metrics to
    url = "https://localhost:8088/services/collector"
-#
-#   ## Timeout for HTTP message
-#   # timeout = "5s"
-#
-#   ## HTTP method, one of: "POST" or "PUT"
-#   # method = "POST"
-#
-#   ## HTTP Basic Auth credentials
-#   # username = "username"
-#   # password = "pa$$word"
-#
-#   ## Optional TLS Config
-#   # tls_ca = "/etc/telegraf/ca.pem"
-#   # tls_cert = "/etc/telegraf/cert.pem"
-#   # tls_key = "/etc/telegraf/key.pem"
-#   ## Use TLS but skip chain & host verification
-#   # insecure_skip_verify = false
-#
-#   ## Data format to output.
-#   ## Each data format has it's own unique set of configuration options, read
-#   ## more about them here:
-#   ## https://github.com/influxdata/telegraf/blob/master/docs/DATA_FORMATS_OUTPUT.md
+
+   ## Timeout for HTTP message
+   # timeout = "5s"
+
+   ## HTTP method, one of: "POST" or "PUT"
+   # method = "POST"
+
+   ## HTTP Basic Auth credentials
+   # username = "username"
+   # password = "pa$$word"
+
+   ## Optional TLS Config
+   # tls_ca = "/etc/telegraf/ca.pem"
+   # tls_cert = "/etc/telegraf/cert.pem"
+   # tls_key = "/etc/telegraf/key.pem"
+   ## Use TLS but skip chain & host verification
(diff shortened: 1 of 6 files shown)


plugins/serializers/splunkmetric/splunkmetric.go
```diff
@@ -0,0 +1,135 @@
...
+
+func (s *serializer) SerializeBatch(metrics []telegraf.Metric) ([]byte, error) {
+
+	var serialized string
+
+	var objects []string
+
+	for _, metric := range metrics {
+		m, err := s.createObject(metric)
+		if err == nil {
```
phemmer: The error should be logged somewhere when not nil. Otherwise metrics are going to be getting dropped and the user won't have any idea why.

plugins/serializers/registry.go
```diff
@@ -73,6 +74,8 @@ func NewSerializer(config *Config) (Serializer, error) {
 		serializer, err = NewGraphiteSerializer(config.Prefix, config.Template, config.GraphiteTagSupport)
 	case "json":
 		serializer, err = NewJsonSerializer(config.TimestampUnits)
+	case "splunkmetric":
+		serializer, err = NewSplunkmetricSerializer(config.TimestampUnits)
```
phemmer: Splunk only supports seconds. The serializer should not be using `config.TimestampUnits`.

plugins/serializers/splunkmetric/splunkmetric.go
```diff
@@ -0,0 +1,135 @@
...
+	for _, m := range objects {
+		serialized = serialized + m + "\n"
+	}
+
+	return []byte(serialized), nil
+}
+
+func (s *serializer) createObject(metric telegraf.Metric) (metricString string, err error) {
+
+	/* Splunk supports one metric per line and has the following required names:
```
phemmer: This is not entirely accurate. Splunk's http event collector is not line-oriented. It's object oriented. You can shove multiple JSON objects on a single line and it's happy to consume them.
See this example: https://docs.splunk.com/Documentation/Splunk/7.1.1/Data/HTTPEventCollectortokenmanagement#Send_multiple_events_to_HEC_in_one_request
```
{"event": "Pony 1 has left the barn"}{"event": "Pony 2 has left the barn"}{"event": "Pony 3 has left the barn", "nested": {"key1": "value1"}}
```

plugins/serializers/splunkmetric/splunkmetric.go
```diff
@@ -0,0 +1,135 @@
...
+	 ** metric_name: The name of the metric
+	 ** _value:      The value for the metric
+	 ** _time:       The timestamp for the metric
+	 ** All other index fields become deminsions.
```
phemmer: I personally don't think this is a good structure to follow. This will prevent using this serializer with pretty much every single input plugin, as none of them follow this format. I think the implementation over on #4185 has a much more flexible implementation.

phemmer:
This is now our 3rd splunk output proposal in a few days :-/
See also:
#4185 
#4300

plugins/serializers/registry.go
ronnocol: Splunk absolutely support sub second timestamps. From the default datetime.xml:
```xml
<define name="_utcepoch" extract="utcepoch, subsecond">
        <!-- update regex before '2017' -->
    <text><![CDATA[((?<=^|[\s#,"=\(\[\|\{])(?:1[012345]|9)\d{8}|^@[\da-fA-F]{16,24})(?:\.?(\d{1,6}))?(?![\d\(])]]></text>
</define>```

Comment (phemmer) [reply plugins/serializers/registry.go]: I didn't say it doesn't. It only supports seconds as the unit. It supports more than that as the precision. Unit != precision.

File: plugins/serializers/splunkmetric/splunkmetric.go (hunk 3 above)
Comment (ronnocol) [reply plugins/serializers/splunkmetric/splunkmetric.go]: I agree, it's not line oriented, but it doesn't support reading an array of JSON objects as metrics. This is OK:
```javascript
{"event": "Pony 1 has left the barn"}{"event": "Pony 2 has left the barn"}{"event": "Pony 3 has left the barn", "nested": {"key1": "value1"}}
```

This is not OK:
```javascript
[{"event": "Pony 1 has left the barn"}{"event": "Pony 2 has left the barn"}{"event": "Pony 3 has left the barn", "nested": {"key1": "value1"}}]
```

phemmer: My point was that developer documentation should be accurate. We shouldn't tell the developers it's "one metric per line" if it's not.

plugins/serializers/splunkmetric/splunkmetric.go
ronnocol: The comment is wrong, it should be `time`, not `_time` (I'll fix), but `metric_name` and `_value` are required fields for the metrics store. Naming the fields this way means you don't need custom props.conf or transforms.conf. In fact this is the same naming convention that is used in the PR you reference. The point of this serializer is to format the metrics into this format so that it can be used with the generic, well tested, HTTP output. Or the file output if you're running telegraf on a machine that is running a Splunk forwarder.

ronnocol:
@phemmer Thank you for the comprehensive review, I appreciate all of the feedback and believe I have addressed all/most of the issues with this latest commit.

I know that there were some PRs for Splunk HEC outputs, but this was specifically written as a serializer so that you could have Splunk compatible metrics generated for any reasonable output (e.g. file or HTTP.) Furthermore, it will allow you to easily redirect different types of metrics to different metrics indexes, which the output based PRs do not allow for. The output based PRs also require that you have/deploy a HEC to be used to get metrics into the Splunk infrastructure vs. being able to use existing infrastructure (forwarders) to get the data into a metrics index.

dkulig:
Looks like it doesn't work properly with the `zookeeper` input, 

```
2018-07-19T12:50:41Z D! Attempting connection to output: file
2018-07-19T12:50:41Z D! Successfully connected to output: file
2018-07-19T12:50:41Z I! Starting Telegraf v1.8.0~caf97dc5
2018-07-19T12:50:41Z I! Loaded inputs: inputs.zookeeper
2018-07-19T12:50:41Z I! Loaded aggregators:
2018-07-19T12:50:41Z I! Loaded processors:
2018-07-19T12:50:41Z I! Loaded outputs: file
2018-07-19T12:50:41Z I! Tags enabled: host=mbpdk
2018-07-19T12:50:41Z I! Agent Config: Interval:10s, Quiet:false, Hostname:"mbpdk", Flush Interval:10s
2018-07-19T12:50:51Z D! Output [file] buffer fullness: 1 / 10000 metrics.
2018-07-19T12:50:51Z E! [serializer.splunkmetric] Dropping invalid metric
2018-07-19T12:50:51Z E! Error writing to output [file]: failed to serialize message: can not parse value
```

my config: 

```
[[outputs.file]]
   files = ["stdout"]
   data_format = "splunkmetric"
   hec_routing = false

[[inputs.zookeeper]]
  servers = [":2181"]
```

arohter:
This looks great, and 👍 for leveraging the existing http plugin.   Can we get this merged and into an official release?

docs/DATA_FORMATS_OUTPUT.md
```diff
@@ -209,3 +210,67 @@ reference the documentation for the specific plugin.
...
+
+## SplunkMetric
+
+The SplunkMetric format translates the telegraf metrics into a JSON format compatible
+with a Splunk Metrics index. Normally you would use this to send the metrics to a
+HTTP Event Collector (HEC), it follows the format specified in the document [here.](http://dev.splunk.com/view/SP-CAAAFDN#json)
+
+```toml
+[[outputs.http]]
+#   ## URL is the address to send metrics to
```
glinton: no need for the initial `#` through line 252

plugins/serializers/splunkmetric/README.md
```diff
@@ -0,0 +1,135 @@
...
+* user
+
+## Using with the HTTP output
+
+To send this data to a Splunk HEC, you can use the HTTP output, there are some custom headers that you need to add
+to manage the HEC authorization, here's a sample config for an HTTP output:
+
+```toml
+[[outputs.http]]
+#   ## URL is the address to send metrics to
```
glinton: again, leading `#` isn't necessary

plugins/serializers/splunkmetric/README.md
```diff
@@ -0,0 +1,135 @@
...
+INDEXED_EXTRACTIONS = json
+KV_MODE = none
+TIMESTAMP_FIELDS = time
+TIME_FORMAT = %s.%3N
+```
+
+An example configuration of a file based output is: 
+
+```toml
+# # Send telegraf metrics to file(s)
```
glinton: same `#`

plugins/serializers/splunkmetric/splunkmetric.go
```diff
@@ -0,0 +1,124 @@
+package splunkmetric
+
+import (
+	"encoding/json"
+	//	"errors"
```
glinton: remove this commented import

plugins/serializers/splunkmetric/splunkmetric.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+	return s, nil
+}
+
+func (s *serializer) Serialize(metric telegraf.Metric) ([]byte, error) {
+
+	m, err := s.createObject(metric)
+	if err != nil {
+		log.Printf("E! [serializer.splunkmetric] Dropping invalid metric: %v [%v]", metric, m)
+		return []byte(""), err
```
glinton: pedantic, but maybe return `nil, err` here

ronnocol:
@danielnelson For the sake of clarity, you want me to remove the text in DATA_FORMATS_OUTPUT and replace it with a link to the README. Possibly merging some of the text from that I put in the DATA_FORMAT_OUTPUT file into the README if additional clarity is required.

danielnelson:
That's right, thank you

plugins/serializers/registry.go
```diff
@@ -73,6 +77,8 @@ func NewSerializer(config *Config) (Serializer, error) {
 		serializer, err = NewGraphiteSerializer(config.Prefix, config.Template, config.GraphiteTagSupport)
 	case "json":
 		serializer, err = NewJsonSerializer(config.TimestampUnits)
+	case "splunkmetric":
```
danielnelson: If we renamed this "splunk" would it cause confusion about the format type?  Since all of the formats are for encoding metrics I feel like the metric suffix may be redundant.

plugins/serializers/splunkmetric/README.md
```diff
@@ -0,0 +1,139 @@
...
+   ## Use TLS but skip chain & host verification
+   # insecure_skip_verify = false
+
+   ## Data format to output.
+   ## Each data format has it's own unique set of configuration options, read
+   ## more about them here:
+   ## https://github.com/influxdata/telegraf/blob/master/docs/DATA_FORMATS_OUTPUT.md
+   data_format = "splunkmetric"
+    ## Provides time, index, source overrides for the HEC
+   hec_routing = true
```
danielnelson: Can you rename this variable `splunk_hec_routing`, due to how the parser variables are injected directly into the plugin we have started prefixing the variables.  Long term, these will probably become tables and the parsers will become more independent.

plugins/serializers/splunkmetric/splunkmetric.go
```diff
@@ -0,0 +1,125 @@
...
+		HecRouting: hec_routing,
+	}
+	return s, nil
+}
+
+func (s *serializer) Serialize(metric telegraf.Metric) ([]byte, error) {
+
+	m, err := s.createObject(metric)
+	if err != nil {
+		log.Printf("D! [serializer.splunkmetric] Dropping invalid metric: %v [%v]", metric, m)
```
danielnelson: We should return this as an error: `return nil, fmt.Errorf("Dropping invalid metric: %s", metric.Name())`

plugins/serializers/splunkmetric/splunkmetric.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+func (s *serializer) SerializeBatch(metrics []telegraf.Metric) ([]byte, error) {
+
+	var serialized []byte
+
+	for _, metric := range metrics {
+		m, err := s.createObject(metric)
+		if err != nil {
+			log.Printf("D! [serializer.splunkmetric] Dropping invalid metric: %v [%v]", metric, m)
```
danielnelson: Should return error.  If there are situations where a metric cannot be serialized but it is not an error, have createObject `return nil, nil` and check if m is nil before appending.

plugins/serializers/splunkmetric/splunkmetric.go
```diff
@@ -0,0 +1,125 @@
...
+		Source string                 `json:"source,omitempty"`
+		Fields map[string]interface{} `json:"fields"`
+	}
+
+	dataGroup := HECTimeSeries{}
+
+	for k, v := range metric.Fields() {
+
+		if !verifyValue(v) {
+			log.Printf("D! Can not parse value: %v for key: %v", v, k)
```
danielnelson: I think here we should just continue on the next field.  Otherwise we will never be able to serialize a metric with a string field.

plugins/serializers/splunkmetric/splunkmetric.go
```diff
@@ -0,0 +1,125 @@
...
+		Event  string                 `json:"event"`
+		Host   string                 `json:"host,omitempty"`
+		Index  string                 `json:"index,omitempty"`
+		Source string                 `json:"source,omitempty"`
+		Fields map[string]interface{} `json:"fields"`
+	}
+
+	dataGroup := HECTimeSeries{}
+
+	for k, v := range metric.Fields() {
```
danielnelson: Use metric.FieldList() so that no allocation is performed.

plugins/serializers/splunkmetric/splunkmetric.go
```diff
@@ -0,0 +1,125 @@
...
+			return nil, err
+		}
+
+		obj := map[string]interface{}{}
+		obj["metric_name"] = metric.Name() + "." + k
+		obj["_value"] = v
+
+		dataGroup.Event = "metric"
+		// Convert ns to float seconds since epoch.
+		dataGroup.Time = float64(metric.Time().UnixNano()) / float64(1000000000)
```
danielnelson: `metric.Time().Unix()` will give you unix time in seconds.

plugins/serializers/splunkmetric/splunkmetric.go
```diff
@@ -0,0 +1,125 @@
...
+		} else {
+			serialized = append(serialized, m...)
+		}
+	}
+
+	return serialized, nil
+}
+
+func (s *serializer) createObject(metric telegraf.Metric) (metricJson []byte, err error) {
+
```
danielnelson: We do a bad job of explaining what is guaranteed in a telegraf.Metric, and actually as I write this I notice some additional checks I need to add.

Here is a brief rundown of things you may or may not need to check:
- metric name may be an empty string
- zero or more tags, tag keys are any non-empty string, tag values may be empty strings
- zero, yes zero, or more fields, field keys are any non-empty string, field values may be any float64 (including NaN, +/-Inf),int64,uint64,string,bool
- time is any time.Time.

The part about tag/field keys not being empty strings is not true right now, but after writing this I am going to ensure this is the case in 1.8.

plugins/serializers/registry.go
ronnocol: I named it splunkmetric because Splunk has two (2) different index types, events and metrics. This is more of an indication that telegraf is feeding a Splunk metrics index vs. an events index. I'd be OK changing it if you really want to.

plugins/serializers/splunkmetric/README.md
ronnocol: will do

plugins/serializers/splunkmetric/splunkmetric.go
ronnocol: I don't want seconds, I want ms but as a float (this is Splunk's spec...) e.g. 1536295485.123

plugins/serializers/registry.go
danielnelson: Thanks for the explanation, I think your decision to call it `splunkmetric` is the right one then.

plugins/serializers/splunkmetric/README.md
danielnelson: Just to keep things consistent, lets call it `splunkmetric_hec_routing`

plugins/serializers/splunkmetric/splunkmetric.go
ronnocol: Thanks, this actually caused me to re-examine the the serializer with several different inputs, and I found a case in which metrics were lost (dropped), so I'm refactoring some of the code to deal with that. (Also, it's been a crazy week...so hope to get this done over the next few days.)

plugins/serializers/splunkmetric/splunkmetric.go
ronnocol: Part of the refactoring mentioned below (thank for the suggestion)

plugins/serializers/splunkmetric/splunkmetric.go
ronnocol: Got it. Will include in the other changes being made.

plugins/serializers/splunkmetric/README.md
ronnocol: ok, will do.

plugins/serializers/registry.go
ronnocol: 👍

plugins/serializers/splunkmetric/README.md
```diff
@@ -0,0 +1,139 @@
+# Splunk Metrics serialzier
```
russorat: spelling

danielnelson:
@ronnocol We are hoping to release 1.8.0-rc1 on Wednesday afternoon, let me know if you think that could be a problem for this pr.

PR #4345, opened 2018-06-26
Allow for force gathering ES cluster stats

PierreF:
Add option to gather Elasticsearch cluster stats from all node (default is old behavior, stats are only gathered from master).

Also documented the fact that gathering only from master node only works if local = true.
Slightly changed behavior when local=false and cluster_stats=true. Previous behavior was undetermined (it gather cluster stats only if the last node is the master node. The order was based on the order of a map which is probably not fixed). New behavior is the documented one: if local = false, we gather from all node.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -113,7 +113,6 @@
 
 ### Features
 
-- [#4345](https://github.com/influxdata/telegraf/pull/4345): Allow to force gathering Elasticsearch cluster stats on non-master node.
 - [#4236](https://github.com/influxdata/telegraf/pull/4236): Add SSL/TLS support to redis input.
 - [#4160](https://github.com/influxdata/telegraf/pull/4160): Add tengine input plugin.
 - [#4262](https://github.com/influxdata/telegraf/pull/4262): Add power draw field to nvidia_smi plugin.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -113,6 +113,7 @@
 
 ### Features
 
+- [#4345](https://github.com/influxdata/telegraf/pull/4345): Allow to force gathering Elasticsearch cluster stats on non-master node.
 - [#4236](https://github.com/influxdata/telegraf/pull/4236): Add SSL/TLS support to redis input.
 - [#4160](https://github.com/influxdata/telegraf/pull/4160): Add tengine input plugin.
 - [#4262](https://github.com/influxdata/telegraf/pull/4262): Add power draw field to nvidia_smi plugin.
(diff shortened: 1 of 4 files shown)


PierreF:
I don't think the circleci error is due to my change. It's failing on dep ensure but I didn't changed anything on dependencies. I suspect a timeout while fetching dependencies.
Can I retry the circleci jobs, if yes how ?

glinton:
I'm thinking we'll update [these lines](https://github.com/influxdata/telegraf/blob/master/.circleci/config.yml#L19-L23) to be `vendor-{{ checksum "Gopkg.lock" }}` so there is something to start from. It is indeed a timeout while fetching dependencies.

danielnelson:
The setting of `isMaster` has a long time race condition: #2650

@PierreF Is there a reason we can't collect cluster_stats when `local=false`?

PierreF:
We can always collect cluster_stats, and (with this PR) as soon as cluster_stats = True, they will be collected.
We can't however collect them only from the master when local=false. This is because we determine the master when we query the node stats, if we does only for the local node (local=true), we can know if the local node is the master. But when local=false and we query all nodes, we have no way to know which node is the local one.

danielnelson:
Makes sense, do you think it is possible for us to know which host is the one we connected to when using `local=true` and `/_nodes/stats`?  I feel like this would be helpful in simplifying the config.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 
 ### Features
 
+- [#4345](https://github.com/influxdata/telegraf/pull/4345): Allow to force gathering Elasticsearch cluster stats on non-master node.
```
glinton: Only suggestion is to remove update to the changelog to prevent merge conflict

PierreF:
Sorry for the delay. I've removed the changelog entry to avoid merge conflict.

About @danielnelson query: the behavior (the PR didn't change this) is:
* with local=true, query /_nodes/_local/stats which include only stats of Elastic node Telegraf is currently connected to
* with local=false, query /_nodes/stats which include stats of all Elastic nodes in the cluster. The output is the same on every nodes, so we don't have any way to find to which node Telegraf connected to.

We could do an additional query on /_node/_local to known the current node ID, and the query /_nodes/stats, but that means one additional query at each gather.

PierreF:
Do you want me to do the additional query on each gather to simplify configuration ? Or is it anything else missing to merge this PR ?

danielnelson:
Let's go with this for now.  I wonder if we should consider dropping local=false mode completely though, I feel like the flexibility of this plugin is complicating things and it might be better to just have one clear way to monitor the cluster.

PR #4347, opened 2018-06-26
Add http path configuration for OpenTSDB plugin

jtlisi:
This PR updates the OpenTSDB to allow for the path to be configured when using the HTTP client. Currently there is no way for telegraf to send data to OpenTSDB servers behind a reverse proxy with non default routes.

### Required for all PRs:

### DIFF
--- a/plugins/outputs/opentsdb/opentsdb_http.go
+++ b/plugins/outputs/opentsdb/opentsdb_http.go
@@ -26,6 +26,7 @@ type openTSDBHttp struct {
 	Scheme    string
 	User      *url.Userinfo
 	BatchSize int
+	Path      string
 	Debug     bool
 
 	metricCounter int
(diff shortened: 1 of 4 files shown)


plugins/outputs/opentsdb/opentsdb_http.go
```diff
@@ -123,7 +124,7 @@ func (o *openTSDBHttp) flush() error {
 		Scheme: o.Scheme,
 		User:   o.User,
 		Host:   fmt.Sprintf("%s:%d", o.Host, o.Port),
-		Path:   "/api/put",
+		Path:   o.Path,
```
glinton: Can you default `o.Path` to `"/api/put"` in order to not break any current implementations.

danielnelson:
After merging I switched the config to use snake_case format, either format will work for now since there is no field tags but this is the preferred style.  2ee374cf501e4ee9951108a61ebd2d15bb08e7d8

Issue #2986, opened 2017-07-05
Make timestamp configurable for JSON parser


bbczeuz:
### Proposal:
- Add two configuration options to the JSON parser (e.g. socket_listener input plugin) that allow the timestamp to be collected from the data itself. e.g:
```
time_key = "time"
time_unit = "us"
```

### Current Behavior:
- The timestamp in the JSON parser is currently being set by the parser (https://github.com/influxdata/telegraf/blob/master/plugins/parsers/json/parser.go#L61)
- With some hack in Kapacitor we might be able to use the `time` field as timestamp...

### Desired Behavior:
- Keep the original timestamp of the data

### Use Case:
- I plan to use telegraf to capture Juniper Network Analytics data. Supporting devices send interface/buffer metrics in JSON or GPB via UDP or TCP to remote devices.
- This would replace snmp polling reducing CPU load and latency
- The existing `socket_listener` input plugin supports the JSON format and tag_keys to process most data
- The original JSON data contain a `time` field with the timestamp of the measurement in microseconds since the beginning of the current Epoch (http://www.juniper.net/documentation/en_US/junos/information-products/pathway-pages/ex4300/network-management.pdf)

### Example Data
```
{"record-type":"traffic-stats","time":1499268651471993,"router-id":"qfx5100-01","port":"et-0/0/50","rxpkt":82227,"rxucpkt":74232,"rxmcpkt":7985,"rxbcpkt":10,"rxpps":11,"rxbyte":6254098,"rxbps":55232,"rxdroppkt":0,"rxcrcerr":0,"txpkt":104340,"txucpkt":96366,"txmcpkt":7955,"txbcpkt":19,"txpps":13,"txbyte":9313288,"txbps":71552,"txdroppkt":0,"txcrcerr":0}
```
### Proposed Example Configuration
```
[[inputs.socket_listener]]
    service_address = "udp://:50019"
    data_format = "json"
    tag_keys = [
        "router-id",
        "port",
        "record-type"
    time_key = "time"
    time_unit = "us"
```

danielnelson:
Lets call it `json_timestamp_units` and `json_timestamp_key` to match existing options in the json serializer.

srclosson:
+1, this would be very helpful for me, trying to access OPC data through a restful api.

petetnt:
This issue proved to be a blocker for use case of Telegraf, would you be interested in a PR?

danielnelson:
@petetnt Absolutely

[...]

PR #4351, opened 2018-06-27
Feature/1363: Additional configuration for JSON parser

maxunt:
closes: https://github.com/influxdata/telegraf/issues/2986
### Required for all PRs:


plugins/parsers/registry.go
```diff
@@ -129,11 +131,13 @@ func NewParser(config *Config) (Parser, error) {
 func NewJSONParser(
 	metricName string,
 	tagKeys []string,
+	fieldKeys []string,
```
glinton: Generally we want to avoid changing exported function signatures, as it destabilizes packages. Perhaps what we could do is add a `newJSONParser` that takes different (or functional) parameters. We could then update those things that call `NewJSONParser` directly to call `NewParser`, which will in turn call `newJSONParser`.

danielnelson:
@maxunt As we discussed, let's have this be `string_keys` and list flattened paths to values that should be added as string fields.

plugins/parsers/json/parser.go
```diff
@@ -54,10 +59,28 @@ func (p *JSONParser) parseObject(metrics []telegraf.Metric, jsonOut map[string]i
 		return nil, err
 	}
 
-	tags, nFields := p.switchFieldToTag(tags, f.Fields)
+	//if time key is specified, set it to time
+	nTime := time.Now().UTC()
+	if p.JSONTimeKey != "" {
+		if p.JSONTimeFormat == "" {
+			err := fmt.Errorf("E! If 'json_time_key' is specified, there must be a 'json_time_format'")
```
danielnelson: Since this is an error and not a log message, don't include the `E!`.  As a nitpick, it would probably be good to try to come up with a shorter error string, perhaps: `use of 'json_time_key' requires 'json_time_format'`?

Remember to do this for the other errors, here is some more info on how "standard" Go error strings should look: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

plugins/parsers/json/parser.go
```diff
@@ -114,6 +137,7 @@ func (p *JSONParser) switchFieldToTag(tags map[string]string, fields map[string]
 
 func (p *JSONParser) Parse(buf []byte) ([]telegraf.Metric, error) {
 
+	//if json_query is specified
```
danielnelson: Remove this comment, it doesn't provide any new information on why and the what can be determined by looking at the code.

docs/DATA_FORMATS_INPUT.md
```diff
@@ -124,11 +138,25 @@ For example, if you had this configuration:
...
+  ## List of tag names to extract from JSON server response
   tag_keys = [
     "my_tag_1",
     "my_tag_2"
   ]
+
+  ## List of field names to extract from JSON and add as string fields
+  # string_fields = []
+
+  ## gjson query path to specify a specific chunk of JSON to be parsed with the above configuration
```
danielnelson: Wrap comments before 78 chars for readability.

plugins/parsers/json/parser.go
```diff
@@ -51,10 +59,28 @@ func (p *JSONParser) parseObject(metrics []telegraf.Metric, jsonOut map[string]i
...
+			return nil, err
+		}
 
-	metric, err := metric.New(p.MetricName, tags, nFields, time.Now().UTC())
+		if f.Fields[p.JSONTimeKey] == nil {
+			err := fmt.Errorf("E! JSON time key could not be found")
+			return nil, err
+		}
+
+		nTime, err = time.Parse(p.JSONTimeFormat, f.Fields[p.JSONTimeKey].(string))
```
danielnelson: Make sure to test the type assertion since it comes from input and could be any type.

plugins/parsers/json/parser_test.go
```diff
@@ -448,22 +449,148 @@ func TestJSONParseNestedArray(t *testing.T) {
...
+	}`
+
+	parser := JSONParser{
+		MetricName: "json_test",
+		TagKeys:    []string{},
+		JSONQuery:  "shares.myArr",
+	}
+
+	_, err := parser.Parse([]byte(testString))
+	assert.Error(t, err)
```
danielnelson: Use require instead of assert (almost) everywhere.  Doesn't really matter here but this will prevent confusing chained errors.

plugins/parsers/json/parser_test.go
```diff
@@ -448,22 +449,148 @@ func TestJSONParseNestedArray(t *testing.T) {
...
+		MetricName:   "json_test",
+		StringFields: []string{"last"},
+		TagKeys:      []string{"first"},
+		JSONQuery:    "obj.friends",
+	}
+
+	metrics, err := parser.Parse([]byte(testString))
+	assert.NoError(t, err)
+	assert.Equal(t, metrics[0].Fields()["last"], "Murphy")
+	assert.Equal(t, 3, len(metrics))
```
danielnelson: Check the length before indexing, this will make the test easier to understand if it fails.

plugins/parsers/json/parser_test.go
```diff
@@ -448,22 +449,148 @@ func TestJSONParseNestedArray(t *testing.T) {
...
+	]`
+
+	parser := JSONParser{
+		MetricName:     "json_test",
+		JSONTimeKey:    "b_time",
+		JSONTimeFormat: "02 Jan 06 15:04 MST",
+	}
+	metrics, err := parser.Parse([]byte(testString))
+	assert.NoError(t, err)
+	assert.Equal(t, false, metrics[0].Time() == metrics[1].Time())
```
danielnelson: Check len(metrics) first, add some more tests to check error cases.

docs/DATA_FORMATS_INPUT.md
```diff
@@ -104,9 +104,23 @@ but can be overridden using the `name_override` config option.
...
+If this path leads to an array of values or single data point an error will be thrown.  If this 
+configuration is specified, only the result of the query will be parsed and returned as metrics.  
+
+Object paths are specified using gjson path format, which is denoted by object keys 
+concatenated with "." to go deeper in nested JSON objects.  
+Additional information on gjson paths can be found here: https://github.com/tidwall/gjson#path-syntax
+
+The JSON data format also supports extracting time values through the config "json_time_key" and "json_time_format". 
+If "json_time_key" is set, "json_time_format" must be specified.  The "json_time_key" describes the name of the field containing time information.  The "json_time_format" must be a recognized Go time format.  
+More info on time formats can be found here: https://golang.org/src/time/format.go
```
danielnelson: Use `https://golang.org/pkg/time/#Parse` as the link.  This other link is still not great but the audience of this documentation is not Go developers.

Try to wrap around 78 chars

docs/DATA_FORMATS_INPUT.md
```diff
@@ -125,11 +143,25 @@ For example, if you had this configuration:
...
     "my_tag_1",
     "my_tag_2"
   ]
+
+  ## List of field names to extract from JSON and add as string fields
+  # string_fields = []
+
+  ## gjson query path to specify a specific chunk of JSON to be parsed with 
+  ## the above configuration. If not specified, the whole file will be parsed. 
+  ## gjson query paths are described here:
```
glinton: Add link to where gjson query paths are described

plugins/parsers/registry.go
```diff
@@ -57,9 +57,20 @@ type Config struct {
 
 	// TagKeys only apply to JSON data
 	TagKeys []string
+	// FieldKeys only apply to JSON
+	StringFields []string
 	// MetricName applies to JSON & value. This will be the name of the measurement.
 	MetricName string
 
+	//holds a gjson path for json parser
```
glinton: When adding comments, add a space before starting comment text (`// This is my comment`)

plugins/parsers/registry.go
```diff
@@ -145,6 +161,28 @@ func NewParser(config *Config) (Parser, error) {
...
 
+func newJSONParser(
+	metricName string,
+	tagKeys []string,
+	stringFields []string,
+	jsonQuery string,
+	timeKey string,
+	timeFormat string,
+	defaultTags map[string]string,
+) (Parser, error) {
```
glinton: Returning an error isn't useful here. Where it's an unexported function not matching any interface, I'd remove it, or remove the function altogether and just instantiate a `JSONParser` on line 120

plugins/parsers/json/parser.go
```diff
@@ -126,7 +177,7 @@ func (p *JSONParser) ParseLine(line string) (telegraf.Metric, error) {
 	}
 
 	if len(metrics) < 1 {
-		return nil, fmt.Errorf("Can not parse the line: %s, for data format: influx ", line)
+		return nil, fmt.Errorf("can not parse the line: %s, for data format: influx ", line)
```
glinton: Also the copypasta, `s/influx/json`

plugins/parsers/registry.go
maxunt: No error is ever returned by the new_Parser functions, but i thought i'd include it here to match the format of the rest of the parsers

plugins/parsers/json/parser.go
maxunt: could you clarify what you mean here?

glinton: Replace "influx" with "json", it appears to be copy paste code

maxunt: ah i see. good catch thanks

docs/DATA_FORMATS_INPUT.md
```diff
@@ -174,6 +207,19 @@ For example, if the following configuration:
...
+
+  ## gjson query path to specify a specific chunk of JSON to be parsed with 
+  ## the above configuration. If not specified, the whole file will be parsed
+  # json_query = ""
+
+  ## holds the name of the tag of timestamp
+  json_time_key = "b_time"
+
+  ## holds the format of timestamp to be parsed
+  json_time_format = "02 Jan 06 15:04 MST"
```
danielnelson: This is an interesting example time format because there it doesn't include a year.  This results in a pared time that is in year 0.  I think after parsing the time we should do something similar to what we did in logparser and use the current year.  While this would be pretty annoying if you actually want to store in a metric from year 0, I think it is much more likely to be what is wanted:
```
if ts.Year() == 0 {
	ts = ts.AddDate(timestamp.Year(), 0, 0)
}
```

PR #4354, opened 2018-06-28
Improve passthrough in prometheus output and input plugins

mmolnar:
Prometheus metrics will get mangled when passed to another telegraf instance using any of the serializers due to the loss of type. This is makes the histogram and summary metrics unusable in prometheus.

This change will store prometheus help and type as metric tags in prometheus input and restore them in prometheus output.

Let me know what you think about this change. My go knowledge is mostly superficial as my main focus is system administration, I will wellcome a code review.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -968,6 +968,6 @@
 [solve-meta]
   analyzer-name = "dep"
   analyzer-version = 1
-  inputs-digest = "024194b983d91b9500fe97e0aa0ddb5fe725030cb51ddfb034e386cae1098370"
+  inputs-digest = "6694ba5b3f987292a075a845c2a2188020c491eb03c2b5eae4f733166f6fe654"
   solver-name = "gps-cdcl"
   solver-version = 1
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -109,10 +109,6 @@
   branch = "master"
 
 [[constraint]]
-  name = "github.com/matttproud/golang_protobuf_extensions"
-  version = "1.0.1"
-
-[[constraint]]
   name = "github.com/Microsoft/ApplicationInsights-Go"
   branch = "master"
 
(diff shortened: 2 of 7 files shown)


danielnelson:
The preservation of TYPE support should be working already, can you open a new issue if you are still having problems in 1.7 or newer?

I would like to add support for HELP at some points, but I don't think we should pass it through as a tag, and instead it should become a new piece of metadata that can be stored with the field, similar in many ways to #3670.  This issue has been a little trickier to implement than desired, as I want to find a memory efficient way to make these optional.

mmolnar:
Serializers do not support preservation of TYPE information so this information is lost when using socket writer/socket listener. You are right that it works inside single instance of telegraf, but when using a chain of telegraf instances then the metrics are mangled.

<prometheus input> -> <telegraf> -> <socket writer> -> [loss of TYPE] -> <socket_listener> -> <telegraf> -> <prometheus output>

In this chain the mertics in prometheus output are mangled and mainly the histogram and summary metrics are unusable inside prometheus because histogram_quantile() function expects the data to be in this format:
```
streaming_storage_latency_seconds_bucket{le="0.05"} 56
streaming_storage_latency_seconds_bucket{le="0.1"} 83
streaming_storage_latency_seconds_bucket{le="0.5"} 143
streaming_storage_latency_seconds_bucket{le="1"} 199
streaming_storage_latency_seconds_bucket{le="5"} 215
streaming_storage_latency_seconds_bucket{le="10"} 215
streaming_storage_latency_seconds_bucket{le="+Inf"} 215
streaming_storage_latency_seconds_sum 40.25
streaming_storage_latency_seconds_count 215
```

Instead the metrics in prometheus output look like this:
```
streaming_storage_latency_seconds_0_05 56
streaming_storage_latency_seconds_0_01 83
streaming_storage_latency_seconds_0_5 143
streaming_storage_latency_seconds_1 199
streaming_storage_latency_seconds_5 215
streaming_storage_latency_seconds_10 215
streaming_storage_latency_seconds_inf 215
streaming_storage_latency_seconds_sum 40.25
streaming_storage_latency_seconds_count 215
```

If you plan to add TYPE and HELP metadata to the serializers then I will look forward to this feature, I just hope it will not be too long until it is implemented.

PR #4363, opened 2018-06-30
Add filecount input plugin

sometimesfood:
This pull request introduces a new input plugin named `filecount`. The `filecount` plugin counts files in a directory that match certain criteria (e.g. "name", "size", "mtime" etc.).

While `filecount` is modeled after the [collectd filecount plugin](https://collectd.org/documentation/manpages/collectd.conf.5.shtml#plugin_filecount), the two plugins do not share any code.

### Required for all PRs:

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,6 +158,7 @@ configuration options.
 * [fibaro](./plugins/inputs/fibaro)
 * [file](./plugins/inputs/file)
 * [filestat](./plugins/inputs/filestat)
+* [filecount](./plugins/inputs/filecount)
 * [fluentd](./plugins/inputs/fluentd)
 * [graylog](./plugins/inputs/graylog)
 * [haproxy](./plugins/inputs/haproxy)
--- a/plugins/inputs/all/all.go
+++ b/plugins/inputs/all/all.go
@@ -31,6 +31,7 @@ import (
 	_ "github.com/influxdata/telegraf/plugins/inputs/fail2ban"
 	_ "github.com/influxdata/telegraf/plugins/inputs/fibaro"
 	_ "github.com/influxdata/telegraf/plugins/inputs/file"
+	_ "github.com/influxdata/telegraf/plugins/inputs/filecount"
 	_ "github.com/influxdata/telegraf/plugins/inputs/filestat"
 	_ "github.com/influxdata/telegraf/plugins/inputs/fluentd"
 	_ "github.com/influxdata/telegraf/plugins/inputs/graylog"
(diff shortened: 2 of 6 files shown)


danielnelson:
This will be handy, but I think the implementation should use `filepath.Walk` to reduce the amount of code.

sometimesfood:
You are right of course. Sorry, I sat on this code for a while and completely forgot that I had planned on refactoring it to use `filepath.Walk` before submitting a pull request.

I'll clean up the required changes later this week.

phemmer:
Since this plugin uses things like "size" and "mtime" as filter parameters, should the related metrics be exported as fields? For example: "min size found", "max size found", "total size", "oldest age", "newest age".
And then if this is done, maybe the [postfix plugin](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/postfix) could be re-written as a wrapper around this one.

danielnelson:
This seems nice, I was a little uneasy with a one field plugin.  Here are some proposed field names for the metrics you mentioned, but I think it would be fine to add these either before or after merging this PR.
- count
- size_bytes
- file_size_bytes_min
- file_size_bytes_max
- file_mtime_min
- file_mtime_max

phemmer:
Is the proposed `file_mtime_min` (and `_max`) an absolute timestamp, or an offset? I don't know that an absolute time would be of use, but maybe.

sometimesfood:
I'm not a huge fan of adding hardcoded min/max fields. The original collectd filecount plugin has an additional aggregate file size field (equivalent to @danielnelson's `size_bytes` field) and I think that would make sense here as well.

Regarding the postfix plugin, I think it would be a better idea to extract shared code and move it into a separate helper that could then be used by both plugins rather than adding all the fields needed by the postfix plugin to the filecount plugin. I could open another feature branch so we can give this a try and see how it would improve the existing codebase.

However, if it's OK with you, I think it would be a good idea to merge this pull request first and add new fields/extract shared code/do whatever you decide to do in a separate pull request. None of the proposed changes would break any existing interfaces and it would be nice to keep the scope of this pull request as small as possible.

danielnelson:
@sometimesfood Can you fix the conflict or is it alright if I push to your branch?

sometimesfood:
@danielnelson Should be fixed now. Sorry, I didn't see your thumbs up reaction; GitHub doesn't send out notifications for those.

danielnelson:
Yes, that will work great.

sometimesfood:
Thanks, done.

PR #4364, opened 2018-06-30
Support StatisticValues in cloudwatch output plugin (#4318)

### LINKED ISSUE #4318: Support StatisticValues in Cloudwatch output plugin
## Feature Request
Cloudwatch MericDatum has the StatisticValues field. We could utilize it with basicstats aggregator results so that we don't need to send raw metrics which could improve efficiency and save API cost.

### Proposal:
Cloudwatch output plugin could parse the field names from basicstats aggregator. If it found all required fields are available(min, max, sum, mean, count), it would use StatisticValues field to send this metric instead of sending them as separated values.

### Current behavior:
Cloudwatch output plugin does not use the StatisticValues field. It would send every field as separated cloudwatch.MericDatum. 

### Desired behavior:
Cloudwatch output plugin could utilize StatisticValues field if all required fields are available.

If we have lots of matrics, it is much more efficient and also saves API cost.

david7482:
Fix #4318 

### Required for all PRs:

Add `enable_statistic_values` flag. If true, this plugin would try to parse required statistic fields and send StatisticSet to Cloudwatch.

### DIFF
--- a/plugins/outputs/cloudwatch/README.md
+++ b/plugins/outputs/cloudwatch/README.md
@@ -36,3 +36,12 @@ Examples include but are not limited to:
 ### namespace
 
 The namespace used for AWS CloudWatch metrics.
+
+### enable_statistic_values
+
+If you have a large amount of metrics, you should consider to send 
+statistic values instead of raw metrics. This would not only improve
+performance but also save AWS API cost. Use `basicstats` aggregator to 
+calculate those required statistic fields (count, min, max, and sum).
+[See Cloudwatch StatisticSet](https://docs.aws.amazon.com/sdk-for-go/api/service/cloudwatch/#StatisticSet).
+This plugin would try to parse those statistic fields and send to Cloudwatch.
--- a/plugins/outputs/cloudwatch/README.md
+++ b/plugins/outputs/cloudwatch/README.md
@@ -37,11 +37,12 @@ Examples include but are not limited to:
 
 The namespace used for AWS CloudWatch metrics.
 
-### enable_statistic_values
+### write_statistics
 
-If you have a large amount of metrics, you should consider to send 
-statistic values instead of raw metrics. This would not only improve
-performance but also save AWS API cost. Use `basicstats` aggregator to 
-calculate those required statistic fields (count, min, max, and sum).
-[See Cloudwatch StatisticSet](https://docs.aws.amazon.com/sdk-for-go/api/service/cloudwatch/#StatisticSet).
-This plugin would try to parse those statistic fields and send to Cloudwatch.
+If you have a large amount of metrics, you should consider to send statistic 
+values instead of raw metrics which could not only improve performance but 
+also save AWS API cost. If enable this flag, this plugin would parse the required 
+[CloudWatch statistic fields](https://docs.aws.amazon.com/sdk-for-go/api/service/cloudwatch/#StatisticSet) 
+(count, min, max, and sum) and send them to CloudWatch. You could use `basicstats` 
+aggregator to calculate those fields. If not all statistic fields are available, 
+all fields would still be sent as raw metrics.
--- a/plugins/outputs/cloudwatch/cloudwatch.go
+++ b/plugins/outputs/cloudwatch/cloudwatch.go
@@ -141,21 +141,6 @@ func (f *valueField) addValue(sType statisticType, value float64) {
 
 func (f *valueField) buildDatum() []*cloudwatch.MetricDatum {
 
-	// Do CloudWatch boundary checking
-	// Constraints at: http://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_MetricDatum.html
-	if math.IsNaN(f.value) {
-		return nil
-	}
-	if math.IsInf(f.value, 0) {
-		return nil
-	}
-	if f.value > 0 && f.value < float64(8.515920e-109) {
-		return nil
-	}
-	if f.value > float64(1.174271e+108) {
-		return nil
-	}
-
 	return []*cloudwatch.MetricDatum{
 		{
 			MetricName: aws.String(strings.Join([]string{f.metricName, f.fieldName}, "_")),
@@ -307,7 +292,7 @@ func BuildMetricDatum(buildStatistic bool, point telegraf.Metric) []*cloudwatch.
 
 		val, ok := convert(v)
 		if !ok {
-			// Only fields with values that can be converted to float64 are supported.
+			// Only fields with values that can be converted to float64 (and within CloudWatch boundary) are supported.
 			// Non-supported fields are skipped.
 			continue
 		}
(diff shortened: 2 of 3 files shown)


plugins/outputs/cloudwatch/cloudwatch.go
```diff
@@ -50,6 +60,14 @@ var sampleConfig = `
...
   ## Namespace for the CloudWatch MetricDatums
   namespace = "InfluxData/Telegraf"
+
+  ## If you have a large amount of metrics, you should consider to send 
+  ## statistic values instead of raw metrics. This would not only improve
+  ## performance but also save AWS API cost. Use basicstats aggregator to
+  ## calculate required statistic fields (count, min, max, and sum) and 
+  ## enable this flag. This plugin would try to parse those fields and 
+  ## send statistic values to Cloudwatch.
+  # enable_statistic_values = false
```
danielnelson: Let's call this option `write_statistics`, this is little shorter, but also communicates that it is something that affects the output data.  So long as you are in here, can you mention in the comment that if a field does not have the required statistic fields then it will still be sent as a custom/raw metric.

plugins/outputs/cloudwatch/cloudwatch.go
```diff
@@ -161,9 +179,28 @@ func PartitionDatums(size int, datums []*cloudwatch.MetricDatum) [][]*cloudwatch
...
+// Otherwise, it would make MetricDatum from each field in a Point.
+func BuildMetricDatum(buildStatistic bool, point telegraf.Metric) []*cloudwatch.MetricDatum {
+
+	// If not enable, just take all metrics as value datums.
+	if !buildStatistic {
+		return BuildValueMetricDatum(point)
+	}
+
+	// Try to parse statisticSet first, then build statistic/value datum accordingly.
+	set, ok := getStatisticSet(point)
```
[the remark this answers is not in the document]
david7482: If getStatisticSet() could not get all required fields, it just keeps the original behavior and sends them as independent metrics. It would generate 4 metrics to cloudwatch: cpu_foo_min, cpu_foo_max, cpu_bar_min and cpu_bar_max.

plugins/outputs/cloudwatch/cloudwatch.go
david7482: okay, will update the field name and comments

plugins/outputs/cloudwatch/cloudwatch.go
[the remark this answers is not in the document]
danielnelson: Sorry, my example was somewhat lazy.  What if there are two full sets of statistics in a single `telegraf.Metric`, does it handle this case?
```
cpu foo_min=1,foo_max=2,foo_sum=4,foo_count=2,bar_min=1,bar_max=2,bar_sum=4,bar_count=2
```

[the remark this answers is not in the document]
david7482: Good point. Will support this multiple sets case.

david7482:
Already supported multiple statistic sets.

david7482:
Requested changes are fixed.

plugins/outputs/cloudwatch/README.md
```diff
@@ -36,3 +36,13 @@ Examples include but are not limited to:
 ### namespace
 
 The namespace used for AWS CloudWatch metrics.
+
+### write_statistics
+
+If you have a large amount of metrics, you should consider to send statistic
```
glinton: Some minor grammar:
 - s/consider to send/consider sending/
 - s/ which could/. This should/
 - s/If enable this flag/If this flag is enabled/

plugins/outputs/cloudwatch/cloudwatch.go
```diff
@@ -27,6 +27,128 @@ type CloudWatch struct {
...
+	}
+}
+
+func (f *statisticField) buildDatum() []*cloudwatch.MetricDatum {
+
+	var datums []*cloudwatch.MetricDatum
+
+	if f.hasAllFields() {
+		// If we have all required fields, we build datum with StatisticValues
+		min, _ := f.values[statisticTypeMin]
```
glinton: You don't need to change this, but just using `min := f.values[statisticTypeMin]` is sufficient, since you are ignoring the bool value anyways.

plugins/outputs/cloudwatch/cloudwatch.go
```diff
@@ -50,6 +172,14 @@ var sampleConfig = `
 
   ## Namespace for the CloudWatch MetricDatums
   namespace = "InfluxData/Telegraf"
+
+  ## If you have a large amount of metrics, you should consider to send statistic
```
glinton: Same grammar suggestions as above.

plugins/outputs/cloudwatch/cloudwatch.go
```diff
@@ -151,67 +281,58 @@ func PartitionDatums(size int, datums []*cloudwatch.MetricDatum) [][]*cloudwatch
...
 
-// Make a MetricDatum for each field in a Point. Only fields with values that can be
-// converted to float64 are supported. Non-supported fields are skipped.
-func BuildMetricDatum(point telegraf.Metric) []*cloudwatch.MetricDatum {
-	datums := make([]*cloudwatch.MetricDatum, len(point.Fields()))
-	i := 0
+// Make a MetricDatum from telegraf.Metric. It would check if all required fields of
+// cloudwatch.StatisticSet are available. If so, it would build MetricDatum from statistic values.
+// Otherwise, fields would still been built independently.
+func BuildMetricDatum(buildStatistic bool, point telegraf.Metric) []*cloudwatch.MetricDatum {
```
glinton: In general, changing an exported function's signature is advised against. Since nothing outside of this package uses it, maybe add a note that `BuildMetricDatum(telegraf.Metric)` is deprecated and create a new (unexported) `buildMetricDatum(bool, Metric)` function.

plugins/outputs/cloudwatch/cloudwatch.go
```diff
@@ -260,6 +381,72 @@ func BuildDimensions(mTags map[string]string) []*cloudwatch.Dimension {
...
+		value = float64(t)
+	case int64:
+		value = float64(t)
+	case uint64:
+		value = float64(t)
+	case float64:
+		value = t
+	case bool:
+		if t {
+			value = 1
```
glinton: I see this is how it was before, but should this be `float64(1)` and `float64(0)` in order to preserve the type when sending to an output? cc @danielnelson

plugins/outputs/cloudwatch/cloudwatch.go
danielnelson: This isn't something we are concerned about right now, maybe someday but not now.  The only API we provide stability for is the TOML file.  I keep meaning to document this, but haven't gotten to it.

plugins/outputs/cloudwatch/cloudwatch.go
danielnelson: This looks okay to me, value is a float64 and we are assigning a const to it.

plugins/outputs/cloudwatch/cloudwatch.go
david7482: If this is not an issue, I could help to `unexport` all exported functions to avoid confusion.
How do you think?

PR #4369, opened 2018-07-02
Improve syslog connection handling

### LINKED ISSUE #4335: Telegraf Syslog plugin.
### Relevant telegraf.conf:
```
[[inputs.syslog]]
#   ## Specify an ip or hostname with port - eg., tcp://localhost:6514, tcp://10.0.0.1:6514
#   ## Protocol, address and port to host the syslog receiver.
#   ## If no host is specified, then localhost is used.
#   ## If no port is specified, 6514 is used (RFC5425#section-4.1).
   server = "tcp://localhost:6514"
#
#   ## TLS Config
#   # tls_allowed_cacerts = ["/etc/telegraf/ca.pem"]
#   # tls_cert = "/etc/telegraf/cert.pem"
#   # tls_key = "/etc/telegraf/key.pem"
#
#   ## Period between keep alive probes.
#   ## 0 disables keep alive probes.
#   ## Defaults to the OS configuration.
#   ## Only applies to stream sockets (e.g. TCP).
#    keep_alive_period = "5m"
#
#   ## Maximum number of concurrent connections (default = 0).
#   ## 0 means unlimited.
#   ## Only applies to stream sockets (e.g. TCP).
#   # max_connections = 1024
#
#   ## Read timeout (default = 500ms).
#   ## 0 means unlimited.
#   # read_timeout = 500ms
#
#   ## Whether to parse in best effort mode or not (default = false).
#   ## By default best effort parsing is off.
#   # best_effort = false
#
#   ## Character to prepend to SD-PARAMs (default = "_").
#   ## A syslog message can contain multiple parameters and multiple identifiers within structured data section.
#   ## Eg., [id1 name1="val1" name2="val2"][id2 name1="val1" nameA="valA"]
#   ## For each combination a field is created.
#   ## Its name is created concatenating identifier, sdparam_separator, and parameter name.
#   # sdparam_separator = "_"
```
### System info:

Ubuntu 16.04.3 x64   Telegraf 1.7

### Steps to reproduce:

1) Install the latest Telegraf build

2) Followed [this topic ](https://community.influxdata.com/t/enabling-the-syslog-log/5604/6) and [read me](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/syslog#rsyslog-integration)

3) Restart Telegraf after editing the config and wait for the data, tail the telegraf logs 

# Expected behavior:
Collect mertics from Syslog and send them to InfluxDB
### Actual behavior:
Telegraf does send the data to InfluxDB but this error appears in the telegaf log:

`2018-06-22T11:19:49Z E! Error in plugin [inputs.syslog]: found EOF, expecting a MSGLEN
2018-06-22T11:19:59Z E! Error in plugin [inputs.syslog]: found EOF, expecting a MSGLEN
2018-06-22T11:20:09Z E! Error in plugin [inputs.syslog]: found EOF, expecting a MSGLEN
2018-06-22T11:20:19Z E! Error in plugin [inputs.syslog]: found EOF, expecting a MSGLEN
2018-06-22T11:20:29Z E! Error in plugin [inputs.syslog]: found EOF, expecting a MSGLEN
2018-06-22T11:20:39Z E! Error in plugin [inputs.syslog]: found EOF, expecting a MSGLEN
2018-06-22T11:20:49Z E! Error in plugin [inputs.syslog]: found EOF, expecting a MSGLEN
2018-06-22T11:20:59Z E! Error in plugin [inputs.syslog]: found EOF, expecting a MSGLEN
2018-06-22T11:21:02Z E! Error in plugin [inputs.syslog]: found EOF, expecting a MSGLEN
2018-06-22T11:21:09Z E! Error in plugin [inputs.syslog]: found EOF, expecting a MSGLEN`
### Additional info:
I think my config is valid, Telegraf starts with no issues and sends the metrics to InfluxDB. I've also tested this with a separate Telegraf instance sending data to a remote InfluxDB. 

If i can provide anything else let me know. 

Regards, 

PhilB

glinton:
This change reads from a syslog client in a separate step from the log parsing. This allows fewer errors in telegraf's logs about exceeding read timeouts. This should also limit the amount of reconnecting syslog clients must do when streaming logs to telegraf.

Previously, the syslog plugin would immediately start reading from a connection and if data was not immediately provided (or a connection was made and closed), `found EOF, expecting a MSGLEN` would be printed to telgraf's log.

By reading the data on the connection in a separate step, we are able to handle connection issues separately from parsing issues, thus resulting in less of them. This will also allow for syslog client connections to stay alive longer.

Resolves #4335

### DIFF
--- a/plugins/inputs/syslog/syslog.go
+++ b/plugins/inputs/syslog/syslog.go
@@ -1,7 +1,6 @@
 package syslog
 
 import (
-	"bytes"
 	"crypto/tls"
 	"fmt"
 	"io"
@@ -281,52 +280,23 @@ func (s *Syslog) handle(conn net.Conn, acc telegraf.Accumulator) {
 	}()
 
 	var p *rfc5425.Parser
-	data := &bytes.Buffer{}
 
 	if s.BestEffort {
-		p = rfc5425.NewParser(data, rfc5425.WithBestEffort())
+		p = rfc5425.NewParser(conn, rfc5425.WithBestEffort())
 	} else {
-		p = rfc5425.NewParser(data)
+		p = rfc5425.NewParser(conn)
 	}
 
-	for {
-		data.Reset()
+	if s.ReadTimeout != nil && s.ReadTimeout.Duration > 0 {
+		conn.SetReadDeadline(time.Now().Add(s.ReadTimeout.Duration))
+	}
 
-		// read the data
+	p.ParseExecuting(func(r *rfc5425.Result) {
+		s.store(*r, acc)
 		if s.ReadTimeout != nil && s.ReadTimeout.Duration > 0 {
 			conn.SetReadDeadline(time.Now().Add(s.ReadTimeout.Duration))
 		}
-
-		n, err := io.Copy(data, conn)
-		if err != nil {
-			// read timeout reached, parse what we have
-			if er, ok := err.(net.Error); ok && er.Timeout() {
-				if n == 0 {
-					continue
-				}
-				goto parseMsg
-			}
-			// client has closed connection, return
-			if err == io.EOF {
-				if n > 0 {
-					goto parseMsg
-				}
-				return
-			}
-			// other error, log and return
-			acc.AddError(fmt.Errorf("failed reading from syslog client - %s", err.Error()))
-			return
-		}
-		// handle client disconnect
-		if n == 0 {
-			return
-		}
-
-	parseMsg:
-		p.ParseExecuting(func(r *rfc5425.Result) {
-			s.store(*r, acc)
-		})
-	}
+	})
 }
 
 func (s *Syslog) setKeepAlive(c *net.TCPConn) error {
--- a/plugins/inputs/syslog/syslog.go
+++ b/plugins/inputs/syslog/syslog.go
@@ -296,12 +296,16 @@ func (s *Syslog) handle(conn net.Conn, acc telegraf.Accumulator) {
 			// read to the tmp var
 			n, err := conn.Read(tmp)
 			if err != nil {
-				// Ignore known/recoverable errors. In contrived tests:
-				// * i/o timeout error - no data to Read() before s.ReadTimeout.Duration expired
-				// * EOF error - connection open/close immediately
-				if er, ok := err.(net.Error); err != io.EOF && (ok && !er.Timeout()) {
-					s.store(rfc5425.Result{Error: fmt.Errorf("Failed reading from syslog client - %s", err.Error())}, acc)
+				// read timeout reached, retry
+				if er, ok := err.(net.Error); ok && er.Timeout() {
+					continue
 				}
+				// client has closed connection, return
+				if err == io.EOF {
+					return
+				}
+				// other error, log and return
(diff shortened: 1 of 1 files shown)


plugins/inputs/syslog/syslog.go
```diff
@@ -279,20 +280,51 @@ func (s *Syslog) handle(conn net.Conn, acc telegraf.Accumulator) {
...
+			}
+			// client has closed connection, return
+			if err == io.EOF {
+				if n > 0 {
+					goto parseMsg
+				}
+				return
+			}
+			// other error, log and return
+			s.store(rfc5425.Result{Error: fmt.Errorf("failed reading from syslog client - %s", err.Error())}, acc)
```
danielnelson: Just use acc.AddError(err) directly.

plugins/inputs/syslog/syslog.go
```diff
@@ -279,20 +280,51 @@ func (s *Syslog) handle(conn net.Conn, acc telegraf.Accumulator) {
...
+			return
+		}
+		// handle client disconnect
+		if n == 0 {
+			return
+		}
+
+	parseMsg:
+		var p *rfc5425.Parser
```
danielnelson: Ideally we would only make one parser for each call to handle.

plugins/inputs/syslog/syslog.go
```diff
@@ -279,20 +280,51 @@ func (s *Syslog) handle(conn net.Conn, acc telegraf.Accumulator) {
...
-	}
+		// read the data
+		if s.ReadTimeout != nil && s.ReadTimeout.Duration > 0 {
+			conn.SetReadDeadline(time.Now().Add(s.ReadTimeout.Duration))
+		}
 
-	p.ParseExecuting(func(r *rfc5425.Result) {
-		s.store(*r, acc)
-	})
+		n, err := io.Copy(data, conn)
```
danielnelson: Won't this read until EOF?  How many messages can syslog send over a single connection.  If it allows persistent connection then this could be very large.

leodido: A receiver MUST use the message length to delimit a syslog messages, not `EOF`.
Ie., 14 octets for the following message `<1>1 - - - - -`, thus `14 <1>1 - - - - -`.

RFC5425 specifies only that receivers SHOULD be able to process messages with lengths up to and including 8192 octets.

There is not maximum.

leodido:
@danielnelson no we cannot push down to go-syslog imho, RFC5425 says nothing about timeout.

danielnelson:
Without duplicating the syslog tcp framing, we can't do the message splitting in Telegraf since we don't know where messages begin or ends.  I think this means we have to send in the connection io.Reader.

So, I think we could handle the error in the ParseExecuting function (`Syslog.store`) by watching for the `found EOF, expecting MSGLEN` message.  The downside of this method is we don't know if the connection was closed, a timeout occured or something else.  This makes good error messages hard to produce.  Also, we would need to match on the error string which is not preferred.  As an aside, it also seems like we need to extend the deadline in this function to prevent errors every $timeout seconds.

However, what I think would be ideal is if go-syslog could return errors when returned from the Read function of the io.Reader provided.  The application could then check the error and handle it in an way depending on the type of the io.Reader.

Issue #4375, opened 2018-07-03
Docker memory usage include cache, so is misleading


easyrasta:
Docker plugin get memory usage which include cache. So after some hours, reported memory usage is totaly wrong.

I check code I see you compute memory usage without cache but is not reported to any metrics
https://github.com/influxdata/telegraf/blob/c5fa6729d356d3ecb19faebe7d068ccf903c1ff8/plugins/inputs/docker/docker.go#L533
Could we replace docker memory usage by a better memory usage or to add new metric like use_nocache ?

See also: 
- https://github.com/moby/moby/issues/10824
- https://github.com/docker/cli/pull/80

danielnelson:
Our code is based on the docker cli code, and we want the behavior be the same as upstream.  I quickly looked at our code and it appears very similar to https://github.com/docker/cli/pull/80 , Can you clarify which field is off?

https://github.com/influxdata/telegraf/blob/c389a68f1925b7f4295ab774cdd39fb2acc425c8/plugins/inputs/docker/docker.go#L533-L535

easyrasta:
In docker cli they replaced memory by mem usage without cache. In telegraf we compute only mem usage without cache for mem percent computing (see https://github.com/docker/cli/pull/80/files )

easyrasta:
https://github.com/docker/cli/blob/f285fe67e916055ea61d45d6249c9946ee27b216/cli/command/container/stats_helpers.go#L107#122

easyrasta:
#4383

PR #4383, opened 2018-07-05
#4375 - use memory usage without memory cache

easyrasta:
### Required for all PRs:


plugins/inputs/docker/stats_helpers.go
```diff
@@ -41,8 +41,8 @@ func calculateCPUPercentWindows(v *types.StatsJSON) float64 {
 
 // calculateMemUsageUnixNoCache calculate memory usage of the container.
 // Page cache is intentionally excluded to avoid misinterpretation of the output.
-func calculateMemUsageUnixNoCache(mem types.MemoryStats) float64 {
-	return float64(mem.Usage - mem.Stats["cache"])
+func calculateMemUsageUnixNoCache(mem types.MemoryStats) uint64 {
```
glinton: @danielnelson may need to weigh in, but I believe changing types isn't supported in certain outputs.

danielnelson: That's true but in this case I believe the field types are retained: `usage` is a uint64 and `usage_percent` is a float64.  I think it would be nice to leave this function returning a float64 though so that our code is closer to the upstream code, and then convert to an uint64 just for the usage field.

Gopkg.lock
```diff
@@ -413,6 +413,217 @@
   ]
   revision = "c43482518d410361b6c383d7aebce33d0471d7bc"
 
+[[projects]]
```
danielnelson: The changes to this file should be removed.

rdxmb:
very good :+1:

PR #4385, opened 2018-07-05
Prevent new tag from being overwritten

### LINKED ISSUE #4374: regex plugin does not support matching on the same key and result key
### Relevant telegraf.conf:
```
[[processors.regex]]
  namepass = ["RandomData"]
  [[processors.regex.tags]]
    key = "ComputerName"
    pattern = "^(.*?)a$"
    replacement = "${1}"
    result_key = "ServerName"
  [[processors.regex.tags]]
    key = "ComputerName"
    pattern = "^(.*?)b$"
    replacement = "${1}"
    result_key = "ServerName"
```

### System info:
Telegraf Version: 1.8.0
OS: Windows Server 2012

### Steps to reproduce:

1. Use any input plugin. Try to strip off the last character

### Expected behavior:
In the above configuration, I would expect that ServerName should be a tag that holds the value of the tag "ComputerName" without the trailing a or b.

### Actual behavior:
Only the the configuration of the last regex is honored. I believe what is happening is that when a subsequent regex match fails, it returns an empty string, which then get's reassigned to the tag, overwriting the value that I was interested in.

### Additional info:
I believe this could be a quick fix to the regex processor plugin. The fix I made is below, and seems to be working:

```
		for _, converter := range r.Tags {
			if value, ok := metric.GetTag(converter.Key); ok {
				k, v := r.convert(converter, value)
				if k != "" && v != "" {
					metric.AddTag(k, v)
				}
			}
		}

		for _, converter := range r.Fields {
			if value, ok := metric.GetField(converter.Key); ok {
				switch value := value.(type) {
				case string:
					k, v := r.convert(converter, value)
					if k != "" && v != "" {
						metric.AddField(r.convert(converter, value))
					}
				}
			}
		}
```

glinton:
This prevents tags from being overwritten by other regex processors.

Resolves #4374

### DIFF
--- a/plugins/processors/regex/regex.go
+++ b/plugins/processors/regex/regex.go
@@ -34,7 +34,7 @@ const sampleConfig = `
   #   key = "request"
   #   ## All the power of the Go regular expressions available here
   #   ## For example, named subgroups
-  #   pattern = "^/api(?P<method>/[\\w/]+)\\S*" 
+  #   pattern = "^/api(?P<method>/[\\w/]+)\\S*"
   #   replacement = "${method}"
   #   ## If result_key is present, a new field will be created
   #   ## instead of changing existing field
@@ -67,7 +67,10 @@ func (r *Regex) Apply(in ...telegraf.Metric) []telegraf.Metric {
 	for _, metric := range in {
 		for _, converter := range r.Tags {
 			if value, ok := metric.GetTag(converter.Key); ok {
-				metric.AddTag(r.convert(converter, value))
+				k, v := r.convert(converter, value)
+				if k != "" && v != "" {
+					metric.AddTag(k, v)
+				}
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


danielnelson:
It appears from the test case that this behavior was added on purpose.  @44px do you have a specific use case where you need this behavior?

I think it makes sense to prevent empty tag/fields keys and tag values, though I'm less sure about field values.  It's imaginable that someone would want an empty string as a field value, but I imagine it is almost always unwanted.

I also think this is kind of a roundabout way to fix the issue.  It prevents a non matching pattern from clearing out fields/tags by preventing empty values, but I think it would be better if we also directly fixed the core issue: when the pattern doesn't match it shouldn't modify the key/value at all (pending response from @44px).

44px:
Nope, I don't rely on this behavior. I added this because it seems more correct to emit empty string instead of original tag/field value when `result_key` specified and pattern doesn't match.

For example, having config like:
```toml
[[processors.regex]]
  [[processors.regex.fields]]
    key = "request"
    pattern = "^/api(?P<method>/[\\w/]+)\\S*"
    replacement = "${method}"
    result_key = "method"
```

And given `request="/favicon.ico"` it should emit `method=""`, not `method="/favicon.ico"`.

> when the pattern doesn't match it shouldn't modify the key/value at all

You're right, it seems like more correct behavior. Plugin should not alter existing or emit new tag/field (depending on `result_key`) if there is no match. Here is my fix with new behavior: https://github.com/influxdata/telegraf/pull/4396

danielnelson:
Fixed in #4396

PR #4391, opened 2018-07-06
Filter on all container names, use proper name for container

### LINKED ISSUE #4376: Docker container name reported incorrectly
### Relevant telegraf.conf:
```
[[inputs.docker]]
  timeout = "5s"
  perdevice = true
  total = false
  container_names = []
```
### System info:
Ubuntu 16.04 LTS
Telegraf v1.7.0 (git: release-1.7 f4d22dd4) installed from https://repos.influxdata.com/ubuntu

### Steps to reproduce:
Create a docker container with a given name (in my case influxdb), make sure name is correct:

```
# docker ps
CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS              PORTS                    NAMES
602937aada3f        influxdb            "/entrypoint.sh infl…"   19 hours ago        Up 19 hours         0.0.0.0:8086->8086/tcp   influxdb
```
(output trimmed)

```
#docker inspect influxdb
[
    {
     {...}
        "Name": "/influxdb",
     }
]
```
(output trimmed)

but when I check for the metrics in influx I get:
```
 show tag values from docker_container_net with key=container_name
name: docker_container_net
key            value
---            -----
container_name grafana/influxdb
```
(output trimmed)

So for some reason the name of the container is prefixed with a different container name 
### Expected behavior:
The name is reported as it's seen by docker

### Actual behavior:
The name gets prepended with another container name

glinton:
Resolves #4376

### DIFF
--- a/plugins/inputs/docker/docker.go
+++ b/plugins/inputs/docker/docker.go
@@ -365,22 +365,17 @@ func (d *Docker) gatherContainer(
 ) error {
 	var v *types.StatsJSON
 	// Parse container name
-	cname := "unknown"
-	match := false
-	if len(container.Names) == 0 { // for tests
-		match = true
-	}
-
-	for i := range container.Names {
-		if !match {
-			match = d.containerFilter.Match(strings.TrimPrefix(container.Names[i], "/"))
-			if match {
-				cname = strings.TrimPrefix(container.Names[i], "/")
-			}
+	var cname string
+	for _, name := range container.Names {
+		trimmedName := strings.TrimPrefix(name, "/")
+		match := d.containerFilter.Match(trimmedName)
+		if match {
+			cname = trimmedName
+			break
 		}
 	}
 
-	if !match {
+	if cname == "" {
 		return nil
 	}
 
@@ -421,9 +416,7 @@ func (d *Docker) gatherContainer(
 	daemonOSType := r.OSType
 
 	// use common (printed at `docker ps`) name for container
-	if cname != strings.TrimPrefix(v.Name, "/") && v.Name != "" {
-		tags["container_name"] = strings.TrimPrefix(v.Name, "/")
-	}
+	tags["container_name"] = strings.TrimPrefix(v.Name, "/")
 
 	// Add labels to tags
 	for k, label := range container.Labels {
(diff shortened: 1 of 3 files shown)


plugins/inputs/docker/docker.go
```diff
@@ -366,9 +366,22 @@ func (d *Docker) gatherContainer(
...
+	}
+
+	for i := range container.Names {
+		if !match {
+			match = d.containerFilter.Match(strings.TrimPrefix(container.Names[i], "/"))
+			if match {
+				cname = strings.TrimPrefix(container.Names[i], "/")
+			}
+		}
+	}
```
danielnelson: I hate to be this specific, but this loop can be simplified:

```go
	var cname string
	for _, name := range container.Names {
		trimmedName = strings.TrimPrefix(name, "/")
		match := d.containerFilter.Match(trimmedName)
		if match {
			cname = trimmedName
			break
		}
	}

	if cname == "" {
		return nil
	}
```

By the way, its a minor thing but I never use the index form of the range loop unless I need it, indexing is harder to read, harder to type, and more prone to panic.

plugins/inputs/docker/docker.go
```diff
@@ -366,9 +366,22 @@ func (d *Docker) gatherContainer(
 	var v *types.StatsJSON
 	// Parse container name
 	cname := "unknown"
-	if len(container.Names) > 0 {
-		// Not sure what to do with other names, just take the first.
-		cname = strings.TrimPrefix(container.Names[0], "/")
+	match := false
+	if len(container.Names) == 0 { // for tests
```
danielnelson: Should fix the tests not to need this.

plugins/inputs/docker/docker.go
```diff
@@ -411,6 +420,11 @@ func (d *Docker) gatherContainer(
 	}
 	daemonOSType := r.OSType
 
+	// use common (printed at `docker ps`) name for container
+	if cname != strings.TrimPrefix(v.Name, "/") && v.Name != "" {
```
danielnelson: Don't check if it is already correct, it will be faster just to set it.  Also if you have never seen an empty Name field in docker then I think you should just set the value without any test.

PR #4397, opened 2018-07-07
support expiration of stale metrics and hide unchanged metrics

Dieken:
In pull mode such as Prometheus, if the source is down, its metrics
are still kept in Telegraf's cache when delete_{gauges,counters,sets,timings}
are false, then Prometheus pull those stale metrics again and again. This
patch hides metrics unchanged since last gather and deletes too old ones.

### Required for all PRs:

### DIFF
--- a/plugins/inputs/statsd/README.md
+++ b/plugins/inputs/statsd/README.md
@@ -3,14 +3,14 @@
 ### Configuration
 
 ```toml
-# Statsd Server
+# Statsd UDP/TCP Server
 [[inputs.statsd]]
-  ## Protocol, must be "tcp", "udp4", "udp6" or "udp" (default=udp)
+  ## Protocol, must be "tcp", "udp", "udp4" or "udp6" (default=udp)
   protocol = "udp"
 
   ## MaxTCPConnection - applicable when protocol is set to tcp (default=250)
   max_tcp_connections = 250
-  
+
   ## Enable TCP keep alive probes (default=false)
   tcp_keep_alive = false
 
@@ -33,6 +33,9 @@
   delete_sets = true
   ## Reset timings & histograms every interval (default=true)
   delete_timings = true
+
+  ## Interval to expire metrics and not be collected, 0 == no expiration (default=1h)
+  expiration_interval = "24h"
 
   ## Percentiles to calculate for timing & histogram stats
   percentiles = [90]
@@ -172,6 +175,7 @@ to allow. Used when protocol is set to tcp.
 - **delete_counters** boolean: Delete counters on every collection interval
 - **delete_sets** boolean: Delete set counters on every collection interval
 - **delete_timings** boolean: Delete timings on every collection interval
+- **expiration_interval** Duration: Expire unchanged old metrics on every collection internal
 - **percentiles** []int: Percentiles to calculate for timing & histogram stats
 - **allowed_pending_messages** integer: Number of messages allowed to queue up
 waiting to be processed. When this fills, messages will be dropped and logged.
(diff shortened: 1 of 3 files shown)


danielnelson:
Can you describe in a little more detail why setting, for example, `delete_counters` along with `expiration_interval` in the prometheus output is not sufficient?

Dieken:
@danielnelson 

* why `delete_counters=true` is not ok to me

I want to graph QPS with Prometheus and Grafana, I have two choices:

(1) delete_counters=true.  telegraf will consume very few memory, nice, but what I record in Prometheus is queries per 10 seconds (telegraf default collecting interval),  I have to graph "value / 10" in Grafana to get QPS,  if some day I want to change telegraf collecting interval, I have to update many Grafana dashboards to change the PromQL clauses.  Actually many don't realize the data points are queries per 10s, not queries per second.

(2) delete_counters=false,  telegraf will accumulate the absolute total queries (not so absolute when telegraf is restarted, but anyway good enough),  so I can use rate() in PromQL to get QPS, no matter what collection interval Telegraf uses,   simple and intuitive.

But now Telegraf will eventually run out of memory, that's why I add expiration_interval to plugin statsd, so dead metrics will be purged.

* why `delete_counters=false` and `expiration_interval` in plugin prometheus is not sufficient.

Currently plugin statsd has a cache to aggregate metrics,  plugin prometheus also has a cache for Prometheus, default TTL is 60s, so that Prometheus can have a maximum poll interval 60s.

Notice they are two different caches,  when "delete_counters" is false,  plugin statsd never clears its cache so it will get OOM sooner or later,  that's why I need `expiration_interval` *too* for plugin statsd. 

Longer the `expiration_interval` setting, less chance a not so active counter gets reset to zero but more chance Telegraf gets OOM. Currently I use default value "1h".

Another purpose of `expiration_interval` for *plugin statsd* is to avoid stale metrics keeping being collected by Prometheus,  for example, a Kubernetes pod is deleted,  no new metric "queries,pod=xx:N|c" will be emit to Telegraf, but Telegraf's plugin statsd doesn't record metric time, its Grather() function always use "now",  so Prometheus will keep getting this stale counter when `delete_counters=false` and plugins statsd doesn't expire metrics, it not only wastes storage of Prometheus but also confuse people.

Actually the patch is more aggressive,  it hides a metric to not expose to plugin prometheus if the metric hasn't been updated since last Gather(),  so Prometheus almost immediately got no stale metric when the source is down. This is more consistent with the semantic of push model for statsd. And Grafana can handle missing data points between two timestamps, Grafana can just draw a straight line between two values, or think the missing values are all zero.

BTW, I don't know why the master branch use true for "delete_*",  the latest v0.10.1 still use "false" as default: https://github.com/influxdata/telegraf/blob/v0.10.1/plugins/inputs/statsd/statsd.go

danielnelson:
Thank you for the detailed explanation of the problem.

> delete_counters=false, telegraf will accumulate the absolute total queries (not so absolute when telegraf is restarted, but anyway good enough)

This seems like a bug, or at best an inconsistency with etsy/statsd.  I believe etsy/statsd resets counters on every interval, and should only resend previous gauge values while sending 0 for counters.  I found a good description on [this page](http://dieter.plaetinck.be/post/25-graphite-grafana-statsd-gotchas/#statsd.counters) of how statsd counters normally work.

Because this fix relies on the current incorrect statsd behavior, I don't think we should include this patch.  I'm not proficient with PromQL, but perhaps you can do a integral on your data to convert to a proper monotonically incrementing counter before applying a rate?

> BTW, I don't know why the master branch use true for "delete_*", the latest v0.10.1 still use "false" as default

Looks like this was done in 1.2 https://github.com/influxdata/telegraf/pull/2155

PR #4405, opened 2018-07-10
nvidia-smi readme update - Support for Windows Powershell

gsxryan:
Added to readme for users that would like to use this plugin on windows / powershell

### Required for all PRs:

- [na ] Has appropriate unit tests.

### DIFF
--- a/plugins/inputs/nvidia_smi/README.md
+++ b/plugins/inputs/nvidia_smi/README.md
@@ -14,6 +14,13 @@ This plugin uses a query on the [`nvidia-smi`](https://developer.nvidia.com/nvid
 # timeout = 5s
 ```
 
+### Windows Users
+Uncomment the bin_path line above and use the following syntax
+
+```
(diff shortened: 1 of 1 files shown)


danielnelson:
Thanks but I'm don't think this is required, it isn't different in Windows vs on other platforms.

gsxryan:
It isn't required, but it sure would have helped me in my search, if it was there.  Windows does not have a bin path.  And telegraf directory formatting isn't using a standard that windows script users are used to.   (Requires quotes where the Linux path didn't, and used forwardslash instead of backslash)

danielnelson:
The lack of quoting appears to actually be a bug in the example config, I'll fix that up real quick.

You can actually use backslashes here, the easiest way is to use single quotes:
```
bin_path = 'C:\Program Files\NVIDIA Corporation/NVSMI\nvidia-smi.exe'
```

gsxryan:
Aha!  Nice.  Thanks for the fix!

Issue #4290, opened 2018-06-14
no clear documentation on swap input plugin


randallt:
### Relevant telegraf.conf:

[[inputs.swap]]

### System info:

[Include Telegraf version, operating system name, and other relevant details]

Telegraf 1.6.4. CentOS 7.4

### Steps to reproduce:

1. Try to find clear documentation on the swap input plugin fields.
2. Fail

### Expected behavior:

I should be able to find clear documentation on what fields the swap input plugin produces, what they mean, and where they come from.

### Actual behavior:

I had to use guess and check mode and look at Go code.

danielnelson:
We should move this plugin out into it's own file while we are at it, and create a new `SWAP_README.md`.  Longer term I would like to move all the "system" plugins out into the main plugins/inputs directory to increase uniformity.

danielnelson:
Closed in #4342

Issue #4400, opened 2018-07-09
collect virtual machine temperature


pytimer:
## Feature Request

Opening a feature request kicks off a discussion.

### Proposal:

`system` inputs plugin add temperature metric.

### Current behavior:

I want to use telegraf collect vm temperature, but i not found inputs plugin to do it.

### Desired behavior:

Telegraf have inputs plugin to do it.

### Use case: [Why is this important (helps with prioritizing requests)]

If telegraf inputs plugin can collect temperature, please tell me. Thanks. :)

danielnelson:
We have the [sensors input](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/sensors) for Linux, but I'm assuming you are interested in this on Windows via gopsutil?  I think this would be good to add, do you want to prepare a PR?

pytimer:
Thanks for your reply.

I read `sensors input`, it use `lm-sensors` to collect sensor metrics, but many linux machine not installed lm-sensors package. 

I already read gopsutil code, linux it read machine file to get [temperature](https://github.com/shirou/gopsutil/blob/master/host/host_linux.go#L594:6), and yesterday i send a [PR](https://github.com/shirou/gopsutil/pull/550) to gopsutil,  use WMI to collect Windows temperature. Is it helpful for Telegraf?

Whether Telegraf use gopsutil to get temperature? If not, you can tell me which plugin can add this feature, i will happly to send a PR to implement it.

danielnelson:
It should be acceptable to add a `temp` plugin that uses this interface from gopsutil.

pytimer:
Add a new `temp` input plugin in `system` package, i understand it right?

If right, which metrics name should be included in this plugin. Do you have any idea?

danielnelson:
> Add a new temp input plugin in system package, i understand it right?

Yes

> If right, which metrics name should be included in this plugin.

Since the current interface from gopsutil is:
```go
type TemperatureStat struct {
	SensorKey   string  `json:"sensorKey"`
	Temperature float64 `json:"sensorTemperature"`
}
```

Something like this should be appropriate:
```
temp,sensor=<sensorKey> temp=<sensorTemperature>
```

pytimer:
Now my PR not merge, if it merged, i want to begin this work. 
Is it ok? :-)

danielnelson:
[...]

PR #4406, opened 2018-07-10
Moved 'system' inputs out as individual inputs

sdomino:
### Required for all PRs:

This is in response to a comment made in #4290. Associated `docs` update [here](https://github.com/influxdata/docs.influxdata.com/pull/1715).

I wasn't sure whether to separate `net` and `netstat`. After looking them over, it seemed like they're intended to fulfill the same purpose for different platforms, and so would remain as parts of the same package.

Ultimately, I based that decision off of reading the following in the `net` README:

```
Different platforms gather the data above with different mechanisms. Telegraf uses the (gopsutil) package, which under Linux reads the /proc/net/dev file. Under freebsd/openbsd and darwin the plugin uses netstat.
```


plugins/inputs/ps.go
```diff
@@ -1,4 +1,4 @@
-package system
+package inputs
```
danielnelson: Can you leave this file in the system package for the time being?  Long term I would like to split it out to the required plugins as well as splitting the interface up, but we should do that after moving the files to reduce the size of the conflict window.

plugins/inputs/mock_PS.go
```diff
@@ -1,4 +1,4 @@
-package system
```
danielnelson: Leave in system package for now.

tstackhouse:
It was mentioned in #4400 that the `system` package may be removed, with 1.8 on the horizon, what would be the equivalent? Assuming the docs will be updated soon?

danielnelson:
This is mostly a code organization change.  From a users point of view the only change is that the README files for the plugins have been moved to a different location.  For example, with the `disk` input:
```diff
- /plugins/inputs/system/DISK_README.md
+ /plugins/inputs/disk/README.md
``` 

There isn't any effect on the configuration file or on plugin availability, and the `system` input plugin is not removed, so there should be no action required.

PR #4424, opened 2018-07-15
Add an exec daemon plugin

jgraichen:
The execd input plugin spawns a long running process and parses its output to gather metrics. The process can be signaled in different ways on each interval.

This can be used to avoid spawning processes on each interval as with the exec plugin or with external programs that report metrics by themselves e.g. in irregular intervals (events). Processes are restarted after termination with a one second pause.

The overhead is minimal for both sides as only `stdout`, `stderr`, and `stdin` or signals are used. Some example daemons in bash, go, and ruby are included.

---

This is my first time writing Go at all. Feel free to comment or change things. I only have this running on linux, but cross compilation to windows works. I have no idea how to unit test this unfortunately.

### Required for all PRs:

### DIFF
--- a/plugins/inputs/execd/execd.go
+++ b/plugins/inputs/execd/execd.go
@@ -40,10 +40,11 @@ type Execd struct {
 	Command string
 	Signal  string
 
-	acc    telegraf.Accumulator
-	cmd    *exec.Cmd
-	parser parsers.Parser
-	stdin  io.WriteCloser
+	acc     telegraf.Accumulator
+	cmd     *exec.Cmd
+	parser  parsers.Parser
+	stdin   io.WriteCloser
+	stopped bool
 }
 
 func (e *Execd) SampleConfig() string {
@@ -72,6 +73,8 @@ func (e *Execd) Start(acc telegraf.Accumulator) error {
 }
 
 func (e *Execd) Stop() {
+	e.stopped = true
+
 	if e.cmd == nil || e.cmd.Process == nil {
 		return
 	}
@@ -121,6 +124,10 @@ func (e *Execd) cmdRun(args []string) error {
 
 	wg.Wait()
 	e.cmd.Wait()
+
+	if e.stopped {
+		return nil
+	}
 
 	log.Printf("E! [inputs.execd] %s terminated. Restart in one second...", e.Command)
 
--- a/plugins/inputs/execd/execd.go
+++ b/plugins/inputs/execd/execd.go
@@ -158,7 +158,7 @@ func (e *Execd) cmdReadOut(out io.Reader) {
 		}
 
 		for _, metric := range metrics {
-			e.acc.AddFields(metric.Name(), metric.Fields(), metric.Tags(), metric.Time())
+			e.acc.AddMetric(metric)
 		}
 	}
 
--- a/plugins/inputs/execd/execd.go
+++ b/plugins/inputs/execd/execd.go
@@ -159,7 +159,7 @@ func (e *Execd) cmdReadErr(out io.Reader, wg *sync.WaitGroup) {
 	scanner := bufio.NewScanner(out)
 
 	for scanner.Scan() {
-		log.Printf("E! [inputs.execd] stderr output: %s", scanner.Text())
+		log.Printf("E! [inputs.execd] stderr output: %q", scanner.Text())
 	}
 
 	if err := scanner.Err(); err != nil {
--- a/plugins/inputs/execd/execd.go
+++ b/plugins/inputs/execd/execd.go
@@ -1,0 +1,176 @@
+package execd
+
+import (
+	"bufio"
+	"fmt"
+	"io"
+	"log"
+	"os/exec"
+	"sync"
+	"time"
+
+	"github.com/kballard/go-shellquote"
+
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/plugins/inputs"
+	"github.com/influxdata/telegraf/plugins/parsers"
+)
+
+const sampleConfig = `
+  ## Program to run as daemon
+  command = "telegraf-smartctl"
+
+  ## Define how the process is signaled on each collection interval.
+
+  ## Valid values are:
+  ##   "none"   : Do not signal anything.
+  ##              The process must output metrics by itself.
+  ##   "STDIN"  : Send a newline on STDIN.
+  ##   "SIGHUP" : Send a HUP signal. Not available on Windows.
+  signal = "none"
+
+  ## Data format to consume.
+  ## Each data format has its own unique set of configuration options, read
+  ## more about them here:
+  ## https://github.com/influxdata/telegraf/blob/master/docs/DATA_FORMATS_INPUT.md
+  data_format = "influx"
+`
+
+type Execd struct {
+	Command string
+	Signal  string
+
+	acc    telegraf.Accumulator
+	cmd    *exec.Cmd
+	parser parsers.Parser
+	stdin  io.WriteCloser
+}
+
+func (e *Execd) SampleConfig() string {
+	return sampleConfig
+}
+
+func (e *Execd) Description() string {
+	return "Run executable as long-running input plugin"
+}
+
+func (e *Execd) SetParser(parser parsers.Parser) {
+	e.parser = parser
+}
+
+func (e *Execd) Start(acc telegraf.Accumulator) error {
+	e.acc = acc
+
+	args, err := shellquote.Split(e.Command)
+	if err != nil || len(args) == 0 {
+		return fmt.Errorf("execd: unable to parse command: %s", err)
+	}
+
+	go e.cmdRun(args)
+
+	return nil
+}
+
+func (e *Execd) Stop() {
+	if e.cmd == nil || e.cmd.Process == nil {
+		return
+	}
+
+	if err := e.cmd.Process.Kill(); err != nil {
+		log.Printf("E! FATAL error killing process: %s", err)
+	}
+}
+
+func (e *Execd) cmdRun(args []string) error {
+	var wg sync.WaitGroup
+
+	if len(args) > 1 {
+		e.cmd = exec.Command(args[0], args[1:]...)
+	} else {
+		e.cmd = exec.Command(args[0])
+	}
+
+	stdin, err := e.cmd.StdinPipe()
+	if err != nil {
+		return fmt.Errorf("execd: error opening stdin pipe: %s", err)
+	}
+
+	e.stdin = stdin
+
+	stdout, err := e.cmd.StdoutPipe()
+	if err != nil {
+		return fmt.Errorf("execd: error opening stdout pipe: %s", err)
+	}
+
+	stderr, err := e.cmd.StderrPipe()
+	if err != nil {
+		return fmt.Errorf("execd: error opening stderr pipe: %s", err)
+	}
+
+	log.Printf("D! execd start program: %s", e.Command)
+
+	err = e.cmd.Start()
+	if err != nil {
+		return fmt.Errorf("execd error starting program: %s", err)
+	}
+
+	wg.Add(2)
+
+	go e.cmdReadOut(stdout, &wg)
+	go e.cmdReadErr(stderr, &wg)
+
+	wg.Wait()
+	e.cmd.Wait()
+
+	log.Printf("E! execd: %s terminated. Restart in one second...", e.Command)
+
+	go func() {
+		<-time.After(time.Second)
+		e.cmdRun(args)
+	}()
+
+	return nil
+}
+
+func (e *Execd) cmdReadOut(out io.Reader, wg *sync.WaitGroup) {
+	defer wg.Done()
+
+	scanner := bufio.NewScanner(out)
+
+	for scanner.Scan() {
+		metrics, err := e.parser.Parse(scanner.Bytes())
+		if err != nil {
+			e.acc.AddError(fmt.Errorf("parse error: %s", err.Error()))
+		}
+
+		for _, metric := range metrics {
+			e.acc.AddFields(metric.Name(), metric.Fields(), metric.Tags(), metric.Time())
+		}
+	}
+
+	if err := scanner.Err(); err != nil {
+		e.acc.AddError(fmt.Errorf("Error reading stdout: %s", err.Error()))
+	}
+}
+
+func (e *Execd) cmdReadErr(out io.Reader, wg *sync.WaitGroup) {
+	defer wg.Done()
+
+	scanner := bufio.NewScanner(out)
+
+	for scanner.Scan() {
+		log.Printf("E! execd stderr output: %s", scanner.Text())
+	}
+
+	if err := scanner.Err(); err != nil {
+		e.acc.AddError(fmt.Errorf("Error reading stderr: %s", err.Error()))
+	}
+}
+
+func init() {
+	inputs.Add("execd", func() telegraf.Input {
+		return &Execd{
+			Signal: "none",
+		}
+	})
+}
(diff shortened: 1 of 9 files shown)


jgraichen:
Sorry for the delay. I've changed things as requested. Thanks for that review.

sgtsquiggs:
I _really_ like this idea. I have had to for telegraf to add my own plugins but having execd would allow me to build and maintain them separately. I love the idea of using a signal to trigger metrics. This would allow the plugin to either use the signal to trigger, or simply stream at-will.

In addition to this I would like to see separately an execd-style parser (stdin -> parse -> stdout) but perhaps the overhead is not worth it; I can easily import telegraf's input plugins into my own execd plugin.

@danielnelson - this seems like what we discussed at InfluxDays; would this suffice?

danielnelson:
> In addition to this I would like to see separately an execd-style parser (stdin -> parse -> stdout) but perhaps the overhead is not worth it; I can easily import telegraf's input plugins into my own execd plugin.

I'd like to have a similar plugin for processors, aggregators, and outputs.  I think the processor variant is what you are thinking of.  On the subject of related features I'd like to have some wrapper code that would allow you to take the same plugin format we use for compiled-in plugins and run it as an process for this plugin.  The idea being that you wouldn't need to make many changes to create a pull request from what starts as an external process.

> this seems like what we discussed at InfluxDays; would this suffice?

We are planning to bring in this pull request but I know I have some, mostly cosmetic, changes I'd like made but I haven't quite found the time to write them all down.

sgtsquiggs:
>I think the processor variant is what you are thinking of.

The specific use-case I am trying to tackle with this style plugin:

consume protobuf data from multicast socket
* Input = socket_listener
* Parser (data_format) = ???

Perhaps a vanilla way to do this would be to use socket_listener + data_format=value. Treat the binary as a string until a processor plugin receives it.

danielnelson:
Yeah, the string -> processor way would work, but I think a execd/subprocess parser would be good as well.

plugins/inputs/execd/execd.go
```diff
@@ -0,0 +1,176 @@
...
+
+	wg.Add(2)
+
+	go e.cmdReadOut(stdout, &wg)
+	go e.cmdReadErr(stderr, &wg)
+
+	wg.Wait()
+	e.cmd.Wait()
+
+	log.Printf("E! execd: %s terminated. Restart in one second...", e.Command)
```
sgtsquiggs: Wouldn't this restart even after being killed via `Stop()` ?

plugins/inputs/execd/execd.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+}
+
+func (e *Execd) cmdReadErr(out io.Reader, wg *sync.WaitGroup) {
+	defer wg.Done()
+
+	scanner := bufio.NewScanner(out)
+
+	for scanner.Scan() {
+		log.Printf("E! execd stderr output: %s", scanner.Text())
```
glinton: Change this to `%q` to prevent a multi-line log

plugins/inputs/execd/execd_unix.go
```diff
@@ -0,0 +1,33 @@
...
+	"fmt"
+	"io"
+	"syscall"
+
+	"github.com/influxdata/telegraf"
+)
+
+func (e *Execd) Gather(acc telegraf.Accumulator) error {
+	if e.cmd == nil || e.cmd.Process == nil {
+		return nil
```
glinton: Is this a good place to restart `cmdRun`?

plugins/inputs/execd/execd_unix.go
```diff
@@ -0,0 +1,33 @@
...
+		e.cmd.Process.Signal(syscall.SIGHUP)
+	case "SIGUSR1":
+		e.cmd.Process.Signal(syscall.SIGUSR1)
+	case "SIGUSR2":
+		e.cmd.Process.Signal(syscall.SIGUSR2)
+	case "STDIN":
+		io.WriteString(e.stdin, "\n")
+	case "none":
+	default:
+		return fmt.Errorf("execd: invalid signal: %s", e.Signal)
```
glinton: Don't include prefix when returning errors.

plugins/inputs/execd/execd.go
```diff
@@ -0,0 +1,176 @@
...
+	log.Printf("D! execd start program: %s", e.Command)
+
+	err = e.cmd.Start()
+	if err != nil {
+		return fmt.Errorf("execd error starting program: %s", err)
+	}
+
+	wg.Add(2)
+
+	go e.cmdReadOut(stdout, &wg)
```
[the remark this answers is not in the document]
jgraichen: Can you elaborate on this? Do you mean in-lining `cmdReadOut`?

(I'm new to Go...)

plugins/inputs/execd/execd_unix.go
jgraichen: It is surly possible.

It would mean that the process can be dead up to the configured `interval`. If the execd program is collecting metrics by itself or listens for events this can be quite long.

jgraichen:
Sorry, I haven't had much time recently. I'll try to tackle the issues but might need a few Go hints along the way. I'm not working regularly with Go.

I'll rebase and add additional commits for the different issues if that is fine. I can squash it when it is at an acceptable level.

plugins/inputs/execd/execd_unix.go
jgraichen: Changed for all returned errors.

plugins/inputs/execd/execd.go
jgraichen: Fixed.

plugins/inputs/execd/execd.go
jgraichen: For now I've added a bool flag that prevents the process from being restarted when the plugin has been stopped.

plugins/inputs/execd/execd.go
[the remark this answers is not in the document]
glinton: do:
```go
go func() {
  e.cmdReadOut(sdtout)
  wg.Done()
}()
```

plugins/inputs/execd/execd_unix.go
glinton: It seems like this is a reasonable trade-off to always restarting (even on `Stop`) I'm curious to hear what @danielnelson thinks

glinton:
>  I can squash it when it is at an acceptable level.

No need to squash, as we squash and merge. Squashing and force pushing just adds un-necessary time to pull request reviews.

plugins/inputs/execd/execd_unix.go
jgraichen: Restarting on `Stop` has been fixed. Processed will not be restarted if the plugin has been stopped (3d603c4).

plugins/inputs/execd/execd.go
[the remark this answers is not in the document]
jgraichen: Thanks. I've changed that for both functions.

jgraichen:
@glinton Any news here?

plugins/inputs/execd/execd.go
```diff
@@ -0,0 +1,186 @@
...
+
+	go func() {
+		e.cmdReadErr(stderr)
+		wg.Done()
+	}()
+
+	wg.Wait()
+	e.cmd.Wait()
+
+	if e.stopped {
```
glinton: I think it would be more straightforward to remove this logic and the 1s delayed restart and just call `e.cmdRun` in `Gather`.

plugins/inputs/execd/execd.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+func (e *Execd) SetParser(parser parsers.Parser) {
+	e.parser = parser
+}
+
+func (e *Execd) Start(acc telegraf.Accumulator) error {
+	e.acc = acc
+
+	args, err := shellquote.Split(e.Command)
```
glinton: Store `args` on the `Execd` struct, then just call `go e.cmdRun()` on line 70

plugins/inputs/execd/execd.go
```diff
@@ -0,0 +1,186 @@
...
+		return fmt.Errorf("unable to parse command: %s", err)
+	}
+
+	go e.cmdRun(args)
+
+	return nil
+}
+
+func (e *Execd) Stop() {
+	e.stopped = true
```
glinton: This would be obsolete when starting command in `Gather`.

Should there be a signal sent to `e.cmd.Process` to tell the command to stop? If so, should it be configurable?

plugins/inputs/execd/execd_unix.go
glinton: Let's move the restart handling to here, like so:
```suggestion
    err := e.cmdRun()
    if err != nil {
      return err
    }
    return nil
```

Same in the Windows version

jgraichen: I would need to run `go e.cmdRun()`. How does this work with the error handling?

plugins/inputs/execd/execd.go
jgraichen: Done.

plugins/inputs/execd/execd.go
jgraichen: It would probably be a good idea. Very nice would be to send TERM, wait for shutdown for a responsable amount of time, and finally kill with KILL. This would give the process some time to e.g. cleanup things.

I'm also just reading into process groups and `cmd.SysProcAttr` for Go, but I have absolutely no idea about Windows.

plugins/inputs/execd/execd.go
jgraichen: Personally I do dislike this as this couples the interval with recovering the process. I do have a few telegraf running with intervals upwards of one hour. They are mainly receiving things.

I can change this nevertheless if there is a solution to the discussion on `func Gather`. What does @danielnelson prefer?

plugins/inputs/execd/execd.go
```diff
@@ -0,0 +1,188 @@
...
+	e.args = args
+	go e.cmdRun()
+
+	return nil
+}
+
+func (e *Execd) Stop() {
+	e.stopped = true
+
+	if e.cmd == nil || e.cmd.Process == nil {
```
danielnelson: There is a race condition here, the process could start at any time during this function or even after the function exits.  What I suggest doing is to setup a WaitGroup and Context in Start() for a goroutine running the command in a loop, then in stop just call the CancelFunc and wait on the WaitGroup.

plugins/inputs/execd/execd.go
danielnelson: Either way works for me, I probably would have done it in Gather but I see your point about how we are using the interval for two functions.

Read my earlier comment about how to stop the process without a race condition first and then what I suggest is to either have the command goroutine sleep at the end of its internal loop or have it read from a channel that is written to by Gather.

If you decide to go with a fixed sleep then I would sleep for 10s, one second is too short IMO.

plugins/inputs/execd/execd.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+func (e *Execd) SetParser(parser parsers.Parser) {
+	e.parser = parser
+}
+
+func (e *Execd) Start(acc telegraf.Accumulator) error {
+	e.acc = acc
+
+	args, err := shellquote.Split(e.Command)
```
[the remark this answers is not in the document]
jgraichen: Done.

plugins/inputs/execd/execd.go
```diff
@@ -0,0 +1,188 @@
...
+	scanner := bufio.NewScanner(out)
+
+	for scanner.Scan() {
+		metrics, err := e.parser.Parse(scanner.Bytes())
+		if err != nil {
+			e.acc.AddError(fmt.Errorf("parse error: %s", err.Error()))
+		}
+
+		for _, metric := range metrics {
+			e.acc.AddFields(metric.Name(), metric.Fields(), metric.Tags(), metric.Time())
```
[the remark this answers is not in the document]
jgraichen: Done.

plugins/inputs/execd/execd.go
jgraichen: Fixed in 0e67fe56. The process is now managed in `runLoop()`. A cancelation context is created in `Start()` and used to stop the process. The termination itself is handled in `runLoop()`.

It seems to work but I'd really like to get an additional review here.

plugins/inputs/execd/execd.go
```diff
@@ -0,0 +1,188 @@
...
+func (e *Execd) Stop() {
+	e.stopped = true
+
+	if e.cmd == nil || e.cmd.Process == nil {
+		return
+	}
+
+	if err := e.cmd.Process.Kill(); err != nil {
+		log.Printf("E! [inputs.execd] FATAL error killing process: %s", err)
+	}
```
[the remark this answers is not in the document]
jgraichen: Should be fixed in 0e67fe56.

plugins/inputs/execd/execd.go
```diff
@@ -0,0 +1,188 @@
...
+
+	if err := scanner.Err(); err != nil {
+		e.acc.AddError(fmt.Errorf("Error reading stderr: %s", err.Error()))
+	}
+}
+
+func init() {
+	inputs.Add("execd", func() telegraf.Input {
+		return &Execd{
+			Signal: "none",
```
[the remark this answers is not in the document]
jgraichen: As in using `""` with the same meaning as `"none"` and removing `none` at all?

plugins/inputs/execd/execd.go
jgraichen: I've made the restart delay configurable in 30498402 with a default of 10 seconds. This should cover all cases. If a daemon needs to be restarted quickly to e.g. not loose events that is possible now.

plugins/inputs/execd/execd.go
jgraichen: Changed in 0e67fe56.

plugins/inputs/execd/execd_unix.go
jgraichen: Restart handling has been moved into a standalone goroutine to avoid a race condition in 0e67fe56.

plugins/inputs/execd/execd.go
```diff
@@ -0,0 +1,188 @@
...
+  signal = "none"
+
+  ## Data format to consume.
+  ## Each data format has its own unique set of configuration options, read
+  ## more about them here:
+  ## https://github.com/influxdata/telegraf/blob/master/docs/DATA_FORMATS_INPUT.md
+  data_format = "influx"
+`
+
+type Execd struct {
```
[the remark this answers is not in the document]
jgraichen: I do like `execd` but have no strong opinion here. I'm fine with renaming if that is wished.

jgraichen:
@danielnelson @glinton @sgtsquiggs I've update the plugin to use a standalone goroutine managing the daemon process to avoid a race condition (thx @danielnelson) and cleanup stopping. The restart delay is configurable now with a default of ten seconds. This should cover all cases including needing a quick restart to e.g. not loose events.

I'd really like to get another review on the new goroutine code. Thanks!

plugins/inputs/execd/execd.go
```diff
@@ -0,0 +1,209 @@
...
+		done := make(chan error, 1)
+		go func() {
+			done <- e.cmdRun()
+		}()
+
+		select {
+		case <-ctx.Done():
+			// Immediately exit process but with a graceful shutdown
+			// period before killing
+			internal.WaitTimeout(e.cmd, 0)
```
ssoroka: this isn't actually stopping the process. We need to send it a signal and close stdin.

jgraichen:
Thanks for all your patience and reviews.

Issue #4331, opened 2018-06-21
Large numbers are rounded.


glinton:
### Relevant telegraf.conf:
```
[[inputs.mysql]]
  servers = ["root:password@tcp(mysql:3306)/"]
  interval_slow = "30m"
```
### System info:

[Include Telegraf version, operating system name, and other relevant details]
 * Telegraf v1.7.0
 * Ubuntu 18.04
### Steps to reproduce:

1. Collect basic stats from mysql
2. Look for the field `max_join_size`

### Expected behavior:
```
mysql> SHOW GLOBAL VARIABLES;
| max_join_size | 18446744073709551615 |
...
max_join_size=18446744073709551615
```
### Actual behavior:
```
max_join_size=18446744073709552000
```
### Additional info:
```
telegraf_1  | mysql_variables,host=3549010a79d2,innodb_version=5.7.22,server=mysql:3306 max_sp_recursion_depth=0,max_prepared_stmt_count=16382,max_relay_log_size=0,max_write_lock_count=18446744073709552000,metadata_locks_cache_size=1024,myisam_data_pointer_size=6,max_points_in_geometry=65536,max_execution_time=0,max_heap_table_size=16777216,max_join_size=18446744073709552000,max_length_for_sort_data=1024,metadata_locks_hash_instances=8,multi_range_count=256,max_error_count=64,max_seeks_for_key=18446744073709552000,max_sort_length=1024,max_tmp_tables=32,max_user_connections=0,min_examined_row_limit=0,max_insert_delayed_threads=20 1529537880000000000
```

glinton:
Influx currently only supports "double precision floating point numbers." That is as precise as a `float64` can get in this case. [See here](https://play.golang.org/p/TD_YdCaIjzx) for an example of float precision.

kyunghoj:
So how can I deal with this issue? Shouldn't I collect this metric?

gangseok514:
@glinton Hello, I'm facing the same issue. In this case, Should I change max setting in mysql? or are there any workaround? Thanks.

glinton:
@kyunghoj I personally see no value in collecting a static value like this, so a non-exact value is likely fine. If you do need to collect it, you can lower the max setting in mysql as @gangseok514 mentioned.

glinton:
[...]

PR #4430, opened 2018-07-17
Fix uint64 value parse issue in mysql plugin

gangseok514:
Solve https://github.com/influxdata/telegraf/issues/4331

### Required for all PRs:

### DIFF
--- a/plugins/inputs/mysql/mysql.go
+++ b/plugins/inputs/mysql/mysql.go
@@ -1629,6 +1629,11 @@ func parseValue(value sql.RawBytes) (interface{}, bool) {
 	if val, err := strconv.ParseInt(string(value), 10, 64); err == nil {
 		return val, true
 	}
+
+	if val, err := strconv.ParseUint(string(value), 10, 64); err == nil {
+		return val, true
+	}
+
 	if val, err := strconv.ParseFloat(string(value), 64); err == nil {
 		return val, true
 	}


gangseok514:
@danielnelson Hello, Could you review this PR?

glinton:
This change would break some existing stacks. For example, you cannot change a value type in influx so any new metrics, properly parsed, would get dropped if the value in question was already stored as a float64.

danielnelson:
With this implementation the field will vary it's type depending on the value, which will cause problems even if we ignore what the current type is.  Perhaps we can find a better method for assigning types than this fallback list.  We should check what type the database driver believes these fields are, does it differentiate between int64, uint64 and float64?

gangseok514:
I understand what you concern. But I couldn't find other problems except @glinton's example.
Anyway, I faced a critical issue related to this.
I'm using fluentd and telegraf sends metric to fluentd. Fluentd cannot parse the value(Bignum too big to convert to unsigned long long) because of this issue.
I had to drop the field manually but the values can be useful for some users.

As you know, mariadb has this value as default in some configuration variables. https://mariadb.com/kb/en/library/server-system-variables/#max_join_size

I think telegraf is the best metric agent. so I'm using it with influxdb very well in my k8s clusters.

But now I need to customize telegraf for this. I hope to contribute this
Thanks

danielnelson:
> Fluentd cannot parse the value(Bignum too big to convert to unsigned long long) because of this issue.

This seems like it might be an independent issue, how are you sending metrics from Telegraf to fluentd?

danielnelson:
So you are using the socket_writer + `data_format = "json"`?

JSON has arbitrary length and precision numbers, not float/int, I think the `.0` is left out if it not needed.  Here are the details of the float to number encoding: https://golang.org/src/encoding/json/encode.go?s=6456:6499#L537

You may want to capture the JSON output from Telegraf and let the fluentd team know about it, to me it sounds like an issue for them to take a look at.

srebhan:
@gangseok514: The mysql-driver supports ColumnTypeScanType() to derive the actual type from the DB/scheme-definition. It includes information about signedness of the column. Instead of guessing the types, we should use this information to parse the value.
So I agree with the other guys that this is not the right fix.

sjwang90:
Closing this issue. If anyone is interesting in fixing this based off the above discussion feel free to submit a PR.
> The mysql-driver supports ColumnTypeScanType() to derive the actual type from the DB/scheme-definition. It includes information about signedness of the column. Instead of guessing the types, we should use this information to parse the value.

PR #4431, opened 2018-07-17
Add ip restriction for the prometheus_client output

dupondje:
Sometimes you want to protect the exported /metrics path from the prometheus_client.
It was already possible to secure it with a password.

This patch adds the option to allow only defined CIDR ranges to access the /metrics.

### DIFF
--- a/plugins/outputs/prometheus_client/prometheus_client.go
+++ b/plugins/outputs/prometheus_client/prometheus_client.go
@@ -5,6 +5,7 @@ import (
 	"crypto/subtle"
 	"fmt"
 	"log"
+	"net"
 	"net/http"
 	"os"
 	"regexp"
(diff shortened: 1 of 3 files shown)


danielnelson:
@dupondje The code here is very good, but I feel like this addition won't be in very wide demand, and I don't want to support too many connection limitations as they are endless.  Why not just set this up in your system firewall instead?

plugins/outputs/prometheus_client/prometheus_client.go
```diff
@@ -110,6 +115,21 @@ func (p *PrometheusClient) basicAuth(h http.Handler) http.Handler {
 			}
 		}
 
+		matched := false
+		remoteIPs, _, _ := net.SplitHostPort(r.RemoteAddr)
+		remoteIP := net.ParseIP(remoteIPs)
+		for _, iprange := range p.IPRange {
+			_, ipNet, _ := net.ParseCIDR(iprange)
```
glinton: Handle the error so the next line doesn't panic if user passes bad configuration

dupondje:
Adjusted the code like requested.
Also put everything inside a if len(p.IPRange) > 0 {, cause we don't need to parse the RemoteIP if we don't set a ip_range restriction ofcourse. Saves some cycles :)

Issue #2108, opened 2016-11-30
CSV data input format (for use with exec, tail)


stanch:
It would be great to have CSV among the supported data input formats. In particular, this allows to leverage the `tail` plugin to integrate with existing tools outputting CSV, instead of e.g. using a named pipe (see also: https://github.com/influxdata/telegraf/issues/2089) as an adapter.

According to the [input format documentation](https://github.com/influxdata/telegraf/blob/master/docs/DATA_FORMATS_INPUT.md), there are four components that need to be extracted:
> 1. Measurement Name
> 2. Tags
> 3. Fields
> 4. Timestamp

Here is the suggested configuration format that allows to map those to CSV columns:
```toml
# The new data format
data_format = "csv"
data_separator = ","
data_quote_char = "\""

# List of column names in the CSV, in order of appearance on each line
data_columns = [
  "ts",
  "system_id",
  "client_id",
  "request_type",
  "latency"
]

# The column containing the measurement name
name_column = "request_type"

# The column containing the timestamp
timestamp_column = "ts"

# List of columns containing tags
tag_columns = [
  "system_id",
  "client_id"
]

# List of columns containing fields
field_columns = [
  "latency"
]
```

Rushi-Kumar:
Is it supported now??

joeltstephen:
Hi @stanch, Is it supported now ?

stanch:
@joeltstephen I am not a contributor, so I don’t know for sure, but the page I linked to in the original issue description still does not mention it, so I assume it’s not supported. I suggest you to look into https://github.com/influxdata/telegraf/tree/master/plugins/inputs/logparser, it might be an alternative solution.

PR #4439, opened 2018-07-19
Feature/2108 - csv parser

maxunt:
closes: https://github.com/influxdata/telegraf/issues/2108

### Required for all PRs:


plugins/parsers/csv/parser.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+func (p *CSVParser) Parse(buf []byte) ([]telegraf.Metric, error) {
+	r := bytes.NewReader(buf)
+	csvReader := csv.NewReader(r)
+	csvReader.FieldsPerRecord = len(p.DataColumns)
+	if p.Delimiter != "" {
+		runeStr := []rune(p.Delimiter)
+		if len(runeStr) > 1 {
+			return nil, fmt.Errorf("delimiter must be a single character, got: %v", p.Delimiter)
```
glinton: Pedantic, but can you use the non-default verb to print (`%s` in this case)

plugins/parsers/csv/parser.go
```diff
@@ -0,0 +1,156 @@
...
+
+	for _, tagName := range p.TagColumns {
+		if recordFields[tagName] == "" {
+			return nil, fmt.Errorf("could not find field: %v", tagName)
+		}
+		tags[tagName] = recordFields[tagName]
+	}
+
+	for _, fieldName := range p.FieldColumns {
+		if recordFields[fieldName] == "" {
```
glinton: Consider using `value, ok := recordFields[fieldName]` then check if `!ok` and return, line 115 becomes unnecessary.

plugins/parsers/csv/parser.go
```diff
@@ -0,0 +1,156 @@
...
+		m, err := p.parseRecord(record)
+		if err != nil {
+			return metrics, err
+		}
+		metrics = append(metrics, m)
+	}
+	return metrics, nil
+}
+
+//does not use any information in header and assumes DataColumns is set
```
glinton: If a comment is directly above an exported function, start it with `// FunctionName ...`

plugins/parsers/registry.go
```diff
@@ -139,12 +150,49 @@ func NewParser(config *Config) (Parser, error) {
...
+func newCSVParser(metricName string,
+	header bool,
+	delimiter string,
+	dataColumns []string,
+	tagColumns []string,
+	fieldColumns []string,
+	nameColumn string,
+	timestampColumn string,
+	timestampFormat string,
+	defaultTags map[string]string) (Parser, error) {
```
glinton: Returning an error isn't useful here. Where it's an unexported function not matching any interface, I'd remove it, or remove the function altogether and just instantiate a `CSVParser` on line 154.

plugins/parsers/csv/parser.go
```diff
@@ -0,0 +1,156 @@
...
+	"encoding/csv"
+	"fmt"
+	"strconv"
+	"time"
+
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/metric"
+)
+
+type CSVParser struct {
```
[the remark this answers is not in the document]
danielnelson: It doesn't look like the quote character can be customized when using the Go csv parser, and we would want to stick to this implementation.

The other options sound great, but I don't think they are must have and we could add them later depending on available time.

danielnelson:
@maxunt Don't forget to rebase/merge so that the unrelated grok documentation is no longer present.

plugins/parsers/csv/parser.go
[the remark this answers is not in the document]
maxunt: I can definitely add features to allow comments and trimming.  I know quote characters aren't supported by the go csv parser.  Regarding column skipping, there is already functionality for that by simply not adding the column name to either `csv_tag_columns` or `csv_field_columns.`  The header row count raises a few issues about how a header with more than one line would be interpreted, unless we decide to skip it entirely.  It would most likely not mesh well with the function to extract column names from the header. We would have to decide how we want to configure that; I think we could probably pair it with the row skipping configuration.

plugins/parsers/csv/parser_test.go
```diff
@@ -0,0 +1,193 @@
...
+		"third":  true,
+		"fourth": "hello",
+	}
+
+	metrics, err := p.Parse([]byte(testCSV))
+	require.NoError(t, err)
+
+	expectedTags := make(map[string]string)
+	goodMetric, err1 := metric.New("test_value", expectedTags, expectedFields, time.Unix(0, 0))
+	returnedMetric, err2 := metric.New(metrics[0].Name(), metrics[0].Tags(), metrics[0].Fields(), time.Unix(0, 0))
```
danielnelson: I would consider adding a TimeFunc field to the parser like this:

https://github.com/influxdata/telegraf/blob/019d265167eac47ed858766d77cc153f92129763/metric/builder.go#L9-L25

In the normal case it will be set to time.Time, and in the testcase it will be func() { return time.Unix(0,0) }

Can you use this function here for comparing metrics?

https://github.com/influxdata/telegraf/blob/019d265167eac47ed858766d77cc153f92129763/testutil/metric.go#L11

plugins/parsers/csv/parser_test.go
```diff
@@ -0,0 +1,193 @@
...
+	require.NoError(t, err1)
+	require.NoError(t, err2)
+
+	//deep equal fields
+	require.True(t, reflect.DeepEqual(goodMetric.Fields(), returnedMetric.Fields()))
```
danielnelson: `require.Equal(t, goodMetric.Fields(), returnedMetric.Fields())`

Call these `expected/actual`, or `want/got`.

plugins/parsers/csv/parser_test.go
```diff
@@ -0,0 +1,193 @@
...
+	expectedFields := map[string]interface{}{
+		"first":  int64(4),
+		"second": 9.9,
+		"third":  true,
+		"fourth": "name_this",
+	}
+
+	metrics, err := p.Parse([]byte(testCSV))
+	require.NoError(t, err)
+	require.Equal(t, true, reflect.DeepEqual(expectedFields, metrics[0].Fields()))
```
danielnelson: `require.Equal(t, expectedFields, metrics[0].Fields())`

Check other tests and make sure you are using this everywhere.

plugins/parsers/csv/parser_test.go
```diff
@@ -0,0 +1,193 @@
...
+		"first":  3.3,
+		"second": int64(4),
+		"third":  true,
+		"fourth": "hello",
+	}
+
+	metrics, err := p.Parse([]byte(testCSV))
+	for k := range metrics[0].Fields() {
+		log.Printf("want: %v, %T", expectedFields[k], expectedFields[k])
+		log.Printf("got: %v, %T", metrics[0].Fields()[k], metrics[0].Fields()[k])
```
danielnelson: Boo! no logging in tests

plugins/parsers/csv/parser.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+	m, err := p.parseRecord(record)
+	if err != nil {
+		return nil, err
+	}
+	return m, nil
+}
+
+func (p *CSVParser) parseRecord(record []string) (telegraf.Metric, error) {
+	recordFields := make(map[string]string)
```
danielnelson: I think you won't need this intermediate map if you make fields implicit.

plugins/parsers/csv/parser.go
```diff
@@ -0,0 +1,181 @@
...
+	for _, fieldName := range p.FieldColumns {
+		value, ok := recordFields[fieldName]
+		if !ok {
+			return nil, fmt.Errorf("could not find field: %v", fieldName)
+		}
+
+		// attempt type conversions
+		if iValue, err := strconv.Atoi(value); err == nil {
+			fields[fieldName] = iValue
+		} else if fValue, err := strconv.ParseFloat(value, 64); err == nil {
```
danielnelson: This will require all floats to have a decimal part to avoid type mismatch errors.  @goller Is this going to work for us in the future?

docs/DATA_FORMATS_INPUT.md
```diff
@@ -2,6 +2,17 @@
 
 Telegraf is able to parse the following input data formats into metrics:
 
+<<<<<<< HEAD
```
danielnelson: This file needs fixed up due to merge issues.  I just merged your updates to the JSON parser so you may need to update again too.

maxunt: im not sure i correctly followed your new format for the INPUT_DATA_FORMATS file when i resolved the merge conflict if you could take a look at that.  I think the csv is missing the proper link to its section

plugins/parsers/csv/parser.go
```diff
@@ -0,0 +1,202 @@
...
+	if len(p.ColumnNames) == 0 {
+		for i := 0; i < p.HeaderRowCount; i++ {
+			header, err := csvReader.Read()
+			if err != nil {
+				return nil, err
+			}
+			//concatenate header names
+			for i := range header {
+				name := header[i]
+				name = strings.Trim(name, " ")
```
danielnelson: I think we may not want to trim the strings.

plugins/parsers/csv/parser.go
```diff
@@ -0,0 +1,202 @@
...
+	TimestampColumn   string
+	TimestampFormat   string
+	DefaultTags       map[string]string
+}
+
+func (p *Parser) compile(r *bytes.Reader) (*csv.Reader, error) {
+	csvReader := csv.NewReader(r)
+	// ensures that the reader reads records of different lengths without an error
+	csvReader.FieldsPerRecord = -1
+	if p.Delimiter != "" {
```
danielnelson: Let's compute the effective delimiter and comment values as part of a New() function, then when Parse is called we can create the reader without needing to worry to redo this work.

plugins/parsers/csv/parser.go
```diff
@@ -0,0 +1,202 @@
...
+		}
+		p.ColumnNames = headerNames[p.SkipColumns:]
+	} else if len(p.ColumnNames) > 0 {
+		// if columns are named, just skip header rows
+		for i := 0; i < p.HeaderRowCount; i++ {
+			csvReader.Read()
+		}
+	} else if p.HeaderRowCount == 0 && len(p.ColumnNames) == 0 {
+		// if there is no header and no DataColumns, that's an error
+		return nil, fmt.Errorf("there must be a header if `csv_data_columns` is not specified")
```
danielnelson: This test can also go in the New function.

plugins/parsers/csv/parser.go
```diff
@@ -0,0 +1,202 @@
...
+func (p *Parser) ParseLine(line string) (telegraf.Metric, error) {
+	r := bytes.NewReader([]byte(line))
+	csvReader, err := p.compile(r)
+	if err != nil {
+		return nil, err
+	}
+
+	// if there is nothing in DataColumns, ParseLine will fail
+	if len(p.ColumnNames) == 0 {
+		return nil, fmt.Errorf("[parsers.csv] data columns must be specified")
```
danielnelson: Can go in New function

plugins/parsers/csv/parser.go
```diff
@@ -0,0 +1,202 @@
...
+func (p *Parser) parseRecord(record []string) (telegraf.Metric, error) {
+	recordFields := make(map[string]interface{})
+	tags := make(map[string]string)
+
+	// skip columns in record
+	record = record[p.SkipColumns:]
+	for i, fieldName := range p.ColumnNames {
+		if i < len(record) {
+			value := record[i]
+			value = strings.Trim(value, " ")
```
danielnelson: Not sure if we want to trim

plugins/parsers/csv/parser.go
```diff
@@ -0,0 +1,202 @@
...
+
+	// add default tags
+	for k, v := range p.DefaultTags {
+		tags[k] = v
+	}
+
+	// will default to plugin name
+	measurementName := p.MetricName
+	if recordFields[p.MeasurementColumn] != nil {
+		measurementName = recordFields[p.MeasurementColumn].(string)
```
danielnelson: This could panic if the column is not a string, perhaps we should pull the value from `record`?

plugins/parsers/csv/parser.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+
+	for _, tagName := range p.TagColumns {
+		if recordFields[tagName] == nil {
+			return nil, fmt.Errorf("could not find field: %v", tagName)
+		}
+		tags[tagName] = recordFields[tagName].(string)
```
danielnelson: Will panic if not a string.

plugins/parsers/csv/parser.go
```diff
@@ -0,0 +1,202 @@
...
+		delete(recordFields, tagName)
+	}
+
+	metricTime := time.Now()
+	if p.TimestampColumn != "" {
+		if recordFields[p.TimestampColumn] == nil {
+			return nil, fmt.Errorf("timestamp column: %v could not be found", p.TimestampColumn)
+		}
+		tStr := recordFields[p.TimestampColumn].(string)
```
danielnelson: Could panic, maybe use `record` with two return value form:
```
timeColumn := record[p.TimestampColumn]
if timeColumn != nil {
    //
}

timeString, ok := col.(string); ok {
}
```

Might be easier to deal with errors if you put this into a function.

plugins/parsers/csv/parser.go
```diff
@@ -0,0 +1,202 @@
...
+		if p.TimestampFormat == "" {
```
danielnelson: You can verify this in the New function, and then treat this as a struct invariant.

PR #4440, opened 2018-07-19
Update docker and start with older API version

### LINKED ISSUE #4434: Docker client version is hardcoded
Hi,

Docker client version is hardcoded in client.go

```
var (
	version        = "1.24"
	defaultHeaders = map[string]string{"User-Agent": "engine-api-cli-1.0"}
)
```

We have system with old docker version:
```
 $ dpkg -l | grep -i docker
ii  docker                               1.5-1                                      amd64        System tray for KDE3/GNOME2 docklet applications
ii  docker-engine                        1.9.1-0~trusty                             amd64        Docker: the open-source application container engine

$ docker version
Client:
 Version:      1.9.1
 API version:  1.21
 Go version:   go1.4.2
 Git commit:   a34a1d5
 Built:        Fri Nov 20 13:12:04 UTC 2015
 OS/Arch:      linux/amd64

Server:
 Version:      1.9.1
 API version:  1.21
 Go version:   go1.4.2
 Git commit:   a34a1d5
 Built:        Fri Nov 20 13:12:04 UTC 2015
 OS/Arch:      linux/amd64
```
The docker input plugin is returning the following error:

`E! Error in plugin [inputs.docker]: Error response from daemon: client is newer than server (client API version: 1.24, server API version: 1.21`

Regards,

glinton:
Resolves #4434 by setting a lower starting api version. This also implements the latest stable docker release lib


plugins/inputs/docker/client.go
```diff
@@ -27,29 +26,26 @@ type Client interface {
...
-	}
-
 	transport := &http.Transport{
 		TLSClientConfig: tlsConfig,
 	}
-	sockets.ConfigureTransport(transport, proto, addr)
 	httpClient := &http.Client{Transport: transport}
 
-	client, err := docker.NewClient(host, version, httpClient, defaultHeaders)
+	client, err := docker.NewClientWithOpts(docker.WithHTTPHeaders(defaultHeaders), docker.WithHTTPClient(httpClient), docker.WithVersion(version), docker.WithHost(host))
```
danielnelson: Can you line wrap this, I aim for ~78 chars.

plugins/inputs/docker/client.go
```diff
@@ -27,29 +26,26 @@ type Client interface {
...
 	if err != nil {
 		return nil, err
 	}
+
+	client.NegotiateAPIVersion(context.Background())
```
danielnelson: Does this override the initial version above?

Gopkg.toml
```diff
@@ -31,46 +29,43 @@
...
-[[constraint]]
-  name = "github.com/docker/go-connections"
-  version = "0.3.0"
-#  version = "0.2.1"
+[[override]]
+  name = "github.com/docker/distribution"
+  revision = "edc3ab29cdff8694dd6feb85cfeb4b5f1b38ed9c" # v18.05.0-ce
+
+[[override]]
+  name = "github.com/vmware/vic"
```
danielnelson: I'm guessing this is pulled in on this pr by accident?

Gopkg.lock
```diff
@@ -884,90 +1111,201 @@
...
   packages = [
     ".",
     "bson",
     "internal/json",
     "internal/sasl",
-    "internal/scram"
+    "internal/scram",
   ]
-  revision = "3f83fa5005286a7fe593b055f0d7771a7dce4655"
+  pruneopts = ""
```
danielnelson: I think we should use a tagged version of dep for this file to avoid issues and churn.  Can you update the makefile to pull a specific version?

plugins/inputs/docker/client.go
glinton: Actually no, it apparently only updates if the server api version is lower than the client version (specified above) since 1.24 is the lowest reported server api version, `NegotiateAPIVersion` doesn't do anything for us. I'll remove it

Gopkg.toml
glinton: No, for whatever reason, `dep` decided to complain this time about `vic` and their dependency on `Sirupsen/logrus` where we already have lowercase `sirupsen/logrus` vendored. This overrides their version to pull from a source where the imports have been updated.

Gopkg.lock
glinton: Great idea

glinton: So, because `dep` imports their `internal` package, updating the makefile to `go get -u gopkg.in/golang/dep.v0` won't work. Their suggested method of installing the latest release is with `curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh` which isn't guaranteed to work on all systems. Though installing a release is better than master.

danielnelson: I won't be able to look myself in the mirror if I suggest `curl | bash` to anyone, but I will update the Makefile in a separate pr.  In the meantime lets just manually install v0.4.1.

glinton: i definitely hear you there, we could always `go get -d g.../dep && cd $GOPATH/.../dep && git checkout v0.4.1 && go install` but i agree, another commit

PR #4441, opened 2018-07-20
Add benchmark input plugin

danielnelson:
This is a quick performance and debugging tool I am using for testing output performance.  This improves on my previous method of spamming data to the socket_listener input, mainly because the data does not have to be reparsed.

### Required for all PRs:


danielnelson:
It has been suggested to me that we should try to improve the name.  Any suggestions would be appreciated.

PR #4448, opened 2018-07-21
inheritance of environment variables for the exec plugin

angelopoerio:
### Required for all PRs:

I think it would be very useful for the processes spawned by the exec plugin to to ihnerit the environment variables. It is very common to use them to pass informations about the state of the application and/or the environment itself. It is the used way in Kubernetes and Mesos/Marathon for example.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ### Release Notes
 
+- With the addition of the standalone `grok` input data format, the
+  `logparser` input plugin has been deprecated in favor of using the `tail`
+  input plugin combined with `data_format="grok"` .
+
 ### New Inputs
 
 - [file](./plugins/inputs/file/README.md) - Contributed by @maxunt
--- a/plugins/inputs/exec/exec.go
+++ b/plugins/inputs/exec/exec.go
@@ -3,6 +3,7 @@ package exec
 import (
 	"bytes"
 	"fmt"
+	"os"
 	"os/exec"
 	"path/filepath"
 	"runtime"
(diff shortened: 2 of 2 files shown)


danielnelson:
From my testing the exec plugin already passes the environment down to the sub process without this change.  @angelopoerio Can you double check that this change is needed?

angelopoerio:
Hi Daniel,

you are right the environment variable get passed in a "normal" environment
but in this case I am using this one:
https://github.com/phusion/baseimage-docker.
It implements under the hood a minimal init system and has a quite
particular way of handling env variables (
https://github.com/phusion/baseimage-docker#modifying_envvars).
This causes the condition that makes the child process not ihnerit the
variables. It is an edge case so better to keep it generic.



2018-07-23 20:46 GMT+02:00 Daniel Nelson <notifications@github.com>:

danielnelson:
I don't understand how the environment could change the environment inheriting works, here is how the Env field is initialized when the process is ran:

https://golang.org/src/os/exec/exec.go?s=5400:5445#L189

It still seems to me that this patch wouldn't change anything.  Can you write a unittest or a sample script that shows the difference?

danielnelson:
@angelopoerio Let me know if I need to reopen this pr.

angelopoerio:
Hi Daniel,

sorry for the late answer. Took some time to investigate better, at the end
I discovered that the environment variables got cleaned for other reasons.
Under the hood golang uses execve() that passes the env of the parent
process to the child (always).

2018-07-28 3:41 GMT+02:00 Daniel Nelson <notifications@github.com>:

PR #4450, opened 2018-07-22
Ipmi.v2.schema

### LINKED ISSUE #3416: impi_sensors should split multiple sensor readings with the same name
## Feature Request

I'm monitoring a Dell R630 with telegraf and ipmi_sensors. The server has two CPU sockets and therefore reports temperatures for both CPUs. It would be nice, if I could access those two readings seperately as they are currently combined into one reading.

### Proposal:

If multiple instances of a sensor are detected, they should be split. Example the Temp sensor:
```
$ ipmitool -I lanplus -H 10.1.100.103 -U root sdr
Password:
SEL              | Not Readable      | ns
Intrusion        | 0x00              | ok
Fan1A            | 5160 RPM          | ok
Fan2A            | 5160 RPM          | ok
Fan3A            | 5040 RPM          | ok
Fan4A            | 4920 RPM          | ok
Fan5A            | 5160 RPM          | ok
Fan6A            | 5040 RPM          | ok
Inlet Temp       | 18 degrees C      | ok
CPU Usage        | 8 percent         | ok
IO Usage         | 0 percent         | ok
MEM Usage        | 0 percent         | ok
SYS Usage        | 9 percent         | ok
Exhaust Temp     | 29 degrees C      | ok
Temp             | 39 degrees C      | ok
Temp             | disabled          | ns
... 
```

### Current behavior:

The output comes from a server, which has only one CPU but it still applies:
```
$ telegraf --config-directory /etc/telegraf/telegraf.d/ --input-filter ipmi_sensor --test
* Plugin: inputs.ipmi_sensor, Collection 1
* Internal: 30s
...
> ipmi_sensor,unit=degrees_c,host=monitoring,name=temp,server=esxi0-mgmt.example.org status=1i,value=42 1509562189000000000
> ipmi_sensor,host=monitoring,name=temp,server=esxi0-mgmt.example.org value=0,status=0i 1509562189000000000
...
```
Two values are inserted for `temp`: 42 and 0. This makes it either difficult (only on CPU) or impossible (two or more CPUs) to find out the temperature of CPU0/CPU1/.../CPUn.

### Desired behavior:

Instead of just one `temp` reading, a second one should be created to hold the second value.
I propose the following naming schemes for further discussion:

1. `temp_0`, `temp_1`, ..., `temp_n`
Could be not that easy to implement, as you would need to look first for multiple instances before recording data.
2. `temp`, `temp_1`, ..., `temp_n`
First measurement is always called by its name. Further instances are suffixed with a number

For backwards compatibility reasons I would suggest the second option. This is also easier to implement, as you just need to create a new one if one with the same name already exists for the current parsing run.

Currently it is not possible to distinguish both temperature readings after the data has left telegraf, as only one measurement exists. Monitoring a multi CPU system's temperatures is therefore not possible. For servers with only one CPU but two sockets it is possbile but difficult to get the temperature as you need to employ e.g. the `max` function of InfluxDB to get only the non-zero value out of the series.

jhg03a:
### Required for all PRs:

Fixes #3416

### DIFF
--- a/plugins/inputs/ipmi_sensor/README.md
+++ b/plugins/inputs/ipmi_sensor/README.md
@@ -8,6 +8,10 @@ If no servers are specified, the plugin will query the local machine sensor stat
 ```
 ipmitool sdr
 ```
+or with the version 2 schema:
+```
+ipmitool sdr elist
+```
 
 When one or more servers are specified, the plugin will use the following command to collect remote host sensor stats:
 
@@ -41,19 +45,36 @@ ipmitool -I lan -H SERVER -U USERID -P PASSW0RD sdr
 
   ## Timeout for the ipmitool command to complete. Default is 20 seconds.
   timeout = "20s"
+
+  ## Schema Version: (Optional, defaults to version 1)
+  schemaVersion = 2
 ```
 
 ### Measurements
 
+Version 1 schema:
 - ipmi_sensor:
   - tags:
     - name
     - unit
+    - host
     - server (only when retrieving stats from remote servers)
   - fields:
-    - status (int)
+    - status (int, 1=ok status_code/0=anything else)
     - value (float)
 
+Version 2 schema:
+- ipmi_sensor:
+  - tags:
+    - name
+    - entity_id (can help uniquify duplicate names)
+    - status_code (two letter code from IPMI documentation)
+    - status_desc (extended status description field)
+    - unit (only on analog values)
+    - host
+    - server (only when retrieving stats from remote)
+  - fields:
+    - value (float)
 
 #### Permissions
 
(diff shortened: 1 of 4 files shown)


plugins/inputs/ipmi_sensor/ipmi.go
```diff
@@ -93,23 +102,32 @@ func (m *Ipmi) parse(acc telegraf.Accumulator, server string) error {
...
+		return parseV2(acc, hostname, string(out))
+	}
+	return parseV1(acc, hostname, string(out))
+}
 
+func parseV1(acc telegraf.Accumulator, hostname string, cmdOut string) error {
 	// each line will look something like
 	// Planar VBAT      | 3.05 Volts        | ok
-	lines := strings.Split(string(out), "\n")
+	lines := strings.Split(cmdOut, "\n")
```
danielnelson: I'll bet this is was an existing bug in Windows due to only allowing CR line endings.  What would be ideal is if we switched to bufio.Scanner here.

plugins/inputs/ipmi_sensor/ipmi.go
```diff
@@ -143,13 +159,76 @@ func (m *Ipmi) parse(acc telegraf.Accumulator, server string) error {
...
+			fields["value"] = 0.0
+			// Extended status descriptions aren't required, in which case for consistency re-use the status code
+			if descriptionResults["status"] != "" {
+				tags["status_desc"] = transform(descriptionResults["status"])
+			} else {
+				tags["status_desc"] = transform(ipmiFields["status_code"])
+			}
+		}
+
+		acc.AddFields("ipmi_sensor", fields, tags, time.Now())
```
danielnelson: Call time.Now() once before the loop, then pass the same time in for all metrics.  Usually doesn't matter due to timestamp rounding but it is still a good idea.

plugins/inputs/ipmi_sensor/ipmi.go
```diff
@@ -143,13 +159,76 @@ func (m *Ipmi) parse(acc telegraf.Accumulator, server string) error {
...
+	}
+
+	return nil
+}
+
+// extractFieldsFromRegex consumes a regex with named capture groups and returns a kvp map of strings with the results
+func extractFieldsFromRegex(regex string, input string) map[string]string {
+	re := regexp.MustCompile(regex)
```
danielnelson: We should compile all the regex once at the package level, then reuse the compiled regex here.

plugins/inputs/ipmi_sensor/ipmi.go
```diff
@@ -143,13 +159,76 @@ func (m *Ipmi) parse(acc telegraf.Accumulator, server string) error {
...
+		if hostname != "" {
+			tags["server"] = hostname
+		}
+		tags["entity_id"] = transform(ipmiFields["entity_id"])
+		tags["status_code"] = trim(ipmiFields["status_code"])
+		fields := make(map[string]interface{})
+		descriptionResults := extractFieldsFromRegex(`^(?P<analogValue>[0-9.]+)\s(?P<analogUnit>.*)|(?P<status>.+)|^$`, trim(ipmiFields["description"]))
+		// This is an analog value with a unit
+		if descriptionResults["analogValue"] != "" && len(descriptionResults["analogUnit"]) >= 1 {
+			fields["value"] = aToFloat(descriptionResults["analogValue"])
```
danielnelson: I think if the value cannot be converted to a float then we should skip the line.

plugins/inputs/ipmi_sensor/README.md
```diff
@@ -68,24 +89,36 @@ KERNEL=="ipmi*", MODE="660", GROUP="telegraf"
...
-ipmi_sensor,unit=volts,name=planar_3.3v status=1i,value=3.28 1458488465012861875
-ipmi_sensor,unit=volts,name=planar_vbat status=1i,value=3.04 1458488465013072508
-ipmi_sensor,unit=rpm,name=fan_1a_tach status=1i,value=2610 1458488465013137932
-ipmi_sensor,unit=rpm,name=fan_1b_tach status=1i,value=1775 1458488465013279896
+ipmi_sensor,name=uid_light,entity_id=23.1,status_code=ok,status_desc=ok,host=foo.bar.com value=0 1517125474000000000
+ipmi_sensor,host=foo.bar.com,name=sys._health_led,entity_id=23.2,status_code=ok,status_desc=ok value=0 1517125474000000000
+ipmi_sensor,status_code=ok,unit=watts,status_desc=presence_detected,host=foo.bar.com,name=power_supply_1,entity_id=10.1 value=110 1517125474000000000
+ipmi_sensor,host=foo.bar.com,name=power_supply_2,entity_id=10.2,status_code=ok,unit=watts,status_desc=presence_detected value=125 1517125474000000000
+ipmi_sensor,name=power_supplies,entity_id=10.3,status_code=ok,status_desc=fully_redundant,host=foo.bar.com value=0 1517125474000000000
+ipmi_sensor,unit=percent,status_desc=transition_to_running,host=foo.bar.com,name=fan_1,entity_id=7.1,status_code=ok value=43.12 1517125474000000000
```
danielnelson: If you regen this output with Telegraf 1.7, it should sort the tags.  You can leave out the host tag since it is added everywhere.

envy:
Hi, sorry for the late reply.

The solution with the entity ids looks good!

Issue #4453, opened 2018-07-23
win_perf_counters '% Processor Time' sent invalid data periodically (every ~2 minutes)


nognomar:
### Relevant telegraf.conf:
https://gist.github.com/nognomar/95c8c0c086106e2fb66e8f9d8940638f
### System info:

[Include Telegraf version, operating system name, and other relevant details]
telegraf 1.7.2
Windows server 2012R2
Processor [Intel-Xeon-Processor-E5-2620](https://ark.intel.com/products/64594/Intel-Xeon-Processor-E5-2620-15M-Cache-2_00-GHz-7_20-GTs-Intel-QPI) (2 processors)

Every 2 (sometime 4) minutes telegraf detect 100% CPU loading. But I can't detect that by other methods (task manager, sysinternals process explorer, **performance monitor**)

I create simple tool in C# and can't handle high cpu usages too:
https://gist.github.com/nognomar/05cbcb0ca8842c537e163557a4f96517

danielnelson:
@vlastahajek Any idea what could cause this?

vlastahajek:
@nognomar, does this happen also on another machine, if you have one?

Could, please, also try the ```typeperf``` tool if it matches what Telegraf detects?
Simply run from following command line: ```typeperf "\Processor(_Total)\% Processor Time"```

vlastahajek:
Also, it's worth to disable refreshing counters configuration by setting performance counters global parameter: ```CountersRefreshInterval="0s"```

vlastahajek:
For easy comparison of the performance counters plugin output and the typeperf output do following two simple steps to get typeperf output to InfluxDb (and display in Chronograf or Grafana):
1. redirect typeperf 's output to file: ```typeperf "\Processor(_Total)\% Processor Time" >c:\tmp\typeperf_cpu.log```
1. setup [the Telegraf's tail plugin](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/tail) to read typeperf output:
````
[[inputs.tail]]
  files = ['c:\tmp\typeperf_cpu.log']
  from_beginning = true
  data_format = "grok"
  grok_patterns = ['"%{DATESTAMP:timestamp:ts-"01/02/2006 15:04:05.000"}","%{NUMBER:value:float}"']
  name_override = "typeperf_cpu"
  grok_timezone = "Local"

danielnelson:
[...]

PR #4483, opened 2018-07-27
Fixing #4453 by disabling counter refresh by default

vlastahajek:
- disabled counter refreshing by default - fix for #4453 
- added debugging prints
- added other troubleshooting tip


vlastahajek:
I've fixed conflict and new commit made running tests again, with good result.

danielnelson:
Am I correct that this would mean that, by default, only instances seen by the initial Gather would be seen by the plugin?  It seems like for dynamic types, such as process, this would be an issue.

sjwang90:
@vlastahajek Thanks for your contribution. Could you address the above question and it will help us figure out how to proceed with this PR.

vlastahajek:
@sjwang90, when disabling the counter refresh interval, counters for dynamic types will not work properly. 
After getting back to this issue after two years with a more objective view, it seems to me that this type of fix is not a good idea.
As an alternative to the intended fix,  it will be worth mentioning the workaround in common problems. 
As other changes in the PR are useful, I'm going to close this PR and create another one with logging and doc improvements.

srebhan:
@vlastahajek cool thank you!

PR #4485, opened 2018-07-28
Add support for configuring an AWS endpoint_url

### LINKED ISSUE #4479: AWS-related plugins should support arbitrary endpoint URLs
### Proposal:

Allow the user to specify the exact host address of the AWS service. This allows communicating to certain AWS services via private VPC connections rather than over the open Internet. See [here](https://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/endpoint-service.html) for a general overview.

It also allows connecting to regions not yet supported by the Telegraf plugins, or to mock AWS endpoints (as offered by many tools such as [localstack](https://github.com/localstack/localstack))

### Current behavior:

The user supplies a region identifier, which maps to a predefined API URL.

### Desired behavior:

The user can supply a custom endpoint URL, which, if using AWS PrivateLink, is generated automatically for the user on an account basis by AWS. This URL is randomized and cannot be mapped ahead of time using region identifiers.

### Use case:

Many users want to talk to AWS services over AWS's internal network via VPC endpoints rather than over the Internet. This is not only more secure, it incurs less latency and cost.

danielnelson:
closes #4479 

### Required for all PRs:

### DIFF
--- a/internal/config/aws/credentials.go
+++ b/internal/config/aws/credentials.go
@@ -9,13 +9,14 @@ import (
 )
 
 type CredentialConfig struct {
-	Region    string
-	AccessKey string
-	SecretKey string
-	RoleARN   string
-	Profile   string
-	Filename  string
-	Token     string
+	Region      string
+	AccessKey   string
+	SecretKey   string
+	RoleARN     string
+	Profile     string
+	Filename    string
+	Token       string
+	EndpointURL string
 }
 
 func (c *CredentialConfig) Credentials() client.ConfigProvider {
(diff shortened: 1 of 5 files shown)


danielnelson:
@jcmcken Do you think you could verify this change for me, here is an [amd64 rpm](https://16016-33258973-gh.circle-artifacts.com/0/build/linux/amd64/telegraf-1.8.0%7E8f878c8c-0.x86_64.rpm), but let me know if you need a different build.

jcmcken:
@danielnelson I'll give it a try

jcmcken:
👍 This works

Issue #932, opened 2016-03-26
procstat: include children / sum totals / pss


zarnovican:
It would be nice if `procstat` had some additional features..
- child processes - Sometimes it is not possible to match easily a service together with all descendants. It would be nice if you could specify something like:
  
  ```
  pid_file = "/var/run/nginx.pid"
  include_children = true
  ```
  
  While specifying only the top process, it would include all direct and indirect descendants in the measurements.
- sum totals - Plugin could optionally do the aggregation at the source. Similar to what 'cpu' plugin does.
  
  ```
  perprocess = false
  totalprocess = true
  ```
  
  This makes a lot of sense for services that fork off short-lived children. Currently, telegraf will create a new measurement in Influx for every new child (it contains child's pid). Totals might be hard to implement, as it is not obvious which metrics are "aggregate"able and how.
- measure PSS too - Plugin could include also stats from Linux `smaps`
  
  ```
  include_smaps = true
  ```
  
  Why? For a group of processes it makes more sense to sum PSS values of each process rather than RSS. PSS is available on Linux from `/proc/<pid>/smaps` (requires root access). Sum of PSS is much closer to answering the question "How much memory is this service using..?"

These three features could be useful each independently. Let me know if you would like to split them into three tickets.

Because those features were missing in current `procstat` plugin, I had to implement them on my own. Here is a crappy bash version of CPU/mem usage.
https://gist.github.com/zarnovican/9840d171d9a9c6de1766

```
# /usr/lib/telegraf/scripts/sumproc.sh -u postgres
sumproc utime=6.9,stime=40.58,nprocess=8i,pss=24326144i
```

powersj:
next steps: need to see which of these still need to be done

zarnovican:
I'm closing the ticket. I have switched to monitoring systemd units istead of individual processes.

If any of the subscribers think, that something should be implemented after all, feel free to reopen the ticket.

PR #4508, opened 2018-08-02
inputs/procstat: add pss metric

tagirb:
Add pss (proportional set size) metric to the `procstat` input.


tagirb:
Addresses the pss part of https://github.com/influxdata/telegraf/issues/932

PR #4515, opened 2018-08-03
copy proxy code from http_response httpjson for proxy use

jeffreyhodge:
httpjson is not able to go through a proxy (whether set in the environment or systemd script).  Copied the code in http_response to httpjson.  

### Required for all PRs:

### DIFF
--- a/plugins/inputs/httpjson/httpjson.go
+++ b/plugins/inputs/httpjson/httpjson.go
@@ -2,6 +2,7 @@ package httpjson
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"net/http"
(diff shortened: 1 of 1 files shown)


danielnelson:
We have deprecated the `httpjson` plugin since the `http` input plugin can do everything it does and more, so we aren't accepting changes to this plugin.

You can use a proxy in the `http` input via the environment variables, but we would also be interested in a PR that adds a `http_proxy` option as well.  It might be easier to use this method for setting the ProxyFunc though:

https://github.com/influxdata/telegraf/blob/0759c8b22b922226fbe5cabf19896c4796babea5/plugins/outputs/influxdb/http.go#L150-L155

PR #4521, opened 2018-08-06
Set version to unknown if not set

### LINKED ISSUE #4516: Wrong version reported after building Telegraf
### Steps to reproduce:

```bash
TELEGRAF_VERSION=1.7.2

go get -d github.com/influxdata/telegraf
cd $GOPATH/src/github.com/influxdata/telegraf
git checkout "$TELEGRAF_VERSION"
make deps
GOOS=linux GOARCH=amd64 go build -o telegraf-"$TELEGRAF_VERSION" ./cmd/telegraf
./telegraf-"$TELEGRAF_VERSION" version
Telegraf v1.7.0~unknown (git: unknown unknown)
```

### Expected behavior:

To see `Telegraf v1.7.2~unknown (git: unknown unknown)`

### Actual behavior:

I see `Telegraf v1.7.0~unknown (git: unknown unknown)`

danielnelson:
Prevents setting the branch version to the last major release, and as a bonus we no longer need to set this variable on release:
```
$ go build -ldflags "-X main.version=1.8.0 -X main.commit=fa08656a -X main.branch=default-version" ./cmd/telegraf
$ ./telegraf version
Telegraf 1.8.0 (git: default-version fa08656a)
```
```
$ go build -ldflags " -X main.commit=fa08656a -X main.branch=default-version" ./cmd/telegraf
$ ./telegraf version
Telegraf unknown (git: default-version fa08656a)
```
```
$ go build ./cmd/telegraf
$ ./telegraf version
Telegraf unknown
```

closes #4516 

### Required for all PRs:

### DIFF
--- a/cmd/telegraf/telegraf.go
+++ b/cmd/telegraf/telegraf.go
@@ -58,21 +58,10 @@ var fService = flag.String("service", "",
 var fRunAsConsole = flag.Bool("console", false, "run as console application (windows only)")
 
 var (
-	nextVersion = "1.8.0"
-	version     string
-	commit      string
-	branch      string
+	version string
+	commit  string
+	branch  string
 )
-
-func init() {
-	// If commit or branch are not set, make that clear.
-	if commit == "" {
-		commit = "unknown"
-	}
-	if branch == "" {
-		branch = "unknown"
-	}
-}
 
 var stop chan struct{}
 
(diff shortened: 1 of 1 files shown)


danielnelson:
> Wouldn't it be better to update the variable nextVersion to match the release tag when a release is made?

I think this is a valid argument but I would prefer to have one less step when it comes to preparing a release.  My philosophy here is that if it isn't a release then it doesn't have a version.  I highly recommend that you use the Makefile to ensure this is set automatically when building a tag.

kaspergrubbe:
Fair enough, thanks for pointing out the Makefile, I missed the versioning that it does 👍 

I noticed that the build-task in the Makefile does not include `-w -s` as you recommended in https://github.com/influxdata/telegraf/issues/4516 is that on purpose?

Thanks.

danielnelson:
We use the ldflags in the official builds because they shrink the size of the binary, but not during development.  You probably want these flags if you are not planning to use a debugger, although I'm not sure exactly how much the functionality is limited by stripping symbols.

The official releases are made using `scripts/build.py`, at some point I would like this script to call the Makefile to create the binaries but it's not a big priority.

PR #4527, opened 2018-08-07
plugins/inputs/procstat: add memory_pss on Linux

tagirb:
### Background
PSS (proportional set size) is a great metric for measuring memory consumption of multi-process applications. Whereas RSS includes the total amount of memory used by shared libraries, PSS will only show a proportional amount divided by all processes using this library.
Thus, RSS of a multi-process application is the sum of the PSS metrics of all its processes.

### MR Content
- refactor procstat to support OS-specific metrics
- add pss as a Linux-specific metric to procstat
- update `procstat/README.md` correspondingly
- bump `gopsutil` version in `Gopkg.toml`

### Required for all PRs:

### DIFF
--- a/plugins/inputs/procstat/process_linux.go
+++ b/plugins/inputs/procstat/process_linux.go
@@ -1,0 +1,25 @@
+package procstat
+
+import (
+	"time"
+
+	"github.com/shirou/gopsutil/cpu"
+	"github.com/shirou/gopsutil/process"
+)
+
+type Process interface {
+	PID() PID
+	Tags() map[string]string
+
+	IOCounters() (*process.IOCountersStat, error)
+	MemoryInfo() (*process.MemoryInfoStat, error)
+	MemoryMaps(bool) (*[]process.MemoryMapsStat, error)
+	Name() (string, error)
+	NumCtxSwitches() (*process.NumCtxSwitchesStat, error)
+	NumFDs() (int32, error)
+	NumThreads() (int32, error)
+	Percent(interval time.Duration) (float64, error)
+	Times() (*cpu.TimesStat, error)
+	RlimitUsage(bool) ([]process.RlimitStat, error)
+	Username() (string, error)
+}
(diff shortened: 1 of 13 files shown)


tagirb:
Closes https://github.com/influxdata/telegraf/issues/932. Aggregation is easily done with basicstats.

glinton:
When Daniel approves we should be ok to go ahead and merge

tagirb:
@danielnelson Sorry to disturb. This PR's been hanging around for 2 months already. Is there anything, where I can help to get it merged? It is a simple, but really valuable thing for monitoring multi-process apps, like php-fpm, for instance.

plugins/inputs/procstat/process_linux.go
```diff
@@ -0,0 +1,25 @@
...
+	"github.com/shirou/gopsutil/process"
+)
+
+type Process interface {
+	PID() PID
+	Tags() map[string]string
+
+	IOCounters() (*process.IOCountersStat, error)
+	MemoryInfo() (*process.MemoryInfoStat, error)
+	MemoryMaps(bool) (*[]process.MemoryMapsStat, error)
```
danielnelson: Instead of one interface per platform, there should only be one interface and then platform specific implementations of the function.

You can add the new function to `type Proc struct`, but instead of implementing for all known platforms, do one implementation for Linux and a do nothing implementation for not linux.

Finally, because the MemoryMapsStat struct varies so much in gopsutil, I would probably make the function more specific.  On non-linux platforms you could define and return a error like `ErrNotImplementedError`:
```
MemoryPSS() (uint64, error)
```

An alternative to above would be wrapping the []process.MemoryMapsStat in an separate type for linux/non-linux with functions to retrieve the various fields.

tagirb:
Thanks for the feedback! I now believe measuring PSS for a group of processes is a wrong way to go. Cgroups is a way better solution.

PR #4528, opened 2018-08-07
Rename processor

goldibex:
### Required for all PRs:

This PR adds a processor plugin called "rename", which simply renames metrics, tags, or fields according to its configuration. It has unit coverage and the documentation has been updated with instructions on how to use it.

We need this functionality because both the `statsd` input and `cloudwatch` output have hardcoded mutually inconsistent metric field names that we needed a simple way to harmonize.

### DIFF
--- a/plugins/processors/rename/rename.go
+++ b/plugins/processors/rename/rename.go
@@ -1,0 +1,101 @@
+package rename
+
+import (
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/plugins/processors"
+	"sync"
+)
+
+const sampleConfig = `
+  ## Measurement, tag, and field renamings are stored in separate sub-tables.
+  ## Specify one sub-table per rename operation.
+  # [[processors.rename.measurement]]
+  #   ## measurement to change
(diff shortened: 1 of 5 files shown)


plugins/processors/rename/rename.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+func (r *Rename) Apply(in ...telegraf.Metric) []telegraf.Metric {
+	r.once.Do(r.init)
+
+	for _, point := range in {
+		if newMeasurementName, ok := r.measurements[point.Name()]; ok {
+			point.SetName(newMeasurementName)
+		}
+		for oldTagName, tagValue := range point.Tags() {
```
danielnelson: We would want to use TagList() here instead of Tags() because it doesn't allocate, but what might be even better is if we skipped the initialization completely and just directly used the `renamer` types:
```
for _, rule := range r.Tag {
	value, ok := point.GetTag(rule.From)
		if ok {
			point.AddTag(rule.To, value)
		}
}
```

goldibex: will do!

PR #4532, opened 2018-08-08
New plugin "systemd_units" for linux

benschweizer:
### Required for all PRs:


plugins/inputs/services/services_windows.go
```diff
@@ -0,0 +1,93 @@
...
+// SampleConfig returns sample configuration options.
+func (services *Services) SampleConfig() string {
+	return `
+  ## no options
+`
+}
+
+// Gather parses wmi outputs and adds counters to the Accumulator
+func (services *Services) Gather(acc telegraf.Accumulator) error {
+	//var dst []Win_32Service
```
glinton: Remove commented code

danielnelson:
One big picture question on this pull request is how should we handle this change with respect to the win_services plugin.

We could either go for a unified plugin for all init systems or we could have specialized plugins for each init system.

If we decided to do a unified plugin then I think we need to provide greater unification between backends.  For example, all the backends should use the same set of state codes and the format of the state should also be shared.

If we go with specialized plugins then I think the consistency requirement can be less rigid.

plugins/inputs/services/services_linux.go
```diff
@@ -0,0 +1,110 @@
...
+  ## The default timeout of 1s for systemctl execution can be overridden here:
+  # timeout = "1s"
+`
+}
+
+// Gather parses systemctl outputs and adds counters to the Accumulator
+func (services *Services) Gather(acc telegraf.Accumulator) error {
+	out, e := services.systemctl(services.Timeout)
+	if e != nil {
+		acc.AddError(e)
```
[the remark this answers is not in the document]
cxcv: ok

plugins/inputs/services/readme.md
```diff
@@ -0,0 +1,51 @@
...
+> services,host=host2.example.com,name=CryptSvc state="Running",status=0i 1533731001000000000
+> services,host=host2.example.com,name=dbupdate state="Stopped",status=2i 1533731001000000000
+...
+```
+
+# Statuses
+
+| Value | Meaning  | Description                             |
+| ----- | -------  | -----------                             |
+| 0     | Ok       | Service state is withour failure        |
```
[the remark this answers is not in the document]
cxcv: ok

plugins/inputs/services/services_windows.go
cxcv: ok

cxcv:
@danielnelson @vlastahajek
Good point, I've missed the win_services plugin, and the intention is the same.
I think the code is similar, but as you stated win_services has a more complete list of states.
Generally spoken, I think a unified plugin would be better, as you can use the same alert rules.
May be it would be good to i) merge the code and ii) keep the defined states.
What do you think?

plugins/inputs/services/services_linux.go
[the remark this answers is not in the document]
glinton: Is there a reason you're not returning the error here `return e`?

cxcv:
@danielnelson @vlastahajek
Meanwhile I had a look into win_services.
The first difference is that win_services uses the go-native service manager while this plugin uses 3rd party wmi. The field names are different with both APIs. Next, win_services keeps more raw-data where this services plugin reduces the output to a more simple ok/warning/critical/unknown scheme. Overall, the win_services module looks more extensive and might be a fit for more use cases.
From there, renaming this module to systemd_units and dropping windows support is an option. Any comments?

vlastahajek:
Win_services plugin was created to answer multiple requests for it and it was validated by users. As you pointed, your implementation is limited in comparing to win_services, but on the other hand it adds status field, which in not available via service manager

If functionalities are unified, it makes sense for me to have plugin that provides same interface to system services on all systems, on the other hand, users of win_services would need to switch, to new plugin, but if done properly , it may not  be a problem.

@danielnelson, WDYT?

danielnelson:
I think what makes the most sense initially is to convert this to a systemd service only plugin.

In the future we could take a stab at making a generalized service input, which would be restricted to a subset of metrics.  Having the specialized plugins will help as we will be able to take a broad look at what data is available on all platforms.  Even with a multiplatform service input it may be useful to have ones specific to the particular init system, to expose more advanced metrics.

cxcv:
I agree that having specific plugins for each system and a generic overall plugin would be nice.
For a pure systemd plugin, I would rename tags/fields to be systemd specific and drop the computed status. Though, what would you do differently for a generic overall services plugin? May be we are already pretty close.

danielnelson:
It seems like the common are the service name and a general service state (stopped, running):
```
services,host=localhost,origin=localhost,name=telegraf,state=running state_code=4i
```

If we could build off of this so that systemd is a superset of these values we could extend the values:
```
services,host=localhost,origin=localhost,name=telegraf.service,state=running,status=warning state_code=4i,status_code=1i
```

We would need to make some larger changes to win_services if we wanted to fit it in.

cxcv:
Hello,

I've set the PR to WIP, I plan to update it in the next days.
As there's a working plugin for windows, I would focus this one on systemd.
A great unified plugin would be nice, may be in a 2nd step afterwards.
Would this be ok for everyone involved?

Cheers
Benjamin

amarao:
Hello!

It's really cool plugin, and I would be happy to see it in the telegraf, but it currently lacks one important metric: `NRestarts` (which says how many times service was restarted due to failure). Could you please add it?

P.S. Example of InfluxDB failed few times due to lack of memory:

```
systemctl show influxdb -p NRestarts
NRestarts=9
systemctl restart influxdb
systemctl show influxdb -p NRestarts
NRestarts=0
```
(as you can see it's been cleared on manual restart).

cxcv:
@danielnelson I've updated the plugin:
- removed redundant windows services checks as they are meanwhile covered by win_services
- added support for systemd unit types beyond services
- returns all fields as reported by systemctl
If there are no further comments, you could merge this.

cxcv:
> It's really cool plugin, and I would be happy to see it in the telegraf, but it currently lacks one important metric: `NRestarts` (which says how many times service was restarted due to failure). Could you please add it?

Nice idea, but we'd need to query each unit. Is it possible to have this in one short, e.g. as an option to list-units ?

cxcv:
@danielnelson 
I've updated the plugin:
- removed redundant code that is meanwhile covered in the win_services plugin
- added support of unit types beyond .service
- returns all available fields (load, active, sub)

If there are no further comments, it's done and it could be merged.
Cheers
Benjamin

plugins/inputs/systemd_units/README.md
```diff
@@ -0,0 +1,136 @@
...
+```
+
+### Metrics
+- systemd_units:
+  - tags:
+    - name (string, unit name)
+  - fields:
+    - loaded (int, see below)
+    - active (int, see below)
+    - running (int, see below)
```
stefanhorning: Shouldn't this list be `load`, `active` and `sub`?

cxcv: you are right thanks for spotting this

Marlinc:
Nice work, this is a great addition to Telegraf. Will try and test this

plugins/inputs/systemd_units/systemd_units_linux.go
```diff
@@ -0,0 +1,227 @@
...
+func (systemd_units *SystemdUnits) Description() string {
+	return "Gather systemd units state"
+}
+
+// SampleConfig returns sample configuration options.
+func (systemd_units *SystemdUnits) SampleConfig() string {
+	return `
+  ## The default timeout of 1s for systemctl execution can be overridden here:
+  # timeout = "1s"
+  # unittype = "service"
```
glinton: Same comments as the readme

plugins/inputs/systemd_units/systemd_units_linux.go
```diff
@@ -0,0 +1,227 @@
...
+
+		data := strings.Fields(line)
+		if len(data) < 4 {
+			acc.AddError(fmt.Errorf("Error parsing line (expected at least 4 fields): %s", line))
+			continue
+		}
+		tags := map[string]string{
+			"name": data[0],
+		}
+		// unit_load_state_table, see
```
glinton: Make these state table consts and pull them out of gather.

cxcv: I've moved them, but maps cannot be const!?

plugins/inputs/systemd_units/README.md
```diff
@@ -10,13 +10,15 @@ This plugin is related to the [win_services module](../win_services/), which
...
 
 In addition to services, this plugin can gather other unit types as well,
-see ```systemctl list-units --all --type help``` for possible options.
+see `systemctl list-units --all --type help` for possible options.
 
 ### Configuration
 ```
 [[inputs.systemd_units]]
-  ## The default timeout of 1s for systemctl execution can be overridden here:
+  ## The default timeout of 1s for systemctl execution can be overridden
```
glinton: Maybe just use `## Set timeout for systemctl execution`

plugins/inputs/systemd_units/README.md
```diff
@@ -10,13 +10,15 @@ This plugin is related to the [win_services module](../win_services/), which
...
+  ## here:
   # timeout = "1s"
+  ## Filter for a specific unit types, default is "service":
```
glinton: Add a newline above this line

plugins/inputs/systemd_units/systemd_units_linux.go
```diff
@@ -24,6 +24,132 @@ type systemctl func(Timeout internal.Duration, UnitType string) (*bytes.Buffer,
 
 const measurement = "systemd_units"
 
+// unit_load_state_table, see
+// https://github.com/systemd/systemd/blob/c87700a1335f489be31cd3549927da68b5638819/src/basic/unit-def.c#L87
+var load_map = map[string]int{
+	// err: 0,
```
glinton: Let's remove all the commented out values, if they aren't being used, it doesn't seem valuable to have lingering commented code lying around.

plugins/inputs/systemd_units/README.md
glinton: Same for `SampleConfig`

PsyanticY:
Hi,  any updates on this

PR #4537, opened 2018-08-09
Add Message Max bytes configuration

otherpirate:
### Required for all PRs:

### DIFF
--- a/plugins/outputs/kafka/kafka.go
+++ b/plugins/outputs/kafka/kafka.go
@@ -37,6 +37,8 @@ type (
 		RequiredAcks int
 		// MaxRetry Tag
 		MaxRetry int
+		// Max Message Bytes
+		MaxMessageBytes int
 
 		Version string `toml:"version"`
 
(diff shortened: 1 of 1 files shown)


danielnelson:
What happens if the message is over the max, does `err := k.producer.SendMessages(msgs)` return an error?

otherpirate:
When message is bigger than MessageMaxBytes config this error occurs:
`2018-08-09T15:47:24Z E! Error writing to output [kafka]: kafka: Failed to produce message to topic <topic>: kafka server: Message was too large, server rejected it to avoid allocation error.`

It's the same error with default value(without configuration)

danielnelson:
It seems like short term we ought to drop these messages, after logging, because otherwise the output plugin will just be stuck trying to send with no potential to send successfully.  I don't think this needs to be done in the PR though.

Long term, I would like to add a max size limitation to the Serializer interface.  This way we could attempt to split the messages, similar to how we can currently limit single lines in the influx serializer.  We would have to consider what this means for each serializer and ultimately it still may impossible to write some metrics in the allotted space.

danielnelson:
cherry-picked to master 61513c64b85b2cb325a49b930f5ae6ff44af4091

PR #4541, opened 2018-08-09
Dockerstate

johnseekins:
### Required for all PRs:

This plugin collects very _basic_ stats about each docker container running (and stopped) on a system. It is far less verbose than the current `docker` input plugin. If you are collecting docker stats for 100s of hosts, this might be more useful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,6 +151,7 @@ configuration options.
 * [dmcache](./plugins/inputs/dmcache)
 * [dns query time](./plugins/inputs/dns_query)
 * [docker](./plugins/inputs/docker)
+* [dockerstate](./plugins/inputs/dockerstate)
 * [dovecot](./plugins/inputs/dovecot)
 * [elasticsearch](./plugins/inputs/elasticsearch)
 * [exec](./plugins/inputs/exec) (generic executable plugin, support JSON, influx, graphite and nagios)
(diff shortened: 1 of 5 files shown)


danielnelson:
@johnseekins Thanks for the contribution, but I don't think we would be willing to add a secondary plugin for docker.  I would prefer to just make it possible to configure the docker input to do less for those with large container infrastructures.

How large is the performance difference if you take the existing plugin and configure it like:
```
# Read metrics about docker containers
[[inputs.docker]]
  gather_services = false
  container_name_include = []
  container_name_exclude = ["*"]
  timeout = "5s"
  perdevice = false
  total = false
  docker_label_include = []
  docker_label_exclude = ["*"]
```

You can enable the internal plugin and check the `internal_gather gather_time_ns` metric to compare the performance.

johnseekins:
What we ran in to wasn't the speed of telegraf, but the amount of data in Influx became a problem. Switching to these smaller stats saved us a good 90% in our InfluxDB instance.
If the settings you show would give similar stats (_just_ state without disk/memory/etc.) for each container, that sounds great. I'll try it out.

danielnelson:
Try it out and let us know, also remember that you can use [measurement filtering](https://github.com/influxdata/telegraf/blob/master/docs/CONFIGURATION.md#measurement-filtering) on any input to remove fields and tags that you are not interested in.  The only thing to be careful about when removing tags is not to remove so much that the source of the data is no long unique, because then you might run into things overwriting each other.

johnseekins:
Looking through the stats that `docker` provides without cpu/mem/disk, this plugin _actually_ produces different results.
We tell the state of _each_ container and the version of that container. This means you can track deployments using this plugin where the current docker plugin doesn't allow for that.
Perhaps modifying the other docker plugin would be better?

johnseekins:
That looks similar enough. We'll pursue this separately. Thanks!

PR #4551, opened 2018-08-13
Parser processor

### LINKED ISSUE #4427: Add logfmt Parser Plugin
## Feature Request

### Proposal:
add a parser for `logfmt`

### Current behavior:
None

### Desired behavior:
the ability to parse logfmt log messages and specify the keys for timestamp (and format), tags, and fields

this would be useful to ingesting influxdb log data

Ayrdrie:
### Required for all PRs:
closes #4427 


plugins/processors/parser/parser.go
```diff
@@ -0,0 +1,73 @@
...
+type Parser struct {
+	config      parsers.Config
+	parseFields []string `toml:"parse_fields"`
+	Parser      parsers.Parser
+}
+
+// holds a default sample config
+var SampleConfig = `
+
+## specify the name of the field[s] whose value will be parsed
```
glinton: Double space the sample config and remove the empty line on line 20.

plugins/processors/parser/parser.go
```diff
@@ -0,0 +1,73 @@
...
+		p.Parser, err = parsers.NewParser(&p.config)
+		if err != nil {
+			log.Printf("E! [processors.parser] could not create parser: %v", err)
+			return metrics
+		}
+	}
+
+	for _, metric := range metrics {
+		for _, key := range p.parseFields {
+			value := metric.Fields()[key]
```
glinton: It might not matter, but we could probably do `if value, ok := metric.Fields()[key]; ok {` here, to ensure there is a field value for the key.

plugins/processors/parser/parser.go
```diff
@@ -0,0 +1,129 @@
...
+	"fmt"
+	"log"
+
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/plugins/parsers"
+	"github.com/influxdata/telegraf/plugins/processors"
+)
+
+type Parser struct {
+	Config      parsers.Config `toml:"config"`
```
danielnelson: Use an embedded field for the parsers.Config:
```

type Parser struct {
	parsers.Config `toml:"config"`
```

This should make it so you can specify the options directly on the Parser object, similar to how it is set in input plugins:
```toml
[[processors.parser]]
  parse_fields = ["message"]
  original = "merge"
  data_format = "json"
  tag_keys = ["verb", "request"]
```

plugins/processors/parser/parser.go
```diff
@@ -0,0 +1,129 @@
...
+
+	name := ""
+
+	for _, metric := range metrics {
+		if n := metric.Name(); n != "" {
+			name = n
+		}
+		sMetrics := []telegraf.Metric{}
+		for _, key := range p.ParseFields {
+			if value, ok := metric.Fields()[key]; ok {
```
danielnelson: Don't use `.Fields()` since it allocates a new map, (I really need to add a comment suggesting that it not be used).  Instead use `.FieldList()`.  You will probably want to reorder the loops since it think ParseFields should always be shorter than Fields.

plugins/processors/parser/parser.go
```diff
@@ -0,0 +1,129 @@
...
+				strVal := fmt.Sprintf("%v", value)
```
danielnelson: If it's not a string then log a warning and skip.

plugins/processors/parser/parser.go
```diff
@@ -0,0 +1,129 @@
...
+				nMetrics, err := p.parseField(strVal)
+				if err != nil {
+					log.Printf("E! [processors.parser] could not parse field %v: %v", key, err)
+					switch p.Original {
+					case "keep":
+						return metrics
```
danielnelson: This doesn't seem right... don't we need to parse all fields before we can return?

plugins/processors/parser/parser.go
```diff
@@ -0,0 +1,129 @@
...
+					case "merge":
+						nMetrics = metrics
```
danielnelson: I'm all confused by the variable names: rMetric, sMetric, nMetric.  Can you try to improve these?   The Metric part is less interesting than the r, s, n parts, since we can easily lookup the type.

plugins/processors/parser/parser.go
```diff
@@ -0,0 +1,129 @@
...
+					}
+				}
+				sMetrics = append(sMetrics, nMetrics...)
+			} else {
+				fmt.Println("key not found", key)
```
danielnelson: I would probably ignore this situation to allow fields to be optional.

plugins/processors/parser/parser.go
```diff
@@ -0,0 +1,129 @@
...
+	if p.Original == "merge" {
+		rMetrics = p.mergeTagsFields(rMetrics...)
+	}
+
+	return p.setName(name, rMetrics...)
+
+}
+
+func (p Parser) setName(name string, metrics ...telegraf.Metric) []telegraf.Metric {
+	if len(metrics) == 0 {
```
danielnelson: Don't need this, if you remove it then it will just range over nothing and return metrics.

plugins/processors/parser/parser.go
```diff
@@ -0,0 +1,129 @@
...
+
+	return metrics
+}
+
+func (p Parser) mergeTagsFields(metrics ...telegraf.Metric) []telegraf.Metric {
+	if len(metrics) == 0 {
+		return nil
+	}
+
+	rMetric := metrics[0]
```
danielnelson: This seems like a problem if more than one metric is passed in to Apply.  Secretly this never happens but the interface doesn't promise this.

glinton: More than one metric does get passed into this function, it is intended for all the tags and fields to get merged down to the first metric

glinton: More than one metric does get passed into this function, it is intended for all the tags and fields to get merged down to the first metric

plugins/processors/parser/parser.go
glinton: Good catch, the tests missed this.

Issue #4576, opened 2018-08-21
Gather Kibana health metrics


yuriy-slobodchikov:
## Feature Request
 Kibana health check plugin
( @danielnelson )

Opening a feature request kicks off a discussion.

### Proposal:
Adding  Kibana health check plugin

### Current behavior:
No Kibana health check via localhost:5601/api/status
https://www.elastic.co/guide/en/kibana/6.3/access.html#status

### Desired behavior:
To have ability to get metrics and plugin health checks from json localhost:5601/api/status
For "state" also should be created additional state_code    (sometimes it is more convenient to have numeric thresholds in grafana)
transformation from 
"green" - 0
"yellow" - 1
"red" - 2
any other - 3

{
  "name": "my-kibana",
  "uuid": "e279c9b7-2125-436b-822e-75e3c179a79f",
  "version": {
    "number": "6.3.0",
    "build_hash": "c360fcacca5594000a2f7c80fe2dcaad5c7ee52c",
    "build_number": 17230,
    "build_snapshot": false
  },
  "status": {
    "overall": {
      "state": "green",
      "title": "Green",
      "nickname": "Looking good",
      "icon": "success",
      "since": "2018-06-28T13:16:31.537Z"
    },
    "statuses": [
      {
        "id": "plugin:kibana@6.3.0",
        "state": "green",
        "icon": "success",
        "message": "Ready",
        "since": "2018-06-28T13:16:31.537Z"
      },
     ...
    ]
  },
  "metrics": {
    "last_updated": "2018-08-21T10:41:51.024Z",
    "collection_interval_in_millis": 5000,
    "uptime_in_millis": 4656520644,
    "process": {
      "mem": {
        "heap_max_in_bytes": 208896000,
        "heap_used_in_bytes": 161749976
      }
    },
    "os": {
      "cpu": {
        "load_average": {
          "1m": 0.0029296875,
          "5m": 0.0146484375,
          "15m": 0.04541015625
        }
      },
      "cgroup": {
        "cpuacct": {
          "control_group": "/system.slice/kibana.service",
          "usage_nanos": 14313207367574
        },
        "cpu": {
          "control_group": "/system.slice/kibana.service",
          "cfs_period_micros": 100000,
          "cfs_quota_micros": -1,
          "stat": {
            "number_of_elapsed_periods": 0,
            "number_of_times_throttled": 0,
            "time_throttled_nanos": 0
          }
        }
      }
    },
    "response_times": {
[...]

PR #4585, opened 2018-08-22
Add Kibana input plugin 

lpic10:
Implements #4576

I tried to keep it simple and only collecting the metrics related to Kibana itself.

### Required for all PRs:


lpic10:
@danielnelson thanks for the review, it should be better now. 
I also added a new field  `requests_per_sec` and fixed some issues with types. 

> any idea if we need to be concerned about elasticsearch version differences?

Yes, indeed this plugin will work on Kibana 6.x (tested from 6.0 to 6.4), but not on previous versions as the status API returns a different structure. For that I added a note in the doc, do you think that's enough?

danielnelson:
That should be good, thanks again!

PR #4596, opened 2018-08-27
updated version of telegraf(1.7+) that supports OpenHardwareMonitor!

Suron12:
Please precompile an updated version of telegraf(1.7+) that supports OpenHardwareMonitor! I only run windows, and DEP doesn't have windows installers. I also have never had to use GO before. 
Thanks,

### DIFF
--- a/README.md
+++ b/README.md
@@ -270,6 +270,7 @@ For documentation on the latest development code see the [documentation index][d
 * [openweathermap](./plugins/inputs/openweathermap)
 * [pf](./plugins/inputs/pf)
 * [pgbouncer](./plugins/inputs/pgbouncer)
+* [open_hardware_monitor](./plugins/inputs/open_hardware_monitor)
 * [phpfpm](./plugins/inputs/phpfpm)
 * [phusion passenger](./plugins/inputs/passenger)
 * [ping](./plugins/inputs/ping)
--- a/plugins/inputs/all/all.go
+++ b/plugins/inputs/all/all.go
@@ -116,6 +116,7 @@ import (
 	_ "github.com/influxdata/telegraf/plugins/inputs/nstat"
 	_ "github.com/influxdata/telegraf/plugins/inputs/ntpq"
 	_ "github.com/influxdata/telegraf/plugins/inputs/nvidia_smi"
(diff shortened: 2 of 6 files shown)


danielnelson:
Try either the [curl installation](https://github.com/golang/dep#installation) or you can build from source, here is how I do it and this should place dep in your $GOPATH/bin directory:
```
go get -d github.com/golang/dep
cd src/github.com/golang/dep
git checkout -q v0.5.0
go install -ldflags="-X main.version=v0.5.0" ./cmd/dep
```

iamsilvio:
@marianob85 are you able to create a updated pull request? I would love to use your plugin but I have to rely on the influxdata repo

aumann:
Does this have a chance of being merged after the remaining issues are solved?

Compilation with this plugin seems to work without problems for us, even after rebasing locally.

We found this plugin to be really useful, as Windows is slow in adding support for newer sensors like temperature sensors or fan speed sensors. OpenHardwareMonitor has drivers for those sensors and is a great addition to the other plugins for newer hardware.

If we get a pointer as to what needs to be done to get this merged, we could try and solve the remaining issues.

sjwang90:
Due to the level of interest in the plugin, I think it would be better suited as an [external plugin](https://github.com/influxdata/telegraf/blob/master/docs/EXTERNAL_PLUGINS.md) that can be used with `execd` to run seamlessly with Telegraf. @marianob85 would you be willing to submit your plugin as an external plugin?

marianob85:
> Due to the level of interest in the plugin, I think it would be better suited as an [external plugin](https://github.com/influxdata/telegraf/blob/master/docs/EXTERNAL_PLUGINS.md) that can be used with `execd` to run seamlessly with Telegraf. @marianob85 would you be willing to submit your plugin as an external plugin?

Hi, I was made this plugin long time ago, but yes, I will try to rewrite it as an external plugin for telegraf.

marianob85:
https://github.com/marianob85/open_hardware_monitor-telegraf-plugin

sjwang90:
Added to list of [external plugins](https://github.com/influxdata/telegraf/blob/master/EXTERNAL_PLUGINS.md)

PR #4476, opened 2018-07-26
re-worked lowercase into a generic strings processor

bsmaldon:
I'm re-opening this as a new pull request because I've rebased to clean up the history and don't see a way to update the original pull request. This is the implementation of the strings processor we've been talking about in pull request #3890.

### Required for all PRs:

- [x] Signed [CLA](https://influxdata.com/community/cla/).
- [x] Associated README.md updated.
- [x] Has appropriate unit tests.

PR #4599, opened 2018-08-27
Char replacer processor

doublerainbo:
### Required for all PRs:

This adds in a new processor plugin that runs a search and replace on all the metric names. Our current use case is that we're scraping off a bunch of data from prometheus and the convention there is to use a `:` as a the separator, but when we send it to an output we would want to use `_` as the separator. The existing plugins match on exact naming match which is what prompted this change

### DIFF
--- a/plugins/processors/replace/README.md
+++ b/plugins/processors/replace/README.md
@@ -1,0 +1,29 @@
+# Replace Processor Plugin
+
+The `replace` processor runs a `strings.replace` on metric names and replaces
+specified chars with replacement chars. Multiple replacement configs can be
+stacked on top of each other and executed sequentially. This is useful when
+an input module uses a different set of metric separator than the output module,
+or if there are parts of the metric name that are not useful.
+
+Note that if the char deletion aspect of this is used, and the entire name is
+deleted, it will not perform the operation and the original name will be left
+as is.
+
+### Configuration:
+
+The following example shows a replacement function where it first replace all
+`_` with `-`, then we replace all `:` with `_`. For example if an incoming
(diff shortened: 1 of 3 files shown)


danielnelson:
This seems handy, but it is also quite specialized.  I wonder if we should combine it with #4476?

doublerainbo:
Open to putting it where it makes sense, please point me to the direction you'd like :)

danielnelson:
I tried to quickly see if I could fit this into the strings processor, but had a few issues modeling the config file.  The difficulty is around field vs tag vs measurement as well as with key vs value.  Additionally, the current and newly added processors all have a different feel and I would like to attempt to find a way where we can provide a bit more consistency.

Give me a few days and I'll come up with a proposal for comments.

danielnelson:
I do think we should merge this into strings, the final plugin could look something like this:
```toml
[[processors.strings]]
  [[processors.strings.replace]]
    measurement = "*"
    old = ":"
    new = "_"
```

The strings processor #4476 is very close to being finished, I think we can have it merged within a week at the most.

My full brain dump on processor configuration is in #4616

doublerainbo:
sounds like a plan, as I haven't worked on the string processor, should i be branching off of that PR and move my code into there or should i stay out of the way and wait until it is done, and see what course of action to take from there. thanks!

danielnelson:
If you are comfortable with it, you could branch the pr and after it is merged do an interactive rebase to remove the duplicate commits since we do a squash and merge.

If that sounds like a lot of fuss to you then I would probably just wait for it to be merged.

danielnelson:
@doublerainbo strings processor is merged into master now

PR #4618, opened 2018-08-31
Update README.md

crazy-canux:
If check interval is less than group by time interval, mean is better than last.
If they are equal, it's the same.

### Required for all PRs:


danielnelson:
I believe the idea behind using `last` is that it is a faster aggregation to calculate.  Another potential perk is that it doesn't change the type of the value from int to float.  I think we should leave this as is.

PR #4628, opened 2018-09-04
add queue_durability property to amqp_consumer input

aepetelin:
### Required for all PRs:

- [+] Signed [CLA](https://influxdata.com/community/cla/).
- [+] Associated README.md updated.
- [-] Has appropriate unit tests.

Added a Queue Durability property to amqp_consumer input plugin.

The amqp_consumer input plugin always creates at RabbitMQ server a queue with durable 'true',
that meas RabbitMQ is saving messages to disk in the queue util they are consumed. Sometimes it is needed to avoid impact of disk I/O and set the queue 'not durable'. It is easy to do in plugin configuration file.

Please review the code if it can be merged.

### DIFF
--- a/plugins/inputs/amqp_consumer/amqp_consumer.go
+++ b/plugins/inputs/amqp_consumer/amqp_consumer.go
@@ -31,7 +31,7 @@ type AMQPConsumer struct {
 
 	// Queue Name
 	Queue           string `toml:"queue"`
-	QueueDurability bool   `toml:"queue_durability"`
+	QueueDurability string `toml:"queue_durability"`
 	// Binding Key
 	BindingKey string `toml:"binding_key"`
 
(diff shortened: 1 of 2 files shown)


plugins/inputs/amqp_consumer/amqp_consumer.go
```diff
@@ -98,10 +99,13 @@ func (a *AMQPConsumer) SampleConfig() string {
   # exchange_arguments = { }
   # exchange_arguments = {"hash_propery" = "timestamp"}
 
-  ## AMQP queue name
+  ## AMQP queue name.
   queue = "telegraf"
 
-  ## Binding Key
+  ## AMQP queue durability.
+  queue_durability = true
```
danielnelson: Can you make this a string that can be either "durable" or "transient", defaulting to "durable".  This will just match up with the exchange_durability option a bit better.

aepetelin: @danielnelson Good idea, just fixed that

Issue #4607, opened 2018-08-29
JSON parser, UNIX time format for timestamp field


lazarberbakov:
Currently, the JSON parser allows parsing of time stamp and configuration of time format by using e.g.
json_time_format = "02 Jan 06 15:04 MST".

### Proposal:
It would be nice if epoch/unix timestamp could also be parsed by specifying whether the timestamp is
given in seconds, milliseconds, microseconds or nanoseconds.

PR #4633, opened 2018-09-04
Added Unix epoch timestamp support for JSON parser

drenizg:
closes: #4607

..., corresponding test and doc updated.

If parsing a Unix epoch timestamp, e.g. 1536092344000, "json_time_format" config must be set to "Unix"; corresponding JSON value must be in milliseconds and unquoted.

### Required for all PRs:

### DIFF
--- a/plugins/parsers/json/parser.go
+++ b/plugins/parsers/json/parser.go
@@ -13,6 +13,7 @@ import (
 	"github.com/influxdata/telegraf/metric"
 	"github.com/tidwall/gjson"
 	"math"
+	"regexp"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


drenizg:
@danielnelson, thanks for the feedback! I just commited the changes.

plugins/parsers/json/parser.go
```diff
@@ -78,14 +79,31 @@ func (p *JSONParser) parseObject(metrics []telegraf.Metric, jsonOut map[string]i
...
+		// unix_ms: epoch is assumed to be in milliseconds and unquoted. Cannot have a decimal part.
+		if strings.EqualFold(p.JSONTimeFormat, "unix") || strings.EqualFold(p.JSONTimeFormat, "unix_ms") {
+			timeEpoch, ok := f.Fields[p.JSONTimeKey].(float64) //Using float64 as that is what the JSONFlattener returns
+			if !ok {
+				err := fmt.Errorf("time: %v could not be converted to float64", f.Fields[p.JSONTimeKey])
+				return nil, err
+			}
+
+			if strings.EqualFold(p.JSONTimeFormat, "unix") {
+				timeInt, timeFractional := math.Modf(timeEpoch)
```
danielnelson: I've been looking at this method and I'm concerned that it will lose precision for some timestamps.  Here is an example testcase I made:
```
func TestModf(t *testing.T) {
	epoch := 1536107026.123456789
	integer, fractional := math.Modf(epoch)
	tm := time.Unix(int64(integer), int64(fractional*1e9)).UTC()
	require.Equal(t, int64(1536107026123456789), tm.UnixNano())
}
```
```
--- FAIL: TestModf (0.00s)
	Error Trace:	main_test.go:140
	Error:      	Not equal: 
	            	expected: 1536107026123456789
	            	actual  : 1536107026123456716
	Test:       	TestModf
```

That said, this is an issue with the json flattener and while we could delay conversion to float, I don't think it is required now.

plugins/parsers/json/parser.go
```diff
@@ -78,14 +79,31 @@ func (p *JSONParser) parseObject(metrics []telegraf.Metric, jsonOut map[string]i
...
-			err := fmt.Errorf("time: %v could not be converted to string", f.Fields[p.JSONTimeKey])
-			return nil, err
-		}
-		nTime, err = time.Parse(p.JSONTimeFormat, timeStr)
-		if err != nil {
-			return nil, err
+		// unix: epoch is assumed to be in seconds and unquoted. Can have a decimal part.
+		// unix_ms: epoch is assumed to be in milliseconds and unquoted. Cannot have a decimal part.
+		if strings.EqualFold(p.JSONTimeFormat, "unix") || strings.EqualFold(p.JSONTimeFormat, "unix_ms") {
+			timeEpoch, ok := f.Fields[p.JSONTimeKey].(float64) //Using float64 as that is what the JSONFlattener returns
```
danielnelson: I think we should support taking this value from a float or a string, if we take it from a string we should use string processing to split the parts to avoid precision loss (see my other comment).

drenizg: Ok, working on it :thumbsup:  I'll try to support both dot and comma as decimal separators (just in case).

drenizg: What I've seen is that when the JSON field is unquoted, the `JSONFlattener` returns a `float64`, so there could be already an unavoidable precision loss, specially if the value has more than a few decimal positions. This can be avoided using string processing as you suggested, so I think we should clarify in the docs that if precision accuracy is required, the value must be sent as string. On the other hand, if that is not required or if not using decimals at all (like in `unix_ms` option), the JSON value can come quoted or unquoted. What do you think?

danielnelson: A note in the docs about potential precision loss would be a nice touch.  Later on, we can explore updating the JSONFlattener code to use `json.Number`.

drenizg: Great! Just commited the changes!

danielnelson:
One last thing, now that we have quite a bit of code for parsing the timestamps I think we should extract it into a new function.

drenizg:
Agreed. Just commited the change. Now there's a new function called `parseUnixTimestamp` with all the parsing logic.

Thidjs:
I'm breaking my head over this. When I use the "unix" time_format, i get the following error;
parsing time "1536303994" as "unix": cannot parse "1536303994" as "unix"
Why can't it parse this timestamp? And more important how to fix ;-)

drenizg:
Hi @Thidjs, are you sure you're using a version with this change in it? If using a nightly build, it could be that it was packaged before the change was merged.

Thidjs:
@drenizg I tested again today with the latest nightly build and it seems to be working now. Thanx!

PR #4634, opened 2018-09-04
Log file related errors in tail and logparser input

### LINKED ISSUE #3296: telegraf logparser fails silently on permissions issues
I spent a bit of time chasing down why I couldn't get the log parser to parse web logs to discover that telegraf didn't have read permissions on the relevant files.

It didn't produce any message on this failure including with debugging on, so this would likely be a feature request to not have it fail silently on errors, especially if you have debug on.

1.4.1-1, Ubuntu 16.04 

### Steps to reproduce:

1.  Create a config file with [[inputs.logparser]] present where telegraf doesn't have read permissions to the listed file(s) it should be parsing.

### Expected behavior:

Some sort of error produced, especially with debugging turned on.

### Actual behavior:
It continues to work as it should for other metrics.

## Feature Request

Do not fail silently on errors. Include errors on STDOUT and/or in the logs, especially when debug is present.

glinton:
Resolves #3296


danielnelson:
I think we want to display an error only if we get a permission denied, and be silent if there are just no files yet.  I think this will require changing the GlobPath type to return errors from Match.

This also makes me wonder if tail supports new files that appear after the plugin is started, can you check into this?

glinton:
Tail doesn't currently support new files that appear after plugin start. I'll update `Match` to return permissions errors

danielnelson:
Uh-oh, we are going to need to add support for new files or undeprecate logparser for 1.8.

glinton:
Merged master then got thinking, maybe we could just log the permission errors right from `globpath.go` and not worry about returning the error like before?

russorat:
closing so that we can regroup on the solution.

PR #4645, opened 2018-09-06
Add influx2.0 output plugin

glinton:


plugins/outputs/influxdb_v2/http.go
```diff
@@ -166,7 +183,33 @@ func (c *httpClient) Write(ctx context.Context, metrics []telegraf.Metric) error
...
+	case http.StatusRequestEntityTooLarge: // 413
+		// LineProtocolLengthError
+		desc = fmt.Sprintf("%s - %s;%d", writeResp, writeResp.Op, writeResp.MaxLength)
+	case http.StatusTooManyRequests, http.StatusServiceUnavailable: // 429, 503
+		retryAfter := resp.Header.Get("Retry-After")
+		retry, err := strconv.Atoi(retryAfter)
+		if err != nil {
+			return fmt.Errorf("Bad value for 'Retry-After': %s", err.Error())
+		}
+		time.Sleep(time.Second * time.Duration(retry))
```
goller: You may want a max sleep just in case the server says you sleep for a long time.

plugins/outputs/influxdb_v2/influxdb.go
```diff
@@ -32,12 +32,15 @@ var sampleConfig = `
...
 
-  ## Organization is the name of the organization you wish to write to.
+  ## Organization is the name of the organization you wish to write to; must exist.
   organization = ""
 
   ## Bucket to the name fo the bucketwrite into; must exist.
   bucket = ""
 
+  ## Precision for the unix timestamps within the body line-protocol.
+  # precision = "ns"
```
goller: I don't think telegraf uses nanoseconds, does it?

plugins/outputs/influxdb_v2/http.go
```diff
@@ -166,7 +183,33 @@ func (c *httpClient) Write(ctx context.Context, metrics []telegraf.Metric) error
...
+		c.Write(ctx, metrics)
+	}
+
+	if xErr := resp.Header.Get("X-Influx-Error"); xErr != "" {
+		desc = fmt.Sprintf("%s - %s", desc, xErr)
```
goller: because you are parsing the JSON error body, I don't think you need X-Influx-Error

plugins/outputs/influxdb_v2/http.go
```diff
@@ -166,7 +183,33 @@ func (c *httpClient) Write(ctx context.Context, metrics []telegraf.Metric) error
...
 
-	desc := resp.Header.Get("X-Influx-Error")
+	writeResp := &genericRespError{}
+	json.NewDecoder(resp.Body).Decode(writeResp)
+	var desc string
+
+	switch resp.StatusCode {
+	case http.StatusBadRequest: // 400
+		// LineProtocolError
+		desc = fmt.Sprintf("%s - %s;%d;%s", writeResp, writeResp.Op, writeResp.Line, writeResp.Err)
```
goller: At this point you probably only care about code, message and line; not op and err

Op/Err is useful as a stack trace

plugins/outputs/influxdb_v2/http.go
```diff
@@ -166,7 +183,33 @@ func (c *httpClient) Write(ctx context.Context, metrics []telegraf.Metric) error
...
+	case http.StatusUnauthorized, http.StatusForbidden: // 401, 403
+		// Error
+		desc = fmt.Sprintf("%s - %s;%s", writeResp, writeResp.Op, writeResp.Err)
```
goller: Probably not op/err here

plugins/outputs/influxdb_v2/http.go
```diff
@@ -166,7 +183,33 @@ func (c *httpClient) Write(ctx context.Context, metrics []telegraf.Metric) error
...
+	case http.StatusRequestEntityTooLarge: // 413
+		// LineProtocolLengthError
+		desc = fmt.Sprintf("%s - %s;%d", writeResp, writeResp.Op, writeResp.MaxLength)
```
goller: Probably not op.

plugins/outputs/influxdb_v2/http.go
```diff
@@ -0,0 +1,285 @@
...
+	Line      int32
+	MaxLength int32
+}
+
+func (g genericRespError) String() string {
+	return fmt.Sprintf("%s: %s", g.Code, g.Message)
+}
+
+func (c *httpClient) Write(ctx context.Context, metrics []telegraf.Metric) error {
+	var err error
```
goller: I don't think you'll need to declare err here.

plugins/outputs/influxdb_v2/influxdb.go
danielnelson: It does, though rounding is enabled by default on non-service inputs.  I think here we should do the same here as we do in the InfluxDB v1 output, and always send in nanosecond without the option to change it.

plugins/outputs/influxdb_v2/http.go
danielnelson: 10s seems reasonable to me as a max sleep.  The big issue with the influxdb output is that, when under load, the next filled batch will almost immediately trigger the next write even after a write timeout.

danielnelson: Oh but don't sleep and write because of #2919, hang on to the time and return.  On the next call to Write check the time and if the amount of time has not passed then return an error.

Let's define a special error type specifically for this, something like CircuitOpenErr?  We can put it in the internal package for now.

plugins/outputs/influxdb_v2/http.go
```diff
@@ -0,0 +1,285 @@
...
+	}
+
+	writeResp := &genericRespError{}
+	json.NewDecoder(resp.Body).Decode(writeResp)
+	var desc string
+
+	switch resp.StatusCode {
+	case http.StatusBadRequest: // 400
+		// LineProtocolError
+		desc = fmt.Sprintf("%s - %s;%d;%s", writeResp, writeResp.Op, writeResp.Line, writeResp.Err)
```
danielnelson: Probably fold the desc lines into the Error() function of genericRespError

plugins/outputs/influxdb_v2/http.go
```diff
@@ -0,0 +1,285 @@
...
+		return err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode == http.StatusNoContent {
+		return nil
+	}
+
+	writeResp := &genericRespError{}
+	json.NewDecoder(resp.Body).Decode(writeResp)
```
danielnelson: Need to check for err

plugins/outputs/influxdb_v2/http.go
```diff
@@ -0,0 +1,285 @@
...
+	var desc string
+
+	switch resp.StatusCode {
```
danielnelson: We need to return and error if we would like to retry, and nil if we want to treat the data as written (and delete it).  I think there are three cases:
- Client error and there is potential for a retry to work; return an error
- Client error and no chance a retry will help: log and return nil
- Server error; return an error

plugins/outputs/influxdb_v2/influxdb.go
glinton: nanoseconds is the default in the platform spec, I'll remove the option to change it

plugins/outputs/influxdb_v2/http.go
glinton: I meant to note to remove this once the spec was fully implemented

plugins/outputs/influxdb_v2/http.go
glinton: Should op/err be removed from the spec for `Error` and `LineProtocolError` if it's not intended to be used anywhere?

plugins/outputs/influxdb_v2/http.go
glinton: Good call, carryover from influxv1 plugin

plugins/outputs/influxdb_v2/http.go
glinton: I only used code/message in there because it's the only shared field across the three different returned error types

plugins/outputs/influxdb_v2/http.go
goller: No it should not be removed the spec. The audience for op/err are people when debugging and doing more thorough analysis of a problem.

plugins/outputs/influxdb_v2/http.go
glinton: ok, it was essentially ignored in the v1 plugin. do we return or what?

danielnelson: Yeah, that sounds right.

plugins/outputs/influxdb_v2/influxdb.go
```diff
@@ -0,0 +1,199 @@
...
+
+func (i *InfluxDB) Connect() error {
+	ctx := context.Background()
+
+	if len(i.URLs) == 0 {
+		i.URLs = append(i.URLs, defaultURL)
+	}
+
+	i.serializer = influx.NewSerializer()
+	i.serializer.SetFieldTypeSupport(influx.UintSupport)
```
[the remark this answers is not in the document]
goller: @danielnelson , yes.  We support uint.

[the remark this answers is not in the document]
danielnelson: Is uint always desired?  If someone imported non-uint data into the database this would be a change of type and could result in data be uninsertable.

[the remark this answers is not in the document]
glinton: Sounds like an issue for db migrations, right? This flag only allows the plugin to add uints if they are configured/used, if i understand correctly.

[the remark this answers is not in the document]
glinton: Plus, with a major version change there are bound to have some incompatibilities. Keeping this always set seems like the best thing since the database supports uints from the getgo.

plugins/outputs/influxdb_v2/http.go
glinton: Does the failed to write metric get cached, timestamp included? If the next call to write is less than the initial retry-time, would that second metric also get cached and be written on the next attempt that is after the retry-time? If the next write also returns a retry-time, is it acceptable to drop that new metric, even though it didn't get a chance to retry?
I put in a max retries limit based on the connection/`httpClient`, but it seemed wrong, would that be useful to put back in?

plugins/outputs/influxdb_v2/http.go
```diff
@@ -0,0 +1,293 @@
...
+		log.Printf("E! [outputs.influxdb_v2] Failed to write metric: %s\n", desc)
+		return nil
+	case http.StatusTooManyRequests, http.StatusServiceUnavailable:
+		retryAfter := resp.Header.Get("Retry-After")
+		retry, err := strconv.Atoi(retryAfter)
+		if err != nil {
+			retry = 0
+		}
+		if retry > defaultMaxWait {
+			log.Println("E! [outputs.influxdb_v2] Failed to write metric: retry interval too long")
```
danielnelson: Make this more like:
```
if retry > defaultMaxWait {
    retry = defaultMaxWait
}

File: plugins/outputs/influxdb_v2/README.md [hunk 13]
```diff
@@ -0,0 +1,48 @@
...
+  ## HTTP Proxy override, if unset values the standard proxy environment
+  ## variables are consulted to determine which proxy, if any, should be used.
+  # http_proxy = "http://corporate.proxy:3128"
+
+  ## HTTP User-Agent
+  # user_agent = "telegraf"
+
+  ## Content-Encoding for write request body, can be set to "gzip" to
+  ## compress body or "identity" to apply no encoding.
+  # content_encoding = "identity"
```
danielnelson: Let's change the default to gzip.

plugins/outputs/influxdb_v2/influxdb.go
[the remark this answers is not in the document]
danielnelson: I don't think a data migration could handle this since the database doesn't know what should be a int/uint.  When it comes to Telegraf generated data, the only way to determine it today is by reading the code.

PR #4654, opened 2018-09-07
Add support for couchdb 2.0+ API

### LINKED ISSUE #4456: Telegraf reporting zeros for Couchdb 2.1.1 stats endpoint
### Relevant telegraf.conf:

 Read CouchDB Stats from one or more servers
 [[inputs.couchdb]]
 Works with CouchDB stats endpoints out of the box
Multiple HOSTs from which to read CouchDB stats:
   hosts = ["http://###.###.###.###:5986/_stats"]
 internal IP address commented out for config sharing

### System info:

 - Telegraf v1.7.2 (git: release-1.7 c6feb056)
 - Linux MYHOSTNAME 2.6.32-696.30.1.el6.x86_64 #1 SMP Tue May 22 03:28:18 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux

### Steps to reproduce:

1. Installed Telegraf
2. Configured Couchdb Plugin

### Expected behavior:
Telegraf should report values as seen on stats page for a node
### Actual behavior:
When running Telegraf test, all key values returned are 0.

### Additional info:

No error messages are seen in the log when running in debug mode.

glinton:
Resolves #4456

Couchdb modified its API for their 2.0 release. This allows telegraf to collect metrics for 2.0.


plugins/inputs/couchdb/couchdb.go
```diff
@@ -128,6 +134,8 @@ func (c *CouchDB) fetchAndInsertData(accumulator telegraf.Accumulator, host stri
 	c.MapCopy(fields, c.generateFields("couchdb_request_time", stats.Couchdb.RequestTime))
 	c.MapCopy(fields, c.generateFields("couchdb_database_reads", stats.Couchdb.DatabaseReads))
 	c.MapCopy(fields, c.generateFields("couchdb_open_os_files", stats.Couchdb.OpenOsFiles))
+	c.MapCopy(fields, c.generateFields("request_time", stats.RequestTime))
+	c.MapCopy(fields, c.generateFields("open_os_files", stats.OpenOsFiles))
```
danielnelson: I think we should reuse the field name from before, even though I prefer without the couchdb prefix.  It looks like the datatype is compatible.

glinton: The only downside with using the same field name prefix is that the suffix will be different. `couchdb_request_time_value` will be the new populated field while `couchdb_request_time_current` will be populated for pre-2.0 versions.

danielnelson: Yeah, your way fits the input data better.

glinton: Then again, maybe using the same prefix is ok since for 2.0 versions, the `.._value` measurement is what matters. Which brings up another question as to what to do with the unused fields (`_mean`, `_max`, etc...) for 2.0 versions

danielnelson: On second look the prefix is coming from structure of the input data, and now they have moved these fields outside of the couchdb subobject, so I think it is okay to mirror this.

On the unset fields we should make sure not to set them, looks like generateFields needs modified.

plugins/inputs/couchdb/README.md
```diff
@@ -62,194 +62,231 @@ httpd statistics:
 
 ### Example output:
 
+**Post Couchdb 2.0**
 ```
-➜  telegraf git:(master) ✗ ./telegraf --config ./config.conf  --input-filter couchdb --test
-* Plugin: couchdb,
- Collection 1
-> couchdb,server=http://localhost:5984/_stats couchdb_auth_cache_hits_current=0,
-couchdb_auth_cache_hits_max=0,
```
danielnelson: Little bit of README cleanup I have been doing:
- Remove the lines with the command and unrelated bits of output
- Remove the `> ` prefix
- Join all the lines into one big line

plugins/inputs/couchdb/couchdb.go
```diff
@@ -3,65 +3,96 @@ package couchdb
...
+	Stddev  JSONFloat64 `json:"stddev"`
+	Min     JSONFloat64 `json:"min"`
+	Max     JSONFloat64 `json:"max"`
+	Value   JSONFloat64 `json:"value"`
+}
+
+type JSONFloat64 float64
+
+// Used to restore Couchdb <2.0 API behavior
+func (j *JSONFloat64) UnmarshalJSON(data []byte) error {
```
danielnelson: Is this for the unset vs 0 value support?  I think this can be done simpler by making each field in metaData a `*float64`.

glinton: I did try it with the `*float64`, in pre 2.0, the JSON response would contain `{"key":null}` but the plugin would set it to 0. Since we don't wan't to set every [empty] key to 0 (as the 2.0 version would be spammed with unset fields), nor use the pointer and have it not parse "null" values in <2, this `UnmarshalJSON` function is needed to preserve the current behavior.

PR #4667, opened 2018-09-10
Align metrics windows to gather interval for cloudwatch input

### LINKED ISSUE #4643: CloudWatch input metrics start time calculated from period, not interval, leading to missing data
Here's something I'm trying to accomplish:

I want to retrieve certain EC2 metrics every 15 minutes. I want to retrieve the metrics in 5 minute periods, so that for a single metric and dimension, I should have 3 data points. Here's an example using AWSCLI:

```console
# aws cloudwatch get-metric-statistics --namespace AWS/EC2 --region us-east-1 --period 300 --start-time $(date -d '15 minutes ago' --iso-8601=s) --end-time $(date --iso-8601=s) --statistics Average --metric-name CPUUtilization --dimensions Name=InstanceId,Value=i-deadbeef
{
    "Datapoints": [
        {
            "Timestamp": "2018-09-06T17:48:00Z",
            "Average": 12.0,
            "Unit": "Percent"
        },
        {
            "Timestamp": "2018-09-06T17:58:00Z",
            "Average": 37.0,
            "Unit": "Percent"
        },
        {
            "Timestamp": "2018-09-06T17:53:00Z",
            "Average": 25.0,
            "Unit": "Percent"
        }
    ],
    "Label": "CPUUtilization"
}

```

Looking at [the plugin code](https://github.com/influxdata/telegraf/blob/master/plugins/inputs/cloudwatch/cloudwatch.go#L363), though, the "start time" is calculated using the ``period``. So, assuming we ignore the ``delay`` setting just to simplify, setting ``period`` to ``5m`` and ``interval`` to ``15m`` gives me a single data point aggregated over 5 minutes from 5 minutes ago (with 2 data points at 10 and 15 minutes ago skipped entirely). 

I'm not sure if this is a bug or not, but it seems like the start time should be using the ``interval``. In such a case, setting ``period`` to ``5m`` and ``interval`` to ``15m`` gives me exactly what I want: a single API call issued every 15 minutes that contains 3 data points (5, 10, 15 minutes ago)

jcmcken:
Fixes #4643 

For consideration, please review

### Required for all PRs:

### DIFF
--- a/plugins/inputs/cloudwatch/cloudwatch.go
+++ b/plugins/inputs/cloudwatch/cloudwatch.go
@@ -36,6 +36,8 @@ type (
 		RateLimit   int               `toml:"ratelimit"`
 		client      cloudwatchClient
 		metricCache *MetricCache
+		windowStart time.Time
+		windowEnd   time.Time
 	}
 
 	Metric struct {
@@ -197,6 +199,11 @@ func (c *CloudWatch) Gather(acc telegraf.Accumulator) error {
 
 	now := time.Now()
 
+	err = c.updateWindow(now)
+	if err != nil {
+		return err
+	}
+
 	// limit concurrency or we can easily exhaust user connection limit
 	// see cloudwatch API request limits:
 	// http://docs.aws.amazon.com/AmazonCloudWatch/latest/DeveloperGuide/cloudwatch_limits.html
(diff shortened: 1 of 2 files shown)


jcmcken:
There didn't seem to be an obvious place that I should update the README, so I made no changes.

plugins/inputs/cloudwatch/cloudwatch.go
```diff
@@ -208,14 +215,32 @@ func (c *CloudWatch) Gather(acc telegraf.Accumulator) error {
...
 
 	return nil
 }
 
+func (c *CloudWatch) updateWindow(relativeTo time.Time) error {
+	defaultTime := time.Time{}
+
+	windowEnd := relativeTo.Add(-c.Delay.Duration)
+
+	if c.windowEnd == defaultTime {
```
danielnelson: Nitpick: `if c.windowEnd.IsZero()`

jcmcken: Didn't know that function was there. I'll update

As for the rounding, is there something you had in mind? It looks like, based on [this](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/cloudwatch_concepts.html#CloudWatchPeriods) that non-custom metrics have a minimum period of 60s. It also says:

>When statistics are aggregated over a period of time, they are stamped with the time corresponding to the beginning of the period.

Seems to indicate it should be aligned to the start of the minute. Is this what you're referring to?

danielnelson: On the rounding, it's probably not something that needs done as I don't see us doing it before.  I was just curious if the times needed to rounded and aligned, but I guess either of these ranges would work fine (second one is not aligned to period) and what really matters is that we cover all times?
```
--start-time 2018-09-06T17:15:00Z --end-time 2018-09-06T17:30:00Z --period 300
--start-time 2018-09-06T17:16:42Z --end-time 2018-09-06T17:31:42Z --period 300
```

jcmcken: I fixed the zero function.

I think we probably don't need to align the data as long as we cover all time periods, like you said.

PR #4669, opened 2018-09-10
Add means to specify server password for redis input

### LINKED ISSUE #2786: Add password option to redis plugin
### Relevant telegraf.conf:
[[inputs.redis]]
    servers = ["tcp://123456$%^@localhost:6379"]
### System info:
debian 8.8
Telegraf  1.2.1

### Steps to reproduce:

telegraf -config /etc/telegraf/telegraf.conf -input-filter redis -test

### Expected behavior:
Can generate right redis info 

### Actual behavior:
* Plugin: inputs.redis, Collection 1
Unable to parse to address 'tcp://123456$%^@localhost:6379': parse tcp://123456$%^@localhost:6379: invalid URL escape "%^"

### Current behavior:
only alphanumerics, the special characters "$-_.+!*'(),", and reserved characters used for their reserved purposes may be used unencoded within a URL

### Desired behavior:
allow other character in password

all password has this problem

### LINKED ISSUE #4663: plugins/inputs/redis password contain "#" parse fail
[[inputs.redis]]
servers=["tcp://:passw@rd#aaa@localhost:6379"]

command：
telegraf --config telegraf.conf --input-filter redis --test

result：
 E! Error in plugin [inputs.redis]: dial tcp: address rd: missing port in address

glinton:
Resolves #4663 
Resolves #2786

### DIFF
--- a/plugins/inputs/redis/redis.go
+++ b/plugins/inputs/redis/redis.go
@@ -2,7 +2,6 @@ package redis
 
 import (
 	"bufio"
-	"errors"
 	"fmt"
 	"io"
 	"log"
@@ -19,8 +18,8 @@ import (
 )
 
 type Redis struct {
-	Servers   []string
-	Passwords []string
+	Servers  []string
+	Password string
 	tls.ClientConfig
 
 	clients     []Client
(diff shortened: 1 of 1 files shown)


plugins/inputs/redis/redis.go
```diff
@@ -18,7 +19,8 @@ import (
 )
 
 type Redis struct {
-	Servers []string
+	Servers   []string
+	Passwords []string
```
danielnelson: Just do a single password, if the password is different across servers you will need to do multiple plugin definitions.

Issue #3319, opened 2017-10-10
cloudwatch output uses ListMetricData


adamchainz:
## Bug report

When starting the Cloudwatch output plugin, it tries to list some metrics, even though this isn't needed for normal operation: https://github.com/influxdata/telegraf/blob/ac1fa05672a3eb9342b23534f22cb5538576e759/plugins/outputs/cloudwatch/cloudwatch.go#L62 . Thus the AWS credentials in use need access to `ListMetricData` even though when in use the plugin only needs `PutMetricData`.

The required permissions aren't even documented for the plugin.

### Relevant telegraf.conf:

```ini
[[outputs.cloudwatch]]
region = "us-east-1"
namespace = "foo"
```

### System info:

Latest telegraf

### Steps to reproduce:

1. Create an AWS IAM user/role with only `PutMetricData` rights
2. Configure cloudwatch output in telegraf, start it

### Expected behavior:

Successful start

### Actual behavior:

It fails

### Additional info:

AWS STS GetSessionToken can be used to test connection without needing any specific permissions: https://docs.aws.amazon.com/sdk-for-go/api/service/sts/#STS.GetSessionToken

danielnelson:
Thanks for the report, I agree on all points.  Is this something you would be able to work on?

adamchainz:
Never used this language before but I’ll give it a... Go

PR #4681, opened 2018-09-11
don't try to reach STS if role_arn is not specified

MichaelCosby:
When telegraf is running inside a VPC, STS (AWS Simple Token Service) may not be reachable. This makes the cloudwatch output plugin fail before it even tries other methods of getting its authentication tokens.

Since the documentation already states that STS is only tried if role_arn is specified, this patch skips STS entirely if the role_arn is blank.
### Required for all PRs:

Since the change I'm making here simply makes the existing README.md true, there was no need to modify it.

I didn't add a unit test; I don't know enough about go yet to know how to mock out the sts module and the existing tests don't have a good example to copy. Assuming that's required for this change, I can learn that...

### DIFF
--- a/plugins/outputs/cloudwatch/cloudwatch.go
+++ b/plugins/outputs/cloudwatch/cloudwatch.go
@@ -210,17 +210,19 @@ func (c *CloudWatch) Connect() error {
 	}
 	configProvider := credentialConfig.Credentials()
 
-	stsService := sts.New(configProvider)
+	if c.RoleARN != "" {
+		stsService := sts.New(configProvider)
 
-	params := &sts.GetCallerIdentityInput{}
+		params := &sts.GetCallerIdentityInput{}
 
-	_, err := stsService.GetCallerIdentity(params)
+		_, err := stsService.GetCallerIdentity(params)
 
-	if err != nil {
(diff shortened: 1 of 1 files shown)


danielnelson:
This code is only used to test if the authentication works, and isn't critical even with other authentication methods.  I think perhaps we should just remove it altogether and rely on errors when we actually try to write.  @adamchainz What do you think?

adamchainz:
> Since the documentation already states that STS is only tried if role_arn is specified, this patch skips STS entirely if the role_arn is blank.

That's not quite true. It's saying the credentials are explicitly taken from STS if `role_arn` is provided.

The check using STS is separate, and something I added because I believed STS is always reachable. Some VPC's being configured to make AWS services unreachable breaks that assumption.

I agree with @danielnelson , removing the check entirely is probably the only sensible course of action. It would still leave my original issue #3319, resolved, just it's a shame configuration can't be validated at startup.

danielnelson:
I opened a PR that removes the check altogether #4695, I don't think it is worth having the check only in the rare case that the `role_arn` is set.

MichaelCosby:
That sounds good. I'll go ahead and drop this pull request.

PR #4686, opened 2018-09-12
String replacer function to strings plugin

doublerainbo:
### Required for all PRs:

In reference to https://github.com/influxdata/telegraf/pull/4599

Adding the ability to replace substrings in a field, tag, or measurement. Also adding the ability to
match on multiple of the resource with '*'. With that active, the `dest` field is ignored and the same data is written back to the same field name to avoid writing all modified data to the same field name.


doublerainbo:
@danielnelson i added a bit more than just the substring replacement in this PR since we need the ability to run a replace on every metric going through, but i'd imagine '*' should be fairly contained since it's likely not a valid name for the existing tags/fields/metrics 🤞 . If you can take a look that'd be awesome. Thanks!

PR #4700, opened 2018-09-15
Google Core IoT

davidgs:
### Required for all PRs:

Resubmitting after changes, cleanup.


jamesmaidment:
@davidgs Given that this could be used as a generic Cloud Pubsub push endpoint - is there a reason it's titled as a Google IoT?

.gitignore
```diff
@@ -3,3 +3,8 @@
 /telegraf.exe
 /telegraf.gz
 /vendor
+plugins/inputs/.DS_Store
+plugins/.DS_Store
+OpenWrt.sparseimage
+telegraf.arm64
+.DS_Store
```
danielnelson: Revert this file, though I suggest adding .DS_Store to your systemwide ignore file, if you don't have one you can add it with:
```
git config --global core.excludesfile $HOME/.gitignore
```

Then add `.DS_Store` to `$HOME/.gitignore`.

plugins/inputs/googlecoreiot/README.md
```diff
@@ -0,0 +1,69 @@
+# Google Core IoT service input plugin
+
+The Google Core IoT listener is a service input plugin that listens for messages sent via HTTP POST from Google's Pub/Sub service.
```
danielnelson: Add link to associated service documentation.  I'm wondering what makes this a Google Core IoT plugin instead of a general pub/sub consumer, can you explain the relationship between these technologies?  Whats the scoop on Google Core IoT vs Google Cloud IoT?

plugins/inputs/googlecoreiot/googlecoreIot_test.go
```diff
@@ -0,0 +1,333 @@
+package googlecoreiot
```
danielnelson: Rename file googlecoreiot_test.go

plugins/inputs/googlecoreiot/googlecoreiot.go
```diff
@@ -0,0 +1,436 @@
...
+	Port            int
+	Precision       string
+	DataFormat      string
+	MeasurementName string
+	handler         *influx.MetricHandler
+
+	tlsint.ServerConfig
+
+	BasicUsername string
+	BasicPassword string
```
danielnelson: Remove any options no longer used, such as Methods, BasicUsername, BasicPassword.  Make sure any exported variable is documented.

plugins/inputs/googlecoreiot/googlecoreiot.go
```diff
@@ -0,0 +1,436 @@
...
+	ServiceAddress  string
+	Path            string
+	Methods         []string
+	ReadTimeout     internal.Duration
+	WriteTimeout    internal.Duration
+	MaxBodySize     internal.Size
+	Port            int
+	Precision       string
+	DataFormat      string
+	MeasurementName string
```
danielnelson: Remove MeasurementName, can be set using the `name_override` setting on any input plugin.

plugins/inputs/googlecoreiot/googlecoreiot.go
```diff
@@ -0,0 +1,436 @@
...
+	return metrics, nil
+
+}
+
+func (h *GoogleListener) serveNotFound(res http.ResponseWriter, req *http.Request) {
+	res.Header().Set("Content-Type", "application/json")
+	res.Header().Set("X-Influxdb-Version", "1.0")
+	res.WriteHeader(http.StatusNotFound)
+	res.Write([]byte(`{"error":"http: not found"}`))
+	return
```
danielnelson: Does Google define how the error response should look?  If not, just return an error code with no body.

plugins/inputs/googlecoreiot/README.md
```diff
@@ -0,0 +1,69 @@
...
+  write_timeout = "10s"
+
+  # precision of the time stamps. can be one of the following:
+  # second, millisecond, microsecond, nanosecond
+  # Default is nanosecond
+  
+  precision = "nanosecond"
+  
+  # Data Format is either line protocol or json
+  protocol="line protocol"
```
danielnelson: Make this plugin a ParserInput by adding the `func (*GoogleListener) SetParser(parser parsers.Parser) {` function.  This will allow the plugin to parse a payload in any supported data format.

Some potential issue with this idea is that the JSONData format defined below is not a supported payload.  Additionally the influx parser does not support changing the precision.  Overall it should be better though as it will make it more consistent with the rest of Telegraf and all formats will be supported, code will be simpler too.

plugins/inputs/googlecoreiot/googlecoreiot.go
```diff
@@ -0,0 +1,436 @@
...
+}
+
+func (h *GoogleListener) serveWrite(res http.ResponseWriter, req *http.Request) {
+	// Check that the content length is not too large for us to handle.
```
danielnelson: This function will be called concurrently, you will need to add appropriate locking for shared data in the GoogleListener struct.

plugins/inputs/googlecoreiot/googlecoreIot_test.go
```diff
@@ -0,0 +1,333 @@
...
+	"github.com/influxdata/telegraf/internal"
+	"github.com/influxdata/telegraf/plugins/parsers"
+	influx "github.com/influxdata/telegraf/plugins/parsers/influx"
+	"github.com/influxdata/telegraf/testutil"
+
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	testMsg          = `{"message":{"attributes":{"deviceId":"myPi","deviceNumId":"2808946627307959","deviceRegistryId":"my-registry","deviceRegistryLocation":"us-central1","projectId":"conference-demos","subFolder":""},"data":"dGVzdGluZ0dvb2dsZSxzZW5zb3I9Ym1lXzI4MCB0ZW1wX2M9MjMuOTUsaHVtaWRpdHk9NjIuODMgMTUzNjk1Mjk3NDU1MzUxMDIzMQ==","messageId":"204004313210337","message_id":"204004313210337","publishTime":"2018-09-14T19:22:54.587Z","publish_time":"2018-09-14T19:22:54.587Z"},"subscription":"projects/conference-demos/subscriptions/my-subscription"}\n`
```
danielnelson: This test file need pretty significant changes.  The test data needs to be created to exercise specific functionality, and can't just be a big pile of actual data, as it makes it impossible to tell what is being tested in each case.  Each test case should test one piece of functionality in googleiotcore.go, if you have lots of similar cases use a table test.  On the testmsgs.gz, we should just gzip the body before sending it to the test function, in code instead of precompressed on disk.  We can talk more about this directly.

.gitignore
davidgs: Done.

plugins/inputs/googlecoreiot/README.md
davidgs: Please speak to Sharjeel and Russ about this.

plugins/inputs/googlecoreiot/README.md
davidgs: If I make this change, then there is a high likelihood that input from the Google Pub/Sub will fail due to precision mis-match, etc.

plugins/inputs/googlecoreiot/googlecoreiot.go
davidgs: The measurement name can be set via the Google Core IoT platform, so it needs to be included here in order for it to parse correctly.

danielnelson:
#5442

PR #4702, opened 2018-09-15
use less goroutine and memory while executing multiple cmds at once

Jaeyo:
### Required for all PRs:

too many goroutine while executing multiple cmds at once


Jaeyo:
@danielnelson can you review again whether your requesting change all adopted?

internal/internal.go
```diff
@@ -153,22 +153,26 @@ func RunTimeout(c *exec.Cmd, timeout time.Duration) error {
...
-	timer := time.NewTimer(timeout)
-	done := make(chan error)
-	go func() { done <- c.Wait() }()
-	select {
-	case err := <-done:
-		timer.Stop()
-		return err
-	case <-timer.C:
-		if err := c.Process.Kill(); err != nil {
+	var err error
```
danielnelson: This error is never assigned to, remove it and the logic checking it below.

internal/internal.go
```diff
@@ -153,22 +153,26 @@ func RunTimeout(c *exec.Cmd, timeout time.Duration) error {
...
+		if err != nil {
 			log.Printf("E! FATAL error killing process: %s", err)
-			return err
+			return
 		}
-		// wait for the command to return after killing it
-		<-done
+	})
+
+	exitErr := c.Wait()
```
danielnelson: If you remove the err above, this is the only error so I would probably just call it `err`.

Jaeyo:
@danielnelson Done! :D

Issue #2159, opened 2016-12-14
Support IPv6 in the ping plugin


kostasb:
**Feature Request**: Support IPv6 ping by adding a configuration option in the plugin to switch to using the ping6 tool instead of ping.

IPv6 ICMP is not supported by the ping tool.

```
ping6 ipv6.google.com
PING6(56=40+8+8 bytes) 2a02:214d:8225:4000:b932:d1e7:533d:be94 --> 2a00:1450:4016:804::200e
16 bytes from 2a00:1450:4016:804::200e, icmp_seq=0 hlim=56 time=43.345 ms
```

Reference:

```
The ping6 utility is intentionally separate from ping.

     There have been many discussions on why we separate ping6 and ping.
     Some people argued that it would be more convenient to uniform the ping
     command for both IPv4 and IPv6.  The followings are an answer to the
     request.

     From a developer's point of view: since the underling raw sockets API is
     totally different between IPv4 and IPv6, we would end up having two types
     of code base.  There would actually be less benefit to uniform the two
     commands into a single command from the developer's standpoint.

     From an operator's point of view: unlike ordinary network applications
     like remote login tools, we are usually aware of address family when
     using network management tools.  We do not just want to know the
     reachability to the host, but want to know the reachability to the host
     via a particular network protocol such as IPv6.  Thus, even if we had a
     unified ping command for both IPv4 and IPv6, we would usually type a
     -6 or -4 option (or something like those) to specify the particular
     address family.  This essentially means that we have two different
     commands.
```

[source](https://www.freebsd.org/cgi/man.cgi?query=ping6&sektion=8#end)

danielnelson:
We need to gather ping6 examples for Linux, BSD, Windows.  Also interesting, on Debian ping6 is a symlink to ping which has a `-6` option.

MarkMartinec:
> From an operator's point of view: unlike ordinary network applications like remote login tools, we are usually aware of address family when using network management tools.  We do not just want to know the reachability to the host, but want to know the reachability to the host via a particular network protocol such as IPv6. Thus, even if we had a unified ping command for both IPv4 and IPv6, we would usually type a -6 or -4 option (or something like those) to specify the particular address family.  This essentially means that we have two different commands.

[...]

PR #4703, opened 2018-09-15
Support IPv6 in the ping plugin

Jaeyo:
### Required for all PRs:

#2159

### DIFF
--- a/plugins/inputs/ping/ping.go
+++ b/plugins/inputs/ping/ping.go
@@ -44,7 +44,7 @@ type Ping struct {
 	Urls []string
 
 	// URLs to ping ipv6 address
-	UrlsV6 []string `toml:urls_v6`
+	UrlsV6 []string `toml:"urls_v6"`
 
 	// host ping function
 	pingHost HostPinger
@@ -89,19 +89,19 @@ func (p *Ping) Gather(acc telegraf.Accumulator) error {
 	// Spin off a go routine for each url to ping
 	for _, url := range p.Urls {
 		wg.Add(1)
-		go p.pingToURL(url, false, wg, acc)
+		go p.pingToURL(url, false, &wg, acc)
 	}
 	for _, url := range p.UrlsV6 {
 		wg.Add(1)
-		go p.pingToURL(url, true, wg, acc)
+		go p.pingToURL(url, true, &wg, acc)
 	}
 
-	wg.wait()
+	wg.Wait()
 
 	return nil
 }
 
-func (p *Ping) pingToURL(u string, isV6 bool, wg sync.WaitGroup, acc telegraf.Accumulator) {
+func (p *Ping) pingToURL(u string, isV6 bool, wg *sync.WaitGroup, acc telegraf.Accumulator) {
 	defer wg.Done()
 	tags := map[string]string{"url": u}
 	fields := map[string]interface{}{"result_code": 0}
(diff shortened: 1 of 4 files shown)


danielnelson:
The way we've been planning to implement this is with a new option to override the ping command used as well as a way to set arguments directly.  If you needed to do some ipv4 and some ipv6 urls then you would define the plugin twice.

```
[[inputs.ping]]
  urls = ["www.example.org"]
  binary = "ping6"
  arguments = ["-6"]
```

Do you think you could change around the options like this?

Jaeyo:
@danielnelson review again please

Jaeyo:
@danielnelson so the existing options for ping input (ex. ping_interval, timeout, deadline) will replaced by "arguments"?

danielnelson:
Yes, I think what would be best is if when `arguments` is empty we us the options that currently exist, otherwise we only use the arguments and ignore `ping_interval`, `timeout`, etc.

Jaeyo:
@danielnelson can you review this?

plugins/inputs/ping/ping.go
```diff
@@ -71,97 +80,106 @@ const sampleConfig = `
...
-
 	// Spin off a go routine for each url to ping
 	for _, url := range p.Urls {
-		wg.Add(1)
-		go func(u string) {
-			defer wg.Done()
-			tags := map[string]string{"url": u}
-			fields := map[string]interface{}{"result_code": 0}
+		p.wg.Add(1)
+		p.pingToURL(url, acc)
```
glinton: Goroutine this - `go p.pingToURL(url, acc)`

plugins/inputs/ping/ping.go
```diff
@@ -173,19 +191,25 @@ func hostPinger(timeout float64, args ...string) (string, error) {
...
-		case "darwin", "freebsd", "netbsd", "openbsd":
+		case "darwin":
 			args = append(args, "-W", strconv.FormatFloat(p.Timeout*1000, 'f', -1, 64))
+		case "freebsd", "netbsd", "openbsd":
+			args = append(args, "-w", strconv.FormatFloat(p.Timeout*1000, 'f', -1, 64))
 		case "linux":
 			args = append(args, "-W", strconv.FormatFloat(p.Timeout, 'f', -1, 64))
 		default:
-			// Not sure the best option here, just assume GNU ping?
+			// not sure the best option here, just assume gnu ping?
```
glinton: Why the case change in certain comments?

plugins/inputs/ping/ping.go
```diff
@@ -196,28 +220,30 @@ func (p *Ping) args(url string, system string) []string {
...
+			args = append(args, "-i", p.Interface)
 		}
 	}
 	args = append(args, url)
 	return args
 }
 
-// processPingOutput takes in a string output from the ping command, like:
+// processpingoutput takes in a string output from the ping command, like:
```
glinton: Revert case here as well

plugins/inputs/ping/ping_windows.go
```diff
@@ -49,14 +59,100 @@ const sampleConfig = `
...
-	bin, err := exec.LookPath("ping")
+func (p *Ping) Gather(acc telegraf.Accumulator) error {
+	if p.Count < 1 {
+		p.Count = 1
+	}
+
+	// Spin off a go routine for each url to ping
+	for _, url := range p.Urls {
+		p.wg.Add(1)
+		p.pingToURL(url, acc)
```
glinton: goroutine this too

Issue #1074, opened 2016-04-21
Docker Plugin not finding Docker socket


hrzbrg:
Hi

after I updated a testing host from 0.11.1 to 0.12.1 the Docker plugin stopped working. The log shows:
```
Error in input [docker]: Cannot connect to the Docker daemon. Is the docker daemon running on this host?
```

But when I run `telegraf -config /etc/telegraf/telegraf.conf -input-filter docker -test` it outputs the values to stdout. 

I also tried the most recent nightly, which produces the same error.

docker info is: 
```
Containers: 5
 Running: 3
 Paused: 0
 Stopped: 2
Images: 69
Server Version: 1.10.3
Storage Driver: aufs
 Root Dir: /var/lib/docker/aufs
 Backing Filesystem: extfs
 Dirs: 182
 Dirperm1 Supported: true
Execution Driver: native-0.2
Logging Driver: json-file
Plugins:
 Volume: local
 Network: bridge null host
Kernel Version: 4.2.0-35-generic
Operating System: Ubuntu 14.04.4 LTS
OSType: linux
Architecture: x86_64
CPUs: 8
Total Memory: 7.796 GiB
```

Regards
hrzbrg

whetherharder:
Looks like permissions problem.
21.04.2016 12:11 пользователь "Sebastian" <notifications@github.com>
написал:

sparrc:
try running as the telegraf user: `sudo -u telegraf telegraf -config /etc/telegraf/telegraf.conf -input-filter docker -test`

hrzbrg:
okay, `sudo usermod -aG docker telegraf` solves it

sparrc:
thanks for posting the solution @hrzbrg

PR #4724, opened 2018-09-20
Add readme note about docker socket permissions

gunnaraasen:
This updates the docker input plugin readme with a section about setting permissions for telegraf to access the docker daemon socket. This should address some of the confusion about why the docker input plugin does not work out of the box ([as seen here](https://github.com/influxdata/telegraf/issues/1074#issuecomment-213040280)).

There should probably be a note about security considerations with the current method. That could be a link to [this page](https://docs.docker.com/install/linux/linux-postinstall/#use-a-different-storage-engine) or [this page](https://docs.docker.com/engine/security/https/) on the Docker docs.

A better solution would be showing a way to give telegraf read-only permissions to the socket. Any suggestions on the "right" way to do that would be appreciated.

### Required for all PRs:

### DIFF
--- a/plugins/inputs/docker/README.md
+++ b/plugins/inputs/docker/README.md
@@ -66,6 +66,31 @@ to gather stats from the [Engine API](https://docs.docker.com/engine/api/v1.24/)
 When using the `"ENV"` endpoint, the connection is configured using the
 [cli Docker environment variables](https://godoc.org/github.com/moby/moby/client#NewEnvClient).
 
+#### Security
+
+Giving telegraf access to the Docker daemon expands the [attack surface](https://docs.docker.com/engine/security/security/#docker-daemon-attack-surface) that could result in an attacker gaining root access to a machine. This is especially relevant if the telegraf configuration can be changed by untrusted users.
+
+#### Docker Daemon Permissions
+
+Typically, telegraf must be given permission to access the docker daemon unix
+socket when using the default endpoint. This can be done by adding the
+`telegraf` unix user (created when installing a Telegraf package) to the
+`docker` unix group with the following command:
+
+```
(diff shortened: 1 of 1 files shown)


plugins/inputs/docker/README.md
```diff
@@ -66,6 +68,26 @@ to gather stats from the [Engine API](https://docs.docker.com/engine/api/v1.24/)
 When using the `"ENV"` endpoint, the connection is configured using the
 [cli Docker environment variables](https://godoc.org/github.com/moby/moby/client#NewEnvClient).
 
+#### Permissions
+
+When using the default unix socket endpoint, telegraf must be given permission
+to access the socket. This can be done by adding telegraf to the docker group
```
glinton: Maybe define "telegraf" as a user
> adding the telegraf user to the docker group

gunnaraasen: I think all our packages already defines a `telegraf` user (and maybe also a `telegraf` group.), but I'll add a note that the user may need to be created if the packages are not used.

glinton: It does already create that user. I was referencing the documentation and how it's assumed that a user knows that you are talking about the `telegraf` user, and not the `telegraf` binary in the docs. A bit pedantic, but there's a slight chance of confusion.

danielnelson:
I think we should leave out the comment in the config file and only have it in the README.

plugins/inputs/docker/README.md
gunnaraasen: Ah, yeah, I misunderstood. I'll call that out.

gunnaraasen:
I made some changes, including a new section to call out the security risks of giving telegraf access to the docker daemon socket.

PR #4730, opened 2018-09-21
Forwarded Records and Offline

tboggiano:
Corrected spelling of Forwarded Records and added correct state status of 6 for offline databses on permise

### Required for all PRs:


danielnelson:
@tboggiano Can you remove the whitespace changes so that there will be less churn in the git history?

@m82labs Can you review?

tboggiano:
I didn't take out the whitepsaces on purpose VSCode automatically removes them. I tried to manually put them back in already and it just removes them again.

m82labs:
I ran all the queries and they all run fine. I would actually be fine with the whitespace changes if it's not a big deal. I confirmed the forwarded records metric returns values as well.

m82labs:
@tboggiano are you able to build and test this on your system? I don't think the whitespace changes will cause trouble when the queries are actually ran through the plugin, but it would need to be tested. If you are unable to build, I can build it locally tonight and test it out.

tboggiano:
@m82labs yes let me build it out and stick on a live server

tboggiano:
Tested both and have data showing up in the dashboards.

m82labs:
Thanks @tboggiano!

PR #4747, opened 2018-09-25
Add hostname to TLS config for SNI support

onurguzel:
x509_cert plugin was creating TLS connection without specifying a
server name for handshake. If there are multiple certificates on the
same host, the first one is received, resulting insecure host warning.

ServerName in tls.Config turns Server Name Indication (SNI) on for the
handshake, which allows virtual hosts on the same server.

### Required for all PRs:


onurguzel:
I only checked the first box, because there is no README.md changes needed and I'm not sure how you feel about badsll.com in unit tests.

PR #4752, opened 2018-09-26
Per-directory file counts in the Filecount plugin

### LINKED ISSUE #4749: Per-directory file counts in the Filecount plugin
### Proposal:

Perhaps the `directory` section of the plugin config becomes a glob...similar to the `file` section?

### Current behavior:

Recursively totals the number of occurrences of certain types/names of files within a higher level directory

### Desired behavior:

I think it would benefit our users to see the file counts by name in each subdirectory of the top level directory you feed the config.

The first use case I could think of for this was counting `tsm` and `tsi` files in an `influxdb/data` directory.  Getting the full count is fine for keeping track, at a high level, of how many data files you're generating.  However, for debugging purposes, it would be nice to know if certain shard directories are acting up (i.e., not compacting enough files, producing too many files, etc.).  The current behavior won't provide this information.

Jaeyo:
### Required for all PRs:

close #4749

### DIFF
--- a/plugins/inputs/filecount/filecount.go
+++ b/plugins/inputs/filecount/filecount.go
@@ -14,6 +14,11 @@ import (
 
 const sampleConfig = `
   ## Directory to gather stats about.
+  ## This accept standard unit glob matching rules, but with the addition of
+  ## ** as a "super asterisk". ie:
+  ##   /var/log/**    -> recursively find all directories in /var/log
+  ##   /var/log/*/*   -> find all directories with a parent dir in /var/log
+  ##   /var/log       -> only count the /var/log dir
   directory = "/var/cache/apt/archives"
 
   ## Only count files that match the name pattern. Defaults to "*".
--- a/plugins/inputs/filecount/filecount.go
+++ b/plugins/inputs/filecount/filecount.go
@@ -216,7 +216,7 @@ func getTargetDirs(directory string) ([]string, error) {
 
 	filtered := []string{}
 	for path, file := range g.Match() {
-		if file.IsDir() == true {
+		if file.IsDir() {
 			filtered = append(filtered, path)
 		}
 	}
--- a/plugins/inputs/filecount/filecount.go
+++ b/plugins/inputs/filecount/filecount.go
@@ -16,9 +16,9 @@ const sampleConfig = `
   ## Directory to gather stats about.
   ## This accept standard unit glob matching rules, but with the addition of
   ## ** as a "super asterisk". ie:
-  ##   /var/log/**    -> recursively find all directories in /var/log
-  ##   /var/log/*/*   -> find all directories with a parent dir in /var/log
-  ##   /var/log       -> only count the /var/log dir
+  ##   /var/log/**    -> recursively find all directories in /var/log and count files in each directories
+  ##   /var/log/*/*   -> find all directories with a parent dir in /var/log and count files in each directories
+  ##   /var/log       -> count all files in /var/log and all of its subdirectories
   directory = "/var/cache/apt/archives"
 
   ## Only count files that match the name pattern. Defaults to "*".
--- a/plugins/inputs/filecount/filecount.go
+++ b/plugins/inputs/filecount/filecount.go
@@ -14,12 +14,16 @@ import (
 
 const sampleConfig = `
   ## Directory to gather stats about.
+  ##   deprecated in 1.9; use the directories option
+  directory = "/var/cache/apt/archives"
+
+  ## Directories to gather stats about.
   ## This accept standard unit glob matching rules, but with the addition of
   ## ** as a "super asterisk". ie:
   ##   /var/log/**    -> recursively find all directories in /var/log and count files in each directories
   ##   /var/log/*/*   -> find all directories with a parent dir in /var/log and count files in each directories
   ##   /var/log       -> count all files in /var/log and all of its subdirectories
-  directory = "/var/cache/apt/archives"
+  directories = ["/var/cache/apt/archives"]
 
   ## Only count files that match the name pattern. Defaults to "*".
   name = "*.deb"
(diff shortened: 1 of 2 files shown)


Jaeyo:
@glinton @danielnelson how about use `directories` than `directory` like file input?

glinton:
Would `directories` take an array of glob strings? Similar to the exec input `commands` vs `command`?

plugins/inputs/filecount/filecount.go
```diff
@@ -14,6 +14,11 @@ import (
 
 const sampleConfig = `
   ## Directory to gather stats about.
+  ## This accept standard unit glob matching rules, but with the addition of
+  ## ** as a "super asterisk". ie:
+  ##   /var/log/**    -> recursively find all directories in /var/log
+  ##   /var/log/*/*   -> find all directories with a parent dir in /var/log
+  ##   /var/log       -> only count the /var/log dir
```
sometimesfood: In this context, "only count the /var/log dir" is a bit misleading. By default, using `/var/log` counts all files in `/var/log` and all of its subdirectories.

In my opinion we should distinguish counting files in a directory from producing separate stats (fields) for a directory.

sometimesfood:
Regarding the docs:

In this context, "only count the /var/log dir" is a bit misleading. By default, using `/var/log` counts all files in `/var/log` and all of its subdirectories.

In my opinion we should distinguish between counting files in a directory and producing separate stats (fields) for a directory. These are not the same.

sometimesfood:
> Would `directories` take an array of glob strings? Similar to the exec input `commands` vs `command`?

Wouldn't it be more consistent to drop "directory" in favour of "directories" (array of globs) altogether?

If a glob expression is used, the filecount plugin has to deal with multiple base directories and report stats for all matching directories accordingly.

sometimesfood:
@Jaeyo Since you removed the `countFn` argument, you should get rid of the `countFunc` type.

sometimesfood:
Also, wouldn't it be a good idea to add some updated unit tests to test the new behaviour?

I hope my comments are not coming across as too nitpicky. I actually quite like the approach taken in this PR, I just think some minor details could be improved a bit.

plugins/inputs/filecount/filecount.go
```diff
@@ -168,34 +197,33 @@ func (fc *FileCount) filter(file os.FileInfo) (bool, error) {
...
 
+func getTargetDirs(directory string) ([]string, error) {
+	g, err := globpath.Compile(directory)
+	if err != nil {
+		return nil, fmt.Errorf("could not compile glob %v: %v", directory, err)
+	}
+
+	filtered := []string{}
+	for path, file := range g.Match() {
+		if file.IsDir() == true {
```
sometimesfood: This should read `if file.IsDir() {` instead of `if file.IsDir() == true {`.

glinton:
>Wouldn't it be more consistent to drop "directory" in favour of "directories" (array of globs) altogether?

I agree, however one major drawback with that is breaking compatibility with config files from older versions.

Jaeyo:
i think my comment lacked a little explanation.
my sugestion is,
1. add "directories" option that is array of glob string.
2. mark "directory" as deprecated.

and like @sometimesfood comment, doc also should be modified.

thanks for all comment.
@glinton

sometimesfood:
> I agree, however one major drawback with that is breaking compatibility with config files from older versions.

Marking "directory" is marked as deprecated in favour of "directories" sounds like a good approach in order to avoid breaking existing configurations.

Jaeyo:
`directories` added

Samuel-BF:
Reading the code, it seems that it reads the same directories several times (in case of recursive ask) : 
if you have the hierarchy /var/log/{apache2/{access/,error/},nginx/{access/,error/},mysql/}, then when you'll ask for [/var/log/**] it will count elements in /var/log/apache2/access directory 3 times (once for /var/log/apache2/access, once for /var/log/apache2 and once for /var/log).

Imho, it would be more efficient to parse once all directories rather than to launch walkFn for each directory found. I may propose a fix, but not before this friday.

Jaeyo:
@Samuel-BF thanks for your suggestion. i would appreciate any fix at your convenience.

Samuel-BF:
Well, I already recoded the plugin without filepath.Walk (using rather recursive functions) for the #4778 PR. 

It seemed simpler to modify it to your work in it (Directories option), that's what I made in 18373c7bf8450c85500b431689808a24804fed74 .

Detail : in fact, your version parses *n+1* times the directories, where *n* is the level of recursion of directories (once in globpath.Match and one time per recursion level).

I made one test comparing our two versions on my system (running it against source tree of telegraf) :
```
$ time ./telegraf --config test.conf --test # PR 4778
[... lots of outputs]
real    0m0,717s
user    0m0,370s
sys     0m0,426s
$ time ./telegraf --config test.conf --test # PR 4752
real    0m3,159s
user    0m1,849s
sys     0m1,725s
```
Of course, the gain is proportional to your folder depth.

sometimesfood:
@Samuel-BF: While I appreciate the better efficiency and the resulting speedup, I dislike the `recursive_print_size` option. Hardcoding this type of behaviour seems like a bad idea to me.

Samuel-BF:
> @Samuel-BF: While I appreciate the better efficiency and the resulting speedup, I dislike the `recursive_print_size` option. Hardcoding this type of behaviour seems like a bad idea to me.

Let me explain my use case.
I'm a sysadmin willing to monitor the use of a shared disk by employees (they all have a personal folder on it, no quota yet). I want to be able to identify folders which takes a lot of place. Without this option, i have to : 
- either output informations for all directories to my influxDB backend (which make a lot of unnecessary data)
- either restrict information to a predefined set of folders (e.g. .../users/*)
An option like recursive_print_size allows me to have a configurable balance between these two extremas.

But I agree, it doesn't seem very elegant, both the term and the logic. Any proposal ? I'll check the processors or agreggators plugins to see if it allows me to drop measurements.

sometimesfood:
I think I understand your use case, but IMHO this should be addressed on a different abstraction level. Using processors or aggregators might be a solution to this type of problem.

I added some comments in your pull request, sorry about splitting up the discussion.

danielnelson:
I added a few comments to #4778 about the proposed options, but we can add other changes from there on top of this work.

PR #4754, opened 2018-09-26
parallel dns_query

### LINKED ISSUE #2547: dns_query should query all servers in parallel
### Proposal:
The `dns_query` input should query all servers in parallel.

### Current behavior:
The `dns_query` plugin queries all server sequentially.

### Desired behavior:
The `dns_query` input should query all servers in parallel.

When diagnosing issues, having all the servers queried at the same time is vitally important. For example if you're querying 2 different DNS servers with a recursive lookup, and there is an issue with the recursion upstream, both servers will be affected. However if it is only a momentary issue, telegraf is going to hit the first DNS server, it'll stall out during recursion until the momentary issue passes, and then it'll proceed to the second server which will return fast because the upstream issue passed. Thus the first DNS server will appear to have a problem, when really the problem is just visible on whatever server is first in the list.

Jaeyo:
### Required for all PRs:

close #2547

### DIFF
--- a/plugins/inputs/dns_query/dns_query.go
+++ b/plugins/inputs/dns_query/dns_query.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net"
 	"strconv"
+	"sync"
 	"time"
 
 	"github.com/miekg/dns"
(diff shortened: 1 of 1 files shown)


danielnelson:
@Jaeyo Thanks again!

PR #4755, opened 2018-09-26
Add http_listener_v2 input plugin (supports multiple data formats)

### LINKED ISSUE #684: httpjson service plugin
I would essentially like an equivalent to Fluentd's [http input plugin](http://docs.fluentd.org/articles/in_http), which listens for incoming connections and expects to receive HTTP POST requests (JSON by default, but format is configurable). Telegraf's httpjson input plugin (like all of them, I gather) instead makes requests to retrieve data from a remote endpoint; I _think_ Telegraf uses the "service plugin" term to distinguish these two classes of data sources.

### LINKED ISSUE #3226: Add input that accepts all input data formats using HTTP
## Feature Request

The current `http_listener` input only accepts influx data format and attempts to behave like an InfluxDB server with respect to parsing, paths, headers, etc.

### Proposal:

Add a plugin alias for the current http_listener plugin: `influxdb_listener`, at some point in the future we remove the old http_listener.

Create a new input from scratch called `http_listener_ng`, or a good name if anyone has ideas.  The initial configuration could look something like:

```toml
[[inputs.http_listener_ng]]
  address = ":8186"

  paths = ["/telegraf"]
  methods = ["POST", "PUT"]

  read_timeout = "10s"
  write_timeout = "10s"

  tls_cert= "/etc/telegraf/server/cert.pem"
  tls_key = "/etc/telegraf/server/key.pem"
  tls_allowed_cacerts = ["/etc/telegraf/clientca.pem"]

  data_format = "influx"
```

### Current behavior:

No plugin exists that accepts the input data formats over HTTP.

### Desired behavior:

Plugin exists.

Allows pushing application metrics to Telegraf over http in any format.

This is commonly needed:
- https://github.com/influxdata/telegraf/pull/3216
- https://community.influxdata.com/t/json-http-plugin-es/2313

jul1u5:
Created `http_listener_v2` based on `http_listener`. Resolves #3226. Major differences:
- `data_format` config option support
- removed `bufferpool.go` from previous implementation

### Required for all PRs:

Resolves #684


plugins/inputs/http_listener_ng/http_listener_ng.go
```diff
@@ -0,0 +1,379 @@
...
+			return
+		}
+		handler(res, req)
+	} else {
+		handler(res, req)
+	}
+}
+
+func init() {
+	parser, _ := parsers.NewInfluxParser()
```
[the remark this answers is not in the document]
jul1u5: I have already added `SetParser(parser)` function. `NewInfluxParser()` is here for setting the default parser.

jul1u5:
@danielnelson is there anything else I should change?

plugins/inputs/http_listener_v2/http_listener_v2.go
```diff
@@ -0,0 +1,304 @@
...
+			return
+		}
+		handler(res, req)
+	} else {
+		handler(res, req)
+	}
+}
+
+func init() {
+	parser, _ := parsers.NewInfluxParser()
```
danielnelson: Remove, the SetParser function will be called with the correct parser.

plugins/inputs/http_listener_v2/http_listener_v2.go
```diff
@@ -0,0 +1,304 @@
...
+
+	inputs.Add("http_listener_v2", func() telegraf.Input {
+		return &HTTPListenerV2{
+			ServiceAddress: ":8186",
+			TimeFunc:       time.Now,
+			Parser:         parser,
```
danielnelson: Remove, the SetParser function will be called with the correct parser.

plugins/inputs/http_listener_v2/http_listener_v2.go
```diff
@@ -0,0 +1,304 @@
...
+		if err != nil {
+			log.Println("D! " + err.Error())
+			badRequest(res, err.Error())
+			return
+		}
+	}
+
+	// Add +1 for EOF
+	buf := make([]byte, h.MaxBodySize+1)
+	n, err := io.ReadFull(body, buf)
```
danielnelson: Use `ioutil.ReadAll` here, less error prone and will avoid allocating a very large MaxBodySize.  The handling of the data will be much easier too, I think there are 3 cases:
1. request too large: return 413
2. any other error: return 500
3. err is nil, continue on to parser and then return 204

plugins/inputs/http_listener_v2/http_listener_v2.go
```diff
@@ -0,0 +1,304 @@
...
+	if req.URL.Path == h.Path {
+		h.AuthenticateIfSet(h.serveWrite, res, req)
+	} else {
+		h.AuthenticateIfSet(http.NotFound, res, req)
+	}
+}
+
+func (h *HTTPListenerV2) serveWrite(res http.ResponseWriter, req *http.Request) {
+	// Check that the content length is not too large for us to handle.
+	if req.ContentLength > h.MaxBodySize {
```
danielnelson: This is a good check but the Content-Length header could potentially not match the actual body, and I'm not sure if the req.ContentLength is just set to the value from the header.  I think we need to use `http.MaxBytesReader` to wrap the r.Body.

plugins/inputs/http_listener_v2/http_listener_v2.go
```diff
@@ -0,0 +1,304 @@
...
+		methodNotAllowed(res)
+		return
+	}
+
+	// Handle gzip request bodies
+	body := req.Body
+	if req.Header.Get("Content-Encoding") == "gzip" {
+		var err error
+		body, err = gzip.NewReader(req.Body)
+		defer body.Close()
```
danielnelson: If an error occurred, body will probably be nil and calling Close might panic, move this under the err check.

plugins/inputs/http_listener_v2/http_listener_v2.go
```diff
@@ -0,0 +1,304 @@
...
+			res.WriteHeader(http.StatusNoContent)
+		}
+	} else {
+		log.Println("D! " + err.Error())
+		// problem reading the request body
+		internalServerError(res)
+	}
+}
+
+func (h *HTTPListenerV2) parse(b []byte) error {
```
danielnelson: I would bring the contents of this function directly into serveWrite, it will ease error handling.  The locking is not required since we are not changing the time precision around like is done in `http_listener`.

plugins/inputs/http_listener_v2/http_listener_v2.go
```diff
@@ -0,0 +1,304 @@
...
+	h.mu.Lock()
+	defer h.mu.Unlock()
+
+	metrics, err := h.Parse(b)
+	if err != nil {
+		return fmt.Errorf("unable to parse: %s", err.Error())
```
danielnelson: Don't add the extra error context here, this is done in the other plugin only to mirror InfluxDB's error messages.

plugins/inputs/http_listener_v2/http_listener_v2_test.go
```diff
@@ -0,0 +1,429 @@
...
+
+	testMsgNoNewline = "cpu_load_short,host=server01 value=12.0 1422568543702900257"
+
+	testMsgs = `cpu_load_short,host=server02 value=12.0 1422568543702900257
+cpu_load_short,host=server03 value=12.0 1422568543702900257
+cpu_load_short,host=server04 value=12.0 1422568543702900257
+cpu_load_short,host=server05 value=12.0 1422568543702900257
+cpu_load_short,host=server06 value=12.0 1422568543702900257
+`
+	testJSONMsgs = `
```
danielnelson: It's not often I get to say write less tests, but don't worry about testing different types of input data, these tests belong with the individual parsers.  The only cases to worry about are:
1. valid input
2. invalid (unparseable) input

Run all your tests against the influx parser.

plugins/inputs/http_listener_v2/http_listener_v2.go
jul1u5: I am not sure when it should return 500.
```go
body = http.MaxBytesReader(res, body, h.MaxBodySize)
bytes, err := ioutil.ReadAll(body)
```
`err` type is `errors.errorString` and as far as I tested `err.Error()` only returns `http: request body too large`. Should I compare `err.Error()` to `http: request body too large` if `err != nil`? It doesn't seem that robust to me to check errors by string comparison.

Right now I return:
- 413 if I see an error after `ioutil.ReadAll()`
- 400 if parsing fails
- 200 if it parses successfully

danielnelson: I think we are pretty much stuck with error string comparison here since the http package doesn't export a type to compare against.  Just reporting 413 on any error from ReadAll would be a reasonable thing to do in this case as well.

400 if parsing fails is the right one, only return 500 if something unexpected happens that shouldn't be possible.  If we ever see a 500 we will know there is a bug in our code.

jul1u5: I still couldn't figure out where to return 500. Can you check the code?

plugins/inputs/http_listener_v2/http_listener_v2.go
```diff
@@ -0,0 +1,281 @@
...
+
+	log.Printf("I! Started HTTP listener V2 service on %s\n", h.ServiceAddress)
+
+	return nil
+}
+
+// Stop cleans up all resources
+func (h *HTTPListenerV2) Stop() {
+	h.mu.Lock()
+	defer h.mu.Unlock()
```
danielnelson: These locks can go too

plugins/inputs/http_listener_v2/http_listener_v2.go
danielnelson: Looks like it won't be needed, there are no spots where a server error can occur now.

(Technically something else could happen during ReadAll but as we discussed above we can ignore this)

danielnelson:
Thanks @jul1u5!  I have also renamed the `http_listener` plugin to the more appropriate `influxdb_listener` in 2d8cda02df64a1d313088efbc9cfe19f61f7c510.

PR #4763, opened 2018-09-27
Add more detailed description for swap fields

dirkaholic:
Especially for the fields in & out it was pretty unclear what they contain as the usual values from `vmstat` returns the number of pages but here we get bytes. 

### Required for all PRs:

### DIFF
--- a/plugins/inputs/swap/README.md
+++ b/plugins/inputs/swap/README.md
@@ -14,14 +14,14 @@ For more information on what swap memory is, read [All about Linux swap space](h
 
 ### Metrics:
 
-- swap
-  - fields:
-    - free (int)
-    - total (int)
-    - used (int)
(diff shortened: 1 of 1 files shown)


danielnelson:
This is helpful information but it doesn't match the style in our [EXAMPLE_README.md](https://github.com/influxdata/telegraf/blob/master/plugins/inputs/EXAMPLE_README.md).

We don't have an established style for how to add a field description, but perhaps we could do it like this to reduce the impact:
```
    - in (int, bytes): data swapped in since last boot calculated from page number
```

If you think that will work I'll add it to the example readme.

dirkaholic:
@glinton @danielnelson Adjusted as suggested

danielnelson:
Thanks

PR #4766, opened 2018-09-27
Unify http_listener error response with influxdb

### LINKED ISSUE #4742: Errors from http_listener should match InfluxDB
### Relevant telegraf.conf:

```toml
[[http_listener]]
  service_address = ":9999"
```

### System info:

N/A

### Steps to reproduce:

1. Send malformed line protocol to http_listener
   - `cpu value=`
   - `cpu value=42\ncpu value=`

### Expected behavior:

Telegraf response should have `X-Influxdb-Error`, while the error strings do
not need to be identical they should contain the prefixes `unable to parse`
and `partial write`:

```
$ echo -e "cpu value=" | curl -v 'http://localhost:8086/write?db=telegraf' --data-binary @-
< HTTP/1.1 400 Bad Request
< Content-Type: application/json
< X-Influxdb-Error: unable to parse 'cpu value=': missing field value
<
{"error":"unable to parse 'cpu value=': missing field value"}
```

```
$ echo -e "cpu value=84\ncpu value=" | curl -v 'http://localhost:8086/write?db=telegraf' --data-binary @-
< HTTP/1.1 400 Bad Request
< Content-Type: application/json
< X-Influxdb-Error: partial write: unable to parse 'cpu value=': missing field value dropped=0
<
{"error":"partial write: unable to parse 'cpu value=': missing field value dropped=0"}
```

### Actual behavior:

```
$ echo -e "cpu value=" | curl -v 'http://localhost:9999/write?db=telegraf' --data-binary @-
< HTTP/1.1 400 Bad Request
< Content-Type: application/json
< X-Influxdb-Version: 1.0
<
{"error":"http: bad request"}
```

```
$ echo -e "cpu value=84\ncpu value=" | curl -v 'http://localhost:9999/write?db=telegraf' --data-binary @-
< HTTP/1.1 400 Bad Request
< Content-Type: application/json
<
{"error":"http: bad request"}
```

glinton:
Resolves #4742

### DIFF
--- a/plugins/inputs/http_listener/http_listener.go
+++ b/plugins/inputs/http_listener/http_listener.go
@@ -311,7 +311,11 @@ func (h *HTTPListener) serveWrite(res http.ResponseWriter, req *http.Request) {
 				return400 = true
 			}
 			if return400 {
-				badRequest(res, err.Error())
+				if err != nil {
+					badRequest(res, err.Error())
+				} else {
+					badRequest(res, "")
+				}
 			} else {
 				res.WriteHeader(http.StatusNoContent)
 			}
--- a/plugins/inputs/http_listener/http_listener.go
+++ b/plugins/inputs/http_listener/http_listener.go
@@ -5,6 +5,7 @@ import (
 	"compress/gzip"
 	"crypto/subtle"
 	"crypto/tls"
+	"fmt"
 	"io"
 	"log"
 	"net"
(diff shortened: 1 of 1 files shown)


danielnelson:
The error strings need to match up with what the influxdb output expects to trigger the cases that will cause an upstream Telegraf to retry.

https://github.com/influxdata/telegraf/blob/af0ef55c0281af67fe2db52d8c700fa8f636010e/plugins/outputs/influxdb/http.go#L320-L335

PR #4769, opened 2018-09-28
Resolved #4768: Added UUID to VMs

### LINKED ISSUE #4768: vSphere plugin: Add the BIOS UUID field to virtual machines
## Feature Request

Add the BIOS UUID field to virtual machines.

### Proposal:
vCenter stores three distinct identifiers for a VM: The moid, the instance UUID and the BIOS UUID. We currently emit the moid, since it can be used (together with the vCenter name) to uniquely identify a VM. 

We are proposing to expose the BIOS UUID as well. This useful since it can also be obtained from within a VM, thus making it possible to correlate metrics collected inside the VM to those collected from vCenter. 

### Current behavior:
The BIOS UUID is not emitted.

### Desired behavior:
Emit the BIOS UUID.

Users want to correlate metrics collected inside the VM using e.g. the Telegraf system collectors with metrics collected from vSphere.

prydin:
### Required for all PRs:

- [ X ] Associated README.md updated.
- [ X ] Has appropriate unit tests.

Resolved #4768: Added UUID to VMs. 
BIOS UUIDs are needed to correlate metrics from inside the VM to metrics gathered from vSphere. They were originally not included and users raised this as an issue. This PR addresses that issue and adds code to emit BIOS UUIDs when available.

### DIFF
--- a/plugins/inputs/vsphere/endpoint.go
+++ b/plugins/inputs/vsphere/endpoint.go
@@ -60,6 +60,7 @@ type objectMap map[string]objectRef
 
 type objectRef struct {
 	name      string
+	altID     string
 	ref       types.ManagedObjectReference
 	parentRef *types.ManagedObjectReference //Pointer because it must be nillable
 	guest     string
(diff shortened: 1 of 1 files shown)


danielnelson:
UUIDs can often be problematic because they, by design, have high cardinality.  We should think about when the BIOS UUID is unique and when it changes.  Is it different each time you start a VM?  Does it ever change during the VMs lifetime?

prydin:
They are the same for the lifetime of the VM. It doesn't change when a machine is restarted.

prydin:
Any indication as to if/when this will be merged? Will it make it into the next maintenance release?

danielnelson:
This is a little bit borderline for a patch release since it adds a new tag but I think it is minor enough and the plugin is new enough that we can add it.

randallt:
@prydin Can you describe the scenario a bit more on why a UUID is required to correlate?

prydin:
The BIOS UUID is the only identifier that's visible both from the vCenter and from inside the guest OS. On Linux, you can get it using the ```dmidecode``` command. I've seen use case where there's a need to correlate metrics from within the guest to what vCenter sees.

PR #1999, opened 2016-11-04
[wip] Add exec output plugin

ewbankkit:
### Required for all PRs:

- [ ] CHANGELOG.md updated (we recommend not updating this until the PR has been approved by a maintainer)
- [x] Sign [CLA](https://influxdata.com/community/cla/) (if not already signed)
- [x] README.md updated (if adding a new plugin)

Very much a Work In Progress but I have it working with an "external" plugin I wrote: https://github.com/ewbankkit/telegraf-signalfx-output.
Much of the code is copied from https://github.com/influxdata/telegraf/blob/master/plugins/inputs/exec/exec.go and much could/should be shared.

TODO: Tests 😄.

Addresses https://github.com/influxdata/telegraf/issues/544.
Relates to https://github.com/influxdata/telegraf/issues/1717.

Let me know if you want me continue working on this.

Issue #1717, opened 2016-09-07
Support External Plugins using Go 1.8 plugin feature


sparrc:
We should support building and loading of external plugins as dynamic libraries, using Go's new "plugin" feature, available in Go 1.8: https://tip.golang.org/pkg/plugin

For this we will likely want to maintain an external repository where we can aggregate "external" plugins which are not part of the telegraf statically-linked build.

Telegraf will be able to load all ".so" plugins from a directory, adding them to the regular plugin registry before the config file is loaded.

sparrc:
we may have experimental support for this in Telegraf 1.2, but since Go 1.8 will not be officially released at that point we will probably only see full support in Telegraf 1.3.

johnrengelman:
@sparrc I have to do some reading on Go 1.8's feature, but have you seen or evaluated this as an alternative? https://github.com/hashicorp/go-plugin

sparrc:
yes, that has been evaluated, IIRC that library doesn't give us anything we don't already have (plugins statically-compiled into a single binary)

johnrengelman:
I believe you can use it to load plugins as external binaries that communicate over network RPC. That's how Terraform used to work (as separate binaries per plugin) before they changed to bundling everything into a single binary. I believe external plugins are still loaded via this mechanism.

johnrengelman:
https://github.com/hashicorp/go-plugin/blob/master/client.go#L78

sparrc:
We could do external binaries, but there really is no need in that case for a plugin framework, IMHO, why not just communicate over plain grpc or text at that point?

the decision to go with dynamic libraries was specifically to avoid managing external processes.

johnrengelman:
Sounds good. Just wanted to offer up some options since plugins would be super beneficial for me. +1 to this plan. :)

daledude:
Intel snap telemetry project also uses external processes via grpc.

sebito91:
Would this supersede the exec plugin moving forward? Or would that stick around?

sparrc:
`exec` will definitely stick around, these .so files will only ever support plugins written in Go, for one thing

seuf:
Does this pr will allow dynamic loading of parser and serializer plugins ? (we are using custom metrics format in our company with custom serializer and parser)
Does all the actual plugins will be provided as external libraries ? Each one in a separate repository ? This will reduce the size of the telegraf binary.

danielnelson:
[...]

PR #4776, opened 2018-09-30
Add exec output plugin

### LINKED ISSUE #544: Add exec plugin as output
Would be nice to be able to execute a script that would take json metrics content and would send it to where it needs to go. (Most of these would help with various http/https api integrations)

Jaeyo:
### Required for all PRs:

Inherit #1999 
Close #544
Relates to #1717


plugins/outputs/exec/exec.go
```diff
@@ -0,0 +1,171 @@
...
+
+	cmd := exec.Command(splitCmd[0], splitCmd[1:]...)
+	cmd.Stdin = &buffer
+	var stderr bytes.Buffer
+	cmd.Stderr = &stderr
+
+	if err := internal.RunTimeout(cmd, e.Timeout.Duration); err != nil {
+		s := stderr.String()
+		if s != "" {
+			log.Printf("D! Command error: %s\n", s)
```
glinton: Instead of `%s`, use `%q` to prevent printing newlines

plugins/outputs/exec/README.md
```diff
@@ -0,0 +1,24 @@
...
+This plugin executes external programs with serialized metrics data passed on stdin, one metric per line.  
+The external program is free to do what it wants with the data.
+
+### Configuration:
+
+```toml
+[[outputs.exec]]
+  # Shell/commands array
+  # Full command line to executable with parameters, or a glob pattern to run all matching files.
+  commands = ["/usr/local/bin/telegraf-output --config /etc/telegraf-output.conf"]
```
danielnelson: We should only support a single command, having more than one in a single plugin will be problematic with the way metrics are buffered for outputs and it doesn't seem like something that needs done very frequently.

Also breaking from the way the exec input was done we should take the arguments for the single command as an array, to avoid issues with quoting and escaping which is problematic across platforms.

glinton:
@Jaeyo can you address that last bit, to just allow a single command instead of a list. thanks!

Jaeyo:
@glinton @danielnelson sorry for late! `support only a single command` applied.

lahsivjar:
@glinton @danielnelson : Are there any blockers for merging this PR?

internal/errchan/errchan.go
```diff
@@ -0,0 +1,35 @@
+package errchan
```
danielnelson: Not used anymore, I think we can remove this file.

plugins/outputs/exec/exec.go
```diff
@@ -0,0 +1,125 @@
...
+
+  ## Data format to output.
+  ## Each data format has it's own unique set of configuration options, read
+  ## more about them here:
+  ## https://github.com/influxdata/telegraf/blob/master/docs/DATA_FORMATS_OUTPUT.md
+  data_format = "influx"
+`
+
+type Exec struct {
+	Command string
```
danielnelson: I still think this should be an array:
```
command = ["echo", "howdy"]
```

This should be less problematic than using shellquote.Split, especially for those on Windows.

plugins/outputs/exec/exec.go
```diff
@@ -0,0 +1,125 @@
...
+
+	cmd := exec.Command(splitCmd[0], splitCmd[1:]...)
+	cmd.Stdin = &buffer
+	var stderr bytes.Buffer
+	cmd.Stderr = &stderr
+
+	if err := internal.RunTimeout(cmd, e.Timeout.Duration); err != nil {
+		s := stderr.String()
+		if s != "" {
+			log.Printf("D! Command error: %q\n", s)
```
danielnelson: Here we should return an generic error, potentially with the exit code.  Maybe `internal.ExitStatus(err)` will be useful.

plugins/outputs/exec/exec.go
```diff
@@ -0,0 +1,125 @@
...
+
+func (e *Exec) Connect() error {
+	return nil
+}
+
+func (e *Exec) Close() error {
+	return nil
+}
+
+func (e *Exec) ProcessCommand(command string, buffer bytes.Buffer) {
```
danielnelson: I don't see where this is called, can it be removed?

glinton:
@Jaeyo how likely is it you could address the feedback this week?

glinton:
We'll pick this up and push it through as there is internal desire for this to make it into the 1.12 release. Thanks for the great start Jaeyo!

Jaeyo:
@glinton sorry for checking late and didn't finishing it. i was busy for other stuff. just feel free to pick this PR.

danielnelson:
Continued on #6267

PR #4592, opened 2018-08-24
Processor/allofem - all the intern processors for tracing

maxunt:
subject to renaming and more tests
### Required for all PRs:

- [x] Signed [CLA](https://influxdata.com/community/cla/).
- [ ] Associated README.md updated.
- [ ] Has appropriate unit tests.

PR #4572, opened 2018-08-20
Fix instance and object name in performance counters with backslashes 

vlastahajek:
closes #4499 

### Overview
replaced complex regexp marching with simple index searching
Added unit test for counterPath parsing

### Required for all PRs:
- [X] Signed [CLA](https://influxdata.com/community/cla/).
- [X] Has appropriate unit tests.

PR #4778, opened 2018-10-01
[inputs/filecount] Add sizes to filecount

Samuel-BF:
# What this PR changes

It alters filecount input plugin, by adding configuration options :
- count_size : also output the total size of matched elements (subfiles/directories of 'directory'). This was [asked in influx community forums](https://community.influxdata.com/t/collect-folder-size-with-telegraf/3059).
- recursive_print : print gathered statistics (count/size) also for subdirectories
- recursive_print_size : minimum size of subdirectory to output it

It modifies the option "size" which was the minimum size in Bytes of files to take into account for counting : it is now a string with units (KiB, MiB...), parsed by [units](https://godoc.org/github.com/alecthomas/units) plugin (side note : this package was listed in dependencies but i haven't found where it were used). This breaks existing configurations which made use of this option, and is introduces by the last commit.

Internally, it uses the more specific 'addGauge' method where 'addFields' was used.

### Required for all PRs:

- [ X ] Associated README.md updated.
- [ X ] Has appropriate unit tests.

### DIFF
--- a/plugins/inputs/filecount/README.md
+++ b/plugins/inputs/filecount/README.md
@@ -10,6 +10,9 @@ Counts files in directories that match certain criteria.
   ## Directory to gather stats about.
   directory = "/var/cache/apt/archives"
 
+  ## Also compute total size of matched elements. Defaults to false.
+  count_size = false
+
   ## Only count files that match the name pattern. Defaults to "*".
   name = "*.deb"
 
@@ -34,6 +37,7 @@ Counts files in directories that match certain criteria.
 
 - filecount
     - count (int)
+    - size (int, in Bytes)
 
 ### Tags:
 
@@ -44,6 +48,6 @@ Counts files in directories that match certain criteria.
 
 ```
 $ telegraf --config /etc/telegraf/telegraf.conf --input-filter filecount --test
-> filecount,directory=/var/cache/apt,host=czernobog count=7i 1530034445000000000
-> filecount,directory=/tmp,host=czernobog count=17i 1530034445000000000
+> filecount,directory=/var/cache/apt,host=czernobog count=7i,size=7438336i 1530034445000000000
+> filecount,directory=/tmp,host=czernobog count=17i,size=28934786i 1530034445000000000
 ```
--- a/plugins/inputs/filecount/README.md
+++ b/plugins/inputs/filecount/README.md
@@ -22,10 +22,12 @@ Counts files in directories that match certain criteria.
   ## Only count regular files. Defaults to true.
   regular_only = true
 
-  ## Only count files that are at least this size in bytes. If size is
-  ## a negative number, only count files that are smaller than the
-  ## absolute value of size. Defaults to 0.
-  size = 0
+  ## Only count files that are at least this size. If size is a 
+  ## negative number, only count files that are smaller than the
+  ## absolute value of size. Defaults to "0B".
+  ## For available units, see :
+  ## https://godoc.org/github.com/alecthomas/units#pkg-constants
+  size = "0B"
 
   ## Only count files that have not been touched for at least this
   ## duration. If mtime is negative, only count files that have been
@@ -36,8 +38,8 @@ Counts files in directories that match certain criteria.
   recursive_print = false
 
   ## Only output directories whose sub elements weighs more than this
-  ## size in bytes. Defaults to 0.
-  recursive_print_size = 0
+  ## size. Defaults to "0B".
+  recursive_print_size = "0B"
 ```
 
 ### Measurements & Fields:
--- a/plugins/inputs/filecount/README.md
+++ b/plugins/inputs/filecount/README.md
@@ -5,7 +5,7 @@ Counts files in directories that match certain criteria.
 ### Configuration:
 
 ```toml
-# Count files in a directory
+# Count files in a directory and compute their size
 [[inputs.filecount]]
   ## Directory to gather stats about.
   directory = "/var/cache/apt/archives"
@@ -31,6 +31,13 @@ Counts files in directories that match certain criteria.
   ## duration. If mtime is negative, only count files that have been
   ## touched in this duration. Defaults to "0s".
   mtime = "0s"
+
+  ## Output stats for every subdirectory. Defaults to false.
+  recursive_print = false
+
+  ## Only output directories whose sub elements weighs more than this
+  ## size in bytes. Defaults to 0.
+  recursive_print_size = 0
 ```
 
 ### Measurements & Fields:
@@ -42,7 +49,7 @@ Counts files in directories that match certain criteria.
 ### Tags:
 
 - All measurements have the following tags:
-    - directory (the directory path, as specified in the config)
+    - directory (the directory path)
 
 ### Example Output:
 
--- a/plugins/inputs/filecount/filecount.go
+++ b/plugins/inputs/filecount/filecount.go
@@ -14,6 +14,9 @@ const sampleConfig = `
   ## Directory to gather stats about.
   directory = "/var/cache/apt/archives"
 
+  ## Also compute total size of matched elements. Defaults to false.
+  count_size = false
+
   ## Only count files that match the name pattern. Defaults to "*".
   name = "*.deb"
 
@@ -36,6 +39,7 @@ const sampleConfig = `
 
 type FileCount struct {
 	Directory   string
+	CountSize   bool
 	Name        string
 	Recursive   bool
 	RegularOnly bool
@@ -169,6 +173,7 @@ func (fc *FileCount) filter(file os.FileInfo) (bool, error) {
 
 func (fc *FileCount) Gather(acc telegraf.Accumulator) error {
 	numFiles := int64(0)
+	totalSize := int64(0)
 	countFn := func(f os.FileInfo) {
 		match, err := fc.filter(f)
 		if err != nil {
@@ -179,16 +184,22 @@ func (fc *FileCount) Gather(acc telegraf.Accumulator) error {
 			return
 		}
 		numFiles++
+		if fc.CountSize {
+			totalSize += f.Size()
+		}
 	}
 	err := count(fc.Directory, fc.Recursive, countFn)
 	if err != nil {
 		acc.AddError(err)
 	}
 
-	acc.AddFields("filecount",
-		map[string]interface{}{
-			"count": numFiles,
-		},
+	gauge := map[string]interface{}{
+		"count": numFiles,
+	}
+	if fc.CountSize {
+		gauge["size"] = totalSize
+	}
+	acc.AddGauge("filecount", gauge,
 		map[string]string{
 			"directory": fc.Directory,
 		})
@@ -199,6 +210,7 @@ func (fc *FileCount) Gather(acc telegraf.Accumulator) error {
 func NewFileCount() *FileCount {
 	return &FileCount{
 		Directory:   "",
+		CountSize:   false,
 		Name:        "*",
 		Recursive:   true,
 		RegularOnly: true,
--- a/plugins/inputs/filecount/filecount.go
+++ b/plugins/inputs/filecount/filecount.go
@@ -8,6 +8,7 @@ import (
 	"github.com/influxdata/telegraf"
 	"github.com/influxdata/telegraf/internal"
 	"github.com/influxdata/telegraf/plugins/inputs"
+	"github.com/alecthomas/units"
 )
 
 const sampleConfig = `
@@ -26,10 +27,12 @@ const sampleConfig = `
   ## Only count regular files. Defaults to true.
   regular_only = true
 
-  ## Only count files that are at least this size in bytes. If size is
-  ## a negative number, only count files that are smaller than the
-  ## absolute value of size. Defaults to 0.
-  size = 0
+  ## Only count files that are at least this size. If size is a 
+  ## negative number, only count files that are smaller than the
+  ## absolute value of size. Defaults to "0B".
+  ## For available units, see :
+  ## https://godoc.org/github.com/alecthomas/units#pkg-constants
+  size = "0B"
 
   ## Only count files that have not been touched for at least this
   ## duration. If mtime is negative, only count files that have been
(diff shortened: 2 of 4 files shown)


plugins/inputs/filecount/README.md
```diff
@@ -19,31 +22,41 @@ Counts files in directories that match certain criteria.
...
-  ## Only count files that are at least this size in bytes. If size is
-  ## a negative number, only count files that are smaller than the
-  ## absolute value of size. Defaults to 0.
-  size = 0
+  ## Only count files that are at least this size. If size is a 
+  ## negative number, only count files that are smaller than the
+  ## absolute value of size. Defaults to "0B".
+  ## For available units, see :
+  ## https://godoc.org/github.com/alecthomas/units#pkg-constants
+  size = "0B"
```
glinton: Will an old config without the `B` qualifier still work with this change?

plugins/inputs/filecount/filecount.go
```diff
@@ -149,6 +152,55 @@ func (fc *FileCount) initFileFilters() {
...
+			}
+			acc.AddGauge("filecount", gauge,
+				map[string]string{
+					"directory": basedir,
+				})
+		}
+	}
+
+	return numFiles, totalSize
+
```
glinton: Remove this empty line

plugins/inputs/filecount/README.md
Samuel-BF: No. That's a breaking change.

I see 2 options to avoid this : 

- either accept string OR int value for size, and treat string with units and ints as a byte value. But I haven't found a way to do such a thing (I tried an interface, but i'm a novice go programmer and i failed to implement it). A tip would be appreciated.
- either add a second option for handling strings with units (but what should be the name ? "size_with_units" ? and also add "recursive_print_size_with_units" ?)

Note that I'll first try to merge parts of the changes into #4752 (recursive print of matched directories).

Samuel-BF:
Commit 18373c7bf8450c85500b431689808a24804fed74 is a kind of merge from #4752 (using Directories option with globs rather than the recursive_print option).
To avoid to re-parse several times the folder structure, it was necessary to make some additions to globpath : 
- make globpath.root accessible
- changing the meaning of globpath.root (if "/etc/telegraf.d" is a directory, its root is "/etc/telegraf.d" and not "/etc" as it was the case before).
- add the ability to test a string (file path) against the glob.

Note that I didn't tested the behavior of "recursive" option, and i suspect that i doesn't work very well (what happens if i set directories = ["/my/path/**"] and recursive = false ? -> i'll test it)

WARNING : this version stills includes **breaking changes** for configuration files (allowing to use units such as "MB" for size field). any opinion about how to fix that point the right way is welcome.

sometimesfood:
> WARNING : this version stills includes **breaking changes** for configuration files (allowing to use units such as "MB" for size field). any opinion about how to fix that point the right way is welcome.

Is this really a "breaking change"? Doesn't your pull request simply add the unit parsing as a new feature instead of breaking old configs? (For the record: I haven't read your source code changes yet, so maybe I misunderstood your comments.)

Also, is there a specific reason why the `count_size` config option is needed? In my opinion the plugin should just always sum up file sizes, I just didn't get around to adding that feature. I think it would be better to omit the config option and always report sizes.

As stated in #4752, I strongly dislike the `recursive_print_size` option. It adds unnecessary clutter to the config in order to cover a relatively rare use case. If you omit the `count_size` option and remove the `recursive_print_size` option, you don't need any new config options at all.

Personally I think it would be a good idea to split this pull request into three separate pull requests:

- One pull request should add support for units to the `size` filter.
- The second pull request should add the `size` metric.
- The third pull request should add support for directory globbing and hence the `directories` option.

I hope this doesn't sound too dismissive, I like (most of) your changes, but IMHO they would benefit from a) cleaning up the config options a bit and b) being split up into separate pull requests.

plugins/inputs/filecount/README.md
```diff
@@ -5,11 +5,23 @@ Counts files in directories that match certain criteria.
...
+  ## Directories to gather stats about.
+  ## This accept standard unit glob matching rules, but with the addition of
+  ## ** as a "super asterisk". ie:
+  ##   /var/log/**    -> recursively find all directories in /var/log and count files in each directories
+  ##   /var/log/*/*   -> find all directories with a parent dir in /var/log and count files in each directories
+  ##   /var/log       -> count all files in /var/log and all of its subdirectories
+  directories = ["/var/cache/apt/archives"]
+
+  ## Also compute total size of matched elements. Defaults to false.
+  count_size = false
```
danielnelson: This operation should be left to query time, or listing the parent directory should be functionally the same.

plugins/inputs/filecount/README.md
```diff
@@ -19,31 +31,38 @@ Counts files in directories that match certain criteria.
...
-  ## Only count files that are at least this size in bytes. If size is
-  ## a negative number, only count files that are smaller than the
-  ## absolute value of size. Defaults to 0.
-  size = 0
+  ## Only count files that are at least this size. If size is a 
+  ## negative number, only count files that are smaller than the
+  ## absolute value of size. Defaults to "0B".
+  ## For available units, see :
+  ## https://godoc.org/github.com/alecthomas/units#pkg-constants
+  size = "0B"
```
danielnelson: I think you can support both types by creating a new type with an UnmarshalTOML method: https://github.com/influxdata/telegraf/blob/master/internal/internal.go#L33

plugins/inputs/filecount/README.md
```diff
@@ -19,31 +31,38 @@ Counts files in directories that match certain criteria.
...
 
   ## Only count files that have not been touched for at least this
   ## duration. If mtime is negative, only count files that have been
   ## touched in this duration. Defaults to "0s".
   mtime = "0s"
+
+  ## Only output directories whose sub elements weighs more than this
+  ## size. Defaults to "0B".
+  recursive_print_size = "0B"
```
danielnelson: This would be better done as a processor.  We kind of toyed around with this in a [threshold processor](https://github.com/influxdata/telegraf/pull/4592/files#diff-8b33a517a228dce95ce6cd88080d8447), but we need to finish spec'ing out the processor.

Samuel-BF:
Thanks @danielnelson for the review.

> This operation should be left to query time, or listing the parent directory should be functionally the same.

I don't understand that point. But, as proposed by @sometimesfood and unless you disagree with his point, I'll suppress this option (the plugin will always send back count & size of elements in directory)

> I think you can support both types by creating a new type with an UnmarshalTOML method: https://github.com/influxdata/telegraf/blob/master/internal/internal.go#L33

Thanks! I'll check it out !

> I would rather use a well tested library like https://github.com/karrick/godirwalk and avoid having custom file traversal, filesystems often have funny edge cases.

If you trust that library, I will try it. The problem here is that we need a recursive processing of directories to compute the total size of a directory (including the size of children elements, subdirectories children elements and so on ...) when recursive is set to "true" (which is the default). The "simple" option (as implemented in #4572) is inefficient when you have a '**' in your expression : 
- You do file traversal to resolve any path containing "**"
- And for each directory, you traverse all descendant directories to compute the total size (this make leaf directories to be processed *n* times where *n* is your directory depth).

I don't see how to efficiently use Walk functions when the function you need has to use the result of the Walk call on child elements. I'll try post-children callbacks from https://github.com/karrick/godirwalk .

I'll propose fixes, probably on friday. And reading the code for globpath.Match (in internals), it seems that some optimizations could be made. I'll propose a fix for that point first.

@sometimesfood  : the initial proposal was split in 3 commits as you proposed. But then I tried to mix with #4752 and it lost its separable logic.

Samuel-BF:
These two commits come back to the initial object of this PR : add 'size' field in outputs (and nothing more : no new option, no breaking changes). It integrates changes made by #4752 branch..

Commit 42c7db3 fixes incorrect recursive=false handling in previous versions and adds corresponding unit test.

If #4836 is merged, then there will be some tweakings to make to this code.

It still includes directories traversal "by hand" (I'll try gordiwalk)

An optimization could be made if, as I suspect, init() is called only at telegraf launch and "Gather" at every agent interval : the GlobPath.Compile operation should be done on init rather than on "Gather", to avoid recompiling expression.

Samuel-BF:
Changes necessitated by globpath optimization made. I consider this code to be ready.
Note that this PR adds size statistics and optimize filecount plugin (removing unnecessary re-parsings).

Samuel-BF:
While trying to [implement a recursive walk](https://github.com/karrick/godirwalk/pull/12) in [godirwalk](https://github.com/karrick/godirwalk/), @chadnetzer pointed a memory-efficient way of doing the work. I'll try to implement it in this PR.

Samuel-BF:
Thanks for the comment.

I tried to recode godirwalk with a [recursive handling of directories](https://github.com/karrick/godirwalk/pull/12) (but failed). @chadnetzer indicated a memory-efficient way to use the PostChildrenCallback to compute directory sizes. I'll try to implement it.

@danielnelson :
> We should return the first error up and out of Gather, since it indicates a bad configuration file.

Right. But how/where ? Is it possible to call "getGlobPaths" (which may be more consistently named initGlobPaths btw) in init function ?
The same logic apply to initFileFilters i think.

Samuel-BF:
Thanks for the comment.

I tried to recode godirwalk with a [recursive handling of directories](https://github.com/karrick/godirwalk/pull/12) (but failed). @chadnetzer indicated a memory-efficient way to use the PostChildrenCallback to compute directory sizes. I'll try to implement it.

@danielnelson :
> We should return the first error up and out of Gather, since it indicates a bad configuration file.

Right. But how/where ? Is it possible to call "getGlobPaths" (which may be more consistently named initGlobPaths btw) in init function ?
The same logic apply to initFileFilters i think.

Samuel-BF:
Thanks for the comment.

I tried to recode godirwalk with a [recursive handling of directories](https://github.com/karrick/godirwalk/pull/12) (but failed). @chadnetzer indicated a memory-efficient way to use the PostChildrenCallback to compute directory sizes. I'll try to implement it.

@danielnelson :
> We should return the first error up and out of Gather, since it indicates a bad configuration file.

Right. But how/where ? Is it possible to call "getGlobPaths" (which may be more consistently named initGlobPaths btw) in init function ?
The same logic apply to initFileFilters i think.

Samuel-BF:
Thanks for the comment.

I tried to recode godirwalk with a [recursive handling of directories](https://github.com/karrick/godirwalk/pull/12) (but failed). @chadnetzer indicated a memory-efficient way to use the PostChildrenCallback to compute directory sizes. I'll try to implement it.

@danielnelson :
> We should return the first error up and out of Gather, since it indicates a bad configuration file.

Right. But how/where ? Is it possible to call "getGlobPaths" (which may be more consistently named initGlobPaths btw) in init function ?
The same logic apply to initFileFilters i think.

Samuel-BF:
Thanks for the comment.

I tried to recode godirwalk with a [recursive handling of directories](https://github.com/karrick/godirwalk/pull/12) (but failed). @chadnetzer indicated a memory-efficient way to use the PostChildrenCallback to compute directory sizes. I'll try to implement it.

@danielnelson :
> We should return the first error up and out of Gather, since it indicates a bad configuration file.

Right. But how/where ? Is it possible to call "getGlobPaths" (which may be more consistently named initGlobPaths btw) in init function ?
The same logic apply to initFileFilters i think.

Samuel-BF:
Thanks for the comment.

I tried to recode godirwalk with a [recursive handling of directories](https://github.com/karrick/godirwalk/pull/12) (but failed). @chadnetzer indicated a memory-efficient way to use the PostChildrenCallback to compute directory sizes. I'll try to implement it.

@danielnelson :
> We should return the first error up and out of Gather, since it indicates a bad configuration file.

Right. But how/where ? Is it possible to call "getGlobPaths" (which may be more consistently named initGlobPaths btw) in init function ?
The same logic apply to initFileFilters i think.

Samuel-BF:
Thanks for the comment.

I tried to recode godirwalk with a [recursive handling of directories](https://github.com/karrick/godirwalk/pull/12) (but failed). @chadnetzer indicated a memory-efficient way to use the PostChildrenCallback to compute directory sizes. I'll try to implement it.

@danielnelson :
> We should return the first error up and out of Gather, since it indicates a bad configuration file.

Right. But how/where ? Is it possible to call "getGlobPaths" (which may be more consistently named initGlobPaths btw) in init function ?
The same logic apply to initFileFilters i think.

Samuel-BF:
I updated with godirwalk. Performance impact is negligible (simple tests using telegraf source code), but the code looks cleaner (and slightly longer).

danielnelson:
> > We should return the first error up and out of Gather, since it indicates
> > a bad configuration file.
>
> Right. But how/where ? Is it possible to call "getGlobPaths" (which may be
> more consistently named initGlobPaths btw) in init function ?  > The same
> logic apply to initFileFilters i think.

We really need an initialization function for plugins, which is something I added in my super secret future config improvement (https://github.com/danielnelson/tgconfig), but we don't have it just yet.  In the meantime it should work as it is if you chain the error up `if err != nil { return err; }`.

plugins/inputs/filecount/filecount.go
```diff
@@ -208,19 +228,30 @@ func (fc *FileCount) filter(file os.FileInfo) (bool, error) {
...
-	if err != nil {
-		return err
+	if fc.globPaths == nil {
+		fc.initGlobPaths(acc)
 	}
 
-	for _, dir := range dirs {
-		fc.count(acc, dir, fc.Recursive)
+	for _, glob := range fc.globPaths {
+		for _, dir := range onlyDirectories(glob.GetRoots()) {
```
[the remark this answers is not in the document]
Samuel-BF: > It seems like we should update glob.GetRoots() to always return a list of directories, this seems more in line with the function name.

I think it's better to rename the function. For filecount, our objects are directories, that's why we are discarding non-directory elements. But other plugins may have the need to recurse by themselves (and not use the globpath.Walk function) to find files and not (only) directories. In the case of patterns such as "a/b/c/d**", they have to examine both files and directories returned by GetRoots.

> Let's add documentation to the new functions in globpath as well.

Done.

Reviewing globpath code, there is inconsistency introduced by my code in Match function regarding the 'Stat' on matched files : 

- when !hasMeta : stat-ing the 'path' prevents returning a non-existent file or a file we cannot stat (is there another error case ?)
- when !HasSuperMeta : also stat-ing results of filepath.Glob . Is it necessary ? What does it prevent ?
- when hasSuperMeta : not stat-ing results of filepath.Walk (and not filtering errors ... which may also be necessary)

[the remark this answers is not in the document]
danielnelson: > I think it's better to rename the function.

Okay, I guess I'm not sure what a good name is here though.  If you think there is a more appropriate one then use that otherwise we can leave as is.

> regarding the 'Stat' on matched files

It may be there there for non-existing files, but I don't think it is needed.  It doesn't make sense to worry too much about errors opening the files as these need to be handled when actually opening them, also it probably just makes it more difficult to report errors later.  I think it is best to remove the stat calls and just return all items.

Samuel-BF:
Well, re-reading `Match()` code with a fresh eye, `os.Stat()` are necessary to return FileInfo map to caller. So i kept them. But actually only 'filestat' input plugins uses the FileInfo field, all other plugins using globpath.Match() discards this information (file, tail and logparser). More optimization could be made to remove the unneeded stats here. I'll prepare a separate PR (also featuring godirwalk in globpath.go).

The new commit are just about Gopkg.lock / .toml.

PR #4781, opened 2018-10-01
Add new config for csv column explict type conversion

### LINKED ISSUE #4775: config data types of fields in CSV input plugin
__Proposal:__  add a config to allow the users to set data types of fields, in CSV input plugin

__Current behavior:__ The CSV input detects the field type by trying int, float, bool and string in turn.
So for the following input file:

    lily, 2
    lucy, 2.1

at the first line, the plugin takes the value `2` as int. And at the second line, it takes the value `2.1` as float.

The different types will make `field type conflict` error in influxdb.

__Desired behavior:__ So can you add a config to allow the users to set data types of fields?

It may be like:

    csv_column_names = ["name", "height"]
    csv_column_types = ["string", "float"]  # add this config

__Use case:__ As said above, the different types will make `field type conflict` error in influxdb.

Requests may be closed if we're not actively planning to work on them.

rudbast:
### Required for all PRs:

Resolves #4775.

### DIFF
--- a/plugins/parsers/csv/parser.go
+++ b/plugins/parsers/csv/parser.go
@@ -21,6 +21,7 @@ type Parser struct {
 	Comment           string
 	TrimSpace         bool
 	ColumnNames       []string
+	ColumnTypes       []string
 	TagColumns        []string
 	MeasurementColumn string
 	TimestampColumn   string
@@ -146,6 +147,35 @@ outer:
 					tags[tagName] = value
 					continue outer
 				}
+			}
+
+			// In case if defined column names & types count don't match.
+			if i < len(p.ColumnTypes) {
+				var val interface{}
+				var err error
+
+				switch p.ColumnTypes[i] {
+				case "int":
+					val, err = strconv.ParseInt(value, 10, 64)
+					if err != nil {
+						return nil, fmt.Errorf("column type: parse int error %s", err)
+					}
+				case "float":
+					val, err = strconv.ParseFloat(value, 64)
+					if err != nil {
+						return nil, fmt.Errorf("column type: parse float error %s", err)
+					}
+				case "bool":
+					val, err = strconv.ParseBool(value)
+					if err != nil {
+						return nil, fmt.Errorf("column type: parse bool error %s", err)
+					}
+				default:
+					val = value
+				}
+
+				recordFields[fieldName] = val
+				continue
 			}
 
 			// attempt type conversions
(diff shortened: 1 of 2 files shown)


plugins/parsers/csv/parser.go
```diff
@@ -148,6 +149,35 @@ outer:
 				}
 			}
 
+			// In case if defined column names & types count don't match.
+			if i < len(p.ColumnTypes) {
```
glinton: Seems like throwing an error if they don't match before here would be a good idea.

rudbast: i've added the throw error statement on `newCSVParser`, seems like it fits there better ?

FAE-MICHAEL:
Hi @danielnelson / @glinton / @rudbast 

I'm trying to use 'csv_column_types' with only datatypes ["float", "string" ...]. 
But this is generating the error:

_**Error parsing /etc/telegraph/telegraph.conf, line 48: field corresponding to `csv_column_types 'is not defined in` * file.File'**_

Has anyone ever experienced this?

Example of use:

[[inputs.file]]
   files = ["/home/STATRACK.csv"]
   data_format = "csv"
   csv_delimiter = "|"
   csv_header_row_count = 0
   csv_column_names = ["REPTIMEOBTAINED", "REPSTATSCACHESIZE", "VMMAXMEM", "VMTOTALMEM", "LASTXACTID"]
   csv_column_types = ["string", "float", "float", "float", "string"]

FAE-MICHAEL:
Hi @rudbast

I'm using the version of telegraf 1.8 1 downloaded from site influxdb. is there any more recent compilated version? Where can I download it? If it does not work, I'll open another problem here ...

rudbast:
@FAE-MICHAEL i think this PR is planned for 1.9 and as of now it hasn't released yet. if you still planned to use this feature (field), i suggest you build your own from current master branch https://github.com/influxdata/telegraf#from-source.

FAE-MICHAEL:
thank you very much for the answers ... I can point the telegraf to the PR already resolved and that will be released in version 1.9 or will I have to wait?

danielnelson:
1.9 is tentatively scheduled by end of November, you can using a [nightly build](https://github.com/influxdata/telegraf#nightly-builds) before then.

FAE-MICHAEL:
Hi @danielnelson and @rudbast

I made the update and the start of the telegraf occurred successfully, but in its log now appears:

Error in plugin [inputs.file]: column type: parse float error strconv.ParseFloat: parsing "18": invalid syntax

Did I just install telegraf-nightly.x86_64.rpm, do I need to update or change some settings?

Thank you

FAE-MICHAEL:
I did a DROP MEASUREMENT and it was created with the correct FIELDTYPE. Now the problem is solved. Thank you so much for your help @danielnelson and @rudbast .

PR #4784, opened 2018-10-01
suggest other configs when qos != 0

rgitzel:
Hi,

The MQTT Consumer plugin config 'qos' (quality-of-service) configuration didn't work as I expected, it wasn't clear that other settings were needed.  This would have helped.

Thanks!

rodney

### Required for all PRs:

- [na] Has appropriate unit tests.


danielnelson:
I think we should also add why this setting is needed, what do you think about this comment based on the mqtt output:
```
  ## QoS policy for messages
  ##   0 = at most once
  ##   1 = at least once
  ##   2 = exactly once
  ##
  ## When using a QoS of 1 or 2, you should enable persistent_session to allow
  ## resuming unacknowledged messages.
```

rgitzel:
@danielnelson Nice idea!

I will do that up tomorrow.

rgitzel:
@danielnelson done.

PR #4793, opened 2018-10-03
Add a Dockerfile that matches influxdata-docker to build images from source

jsternberg:
The images in influxdata-docker are meant to be built by downloading
official releases. Sometimes, it is useful to build directly from source
when you need an unofficial release. These images are meant to be used
then using multi-stage builds so that it can build from source and then
copy the results to images that match the official counterpart.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,18 @@
+FROM golang:1.11.0 as builder
+ENV DEP_VERSION 0.5.0
+RUN curl -fsSL -o /usr/local/bin/dep https://github.com/golang/dep/releases/download/v${DEP_VERSION}/dep-linux-amd64 && chmod +x /usr/local/bin/dep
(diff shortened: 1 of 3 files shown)


danielnelson:
Let's move the files under the scripts directory, remove the `dev.docker` file, and add the build command to the Makefile.

jsternberg:
It's probably easier if it's in the base directory because then it's easier to find and build as if it were a nightly.

Or do you mean just move the docker folder itself under the scripts folder so you have `scripts/docker` and keep the Dockerfile at the root?

danielnelson:
I'm trying to avoid adding files to the top of the repo, so I prefer to have them all moved under scripts: `scripts/stretch.docker`, `scripts/alpine.docker`, `scripts/entrypoint.sh`.

PR #4798, opened 2018-10-03
Allow connecting to prometheus via socket

glinton:
Based on, and closes #2974

### DIFF
--- a/plugins/inputs/prometheus/README.md
+++ b/plugins/inputs/prometheus/README.md
@@ -28,6 +28,8 @@ in Prometheus format.
   # insecure_skip_verify = false
 ```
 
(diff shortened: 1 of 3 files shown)


danielnelson:
I think there is a race condition if there are both http and unix socket connections, since the dialer is being replaced.  It seems we would need one transport per unix socket plus one shared for http.

One way we could deal with this is by allowing only a single socket:
```
[[inputs.prometheus]]
  urls = ["http://localhost/metrics"]
  unix_socket = "/var/run/prometheus.sock"
```

This is basically how curl added support for unix sockets.  If you need multiple sockets you would need to define multiple inputs.  The path would be defined in the urls, and this could support query parameters or even TLS, though I'm unsure how useful those are.  You could still have multiple urls, but this would only be useful if they have differing paths because all calls would go to the socket.

If we did this a plugin would only need a single transport and I think it would keep things simple.

Issue #4699, opened 2018-09-15
Datadog Plugin Spikes CPU 


smicha:
When I enabled the datadog plugin  the cpus on our servers (windows ec2 ) went up from 6% to 35% . Is this normal or am I missing something .  I am running a binary that I built  with the below command 

 GOOS=windows GOARCH=amd64 make telegraf and Installed it .

danielnelson:
@smicha I see that this plugin does a few operations that are somewhat expensive, in particular it calls `Metric.Fields()` and `Metric.Tags()` instead of `FieldList()` and `TagList()`.  We should update the output to use the more efficient methods.

smicha:
Thanks . I will submit a PR for it tomorrow

danielnelson:
Closed in #4803

PR #4803, opened 2018-10-03
#4699 Improve how the datadog plugin uses Metric.Fields and Metric.Tags

kevinconaway:
### Required for all PRs:

- [Y ] Signed [CLA](https://influxdata.com/community/cla/).
- [ NA] Associated README.md updated.
- [ NA] Has appropriate unit tests.

@danielnelson This PR addresses the feedback in #4699.  I replaced the use of `Metric.Fields` with `Metric.FieldList` and reduced how often `Metric.Tags` is called (it didn't look like using `TagList` made sense)

Since the changes were internal, I left the existing tests as is.

### DIFF
--- a/plugins/outputs/datadog/datadog.go
+++ b/plugins/outputs/datadog/datadog.go
@@ -7,7 +7,6 @@ import (
 	"log"
 	"net/http"
 	"net/url"
-	"sort"
 	"strings"
 
 	"github.com/influxdata/telegraf"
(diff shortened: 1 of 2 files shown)


plugins/outputs/datadog/datadog.go
```diff
@@ -76,6 +76,10 @@ func (d *Datadog) Write(metrics []telegraf.Metric) error {
 
 	for _, m := range metrics {
 		if dogMs, err := buildMetrics(m); err == nil {
+			tagMap := m.Tags()
+			metricTags := buildTags(tagMap)
```
danielnelson: This is much better but if you want to take it even further you could use TagList() to build the tags list.  TagList is always sorted so you can avoid the sort as well.

danielnelson:
Tests should pass if you add the changes made by `make fmt`.

kevinconaway:
@danielnelson Thanks for the feedback!

I made the changes that you suggested for `buildTags` to use `Metric.TagsList()`.  I also removed the call to `Metric.Tags()` in favor of just looking up the `host` tag directly on the metric.  There was no need to a new map just to find that one tag.

PR #4807, opened 2018-10-04
Implement HTTP Output Compression

### LINKED ISSUE #4472: HTTP Output Compression
## Feature Request
Enable HTTP Compression for outgoing HTTP

### Proposal:
Enable HTTP Compression for outgoing HTTP. Really only makes sense for POST with batched data.
Could be set similar to influxdb OUTPUT config (though 'identity' doesn't make sense in straight HTTP context). Might even be able to copy/paste code to implement from influxdb plugin.
```
## HTTP Content-Encoding for write request body, can be set to "gzip" to
## compress body or "identity" to apply no encoding.
  # content_encoding = "identity"
```
### Current behavior:
Uncompressed body is the only option.
### Desired behavior:
1. Enable Compression (gzip, deflate, br, etc) through telegraf.conf
2. Set correct `Content-Encoding` HTTP header to match chosen compression.
3. Compress payload
We use Telegraf as sidecar for docker containers on machines that have terrible bandwidth. Every byte saved helps.

mihaitodor:
Fixes #4472.

TODO:
- [x] Should there be some refactoring done to avoid duplicating `compressWithGzip` here, since it's already present [in influxdb](https://github.com/influxdata/telegraf/blob/6bf5643351c86aa09ae4819816ec783b8a4e1c2f/plugins/outputs/influxdb/http.go#L384) and [in influxdb_v2](https://github.com/influxdata/telegraf/blob/6bf5643351c86aa09ae4819816ec783b8a4e1c2f/plugins/outputs/influxdb_v2/http.go#L261)?
- [x] Would you be interested in caching gzip writers similar to [here](https://github.com/Nitro/lazyraster/blob/318cb9c229e728d756299e34bb06386d88082c08/http.go#L463-L490) for better performance?
- [x] Do you really want to support "deflate" compression besides "gzip"?
- [x] If "br" stands for "brotli", it's not supported natively yet: https://github.com/google/brotli/issues/182 Do you want to bring in a 3rd party package for this if support for it is required?

### Required for all PRs:

### DIFF
--- a/plugins/outputs/http/README.md
+++ b/plugins/outputs/http/README.md
@@ -44,4 +44,8 @@ data formats.  For data_formats that support batching, metrics are sent in batch
   # [outputs.http.headers]
   #   # Should be set manually to "application/json" for json data_format
   #   Content-Type = "text/plain; charset=utf-8"
+
+  ## HTTP Content-Encoding for write request body, can be set to "gzip" to
+  ## compress body or "identity" to apply no encoding.
+  # content_encoding = "identity"
 ```
--- a/plugins/outputs/http/http.go
+++ b/plugins/outputs/http/http.go
@@ -2,8 +2,10 @@ package http
 
 import (
 	"bytes"
+	"compress/gzip"
 	"context"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"net/http"
 	"strings"
@@ -55,6 +57,10 @@ var sampleConfig = `
   # [outputs.http.headers]
   #   # Should be set manually to "application/json" for json data_format
   #   Content-Type = "text/plain; charset=utf-8"
+
+  ## HTTP Content-Encoding for write request body, can be set to "gzip" to
+  ## compress body or "identity" to apply no encoding.
+  # content_encoding = "identity"
 `
 
 const (
@@ -64,16 +70,17 @@ const (
 )
 
 type HTTP struct {
-	URL          string            `toml:"url"`
-	Timeout      internal.Duration `toml:"timeout"`
-	Method       string            `toml:"method"`
-	Username     string            `toml:"username"`
-	Password     string            `toml:"password"`
-	Headers      map[string]string `toml:"headers"`
-	ClientID     string            `toml:"client_id"`
-	ClientSecret string            `toml:"client_secret"`
-	TokenURL     string            `toml:"token_url"`
-	Scopes       []string          `toml:"scopes"`
+	URL             string            `toml:"url"`
+	Timeout         internal.Duration `toml:"timeout"`
+	Method          string            `toml:"method"`
+	Username        string            `toml:"username"`
+	Password        string            `toml:"password"`
+	Headers         map[string]string `toml:"headers"`
+	ClientID        string            `toml:"client_id"`
+	ClientSecret    string            `toml:"client_secret"`
+	TokenURL        string            `toml:"token_url"`
+	Scopes          []string          `toml:"scopes"`
+	ContentEncoding string            `toml:"content_encoding"`
 	tls.ClientConfig
 
 	client     *http.Client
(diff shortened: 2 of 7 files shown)


glinton:
1. I think that's a great idea, though obviously a bit more involved
1. Also likely worth implementing some form of writer caching
1. Personally, I'm fine with just gzip
1. I don't think it's worth adding, but I think having some generic compressor/writer pool that can be extended to add deflate/br sometime in the future wouldn't be a bad idea.

I'm curious to know, and would wait to find out, what @danielnelson has to say on the matter though.

danielnelson:
We could move the shared code under `internal`, it sounds like it would be a good idea.

I would rather avoid a pool for gzip.Writer.  If it is a performance bottleneck, it may make sense to reuse the Writer between calls to Write, within a plugin instance, but I would want to see some benchmarks showing the gzip.Writer as a significant portion of write time first.

On brotli, I don't think it has enough adoption to warrant, I'm just hearing of it for the first time now.

mihaitodor:
@glinton @danielnelson Thanks for answering my questions! I have implemented `1` in a53e591.

Studying this `CompressWithGzip` function in more detail, I think its design is not safe. `io.Copy(gzipWriter, data)` is executed in the background, which means that the internal goroutine will be still running when `CompressWithGzip` returns, so the caller will never know if `io.Copy(gzipWriter, data)` returns an error eventually. Do you have any insight into why it was designed this way? I'd probably consider returning an error channel to the caller instead and let it wait for it in a select statement. WDYT?

Regarding `2`, I'd rather leave the pool for `gzip.Writer` for a future PR, if you're happy with the current performance. I'm not too keen to spend time on coming up with benchmarks for this right now.

Regarding `3` and `4`, I asked because #4472 asks explicitly for "gzip, deflate, br, etc", so wasn't sure what your expectations are. Glad to hear that you're OK with having only `gzip`. I think it should be enough for most situations.

plugins/outputs/http/http.go
```diff
@@ -162,7 +168,17 @@ func (h *HTTP) Write(metrics []telegraf.Metric) error {
 }
 
 func (h *HTTP) write(reqBody []byte) error {
-	req, err := http.NewRequest(h.Method, h.URL, bytes.NewBuffer(reqBody))
+	var reqBodyBuffer io.Reader = bytes.NewBuffer(reqBody)
+	var err error
+
+	if h.ContentEncoding == "gzip" && h.Method == http.MethodPost {
```
danielnelson: Why only for POST method?

plugins/outputs/http/README.md
```diff
@@ -44,4 +44,8 @@ data formats.  For data_formats that support batching, metrics are sent in batch
   # [outputs.http.headers]
   #   # Should be set manually to "application/json" for json data_format
   #   Content-Type = "text/plain; charset=utf-8"
+
+  ## HTTP Content-Encoding for write request body, can be set to "gzip" to
+  ## compress body or empty string to not apply any encoding.
+  # content_encoding = ""
```
danielnelson: This is minor but can you document like the influxdb option with the `identity` setting as default, no code needs to change:
```
  ## HTTP Content-Encoding for write request body, can be set to "gzip" to
  ## compress body or "identity" to apply no encoding.
  # content_encoding = "identity"
```

plugins/outputs/http/http.go
mihaitodor: That's what was requested in #4472. It will add a bit of overhead to requests which don't have a body, but I wouldn't normally consider such small optimisations. Maybe @ madisonleavo can offer additional insight.

danielnelson: All of the requests currently have a body in this plugin, and it is hard to imagine a new data format that wouldn't produce it, perhaps a null data format?  At any rate we currently do POST and PUT and I think they should both use the encoding if it is requested.

mihaitodor:
@danielnelson

> The setting of the error in CompressWithGzip is definitely wrong, but I don't think an error channel would be the right approach here as we really want any reader errors passed though the Pipe, this way the user of the pipe (http package in this case) gets the error. This would give consistent behavior with other errors in the Reader.

Indeed, you are right! I got carried away thinking that it would be enough to return an error in the `write` method of the `HTTP` output plugin.

How about calling [`CloseWithError`](https://golang.org/pkg/io/#PipeWriter.CloseWithError) on the `PipeWriter`? I can simply pass into it the error returned by `io.Copy`, because:
> subsequent reads from the read half of the pipe will return no bytes and the error err, or EOF if err is nil.

> Also, fixing the error handling is optional for this PR, since it wouldn't be a step backwards.
I agree. Will open a new issue.

danielnelson:
That sounds perfect

plugins/outputs/http/http.go
mihaitodor: Cool! Done in 26ee478.

mihaitodor:
@danielnelson OK, I believe I have addressed all the required changes in this PR. Please let me know if I missed anything.

danielnelson:
Did you do the CloseWithError idea, I'm not seeing it?

mihaitodor:
Oops, communication issue. I thought you preferred to see a new PR for it. Added it in 11cdf77.

danielnelson:
Thanks!

PR #4809, opened 2018-10-04
Add ServiceNow Metrics serializer

JefMuller:
### Required for all PRs:

### DIFF
--- a/plugins/serializers/nowmetric/nowmetric.go
+++ b/plugins/serializers/nowmetric/nowmetric.go
@@ -1,0 +1,132 @@
+package nowmetric
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"time"
+
+	"github.com/influxdata/telegraf"
+)
+
+type serializer struct {
+	TimestampUnits time.Duration
+}
+
+const METRICFMT string = "{ \"metric_type\": \"%s\", \"resource\": \"%s\", \"node\": \"%s\", \"value\": %.2f, \"timestamp\": %d, \"ci2metric_id\": { \"node\": \"%s\" }, \"source\": \"Telegraf\" }"
+
+// field 1, resourcename, hostname, field 2, timestamp, hostname
+//
+func NewSerializer(timestampUnits time.Duration) (*serializer, error) {
+	s := &serializer{
+		TimestampUnits: truncateDuration(timestampUnits),
+	}
+	return s, nil
+}
+
+func (s *serializer) Serialize(metric telegraf.Metric) ([]byte, error) {
+	serialized := s.createObject(metric)
+	if serialized == nil {
+		return []byte{}, nil
+	}
+	return serialized, nil
+}
+
+func (s *serializer) SerializeBatch(metrics []telegraf.Metric) ([]byte, error) {
+
+	objects := make([]byte, 0)
+	for _, metric := range metrics {
+		m := s.createObject(metric)
+		objects = append(objects, m...)
+	}
+
+	if objects == nil {
+		return []byte{}, nil
+	}
+	return objects, nil
+
+}
+
+func (s *serializer) createObject(metric telegraf.Metric) []byte {
+	var payload string
+	var resourcename string
+	var hostname string
+	var utime int64
+
+	// Process Tags to extract node & resource name info
+	for _, tag := range metric.TagList() {
+		key := tag.Key
+		value := tag.Value
+
+		if key == "" || value == "" {
+			continue
+		}
+
+		if key == "objectname" {
+			resourcename = value
+		}
+		if key == "host" {
+			hostname = value
+		}
+	}
+
+	// Format timestamp to UNIX epoch
+	utime = (metric.Time().UnixNano() / int64(s.TimestampUnits)) * 1000
+
+	nbdatapoint := 0
+	// Loop of fields value pair and build datapoint for each of them
+	for _, field := range metric.FieldList() {
+		// field.Key, field.Value
+		metrictype := field.Key
+		metricvalue := field.Value
+
+		if !verifyValue(metricvalue) {
+			// Ignore String
+			continue
+		}
+
+		if metrictype == "" || metricvalue == "" {
+			continue
+		}
+		// Params : metrictype, resourcename, hostname, metricvalue, utime, hostname
+		if nbdatapoint >= 1 {
+			payload = payload + ",\n"
+		}
+		payload = payload + fmt.Sprintf(METRICFMT, metrictype, resourcename, hostname, metricvalue, utime, hostname)
+		nbdatapoint++
+	}
+
+	raw := make([]byte, 0)
+	raw = append(raw, "[ "...)
+	raw = append(raw, payload...)
+	raw = append(raw, " ]\n"...)
+
+	out := new(bytes.Buffer)
+	json.HTMLEscape(out, raw)
+
+	return out.Bytes()
+}
+
+func truncateDuration(units time.Duration) time.Duration {
+	// Default precision is 1s
+	if units <= 0 {
+		return time.Second
+	}
+
+	// Search for the power of ten less than the duration
+	d := time.Nanosecond
+	for {
+		if d*10 > units {
+			return d
+		}
+		d = d * 10
+	}
+}
+
+func verifyValue(v interface{}) bool {
+	switch v.(type) {
+	case string:
+		return false
+	}
+	return true
+}
--- a/plugins/serializers/nowmetric/nowmetric_test.go
+++ b/plugins/serializers/nowmetric/nowmetric_test.go
@@ -1,0 +1,195 @@
+package nowmetric
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/metric"
+)
+
+func MustMetric(v telegraf.Metric, err error) telegraf.Metric {
+	if err != nil {
+		panic(err)
+	}
+	return v
+}
+
+func TestSerializeMetricFloat(t *testing.T) {
+	now := time.Now()
+	tags := map[string]string{
+		"cpu": "cpu0",
+	}
+	fields := map[string]interface{}{
+		"usage_idle": float64(91.5),
+	}
+	m, err := metric.New("cpu", tags, fields, now)
+	assert.NoError(t, err)
+
+	s, _ := NewSerializer(0)
+	var buf []byte
+	buf, err = s.Serialize(m)
+	assert.NoError(t, err)
+	expS := []byte(fmt.Sprintf(`{"fields":{"usage_idle":91.5},"name":"cpu","tags":{"cpu":"cpu0"},"timestamp":%d}`, now.Unix()) + "\n")
+	assert.Equal(t, string(expS), string(buf))
+}
+
+func TestSerialize_TimestampUnits(t *testing.T) {
+	tests := []struct {
+		name           string
+		timestampUnits time.Duration
+		expected       string
+	}{
+		{
+			name:           "default of 1s",
+			timestampUnits: 0,
+			expected:       `{"fields":{"value":42},"name":"cpu","tags":{},"timestamp":1525478795}`,
+		},
+		{
+			name:           "1ns",
+			timestampUnits: 1 * time.Nanosecond,
+			expected:       `{"fields":{"value":42},"name":"cpu","tags":{},"timestamp":1525478795123456789}`,
+		},
+		{
+			name:           "1ms",
+			timestampUnits: 1 * time.Millisecond,
+			expected:       `{"fields":{"value":42},"name":"cpu","tags":{},"timestamp":1525478795123}`,
+		},
+		{
+			name:           "10ms",
+			timestampUnits: 10 * time.Millisecond,
+			expected:       `{"fields":{"value":42},"name":"cpu","tags":{},"timestamp":152547879512}`,
+		},
+		{
+			name:           "15ms is reduced to 10ms",
+			timestampUnits: 15 * time.Millisecond,
+			expected:       `{"fields":{"value":42},"name":"cpu","tags":{},"timestamp":152547879512}`,
+		},
+		{
+			name:           "65ms is reduced to 10ms",
+			timestampUnits: 65 * time.Millisecond,
+			expected:       `{"fields":{"value":42},"name":"cpu","tags":{},"timestamp":152547879512}`,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			m := MustMetric(
+				metric.New(
+					"cpu",
+					map[string]string{},
+					map[string]interface{}{
+						"value": 42.0,
+					},
+					time.Unix(1525478795, 123456789),
+				),
+			)
+			s, _ := NewSerializer(tt.timestampUnits)
+			actual, err := s.Serialize(m)
+			require.NoError(t, err)
+			require.Equal(t, tt.expected+"\n", string(actual))
+		})
+	}
+}
+
+func TestSerializeMetricInt(t *testing.T) {
+	now := time.Now()
+	tags := map[string]string{
+		"cpu": "cpu0",
+	}
+	fields := map[string]interface{}{
+		"usage_idle": int64(90),
+	}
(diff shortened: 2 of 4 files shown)


plugins/serializers/nowmetric/nowmetric.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	// Format timestamp to UNIX epoch
+	utime = (metric.Time().UnixNano() / int64(s.TimestampUnits)) * 1000
+
+	nbdatapoint := 0
+	// Loop of fields value pair and build datapoint for each of them
+	for _, field := range metric.FieldList() {
+		// field.Key, field.Value
+		metrictype := field.Key
```
glinton: Just check and use `field.Key` and `field.Value`

plugins/serializers/registry.go
```diff
@@ -1,124 +1,131 @@
-package serializers
```
glinton: Why does this file have so many "changes" when it appears to be the same file?

nicolas17: LF vs CRLF newlines?

JefMuller:
What does this mean?

glinton:
It's because that user left a comment. Likely about the need to run `make fmt` and commit the changes so the tests pass. Your editor might be changing the newline type from a simple line feed (LF) to carriage return/line feed (CRLF).

glinton:
Thanks for addressing that so fast @JefMuller. As you may have noticed, the tests are failing elsewhere now - can be seen about midway through [this page](https://circleci.com/gh/influxdata/telegraf/20562?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link).

JefMuller:
Hi,
Wondering if there is chance to get this one integrated in the core?
In case, yes, what is your plan from the timeline perspective?
Thx

glinton:
It might make it into 1.9 or 1.10, I still need to talk to @danielnelson about when he thinks this serializer will fit in.

JefMuller:
Cool
Thx for the feedback

Sent from my iPhone

On 17 Oct 2018, at 19:15, Greg <notifications@github.com<mailto:notifications@github.com>> wrote:

[External Email]

It might make it into 1.9 or 1.10, I still need to talk to @danielnelson<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_danielnelson&d=DwMCaQ&c=Zok6nrOF6Fe0JtVEqKh3FEeUbToa1PtNBZf6G01cvEQ&r=rXT_Chyhukb3Yc3G1h2Zzr-92bGKF8RC0ExztjpwQ61LFBha_LgtWdqC1brsVUMm&m=lDEQK0SIfsMwIO7rmxxi6PdqCujfdPgeK2MUOY9hqqo&s=GDygd6iekU02YEg-GHwRfwK9X82hZ7B2S_B8JIP2wiI&e=> about when he thinks this serializer will fit in.

JefMuller:
Hi Greg,

Did you get a chance to talk to the person in charge about this contribution?
I’m asking since I would like to promote this internally at ServiceNow as soon as I have a timeline from you.

Regards
Jef

JefMuller:
Hi Greg,

Wondering if you could provide me an estimation when will this be integrated in the core?

Thanks
Regards
Jef

glinton:
Sorry, we got pretty bogged down getting the 1.9 release candidate out. I'm not certain on the timeline for this plugin, @danielnelson may have a better idea though.

JefMuller:
Hi Greg,
Thanks for feedback.
I understand.
Just wanted to have an idea about your plan.

Regards
Jef

danielnelson:
@russorat Can you provide a target milestone for this issue?

danielnelson:
@JefMuller It seems like since this plugin would almost always be used with the http output, instead of having a servicenow data_format it should instead be a standalone output.  This will make it easier to configure the plugin as you won't need to set headers/method/data_format/path, the output would know how to set these parameters:
```
[[outputs.servicenow]]
  ## URL is the address to send metrics to
  url = "http://ec2-52-57-154-43.eu-central-1.compute.amazonaws.com:9082"

  ## Write Timeout
  # timeout = "5s"

  ## Credentials
  # username = 'evt.integration'
  # password = 'ServiceN0w!'

  ## Optional TLS Config
```

plugins/serializers/nowmetric/nowmetric.go
```diff
@@ -0,0 +1,122 @@
...
+	"time"
+
+	"github.com/influxdata/telegraf"
+)
+
+type serializer struct {
+	TimestampUnits time.Duration
+}
+
+const METRICFMT string = "{ \"metric_type\": \"%s\", \"resource\": \"%s\", \"node\": \"%s\", \"value\": %v, \"timestamp\": %d, \"ci2metric_id\": { \"node\": \"%s\" }, \"source\": \"Telegraf\" }"
```
danielnelson: This is just a quick lookthrough, not a full review, but this needs to be changed to use the json.Marshal function on a struct or array of structs, as this won't do suitable escaping.

danielnelson:
The splunkmetrics serializer was added to allow using splunk format with multiple output plugins, but in the case of servicenow my understanding is that users will always want to use the http output.  Having it as a serializer requires the users to configure the http output in ways that we could easily handle for them with a servicenow output.

JefMuller:
Hi Daniel,

We have basically the same requirements.
Write the output to a file is useful if the customer doesn’t want to use the push approach and rather collect the information using the pull connector.

Can I assume that the serializer once using json.Marshal is a valid option for integrating this piece in the core project?

Thanks
Regards
Jef

danielnelson:
Could you send me some links on the pull connector?

JefMuller:
Sure.

https://docs.servicenow.com/bundle/london-it-operations-management/page/product/event-management/task/configure-metric-connectors.html


Sent from my iPhone

On 30 Nov 2018, at 18:58, Daniel Nelson <notifications@github.com<mailto:notifications@github.com>> wrote:

[External Email]


Could you send me some links on the pull connector?

danielnelson:
It looks like all of the existing pull connectors read from other services, is there an existing one that reads from file or would it require a custom connector?

JefMuller:
Hi Daniel,

The metric connector logic is implement in a MID script file.
Some of them are reading from a file generated by command executed against the 3rd party. (The SCOM one for ex)
For Madrid release, we also have a plan for an XML “generic” connector.

But we don’t have a connector for Telegraf yet.
Might be my next exercise. Don’t know yet.

Besides that, I have applied your recommendation “json.Marshal function + struct” and tested against London release as well as updated the documentation due to slight change on our side.

Let me know if there is a chance to see this in your standard build or not.

Thanks
Regards
Jef

plugins/serializers/nowmetric/nowmetric.go
```diff
@@ -0,0 +1,152 @@
+package nowmetric
```
danielnelson: Should this parser be called "servicenow"?  When I google for nowmetric I don't get any relevant hits, can you explain a little more?

plugins/serializers/nowmetric/nowmetric.go
```diff
@@ -0,0 +1,152 @@
...
+		oimetric.CiMapping = cimapping
+
+		metricJson, err = json.Marshal(oimetric)
+		payload = append(payload, metricJson...)
+
+		nbdatapoint++
+	}
+
+	raw := make([]byte, 0)
+	raw = append(raw, "[ "...)
```
danielnelson: The array handling should also be part of the Marshal, build an array of metric objects and then you will just need a single call to Marshal and no need to append newlines or other text.

plugins/serializers/nowmetric/nowmetric.go
```diff
@@ -0,0 +1,152 @@
...
+			oimetric.Resource = tag.Value
+		}
+
+		if tag.Key == "host" {
+			oimetric.Node = tag.Value
+		}
+	}
+
+	// Format timestamp to UNIX epoch
+	oimetric.Timestamp = (metric.Time().UnixNano() / int64(s.TimestampUnits)) * 1000
```
danielnelson: Why * 1000?  When I read these docs it says the timestamp should always be in milliseconds.  Should we remove the option and have `oimetric.Timestamp = metric.Time().UnixNano() / int64(time.Milliseconds)`


https://docs.servicenow.com/bundle/london-it-operations-management/page/product/event-management/reference/mid-POST-metrics.html

plugins/serializers/nowmetric/nowmetric.go
```diff
@@ -0,0 +1,152 @@
...
+
+	nbdatapoint := 0
+	// Loop of fields value pair and build datapoint for each of them
+	for _, field := range metric.FieldList() {
+		if !verifyValue(field.Value) {
+			// Ignore String
+			continue
+		}
+
+		if field.Key == "" || field.Value == "" {
```
danielnelson: The value will never be an empty string because the previous check skips strings.

plugins/serializers/nowmetric/nowmetric.go
```diff
@@ -0,0 +1,152 @@
...
+}
+*/
+
+type OIMetric struct {
+	Metric    string                 `json:"metric_type"`
+	Resource  string                 `json:"resource"`
+	Node      string                 `json:"node"`
+	Value     interface{}            `json:"value"`
+	Timestamp int64                  `json:"timestamp"`
+	CiMapping map[string]interface{} `json:"ci2metric_id"`
```
danielnelson: Make this a map[string]string if possible, documentation seems to indicate that this can be a string only.

plugins/serializers/nowmetric/nowmetric.go
```diff
@@ -0,0 +1,152 @@
...
+		}
+
+		if nbdatapoint >= 1 {
+			payload = append(payload, ",\n"...)
+		}
+		oimetric.Metric = field.Key
+		oimetric.Value = field.Value
+
+		cimapping := map[string]interface{}{}
+		cimapping["node"] = oimetric.Node
```
danielnelson: This may be unset if the metric doesn't have a "host" tag.

JefMuller:
Hi Daniel,

Thanks for feedback.
I will do my best to update this next week.

Regards
Jef

plugins/serializers/nowmetric/nowmetric.go
JefMuller: Hi, 
Asked myself the same question...
But ServiceNow usually means "Incident Management" integration which is a complete different story.
nowmetric is the shortest i found, that is branding compliant. (snow is not allowed)

plugins/serializers/nowmetric/nowmetric.go
JefMuller: Ok got it. Thanks for the tips

plugins/serializers/nowmetric/nowmetric.go
JefMuller: Ok

plugins/serializers/nowmetric/nowmetric.go
JefMuller: fixed

plugins/serializers/nowmetric/nowmetric.go
```diff
@@ -0,0 +1,152 @@
...
+	Timestamp int64             `json:"timestamp"`
+	CiMapping map[string]string `json:"ci2metric_id"`
+	Source    string            `json:"source"`
+}
+
+type OIMetrics []OIMetric
+
+func NewSerializer(timestampUnits time.Duration) (*serializer, error) {
+	s := &serializer{
+		TimestampUnits: truncateDuration(timestampUnits),
```
danielnelson: This option is no longer used, so we should remove it from the serializer and also remove the truncateDuration function.

plugins/serializers/nowmetric/nowmetric.go
```diff
@@ -0,0 +1,152 @@
...
+		if oimetric.Node != "" {
+			cimapping := map[string]string{}
+			cimapping["node"] = oimetric.Node
+			oimetric.CiMapping = cimapping
+		}
+
+		allmetrics = append(allmetrics, oimetric)
+	}
+
+	metricsJson, err := json.Marshal(allmetrics)
```
danielnelson: Return both of these values, function signature should be `func (s *serializer) createObject(metric telegraf.Metric, err error) ([]byte, error)`, then you can use it in the parent functions to test for errors and return the error if it occurs.

plugins/serializers/nowmetric/nowmetric_test.go
```diff
@@ -0,0 +1,184 @@
...
+			},
+			time.Unix(0, 0),
+		),
+	)
+
+	metrics := []telegraf.Metric{m, m}
+	s, _ := NewSerializer(0)
+	buf, err := s.SerializeBatch(metrics)
+	require.NoError(t, err)
+	require.Equal(t, []byte(`[{"metric_type":"value","resource":"","node":"","value":42,"timestamp":0,"ci2metric_id":null,"source":"Telegraf"}][{"metric_type":"value","resource":"","node":"","value":42,"timestamp":0,"ci2metric_id":null,"source":"Telegraf"}]`), buf)
```
danielnelson: This does not seem correct to me, shouldn't the metrics all be in one array: `[{}, {}]` instead of `[{}][{}]`?

danielnelson: I think the SerializeBatch function is not producing the output the servicenow API requires.  The current output with multiple arrays is not even valid JSON: ` [{}][{}]`.  I would expect to output multiple objects within a single array like: `[{}, {}]`.

JefMuller: Ok i get your point & I need your advice for this one..
I reproduced what i have seen for other serializer, but this test is not correct.  

SerializeBatch is expecting a metric array as parameter...
As this function just loop over the array and call the marshal function for each metric, it results in this string.

What would you recommend?

a. Should i merge the metric array in one metric object before marshalling?
b. Or Modify the SerializeBatch test, to ingest only one metric object but with multiples metric.FieldList() ?

I have tested the actual code on both Windows and Linux platform and I'm able to ingest all metrics on ServiceNow side successfully. So i assume the issue is "limited" to the test function.

JefMuller: For which use case or situation is the SerializeBatch function called?

JefMuller: I mean I'm testing my agent with Windows Performance Counters plugin setup and against a standard AWS Linux box then i compare the metric stream in ServiceNow and the file output. 
Is SerializeBatch invoked in these test case? I have the feeling that for both case, it's the Serialize function which is invoked. How can SerializeBatch be invoked?

danielnelson: It is up to the output plugin to choose between Serialize and SerializeBatch, of interest though is that the HTTP output uses SerializeBatch while the file output uses Serialize.

JefMuller: Ok. This makes it much more clear…. Thanks.
The MID listener didn't returned any wrong code with this payload, but some metrics got probably ignored.
I have modified the SerializeBatch function and I'm currently running a bunch of tests. They look positive so far.

Continuous-integration tests are failing outside of my code. Is this an issue on my side?

JefMuller: Except continuous integration tests, the functional one are ok now.
Let me know how to fix others.
Thanks for your help

JefMuller:
Thanks Daniel for your help and support on this.
I assume you are ok if I mention you in my blog regarding this implementation.

Can you eventually share your release plan? What is your target date for 1.10?

danielnelson:
Thanks @JefMuller, yeah feel free to mention on the blog, we keep the latest planned release date in the milestone here on Github: https://github.com/influxdata/telegraf/milestones, but keep in mind they do sometimes change.

hwaastad:
@JefMuller, thanks for adding this serializer.
However, would you consider option to add all tags to ci2metric_id?

JefMuller:
@hwaastad : What would be the benefit? Why the ask?

hwaastad:
@JefMuller, thanks for your reply.
I was thinking since the ci2metric_id, by spec, is a list of key-value pairs, It would enhance the indentification process in servicenow.
an example could be snmp collection on interfaces, which has metric_name: ifhcoutoctets, node: ip/mac, resource(objectname): interface name anc ci2metric_id: node: MAC, interface-index: int
It would make it easier to add different event rules filters.
Just thinking flexibility :-)

PR #4815, opened 2018-10-05
update the comment

kelwang:
### Required for all PRs:

update the bucket comment

### DIFF
--- a/plugins/outputs/influxdb_v2/influxdb.go
+++ b/plugins/outputs/influxdb_v2/influxdb.go
@@ -35,7 +35,7 @@ var sampleConfig = `
   ## Organization is the name of the organization you wish to write to; must exist.
   organization = ""
 
-  ## Bucket to the name fo the bucketwrite into; must exist.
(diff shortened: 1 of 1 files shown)


plugins/outputs/influxdb_v2/influxdb.go
```diff
@@ -35,7 +35,7 @@ var sampleConfig = `
   ## Organization is the name of the organization you wish to write to; must exist.
   organization = ""
 
-  ## Bucket to the name fo the bucketwrite into; must exist.
+  ## Destination bucket to write.
```
danielnelson: Thanks, but I feel that the working is still not quite right, since we aren't writing the bucket we are writing into it.  Perhaps:
```
## Destination bucket to write to.
```
or
```
## Bucket to write into.
```

@stevebang Any thoughts?

stevebang: I think that "Destination bucket to write into" is the clearest.

PR #4820, opened 2018-10-07
Add cookie support to input and output HTTP plugins

### LINKED ISSUE #4813: Plugin HTTP with cookies
## Feature Request

Some REST API environment uses authentication through GET request and generate a cookie to use for the next GET command.

### Proposal:

When the server return a cookie in the HTTP reply, store the cookie in a jar and use it for the next requests in the plugin.

### Current behavior:

No cookies are stored, so all the sessions are considered as a new session, so all the GET requests return a 401 error becaused session must authenticate first.
All the login event done with a GET request returns a 200 ok and create a login session in the remote server.

### Desired behavior:

When a cookie is received in a reply, store the cookie in a jar to reuse it on the next request.
Use the cookie received to identify the session used to get information from REST GET requests.
In the list of urls, the first entry is a login url, then the different urls to check, then the logout url.
empty the cookie jar after running the plugin.

We can gather information from REST API when Authorization header is used, but not when the API uses cookies.  Since the authorisation part is not standardised, having this option would enlarge the number of APIs supported.

mihaitodor:
I believe this fixes #4813.

Note: It would be great to test these changes against a real REST API environment which uses this authentication mechanism. Not sure how to get access to one (cc @Touchedegris).

Questions:
- [ ] The test code is mostly duplicated for both plugins, so I'm considering to hide parts of it under `testutil`. WDYT?
- [ ] Is there any place in the docs where this new feature needs to be mentioned?

### Required for all PRs:

### DIFF
--- a/plugins/inputs/http/http.go
+++ b/plugins/inputs/http/http.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"net/http/cookiejar"
 	"strings"
 	"sync"
 	"time"
@@ -14,6 +15,7 @@ import (
 	"github.com/influxdata/telegraf/internal/tls"
 	"github.com/influxdata/telegraf/plugins/inputs"
 	"github.com/influxdata/telegraf/plugins/parsers"
+	"golang.org/x/net/publicsuffix"
 )
 
 type HTTP struct {
@@ -29,7 +31,8 @@ type HTTP struct {
 
 	Timeout internal.Duration
 
-	client *http.Client
+	cookieJar *cookiejar.Jar
+	client    *http.Client
 
 	// The parser will automatically be set by Telegraf core code because
 	// this plugin implements the ParserInput interface (i.e. the SetParser method)
@@ -86,6 +89,14 @@ func (h *HTTP) Gather(acc telegraf.Accumulator) error {
 		return errors.New("Parser is not set")
 	}
 
+	var err error
+	if h.cookieJar == nil {
+		h.cookieJar, err = cookiejar.New(&cookiejar.Options{PublicSuffixList: publicsuffix.List})
+		if err != nil {
+			return fmt.Errorf("failed to create cookie jar: %s", err)
+		}
+	}
+
 	if h.client == nil {
 		tlsCfg, err := h.ClientConfig.TLSConfig()
 		if err != nil {
@@ -97,6 +108,7 @@ func (h *HTTP) Gather(acc telegraf.Accumulator) error {
 				Proxy:           http.ProxyFromEnvironment,
 			},
 			Timeout: h.Timeout.Duration,
+			Jar:     h.cookieJar,
 		}
 	}
 
--- a/plugins/inputs/http/http_test.go
+++ b/plugins/inputs/http/http_test.go
@@ -1,15 +1,24 @@
 package http_test
 
 import (
+	"fmt"
 	"net/http"
 	"net/http/httptest"
+	"net/url"
 	"testing"
+	"time"
 
 	plugin "github.com/influxdata/telegraf/plugins/inputs/http"
 	"github.com/influxdata/telegraf/plugins/parsers"
 	"github.com/influxdata/telegraf/testutil"
 	"github.com/stretchr/testify/require"
 )
+
+const simpleJSON = `
+{
+    "a": 1.2
+}
+`
 
 func TestHTTPwithJSONFormat(t *testing.T) {
 	fakeServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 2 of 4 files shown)


Touchedegris:
If you have access to JIRA, it looks like they have REST API authentication with cookies:
https://developer.atlassian.com/cloud/jira/platform/jira-rest-api-cookie-based-authentication/

I am using it to replace SNMP to monitor LAN switches that support REST.  It is the first time I would try new code so I would be glad to help if I get guidance on how to install it and test it.  This week is a bad timing for testing but next week should be good.  As soon as I can test it I will let you know the results :)

mihaitodor:
I'm not sure how I could use Jira to perform such tests, but I'd be grateful if you can have a go at trying to test this. It should work just fine, based on the unit tests that I added.

Touchedegris:
It is the first time I would have to test code.  Could you please let me know what to do to install the test code?  Just want to make sure I do it properly.

mihaitodor:
OK, are you familiar with installing the Go compiler and building Go code and would you be able to clone this Git repository on your machine? If not, I could provide a pre-built executable for you. Just let me know on what operating system it needs to run.

Touchedegris:
I am not familiar with go compiling.  I can test it with either CentOS or Ubuntu.

Thanks!

danielnelson:
Here are some links to the automated builds of this branch on CircleCI:
- [telegraf_1.9.0%7E22d36e86-0_amd64.deb](https://20889-33258973-gh.circle-artifacts.com/0/build/linux/amd64/telegraf_1.9.0%7E22d36e86-0_amd64.deb)
- [telegraf-1.9.0%7E22d36e86-0.x86_64.rpm](https://20889-33258973-gh.circle-artifacts.com/0/build/linux/amd64/telegraf-1.9.0%7E22d36e86-0.x86_64.rpm)

mihaitodor:
@Touchedegris Any luck testing those builds? Please let us know if you need any help.

Touchedegris:
Hi,
I finally got time to test it today.  Here is what I found:

1- The URLs are sent in a random order.  To be sucessful, the "login" URL sould be launched first, then the URLs to get the information, then the "logout" URL.  By doing so, we make sure the session is active before we send any REST GET requests, and closes the session at the end.  I dont know if it would be appropriate to create an option in the plugin or simply ask the plugin to launch the requests in order of appearance in the config file.

2- Even when launched in the proper order, I could not find the Cookie header in the GET requests.  I can see the server returns a SET-COOKIE header.  I was using the telegraf --test option as my influxdb server is not available at the time.  I don't know if it makes a difference though...

I used Ubuntu 18.04.1LTS with the .deb package.  I did not test with CentOS.

mihaitodor:
OK, thanks for testing @Touchedegris! Sorry to hear it didn't work as expected.

For 1, I was wandering about that... I'm not sure how the URLs are picked up from the config, but I can have a look. Would you mind sharing the (anonymised) config file and the command line that you used to start telegraf?

For 2, I'll try to build a dummy server and check if I can replicate the issue.

Touchedegris:
I was looking at the wireshark trace and I also noticed that Telegraf sends the 3 requests all together, so obviously it cannot use a cookie as all the requests are sent before I receive the cookie...  In this case, I think that the login URL option makes more sense.

Touchedegris:
[telegraf_http.txt](https://github.com/influxdata/telegraf/files/2503630/telegraf_http.txt)

mihaitodor:
Thanks for sharing that config! In the mean time, I see the code [here](https://github.com/influxdata/telegraf/blob/55f3645e8cbb353afacfae00dec7e74663387f4d/plugins/inputs/http/http.go#L106) simply fetches all the URLs in parallel as you said, which is definitely not what you want in this case.

@danielnelson @glinton I'm thinking to add an optional parameter for this plugin called `login_url` and make an initial call to [`gatherURL`](https://github.com/influxdata/telegraf/blob/55f3645e8cbb353afacfae00dec7e74663387f4d/plugins/inputs/http/http.go#L131) using this URL if it's set and `h.client == nil` in the [`Gather`](https://github.com/influxdata/telegraf/blob/55f3645e8cbb353afacfae00dec7e74663387f4d/plugins/inputs/http/http.go#L84) method. Regarding logging out, I'm not sure when that should happen. I see that there are `ServiceInput` plugins and I'm wondering if maybe there should be an `http_service` input plugin for this scenario. WDYT?

danielnelson:
Sounds like a fair bit of extra complexity to support this type of workflow, and in the JIRA case this flow is already deprecated in favor of basic auth.  What will happen if we get logged out, we would need to detect the login error and restart the process.  @Touchedegris any chance the switches you are interfacing with support another authentication method?

danielnelson:
Yeah, let's close, right now I think doing the full lifecycle will be too much additional complexity to justify when considering how much use this feature would get.

Thanks for looking into this, it was still very helpful so we could make the determination.

stegenfeldt:
Oh, no. I am too late. Cookie based authentication is used by a bunch of IBM API's as well. I am currently working on getting IBM Spectrum Control data into InfluxDB and it is using cookie based auth. 

Ah, well. I guess you won't re-open this due to the closing comments, but I want to post this anyway, for awareness.

danielnelson:
@stegenfeldt I'm not entirely closed off to reopening this, does the API you are using have a separate login URL and return 401 or 403 if the cookie is not set?

stegenfeldt:
@danielnelson Yes, it does. You get a 401 directing you to the login form (URL-encoded POST, j_username, j_password fields) if you lack the session and auth cookies. If you have a session cookie, but don't carry an authentication cookie you will get a 403 if I've figured it out correctly. 
I have solved it for now, using inputs.exec and a shell-script (with curl) and a python-script to parse and convert output to InfluxDB Line Protocol since I am completely new to Go, but the authentication flow against an IBM Spectrum Control server is basically:

- POST to https://servername:9569/srm/j_security_check
- save cookies to cookiejar
- pass cookiejar with every subsequent request against API

A bit of googling tells me that j_security_check is quite common in JSP/Tomcat applications. 

I gave Go a try, but the very limited time I had in the sprint really didn't give me a lot of time to indulge in a new language, unfortunately. Never really grasped how to manage cookiejars in Go. :)

danielnelson:
Sounds like [this](https://docs.oracle.com/cd/E19226-01/820-7627/bncbq/) flavor of form based authentication.  The issue is that this method is not formalized, likely it would work for Java Servlets but not much else.  For example, the way the credentials are sent wouldn't work with the original case with Jira, because it requires sending the username/password as JSON.

Maybe it would make more sense to write a proxy server that handles the form based authentication and provide a standard way of authenticating to Telegraf?

bschreze:
Hi all,
sorry for hitting the same thing again. I am working with an storage virtualization solution called Falconstor. The restAPI unfortunately only accepts cookie authentication.
It is exactly like @stegenfeldt wrote in hist post.
1. Post to /ipstor/auth/login
2. save cookie (session id)
3. pass cookie (session id) for the next requests
4. pass cookie (session id) to logout
I had a discussion with the vendor but they are not interested to change things to bearer token.

**SAMPLE USAGE**
```
POST /ipstor/auth/login
data:{
"username":"root",
"password":"mypassword",
"server":"10.3.2.67"
}
```

**SAMPLE RESPONSE**
```
Cookie Data: session_id=7c266058-f3b0-40d5-9656-6dddd1469be7
{"rc":0,"id":"7c266058-f3b0-40d5-9656-6dddd1469be7","type":"root"}
```

Hope someone could help. It seems that the only alternative is to login to each Falconstor server and run script and to "manually" tag the performance and usage statistics before importing to influxdb.

Regards
BSchreze

stegenfeldt:
I ended up writing my own exec probe for my checks, in Go. 
It is VERY specific to IBM Spectrum Control, but could probably be reworked to something that is more flexible. In the case of IBM Spectrum Control, I have to do requests in series with data from the previous request, so I am not sure HOW flexible it can get. For example; to get hypervisor disk sizes, I first have to authenticate to the form for the cookiejar, then I have to probe for hypervisors, then I have to loop through each hypervisor-ID to get connected disks, then probe each disk for they metrics. It also requires a lot of data-washing since it gladly mixes numerics and text. 
Like, "empty"->0..99->"full" and stuff like that. 

My point is, I could probably convince my customer to share the code for the probe we made, but I would need someone more fluid in golang to make this into a proper plugin. Perhaps a dedicated IBM Spectrum Control plugin in this case. 

Relating back to the topic at hand. 
What I did to get the cookies working is adding a cookiejar to the http object (struct?). Like so:
```go
srmRootURL := "https://apiserver:9569/srm"
loginURL := srmRootURL + "/j_security_check"
restAPIBase := srmRootURL + "/REST/api/v1/"

httpClient := newHTTPClient()
jar, _ := cookiejar.New(nil) //TODO: Move into srmclient.go
httpClient.Jar = jar

authCookies, statusCode, err := loginToSRM(loginURL, username, password, httpClient)
```
`newHTTPClient()` just creates a simple `http.client` and sets stuff like timeouts and disables certificate verification in this particular case. 

The `loginToSRM` is also quite simple, nothing magical here. It's... in need of improvement, but the budget wasn't there for me to do that yet. 
```go
func loginToSRM(loginURL string, username string, password string, client *http.Client) (cookies []*http.Cookie, statusCode int, err error) {
	postValues := url.Values{}
	postValues.Add("j_username", username)
	postValues.Add("j_password", password)

	postResponse, err := client.Post(loginURL, "application/x-www-form-urlencoded", strings.NewReader(postValues.Encode()))

	if err != nil {
		// error handling here, i guess
		// fmt.Println("Error on http.Post")
	}

	defer postResponse.Body.Close()
	cookies = postResponse.Cookies()
	statusCode = postResponse.StatusCode

	return cookies, statusCode, err
}
```
This is the first code I've ever written i Go, so... pardon my non-existent error handling. :) 

Anyway, the key parts in this is to make sure all calls to the same system uses the same `*http.client` that has the proper cookiejar. And that should be doable in the HTTP-plugins if there was an "auth" section in the config for the probe. 
It wouldn't help me in this case due to the silly IBM APIs, and maybe you've already done this. But in case this would help anyone, well there you go.

Cheers!

And thanks for all the work on the telegraf agent, i quite like it.

danielnelson:
@bschreze Looks like the Falconstor example is similar to the Jira method.  For now, I recommend writing a script for the exec plugin that handles this, perhaps using curl to make the HTTP requests.

@stegenfeldt Thanks for sharing, and glad it wasn't too difficult to add.

PR #4842, opened 2018-10-10
Report timeout as error in dns_query plugin

kostya-sh:
### Required for all PRs:


danielnelson:
I'm not sure about returning timeout as an error, I think we had it as an error but it was removed on purpose.  The argument was made that since this plugins main job is to measure query time, a timeout shouldn't be logged, only reported in the metrics.

Similar changes were made to the net_response and http_response plugins.

kostya-sh:
I see, I will update the test then to test this as it fails at the moment. It looks like these tests do not run as part of CI build.

PR #4849, opened 2018-10-11
Don't pass empty measurement name to AddFields

### LINKED ISSUE #4816: logparser broken on 1.8.x
### Relevant telegraf.conf:
```toml
[[inputs.logparser]]
  files = ["/var/log/nginx/cdn_access.log"]
  from_beginning = true
  name_override = "nginx_access_log"
  [inputs.logparser.grok]
    patterns = ["%{COMBINED_LOG_FORMAT}"]
```

### System info:
telegraf 1.8.x (tested on .0 and .1), Ubuntu 16.04.5

### Steps to reproduce:

1. Define a standard nginx 'combined' logfile
2. configure telegraf with a logparser as per above
3. restart telegraf

### Expected behavior:

telegraf should send metrics from the parsed logfile to the defined 'output' (the above config works with telegraf 1.7.x, tested with .0, .1, and .4) 

### Actual behavior:

telegraf 1.8.x's logfile shows it loads the 'inputs.logparser', but does not actually send metrics.  No additional info, error, or debug output is provided.

glinton:
Resolves #4816

### DIFF
--- a/plugins/inputs/logparser/logparser.go
+++ b/plugins/inputs/logparser/logparser.go
@@ -131,8 +131,14 @@ func (l *LogParserPlugin) Start(acc telegraf.Accumulator) error {
 	l.done = make(chan struct{})
 	l.tailers = make(map[string]*tail.Tail)
 
+	mName := "logparser"
+	if l.GrokConfig.MeasurementName != "" {
+		mName = l.GrokConfig.MeasurementName
+	}
+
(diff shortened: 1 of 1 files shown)


danielnelson:
We should configure the parser to produce metrics with names, let's set the measurement name to the defealt in Start, make sure that ParseLine returns the metric with a name and we can call m.Name() to pass it to acc.AddFields.

PR #4852, opened 2018-10-12
Feature : size options can come with units (KB, MiB, ...)

Samuel-BF:
# What it does

Some options can now be expressed either as int (without quotes, representing the value in Bytes), either as strings with units ("100KiB", "-5MB" ...). These options are : 
- inputs/filecount: size (minimum or maximum size of file to consider)
- inputs/http_listener: max_body_size and max_line_size
- inputs/socket_listener : read_buffer_size
- ouputs/influxdb : udp_payload (maximum packet size for UDP packet)

I've not modified max_message_bytes from outputs/kafka because 1. the parallel option in kafka is also expressed in bytes without units and 2. it would be necessary to rename the option.

The default configuration is also updated to show examples with units.

# How it works

It adds the internal.Size type, which can be used by plugins. The configuration option is processed by (*Size) UnmarshalTOML function, which uses [units package](https://godoc.org/github.com/alecthomas/units).

### Required for all PRs:


danielnelson:
Looks good, just need to resolve the conflicts caused when we aliased `http_listener` to `influxdb_listener` and moved the code.  The new `http_listener_v2` plugin could probably take advantage of this change.

Samuel-BF:
@danielnelson : done.

PR #4856, opened 2018-10-14
Extract end-entity certificate data only

woodsb02:
The previous code was iterating over each certificate in the chain and over-writing the "fields" variable with each subsequent certificate. This had the effect of only extracting the data from the last certificate in the chain (typically the issuing authority certificate), when actually the user is likely expecting the first (the end-entity certificate).

### Required for all PRs:

### DIFF
--- a/plugins/inputs/x509_cert/x509_cert.go
+++ b/plugins/inputs/x509_cert/x509_cert.go
@@ -143,11 +143,10 @@ func (c *X509Cert) Gather(acc telegraf.Accumulator) error {
 			"source": location,
 		}
 
-		for _, cert := range certs {
-			fields := getFields(cert, now)
+		// Only extract data from the end-entity certificate (first in the chain)
+		fields := getFields(certs[0], now)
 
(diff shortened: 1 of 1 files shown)


plugins/inputs/x509_cert/x509_cert.go
```diff
@@ -143,11 +143,10 @@ func (c *X509Cert) Gather(acc telegraf.Accumulator) error {
 			"source": location,
 		}
 
-		for _, cert := range certs {
-			fields := getFields(cert, now)
+		// Only extract data from the end-entity certificate (first in the chain)
+		fields := getFields(certs[0], now)
```
danielnelson: Test the length of certs before indexing, just in case.

danielnelson:
related https://github.com/influxdata/telegraf/pull/4873

jtyr:
Having the option to report all or a selected cert (by default the last one) would be nice.

danielnelson:
I think we will do https://github.com/influxdata/telegraf/pull/4873 instead which reports all certificates with additional tags.

Issue #3008, opened 2017-07-11
insecure_skip_verify not supported for influxdb input plugin


kjschnei001:
## Bug report
Despite being *sort of* documented as part of the InfluxDB input plugin, https://github.com/influxdata/telegraf/tree/master/plugins/inputs/influxdb, the `insecure_skip_verify` option is not actually supported.  

### Relevant telegraf.conf:
```
[[inputs.influxdb]]
insecure_skip_verify = true
urls = ["https://127.0.0.1:8084/debug/vars"]
```

### System info:
Telegraf v1.3.3 (git: release-1.3 46db92aad3e68af04f1732598ac89f6b9b11daf8)
CentOS Linux release 7.2.1511 (Core)

### Steps to reproduce:

1. Configure the influxdb telegraf input plugin with `insecure_skip_verify`

### Expected behavior:
`insecure_skip_verify = true` will disable ssl verification of the influxdb certificate when the plugin connects to monitor it

### Actual behavior:
```
2017/07/11 21:09:05 E! Error parsing /etc/telegraf/telegraf.d/default_inputs.conf, line 10: field corresponding to `insecure_skip_verify' is not defined in `*influxdb.InfluxDB'
```

danielnelson:
Support for this is only added in master, and will be in 1.4.0. #2773

If you are following links from https://docs.influxdata.com, note that they are linked to the master branch and are not specific to the version selected.  Sorry about this confusion.

danielnelson:
I opened an issue about the links to the docs https://github.com/influxdata/docs.influxdata.com/issues/1195

nickidw:
> Support for this is only added in master, and will be in 1.4.0. #2773
> 
> If you are following links from https://docs.influxdata.com, note that they are linked to the master branch and are not specific to the version selected. Sorry about this confusion.

I'm running 1.8.2 on Windows, and the error message indicates it's not yet supported.

`C:\Program Files\Telegraf>telegraf --console --test
2018/10/14 18:33:15 I! Using config file: C:\Program Files\Telegraf\telegraf.con
f
2018/10/14 18:33:15 E! Error parsing C:\Program Files\Telegraf\telegraf.conf, li
ne 79: field corresponding to `**insecure-skip-verify**' is not defined in `*influxd
b.InfluxDB'

C:\Program Files\Telegraf>`

nickidw:
If I remove the option I get:
[...]

PR #4857, opened 2018-10-14
Update README.md

nickidw:
Fixed InsecureSkipVerify setting name

### Required for all PRs:


danielnelson:
The `insecure_skip_verify` name should be correct, but if you are having trouble getting it to work let me know.

nickidw:
As mentioned at https://github.com/influxdata/telegraf/issues/3008#issuecomment-429641069 it does not work for me when using 1.8.2 on Windows with **insecure_skip_verify**. Changing the option to the capitalised version **InsecureSkipVerify** without underscores works. This seems to be expected according to https://golang.org/pkg/crypto/tls/#Config

nickidw:
Forget what I said, I did not see that somehow I had dashes instead of underscores. Apologies.

PR #4873, opened 2018-10-17
Add certificate subject as a tag

onurguzel:
If there is a certificate chain in input source, all certificates are
returned with the same source tag. Subject of the certificate is added
to use all certs in the chain as an input.

### Required for all PRs:

### DIFF
--- a/plugins/inputs/x509_cert/x509_cert.go
+++ b/plugins/inputs/x509_cert/x509_cert.go
@@ -4,6 +4,7 @@ package x509_cert
 import (
 	"crypto/tls"
 	"crypto/x509"
+	"crypto/x509/pkix"
 	"encoding/pem"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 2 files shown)


danielnelson:
I think we should call this tag `cn` and it could be a nice touch to add other fields if set such as  `o`, `ou`: https://tools.ietf.org/html/rfc2253#section-2.3

onurguzel:
> I think we should call this tag `cn`

At first, I was using the RDN with [`Subject.String()`](https://golang.org/pkg/crypto/x509/pkix/#Name.String) (Go 1.10+), and then I modified my commit to use CommonName to be able to build on Go 1.9, leaving tag name `subject`. Should I rename to `common_name` or the abbreviation `cn`?

>  and it could be a nice touch to add other fields if set such as `o`, `ou`

Sounds nice. What do you think about a `c` (country) tag?

danielnelson:
Thanks, I'll update the readme and I'm going to add this to 1.8.3 since this is critical for the plugin to work properly.

PR #4874, opened 2018-10-17
Fix segfault in x509_cert

mcfedr:
- When no tls params are in config, tlsCfg will be nil
- fix setting hostname in tlsCfg it will contain the port number and never correctly match server certs
- Add a test that shows that it now works
- Fix readme as port number is required in soruces

Its currently possible to workaround the crash by adding `insecure_skip_verify = true` to the conf

### Required for all PRs:

### DIFF
--- a/plugins/inputs/x509_cert/README.md
+++ b/plugins/inputs/x509_cert/README.md
@@ -10,7 +10,7 @@ file or network connection.
 # Reads metrics from a SSL certificate
 [[inputs.x509_cert]]
   ## List certificate sources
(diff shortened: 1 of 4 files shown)


astellingwerf:
@glinton , when can I expected a Docker image with this fix (telegraf:1.8.2-alpine)? The milestone was closed 7 days ago, but I'm unable to use it.

glinton:
Oops, it seems circle ci was having trouble hitting mit's gpg servers and the tests failed, I've restarted them and will keep a close eye on it today and make sure it gets done today. Thanks for the heads up/reminder.

Issue #4757, opened 2018-09-27
Support IPVS metrics


amoghe:
## Feature Request

I'd like to implement an input plugin for reporting IPVS metrics (for "virtual_server" and "real_server").

### Proposal:

Proposal is to add this in phases. 

## Phase 1 
Add metrics to track "virtual_services". They are namespaced under "virtual_server" so that we can disinguish between them and the "real_server" metrics (see phase 2). 

These are the readily available metrics:
- Connections (`virtual_server.connections`) 
- PacketsIn (`virtual_server.pkts_in`)
- PacketsOut (`virtual_server.pkts_out`)
- BytesIn (`virtual_server.bytes_in`)
- BytesOut  (`virtual_server.bytes_out`)
- CPS (`virtual_server.cps`)
- PPSIn (`virtual_server.pps_in`)
- PPSOut (`virtual_server.pps_out`)
- BPSIn (`virtual_server.bps_in`)
- BPSOut (`virtual_server.bps_out`)

Each metric has the following tags:
- `address`
- `protocol`
- `port`
- `fwmark` (if it is non zero)

## Phase 2
Add metrics to track each of the above metrics, except this time for each "real_server" that backs each of the above "virtual_server"

Each "destination"s metrics will need to contain an additional tag that indicates which "service" it was backing. For this we add
- `service` : concat(`protocol`, `address`, `port`, `fwmark`) with some separator (`-`)

### Use case: [Why is this important (helps with prioritizing requests)]

Help monitoring linux virtual servers.

amoghe:
The terminology is different depending on where you look. The golang libraries call them "Services" and "Destinations" , however the IPVS terminology is "Virtual Server" and "Real Server".

A virtual server is configured to be backed by one or more real servers (that can actually handle the traffic). The virtual server is just a load balancer.

The virtual server can be configured to schedule connections based on proto/addr/port/fwmark (addr+port being commonly used).

These metrics are exposed via the kernel over a netlink interface. There is a pure-go library to deal with netlink sockets and parsing these messages here - github.com/mqliang/libipvs however I've forked it to add a `Close()` method (to allow for graceful error handling and reopen/retry). That is available here - https://github.com/amoghe/libipvs. There is a PR open to merge it back. Once that happens we can switch to that.

amoghe:
I'll switch to using `virtual_server` and `real_server` in the code and documentation. I think it is better to align the terminology with IPVS one than the one used by the libraries.

[...]

Issue #4413, opened 2018-07-12
Standardize on tag representing the source host of the metrics


danielnelson:
### Proposal:

The `host` tag in Telegraf reports the hostname of the system running Telegraf, but when using a plugin that gathers from another service on the network, it can be helpful to know the hostname running the service.

Use `source` as the tag name and a domain name as the value.  The name `source` mirrors Wavefront style (#4402).  The `host` tag would be unchanged, which means some metrics would begin having both tags: `host=example.org,source=example.org`.

The new tag would, of course, be a convention only.  Users would be free to add any value to this tag or tagexclude it completely.  There wouldn't be any special syntax added to the telegraf.conf as was done with the `host` tag, and each plugin would be responsible for adding the appropriate value.

### Current behavior:

Variety of tags are used, with a variety of value formats, here is a sample:
```
docker,engine_host=debian-stretch-docker
nginx,server=localhost,port=80
mongodb,hostname=127.0.0.1:27017
prometheus,url=http://example.org:9273/metrics
```

### Desired behavior:
```
docker,source=debian-stretch-docker
nginx,source=localhost,port=80
mongodb,source=127.0.0.1,port=27017
prometheus,source=example.org,port=9273,resource=/metrics,scheme=http
```

### Use case:

Unification of the tag name and format will allow lookup of metrics from a particular hostname.

danielnelson:
cc @glinton @goller @phemmer @puckpuck @lpic10

phemmer:
Been pondering this since the ping, and I'm still unsure about the name. I think standardizing on something is a good idea, as it is a very common occurrence that the metrics being recorded aren't about the host telegraf is running on. It's just the name `source` that has me a little hesitant. I just keep thinking of scenarios where telegraf is monitoring communication between 2 entities, one of them a source of the communication, one of them a destination. In this case, "source" might not represent where the metrics are being recorded from.
The only 2 terms that pop into mind are "origin", and "agent". Though I'm not too fond of either. "origin" kinda suffers the same problem as "source". The term "agent" is stolen from SNMP nomenclature, and isn't very self-explanatory.

danielnelson:
I also think "agent" would cause confusion with the existing "host" tag, since we promote Telegraf as an agent.  That said I'm not sure the proposed "source" and "host" are any better, and I could imagine either name being used to refer to the host running Telegraf. In many ways I would rather swap the meaning of agent/source with host, but of course it would be too disruptive so we won't be doing that.

[...]

PR #4890, opened 2018-10-19
IPVS input plugin

amoghe:
issue #4757

### Required for all PRs:

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -238,6 +238,6 @@
   branch = "master"
   name = "github.com/docker/libnetwork"
 
-[[constraint]]
+[[override]]
   name = "github.com/vishvananda/netlink"
   revision = "b2de5d10e38ecce8607e6b438b6d174f389a004e"
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -237,3 +237,7 @@
 [[constraint]]
   branch = "master"
   name = "github.com/docker/libnetwork"
+
+[[constraint]]
+  name = "github.com/vishvananda/netlink"
+  revision = "b2de5d10e38ecce8607e6b438b6d174f389a004e"
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -233,3 +233,7 @@
 [[constraint]]
   branch = "master"
   name = "golang.org/x/oauth2"
+
+[[constraint]]
+  branch = "master"
+  name = "github.com/docker/libnetwork"
--- a/plugins/inputs/ipvs/README.md
+++ b/plugins/inputs/ipvs/README.md
@@ -5,7 +5,8 @@ metrics about ipvs virtual and real servers.
 
 ## Configuration
 
-None
+[[inputs.ipvs]]
+  # no configuration
 
 ## Permissions
 
--- a/plugins/inputs/ipvs/README.md
+++ b/plugins/inputs/ipvs/README.md
@@ -1,0 +1,25 @@
+# IPVS Input Plugin (Linux)
+
+The IPVS input plugin uses the linux kernel netlink socket interface to gather
+metrics about ipvs virtual and real servers.
+
+## Configuration
+
+None
+
+## Permissions
+
+Assuming you installed the telegraf package via one of the published packages,
+the process will be running as the `telegraf` user. However, in order for this
+plugin to communicate over netlink sockets it needs the telegraf process to be
+running as `root` (or some user with `CAP_NET_ADMIN` and `CAP_NET_RAW`). Be sure
+to ensure these permissions before running telegraf with this plugin included.
+
+## Sample Output
+
+This is what you can expect the emitted metrics to look like
+
+```
+{"fields":{"bytes_in":0,"bytes_out":0,"connections":0,"cps":0,"pkts_in":0,"pkts_out":0,"pps_in":0,"pps_out":0},"name":"ipvs_virtual_server","tags":{"address_family":"inet","netmask":"32","sched":"rr","address":"172.18.64.234","port":"9000","protocol":"tcp"},"timestamp":1539810710}
+{"fields":{"bytes_in":0,"bytes_out":0,"connections":0,"cps":0,"pkts_in":0,"pkts_out":0,"pps_in":0,"pps_out":0},"name":"ipvs_virtual_server","tags":{"address_family":"inet","netmask":"32","sched":"rr","fwmark":"47"},"timestamp":1539810710}
+```
--- a/plugins/inputs/ipvs/README.md
+++ b/plugins/inputs/ipvs/README.md
@@ -21,6 +21,6 @@ to ensure these permissions before running telegraf with this plugin included.
 This is what you can expect the emitted metrics to look like
 
 ```
-{"fields":{"bytes_in":0,"bytes_out":0,"connections":0,"cps":0,"pkts_in":0,"pkts_out":0,"pps_in":0,"pps_out":0},"name":"ipvs_virtual_server","tags":{"address_family":"inet","netmask":"32","sched":"rr","address":"172.18.64.234","port":"9000","protocol":"tcp"},"timestamp":1539810710}
-{"fields":{"bytes_in":0,"bytes_out":0,"connections":0,"cps":0,"pkts_in":0,"pkts_out":0,"pps_in":0,"pps_out":0},"name":"ipvs_virtual_server","tags":{"address_family":"inet","netmask":"32","sched":"rr","fwmark":"47"},"timestamp":1539810710}
+ipvs_virtual_server,address=172.18.64.234,address_family=inet,netmask=32,port=9000,protocol=tcp,sched=mh_418 bytes_out=0i,pps_in=0i,pps_out=0i,cps=0i,pkts_in=0i,pkts_out=0i,connections=0i,bytes_in=0i 1540407540000000000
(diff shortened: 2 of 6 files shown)


plugins/inputs/ipvs/README.md
```diff
@@ -0,0 +1,25 @@
+# IPVS Input Plugin (Linux)
+
+The IPVS input plugin uses the linux kernel netlink socket interface to gather
+metrics about ipvs virtual and real servers.
+
+## Configuration
+
+None
```
danielnelson: Add this as a ```toml block, it is the output of `./telegraf -usage ipvs` without the plugin description.
```
[[inputs.ipvs]]
  # no configuration
```

plugins/inputs/ipvs/ipvs.go
```diff
@@ -0,0 +1,114 @@
...
+		}
+		acc.AddGauge("ipvs_virtual_server", fields, serviceTags(s))
+	}
+
+	return nil
+}
+
+// helper: given a Service, return tags that identify it
+func serviceTags(s *ipvs.Service) map[string]string {
+	ret := map[string]string{
```
danielnelson: Add a tag `source` and set it to `os.Hostname()`, this probably seems somewhat pointless, but it is for #4413 and will be nice when we get all plugins supporting it.

Gopkg.toml
```diff
@@ -233,3 +233,11 @@
 [[constraint]]
   branch = "master"
   name = "golang.org/x/oauth2"
+
+[[constraint]]
+  branch = "master"
+  name = "github.com/docker/libnetwork"
+
+[[constraint]]
```
danielnelson: I'm no dep expert, so be skeptical of this comment, but should we change this to an override to indicate that its not a direct dependency?

glinton:
Replaces #4792

plugins/inputs/ipvs/README.md
amoghe:
```suggestion
[[inputs.ipvs]]
  # no configuration
```

plugins/inputs/ipvs/README.md
```diff
@@ -0,0 +1,25 @@
...
+running as `root` (or some user with `CAP_NET_ADMIN` and `CAP_NET_RAW`). Be sure
+to ensure these permissions before running telegraf with this plugin included.
+
+## Sample Output
+
+This is what you can expect the emitted metrics to look like
+
+```
+{"fields":{"bytes_in":0,"bytes_out":0,"connections":0,"cps":0,"pkts_in":0,"pkts_out":0,"pps_in":0,"pps_out":0},"name":"ipvs_virtual_server","tags":{"address_family":"inet","netmask":"32","sched":"rr","address":"172.18.64.234","port":"9000","protocol":"tcp"},"timestamp":1539810710}
+{"fields":{"bytes_in":0,"bytes_out":0,"connections":0,"cps":0,"pkts_in":0,"pkts_out":0,"pps_in":0,"pps_out":0},"name":"ipvs_virtual_server","tags":{"address_family":"inet","netmask":"32","sched":"rr","fwmark":"47"},"timestamp":1539810710}
```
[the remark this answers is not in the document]
amoghe: Done

Gopkg.toml
amoghe: Yes, as per my understanding of https://github.com/golang/dep/blob/master/docs/Gopkg.toml.md , it makes sense to capture this as an `override`

amoghe:
```suggestion
[[override]]
```

plugins/inputs/ipvs/ipvs.go
amoghe: Since these are not metrics from a remote service (as described in #4413 ), I dont think this plugin should emit that tag.

As I understand that issue - the problem arises when metrics from a remote host (not the same host that telegraf is running on) are received and then forwarded by telegraf, we want to have a way to know what `data source` the metrics originated on.

However in this case, the metrics are from the host that telegraf is running on (like, cpu/mem/network). So the `agent.omit_hostname` (and related) config knobs should already provide this.

Please correct me if I'm missing something.

danielnelson: In this case the `source` and `host` tag have the same value, but still would have different meanings.  Adding it in this case would just be for uniformity with other plugins with the goal that eventually all metrics would contain both of these tags.

With both tags you will be able to query either by the Telegraf system that reported the metric or by the system that the metric is about, without needing to check which is supported for each plugin, and will be able to list all metrics for a hostname more easily.

plugins/inputs/ipvs/ipvs.go
```diff
@@ -0,0 +1,114 @@
...
+func addressFamilyToString(af uint16) string {
+	switch af {
+	case syscall.AF_INET:
+		return "inet"
+	case syscall.AF_INET6:
+		return "inet6"
+	default:
+		return fmt.Sprintf("%d", af)
+	}
+
```
glinton: nitpick: remove this empty line

plugins/inputs/ipvs/ipvs.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// SampleConfig returns a sample configuration for this input plugin
+func (i *IPVS) SampleConfig() string {
+	return ``
+}
+
+// Gather gathers the stats
+func (i *IPVS) Gather(acc telegraf.Accumulator) error {
+
```
glinton: nitpick: remove this empty line

plugins/inputs/ipvs/ipvs.go
amoghe: @danielnelson  - I think I understand.

I think this conversation is probably to be had on #4413, but I'll continue here for now...

I'd like to point out that the nice thing about the `host` tag is/was that it defaulted to `os.Hostname` but was overridable via the config file. This thoughtful knob was immensely useful in situations where we deploy telegraf to large deployments where the logical name is better understood than the actual dns/hostname given to the system.

After reading the conversation on #4413 (which wandered in the direction of selecting an appropriate name) I still feel like the `source` tag should be added by telegraf when one is not provided by the input plugin (using whatever value it uses for the `host` tag). This makes a lot of sense for all the metrics that originate on the host itself and frees the plugin author from needing to think about this. 

I'd argue that the majority of the metrics collected by the input plugins are from the local host itself. There are fewer plugins that collect metrics from over the network (statsd, wavefront, etc).

If we proceed in this direction, we potentially run into a situation where every input plugin can now handle this tag with subtle differences. In one case, as a plugin author, I may choose to accept this as a config element so as to allow users to specify an alternate value, in other cases the user may override this using `[[inputs.foobar.tags]]`, thus making this redundant. In other cases, Ithe burden to add a `tagdrop` now falls on the user (where she doesn't want this tag). We're now beginning to fray one of telegrafs greatest strengths - its simplicity.

I'd hope that the burden of doing this falls only on the input plugins who regularly collect metrics from over the network (statsd, wavefront, etc).

Has a final decision been reached on this? If so, then I'll go ahead and add this as you suggest. If not, I'd like to participate in that discussion before I add this.

plugins/inputs/ipvs/ipvs.go
amoghe: Done

plugins/inputs/ipvs/ipvs.go
amoghe: Done

amoghe:
@glinton  - not sure what tests would be appropriate here. As you can see, the `Gather` mainly copies data between the response from `docker/libnetwork` struct into the `telegraf.Accumulator`. There are some helper funcs to translate `uint16`s to `strings`, but I'm not sure those tests would add much value. 

Thoughts?

plugins/inputs/ipvs/ipvs.go
danielnelson: There hasn't been any discussion about automatically adding the tag, so I would comment on 4413 about the idea and any details.  I'm not opposed to the idea, but it seems that if we automatically add the tag we need to go through all the remote plugins in one pass and add in the right value.   Otherwise on top of it being incorrect I think it will cause confusion about what the tag means.  I'd say it's about a 50/50 split between localhost only and plugins that can support a remote host on the input plugins.

For plugins that can collect remotely it will have to be the plugin who sets the value, as we don't have a uniform way to determine it from looking at the config and sometimes hostnames are dynamic.  Uniformity in handling will just have to be enforced manually for this reason, and that means we can't totally solve the issue by adding it automatically.

Filtering it out is pretty easy, just add a tagexclude on the outputs, but I think this is something that very few will want.  In general adding a tag like this doesn't increase series cardinality because it either doesn't vary with respect to the the host or there already is identifying information on the metric that it doesn't vary against.

We aren't very far along on adding this tag, so we can leave it out of this plugin for now.  Once we have proper documentation in place for how it should work then we will begin requiring it in new plugins.

glinton:
good point, they really wouldn't add too much value in this case.

danielnelson:
Not required, but my ideas on what tests could be added.  I would say the main thing you would want to test here is the state machine in Gather and how it interacts with `ipvs.Handle`.  Adding a happy path test set of data returned from services could be useful as both an example and if bugs are found we will have the basic structure available to add additional cases.

amoghe:
@danielnelson  I'll update #4413 with my thoughts/ideas on where we could go with this tag.

Thanks for approving this, should we go ahead and merge this right now? Or should I add the `real_server` metrics and then we merge after that? Originally I had intended to do this in 2 (or 3) distinct phases, and this the first.

danielnelson:
Let's do the rest of the work in a follow up PR.

PR #4915, opened 2018-10-24
Add means for config to 'work' (not cause an exit) cross platform

### LINKED ISSUE #2652: processes input means config is not cross-platform
### Relevant telegraf.conf:

```
 # Get the number of processes and group them by status
 #[[inputs.processes]]
 #  # no configuration
```

### System info:

1.2.1, macOS & Windows 10.

### Steps to reproduce:

I had a working telegraf config already, and attempted to run it on a Windows PC on my network to get additional "coverage".

### Expected behavior:

The same config works across multiple hosts.

### Actual behavior:

I was greeted by this:

> 2017/04/12 01:12:59 E! Error parsing E:\telegraf\telegraf.conf, Undefined but requested input: processes

### Additional info:

It was quite easy to get past (comment out the processes input plugin in the config), but it'd be nice if the same config worked without issues across platforms. Is this possible?

glinton:
Potential solution for #2652 

This adds a generic service input plugin that will print a custom error message on `Start()`, allowing a shared config to not prevent telegraf from starting up.

### DIFF
--- a/plugins/inputs/invalid_plugin.go
+++ b/plugins/inputs/invalid_plugin.go
@@ -1,0 +1,39 @@
+package inputs
+
+import (
+	"log"
+
+	"github.com/influxdata/telegraf"
+)
+
+// InvalidPlugin is used when a plugin may be invalid for a
+// platform or other reason.
+type InvalidPlugin struct {
+	InvalidReason string
+	OrigDesc      string
+	OrigSampleCfg string
+}
+
+// Description should be inherited from the parent plugin.
+func (i *InvalidPlugin) Description() string {
+	return i.OrigDesc
+}
+
+// SampleConfig should be inherited from the parent plugin.
+func (i *InvalidPlugin) SampleConfig() string {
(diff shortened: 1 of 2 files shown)


danielnelson:
This will work well for plugins with no options, but I think less so the more options a plugin has.  For example, "supporting" win_perf_counters on linux would require duplicating the config options somewhere or moving them into a file compiled on all platforms.

If we are going to move config code into files compiled on all platforms, which I think we should, then we might as well just split out the Gather/Start/Stop functions:

processes.go: all functions except Gather/Start/Stop
processes_linux.go: Real Gather/Start/Stop
processes_notlinux.go: Stub Gather/Start/Stop

One minor issue is the switch this makes from a regular Input to a ServiceInput.  This is nice because we can trivially log once, but it would have the side effect of printing a confusing message when ran with `--test`.  I do like how it will log earlier, but it might be better to use a guard on Gather to log only the first time its called.

glinton:
Very good point

Issue #2055, opened 2016-11-17
Telegraf InfluxDB udp output does not set database properly


evanccnyc:
## Bug report
  On Telegraf 1.1.1 it appears that the database is not set properly in a UDP packet. Its set properly in HTTP.

### Relevant telegraf.conf:
```
...
[[outputs.influxdb]]
  ## The full HTTP or UDP endpoint URL for your InfluxDB instance.
  ## Multiple urls can be specified as part of the same cluster,
  ## this means that only ONE of the urls will be written to each interval.
  # urls = ["udp://localhost:8089"] # UDP endpoint example
  urls = ["udp://10.10.25.84:9096"] # required
  ## The target database for metrics (telegraf will create it if not exists).
  database = "abcdefg" # required
...
```
### System info:
Ubuntu 14.04
Telegraf v1.1.1 (git: release-1.1.0 94de9dca1fc6efb3a4bf3ec6869c356278c6755a)

### Steps to reproduce:

1. Create InfluxDB output with UDP settings in telegraf.conf
2. Specify database in config
3. Restart Telegraf
4. Wait till data is sent, check influxdb, data points/database are not created properly

### Expected behavior:
Data & database created and data sent properly

### Actual behavior:
No data

### Additional info:

Here is the packet capture of one of the metrics being sent from Telegraf to influx:

```
19:42:00.327605 IP 10.10.22.26.52875 > 10.10.25.84.9096: UDP, length 199
	0x0000:  4500 00e3 4649 4000 4011 b03f 0a0a 161a  E...FI@.@..?....
	0x0010:  0a0a 1954 ce8b 2388 00cf d49c 7072 6f63  ...T..#.....proc
	0x0020:  6573 7365 732c 686f 7374 3d74 656c 6567  esses,host=teleg
	0x0030:  7261 6672 656c 6179 2d73 7461 672d 7573  rafrelay-stag-us
	0x0040:  2d65 6173 742d 3163 2d69 2d34 6139 3961  -east-1c-i-4axxa
	0x0050:  6262 xxxx xxxx xxxx xxxx xxxx 2e73 7461  bb2.xxx.sta
	0x0060:  6720 626c 6f63 6b65 643d 3069 2c70 6167  g.blocked=0i,pag
	0x0070:  696e 673d 3069 2c72 756e 6e69 6e67 3d30  ing=0i,running=0
	0x0080:  692c 736c 6565 7069 6e67 3d31 3239 692c  i,sleeping=129i,
[...]

PR #4974, opened 2018-11-08
Clarify UDP influxdb endpoint

vandot:
Additional explanation for UDP influxdb endpoint.

### DIFF
--- a/plugins/outputs/influxdb/README.md
+++ b/plugins/outputs/influxdb/README.md
@@ -16,6 +16,7 @@ The InfluxDB output plugin writes metrics to the [InfluxDB v1.x] HTTP or UDP ser
   # urls = ["http://127.0.0.1:8086"]
 
   ## The target database for metrics; will be created as needed.
+  ## For UDP url endpoint database needs to be configured on server side.
   # database = "telegraf"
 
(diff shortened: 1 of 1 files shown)


plugins/outputs/influxdb/README.md
```diff
@@ -16,6 +16,7 @@ The InfluxDB output plugin writes metrics to the [InfluxDB v1.x] HTTP or UDP ser
   # urls = ["http://127.0.0.1:8086"]
 
   ## The target database for metrics; will be created as needed.
+  ## For UDP url endpoint database needs to be configured on server side.
```
glinton: Place this above the `urls` token. Also, it may read more clearly as:
`Server must be configured to listen on a UDP socket in order for 'UDP' to work here`

vandot: @glinton this docs change is about database configuration, if you change database parameter on the client side for UDP endpoint your data will be still written to the database that is configured on the server side. For HTTP and UNIX it is possible to change database on the client side. The only place that mentions this behavior is in this ticket [#2055](https://github.com/influxdata/telegraf/issues/2055#issuecomment-261350359)

glinton: Ok, I understand now. It was not immediately clear that is what was meant by "configured".

PR #4977, opened 2018-11-09
refactor(stackdriver): Update stackdriver input plugin

WuHan0608:
### Required for all PRs:

### DIFF
--- a/plugins/inputs/stackdriver/README.md
+++ b/plugins/inputs/stackdriver/README.md
@@ -1,0 +1,100 @@
+# Stackdriver Input Plugin
+
+Stackdriver scrapes metrics from Google Cloud Monitoring.
+
+### Credentials
+
+This plugin uses a [Google Service Account](https://cloud.google.com/docs/authentication/getting-started) to interacting with the Stackdriver Monitoring API.
+
+### Configuration:
+
+```toml
+[[inputs.stackdriver]]
+  ## GCP Project
+  project = "projects/erudite-bloom-151019"
+  
+  ## API rate limit. On a default project, it seems that a single user can make
+  ## ~14 requests per second. This might be configurable. Each API request can
+  ## fetch every time series for a single metric type, though -- this is plenty
+  ## fast for scraping all the builtin metric types (and even a handful of
+  ## custom ones) every 60s.
+  # rateLimit = 14
+  
+  ## Every query to stackdriver queries for data points that have timestamp t
+  ## such that: (now() - lookbackSeconds) <= t < now(). Note that influx will
+  ## de-dedupe points that are pulled twice, so it's best to be safe here, just
+  ## in case it takes GCP awhile to get around to recording the data you seek.
+  # lookbackSeconds = 600
+  
+  ## Collection Delay Seconds (required - must account for metrics availability via Stackdriver Monitoring API)
+  # delaySeconds = 60
+  
+  ## Recommended: use metric 'interval' that equals to 'lookbackSeconds' to avoid
+  ## gaps or overlap in pulled data
+  interval = "10m"
+  
+  ## Configure the TTL for the internal cache of timeseries requests.
+  ## Defaults to 1 hr if not specified
+  # cacheTTLSeconds = 3600
+  
+  ## Sets whether or not to scrape all bucket counts for metrics whose value
+  ## type is "distribution". If those ~70 fields per metric
+  ## type are annoying to you, try out the distributionAggregationAligners
+  ## configuration option, wherein you may specifiy a list of aggregate functions
+  ## (e.g., ALIGN_PERCENTILE_99) that might be more useful to you.
+  # scrapeDistributionBuckets = true
+  
+  ## Excluded GCP metric types. Any string prefix works.
+  ## Only declare either this or includeMetricTypePrefixes
+  excludeMetricTypePrefixes = [
+    "agent",
+    "aws",
+    "custom"
+  ]
+  
+  ## *Only* include these GCP metric types. Any string prefix works
+  ## Only declare either this or excludeMetricTypePrefixes
+  # includeMetricTypePrefixes = nil
+  
+  ## Excluded GCP metric and resource tags. Any string prefix works.
+  ## Only declare either this or includeTagPrefixes
+  excludeTagPrefixes = [
+    "pod_id",
+  ]
+  
+  ## *Only* include these GCP metric and resource tags. Any string prefix works
+  ## Only declare either this or excludeTagPrefixes
+  # includeTagPrefixes = nil
+  
+  ## Declares a list of aggregate functions to be used for metric types whose
+  ## value type is "distribution". These aggregate values are recorded in the
+  ## distribution's measurement *in addition* to the bucket counts. That is to
+  ## say: setting this option is not mutually exclusive with
+  ## scrapeDistributionBuckets.
+  distributionAggregationAligners = [
+    "ALIGN_PERCENTILE_99",
+    "ALIGN_PERCENTILE_95",
+    "ALIGN_PERCENTILE_50",
+  ]
+  
+  ## The filter string consists of logical AND of the 
+  ## resource labels and metric labels if both of them
+  ## are specified. (optional)
+  ## See: https://cloud.google.com/monitoring/api/v3/filters
+  ## Declares resource labels to filter GCP metrics
+  ## that match any of them.
+  # [[inputs.stackdriver.filter.resourceLabels]]
+  #   key = "instance_name"
+  #   value = 'starts_with("localhost")'
+  
+  ## Declares metric labels to filter GCP metrics
+  ## that match any of them.
+  # [[inputs.stackdriver.filter.metricLabels]]
+  #   key = "device_name"
+  #   value = 'one_of("sda", "sdb")'
+```
+
+### Tips
+
+- If `includeMetricTypePrefixes` field is specified, this plugin will add filter string into list metric descriptors request which is more efficient.
+- If `includeMetricTypePrefixes` field is left blank, this plugin will fetch all metric descriptors. Usually this will cause more API cost.
--- a/plugins/inputs/stackdriver/README.md
+++ b/plugins/inputs/stackdriver/README.md
@@ -10,7 +10,7 @@ This plugin uses a [Google Service Account](https://cloud.google.com/docs/authen
 
 ```toml
 [[inputs.stackdriver]]
-  ## GCP Project
+  ## GCP Project (required - must be prefixed with "projects/")
   project = "projects/{project_id_or_number}"
   
   ## API rate limit. On a default project, it seems that a single user can make
--- a/plugins/inputs/stackdriver/README.md
+++ b/plugins/inputs/stackdriver/README.md
@@ -11,7 +11,7 @@ This plugin uses a [Google Service Account](https://cloud.google.com/docs/authen
 ```toml
 [[inputs.stackdriver]]
   ## GCP Project
-  project = "projects/erudite-bloom-151019"
+  project = "projects/{project_id_or_number}"
   
   ## API rate limit. On a default project, it seems that a single user can make
   ## ~14 requests per second. This might be configurable. Each API request can
@@ -20,18 +20,20 @@ This plugin uses a [Google Service Account](https://cloud.google.com/docs/authen
   ## custom ones) every 60s.
   # rateLimit = 14
   
-  ## Every query to stackdriver queries for data points that have timestamp t
-  ## such that: (now() - lookbackSeconds) <= t < now(). Note that influx will
-  ## de-dedupe points that are pulled twice, so it's best to be safe here, just
-  ## in case it takes GCP awhile to get around to recording the data you seek.
-  # lookbackSeconds = 600
-  
   ## Collection Delay Seconds (required - must account for metrics availability via Stackdriver Monitoring API)
   # delaySeconds = 60
-  
-  ## Recommended: use metric 'interval' that equals to 'lookbackSeconds' to avoid
-  ## gaps or overlap in pulled data
-  interval = "10m"
+
+  ## The first query to stackdriver queries for data points that have timestamp t
+  ## such that: (now() - delaySeconds - lookbackSeconds) <= t <= (now() - delaySeconds).
+  ## The subsequence queries to stackdriver query for data points that have timestamp t
+  ## such that: lastQueryEndTime <= t <= (now() - delaySeconds).
+  ## Note that influx will de-dedupe points that are pulled twice,
+  ## so it's best to be safe here, just in case it takes GCP awhile
+  ## to get around to recording the data you seek.
+  # lookbackSeconds = 120
+
+  ## Metric collection period
+  interval = "1m"
   
   ## Configure the TTL for the internal cache of timeseries requests.
   ## Defaults to 1 hr if not specified
--- a/plugins/inputs/stackdriver/stackdriver.go
+++ b/plugins/inputs/stackdriver/stackdriver.go
@@ -28,7 +28,7 @@ const (
 	description  = "Plugin that scrapes Google's v3 monitoring API."
 	sampleConfig = `
 ## GCP Project
-project = "projects/erudite-bloom-151019"
+project = "projects/{project_id_or_number}"
 
 ## API rate limit. On a default project, it seems that a single user can make
 ## ~14 requests per second. This might be configurable. Each API request can
@@ -37,18 +37,20 @@ project = "projects/erudite-bloom-151019"
 ## custom ones) every 60s.
 # rateLimit = 14
 
-## Every query to stackdriver queries for data points that have timestamp t
-## such that: (now() - lookbackSeconds) <= t < now(). Note that influx will
-## de-dedupe points that are pulled twice, so it's best to be safe here, just
-## in case it takes GCP awhile to get around to recording the data you seek.
-# lookbackSeconds = 600
-
 ## Collection Delay Seconds (required - must account for metrics availability via Stackdriver Monitoring API)
 # delaySeconds = 60
 
-## Recommended: use metric 'interval' that is not greater than 'lookbackSeconds' to avoid
-## gaps in pulled data
-interval = "10m"
+## The first query to stackdriver queries for data points that have timestamp t
+## such that: (now() - delaySeconds - lookbackSeconds) <= t <= (now() - delaySeconds).
+## The subsequence queries to stackdriver query for data points that have timestamp t
+## such that: lastQueryEndTime <= t <= (now() - delaySeconds).
+## Note that influx will de-dedupe points that are pulled twice,
+## so it's best to be safe here, just in case it takes GCP awhile
+## to get around to recording the data you seek.
+# lookbackSeconds = 120
+
+## Metric collection period
+interval = "1m"
 
 ## Configure the TTL for the internal cache of timeseries requests.
 ## Defaults to 1 hr if not specified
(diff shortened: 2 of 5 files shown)


plugins/inputs/stackdriver/README.md
```diff
@@ -0,0 +1,100 @@
...
+  ## such that: (now() - lookbackSeconds) <= t < now(). Note that influx will
+  ## de-dedupe points that are pulled twice, so it's best to be safe here, just
+  ## in case it takes GCP awhile to get around to recording the data you seek.
+  # lookbackSeconds = 600
+  
+  ## Collection Delay Seconds (required - must account for metrics availability via Stackdriver Monitoring API)
+  # delaySeconds = 60
+  
+  ## Recommended: use metric 'interval' that equals to 'lookbackSeconds' to avoid
+  ## gaps or overlap in pulled data
```
JanekLehr: Hi, it's great to see this plugin get more traction now!

Why is it bad to have overlap? InfluxDB de-dupes points that re pulled twice, as is describe in the comments for `lookbackSeconds`. Based on that comment it sounds like metrics come in late sometimes (`delaySeconds`), so it can be good to have overlap to fill in gaps. Does your experience show otherwise?

WuHan0608: Thanks for your advice. The comment is updated.

JanekLehr: Thanks! The updated comment make sense to me.

sciurus:
What is needed to push this feature over the finish line and get it merged? I should be able to test this PR soon if that is helpful.

plugins/inputs/stackdriver/README.md
sciurus: I think the comments are still misleading. Based on https://github.com/influxdata/telegraf/pull/4977/files#diff-7f20909bbc5d4389bc3295b1477ec408R288 the window that data is pulled from is not

`(now() - lookbackSeconds) <= t < now()`

but rather `(now() - lookbackSeconds) <= t < (now() - delaySeconds)`

right?

Won't the way this is calculated leave a gap with missing data?

Imagine i have the defaults of

```
delaySeconds = 60
lookbackseconds == 600
interval == 10m
```
That means if i start telegraf 

at 2:00:00 it will pull data from 1:50:00 to 1:59:00
at 2:10:00 it will pull data from 2:00:00 to 2:09:00

and I missed the data from 1:59:00 to 2:00:00.

Might the fix be to change the window to 

`(now() - lookbackSeconds - delaySeconds) <= t < (now() - delaySeconds)`

?

If you go with that, consider renaming lookbackSeconds to something like windowWidth.

plugins/inputs/stackdriver/README.md
```diff
@@ -0,0 +1,100 @@
...
+  interval = "10m"
```
sciurus: This means by default telegraf will only publish metrics from stackdriver every ten minutes, right? That's quite a long delay before data is available.

plugins/inputs/stackdriver/README.md
```diff
@@ -0,0 +1,100 @@
...
+### Credentials
+
+This plugin uses a [Google Service Account](https://cloud.google.com/docs/authentication/getting-started) to interacting with the Stackdriver Monitoring API.
+
+### Configuration:
+
+```toml
+[[inputs.stackdriver]]
+  ## GCP Project
+  project = "projects/erudite-bloom-151019"
```
sciurus: Consider documenting that the `projects/` prefix is required by goggle. It took quite some digging for me to find that format used by the stackdriver api is `"projects/{project_id_or_number}"`

plugins/inputs/stackdriver/README.md
WuHan0608: The comment is updated.

sciurus:
So far this plugin is not working for me.

I'm unable to build just this branch. I am able to build it after merging in master.

I've ran it on a GCP instance with full access to the Stackdriver Monitoring API. This instance is in a test project with Stackdriver set up but not much data. I configured this plugin with

```
[[inputs.stackdriver]]
project = "projects/redacted"
distributionAggregationAligners = [
        "ALIGN_PERCENTILE_99",
        "ALIGN_PERCENTILE_95",
        "ALIGN_PERCENTILE_50",
]
```

but otherwise left its values at their default. I have telegraf configured to output to a file.

When I started telefgraf, it quickly jumped to 99% cpu utilization and stays there. It printed

```
bpitts@telegraf-stackdriver-test:~/stackdriver_test$ ./telegraf --config telegraf.conf
2019-01-08T17:20:58Z I! Starting Telegraf
2019-01-08T17:20:58Z I! Loaded inputs: inputs.system inputs.stackdriver
2019-01-08T17:20:58Z I! Loaded aggregators:
2019-01-08T17:20:58Z I! Loaded processors:
2019-01-08T17:20:58Z I! Loaded outputs: file
2019-01-08T17:20:58Z I! Tags enabled: host=telegraf-stackdriver-test
2019-01-08T17:20:58Z I! [agent] Config: Interval:10s, Quiet:false, Hostname:"telegraf-stackdriver-test", Flush Interval:10s
2019-01-08T17:21:10Z W! [agent] input "inputs.stackdriver" did not complete within its interval
2019-01-08T17:21:20Z W! [agent] input "inputs.stackdriver" did not complete within its interval
```

The message about the inputs.stackdriver not completing within its interval repeats every ten seconds, which makes me think the default plugin-specific interval of `interval = "1m"` is not working properly.

I let this run for five minutes. The 99% cpu and log message continued this entire time. Telegraf did not respond to SIGINT or SIGTERM so I killed it with SIGKILL. The output file did not contain any metrics from stackdriver.

I uncommented `interval = "1m"` in the config and started telegraf again. It again went to 99% cpu utilization. This time, the message about not completing within interval was printed every minute.

```
bpitts@telegraf-stackdriver-test:~/stackdriver_test$ ./telegraf --config telegraf.conf
2019-01-08T17:30:40Z I! Starting Telegraf
2019-01-08T17:30:40Z I! Loaded inputs: inputs.system inputs.stackdriver
2019-01-08T17:30:40Z I! Loaded aggregators:
2019-01-08T17:30:40Z I! Loaded processors:
2019-01-08T17:30:40Z I! Loaded outputs: file
2019-01-08T17:30:40Z I! Tags enabled: host=telegraf-stackdriver-test
2019-01-08T17:30:40Z I! [agent] Config: Interval:10s, Quiet:false, Hostname:"telegraf-stackdriver-test", Flush Interval:10s
2019-01-08T17:32:00Z W! [agent] input "inputs.stackdriver" did not complete within its interval
2019-01-08T17:33:45Z W! [agent] input "inputs.stackdriver" did not complete within its interval
```

Again, I had to kill it with SIGKILL and there were no metrics from stackdriver in the output.

I'll bump the interval to 10m and leave it running to see what happens.

sciurus:
After changing the interval to 10m, what I see is telegraf's CPU usage spikes to 99% for two minutes every ten minutes, then goes back to idle. I still don't have any metrics from stackdriver in my output.

sciurus:
@WuHan0608 any thought on what I reported earlier?

I have some time to test this again, so I started with changing the configuration to

```
interval = "10s"
includeMetricTypePrefixes = ["compute.googleapis.com/instance/cpu/utilization"]
```

As I understand the code, this means that we'll use a filter when calling getting metric descriptors so we only find this one (instead of the 1,376 I see when i test hitting that api), and then we'll only try to list this single time series.

Limiting to to just this descriptor/series makes cpu utilization negligible and the plugin completes within ten seconds. However, I still get the error about a filter matching more than one metric, which means the call to list the time series must be broken somehow.

plugins/inputs/stackdriver/stackdriver.go
```diff
@@ -0,0 +1,690 @@
...
+	// field just holds the value of the field name.)
+	fieldPrefix string
+	// The GCP API request that we'll use to fetch data for this time series.
+	listTimeSeriesRequest *monitoringpb.ListTimeSeriesRequest
+}
+
+// Generate filter string for ListTimeSeriesRequest
+func (s *Stackdriver) newListTimeSeriesFilter(metricType string) string {
+	if s.Filter == nil {
+		return ""
```
sciurus: This is what is causing the "The provided filter matches more than one metric." error that I saw. The filterString defined in line 326 should be returned when the user hasn't defined any additional filters. We always have to filter by a single metric type.

After I made this change in https://github.com/sciurus/telegraf/commit/352a57902ff122b9267a054b9e855b475b6d062a I got metrics from stackdriver in my output.

plugins/inputs/stackdriver/stackdriver.go
```diff
@@ -0,0 +1,690 @@
...
+## Note that influx will de-dedupe points that are pulled twice,
+## so it's best to be safe here, just in case it takes GCP awhile
+## to get around to recording the data you seek.
+# lookbackSeconds = 120
+
+## Metric collection period
+interval = "1m"
+
+## Configure the TTL for the internal cache of timeseries requests.
+## Defaults to 1 hr if not specified
```
sciurus: I don't think this actually defaults to 1 hour. I could be missing it, but I don't see any code that sets that default.

The function newTimeSeriesConf should be called per-timeseries every interval if there is no cache, but only once per cacheTTLSeconds if there is a cache. I added a debugging statement to this function

When I  specified an interval of 1m and did not set cacheTTLSeconds, my debugging statement appeared once per timeseries per minute. That shows there was actually no hour-long cache of the metric names. Metric data was output properly.

When I specified an interval of 1m and set `cacheTTLSeconds = 3600`, my debugging statment ran once per timeseries and then did not run again in the following minutes. That suggests the cache is there now. However, the plugin stopped outputting any metrics from stackdriver!

sciurus: Actually, I take back the portion about enabling a cache causing stackdriver metric output to stop. Something else has gone wrong, because now with cache commented out again my earlier change isn't working anymore.

sciurus:
With my change to fix the filter, occasionally I can get metrics out of this. Most of the time I don't, though, and I'm not sure why.

I added some debugging print statements to scrapeTimeSeries. I see my log line right before we make the request, but my log lines that should trigger on success or failure are never printed.

https://github.com/sciurus/telegraf/blob/15efab847ec37439c532c5ffd222a51d90087ec1/plugins/inputs/stackdriver/stackdriver.go#L578-L586

danielnelson:
@sciurus I don't have any news yet, but I am looking at this plugin now and I'll try to get to the bottom of the issues you are experiencing.

sciurus:
Actually, I think the problem is that for the metric I'm testing with, stackdriver is typically delaying it by ~3 minutes. So with delaySeconds of 60 and interval of 1m I would keep pulling data from 60 to 120 seconds ago, which was usually empty. After changing delaySeconds to 300 and keeping interval at 1m I'm always fetching data.

My attempt at adding a debugging line for a successful response must have been wrong; I guess that code isn't triggered if the response is empty.

I'll try to test some more and provide more feedback on the interactions of interval, delaySeconds, and lookbackSeconds tomorrow.

danielnelson:
I'm using this for the "The provided filter matches more than one metric." message.

```diff
diff --git a/plugins/inputs/stackdriver/stackdriver.go b/plugins/inputs/stackdriver/stackdriver.go
index 7eed706a..cc7a940f 100644
--- a/plugins/inputs/stackdriver/stackdriver.go
+++ b/plugins/inputs/stackdriver/stackdriver.go
@@ -313,10 +310,6 @@ type timeSeriesConf struct {
 
 // Generate filter string for ListTimeSeriesRequest
 func (s *Stackdriver) newListTimeSeriesFilter(metricType string) string {
-       if s.Filter == nil {
-               return ""
-       }
-
        functions := []string{
                "starts_with",
                "ends_with",
@@ -325,6 +318,10 @@ func (s *Stackdriver) newListTimeSeriesFilter(metricType string) string {
        }
        filterString := fmt.Sprintf(`metric.type = "%s"`, metricType)
 
+       if s.Filter == nil {
+               return filterString
+       }
+
```

sciurus:
Thanks for starting to take a look at this! I've uploaded my notes on the code to https://gist.github.com/sciurus/0cd53dcea1bf8343f8dd635ed07070dd . Since I'm not proficient with golang it was helpful for me to write things down.

## CPU Usage and Metric Exclusions

Regarding sustained high CPU utilization, it is happening when lots of scrapeTimeSeries are running. If you don't supply either includeMetricTypePrefixes or excludeMetricTypePrefixes then you end up with nearly 1500 series to scrape each interval. We can't improve the time it takes to make these calls due to the rate limiting, and I'm not sure if it's expected for them to use as much CPU as they do. Of course, most of these time series won't have data for any given stackdriver project. I think a good safeguard would be to both

1) In the README, encourage people to configure includeMetricTypePrefixes or excludeMetricTypePrefixes appropriately for their environment
2) Set excludeMetricTypePrefixes to exclude "aws" and "agent" by default. I think those are the least likely to be used, and when I check the metric descriptors api they account for 917 of the 1376 entries.

## Collection Windows

On the topic of the data collection window, my goal is to get data from Stackdriver into InfluxDB as quickly as possible. I think the current approach is more complicated than it needs to be and actually makes my goal harder.  

Stackdriver makes new data for a timeseries available once per minute, but that data will be from a variable amount of time in the past. The stackdriver API requires you to give a time range to collect data from. In the current strategy for dealing with this three things affect calculating the window: the interval, lookbackSeconds, and delaySeconds. The calculations are

initial window start = now - delaySeconds - lookbackSeconds
subsequent window starts = previous window end
window end =  now - delaySeconds

Imagine you have

interval = 1m
delaySeconds = 60
lookbackSeconds = 300

and start telegraf at 2:00

2:00 - collect data from 1:54 to 1:59
2:01 - collect data from 1:59 to 2:00
2:02 - collect data from 2:00 to 2:01

Note that lookBackSeconds is effectively pointless, since it is only applied on the first collection. What really determines the length of the window is your interval.

If your data is delayed more than 1 minute, all of those collections would have gathered no data! Let's say our data is delayed randomly by 2 to 4 minutes. So we bump delaySeconds to 5 minutes to be safe.

interval = 1m
delaySeconds = 300
lookbackSeconds = 300

and start telegraf at 2:00

2:00 - collect data from 1:50 to 1:55
2:01 - collect data from 1:55 to 1:56
2:02 - collect data from 1:57 to 2:02

Now we get data, but even if it was available in Stackdriver within 2 minutes we had to wait 5 minutes to fetch it!

### Alternate idea

What if instead, we did

initial window start = now - lookbackSeconds
subsequent window starts = previous window end - lookbackSeconds
window end =  now

Now that we always apply lookbackSeconds, there is no need to keep delaySeconds around.

Let's say data is arriving after a 2 to 4 minute delay, and we have

interval = 1m
lookbackSeconds = 300

with this approach if we start telegraf at 02:00

2:00 - collect data from 1:55 to 2:00
2:01 - collect data from 1:56 to 2:01
2:02  - collect data from 1:57 to 2:02

Now we always get the latest available data, and we backfill any older data as long as it was available within lookbackSeconds.

This does send some duplicate data points to InfluxDB and relies on it to deduplicate them. I'm not sure if this is a big downside.

sciurus:
I've been running my branch on a real GCP project for a day now and it's working well. That let me test the distribution support; both pulling from buckets and setting aligners works.

I was wondering if it would get any slower working with more timeseries that have actual data, but it doesn't seem to. In both my dummy project and the real project, once I set excludeMetricTypePrefixes to "aws" and "agent" scraping all the remaining series takes 39 seconds. If I filter down to just the metric types this project has, scraping those series takes just 11 seconds.

danielnelson:
One advantage of the lookback only method you suggest would be that data is less likely to be missed, just increase the lookback in order to have more reliability at the expense of more data transferred.  It would also provide excellent latency since we would always attempt to pull the latest metrics each interval.

The disadvantage is duplicate points and with it API cost.  The duplicate points would mean additional load on any services downstream of Telegraf, such as messaging systems and InfluxDB, and somewhat odd behavior when debugging with the file output.  Downloading metrics appear to be [billable][] on a per MiB basis, so duplicating points could be an issue if you are pulling down many non GCP metrics.

The way we have settled on in the Cloudwatch input is to have just `delay`.  Points are always gathered since the last time the plugin ran, under normal circumstances this is the same as the interval, but could be larger if there are collection delays.  There are no duplicate points created, but you always must wait the delay to see points.

However, the concerns are not exactly the same.  In Cloudwatch you want to optimise for number of requests due to the request count billing, which encourages a large interval, but in Stackdriver we care more about payload size, though there is also the request [quotas].

I think the current `delay` plus `lookback` variables give the best of both worlds.  It can be setup for low latency with duplicate points (delay=0, lookback=5m), or it can be set to reduce billing with a greater delay (delay=5m, lookback=1m).  The only change I would propose it to have the lookback use an automatic window size like the Cloudwatch plugin does, unless manually set.

[billable]: https://cloud.google.com/stackdriver/pricing#stackdriver_monitoring_services
[quotas]: https://cloud.google.com/monitoring/quotas

sciurus:
@danielnelson 

Thanks for your feedback on the collection timing. I think your idea of keeping both knobs could work. Like you said, that allows tuning them to balance data freshness vs data transferred. If I understand what you mean by "automatic window", the idea is that if you do specify a window it is that specific width, but if you don't specify one your window width will become your interval.

So that means with window set the formulas are

initial start = now - delay - window
subsequent starts = now - delay - window
window end = now - delay

And without a window set the formulas are

initial start = now - delay - 1m
subsequent starts = previous end
window end = now - delay

(1m is the value you suggested hard-coding above)

I think we could simplify the implementation by making window equal to interval when window is unset or 0. Then we could reuse the same formulas for both.

Some example configurations and timings would be

For low latency with duplicate points
interval=1m
delay=0
window=5m

2:00 - collect data from 1:55 to 2:00
2:01 - collect data from 1:56 to 2:01
2:02 - collect data from 1:57 to 2:02

interval=2m
delay=1m
window=5m

2:00 - collect data from 1:54 to 1:59
2:02 - collect data from 1:56 to 2:01
2:04 - collect data from 1:58 to 2:03


To reduce billing with a greater delay

interval=1m
delay=5m
window=0 

2:00 - collect data from 1:54 to 1:55
2:01 - collect data from 1:55 to 1:56
2:02 - collect data from 1:56 to 1:57

interval=2m
delay=5m
window=0 

2:00 - collect data from 1:54 to 1:55 (or if window = interval, 1:53 to 1:55)
2:02 - collect data from 1:55 to 1:57
2:04 - collect data from 1:57 to 1:59

sciurus:
@WuHan0608 can you chime in on whether you're planning to continue work on this PR to address Daniel's feedback? If not, I can ask some of my more golang-proficient teammates to help me do it.

WuHan0608:
I will work on it on this weenkend.

danielnelson:
> I think we could simplify the implementation by making window equal to interval when window is unset or 0. Then we could reuse the same formulas for both.

Plugins do not have access to the value of `interval`, but it might need calculated anyway because it could introduce small gaps that are not collected just due to minor fluctuation in exactly when the plugin runs.

Legogris:
FWIW we are experiencing the same problems as here but for the output plugin - we use it submit application metrics. Tweaking the interval and window resolves it but that can be temporary as an increase in datapoints can introduce the error again.

danielnelson:
@WuHan0608 We are finishing up work on the 1.10 release during this next week and really want to include your plugin.  If it's alright, I will take care of the remaining items on the to do list and then pass it back your way for a final review.

sciurus:
@danielnelson ping me when you're done and I'll be happy to test.

danielnelson:
I made most of the changes suggested in my code review on #5406, along with the new window + delay options discussed above.  I did end up leaving the client response goroutines, which might be helpful in some cases and anyway aren't causing any harm.

@sciurus Would definitely take you up on your offer for testing and review.

sciurus:
Absolutely! I should be able to test this week.

danielnelson:
This pull request was included in #5406, thanks @WuHan0608

Issue #5006, opened 2018-11-19
splunkmetric serializer doesn't properly support bool values


ronnocol:
### Relevant telegraf.conf:
 [[outputs.file]]
   files = ["stdout"]
   data_format = "splunkmetric"
[[inputs.docker]]

### System info:
Telegraf unknown (git: master f1758489)
Linux or OSX
Splunk 7.1+ (with metrics support)

### Steps to reproduce:

1. Run telegraf with an input that produces a bool (e.g. inputs.docker)
2. Watch Splunk be sad with a non-numeric metric value

### Expected behavior:
The metric is either converted into a valid value (1 for true, 0 for false) or dropped like string values are.

### Actual behavior:
Serializer tries to send the metric to Splunk causing an ingest error.

### Additional info:
Reported to me on the Splunk Slack channels by guilhem.marchand

PR #5008, opened 2018-11-19
Refactor value verification to manage bools.

ronnocol:
Strings are still dropped, bools are set to 1 for true and 0 for false
closes: #5006

### Required for all PRs:

- [n/a] Associated README.md updated.

### DIFF
--- a/plugins/serializers/splunkmetric/splunkmetric.go
+++ b/plugins/serializers/splunkmetric/splunkmetric.go
@@ -68,23 +68,16 @@ func (s *serializer) createObject(metric telegraf.Metric) (metricGroup []byte, e
 
 	for _, field := range metric.FieldList() {
 
-		switch field.Value.(type) {
-		case string:
+		value, valid := verifyValue(field.Value)
+
+		if !valid {
 			log.Printf("D! Can not parse value: %v for key: %v", field.Value, field.Key)
 			continue
-		case bool:
-			if field.Value == bool(true) {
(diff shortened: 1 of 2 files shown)


plugins/serializers/splunkmetric/splunkmetric.go
```diff
@@ -68,9 +68,18 @@ func (s *serializer) createObject(metric telegraf.Metric) (metricGroup []byte, e
...
 
-		if !verifyValue(field.Value) {
+		switch field.Value.(type) {
+		case string:
 			log.Printf("D! Can not parse value: %v for key: %v", field.Value, field.Key)
 			continue
+		case bool:
+			if field.Value == bool(true) {
+				// Store 1 for a "true" value
+				field.Value = 1
```
danielnelson: Can you change this to copy to a local?  I'm planning to change around FieldList() to prevent modification this way, it will give more predictable behavior during retries and harder to use the type incorrectly.

Might be easiest to rename `verifyValue` and have it return a value + bool so we don't have to switch twice:
```
func convertValue(v interface{}) (interface{}, bool)
```

ronnocol: went back to having it called verifyValue() but I believe I implemented it in a fashion compatible with your request.

ronnocol: (I don't know what the protocol is for who should resolve the conversation)

PR #5018, opened 2018-11-21
Added Network Interfaces with tagdrop example

gregvolk:
I spent several hours trying to figure out how to use tagdrop to filter out network interfaces that never pass traffic. I think this section is a worthwhile addition to the win_perf_counters documentation because a tagdrop example doesn't appear anywhere else in it.

### Required for all PRs:

### DIFF
--- a/plugins/inputs/win_perf_counters/README.md
+++ b/plugins/inputs/win_perf_counters/README.md
@@ -321,6 +321,29 @@ if any of the combinations of ObjectName/Instances/Counters are invalid.
 ```
 
 
+### Network Interfaces with a tagdrop blacklist
+Windows can have lots of interfaces that never pass any traffic. In the config below we use tagdrop to skip interfaces that start with isatap and Local (Local Area Connection).
+```
+# Network inteface metrics
+[[inputs.win_perf_counters]]
+  [inputs.win_perf_counters.tagdrop]
+    instance = ["isatap*", "Local*"]
+  [[inputs.win_perf_counters.object]]
+    ObjectName = "Network Interface"
+    Instances = ["*"]
+    Counters = [
+      "Packets Received Discarded",
+      "Packets Received Errors",
+      "Packets Outbound Discarded",
+      "Packets Output Errors",
+      "Bytes Received/sec",
+      "Bytes Sent/sec",
+      "Packets Received/sec",
+      "Packets Sent/sec"
+    ]
+    Measurement = "win_net"
+```    
+
 ### Process
 ```
   [[inputs.win_perf_counters.object]]


danielnelson:
I agree that this isn't very clear, and that the win_perf_counters readme really needs redone too, but this option could be used on any input plugin, and I would hate to add it to the readme for all ~150 plugins.

From our previous conversation it sounds like you may have read over the [CONFIGURATION.md](https://github.com/influxdata/telegraf/blob/master/docs/CONFIGURATION.md) document but were still unclear on how to use the option?  Do you think we just need to improve the examples in that document?

gregvolk:
The confusing part to me in this case was that tagdrop can only be used between inputs.win_perf_counters and inputs.win_perf_counters.object and not after inputs.win_perf_counters.object. It makes sense now that I have it working but I think an example that specifically shows the usage would be helpful. I'm happy to move it to the main CONFIGURATION.md file if you think that is a better place for it.

Additionally, [this part of CONIFIGURATION.md](https://github.com/influxdata/telegraf/blob/master/docs/CONFIGURATION.md#input-config-tagpass-and-tagdrop) specifically says

> NOTE tagpass and tagdrop parameters must be defined at the end of the plugin definition, 

but in my example the tagdrop section must be towards the top of the input plugin definition otherwise an error is thrown.

glinton:
I do think that the `CONFIGURATION.md` is a better place for this. You could make it more generic (though still use `win_perf_counters`) by improving the note by adding to "end of the plugin definition" that it must be defined "before any children config elements" of a plugin (`.object` is a child of the `win_perf_counters` plugin)

danielnelson:
In this case the order doesn't matter, it can be used after the object as well:
```
[[inputs.win_perf_counters]]
    ObjectName = "Network Interface"
    Instances = ["*"]
    Counters = [
      "Packets Received Discarded",
    ]
    Measurement = "win_net"
  [inputs.win_perf_counters.tagdrop]
    instance = ["isatap*", "Local*"]
```

However, imagine a plugin with one option `foo`:
```
[[inputs.fake]]
  foo = "bar"
```

If you want to use tagdrop here you must put it at the end:
```
[[inputs.fake]]
  foo = "bar"
  [inputs.fake.tagdrop]
    baz = ["a"]
```

If you were to put it at the beginning then foo would fall under the tagdrop table:
```
[[inputs.fake]]
  # bad!
  [inputs.fake.tagdrop]
    baz = ["a"]
    foo = "bar"
```

Since the `win_perf_counters` config you are using has no top level options the order doesn't matter, but in general it should probably be placed at the end of the plugin.

gregvolk:
Sounds good. I will submit a PR to add the following win_perf_counters tagdrop example to CONFIGURATION.md

```
[[inputs.win_perf_counters]]
    ObjectName = "Network Interface"
    Instances = ["*"]
    Counters = [
      "Packets Received Discarded",
    ]
    Measurement = "win_net"
  [inputs.win_perf_counters.tagdrop]
    instance = ["isatap*", "Local*"]
```

danielnelson:
Closed in #5061

PR #5024, opened 2018-11-22
Switch CPU from field to tag in Interrupts input plugin (#4999)

### LINKED ISSUE #4999: Switch CPU from field to tag in Interrupts input plugin
Currently the inputs.interrupts plugin uses fields to store interrupt counters. This prevents users from grouping, filtering, discriminating and otherwise manipulating data that tags are normally used for.
For example in Grafana you won't be able to filter by chosen CPUs simply because they are fields not tags.

I propose changing CPU from field to a tag and introducing COUNT field.  
The required change here is very simple and I have a PR ready if you're interested.

bitcharmer:
### Required for all PRs:

Resolves #4999

### DIFF
--- a/plugins/inputs/interrupts/README.md
+++ b/plugins/inputs/interrupts/README.md
@@ -5,7 +5,7 @@ The interrupts plugin gathers metrics about IRQs from `/proc/interrupts` and `/p
 ### Configuration
 ```
 [[inputs.interrupts]]
-  # To report cpus as tags instead of fields use cpus_as_tags
+  # To report cpus as tags instead of fields use cpu_as_tags
     # cpu_as_tags = false
   #
   ## To filter which IRQs to collect, make use of tagpass / tagdrop, i.e.
@@ -19,10 +19,10 @@ There are two measurements reported by this plugin.
 - `soft_interrupts` gathers metrics from the `/proc/softirqs` file
 
 ### Fields
-For cpus_as_tags=false (default):
+For cpu_as_tags=false (default):
 - CPUx: the amount of interrupts for the IRQ handled by the CPU
 - Total: sum of interrupts for the IRS for all CPUs
-For cpus_as_tags=true ():
+For cpu_as_tags=true ():
 - Count: the amount of interrupts for the IRQ handled by CPU described in CPU tag
 
 ### Tags
--- a/plugins/inputs/interrupts/README.md
+++ b/plugins/inputs/interrupts/README.md
@@ -16,20 +16,20 @@ There are two measurements reported by this plugin.
 - `soft_interrupts` gathers metrics from the `/proc/softirqs` file
 
 ### Fields
-- CPUx: the amount of interrupts for the IRQ handled by that CPU
-- total: total amount of interrupts for all CPUs
+- Count: the amount of interrupts for the IRQ handled by CPU described in CPU tag
 
 ### Tags
 - irq: the IRQ
 - type: the type of interrupt
 - device: the name of the device that is located at that IRQ
+- cpu: the CPU
 
 ### Example Output
 ```
 ./telegraf --config ~/interrupts_config.conf --test
 * Plugin: inputs.interrupts, Collection 1
-> interrupts,irq=0,type=IO-APIC,device=2-edge\ timer,host=hostname CPU0=23i,total=23i 1489346531000000000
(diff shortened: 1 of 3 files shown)


danielnelson:
Looks good, but we will need a few things to manage the change since it can break queries.

Can you add a new option to the plugin called `metric_version`, the default value should be `metric_version = 1` and if the it is either unset or set to 1 then cpu will still be a field with the old format.  If the value is greater than or equal to `2` then cpu should be a tag.

In the sample config document the option like:
```
## Selects the metric version to maintain backwards compatiblity.
##   recommeneded setting for new plugin use is 2; version 1 is deprecated in 1.10
# metric_version = 1
```

bitcharmer:
Due to your request to preserve old behaviour it was easier to go back to the original design and just add some conditional logic for gathering tags and fields.
The two parallel data formatting options increased testing complexity so I decided to restructure tests and clean up a little.
No need to deprecate original behaviour; both options can coexist.
I decided to introduce `tags_as_cpus` config parameter, which I think is more descriptive and clearer than `version`. Also it will allow you to switch the default value easily with one of the future Telegraf releases.
Thanks

bitcharmer:
@danielnelson is this PR acceptable for you guys?

plugins/inputs/interrupts/interrupts.go
```diff
@@ -12,7 +12,13 @@ import (
 	"github.com/influxdata/telegraf/plugins/inputs"
 )
 
-type Interrupts struct{}
+type Interrupts struct {
+	CpusAsTags bool
+}
+
+func NewInterrupts() *Interrupts {
```
danielnelson: Remove since you didn't end up using.

plugins/inputs/interrupts/interrupts.go
```diff
@@ -29,7 +35,10 @@ func NewIRQ(id string) *IRQ {
 const sampleConfig = `
   ## To filter which IRQs to collect, make use of tagpass / tagdrop, i.e.
   # [inputs.interrupts.tagdrop]
-    # irq = [ "NET_RX", "TASKLET" ]
+	# irq = [ "NET_RX", "TASKLET" ]
+  #
+  # To report cpus as tags instead of fields use cpus_as_tags
+  # cpus_as_tags = false
```
danielnelson: Move this option above the tagdrop usage example, if it is below it will be part of the tagdrop table.  Also in the README.

plugins/inputs/interrupts/interrupts_test.go
```diff
@@ -2,189 +2,167 @@ package interrupts
...
-		},
-		{
-			ID: "TASKLET", Cpus: []int64{int64(205), int64(0)},
-			Total: int64(205),
-		},
+// =====================================================================================
+//	Setup and helper functions
+// =====================================================================================
+
+type MockAccumulator struct {
```
danielnelson: Optional: consider using testutil.Accumulator

plugins/inputs/interrupts/README.md
```diff
@@ -8,6 +8,10 @@ The interrupts plugin gathers metrics about IRQs from `/proc/interrupts` and `/p
   ## To filter which IRQs to collect, make use of tagpass / tagdrop, i.e.
   # [inputs.interrupts.tagdrop]
     # irq = [ "NET_RX", "TASKLET" ]
+  #
+  # To report cpus as tags instead of fields use cpus_as_tags
+  # cpus_as_tags = false
```
danielnelson: Can you tweak this to `cpu_as_tag` (will match better with unbound plugin's `thread_as_tag`)?

bitcharmer:
Hi @danielnelson thanks for the review. Each requested change made sense to me so I followed all your suggestions, made changes and pushed.
Let me know if you find anything else.

danielnelson:
I tweaked the option from `cpu_as_tags` -> `cpu_as_tag` and updated the readme which didn't have the latest style yet.

plugins/inputs/interrupts/README.md
tushar2013: i am using telegraf version 1.9.2 and this is not working.
Error parsing **file.conf**, line 76: field corresponding to `cpu_as_tag' is not defined in `*interrupts.Interrupts'

danielnelson: You need at least version 1.10.  We try to mark the milestone a feature will be released in on the pull request, you should be able to see it on the right sidebar.

PR #5040, opened 2018-11-26
added result_code field

LTKH:
Hello, 
added the result_code field to the procstate input plugin, similar to the http_response and net_response plugins.
result_code (int, success = 0, not_running = 1, error_getting_process = 2)
Please consider the request.


LTKH:
went to think

PR #5045, opened 2018-11-27
Nsd input

### LINKED ISSUE #7401: Add nsd support
## Feature Request

nsd is a authoritative-only DNS server:
https://www.nlnetlabs.nl/projects/nsd/about/

It already has support for exporting metrics via the nsd-control cli utility.
Adding support for this should relatively straightforward.

Currently, we using the exec input with a custom script to support this:

```
#!/bin/sh
MEASUREMENT=nsd
# the time measurements are in float so we sort that out
measurements=$(sudo nsd-control stats_noreset | grep -v ^time | perl -p -e 's/\n/i,/' | sed s/,$// )
echo "$MEASUREMENT $measurements"
```

It sorts out the float values (as we don't really need them), and is a bit fast and dirty, but it works.

It relies on the following sudo entry:

```
telegraf ALL=(ALL) NOPASSWD: /usr/sbin/nsd-control stats_noreset
```

telegraf.conf entry:
```
[inputs.exec]
commands = ["/usr/local/bin/nsd-stats"]
data_format = "influx"
```

### Proposal:

Add input specifically for nsd.

### Use case:

Mostly looking at request / sec to track usage.

izolight:
I made an input plugin for the nsd DNS Server(https://nlnetlabs.nl/projects/nsd/about/). The plugin is based on the unbound Plugin(https://github.com/influxdata/telegraf/tree/master/plugins/inputs/unbound) since they use the same format for the statistics.

closes #7401

### Required for all PRs:


plugins/inputs/nsd/README.md
```diff
@@ -0,0 +1,173 @@
...
+    - num_answer_wo_aa
+    - num_rxerr
+    - num_txerr
+    - num_raxfr
+    - num_truncated
+    - num_dropped
+    - zone_master
+    - zone_slave
+    
+- nsd_thread
```
danielnelson: This won't actually exist will it?

plugins/inputs/nsd/README.md
```diff
@@ -0,0 +1,173 @@
...
+ 
+   ## The default timeout of 1s can be overriden with:
+   # timeout = "1s"
+ 
+   ## When set to true, thread metrics are tagged with the thread id.
+   ##
+   ## The default is false for backwards compatibility, and will be change to
+   ## true in a future version.  It is recommended to set to true on new
+   ## deployments.
+   thread_as_tag = false
```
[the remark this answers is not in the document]
izolight: In nsd the only per thread metric available is the number of queries, which has the prefix serverX.
So i wasn't sure if this should be a tag by default or that we should give a choice. But now that i looked through the unbound commit history i saw that it was only because of legacy versions of the plugin and probably should be true by default here.

plugins/inputs/nsd/nsd_test.go
```diff
@@ -0,0 +1,335 @@
...
+var parsedFullOutputThreadAsTagMeasurementNsd = map[string]interface{}{
+	"num_queries":         float64(32),
+	"time_boot":           float64(340867.515436),
+	"time_elapsed":        float64(3522.901971),
+	"size_db_disk":        float64(11275648),
+	"size_db_mem":         float64(5910672),
+	"size_xfrd_mem":       float64(83979048),
+	"size_config_disk":    float64(0),
+	"size_config_mem":     float64(15600),
+	"num_type_A":          float64(24),
```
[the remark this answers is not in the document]
izolight: only two of the values are actually floats. I changed it to handle these with a simple if

plugins/inputs/nsd/README.md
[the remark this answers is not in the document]
danielnelson: Is this the `server0.queries=32` option?  Yeah, we should make this true by default, no need for an option.  If it is this server option then we could have it be `nsd,server=0 queries=32i`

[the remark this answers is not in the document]
izolight: i made it default, but i need a separate measurement for it "nsd_server". Or is there any other way i can pass the test hasTags() with 2 measurements with the same name but one with tag and the other without?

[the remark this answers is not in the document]
danielnelson: Probably the easiest way is to get the `acc.Metrics`, check which one has the tag, and then compare them against an expected metric created with `testutil.MustMetric` using `testutil.RequireMetricEqual`.

plugins/inputs/nsd/nsd.go
```diff
@@ -0,0 +1,179 @@
...
+					fieldsThreads[threadID][field] = fieldValue
+				}
+			}
+		} else {
+			field := strings.Replace(stat, ".", "_", -1)
+			fields[field] = fieldValue
+		}
+	}
+
+	acc.AddFields("nsd", fields, nil)
```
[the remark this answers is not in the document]
izolight: i added the tags with a default of 8952 for the control port and loopback detection.

plugins/inputs/nsd/README.md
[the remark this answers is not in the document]
izolight: I made a check to find the metric with the tag (both have the same name now)

README.md
```diff
@@ -225,6 +225,8 @@ For documentation on the latest development code see the [documentation index][d
 * [nginx_upstream_check](./plugins/inputs/nginx_upstream_check)
 * [nginx_vts](./plugins/inputs/nginx_vts)
 * [nsq_consumer](./plugins/inputs/nsq_consumer)
+* [nginx_vts](./plugins/inputs/nginx_vts)
```
glinton: Remove this addition

etc/telegraf.conf
```diff
@@ -2877,6 +2877,27 @@
 #   ## HTTP response timeout (default: 5s)
 #   response_timeout = "5s"
 
+# # A plugin to collect stats from the NSD DNS server
```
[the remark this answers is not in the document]
glinton: Revert this file

plugins/inputs/nsd/README.md
```diff
@@ -0,0 +1,164 @@
...
+    - num_truncated
+    - num_dropped
+    - zone_master
+    - zone_slave
+    
+- nsd_server
+  - tags:
+    - server
+  - fields:
+    - queries
```
glinton: Is this the same as `num_queries` above? if so, it may be best to add `source` as a tag to the above metric and give it the value of `server` from this one.

plugins/inputs/nsd/nsd.go
```diff
@@ -0,0 +1,200 @@
...
+
+	fields := make(map[string]interface{})
+	fieldsServers := make(map[string]map[string]interface{})
+
+	host, port, err := net.SplitHostPort(s.Server)
+	// the port is nil since we already checked for other errors in the nsdRunner
+	if err != nil {
+		port = "8952"
+	}
+	_, localhostNet, _ := net.ParseCIDR("127.0.0.1/8")
```
glinton: Make these package-wide variables and set them elsewhere (`init` perhaps) to avoid setting them every call to `Gather`

plugins/inputs/nsd/nsd.go
```diff
@@ -0,0 +1,200 @@
...
+func (s *Nsd) Gather(acc telegraf.Accumulator) error {
+	out, err := s.run(s.Binary, s.Timeout, s.UseSudo, s.Server)
+	if err != nil {
+		return fmt.Errorf("error gathering metrics: %s", err)
+	}
+
+	fields := make(map[string]interface{})
+	fieldsServers := make(map[string]map[string]interface{})
+
+	host, port, err := net.SplitHostPort(s.Server)
```
glinton: Save these to the struct and set them only if they are not set to avoid `SplitHostPort` every call to `Gather`

plugins/inputs/nsd/nsd.go
```diff
@@ -0,0 +1,200 @@
...
+	// the port is nil since we already checked for other errors in the nsdRunner
+	if err != nil {
+		port = "8952"
+	}
+	_, localhostNet, _ := net.ParseCIDR("127.0.0.1/8")
+	_, localhostNetv6, _ := net.ParseCIDR("::1/128")
+	// this should only occur in tests
```
glinton: What in particular "should only occur in tests" and is there a way to modify the tests to not require test-specific code in the application's code?

plugins/inputs/nsd/nsd.go
```diff
@@ -0,0 +1,200 @@
...
+	} else if localhostNet.Contains(hostIP) || localhostNetv6.Contains(hostIP) {
+		hostname, err := os.Hostname()
+		if err != nil {
+			host = "localhost"
+		} else {
+			host = hostname
+		}
+	}
+
+	tags := map[string]string{"source": host, "port": port}
```
glinton: Document these tags in the readme

plugins/inputs/nsd/nsd.go
```diff
@@ -0,0 +1,200 @@
...
+
+	host, port, err := net.SplitHostPort(s.Server)
+	// the port is nil since we already checked for other errors in the nsdRunner
+	if err != nil {
+		port = "8952"
+	}
+	_, localhostNet, _ := net.ParseCIDR("127.0.0.1/8")
+	_, localhostNetv6, _ := net.ParseCIDR("::1/128")
+	// this should only occur in tests
+	hostIP := net.ParseIP(host)
```
glinton: Do this when/where you `SplitHostPort` as it will not change on each call to gather

plugins/inputs/nsd/nsd.go
```diff
@@ -0,0 +1,200 @@
...
+			fields[field] = fieldValue
+		}
+	}
+
+	acc.AddFields("nsd", fields, tags)
+
+	if len(fieldsServers) > 0 {
+		for thisServerID, thisServerFields := range fieldsServers {
+			tags["server"] = thisServerID
+			acc.AddFields("nsd", thisServerFields, tags)
```
glinton: According to the Readme, this shoudl be `nsd_server`.

plugins/inputs/nsd/nsd_test.go
```diff
@@ -0,0 +1,241 @@
...
+	assert.Len(t, acc.Metrics, 2)
+	assert.Equal(t, acc.NFields(), 89)
+
+	tags := map[string]string{
+		"server": "0",
+		"source": "localhost",
+		"port":   "8952",
+	}
+	acc.AssertContainsFields(t, "nsd", parsedFullOutput)
+	acc.AssertContainsTaggedFields(t, "nsd", parsedFullOutputServerAsTagMeasurementNsdServers, tags)
```
glinton: Obviously tests would need updated after changing the measurement name and the above updates.

README.md
izolight: I'm confused, why would I need to remove the nginx_vts line?

plugins/inputs/nsd/README.md
izolight: it is the number of queries answered by this server thread, whereas the num_queries is the total of all threads. this seems to be the only per thread metric.

dwerder:
Is this staled? It would be nice to get this plugin.

sjwang90:
NSD Input Plugin added in #7822

izolight:
interesting that the merged PR is basically a verbatim copy of my PR (with some code removed in subsequent commits).
Any reason why the newer MR was merged and not my existing?

PR #5054, opened 2018-11-28
Improve docs to clarify common issues

### LINKED ISSUE #5026: json_time_format documentation needs to explicitly define the date to use in the "reference time"
### System info:

Ubuntu 18.10
Telegraf 1.9.0 (git: HEAD 2bf21c0d)

### Steps to reproduce:
Create test json file ("`telegraf-test.json`") with the following included in the content (note the date is in the format "YYYY-**MM-DD** HH-MM-SS.000+0800" not "YYYY-**DD-MM** HH-MM-SS.000+0800" (i.e. UK format rather than US format?) which the input parser seems to expect

```
...
"date" : "2016-11-18 12:00:00.000+0800",
...
```

Create following telegraf config file: `telegraf-test-json.conf`
```
[[inputs.file]]
  files = ["telegraf-test.json"]

  ## Data format to consume.
  ## more about them here:
  data_format = "json"

  ## Query is a GJSON path that specifies a specific chunk of JSON to be
  ## parsed, if not specified the whole document will be parsed.
  ##
  ## GJSON query paths are described here:
  ##   https://github.com/tidwall/gjson#path-syntax
  json_query = "**JSON_QUERY**"

  ## Tag keys is an array of keys that should be added as tags.
  tag_keys = [ "test" ]

  ## String fields is an array of keys that should be added as string fields.
 #json_string_fields = []

  json_string_fields = [ "date" ]

  ## Name key is the key to use as the measurement name.
  json_name_key = "test"

  ## Time key is the key containing the time that should be used to create the
  ## metric.
  json_time_key = "date"


  ## Time format is the time layout that should be used to interprete the
  ## json_time_key.  The time must be `unix`, `unix_ms` or a time in the
  ## "reference time".
  ##   ex: json_time_format = "Mon Jan 2 15:04:05 -0700 MST 2006"
  ##       json_time_format = "2006-01-02T15:04:05Z07:00"
  ##       json_time_format = "unix"
  ##       json_time_format = "unix_ms"
  json_time_format = "2016-11-18 12:00:00.000+0800"

```

Run the telegraf test as:
`telegraf --config telegraf-test-json.conf`

The following error is returned:

```
USER@HOST:~/telegraf-tests/json-importer$ telegraf --config telegraf-test-json.conf
2018-11-23T08:17:10Z I! Starting Telegraf 1.9.0
2018-11-23T08:17:10Z E! [telegraf] Error running agent: parsing time "2016-11-18 12:00:00.000+0800": month out of range

```

### Expected behavior:
The date is parsed correctly as expected

### Actual behavior:
When parsing the date, the error is thrown "month out of range"

### Additional info:

### LINKED ISSUE #5031: inputs.system: document /var/run/utmp no such file workaround
### Relevant telegraf.conf:
```
[[inputs.system]]
```

### System info:
Ubuntu 18.04 docker host.
`telegraph:alpine` docker image (I believe it is `1.8.3-alpine`)

### Steps to reproduce:

1. Run the docker image with `[[inputs.system]]` config.

### Expected behavior:
No errors.

### Actual behavior:
Error repeated with every interval.
```
E! Error in plugin [inputs.system]: open /var/run/utmp: no such file or directory
```

### Additional info:
I attempted to set HOST_VAR environment variable, only to realize that my host also doesn't have a `/var/run/utmp`.  I'm not sure why telegraf is looking for this file, but it doesn't exist in the docker container or on my host, and I don't see a way to disable looking for this file.

### LINKED ISSUE #5032: logparser/grok parsing documentation needs to include syntax when using raw grok patterns
### Relevant telegraf.conf:
```# # Stream and parse log file(s).
  [[inputs.logparser]]
#   ## Log files to parse.
#   ## These accept standard unix glob matching rules, but with the addition of
#   ## ** as a "super asterisk". ie:
#   ##   /var/log/**.log     -> recursively find all .log files in /var/log
#   ##   /var/log/*/*.log    -> find all .log files with a parent dir in /var/log
#   ##   /var/log/apache.log -> only tail the apache log file
    files = ["C:\\Users\\paperspace\\Saved Games\\DCS.openbeta\\Logs\\dcs.log"]
#
#   ## Read files that currently exist from the beginning. Files that are created
#   ## while telegraf is running (and that match the "files" globs) will always
#   ## be read from the beginning.
    from_beginning = true
#
#   ## Method used to watch for file updates.  Can be either "inotify" or "poll".
    watch_method = "poll"
#
#   ## Parse logstash-style "grok" patterns:
    [inputs.logparser.grok]
#     ## This is a list of patterns to check the given log file(s) for.
#     ## Note that adding patterns here increases processing time. The most
#     ## efficient configuration is to have one pattern per logparser.
#     ## Other common built-in patterns are:
#     ##   %{COMMON_LOG_FORMAT}   (plain apache & nginx access logs)
#     ##   %{COMBINED_LOG_FORMAT} (access logs + referrer & agent)
      patterns = ["%{GREEDYDATA}"]
```
### System info:

Windows 10

### Expected behavior:
Everything in every line to be matched

### Actual behavior:
Debug logging reports no grok found for every line in the log

### Additional info:
##Example logs:
```
=== Log opened UTC 2018-11-23 01:30:13
2018-11-23 01:30:13.624 INFO    DCS: Command line: "C:\Program Files\Eagle Dynamics\DCS World OpenBeta\bin\DCS.exe" 
2018-11-23 01:30:13.624 INFO    DCS: DCS/2.5.3.24436 (x86_64; Windows NT 6.1.7601)
2018-11-23 01:30:13.640 INFO    DCS: Hwid: 6W_4iFG9Z05PNGJTZdbDFKVShPgFW24xP5HBUA
2018-11-23 01:30:13.640 INFO    DCS: DCS revision: 144434
2018-11-23 01:30:13.640 INFO    DCS: Renderer revision: 18247
2018-11-23 01:30:13.640 INFO    DCS: Terrain revision: 18274
2018-11-23 01:30:13.640 INFO    DCS: CPU cores: 8, threads: 8, System RAM: 32751 MB, Pagefile: 65501 MB
2018-11-23 01:30:13.660 INFO    EDCORE: (dDispatcher)enterToState_:0
2018-11-23 01:30:13.665 INFO    Dispatcher: 2018/11/22 20:30 V1803061700
2018-11-23 01:30:13.669 INFO    INPUT: Device [Keyboard] created deviceId = -1
2018-11-23 01:30:13.669 INFO    INPUT: Device [Mouse] created deviceId = -1
2018-11-23 01:30:13.680 INFO    SOUND: Using driver: xaudio27
2018-11-23 01:30:13.682 INFO    SOUND: Found 1 available audio device(s):
2018-11-23 01:30:13.682 INFO    SOUND:   0: ID: "{0.0.0.00000000}.{1d683da7-071a-4a59-ada9-aaf8c17006f3}", Name: "CABLE Input (VB-Audio Virtual Cable)"
2018-11-23 01:30:13.777 INFO    SOUND: loaded 1313 sdefs from "sounds\sdef"
2018-11-23 01:30:14.385 INFO    NET: Login success.
2018-11-23 01:30:15.344 INFO    NET: Got auth data.
2018-11-23 01:30:15.357 INFO    DCS: Successfully got authorization data.
2018-11-23 01:30:15.359 INFO    DCS: Loading installed modules...
2018-11-23 01:30:16.703 INFO    Scripting: OBSOLETE mount_vfs_sound_path() used in ./CoreMods/aircraft/AJS37/entry.lua
2018-11-23 01:30:16.809 INFO    Scripting: PLUGINS START-------------------------------------------------
2018-11-23 01:30:16.826 INFO    SOUND: loaded 15 sdefs from ".\coremods\aircraft\av8bna\sounds\sdef"
2018-11-23 01:30:16.842 INFO    SOUND: loaded 9 sdefs from ".\coremods\aircraft\c-101\sounds\sdef"
2018-11-23 01:30:16.845 INFO    SOUND: loaded 12 sdefs from ".\coremods\aircraft\chinaassetpack\sounds\sdef"
2018-11-23 01:30:16.854 INFO    SOUND: loaded 19 sdefs from ".\coremods\aircraft\m-2000c\sounds\sdef"
2018-11-23 01:30:16.857 INFO    SOUND: loaded 10 sdefs from ".\coremods\aircraft\mig-21bis\sounds\sdef"
2018-11-23 01:30:16.863 INFO    SOUND: loaded 2 sdefs from ".\coremods\aircraft\sa342\sounds\sdef"
2018-11-23 01:30:16.872 INFO    Scripting: PLUGINS DONE--------------------------------------------------
2018-11-23 01:30:17.287 INFO    SECURITYCONTROL: Loaded ./CoreMods/aircraft/AV8BNA/bin/AV8B_AI.dll
2018-11-23 01:30:17.294 ERROR   EDCORE: LoadLibrary failed for ./CoreMods/services/CaptoGlove/bin/edCaptoGlove.dll: (126) The specified module could not be found.
```

## Patterns that should work
```%{TIMESTAMP_ISO8601} %{LOGLEVEL} %{GREEDYDATA}```
```%{GREEDYDATA}```

Neither works in Telegraf, but both work as tested here  http://grokconstructor.appspot.com/do/match#result

### LINKED ISSUE #5053: telegraf's jolokia 2 documentation needs correction for collecting GC related metrics
### Relevant telegraf.conf:
https://github.com/influxdata/telegraf/blob/master/plugins/inputs/jolokia2/examples/java.conf

### System info:

Telegraf documentation on master

### Steps to reproduce:

The line # 17 in java.conf needs to be changed from:
`mbean    = "java.lang:name=G1*,type=GarbageCollector"`
to
`mbean    = "java.lang:name=*,type=GarbageCollector"`

As it will not be able to collect necessary metrics when searching with `G1*`

### Additional info:
Minor documentation error or typo - needs to be modified.

### LINKED ISSUE #5079: Redis plugin: document 'password' option in README
Hello, i try to use telegraf with redis but i got error - [inputs.redis]: Error in plugin: NOAUTH Authentication required.
It is my config:
```
[[inputs.redis]]
  ## specify servers via a url matching:
  ##  [protocol://][:password]@address[:port]
  ##  e.g.
  ##    tcp://localhost:6379
  ##    tcp://:password@192.168.99.100
  ##
  ## If no servers are specified, then localhost is used as the host.
  ## If no port is specified, 6379 is used
  servers = ["tcp://localhost:6379"]
```
Should i use only with login/password?
Thank you.

### PR DESCRIPTION
**Pull request by glinton**
Closes #5026
Closes #5031 
Closes #5032
Closes #5053 
Closes #5079 
Closes #5029

### CODE REVIEW

File: plugins/inputs/system/README.md [hunk 1]
```diff
@@ -3,6 +3,10 @@
 The system plugin gathers general stats on system load, uptime,
 and number of users logged in. It is similar to the unix `uptime` command.
 
+**Note**: If using the system plugin inside a docker container, you must
```
danielnelson: Let's document this at https://hub.docker.com/_/telegraf/ instead, I'm not sure that it needs to be this complicated either:

https://github.com/influxdata/telegraf/blob/master/docs/FAQ.md#q-how-can-i-monitor-the-docker-engine-host-from-within-a-container

I think if you show an example it will be easier to follow.

plugins/parsers/grok/influx_patterns.go
```diff
@@ -13,6 +13,7 @@ const DEFAULT_PATTERNS = `
 #   https://www.elastic.co/guide/en/logstash/current/plugins-filters-grok.html
 # All default logstash patterns are supported, these can be viewed here:
 #   https://github.com/logstash-plugins/logstash-patterns-core/blob/master/patterns/grok-patterns
+# When using default logstash patterns, a semantic name is required.
```
danielnelson: I'm not sure anyone would ever see this file, in fact maybe we should remove this comment altogether?  We could add it to https://github.com/influxdata/telegraf/blob/master/plugins/inputs/logparser/README.md and https://github.com/influxdata/telegraf/blob/master/plugins/parsers/grok/README.md.

That said, I don't think this is quite right.  A semantic name is only required if you wish to capture the pattern.  Perhaps in the paragraph above we should say something along the lines of "If a pattern does not have a semantic name it will not be captured".

docs/FAQ.md
```diff
@@ -10,7 +10,7 @@ docker run --name telegraf
 	-v /etc:/hostfs/etc:ro
 	-v /proc:/hostfs/proc:ro
 	-v /sys:/hostfs/sys:ro
-	-v /var/run/utmp:/var/run/utmp:ro
+	-v /run/utmp:/var/run/utmp:ro
```
danielnelson: I don't think we should change this, even though they are the same files usually the actual file checked is /var/run/utmp.

plugins/inputs/win_perf_counters/README.md
```diff
@@ -384,6 +389,30 @@ if any of the combinations of ObjectName/Instances/Counters are invalid.
     #IncludeTotal=false #Set to true to include _Total instance when querying for all (*).
 ```
 
+### Multiple plugins with unique tags
```
danielnelson: Don't add this, creates a bad precedent for documenting this general option everywhere.

docs/FAQ.md
MicahZoltu: If you bind mount `/var/run/utmp` to `/var/run/utmp`, you will only mount the _link_, you will not mount the actual volume.  Inside the docker container you will have a link from `/var/run/utmp` to `/run/utmp`, but `/run/utmp` will be _empty_ inside the container.

You could do
```
--mount 'type=bind,source=/var/run/utmp,dst=/var/run/utmp'
--mount 'type=bind,source=/run/utmp,dst=/run/utmp'
```
which would result in the container getting both the symbolic link mounted _and_ the destination of the link mounted, though this adds complexity to the command line.

danielnelson: On my Debian/Gentoo/Centos systems `/var/run` is a symlink to `/run`, but `/var/run/utmp` is the actual file, and mounting this file does work.  Since `/var/run/utmp` is the file that is actually used, mounting any other file would be making an assumption about the filesystem layout on the image.

If `/var/run/utmp` itself is actually a symlink on your system, can you let me know the distro/version?

MicahZoltu:
Ubuntu 18.04

docs/FAQ.md
glinton: I believe what originally happened, thinking more about it, was that `/var/run` was mounted, instead of `/var/run/utmp` as described in the `docs/FAQ`.
Ubuntu 18.04:
```
$ ls -lah /var/run/utmp 
-rw-rw-r-- 1 root utmp 1.2K Dec  5 07:58 /var/run/utmp
$ ls -lah /run/utmp
-rw-rw-r-- 1 root utmp 1.2K Dec  5 07:58 /run/utmp
```

quantenschaum:
The link to built-in pattern in 
https://github.com/influxdata/telegraf/blob/release-1.9/plugins/inputs/logparser/README.md yields 404. It should point to https://github.com/influxdata/telegraf/blob/master/plugins/parsers/grok/influx_patterns.go

danielnelson:
@quantenschaum Thanks for the heads up, I have updated the link.

sslupsky:
Do the references to to logstash NUMBER pattern actually work?

https://github.com/influxdata/telegraf/blob/91cd17fd4094a15fdbea3121c73757b8d33f15d2/plugins/parsers/grok/influx_patterns.go#L46

I tried to use NUMBER but I receive a compile error:

Error: error parsing regexp: invalid or unsupported Perl syntax: `(?<`

I looked further into this, Google's re2 doc says it does not support look ahead or look behind or atomic groups.  The logstash NUMBER pattern uses both negative look behind "(?<!" and and atomic groups "(?>".  So, it appears to me that the NUMBER pattern will not work with telegraf?

danielnelson:
@sslupsky Yes I believe it works, we actually use a library for the grok functionality, and it has special Go versions of some patterns: https://github.com/vjeantet/grok/tree/master/patterns

Here is a quick test I ran on the NUMBER pattern:
```
[[inputs.logparser]]
  files = ["test"]
  from_beginning = true
  [inputs.logparser.grok]
    patterns = ["%{NUMBER:value:int}"]
```

```
$ cat test
42
```

```
logparser,host=loaner,path=test value=42i 1552337750601946347
```

sslupsky:
@danielnelson Thanks for the clarification.  At first, I thought I had to include the logstash grok patterns which caused my initial grief.  After I figured out they were already baked in I deleted the logstash patterns file from my config.  I was having a difficult time finding where the baked in patterns  were.  Your link clears that up.  Thank you.

If I can make a suggestion, please include this link in the readme.

danielnelson:
@sslupsky I have clarified this in the docs.

docs/FAQ.md
peret2000: > If you bind mount `/var/run/utmp` to `/var/run/utmp`, you will only mount the _link_, you will not mount the actual data. Inside the docker container you will have a link from `/var/run/utmp` to `/run/utmp`, but `/run/utmp` will be _empty_ inside the container.
> 
> You could do
> 
> ```
> --mount 'type=bind,source=/var/run/utmp,dst=/var/run/utmp'
> --mount 'type=bind,source=/run/utmp,dst=/run/utmp'
> ```
> 
> which would result in the container getting both the symbolic link mounted _and_ the destination of the link mounted, though this adds complexity to the command line.

That is the good solution, but I do not see it in the docs. It has been very hard to find this solution, already explained in #5031

In [[[inputs.system]] README](https://github.com/influxdata/telegraf/blob/master/plugins/inputs/system/README.md) tells a posible neet to add user to group `utmp`, but I have not found the relationship with the problem.

PR #5066, opened 2018-11-29
New plugin: chronyc

flicker581:
New plugin for fetching almost all possible metrics from chrony. Older one gives only a few, but is compatible with some older versions of chrony.

I've created it to monitor my own stratum 1 NTP server.

### Required for all PRs:

- [*] Signed [CLA](https://influxdata.com/community/cla/).
- [*] Associated README.md updated.
- [*] Has appropriate unit tests.


flicker581:
Can I do something to push this into upstream?

sjwang90:
@flicker581 Would you be willing to submit this as an [external plugin](https://github.com/influxdata/telegraf/blob/master/docs/EXTERNAL_PLUGINS.md) that can be used with `execd` to run seamlessly with Telegraf. I believe this would be the best path forward as this plugin doesn't have the highest interest and has fallen back in our priority. External plugins also don't require the typical review process. 

Closing this for now, please let us know if you have any questions.

PR #5069, opened 2018-11-30
Add running field to procstat_lookup

danielnelson:
This field indicates the number of processes that were actually found, as opposed to pid_count which tells you how many pids were found and may not be what you are after with some query methods.

closes #5043 

### Required for all PRs:

### DIFF
--- a/plugins/inputs/procstat/procstat.go
+++ b/plugins/inputs/procstat/procstat.go
@@ -93,6 +93,7 @@ func (p *Procstat) Gather(acc telegraf.Accumulator) error {
 		case "pgrep":
 			p.createPIDFinder = NewPgrep
 		default:
+			p.PidFinder = "pgrep"
 			p.createPIDFinder = defaultPIDFinder
 		}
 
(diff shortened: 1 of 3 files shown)


danielnelson:
@LTKH Can you give this changeset a shot and let me know if it works out okay?

LTKH:
@danielnelson Tested. If the pid file is missing, `running` returns nothing. If it is possible to make `running` always return zero, even if the pid file is not found?

danielnelson:
I made a few changes, it is always showing the running count now but I also added a result tag+result_code field for the status of the lookup.  If the lookup fails the metric will still be emitted as well.

LTKH:
@danielnelson Tested. Works perfect! Thank you.

plugins/inputs/procstat/procstat.go
```diff
@@ -93,6 +93,7 @@ func (p *Procstat) Gather(acc telegraf.Accumulator) error {
 		case "pgrep":
 			p.createPIDFinder = NewPgrep
 		default:
+			p.PidFinder = "pgrep"
```
glinton: Does this fit better in the init function, as it seems to be a standard location for object defaults?

danielnelson: I tried this out, it seems good but then it moves it away from the setting of `p.createPIDFinder`, which it would need updated alongside of.  We could set that in the init function too, but then we would need to introduce a initialized boolean or such.  We can rework this when we add constructor functions for the plugins (part of the config prototype).

Issue #5114, opened 2018-12-05
Wavefront data format rejects delta metrics


snltd:
### Relevant telegraf.conf:

```
[[inputs.socket_listener]]
  service_address = "unix:///tmp/telegraf.sock"
  data_format = "wavefront"
```
### System info:

Telegraf 1.9.0

### Steps to reproduce:

1. Start telegraf with above config
2. Send a delta metric via the Unix domain socket. Using the Wavefront CLI:

```
$ wf write point -i -u unix -S /tmp/telegraf.sock example.delta 1
```
### Expected behavior:

Delta metric is forwarded to Wavefront proxy and thence to Wavefront.

### Actual behavior:

Metric is not forwarded to proxy: this message in `telegraf.log`

```2018-12-05T18:09:36Z E! [inputs.socket_listener]: Error in plugin: unable to parse incoming line: invalid metric value ```

glinton:
Can you retry with the metric name in double quotes as [their docs suggest](https://docs.wavefront.com/wavefront_data_format.html)?
`wf write point -i -u unix -S /tmp/telegraf.sock "example.delta" 1`

PR #5115, opened 2018-12-05
stop Wavefront parser rejecting delta metrics

snltd:
In response to issue #5114 .

With this patch, Telegraf accepts and forwards delta metrics correctly using a Wavefront format listener.


glinton:
Seeing that they reference [delta counters here](https://docs.wavefront.com/delta_counters.html), it seems like this is valid. I'm surprised they don't update their [format fields docs](https://docs.wavefront.com/wavefront_data_format.html#wavefront-data-format-fields) to have the delta character as a valid character.

danielnelson:
@puckpuck Can you review?

puckpuck:
Sorry I did review when the PR was made, but forgot to comment back.  

Changes look good to me, and in line with what we need to do.

Issue #3663, opened 2018-01-12
Expected udev properties cannot be used in diskio input


cornerot:
### Relevant telegraf.conf:

[[inputs.diskio]]
  device_tags = ["DEVTYPE", "ID_FS_TYPE"]

### System info:

Telegraf v1.5.0 (git: release-1.5 a1668bbf)
CentOS Linux release 7.4.1708 (Core)

### Steps to reproduce:

1. Add "device_tags = ["DEVTYPE", "ID_FS_TYPE"]" to the [[inputs.diskio]] section
2. udevadm info -q property -n /dev/sda1 | egrep 'DEVTYPE|ID_FS_TYPE'

```
DEVTYPE=partition
ID_FS_TYPE=linux_raid_member
```

3. telegraf --test | grep sda1 | sed 's|,|,\n|g;s| |\n|g'

```
2018/01/12 12:15:24 I! Using config file: /etc/telegraf/telegraf.conf
>
diskio,
name=sda1,
ID_FS_TYPE=linux_raid_member,
host=example.com
reads=1810738130i,
read_bytes=44825536246784i,
write_time=502432468i,
io_time=196321694i,
iops_in_progress=0i,
writes=158223504i,
write_bytes=1859747491328i,
read_time=1446620992i,
weighted_io_time=1948537468i
1515759325000000000
```

### Expected behavior:

DEVTYPE and ID_FS_TYPE tags in the telegraf output

### Actual behavior:

ID_FS_TYPE only

### Additional info:

telegraf ignoring all device_tags in the [[inputs.diskio]] section which does not contain underscore: DEVNAME, DEVTYPE, SUBSYSTEM, etc.

danielnelson:
It looks like the suggest `udevadm` command shows more fields than we can actually get from the filesystem, here is how we get these fields:

Get the major/minor device numbers:
```
$ ls -l /dev/vda
brw-rw---- 1 root disk 254, 0 Jan 11 17:26 /dev/vda
```

Then we use this file as source of the device_tags:
```
$ cat /run/udev/data/b254\:0
```

I'm not sure if there is a good source of these properties outside of this method, other than running `udevadm` which would be too slow.

danielnelson:
Although this data is cached, so maybe we could use `udevadm`.

phemmer:
I wouldn't recommend caching udevadm, as the device could get replaced between polling cycles, and you wouldn't know.
It should be easy instead to just calculate the `DEV*` fields.
[...]

PR #5116, opened 2018-12-05
Include DEVLINKS in available diskio udev properties

dbergmanskytap:
Partial fix for https://github.com/influxdata/telegraf/issues/3663. The diskio plugin claims to grab udev properties when specified, but can't actually grab all of them due to the way it is written. One property it didn't grab is "DEVLINKS." Our company, Skytap, has a business need for using diskio to tag with DEVLINKS, and it was only a few lines of code to add that capability.

### Required for all PRs:

### DIFF
--- a/plugins/inputs/diskio/README.md
+++ b/plugins/inputs/diskio/README.md
@@ -19,6 +19,9 @@ The diskio input plugin gathers metrics about disk traffic and timing.
   ## Currently only Linux is supported via udev properties. You can view
   ## available properties for a device by running:
   ## 'udevadm info -q property -n /dev/sda'
+  ## Note: Most, but not all, udev properties can be accessed this way. Properties
+  ## that are currently accessible include the "IS_*" series and the "DEVLINKS" 
+  ## property. For more info see https://github.com/influxdata/telegraf/issues/3663
   # device_tags = ["ID_FS_TYPE", "ID_FS_USAGE"]
   #
   ## Using the same metadata source as device_tags, you can also customize the
--- a/plugins/inputs/diskio/README.md
+++ b/plugins/inputs/diskio/README.md
@@ -20,8 +20,9 @@ The diskio input plugin gathers metrics about disk traffic and timing.
   ## available properties for a device by running:
   ## 'udevadm info -q property -n /dev/sda'
   ## Note: Most, but not all, udev properties can be accessed this way. Properties
-  ## that are currently accessible include the "IS_*" series and the "DEVLINKS" 
-  ## property. For more info see https://github.com/influxdata/telegraf/issues/3663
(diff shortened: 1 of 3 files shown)


plugins/inputs/diskio/diskio_linux_test.go
```diff
@@ -14,6 +14,7 @@ import (
 var nullDiskInfo = []byte(`
 E:MY_PARAM_1=myval1
 E:MY_PARAM_2=myval2
+S:foo/bar/devlink foo/bar/devlink1
```
danielnelson: Shouldn't this be multiple lines like:
```
S:foo/bar/devlink
S:foo/bar/devlink1
```

plugins/inputs/diskio/diskio_linux_test.go
```diff
@@ -47,6 +48,7 @@ func TestDiskInfo(t *testing.T) {
 	require.NoError(t, err)
 	assert.Equal(t, "myval1", di["MY_PARAM_1"])
 	assert.Equal(t, "myval2", di["MY_PARAM_2"])
+	assert.Equal(t, "foo/bar/devlink foo/bar/devlink1", di["DEVLINKS"])
```
danielnelson: Based on running udevadm we should prepend "/dev/" to the links: `/dev/foo/bar/devlink /dev/foo/bar/devlink1`.  I'm not sure if there are situations where the prefix would be something other than `/dev`, maybe in docker, but I think we could assume it for now unless we have a way to determine this.

plugins/inputs/diskio/README.md
```diff
@@ -19,6 +19,9 @@ The diskio input plugin gathers metrics about disk traffic and timing.
   ## Currently only Linux is supported via udev properties. You can view
   ## available properties for a device by running:
   ## 'udevadm info -q property -n /dev/sda'
+  ## Note: Most, but not all, udev properties can be accessed this way. Properties
+  ## that are currently accessible include the "IS_*" series and the "DEVLINKS"
```
danielnelson: Should be able to access any of the `E:` values, so that would be at least `ID*` and `DM*` and probably others.  Might be easier to document fields that cannot be looked up: DEVTYPE, DEVNAME, DEVPATH.

dbergmanskytap: Makes sense.

plugins/inputs/diskio/diskio_linux_test.go
dbergmanskytap: Yes, my mistake. I will update accordingly.

dbergmanskytap:
@danielnelson I think you should be satisfied now

dbergmanskytap:
Woohoo! Thanks!

danielnelson:
Thank you!

PR #5125, opened 2018-12-09
Open weather map input plugin

regel:
### Required for all PRs:

Hi, it's a first implementation of the Open Weather Map plugin. This plugin pulls the current weather information and forecasts for a number of cities using openweathermap.org API.

This first implementation is missing snow information, needs more testing on the snow data reported by openweathermap.org before pushing one extra commit.

Let me know your comments.

### DIFF
--- a/plugins/inputs/openweathermap/README.md
+++ b/plugins/inputs/openweathermap/README.md
@@ -1,0 +1,145 @@
+# Telegraf Plugin: openweathermap
+
+OpenWeatherMap provides the current weather and forecasts for more than 200,000 cities. To use this plugin you will need a token. For more information [click here](https://openweathermap.org/appid).
+
+Find city identifiers in this [list](http://bulk.openweathermap.org/sample/city.list.json.gz)
+
+### Configuration:
+
+```
+[[inputs.openweathermap]]
+  ## Root url of API to pull stats
+  base_url = "http://api.openweathermap.org/data/2.5/"
+  # Your personal user token from openweathermap.org
+  app_id = "xxxxxxxxxxxxxxxxxxxxxxx"
+  # List of city identifiers
+  cities = ["2988507", "519188"]
+  # HTTP response timeout (default: 5s)
+  response_timeout = "5s"
+```
+
+### Measurements & Fields:
+
+- weather
+  - humidity
+  - temperature
+  - pressure
+  - rain
+  - wind.speed
+  - wind.deg
+
+### Tags:
+
+- weather
+  - server
+  - port
+  - base_url
+  - city_id
+
+### Example Output:
+
+Using this configuration:
+```
+[[inputs.openweathermap]]
+  base_url = "http://api.openweathermap.org/data/2.5/"
+  app_id = "change_this_with_your_appid"
+  cities = ["2988507", "519188"]
+  response_timeout = "5s"
+```
+
+When run with:
+```
+./telegraf -config telegraf.conf -input-filter openweathermap -test
+```
+
+It produces data similar to:
+```
+> weather,base_url=http://api.openweathermap.org/data/2.5/,city_id=519188,host=localhost,port=80,server=api.openweathermap.org humidity=85i,pressure=1015,rain=0,temperature=-4.399999999999977,wind.deg=130,wind.speed=2 1544214600000000000
+> weather,base_url=http://api.openweathermap.org/data/2.5/,city_id=2988507,host=localhost,port=80,server=api.openweathermap.org humidity=81i,pressure=1013,rain=0,temperature=6.6299999999999955,wind.deg=250,wind.speed=6.2 1544212800000000000
+> weather,base_url=http://api.openweathermap.org/data/2.5/,city_id=519188,host=localhost,port=80,server=api.openweathermap.org humidity=85i,pressure=1009.95,rain=0,temperature=-5.310000000000002,wind.deg=1.50009,wind.speed=1.3 1544216400000000000
+> weather,base_url=http://api.openweathermap.org/data/2.5/,city_id=519188,host=localhost,port=80,server=api.openweathermap.org humidity=81i,pressure=1009.6,rain=0,temperature=-7.139999999999986,wind.deg=327.5,wind.speed=0.42 1544227200000000000
+> weather,base_url=http://api.openweathermap.org/data/2.5/,city_id=519188,host=localhost,port=80,server=api.openweathermap.org humidity=86i,pressure=1009.32,rain=0,temperature=-7.8700000000000045,wind.deg=264.5,wind.speed=0.86 1544238000000000000
+> weather,base_url=http://api.openweathermap.org/data/2.5/,city_id=519188,host=localhost,port=80,server=api.openweathermap.org humidity=89i,pressure=1009.25,rain=0,temperature=-8.21999999999997,wind.deg=71.5003,wind.speed=0.17 1544248800000000000
+> weather,base_url=http://api.openweathermap.org/data/2.5/,city_id=519188,host=localhost,port=80,server=api.openweathermap.org humidity=91i,pressure=1008.57,rain=0,temperature=-6.658999999999992,wind.deg=110.501,wind.speed=1.27 1544259600000000000
+> weather,base_url=http://api.openweathermap.org/data/2.5/,city_id=519188,host=localhost,port=80,server=api.openweathermap.org humidity=91i,pressure=1008.05,rain=0,temperature=-6.55499999999995,wind.deg=126.5,wind.speed=1.41 1544270400000000000
+> weather,base_url=http://api.openweathermap.org/data/2.5/,city_id=519188,host=localhost,port=80,server=api.openweathermap.org humidity=84i,pressure=1007.72,rain=0,temperature=-7.617999999999995,wind.deg=149.504,wind.speed=1.66 1544281200000000000
+> weather,base_url=http://api.openweathermap.org/data/2.5/,city_id=519188,host=localhost,port=80,server=api.openweathermap.org humidity=87i,pressure=1006.51,rain=0,temperature=-7.324999999999989,wind.deg=151.501,wind.speed=2.41 1544292000000000000
+> weather,base_url=http://api.openweathermap.org/data/2.5/,city_id=519188,host=localhost,port=80,server=api.openweathermap.org humidity=90i,pressure=1005.78,rain=0,temperature=-7.103999999999985,wind.deg=170.506,wind.speed=3.21 1544302800000000000
+> weather,base_url=http://api.openweathermap.org/data/2.5/,city_id=519188,host=localhost,port=80,server=api.openweathermap.org humidity=89i,pressure=1004.9,rain=0,temperature=-8.407999999999959,wind.deg=167,wind.speed=3.47 1544313600000000000
+> weather,base_url=http://api.openweathermap.org/data/2.5/,city_id=519188,host=localhost,port=80,server=api.openweathermap.org humidity=89i,pressure=1003.49,rain=0,temperature=-9.12299999999999,wind.deg=157.504,wind.speed=3.86 1544324400000000000
+> weather,base_url=http://api.openweathermap.org/data/2.5/,city_id=519188,host=localhost,port=80,server=api.openweathermap.org humidity=93i,pressure=1002.54,rain=0,temperature=-7.638999999999953,wind.deg=156.501,wind.speed=3.99 1544335200000000000
+> weather,base_url=http://api.openweathermap.org/data/2.5/,city_id=519188,host=localhost,port=80,server=api.openweathermap.org humidity=96i,pressure=1001.59,rain=0,temperature=-5.065999999999974,wind.deg=158.006,wind.speed=4.09 1544346000000000000
+> weather,base_url=http://api.openweathermap.org/data/2.5/,city_id=519188,host=localhost,port=80,server=api.openweathermap.org humidity=94i,pressure=1000.56,rain=0,temperature=-3.4350000000000023,wind.deg=154,wind.speed=3.81 1544356800000000000
+> weather,base_url=http://api.openweathermap.org/data/2.5/,city_id=519188,host=localhost,port=80,server=api.openweathermap.org humidity=93i,pressure=999.59,rain=0,temperature=-2.70799999999997,wind.deg=151.501,wind.speed=3.71 1544367600000000000
+> weather,base_url=http://api.openweathermap.org/data/2.5/,city_id=519188,host=localhost,port=80,server=api.openweathermap.org humidity=95i,pressure=998.53,rain=0,temperature=-1.6970000000000027,wind.deg=159,wind.speed=3.11 1544378400000000000
+> weather,base_url=http://api.openweathermap.org/data/2.5/,city_id=519188,host=localhost,port=80,server=api.openweathermap.org humidity=94i,pressure=998.21,rain=0,temperature=-0.7939999999999827,wind.deg=182.504,wind.speed=2.56 1544389200000000000
+> weather,base_url=http://api.openweathermap.org/data/2.5/,city_id=519188,host=localhost,port=80,server=api.openweathermap.org humidity=97i,pressure=998.73,rain=0,temperature=-0.39299999999997226,wind.deg=210.503,wind.speed=2.45 1544400000000000000
+> weather,base_url=http://api.openweathermap.org/data/2.5/,city_id=519188,host=localhost,port=80,server=api.openweathermap.org humidity=95i,pressure=998.96,rain=0,temperature=-0.1379999999999768,wind.deg=211.503,wind.speed=2.83 1544410800000000000
+> weather,base_url=http://api.openweathermap.org/data/2.5/,city_id=519188,host=localhost,port=80,server=api.openweathermap.org humidity=98i,pressure=999.72,rain=0,temperature=-0.19299999999998363,wind.deg=197.001,wind.speed=3.12 1544421600000000000
+> weather,base_url=http://api.openweathermap.org/data/2.5/,city_id=519188,host=localhost,port=80,server=api.openweathermap.org humidity=99i,pressure=1000.21,rain=0,temperature=0.382000000000005,wind.deg=184,wind.speed=3.41 1544432400000000000
+> weather,base_url=http://api.openweathermap.org/data/2.5/,city_id=519188,host=localhost,port=80,server=api.openweathermap.org humidity=96i,pressure=1000.77,rain=0,temperature=0.492999999999995,wind.deg=179.502,wind.speed=3.67 1544443200000000000
+> weather,base_url=http://api.openweathermap.org/data/2.5/,city_id=519188,host=localhost,port=80,server=api.openweathermap.org humidity=97i,pressure=1001.33,rain=0,temperature=-0.4590000000000032,wind.deg=178.504,wind.speed=3.62 1544454000000000000
+> weather,base_url=http://api.openweathermap.org/data/2.5/,city_id=519188,host=localhost,port=80,server=api.openweathermap.org humidity=97i,pressure=1001.75,rain=0,temperature=-0.8369999999999891,wind.deg=177.504,wind.speed=3.3 1544464800000000000
+> weather,base_url=http://api.openweathermap.org/data/2.5/,city_id=519188,host=localhost,port=80,server=api.openweathermap.org humidity=97i,pressure=1002.17,rain=0,temperature=-0.7069999999999936,wind.deg=167.001,wind.speed=2.86 1544475600000000000
+> weather,base_url=http://api.openweathermap.org/data/2.5/,city_id=519188,host=localhost,port=80,server=api.openweathermap.org humidity=97i,pressure=1002.29,rain=0,temperature=-0.39400000000000546,wind.deg=152.003,wind.speed=2.41 1544486400000000000
+> weather,base_url=http://api.openweathermap.org/data/2.5/,city_id=519188,host=localhost,port=80,server=api.openweathermap.org humidity=98i,pressure=1002.04,rain=0,temperature=-0.18099999999998317,wind.deg=134.502,wind.speed=2.31 1544497200000000000
+> weather,base_url=http://api.openweathermap.org/data/2.5/,city_id=519188,host=localhost,port=80,server=api.openweathermap.org humidity=97i,pressure=1002.09,rain=0,temperature=0.05600000000004002,wind.deg=136.501,wind.speed=2.72 1544508000000000000
+> weather,base_url=http://api.openweathermap.org/data/2.5/,city_id=519188,host=localhost,port=80,server=api.openweathermap.org humidity=98i,pressure=1002.56,rain=0,temperature=0.5660000000000309,wind.deg=146.001,wind.speed=2.82 1544518800000000000
+> weather,base_url=http://api.openweathermap.org/data/2.5/,city_id=519188,host=localhost,port=80,server=api.openweathermap.org humidity=98i,pressure=1002.84,rain=0,temperature=0.6129999999999995,wind.deg=136.002,wind.speed=2.32 1544529600000000000
+> weather,base_url=http://api.openweathermap.org/data/2.5/,city_id=519188,host=localhost,port=80,server=api.openweathermap.org humidity=100i,pressure=1003.12,rain=0,temperature=0.36299999999999955,wind.deg=129,wind.speed=2.66 1544540400000000000
+> weather,base_url=http://api.openweathermap.org/data/2.5/,city_id=519188,host=localhost,port=80,server=api.openweathermap.org humidity=99i,pressure=1002.9,rain=0,temperature=0.4159999999999968,wind.deg=120.502,wind.speed=2.76 1544551200000000000
+> weather,base_url=http://api.openweathermap.org/data/2.5/,city_id=519188,host=localhost,port=80,server=api.openweathermap.org humidity=98i,pressure=1002.32,rain=0,temperature=0.19400000000001683,wind.deg=116.003,wind.speed=3.01 1544562000000000000
+> weather,base_url=http://api.openweathermap.org/data/2.5/,city_id=519188,host=localhost,port=80,server=api.openweathermap.org humidity=97i,pressure=1001.47,rain=0,temperature=-0.7029999999999745,wind.deg=102.003,wind.speed=3.3 1544572800000000000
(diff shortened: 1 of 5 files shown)


regel:
Signed CLA.

Over to you for feedback

plugins/inputs/openweathermap/openweathermap.go
```diff
@@ -0,0 +1,276 @@
...
+		err = gatherStatusUrl(bufio.NewReader(resp.Body), n.getTags(addr), acc)
+		resp.Body = ioutil.NopCloser(bytes.NewBuffer(body))
+		err = gatherWeatherUrl(bufio.NewReader(resp.Body), n.getTags(addr), acc)
+		return err
+	default:
+		return fmt.Errorf("%s returned unexpected content type %s", addr.String(), contentType)
+	}
+}
+
+func (n *OpenWeatherMap) getTags(addr *url.URL) map[string]string {
```
glinton: These tags don't seem meaningful, consider removing this function.

plugins/inputs/openweathermap/openweathermap.go
```diff
@@ -0,0 +1,276 @@
...
+				"temperature": e.Main.Temp - 273.15, // Kelvin to Celsius
+			},
+			tags,
+			tm)
+	}
+}
+
+func init() {
+	inputs.Add("openweathermap", func() telegraf.Input {
+		return &OpenWeatherMap{}
```
glinton: Set `ResponseTimeout` default here

plugins/inputs/openweathermap/README.md
```diff
@@ -0,0 +1,145 @@
...
+  - wind.speed
+  - wind.deg
+
+### Tags:
+
+- weather
+  - server
+  - port
+  - base_url
+  - city_id
```
glinton: Maybe add `city_name` as a tag

plugins/inputs/openweathermap/openweathermap.go
```diff
@@ -0,0 +1,276 @@
...
+	if err := dec.Decode(status); err != nil {
+		return fmt.Errorf("Error while decoding JSON response: %s", err)
+	}
+	status.Gather(tags, acc)
+	return nil
+}
+
+func gatherWeatherUrl(r *bufio.Reader, tags map[string]string, acc telegraf.Accumulator) error {
+	dec := json.NewDecoder(r)
+	status := &Status{}
```
glinton: Making the current weather part of a forecast in order to process feels wrong. Consider revising.

plugins/inputs/openweathermap/openweathermap_test.go
```diff
@@ -0,0 +1,280 @@
...
+import (
+	"fmt"
+	"net"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"testing"
+
+	"github.com/influxdata/telegraf/testutil"
+	//"github.com/stretchr/testify/assert"
```
glinton: Remove comment

glinton:
@regel do you have an ETA on when you can address the PR feedback?

regel:
Hi Greg, I'm okay with the changes requested. I will free some time, ETA e/o April.

plugins/inputs/openweathermap/README.md
```diff
@@ -0,0 +1,145 @@
...
+```
+[[inputs.openweathermap]]
+  ## Root url of API to pull stats
+  base_url = "http://api.openweathermap.org/data/2.5/"
+  # Your personal user token from openweathermap.org
+  app_id = "xxxxxxxxxxxxxxxxxxxxxxx"
+  # List of city identifiers
+  cities = ["2988507", "519188"]
+  # HTTP response timeout (default: 5s)
+  response_timeout = "5s"
```
russorat: could you also add an option for units? i noticed a little later on you were doing a conversion from Kelvin to Celsius: https://openweathermap.org/current#data

plugins/inputs/openweathermap/openweathermap.go
```diff
@@ -0,0 +1,276 @@
...
+		tm := time.Unix(e.Dt, 0)
+		acc.AddFields(
+			"weather",
+			map[string]interface{}{
+				"rain":        e.Rain.Rain3,
+				"wind.deg":    e.Wind.Deg,
+				"wind.speed":  e.Wind.Speed,
+				"humidity":    e.Main.Humidity,
+				"pressure":    e.Main.Pressure,
+				"temperature": e.Main.Temp - 273.15, // Kelvin to Celsius
```
russorat: adding units format would remove the need for this conversion

plugins/inputs/openweathermap/openweathermap.go
```diff
@@ -0,0 +1,276 @@
...
+
+	if n.client == nil {
+		client, err := n.createHttpClient()
+		if err != nil {
+			return err
+		}
+		n.client = client
+	}
+
+	for _, city := range n.Cities {
```
russorat: You can batch these calls for city data: https://openweathermap.org/current#severalid

plugins/inputs/openweathermap/openweathermap.go
```diff
@@ -0,0 +1,276 @@
...
+	Clouds struct {
+		All int64 `json:"all"`
+	} `json:"clouds"`
+	Main struct {
+		GrndLevel float64 `json:"grnd_level"` // empty for weather/
+		Humidity  int64   `json:"humidity"`
+		SeaLevel  float64 `json:"sea_level"` // empty for weather/
+		Pressure  float64 `json:"pressure"`
+		Temp      float64 `json:"temp"`
+		TempMax   float64 `json:"temp_max"`
```
russorat: why not add TempMax and TempMin as fields well?

russorat:
very cool little plugin! I was able to get it working and build a simple dashboard.

image

plugins/inputs/openweathermap/README.md
regel: I'm not sure, what do you have in mind? @russorat is there a new option in telegraf to declare "Kelvin" anywhere?

plugins/inputs/openweathermap/openweathermap.go
regel: OK, good point. I need a bit more time to test this.

plugins/inputs/openweathermap/openweathermap.go
regel: See previous question

plugins/inputs/openweathermap/README.md
regel: I was concerned about homonyms, that's why I did not use city_name as a tag.

plugins/inputs/openweathermap/openweathermap.go
regel: Both JSON structures given by the OWM API are nearly the same (weather info structure in the past, present, future), that's why I'm reusing the code to parse the two API responses.

regel:
> Thanks for this!
> Can you give a use case for collecting the forecast data? To me, the current weather is much more meaningful and would simplify the plugin's logic. If the forecast data is imperative, consider tagging it as "forecast" and make it's collection optional.

I've made the forecast collection optional and tagged the data accordingly. Having an up to date weather forecast is the reason why I wrote this plugin in the first place.

plugins/inputs/openweathermap/openweathermap.go
regel: I will test again, I don't remember why.

regel:
I rebased on master and the go 1.11 check fails. Then, I rebased on 1.10 and now the AppVeyor check fails, but I don't think it has to do with the OWM plugin. Let me know how you'd like to resolve this.

plugins/inputs/openweathermap/openweathermap.go
regel: OK found it, `temp_min` and `temp_max` are optional in the current weather API response. See the [interpretation](https://openweathermap.org/forecast16). These parameters may be applicable to large cities only.

>Please, do not confuse min/max parameters in current weather API and forecast API. In current weather API temp_min and temp_max are optional parameters mean min / max temperature in the city at the current moment to see deviation from current temp just for your reference. For large cities and megalopolises geographically expanded it might be applicable. In most cases both temp_min and temp_max parameters have the same volume as 'temp'. Please, use temp_min and temp_max parameters in current weather API optionally.

On the contrary, temp.min (not temp_min) and temp.max (not temp_max) given by the _forecast_ API are respectively the daily min temperature, and daily max temperature. Adding this info seems straightforward.

plugins/inputs/openweathermap/openweathermap.go
regel: done in the last commit

plugins/inputs/openweathermap/README.md
regel: Do you mean adding one config option that will control the `units=metric`url parameter?

glinton:
Please rebase this against master

regel:
> Please rebase this against master

Rebase completed and all checks in green. We have one last outstanding question with Russ about the unit metric option, how do you see this extra option?

plugins/inputs/openweathermap/README.md
russorat: sorry i wasn't clear. yeah i mean the option in the config to control the metrics returned from the api. So something that controls `units=metric` in the call to the API.

plugins/inputs/openweathermap/openweathermap.go
russorat: Also, remove this conversion.

plugins/inputs/openweathermap/README.md
```diff
@@ -0,0 +1,67 @@
...
+  ## Root url of API to pull stats
+  # base_url = "http://api.openweathermap.org/data/2.5/"
+  ## Your personal user token from openweathermap.org
+  # app_id = "xxxxxxxxxxxxxxxxxxxxxxx"
+  ## List of city identifiers
+  # cities = ["2988507", "519188"]
+  ## HTTP response timeout (default: 5s)
+  # response_timeout = "5s"
+  ## Query the next ten days forecast (default: true)
+  # forecast_enable = true
```
danielnelson: I feel like a user would want to grab the forecast on a different interval, maybe every 3 hours, while getting the current weather every 10m.  It could be easier to use separate instances of this plugin at the same time if we could disable the weather lookup, and run one or both options:
```toml
[[inputs.openweathermap]]
  interval = "10m"
  app_id = "xxxxxxxxxxxxxxxxxxxxxxx"
  gather = ["weather"]

[[inputs.openweathermap]]
  interval = "1h"
  app_id = "xxxxxxxxxxxxxxxxxxxxxxx"
  gather = ["5d_forecast"]
```

I'm using an array of items in the example in case we want to add more lookups later.

plugins/inputs/openweathermap/openweathermap.go
danielnelson: We can make these struct fields pointers and then test if they are nil before adding.  This will allow them to be added if they exist.

plugins/inputs/openweathermap/README.md
```diff
@@ -0,0 +1,67 @@
...
+  - tags:
+    - city_id
+    - forecast
+
+### Example Output:
+
+Using this configuration:
+```toml
+[[inputs.openweathermap]]
+  base_url = "http://api.openweathermap.org/data/2.5/"
```
danielnelson: Lets use `https` as the default too, seems to work and will hide the users api key a bit better.

plugins/inputs/openweathermap/openweathermap.go
```diff
@@ -0,0 +1,291 @@
...
+		status.City.Name = e.Name
+		status.Gather(tags, acc)
+	}
+	return nil
+}
+
+func (s *Status) Gather(tags map[string]string, acc telegraf.Accumulator) {
+	tags["city_id"] = strconv.FormatInt(s.City.Id, 10)
+	for _, e := range s.List {
+		tm := time.Unix(e.Dt, 0)
```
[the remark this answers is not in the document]
regel: > This appears to be the timestamp being forecasted, as opposed to the time the forecast was made. As such, we probably shouldn't use it as the timestamp, since this will make it impossible to track changes in the forecast over time since the forecast will be overwritten each time.

It's intentional: if you visualize the data it's useful to always have the most up to date info on screen.

One way we can compare the changes in the forecast without losing information is tagging. If we tag each data point returned by the API:

- First point (+3h) is tagged `forecast=3h`
- Second point (+6h) is tagged `forecast=6h`
- etc

Eventually, one can select the current weather, the `forecast=24h`, and `forecast=48h`, this makes three time series that are easy to visualize and compare.

plugins/inputs/openweathermap/README.md
regel: OK good idea!

One tiny issue is that gather seems to conflict (`type OpenWeatherMap has both field and method named Gather`) with the function name, so I'm naming the parameter `fetch`

plugins/inputs/openweathermap/openweathermap.go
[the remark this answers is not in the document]
regel: Changes in the last commit. I'm running the plugin for a couple of hours to show the results.

plugins/inputs/openweathermap/openweathermap.go
```diff
@@ -0,0 +1,291 @@
...
+		return fmt.Errorf("%s returned HTTP status %s", addr.String(), resp.Status)
+	}
+	contentType := strings.Split(resp.Header.Get("Content-Type"), ";")[0]
+	switch contentType {
+	case "application/json":
+		body, _ := ioutil.ReadAll(resp.Body)
+		resp.Body = ioutil.NopCloser(bytes.NewBuffer(body))
+		err = gatherStatusUrl(bufio.NewReader(resp.Body), tags, acc)
+		resp.Body = ioutil.NopCloser(bytes.NewBuffer(body))
+		err = gatherWeatherUrl(bufio.NewReader(resp.Body), tags, acc)
```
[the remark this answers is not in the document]
regel: Solved. It's a lot easier since `group` and `forecast` both return a list with the same info.

plugins/inputs/openweathermap/README.md
```diff
@@ -0,0 +1,67 @@
...
+When run with:
+```
+./telegraf -config telegraf.conf -input-filter openweathermap -test
+```
+
+It produces data similar to:
+```
+> weather,city_id=4303602,forecast=false humidity=51i,pressure=1012,rain=0,temperature=16.410000000000025,wind_deg=170,wind_speed=2.6 1556393944000000000
+> weather,city_id=2988507,forecast=false humidity=87i,pressure=1020,rain=0,temperature=7.110000000000014,wind_deg=260,wind_speed=5.1 1556393841000000000
+> weather,city_id=2988507,forecast=true humidity=69i,pressure=1020.38,rain=0,temperature=5.650000000000034,wind_deg=268.456,wind_speed=5.83 1556398800000000000
```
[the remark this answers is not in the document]
regel: There is only one weather, whether it's in the past, present or future so I'd like to keep one unique `weather` measurement.

plugins/inputs/openweathermap/openweathermap.go
regel: I've run more tests today. `temp_min` and `temp_max` are nearly equal to `temp` all the time, as suggested in OWM documentation.

I could not find in the API responses the temp {} structure mentioned in the docs.

regel:
Daniel, Greg, Russ, I believe all the key points are resolved in the last revision. Thank you for your excellent feedback, much appreciated!

You might want to check the `AssertContainsTaggedFields`function in detail. The implementation did not match the semantics: the code did not check all metrics, compared to `AssertDoesNotContainsTaggedFields` that calls `reflect.DeepEqual` and therefore continues to check the next metric. I've done a minor update for the sake of correctness and being able to test this condition.

regel:
openweathermap-forecasts

A quick screenshot that compares the 5d forecast (blue line, tag forecast=*), with t+6h forecast, and t+24h forecast.

The forecast=* series is overwritten every time the plugin collects data. Having one series that contains all future data points helps to maintain the initial intention.

All other t+x series keep a snapshot of how accurate the "+x" forecast was and they can be compared with current weather info. Unfortunately, I don't know how to select "empty" tag values and it's not obvious in the UI. The "current" weather to close this loop could have the forecast="0h" tag, or forecast=* meaning the current weather intentionally overwrites data in the present. In the latter scenario, a single query using forecast=* can return past, present, and future data seamlessly.

regel:
image

Coffee time weathermen? What do we do now?

plugins/inputs/openweathermap/openweathermap.go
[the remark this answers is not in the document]
danielnelson: This seems like a really good solution, makes it very easy to line up all the data for a forecasted time.

plugins/inputs/openweathermap/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+When run with:
+```
+./telegraf -config telegraf.conf -input-filter openweathermap -test
+```
+
+It produces data similar to:
+```
+> weather,city_id=4303602,forecast=false humidity=51i,pressure=1012,rain=0,temperature=16.410000000000025,wind_degrees=170,wind_speed=2.6 1556393944000000000
+> weather,city_id=2988507,forecast=false humidity=87i,pressure=1020,rain=0,temperature=7.110000000000014,wind_degrees=260,wind_speed=5.1 1556393841000000000
```
danielnelson: First two lines of this sample are with the old forecast tag style.

danielnelson:
> You might want to check the AssertContainsTaggedFieldsfunction in detail. 

Looks good, it seems like we only need your corrected implementation?  If so can you update it so we don't have the nested version.

> Unfortunately, I don't know how to select "empty" tag values and it's not obvious in the UI.

You can use queries like these to select tags with or without a value, but it is a bit non-intuitive so we generally try to avoid them.  I'm also not sure how well these forms work with different visualization tools.
```
select * from weather where forecast =~ /./
select * from weather where forecast !~ /./
```

I'll let you decide if you prefer this over `forecast=*`

regel:
> > You might want to check the AssertContainsTaggedFieldsfunction in detail.
> 
> Looks good, it seems like we only need your corrected implementation? If so can you update it so we don't have the nested version.
> 

Hi Daniel, I've used a nested call for backward compatibility. I assume that the function was written this way for developer happiness. It's nice to have the `diff` printed out when points are not strictly equal, that's the reason why I used the nested call. I let you decide if there is a better way to write this.

> ```
> select * from weather where forecast =~ /./
> select * from weather where forecast !~ /./
> ```
> 
> I'll let you decide if you prefer this over `forecast=*`

I've pushed one last commit, `forecast=*` is easier to use with the UI tools.

Have a good day!

plugins/inputs/openweathermap/openweathermap.go
```diff
@@ -0,0 +1,291 @@
...
+	resp, err := n.client.Get(addr.String())
+
+	if err != nil {
+		return fmt.Errorf("error making HTTP request to %s: %s", addr.String(), err)
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("%s returned HTTP status %s", addr.String(), resp.Status)
+	}
+	contentType := strings.Split(resp.Header.Get("Content-Type"), ";")[0]
```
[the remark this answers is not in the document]
glinton: > This is a bit strict with respect to whitespace, I suggest just checking strings.Contains(resp.Header.Get("Content-Type"), "application/json"). For the error message you can just return the full header contents.

Agreed

plugins/inputs/openweathermap/openweathermap.go
```diff
@@ -0,0 +1,305 @@
...
+			var u *url.URL
+			var addr *url.URL
+
+			for _, city := range n.CityId {
+				u, err = url.Parse(fmt.Sprintf("/data/2.5/forecast?id=%s&APPID=%s&units=%s", city, n.AppId, units))
+				if err != nil {
+					acc.AddError(fmt.Errorf("Unable to parse address '%s': %s", u, err))
+					continue
+				}
+				addr = base.ResolveReference(u)
```
glinton: Since this is always used with `.String()` change `addr` to a string and set `addr = base.ResolveReference(u).String()`, passing just the string into `gatherUrl`

same for line 106, 120, 122, etc...

plugins/inputs/openweathermap/openweathermap.go
```diff
@@ -0,0 +1,305 @@
...
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("%s returned HTTP status %s", addr.String(), resp.Status)
+	}
+	contentType := strings.Split(resp.Header.Get("Content-Type"), ";")[0]
+	switch contentType {
+	case "application/json":
+		err = gatherWeatherUrl(bufio.NewReader(resp.Body), forecast, acc)
+		return err
```
glinton: nit: just `return gatherWeatherUrl(bufio.NewReader(resp.Body), forecast, acc)` instead of assigning it to `err` first.

plugins/inputs/openweathermap/openweathermap.go
```diff
@@ -0,0 +1,305 @@
...
+	default:
+		return fmt.Errorf("%s returned unexpected content type %s", addr.String(), contentType)
+	}
+}
+
+type WeatherEntry struct {
```
glinton: Remove `Dttxt`, `Clouds`, `GrndLevel`, `SeaLevel`, `TempMax`, `TempMin`, `Sys`, `Weather`, `Coord`, and `Visibility` as they are unused.

plugins/inputs/openweathermap/openweathermap.go
```diff
@@ -0,0 +1,305 @@
...
+	Id    int64  `json:"id"`
+	Name  string `json:"name"`
+	Coord struct {
+		Lat float64 `json:"lat"`
+		Lon float64 `json:"lon"`
+	} `json:"coord"`
+	Visibility int64 `json:"visibility"`
+}
+
+type Status struct {
```
glinton: Remove `Coord` and `Country`as they are unused. Also `Name` if it's not going to be added as a tag (could always be removed with [`tagexclude`](https://github.com/influxdata/telegraf/blob/master/docs/CONFIGURATION.md#modifiers)).

plugins/inputs/openweathermap/openweathermap.go
```diff
@@ -0,0 +1,305 @@
...
+		Name    string `json:"name"`
+	} `json:"city"`
+	List []WeatherEntry `json:"list"`
+}
+
+func gatherWeatherUrl(r *bufio.Reader, forecast bool, acc telegraf.Accumulator) error {
+	dec := json.NewDecoder(r)
+	status := &Status{}
+	if err := dec.Decode(status); err != nil {
+		return fmt.Errorf("Error while decoding JSON response: %s", err)
```
glinton: nit: start with lowercase on returned errors -> `return fmt.Errorf("error while...`

testutil/accumulator.go
```diff
@@ -342,6 +340,26 @@ func (a *Accumulator) AssertContainsTaggedFields(
 	assert.Fail(t, msg)
 }
 
+func (a *Accumulator) AssertContainsTaggedFields(
```
glinton: What is the reason for this change? If it's crufty, please revert this file.

DRailleur:
Hi,

Could you add a "lang" parameter in order to collect a localized description field ?

Thanks

danielnelson:
@DRailleur Could you open a new feature request issue?  If you can include a link to the relevant API documentation that would be helpful as well.

plugins/inputs/openweathermap/openweathermap.go
PatrickGlatz: Is it not possible to reintroduce the especially `temp_min` and `temp_max` for the forecast? And do you have an idea by when  `snow_1h` / `snow_3h` will be revised again for usage? You guys mentioned you would like to have a look at the JSON reply when it is returned by openweathermap.

Finally, why are all `Weather` (including `weather_id`, `weather_main`, `weather_description`, `weather_icon`) not showing up in influxdb although the values are returned in the json from openweathermap? From what i saw was, that these metrics are included if the length of the list is greater 0, meaning that normally this should be incorporated in the influx entry. This is not the case for me. Can you give me a hint? 

Thanks in advance!

PatrickGlatz:
> openweathermap-forecasts
> 
> A quick screenshot that compares the 5d forecast (blue line, tag forecast=*), with t+6h forecast, and t+24h forecast.
> 
> The forecast=* series is overwritten every time the plugin collects data.
> 
> All other t+x series keep a snapshot of how accurate the "+x" forecast was and they can be compared with current weather info. Unfortunately, I don't know how to select "empty" tag values and it's not obvious in the UI. The "current" weather to close this loop could have the forecast="0h" tag, or forecast=* meaning the current weather intentionally overwrites data in the present. In the latter scenario, a single query using forecast=* can return past, present, and future data seamlessly.

But this seems not to be the current implementation now, does it? I am having massively problems making a correct query on influxdb. The plugin writes for every forecast timepoint a different forecast string (e.g. 120h, 117h, 114h, 111h, ...), whereas '*' does only illustrate the current weather. For me, it is not possible right now to query influxdb with the latest forecast of the weather. Selecting all the 120h values is not correct as the only the farest point is valid. The other points would have to have tags 117h, 114h, 111h, ...

How does a valid query look like, which shows the latest forecast? Thx.

danielnelson:
The timestamps are set to the forecasted time.  So, for example, the 3 hour forecast `forecast=3h` would have the timestamp approx 3 hours in the future.

It sounds like you are wanting to grab all the latest forecasts, I think this should be the last point in each series.  You could ignore the `forecast=*` series or exclude it by adding another where clause.
```
select * from weather where time > now() - 7d group by forecast order by desc limit 1
```

> I would have expected, that an additional tag with entry timestamp would be beneftitial for clearly identifying, when a value was entered, something like request_time.

This might be nice, but it would significantly increase the cardinality of the series, as this tag would continually have new timestamps.

PatrickGlatz:
Hey @danielnelson, 

that is exactly what i want and i had that already but was not able to visualize this in grafana. Any hints or ideas how to manage this?

Happy to hear from you (or others) soon,

Thanks in advance
Patrick

Btw: sorry for the late replay, was on business trip.

danielnelson:
I'm not really sure what the best way to view this in Grafana would be, but I suspect you would have best luck with a table view.  I'd be open to ideas on schemas that would work better too.

I see your [comment][] on the community site, we should probably use that thread for discussion until we have something more concrete, in which case it would make sense to create a new issue.

[comment]: https://community.influxdata.com/t/grafana-plotting-openweathermap-forecast-from-influxdb-telegraf/12029

Issue #2776, opened 2017-05-08
[feature-request] Google PubSub Producer/Consumer


jackzampolin:
## Feature Request

Opening a feature request kicks off a discussion.

### Proposal:

Telegraf currently has producer and consumer plugins for a number of different queueing solutions. Google PubSub, a widely used event bus, is not currently supported.

### Use case: 

Lots of sensor monitoring and other use cases utilize Google PubSub for metrics shipping. Native hooks into that from telegraf would make it much easier to use with the InfluxDB ecosystem.

[Golang Bindings](https://github.com/GoogleCloudPlatform/golang-samples/tree/master/pubsub)

gunnaraasen:
@danielnelson I think this can be closed now that both the input and output are merged.

PR #5136, opened 2018-12-11
PubSub Input Plugin

emilymye:
### Required for all PRs:

Related to #2776


emilymye:
I think the biggest things I wanted feedback on are:

- _Supporting input topics with/instead of subscriptions_: Users wouldn't have to create their own subscription to use with Telegraf, but users may want to have longer-lasting subscriptions and cleanup would be mandatory. If a topic has a subscription, it will not clear a message until all subscriptions have each ACKed. I think if we do want to support both, at the very least preventing people from using both `topics` and `subscriptions` in config at the same time.
- _Single vs multiple topics/subscriptions_: Should we support multiple topics/subs? Or should each agent only pull from one and multiple subscriptions have multiple agents?

plugins/inputs/pubsub/pubsub.go
```diff
@@ -0,0 +1,290 @@
...
+	defaultMaxUndeliveredMessages = 1000
+)
+
+type PubSub struct {
+	CredentialsFile string   `toml:"credentials_file"`
+	Project         string   `toml:"google_project"`
+	Subscriptions   []string `toml:"subscriptions"`
+
+	// Subscription ReceiveSettings
+	MaxExtensionSec        int `toml:"max_extension_seconds"`
```
emilymye: If there is a way to TOML a duration, please let me know!

plugins/inputs/pubsub/pubsub.go
```diff
@@ -0,0 +1,290 @@
...
+			MaxOutstandingBytes:    ps.MaxOutstandingBytes,
+		}
+		ps.subRefs[subRef.ID()] = subRef
+	}
+
+	return nil
+}
+
+func gcpUserAgentString() string {
+	return fmt.Sprintf("InfluxData Telegraf/%s (%s)", internal.Version(), runtime.Version())
```
emilymye: I was hoping to reuse this as well as some of the client setup code for the output plugin. It might make sense to also use this in existing GCP plugins (i.e. Stackdriver). Is there somewhere I can add shared utils?

plugins/inputs/pubsub/pubsub.go
danielnelson: You can use `internal.Duration`, here is [an example](https://github.com/influxdata/telegraf/blob/master/plugins/inputs/http/http.go#L30).

plugins/inputs/pubsub/pubsub.go
danielnelson: We have [something similar](https://github.com/influxdata/telegraf/blob/master/plugins/outputs/influxdb/http.go#L139) in a few plugins such as the influxdb output.  If you like you can add a function to `internal/internal.go` `func ProductToken() string` and have it return `Telegraf/1.9.1`.  

The idea here is to send a [http product token](https://tools.ietf.org/html/rfc2616#section-3.8), I wouldn't mind also including the Go version: `Telegraf/1.9.1 Go/1.11.3` however I don't see a very good way to get the Go version string into this format.

danielnelson:
> Supporting input topics with/instead of subscriptions

Would this require us to make a subscription on the topic in Telegraf and clean it up afterwards?

> Single vs multiple topics/subscriptions

I think this depends on how many subscriptions we thing a typical user might need to subscribe to.  If it is more likely that they will subscribe to only a few then it makes more sense IMO to allow a single subscription per plugin.  This will ensure that there is no unwanted interactions between the different subscriptions as each will be in its own plugin.

If you expect that users may need to subscribe to dozens of topics/subscriptions it may be better to allow them in a single plugin.  In the mqtt_consumer, for example, the topic by convention encodes information about the reading, and so you often need to subscribe to many topics using a wildcard subscription.

emilymye:
> Would this require us to make a subscription on the topic in Telegraf and clean it up afterwards?
> 
Yeah, this sounds like more trouble than it is worth? In addition to clean-up, users might not want their telegraf GCP credentials to have access to subscription creation/deletion
> > Single vs multiple topics/subscriptions

> If it is more likely that they will subscribe to only a few then it makes more sense IMO to allow a single subscription per plugin.

We talked a little about this and it seems like users tend to use fewer, so I'll probably switch to one subscription.

> If you expect that users may need to subscribe to dozens of topics/subscriptions it may be better to allow them in a single plugin. In the mqtt_consumer, for example, the topic by convention encodes information about the reading, and so you often need to subscribe to many topics using a wildcard subscription.

danielnelson:
> > Would this require us to make a subscription on the topic in Telegraf and clean it up afterwards?

> Yeah, this sounds like more trouble than it is worth? In addition to clean-up, users might not want their telegraf GCP credentials to have access to subscription creation/deletion

It may not be possible to do the cleanup always as well, such as in the case of a lost network connection or sigkill.  If the subscription creation is automatic there will be a expectation that cleanup will automatic, and if failure to do so is a problem (especially if it involves any service fees) we may not want to involve it.  If we have good instructions in the README and appropriate error messages when it the subscription does not exist, I think it won't be a large hurdle to setup manually.

emilymye:
@danielnelson I made the discussed changes, added some tests and a README. I think it's in a good place to start review. 

Note: I added wrappers for the PubSub subscription/messages. I don't really like creating structs just to mock them out for tests, but this very annoying private function in [Google pubsub.Message](https://github.com/googleapis/google-cloud-go/blob/master/pubsub/message.go#L55) causes a panic on Ack() if we use a externally-generated Message

plugins/inputs/pubsub/pubsub.go
```diff
@@ -0,0 +1,291 @@
+package pubsub
```
danielnelson: We should rename this plugin to qualify it further since we also want would support other pub/sub systems and the name `pubsub` is too generic.  A couple name ideas I have are `cloud_pubsub`, `google_pubsub`, `google_cloud_pubsub`.

We should also consider how a push delivery endpoint should be named (#4700).  Since these two methods don't have much shared configuration and the API is completely different, I think they should remain separate plugins.  With this in mind maybe we should have `cloud_pubsub_pull` and `cloud_pubsub_push`?  If you think it is better to emphasize one of these more then maybe it should lopsided: `cloud_pubsub` and `cloud_pubsub_push`.

plugins/inputs/pubsub/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+### Configuration
+
+This section contains the default TOML to configure the plugin.  You can
+generate it using `telegraf --usage pubsub`.
+
+```toml
+[[inputs.pubsub]]
+  ## Name of Google Cloud Platform (GCP) Project owning PubSub subscriptions
+  google_project = "my-project"
```
danielnelson: I think name this just `project`.

plugins/inputs/pubsub/pubsub.go
```diff
@@ -0,0 +1,291 @@
...
+	}
+
+	if len(metrics) == 0 {
+		msg.Ack()
+		return nil
+	}
+
+	if len(ps.SubscriptionTag) > 0 {
+		for _, m := range metrics {
+			m.AddTag(ps.SubscriptionTag, ps.sub.ID())
```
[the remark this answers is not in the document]
emilymye: Subscription name; Essentially the last slash-delimited token in the full name`projects/$PROJ/subscriptions/$SUB` (see [code](https://github.com/googleapis/google-cloud-go/blob/master/pubsub/subscription.go#L61) -  ID() just splits after the last `/`). I might switch this to use `ps.sub.String()` here for the resource name.

plugins/inputs/pubsub/pubsub.go
emilymye: I think google_pubsub and google_pubsub_push work - @gunnar @josephlewis42 opinions?

plugins/inputs/pubsub/README.md
```diff
@@ -0,0 +1,86 @@
...
+  
+  ## Maximum number of seconds for which a PubSub subscription
+  ## should auto-extend the PubSub ACK deadline for each message. If less than
+  ## 0, auto-extension is disabled.
+  # max_extension = 0
+
+  ## Maximum number of unprocessed messages in PubSub (unacknowledged but not 
+  ## yet expired in PubSub). A value of 0 is treated as the default 
+  ## PubSub value. Negative values will be treated as unlimited.
+  # max_outstanding_messages = 0
```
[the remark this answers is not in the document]
emilymye: yeah, i think it might make sense if you want to pipe more through telegraf but are limited by the default amount in PubSub, so you just set this value to very large and then manage it in Telegraf instead? This also might be a "it's there so might as well" situation

plugins/inputs/pubsub/pubsub.go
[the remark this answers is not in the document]
danielnelson: Just the subscription name would be preferred, as it keeps the project/subscription data in a normalized form and there is little performance to gain by combining them.

Considering that we have settled on the plugin supporting only a single project and subscription perhaps we should instead use the tags option.  This method for adding additional tags can be used with any input plugin.  One drawback is that some settings must be repeated, but it gives full control over the tags and can be extended easily, without introducing any options that need to be handled by the plugin itself.
```toml
[[inputs.google_pubsub]]
  project = "foo"
  subscription = "bar"
  [inputs.google_pubsub.tags]
    project = "foo"
    subscription = "bar"
```

Where this method could not work is if the tag values are dynamically set by the plugin, but this is not something that comes up in this plugin.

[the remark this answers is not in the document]
emilymye: Sounds good! I'll remove subscription_tag

emilymye:
> One high level issue with the tests is that since they use unexported methods to control the plugin, the coverage is quite low.

Yeah, I pretty much stole the kafka tests for this and should probably have tried harder :/ Part of this is that the PubSub libraries are notoriously hard to test because of how they expose things, and their mock library for [testing](https://godoc.org/cloud.google.com/go/pubsub/pstest) just...doesn't work. The only way to do non-integration tests involves lots of mocking/interface fun, especially for testing the public methods on ServiceInput. I'll take a stab at it.

plugins/inputs/pubsub/pubsub.go
josephlewis42: I think Google's branding guidelines would recommend `cloud_pubsub` or `google_cloud_pubsub`.

I like the idea of the lopsided pull vs push, especially with pull being the more common case. That should get them sorted nicely lexicographically for documentation too.

The hardest part of any project is naming :grin:

emilymye:
@danielnelson thank you for the initial review! Would you mind taking a look at my new changes when you get a chance? also adding @gunnaraasen

fuzzyami:
Hey guys, apologies for snatching the discussion here. thanks for this much-needed output plugin. I do have a question though: the default config for this plugin suggests sending the metrics in the influx format to pubsub. I know (because I tested it) that the json format is correctly consumed, but I am wondering whether influx db format works as well. In addition, I opened a PR (https://github.com/influxdata/telegraf/pull/5378) for some other changes to the default config.

danielnelson:
@fuzzyami It should work, just set the default `data_format = "influx"` to output in InfluxDB line protocol, let us know if you run into any difficulties.  Thanks for the fix to the sample config!

andrekeane:
Hey, great work on the plugin. Works great. Quick question: is there any way to add pubsub metadata (mainly attributes) to the cloud_pubsub (pull) plugin? I can only seem to find a config option in the push version (AddMeta).

PR #5161, opened 2018-12-19
Wavefront output enhancements

puckpuck:
### Required for all PRs:

Updated connection and communication to use Wavefront's [go-sdk](https://github.com/wavefrontHQ/wavefront-sdk-go).  Benefits include:
- Improved performance for higher throughput
- Can be used with Wavefront Direction Ingestion API (https)
- Wavefront proxy can still be used, and http is now supported (in addition to tcp)

Also marked string_to_number feature as deprecated since this is now done using the enum plugin.  Goal is to remove it completely in Telegraf 2.


plugins/outputs/wavefront/wavefront_test.go
```diff
@@ -264,6 +269,7 @@ func TestBuildValueString(t *testing.T) {
 
 }
 
+/*
```
danielnelson: Delete or update this test as appropriate.

plugins/outputs/wavefront/wavefront.go
```diff
@@ -92,44 +98,55 @@ type MetricPoint struct {
...
-
 	for _, m := range metrics {
-		for _, metricPoint := range buildMetrics(m, w) {
-			metricLine := formatMetricPoint(metricPoint, w)
-			_, err := connection.Write([]byte(metricLine))
+		for _, point := range buildMetrics(m, w) {
+			/*
+				metricLine := formatMetricPoint(point, w)
+				_, err := connection.Write([]byte(metricLine))
+			*/
```
danielnelson: Remove commented out code.

plugins/outputs/wavefront/wavefront.go
```diff
@@ -40,43 +40,49 @@ var sanitizedChars = strings.NewReplacer(
...
-  host = "wavefront.example.com"
+  ## Url for Wavefront Direct Ingestion or using HTTP with Wavefront Proxy
+  ## If using Wavefront Proxy, also specify port. example: http://proxyserver:2878
+  url = "https://metrics.wavefront.com"
+
+  ## Authentication Token for Wavefront. Only required if using Direct Ingestion
+  #token = "DUMMY_TOKEN"  
+  
+  ## DNS name of the wavefront proxy server. Do not use if url is specified
+  #host = "wavefront.example.com"
```
danielnelson: Is this option deprecated now in favor of url?

Gopkg.toml
```diff
@@ -246,6 +246,10 @@
   name = "github.com/vishvananda/netlink"
   revision = "b2de5d10e38ecce8607e6b438b6d174f389a004e"
 
+[[constraint]]
+  name = "github.com/wavefronthq/wavefront-sdk-go"
+  version = "v0.9.0"
```
danielnelson: There should be some changes to Gopkg.lock as well, which is why the tests aren't working, just need to run `dep ensure` once.

plugins/outputs/wavefront/wavefront.go
```diff
@@ -92,44 +98,55 @@ type MetricPoint struct {
...
+	if len(w.StringToNumber) > 0 {
+		log.Print("W! [outputs.wavefront] The string_to_number option is deprecated; please use the enum processor instead")
+	}
+
+	if w.Url != "" {
+		log.Printf("D! [outputs.wavefront] connecting over http/https using Url: %s", w.Url)
+		sender, err := wavefront.NewDirectSender(&wavefront.DirectConfiguration{
+			Server:               w.Url,
+			Token:                w.Token,
+			FlushIntervalSeconds: 5,
```
danielnelson: Telegraf considers returning from Write() without error to be successful delivery of metrics, but with the extra layer of buffering we don't actually know if the metric was sent.  There may be some reduction in durability because of this, might be worth considering a sender implementation that sends immediately without buffering?

plugins/outputs/wavefront/wavefront_test.go
puckpuck: Should TestBuildValueString should remain since the functionality is deprecated but not removed yet?

Did delete commented out test below (TestFormatMetricPoint)

danielnelson: We can remove TestBuildValueString once the option is removed, for now its best to leave it.

Issue #3900, opened 2018-03-17
enhancement - inputs/system/diskio_linux - configurable udev data path


ryanbreed:
from the nightly commit 76ce71f, the linux diskio plugin collects its info cache from a hard-coded path:

https://github.com/influxdata/telegraf/blob/76ce71f7fa0ef46a700be4e51826b6be66dd8edd/plugins/inputs/system/diskio_linux.go#L17

this seems to vary by distribution - on amzn-linux 1 at least, this is available in `/dev/.udev/data`, and cannot be configured via udevd. Here's the base AMI I'm using:

    image_name="amzn-ami-hvm"
    image_version="2017.09"
    image_arch="x86_64"
    image_file="amzn-ami-hvm-2017.09.1.20171120-x86_64.ext4.gpt"
    image_stamp="03dc-49fe"
    image_date="20171120221247"
    recipe_name="amzn ami"
    recipe_id="c9c871d2-7d71-8baa-e873-c70f-cbbe-ddb2-79ee86e7"

ryanbreed:
@danielnelson - although i'm new to go, I think I could throw a pr your way if you'd let me know what the expectations would be on the testing side. I think that the current implementation fails [gracefully](https://github.com/influxdata/telegraf/blob/76ce71f7fa0ef46a700be4e51826b6be66dd8edd/plugins/inputs/system/diskio_linux.go#L24) [enough](https://github.com/influxdata/telegraf/blob/76ce71f7fa0ef46a700be4e51826b6be66dd8edd/plugins/inputs/system/diskio_linux.go#L48) since ENOENT/EACCESS/etc get propagated already (obvs), and don't cause any uncontained instability as far as I can tell.

I think it's a one-line addition to the [DiskIO Type](https://github.com/influxdata/telegraf/blob/76ce71f7fa0ef46a700be4e51826b6be66dd8edd/plugins/inputs/system/diskio.go#L18) to relocate the module-scoped var into the config if I'm reading things right. The rest would be updating the info cache population to pull from the config type [instead of the module-scoped var](https://github.com/influxdata/telegraf/blob/76ce71f7fa0ef46a700be4e51826b6be66dd8edd/plugins/inputs/system/diskio_linux.go#L39). 

I'd be pleased as punch to squish this one unless other plans are in the mix.

danielnelson:
[...]

PR #5180, opened 2018-12-21
Added support for older non-systemd udev mount points.

red15:
Solving issue #3900

### Required for all PRs:

### DIFF
--- a/plugins/inputs/diskio/diskio_linux.go
+++ b/plugins/inputs/diskio/diskio_linux.go
@@ -11,12 +11,9 @@ import (
 )
 
 type diskInfoCache struct {
-	modifiedAt   int64 // Unix Nano timestamp of the last modification of the device. This value is used to invalidate the cache
-	udevDataPath string
-	values       map[string]string
+	modifiedAt int64 // Unix Nano timestamp of the last modification of the device. This value is used to invalidate the cache
+	values     map[string]string
 }
-
-var udevPath = "/run/udev/data"
 
 func (s *DiskIO) diskInfo(devName string) (map[string]string, error) {
 	var err error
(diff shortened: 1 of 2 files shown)


plugins/inputs/diskio/diskio_linux.go
```diff
@@ -15,7 +16,28 @@ type diskInfoCache struct {
 	values       map[string]string
 }
 
-var udevPath = "/run/udev/data"
+var udevPath = []string{"/run/udev/data", "/dev/.udev/db"}
+
+func getUdevDataPath(devName string, major uint64, minor uint64) (out string, err error) {
```
danielnelson: So there are 4 paths can will be checked, for example:
- /run/udev/data/b8:0
- /run/udev/data/block:vda
- /dev/.udev/db/b8:0
- /dev/.udev/db/block:vda

Is this correct, I thought there was only these 2 that needed checked:
- /run/udev/data/b8:0
- /dev/.udev/db/block:vda

plugins/inputs/diskio/diskio_linux.go
```diff
@@ -15,7 +16,28 @@ type diskInfoCache struct {
...
+	for _, udev := range udevPath {
+		// First try the systemd-path
+		var path = fmt.Sprintf("%s/b%d:%d", udev, major, minor)
+		f, err := os.Open(path)
```
danielnelson: To avoid needing to open the file twice, lets have this function return the contents of the file.  It looks like we don't actually need the `udevDataPath` in the `diskInfoCache`.

plugins/inputs/diskio/diskio_linux.go
red15: You are correct that in a worst case scenario there are 4 path's that are being checked, this is because we cannot link the `b8:0` or `block:vda` to the udevPath[0] or udevPath[1] entry.

I experimented with hardcoding the paths and device format but this breaks the test scenarios.

plugins/inputs/diskio/diskio_linux.go
red15: If indeed the `udevDataPath` is not needed inside the `diskInfoCache` then returning the content would indeed be a slightly better flow, I couldn't judge if it was required or not but decided to minimize the changes.

PR #5191, opened 2018-12-25
Issue #5164 Input plugin for Neptune Apex aquarium controller.

### LINKED ISSUE #5164: Input plugin for Neptune Apex's XML data
I'm planning to implement this and want to see if a plugin aimed at a niche market would be accepted for inclusion in the telegraf project. There are supposedly 800k Apex in the wild owned by somewhat technology-savvy people but not necessary savvy enough to setup a full stack without a comprehensive guide. 

### Proposal:
I would like to write a plugin that parses the XML output of a Neptune Apex aquarium controller. An Apex acts as a life support controller for complex aquarium management. It can be used to gather various chemical data (pH, salinity, temperature, etc), gather energy data per outlet (Current status, output in A, output in W). The Apex controller itself can control devices directly connected to it but the offering is somewhat limited and my ultimate goal is to scrape the live status.xml page and publish it via MQTT.

### Current behavior:
Telegraf does not support the Apex, nor XML collection

### Desired behavior:
The ideal situation would be to implement https://github.com/influxdata/telegraf/issues/1758 however a universal XML input is harder to implement correctly when it must work with any valid XML. For example, how to differentiate between attributes and sub-objects on the line protocol?
I would like this plugin to be specific to the Neptune Apex so some assumptions will made.

With very few exceptions, the Apex controller can only control what is directly attached to its smart power switch. Allowing this input opens its users to a plethora of smart reporting features and more importantly, home automation with MQTT.
The current method of cooling down an aquarium revolves around starting a fan above it to trigger evaporative cooling and if that doesn't work, turn a chiller on. Chillers are somewhat inefficient and since they're usually sitting by the aquarium, they raise the air temperature to lower the water temperature. It would be better to control a smart thermometer and cool the house with an exterior condenser. Other use might include starting an air exchanger when the pH of the aquarium drops (Indicating a somewhat higher CO2 level in the air).

MaxRenaud:
### Required for all PRs:

- [ ✓] Signed [CLA](https://influxdata.com/community/cla/).
- [ ✓] Associated README.md updated.
- [ ✓] Has appropriate unit tests.

This PR closes #5164 and adds an input plugin for the Neptune Apex aquarium controller. General format was taken from the "httpjson" input plugin.
I've been running this at home for 24 hours and all stats look good.

### DIFF
--- a/plugins/inputs/neptune_apex/README.md
+++ b/plugins/inputs/neptune_apex/README.md
@@ -1,0 +1,145 @@
+# neptune_apex Input Plugin
+
+The neptune_apex Input Plugin collects real-time data from the Apex's status.xml page.
+
+### Configuration
+
+```toml
+[[inputs.neptune_apex]]
+  ## The Neptune Apex plugin reads the publicly available status.xml data from a local Apex.
+  ## Measurements will be logged under "apex".
+
+  ## The hostname/IP of the local Apex(es). If you specify more than one server, they will
+  ## be differentiated by the "hostname" tag.
+  servers = [
+    "apex.local",
+  ]
+
+  ## The response_timeout specifies how long to wait for a reply from the Apex.
+  #response_timeout = "5s"
+
+```
+
+### Metrics
+
+The [Neptune Apex](https://www.neptunesystems.com/) controller family allows an aquarium hobbyist to monitor and control
+their tanks based on various probes. The data is taken directly from the /cgi-bin/status.xml at the interval specified 
+in the telegraf.conf configuration file. 
+
+No manipulation is done on any of the fields to ensure future changes to the status.xml do not introduce conversion bugs
+to this plugin. When reasonable and predictable, some tags are derived to make graphing easier and without front-end 
+programming. These tags are clearly marked in the list below and should be considered a convenience rather than authoritative.
+
+- neptune_apex (All metrics have this measurement name)
+  - tags:
+    - host (mandatory, string) is the host on which telegraf runs.
+    - hostname (mandatory, string) contains the hostname of the apex device. This can be used to differentiate between 
+    different units. By using the hostname instead of the serial number, replacements units won't disturb graphs.
+    - type (mandatory, string) maps to the different types of data. Values can be "controller" (The Apex controller 
+    itself), "probe" for the different input probes, or "output" for any physical or virtual outputs. The Watt and Amp 
+    probes attached to the physical 120V outlets are aggregated under the output type.
+    - probe_type (optional, string) contains the probe type as reported by the Apex.
+    - name (optional, string) contains the name of the probe or output.
+    - output_id (optional, string) represents the internal unique output ID. This is different from the device_id.
+    - device_id (optional, string) maps to either the aquabus address or the internal reference.
+    - output_type (optional, string) categorizes the output into different categories. This tag is DERIVED from the 
+    device_id. Possible values are: "variable" for the 0-10V signal ports, "outlet" for physical 120V sockets, "alert" 
+    for alarms (email, sound), "virtual" for user-defined outputs, and "unknown" for everything else.
+  - fields:
+    - value (float, various unit) represents the probe reading.
+    - state (string) represents the output state as defined by the Apex. Examples include "AOF" for Auto (OFF), "TBL" 
+    for operating according to a table, and "PF*" for different programs.
+    - amp (float, Ampere) is the amount of current flowing through the 120V outlet.
+    - watt (float, Watt) represents the amount of energy flowing through the 120V outlet.
+    - xstatus (string) indicates the xstatus of an outlet. Found on wireless Vortech devices.
+    - hardware (string, controller hardware version)
+    - software (string, software version)
+    - power_failed (string, date time) when the controller last lost power.
+    - power_restored (string, date time) when the controller last powered on.
+    - serial (string, serial number)
+    - timezone (float, timezone offset)
+   - time:
+     - The time used for the metric is parsed from the status.xml page. This helps when cross-referencing events with 
+     the local system of Apex Fusion. Since the Apex uses NTP, this should not matter in most scenarios. 
+     
+
+### Sample Queries
+
+
+Get the max, mean, and min for the temperature in the last hour:
+```
+SELECT mean("value") FROM "neptune_apex" WHERE ("probe_type" = 'Temp') AND time >= now() - 6h GROUP BY time(20s)
+```
+
+### Troubleshooting
+
+#### sendRequest failure
+This indicates a problem communicating with the local Apex controller. If on Mac/Linux, try curl:
+```
+$ curl apex.local/cgi-bin/status.xml
+```
+to isolate the problem.
+
+#### parseXML errors
+Ensure the XML being returned is valid. If you get valid XML back, open a bug request.
+
+#### Missing fields/data
+The neptune_apex plugin is strict on its input to prevent any conversion errors. If you have fields in the status.xml
+output that are not converted to a metric, open a feature request and paste your whole status.xml
+
+### Example Output
+
+```
+> neptune_apex,host=ubuntu,hostname=apex hardware="1.0",power_failed="12/15/2018 11:00:00",power_restored="12/15/2018 16:31:15",serial="AC5:12345",software="5.04_7A18",timezone=-8 1545716958000000000
+> neptune_apex,device_id=base_Var1,host=ubuntu,hostname=apex,name=VarSpd1_I1,output_id=0,output_type=variable,type=output state="PF1" 1545716958000000000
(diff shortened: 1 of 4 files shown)


plugins/inputs/neptune_apex/README.md
```diff
@@ -0,0 +1,145 @@
...
+
+```toml
+[[inputs.neptune_apex]]
+  ## The Neptune Apex plugin reads the publicly available status.xml data from a local Apex.
+  ## Measurements will be logged under "apex".
+
+  ## The hostname/IP of the local Apex(es). If you specify more than one server, they will
+  ## be differentiated by the "hostname" tag.
+  servers = [
+    "apex.local",
```
danielnelson: Let's have this take a base url in case we want to add support for TLS later.  Even though the api is over HTTP, its possible that a user will want to use a proxy to access the server.

plugins/inputs/neptune_apex/README.md
```diff
@@ -0,0 +1,145 @@
...
+in the telegraf.conf configuration file. 
+
+No manipulation is done on any of the fields to ensure future changes to the status.xml do not introduce conversion bugs
+to this plugin. When reasonable and predictable, some tags are derived to make graphing easier and without front-end 
+programming. These tags are clearly marked in the list below and should be considered a convenience rather than authoritative.
+
+- neptune_apex (All metrics have this measurement name)
+  - tags:
+    - host (mandatory, string) is the host on which telegraf runs.
+    - hostname (mandatory, string) contains the hostname of the apex device. This can be used to differentiate between
```
danielnelson: How do you feel about calling this `source`, this is tag that we are planning to standardize on to represent the hostname of the device being monitored, see #4413.

plugins/inputs/neptune_apex/README.md
```diff
@@ -0,0 +1,145 @@
...
+    for operating according to a table, and "PF*" for different programs.
+    - amp (float, Ampere) is the amount of current flowing through the 120V outlet.
+    - watt (float, Watt) represents the amount of energy flowing through the 120V outlet.
+    - xstatus (string) indicates the xstatus of an outlet. Found on wireless Vortech devices.
+    - hardware (string, controller hardware version)
+    - software (string, software version)
+    - power_failed (string, date time) when the controller last lost power.
+    - power_restored (string, date time) when the controller last powered on.
+    - serial (string, serial number)
+    - timezone (float, timezone offset)
```
danielnelson: Consider if timezone offset is needed, generally we would just store all times in UTC, but I can imagine some cases where this might be useful (extremely coarse device location?).

plugins/inputs/neptune_apex/README.md
```diff
@@ -0,0 +1,145 @@
...
+    - value (float, various unit) represents the probe reading.
+    - state (string) represents the output state as defined by the Apex. Examples include "AOF" for Auto (OFF), "TBL" 
+    for operating according to a table, and "PF*" for different programs.
+    - amp (float, Ampere) is the amount of current flowing through the 120V outlet.
+    - watt (float, Watt) represents the amount of energy flowing through the 120V outlet.
+    - xstatus (string) indicates the xstatus of an outlet. Found on wireless Vortech devices.
+    - hardware (string, controller hardware version)
+    - software (string, software version)
+    - power_failed (string, date time) when the controller last lost power.
+    - power_restored (string, date time) when the controller last powered on.
```
danielnelson: Usually we recommend to store times as nanosecond since epoch, but have been extremely consistent on this in the past, these are more efficient to store in the database.  With our new query language "flux" it is possible to format a time in a human readable format in the query, although currently with InfluxQL you would need to do it outside the query language.

plugins/inputs/neptune_apex/README.md
```diff
@@ -0,0 +1,145 @@
...
+    for alarms (email, sound), "virtual" for user-defined outputs, and "unknown" for everything else.
+  - fields:
+    - value (float, various unit) represents the probe reading.
+    - state (string) represents the output state as defined by the Apex. Examples include "AOF" for Auto (OFF), "TBL" 
+    for operating according to a table, and "PF*" for different programs.
+    - amp (float, Ampere) is the amount of current flowing through the 120V outlet.
+    - watt (float, Watt) represents the amount of energy flowing through the 120V outlet.
+    - xstatus (string) indicates the xstatus of an outlet. Found on wireless Vortech devices.
+    - hardware (string, controller hardware version)
+    - software (string, software version)
```
danielnelson: Consider making the software and hardware version tags, unless they are frequently changing.  It should be somewhat more efficient to store and could allow grouping by version.

plugins/inputs/neptune_apex/README.md
```diff
@@ -0,0 +1,145 @@
...
+
+```
+
+### Metrics
+
+The [Neptune Apex](https://www.neptunesystems.com/) controller family allows an aquarium hobbyist to monitor and control
+their tanks based on various probes. The data is taken directly from the /cgi-bin/status.xml at the interval specified 
+in the telegraf.conf configuration file. 
+
+No manipulation is done on any of the fields to ensure future changes to the status.xml do not introduce conversion bugs
```
danielnelson: Really smart :1st_place_medal:

plugins/inputs/neptune_apex/neptune_apex.go
```diff
@@ -0,0 +1,265 @@
...
+
+	mainFields := map[string]interface{}{
+		"software":       r.SoftwareVersion,
+		"hardware":       r.HardwareVersion,
+		"serial":         r.Serial,
+		"timezone":       r.Timezone,
+		"power_failed":   r.PowerFailed,
+		"power_restored": r.PowerRestored,
+	}
+	acc.AddFields(Measurement, mainFields, map[string]string{"hostname": r.Hostname, "type": "controller"}, reportTime)
```
danielnelson: Can you line wrap around around 78 chars through the pr?

plugins/inputs/neptune_apex/neptune_apex.go
```diff
@@ -0,0 +1,265 @@
...
+		// Find Amp and Watt probes and add them as fields. Remove the redundant probe.
+		if pos := findProbe(fmt.Sprintf("%sW", o.Name), r.Probe); pos > -1 {
+			value, err := strconv.ParseFloat(strings.TrimSpace(r.Probe[pos].Value), 64)
+			if err != nil {
+				acc.AddError(
+					fmt.Errorf("cannot convert string value %q to float64: %v", r.Probe[pos].Value, err))
+				continue // Skip the whole outlet.
+			}
+			fields["watt"] = value
+			r.Probe[pos] = r.Probe[len(r.Probe)-1]
```
danielnelson: This could panic if r.Probe length is 0, is this possible such in the case of unexpected XML?

plugins/inputs/neptune_apex/README.md
MaxRenaud: Done.

plugins/inputs/neptune_apex/README.md
MaxRenaud: Sure thing!

plugins/inputs/neptune_apex/README.md
MaxRenaud: We don't need to store the offset. It's used to calculate the time we add to the metric and the power_failed and power_restored. It was added because it was contained in the XML but it's more meta-data than data anyway. I removed it.

plugins/inputs/neptune_apex/README.md
MaxRenaud: Ok, no problem.
FYI, grafana supports showing this as time if we divide by 1,000 and set its type to "Time". That was my use case and this workaround works.

plugins/inputs/neptune_apex/neptune_apex.go
MaxRenaud: Done.
I'm assuming 78 + \r + \n == the golden 80 char/line?
I would recommend adding this requirement to the CONTRIBUTING.md file as it's usually not a requirement for Go projects (And the tools aren't built for it).

plugins/inputs/neptune_apex/README.md
MaxRenaud: Done as it probably conforms to influxdb's best practices.

plugins/inputs/neptune_apex/neptune_apex.go
MaxRenaud: That statement is part of an if block that guards against r.Probe from having 0 elements.

MaxRenaud:
Thanks for the review. Addressed all comments except for the one about the panic (See my comment).

plugins/inputs/neptune_apex/neptune_apex.go
danielnelson: 78 is just what I have my text editor set to, I'm not sure why, maybe I read [rfc2822](https://tools.ietf.org/html/rfc2822#section-2.1.1) at some point, anyway the exact number is not super important for me but its helpful to be able to see the full line without move horizontally and I often have several files up side by side.

I added some info about this to the various plugin pages: 1d6ff4fe4c3b535da56565a3271aece463123794

PR #5202, opened 2018-12-28
GCP Cloud Pubsub Output Plugin

emilymye:
Related to #2776.

### Required for all PRs:

I want to submit this after https://github.com/influxdata/telegraf/pull/5136 goes in.

### DIFF
--- a/plugins/outputs/cloud_pubsub/README.md
+++ b/plugins/outputs/cloud_pubsub/README.md
@@ -1,0 +1,61 @@
+# Google Cloud PubSub Output Plugin
+
+The GCP PubSub plugin publishes metrics to a [Google Cloud PubSub][pubsub] topic
+as one of the supported [output data formats][].
+
+
+### Configuration
+
+This section contains the default TOML to configure the plugin.  You can
+generate it using `telegraf --usage pubsub`.
+
+```toml
+[[inputs.pubsub]]
+  ## Required. Name of Google Cloud Platform (GCP) Project that owns
+  ## the given PubSub subscription.
+  project = "my-project"
+
+  ## Required. Name of PubSub subscription to ingest metrics from.
+  subscription = "my-subscription"
+
+  ## Required. Data format to consume.
+  ## Each data format has its own unique set of configuration options.
+  ## Read more about them here:
+  ## https://github.com/influxdata/telegraf/blob/master/docs/DATA_FORMATS_INPUT.md
+  data_format = "influx"
+
+  ## Optional. Filepath for GCP credentials JSON file to authorize calls to 
+  ## PubSub APIs. If not set explicitly, Telegraf will attempt to use 
+  ## Application Default Credentials, which is preferred. 
+  # credentials_file = "path/to/my/creds.json"
+
+  ## Optional. PubSub attributes to add to metrics.
+  # [[inputs.pubsub.attributes]]
+  #   my_attr = "tag_value"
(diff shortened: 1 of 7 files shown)


plugins/outputs/cloud_pubsub/README.md
```diff
@@ -0,0 +1,61 @@
...
+  data_format = "influx"
+
+  ## Optional. Filepath for GCP credentials JSON file to authorize calls to 
+  ## PubSub APIs. If not set explicitly, Telegraf will attempt to use 
+  ## Application Default Credentials, which is preferred. 
+  # credentials_file = "path/to/my/creds.json"
+
+  ## Optional. PubSub attributes to add to metrics.
+  # [[inputs.pubsub.attributes]]
+  #   my_attr = "tag_value"
```
danielnelson: Move this option to the end, if it is uncommented the remaining options will fall under it in the TOML document.

plugins/outputs/cloud_pubsub/README.md
```diff
@@ -0,0 +1,61 @@
...
+  ## Optional. Send a request to PubSub (i.e. actually publish a batch)
+  ## when it has this many PubSub messages. If send_batched is true,
+  ## this is ignored and treated as if it were 1
+  publish_byte_threshold = 1000000
+
+  ## Optional. Specifically configures requests made to the PubSub API.
+  publish_num_go_routines = 50
+
+  ## Optional. Specifies a timeout for requests to the PubSub API.
+  publish_timeout = "10m"
```
danielnelson: Comment out these values since they have defaults, but make sure they are set in the plugin (probably just in the init() function).

10m is definitely on the high end for a timeout, this could cause Telegraf to shutdown slowly as we don't have a way to signal a cancel to the output currently.  Usually we have timeouts in the ~10s range, could this be reduced?

plugins/outputs/cloud_pubsub/README.md
```diff
@@ -0,0 +1,61 @@
...
+  ## this is ignored and treated as if it were 1.
+  publish_count_threshold = 1000
+
+  ## Optional. Send a request to PubSub (i.e. actually publish a batch)
+  ## when it has this many PubSub messages. If send_batched is true,
+  ## this is ignored and treated as if it were 1
+  publish_byte_threshold = 1000000
+
+  ## Optional. Specifically configures requests made to the PubSub API.
+  publish_num_go_routines = 50
```
danielnelson: Seems quite high.

plugins/outputs/cloud_pubsub/pubsub.go
```diff
@@ -0,0 +1,258 @@
...
+	if ps.stubTopic == nil {
+		return ps.initPubSubClient()
+	} else {
+		return nil
+	}
+}
+
+func (ps *PubSub) Close() error {
+	if ps.t != nil {
+		ps.t.Stop()
```
danielnelson: Ideally we would wait here.

plugins/outputs/cloud_pubsub/README.md
emilymye: The default is actually 60 seconds - would you want to reduce this further?

plugins/outputs/cloud_pubsub/README.md
emilymye: Reduced heavily!

PR #5210, opened 2019-01-02
[inputs.ceph] added the read and write op per second fields

svenwiltink:
### Required for all PRs:

### DIFF
--- a/plugins/inputs/ceph/README.md
+++ b/plugins/inputs/ceph/README.md
@@ -135,6 +135,8 @@ All fields are collected under the **ceph** measurement and stored as float64s. 
 
 * ceph\_pool\_stats
   * op\_per\_sec (float)
+  * read_op\_per\_sec (float)
+  * write_op\_per\_sec (float)
   * read\_bytes\_sec (float)
   * write\_bytes\_sec (float)
   * recovering\_object\_per\_sec (float)
--- a/plugins/inputs/ceph/README.md
+++ b/plugins/inputs/ceph/README.md
@@ -109,6 +109,8 @@ All fields are collected under the **ceph** measurement and stored as float64s. 
   * data\_bytes (float)
   * num\_pgs (float)
   * op\_per\_sec (float)
+  * read_op\_per\_sec (float)
+  * write_op\_per\_sec (float)
   * read\_bytes\_sec (float)
   * version (float)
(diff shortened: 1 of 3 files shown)


svenwiltink:
the `op_per_sec` field is no longer exported by ceph, but I am reluctant to remove it as older version might still do it.

svenwiltink:
I will also add it to the osd pool stats output

svenwiltink:
requires https://github.com/influxdata/telegraf/pull/5213

plugins/inputs/ceph/ceph.go
```diff
@@ -345,6 +345,8 @@ type CephStatus struct {
 		ReadBytesSec  float64 `json:"read_bytes_sec"`
 		WriteBytesSec float64 `json:"write_bytes_sec"`
 		OpPerSec      float64 `json:"op_per_sec"`
+		ReadOpPerSec  float64 `json:"read_op_per_sec"`
+		WriteOpPerSec float64 `json:"write_op_per_sec"`
```
danielnelson: Let's make the three fields a `*float64`, if the value is nil it will be skipped over when adding it to the accumulator.  Can you also add a comment with the version of ceph that stopped exporting the field here as well.

plugins/inputs/ceph/README.md
```diff
@@ -133,6 +135,8 @@ All fields are collected under the **ceph** measurement and stored as float64s.
 
 * ceph\_pool\_stats
   * op\_per\_sec (float)
```
danielnelson: Add the version of Ceph that removed this field, something like: `* op\_per\_sec (float, ceph < 14)`.

svenwiltink:
The commit that changed this was https://github.com/ceph/ceph/commit/c01a314da3fa574fc3314089090f131973e9d714. This belongs to version v1.10, so I will update the readme accordingly

plugins/inputs/ceph/ceph.go
svenwiltink: One issue with making these 3 fields pointers is that if there isn't any client I/O going on ceph doesn't report the fields. This means we can't differentiate between zero I/O and no exported data, but we do always want to report client I/O

danielnelson: Okay, that being the case I agree we should leave it as is.

PR #5254, opened 2019-01-05
MQTT metric name assignment using regular expressions

ander26:
### Required for all PRs:

Added optional field for establishing the metric's name using regular expressions. Currently, two regular expressions are available to perform the assignment of names since those were the options I needed. However, the addition of a new option where the user could enter the regular expression could be evaluated. The behavior has been tested and works correctly.

Here are the options more detailed: 

Option for determining the name the metric will take
0 = Default behaviour. The metric will receive "mqtt_consumer" as name
1 = The name of the metric will be conformed by the last string of the topic. In case the topic is composed by a single string, that string will be taken as the name of the metric (For example if the topic is test/test1, the metric's name will be test1)
2 = The name of the metric will be conformed by the concatenation of all the strings of the topic separated by "_", except from the first one. In case the topic is composed by a single string, that string will be taken as the name of the metric (For example if the topic is test/test1/test2, the metric's name will be test1_test2)


ander26:
Hi @danielnelson ,

Just to clarify the approach we are proposing, we would like to avoid having multiple mqtt inputs and being able to adapt the name of the measurement (not metric, sorry). So, if we had:
[[inputs.mqtt_consumer]]
  servers = ["tcp://localhost:1883"]
  qos = 1
  connection_timeout = "30s"
  topics = [
    "ander/topic1",
    "ander/topic2",
    "ander/test/#"
  ]
  persistent_session = false
  client_id = ""

  data_format = "json"
  json_time_key = "timestamp"
  json_time_format = "unix_ms"

We would be receiving all the topics with measurement name "mqtt_consumer". The workaround is to have multiple inputs with name overrides, such as:

[[inputs.mqtt_consumer]]
  servers = ["tcp://localhost:1883"]
  qos = 1
  connection_timeout = "30s"
  topics = [
    "ander/topic1"
  ]
  name_override = "topic1"
  persistent_session = false
  client_id = ""

  data_format = "json"
  json_time_key = "timestamp"
  json_time_format = "unix_ms"

[...]

What we propose is the following:

[[inputs.mqtt_consumer]]
  servers = ["tcp://localhost:1883"]
  qos = 1
  connection_timeout = "30s"
  topics = [
    "ander/#"
  ]

  ## This option determines metrics name
  ## 0 - Metric will receive "mqtt_consumer" name
  ## 1 - Metric will receive the last string of the topic (for example if the topic is test/test1, name will be test1)
  ## 2 - Metric will be conformed by the strings in the topic (for example if the topic is test/test1/test2, name will be test1_test2)

  topic_name = 2

  persistent_session = false
  client_id = ""

  data_format = "json"
  json_time_key = "timestamp"
  json_time_format = "unix_ns"

So that if we receive serveral topics they would be redirected to our postgresql db with to the corresponding measurements (if using topic_name parameter = 2):

ander/topic1 -> topic1
ander/topic2 -> topics2
ander/test/topic3 -> test_topic3

Using topic_name = 1:

ander/topic1 -> topic1
ander/topic2 -> topics2
ander/test/topic3 -> topic3

Using topic_name = 0:

ander/topic1 -> mqtt_consumer
ander/topic2 -> mqtt_consumer
ander/test/topic3 -> mqtt_consumer

Maybe everything was clear to you, but rereading my request and your response we were not sure our proposal was correctly understood.

If you still think is not suitable for the current version we will keep in our fork. Thank you very much for your time and keep the good work ;-).

PS: sorry for the loooooong message.

danielnelson:
Sorry about the slow followup, this does make sense if the objects in each topic have a different shape, I'd like to have the transformation done by processors though, because exactly how it is handled needs to be very flexible.  What do you think about something like this:

```toml
[[inputs.mqtt_consumer]]
  servers = ["tcp://localhost:1883"]
  qos = 1
  connection_timeout = "30s"
  topics = [
  "ander/#"
  ]

  persistent_session = false
  client_id = ""

  data_format = "json"
  json_time_key = "timestamp"
  json_time_format = "unix_ns"

[[processors.regex]]
  namepass = ["mqtt_consumer"]
  order = 1

  # topic=ander/test/topic1 -> topic=test/topic1
  [[processors.regex.tags]]
    key = "topic"
    pattern = "^[^/]+/(.*)"
    replacement = "${1}"
  # topic=test/topic1 -> topic=test_topic1
  [[processors.regex.tags]]
    key = "topic"
    pattern = "/"
    replacement = "_"

# doesn't work yet; proposed addition
# mqtt_consumer,topic=ander_topic1 field1=value1 -> ander_topic1 field1=value1
[[processors.converter]]
  namepass = ["mqtt_consumer"]
  order = 2
  [processors.converter.tag]
    measurement = ["topic"]
```

sjwang90:
Seems like the regex processor solution above should address your use case. Please feel free to re-open if there is additional enhancement you would like to add that the `processors.regex` may not address.

PR #4380, opened 2018-07-04
Multiplier plugin

Alessar:
### Required for all PRs:

- [x] Signed [CLA](https://influxdata.com/community/cla/).
- [x] Associated README.md updated.
- [x] Has appropriate unit tests.

Issue #3348, opened 2017-10-17
Gather information from /sys/class/block/*/queue


phemmer:
## Feature Request

Opening a feature request kicks off a discussion.

### Proposal:
Gather metrics from `/sys/class/block/*/queue`

### Current behavior:
No such data

### Desired behavior:
Gather the data.

### Use case: [Why is this important (helps with prioritizing requests)]
The main metric I'm interested in right now is the `nr_requests`, so we can get the maximum number of queued requests on a device, and generate an alarm when we reach a certain percentage.

I'm unsure if this is something that we should add to telegraf. The data in this directory doesn't really change over time. Some of the values only change under human control, and others don't change ever for that specific block device. But having the data would allow users to write alerts without having to know the values present on every block device on every server.

phemmer:
An alternative might be a "file" or "directory" input plugin. This plugin would be given a path (in this case `/sys/class/block/*/queue`), and it would poll all the files in that path (or single file if not a directory), and record each file as a field. The wildcard in the path could also be extracted and added as a tag somehow.

danielnelson:
I guess this is not part of iops_in_progress from diskio, it seems like something that would ideally be in there:
> iops_in_progress:
> This value counts the number of I/O requests that have been issued to
> the device driver but have not yet completed.  It does not include I/O
> requests that are in the queue but not yet issued to the device driver.

I think the file/directory plugin idea could be useful.  Maybe the input could use `data_format` and in this case it would just be `data_format="value"`?  It would essentially be the tail input without tailing.  I do worry though that the field/tag naming will not be very flexible, you almost need something like the graphite template.

phemmer:
> I guess this is not part of iops_in_progress from diskio, it seems like something that would ideally be in there.

For that one specific metric, it might go in there sure. If we want to do that and punt the issue on the rest of the values, it would satisfy my use case.

danielnelson:
I would love to have the file input as well, seems handy for debugging vs what I do now with socket_listener: `echo 'foo bar=42' | nc -u localhost 8094`, so I'm open to either as well.

danielnelson:
I'm also unsure if it is a good idea in general for a plugin to gather from multiple sources.  Diskio currently is essentially `/proc/diskstats`, would adding `/sys/class/block/*/queue` make it less cohesive?  Different sources would be more likely to have different units or, perhaps, underlying sampling rates.  If it is a concern, it would be only a guideline, not a rule.

[...]

PR #5256, opened 2019-01-06
Add Multi-File Input Plugin

martin2250:
### Required for all PRs:

Hi everyone,

I want to use telegraf to pull data from I2C sensors on a Raspberry Pi. The sensor I'm using is not supported by lm_sensors, so I wrote my own plugin to read this (any pretty much any other) sensor. See the README for how it works.

### unit tests
are not done yet, because this is my first go project and I don't know how all this works. I based this plugin on the file input plugin, so possibly something similar to that? I'd appreciate any help.

Thanks!


plugins/inputs/device/device.go
```diff
@@ -0,0 +1,167 @@
...
+			fields[field.Name] = value
+		}
+
+		acc.AddGauge("device", fields, tags, now)
+	}
+	return nil
+}
+
+func init() {
+	types := map[string]DeviceDescriptor{
```
JesusIslam: I don't think you should insert this default devices as not every people who use this has ina219 and bme280.

CMIIW.

martin2250: I plan to add either more of these predefined devices (these two are the only ones I have on hand to test) or make the device descriptor completely customizable via telegraf.conf. This list is not at all exhaustive.
I'm not sure which option would fit better to the overall project, maybe it would also be better to move the type definitions to a new file to not clutter the actual code.

JesusIslam: I personally prefer to have a customizable device descriptors (just provide a commented one for example in the sample config) rather than a list of predefined ones. 

As for the unit test, I think you should make mock device and look at other input plugins as for how it should be tested.

martin2250: Your're right that's probably the better solution.

How would I make a mock device? Just provide a directory with some regular text files like in [the file plugin](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/file/dev/testfiles)? Quick question: the docker-compose.yml in that folder has nothing to do with the go tests, are both required?

JesusIslam: If that's how your plugin works, then you can do that.

The docker-compose.yml from input file plugin is clearly just to test locally judging by the directory's name `dev`. I don't know why he put it there though.

danielnelson:
This plugin ends up being essentially the file input with `value` data_format, perhaps it would make more sense to have this be specifically a raspberry-pi plugin?

Another possibility would be for this to just be a set of example configurations similar to what we have with the [jolokia2 plugin](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/jolokia2#example-configurations).  

One thing the file input can't do is apply change the units of a value, this might be best done in a processor such as in https://github.com/influxdata/telegraf/pull/4380.

JesusIslam:
> perhaps it would make more sense to have this be specifically a raspberry-pi plugin?

AFAIK this is done using `i2c-dev` so essentially should be available for any SBC with GPIO.

martin2250:
@danielnelson, @JesusIslam it does not use any hardware or driver specific interfaces. It is, like you said, a more fancy file plugin that adds
1) multiple fields per point
2) multiplicative factors ("19200" -> 19.2)

You can just as well use it for devices connected via SPI/GPIO, monitoring partitions (eg. `/sys/fs/ext4/nvme0n1p2/`) or anything else that exposes values in files. It's just much more elegant than using a separate file input for each field and using processor plugins to format the points into a more appropriate format.

danielnelson:
Can you show me an example of the multiple named fields per point?  I'm trying to determine if it would be possible to replicate this with the existing plugin set.

martin2250:
> Can you show me an example of the multiple named fields per point? I'm trying to determine if it would be possible to replicate this with the existing plugin set (other than the multiplication).

Consider the BME280 driver: it provides (amongst others) three files in `/sys/bus/i2c/devices/1-0076/iio:device0`:
`in_pressure_input`: `100.323976562`
`in_temp_input`: `19400`
`in_humidityrelative_input`: `51165`

This plugin will output points like this when configured for the bme280 (right now still using the hardcoded devices):
```
device,device=/sys/bus/i2c/devices/1-0076/iio:device0,host=mserver,type=bme280 humidity_relative=51.166000000000004,temperature=19.42,pressure=100.322230468 1546982518000000000
```
(the exact values changed slightly while typing this comment, you get the idea)

danielnelson:
I played around with the file input, and while you can collect a directory full of files using `data_format = "value"` and a glob filename, it doesn't produce a very good result.

This pull request also makes me think of the directory plugin suggested by @phemmer in #3348, I think the method suggested there could be more useful.

What if we had something, somewhat inspired by the snmp input, like this:
```toml
[[inputs.multifile]]
  name_override = "bme280"
  [inputs.multifile.tags]
    device = "device0"

  [[inputs.multifile.file]]
    file = "/sys/bus/i2c/devices/1-0076/iio:device0/in_pressure_input"
    dest = "pressure"
    conversion = "float"
  [[inputs.multifile.file]]
    file = "/sys/bus/i2c/devices/1-0076/iio:device0/in_temp_input"
    dest = "temperature"
    conversion = "float"
  [[inputs.multifile.file]]
    file = "/sys/bus/i2c/devices/1-0076/iio:device0/in_humidityrelative_input"
    dest = "humidityrelative"
    conversion = "float"
```

If `dest` is not defined, the default could be `$(basename $file)`, if no conversion then it would be a string.  Conversion types would be int, float, string, bool, tag, measurement.

Suggestions for a better plugin name would be appreciated.

A couple things are gone in my idea: static definition and factor:

I think that due to the number of different devices and variations that exist as well as differences between kernel versions it would be nice to have a solution that could be used without it being defined statically in code.

The factor seems to be used to change units, but this seems to be mostly a preference, who is to say the values should be reported in volts or millivolts.  Change we make would be overriding what the device driver author decided, which I think are generally what ought to be reported.  I would move the multiplication factor part out to a processor.

martin2250:
I really like your config structure and the name "multifile" describes the plugins function perfectly, I see no need to change that.

Few thoughts:
- I'd suggest adding a 'basedir' value to `[[inputs.multifile]]` to avoid repeating the device path for every file (if the plugin is used for a device). It would default to `/`, so the paths are still absolute if left out.
- Add an option that changes the behavior from "fail completely if a single file can't be read/parsed" to "leave out a field if the file fails to read/parse"
- Do you think allowing a file to override the measurement name is a good idea? I can't think of a scenario where this would be used.

I think the factor should remain part of the plugin. The format which is reported should not depend on the driver implementation. The units used for different physical quantities might be consistent across the linux kernel (they're not, eg. [kernel/power](https://www.kernel.org/doc/Documentation/power/power_supply_class.txt) uses µA to measure current, while the INA219 kernel driver reports current in units of 100µA or power in units of 100nW (?!)), but many other data sources use other units (eg. the [temp input plugin](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/temp) uses °C, while many kernel drivers report temperature in 1/1000 °C)  
Outsourcing the conversion to a processor plugin would only serve to make the config file less readable.

martin2250:
I implemented the plugin as outlined in the last two comments, let me know what you think!

I'll add unit tests as soon as you're ok with the current state

plugins/inputs/multifile/multifile.go
```diff
@@ -0,0 +1,160 @@
...
+	m.initialized = true
+}
+
+func (m *MultiFile) Gather(acc telegraf.Accumulator) error {
+	m.init()
+	now := time.Now()
+	fields := make(map[string]interface{})
+	tags := make(map[string]string)
+
+	for key, value := range m.Tags {
```
danielnelson: Tags setting is actually built in on any input plugin, you can remove this code and it will still work.

martin2250:
Sure thing!

plugins/inputs/multifile/multifile.go
martin2250: I needed quite a while to get this working, can you tell me if I can also remove [the Tags Array in Line 23]?(https://github.com/influxdata/telegraf/pull/5256/files/44f0acfe259becb9d4db6ee8b35e91a770ad6152#diff-4e89817cac1e51f96268b11e514a37feR23)

martin2250:
Ok, multifile should be ready to merge now.

plugins/inputs/multifile/dev/bool.txt
```diff
@@ -0,0 +1 @@
+true
```
danielnelson: Call this directory `testdata`

plugins/inputs/multifile/multifile.go
```diff
@@ -0,0 +1,149 @@
...
+
+		if err != nil {
+			if m.FailEarly {
+				return err
+			}
+			continue
+		}
+
+		if value == nil {
+			return errors.New(fmt.Sprintf("invalid conversion %v", file.Conversion))
```
danielnelson: When would this be hit?  I can think of the instance where float conversion fails but I think that is unintentional.

plugins/inputs/multifile/multifile.go
```diff
@@ -0,0 +1,149 @@
...
+
+		if file.Conversion == "tag" {
+			tags[file.Dest] = vStr
+			continue
+		}
+
+		var value interface{}
+
+		var d int = 0
+		if _, err := fmt.Sscanf(file.Conversion, "float(%d)", &d); err == nil || file.Conversion == "float" {
```
danielnelson: This will not set err correctly for the test below because a new err is created scoped to the if statement.

plugins/inputs/multifile/multifile.go
martin2250: This covers the case where an unsupported value is used for the conversion setting. I could also solve this by a long if/else if chain, but I think this is a bit more elegant and could potentially catch other errors.

plugins/inputs/multifile/multifile.go
martin2250: thanks!

plugins/inputs/multifile/dev/bool.txt
martin2250: done

martin2250:
how's the status on the PR?

martin2250:
thanks :)

PR #5264, opened 2019-01-08
Add raw counters for MongoDB metrics

### LINKED ISSUE #3121: Gather raw metrics in MongoDB input
The MongoDB plugin records a few metrics in InfluxDB as "X per seconds" rather than storing the counter value. The metrics should be counters, not gauge and the "per seconds" should be measured at analysis time instead.

Having the counter, which is what the MongoDB `serverStatus` command returns for operations, queries, etc, is more flexible. If you miss a few records of data, you still have a correct value over time.

I really would like to see the counter values rather than the per-sec values.

bozaro:
closes #3121

Unfortunately MongoDB plugin convert counter metrics to gauge. To eliminate this omission this PR adds raw counter MongoDB metrics for per second gauge metrics.

For example: add `repl_updates_cnt` counter for `repl_updates_per_sec` gauge metric.


plugins/inputs/mongodb/mongodb_data.go
```diff
@@ -31,28 +31,43 @@ func NewMongodbData(statLine *StatLine, tags map[string]string) *MongodbData {
 }
 
 var DefaultStats = map[string]string{
+	"inserts_cnt":               "InsertCnt",
```
danielnelson: Let's name the fields without the `_cnt` where possible, trying to stick to the MongoDB naming (but still in snake_case), as listed in mongostat.go.  So this would be `insert` instead of `inserts`.

For the fields which would collide with the existing field, such as `cursor_timed_out`, then the suffix will need to be added but write it out like `cursor_timed_out_count`.

bozaro: I renamed the added metrics.

bozaro:
Thanks a lot!

PR #5265, opened 2019-01-08
Add Graphite output Template Map

dmitryilyin:
Add the template map feature to the Graphite output. It allows a user to
select which metric should be serialized by which template either
by metric name of by the specified tag value instead of using only
a single template for every metric.
Unmatched metrics can either be processed by the default template
of be excluded from the output.

### Required for all PRs:


dmitryilyin:
Yes, that's how i'm doing it... but if there are many different templates for different metrics it turns to a very long list of graphite outputs... and all those namepass or tagpass, thats nightmare to manage.

Also they are working independently and it's not a good thing performance-wise if i'm correct.

This method does not allow to set different prefix for different templates, otherwise it has the same functionality but with much more human-friendly configuration format.

dmitryilyin:
Also combining both methods can provide a great flexibility.

danielnelson:
> Also they are working independently and it's not a good thing performance-wise if i'm correct.

Should be very similar, Graphite would need an extra connection but that shouldn't be a problem unless you have, at the very least, thousands of connections.  There are details to consider about with the metric buffer, every output gets its own buffer and since they are not combined you may want adjust the sizes.

Can you show me what your current output configs look like?

dmitryilyin:
Well, something like this:

`
[[outputs.graphite]]
  servers = ["127.0.0.1:2003"]
  prefix = "hosts."
  template = "host.input_type.field"
  timeout = 5
  [outputs.graphite.tagpass]
    input_type = ["mem", "swap", "kernel", "system", "processes", "netstat"]

[[outputs.graphite]]
  servers = ["127.0.0.1:2003"]
  prefix = "hosts."
  template = "host.input_type.cpu.field"
  timeout = 5
  [outputs.graphite.tagpass]
    input_type = ["cpu"]

[[outputs.graphite]]
  servers = ["127.0.0.1:2003"]
  prefix = "hosts."
  template = "host.input_type.path.field"
  timeout = 5
  [outputs.graphite.tagpass]
    input_type = ["disk"]
`

But a lot more of these blocks.

Although there is a ability to specify prefix, timeout and buffer/batch size per output...
Well, maybe a little shortening of configuration is not really worth it.

danielnelson:
I think I could get on board if this was added to the graphite serializer, then exposed in the graphite output.  If you want to do this we should try to make it work similar to the graphite parser, using a list for the templates.  I also think we should require a default pattern, so we aren't duplicating the normal filtering mechanism.

PR #5269, opened 2019-01-09
Split statsd parser out into standalone parser without aggregation

j2gg0s:
### Required for all PRs:


danielnelson:
Conceptually the statsd-ness of the metrics would end after parsing, at which point you have only telegraf.Metric, so I'm not in favor of having an aggregator that operates specifically on metrics from the statsd parser.  I think ideally we would use a combination of specifically configured existing aggregators (probably basicstats and histogram) and then also add in new functionality that is not found either to these plugins or more likely to new plugins.

Metrics would then need to be added in to aggregators using the [metric filtering](https://github.com/influxdata/telegraf/blob/master/docs/CONFIGURATION.md#metric-filtering) options and it would be up to the user to set these as needed, though we could certainly show an example somewhere in the docs to replicate the statsd aggregation behavior.

Picking the easiest type, you could sum counters with something like this:

```toml
[[aggregators.basicstats]]
  stats = ["sum"]
  drop_original = true
  [aggregators.basicstats.tagpass]
    statsd_type = ["c"]
```

I think supporting all the types would involve:
- counters: existing basicstats aggregators
- gauge: new aggregator that can do last/first, similar to the influxql functions.
- timers: existing basicstats aggregators + new percentile aggregator
- sets: existing valuecounter aggregator

j2gg0s:
@danielnelson 

Agree, and there are some questions.

- sets need new aggregator, it count distinct value in a period, not number of occurrences
- as [reference](https://github.com/etsy/statsd/blob/master/docs/metric_types.md#gauges) in `inputs.statsd`, gauge support sign. This should be support in new aggregator.
- percentile must have the ability to calculate basictats's field, like sum, mean. so why we should use basicstats + percentile?

danielnelson:
All very good points,

> sets need new aggregator, it count distinct value in a period, not number of occurrences

Yes, good catch

> gauge support sign. This should be support in new aggregator.

Oh this does somewhat complicate things, as telegraf.Metric doesn't have a way to express a relative value.  It is always possible to encode more data into a tag but it would make the aggregator very specific to statsd.

One thing I wonder is if we could somehow push this back into the parser and only output absolute values, but I see some complexity around how long to cache for and the delete_gauges options while keeping the Parser interface.

> percentile must have the ability to calculate basictats's field, like sum, mean. so why we should use basicstats + percentile?

This is true if we want to support all the percentile features from statsd, which I don't think we are doing today, we ought to have upper/lower for each percentile but the statsd docs show only a single value for the timer.  It makes sense to have a percentile or quantile aggregator that handles all of this.

Let's just do the parser in this PR and we can open separate pull requests for aggregators?

plugins/parsers/statsd/parser.go
```diff
@@ -0,0 +1,272 @@
...
+
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/metric"
+	"github.com/influxdata/telegraf/plugins/parsers/graphite"
+)
+
+const (
+	defaultFieldName = "value"
+	defaultSeparator = "_"
+	defaultTagKey    = "statsd_type"
```
danielnelson: I think we should use `metric_type` to remain more compatible with stats parsed from the statsd input, along with the long names: `counter`, `gauge`, etc.

plugins/parsers/statsd/parser.go
```diff
@@ -0,0 +1,272 @@
...
+		v, err := strconv.ParseInt(value, 10, 64)
+		if err != nil {
+			v2, err2 := strconv.ParseFloat(value, 64)
+			if err2 != nil {
+				return nil, errors.New("Error Parsing statsd line")
+			}
+			v = int64(v2)
+		}
+		return v, nil
+	case "s", "g":
```
danielnelson: I think we should parse gauge as a float64, but instead add an additional tag if it is a relative value:
```
foo,metric_type=gauge,operation=+ value=42
```

plugins/parsers/statsd/parser.go
j2gg0s: Done that.

plugins/parsers/statsd/parser.go
j2gg0s: - Change type to float64.
- use "additive" may be better? 

`foo:50|g` --> `foo,metric_type=gauge value=50`
`foo:+50|g` --> `foo,metric_type=gauge,operation=additive value=50`
`foo:-50|g` --> `foo,metric_type=gauge,operation=additive value=-50`

danielnelson: Sure, and subtractive for `-`?

danielnelson: Oh nevermind, yeah your example is fine.

danielnelson:
@j2gg0s I've been meaning to review again but I've been very busy, sorry.  Can you let me know why you closed the PR?

j2gg0s:
LOL 
I have some free time this morning, so i tried to sort out my account and closed some issue and pr.

plugins/parsers/statsd/README.md
```diff
@@ -0,0 +1,136 @@
...
+  # service_address = "udp4://:8094"
+  # service_address = "udp6://:8094"
+  # service_address = "unix:///tmp/telegraf.sock"
+  # service_address = "unixgram:///tmp/telegraf.sock"
+
+  ## Data format to consume.
+  ## Each data format has its own unique set of configuration options, read
+  ## more about them here:
+  ## https://github.com/influxdata/telegraf/blob/master/docs/DATA_FORMATS_INPUT.md
+  data_format = "influx"
```
srebhan, line 25: This probably should be `statsd`!?

PR #5293, opened 2019-01-15
Collect from newly discovered/launched pods

### LINKED ISSUE #5281: Telegraf prometheus doesn't detect pods added and deleted
### Relevant telegraf.conf:
Input plugin: Prometheus
Output plugin : OpenTSDB

### System info:
Telegraf version : 1.9.1
Platform: Openshift 3.7

### Steps to reproduce:

1. Generic configuration of Prometheus plugin and OpenTSDB plugin
2. monitoring_pods is enabled in the Prometheus plugin

### Expected behavior:
Metrics are retrieved successfully from all pods that contain Prometheus annotations, even if I added a pod or deleted a pod, telegraf picks up on that and start/stop retrieving metrics from the pod.

### Actual behavior:
Metrics are retrieved successfully from all pods that are present the minute I deploy telegraf, but if I add a pod then telegraf doesn't detect it and if I delete a pod then telegraf starts showing errors because the pod's IP is not present anymore so it can't connect to it.

glinton:
Stop collecting from deleted pods

Resolves #5281

### DIFF
--- a/plugins/inputs/prometheus/kubernetes.go
+++ b/plugins/inputs/prometheus/kubernetes.go
@@ -45,6 +45,7 @@ func (p *Prometheus) start(ctx context.Context) error {
 		if err != nil {
 			return fmt.Errorf("Failed to get current user - %v", err)
 		}
+
 		configLocation := filepath.Join(u.HomeDir, ".kube/config")
 		if p.KubeConfig != "" {
 			configLocation = p.KubeConfig
@@ -98,13 +99,42 @@ func (p *Prometheus) watch(ctx context.Context, client *k8s.Client) error {
 
 			switch eventType {
 			case k8s.EventAdded:
-				registerPod(pod, p)
-			case k8s.EventDeleted:
-				unregisterPod(pod, p)
+				if podReady(pod.Status.GetContainerStatuses()) {
+					registerPod(pod, p)
+				}
 			case k8s.EventModified:
+				if pod.Metadata.GetDeletionTimestamp() != nil && podConditionReady(pod.Status.GetConditions()) {
+					unregisterPod(pod, p)
+				} else if podReady(pod.Status.GetContainerStatuses()) && podConditionReady(pod.Status.GetConditions()) {
+					registerPod(pod, p)
+				}
 			}
 		}
 	}
+}
+
+func podReady(statuss []*corev1.ContainerStatus) bool {
+	if len(statuss) == 0 {
+		return false
+	}
+	for _, cs := range statuss {
+		if !cs.GetReady() {
+			return false
+		}
+	}
+	return true
+}
+
+func podConditionReady(statuss []*corev1.PodCondition) bool {
+	if len(statuss) < 3 {
+		return false
+	}
+	for _, pc := range statuss {
+		if pc.GetType() == "Ready" && pc.GetStatus() == "False" {
+			return false
+		}
+	}
+	return true
 }
 
 func registerPod(pod *corev1.Pod, p *Prometheus) {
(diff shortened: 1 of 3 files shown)


plugins/inputs/prometheus/kubernetes.go
```diff
@@ -98,15 +99,44 @@ func (p *Prometheus) watch(ctx context.Context, client *k8s.Client) error {
...
-			case k8s.EventDeleted:
-				unregisterPod(pod, p)
+				if podReady(pod.Status.GetContainerStatuses()) {
+					registerPod(pod, p)
+				}
 			case k8s.EventModified:
+				if pod.Metadata.GetDeletionTimestamp() != nil && podConditionReady(pod.Status.GetConditions()) {
+					unregisterPod(pod, p)
+				} else if podReady(pod.Status.GetContainerStatuses()) && podConditionReady(pod.Status.GetConditions()) {
+					registerPod(pod, p)
```
danielnelson: What if the pod was already registered in EventAdded?  Since `p.kubernetesPods` is a list won't we add it twice?  I think this might be simpler if we change it to a map.

plugins/inputs/prometheus/kubernetes.go
```diff
@@ -98,15 +99,44 @@ func (p *Prometheus) watch(ctx context.Context, client *k8s.Client) error {
...
-				registerPod(pod, p)
-			case k8s.EventDeleted:
-				unregisterPod(pod, p)
+				if podReady(pod.Status.GetContainerStatuses()) {
+					registerPod(pod, p)
+				}
 			case k8s.EventModified:
+				if pod.Metadata.GetDeletionTimestamp() != nil && podConditionReady(pod.Status.GetConditions()) {
+					unregisterPod(pod, p)
+				} else if podReady(pod.Status.GetContainerStatuses()) && podConditionReady(pod.Status.GetConditions()) {
```
danielnelson: Can the pod be not ready when the containers are ready or vis-versa?  Why doesn't this match the unregister condition on line 106?  Do we always receive a modified event in the case that the pod is removed?

plugins/inputs/prometheus/kubernetes.go
```diff
@@ -98,15 +99,44 @@ func (p *Prometheus) watch(ctx context.Context, client *k8s.Client) error {
...
+	}
+	return true
+}
+
+func podConditionReady(statuss []*corev1.PodCondition) bool {
+	if len(statuss) < 3 {
+		return false
+	}
+	for _, pc := range statuss {
+		if pc.GetType() == "Ready" && pc.GetStatus() == "False" {
```
danielnelson: This looks brittle, relying on the number of conditions and finding a Ready state... Is there only one condition with GetStatus() == "True" at a time?

glinton: Any and all can have different or the same status values.

plugins/inputs/prometheus/kubernetes.go
glinton: I guess I didn't try a pod restart, which would trigger a modify event on an already registered pod.. I'll test that and see.

plugins/inputs/prometheus/kubernetes.go
glinton: I know when I didn't have them, a newly created pod would be registered, unregistered, and then registered again before it was "up"

plugins/inputs/prometheus/kubernetes.go
glinton: This particular case is handled now. Though it would only be triggered if kubernetes didn't initiate the deletion of the old pod (`docker kill mypod...`).

plugins/inputs/prometheus/kubernetes.go
glinton: Not sure what I was doing to think they were different, they seem to be true at the same time. We do not always receive a modified event in the case that a pod is removed (external deletion/`docker rm`, etc...), but we always do when k8s requests the deletion.

plugins/inputs/prometheus/kubernetes.go
glinton: Couldn't think of a great way to avoid checking the length, so I just went with the less brittle way to check.

If `len(pod.Status.GetContainerStatuses()) == 0` then there are no containers in the pod and it is not ready, thus `false`. If any containers are not "ready", the pod appears to not be fully ready either.

Issue #4666, opened 2018-09-10
Support for Windows containers


olljanat:
## Feature Request
Currently containerized version of Telegraf is only available for Linux and it would be very useful to have it for native [Windows containers](https://docs.microsoft.com/en-us/virtualization/windowscontainers/about/) too.

### Proposal:
* Make sure that Docker plugin is able to communicate with Windows containers
* Implement Dockerfile for Windows containers:  https://github.com/influxdata/influxdata-docker/issues/145

### Use case:
Windows container host monitoring.

## Extra comments
~~I can start look this one but let me know if you have any requirements/comments how you want it to be implemented.~~

danielnelson:
@olljanat That would be great if you could look into this.  I haven't researched this at all but the main question I have is if the image would have to be built on a Windows machine or if it can be created from Linux.

olljanat:
@danielnelson it need to be created on Windows. That is only way.

Btw. Is there some reason that Docker client is set to use 1.21 version instead of some newer one? (https://github.com/influxdata/telegraf/blob/master/plugins/inputs/docker/client.go#L14)
1.24 is first one which is supported by Windows containers.

glinton:
The docker client is set to 1.21 to support older docker versions. Iirc, it will still work with newer server versions, let me know if you find differently.

zentron:
We are hitting this issue with version `18.03.1-ee-3` of the docker engine. It currently requires minimum of `1.24`

Interestingly it looks like it _was_ using `1.24` but was downgraded.

Would it be possible to get this version to be configurable?

danielnelson:
As a workaround, you should be able to use the environment configuration to manually set the api version, in your environment set these variables as needed:
```
// DOCKER_HOST to set the url to the docker server.
// DOCKER_API_VERSION to set the version of the API to reach, leave empty for latest.
// DOCKER_CERT_PATH to load the TLS certificates from.
// DOCKER_TLS_VERIFY to enable or disable TLS verification, off by default.
```

Then use `endpoint = "ENV"` in the plugin configuration:
```
[[inputs.docker]]
  endpoint = "ENV"
```

We should investigate if we can use `NegotiateAPIVersion` and verify that it will work with 1.8 as well as the latest docker version.

TerraVenil:
Hi, @danielnelson. Suggested workaround not worked for me, seems like env variables DOCKER_HOST/DOCKER_API_VERSION ignored by telegraf service.

[...]

PR #5302, opened 2019-01-17
Set windows docker client version to lowest supported

glinton:
As pointed out in #4666, the lowest docker version windows supports is `1.24`. `NegotiateAPIVersion` isn't an option because it only downgrades the client to the server version, since the client in telegraf is set lower than docker will ever report, it does nothing.

Ideally we'd quit supporting EOL docker versions that require older docker clients and go back to using `1.24` for all distributions. Would revert change made in #4434 to support deprecated versions of docker :flushed:.

### DIFF
--- a/plugins/inputs/docker/client.go
+++ b/plugins/inputs/docker/client.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/tls"
 	"net/http"
+	"runtime"
 
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/api/types/swarm"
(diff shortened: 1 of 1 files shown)


danielnelson:
I don't think this fix can work, since docker is often collecting over the network.  For example, you may be collecting from Windows Docker over the network from Linux.

olljanat:
@glinton I think that if you add another check that if host value starts with **npipe:** then it is bullet proof as that is possible only if Telegraf is running on Windows and it is connecting to local Docker installation.

PR #5316, opened 2019-01-18
Removing auth from /ping route

JeffAshton:
Adding ability to disable authentication on ping route.  Load balancers like AWS's ALBs don't support basic auth.

### Required for all PRs:

### DIFF
--- a/plugins/inputs/influxdb_listener/http_listener.go
+++ b/plugins/inputs/influxdb_listener/http_listener.go
@@ -229,9 +229,7 @@ func (h *HTTPListener) ServeHTTP(res http.ResponseWriter, req *http.Request) {
 		h.PingsRecv.Incr(1)
 		defer h.PingsServed.Incr(1)
 		// respond to ping requests
-		h.AuthenticateIfSet(func(res http.ResponseWriter, req *http.Request) {
-			res.WriteHeader(http.StatusNoContent)
-		}, res, req)
+		res.WriteHeader(http.StatusNoContent)
 	default:
 		defer h.NotFoundsServed.Incr(1)
 		// Don't know how to respond to calls to other endpoints


danielnelson:
It's the healthcheck that doesn't support authentication?

JeffAshton:
That's correct.  Here's the docs,, 

https://docs.aws.amazon.com/elasticloadbalancing/latest/application/target-group-health-checks.html

'm currently matching on the 401 response, but this shows up as auth failures in the internal stats, and I want to monitor this separately.

danielnelson:
How would you feel about using a separate http_listener_v2 as the ping endpoint?  This way we don't need to add a new option that InfluxDB itself doesn't have.
```
[[inputs.http_listener_v2]]
  service_address = ":8681"
  path = "/ping"
```

JeffAshton:
So from my testing, influx doesn't actually do auth on the ping route:

```
influx-db:
    image: "influxdb:1.7.3-alpine"
    environment:
      INFLUXDB_HTTP_AUTH_ENABLED: 'true'
      INFLUXDB_ADMIN_USER: 'admin'
      INFLUXDB_ADMIN_PASSWORD: 'password'
    ports:
      - "18086:8086"
    restart: always
```

```
curl -Is http://localhost:18086/ping
HTTP/1.1 204 No Content
Content-Type: application/json
Request-Id: 1a3f533a-1f2d-11e9-8005-0242ac110005
X-Influxdb-Build: OSS
X-Influxdb-Version: 1.7.3
X-Request-Id: 1a3f533a-1f2d-11e9-8005-0242ac110005
Date: Wed, 23 Jan 2019 16:36:58 GMT
```

danielnelson:
Let's just remove the auth from the `/ping` endpoint, without any option.  The goal for this plugin is to act as much like InfluxDB as possible and anytime we can avoid additional options I'm in favor.

Add to that list `/debug/pprof`, which can at least be disabled in the config.  I found this issue https://github.com/influxdata/influxdb/issues/5305 which mentions `/debug/vars`, but not these other endpoints.

JeffAshton:
Sounds good.  Will update the PR tomorrow.

plugins/inputs/influxdb_listener/http_listener.go
```diff
@@ -229,9 +229,7 @@ func (h *HTTPListener) ServeHTTP(res http.ResponseWriter, req *http.Request) {
 		h.PingsRecv.Incr(1)
 		defer h.PingsServed.Incr(1)
 		// respond to ping requests
-		h.AuthenticateIfSet(func(res http.ResponseWriter, req *http.Request) {
-			res.WriteHeader(http.StatusNoContent)
-		}, res, req)
+		res.WriteHeader(http.StatusNoContent)
```
JeffAshton: Covered by existing test case:

https://github.com/influxdata/telegraf/blob/0f387674407a7ca9dbef97fa15a09a73fe967df9/plugins/inputs/influxdb_listener/http_listener_test.go#L413

PR #5320, opened 2019-01-21
Fix Consul "datacenter" typo

marceloalmeida:
### Required for all PRs:


danielnelson:
Thanks for the pull request, would you be able to keep support for the old name by having both Datacenter and Datacentre on the struct, and then copy the old value into the new one?

marceloalmeida:
Thanks for the input @danielnelson, already did the fix to keep the backward compatibility with old configurations.

PR #5323, opened 2019-01-22
Add azure_storage_queue input plugin

mjiderhamn:
### Required for all PRs:

### DIFF
--- a/plugins/inputs/azure_storage_queue/README.md
+++ b/plugins/inputs/azure_storage_queue/README.md
@@ -13,18 +13,23 @@ This plugin gathers sizes of Azure Storage Queues.
   ## Required Azure Storage Account access key
   azure_storage_account_key = "storageaccountaccesskey"
   
-  ## Uncomment to disable peeking age of oldest message (saves time)
-  # peek_oldest_message_age = false
+  ## Set to false to disable peeking age of oldest message (executes faster)
+  # peek_oldest_message_age = true
 ```
 
-### Measurements & Fields:
+### Metrics
+- azure_storage_queues
+  - tags:
+    - name
+    - storage_account
+  - fields:
+    - size (integer, count)
+    - oldest_message_age (integer, seconds) Age of message at the head of the queue, in seconds.
+      Requires `peek_oldest_message_age` to be configured to `true`.
+      
+### Example Output
 
-- azure_storage_queues:
-  - size
-  - oldest_message_age - Age of message at the head of the queue, in seconds
-
-### Tags:
-
-- azure_storage_queues:
-  - name
-  - storage_account
+```
+azure_storage_queues,name=myqueue,storage_account=mystorageaccount oldest_message_age=132i,size=7i 1565970503000000000
+azure_storage_queues,name=myemptyqueue,storage_account=mystorageaccount size=0i 1565970502000000000
+```
--- a/plugins/inputs/azure_storage_queue/README.md
+++ b/plugins/inputs/azure_storage_queue/README.md
@@ -24,12 +24,12 @@ This plugin gathers sizes of Azure Storage Queues.
     - storage_account
   - fields:
     - size (integer, count)
-    - oldest_message_age (integer, seconds) Age of message at the head of the queue, in seconds.
+    - oldest_message_age_ns (integer, nanoseconds) Age of message at the head of the queue.
       Requires `peek_oldest_message_age` to be configured to `true`.
       
 ### Example Output
 
 ```
-azure_storage_queues,name=myqueue,storage_account=mystorageaccount oldest_message_age=132i,size=7i 1565970503000000000
+azure_storage_queues,name=myqueue,storage_account=mystorageaccount oldest_message_age=799714900i,size=7i 1565970503000000000
 azure_storage_queues,name=myemptyqueue,storage_account=mystorageaccount size=0i 1565970502000000000
 ```
--- a/plugins/inputs/azure_storage_queue/README.md
+++ b/plugins/inputs/azure_storage_queue/README.md
@@ -8,10 +8,10 @@ This plugin gathers sizes of Azure Storage Queues.
 # Description
 [[inputs.azure_storage_queue]]
   ## Required Azure Storage Account name
-  azure_storage_account_name = "mystorageaccount"
+  account_name = "mystorageaccount"
 
   ## Required Azure Storage Account access key
-  azure_storage_account_key = "storageaccountaccesskey"
+  account_key = "storageaccountaccesskey"
   
   ## Set to false to disable peeking age of oldest message (executes faster)
   # peek_oldest_message_age = true
@@ -20,8 +20,8 @@ This plugin gathers sizes of Azure Storage Queues.
 ### Metrics
 - azure_storage_queues
   - tags:
-    - name
-    - storage_account
+    - queue
+    - account
   - fields:
     - size (integer, count)
     - oldest_message_age_ns (integer, nanoseconds) Age of message at the head of the queue.
@@ -30,6 +30,6 @@ This plugin gathers sizes of Azure Storage Queues.
 ### Example Output
 
 ```
-azure_storage_queues,name=myqueue,storage_account=mystorageaccount oldest_message_age=799714900i,size=7i 1565970503000000000
-azure_storage_queues,name=myemptyqueue,storage_account=mystorageaccount size=0i 1565970502000000000
+azure_storage_queues,queue=myqueue,account=mystorageaccount oldest_message_age=799714900i,size=7i 1565970503000000000
+azure_storage_queues,queue=myemptyqueue,account=mystorageaccount size=0i 1565970502000000000
 ```
--- a/plugins/inputs/azure_storage_queue/README.md
+++ b/plugins/inputs/azure_storage_queue/README.md
@@ -12,12 +12,16 @@ This plugin gathers sizes of Azure Storage Queues.
 
   ## Required Azure Storage Account access key
   azure_storage_account_key = "storageaccountaccesskey"
+  
+  ## Uncomment to disable peeking age of oldest message (saves time)
+  # peek_oldest_message_age = false
 ```
 
 ### Measurements & Fields:
 
 - azure_storage_queues:
   - size
+  - oldest_message_age - Age of message at the head of the queue, in seconds
 
 ### Tags:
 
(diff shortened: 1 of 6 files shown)


danielnelson:
Looks like we need to run `dep ensure` once to fix the CI tests.

mjiderhamn:
Is there anything I can do to keep this ball rolling?

danielnelson:
Planning to take a closer look during the 1.11 release cycle, but can you try to get the tests passing?  `dep ensure` should modify the Gopkg.lock file and I think that will do it.

mjiderhamn:
@danielnelson `Gopkg.lock` [has been updated](https://github.com/influxdata/telegraf/pull/5323/commits/44cd49882e763a8270fca736ec1ad156380bb051).

Do I need to update the [Circle CI config](https://github.com/influxdata/telegraf/blob/master/.circleci/config.yml) somehow? Or it related to [the Docker image](https://quay.io/repository/influxdb/telegraf-ci)???
(I'm sorry that I'm relatively new to both Go build pipeline and Circle CI.)

glinton:
Did you add the missing dependency with `dep ensure -add github.com/Azure/azure-pipeline-go/pipeline`?

mjiderhamn:
Apparently, Microsoft does not support Go 1.9
```
vendor/github.com/Azure/azure-storage-queue-go/azqueue/sas_service.go:76:8: undefined: strings.Builder
vendor/github.com/Azure/azure-storage-queue-go/azqueue/url_queue.go:109:8: undefined: strings.Builder
```

What is the reason Telegraf needs to be Go 1.9 compatible?

danielnelson:
We are trying to keep Go 1.9 support because it is the last version that produces binaries that work on RHEL5.  Don't worry about the failure for now, we will either drop support for Go 1.9 in the next release, or find a way to exclude this plugin from the build for Go 1.9.

danielnelson:
We have dropped Go 1.9 support, if you rebase against master you should pick up the change.

Hypothetically, if we had a [Azure Monitor](https://docs.microsoft.com/en-us/azure/storage/common/storage-metrics-in-azure-monitor) input, similar to the existing cloudwatch input, could it do everything this plugin does?  I have heard a rumor that @chobbs has an experimental input plugin for Azure Monitor, we just need to convince him to open a PR.

mjiderhamn:
AFAIK the Azure Metrics won't allow you to get the age of the oldest message. And I'm uncertain how you'd get the message count _per queue_ (would you have to use filters and query one queue at a time? Can you get the queue names through the Metrics API???).
In addition to that, the authentication model is different. So it may make sense to have both plugins, even if data provided would be partially overlapping.

I've been a bit keen on creating a more generic Azure plugin similar to [Datadog Azure integration](https://docs.datadoghq.com/integrations/azure/?tab=azurecliv20#data-collected). If @chobbs has anything started and/or would like to join forces, feel free to reach out to me.

plugins/inputs/azure_storage_queue/azure_storage_queue.go
```diff
@@ -0,0 +1,142 @@
...
+	acc.AddFields("azure_storage_queues", records, tags)
+}
+
+func (a *AzureStorageQueue) Gather(acc telegraf.Accumulator) error {
+	serviceURL, err := a.GetServiceURL()
+	if err != nil {
+		return err
+	}
+
+	ctx := context.TODO() // This example uses a never-expiring context
```
glinton: Remove comment

plugins/inputs/azure_storage_queue/azure_storage_queue.go
```diff
@@ -0,0 +1,142 @@
...
+
+	for marker := (azqueue.Marker{}); marker.NotDone(); {
+		queuesSegment, err := serviceURL.ListQueuesSegment(ctx, marker,
+			azqueue.ListQueuesSegmentOptions{
+				Detail: azqueue.ListQueuesSegmentDetails{Metadata: false},
+			})
+		if err != nil {
+			// log.Fatal(err)
```
glinton: Remove unused code, if it is intended that operation continue, you can call `log.Printf("E! [inputs.azure_storage_queue] some optional description: %s", err.Error())` followed by `continue`

Same for line 118 and 125.

plugins/inputs/azure_storage_queue/azure_storage_queue.go
```diff
@@ -0,0 +1,142 @@
...
+		pipeline := azqueue.NewPipeline(credential, azqueue.PipelineOptions{})
+
+		serviceURL := azqueue.NewServiceURL(*_url, pipeline)
+		a.serviceURL = &serviceURL
+	}
+	return *a.serviceURL, nil
+}
+
+func (a *AzureStorageQueue) GatherQueueMetrics(acc telegraf.Accumulator, queueItem azqueue.QueueItem, properties *azqueue.QueueGetPropertiesResponse, peekedMessage *azqueue.PeekedMessage) {
+	records := make(map[string]interface{})
```
glinton: nitpick: call this `fields`

plugins/inputs/azure_storage_queue/README.md
```diff
@@ -0,0 +1,30 @@
...
+  
+  ## Uncomment to disable peeking age of oldest message (saves time)
+  # peek_oldest_message_age = false
+```
+
+### Measurements & Fields:
+
+- azure_storage_queues:
+  - size
+  - oldest_message_age - Age of message at the head of the queue, in seconds
```
glinton: I think we're trying to standardize on nanoseconds, can you append `_ns` to this and make sure it's stored as nanoseconds?

plugins/inputs/azure_storage_queue/README.md
```diff
@@ -0,0 +1,30 @@
...
+# Description
+[[inputs.azure_storage_queue]]
+  ## Required Azure Storage Account name
+  azure_storage_account_name = "mystorageaccount"
+
+  ## Required Azure Storage Account access key
+  azure_storage_account_key = "storageaccountaccesskey"
+  
+  ## Uncomment to disable peeking age of oldest message (saves time)
+  # peek_oldest_message_age = false
```
glinton: Change this to true (as it's the default in the code)

plugins/inputs/azure_storage_queue/README.md
glinton: Add note that this field is dependent on `peek_oldest_message_age`

mjiderhamn:
Thanks for your comments @glinton - please see updates

plugins/inputs/azure_storage_queue/README.md
glinton: Change this value to `true` in the `SampleConfig` as well

plugins/inputs/azure_storage_queue/azure_storage_queue.go
```diff
@@ -0,0 +1,132 @@
...
+			if err != nil {
+				log.Printf("E! [inputs.azure_storage_queue] Error getting properties for queue %s: %s", queueItem.Name, err.Error())
+				continue
+			}
+			var peekedMessage *azqueue.PeekedMessage
+			if a.PeekOldestMessageAge {
+				messagesURL := queueURL.NewMessagesURL()
+				messagesResponse, err := messagesURL.Peek(ctx, 1)
+				if err != nil {
+					log.Printf("E! [inputs.azure_storage_queue] Error peeking queue %s: %s", queueItem.Name, err.Error())
```
glinton: Should there be a `continue` after this so `a.GatherQueueMetrics isn't called?

glinton:
Thanks for your changes. If you feel that `continue` is needed, please refrain from force pushing when making that change. We squash and merge so there is no benefit to force pushing. Thanks

plugins/inputs/azure_storage_queue/azure_storage_queue.go
mjiderhamn: My intent is that if we fail to collect the oldest message age, we still want to report queue size (as if `PeekOldestMessageAge` was false).
Potentially that means a warning is more appropriate than an error message?

mjiderhamn:
I see your point on the no of dependencies it brings in @danielnelson , but I'm not sure I agree.

First of all it's usually good to avoid trying to reinvent the wheel, and use vendor provided libs where available. This is already done for other Azure plugins such as `azure_monitor` and `application_insights` outputs. How is this different? Do those plugins provide more "features per dependency"?

Secondly, since there is a lot of data to be gathered from Azure, it would seem likely that these dependencies may be used in other future plugins. Assume we use a homegrown solution in this case. Then we add another input plugin, with another homegrown solution. And another. And maybe yet another. When would it be motivated to switch to the vendor provided dependencies?

Even though the queue API interaction is feature specific and the small part of it that we do use could be homegrown, I do think it makes sense to reuse the vendor supplied auth API.

Looking at the dependency tree, it seems a lot of them are pulled in just to [support detecting Windows proxy settings](https://github.com/mattn/go-ieproxy). Would removing that dependency make you less reluctant? In that case, what are our options?

_(I hope that you will excuse the fact that I'm still new to the Go ecosystem and colored by 15+ years of Java.)_

Is it possible to override the proxy detection dependency with a custom implementation of the same API and exclude the transitive dependencies (as you would be able to do with Maven)?

Does Telegraf use other means of proxy detection that we can replace this with? Or could we decide not to support proxy detection?

danielnelson:
We can do this with the reasoning that we want to do some more related plugins that would use these dependencies.  Little bit more on my thoughts below for those who come later.

> This is already done for other Azure plugins such as azure_monitor and application_insights outputs. How is this different? Do those plugins provide more "features per dependency"?

Yes, and we are not utilizing much functionality from these libraries.

> Secondly, since there is a lot of data to be gathered from Azure, it would seem likely that these dependencies may be used in other future plugins.

This is probably the best argument for allowing this, we will likely want to add more support for storage queues, for example a `azure_storage_queue_consumer` input and a `azure_storage_queue` output.  Similar to the AWS Kinesis or Google pubsub plugins.

> Then we add another input plugin, with another homegrown solution. And another. And maybe yet another. When would it be motivated to switch to the vendor provided dependencies?

It's a gray area judgement call based on value gained from the dependency vs cost to implement.

> Is it possible to override the proxy detection dependency with a custom implementation of the same API and exclude the transitive dependencies (as you would be able to do with Maven)?

Technically possible but not something that would be considered best practice in Go, and not something we would consider.  Most of these dependencies we already have so its not as bad as it looks, so it would add:
- github.com/Azure/azure-storage-queue-go
- github.com/Azure/azure-pipeline-go
- github.com/mattn/go-ieproxy (not even added now as azure-storage-queue-go depends on an older version of azure-pipeline-go; but we can expect it in the future)

PR #5324, opened 2019-01-22
Add result type to regex processor

sgtsquiggs:
I really needed the ability to add/update a tag based on field content; the easiest path seemed to be modifying the regex processor. Perhaps someone else will need this functionality as well?

### Required for all PRs:


danielnelson:
You should be able to convert the field to a tag using the converter processor either before or after applying the regex processor, make sure to use the `order` options to control which order they are applied in.

Issue #5344, opened 2019-01-25
Add Serializer for Carbon2


frankreno:
## Feature Request

Opening a feature request to kick off a discussion.

### Proposal:

I would like to add a Serializer for the Carbon2 format defined in the[ Metric2.0 spec](http://metrics20.org/implementations/).  Per the contribution guidelines, I am opening this issue first but do have a PR that will be ready for review shortly.

### Current behavior:

Currently, there is no support for Carbon2.

### Desired behavior:

As a Telegraf user, I want to have the ability to serialize a Telegraf metric to Carbon2 for applications that support that Carbon2 format.

### Use case: [Why is this important (helps with prioritizing requests)]

Various tools and projects support Carbon2 today:

- [Statsdaemon](https://github.com/raintank/statsdaemon)
- [Graph-Explorer](http://vimeo.github.io/graph-explorer/)
- [Metricsd](https://github.com/josegonzalez/metricsd)
- [Carbon-tagger](https://github.com/vimeo/carbon-tagger)
- [Structured Metrics](https://github.com/vimeo/graph-explorer/tree/master/graph_explorer/structured_metrics)
- [Sumo Logic](https://help.sumologic.com/03Send-Data/Sources/02Sources-for-Hosted-Collectors/HTTP-Source/Upload-Metrics-to-an-HTTP-Source#Content-Type_headers_for_metrics)

The above applications and services, as well as any future vendors who decide to implement support for the Carbon2 metrics format,  could benefit from this addition.

frankreno:
Fixed by #5345

PR #5345, opened 2019-01-25
Add carbon2 serializer

frankreno:
### Required for all PRs:

This PR addresses issue #5344 by adding a Carbon2 serializer.  Carbon2 is a metrics format supported by multiple applications and vendors.  Implementing it as a serializer would allow it to integrate with the existing output plugin ecosystem as well as integrate with any additional vendors who want to add support for the Carbon2 format.

### DIFF
--- a/plugins/serializers/carbon2/carbon2.go
+++ b/plugins/serializers/carbon2/carbon2.go
@@ -21,7 +21,7 @@ func (s *serializer) Serialize(metric telegraf.Metric) ([]byte, error) {
 }
 
 func (s *serializer) SerializeBatch(metrics []telegraf.Metric) ([]byte, error) {
-	var batch strings.Builder
+	var batch bytes.Buffer
 	for _, metric := range metrics {
 		batch.WriteString(s.createObject(metric))
 	}
--- a/plugins/serializers/carbon2/carbon2.go
+++ b/plugins/serializers/carbon2/carbon2.go
@@ -1,6 +1,7 @@
 package carbon2
 
 import (
+	"bytes"
 	"fmt"
 	"github.com/influxdata/telegraf"
 	"strconv"
@@ -28,7 +29,7 @@ func (s *serializer) SerializeBatch(metrics []telegraf.Metric) ([]byte, error) {
 }
 
 func (s *serializer) createObject(metric telegraf.Metric) string {
-	var m strings.Builder
+	var m bytes.Buffer
 	for fieldName, fieldValue := range metric.Fields() {
 		if isNumeric(fieldValue) {
 			m.WriteString("metric=")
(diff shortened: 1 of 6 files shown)


frankreno:
Ok, looks like we need to support Go 1.9 which does not have StringBuilders. I'll refactor code to account for that.  Any idea on when Telegraf will drop support for 1.9?

plugins/serializers/carbon2/carbon2.go
```diff
@@ -0,0 +1,66 @@
...
+func (s *serializer) createObject(metric telegraf.Metric) string {
+	var m bytes.Buffer
+	for fieldName, fieldValue := range metric.Fields() {
+		if isNumeric(fieldValue) {
+			m.WriteString("metric=")
+			m.WriteString(strings.Replace(metric.Name(), " ", "_", -1))
+			m.WriteString(" field=")
+			m.WriteString(strings.Replace(fieldName, " ", "_", -1))
+			m.WriteString(" ")
+			for k, v := range metric.Tags() {
```
danielnelson: Use TagList() instead, this will give you the tags in sorted order, since the specifications says:

> For intrinsic tags, pick an order and stick with it, otherwise the metric id will change

This will also more performance because it doesn't require a map to be created for each field.

plugins/serializers/carbon2/carbon2.go
```diff
@@ -0,0 +1,66 @@
...
+func (s *serializer) Serialize(metric telegraf.Metric) ([]byte, error) {
+	return []byte(s.createObject(metric)), nil
+}
+
+func (s *serializer) SerializeBatch(metrics []telegraf.Metric) ([]byte, error) {
+	var batch bytes.Buffer
+	for _, metric := range metrics {
+		batch.WriteString(s.createObject(metric))
+	}
+	return []byte(batch.String()), nil
```
danielnelson: Since the bytes.Buffer is no longer modified, you can return return batch.Bytes() to avoid a copy.

sgmeyer:
🥇 I tested this out with numerous inputs and it worked quite well.  I would love to see this merged as our project would like to use this w/ SumoLogic instead of the graphite serializer.  This makes Telegraf/Sumo configuration trivial w/ HTTP output.

plugins/serializers/carbon2/carbon2.go
```diff
@@ -0,0 +1,66 @@
...
+				m.WriteString(" ")
+			}
+			m.WriteString(" ")
+			m.WriteString(fmt.Sprintf("%v", fieldValue))
+			m.WriteString(" ")
+			m.WriteString(strconv.FormatInt(metric.Time().Unix(), 10))
+			m.WriteString("\n")
+		}
+	}
+	return m.String()
```
danielnelson: May be worth returning as []byte to avoid a copy here, or passing the bytes.Buffer in from the calling function.

frankreno:
@danielnelson thanks much for the review and suggestions.  Just pushed the updates, will monitor tests but local tests passed.

danielnelson:
Nice work, thanks!

Issue #3573, opened 2017-12-12
Logparser plugin don't process new lines after telegraf configuration reload.


Isonami:
## Bug report
The newly started telegraf works well, but after configuration reload via SIGHUP, new lines in log aren't parsed or sent to output plugin. Other metrics aren't affected.
### Relevant telegraf.conf:
/etc/telegraf/telegraf.conf
```
[global_tags]

[agent]
  hostname = "hostname"
  omit_hostname = false
  interval = "30s"
  round_interval = true
  metric_batch_size = 1000
  metric_buffer_limit = 10000
  collection_jitter = "0s"
  flush_interval = "30s"
  flush_jitter = "0s"
  logfile = ""
  debug = false
  quiet = false

#
# OUTPUTS:
#
[[outputs.influxdb]]
  database = "telegraf"
  insecure_skip_verify = true
  urls = ["https://influx.hostname"]

#
# INPUTS:
#
[[inputs.cpu]]
  percpu = true
  totalcpu = true
[[inputs.disk]]
[[inputs.io]]
[[inputs.mem]]
[[inputs.net]]
[[inputs.swap]]
[[inputs.system]]
```
/etc/telegraf/telegraf.d/logparser.conf
```
[[inputs.logparser]]
  files = ["/var/log/nginx/web-*-access.log"]
  from_beginning = false
[inputs.logparser.grok]
  measurement = "nginx_access_log"
  patterns = ["\\[[0-9a-z-]+\\] %{CLIENT:client_ip} - %{NOTSPACE} \\[%{HTTPDATE:timestamp:ts-httpd}\\] \"(?:%{WORD:verb:tag} %{NOTSPACE:request}(?: HTTP/%{NUMBER:http_version:float})?|%{DATA})\" %{NUMBER:resp_code:tag} (?:%{NUMBER:resp_bytes:int}|-) \"%{NOTSPACE:referer}\" \"[^\"]+\" \"(?:%{CLIENT:forwarded_for}|-)\" \"(?:%{CLIENT:real_ip}|-)\" %{NOTSPACE} rt=%{NUMBER:request_time:float} uht=(?:%{NUMBER:upstream_header_time:float}|-) urt=(?:%{NUMBER:upstream_response_time:float}|-)"]
  timezone = ""
```
### System info:
Centos 7.2
Telegraf 1.4.3,1.4.5

### Steps to reproduce:

1. Start telegraf
[...]

PR #5347, opened 2019-01-25
Update tail and logparser plugins to use a different tail library

sgtsquiggs:
This new tail plugin fixes the current `tail` plugin problem where it stops reading new lines after telegraf reloads configuration (via SIGHUP). This is documented in ticket #3573 for `logparser` but also affects `tail` since they both use the same underlying log tailing engine.

More presenting this as an idea rather than a real pull request. Ideally both `tail` and `logparser` use the same underlying engine and that engine works after a SIGHUP :smile: 

`tail2` is a copy of `tail` with the change:
* [github.com/influxdata/tail](https://github.com/influxdata/tail) -> [github.com/sgtsquiggs/tail](https://github.com/sgtsquiggs/tail)

github.com/sgtsquiggs/tail is a fork of [github.com/google/mtail](https://github.com/google/mtail) modified to act as a library without unnecessary dependencies

I welcome feedback! Thank you!

### Required for all PRs:

### DIFF
--- a/plugins/inputs/logparser/logparser.go
+++ b/plugins/inputs/logparser/logparser.go
@@ -3,12 +3,19 @@
 package logparser
 
 import (
+	"fmt"
 	"log"
 	"strings"
 	"sync"
+	"time"
 
-	"github.com/influxdata/tail"
+	taillog "github.com/sgtsquiggs/tail/logger"
+	"github.com/sgtsquiggs/tail/logline"
+	"github.com/sgtsquiggs/tail/tailer"
+	"github.com/sgtsquiggs/tail/watcher"
+
 	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/internal"
 	"github.com/influxdata/telegraf/internal/globpath"
 	"github.com/influxdata/telegraf/plugins/inputs"
 	"github.com/influxdata/telegraf/plugins/parsers"
@@ -16,7 +23,8 @@ import (
 )
 
 const (
-	defaultWatchMethod = "inotify"
+	defaultPollInterval = time.Millisecond * 250
+	defaultWatchMethod  = "inotify"
 )
 
 // LogParser in the primary interface for the plugin
@@ -40,10 +48,12 @@ type LogParserPlugin struct {
 	Files         []string
 	FromBeginning bool
 	WatchMethod   string
+	PollInterval  internal.Duration
 
-	tailers map[string]*tail.Tail
-	lines   chan logEntry
-	done    chan struct{}
+	watcher watcher.Watcher
+	tailer  *tailer.Tailer
+	lines   chan *logline.LogLine
+	files   map[string]bool
 	wg      sync.WaitGroup
 	acc     telegraf.Accumulator
 
@@ -69,6 +79,9 @@ const sampleConfig = `
 
   ## Method used to watch for file updates.  Can be either "inotify" or "poll".
   # watch_method = "inotify"
+
+  ## Poll interval. Used when watch_method is set to "poll".
+  # poll_interval = "250ms"
 
   ## Parse logstash-style "grok" patterns:
   [inputs.logparser.grok]
@@ -122,7 +135,7 @@ func (l *LogParserPlugin) Gather(acc telegraf.Accumulator) error {
 	defer l.Unlock()
 
 	// always start from the beginning of files that appear while we're running
-	return l.tailNewfiles(true)
+	return l.tailNewFiles(true)
 }
 
 // Start kicks off collection of stats for the plugin
@@ -131,9 +144,32 @@ func (l *LogParserPlugin) Start(acc telegraf.Accumulator) error {
 	defer l.Unlock()
 
 	l.acc = acc
-	l.lines = make(chan logEntry, 1000)
-	l.done = make(chan struct{})
-	l.tailers = make(map[string]*tail.Tail)
+	l.lines = make(chan *logline.LogLine)
+	l.files = make(map[string]bool)
+
+	var poll bool
+	if l.WatchMethod == "poll" {
+		poll = true
+	}
+
+	var err error
+	l.watcher, err = watcher.NewLogWatcher(l.PollInterval.Duration, !poll, watcher.Logger(taillog.DiscardingLogger))
+	if err != nil {
+		defer close(l.lines)
+		return err
+	}
+
+	opts := []tailer.Option{tailer.Logger(taillog.DiscardingLogger)}
+	if l.FromBeginning {
+		opts = append(opts, tailer.OneShot)
+	}
+
+	l.tailer, err = tailer.New(l.lines, l.watcher, opts...)
+	if err != nil {
+		defer l.watcher.Close()
+		defer close(l.lines)
+		return err
+	}
 
 	mName := "logparser"
 	if l.GrokConfig.MeasurementName != "" {
@@ -152,56 +188,33 @@ func (l *LogParserPlugin) Start(acc telegraf.Accumulator) error {
 		DataFormat:             "grok",
 	}
 
-	var err error
 	l.GrokParser, err = parsers.NewParser(config)
 	if err != nil {
 		return err
 	}
 
 	l.wg.Add(1)
-	go l.parser()
+	go l.parse()
 
-	return l.tailNewfiles(l.FromBeginning)
+	return l.tailNewFiles(l.FromBeginning)
 }
 
 // check the globs against files on disk, and start tailing any new files.
 // Assumes l's lock is held!
-func (l *LogParserPlugin) tailNewfiles(fromBeginning bool) error {
-	var seek tail.SeekInfo
-	if !fromBeginning {
-		seek.Whence = 2
-		seek.Offset = 0
-	}
+func (l *LogParserPlugin) tailNewFiles(fromBeginning bool) error {
 
-	var poll bool
-	if l.WatchMethod == "poll" {
-		poll = true
-	}
-
-	// Create a "tailer" for each file
 	for _, filepath := range l.Files {
 		g, err := globpath.Compile(filepath)
 		if err != nil {
-			log.Printf("E! Error Glob %s failed to compile, %s", filepath, err)
-			continue
+			l.acc.AddError(fmt.Errorf("E! Error Glob %s failed to compile, %s", filepath, err))
 		}
-		files := g.Match()
-
-		for _, file := range files {
-			if _, ok := l.tailers[file]; ok {
+		for _, file := range g.Match() {
+			if _, ok := l.files[file]; ok {
 				// we're already tailing this file
 				continue
 			}
 
-			tailer, err := tail.TailFile(file,
-				tail.Config{
-					ReOpen:    true,
-					Follow:    true,
-					Location:  &seek,
-					MustExist: true,
-					Poll:      poll,
-					Logger:    tail.DiscardingLogger,
-				})
+			err := l.tailer.TailPath(file)
 			if err != nil {
 				l.acc.AddError(err)
 				continue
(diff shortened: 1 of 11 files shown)


danielnelson:
It would be have some of these long standing bugs fixed, but we will need to do more testing than usual to avoid any regressions.  Logparser is going to be deprecated, everything it does can be done by the tail plugin.

Can you submit this as a replacement for the tail plugin?

danielnelson:
I did a quick fix to skip the Go1.9 build, we will probably remove support for this version of Go in 1.12.0.
https://github.com/influxdata/telegraf/commit/42936e1e917bf086506fd541d39eddbba6ccb987

Planning to distribute this PR out to those having issues with the current tail plugin so we can begin gathering some usage reports.

sgtsquiggs:
I need to update this; I've come across and fixed a few bugs since submitting.

sgtsquiggs:
I am reworking this PR to replace inputs/tail and modify inputs/logparser.

sgtsquiggs:
This is ready for review!

sgtsquiggs:
I _could_ make this build for 1.9 by extracting `os.SetReadDeadline` out so I can tag it

sgtsquiggs:
For maximum compatibility I could modify the tail library to build w/o fsnotify for solaris. It would be poll only.

sgtsquiggs:
Rebasing this change on master

sgtsquiggs:
The tail library used in this change (github.com/sgtsquiggs/tail) is fairly verbose, especially at `Info` log level. I have [disabled logging in the library](https://github.com/sgtsquiggs/telegraf/blob/19c49ccbd5e3b41ff5b4f363c6fb9d92aa75183b/plugins/inputs/tail/tail.go#L103) and will only log errors that bubble up to telegraf:

https://github.com/sgtsquiggs/telegraf/blob/19c49ccbd5e3b41ff5b4f363c6fb9d92aa75183b/plugins/inputs/tail/tail.go#L118
https://github.com/sgtsquiggs/telegraf/blob/19c49ccbd5e3b41ff5b4f363c6fb9d92aa75183b/plugins/inputs/tail/tail.go#L147
https://github.com/sgtsquiggs/telegraf/blob/19c49ccbd5e3b41ff5b4f363c6fb9d92aa75183b/plugins/inputs/tail/tail.go#L208

plugins/inputs/tail/tail.go
```diff
@@ -179,43 +183,35 @@ func (t *Tail) receiver(parser parsers.Parser, tailer *tail.Tail) {
...
 
 func (t *Tail) Stop() {
 	t.Lock()
 	defer t.Unlock()
 
-	for _, tailer := range t.tailers {
-		err := tailer.Stop()
-		if err != nil {
-			t.acc.AddError(fmt.Errorf("E! Error stopping tail on file %s\n", tailer.Filename))
+	err := t.tailer.Close()
```
danielnelson: Can you describe what, if anything, on tailer can be used concurrently?  Maybe we don't need the locks anymore.  Telegraf will never call Start(), Gather(), Stop() at the same time, so we only need to worry about Stop and any processing goroutines.

plugins/inputs/tail/tail.go
```diff
@@ -179,43 +183,35 @@ func (t *Tail) receiver(parser parsers.Parser, tailer *tail.Tail) {
...
+	if err != nil {
+		t.acc.AddError(fmt.Errorf("E! Error closing tailer, Error: %s\n", err))
+	} else {
+		for file := range t.files {
+			log.Printf("D! [inputs.logparser] tail removed for file: %v", file)
```
danielnelson:
```diff
- inputs.logparser
+ inputs.tail
```

plugins/inputs/tail/tail.go
```diff
@@ -86,89 +91,88 @@ func (t *Tail) Start(acc telegraf.Accumulator) error {
...
-				tail.Config{
-					ReOpen:    true,
-					Follow:    true,
-					Location:  seek,
-					MustExist: true,
-					Poll:      poll,
-					Pipe:      t.Pipe,
-					Logger:    tail.DiscardingLogger,
-				})
+			err := t.tailer.TailPath(file)
```
danielnelson: If the file doesn't exist this returns an error, normally globpath.Match only returns matches, but if no globs are in use it just returns the path.  Even if we changed Match then its possible that the file would go away before we can tail, so I think we need to check if the file is not found or permission error occurs here and just ignore it.

I think ideally this function could return one of these known errors https://golang.org/pkg/os/#pkg-variables

plugins/inputs/logparser/logparser.go
```diff
@@ -131,9 +145,32 @@ func (l *LogParserPlugin) Start(acc telegraf.Accumulator) error {
 	defer l.Unlock()
 
 	l.acc = acc
-	l.lines = make(chan logEntry, 1000)
-	l.done = make(chan struct{})
-	l.tailers = make(map[string]*tail.Tail)
+	l.lines = make(chan *logline.LogLine)
```
danielnelson: Really sorry about this, but actually let's just leave logparser on hpcloud/tail since I'm going to deprecate it anyway.

plugins/inputs/tail/tail.go
```diff
@@ -86,89 +91,88 @@ func (t *Tail) Start(acc telegraf.Accumulator) error {
...
 	var poll bool
 	if t.WatchMethod == "poll" {
 		poll = true
 	}
 
-	// Create a "tailer" for each file
+	var err error
+	t.watcher, err = watcher.NewLogWatcher(t.PollInterval.Duration, !poll, watcher.Logger(taillog.DiscardingLogger))
+	if err != nil {
+		defer close(t.lines)
```
danielnelson: I'd skip defer here and elsewhere in this function: ordering is more confusing, it is more expensive, and we are going to return anyway.

plugins/inputs/tail/tail.go
```diff
@@ -86,89 +91,88 @@ func (t *Tail) Start(acc telegraf.Accumulator) error {
...
+	t.tailer, err = tailer.New(t.lines, t.watcher, opts...)
+	if err != nil {
+		defer t.watcher.Close()
+		defer close(t.lines)
+		return err
+	}
+
+	parser, err := t.parserFunc()
+	if err != nil {
+		defer t.tailer.Close()
```
danielnelson: Do we need to also close lines/watcher?

plugins/inputs/tail/tail.go
```diff
@@ -1,4 +1,5 @@
 // +build !solaris
+// +build go1.10
```
danielnelson: I removed support for Go 1.9 in #6026, so you can remove this now.

sgtsquiggs:
Hi @danielnelson, I'm on vacation from June 21-29; I'll get on this when I'm back!

danielnelson:
I also notice that with inotify it is a little slow to shutdown, with strace or turning on logging it appears to watch all files recursively in the entire directory.  We will have to find a way to reduce this.

plugins/inputs/tail/README.md
```diff
@@ -36,8 +40,6 @@ The plugin expects messages in one of the
   files = ["/var/mymetrics.out"]
   ## Read file from beginning.
   from_beginning = false
-  ## Whether file is a named pipe
-  pipe = false
```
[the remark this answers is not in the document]
sgtsquiggs: No; the tail library checks the file mode for `io.ModeNamedPipe` and acts accordingly.

plugins/inputs/tail/tail.go
```diff
@@ -86,89 +91,88 @@ func (t *Tail) Start(acc telegraf.Accumulator) error {
...
-			parser, err := t.parserFunc()
-			if err != nil {
-				t.acc.AddError(fmt.Errorf("error creating parser: %v", err))
-			}
-
 			// create a goroutine for each "tailer"
-			t.wg.Add(1)
-			go t.receiver(parser, tailer)
-			t.tailers[tailer.Filename] = tailer
+			t.files[file] = true
```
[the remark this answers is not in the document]
sgtsquiggs: Currently the only way `tail` stops following a file is if that file is deleted. However, I should be removing files from that map when `Stop()` is called. I'll fix that.


### To potentially solve the potential slow resource leak problem:

The tail library I am using supports garbage collection, but is disabled by default: https://github.com/sgtsquiggs/tail/blob/master/tailer/tail.go#L385-L418

This may be useful when tailing a logrotated file:

assuming `paths = ["/var/log/apache.log"]`, once `apache.log` rotates to `apache.log.1` telegraf will continue tailing `apache.log.1` and will additionally tail `apache.log` when once Gather is next called. With GC, the tail lib will stop tailing `apache.log.1` after 24h of inactivity. It won't restart tailing `apache.log.1` unless it matches a glob or specific pathname.

This garbage collection feature could be exposed as a interval - "stop tailing files after **interval** of inactivity", which would reset when the plugin is started (upon start or after SIGHUP).

plugins/inputs/tail/tail.go
sgtsquiggs: The tail lib handles glob patterns itself; perhaps I remove the globbing code from the plugin and rely on the tail library.
https://github.com/sgtsquiggs/tail/blob/master/tailer/tail.go#L152-L181

I wrote it the way it is now so that I could easily track which files are being watched in the plugin. I believe a better way would be to have the tail lib expose callbacks for when a file is watched/unwatched. Additionally it would allow the plugin to accurately report when a file is deleted.

sgtsquiggs: W/r to the errors; the tail lib is [wrapping](https://godoc.org/github.com/pkg/errors#Wrapf) the os errors with a custom error; I can check for underlying error and ignore.

plugins/inputs/tail/tail.go
sgtsquiggs: t.tailer.Close() will close lines and watcher; I'll add comments noting this.

plugins/inputs/tail/tail.go
```diff
@@ -86,89 +91,88 @@ func (t *Tail) Start(acc telegraf.Accumulator) error {
...
+	t.files = make(map[string]bool)
 
 	var poll bool
 	if t.WatchMethod == "poll" {
 		poll = true
 	}
 
-	// Create a "tailer" for each file
+	var err error
+	t.watcher, err = watcher.NewLogWatcher(t.PollInterval.Duration, !poll, watcher.Logger(taillog.DiscardingLogger))
```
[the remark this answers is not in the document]
sgtsquiggs: Noted! Forgot that fsnotify wraps more than just inotify.

sgtsquiggs:
TODO: I also need to update the provided `etc/telegraf.conf` with updated config

sgtsquiggs:
@danielnelson I think I am going to abandon this request. The underlying issue (https://github.com/influxdata/telegraf/issues/3573) I was attempting to solve seems to be resolved as of 1.11.0. I can no longer reproduce.

The tail library (https://github.com/sgtsquiggs/tail) handles watching very differently than hpcloud/tail. It expects to handle all pattern matching itself, and **recursively** watches parent directories _by default_ to ensure it finds new files matching those patterns. It would take a lot more work to make it work more similarly to hpcloud/tail. For what I need, the existing tail plugin works fine.


I think my time would be better spent improving the existing tail plugin rather than replacing it. One feature I would like to add is the ability for telegraf to resume tailing a file from where it left off after receiving SIGHUP.



Thoughts?

danielnelson:
We can set this pull request aside for now and we can always pick it up again later if it makes sense.

Right now the open issues on logparser/tail have the label `area/grok`.  I should probably split this into `area/grok` and `area/tail`, but take a look through the open issues to make sure nothing too major will hit you.

PR #5365, opened 2019-02-01
Improve testutil AssertDoesNotContainsTaggedFields

gracedo:
Improved the testutil accumulator's `AssertDoesNotContainsTaggedFields` method to allow testing for specific fields that don't have a set of tags (tags that may exist in the same measurement under different fields). Otherwise such a test within the same measurement type isn't possible (checking that the same set of tags exists on some fields but not others).

I think the method as is would have been more accurately named `AssertDoesNotContainsTags` since it is mainly testing for tags. If it's better to instead create a `AssertDoesNotContainsTags` method with the original logic, and `AssertDoesNotContainsTaggedFields` with the changes below, I am happy to make that change!

### Required for all PRs:

### DIFF
--- a/testutil/accumulator.go
+++ b/testutil/accumulator.go
@@ -340,9 +340,10 @@ func (a *Accumulator) AssertDoesNotContainsTaggedFields(
 			continue
 		}
 
-		if p.Measurement == measurement {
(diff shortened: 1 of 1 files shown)


danielnelson:
Thanks!

Issue #5364, opened 2019-02-01
Error writing out of order points to Stackdriver output


Legogris:
### Relevant telegraf.conf:
```
[agent]
    interval = "30s"
    round_interval = true
    flush_interval = "60s"
    flush_jitter = "0s"
    debug = false
    hostname = ""
```

### System info:
Telegraf 1.9.0 (git: HEAD 2bf21c0d)
OS: Amazon Linux (`4.14.77-81.59.amzn2.x86_64 #1 SMP Mon Nov 12 21:32:48 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux`)

Stackdriver requires points to be written in order - it seems that this is not guaranteed by the stackdriver output plugin:

```
2019-01-29T00:30:00Z E! [output.stackdriver] unable to write to Stackdriver: rpc error: code = InvalidArgument desc = One or more TimeSeries could not be written: Points must be written in order. One or more of the points specified had an older end time than the most recent point.: timeSeries[0]
2019-01-29T00:30:00Z E! [agent] Error writing to output [stackdriver]: rpc error: code = InvalidArgument desc = One or more TimeSeries could not be written: Points must be written in order. One or more of the points specified had an older end time than the most recent point.: timeSeries[0]
2019-01-29T00:31:00Z E! [output.stackdriver] unable to write to Stackdriver: rpc error: code = InvalidArgument desc = One or more TimeSeries could not be written: Points must be written in order. One or more of the points specified had an older end time than the most recent point.: timeSeries[0]
2019-01-29T00:31:00Z E! [agent] Error writing to output [stackdriver]: rpc error: code = InvalidArgument desc = One or more TimeSeries could not be written: Points must be written in order. One or more of the points specified had an older end time than the most recent point.: timeSeries[0]
2019-01-29T00:32:00Z E! [output.stackdriver] unable to write to Stackdriver: rpc error: code = InvalidArgument desc = One or more TimeSeries could not be written: Points must be written in order. One or more of the points specified had an older end time than the most recent point.: timeSeries[0]
2019-01-29T00:32:00Z E! [agent] Error writing to output [stackdriver]: rpc error: code = InvalidArgument desc = One or more TimeSeries could not be written: Points must be written in order. One or more of the points specified had an older end time than the most recent point.: timeSeries[0]
[...]

PR #5385, opened 2019-02-06
Stackdriver sort order

danielnelson:
This change attempts to mitigate #5364 by sorting the batch, but doesn't solve the issue.  Remaining issues are documented in the README and some errors are ignored, resulting in some points being dropped, but attempting to keep the output from becoming stuck.

related #5364

### Required for all PRs:

### DIFF
--- a/plugins/outputs/stackdriver/stackdriver.go
+++ b/plugins/outputs/stackdriver/stackdriver.go
@@ -41,8 +41,9 @@ const (
 	// MaxInt is the max int64 value.
 	MaxInt = int(^uint(0) >> 1)
 
-	errStringPointsOutOfOrder = "One or more of the points specified had an older end time than the most recent point"
-	errStringPointsTooOld     = "Data points cannot be written more than 24h in the past"
+	errStringPointsOutOfOrder  = "One or more of the points specified had an older end time than the most recent point"
+	errStringPointsTooOld      = "Data points cannot be written more than 24h in the past"
+	errStringPointsTooFrequent = "One or more points were written more frequently than the maximum sampling period configured for the metric"
 )
 
 var sampleConfig = `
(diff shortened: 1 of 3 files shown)


danielnelson:
@Legogris This isn't a full fix for the issue, but I think it should help.  Do you think you could try it out and see if you run into any major problem?  Here are some packages:

- [telegraf-1.10.0~f48d387e-0.x86_64.rpm](https://27039-33258973-gh.circle-artifacts.com/0/build/linux/amd64/telegraf-1.10.0~f48d387e-0.x86_64.rpm)
- [telegraf_1.10.0~f48d387e-0_amd64.deb](https://27039-33258973-gh.circle-artifacts.com/0/build/linux/amd64/telegraf_1.10.0~f48d387e-0_amd64.deb)
- [telegraf-1.10.0~f48d387e_linux_amd64.tar.gz](https://27039-33258973-gh.circle-artifacts.com/0/build/linux/amd64/telegraf-1.10.0~f48d387e_linux_amd64.tar.gz)

Legogris:
Awesome @danielnelson, will try it out :)

Legogris:
@danielnelson Been running it for a bit, no issues so far. Once this is included in a release we will be able to roll it out to production and see how it behaves there. Really appreciate the quick patch, thanks! We do have an interval of 30s so if we see issues after this we will look into increasing it to 60s or using `basicstats`.

plugins/outputs/stackdriver/stackdriver.go
```diff
@@ -70,17 +76,31 @@ func (s *Stackdriver) Connect() error {
 	return nil
 }
 
+// sorted returns a copy of the metrics in time ascending order
+func sorted(metrics []telegraf.Metric) []telegraf.Metric {
+	batch := make([]telegraf.Metric, 0, len(metrics))
+	for i := len(metrics) - 1; i >= 0; i-- {
+		batch = append(batch, metrics[i])
+	}
+	sort.Slice(batch, func(i, j int) bool {
```
glinton: Why not just `sort.Slice` `metrics`, instead of duplicating it into `batch`?

danielnelson: That's a good question, the reason is that outputs are not allowed to modify the batch because if an error occurs the metrics will not be restored to the metric buffer correctly.  This is an unfortunate aspect of the output interface, it does save a copy of the buffer but it might be worth the copy to prevent misuse.

PR #5392, opened 2019-02-07
Add `fieldvaluepass` and `fieldvaluedrop` to select by field value

### LINKED ISSUE #3904: Removing entire metric base on a field value.
### Proposal:
Add option that removes an entire metric based on a field value. Add option that removes a field based on value (I don't need it currently but it is good to have). 
### Current behavior:
We can't drop entire metric based on field. There is [workaround ](https://github.com/influxdata/telegraf/issues/2176#issuecomment-372974585) to drop all fields but it doesn't look like good approach and causes error messages in telegraf.log file. 
### Desired behavior:
Be able to drop entire metric based on field. Regex support to match at least the beginning of the field is desirable to drop same messages with differences in IPs, ID's, etc. 
### Use case: 
Excluding some logs from being parsed and recorded to influxdb based on error message field. I can't drop all logs from certain program. I have to filter them based on message.
Deeper overview of this [use case](https://github.com/influxdata/telegraf/issues/2176#issuecomment-369565606)

sgtsquiggs:
Resolves #3904.

Implements a field value filter as is currently implemented for tags.

The new filter names are `fieldvaluepass` and `fieldvaluedrop`. They work the same as `tagpass` and `tagdrop`, but for fields.

This implementation is limited. It treats all non-string field values as strings via `fmt.Sprintf("%v", value)`.

### Required for all PRs:

### DIFF
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -382,6 +382,15 @@ a pattern in this list are emitted.
 The inverse of `namepass`.  If a match is found the metric is discarded. This
 is tested on metrics after they have passed the `namepass` test.
 
+- **fieldvaluepass**:
+A table mapping field keys to arrays of glob pattern strings.  Only metrics
+that contain a field key in the table and a field value matching one of its
+patterns is emitted.
+
+- **fieldvaluedrop**:
+The inverse of `fieldvaluepass`.  If a match is found the metric is discarded. This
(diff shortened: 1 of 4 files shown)


danielnelson:
Thanks for sharing this code but I think we will probably not merge since I would like to do something more comprehensive for filtering.  A workaround that helps in most cases is to use the converter processor to convert the field to a tag and then you can use tagdrop.

PR #5396, opened 2019-02-08
Change log level of buffer too small error

glinton:
Provides the ability to sort/search for errors that relate to the input being oversized.


danielnelson:
I think we should add a selfstat with a counter field `entity_too_large` instead.  I don't want to make this warning too noisy since there are limited methods for silencing it right now.

PR #5402, opened 2019-02-11
Add connection info per process to procstat

adrianlzt:
Only added where gopsutil supports returning connections by PID: linux,
freebsd and darwin.

Add these new fields, following the same naming as netstat:
tcp_close
tcp_close_wait
tcp_closing
tcp_established
tcp_fin_wait1
tcp_fin_wait2
tcp_last_ack
tcp_listen
tcp_none
tcp_syn_recv
tcp_syn_sent
tcp_time_wait
udp_socket

### Required for all PRs:


danielnelson:
I think we need to begin adding monitoring components to this plugin, to allow users to be able to skip collection of metrics they are not interested in.  How about we add a new option:
```
metrics_include = ["connections"]
```

If the option is unset then we collect only the current items.

adrianlzt:
We can do the same with:
```
fielddrop = ["tcp_*", "udp_socket"]]

danielnelson:
I more concerned about slowing down the collection speed as we parse more files.

adrianlzt:
Superseded by #8755

Issue #5404, opened 2019-02-11
Aggregate Stackdriver output requests


Legogris:
## Feature Request

Stackdriver currently send a single request for each TimeSeries in a `write`. This will hit Stackdriver rate limits pretty early. The feedback we got from Google was that we were abusing their API using the current version.

### Proposal:
Aggregate separate TimeSeries in batch requests

### Current behavior:
Each TimeSeries is sent as a single request: https://github.com/influxdata/telegraf/blob/516a5898fd22f3480e506714f1244c043d29bfd3/plugins/outputs/stackdriver/stackdriver.go#L176

### Desired behavior:
TimeSeries are aggregated together. The reason this is not as simple as just aggregating all TimeSeries passed to `write` is that if more than one of these contain points for the same TimeSeries, this will yield an error: `Duplicate TimeSeries encountered. Only one point can be written per TimeSeries per request.` – refers to an attempt to write more than one point to a single time series in the same request, which is not allowed. 

It should also be noted that each request can contain a maximum of 200 TimeSeries.

danielnelson:
It seems that we first need to group all metrics by minute, and then aggregate the metrics.  Do you have any thoughts as to what aggregation would be most appropriate?  We may want to just do a "last" aggregation to send the most recent point.

Just to clarify with respect to the 200 time series per request, you are suggesting that we pack the full 200 time series into the request before calling CreateTimeSeries instead of the current strategy of sending one telegraf.Metric per call?

Legogris:
> It seems that we first need to group all metrics by minute, and then aggregate the metrics. Do you have any thoughts as to what aggregation would be most appropriate? We may want to just do a "last" aggregation to send the most recent point.
I'm really not the expert here, but is there anything special about minutes here?


> Just to clarify with respect to the 200 time series per request, you are suggesting that we pack the full 200 time series into the request before calling CreateTimeSeries instead of the current strategy of sending one telegraf.Metric per call?
Ideally each request should contain as many TimeSeries as possible, just wanted to put it in case an implementer misses this part and attempts to send over 200 TimeSeries per request.

Legogris:
I made a first attempt in #5407, happy to hear your thoughts on the best way to go about this @danielnelson .

danielnelson:
In the [quotas](https://cloud.google.com/monitoring/quotas#custom_metrics_quotas) documentation I found:

> Rate at which data can be written to a single time series: one point per minute

Legogris:
[...]

PR #5407, opened 2019-02-11
Aggregate multiple Stackdriver TimeSeries in requests

Legogris:

This is a first attempt at #5404. This is not yet perfect, as it just does an iteration over the metrics/fields, adding TimeSeries until adding the next one would mean a duplicate, in which case it sends off the batch so far and then proceeds with a new request.
This should still be an improvement over the current situation (it is not yet properly tested in practice, though)

### DIFF
--- a/plugins/outputs/stackdriver/stackdriver.go
+++ b/plugins/outputs/stackdriver/stackdriver.go
@@ -113,19 +113,25 @@ func sorted(metrics []telegraf.Metric) []telegraf.Metric {
 	return batch
 }
 
-func getFieldHash(m telegraf.Metric, f *telegraf.Field) uint64 {
+type timeSeriesBuckets map[uint64][]*monitoringpb.TimeSeries
+
+func (tsb timeSeriesBuckets) Add(m telegraf.Metric, f *telegraf.Field, ts *monitoringpb.TimeSeries) {
 	h := fnv.New64a()
 	h.Write([]byte(m.Name()))
-	h.Write([]byte("\n"))
+	h.Write([]byte{'\n'})
 	h.Write([]byte(f.Key))
-	h.Write([]byte("\n"))
+	h.Write([]byte{'\n'})
 	for key, value := range m.Tags() {
 		h.Write([]byte(key))
-		h.Write([]byte("\n"))
+		h.Write([]byte{'\n'})
 		h.Write([]byte(value))
-		h.Write([]byte("\n"))
+		h.Write([]byte{'\n'})
 	}
-	return h.Sum64()
+	k := h.Sum64()
+
+	s := tsb[k]
+	s = append(s, ts)
+	tsb[k] = s
 }
 
 // Write the metrics to Google Cloud Stackdriver.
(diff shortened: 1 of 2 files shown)


nicbaz:
hello @danielnelson,

Referring to #5404 what are your thoughts on this implementation?

plugins/outputs/stackdriver/stackdriver.go
```diff
@@ -151,25 +171,49 @@ func (s *Stackdriver) Write(metrics []telegraf.Metric) error {
...
+	}
 
-		if len(timeSeries) < 1 {
-			continue
+	// process the buckets in order
+	keys := make([]uint64, 0, len(buckets))
+	for k := range buckets {
+		keys = append(keys, k)
+	}
+	sort.Slice(keys, func(i, j int) bool { return keys[i] < keys[j] })
```
danielnelson: Why do we need to sort?

plugins/outputs/stackdriver/stackdriver.go
```diff
@@ -151,25 +171,49 @@ func (s *Stackdriver) Write(metrics []telegraf.Metric) error {
...
+
+	for len(buckets) != 0 {
+		// can send up to 200 time series to stackdriver
+		timeSeries := make([]*monitoringpb.TimeSeries, 0, 200)
+		for _, k := range keys {
+			s := buckets[k]
+			timeSeries = append(timeSeries, s[0])
+			if len(s) == 1 {
+				delete(buckets, k)
```
danielnelson: This removes the key from the buckets map, but it will still come up on subsequent iterations over keys.  Try the new test with these metrics and it should panic:
```
	metrics := []telegraf.Metric{
		testutil.MustMetric("cpu",
			map[string]string{
				"foo": "bar",
			},
			map[string]interface{}{
				"value": 42,
			},
			time.Unix(2, 0),
		),
		testutil.MustMetric("cpu",
			map[string]string{
				"foo": "foo",
			},
			map[string]interface{}{
				"value": 43,
			},
			time.Unix(1, 0),
		),
		testutil.MustMetric("cpu",
			map[string]string{
				"foo": "bar",
			},
			map[string]interface{}{
				"value": 43,
			},
			time.Unix(1, 0),
		),
	}
```

plugins/outputs/stackdriver/stackdriver.go
nicbaz: In order to have a predictable output (request order) since maps ordering is not guaranteed.

plugins/outputs/stackdriver/stackdriver.go
nicbaz: looking into this

plugins/outputs/stackdriver/stackdriver.go
danielnelson: Okay

plugins/outputs/stackdriver/stackdriver.go
nicbaz: good catch, wasn't an issue before using a sorted list of keys :sweat_smile:

nicbaz:
@danielnelson fixed :+1:

danielnelson:
@Legogris I should have checked this already, but I don't have you listed as having signed our CLA.  Could you fill it out, if you already did it once could you do it again as something must have gone wrong on our end.

https://www.influxdata.com/legal/cla/

Legogris:
> @Legogris I should have checked this already, but I don't have you listed as having signed our CLA. Could you fill it out, if you already did it once could you do it again as something must have gone wrong on our end.
> 
> https://www.influxdata.com/legal/cla/

@danielnelson Signed (again, I think)! :+1:

PR #5413, opened 2019-02-12
Check if SNMP6 directory exists (#5403)

### LINKED ISSUE #5403: [inputs/nstat] error on snmp6 when ipv6 is disabled
### Relevant telegraf.conf:

```
[[inputs.nstat]]
```
### System info:

Debian 4.9.110-3+deb9u4 (2018-08-21), Telegraf 1.9.4

### Steps to reproduce:

1. Linux kernel without IPv6 support ` ipv6.disable=1`

### Expected behavior:

Silently ignore missing `/proc/net/snmp6`.

### Workaround

~~redirect `snmp6` to `snmp` and then ignore it.~~

~~`[[inputs.nstat]]`~~
~~`  proc_net_snmp6 = "/proc/net/snmp"`~~
~~`  namedrop = "snmp6"`~~

Not a good idea. Incorrectly parsed `snmp6` fields with enormous cardinality:

```
> SHOW FIELD KEY CARDINALITY FROM nstat
name: nstat
count
-----
389040
```

### Actual behavior:

```
E! [inputs.nstat]: Error in plugin: open /proc/net/snmp6: no such file or directory
```

deric:
Silently ignore missing `/proc/stat/snmp6` directory

### Required for all PRs:

Resolves #5403

### DIFF
--- a/plugins/inputs/nstat/nstat.go
+++ b/plugins/inputs/nstat/nstat.go
@@ -84,15 +84,14 @@ func (ns *Nstat) Gather(acc telegraf.Accumulator) error {
 	}
 
 	// collect SNMP6 data, if SNMP6 directory exists (IPv6 enabled)
-	if _, err := os.Stat(ns.ProcNetSNMP6); err == nil {
-		snmp6, err := ioutil.ReadFile(ns.ProcNetSNMP6)
-		if err != nil {
-			return err
-		}
+	snmp6, err := ioutil.ReadFile(ns.ProcNetSNMP6)
+	if err == nil {
 		err = ns.gatherSNMP6(snmp6, acc)
 		if err != nil {
 			return err
 		}
+	} else if !os.IsNotExist(err) {
+		return err
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


plugins/inputs/nstat/nstat.go
```diff
@@ -83,14 +83,16 @@ func (ns *Nstat) Gather(acc telegraf.Accumulator) error {
...
-	snmp6, err := ioutil.ReadFile(ns.ProcNetSNMP6)
-	if err != nil {
-		return err
-	}
-	err = ns.gatherSNMP6(snmp6, acc)
-	if err != nil {
-		return err
+	// collect SNMP6 data, if SNMP6 directory exists (IPv6 enabled)
+	if _, err := os.Stat(ns.ProcNetSNMP6); err == nil {
+		snmp6, err := ioutil.ReadFile(ns.ProcNetSNMP6)
```
danielnelson: This introduces a race condition between the `os.Stat` call and the `os.Open` call within `ioutil.ReadFile`.  This is one of those situations where you need to ask forgiveness instead of permission, can you try this:
```go
	snmp6, err := ioutil.ReadFile(ns.ProcNetSNMP6)
	if err != nil && !os.IsNotExist(err) {
		return err
	}
```

deric: Right, good catch. I've refactored the code.

PR #5433, opened 2019-02-14
Loggregator forwarder agent input plugin

### LINKED ISSUE #5432: [Feature Request] Provide plugin to consume Cloud Foundry metrics into Telegraf via the Loggregator API
## Feature Request

### Proposal:

We are proposing a plugin that will allow Telegraf to consume metrics from the Cloud Foundry platform. Cloud Foundry uses an open source API for Metrics and Logs known as "Loggregator". In more recent versions of the Cloud Foundry platform, a new pattern has been introduced known as the [Loggregator Forwarder Agent](https://github.com/cloudfoundry/loggregator-agent-release). This architecture allows logs and metrics coming off a single VM to be multiplexed to N number of downstream consumers who implement the Loggregator API. By developing a plugin for Telegraf that implements this API, we hope to enable Cloud Foundry operators to deploy a Telegraf process to each of their Cloud Foundry VMs, thus allowing them to harness the power of Telegraf and output their metrics to any of the many support Telegraf outputs.

### Current behavior:

This plugin does not currently exist.

### Desired behavior:

This plugin exists and can be turned on for anyone using Telegraf.

A Cloud Foundry operators can deploy a Telegraf process to each of their Cloud Foundry VMs and export metrics to any of the Telegraf outputs.

trevorwhitney:
### Required for all PRs:

Fixes #5432


aegershman:
This is really interesting. Is the intent for Cloud Foundry operators to configure this as a bosh vm extension? Or would this basically come for free as a part loggregator? Noticing it's listening on localhost only (https://github.com/influxdata/telegraf/pull/5433/files#diff-7242947cb0ee2aba6e3d1d79b55f2d5cR76) so I'm assuming this is to be run as a bosh-managed process and not something like an app pulling from the firehose?

Looking forward to seeing how this will be used in interesting ways by teams

trevorwhitney:
@aegershman the plan, for now, is that Telegraf would be setup as a BOSH runtime configuration targeting all VMs with the Loggregator forwarder agent process. This would install Telegraf on all VMs with a Loggregator forwarder agent, which will pick up the Telegraf Loggregator forwarder agent's ingress port from a configuration file. Our longterm hope is that if we can prove this pattern via a runtime configuration, we may eventually be able to PR Telegraf into `cf-deployment` itself.

trevorwhitney:
Hey all, quick update on this. We have been unable to rely on the availability of feature in CF that this plugin depends on, and have been focusing most of our efforts on PR #5471. This might still be valuable for the open source cloud foundry community, but it has fallen behind in features compared to the RLP input, so I just wanted to double check that everyone was still comfortable moving forward with this? I think it is still of production quality, but has fallen in our priority list.

danielnelson:
What would be the argument for having both of these plugins instead of only #5471?

PR #5436, opened 2019-02-14
add prometheus histogram aggregator plugin

### LINKED ISSUE #5435: [Feature Request] Add histogram aggregator compatible with Prometheus
## Feature Request

### Proposal:

We are proposing a new aggregator plugin that will aggregate gauge metrics into Prometheus histograms. This will allow the aggregation of asynchronous metrics from any input into Prometheus compatible histograms (that can be used with any output but are designed specifically for the Prometheus output).

### Current behavior:

This plugin doesn't currently exist. The existing histogram aggregator does not output `Histogram` type metrics. It also doesn't keep track of sum or count.

### Desired behavior:

A histogram aggregator that can build Prometheus compatible histograms based on gauge metrics. 

### Use case:

Aggregate asynchronous gauge metrics to reduce output size and avoid earlier values being overwritten. Example: an app emits a gauge metric for request duration each time a request is made to its API. Using an output like the Prometheus output where only the last gauge value received is report, there will be a large amount of metric loss. Additionally, if the app is used heavily, there could be a huge number of these metrics making aggregation important.

kitemongerer:
### Required for all PRs:

Fixes #5435

### DIFF
--- a/README.md
+++ b/README.md
@@ -323,6 +323,7 @@ For documentation on the latest development code see the [documentation index][d
 * [basicstats](./plugins/aggregators/basicstats)
 * [minmax](./plugins/aggregators/minmax)
 * [histogram](./plugins/aggregators/histogram)
+* [prometheus_histogram](./plugins/aggregators/prometheus_histogram)
 * [valuecounter](./plugins/aggregators/valuecounter)
 
 ## Output Plugins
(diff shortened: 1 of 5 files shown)


danielnelson:
Closing, see https://github.com/influxdata/telegraf/issues/5435#issuecomment-571871206

Issue #4942, opened 2018-10-31
influx format rejects newlines in quoted field values


awelzel:
Hey hey,

I am trying to send log messages through telegraf via UDP in the influx line format. It appears the telegraf influx data parser does not accept newlines inside field values - not even inside quoted field values. I believe this should be allowed. In fact, using the same steps to reproduce the issues directly with an influxdb udp listener setup works as expected - newlines are stored in the database.

### Relevant telegraf.conf:
```
[[inputs.socket_listener]]
  service_address = "udp://:9089"
  data_format = "influx"
```

### influxdb.conf for comparison
```
[[udp]]
  enabled = true
  bind-address = ":8089"
```

### System info:

Telegraf: master (d2717f7)
Debian buster/sid

### Steps to reproduce:
```
    # Sending to telegraf. Note the -e
    $ echo -e 'test_logs message="Hey!\n"' | nc -u localhost 9089

    # telegraf log:
    2018-10-31T12:10:10Z E! Error in plugin [inputs.socket_listener]: unable to parse incoming packet: metric parse error: expected field at offset 23: "test_logs message=\"Hey!\n\"\n"
```
Sending to influxdb directly works as expected:
```
    $ echo -e 'test_logs message="Hey!\nhow are you?"' | nc -u localhost 8089
    > select * from test_logs;
    name: test_logs
    time                message
    ----                -------
    1540988837240091217 Hey!
    how are you?
```

### Expected behavior:

The influx parser should accept newlines inside quoted field values.

### Actual behavior:

The parser rejects newlines inside quoted field values.

### Additional info:
From what I could gather from the code: Telgraf simply scans for a newline character in the received data to determine where a influx line starts and terminates. The influxdb code seems more advanced in that it handles quoted fields: https://github.com/influxdata/influxdb/blob/master/models/points.go#L1081

### Background:
I'm attempting to send log messages (and particularly tracebacks with newlines) using a Python logging handler [1]. Previously I had only tested against influxdb directly, but in the meantime have a use case to send logs through telegraf which should then get forwarded to influxdb.

[1] https://github.com/awelzel/influxpy

danielnelson:
As a workaround string fields support the `\n` escape to insert a CR.

awelzel:
> As a workaround string fields support the `\n` escape to insert a CR.

Daniel, thanks for the quick reply! Hmm, I have just tried sending a ``\n`` (note the missing ``-e`` for ``echo``).
```
[...]

PR #5439, opened 2019-02-15
Fix InfluxDB UDP line splitting (#5438)

### LINKED ISSUE #5438: InfluxDB UDP output plugin not splitting large measurements
Telegraf's UDP output is not splitting outgoing measurement lines based on the udp_payload. It seems like the this block is never executed as serializer alway exists:  https://github.com/influxdata/telegraf/blob/master/plugins/outputs/influxdb/udp.go#L46

### Relevant telegraf.conf:
```
[[outputs.influxdb]]
  urls = ['udp://metrics-udp:22001/']
  database = "tg_udp"
  precision = "s"
  timeout = "5s"
  user_agent = "telegraf"
  udp_payload = 1400
```
### System info:
Tested on Telegraf 1.9.4 but issue seems to be still in HEAD 

### Steps to reproduce:

1. Set a udp_payload value in the [[outputs.influxdb]] plugin which is lower the size of some measurement lines.
2. Check the UDP datagram sizes with tcpdump.

### Expected behavior:
The datagrams should be split to be under the udp_payload size.

### Actual behavior:
The udp_payload parameter is disregarded and no split is performed. Measurements larger than udp_payload size are sent as single datagrams. For very large measurements a "message too long" error is produced.

oplehto:
### Required for all PRs:

Fixes #5438 

- Set MaxLineBytes correctly
- Split the data to utilize as much of the datagram as possible (even with MaxLineBytes set correctly each metric that did not fit into the first datagram was split up separately)
- Split the byte slice from serializer to separate UDP datagrams if there are multiple lines

### DIFF
--- a/plugins/outputs/influxdb/udp.go
+++ b/plugins/outputs/influxdb/udp.go
@@ -1,6 +1,8 @@
 package influxdb
 
 import (
+	"bufio"
+	"bytes"
 	"context"
 	"fmt"
 	"log"
@@ -45,9 +47,9 @@ func NewUDPClient(config *UDPConfig) (*udpClient, error) {
 	serializer := config.Serializer
 	if serializer == nil {
 		s := influx.NewSerializer()
-		s.SetMaxLineBytes(config.MaxPayloadSize)
 		serializer = s
 	}
+	serializer.SetMaxLineBytes(size)
 
 	dialer := config.Dialer
 	if dialer == nil {
@@ -96,7 +98,11 @@ func (c *udpClient) Write(ctx context.Context, metrics []telegraf.Metric) error 
 			continue
 		}
 
-		_, err = c.conn.Write(octets)
+		scanner := bufio.NewScanner(bytes.NewReader(octets))
+		scanner.Split(scanLines)
+		for scanner.Scan() {
+			_, err = c.conn.Write(scanner.Bytes())
+		}
 		if err != nil {
 			c.conn.Close()
 			c.conn = nil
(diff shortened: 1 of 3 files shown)


plugins/outputs/influxdb/udp.go
```diff
@@ -96,7 +98,11 @@ func (c *udpClient) Write(ctx context.Context, metrics []telegraf.Metric) error
 			continue
 		}
 
-		_, err = c.conn.Write(octets)
+		scanner := bufio.NewScanner(bytes.NewReader(octets))
+		scanner.Split(scanLines)
+		for scanner.Scan() {
+			_, err = c.conn.Write(scanner.Bytes())
+		}
```
danielnelson: For 1.10, I think I will experiment with doing a metric splitter before the Serialize step, so we don't need to parse the serialized data.  There might be some duplication of the serialize code but I think it will be worth it, and also string fields can contain unescaped newlines in line protocol so there is a potential flaw with this (#4942).

I think we are going to do a 1.9.5, and this will be a more appropriate change for it.

plugins/serializers/influx/influx_test.go
```diff
@@ -275,6 +275,22 @@ var tests = []struct {
 		),
 		output: []byte("cpu abc=123i 1519194109000000042\ncpu def=456i 1519194109000000042\n"),
 	},
+	{
+		name:     "split_fields_overflow",
```
danielnelson: This test passes without the change to the serializer, is it possible to modify it to require the change?

plugins/outputs/influxdb/udp.go
oplehto: I completely agree doing the change you proposed is a sensible approach. This was just a stopgap to get things working with minimal modifications to the existing code.

plugins/serializers/influx/influx_test.go
oplehto: The test now fails if the change is not implemented

PR #5458, opened 2019-02-20
Retry restarting receiver on PubSub service error

emilymye:
### Required for all PRs:

### DIFF
--- a/plugins/inputs/cloud_pubsub/pubsub.go
+++ b/plugins/inputs/cloud_pubsub/pubsub.go
@@ -131,7 +131,7 @@ func (ps *PubSub) Stop() {
 // subscription.Receive() up and running while the plugin has not been stopped.
 func (ps *PubSub) receiveWithRetry() {
 	err := ps.startReceiver()
-	
+
 	for err != nil && ps.ctx.Err() == nil {
 		log.Printf("E! Receiver for subscription %s exited with error: %v", ps.sub.ID(), err)
 
@@ -303,6 +303,12 @@ const sampleConfig = `
   ## PubSub APIs. If not set explicitly, Telegraf will attempt to use
   ## Application Default Credentials, which is preferred.
   # credentials_file = "path/to/my/creds.json"
+
+  ## Optional. Number of seconds to wait before attempting to restart the 
+  ## PubSub subscription receiver after an unexpected error. 
+  ## If the streaming pull for a PubSub Subscription fails (receiver),
+  ## the agent attempts to restart receiving messages after this many seconds.
+  # retry_delay_seconds = 5
 
   ## Optional. Maximum byte length of a message to consume.
   ## Larger messages are dropped with an error. If less than 0 or unspecified,
--- a/plugins/inputs/cloud_pubsub/pubsub.go
+++ b/plugins/inputs/cloud_pubsub/pubsub.go
@@ -43,7 +43,6 @@ type PubSub struct {
 	sub     subscription
 	stubSub func() subscription
 
-	ctx    context.Context
 	cancel context.CancelFunc
 
 	parser parsers.Parser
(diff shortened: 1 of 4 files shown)


plugins/inputs/cloud_pubsub/pubsub.go
```diff
@@ -78,35 +84,37 @@ func (ps *PubSub) Start(ac telegraf.Accumulator) error {
 		return fmt.Errorf(`"project" is required`)
 	}
 
-	cctx, cancel := context.WithCancel(context.Background())
+	ps.sem = make(semaphore, ps.MaxUndeliveredMessages)
+	ps.acc = ac.WithTracking(ps.MaxUndeliveredMessages)
+
+	// Create top-level context with cancel that will be called on Stop().
+	ctx, cancel := context.WithCancel(context.Background())
+	ps.ctx = ctx
```
danielnelson: Don't assign the context to the struct, instead pass it down to the goroutines below and save only the cancel function.

plugins/inputs/cloud_pubsub/pubsub.go
```diff
@@ -119,13 +127,46 @@ func (ps *PubSub) Stop() {
...
+
+		log.Printf("I! Waiting %d seconds before attempting to restart receiver...", delay)
+		time.Sleep(time.Duration(delay) * time.Second)
+
+		err = ps.startReceiver()
+	}
+}
+
+func (ps *PubSub) startReceiver() error {
+	cctx, ccancel := context.WithCancel(ps.ctx)
```
danielnelson: Move the creation of the cancel function into receiveOnSub, and just pass the context down.

plugins/inputs/cloud_pubsub/pubsub.go
```diff
@@ -119,13 +127,46 @@ func (ps *PubSub) Stop() {
...
 			ps.acc.AddError(fmt.Errorf("unable to add message from subscription %s: %v", ps.sub.ID(), err))
 		}
 	})
-	ps.acc.AddError(fmt.Errorf("receiver for subscription %s exited: %v", ps.sub.ID(), err))
+	if err != nil {
+		ps.acc.AddError(fmt.Errorf("receiver for subscription %s exited: %v", ps.sub.ID(), err))
+	} else {
+		log.Printf("I! subscription pull ended (no error, most likely stopped)")
+	}
+	ccancel()
```
danielnelson: I'm confused here, can you explain what this cancel function is doing?  It seems like the operation is over at this point?

plugins/inputs/cloud_pubsub/pubsub.go
```diff
@@ -119,13 +127,46 @@ func (ps *PubSub) Stop() {
...
 }
 
-func (ps *PubSub) subReceive(cctx context.Context) {
+// startReceiver is called within a goroutine and manages keeping a
+// subscription.Receive() up and running while the plugin has not been stopped.
+func (ps *PubSub) receiveWithRetry() {
+	err := ps.startReceiver()
+
+	for err != nil && ps.ctx.Err() == nil {
+		log.Printf("E! Receiver for subscription %s exited with error: %v", ps.sub.ID(), err)
```
danielnelson: On all the logging calls, start them with: `E! [inputs.cloud_pubsub]` (or `I!`, `D!`).  We are looking to simplify this in the future but for now this helps keep straight where the log message is coming from.

plugins/inputs/cloud_pubsub/pubsub.go
emilymye: I'm just making sure the child context I create per subscription receive is canceled. I believe this needs to be done to guarantee that some message handling is finished at the end of the receive (i.e. if receive fails), and it doesn't hurt even if the Google lib already does it. This was previously being done on the top-level cancel during Stop().

PR #5461, opened 2019-02-21
update wavefront-sdk-go version

puckpuck:
### Required for all PRs:

Update dep version dependency reference for Wavefront to use major range instead of specific version.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -211,6 +211,14 @@
   version = "v2.1.13"
 
 [[projects]]
+  digest = "1:e5691038f8e87e7da05280095d968e50c17d624e25cca095d4e4cd947a805563"
+  name = "github.com/caio/go-tdigest"
+  packages = ["."]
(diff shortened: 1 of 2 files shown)


danielnelson:
Do you need to update to 0.9.2 now as well?  If so can you run `dep ensure` and commit the Gopkg.lock file.

puckpuck:
updated gopkg.lock as well.  I only captured updates to the wavefront portion as to not disrupt anything else.

I'm curious as to what is the process as to when (if ever) a `dep ensure -update` is performed to update all dependencies to latest versions?

danielnelson:
> I'm curious as to what is the process as to when (if ever) a dep ensure -update is performed to update all dependencies to latest versions?

I am not updating versions across the board, it just creates too much bug churn, only when a change requires a newer version to operate or if there is a known bug fix in the newer version.  One exception to this is gopsutil, which I usually update for every release.

PR #5464, opened 2019-02-21
For Prometheus Input add ability to query Consul Service catalog

mmolnar:
### Required for all PRs:

This PR adds the ability to query Consul Catalog for published prometheus metric endpoints.

Tests are not my strong suit, I will add them if this PR makes sense and there are no other issues with it.

### DIFF
--- a/plugins/inputs/prometheus/consul.go
+++ b/plugins/inputs/prometheus/consul.go
@@ -1,0 +1,208 @@
+package prometheus
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"net/url"
+	"strings"
+	"text/template"
+	"time"
+
+	"github.com/hashicorp/consul/api"
+	"github.com/influxdata/telegraf/config"
+)
+
+type ConsulConfig struct {
+	// Address of the Consul agent. The address must contain a hostname or an IP address
+	// and optionally a port (format: "host:port").
+	Enabled       bool            `toml:"enabled"`
+	Agent         string          `toml:"agent"`
+	QueryInterval config.Duration `toml:"query_interval"`
+	Queries       []*ConsulQuery  `toml:"query"`
+}
+
+// One Consul service discovery query
+type ConsulQuery struct {
+	// A name of the searched services (not ID)
+	ServiceName string `toml:"name"`
+
+	// A tag of the searched services
+	ServiceTag string `toml:"tag"`
+
+	// A DC of the searched services
+	ServiceDc string `toml:"dc"`
+
+	// A template URL of the Prometheus gathering interface. The hostname part
+	// of the URL will be replaced by discovered address and port.
+	ServiceURL string `toml:"url"`
+
+	// Extra tags to add to metrics found in Consul
+	ServiceExtraTags map[string]string `toml:"tags"`
+
+	serviceURLTemplate       *template.Template
+	serviceExtraTagsTemplate map[string]*template.Template
+
+	// Store last error status and change log level depending on repeated occurence
+	lastQueryFailed bool
+}
+
+func (p *Prometheus) startConsul(ctx context.Context) error {
+	consulAPIConfig := api.DefaultConfig()
+	if p.ConsulConfig.Agent != "" {
+		consulAPIConfig.Address = p.ConsulConfig.Agent
+	}
+
+	consul, err := api.NewClient(consulAPIConfig)
+	if err != nil {
+		return fmt.Errorf("cannot connect to the Consul agent: %v", err)
+	}
+
+	// Parse the template for metrics URL, drop queries with template parse errors
+	i := 0
+	for _, q := range p.ConsulConfig.Queries {
+		serviceURLTemplate, err := template.New("URL").Parse(q.ServiceURL)
+		if err != nil {
+			p.Log.Errorf("Could not parse the Consul query URL template (%s), skipping it. Error: %s", q.ServiceURL, err)
+			continue
+		}
+		q.serviceURLTemplate = serviceURLTemplate
+
+		// Allow to use join function in tags
+		templateFunctions := template.FuncMap{"join": strings.Join}
+		// Parse the tag value templates
+		q.serviceExtraTagsTemplate = make(map[string]*template.Template)
+		for tagName, tagTemplateString := range q.ServiceExtraTags {
+			tagTemplate, err := template.New(tagName).Funcs(templateFunctions).Parse(tagTemplateString)
+			if err != nil {
+				p.Log.Errorf("Could not parse the Consul query Extra Tag template (%s), skipping it. Error: %s", tagTemplateString, err)
+				continue
+			}
+			q.serviceExtraTagsTemplate[tagName] = tagTemplate
+		}
+		p.ConsulConfig.Queries[i] = q
+		i++
+	}
+	// Prevent memory leak by erasing truncated values
+	for j := i; j < len(p.ConsulConfig.Queries); j++ {
+		p.ConsulConfig.Queries[j] = nil
+	}
+	p.ConsulConfig.Queries = p.ConsulConfig.Queries[:i]
+
+	catalog := consul.Catalog()
+
+	p.wg.Add(1)
+	go func() {
+		// Store last error status and change log level depending on repeated occurence
+		var refreshFailed = false
+		defer p.wg.Done()
+		err := p.refreshConsulServices(catalog)
+		if err != nil {
+			refreshFailed = true
+			p.Log.Errorf("Unable to refreh Consul services: %v", err)
+		}
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-time.After(time.Duration(p.ConsulConfig.QueryInterval)):
+				err := p.refreshConsulServices(catalog)
+				if err != nil {
+					message := fmt.Sprintf("Unable to refreh Consul services: %v", err)
+					if refreshFailed {
+						p.Log.Debug(message)
+					} else {
+						p.Log.Warn(message)
+					}
+					refreshFailed = true
+				} else if refreshFailed {
+					refreshFailed = false
+					p.Log.Info("Successfully refreshed Consul services after previous errors")
+				}
+			}
+		}
+	}()
+
+	return nil
+}
+
+func (p *Prometheus) refreshConsulServices(c *api.Catalog) error {
+	consulServiceURLs := make(map[string]URLAndAddress)
+
+	p.Log.Debugf("Refreshing Consul services")
+
+	for _, q := range p.ConsulConfig.Queries {
+		queryOptions := api.QueryOptions{}
+		if q.ServiceDc != "" {
+			queryOptions.Datacenter = q.ServiceDc
+		}
+
+		// Request services from Consul
+		consulServices, _, err := c.Service(q.ServiceName, q.ServiceTag, &queryOptions)
+		if err != nil {
+			return err
+		}
+		if len(consulServices) == 0 {
+			p.Log.Debugf("Queried Consul for Service (%s, %s) but did not find any instances", q.ServiceName, q.ServiceTag)
+			continue
+		}
+		p.Log.Debugf("Queried Consul for Service (%s, %s) and found %d instances", q.ServiceName, q.ServiceTag, len(consulServices))
+
+		for _, consulService := range consulServices {
+			uaa, err := p.getConsulServiceURL(q, consulService)
+			if err != nil {
+				message := fmt.Sprintf("Unable to get scrape URLs from Consul for Service (%s, %s): %s", q.ServiceName, q.ServiceTag, err)
+				if q.lastQueryFailed {
+					p.Log.Debug(message)
+				} else {
+					p.Log.Warn(message)
+				}
+				q.lastQueryFailed = true
+				break
+			}
+			if q.lastQueryFailed {
+				p.Log.Infof("Created scrape URLs from Consul for Service (%s, %s)", q.ServiceName, q.ServiceTag)
+			}
+			q.lastQueryFailed = false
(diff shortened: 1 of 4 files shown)


danielnelson:
Thanks the PR, this looks very interesting and useful.  I'm really busy finalizing 1.10 up now so I'll have to take a closer look in a few weeks.

plugins/inputs/prometheus/prometheus.go
```diff
@@ -30,6 +31,9 @@ type Prometheus struct {
 	// Location of kubernetes config file
 	KubeConfig string
 
+	// Consul SD configuration
+	ConsulConfig ConsulConfig `toml:"consul,ommitempty"`
```
GeorgeMac: Typo: `omitempty`

plugins/inputs/prometheus/consul.go
```diff
@@ -0,0 +1,174 @@
...
+		Address:        s.Address,
+		Datacenter:     s.Datacenter,
+		NodeMeta:       s.NodeMeta,
+		ServiceID:      s.ServiceID,
+		ServiceAddress: s.ServiceAddress,
+		ServiceTags:    s.ServiceTags,
+		ServiceMeta:    s.ServiceMeta,
+		ServicePort:    s.ServicePort,
+	}
+}
```
srebhan, line 42: Is there a reason not to use `api.CatalogService` directly? Your declaration is a one-to-one translation...

plugins/inputs/prometheus/consul.go
```diff
@@ -0,0 +1,174 @@
...
+func (p *Prometheus) startConsul(ctx context.Context) error {
+	config := api.DefaultConfig()
+	if p.ConsulConfig.Agent != "" {
+		config.Address = p.ConsulConfig.Agent
+	}
+
+	consul, err := api.NewClient(config)
+	if err != nil {
+		log.Printf("E! [inputs.prometheus] Cannot connect to the Consul agent. Error: %s", err)
+		return nil
```
srebhan, line 53: Please avoid logging with `log` and rather use the telegraf logging-framework instead. However, here I would rather like to see
```suggestion
		return fmt.Errorf("cannot connect to the Consul agent: %v", err)
```

plugins/inputs/prometheus/consul.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	// Parse the template for metrics URL
+	for i, q := range p.ConsulConfig.Queries {
+		serviceURLTemplate, err := template.New("URL").Parse(q.ServiceURL)
+		if err != nil {
+			log.Printf("E! [inputs.prometheus] Could not parse the Consul query URL template (%s), skipping it. Error: %s", q.ServiceURL, err)
```
srebhan, line 60: Please use
```suggestion
			p.Log.Errorf("Could not parse the Consul query URL template (%s), skipping it. Error: %s", q.ServiceURL, err)
```
instead!

plugins/inputs/prometheus/consul.go
```diff
@@ -0,0 +1,174 @@
...
+		p.ConsulConfig.Queries[i].serviceURLTemplate = serviceURLTemplate
+
+		// Allow to use join function in tags
+		templateFunctions := template.FuncMap{"join": strings.Join}
+		// Parse the tag value templates
+		p.ConsulConfig.Queries[i].serviceExtraTagsTemplate = make(map[string]*template.Template)
+		for tagName, tagTemplateString := range q.ServiceExtraTags {
+			tagTemplate, err := template.New(tagName).Funcs(templateFunctions).Parse(tagTemplateString)
+			if err != nil {
+				log.Printf("E! [inputs.prometheus] Could not parse the Consul query Extra Tag template (%s), skipping it. Error: %s", tagTemplateString, err)
```
srebhan, line 72: See above. Please use `p.Log` also for all other instances of `log`.

plugins/inputs/prometheus/consul.go
```diff
@@ -0,0 +1,174 @@
...
+
+	catalog := consul.Catalog()
+
+	p.wg.Add(1)
+	go func() {
+		defer p.wg.Done()
+		err := p.refreshConsulServices(catalog)
+		if err != nil {
+			log.Printf("E! [inputs.prometheus] Unable to refresh Consul services: %v", err)
+		}
```
srebhan, line 88: Does it really make sense to unconditionally continue here?

plugins/inputs/prometheus/consul.go
```diff
@@ -0,0 +1,174 @@
...
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-time.After(p.ConsulConfig.QueryInterval.Duration):
+				err := p.refreshConsulServices(catalog)
+				if err != nil {
+					log.Printf("E! [inputs.prometheus] Unable to refresh Consul services: %v", err)
```
srebhan, line 96: You might spam the log file with those errors if e.g. the service cannot be reached...

plugins/inputs/prometheus/consul.go
```diff
@@ -0,0 +1,174 @@
...
+	var consulServiceInfo *ConsulServiceInfo
+	consulServiceURLs := make(map[string]URLAndAddress)
+
+	log.Printf("D! [inputs.prometheus] Refreshing Consul services")
+
+	for _, q := range p.ConsulConfig.Queries {
+		// Template parse failed, ignore this Service Query
+		if q.serviceURLTemplate == nil {
+			continue
+		}
```
srebhan, line 116: Can we please check this in `startConsul()` and remove the template right away.

plugins/inputs/prometheus/consul.go
```diff
@@ -0,0 +1,174 @@
...
+
+		log.Printf("D! [inputs.prometheus] Queried Consul for Service (%s, %s)", q.ServiceName, q.ServiceTag)
+
+		for _, consulService := range consulServices {
+			consulServiceInfo = convertConsulService(consulService)
+			buffer.Reset()
+			err = q.serviceURLTemplate.Execute(&buffer, consulServiceInfo)
+			if err != nil {
+				log.Printf("E! [inputs.prometheus] Could not execute Consul Service URL template, skipping it. Error: %s", err)
+
```
srebhan, line 137: Remove newline.

plugins/inputs/prometheus/prometheus.go
srebhan: Please omit the `omitempty` altogether!

plugins/inputs/prometheus/prometheus.go
```diff
@@ -46,6 +50,37 @@ type Prometheus struct {
...
+
+	// List of consul services to scrape
+	consulServices map[string]URLAndAddress
+}
+
+type ConsulConfig struct {
+	// Address of the Consul agent. The address must contain a hostname or an IP address
+	// and optionally a port (format: "host:port").
+	Enabled       bool              `toml:"enabled"`
+	Agent         string            `toml:"agent,omitempty"`
```
srebhan, line 62: No `omitempty` please as it will result in an empty string anyway.

plugins/inputs/prometheus/prometheus.go
```diff
@@ -46,6 +50,37 @@ type Prometheus struct {
...
+	// A template URL of the Prometheus gathering interface. The hostname part
+	// of the URL will be replaced by discovered address and port.
+	ServiceURL string `toml:"url"`
+
+	// Extra tags to add to metrics found in Consul
+	ServiceExtraTags map[string]string `toml:"tags"`
+
+	serviceURLTemplate       *template.Template
+	serviceExtraTagsTemplate map[string]*template.Template
 }
```
srebhan, line 84: It might make sense to move these declarations into `consul.go` as they are mostly used there.

plugins/inputs/prometheus/prometheus.go
```diff
@@ -306,18 +358,29 @@ func (p *Prometheus) gatherURL(u URLAndAddress, acc telegraf.Accumulator) error
...
-		var ctx context.Context
-		ctx, p.cancel = context.WithCancel(context.Background())
-		return p.start(ctx)
+		err = p.startK8s(ctx)
+		if err != nil {
+			return err
+		}
 	}
-	return nil
+	return err
```
srebhan, line 379: I think `return nil` is correct here.

plugins/inputs/prometheus/prometheus.go
```diff
@@ -306,18 +358,29 @@ func (p *Prometheus) gatherURL(u URLAndAddress, acc telegraf.Accumulator) error
...
+		}
+	}
 	if p.MonitorPods {
-		var ctx context.Context
-		ctx, p.cancel = context.WithCancel(context.Background())
-		return p.start(ctx)
+		err = p.startK8s(ctx)
+		if err != nil {
+			return err
+		}
```
srebhan, line 377: Please use
```suggestion
		if err := p.startK8s(ctx); err != nil {
			return err
		}
```

srebhan (commented):
The code looks quite good already. I have some comments in the code, especially about using or rather not using `log`. Please also rebase the code to the current master!

srebhan:
@mmolnar any chance you can work on this?

mmolnar:
Thank you for the review. I will try to find time to work on it.

srebhan:
@mmolnar did you have time to take a look?

plugins/inputs/prometheus/consul.go
mmolnar: Probably not, originally I wanted to limit the number of fields that can be used in the template, but that might not be necessary.

plugins/inputs/prometheus/consul.go
mmolnar: Consul might be available later and services will start to be queried without the need to restart telegraf.

plugins/inputs/prometheus/consul.go
mmolnar: I will be dropping queries with parse errors altogether in startConsul

plugins/inputs/prometheus/consul.go
srebhan: Hmm how about making it a `warn` the first time and then switch to `debug`? It's a difficult trade-off here: You want the user to know that something is broken, but at the same time you might spam the log-file (depending on the length of `QueryInterval`) causing all sorts of issues... TBH I'm not completely sure what the best approach is here. We only report some errors (e.g. file does not exist) for some other plugins and show a success if any. This way the user sees that something is wrong and how long the "error state" persisted. Do you think we can do something similar here?

plugins/inputs/prometheus/consul.go
```diff
@@ -0,0 +1,176 @@
...
+					p.Log.Infof("Unable to refresh Consul services: %v", err)
+				}
+			}
+		}
+	}()
+
+	return nil
+}
+
+func (p *Prometheus) refreshConsulServices(c *api.Catalog) error {
```
srebhan, line 111: With the current implementation this function can only return `nil` for error. You might either refine your success condition (e.g. at least one query needs to succeed) and return an error otherwise **or** remove the return value altogether.

srebhan (commented):
Almost perfect. :-) Thanks @mmolnar for keep this going! I added a comment regarding the potential "log-spam" issue and additionally found that currently `refreshConsulServices()` always returns `nil` as error. Can you please take a look there if we can come up with a refined "success" definition (see my comment in the code)?

plugins/inputs/prometheus/consul.go
```diff
@@ -0,0 +1,209 @@
...
+					p.Log.Debug(message)
+				} else {
+					p.Log.Warn(message)
+				}
+				q.lastQueryFailed = true
+				break
+			} else if q.lastQueryFailed {
+				q.lastQueryFailed = false
+				p.Log.Infof("Created scrape URLs from Consul for Service (%s, %s)", q.ServiceName, q.ServiceTag)
+			}
```
srebhan, line 167: I think you can drop the `else` and only use `if` as you `break` in the if clause above. You maybe also want to move the `q.lastQueryFailed = false` part out of the if, just to be sure we reset the thing. :-)

plugins/inputs/prometheus/consul.go
mmolnar: I left the error return value there but at the moment the only case when it is used is when the consul agent is not running.

All other errors about are skipped with error logged to as debug, warning or error.

srebhan (commented):
Wow that was quick!
One cosmetic thing: can you please move the reset of `lastQueryFailed` out of the if clause. I'm always getting nervous when dealing with states, so it's better to involve the least number of branches possible...  :-)

plugins/inputs/prometheus/consul.go
mmolnar: This should be resolved now, the code got a bit more cluttered, but I could not think of a way to make it look nicer.

srebhan:
This leave the linter issues (see https://github.com/influxdata/telegraf/pull/5464/checks?check_run_id=2820187786)
You can check your branch locally running `make lint-branch`...

mmolnar:
> This leaves the linter issues (see https://github.com/influxdata/telegraf/pull/5464/checks?check_run_id=2820187786)
> You can check your branch locally running `make lint-branch`...

Sorry for that, should be working correctly now.

srebhan:
!retry-failed

srebhan:
!retry-failed

srebhan (approved):
@mmolnar no worries. The code looks good to me. Thanks for keeping this going!

akrantz01 (changes requested):
The code looks good to me. However, could you add the example config and some documentation to the README?

mmolnar:
I must have messed something up while rebasing and pushing the requested changes to my fork...
It was not my intention to close this PR.

The changes are still available in the master branch of my fork.

akrantz01 (approved):
LGTM!

helenosheaa (approved):
Looks good, thanks for working on this!

PR #5471, opened 2019-02-22
Add loggregator_rlp input plugin

robertjsullivan:
### Required for all PRs:


trevorwhitney:
Hey @glinton, just wanted to bump this thread. We have been actively developing and using this plugin and already have some updates to it. Are we still on track to ship in 1.11.0? Is there anything else you need from us? Should we hold off on the updates until after we ship this first version, or should we add them to this PR before things go through?

Thanks!

glinton:
I'd say go ahead and add them into this PR. Once you do, I'll do an initial review and I believe this could make it into 1.11

trevorwhitney:
@glinton updated the PR, thanks!

plugins/inputs/loggregator_rlp/envelope_writer.go
```diff
@@ -0,0 +1,247 @@
...
+	w.incrementMetric(w.ingressedEnvelopes, 1)
+}
+
+func (w *EnvelopeWriter) handleEnvelopes() {
+	for {
+		select {
+		case <-w.stop:
+			return
+		default:
+			e, ok := w.envelopesQueue.TryNext()
```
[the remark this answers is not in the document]
trevorwhitney: @glinton not sure I understand the question? The queue in question is a [diode](https://github.com/cloudfoundry/go-diodes) and will drop envelopes when under load so as not to slow down ingestion. Is your question around whether or not we need a queue at all, or in how we're using it in this case?

plugins/inputs/loggregator_rlp/rlp.go
```diff
@@ -0,0 +1,179 @@
...
+	"code.cloudfoundry.org/go-loggregator"
+	"code.cloudfoundry.org/go-loggregator/rpc/loggregator_v2"
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/plugins/inputs"
+)
+
+type LoggregatorRLPInput struct {
+	TlsCommonName           string   `toml:"tls_common_name"`
+	RlpAddress              string   `toml:"rlp_address"`
+	InternalMetricsInterval string   `toml:"internal_metrics_interval"`
```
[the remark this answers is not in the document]
trevorwhitney: great suggestion, fixed!

plugins/inputs/loggregator_rlp/envelope_writer.go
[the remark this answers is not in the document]
glinton: I think I was just concerned about `TryNext` returning nothing for many iterations, but really I'm not certain there's a better way.

trevorwhitney:
@russorat why did the milestone get changed? Can we no longer expect this to land in 1.11.0?

danielnelson:
Unfortunately, we won't have time to look into this for 1.11.

plugins/inputs/loggregator_rlp/rlp.go
```diff
@@ -0,0 +1,178 @@
...
+			},
+		}
+	default:
+		panic("Invalid envelope type " + envelopeType)
+	}
+}
+
+func (l *LoggregatorRLPInput) Stop() {
+	log.Printf("Info: Stopping RLP listener")
+	l.stopRlpConsumer()
```
danielnelson: Once this functions returns, it is illegal to add any more metrics to the accumulator, this means we need to wait for all goroutines to exit.  Probably can be solved with a `sync.WaitGroup`.

plugins/inputs/loggregator_rlp/rlp.go
```diff
@@ -0,0 +1,178 @@
...
+
+	"github.com/influxdata/telegraf/internal/tls"
+
+	"code.cloudfoundry.org/go-loggregator"
+	"code.cloudfoundry.org/go-loggregator/rpc/loggregator_v2"
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/plugins/inputs"
+)
+
+type LoggregatorRLPInput struct {
```
danielnelson: Call this `LoggregatorRLP`

plugins/inputs/loggregator_rlp/rlp.go
```diff
@@ -0,0 +1,178 @@
...
+  source_id_filters = ['router']
+`
+}
+
+func (l *LoggregatorRLPInput) Gather(_ telegraf.Accumulator) error {
+	return nil
+}
+
+func (l *LoggregatorRLPInput) Start(acc telegraf.Accumulator) error {
+	envelopeWriter := NewEnvelopeWriter(acc, l.InternalMetricsInterval.Duration)
```
danielnelson: I don't see the purpose of this, just call a function in the consumer goroutine on line 103 that will add the data to the accumulator.  This will dramatically simplify this plugin and eliminate the need for the internal metrics and ring buffer.  The ring buffer duplicates functionality that Telegraf gives you for free and introduces another buffer and with it the potential for dropped metrics.

plugins/inputs/loggregator_rlp/README.md
```diff
@@ -0,0 +1,39 @@
...
+  rlp_address = "bar"
+
+  ## A string duration for how frequently to report internal metrics
+  internal_metrics_interval = "30s"
+```
+
+### Known Issues
+
+Counters with only delta values are currently dropped.
+The plugin supports Counter envelopes with totals, Gauge envelopes and `http` Timer envelopes.
```
danielnelson: It would be helpful to see some example data so we can evaluate the schema.

plugins/inputs/loggregator_rlp/envelope_writer.go
```diff
@@ -0,0 +1,249 @@
...
+		return
+	}
+
+	if counter.GetTotal() == 0 && counter.GetDelta() > 0 {
+		w.incrementMetric(w.droppedDeltaCounters, 1)
+		return
+	}
+
+	fields := map[string]interface{}{
+		"counter": counter.GetTotal(),
```
danielnelson: The name of this field indicates that we may not be using the measurement and field names using the current Telegraf style.  Most likely we would want to set the field name to `counter.GetName()` and the measurement would be the same across all metrics, perhaps `loggregator_rlp`.  Add some example data and we can take a closer look.

sjwang90:
Closing this in favor of #7773. Feel free to open back up if necessary.

PR #5478, opened 2019-02-25
feat(http): add token auth

kelwang:
Add a token auth option to http out plugin for idpe billing metrics
### Required for all PRs:


danielnelson:
I think we just want to use the headers option for this, since it isn't really a standard auth method.

PR #5483, opened 2019-02-25
Log permission error rather than add to filecount input

### LINKED ISSUE #5449: Filecount: directory permission denied blocks input regardless of recurse parameter
### System info:
Telegraf 1.9.4 (git: HEAD 4da8d0a4)
Ubuntu 16.04

### Steps to reproduce:
1. create this architecture in /tmp:
> foo
> |-- [-rw-rw-r--]  aaa
> `-- [d---------]  bar

   `foo` and `bar` are directories
   `aaa` is a regular file
   `bar` cannot be read by telegraf user (here, it is simulated by a 000 chmod)
2. Create the file `/tmp/aaa.conf` with this content:
> [[inputs.filecount]]
>   directory = "/tmp/foo/"
>   name = "aaa"
>   recursive = false
3. execute config file: `telegraf --config /tmp/aaa.conf --test`

### Expected behavior:
Telegraf process should return: 
    > filecount,directory=/tmp/foo/,host=computer count=1i 1550582182000000000

### Actual behavior:
Telegraf process returns:
2019-02-19T13:17:01Z E! [inputs.filecount]: Error in plugin: open /tmp/foo/bar: permission denied

### Additional info:
Since recurse is set to false, telegraf should not care about subdirectories, especially, it should not go into error. 
I am still looking for a workaround.

glinton:
Resolves #5449

### DIFF
--- a/plugins/inputs/filecount/filecount.go
+++ b/plugins/inputs/filecount/filecount.go
@@ -19,7 +19,7 @@ import (
 const sampleConfig = `
   ## Directory to gather stats about.
   ##   deprecated in 1.9; use the directories option
-  directory = "/var/cache/apt/archives"
+  # directory = "/var/cache/apt/archives"
 
   ## Directories to gather stats about.
   ## This accept standard unit glob matching rules, but with the addition of
--- a/plugins/inputs/filecount/filecount.go
+++ b/plugins/inputs/filecount/filecount.go
@@ -1,16 +1,19 @@
 package filecount
 
 import (
+	"log"
 	"os"
 	"path/filepath"
 	"strings"
 	"time"
 
+	"github.com/karrick/godirwalk"
+	"github.com/pkg/errors"
+
 	"github.com/influxdata/telegraf"
 	"github.com/influxdata/telegraf/internal"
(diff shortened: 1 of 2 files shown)


plugins/inputs/filecount/filecount.go
```diff
@@ -203,6 +208,13 @@ func (fc *FileCount) count(acc telegraf.Accumulator, basedir string, glob globpa
 		Callback:             walkFn,
 		PostChildrenCallback: postChildrenFn,
 		Unsorted:             true,
+		ErrorCallback: func(osPathname string, err error) godirwalk.ErrorAction {
+			if os.IsPermission(errors.Cause(err)) {
+				log.Println("I! [inputs.filecount]", err)
```
danielnelson: Use `D!`, since I think most will just want to ignore this error.  Can you paste an example of how the error looks here, we might need some leading context.

glinton: Meant to do that,
```
2019-02-25T23:33:30Z D! [inputs.filecount] cannot ReadDirents: cannot Open: open /tmp/foo/bar: permission denied
```

PR #5488, opened 2019-02-26
Add option to disable timestamp adjustment in grok parser

### LINKED ISSUE #5394: Data Duplication in InfluxDB
Hi All,

Need help in scenario that we are facing:-
We are loading data into influxdb parsing via telegraf.
We have one record in source file. Please find below snapshot of the source record-

image

After data loading above source file into InfluxDB we are getting two records when we are trying to query it.
Please find below snapshot after querying the result set of influxdb
image

It seems that influxdb creates duplicate records itself with another timestamp i.e 
image

The above scenario is observed sometimes but not in each and every data loading source file.

Please help us.

Warm regrards
//Ashlesh

glinton:
Resolves #5394 by adding `disable_time_mod` option to grok parser.

### DIFF
--- a/plugins/parsers/grok/parser.go
+++ b/plugins/parsers/grok/parser.go
@@ -86,6 +86,9 @@ type Parser struct {
 	Timezone string
 	loc      *time.Location
 
+	// DisableTimeMod will not increment the timestamp if there is a duplicate.
+	DisableTimeMod bool
+
 	// typeMap is a map of patterns -> capture name -> modifier,
 	//   ie, {
 	//          "%{TESTLOG}":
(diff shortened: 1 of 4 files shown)


danielnelson:
Let's call the option `grok_unique_timestamp` with possible values "auto" and "disable", to leave us open to further options, default is "auto".  This code won't work as it is, you need to hook up the options in registry.go and config.go.

plugins/parsers/grok/parser.go
```diff
@@ -358,6 +365,10 @@ func (p *Parser) ParseLine(line string) (telegraf.Metric, error) {
 		return nil, fmt.Errorf("grok: must have one or more fields")
 	}
 
+	if p.UniqueTimestamp != "auto" {
+		return metric.New(p.Measurement, tags, fields, timestamp)
```
danielnelson: This is backwards, isn't it?

danielnelson: Nevermind, not equals

PR #5490, opened 2019-02-27
Influxdb output database routing tag

danielnelson:
### Required for all PRs:

### DIFF
--- a/plugins/outputs/influxdb/README.md
+++ b/plugins/outputs/influxdb/README.md
@@ -19,6 +19,10 @@ The InfluxDB output plugin writes metrics to the [InfluxDB v1.x] HTTP or UDP ser
   ## For UDP url endpoint database needs to be configured on server side.
   # database = "telegraf"
 
+  ## The value of this tag will be used to determine the database.  If this
+  ## tag is not set the 'database' option is used as the default.
+  # database_tag = ""
+
   ## If true, no CREATE DATABASE queries will be sent.  Set to true when using
   ## Telegraf with a user without permissions to create databases or when the
   ## database already exists.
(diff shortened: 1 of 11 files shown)


voiprodrigo:
Is it possible to drop the tag after routing is established?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

danielnelson:
I didn't include this.

amarbajric:
@danielnelson Is there any possibility at the moment to drop the tag that was used for dynamic routing with `database_tag`? I have tried to use `tagexclude` without success. I guess `database_tag` and `tagexclude` are not intended to work together by now?

danielnelson:
Unfortunately, the general metric filtering options are applied before `database_tag`, so they can't be used together.  This means we will need to add a new option specifically for this :(.

amarbajric:
Ah I see, that's what I was thinking as telegraf was simply ignoring the `tagexclude` property in the configuration. Yes I will open an issue for this feature request as I think that others also would like this specific functionality! Thank you for your fast response :)

x70b1:
```
[[outputs.influxdb]]
  username = "telegraf"
  password = "xxxxx"
  timeout = "5s"

  database = "Database1"
  database_tag = "influxdb_database"
  tagexclude = ["influxdb_database"]

[[inputs.exec]]
  commands = [
    "bash command.sh"
  ]

  interval = "1m"
  timeout = "20s"
  data_format = "influx"

  [inputs.exec.tags]
    influxdb_database = "Database1"

```

Sounds weird, but `tagexclude` works there for me? How is that possible?
I tested it.

danielnelson:
Tag exclude will work to remove the tag, but then it won't be available for the `database_tag` option to use and it will just send everything to the default database `Database1`.

x70b1:
That makes sense. Thanks for the clarification!

PR #5497, opened 2019-02-28
Open log file in append mode.

sergeykhegay:
Problem:
When log file didn't exist the agent used to create a file without `O_APPEND` mode. This led to the file growing indefinitely, because the writes ignored the truncated file (by the current `logrotate` settings) and did not update the fd offset.

When you write to the offset bigger than the current file size, the file is expanded by the OS (at least Linux), the gap is filled with `\0`.

Note:
This behavior doesn't apply when the log file has already been created before Telegraf starts. The file opened in the proper append mode then.

Fix:
Always open in `O_APPEND` mode.

### Required for all PRs:

- [ x] Signed [CLA](https://influxdata.com/community/cla/).
- [ ] ~Associated README.md updated.~ Not needed

### DIFF
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -50,16 +50,10 @@ func SetupLogging(debug, quiet bool, logfile string) {
 
 	var oFile *os.File
 	if logfile != "" {
-		if _, err := os.Stat(logfile); os.IsNotExist(err) {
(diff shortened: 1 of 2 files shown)


danielnelson:
Thanks!

Issue #5503, opened 2019-02-28
When using splunkmetic serializer with vsphere input, source is not allowed to be overridden


dgshue:
## Bug/Feature Request

When using the splunkmetric serializer with vsphere input, vsphere input adds a tag of "source" which is later interpreted as a the splunk "source" in the payload.  This should be allowed to be overridden using tags.

Additionally, the splunkmetic serializer does not include "sourcetype" as an included field in the payload. Would like to add this as a tag.

Opening a feature request kicks off a discussion.

### Proposal:
Add SourceType as part of the HECTimeSeries struct and allow to be changed as a tag to the input or global_tags.

Suggest changing the splunkmetric to look for specific named tags for the serializer instead of something generic as "source" or "sourcetype" to "splunkmetric_source" and "splunkmetric_sourcetype".

### Current behavior:
Adding tags to vsphere input as follows:

  [inputs.vsphere.tags]
	
	host = "overide.host.com"
	index = "telegraf_metrics"
	sourcetype = "its:esx:metrics"
	source = "telegraf_agent"

Outputs JSON:

{  
   "time":1551372240,
   "event":"metric",
   "host":"overide.host.com",
   "index":"telegraf_metrics",
   "source":"VMSERVERNAME",
   "fields":{  
      "_value":0,
      "clustername":"CLUSTER",
      "dcname":"DCNAME",
      "disk":"DISKID",
      "esxhostname":"HOSTNAME",
      "guest":"GUESTOS",
      "metric_name":"METRICNAME",
      "moid":"VMID",
      "uuid":"UUID",
      "vcenter":"VCENTERNAME",
      "vmname":"VMNAME"
   }
}

### Desired behavior:

Adding tags to vsphere input as follows:

  [inputs.vsphere.tags]
	
	splunkmetric_host = "overide.host.com"
	splunkmetric_index = "telegraf_metrics"
	splunkmetric_sourcetype = "its:esx:metrics"
	splunkmetric_source = "telegraf_agent"

Outputs JSON:

{  
   "time":1551372240,
   "event":"metric",
   "host":"overide.host.com",
   "index":"telegraf_metrics",
   "source":"telegraf_agent",
   "sourcetype":"its:esx:metrics",
   "fields":{  
      "_value":0,
      "clustername":"CLUSTER",
      "dcname":"DCNAME",
      "disk":"DISKID",
      "esxhostname":"HOSTNAME",
      "guest":"GUESTOS",
[...]

PR #5504, opened 2019-02-28
Adding support for sourcetype in Splunkmetric serializer and standardizing tags for other metadata fields

dgshue:
### Required for all PRs:

[Bugfix/5503](https://github.com/influxdata/telegraf/issues/5503)

This will have impact on any previous config files as users will have to update their tags to reflect the new tag names.  Other than that it's pretty straight forward. Welcome any feedback, first time contributor so bare with me.

### DIFF
--- a/plugins/serializers/splunkmetric/README.md
+++ b/plugins/serializers/splunkmetric/README.md
@@ -73,11 +73,24 @@ to manage the HEC authorization, here's a sample config for an HTTP output:
 ## Overrides
 You can override the default values for the HEC token you are using by adding additional tags to the config file.
 
-The following aspects of the token can be overriden with tags:
+The following metadata will be included in the JSON payload and can be overridden with specific tags for the expected and/or optional fields:
+* host
 * index
 * source
+* sourcetype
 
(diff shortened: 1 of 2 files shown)


danielnelson:
I think we should use the `sourcetype` tag but not rename any of the existing tags, this should fit in with the existing implementation without breaking anything.  Alternately, I could see these as being configurable options on the parser:
```
data_format = "splunkmetric"
splunkmetric_sourcetype_tag = "splunkmetric_sourcetype"
```

ronnocol:
I'm with @danielnelson here, breaking the existing configuration syntax could be significantly painful for people who have large telegraf deployments. I think just adding sourcetype as an overridable field and generally adopting https://github.com/influxdata/telegraf/issues/3362 should be sufficient to resolve the issue without requiring config changes. 

There are Splunk based apps that are widely deployed that have the existing configuration schema included in them, I'm not saying it can't change, but it would first have to be deprecated before being removed.

srebhan:
@dgshue any update?

powersj:
Hi,

Thanks!

PR #5517, opened 2019-03-02
Gather all mysql channels

utking:
### Required for all PRs:


plugins/inputs/mysql/mysql.go
```diff
@@ -588,9 +592,11 @@ func (m *Mysql) gatherSlaveStatuses(db *sql.DB, serv string, acc telegraf.Accumu
 	tags := map[string]string{"server": servtag}
 	fields := make(map[string]interface{})
 
-	// to save the column names as a field key
-	// scanning keys and values separately
-	if rows.Next() {
+	// for each channel record
+	for rows.Next() != false {
```
glinton: The `!= false` is redundant, just use `for rows.Next() {`

plugins/inputs/mysql/mysql.go
```diff
@@ -575,7 +579,7 @@ func (m *Mysql) gatherGlobalVariables(db *sql.DB, serv string, acc telegraf.Accu
 // When the server is slave, then it returns only one row.
 // If the multi-source replication is set, then everything works differently
 // This code does not work with multi-source replication.
-func (m *Mysql) gatherSlaveStatuses(db *sql.DB, serv string, acc telegraf.Accumulator) error {
+func (m *Mysql) gatherSlaveStatuses(db *sql.DB, serv string, acc telegraf.Accumulator, gatherAllChannels bool) error {
```
glinton: No need to pass `gatherAllChannels` as it is part of the `Mysql` struct; just use `m.GatherAllSlaveChannels` where needed

utking: Removed in the most recent commit

plugins/inputs/mysql/mysql.go
utking: Removed the redundancy

danielmotaleite:
any change of adding support for mariadb and so fixing bug https://github.com/influxdata/telegraf/issues/6395

spumer:
any updates?

utking:
@glinton - I've updated the PR with MariaDB-related things. I really hope it won't be waiting for another half a year before something else is changed and we need to fix it again and again.
cc @danielmotaleite @spumer

plugins/inputs/mysql/README.md
```diff
@@ -196,7 +202,9 @@ measurement name.
 * Slave status - metrics from `SHOW SLAVE STATUS` the metrics are gathered when
 the single-source replication is on. If the multi-source replication is set,
 then everything works differently, this metric does not work with multi-source
-replication.
+replication, unless you set `gather_all_slave_channels = true`. For MariaDB,
+`mariadb_dialect = true` should be set to address the field names and commands
```
glinton, line 206: Does percona have it's own dialect as well? Maybe this would be better suited as an enum, something like `sql_dialect = "mariadb"` where the default/blank value is `"mysql"`, for extensibility.

plugins/inputs/mysql/mysql.go
```diff
@@ -641,11 +658,24 @@ func (m *Mysql) gatherSlaveStatuses(db *sql.DB, serv string, acc telegraf.Accumu
...
+					channelName = string(*vals[i].(*sql.RawBytes))
+				}
+				tags["channel"] = channelName
+			} else if value, ok := m.parseValue(*vals[i].(*sql.RawBytes)); ok {
 				fields["slave_"+col] = value
 			}
 		}
 		acc.AddFields("mysql", fields, tags)
+
+		if !m.GatherAllSlaveChannels {
```
glinton, line 676: It seems like `break`ing early would disturb current functionality. Since the loop logic is the same as before, we should expect someone is parsing subsequent rows.

glinton (commented):
Sorry for the long delay, i haven't been on the telegraf team for a while. @sjwang90 will likely know more about when this can fit in their timeline, so i won't approve or block this.

plugins/inputs/mysql/README.md
utking: No, Percona uses MySQL's one.

plugins/inputs/mysql/mysql.go
utking: Breaking early is what we have right now (with if instead of while) when we do not check all channels. So, this if does the same early break.

plugins/inputs/mysql/mysql.go
```diff
@@ -609,7 +618,13 @@ func (m *Mysql) parseGlobalVariables(key string, value sql.RawBytes) (interface{
...
 func (m *Mysql) gatherSlaveStatuses(db *sql.DB, serv string, acc telegraf.Accumulator) error {
 	// run query
-	rows, err := db.Query(slaveStatusQuery)
+	var rows *sql.Rows
+	var err error
+	if m.MariadbDialect {
+		rows, err = db.Query(slaveStatusQueryMariadb)
+	} else {
+		rows, err = db.Query(slaveStatusQuery)
+	}
```
srebhan, line 627: Can we please do this branching once e.g. in `InitMysql()` to make this more simple and readable!?

plugins/inputs/mysql/mysql.go
```diff
@@ -641,11 +658,24 @@ func (m *Mysql) gatherSlaveStatuses(db *sql.DB, serv string, acc telegraf.Accumu
 			if m.MetricVersion >= 2 {
 				col = strings.ToLower(col)
 			}
-			if value, ok := m.parseValue(*vals[i].(*sql.RawBytes)); ok {
+
+			if m.GatherAllSlaveChannels &&
+				(strings.ToLower(col) == "channel_name" || strings.ToLower(col) == "connection_name") {
+				// Since the default channel name is empty, we need this block
+				channelName := "default"
+				if len(*vals[i].(*sql.RawBytes)) > 0 {
```
srebhan, line 666: Is `vals[i]` _always_ of type `*sql.RawBytes`? Otherwise this type assertion might panic...

plugins/inputs/mysql/mysql.go
srebhan: Can you please add a small comment here saying something like
```golang
// Only the first row is relevant if not all slave-channels should be gathered, so break here and skip the remaining rows
```

srebhan (commented):
Had some small comments in the code. The most relevant one is regarding the type-assertion to `*sql.RawBytes`.

plugins/inputs/mysql/mysql.go
utking: First of all, as you can see, I did not add this line. It was there from the beginning. Second, the definition of the type is "type RawBytes []byte" - any SQL response can be converted to this type, right?

plugins/inputs/mysql/mysql.go
utking: Moved the check to InitMysql()

plugins/inputs/mysql/mysql.go
utking: Added the comment

utking:
@srebhan I've addressed your comments. Could you please check it now?

plugins/inputs/mysql/mysql.go
srebhan: Ok, seems reasonable for me.

srebhan (approved):
Looks good to me.

sspaink (approved):

Issue #5451, opened 2019-02-19
restart needed for system.ncpus to update


caapos:
This https://github.com/influxdata/telegraf/blob/master/plugins/inputs/system/system.go#L37 calls a GO system function https://golang.org/pkg/runtime/#NumCPU, which has undesired bahaviour. To force the system.ncpus metric to update in dashboard after change in OS, it is needed to restart telegraf process.

danielnelson:
We should probably start using gopsutil for this, but they are currently getting cpu count in the same way.

https://github.com/shirou/gopsutil/blob/9294781cc96294e42b8cb19de82a0ff3b19bb056/cpu/cpu.go#L70-L72

Can you describe what type of system you have that has changes to the cpu count, so we can make a good argument for the necessity of this code, and then I'll open an issue on gopsutil.

caapos:
We are running our monitored environments on CentOS virtual machines, ufortunately I do not have enough information about the underlying system configuration, but the parameters like te number of CPUs or RAM of individual VMs can be changed without the need to restart the VM. If you need more information, please give me a specific question and I will try to answer.

danielnelson:
Do you know what the hypervisor is?

caapos:
I will find out and send it to you, but the bug would reproduce on ANY hypervisor, because of the nature of the code for checking the number of CPUs.

danielnelson:
True, I'm just trying to make the use case a little more concrete with a specific example.  I just looked up how to do this with kvm + libvirt and was able to do this in a test VM, so I think this will be a good enough example and also helpful for testing.

caapos:
Thanks a lot, I will give you specific information about the hypervisor when the rest of our team is awake :)

danielnelson:
Support for updating the number of CPUs dynamically has been added, though we only have Linux support for now.

Lomanic:
> we only have Linux support for now

FYI [darwin](https://github.com/shirou/gopsutil/blob/58dc3b28aaabbc5d4fdd09a8146279276305aa10/cpu/cpu_darwin.go#L89-L103) and [windows](https://github.com/shirou/gopsutil/blob/58dc3b28aaabbc5d4fdd09a8146279276305aa10/cpu/cpu_windows.go#L226-L255) are now also supported in gopsutil.

danielnelson:
Awesome job @Lomanic, Telegraf really couldn't exist without you and @shirou's hard work on gopsutil.

[...]

PR #5519, opened 2019-03-04
Interrupts+spurious

richardelling:
Adds an optional `spurious_interrupts` measurement for tracking spurious interrupts.

Also, removes metrics for phantom CPUs: those who may be reported in the /proc/softirqs, but might not actually exist and only include zeroes. See also https://github.com/influxdata/telegraf/issues/5451

### Required for all PRs:


plugins/inputs/interrupts/interrupts.go
```diff
@@ -90,14 +98,63 @@ scan:
...
+		if i > validCpuIndex {
+			validCpuIndex = i
+		}
+	}
+	// remove data for any CPUs above the validCpuIndex
+	validCpuCount := validCpuIndex + 1
+	for i := 0; i < len(irqs); i++ {
+		if len(irqs[i].Cpus) > validCpuCount {
+			irqs[i].Cpus = append(irqs[i].Cpus[:validCpuCount])
+		}
```
danielnelson: Can you explain this hunk?  The comments are nice but they don't tell me the why behind it.

plugins/inputs/interrupts/README.md
```diff
@@ -13,65 +14,99 @@ The interrupts plugin gathers metrics about IRQs from `/proc/interrupts` and `/p
...
+| device | description of device |
+| irq | IRQ name |
+| type | interrupt type |
+
+##### spurious_interrupts fields 
+| field | type | units | description |
+|:---:|:---:|:---:|---|
+| count | counter | events | number of times the interrupt has been handled (modulo 100,000) |
+| total | counter | events | total number of times the interrupt has been handled |
+| unhandled | counter | events | number of times an interrupt was not handled |
```
danielnelson: This looks nice but we need to stick to the previous format style to match https://raw.githubusercontent.com/influxdata/telegraf/master/plugins/inputs/EXAMPLE_README.md

plugins/inputs/interrupts/README.md
```diff
@@ -13,65 +14,99 @@ The interrupts plugin gathers metrics about IRQs from `/proc/interrupts` and `/p
   ## deployments.
   # cpu_as_tag = false
 
+  ## spurious interrupt counters can be collected
+  # spurious = false
```
danielnelson: Let's just enable this by default, and rely on the [metric filtering](https://github.com/influxdata/telegraf/blob/master/docs/CONFIGURATION.md#metric-filtering) for those who don't want spurious counts per IRQ.

plugins/inputs/interrupts/interrupts.go
```diff
@@ -112,16 +169,19 @@ func (s *Interrupts) Gather(acc telegraf.Accumulator) error {
...
 		f, err := os.Open(file)
 		if err != nil {
-			acc.AddError(fmt.Errorf("Could not open file: %s", file))
+			acc.AddError(fmt.Errorf("could not open file: %s", file))
 			continue
 		}
-		defer f.Close()
-		irqs, err := parseInterrupts(f)
+		irqs, err := parseInterrupts(f, s.Spurious)
+		_ = f.Close()
```
danielnelson: The previous defer method is actually preferred, since it will be less likely to be broken in future code updates.

plugins/inputs/interrupts/interrupts.go
richardelling: Sure, it covers a case that not everyone will see in their distro, so it can be confusing if you've never seen it.

plugins/inputs/interrupts/README.md
richardelling: will do, though I will assert your standard is ugly and takes up too much screen space :-)

plugins/inputs/interrupts/README.md
richardelling: ok

plugins/inputs/interrupts/interrupts.go
richardelling: The problem with the previous method is that the static code checkers complain about the `f.close()` return code being ignored, which it is. Two choices:

1. add proper func for `defer f.Close()` to explicitly ignore the return value
2. add comment for static checkers to ignore the ignored return value error

I'm fine either way. Do you have a preference?

danielnelson: I'm not our goal to keep static code checkers quiet, outside of the ones required by the integration tests (go vet in particular).  So I'd rather just leave it as before, I don't find assigning to `_` to be helpful in explaining the code.

richardelling: I think I'll prefer to f.Close() in the loop. If defer is in a loop, we trigger possible leak detection because the defer's closure is the function, not the loop. In this particular location inside `Gather`, we loop twice, not a real problem. In `parseInterrupts` we open one file per irq, this can easily grow into the hundreds of open file descriptors, so it is better to close in the loop.

danielnelson: Good point, this is possibly an indication that we should extract a function.

richardelling: yeah, I'm not particularly happy with the overall design, lemme try a more detailed refactor, including some improved testing

srebhan:
@richardelling any update on this? The code looks quite nice already and it would be a pitty if it bit-rots like this.

srebhan:
@richardelling do you see a chance to work on this PR?

srebhan:

richardelling:
@srebhan thanks for the ping, I'll not have time until next month at the earliest. I

srebhan:
@richardelling thank you for letting me know. Please drop me a note once you get to it (or decide to drop this PR). :-)

richardelling:
@srebhan closing this in favor of a newer PR #9496

Issue #5527, opened 2019-03-04
Allow using floating point numbers in statsd percentiles


Pitxyoki:
## Feature Request

Opening a feature request kicks off a discussion.

### Proposal:
We should be able to define percentiles with decimal places when using statsd as the source of input.

### Current behavior:
If we try to define a percentile array with floating point numbers, for example: `[99.0, 99.9, 99.99, 100.0]`, Telegraf won't even start and will stop with the following error:

> 2019-03-04T21:04:26Z I! Starting Telegraf 1.9.2
> 2019-03-04T21:04:26Z I! Using config file: /etc/telegraf/telegraf.conf
> 2019-03-04T21:04:26Z E! [telegraf] Error running agent: Error parsing /etc/telegraf/telegraf.conf, line 1579: statsd.Statsd.Percentiles: `int' is not float32 or float64

### Desired behavior:
We should be able to define more fine-grained percentiles. 

### Use case:
After getting more than 100 data points per submission interval, it becomes important to see percentiles between 99.0 and 100.0.

danielnelson:
Closed in #5572

PR #5528, opened 2019-03-04
Floats in statsd percentiles

Pitxyoki:
### Required for all PRs:

This addresses #5527. Note that this will be a breaking change for those using the plugin with the percentile array configured with integer numbers. It will now fail in those cases and only work with floats.

The already-present Unit Tests were only slightly changed to accept the correct type. No new tests were added.
A signed CLA should be on its way.

### DIFF
--- a/plugins/inputs/statsd/README.md
+++ b/plugins/inputs/statsd/README.md
@@ -34,8 +34,9 @@
   ## Reset timings & histograms every interval (default=true)
   delete_timings = true
 
-  ## Percentiles to calculate for timing & histogram stats
-  percentiles = [90]
+  ## Percentiles to calculate for timing & histogram stats.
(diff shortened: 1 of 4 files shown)


glinton:
Can you make the pr against master?

danielnelson:
For the backwards compatibility issue, we can define a number type.  Place this in `internal/internal.go`:
```go
type Number struct {
	Value float64
}

func (n *Number) UnmarshalTOML(b []byte) error {
	value, err := strconv.ParseFloat(string(b), 64)
	if err != nil {
		return err
	}

	n.Value = value
	return nil
}
```

Then use a `[]internal.Number` as the type in the plugin.

Pitxyoki:
Thank you for your feedback. This will continue in #5572.

PR #5542, opened 2019-03-06
Add percent_packet_loss field to ping input on error

### LINKED ISSUE #5499: [[inputs.ping]] percent_packet_loss not written to InfluxDB when ping command fails
### Relevant telegraf.conf:

```
[[inputs.ping]]
  urls = [
    "host-123",
    "host-124",
  ]
```
### System info:

telegraf version: 1.9.5
operating system: Ubuntu 16.04.5 LTS

### Steps to reproduce:

We used to monitor all the hosts in our infrastructure using the `percent_packet_loss` returned from the ping plugin. When the value would exceed a given treshold an alert was triggered. Formerly it appeared to always write the percent_packet_loss even when the ping command failed. Since some of the latest updates (#4703 looks like a candidate) it seems to return early and write the `result_code` but does not fill the `percent_packet_loss` which leads to the respective alert not being triggered.

### Expected behavior:

`percent_packet_loss` is always 100 when an error occurs during the ping command execution.

```
time                percent_packet_loss result_code
----                ------------------- -----------
1551250190000000000 100                 2
1551250200000000000 100                 2
```

### Actual behavior:

`percent_packet_loss` is always empty when an error occurs during the ping command execution.

```
time                percent_packet_loss result_code
----                ------------------- -----------
1551255636000000000                     2
1551255646000000000                     2
```

### Additional info:

Is it really expected that `percent_packet_loss` is empty in case of an ping error ? I would expect that `percent_packet_loss` is always 100 when an error occurs during ping execution. If that assumption is correct (and as it was like that earlier) I would try to implement a fix.

glinton:
Resolves #5499

### DIFF
--- a/plugins/inputs/ping/ping.go
+++ b/plugins/inputs/ping/ping.go
@@ -114,6 +114,7 @@ func (p *Ping) pingToURL(u string, acc telegraf.Accumulator) {
 	if err != nil {
 		acc.AddError(err)
 		fields["result_code"] = 1
+		fields["percent_packet_loss"] = float64(100.0)
 		acc.AddFields("ping", fields, tags)
 		return
 	}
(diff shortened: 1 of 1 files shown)


danielnelson:
Closing, instead let's work on improving the timeout log message.

PR #5543, opened 2019-03-06
Base64-encoding/decoding parameters for Google Cloud PubSub plugins

emilymye:
### Required for all PRs:

This enables several GCP services that can and only output base64-encoded JSON to PubSub (and thus input into Telegraf/TICK). Enabled base-64 encoded for output of metrics just because symmetry is nice and it's suggested that PubSub wants base64-encoded data no matter what.

### DIFF
--- a/plugins/inputs/cloud_pubsub/README.md
+++ b/plugins/inputs/cloud_pubsub/README.md
@@ -79,7 +79,7 @@ and creates metrics using one of the supported [input data formats][].
   ## Optional. If true, Telegraf will attempt to base64 decode the 
   ## PubSub message data before parsing. Many GCP services that
   ## output JSON to Google PubSub base64-encode the JSON payload.
-  # base_64_data = true
+  # base64_data = false
 ```
 
 ### Multiple Subscriptions and Topics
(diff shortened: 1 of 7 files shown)


plugins/inputs/cloud_pubsub/README.md
```diff
@@ -75,6 +75,11 @@ and creates metrics using one of the supported [input data formats][].
   ## 1. Note this setting does not limit the number of messages that can be
   ## processed concurrently (use "max_outstanding_messages" instead).
   # max_receiver_go_routines = 0
+  
+  ## Optional. If true, Telegraf will attempt to base64 decode the 
+  ## PubSub message data before parsing. Many GCP services that
+  ## output JSON to Google PubSub base64-encode the JSON payload.
+  # base_64_data = true
```
danielnelson: Minor nit, call this `base64_data`, `base64_payload`, or `base64_message`.

Since the default is false, document it like:
```
  ## Blah blah blah
  # base64_data = false
```

emilymye:
@danielnelson This is actually only setting encoded/decoded values for pubsub.Message.Data, which corresponds to the "data" field in a JSON representation of the message sent to the API.

PR #5544, opened 2019-03-07
Collect cloudwatch stats in a more efficient and cost effective manner

### LINKED ISSUE #3255: Error in plugin [inputs.cloudwatch]: RequestError: send request failed
## Bug report
CloudWatch input seems to leave many TIME_WAIT connections which eventually causes it to return too many open files error.

### Relevant telegraf.conf:
60 or so different cloudwatch configs (we use Puppet so the files are separated) for various namespaces (AWS/AutoScaling, AWS/ECS, AWS/RDS)
```
[[inputs.cloudwatch]]
  delay = "1m"
  interval = "1m"
  namespace = "AWS/RDS"
  period = "1m"
  profile = "my_profile"
  ratelimit = 25
  region = "us-east-1"
```

### System info:
telegraf-1.3.5-1.x86_64
CentOS Linux release 7.2.1511 (Core)

### Steps to reproduce:
Create telegraf config with many CloudWatch inputs. I have 60 configured.

### Expected behavior: 
TCP connections should be handled in a matter where TIME_WAITs don't linger (reused or closed faster perhaps).

### Actual behavior: 
TCP connections seem to be left open which causes TIME_WAITs and uses up open files. Telegraf log shows many lines like:

```
Sep 20 19:27:08 my_host telegraf[23200]: 2017-09-20T19:27:08Z E! Error in plugin [inputs.cloudwatch]: RequestError: send request failed
Sep 20 19:27:08 my_host telegraf[23200]: caused by: Post https://monitoring.us-east-1.amazonaws.com/: dial tcp xx.xx.xx.xx:443: socket: too many open files
```

### Additional info:
Past week of TCP connections. The drop is me experimenting with different kernel parameters.
screenshot at sep 20 15-54-57

I've tried tweaking network stack but didn't seem to help.

```
$ sysctl net.ipv4.tcp_fin_timeout net.ipv4.tcp_tw_recycle net.ipv4.tcp_tw_reuse fs.file-max
net.ipv4.tcp_fin_timeout = 1
net.ipv4.tcp_tw_recycle = 0
net.ipv4.tcp_tw_reuse = 0
fs.file-max = 500000
```
These settings don't seem to be honored. Since this is systemd system adding `LimitNOFILE=65536` to the telegraf.service unit file works to increase open files. However, this requires a manual change to a file that in the telegraf package and doesn't really fix the TIME_WAIT issue. On a side, I see that `LimitNOFILE=65536` is in the influxdb.service unit file. Maybe it makes sense to add for telegraf too?

### LINKED ISSUE #5420: CloudWatch Input: Use GetMetricData vs. GetMetricStatistics API
## Feature Request
Would it be possible to use GetMetricData vs. GetMetricStatistics calls for CloudWatch input?

GetMetricData is more efficient vs. GetMetricStatistics and is [recommended by AWS](https://aws.amazon.com/premiumsupport/knowledge-center/cloudwatch-getmetricdata-api/). Switching calls will save money and allow large namespaces to be consumed more efficiently and timely.

From AWS support:
> The Service Team found that in the third party tool that was provided, the Go program is making GetMetricStatistics API calls instead of GetMetricData. They recommend using GetMetricData since it's always efficient and it's the recommended solution for metric retrieval over GetMetricStatistics. GetMetricData is also more cost effective. You can ask for more data points for a single metric by using GetMetricData rather than by using GetMetricStatistics. To get the same number of data points, you must perform 70 GetMetricStatistics calls. Here is a reference with more information: https://aws.amazon.com/premiumsupport/knowledge-center/cloudwatch-getmetricdata-api/

### Proposal:
Use GetMetricData vs. GetMetricStatistics calls for CloudWatch input.

### Current behavior:
GetMetricStatistics calls are used

### Desired behavior:
Use GetMetricData calls

Real example:

AWS/SQS in one of our accounts has ~40k MetricName's which takes ~15m to complete using:

```TOML
[[inputs.cloudwatch]]
delay = "1m"
fieldpass = ["approximate_number_of_messages_visible_maximum"]
interval = "5m"
namespace = "AWS/SQS"
period = "5m"
ratelimit = 20
```

For the sake of argument, let's say we need all that data. Rate limit can be increased, but I also have 12 other namespaces that need to be consumed.

Rates can be increased on AWS side but if consuming can be more efficient, that would be preferred.

### LINKED ISSUE #5609: Cloudwatch Input: min/max/average/sum etc selective statistics
## Feature Request

It would be beneficial (for AWS costs) to be able to select specific "statistics" only for the given metrics/dimensions, so when pulling a high volume of metrics from Cloudwatch you can keep costs under control.

### Proposal:

Eg. you may only care about `average` and `max`, but not `min` or `sum`. This would halve the % of retrievals involved.

https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_GetMetricData.html seems to provide the ability to specify, there also seems to be some on https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_GetMetricStatistics.html

### Current behavior:

Retrieves all "statistics" for a given metric / dimension, which is somewhere around 6~ each.

### Desired behavior:

Selectively pull the ones you want.

End user cost, plus this will have a rate limit benefit also (efficiency).

glinton:
Resolves #5420 
Resolves #5609
closes #3255

Due to the structure of the data returned by `GetMetricData` (as opposed to `GetMetricStatistics`) metrics will no longer be tagged with the unit type.

This version makes one call to the AWS api and creates metrics based on the dimensions and specified metric names from the config file.

Config:
```toml
  [[inputs.cloudwatch.metrics]]
    names = ["NetworkIn"]
    [[inputs.cloudwatch.metrics.dimensions]]
      name = "AutoScalingGroupName"
      value = "k8s-ci"
  [[inputs.cloudwatch.metrics]]
    names = ["NetworkOut"]
    [[inputs.cloudwatch.metrics.dimensions]]
      name = "AutoScalingGroupName"
      value = "k8s-frontline"
```
Output:
```
cloudwatch_aws_ec2,AutoScalingGroupName=k8s-frontline,region=us-east-1 network_out_sum=6625229694,network_out_average=3312614847,network_out_maximum=3345284533,network_out_minimum=3279945161 1551919140000000000
cloudwatch_aws_ec2,AutoScalingGroupName=k8s-ci,region=us-east-1 network_in_minimum=1167141,network_in_sum=45580787,network_in_average=15193595.666666666,network_in_maximum=42813872 1551919140000000000
```

Todo:
 - [x] tests
 - [x] handle pagination
 - [x] handle window better


CpuID:
> Due to the structure of the data returned by GetMetricData (as opposed to GetMetricStatistics) metrics will no longer be tagged with the unit type.

❤️ I planned to rename this out anyway tbh

danielnelson:
@glinton regarding the unit type, I'm not super attached to it but it seems we still know that information when we make the query, are you still thinking to remove it?

plugins/inputs/cloudwatch/README.md
```diff
@@ -68,12 +68,20 @@ API endpoint. In the following order the plugin will attempt to authenticate.
   ## See http://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/cloudwatch_limits.html
   ratelimit = 200
 
+  ## Namespace-wide statistic filters (only gets used if no metrics are defined). Optional.
+  # statistic_exclude = [ "average", "sum", min", "max", sample_count" ]
+  # statistic_include = [ "average", "sum", min", "max", sample_count" ]
```
CpuID: I think `min` and `max` need to be `minimum` and `maximum` here in certain use cases? noticed `max` doesn't return under the `AWS/ECS` namespace if I specify it (silently omits), whereas `maximum` works fine

skimming the code below, possibly just a docs bug?

plugins/inputs/cloudwatch/README.md
```diff
@@ -68,12 +68,20 @@ API endpoint. In the following order the plugin will attempt to authenticate.
...
+
   ## Metrics to Pull (optional)
   ## Defaults to all Metrics in Namespace if nothing is provided
   ## Refreshes Namespace available metrics every 1h
   [[inputs.cloudwatch.metrics]]
     names = ["Latency", "RequestCount"]
 
+    ## Statistic filters for metric.  Optional.
+    # statistic_exclude = [ "average", "sum", min", "max", sample_count" ]
+    # statistic_include = [ "average", "sum", min", "max", sample_count" ]
```
CpuID: see above, re `min`/`max` vs `minimum`/`maximum`

plugins/inputs/cloudwatch/cloudwatch.go
```diff
@@ -311,62 +395,101 @@ func (c *CloudWatch) fetchNamespaceMetrics() ([]*cloudwatch.Metric, error) {
...
+		// todo: group these once beforehand if possible to avoid looping through all results metrics number of times
+		results := getResults(*metric, metricDataResults)
+		for _, result := range results {
+			for _, dimension := range metric.Dimensions {
+				tags[snakeCase(*dimension.Name)] = *dimension.Value
+			}
+			if len(result.Timestamps) == 0 || result.Timestamps[0] == nil {
+				continue
+			}
+			// todo: determine if we need to handle multiple values.
```
CpuID: maybe? https://github.com/influxdata/telegraf/issues/5420#issuecomment-475783462 :)

CpuID:
> @glinton regarding the unit type, I'm not super attached to it but it seems we still know that information when we make the query, are you still thinking to remove it?

I think the main upside to retaining it would be retaining namespace compatibility for existing users. ie, user upgrades, things "just work" in terms of other subsystems/dashboards that expect a pre-existing metric namespace.

maybe having a feature flag to disable the units would be extremely nice though...

plugins/inputs/cloudwatch/cloudwatch.go
glinton: I'm not certain we'd gain any benefit from parsing these. I think I'm going to stop logging that and just always take the first value. If we looped through, it would take the last value as the field name will always be the same. Since the timestamp is re-updated on each result, it would have the same timestamp as well.

glinton:
I'm not certain that we know the unit type at query time. What about adding a config option to set a unit type, as the user may know what units a metric/namespace uses? If it's not set, there is no unit type tag.

danielnelson:
Don't worry about the unit tag then, we will add it to the release notes, there are other ways one could manually attach tags.

plugins/inputs/cloudwatch/cloudwatch.go
danielnelson: I believe this is something we need to support or we would break  https://github.com/influxdata/telegraf/issues/4643.

CpuID:
@glinton can you drop some "napkin math" in here re how you foresee the "cost effective manner" part playing out?

background: I've been running this branch in "protoduction" for a few days, reading a reasonable volume of metrics from CW:

(4 `inputs.cloudwatch` stanzas, EBS (5m), ECS (1m), EC2 (5m), RDS (5m), most selectively, ECS all dimensions/statistics)

```
2019-03-25T19:00:03Z D! [outputs.graphite] wrote batch of 352 metrics in 72.25408ms
2019-03-25T19:00:03Z D! [outputs.graphite] buffer fullness: 1 / 10000 metrics. 
2019-03-25T19:01:00Z D! [outputs.graphite] wrote batch of 593 metrics in 106.623415ms
2019-03-25T19:01:00Z D! [outputs.graphite] buffer fullness: 0 / 10000 metrics. 
2019-03-25T19:02:03Z D! [outputs.graphite] wrote batch of 664 metrics in 178.746237ms
2019-03-25T19:02:03Z D! [outputs.graphite] buffer fullness: 3 / 10000 metrics. 
2019-03-25T19:03:03Z D! [outputs.graphite] wrote batch of 636 metrics in 109.464365ms
2019-03-25T19:03:03Z D! [outputs.graphite] buffer fullness: 1 / 10000 metrics. 
2019-03-25T19:04:01Z D! [outputs.graphite] wrote batch of 376 metrics in 63.310413ms
2019-03-25T19:04:01Z D! [outputs.graphite] buffer fullness: 3 / 10000 metrics. 
2019-03-25T19:05:01Z D! [outputs.graphite] wrote batch of 260 metrics in 199.167616ms
2019-03-25T19:05:01Z D! [outputs.graphite] buffer fullness: 5 / 10000 metrics. 
2019-03-25T19:06:04Z D! [outputs.graphite] wrote batch of 772 metrics in 115.374741ms
2019-03-25T19:06:04Z D! [outputs.graphite] buffer fullness: 2 / 10000 metrics. 
2019-03-25T19:07:01Z D! [outputs.graphite] wrote batch of 481 metrics in 52.644945ms
2019-03-25T19:07:01Z D! [outputs.graphite] buffer fullness: 1 / 10000 metrics. 
2019-03-25T19:08:03Z D! [outputs.graphite] wrote batch of 689 metrics in 212.339554ms
2019-03-25T19:08:03Z D! [outputs.graphite] buffer fullness: 4 / 10000 metrics. 
2019-03-25T19:09:03Z D! [outputs.graphite] wrote batch of 467 metrics in 86.771158ms
2019-03-25T19:09:03Z D! [outputs.graphite] buffer fullness: 3 / 10000 metrics. 
2019-03-25T19:10:01Z D! [outputs.graphite] wrote batch of 160 metrics in 39.880934ms
2019-03-25T19:10:01Z D! [outputs.graphite] buffer fullness: 1 / 10000 metrics. 
2019-03-25T19:11:01Z D! [outputs.graphite] wrote batch of 697 metrics in 163.696045ms
2019-03-25T19:11:01Z D! [outputs.graphite] buffer fullness: 4 / 10000 metrics. 
2019-03-25T19:12:02Z D! [outputs.graphite] wrote batch of 592 metrics in 171.005482ms
2019-03-25T19:12:02Z D! [outputs.graphite] buffer fullness: 5 / 10000 metrics. 
2019-03-25T19:13:04Z D! [outputs.graphite] wrote batch of 687 metrics in 207.710244ms
2019-03-25T19:13:04Z D! [outputs.graphite] buffer fullness: 4 / 10000 metrics. 
2019-03-25T19:14:02Z D! [outputs.graphite] wrote batch of 407 metrics in 47.990755ms
2019-03-25T19:14:02Z D! [outputs.graphite] buffer fullness: 1 / 10000 metrics. 
2019-03-25T19:15:03Z D! [outputs.graphite] wrote batch of 329 metrics in 75.403892ms
2019-03-25T19:15:03Z D! [outputs.graphite] buffer fullness: 1 / 10000 metrics. 
```

When I enabled ECS only last week on telegraf master, it took our Cloudwatch usage from $2.12/day (unrelated usage, baseline) to $9~/day ($6.88~/day in telegraf usage, $206.40~/mth. I then enabled the other 3 stanzas and used this branch, left it going over the weekend and checked our daily usage, $10.26/day ($8.14~/day in telegraf usage, $244.20~/mth)

I'm wondering if my assumptions on how `GetMetricData` is billed are wrong here...? how do you interpret the pricing model?

----------------------------

My equivalent "napkin math", which I feel may be off:

Threw some print debugging in my local build of the https://github.com/influxdata/telegraf/pull/5544 branch, to see how many `GetMetricData` calls were fired off:

- for the 235~ per minute metrics for `AWS/ECS`, I saw 8 API calls total. the bulk collection vs the old `GetMetricStatistic` is *so* much more efficient
- using https://aws.amazon.com/cloudwatch/pricing/

> You can request up to five statistics for the same metric in a single GetMetricData API request. Additional statistics are billed as an additional metric.
> $0.01/1,000 metrics requested using GetMetricData
> $0.01/1,000 GetMetricStatistics, ListMetrics, PutMetricData, GetDashboard, ListDashboards, PutDashboard and DeleteDashboards requests

- in the code, the breakout comment says `// loop through metrics and send groups of 100 at a time to gatherMetrics`
- so the code does 100 metrics per `GetMetricData` call, vs a single metric in the current "released" code per `GetMetricStatistic` call
- some math:
  - 235~ metrics per min divided by 5 = 47 billable `GetMetricData` API calls. Could be slightly higher depending on the qty of data queries per API call, but let's go with that for napkin math sake
  - thats 2030400 billable `GetMetricData` API calls per month
  - 2030.4 x $0.01 = $20.304/mth
- the ECS metrics are per minute, the EBS + RDS + EC2 metrics are per 5mins so retrieved 80% less often, a minority of the overall usage.

what we are seeing right now:

- 391~ x 1 `GetMetricStatistic` calls, per min
  - 40% higher due to no statistic filtering, so we get sum/min/max/avg/sample_count, instead of just avg/max/sample_count
- thats 16891200 billable `GetMetricStatistic` API calls per mth, for ECS metrics alone
- $168.912/mth, or $5.6304/day
- this lines up pretty closely with the $7~/day I was seeing for ECS + EC2 + RDS + EBS metrics (last 24hrs $9~ vs a couple of days ago $2~ - the $2 is mostly cloudwatch-to-graphite)

and after deploying with the 5min metrics added also (3 extra stanzas):

- all the other 5min AWS metrics enabled are about 665280 stats per month. the 5min metric count is something like 1k~, some napkin math tells me $17.28~/mth for those.

I'd say it's pretty safe to say that https://github.com/influxdata/telegraf/pull/5544 will bring our telegraf Cloudwatch cost down to something like $37~/mth instead of $206~/mth for latest released telegraf

glinton:
When I started, I didn't read the pricing docs, but it makes it sound like it'll be the same cost for either way?

>$0.01/1,000 metrics requested using GetMetricData
>$0.01/1,000 GetMetricStatistics, ListMetrics, PutMetricData, GetDashboard, ListDashboards, PutDashboard and DeleteDashboards requests

Sure, we make up to 20-100x fewer calls at best using `GetMetricData.

Say you want stats for 100 metrics. Using `GetMetricStatistic` API, it would be one call per metric and you would get the 5 stats back (max, min, avg, sum, count), resulting in 100 API calls. (statistics API is billed per request)
Using `GetMetricData` and no filtering, it would be 5 calls, because you can request 100 data points per call (20 metrics with 5 stats per request = 100 data points)
Using `GetMetricData` and only requesting the `average` of all 100 metrics, it would be 1 request.

What has me confused now is that the pricing for `GetMetricData` is per metric requested. Sure it's more efficient, but the pricing doc makes it sound like it won't actually save any money.

[In their knowledge-center](https://aws.amazon.com/premiumsupport/knowledge-center/cloudwatch-getmetricdata-api/), it says:

>GetMetricData is cheaper. You can ask for more data points for a single metric by using GetMetricData rather than by using GetMetricStatistics. To get the same number of data points, you must perform 70 GetMetricStatistics calls.

Which makes it sound not only cheaper computationally, but financially as well.

I know docs docs are eventually consistent, so there's a chance (and it seems you've experienced it) that this new API call does save money while making fewer requests.

CpuID:
> I know docs docs are eventually consistent, so there's a chance (and it seems you've experienced it) that this new API call does save money while making fewer requests.

I wonder if it's really only a saving for someone retrieving bulk metrics, ie more than 1 data point per interval? I guess one route there if you are willing to accept a delayed retrieval, is tune the configuration (still thinking through the exact combination) to pull say 5 x 1min metrics, once per 5mins using `GetMetricData` since that API call permits it? That would reduce the retrieval cost substantially, at the expense of retrieval "latency".

I think I concur with your thoughts, from my initial reading. Admittedly I did just wake up though, so if this opinion changes... :)

glinton:
That's a great point.. I'll think about a way to preserve the timestamps when and if multiple data points are returned.

CpuID:
@glinton I see you removed WIP from the title, are you considering this ready to merge?

glinton:
@CpuID pending review, yes I feel it's ready to merge

plugins/inputs/cloudwatch/cloudwatch.go
```diff
@@ -204,22 +248,58 @@ func (c *CloudWatch) Gather(acc telegraf.Accumulator) error {
...
+	wg.Add(1)
+	go func(inm []*cloudwatch.MetricDataQuery) {
+		defer wg.Done()
+		result, err := c.gatherMetrics(c.getDataInputs(inm))
+		if err != nil {
+			acc.AddError(err)
+			// return
+		}
+		results = append(results, result...)
+	}(queries[(groups * 100):])
```
goller: I think remainder case can be a problem if len(queries) % 100 == 0.

plugins/inputs/cloudwatch/cloudwatch.go
```diff
@@ -311,89 +396,192 @@ func (c *CloudWatch) fetchNamespaceMetrics() ([]*cloudwatch.Metric, error) {
...
+	resp, err := c.client.GetMetricData(params)
 	if err != nil {
-		return err
+		return nil, errors.New("Failed to get metric data - " + err.Error())
 	}
 
-	for _, point := range resp.Datapoints {
+	// todo: determine if we need to handle pagination.
+	if resp.NextToken != nil {
+		log.Println("[inputs.cloudwatch] W! UNHANDLED PAGINATED RESULTS!!", *resp.NextToken)
```
goller: when I've written AWS code I typically put it in a for {} loop and break when NextToken is nil.

@gianarb how are you doing AWS paging code now-a-days?  Do you just do a forever loop?

plugins/inputs/cloudwatch/cloudwatch.go
```diff
@@ -311,89 +396,192 @@ func (c *CloudWatch) fetchNamespaceMetrics() ([]*cloudwatch.Metric, error) {
...
-			"unit":   snakeCase(*point.Unit),
 		}
 
-		for _, d := range metric.Dimensions {
-			tags[snakeCase(*d.Name)] = *d.Value
+		type metric struct {
+			fields map[string]interface{}
+			tags   map[string]string
 		}
+		thing := map[time.Time]metric{}
```
goller: heh... looks like a leftover here... my "foo" is the word howdy, btw :)

plugins/inputs/cloudwatch/cloudwatch.go
```diff
@@ -311,89 +396,192 @@ func (c *CloudWatch) fetchNamespaceMetrics() ([]*cloudwatch.Metric, error) {
...
+		for _, result := range metricDataResults {
+			if nameMatch(query.id, *result.Id) {
+				for _, dimension := range query.metric.Dimensions {
+					tags[snakeCase(*dimension.Name)] = *dimension.Value
+				}
+				if len(result.Timestamps) != len(result.Values) {
+					log.Println("[inputs.cloudwatch] W! MISMATCHED TIMESTAMP/VALUE LENGTH!!", len(result.Timestamps), len(result.Values))
```
goller: As an aside, when would this happen?

plugins/inputs/cloudwatch/cloudwatch.go
```diff
@@ -311,89 +396,192 @@ func (c *CloudWatch) fetchNamespaceMetrics() ([]*cloudwatch.Metric, error) {
...
-}
-
 func formatMeasurement(namespace string) string {
 	namespace = strings.Replace(namespace, "/", "_", -1)
 	namespace = snakeCase(namespace)
-	return fmt.Sprintf("cloudwatch_%s", namespace)
+	return "cloudwatch_" + namespace
 }
 
 func snakeCase(s string) string {
```
goller: What do you think of calling this function something like `sanitizeName` or `encode` or something ask it is doing more than just `snakeCase`?

plugins/inputs/cloudwatch/cloudwatch.go
```diff
@@ -311,89 +396,192 @@ func (c *CloudWatch) fetchNamespaceMetrics() ([]*cloudwatch.Metric, error) {
...
+	id     string
+}
+
+// get all of the possible queries so we can send groups of 100
+func (c *CloudWatch) getDataQueries(filteredMetrics []filteredMetric) ([]*cloudwatch.MetricDataQuery, error) {
+	if c.queryCache != nil && c.metricCache != nil && c.metricCache.IsValid() {
+		return c.queryCache, nil
+	}
+
+	// todo: lock
```
goller: Does this todo need to happen?

plugins/inputs/cloudwatch/cloudwatch_test.go
```diff
@@ -6,48 +6,98 @@ import (
...
+				},
+			},
+		},
+	}, nil
 }
 
+func TestSnakeCase(t *testing.T) {
+	assert.Equal(t, "cluster_name", snakeCase("Cluster Name"))
+	assert.Equal(t, "broker_id", snakeCase("Broker ID"))
+}
```
goller: nit ... add a space between these functions.

plugins/inputs/cloudwatch/cloudwatch.go
```diff
@@ -58,7 +68,7 @@ type (
 
 	cloudwatchClient interface {
 		ListMetrics(*cloudwatch.ListMetricsInput) (*cloudwatch.ListMetricsOutput, error)
-		GetMetricStatistics(*cloudwatch.GetMetricStatisticsInput) (*cloudwatch.GetMetricStatisticsOutput, error)
+		GetMetricData(*cloudwatch.GetMetricDataInput) (*cloudwatch.GetMetricDataOutput, error)
```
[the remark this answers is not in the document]
glinton: These are wrappers to the cloudwatch client's functions. This interface is present for testing and would break if we changed that, as the cw client doesn't implement `GetMetrics`

plugins/inputs/cloudwatch/cloudwatch.go
```diff
@@ -19,14 +21,16 @@ import (
...
-		Profile     string `toml:"profile"`
-		Filename    string `toml:"shared_credential_file"`
-		Token       string `toml:"token"`
-		EndpointURL string `toml:"endpoint_url"`
+		Region           string   `toml:"region"`
+		AccessKey        string   `toml:"access_key"`
+		SecretKey        string   `toml:"secret_key"`
+		RoleARN          string   `toml:"role_arn"`
+		Profile          string   `toml:"profile"`
+		Filename         string   `toml:"shared_credential_file"`
```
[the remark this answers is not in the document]
glinton: Since this a pre-existing plugin, changing options isn't smiled upon. Perhaps for 2.0? Unless it makes sense to add in `CredentialPath` and start deprecating `Filename`. This is slated for a minor release, so I'm fine with some config options improving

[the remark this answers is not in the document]
glinton: I guess to users, it's still `shared_credential_file`... I'll update the code to be more clear and use `CredentialPath`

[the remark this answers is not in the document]
glinton: It was likely `Filename` because that's what it is in our aws `CredentialConfig`

plugins/inputs/cloudwatch/cloudwatch.go
```diff
@@ -36,13 +40,19 @@ type (
 		RateLimit   int               `toml:"ratelimit"`
 		client      cloudwatchClient
 		metricCache *MetricCache
+		queryCache  []*cloudwatch.MetricDataQuery
+		queries     []queryData
 		windowStart time.Time
 		windowEnd   time.Time
+
+		Debug bool `toml:"debug"`
```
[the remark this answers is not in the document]
glinton: Same, I put it in, realizing now undocumented, to match the other amazon plugins.

From kinesis output:
`## debug will show upstream aws messages`

plugins/inputs/cloudwatch/cloudwatch.go
glinton: Also how often do you get paginated results, so far neither I nor those testing these changes have come across paginated results.

plugins/inputs/cloudwatch/cloudwatch.go
glinton: Not sure, figured I'd log and be told when it did... I suppose a `todo: check aws code to ensure timestamps and values always match up` would have served the same purpose

plugins/inputs/cloudwatch/cloudwatch.go
glinton: Realizing this is terrible, but I generally don't change pre-existing working code. Maybe my justification was that it would be easier for someone to come back to and not be surprised we aren't "formatting the measurement" anymore. I'm concluding now that I'm in the camp of "improve-all-the-code".

plugins/inputs/cloudwatch/cloudwatch.go
glinton: I guess not, as it's used asynchronously

plugins/inputs/cloudwatch/cloudwatch_test.go
glinton: Can't believe I missed that

plugins/inputs/cloudwatch/cloudwatch.go
glinton: ooh, you are right

glinton: Huh, surprisingly no.
```go
fmt.Println(strings.Repeat("b", 100)[100:])
```
Is completely valid, must be due to the ranging.

plugins/inputs/cloudwatch/cloudwatch.go
```diff
@@ -311,89 +393,196 @@ func (c *CloudWatch) fetchNamespaceMetrics() ([]*cloudwatch.Metric, error) {
...
+						Period: aws.Int64(int64(c.Period.Duration.Seconds())),
+						Stat:   aws.String(cloudwatch.StatisticSampleCount),
+					},
+				})
+			}
+		}
+	}
+
+	if len(dataQueries) == 0 {
+		return nil, errors.New("No metrics found to collect")
```
danielnelson: `no metrics found to collect`

plugins/inputs/cloudwatch/README.md
```diff
@@ -68,12 +68,22 @@ API endpoint. In the following order the plugin will attempt to authenticate.
...
+  # statistic_exclude = [ "average", "sum", minimum", "maximum", sample_count" ]
+  # statistic_include = [ "average", "sum", minimum", "maximum", sample_count" ]
+
   ## Metrics to Pull (optional)
   ## Defaults to all Metrics in Namespace if nothing is provided
   ## Refreshes Namespace available metrics every 1h
   [[inputs.cloudwatch.metrics]]
     names = ["Latency", "RequestCount"]
-
+
```
danielnelson: Clear trailing whitespace

plugins/inputs/cloudwatch/README.md
```diff
@@ -68,12 +68,22 @@ API endpoint. In the following order the plugin will attempt to authenticate.
   ## See http://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/cloudwatch_limits.html
   ratelimit = 200
 
+  ## Namespace-wide statistic filters (only gets used if no metrics are defined). These
```
danielnelson: I would expect these to be applied to metrics unless overridden on a per-metric level.  Don't mention that these are optional, having them commented out is enough signaling.  Show the default settings:
```
  # statistic_include = ["average", "sum", minimum", "maximum", sample_count"]
  # statistic_exclude = []
```

plugins/inputs/cloudwatch/README.md
```diff
@@ -168,5 +178,5 @@ aws cloudwatch get-metric-statistics --namespace AWS/EC2 --region us-east-1 --pe
 
 ```
 $ ./telegraf --config telegraf.conf --input-filter cloudwatch --test
-> cloudwatch_aws_elb,load_balancer_name=p-example,region=us-east-1,unit=seconds latency_average=0.004810798017284538,latency_maximum=0.1100282669067383,latency_minimum=0.0006084442138671875,latency_sample_count=4029,latency_sum=19.382705211639404 1459542420000000000
+> cloudwatch_aws_elb,load_balancer_name=p-example,region=us-east-1 latency_average=0.004810798017284538,latency_maximum=0.1100282669067383,latency_minimum=0.0006084442138671875,latency_sample_count=4029,latency_sum=19.382705211639404 1459542420000000000
```
danielnelson: Need to update line about units above.

plugins/inputs/cloudwatch/cloudwatch.go
```diff
@@ -311,89 +396,192 @@ func (c *CloudWatch) fetchNamespaceMetrics() ([]*cloudwatch.Metric, error) {
...
 
-		for _, d := range metric.Dimensions {
-			tags[snakeCase(*d.Name)] = *d.Value
+		type metric struct {
+			fields map[string]interface{}
+			tags   map[string]string
 		}
+		thing := map[time.Time]metric{}
+		for _, result := range metricDataResults {
+			if nameMatch(query.id, *result.Id) {
```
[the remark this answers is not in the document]
glinton: IDs are required to be unique, so I added the prefix so we could still have a logical group of statistics but have unique id's.

plugins/inputs/cloudwatch/cloudwatch.go
glinton: https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_MetricDataResult.html

plugins/inputs/cloudwatch/cloudwatch.go
```diff
@@ -311,89 +393,196 @@ func (c *CloudWatch) fetchNamespaceMetrics() ([]*cloudwatch.Metric, error) {
...
-			"unit":   snakeCase(*point.Unit),
 		}
 
-		for _, d := range metric.Dimensions {
-			tags[snakeCase(*d.Name)] = *d.Value
+		type metric struct {
+			fields map[string]interface{}
+			tags   map[string]string
 		}
+		results := map[time.Time]metric{}
```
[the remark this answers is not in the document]
glinton: This was to used for adding multiple statistic values from a selected metric to a new measurement (multiple values and timestamps can be returned if the duration is configured correctly). Using a grouper should remove the need for this.

glinton:
@druchoo @CpuID can you test this build? there are some caching improvements and modifications to how metrics are created:

- [windows x64](https://30917-33258973-gh.circle-artifacts.com/0/build/windows/amd64/telegraf-1.11.0~aeda8ac8_windows_amd64.zip)
- [linux x64](https://30917-33258973-gh.circle-artifacts.com/0/build/linux/amd64/telegraf-1.11.0~aeda8ac8_linux_amd64.tar.gz)
- [freebsd x64](https://30917-33258973-gh.circle-artifacts.com/0/build/freebsd/amd64/telegraf-1.11.0~aeda8ac8_freebsd_amd64.tar.gz)

druchoo:
Initial testing of build looks good. I plan on taking a closing look at each namespace tomorrow. We're doing some cost analysis first. Notes I have now:

- It's been running ~1d and we're not really seeing favorable results for cost vs the GetMetricStatistics version. Cost Explorer is showing roughly same cost. Granted there's no filtering in any configs but would have thought there would be savings from just switching API calls. I was hopeful at least from the discussion and napkin math above. 
- I believe [ratelimit section](https://github.com/influxdata/telegraf/compare/feature/5420#diff-813170881edf56bceaa182f8bfd80a65R69) in README needs to be updated. GetMetricData is 50 TPS vs 400 for GetMetricStatistics.
- Can you elaborate a bit on the caching (`cache_ttl`)?

glinton:
* Since the napkin math, I've been worried this was a waste of time. I do believe that cost savings can be seen at the cost of not having immediate stats. Can you try increasing the `interval` to 5-10x what you currently have it set to, but leave the `period` the same (I assume it's `1m), this should limit the number of metrics requested, but will return all backfilled data at 1 minute intervals.

* Thanks, I'll make that update to the README.

* The caching applies to how frequently metrics are listed from a namespace. If you don't have any metrics defined for a namespace, this plugin lists the available metrics no more frequently than `cache_ttl`. This setting also serves as ttl for other cache's used in the plugin to limit processing.

druchoo:
@glinton Totally not a waste of time :-). The performance/speed benefit alone is insane. Before it wasn't practical to consume larger namespaces due to time it took to complete.

Increasing interval showed cost savings with `GetMetricStatistics`. I imagine it will be the same for `GetMetricData`. I think we'll start with actually filtering our metrics and dimensions. Depending on how that goes we can play around with interval.

Maybe worth mentioning.. In many cases, all available metrics are not useful. Similarly, although multiple statistics are available for a given metric, there may only be one meaningful one. Unfortunately, you need to read the CloudWatch docs for each namespace.

Regardless of cost savings, thank you for working on this!

danielnelson:
@druchoo Is this helping with #3255 at all?

CpuID:
> @druchoo @CpuID can you test this build? there are some caching improvements and modifications to how metrics are created:

OOO, I can test next wk if required

CpuID:
ya I think the filtering of metrics helps a lot for $ reasons over the prior implementation.

and I think with longer intervals, $ savings are possible.

this method is definitely the more efficient way to bulk retrieve from CW, so I think it's definitely not a waste

druchoo:
@danielnelson, although TCP connections are way down I don't think the underlying issue is resolved. Connections are still left open but it's looks to be very unlikely this would be an issue now. Just another plus for GetMetricData :-) 

Not sure what to do with #3255. Close I guess?

plugins/inputs/cloudwatch/cloudwatch.go
```diff
@@ -1,67 +1,98 @@
...
+
+	// queryCache caches generated queries.
+	queryCache struct {
+		ttl     time.Duration
+		built   time.Time
+		queries []*cloudwatch.MetricDataQuery
+	}
+
+	// filteredCache caches defined metrics and their filters.
+	filteredCache struct {
```
danielnelson: Should we combine all these caches into a single cache since they all have the same lifetime?  Do we need both filteredMetric and metricCache, as they both seem to hold the `*cloudwatch.Metric`?

plugins/inputs/cloudwatch/cloudwatch.go
```diff
@@ -239,168 +412,208 @@ func (c *CloudWatch) updateWindow(relativeTo time.Time) error {
...
 }
 
-/*
- * Formatting helpers
- */
-func formatField(metricName string, statistic string) string {
-	return fmt.Sprintf("%s_%s", snakeCase(metricName), snakeCase(statistic))
+func init() {
+	inputs.Add("cloudwatch", func() telegraf.Input {
+		ttl, _ := time.ParseDuration("1hr")
```
danielnelson: "1h" is the right string, but actually since this is a "constant" just do `ttl := time.Hour`, probably without variable assignment

plugins/inputs/cloudwatch/cloudwatch.go
```diff
@@ -239,168 +412,208 @@ func (c *CloudWatch) updateWindow(relativeTo time.Time) error {
...
-		token = resp.NextToken
-		more = token != nil
-	}
-
-	c.metricCache = &MetricCache{
-		Metrics: metrics,
-		Fetched: time.Now(),
-		TTL:     c.CacheTTL.Duration,
+		results = append(results, resp.MetricDataResults...)
+		if resp.NextToken == nil {
```
danielnelson: We do a very similar pagination above for ListMetrics, but the style is different here vs there.  If there isn't a reason for this then we should use the same style (for the record I like this one more).

plugins/inputs/cloudwatch/cloudwatch.go
```diff
@@ -239,168 +412,208 @@ func (c *CloudWatch) updateWindow(relativeTo time.Time) error {
...
-	for _, point := range resp.Datapoints {
+	for _, result := range metricDataResults {
 		tags := map[string]string{
 			"region": c.Region,
-			"unit":   snakeCase(*point.Unit),
 		}
 
-		for _, d := range metric.Dimensions {
-			tags[snakeCase(*d.Name)] = *d.Value
+		if dimensions, ok := c.queryDimensions[*result.Id]; ok {
```
danielnelson: Consider having queryDimensions and calling it tags, just be a map[string]string so this can be skipped.

danielnelson:
@druchoo I marked this PR to close #3255 issue when merged.

CpuID:
🎉

PR #5547, opened 2019-03-07
Add file rotation based on file age to file output plugin

flexd:
### Required for all PRs:

I have not touched the unit tests at all (yet)

This replaces #3922


flexd:
Err, there we go. Accidentally put in a older file_test.go and broke Write() a bit. This compiles for me, and appears to be working like it should. FYI: It's next to impossible to compile Telegraf with only 2GB ram :-(

glinton:
Is there any way you can push these changes to the 3922 to preserve the history? Or is this a fundamentally different approach?

flexd:
@glinton uh, I could in theory do that I guess? This is basically `rotate.go` from the other pull-request with a number of changes and then I added the rotate_max_age field to `file.go`, in the file plugin.

The other pull-request #3922 adds a whole new plugin `rotatingfile` that I would have to remove again, with a bunch of reverting commits or a force-push. Also I have deleted the local branch for that, but I guess it would be possible to get that back by pulling it from the pull-request.

Isn't history preserved enough by having the first pull-request saying this one takes over?

flexd:
Does the thumbs up mean everything is good? :-)

danielnelson:
@flexd We added rotation based on this pull request to the agent logfile, I think we can update this pull request to use the rotating io.Writer now in https://github.com/influxdata/telegraf/tree/master/internal/rotate.

I'd like to include this in the 1.11 release, would you have time to update the pull request?

flexd:
That sounds reasonable! I will make the required changes :-)

flexd:
Uh, that should do it! I think I might have messed up the rebase there with all these commits showing up :-/ Any suggestions?

danielnelson:
I think you will need to do an interactive rebase `git rebase -i master` and then remove the unrelated commits keeping only your new changes.

flexd:
That looks like it fixed it. :-)

plugins/outputs/file/README.md
```diff
@@ -8,6 +8,9 @@ This plugin writes telegraf metrics to files
   ## Files to write to, "stdout" is a specially handled file.
   files = ["stdout", "/tmp/metrics.out"]
 
+  ## If this is defined, files will be rotated by the time.Duration specified
+  #rotate_max_age = "1m"
```
glinton: nit: add a space after the `#`

plugins/outputs/file/file.go
```diff
@@ -23,6 +26,9 @@ var sampleConfig = `
   ## Files to write to, "stdout" is a specially handled file.
   files = ["stdout", "/tmp/metrics.out"]
 
+  ## If this is defined, files will be rotated by the time.Duration specified
+  #rotate_max_age = "1m"
```
glinton: same

plugins/outputs/file/file.go
```diff
@@ -76,19 +97,19 @@ func (f *File) Description() string {
...
 			return fmt.Errorf("failed to serialize message: %s", err)
 		}
 
-		for _, writer := range f.writers {
-			_, err = writer.Write(b)
-			if err != nil && writer != os.Stdout {
-				writeErr = fmt.Errorf("E! failed to write message: %s, %s", b, err)
-			}
+		_, err = f.writer.Write(b)
+		if err != nil && f.writer != os.Stdout {
```
glinton: You can remove the second half of the check, as it will always be `true`

`fmt.Println(io.MultiWriter([]io.Writer{os.Stdout}...) == os.Stdout)`

flexd:
That last check was leftover from before the io.MultiWriter was used I think :-) Used to be a slice of writers

PR #5566, opened 2019-03-11
[inputs/diskio] Fix how major and minor identifiers of block devices are read.

Soulou:
The current implementation assure that the major and the minor are
coded on one byte. But they are not:

```
brw-rw----  1 root disk    252, 290 Feb 25 11:36 dm-290
```

290 as minor in this example is over 1 byte.

So after wondering why all my devices iops weren't correctly stored,
I found out that several points were added for some disks. For `dm-290`
it was overriding `252:34`, instead of getting udev stats for `252:290`.

The solution is here:
https://sites.uclouvain.be/SystInfo/usr/include/sys/sysmacros.h.html

The implementation is directly taken from this, fixing my bug.

### DIFF
--- a/plugins/inputs/diskio/diskio_linux.go
+++ b/plugins/inputs/diskio/diskio_linux.go
@@ -35,8 +35,8 @@ func (s *DiskIO) diskInfo(devName string) (map[string]string, error) {
 		return ic.values, nil
 	}
 
-	major := unix.Major(stat.Rdev)
-	minor := unix.Minor(stat.Rdev)
+	major := unix.Major(uint64(stat.Rdev))
+	minor := unix.Minor(uint64(stat.Rdev))
 	udevDataPath := fmt.Sprintf("%s/b%d:%d", udevPath, major, minor)
 
 	di := map[string]string{}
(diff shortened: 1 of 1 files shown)


danielnelson:
Could we use these functions to extract the major/minor numbers?

https://github.com/golang/sys/blob/c8c8c57fd1e1a977c10821a559dda983a9a9a3b5/unix/dev_linux.go#L21-L32

Soulou:
Definitely, I'm doing it now

Soulou:
done

danielnelson:
@Soulou could you look into this build error on mipsle?
```
[ERROR] run: Command 'GOOS=linux GOARCH=mipsle go build -o /go/src/github.com/influxdata/telegraf/build/linux/mipsel/telegraf  -ldflags="-w -s -X main.branch=pull/5566 -X main.commit=2118fbe7" ./cmd/telegraf' failed with error: # github.com/influxdata/telegraf/plugins/inputs/diskio
plugins/inputs/diskio/diskio_linux.go:38:26: cannot use stat.Rdev (type uint32) as type uint64 in argument to unix.Major
plugins/inputs/diskio/diskio_linux.go:39:26: cannot use stat.Rdev (type uint32) as type uint64 in argument to unix.Minor
```

Soulou:
done :)

PR #5569, opened 2019-03-11
Add conversion between a string hexadecimal value and integer

Prygan:
### Required for all PRs:

This commit add a conversion between a string hexadecimal (e.g: '0x2A') and an integer (e.g: 42). It's some change I needed because, for a project, I need to consume metrics from an HTTP endpoint. Those metrics are in hexadecimal.

I did not modified the README yet (not sure I had to mention this new case). I can do it if needed.

If someone sees how to make this commit "cleaner" in term of how it's coded, feel free to give your advices ! :smiley:


plugins/processors/converter/converter.go
```diff
@@ -345,11 +346,22 @@ func toInteger(v interface{}) (int64, bool) {
 			return 0, true
 		}
 	case string:
-		result, err := strconv.ParseFloat(value, 64)
-		if err != nil {
-			return 0, false
+		if strings.HasPrefix(value, "0x") {
```
danielnelson: Instead of this check, just try ParseInt and then fallback to ParseFloat if it fails, this will give us octal support and allow `0XFF` to work as well.  Also, can you add a similar conversion for the unsigned type as well?

Prygan: Hey @danielnelson , thanks for you advices. That's now done (refactoring + string hexa to uint converter). Unfortunately, there is one issue in the CI. CircleCI is OK but Appveyor failed. I may be wrong but I think it's not because of my code (the error is `failed to write dep tree: failed to export golang.org/x/crypto: 
	(1) failed to list versions for https://go.googlesource.com/crypto: fatal: remote error: Internal Server` :thinking: )

danielnelson: Sorry about that, just having some issues lately with our builds on appveyor, I'm sure its unrelated.

PR #5572, opened 2019-03-12
Floats in statsd percentiles (rebased with master)

Pitxyoki:
### Required for all PRs:

This addresses #5527. Additionally to the previous PR (#5528), here I'm trying not to do any breaking change:
* those that used the plugin previously with the percentile array configured with only integer numbers may continue to do so (e.g.: `[50, 90, 100]`);
* it is now possible to also configure the same array with only floats (e.g.: `[50.0, 90.0, 99.0, 99.9, 100.0]`).
* it is not possible to configure it with a mix of both (e.g.: `[50, 90, 99, 99.9, 100]`), which I think is reasonable.

This is now targetted at the current `master` branch.
The already-present Unit Tests were only slightly changed to accept the correct type. No new tests were added.


Pitxyoki:
The circleci tests seem to have failed due to errors in quay.io. Can we re-trigger them?

Pitxyoki:
By the way: what is the process for having this merged into the next minor release for, say, 1.10 and 1.9? I'd like to see this in at least those versions, as those are the current and the likely next versions I'm using.
Should I create different pull requests against the respective branch of each of those releases or do you have some other process for that? 

Thanks!

glinton:
We generally only backport bugfixes, and only to the latest release (1.10 as of now).  Features are added to the next minor release (1.11 as of now).

Pitxyoki:
Hi,
Do you need anything else needed before this can be merged onto master? I'd like to be sure this will get released on upcoming versions of telegraf.

Thank you.

danielnelson:
Can you add a unit test that exercises a non-even floating point value, such as the 99.5th percentile?

Pitxyoki:
> Can you add a unit test that exercises a non-even floating point value, such as the 99.5th percentile?

@danielnelson: Added, along with other marginal cases.

PR #5573, opened 2019-03-12
Allow only a config directory to be used

### LINKED ISSUE #5571: Telegraf requires config even when using a config-directory
_From @bmagistro on April 26, 2018 17:44_

I was referencing #486 for information about using telegraf.d, specifically 

> Number (1) is probably a bug, if there is a valid file in telegraf.d we should be able to start Telegraf without specifying a singular .conf file.

While this setup may be "non-standard", we have shifted all configuration items to be loaded from files in telegraf.d.  When initially tried, I had omitted /etc/telegraf/telegraf.conf and that will cause telegraf to fail to start.  To solve this, adding an empty telegraf.conf file allows everything to work and all config files to live in telegraf.d.

From looking at this further the source comes from the telegraf.service file specifying --config and --config-dir.

_Copied from original issue: influxdata/docs.influxdata.com#1589_

glinton:
No longer require a config file to be specified/found if only a config
directory is used.

Resolves #5571

### DIFF
--- a/agent/agent_test.go
+++ b/agent/agent_test.go
@@ -24,35 +24,35 @@ func TestAgent_OmitHostname(t *testing.T) {
 func TestAgent_LoadPlugin(t *testing.T) {
 	c := config.NewConfig()
 	c.InputFilters = []string{"mysql"}
-	err := c.LoadConfig("../internal/config/testdata/telegraf-agent.toml")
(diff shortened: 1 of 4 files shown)


glinton:
Closing in favor of #5703

PR #5585, opened 2019-03-14
Add ability to filter printed headers in `telegraf config`

### LINKED ISSUE #5584: Please change the behaviour of config generation
## Feature Request
The file ```/etc/telegraf/telegraf.conf``` contains more than 5,000 lines and is therefore not maintainable after a minor or major release. 

### Current behavior:
The command ```telegraf --input-filter cpu --output-filter influxdb config``` creates a configuration containing the selected input/output plugins. 

### Desired behavior:
The command ```telegraf --input-filter cpu config``` should only create the configuration for the selected input plugin. 

### Use case: 
I want to use the command ```telegraf --input-filter cpu config``` to store the configuration in the file ```/etc/telegraf/telegraf.d/cpu.conf```. Then I am able to manipulate the plugin parameters with a config management tool like Puppet.

glinton:
Resolves #5584 

This adds `header-filter` as an option to filter the printing of the `global_tags` and `agent` sections of the config file. By setting to "none", config headers will be skipped, allowing the creation of minimal config files.

```
$ telegraf \
  --output-filter none \
  --processor-filter none \
  --aggregator-filter none \
  --input-filter cpu \
  --header-filter none config

[[inputs.cpu]]
  percpu = true
  totalcpu = true
  collect_cpu_time = false
  ## If true, compute and report the sum of all non-idle CPU states.
  report_active = false

```


danielnelson:
Would it be possible to call this `--section-filter` and allow it to skip the plugin sections too?  So if you have `telegraf --section-filter agent` it only prints the agent section, and if you have `telegraf --section-filter input` it doesn't show the 3 line comment header for the other plugin types?

glinton:
That would work, though I do like the ability to remove all unneeded comments in a generated file without piping to a text processor.

danielnelson:
I think we should have this so that this command only prints the `[agent]` section, no inputs/outputs etc.
```
./telegraf --section-filter agent config
```

Also should support sections `inputs`, `outputs`, `processors`, `outputs`:
```
./telegraf --section-filter inputs:processors config
```

Only show the header for a section when the section is enabled.

glinton:
Ok, now I see what you're saying. I understood the section-filter was to only control the other (currently non-configurable) sections agent, and global-tags. To just show the agent section, with changes up to now on this branch, you would need to specify:
```
./telegraf --section-filter agent  --input-filter none --output-filter none  --aggregator-filter none --processor-filter none config
```
Which is a lot of configuration. Making the change now.

internal/config/config.go
```diff
@@ -32,6 +32,10 @@ import (
 )
 
 var (
+	// Default sections
+	sectionDefaults = []string{"agent", "global_tags", "output",
+		"processor", "aggregator", "input"}
```
danielnelson: I think maybe we should switch the plugin sections to be plural: `--section-filter inputs:outputs:processors:aggregators`.

PR #5587, opened 2019-03-15
Input Plugin: GitHub

rawkode:
### Required for all PRs:

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -520,6 +520,22 @@
   version = "v0.2.0"
 
 [[projects]]
+  digest = "1:e38ad2825940d58bd8425be40bcd4211099d0c1988c158c35828197413b3cf85"
+  name = "github.com/google/go-github"
+  packages = ["github"]
+  pruneopts = ""
+  revision = "7462feb2032c2da9e3b85e9b04e6853a6e9e14ca"
+  version = "v24.0.1"
+
+[[projects]]
+  digest = "1:cea4aa2038169ee558bf507d5ea02c94ca85bcca28a4c7bb99fd59b31e43a686"
+  name = "github.com/google/go-querystring"
+  packages = ["query"]
+  pruneopts = ""
+  revision = "44c6ddd0a2342c386950e880b658017258da92fc"
+  version = "v1.0.0"
+
+[[projects]]
   digest = "1:c1d7e883c50a26ea34019320d8ae40fad86c9e5d56e63a1ba2cb618cef43e986"
   name = "github.com/google/uuid"
   packages = ["."]
@@ -1556,6 +1572,7 @@
     "github.com/golang/protobuf/ptypes/empty",
     "github.com/golang/protobuf/ptypes/timestamp",
     "github.com/google/go-cmp/cmp",
+    "github.com/google/go-github/github",
     "github.com/gorilla/mux",
     "github.com/harlow/kinesis-consumer",
     "github.com/harlow/kinesis-consumer/checkpoint/ddb",
--- a/plugins/inputs/github/README.md
+++ b/plugins/inputs/github/README.md
@@ -19,9 +19,6 @@ The [GitHub](https://www.github.com) input plugin gathers statistics from GitHub
 
 ### Metrics:
 
-For more details about collected metrics reference the [HAProxy CSV format
-documentation](https://cbonte.github.io/haproxy-dconv/1.8/management.html#9.1).
-
 - github_repository
   - tags:
     - `full_name` - The full name of the repository, including owner/organization
--- a/plugins/inputs/github/README.md
+++ b/plugins/inputs/github/README.md
@@ -11,7 +11,7 @@ The [GitHub](https://www.github.com) input plugin gathers statistics from GitHub
     ]
 
     ## Optional: Unauthenticated requests are limited to 60 per hour
-    api_key = "SOME_KEY"
+    access_token = "personal access token"
 
     ## Optional: Default 5s
     http_timeout = "5s"
--- a/plugins/inputs/github/README.md
+++ b/plugins/inputs/github/README.md
@@ -41,6 +41,6 @@ The [GitHub](https://www.github.com) input plugin gathers statistics from GitHub
 ### Example Output:
 
 ```
-github,full_name=influxdata/telegraf,language=Go,license=MIT\ License,name=telegraf,owner=influxdata stars=6401i,forks=2421i,open_issues=722i,size=22611i 1552651811000000000
-github_rate_limit,host=p4x-639.lan remaining=59i,limit=60i 1552653551000000000
+github,full_name=influxdata/telegraf,name=telegraf,owner=influxdata,language=Go,license=MIT\ License stars=6401i,forks=2421i,open_issues=722i,size=22611i 1552651811000000000
+github_rate_limit, remaining=59i,limit=60i 1552653551000000000
 ```
--- a/plugins/inputs/github/README.md
+++ b/plugins/inputs/github/README.md
@@ -21,7 +21,6 @@ The [GitHub](https://www.github.com) input plugin gathers statistics from GitHub
 
 - github_repository
   - tags:
-    - `full_name` - The full name of the repository, including owner/organization
     - `name` - The repository name
     - `owner` - The owner of the repository
     - `language` - The primary language of the repository
--- a/plugins/inputs/github/README.md
+++ b/plugins/inputs/github/README.md
@@ -44,5 +44,5 @@ The [GitHub](https://www.github.com) input plugin gathers statistics from GitHub
 
 ```
 github,full_name=influxdata/telegraf,name=telegraf,owner=influxdata,language=Go,license=MIT\ License stars=6401i,forks=2421i,open_issues=722i,size=22611i 1552651811000000000
-github_rate_limit,access_token=Unauthenticated remaining=59i,limit=60i,blocks=0i 1552653551000000000
+internal_github,access_token=Unauthenticated rate_limit_remaining=59i,rate_limit_limit=60i,rate_limit_blocks=0i 1552653551000000000
 ```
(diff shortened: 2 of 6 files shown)


rawkode:
Still need to:

- Tests
- API Token Support
- Wildcard Support (influxdata/*)

plugins/inputs/github/README.md
```diff
@@ -0,0 +1,45 @@
...
+    - `name` - The repository name
+    - `owner` - The owner of the repository
+    - `language` - The primary language of the repository
+    - `license` - The license set for the repository
+  - fields:
+    - `stars` (int)
+    - `forks` (int)
+    - `open_issues` (int)
+    - `size` (int)
+- github_rate_limit
```
danielnelson: This meta metric would best be reported as an `internal` (aka `selfstat`) metric.  These are only emitted if the `internal` plugin is enabled, check the statsd plugin for an example.

plugins/inputs/github/README.md
```diff
@@ -0,0 +1,45 @@
...
+### Configuration:
+
+```toml
+[[inputs.github]]
+    repositories = [
+      "owner/repository",
+    ]
+
+    ## Optional: Unauthenticated requests are limited to 60 per hour
+    access_token = "personal access token"
```
danielnelson: We have some information on [SampleConfig](https://github.com/influxdata/telegraf/wiki/SampleConfig) style on the wiki.  Use a 2 space indent and comment out any optional settings showing the default value.  No need to list them as optional or specify the default in the comment.

plugins/inputs/github/README.md
```diff
@@ -0,0 +1,45 @@
+# GitHub Input Plugin
+
+The [GitHub](https://www.github.com) input plugin gathers statistics from GitHub repositories.
+
+### Configuration:
+
+```toml
+[[inputs.github]]
+    repositories = [
+      "owner/repository",
```
danielnelson: `influxdata/telegraf` would be a good default.  Super nit: just do it as a single line `repositories = ["influxdata/telegraf"]`

plugins/inputs/github/README.md
```diff
@@ -0,0 +1,45 @@
...
+  - tags:
+  - fields:
+    - `limit` - How many requests you are limited to (per hour)
+    - `remaining` - How many requests you have remaining (per hour)
+
+### Example Output:
+
+```
+github,full_name=influxdata/telegraf,language=Go,license=MIT\ License,name=telegraf,owner=influxdata stars=6401i,forks=2421i,open_issues=722i,size=22611i 1552651811000000000
+github_rate_limit,host=p4x-639.lan remaining=59i,limit=60i 1552653551000000000
```
danielnelson: Make sure this matches the listing above.

plugins/inputs/github/README.md
danielnelson: Github pro tip: to avoid weird vertical spacing alternate the style of the first list char:
```
- foo
  - bar
+ xyzzy
  - baz
```

Here is an example of how it looks if you **don't** do this: https://github.com/influxdata/telegraf/tree/master/plugins/inputs/dcos#metrics

Here is where we did it right:
https://github.com/influxdata/telegraf/tree/master/plugins/inputs/kube_inventory#metrics

plugins/inputs/github/github.go
```diff
@@ -0,0 +1,168 @@
...
+
+  ## API Key for GitHub API requests
+  api_key = ""
+
+  ## Timeout for GitHub API requests
+  http_timeout = "5s"
+`
+
+// NewGitHub returns a new instance of the GitHub input plugin
+func NewGitHub() *GitHub {
```
danielnelson: Doesn't look like this is used.  I would probably not use it and set the HTTPTimeout in the `init()` function.

plugins/inputs/github/github.go
```diff
@@ -0,0 +1,168 @@
...
+	var githubClient *gh.Client
+
+	if github.AccessToken != "" {
+		ctx := context.Background()
+		ts := oauth2.StaticTokenSource(
+			&oauth2.Token{AccessToken: github.AccessToken},
+		)
+		tc := oauth2.NewClient(ctx, ts)
+
+		githubClient = gh.NewClient(tc)
```
danielnelson: We will want to assign a http.Client here, to avoid using the default client/transport.  This will also give you a way to use your http_timeout option.  Check the `http` output plugin for an example of doing this with the oauth2 package.

plugins/inputs/github/github.go
```diff
@@ -0,0 +1,168 @@
...
+	return sampleConfig
+}
+
+// Description returns the plugin description.
+func (github *GitHub) Description() string {
+	return "Read repository information from GitHub, including forks, stars, and more."
+}
+
+// Create HTTP Client
+func (github *GitHub) createGitHubClient() (*gh.Client, error) {
```
danielnelson: Go style is generally to use a single letter for the function receiver, I would probably do this and as a bonus you won't have to alias the github package.

Gopkg.lock
```diff
@@ -519,6 +519,22 @@
   revision = "3af367b6b30c263d47e8895973edcca9a49cf029"
   version = "v0.2.0"
 
+[[projects]]
+  digest = "1:e38ad2825940d58bd8425be40bcd4211099d0c1988c158c35828197413b3cf85"
+  name = "github.com/google/go-github"
```
danielnelson: Just an observation for now: this is a fairly big library, moving us from 60MB -> 62MB to the Telegraf binary, on the large side based on what we are doing here.

plugins/inputs/github/github.go
```diff
@@ -0,0 +1,166 @@
...
+
+			ctx := context.Background()
+
+			owner, repository, err := splitRepositoryName(repositoryName)
+			if err != nil {
+				log.Printf("E! [github]: %v", err)
+				return
+			}
+
+			repositoryInfo, response, err := g.githubClient.Repositories.Get(ctx, owner, repository)
```
goller: I think you may want to do something if err is not a rate limit error.

plugins/inputs/github/github.go
```diff
@@ -0,0 +1,166 @@
...
+
+			if _, ok := err.(*github.RateLimitError); ok {
+				log.Printf("E! [github]: %v of %v requests remaining", response.Rate.Remaining, response.Rate.Limit)
```
goller: Perhaps still record the rate limiting metrics when the error occurs?

rawkode:
@danielnelson @goller Hopefully I've addressed all the issues 👍

rawkode:
I had hoped that by resolving the issues I could re-request a review, but the button still isn't showing 🤷‍♂️

plugins/inputs/github/README.md
```diff
@@ -0,0 +1,48 @@
...
+  - fields:
+    - `limit` - How many requests you are limited to (per hour)
+    - `remaining` - How many requests you have remaining (per hour)
+    - `blocks` - How many requests have been blocked due to rate limit
+
+### Example Output:
+
+```
+github,full_name=influxdata/telegraf,name=telegraf,owner=influxdata,language=Go,license=MIT\ License stars=6401i,forks=2421i,open_issues=722i,size=22611i 1552651811000000000
+github_rate_limit,access_token=Unauthenticated remaining=59i,limit=60i,blocks=0i 1552653551000000000
```
danielnelson: Update for `internal` metric style.

I would use the repo name and owner as tags, this will give you number of blocks per repo and sidesteps the access_token encoding.

plugins/inputs/github/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+  ## Optional: Default 5s.
+  # http_timeout = "5s"
+```
+
+### Metrics:
+
+- github_repository
+  - tags:
+    - `full_name` - The full name of the repository, including owner/organization
```
danielnelson: I think we should remove this since we are reporting the owner and name.

plugins/inputs/github/github.go
```diff
@@ -0,0 +1,185 @@
...
+			}
+
+			tokenTags := map[string]string{
+				"access_token": g.obfusticatedToken,
+			}
+
+			repositoryInfo, response, err := g.githubClient.Repositories.Get(ctx, owner, repository)
+
+			if _, ok := err.(*github.RateLimitError); ok {
+				g.RateLimitErrors = selfstat.Register("github", "rate_limit_blocks", tokenTags)
```
danielnelson: Register all the selfstat only once, maybe up where you init the client.

plugins/inputs/github/github_test.go
```diff
@@ -0,0 +1,109 @@
...
+
+	ownerName := "influxdata"
+	owner := gh.User{Login: &ownerName}
+
+	fullName := "influxdata/influxdb"
+	repositoryName := "influxdb"
+
+	language := "Go"
+
+	repository := gh.Repository{FullName: &fullName, Name: &repositoryName, License: &license, Owner: &owner, Language: &language}
```
danielnelson: Nitpick, wrap this line and in next test.

plugins/inputs/github/github_test.go
```diff
@@ -0,0 +1,109 @@
...
+
+	correctTagsReturn := map[string]string{
+		"full_name": fullName,
+		"owner":     ownerName,
+		"name":      repositoryName,
+		"language":  language,
+		"license":   licenseName,
+	}
+
+	require.True(t, reflect.DeepEqual(getTagsReturn, correctTagsReturn))
```
danielnelson:
```
require.Equal(t, expected, actual)
```

plugins/inputs/github/README.md
rawkode: I've updated the internal metric example.

I don't think we should use the repo name and owner as tags, as the block is associated with the access token, the repo is irrelevant.

danielnelson: I think there are a couple arguments for repo:
- some users may be uncomfortable with access_token encoding, even though it is limited.
- there isn't always an access_token, requiring a special value.  this isn't necessarily bad but it is a code smell to me.
- it may be interesting to know which repository was blocked during a time period

I'll leave the final decision for you, I'm okay with how it is as well.

rawkode: I can see both sides, but as the block is directly linked to the access token; I'm keen to leave it as is and maybe we can collect some user feedback.

Happy for this to be merged as is. Thanks for your feedback and help during this, appreciated.

PR #5596, opened 2019-03-17
Rename testdata dir with double dots

keitwb:
Unfortunately due to a security check in Go modules (as of the latest
release), the existance of '..' in a file path in a repository causes
the module to be rejected when adding it with `go get`:
https://github.com/golang/go/issues/28001#issuecomment-443309655.

This renames the testdata directory with the double dots and augments
the test code to rename it on the fly back to having two dots to get the
appropriate test behavior. This means the tests cannot run on a
read-only filesystem but hopefully that is not a requirement.

### Required for all PRs:


danielnelson:
@keitwb Thanks for the heads up on this, I think I will just remove this directory for now rather than workaround with rename.  eeb0e094c291049aac22c577800fff49663b46c7

PR #5630, opened 2019-03-26
Fix basedir check and parent dir extraction in filecount input

### LINKED ISSUE #5621: Filecount not working on Windows
### Relevant telegraf.conf:
```
[[inputs.filecount]]
  ## Directory to gather stats about.
  ##   deprecated in 1.9; use the directories option
  # directory = "/var/cache/apt/archives"

  ## Directories to gather stats about.
  ## This accept standard unit glob matching rules, but with the addition of
  ##   /var/log/**    -> recursively find all directories in /var/log and count files in each directories
  ##   /var/log/*/*   -> find all directories with a parent dir in /var/log and count files in each directories
  ##   /var/log       -> count all files in /var/log and all of its subdirectories
  directories = ["c:/temp"]

  ## Only count files that match the name pattern. Defaults to "*".
  name = "*"

  ## Count files in subdirectories. Defaults to true.
  recursive = false

  ## Only count regular files. Defaults to true.
  regular_only = true

  ## Only count files that are at least this size. If size is
  ## a negative number, only count files that are smaller than the
  ## absolute value of size. Acceptable units are B, KiB, MiB, KB, ...
  ## Without quotes and units, interpreted as size in bytes.
  size = "0B"

  ## Only count files that have not been touched for at least this
  ## duration. If mtime is negative, only count files that have been
  ## touched in this duration. Defaults to "0s".
  mtime = "0s"
```

### System info:

Telegraf 1.10.0~rc1 (git: HEAD 2d2abe29)
Windows 10 x64

### Steps to reproduce:

1. Include filecount plugin config with any Windows path specified
2. Run telegraf --debug --config <your config> --test

### Expected behavior:
Should see filecount counters in output

### Actual behavior:
Filecount counters never listed.

### Additional info:

Telegraf reports in log that filecount plugin was detected, just no output is ever shown.  Have tried several folders all containing files.

danielnelson:
Not just a Windows bug, previously we would panic on paths without a `/`, and it was possible to not detect the top directory if it was aliased with a different path.

closes #5621

### Required for all PRs:

### DIFF
--- a/plugins/inputs/filecount/filecount.go
+++ b/plugins/inputs/filecount/filecount.go
@@ -4,16 +4,14 @@ import (
 	"log"
 	"os"
 	"path/filepath"
-	"strings"
 	"time"
-
-	"github.com/karrick/godirwalk"
-	"github.com/pkg/errors"
 
 	"github.com/influxdata/telegraf"
 	"github.com/influxdata/telegraf/internal"
 	"github.com/influxdata/telegraf/internal/globpath"
 	"github.com/influxdata/telegraf/plugins/inputs"
+	"github.com/karrick/godirwalk"
+	"github.com/pkg/errors"
 )
 
 const sampleConfig = `
@@ -157,7 +155,8 @@ func (fc *FileCount) count(acc telegraf.Accumulator, basedir string, glob globpa
 	childSize := make(map[string]int64)
 
 	walkFn := func(path string, de *godirwalk.Dirent) error {
-		if path == basedir {
+		rel, err := filepath.Rel(basedir, path)
+		if err == nil && rel == "." {
 			return nil
 		}
 		file, err := os.Stat(path)
@@ -173,7 +172,7 @@ func (fc *FileCount) count(acc telegraf.Accumulator, basedir string, glob globpa
 			return nil
 		}
 		if match {
-			parent := path[:strings.LastIndex(path, "/")]
+			parent := filepath.Dir(path)
 			childCount[parent]++
 			childSize[parent] += file.Size()
 		}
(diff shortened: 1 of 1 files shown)


SimonMcN:
This does not work for me on windows, was the fix tested on windows, or just linux please ?
2019-07-02T13:32:03Z I! Loaded inputs: filecount win_services win_perf_counters
2019-07-02T13:32:03Z I! Loaded aggregators: 
2019-07-02T13:32:03Z I! Loaded processors: 
2019-07-02T13:32:03Z I! Loaded outputs: influxdb
2019-07-02T13:32:03Z I! Tags enabled: ENV=PROD host=DLS07-P
2019-07-02T13:32:03Z I! [agent] Config: Interval:10s, Quiet:false, Hostname:"DLS07-P", Flush Interval:10s
2019-07-02T13:32:58Z I! [agent] Hang on, flushing any cached metrics before shutdown

tested with:
C:\telegraf\telegraf.exe --config c:\telegraf\prod.conf --config-directory c:\telegraf\conf.d --test --debug

[[inputs.filecount]]
  directories = ["E:/mydirectory/processed/Failed"]
  name = "*"
  recursive = true
  regular_only = true
  size = "0B"
  mtime = "0s"

SimonMcN:
telegraf.exe V1.11.0

danielnelson:
@SimonMcN Most likely I only tested on Linux, I'll reopen the issue and try it out when I have a Windows system available.

SimonMcN:
I'd appreciate it.  I know the Linux and windows filesystems have different mount point syntax so it would be easy to break it by mistake..... Either that or I am incapable 🤔😐😋

danielnelson:
Looks to be related to the path separators, I believe using backslashes will workaround this issue until the fix is released.  You can either escape the backslashes `\\` or use single quote "literal" strings:
```toml
directories = ['E:\mydirectory\processed\Failed']

danielnelson:
Fix for the separator issue is here: https://github.com/influxdata/telegraf/pull/6077

PR #5641, opened 2019-03-27
[aggregators/histogram] Add option to reset buckets on flush

oplehto:
### Required for all PRs:

For some use cases it is desireable for the histogram buckets to be reset between operations. 

A good example is collecting data from processes using the PID as a tag where the cardinality may become very high and series can become "stale". With the current implementation, series for already dead processes will still continue to be reported on. 

This functionality could be further improved by adding a reset interval and per-subconfig reset options but I expect that for most use cases just having everything reset on flush is a simple and effective solution. 

@vlamug It would be good to hear your thoughts on this

### DIFF
--- a/plugins/aggregators/histogram/histogram.go
+++ b/plugins/aggregators/histogram/histogram.go
@@ -73,6 +73,10 @@ var sampleConfig = `
   ## aggregator and will not get sent to the output plugins.
   drop_original = false
 
+  ## If true, the histogram will be reset on flush instead
+  ## of accumulating the results.
+  reset = false
+
   ## Example config that aggregates all fields of the metric.
   # [[aggregators.histogram.config]]
   #   ## The set of buckets.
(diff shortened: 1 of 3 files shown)


plugins/aggregators/histogram/histogram.go
```diff
@@ -201,9 +206,16 @@ func (h *HistogramAggregator) groupField(
 	)
 }
 
-// Reset does nothing, because we need to collect counts for a long time, otherwise if config parameter 'reset' has
-// small value, we will get a histogram with a small amount of the distribution.
-func (h *HistogramAggregator) Reset() {}
+// Reset does nothing by default, because we need to collect counts for a long time,
+// otherwise if config parameter 'reset' has small value, we will get a histogram
+// with a small amount of the distribution. However in some use cases a reset is useful.
```
vlamug: `if config parameter 'reset' has 'false' value,`

Also delete empty line below please.

PR #5645, opened 2019-03-28
Add support for time precision when using influxdb as output

StephenSorriaux:
This allows to set the time precision (s, ms, us, ns) when sending metrics to InfluxDB. Previously, only ns was possible.

### Required for all PRs:

### DIFF
--- a/plugins/outputs/influxdb/influxdb.go
+++ b/plugins/outputs/influxdb/influxdb.go
@@ -48,6 +48,7 @@ type InfluxDB struct {
 	SkipDatabaseCreation bool              `toml:"skip_database_creation"`
 	InfluxUintSupport    bool              `toml:"influx_uint_support"`
 	tls.ClientConfig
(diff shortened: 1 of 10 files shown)


plugins/outputs/influxdb/influxdb.go
```diff
@@ -48,6 +48,7 @@ type InfluxDB struct {
 	SkipDatabaseCreation bool              `toml:"skip_database_creation"`
 	InfluxUintSupport    bool              `toml:"influx_uint_support"`
 	tls.ClientConfig
+	TimePrecision string `toml:"time_precision"`
 
 	Precision string // precision deprecated in 1.0; value is ignored
```
glinton: I'm not sure the reason, but I believe this functionality was deprecated intentionally.

danielnelson:
You can use the agent `precision` to round values instead.  We actually used to have this functionality but it was removed #1368 because it was no longer needed.

PR #5649, opened 2019-03-29
Add additional examples for SNMPv3

js-mode:
There are a lot of examples out there for SNMP version 1/2c but more and more people are using V3.   Think it may be helpful for others looking for documentation to see snmp examples against Cisco switches that enforce snmpv3.

### Required for all PRs:

### DIFF
--- a/plugins/inputs/snmp/CONFIG-EXAMPLES.md
+++ b/plugins/inputs/snmp/CONFIG-EXAMPLES.md
@@ -63,3 +63,69 @@ Here is the configuration that you add to your `telegraf.conf`:
       oid = "IF-MIB::ifDescr"
       is_tag = true
 ```
+
+
+See below for SNMPv3 Examples of configs against Cisco Switches: 
+
+```
+[[inputs.snmp]]
+  agents = [AGENTS]
+  version = 3
+  auth_password = "TOKEN"
+  auth_protocol = "MD5"
+  priv_password = "TOKEN"
+  priv_protocol = "AES"
+  sec_name = "test"
+  sec_level = "authPriv"
+  max_repetitions = 10
+  timeout = "10s"
+  retries = 1
+  name_prefix = "dev."
+  
+  [[inputs.snmp.field]]
+    name = "hostname"
+    oid = ".1.3.6.1.2.1.1.5.0"
+    is_tag = true
+  [inputs.snmp.tagdrop]
+    ifName=["Vlan[0-9]*", "loopback*"]
+    fieldpass = ["ifInUcastPkts", "ifInNUcastPkts", "ifInDiscards", "ifInErrors", "ifOutUcastPkts", "ifOutNUcastPkts", "ifOutDiscards", "ifOutErrors", "ifType", "ifMtu", "ifSpeed", "ifAdminStatus", "ifOperStatus"]
+  [[inputs.snmp.table]]
+    name = "interfaceStats"
+    inherit_tags = ["hostname"]
+  [inputs.snmp.tagdrop]
+    ifAlias = [""]
+    [[inputs.snmp.table.field]]
+      name = "ifAlias"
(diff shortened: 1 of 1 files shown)


plugins/inputs/snmp/CONFIG-EXAMPLES.md
```diff
@@ -63,3 +63,69 @@ Here is the configuration that you add to your `telegraf.conf`:
...
+    name = "hostname"
+    oid = ".1.3.6.1.2.1.1.5.0"
+    is_tag = true
+  [inputs.snmp.tagdrop]
+    ifName=["Vlan[0-9]*", "loopback*"]
+    fieldpass = ["ifInUcastPkts", "ifInNUcastPkts", "ifInDiscards", "ifInErrors", "ifOutUcastPkts", "ifOutNUcastPkts", "ifOutDiscards", "ifOutErrors", "ifType", "ifMtu", "ifSpeed", "ifAdminStatus", "ifOperStatus"]
+  [[inputs.snmp.table]]
+    name = "interfaceStats"
+    inherit_tags = ["hostname"]
+  [inputs.snmp.tagdrop]
```
danielnelson: This needs to be combined with the `[inputs.snmp.tagdrop]` item above, a plugin can only have one tagdrop table.    I'm somewhat surprised this isn't a syntax error.

plugins/inputs/snmp/CONFIG-EXAMPLES.md
```diff
@@ -63,3 +63,69 @@ Here is the configuration that you add to your `telegraf.conf`:
...
+  retries = 1
+  name_prefix = "dev."
+  
+  [[inputs.snmp.field]]
+    name = "hostname"
+    oid = ".1.3.6.1.2.1.1.5.0"
+    is_tag = true
+  [inputs.snmp.tagdrop]
+    ifName=["Vlan[0-9]*", "loopback*"]
+    fieldpass = ["ifInUcastPkts", "ifInNUcastPkts", "ifInDiscards", "ifInErrors", "ifOutUcastPkts", "ifOutNUcastPkts", "ifOutDiscards", "ifOutErrors", "ifType", "ifMtu", "ifSpeed", "ifAdminStatus", "ifOperStatus"]
```
danielnelson: A key "belongs" to the previously defined table, which means this falls under the tagdrop table.  For it to work as expected it needs to belong to the plugin table `[[inputs.snmp]]`.

danielnelson:
Thanks but I don't think we will add this to the docs since it is so similar to the existing example in functionality (getting network interface stats).  The settings for SNMPv3 can vary quite a bit too so I'm not sure how applicable these ones are for others.

plugins/inputs/snmp/CONFIG-EXAMPLES.md
js-mode: Not totally sure what you mean?  I realize that there was an extra [inputs.snmp.tagdrop] on line 95-96 and removed this.  Not sure why it was not producing an error, but could be because the value on line 96 was empty?

Also when you say "A key 'belongs`' to the previously defined table" -- what key are you referring to exactly?  There is an `[[inputs.snmp.table]]` on line 92 named "interfaceStats" -- so are you saying I need to move up the tag drop?   How do I go about putting the tag drop in the [[inputs.snmp]] table?  Is it just a matter of moving it up to above the first instance of the `[[inputs.snmp.field]]` entry on line 85?

I am a little confused on what key you are referring to and which table it should be put under?  Can you please clarify?

danielnelson: In this example the `fieldpass` key is under the `tagdrop` table, in order for it to work though it needs to be part of the plugins table: `[[inputs.snmp]]`.  Indention has no meaning, its just for readability, so yes it needs to be above any subtables.

The config file is using [TOML](https://github.com/toml-lang/toml), so I suggest reading through this document for a better understanding.

PR #5670, opened 2019-04-03
add wavefront serializer plugin

puckpuck:
### Required for all PRs:

Adds serializer support for Wavefront data format.  Has support for a limited number of settings from the equivalent output plugin.  Settings have a `wavefront_` prefix on them to ensure they will not collide with other settings.

Since serializer settings are still part of internal config, this did require a few changes to some internal packages.

### DIFF
--- a/plugins/serializers/wavefront/wavefront.go
+++ b/plugins/serializers/wavefront/wavefront.go
@@ -1,0 +1,196 @@
+package wavefront
+
+import (
+	"bytes"
+	"fmt"
+	"log"
+	"strconv"
+	"strings"
+
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/plugins/outputs/wavefront"
+)
+
+// WavefrontSerializer : WavefrontSerializer struct
+type WavefrontSerializer struct {
+	Prefix         string
+	UseStrict      bool
+	SourceOverride []string
+}
+
+// catch many of the invalid chars that could appear in a metric or tag name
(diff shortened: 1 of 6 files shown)


plugins/serializers/wavefront/wavefront.go
```diff
@@ -0,0 +1,196 @@
...
+		name = pathReplacer.Replace(name)
+
+		metric := &wavefront.MetricPoint{
+			Metric:    name,
+			Timestamp: m.Time().Unix(),
+		}
+
+		metricValue, buildError := buildValue(value, metric.Metric)
+		if buildError != nil {
+			log.Printf("E! Serializer [wavefront] %s\n", buildError.Error())
```
danielnelson: I think this should be a debug level message, or possibly silent at least for string fields.
```
D! [serializers.wavefront] Serialization error: %v
```

puckpuck: I moved log output to buildValue and only do a debug log when the default case is triggered.  Strings will silently error but continue to next metric.

danielnelson:
I updated the options in the README after merging: ca99569e6f19502894409de960d504fded9f7b9a.  Next step, update the Wavefront output to use this serializer?

puckpuck:
The output plugin leverages the wavefront go SDK which serializes the metrics before sending to Wavefront.  Unless we broke out the serialization inside the SDK I don’t believe we can share the same codebase :(

danielnelson:
Oh, that's too bad, perhaps something that can be done later on.

PR #3131, opened 2017-08-17
Added Cmdline to procstat

akrus:
### Required for all PRs:

- [X] Signed [CLA](https://influxdata.com/community/cla/).
- [ ] Associated README.md updated.
- [X] Has appropriate unit tests.

As a first approach to #1873, this PR adds cmdline with full command line of process as field. Not sure what should be updated in README.md. Adding as a tag is easy, but needs test change, so I'm afraid I need some assistance there.

PR #5681, opened 2019-04-04
adding cmdline procstat

### LINKED ISSUE #1873: get full process path via procstat as TAG
### System info:

telegraf-1.0.0-1.x86_64
RHEL6
## Feature Request

include command_line as a field collected from procstat . 
### Proposal:

get the full command line from  /proc/*/cmdline . return the contents of this file as a tag.
### Current behavior:

returns only the short name for the process and pid as the identifier of the process. 
### Desired behavior:

list the full process path as tag so that you may identify which process is which 

for example you run a java application, procstat current captures "java" as the process name. This is fine if your run one of something but it very common in enterprise architecture to run multiples of something with different configs file/arguments . Capturing only java make it impossible (or very hard) to differentiate between them for resource usage.

scottprichard:
closes #1873

### Required for all PRs:

### DIFF
--- a/plugins/inputs/procstat/README.md
+++ b/plugins/inputs/procstat/README.md
@@ -41,6 +41,9 @@ Processes can be selected for monitoring using one of several methods:
   ## Field name prefix
   # prefix = ""
 
+  ## When true add the full cmdline as a tag.
+  # cmdline_tag = false
+
   ## Add PID as a tag instead of a field; useful to differentiate between
   ## processes whose tags are otherwise the same.  Can create a large number
   ## of series, use judiciously.
@@ -72,6 +75,7 @@ implemented as a WMI query.  The pattern allows fuzzy matching using only
 - procstat
   - tags:
     - pid (when `pid_tag` is true)
+    - cmdline (when 'cmdline_tag' is true)
     - process_name
     - pidfile (when defined)
     - exe (when defined)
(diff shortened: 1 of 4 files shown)


scottprichard:
pulled master in...same thing as:
Added Cmdline to procstat #3131

scottprichard:
ready to merge, would be nice to get someone with write access to review this and merge it in

scottprichard:
@danielnelson :pray:

danielnelson:
PR is good but I am a little nervous about this value due to the potential length and contents of this data.  Let's do as suggested in https://github.com/influxdata/telegraf/issues/1873#issuecomment-439895335 and add a new option that, when enabled, will add this data as a *tag*.  The option would be disabled by default.
```
## When true add the full cmdline as a tag.
# cmdline_tag = false
```

Also make sure to update the README to list the new item.

scottprichard:
@danielnelson can you review ?

danielnelson:
Great, thank you!

scottprichard:
@danielnelson not in 1.10.3 ? when can I expect 1.11.0 to be cut ? or can this get put into the next minor ?

danielnelson:
1.11 should be released near the beginning of June, this isn't something we would put into a patch release but it should be easy to create a [custom package with docker](https://github.com/influxdata/telegraf/wiki/Packaging) or you can use one of the [nightly builds](https://github.com/influxdata/telegraf#nightly-builds)

PR #5682, opened 2019-04-05
New aggregator: Firstlast

oplehto:
### Required for all PRs:

A new aggregator that can be used to emit first and last metrics of a discrete series (that sends metrics for a finite amount of time). For example if we want to get the first and final values of a counter associated with a network flow or process statistics from `procstat`. 

This was developed for a specific in-house use case but it might be useful to the community at-large.


danielnelson:
Added with modification in #5820

PR #5683, opened 2019-04-05
Synproxy input plugin

rfrenayworldstream:
### Required for all PRs:

- [ ]
[influxdata_ccla_190404.pdf](https://github.com/influxdata/telegraf/files/3047208/influxdata_ccla_190404.pdf)
 Signed [CLA](https://influxdata.com/community/cla/).
plugins/inputs/synproxy/README.md
plugins/inputs/synproxy/synproxy_test.go

### DIFF
--- a/plugins/inputs/synproxy/synproxy.go
+++ b/plugins/inputs/synproxy/synproxy.go
@@ -37,8 +37,18 @@ func (k *Synproxy) Gather(acc telegraf.Accumulator) error {
 	return nil
 }
 
+func inSlice(haystack []string, needle string) bool {
+	for _, val := range haystack {
+		if needle == val {
+			return true
+		}
+	}
+	return false
+}
+
 func (k *Synproxy) getSynproxyStat() (map[string]interface{}, error) {
 	var hname []string
+	counters := []string{"entries", "syn_received", "cookie_invalid", "cookie_valid", "cookie_retrans", "conn_reopened"}
 	fields := make(map[string]interface{})
 
 	// Open synproxy file in proc filesystem
@@ -48,27 +58,35 @@ func (k *Synproxy) getSynproxyStat() (map[string]interface{}, error) {
 	}
 	defer file.Close()
 
-	// Read result
+	// Initialise expected fields
+	for _, val := range counters {
+		fields[val] = uint32(0)
+	}
+
 	scanner := bufio.NewScanner(file)
+	// Read header row
 	if scanner.Scan() {
 		line := scanner.Text()
 		// Parse fields separated by whitespace
 		dataFields := strings.Fields(line)
 		for _, val := range dataFields {
+			if !inSlice(counters, val) {
+				val = ""
+			}
 			hname = append(hname, val)
-			fields[val] = uint32(0)
 		}
 	}
-	if len(fields) == 0 {
+	if len(hname) == 0 {
 		return nil, fmt.Errorf("invalid data")
 	}
+	// Read data rows
 	for scanner.Scan() {
 		line := scanner.Text()
 		// Parse fields separated by whitespace
 		dataFields := strings.Fields(line)
 		// If number of data fields do not match number of header fields
-		if len(dataFields) != len(fields) {
-			return nil, fmt.Errorf("invalid number of columns in data, expected %d found %d", len(fields),
+		if len(dataFields) != len(hname) {
+			return nil, fmt.Errorf("invalid number of columns in data, expected %d found %d", len(hname),
 				len(dataFields))
 		}
 		for i, val := range dataFields {
@@ -78,13 +96,15 @@ func (k *Synproxy) getSynproxyStat() (map[string]interface{}, error) {
 			if err != nil {
 				return nil, fmt.Errorf("invalid value '%s' found", val)
 			}
-			fields[hname[i]] = fields[hname[i]].(uint32) + uint32(x)
+			if hname[i] != "" {
+				fields[hname[i]] = fields[hname[i]].(uint32) + uint32(x)
+			}
 		}
 	}
 	return fields, nil
 }
 
-func GetHostProc() string {
+func getHostProc() string {
 	procPath := "/proc"
 	if os.Getenv("HOST_PROC") != "" {
 		procPath = os.Getenv("HOST_PROC")
@@ -95,7 +115,7 @@ func GetHostProc() string {
 func init() {
 	inputs.Add("synproxy", func() telegraf.Input {
 		return &Synproxy{
-			statFile: path.Join(GetHostProc(), "/net/stat/synproxy"),
+			statFile: path.Join(getHostProc(), "/net/stat/synproxy"),
 		}
 	})
 }
--- a/plugins/inputs/synproxy/synproxy.go
+++ b/plugins/inputs/synproxy/synproxy.go
@@ -66,19 +66,19 @@ func (k *Synproxy) getSynproxyStat() (map[string]interface{}, error) {
 		line := scanner.Text()
 		// Parse fields separated by whitespace
 		dataFields := strings.Fields(line)
+		// If number of data fields do not match number of header fields
+		if len(dataFields) != len(fields) {
+			return nil, fmt.Errorf("invalid number of columns in data, expected %d found %d", len(fields),
+				len(dataFields))
+		}
 		for i, val := range dataFields {
 			// Convert from hexstring to int32
 			x, err := strconv.ParseUint(val, 16, 32)
 			// If field is not a valid hexstring
 			if err != nil {
 				return nil, fmt.Errorf("invalid value '%s' found", val)
-			// If index is out of boundary
-			} else if i >= len(fields) {
-				return nil, fmt.Errorf("value '%s' out of column boundary", val)
-			// If field is a valid hexstring and index not out of boundary
-			} else {
-				fields[hname[i]] = fields[hname[i]].(uint32) + uint32(x)
 			}
+			fields[hname[i]] = fields[hname[i]].(uint32) + uint32(x)
 		}
 	}
 	return fields, nil
--- a/plugins/inputs/synproxy/synproxy.go
+++ b/plugins/inputs/synproxy/synproxy.go
@@ -20,7 +20,7 @@ type Synproxy struct {
 }
 
 func (k *Synproxy) Description() string {
-	return fmt.Sprintf("Get synproxy statistics from %s", k.statFile)
+	return "Get synproxy statistics from /proc/net/stat/synproxy"
 }
 
 func (k *Synproxy) SampleConfig() string {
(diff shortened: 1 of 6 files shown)


plugins/inputs/synproxy/synproxy.go
```diff
@@ -0,0 +1,120 @@
+// Copyright (c) 2019 WorldStream B.V. All rights reserved.
```
danielnelson: Can you remove this copyright notice, all code in this repo needs to be covered under the top level LICENSE.md.

plugins/inputs/synproxy/synproxy.go
```diff
@@ -0,0 +1,120 @@
...
+			}
+		}
+	}
+	return fields, nil
+}
+
+func init() {
+	inputs.Add("synproxy", func() telegraf.Input {
+		return &Synproxy{
+			statFile: "/proc/net/stat/synproxy",
```
danielnelson: We should add support for the HOST_PROC environment variable, take a look at the [linux_sysctl_fs][1].

[1]: https://github.com/influxdata/telegraf/blob/master/plugins/inputs/linux_sysctl_fs/linux_sysctl_fs.go#L85

plugins/inputs/synproxy/synproxy.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+func (k *Synproxy) getSynproxyStat() (map[string]interface{}, error) {
+	var hname []string
+	fields := make(map[string]interface{})
+
+	// Open synproxy file in proc filesystem
+	file, err := os.Open(k.statFile)
+	if err != nil {
+		return nil, fmt.Errorf("synproxy: %s does not exist!", k.statFile)
```
danielnelson: Telegraf will automatically add the information about which plugin the error comes from, so no need to do that ourselves.  Normally you will want to append the original error, and make sure not to duplicate any message.  I think in this case you will want to return the original error, since it contains the details.  If you do this the logged error will be:
```
2019-10-19T00:11:58Z E! [inputs.synproxy] Error in plugin: open /proc/net/stat/synproxy: no such file or directory
```


For the other errors, remove the `synproxy` prefix, start the error string with a lowercase letter, don't end it with a `!`, and consider if another error should be appended with `%v`.

plugins/inputs/synproxy/synproxy_notlinux.go
```diff
@@ -0,0 +1,3 @@
+// +build !linux
+
+package synproxy
```
danielnelson: For unsupported platforms create an empty plugin that prints an error, the ethtool pr has a really nice example:

https://github.com/influxdata/telegraf/pull/5865/files#diff-980231a87c3954198c19881c18fd0126

plugins/inputs/synproxy/synproxy.go
```diff
@@ -0,0 +1,120 @@
...
+		for _, val := range dataFields {
+			hname = append(hname, val)
+			fields[val] = uint32(0)
+		}
+	}
+	if len(fields) == 0 {
+		return nil, fmt.Errorf("synproxy: Invalid data!")
+	}
+	for scanner.Scan() {
+		line := scanner.Text()
```
danielnelson: I'm having some difficulties finding documentation for this file, what does each row in the table represent?

plugins/inputs/synproxy/synproxy.go
```diff
@@ -0,0 +1,120 @@
...
+		// Parse fields separated by whitespace
+		dataFields := strings.Fields(line)
+		for i, val := range dataFields {
+			// Convert from hexstring to int32
+			x, err := strconv.ParseUint(val, 16, 32)
+			// If field is not a valid hexstring
+			if err != nil {
+				return nil, fmt.Errorf("synproxy: Invalid value '%s' found!", val)
+				//	hname = append(hname, val)
+				//	fields[val] = 0
```
danielnelson: Remove any commented out code

plugins/inputs/synproxy/synproxy.go
```diff
@@ -0,0 +1,120 @@
...
+		line := scanner.Text()
+		// Parse fields separated by whitespace
+		dataFields := strings.Fields(line)
+		for _, val := range dataFields {
+			hname = append(hname, val)
+			fields[val] = uint32(0)
+		}
+	}
+	if len(fields) == 0 {
+		return nil, fmt.Errorf("synproxy: Invalid data!")
```
danielnelson: This depends on what the rows mean, but I think likely we want to remove this error and return 0 values.

plugins/inputs/synproxy/synproxy.go
```diff
@@ -0,0 +1,120 @@
...
+	defer file.Close()
+
+	// Read result
+	scanner := bufio.NewScanner(file)
+	if scanner.Scan() {
+		line := scanner.Text()
+		// Parse fields separated by whitespace
+		dataFields := strings.Fields(line)
+		for _, val := range dataFields {
+			hname = append(hname, val)
```
danielnelson: Since we already know the column names, we can get away without parsing them at runtime.  Just create a slice of the header names and check that there are the expected number of columns.

plugins/inputs/synproxy/synproxy_test.go
```diff
@@ -0,0 +1,146 @@
...
+
+	acc := testutil.Accumulator{}
+	err := k.Gather(&acc)
+	assert.Error(t, err)
+	assert.Contains(t, err.Error(), "Invalid value")
+}
+
+func TestNoSynproxyFile(t *testing.T) {
+	tmpfile := makeFakeSynproxyFile([]byte(SynproxyFile_Normal))
+	os.Remove(tmpfile)
```
danielnelson: defer os.Remove

plugins/inputs/synproxy/synproxy.go
rfrenayworldstream: Ok clear, the noticed is removed.

plugins/inputs/synproxy/synproxy.go
rfrenayworldstream: Nice solution! updated now.

plugins/inputs/synproxy/synproxy.go
rfrenayworldstream: Done

plugins/inputs/synproxy/synproxy_test.go
rfrenayworldstream: This function will test behavior when the expected file is missing. Therefore we remove the file before ending the test. I added comment in the source to make this more clear.

plugins/inputs/synproxy/synproxy.go
rfrenayworldstream: Done

plugins/inputs/synproxy/synproxy.go
rfrenayworldstream: Synproxy is a Linux netfilter module used for SYN attack mitigation. Each row represents a core dedicated to process the IP traffic. Each column represents the counter values. For example "syn_received" represent the number of SYN packets received by the core.

The number of columns is fixed, but new counters might be added in a future kernel release.
The number of rows depends on the hardware used on the system.

plugins/inputs/synproxy/synproxy.go
rfrenayworldstream: This would break the plugin if a counter would be added in a future release of the kernel.

plugins/inputs/synproxy/synproxy_notlinux.go
rfrenayworldstream: Thanks, using an empty plugin which prints an error.

plugins/inputs/synproxy/synproxy.go
rfrenayworldstream: It would be an unlikely event, but if the number of columns is zero, the data would be invalid.

plugins/inputs/synproxy/synproxy.go
```diff
@@ -0,0 +1,101 @@
...
+		// Parse fields separated by whitespace
+		dataFields := strings.Fields(line)
+		for i, val := range dataFields {
+			// Convert from hexstring to int32
+			x, err := strconv.ParseUint(val, 16, 32)
+			// If field is not a valid hexstring
+			if err != nil {
+				return nil, fmt.Errorf("invalid value '%s' found", val)
+			// If index is out of boundary
+			} else if i >= len(fields) {
```
glinton: Do this check once on line 69:
```go
if len(dataFields) > len(fields) {
```

plugins/inputs/synproxy/synproxy.go
```diff
@@ -0,0 +1,101 @@
...
+				return nil, fmt.Errorf("value '%s' out of column boundary", val)
+			// If field is a valid hexstring and index not out of boundary
+			} else {
```
glinton: No need for the `else` stanza if the `if` blocks `return`

rfrenayworldstream:
> Just some small things from me, thanks.

Thanks for your suggestions!

plugins/inputs/synproxy/synproxy.go
danielnelson: If the kernel updated this file I expect they would append the new columns to the right, and keep the current ones in the existing order.  If we parse only the known columns we can expect the plugin to continue to work.

The main reason I ask for this change though is that we want to prevent the dynamic creation of new fields, only the fields documented in the README should ever be created.  This allows us to properly name and type the fields before start sending them to the database.

The plugin would most likely stop working anyway, we can't expect new columns to have the same formatting as the old, and because we are checking for a specific number of columns on line 70 we would begin producing an error.  If you want to make this more resilient to new field additions you could compare expected fields >= actual fields.

plugins/inputs/synproxy/synproxy.go
danielnelson: Is there a webpage or manpage with this information we can link to in the README?

plugins/inputs/synproxy/synproxy.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/influxdata/telegraf/plugins/inputs"
+)
+
+type Synproxy struct {
+	// Synproxy stats filename (proc filesystem)
+	statFile string
+}
+
+func (k *Synproxy) Description() string {
+	return fmt.Sprintf("Get synproxy statistics from %s", k.statFile)
```
danielnelson: Make this a static string with the default location, we wouldn't want the description to change based on the environment variables set.

plugins/inputs/synproxy/synproxy.go
rfrenayworldstream: The synproxy counters are not described at all. However there is a manpage mentioning the synproxy iptables extension `man iptables-extenions`. I have added this to the README to make it more clear how to use synproxy.

plugins/inputs/synproxy/synproxy.go
rfrenayworldstream: Done

plugins/inputs/synproxy/synproxy.go
rfrenayworldstream: I understand you concern, it would be undesirable indeed if new counters would be added automatically.
I have added some code to ignore new fields and even allow the order to be changed.
Also the test for these conditions are added.

It might be considered an overkill since we do not expect these changes. But at least this plugin will be bullet proof for changes. :)

PR #5693, opened 2019-04-08
Kinesis input compressed metrics

morfien101:
### Required for all PRs:

###############################################

This code change is a supplement to PR #5588

With this you can now download the compressed metrics using the Telegraf Kinesis Input plugin.

### DIFF
--- a/plugins/inputs/kinesis_consumer/kinesis_consumer.go
+++ b/plugins/inputs/kinesis_consumer/kinesis_consumer.go
@@ -23,6 +23,7 @@ import (
 const (
 	compressionGZIP   = "gzip"
 	compressionSnappy = "snappy"
+	compressionNone   = "none"
 )
 
 type (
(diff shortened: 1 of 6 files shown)


plugins/inputs/kinesis_consumer/kinesis_consumer.go
```diff
@@ -39,6 +44,8 @@ type (
 		ShardIteratorType      string    `toml:"shard_iterator_type"`
 		DynamoDB               *DynamoDB `toml:"checkpoint_dynamodb"`
 		MaxUndeliveredMessages int       `toml:"max_undelivered_messages"`
+		CompressedMetrics      bool      `toml:"compressed_metrics"`
```
glinton: Let's remove this `bool` and use the `compression_type` only.

plugins/inputs/kinesis_consumer/kinesis_consumer.go
```diff
@@ -237,7 +251,42 @@ func (k *KinesisConsumer) Start(ac telegraf.Accumulator) error {
...
+		decompressedData, err = decompressSnappy(input)
+		if err != nil {
+			return nil, err
+		}
+	case compressionGZIP:
+		decompressedData, err = decompressGZip(input)
+		if err != nil {
+			return nil, err
+		}
+	default:
```
glinton: Default to uncompressed and set `decompressedData = input`; add a case for a compression type of "none" and do the same.

morfien101: I'm happy to do this. Can we do the same for the output plugin in that case. #5588

sjwang90:
Closing this since we added this as an external plugin to #5588. Feel free to re-open if that's not the case.

PR #5695, opened 2019-04-08
init.sh: prevent potential infinite loop

sajoupa:
### Required for all PRs:

Hi,

The init.sh file which is used in packages, has a potential infinite loop. This should prevent it.

--
Laurent

### DIFF
--- a/scripts/init.sh
+++ b/scripts/init.sh
@@ -153,16 +153,22 @@ case $1 in
             pidofproc -p $pidfile $daemon > /dev/null 2>&1 && status="0" || status="$?"
             if [ "$status" = 0 ]; then
                 # periodically signal until process exists
-                while true; do
+                retries=0
+                while [ $retries -le 15 ]; do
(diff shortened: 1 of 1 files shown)


danielnelson:
On side effect of this change is that reload will now fail if it takes more than 30s for Telegraf to shutdown. 
Can you explain the issue that this was causing in more depth?

sajoupa:
No specific issue that I encountered, just thought it'd be safer to make sure that an infinite loop couldn't happen.
I picked 30s because from my experience it's more than enough for telegraf to shutdown, but it could be higher.

adrianlzt:
I don't have any measurements, but I would say that 30s could be not enough.

It could be safer to wait some time, then do a kill -9 and check that it have stopped

danielnelson:
I don't think we should make this change.  The user running the script `invoke-rc.d telegraf stop` can always ctrl-c the script to stop waiting and I think we should leave `kill -9` to the admin since I would hate to kill the process while it is successfully shutting down, losing all the metrics.

Issue #5306, opened 2019-01-17
Add namespace configuration to Prometheus plugin


simoelmou:
## Feature Request

Opening a feature request kicks off a discussion.

### Proposal:
Add namespace configuration to Prometheus plugin

### Current behavior:
Currently, Prometheus input plugin allows retrieving metrics from the pods of all the namespaces in a cluster.

### Desired behavior:
There should be a configuration where you can specify the namespace(s) that you want to retrieve metrics from. This would allow deploying telegraf in the case where the permissions are limited to only some certain namespaces.

### Use case: [Why is this important (helps with prioritizing requests)]
This is a quick win because it's simple to implement and also allows to use Prometheus input plugin in a limited environment where it's usually the case in enterprises.

goller:
@simoelmou I like that idea.  It also helps scale in a way as different telegrafs could be configured for different namespaces.

danielnelson:
It may be possible to do this currently with a properly configured `kube_config` file, it looks like you will need to define a context and set the current-context option to select it.

simoelmou:
@danielnelson In an Openshift environment that would be complicated to put in place, especially in a production cluster. At the moment I created a fork of the plugin that would only monitor the pods of the namespace where telegraf is deployed.

danielnelson:
Closed in #5697

guai:
Hi
Is there an option so telegraf watched only for the namespace it is deployed in?

PR #5697, opened 2019-04-08
Add optional namespace restriction to prometheus input plugin

Benjamintf1:
Allow prometheus to operate as a namespaced pod with namespace level permissions. This addresses https://github.com/influxdata/telegraf/issues/5306

### Required for all PRs:

- [√] Signed [CLA](https://influxdata.com/community/cla/).
- [√] Associated README.md updated.
- [√] Has appropriate unit tests.

### DIFF
--- a/plugins/inputs/prometheus/README.md
+++ b/plugins/inputs/prometheus/README.md
@@ -24,8 +24,9 @@ in Prometheus format.
   ## - prometheus.io/path: If the metrics path is not /metrics, define it with this annotation.
   ## - prometheus.io/port: If port is not 9102 use this annotation
   # monitor_kubernetes_pods = true
-  ## Only monitor one namespace
-  # monitor_kubernetes_pods_namespace = "test-namespace"
+  ## Restricts Kubernetes monitoring to a single namespace
+  ##   ex: monitor_kubernetes_pods_namespace = "default"
+  # monitor_kubernetes_pods_namespace = ""
 
   ## Use bearer token for authorization. ('bearer_token' takes priority)
   # bearer_token = "/path/to/bearer/token"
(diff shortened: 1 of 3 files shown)


Benjamintf1:
We see that there was a previous issue about this. Seeing their solution, it seems like it could be reasonable/convenient to have it be a toggle and ONLY apply to the namespace of the telegraf config, but we also think the usecases that this supports are wider.

Benjamintf1:
We seem to be having failed tests unrelated to our changes.. I tried to close/open this pr to run the tests again, this does not seem to re-run tests.

plugins/inputs/prometheus/README.md
```diff
@@ -24,6 +24,8 @@ in Prometheus format.
   ## - prometheus.io/path: If the metrics path is not /metrics, define it with this annotation.
   ## - prometheus.io/port: If port is not 9102 use this annotation
   # monitor_kubernetes_pods = true
+  ## Only monitor one namespace
+  # monitor_kubernetes_pods_namespace = "test-namespace"
```
danielnelson: We try to use the default value in the commented out option, but maybe we should show an example setting, does this look right?
```
  ## Restrict Kubernetes monitoring to one namespace.
  ##   ex: monitor_kubernetes_pods_namespace = "default"
  # monitor_kubernetes_pods_namespace = ""
```

https://github.com/influxdata/telegraf/wiki/SampleConfig#style

Benjamintf1:
Done!

danielnelson:
Thanks, I'll switch that option after we merge this.  I don't see your github name in our records as having signed the CLA, could you sign it?  If you already signed, sorry something didn't work, its okay to do it again.

Benjamintf1:
I believe we should be on the pivotal corporate cla. I'm double checking some stuff though as well.

danielnelson:
I don't see a completed CCLA for Pivotal on my side either, but even so it is also required to do the individual CLA.  There is a more info on needing both here: https://www.influxdata.com/legal/ccla/

Benjamintf1:
Ok, I should have just signed the individual cla.

PR #5703, opened 2019-04-10
Don't require a telegraf.conf when config-directory specified

### LINKED ISSUE #5571: Telegraf requires config even when using a config-directory

### LINKED ISSUE #5656: Support --config-directory in Docker
## Feature Request

Would like an option to use --config-directory in docker rather than one telegraf.conf

### Proposal:
Map host/container directory and default config will be stored here and --config-directory used.

### Current behavior:
defaults to /etc/telegraf/telegraf.conf

allows for cleaner management of plugins and other services

glinton:
Resolves #5656 
Resolves #5571 

(Replaces #5573)

### DIFF
--- a/cmd/telegraf/telegraf.go
+++ b/cmd/telegraf/telegraf.go
@@ -120,7 +120,7 @@ func runAgent(ctx context.Context,
 	c.OutputFilters = outputFilters
 	c.InputFilters = inputFilters
 	err := c.LoadConfig(*fConfig)
-	if err != nil {
+	if err != nil && *fConfigDirectory == "" {
 		return err
 	}
 


cmd/telegraf/telegraf.go
```diff
@@ -120,7 +120,7 @@ func runAgent(ctx context.Context,
 	c.OutputFilters = outputFilters
 	c.InputFilters = inputFilters
 	err := c.LoadConfig(*fConfig)
-	if err != nil {
+	if err != nil && *fConfigDirectory == "" {
```
danielnelson: This would ignore parse errors in the initial config file, which should be fatal.  I think to fix this issue properly we need to split the functions reading the configuration files from the function parsing them.

Errors reading a file can be ignored/logged, so long as inputs+outputs are found.  But errors parsing a file should be reported and be fatal.

glinton: Ooh, good point. Was #5573 more what you'd like to see?

danielnelson: I think we need the split in responsibility in order to give the error messages we want, and the http config option complicates this further.  I can't really understand #5573, it might work but it doesn't have a logical flow to me.  Also, I would only use a bool in a function parameter if it were the only argument.

This is trickier than it looks, and I think we should put it on ice until we have time to revamp the config loading in depth.

russorat:
can this be closed?

PR #5713, opened 2019-04-11
Enhance HTTP connection options for phpfpm input plugin

### LINKED ISSUE #5561: Support insecure_skip_verify in input.phpfpm
## Feature Request
Support insecure_skip_verify in input.phpfpm

### Proposal:
input.phpfpm should allow skipping TLS certificate verification by having an `insecure_skip_verify = true|false` configuration option.

### Current behavior:
input.phpfpm always fails when trying to access a php_fpm status page reverse-proxied using a self-signed certificate.

### Desired behavior:
Users should be able to skip certificate verification for self-signed certs in known safe environments.

### Use case:
A lot of other plugins already have this feature. Users may want to gather statistics on php-fpm instances running behind reverse proxies that use self-signed certificates for internal connections.

glinton:
Resolves #5561

### DIFF
--- a/plugins/inputs/phpfpm/phpfpm.go
+++ b/plugins/inputs/phpfpm/phpfpm.go
@@ -13,6 +13,8 @@ import (
 	"sync"
 
 	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/internal"
+	"github.com/influxdata/telegraf/internal/tls"
 	"github.com/influxdata/telegraf/plugins/inputs"
 )
 
@@ -35,7 +37,9 @@ type metric map[string]int64
 type poolStat map[string]metric
 
 type phpfpm struct {
-	Urls []string
+	Urls            []string
+	ResponseTimeout internal.Duration
+	tls.ClientConfig
 
 	client *http.Client
 }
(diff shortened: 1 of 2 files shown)


plugins/inputs/phpfpm/phpfpm.go
```diff
@@ -96,8 +110,18 @@ func (g *phpfpm) Gather(acc telegraf.Accumulator) error {
...
 func (g *phpfpm) gatherServer(addr string, acc telegraf.Accumulator) error {
 	if g.client == nil {
-		client := &http.Client{}
-		g.client = client
+		tlsCfg, err := g.ClientConfig.TLSConfig()
+		if err != nil {
+			return err
+		}
+		tr := &http.Transport{
+			ResponseHeaderTimeout: g.ResponseTimeout.Duration,
```
danielnelson: This line is covered by the Client.Timeout so it can be removed.

glinton: I suppose they do, though they serve different purposes. We should also clean it up from httpjson input when we check that next.

danielnelson: The httpjson plugin is deprecated so I wouldn't worry about it.  This timeout is great if you are streaming the response body but otherwise its just simpler to use client timeout vs contemplating all the possible timeouts.  Just my 2 cents.

plugins/inputs/phpfpm/phpfpm.go
```diff
@@ -58,9 +62,19 @@ var sampleConfig = `
   ##       "fcgi://10.0.0.12:9000/status"
   ##       "cgi://10.0.10.12:9001/status"
   ##
-  ## Example of multiple gathering from local socket and remove host
+  ## Example of multiple gathering from local socket and remote host
   ## urls = ["http://192.168.1.20/status", "/tmp/fpm.sock"]
   urls = ["http://localhost/status"]
+
+  ## Set response timeout for http[s] urls (default 5 seconds)
+  # response_timeout = "5s"
```
danielnelson: Now that you mentioned this was based on httpjson... we should probably clean this up.  Let's do:
```
  ## Duration allowed to complete HTTP requests.
  # timeout = "5s"
```

Key features here are not explicitly saying the default, instead we rely on having it shown in the example value.  Also we use the "duration" to indicate that it is a duration string and use a complete sentence.

PR #5726, opened 2019-04-15
Fixed interval estimation issues

### LINKED ISSUE #5619: Telegraf 1.10.1 default transform vsphere data from 20s to 60s sample.
### Relevant telegraf.conf:
```
# Global tags can be specified here in key="value" format.
[global_tags]
  # dc = "us-east-1" # will tag all metrics with dc=us-east-1
  # rack = "1a"
  ## Environment variables can be used as tags, and throughout the config file
  # user = "$USER"


# Configuration for telegraf agent
[agent]
  ## Default data collection interval for all inputs
  interval = "20s"
  ## Rounds collection interval to 'interval'
  ## ie, if interval="10s" then always collect on :00, :10, :20, etc.
  round_interval = true

  ## Telegraf will send metrics to outputs in batches of at most
  ## metric_batch_size metrics.
  ## This controls the size of writes that Telegraf sends to output plugins.
  metric_batch_size = 10000

  ## For failed writes, telegraf will cache metric_buffer_limit metrics for each
  ## output, and will flush this buffer on a successful write. Oldest metrics
  ## are dropped first when this buffer fills.
  ## This buffer only fills when writes fail to output plugin(s).
  metric_buffer_limit = 1000000

  ## Collection jitter is used to jitter the collection by a random amount.
  ## Each plugin will sleep for a random time within jitter before collecting.
  ## This can be used to avoid many plugins querying things like sysfs at the
  ## same time, which can have a measurable effect on the system.
  collection_jitter = "0s"

  ## Default flushing interval for all outputs. Maximum flush_interval will be
  ## flush_interval + flush_jitter
  flush_interval = "5s"
  ## Jitter the flush interval by a random amount. This is primarily to avoid
  ## large write spikes for users running a large number of telegraf instances.
  flush_jitter = "0s"

  ## By default or when set to "0s", precision will be set to the same
  ## timestamp order as the collection interval, with the maximum being 1s.
  ##   ie, when interval = "10s", precision will be "1s"
  ##       when interval = "250ms", precision will be "1ms"
  ## Precision will NOT be used for service inputs. It is up to each individual
  ## service input to set the timestamp at the appropriate precision.
  ## Valid time units are "ns", "us" (or "µs"), "ms", "s".
  precision = ""
  ## Logging configuration:
  ## Run telegraf with debug log messages.
  debug = true
  ## Run telegraf in quiet mode (error log messages only).
  quiet = false
  ## Specify the log file name. The empty string means to log to stderr.
  logfile = "/var/log/telegraf/telegraf.log"

  ## Override default hostname, if empty use os.Hostname()
  hostname = ""
  ## If set to true, do no set the "host" tag in the telegraf agent.
  omit_hostname = false


# Configuration for sending metrics to InfluxDB
[[outputs.influxdb]]
  ## The full HTTP or UDP URL for your InfluxDB instance.
  ##
  ## Multiple URLs can be specified for a single cluster, only ONE of the
  ## urls will be written to each interval.
  # urls = ["unix:///var/run/influxdb.sock"]
  # urls = ["udp://127.0.0.1:8089"]

  ## The target database for metrics; will be created as needed.
  ## For UDP url endpoint database needs to be configured on server side.
   database = "VC"

  ## The value of this tag will be used to determine the database.  If this
  ## tag is not set the 'database' option is used as the default.
  # database_tag = ""

  ## If true, no CREATE DATABASE queries will be sent.  Set to true when using
  ## Telegraf with a user without permissions to create databases or when the
  ## database already exists.
  # skip_database_creation = false

  ## Name of existing retention policy to write to.  Empty string writes to
  ## the default retention policy.  Only takes effect when using HTTP.
  # retention_policy = ""

  ## Write consistency (clusters only), can be: "any", "one", "quorum", "all".
  ## Only takes effect when using HTTP.
  # write_consistency = "any"

  ## Timeout for HTTP messages.
   timeout = "10s"

  ## HTTP Basic Auth
   username = "username"
   password = "password"

  ## HTTP User-Agent
  # user_agent = "telegraf"

  ## UDP payload size is the maximum packet size to send.
  # udp_payload = "512B"
# # Read metrics from VMware vCenter
 [[inputs.vsphere]]
#   ## List of vCenter URLs to be monitored. These three lines must be uncommented
#   ## and edited for the plugin to work.
   vcenters = [ "https://vcenter/sdk" ]
   username = "username"
   password = "password"
#
#   ## VMs
#   ## Typical VM metrics (if omitted or empty, all metrics are collected)
#   vm_include = [ "/*/vm/**"] # Inventory path to VMs to collect (by default all are collected)
   vm_metric_include = [
        "cpu.usage.average",
        "cpu.ready.summation",
        "cpu.costop.summation",
        "mem.usage.average",
        "mem.swapped.average",
        "net.bytesTx.average",
        "net.bytesRx.average",
        "virtualDisk.totalWriteLatency.average",
        "virtualDisk.totalReadLatency.average",
        "disk.numberReadAveraged.average",
        "disk.numberWriteAveraged.average",
        "disk.read.average",
        "disk.write.average",
        "sys.osUptime.latest",
   ]

#   # vm_metric_exclude = [] ## Nothing is excluded by default
#   # vm_instances = true ## true by default
#   ## whether or not to force discovery of new objects on initial gather call before collecting metrics
#   ## when true for large environments this may cause errors for time elapsed while collecting metrics
#   ## when false (default) the first collection cycle may result in no or limited metrics while objects are discove$
    force_discover_on_init = true
#
#   ## the interval before (re)discovering objects subject to metrics collection (default: 300s)
#   # object_discovery_interval = "300s"
#
#   ## timeout applies to any of the api request made to vcenter
#   # timeout = "60s"
#
#   ## Optional SSL Config
#   # ssl_ca = "/path/to/cafile"
#   # ssl_cert = "/path/to/certfile"
#   # ssl_key = "/path/to/keyfile"
#   ## Use SSL but skip chain & host verification
    insecure_skip_verify = true
```
### System info:

Telegraf: 1.10.1
OS: ubuntu 18.04.02
vSphere: 6.5
### Expected behavior:
Possibility to select sample or leave unchanged.

### Actual behavior:
image
### Additional info:

After upgrade telegraf from version 1.9.5 to 1.10.1 all collected real-time vsphere metrics are transform instead of 20s to 60s samples. I tried added param ` interval = "20s"` but it not works. **Is possible change settings for turn off default transformation to 60s sample?**

How to get back metrics with 20s sample?

prydin:
closes #5619

### Required for all PRs:

Fixed a bug in the interval estimation logic that caused sample rates to be limited to 60s.


prydin:
@glinton I'm not sure this should be classified as an enhancement. It's a bug that prevents collection at less than 60s interval.

glinton:
lol, good catch. Thanks

danielnelson:
Builds from CI for testing:

- [telegraf-1.11.0~6d48257b-0.x86_64.rpm](https://31272-33258973-gh.circle-artifacts.com/0/build/linux/amd64/telegraf-1.11.0~6d48257b-0.x86_64.rpm)
- [telegraf_1.11.0~6d48257b-0_amd64.deb](https://31272-33258973-gh.circle-artifacts.com/0/build/linux/amd64/telegraf_1.11.0~6d48257b-0_amd64.deb)
- [telegraf-1.11.0~6d48257b_windows_amd64.zip](https://31272-33258973-gh.circle-artifacts.com/0/build/windows/amd64/telegraf-1.11.0~6d48257b_windows_amd64.zip)

danielnelson:
@prydin Can you help me out with some context on why we need to estimate the interval and align samples into buckets?

prydin:
> @prydin Can you help me out with some context on why we need to estimate the interval and align samples into buckets?
> 
> I'm putting this in 1.11.0 for now, but will maybe move it into 1.10.

If vCenter is under high load, it can sometimes delay the publishing of metrics. For example, if at 12:00PM I ask for the last minute of samples (i.e. the period 11:59AM--12:00PM), some samples that should have been generated for that period aren't available yet. When this happens, the plugin makes a note of that. Now, let's say we're asking for samples again at 12:01PM. Normally we would ask for samples for the period 12:00PM--12:01PM, but since we know that some samples were missing from the last collection, we extend the period to 11:59AM--12:01AM. 

So far so good. The problem is that the data we get when asking for a longer time range like that will come in 20s period samples. But if our sampling period is set to 60s, we don't want to send more samples than we promised to the outputs, since this could cause scalability issues. So we have to reduce the resolution to whatever the sample period was set to in Telegraf. The problem is that this parameter isn't available to plugins, so we have to estimate it. We do that simply by measuring the time between consecutive calls to Gather(). 

Finally, once we know the sample period, we normalize the samples into 60s buckets through a simple averaging algorithm.

danielnelson:
Thanks for the explanation, sounds good.  The rest of this comment is just some thoughts about what a plugin should do and how other plugins have dealt with this issue to maybe get you thinking.

I wouldn't neccessarily consider it a problem if an input generated multiple samples per interval.  I believe the cloudwatch input does this with certain configurations.  The thing an input shouldn't do is multiple polling operations: query, wait, query.

The primary use of the interval on the input side is to reduce the load of monitoring on the monitored service.  Reducing the amount of output data can always be done by pairing the input with an aggregator.  Still, it could be nice to have control over the resolution of the data and I think this could be done with an new option if it makes sense for the resolution to be different from the interval.

I should mention that the only other plugin I know of that aggregates over the interval size is the statsd plugin.

Another technique used by the cloudwatch/stackdriver input for something similar is a `delay` option.  This is a fixed size delay to the sample period, and one nice thing about this method is that it doesn't attempt to recover from missing data, which could put additional load on the server at the very time that it cannot keep up.  The downside is always having the delay and potentially having "missing" data during high load.

I could also imagine a world where the vsphere plugin takes the last 20s sample every minute (or another interval) and reports only that.  If the data wasn't found it would just skip it.  This would depend on the metrics from the vsphere server being a certain type, for example it would only work nicely if it had non-reseting counters so you could still apply a sum aggregation on query even with "missing" samples.

Issue #5398, opened 2019-02-09
[[inputs.system]] /var/run/utmp: no such file... when using Alpine as base Image


echtelerp:
some informations to #5283 and the older issue #5031: and as asked by @danielnelson 
here the Issue 

### Relevant telegraf.conf:
used the standard, did no changes. just ran the two commands provided in the readme of Dockerhub:
`$ docker run -d --name influxdb -p 8083:8083 -p 8086:8086 influxdb`
`$ docker run --net=container:influxdb telegraf:alpine`

### System info:
Host MacOs, Docker Version 2.0.0.2

[Include Telegraf version, operating system name, and other relevant details]

```
DockerEnvironment$ docker run --net=container:influxdb telegraf:alpine
2019-02-08T21:54:40Z I! Starting Telegraf 1.9.4
2019-02-08T21:54:40Z I! Using config file: /etc/telegraf/telegraf.conf
2019-02-08T21:54:40Z I! Loaded inputs: inputs.mem inputs.processes inputs.swap inputs.system inputs.cpu inputs.disk inputs.diskio inpu
ts.kernel
2019-02-08T21:54:40Z I! Loaded aggregators: 
2019-02-08T21:54:40Z I! Loaded processors: 
2019-02-08T21:54:40Z I! Loaded outputs: influxdb
2019-02-08T21:54:40Z I! Tags enabled: host=9ab66945f415
2019-02-08T21:54:40Z I! [agent] Config: Interval:10s, Quiet:false, Hostname:"9ab66945f415", Flush Interval:10s
2019-02-08T21:54:50Z E! [inputs.system]: Error in plugin: open /var/run/utmp: no such file or directory
...
...
```

### Steps to reproduce:

1. ...`$ docker run -d --name influxdb -p 8083:8083 -p 8086:8086 influxdb`
2. ...`$ docker run --net=container:influxdb telegraf:alpine`
watch the command line output. 

### Expected behavior:
run without errors

### Actual behavior:
`2019-02-08T21:54:50Z E! [inputs.system]: Error in plugin: open /var/run/utmp: no such file or directory`

### Additional info:

I did dig deeper in this problem. 
Alpine as being such a minimal image somehow does not provide the file, that [[inputs.system]] is looking for. 
The command uptime is working. 
```
bash-4.4# uptime
 22:11:56 up  6:36,  0 users,  load average: 0.44, 0.43, 0.30
```
I did not find a solution how to load the utmp corresponding program via apk. 

[...]

Issue #5283, opened 2019-01-14
HOST_VAR not respected in docker.inputs


rantoniuk:
Looking at #5031 we should be configuring `[[inputs.docker]]` plugin with HOST_VAR for monitoring host system. One can obviously override `#  endpoint = "unix:///var/run/docker.sock"` to include `/host` but I guess that is inconsistency in configuration?

danielnelson:
This is an interesting point that I haven't considered.  The `HOST_VAR`, `HOST_PROC` variables are only used by plugins that don't expose the locations of the files they parse, while the docker plugin doesn't know where the socket is and just uses the given path.  I also think the environment variables are not very ideal, I would rather if these locations were set in the config file.

I understand how this would be confusing but I think we shouldn't fix this in the docker input, and instead work on making the host prefix more configurable from the config file.

echtelerp:
some informations to #5283 and the older issue #5031: 

### Relevant telegraf.conf:
used the standard, did no changes. just ran the two commands provided in the readme of Dockerhub:
`$ docker run -d --name influxdb -p 8083:8083 -p 8086:8086 influxdb`
`$ docker run --net=container:influxdb telegraf:alpine`

### System info:
Host MacOs, Docker Version 2.0.0.2

[Include Telegraf version, operating system name, and other relevant details]

```
DockerEnvironment$ docker run --net=container:influxdb telegraf:alpine
2019-02-08T21:54:40Z I! Starting Telegraf 1.9.4
2019-02-08T21:54:40Z I! Using config file: /etc/telegraf/telegraf.conf
2019-02-08T21:54:40Z I! Loaded inputs: inputs.mem inputs.processes inputs.swap inputs.system inputs.cpu inputs.disk inputs.diskio inpu
ts.kernel
2019-02-08T21:54:40Z I! Loaded aggregators: 
2019-02-08T21:54:40Z I! Loaded processors: 
2019-02-08T21:54:40Z I! Loaded outputs: influxdb
2019-02-08T21:54:40Z I! Tags enabled: host=9ab66945f415
2019-02-08T21:54:40Z I! [agent] Config: Interval:10s, Quiet:false, Hostname:"9ab66945f415", Flush Interval:10s
2019-02-08T21:54:50Z E! [inputs.system]: Error in plugin: open /var/run/utmp: no such file or directory
...
...
```

### Steps to reproduce:

1. ...`$ docker run -d --name influxdb -p 8083:8083 -p 8086:8086 influxdb`
[...]

Issue #3837, opened 2018-02-26
[[inputs.system]] fails on SuSE with permission denied for utmp


Crapworks:
### Relevant telegraf.conf:

```ini
[[inputs.system]]
```

### System info:

Telegraf v1.5.2 (git: release-1.5 67440c95)
SuSE 11.3 (x86_64) (Kernel 3.0.93-0.8-xen) 

### Steps to reproduce:

1. Enable System Input Plugin
2. Run Telegraf

### Expected behavior:

Gather metrics about load, uptime, etc.

### Actual behavior:

Error:
```
Error in Plugin [inputs.system]: open /var/run/utmp: permission denied
```

### Additional info:

```
ls -lha /var/run/utmp
-rw-r----- 1 root utmp 23K Feb 26 10:24 /var/run/utmp
```

### Proposal:

It seems that this error is coming from the gopsutil library. It would be nice to be able to disable that specific feature to avoid reading utmp, or if reading it fails, still gather the metrics that are note dependent on that file being readable.

I am aware that I can also add the Telegraf user to the utmp group, but I would rather have this work out of the box.

PR #5734, opened 2019-04-17
add flag to remove uptime and n_users

ehlerst:
This PR allows setting of flags to turn them off for those who wish.

### Required for all PRs:

https://github.com/influxdata/telegraf/issues/5398
https://github.com/influxdata/telegraf/issues/5283
https://github.com/influxdata/telegraf/issues/3837
more...

Flags set to true

```
system,host=some-clever-name load1=2.18,load5=3.31,load15=3.45,n_cpus=8i 1555475800000000000
```

Flags set to false (still the default)
```
system,host=some-clever-name uptime_format=" 0:34" 1555475750000000000
system,host=some-clever-name uptime=2095i 1555475750000000000
system,host=some-clever-name load1=2.36,load5=3.5,load15=3.52,n_cpus=8i,n_users=1i 1555475750000000000
```

### DIFF
--- a/plugins/inputs/system/README.md
+++ b/plugins/inputs/system/README.md
@@ -8,8 +8,12 @@ and number of users logged in. It is similar to the unix `uptime` command.
 ```toml
 # Read metrics about system load & uptime
 [[inputs.system]]
-  # no configuration
+  ## Uncomment the following line to disable uptime collection
+  # skip_uptime = true
+  ## Uncomment the following line to disable uptime collection
+  # skip_users = true
 ```
+
 #### Permissions:
 
 The `n_users` field requires read access to `/var/run/utmp`, and may require
--- a/plugins/inputs/system/system.go
+++ b/plugins/inputs/system/system.go
@@ -15,20 +15,30 @@ import (
 	"github.com/shirou/gopsutil/load"
 )
 
-type SystemStats struct{}
+type SystemStats struct {
+	SkipUsers  bool
+	SkipUptime bool
+}
+
+var SystemsampleConfig = `
+
+## Uncomment to remove deprecated metrics.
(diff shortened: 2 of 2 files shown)


plugins/inputs/system/README.md
```diff
@@ -8,8 +8,12 @@ and number of users logged in. It is similar to the unix `uptime` command.
 ```toml
 [[inputs.system]]
-  # no configuration
+  ## Uncomment the following line to disable uptime collection
+  # skip_uptime = true
```
glinton: Let's make this a filter. I think just an exclude filter is sufficient. Allowed values would be `uptime` and `users`.
```
filter.NewIncludeExcludeFilter([]string{}, SomethingExclude)
```
The docker input plugin uses filters and may be a good reference.

plugins/inputs/system/system.go
```diff
@@ -15,20 +15,30 @@ import (
...
 )
 
-type SystemStats struct{}
+type SystemStats struct {
+	SkipUsers  bool
+	SkipUptime bool
+}
+
+var SystemsampleConfig = `
+
```
glinton: Remove this blank line.

danielnelson:
I feel like this is too many options for such a small plugin.

The only remaining issue with utmp that I know of is when it doesn't exist.  I think it would be a good change to skip the error logging if `os.IsNotExist(err)` and add a note to the README that it requires this file.

With uptime, I haven't seen any isses at all, if its just something you don't want you can add it to the fielddrop option:
```
[[inputs.system]]
  fielddrop = ["uptime*"]
```

ehlerst:
I like @danielnelson idea better, https://github.com/influxdata/telegraf/pull/5742 to replace this.

PR #5741, opened 2019-04-17
Fix docs for metric buffer limit to reflect current behavior

danielnelson:
This corrects an out of date description of how the metric buffers work.

### Required for all PRs:


danielnelson:
We don't list any other default values there.  I'm happy with the default being shown in the config file, and while normally I would say this should be commented out, and I could be convinced of this, I think the agent section is special enough to leave it uncommented.
```
flush_interval = "10s"
```

PR #5742, opened 2019-04-17
skip when utmp is missing

ehlerst:
### Required for all PRs:

Replacement for: https://github.com/influxdata/telegraf/pull/5734

### DIFF
--- a/plugins/inputs/system/system.go
+++ b/plugins/inputs/system/system.go
@@ -48,7 +48,7 @@ func (_ *SystemStats) Gather(acc telegraf.Accumulator) error {
 	} else if os.IsNotExist(err) {
 		log.Printf("D! [inputs.system] Error reading users: %v", err)
 	} else if os.IsPermission(err) {
-		return err
+		log.Printf("D! [inputs.system] %v", err)
 	}
 
 	now := time.Now()
--- a/plugins/inputs/system/system.go
+++ b/plugins/inputs/system/system.go
@@ -16,9 +16,7 @@ import (
 	"github.com/shirou/gopsutil/load"
 )
 
-type SystemStats struct {
-	WarnOnMissing bool
-}
+type SystemStats struct{}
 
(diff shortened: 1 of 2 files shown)


ehlerst:
@danielnelson is this kind of what you proposed? I took inspiration from the perf metrics plugins warning log entries.

Should the warning be enabled by default? I have no preference here.

plugins/inputs/system/system.go
```diff
@@ -44,7 +49,9 @@ func (_ *SystemStats) Gather(acc telegraf.Accumulator) error {
 	users, err := host.Users()
 	if err == nil {
 		fields["n_users"] = len(users)
-	} else if !os.IsPermission(err) {
+	} else if os.IsNotExist(err) && s.WarnOnMissing {
+		log.Printf("[inputs.system]: %s\n", err.Error())
```
danielnelson: Let's do debug level:
```
log.Printf("D! [inputs.system] Error reading users: %v, err")
```

plugins/inputs/system/system.go
```diff
@@ -25,10 +28,12 @@ func (_ *SystemStats) SampleConfig() string {
 	return `
   ## Uncomment to remove deprecated metrics.
   # fielddrop = ["uptime_format"]
+  ## Print out when utmp is not available for n_users
+  # warn_on_missing = true
```
danielnelson: I think we don't need an option, let's just do it unconditionally but at a level that won't be too annoying.

ehlerst: Removed

plugins/inputs/system/system.go
ehlerst: updated

ehlerst:
I think this may be gtg, i'm not sure how to make a Unit test for this since its calling a system file we probably don't want to fiddle with. Let me know.

PR #5743, opened 2019-04-18
add tag filter to consul plugin

GregoryDidenko:
### Required for all PRs:

### DIFF
--- a/plugins/inputs/consul/consul.go
+++ b/plugins/inputs/consul/consul.go
@@ -3,10 +3,12 @@ package consul
 import (
 	"net/http"
 	"strings"
+	"fmt"
 
 	"github.com/hashicorp/consul/api"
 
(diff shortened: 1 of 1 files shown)


glinton:
Seems useful, but what is the benefit of this over the [`taginclude`](https://github.com/influxdata/telegraf/blob/master/docs/CONFIGURATION.md#modifiers) metric filter available for all plugins?

GregoryDidenko:
this option ignores only tags and writes records without the tag. Does the taginclude metric filter ignore tags with records?

srebhan:
@HryhoriiDidenko any news on this?

srebhan:
I think this functionality can be achieved with `taginclude`. It also supports globbing, so I don't see the point in this PR. @HryhoriiDidenko if you have any use-case not covered by `taginclude`, please let me know!

PR #5749, opened 2019-04-20
Collect query params in HTTP listener input plugin

### LINKED ISSUE #5723: Collect query parameters in http listener input plugin
## Feature Request

### Proposal:

Add support for collecting query parameters in **http_listener_v2** input plugin

### Current behavior:

Currently the plugin only collects request body in the selected data format

### Desired behavior:

Add support to collect tags and values from query parameters when handling incoming requests

### Use case:

I have an IoT device setup to send measurements to a URL, appending the data in query parameters (meant to be used with [ThingSpeak API](https://www.mathworks.com/help/thingspeak/writedata.html)). The device can be configured to send requests to any URL, but the data format can't be changed.

Currently, I've written a simple proxy to read the query parameters and resend them JSON formatted to a telegraf listener, but it'll be more convenient to send them directly to telegraf.

byonchev:
### Required for all PRs:

Closes #5723


byonchev:
I don't think the failed test is related to the changes - can we restart the build?

danielnelson:
Not digging into the code yet, I'm concerned about how this design interacts with the `data_format` option when no body is included, as it essentially provides another data_format instead of using the existing system.

It also seems that it would change the way that parsing an empty body works, instead of an empty body producing no metrics it would produce a new metric with potentially a very different layout to the normal metrics.  I think one would either want it to create metrics purely from the query string, or create the metrics from the body augmenting with data from the query string, but I don't think one would want the mode to switch at runtime based on the body.

I would be more comfortable if this capability was removed, or perhaps if we had a `data_format` that parsed `application/x-www-form-urlencoded` data?  This would require the data to be included in the body of a POST request, but it would fit into the parser system smoothly.

byonchev:
Yes, this seems fair. Let's wrap it up - what do you think about 2 changes:
- A config param for the source (body/query/maybe headers in the future?)
- Parser for `application/x-www-form-urlencoded`

That way users will be able to specify single source of truth for the metrics and form data requests will also be working when the new parser + body source is used.

danielnelson:
There would still be a degree of incompatibility between the `data_format` and an option to select the data source.  What would happen if a user had `data_format = "influx"` and the query string as the source?

Looking at the docs for [thingspeak](https://www.mathworks.com/help/thingspeak/writedata.html), it should be possible to send the data in a `POST`.  What if we just did the parser?

byonchev:
Instead of using query values, we can use [URL's](https://golang.org/pkg/net/url/#URL) RawQuery field and feed it to the data parser. That way you can use whatever parser you want, the only requirement being the part after "?" in the URL is encoded.

For example with the influx parser - https://play.golang.org/p/OQ_-CI1klYD .

danielnelson:
That sounds reasonable.

byonchev:
I've pushed the changes we discussed.

byonchev:
@danielnelson A friendly ping for the review on this

plugins/inputs/http_listener_v2/http_listener_v2.go
```diff
@@ -208,19 +257,33 @@ func (h *HTTPListenerV2) serveWrite(res http.ResponseWriter, req *http.Request)
...
-		return
+		return nil, false
 	}
-	for _, m := range metrics {
-		h.acc.AddFields(m.Name(), m.Fields(), m.Tags(), m.Time())
+
+	return []byte(query), true
+}
+
+func appendToMetric(m telegraf.Metric, tags map[string]string, fields map[string]float64) {
```
danielnelson: I don't see where this function is used.

plugins/parsers/registry.go
```diff
@@ -141,6 +142,10 @@ type Config struct {
 	CSVTimestampColumn   string   `toml:"csv_timestamp_column"`
 	CSVTimestampFormat   string   `toml:"csv_timestamp_format"`
 	CSVTrimSpace         bool     `toml:"csv_trim_space"`
+
+	// FormData configuration
+	FormDataTagKeys     []string `toml:"form_data_tag_keys"`
+	FormDataAllowedKeys []string `toml:"form_data_allowed_keys"`
```
danielnelson: I don't think we will need the `form_data_allowed_keys` option, one could use the fielddrop/fielddrop options instead.

danielnelson:
@byonchev I'm going to rename the parser `form_urlencoded` to differentiate it from `multipart/form-data`

PR #5754, opened 2019-04-23
procstat err if no pids found

arindamchoudhury:
### Required for all PRs:

### DIFF
--- a/plugins/inputs/procstat/procstat.go
+++ b/plugins/inputs/procstat/procstat.go
@@ -27,7 +27,6 @@ type Procstat struct {
 	Exe         string
 	Pattern     string
 	Prefix      string
-	CmdLineTag  bool `toml:"cmdline_tag"`
 	ProcessName string
 	User        string
 	SystemdUnit string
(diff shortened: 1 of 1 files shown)


danielnelson:
So this would change the procstat_lookup measurement to have a `lookup_error`?  I don't think this is wanted, the lookup succeeded with no matches.

arindamchoudhury:
this helps to notify when a process dies.

danielnelson:
What about the current `procstat_lookup` behavior, does it not work for you, why not?
```
procstat_lookup,exe=foo,pid_finder=pgrep,process_name=bar,result=success pid_count=0i,result_code=0i,running=0i 1556136150000000000
```

arindamchoudhury:
now, it does not send any data if pid is not found. so we can not make a query to check status of a process.

danielnelson:
Can you show your configuration and how to reproduce?  It should always emit the `procstat_lookup` measurement.

danielnelson:
I'm closing but we can discuss further, want to make sure we have a solution that works for you.

Issue #5746, opened 2019-04-18
Plugin: procstat: PidFile doesn't verify process is still running


rawkode:
### Relevant telegraf.conf:

```toml
[[inputs.procstat]]
  pid_file = "/tmp/mongodb.pid"
```

### System info:

Linux 5.x

### Steps to reproduce:

1. Run mongod --pidfile=/tmp/mongodb.pid
2. Kill -9 <mongd pid>

### Expected behavior:

State would change to `running=0i`

### Actual behavior:

State continues to be `running=1i`

glinton:
What telegraf version?

Ritesh-Sharma:
Hey@glinton, mentioned issue can be reproducible with below version.

$ telegraf version 
Telegraf 1.10.3 (git: HEAD 294bb666)

danielnelson:
Closed in #5768

PR #5768, opened 2019-04-25
Check if a process is running using it's name

fuegas:
### Required for all PRs:

This PR relates to #5746. When you provide a `pid_file` that exists, has a (int32) number in it of a non-existing process, it will report the process as running right from the start of the agent:
```
> procstat_lookup,host=Kita.local,pid_finder=pgrep,pidfile=/tmp/process.pid,result=success pid_count=1i,result_code=0i,running=1i 1556141065000000000
```

A process is marked running if the [procs map has at least one item](https://github.com/influxdata/telegraf/blob/master/plugins/inputs/procstat/procstat.go#L137):
```golang
"running":     len(procs),
```
This map is build/filled by [updateProcesses](https://github.com/influxdata/telegraf/blob/master/plugins/inputs/procstat/procstat.go#L284):
```golang
info, ok := prevInfo[pid]
if ok {
  procs[pid] = info
}
...
proc, err := p.createProcess(pid)
if err != nil {
  // No problem; process may have ended after we found it
  continue
}
procs[pid] = proc
```
Because the PID is specified by a `pid_file` we have a valid `PID` (`int32`). This results in a valid `Process` from `p.createProcess(pid)`, regardless whether the process exists or not. This `Process` is inserted into the `procs` map resulting in a map that has 1 or more entries. So if you provide a `pid_file` that has a number in it, `procstat` will _always_ return `running=1i`.

In this PR I assume processes without name are rare and the chance that you want to monitor a process with no name is even rarer, a process without (or empty) name is seen as not running.

### DIFF
--- a/plugins/inputs/procstat/procstat.go
+++ b/plugins/inputs/procstat/procstat.go
@@ -287,11 +287,19 @@ func (p *Procstat) updateProcesses(pids []PID, tags map[string]string, prevInfo 
 	for _, pid := range pids {
 		info, ok := prevInfo[pid]
 		if ok {
+			// Assumption: if a process has no name, it probably does not exist
+			if name, _ := info.Name(); name == "" {
+				continue
+			}
 			procs[pid] = info
 		} else {
 			proc, err := p.createProcess(pid)
 			if err != nil {
 				// No problem; process may have ended after we found it
+				continue
+			}
+			// Assumption: if a process has no name, it probably does not exist
+			if name, _ := proc.Name(); name == "" {
 				continue
 			}
 			procs[pid] = proc


sgtsquiggs:
Is this really necessary?

`updateProcesses` depends on gopsutil's `NewProcess` to check if the process exists, which achieves this by attempting to open `/proc/<pid>`.

fuegas:
That is true for the [Linux](https://github.com/shirou/gopsutil/blob/master/process/process_linux.go#L72) adapter. However, on [Darwin](https://github.com/shirou/gopsutil/blob/master/process/process_darwin.go#L646), [FreeBSD](https://github.com/shirou/gopsutil/blob/master/process/process_freebsd.go#L514), [OpenBSD](https://github.com/shirou/gopsutil/blob/master/process/process_openbsd.go#L499) and [Windows](https://github.com/shirou/gopsutil/blob/master/process/process_windows.go#L625) the process is only created as a struct and no checks are done. So on those platforms the process from a pidfile always shows `running=1i` regardless of the actual state.

sgtsquiggs:
That is unfortunate. For linux/bsd platforms we would confirm that the process exists via
```go
err := process.Signal(syscall.Signal(0))
```
assuming process is type `os.Process`.

For windows, `os.FindProcess` actually checks to see if the process exists.

plugins/inputs/procstat/procstat.go
```diff
@@ -287,13 +287,21 @@ func (p *Procstat) updateProcesses(pids []PID, tags map[string]string, prevInfo
 	for _, pid := range pids {
 		info, ok := prevInfo[pid]
 		if ok {
+			// Assumption: if a process has no name, it probably does not exist
+			if name, _ := info.Name(); name == "" {
```
danielnelson: Can we use `err != nil` here instead?  I feel like this would be safer than assuming a process cannot have a name.

fuegas: I've changed the lines below to:
```golang
// Assumption: if a process has no name, it probably does not exist
name, err := proc.Name()
if err != nil {
	fmt.Println("proc.Name() gave error: ", err)
}
if name == "" {
	fmt.Println("proc.Name() was empty, but no error!")
	continue
}
procs[pid] = proc
```
It results in the following output:
```
┌─[Kita][⚡ 93%][10:58:49] in ~/Projects/go/src/github.com/influxdata/telegraf
└─⎇ verify-pidfile-process-exists U:1 ?:1 ✗ ⋗ ./telegraf --config telegraf.conf --test
2019-04-26T08:58:59Z I! Starting Telegraf
proc.Name() was empty, but no error!
proc.Name() was empty, but no error!
> procstat_lookup,host=Kita.local,pid_finder=pgrep,pidfile=/tmp/process.pid,result=success pid_count=1i,result_code=0i,running=0i 1556269140000000000
```
So we do not receive an error, even when the process does not exist.

danielnelson: Which operating system is that with?

fuegas: This is on Darwin.

danielnelson: I wonder if this could be a bug in the Darwin implementation of gopsutil.  I tried this on FreeBSD, which has a very similar implementation, and I did get a "no such process" `syscall.Errno` error.  Can you verify that the sysctl syscall is returning no error?

https://github.com/shirou/gopsutil/blob/master/process/process_darwin.go#L627

For comparision, FreeBSD call is here:

https://github.com/shirou/gopsutil/blob/master/internal/common/common_freebsd.go#L54

fuegas: I get the following output when I add `fmt.Println("Syserr is: ", syserr)` to `process_darwin.go`:
```
Syserr is:  errno 0
proc.Name() was empty, but no error!
```
So it does not throw an error...

danielnelson: Lets go with what we have here then.

Issue #1634, opened 2016-08-16
ntp plugin crash


gmr:
A parse error in the ntp plugin caused a telegraf agent to crash.

### Relevant telegraf.conf:
```
[[inputs.ntpq]]
#   ## If false, set the -n ntpq flag. Can reduce metric gather time.
#   dns_lookup = true
```
### System info:

Telegraf - version 0.13.1
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=14.04
DISTRIB_CODENAME=trusty
DISTRIB_DESCRIPTION="Ubuntu 14.04.5 LTS"

### Steps to reproduce:

I'm guessing an unexpected response from the call the agent makes to grab NTP info causes the crash to happen.

### Expected behavior:

Agent doesn't crash

### Actual behavior:

Agent Crashes

### Additional info:

```
2016/08/15 09:46:00 ERROR ntpq: parsing int: 77.759
panic: runtime error: index out of range

goroutine 2202595 [running]:
panic(0x107ebc0, 0xc820010080)
	/usr/local/go/src/runtime/panic.go:481 +0x3e6
github.com/influxdata/telegraf/plugins/inputs/ntpq.(*NTPQ).Gather(0xc8204da660, 0x7f5ecebaa368, 0xc82078c000, 0x0, 0x0)
	/home/ubuntu/telegraf-build/src/github.com/influxdata/telegraf/plugins/inputs/ntpq/ntpq.go:175 +0xc90
github.com/influxdata/telegraf/agent.gatherWithTimeout.func1(0xc820558120, 0xc8201d0c60, 0xc82078c000)
	/home/ubuntu/telegraf-build/src/github.com/influxdata/telegraf/agent/agent.go:174 +0x73
created by github.com/influxdata/telegraf/agent.gatherWithTimeout
	/home/ubuntu/telegraf-build/src/github.com/influxdata/telegraf/agent/agent.go:175 +0xe0
```

sparrc:
thanks for the report, I was able to reproduce in a unit test and I'll have a fix up in master shortly

PR #5782, opened 2019-04-30
Skip lines with missing refid and other fields

danielnelson:
Fix for report of lines with missing refid:
```
+10.0.0.1 16 u 1048 1024 376 15.384 -0.538 0.943
```

I believe this is the actual cause of #1634 based on the error message, which is why I'm removing the test case created for that issue.  The panic from that issue would also no longer occur due to other changes to the plugin since then.

### Required for all PRs:


danielnelson:
The missing refid is believed to be the issue described in http://bugs.ntp.org/show_bug.cgi?id=3484 and should be fixed in ntp-4.2.8p12, I added a comment to the unit test with this information.

Issue #5224, opened 2019-01-03
Add support for dogstatsd events


russorat:
### Proposal:
the current statsd input plugin has the ability to parse data dog tags. dogstatsd clients also have the ability to send events in a specific datadog format. https://docs.datadoghq.com/developers/dogstatsd/data_types/#events

This would either add the ability to parse those specific datadog events in the current statsd plugin, or move everything into a specific dogstatsd input plugin.

### Current behavior:
datadog events are not parsed correctly from the statsd input plugin

### Desired behavior:
either the current statsd plugin accepts and parses those events, or a new plugin is created specifically for dogstatsd

### Use case: [Why is this important (helps with prioritizing requests)]
this is helpful for customers that want to send data to multiple monitoring solutions without having to change their existing application code using dogstatsd client code.

danielnelson:
From on the [Events API documenation](https://docs.datadoghq.com/api/?lang=bash#post-an-event), the example event is:
```
{
      "title": "Did you hear the news today?",
      "text": "Oh boy!",
      "priority": "normal",
      "tags": ["environment:test"],
      "alert_type": "info"
}
```

This could be received using the `http_listener_v2` input configured with the json parser and to collect all the string fields.  This could then be sent to InfluxDB or other outputs that support a string type, so maybe we just need to document a configuration example.

If we wanted to support send this type of metric to Datadog we would need a Datadog Events output.

russorat:
@danielnelson i think the format for an event sent by dogstatsd client is slightly different. See the java client code here: https://github.com/DataDog/java-dogstatsd-client/blob/master/src/main/java/com/timgroup/statsd/NonBlockingStatsDClient.java#L921

danielnelson:
Got it.  I couldn't find any specification for this format but this appears to be the parser for Datadog events in statsd extension: https://github.com/DataDog/datadog-agent/blob/master/pkg/dogstatsd/parser.go#L152

schahal:
@danielnelson, exactly! that's the parser of the [Events datagram](https://docs.datadoghq.com/developers/dogstatsd/data_types/#events) :) I submitted this PR so that those functions become exportable: https://github.com/DataDog/datadog-agent/pull/2839

In the meantime, I added this PR on my own fork (I don't want to PR that branch to influxdata/telegraf:master until the exportable functions are merged by DD above and dependencies can be resolved). 

That is, in final commit, we can change:
```
import (
    ...
[...]

PR #5791, opened 2019-05-01
Statsd datadog

docmerlin:
closes https://github.com/influxdata/telegraf/issues/5224

This adds Datadog events to the Statsd input plugin.

### DIFF
--- a/plugins/inputs/statsd/README.md
+++ b/plugins/inputs/statsd/README.md
@@ -34,7 +34,6 @@
   ## Reset timings & histograms every interval (default=true)
   delete_timings = true
 
-
   ## Percentiles to calculate for timing & histogram stats
   percentiles = [90]
 
@@ -183,7 +182,6 @@ to allow. Used when protocol is set to tcp.
 - **delete_counters** boolean: Delete counters on every collection interval
 - **delete_sets** boolean: Delete set counters on every collection interval
 - **delete_timings** boolean: Delete timings on every collection interval
-
 - **percentiles** []int: Percentiles to calculate for timing & histogram stats
 - **allowed_pending_messages** integer: Number of messages allowed to queue up
 waiting to be processed. When this fills, messages will be dropped and logged.
--- a/plugins/inputs/statsd/README.md
+++ b/plugins/inputs/statsd/README.md
@@ -43,6 +43,7 @@
   ## Parses tags in the datadog statsd format
   ## http://docs.datadoghq.com/guides/dogstatsd/
   parse_data_dog_tags = false
+  parse_data_dog_events = false
 
   ## Statsd data translation templates, more info can be read here:
   ## https://github.com/influxdata/telegraf/blob/master/docs/TEMPLATE_PATTERN.md
@@ -185,6 +186,7 @@ the accuracy of percentiles but also increases the memory usage and cpu time.
 - **templates** []string: Templates for transforming statsd buckets into influx
 measurements and tags.
 - **parse_data_dog_tags** boolean: Enable parsing of tags in DataDog's dogstatsd format (http://docs.datadoghq.com/guides/dogstatsd/)
+- **parse_data_dog_events** boolean: Enable parsing of events in DataDog's dogstatsd format (http://docs.datadoghq.com/guides/dogstatsd/)
 
 ### Statsd bucket -> InfluxDB line-protocol Templates
 
--- a/plugins/inputs/statsd/README.md
+++ b/plugins/inputs/statsd/README.md
@@ -193,7 +193,7 @@ the accuracy of percentiles but also increases the memory usage and cpu time.
 - **templates** []string: Templates for transforming statsd buckets into influx
 measurements and tags.
 - **parse_data_dog_tags** boolean: Enable parsing of tags in DataDog's dogstatsd format (http://docs.datadoghq.com/guides/dogstatsd/)
-- **parse_data_dog_events** boolean: Enable parsing of events in DataDog's dogstatsd format (http://docs.datadoghq.com/guides/dogstatsd/)
+- **datadog_extensions** boolean: Enable parsing of DataDog's extensions to dogstatsd format (http://docs.datadoghq.com/guides/dogstatsd/)
 
 ### Statsd bucket -> InfluxDB line-protocol Templates
 
--- a/plugins/inputs/statsd/README.md
+++ b/plugins/inputs/statsd/README.md
@@ -34,6 +34,7 @@
   ## Reset timings & histograms every interval (default=true)
   delete_timings = true
 
+
   ## Percentiles to calculate for timing & histogram stats
   percentiles = [90]
 
@@ -42,8 +43,13 @@
 
   ## Parses tags in the datadog statsd format
   ## http://docs.datadoghq.com/guides/dogstatsd/
+  ## deprecated in 1.10; use datadog_extensions option instead
   parse_data_dog_tags = false
-  parse_data_dog_events = false
+
+  ## Parses extensions to statsd in the datadog statsd format
+  ## currently supports metrics and datadog tags.
+  ## http://docs.datadoghq.com/guides/dogstatsd/
+  datadog_extensions = false
 
   ## Statsd data translation templates, more info can be read here:
   ## https://github.com/influxdata/telegraf/blob/master/docs/TEMPLATE_PATTERN.md
@@ -177,6 +183,7 @@ to allow. Used when protocol is set to tcp.
 - **delete_counters** boolean: Delete counters on every collection interval
 - **delete_sets** boolean: Delete set counters on every collection interval
 - **delete_timings** boolean: Delete timings on every collection interval
+
 - **percentiles** []int: Percentiles to calculate for timing & histogram stats
 - **allowed_pending_messages** integer: Number of messages allowed to queue up
 waiting to be processed. When this fills, messages will be dropped and logged.
--- a/plugins/inputs/statsd/datadog.go
+++ b/plugins/inputs/statsd/datadog.go
@@ -1,13 +1,14 @@
 package statsd
 
+// this is adapted from datadog's apache licensed version at
+// https://github.com/DataDog/datadog-agent/blob/fcfc74f106ab1bd6991dfc6a7061c558d934158a/pkg/dogstatsd/parser.go#L173
+
 import (
 	"errors"
 	"fmt"
 	"strconv"
 	"strings"
 	"time"
-
-	tmetric "github.com/influxdata/telegraf/metric"
 )
 
 const (
@@ -22,8 +23,6 @@ const (
 
 var uncommenter = strings.NewReplacer("\\n", "\n")
 
-// this is adapted from datadog's apache licensed version at
-// https://github.com/DataDog/datadog-agent/blob/fcfc74f106ab1bd6991dfc6a7061c558d934158a/pkg/dogstatsd/parser.go#L173
 func (s *Statsd) parseEventMessage(now time.Time, message string, defaultHostname string) error {
 	// _e{title.length,text.length}:title|text
 	//  [
@@ -53,7 +52,9 @@ func (s *Statsd) parseEventMessage(now time.Time, message string, defaultHostnam
 	if err != nil {
 		return fmt.Errorf("Invalid message format, could not parse title.length: '%s'", rawLen[0])
 	}
-
+	if len(rawLen[1]) < 1 {
+		return fmt.Errorf("Invalid message format, could not parse text.length: '%s'", rawLen[0])
+	}
 	textLen, err := strconv.ParseInt(rawLen[1][:len(rawLen[1])-1], 10, 64)
 	if err != nil {
 		return fmt.Errorf("Invalid message format, could not parse text.length: '%s'", rawLen[0])
@@ -70,28 +71,16 @@ func (s *Statsd) parseEventMessage(now time.Time, message string, defaultHostnam
 		return fmt.Errorf("Invalid event message format: empty 'title' or 'text' field")
 	}
 
-	// Handle hostname, with a priority to the h: field, then the host:
-	// tag and finally the defaultHostname value
-	// Metadata
-	m := event{
-		name: rawTitle,
-	}
-	m.tags = make(map[string]string, strings.Count(message, ",")+2) // allocate for the approximate number of tags
-	m.fields = make(map[string]interface{}, 9)
-	m.fields["alert_type"] = eventInfo // default event type
-	m.fields["text"] = uncommenter.Replace(string(rawText))
-	// host is a magic tag in the system, and it expects it to replace the result of h: if it is present
-	// telegraf will add a"host" tag anyway with different meaning than dogstatsd, so we need to use source instead of host.
-
-	m.tags["source"] = defaultHostname
-	m.fields["priority"] = priorityNormal
-	m.ts = now
+	name := rawTitle
+	tags := make(map[string]string, strings.Count(message, ",")+2) // allocate for the approximate number of tags
+	fields := make(map[string]interface{}, 9)
+	fields["alert_type"] = eventInfo // default event type
+	fields["text"] = uncommenter.Replace(string(rawText))
+	tags["source"] = defaultHostname // Use source tag because host is reserved tag key in Telegraf.
+	fields["priority"] = priorityNormal
+	ts := now
 	if len(message) < 2 {
-		newM, err := tmetric.New(m.name, m.tags, m.fields, m.ts)
-		if err != nil {
-			return err
-		}
-		s.acc.AddMetric(newM)
+		s.acc.AddFields(name, fields, tags, ts)
 		return nil
 	}
 
@@ -106,45 +95,43 @@ func (s *Statsd) parseEventMessage(now time.Time, message string, defaultHostnam
 			if err != nil {
 				continue
 			}
-			m.fields["ts"] = ts
+			fields["ts"] = ts
 		case "p:":
 			switch rawMetadataFields[i][2:] {
 			case priorityLow:
-				m.fields["priority"] = priorityLow
+				fields["priority"] = priorityLow
 			case priorityNormal: // we already used this as a default
 			default:
 				continue
 			}
 		case "h:":
-			m.tags["source"] = rawMetadataFields[i][2:]
+			tags["source"] = rawMetadataFields[i][2:]
 		case "t:":
 			switch rawMetadataFields[i][2:] {
 			case eventError, eventWarning, eventSuccess, eventInfo:
-				m.fields["alert_type"] = rawMetadataFields[i][2:] // already set for info
+				fields["alert_type"] = rawMetadataFields[i][2:] // already set for info
 			default:
 				continue
 			}
 		case "k:":
-			m.tags["aggregation_key"] = rawMetadataFields[i][2:]
+			tags["aggregation_key"] = rawMetadataFields[i][2:]
 		case "s:":
-			m.fields["source_type_name"] = rawMetadataFields[i][2:]
+			fields["source_type_name"] = rawMetadataFields[i][2:]
 		default:
 			if rawMetadataFields[i][0] == '#' {
-				parseDataDogTags(m.tags, rawMetadataFields[i][1:])
+				parseDataDogTags(tags, rawMetadataFields[i][1:])
 			} else {
 				return fmt.Errorf("unknown metadata type: '%s'", rawMetadataFields[i])
 			}
 		}
 	}
-	if host, ok := m.tags["host"]; ok {
-		delete(m.tags, "host")
-		m.tags["source"] = host
+	// Use source tag because host is reserved tag key in Telegraf.
+	// In datadog the host tag and `h:` are interchangable, so we have to chech for the host tag.
+	if host, ok := tags["host"]; ok {
+		delete(tags, "host")
+		tags["source"] = host
 	}
-	newM, err := tmetric.New(m.name, m.tags, m.fields, m.ts)
-	if err != nil {
-		return err
-	}
-	s.acc.AddMetric(newM)
+	s.acc.AddFields(name, fields, tags, ts)
 	return nil
 }
 
--- a/plugins/inputs/statsd/datadog.go
+++ b/plugins/inputs/statsd/datadog.go
@@ -1,10 +1,10 @@
 package statsd
 
 import (
-	"bytes"
 	"fmt"
 	"log"
 	"strconv"
+	"strings"
 	"time"
 )
 
@@ -15,7 +15,7 @@ const (
 
 // this is adapted from datadog's apache licensed version at
 // https://github.com/DataDog/datadog-agent/blob/fcfc74f106ab1bd6991dfc6a7061c558d934158a/pkg/dogstatsd/parser.go#L173
-func (s *Statsd) parseDataDogEventMessage(message []byte, defaultHostname string) error {
+func (s *Statsd) parseDataDogEventMessage(now time.Time, message string, defaultHostname string) error {
 	// _e{title.length,text.length}:title|text
 	//  [
 	//   |d:date_happened
@@ -25,25 +25,27 @@ func (s *Statsd) parseDataDogEventMessage(message []byte, defaultHostname string
 	//   |s:source_type_nam
 	//   |#tag1,tag2
 	//  ]
-
-	messageRaw := bytes.SplitN(message, []byte(":"), 2)
+	//
+	//
+	// tag is key:value
+	messageRaw := strings.SplitN(message, ":", 2)
 	if len(messageRaw) < 2 || len(messageRaw[0]) < 7 || len(messageRaw[1]) < 3 {
 		return fmt.Errorf("Invalid message format")
 	}
 	header := messageRaw[0]
 	message = messageRaw[1]
 
-	rawLen := bytes.SplitN(header[3:], []byte(","), 2)
+	rawLen := strings.SplitN(header[3:], ",", 2)
 	if len(rawLen) != 2 {
 		return fmt.Errorf("Invalid message format")
 	}
 
-	titleLen, err := strconv.ParseInt(string(rawLen[0]), 10, 64)
+	titleLen, err := strconv.ParseInt(rawLen[0], 10, 64)
 	if err != nil {
 		return fmt.Errorf("Invalid message format, could not parse title.length: '%s'", rawLen[0])
 	}
 
-	textLen, err := strconv.ParseInt(string(rawLen[1][:len(rawLen[1])-1]), 10, 64)
+	textLen, err := strconv.ParseInt(rawLen[1][:len(rawLen[1])-1], 10, 64)
 	if err != nil {
 		return fmt.Errorf("Invalid message format, could not parse text.length: '%s'", rawLen[0])
 	}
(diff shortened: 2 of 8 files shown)


plugins/inputs/statsd/datadog.go
```diff
@@ -0,0 +1,173 @@
...
+				switch rawMetadataFields[i][2:] {
+				case priorityLow:
+					m.fields["priority"] = priorityLow
+				case priorityNormal: // we already used this as a default
+				default:
+					log.Printf("W! [inputs.statsd] skipping priority")
+					continue
+				}
+			case "h:":
+				m.tags["source"] = rawMetadataFields[i][2:]
```
goller: @danielnelson What's typical for a source host in telegraf?

@docmerlin was "source" used in other places too?

plugins/inputs/statsd/datadog.go
```diff
@@ -0,0 +1,173 @@
...
+					m.fields["alert-type"] = "warning"
+				case "success":
+					m.fields["alert-type"] = "success"
+				case "info": // already set for info
+				default:
+					log.Printf("W! [inputs.statsd] skipping alert type")
+					continue
+				}
+			case "k:":
+				// TODO(docmerlin): does this make sense?
```
goller: @docmerlin are you thinking that this may not need to be included?

plugins/inputs/statsd/datadog.go
```diff
@@ -0,0 +1,173 @@
...
+				m.tags["aggregation-key"] = rawMetadataFields[i][2:]
+			case "s:":
+				m.fields["source-type-name"] = rawMetadataFields[i][2:]
+			default:
+				if rawMetadataFields[i][0] == '#' {
+					parseDataDogTags(m.tags, rawMetadataFields[i][1:])
+				} else {
+					log.Printf("W! [inputs.statsd] unknown metadata type: '%s'", rawMetadataFields[i])
```
goller: @danielnelson so, the original source code would optimistically continue parsing and warn about bad values within the even message. 

Does this fit telegraf?  Should these warnings be errors or are logs ok?

plugins/inputs/statsd/datadog.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+	m.tags = make(map[string]string, strings.Count(message, ",")+2) // allocate for the approximate number of tags
+	m.fields = make(map[string]interface{}, 9)
+	m.fields["alert-type"] = "info" // default event type
+	m.fields["text"] = uncommenter.Replace(string(rawText))
+	m.tags["source"] = defaultHostname
+	m.fields["priority"] = priorityNormal
+	m.ts = now
+	if len(message) == 0 {
+		s.events = append(s.events, m)
```
goller: You lock below before manipulating s.events; do you need to here as well?

plugins/inputs/statsd/datadog.go
```diff
@@ -0,0 +1,173 @@
...
+
+func parseDataDogTags(tags map[string]string, message string) {
+	start, i := 0, 0
+	var k string
+	var inVal bool // check if we are parsing the value part of the tag
+	for i = range message {
+		if message[i] == ',' {
+			if k == "" {
+				k = message[start:i]
+				tags[k] = ""
```
goller: I've done empty string tags before in the zipkin plugin, however, it makes it pretty difficult to query for at least with influxdb.

@danielnelson how are label type tags typically done in telegraf?

plugins/inputs/statsd/datadog.go
```diff
@@ -0,0 +1,173 @@
...
+	"strings"
+	"time"
+)
+
+const (
+	priorityNormal = "normal"
+	priorityLow    = "low"
+)
+
+var uncommenter = strings.NewReplacer("\\n", "\n")
```
goller: @danielnelson in datadog events newlines are encoded as \\n (literally 2 characters).  Should this be transformed into line feed (0x0A) for the tag?

plugins/inputs/statsd/datadog.go
```diff
@@ -0,0 +1,173 @@
...
+
+	// Handle hostname, with a priority to the h: field, then the host:
+	// tag and finally the defaultHostname value
+	// Metadata
+	m := cachedEvent{
+		name: rawTitle,
+	}
+	m.tags = make(map[string]string, strings.Count(message, ",")+2) // allocate for the approximate number of tags
+	m.fields = make(map[string]interface{}, 9)
+	m.fields["alert-type"] = "info" // default event type
```
goller: all fields and tag names should be snake case I believe.

plugins/inputs/statsd/datadog.go
```diff
@@ -0,0 +1,173 @@
...
+				default:
+					log.Printf("W! [inputs.statsd] skipping priority")
+					continue
+				}
+			case "h:":
+				m.tags["source"] = rawMetadataFields[i][2:]
+			case "t:":
+				switch rawMetadataFields[i][2:] {
+				case "error":
+					m.fields["alert-type"] = "error"
```
goller: snake

plugins/inputs/statsd/datadog.go
```diff
@@ -0,0 +1,173 @@
...
+				case "success":
+					m.fields["alert-type"] = "success"
+				case "info": // already set for info
+				default:
+					log.Printf("W! [inputs.statsd] skipping alert type")
+					continue
+				}
+			case "k:":
+				// TODO(docmerlin): does this make sense?
+				m.tags["aggregation-key"] = rawMetadataFields[i][2:]
```
goller: snake

plugins/inputs/statsd/datadog.go
```diff
@@ -0,0 +1,173 @@
...
+			case "s:":
+				m.fields["source-type-name"] = rawMetadataFields[i][2:]
```
goller: snake

plugins/inputs/statsd/datadog.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+	s.Lock()
+	s.events = append(s.events, m)
+	s.Unlock()
+	return nil
+}
+
+func parseDataDogTags(tags map[string]string, message string) {
+	start, i := 0, 0
+	var k string
```
goller: @danielnelson because the tag key names are being dynamically created here, should they be converted to snake case?  Do the output plugins do this ?

plugins/inputs/statsd/README.md
```diff
@@ -43,6 +43,7 @@
   ## Parses tags in the datadog statsd format
   ## http://docs.datadoghq.com/guides/dogstatsd/
   parse_data_dog_tags = false
+  parse_data_dog_events = false
```
goller: Probably a comment in the same way as the dog_tags.

plugins/inputs/statsd/README.md
```diff
@@ -43,6 +43,7 @@
   ## Parses tags in the datadog statsd format
```
goller: You'll need to document delete_events parameter as well.

plugins/inputs/statsd/README.md
goller: also... do you think it is possible to document the events translation in the readme?

plugins/inputs/statsd/datadog.go
glinton: [This is the discussion](https://github.com/influxdata/telegraf/issues/4413) on standardizing what `source` should be set to when used as a tag.

plugins/inputs/statsd/datadog.go
glinton: As far as I know, the outputs do not do this

plugins/inputs/statsd/datadog.go
glinton: What is the value of having an empty tag?

plugins/inputs/statsd/datadog.go
```diff
@@ -0,0 +1,173 @@
...
+					continue
+				}
+				m.fields["ts"] = ts
+			case "p:":
+				switch rawMetadataFields[i][2:] {
+				case priorityLow:
+					m.fields["priority"] = priorityLow
+				case priorityNormal: // we already used this as a default
+				default:
+					log.Printf("W! [inputs.statsd] skipping priority")
```
glinton: Remove or make this a debug log.

plugins/inputs/statsd/datadog.go
```diff
@@ -0,0 +1,173 @@
...
+			default:
+				if rawMetadataFields[i][0] == '#' {
+					parseDataDogTags(m.tags, rawMetadataFields[i][1:])
+				} else {
+					log.Printf("W! [inputs.statsd] unknown metadata type: '%s'", rawMetadataFields[i])
+				}
+			}
+		}
+	}
+	// host is a magic tag in the system, and it expects it to replace the result of h: if it is present
```
glinton: Based on the comments above (line 66) shouldn't this be below line 76 but above 108?

plugins/inputs/statsd/datadog.go
```diff
@@ -0,0 +1,173 @@
...
+	m.fields["text"] = uncommenter.Replace(string(rawText))
+	m.tags["source"] = defaultHostname
+	m.fields["priority"] = priorityNormal
+	m.ts = now
+	if len(message) == 0 {
+		s.events = append(s.events, m)
+		return nil
+	}
+
+	if len(message) > 1 {
```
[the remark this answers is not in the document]
docmerlin: If len(message)==1 then it is an invalid message, I see what you are saying though.  I will simplify it and catch that case.

plugins/inputs/statsd/datadog.go
docmerlin: I don't know but Datadog events tests in their agent specifically makes allowances for them.  We wouldn't have spec compatibility if we didn't allow them.  Should we handle them some other way?

plugins/inputs/statsd/statsd.go
```diff
@@ -468,12 +496,16 @@ func (s *Statsd) parser() error {
...
+		case in := <-s.in:
+			lines := strings.Split(in.Buffer.String(), "\n")
+			s.bufPool.Put(in.Buffer)
 			for _, line := range lines {
 				line = strings.TrimSpace(line)
-				if line != "" {
+				switch {
+				case line == "":
+				case s.ParseDataDogEvents && len(line) > 2 && line[:2] == "_e":
+					s.parseEventMessage(time.Now(), line, in.Addr.String())
```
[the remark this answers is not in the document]
docmerlin: Good point!

plugins/inputs/statsd/datadog.go
danielnelson: I think we will want to return the event, then add it to the accumulator passed into Start(acc), use metric.New() to create the event and acc.AddMetric() to add it.  No need to hang on to it until Gather is called because events are not aggregated like normal statsd messages.

plugins/inputs/statsd/datadog.go
danielnelson: I think if the message is malformed an error should be returned up from here and then logged.  Doing a best effort parse doesn't make sense unless we find that the datadog agent regularly emits bad messages.

plugins/inputs/statsd/datadog.go
danielnelson: I don't think it is allowed to have an empty tag value in InfluxDB, we just toss them at serialize time.

plugins/inputs/statsd/statsd.go
```diff
@@ -61,6 +62,7 @@ type Statsd struct {
 	DeleteCounters bool
 	DeleteSets     bool
 	DeleteTimings  bool
+	DeleteEvents   bool
```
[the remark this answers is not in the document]
danielnelson: This field is not needed because there is no aggregation of events.

plugins/inputs/statsd/datadog_test.go
```diff
@@ -0,0 +1,353 @@
...
+func TestEventGather(t *testing.T) {
+	acc := &testutil.Accumulator{}
+
+	now := time.Now()
+	s := NewTestStatsd()
+	err := s.parseEventMessage(now, "_e{10,9}:test title|test text", "default-hostname")
+	require.Nil(t, err)
+	err = s.parseEventMessage(now.Add(1), "_e{10,24}:test title|test\\line1\\nline2\\nline3", "default-hostname")
+	require.Nil(t, err)
+	err = s.parseEventMessage(now.Add(2), "_e{10,9}:test title|test text|d:21", "default-hostname")
```
danielnelson: This, and a few other places in this file, would be a great place to use a table test.  Please also use require.Equal instead of assert.Equal more or less everywhere.

plugins/inputs/statsd/statsd.go
```diff
@@ -468,12 +496,16 @@ func (s *Statsd) parser() error {
...
-			s.bufPool.Put(buf)
+		case in := <-s.in:
+			lines := strings.Split(in.Buffer.String(), "\n")
+			s.bufPool.Put(in.Buffer)
 			for _, line := range lines {
 				line = strings.TrimSpace(line)
-				if line != "" {
+				switch {
+				case line == "":
+				case s.ParseDataDogEvents && len(line) > 2 && line[:2] == "_e":
```
danielnelson: `strings.HasPrefix(`

plugins/inputs/statsd/statsd.go
```diff
@@ -831,7 +845,7 @@ func (s *Statsd) handler(conn *net.TCPConn, id string) {
 			b.WriteByte('\n')
 
 			select {
-			case s.in <- b:
+			case s.in <- input{Buffer: b, Time: time.Now(), Addr: conn.RemoteAddr()}:
```
danielnelson: On the source tag, it should only have the hostname/ip without the port.  I think this does host:port, so we will need to remove the port.

plugins/inputs/statsd/README.md
danielnelson: For enabling this feature, lets add a new field `datadog_extensions = false` which will control both of these, then let's deprecate the old option but if it is true we will set the new option true as well.  The new option will control both of these settings.

https://github.com/influxdata/telegraf/wiki/Deprecation#deprecate-options

plugins/inputs/statsd/datadog.go
```diff
@@ -0,0 +1,173 @@
...
+
+const (
+	priorityNormal = "normal"
+	priorityLow    = "low"
+)
+
+var uncommenter = strings.NewReplacer("\\n", "\n")
+
+// this is adapted from datadog's apache licensed version at
+// https://github.com/DataDog/datadog-agent/blob/fcfc74f106ab1bd6991dfc6a7061c558d934158a/pkg/dogstatsd/parser.go#L173
```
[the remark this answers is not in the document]
danielnelson: I think move the comment to the very top of the file and also add it to LICENSE_OF_DEPENDENCIES.  We also need to handle the NOTICE file, but I'll work this out as I need to look over the existing apache 2 libraries.

plugins/inputs/statsd/datadog.go
docmerlin: Telegraf generally tries to ignore errors and keep working where it can.

plugins/inputs/statsd/datadog.go
docmerlin: I was not sure if it should be a tag or field.  Talked it over and it should be a tag.

plugins/inputs/statsd/datadog_test.go
docmerlin: The reason I didn't use a table test was to make it more similar to their tests.  I wanted to ensure compatibility so I copied their tests over then converted them to our setup.

docmerlin: I could switch, them over now though.

plugins/inputs/statsd/datadog.go
```diff
@@ -0,0 +1,173 @@
...
+	if len(rawLen) != 2 {
+		return fmt.Errorf("Invalid message format")
+	}
+
+	titleLen, err := strconv.ParseInt(rawLen[0], 10, 64)
+	if err != nil {
+		return fmt.Errorf("Invalid message format, could not parse title.length: '%s'", rawLen[0])
+	}
+
+	textLen, err := strconv.ParseInt(rawLen[1][:len(rawLen[1])-1], 10, 64)
```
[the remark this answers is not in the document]
docmerlin: We have a tests for that 

```
	// missing title or text length
	err = s.parseEventMessage(now, "_e{5,}:title|text", "default-hostname")
	require.Error(t, err)

	err = s.parseEventMessage(now, "_e{,4}:title|text", "default-hostname")
	require.Error(t, err)

	err = s.parseEventMessage(now, "_e{}:title|text", "default-hostname")
	require.Error(t, err)

	err = s.parseEventMessage(now, "_e{,}:title|text", "default-hostname")
	require.Error(t, err)
```
Check that.

plugins/inputs/statsd/datadog_test.go
danielnelson: I would switch them to the format we prefer, I'm don't anticipate that we will want to merge the code frequently.

plugins/inputs/statsd/datadog.go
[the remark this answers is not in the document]
danielnelson: What about this one? `"_e{100,:title|text"`

plugins/inputs/statsd/statsd.go
```diff
@@ -499,24 +533,8 @@ func (s *Statsd) parseStatsdLine(line string) error {
...
-						v = ""
-					case 2:
-						k = ts[0]
-						v = ts[1]
-					}
-					if k != "" {
-						lineTags[k] = v
-					}
-				}
+				fmt.Println("************************", segment[1:])
```
goller: Looks like an extra debugging println slipped in

plugins/inputs/statsd/datadog.go
```diff
@@ -0,0 +1,187 @@
...
+				m.fields["priority"] = priorityLow
+			case priorityNormal: // we already used this as a default
+			default:
+				continue
+			}
+		case "h:":
+			m.tags["source"] = rawMetadataFields[i][2:]
+		case "t:":
+			switch rawMetadataFields[i][2:] {
+			case "error", "warning", "success", "info":
```
goller: Would you change this to use your consts?

plugins/inputs/statsd/datadog.go
```diff
@@ -0,0 +1,187 @@
...
+			return err
+		}
+		s.acc.AddMetric(newM)
+		return nil
+	}
+
+	rawMetadataFields := strings.Split(message[1:], "|")
+	for i := range rawMetadataFields {
+		if len(rawMetadataFields[i]) < 2 {
+			log.Printf("W! [inputs.statsd] too short metadata field")
```
goller: Should this return an error?

plugins/inputs/statsd/statsd.go
docmerlin: Thanks!

plugins/inputs/statsd/statsd.go
```diff
@@ -489,6 +520,9 @@ func (s *Statsd) parseStatsdLine(line string) error {
 
 	lineTags := make(map[string]string)
 	if s.ParseDataDogTags {
+		log.Printf("I! WARNING statsd: parse_data_dog_tags config option is deprecated,  please use datadog_extensions instead")
```
danielnelson: Something like this, with W! and the [inputs.statsd] bit.  Also would be nice to wrap this in the code near 80 chars.
```
log.Println("W! [inputs.statsd] The parse_data_dog_tags option has been deprecated, please use datadog_extensions instead.")
```

Move this up to line 325, don't use the s.ParseDataDogTags field any other places in the code.

plugins/inputs/statsd/statsd.go
```diff
@@ -807,7 +823,17 @@ func (s *Statsd) handler(conn *net.TCPConn, id string) {
...
 	}()
-
+	addr := conn.RemoteAddr()
+	a, err := url.Parse(addr.String())
+	var host string
+	if err != nil {
+		// this should never happen because the conn handler should give us parsable addresses,
+		// but if it does we will know
+		host = "badhost"
+		log.Printf("E! failed to parse %s\n", addr)
```
danielnelson: Either close the connection or drop the packet (tcp vs udp), or just silently don't set a remote host.

plugins/inputs/statsd/statsd.go
```diff
@@ -261,41 +287,43 @@ func (s *Statsd) Gather(acc telegraf.Accumulator) error {
...
+		acc.AddFields(m.name, fields, m.tags, now)
 	}
 	if s.DeleteTimings {
 		s.timings = make(map[string]cachedtimings)
 	}
 
-	for _, metric := range s.gauges {
-		acc.AddGauge(metric.name, metric.fields, metric.tags, now)
+	for _, m := range s.gauges {
+		acc.AddGauge(m.name, m.fields, m.tags, now)
```
danielnelson: This is fine but normally don't change this type of stuff, it makes review and tracing history more difficult.

plugins/inputs/statsd/statsd.go
```diff
@@ -203,7 +225,7 @@ const sampleConfig = `
   delete_sets = true
   ## Reset timings & histograms every interval (default=true)
   delete_timings = true
-
+
```
danielnelson: remove trailing spaces

plugins/inputs/statsd/statsd.go
```diff
@@ -214,6 +236,10 @@ const sampleConfig = `
   ## http://docs.datadoghq.com/guides/dogstatsd/
   parse_data_dog_tags = false
 
+  ## Parses events in the datadog statsd format
+  ## http://docs.datadoghq.com/guides/dogstatsd/
+  parse_data_dog_events = false
```
danielnelson: Update for new option; don't include this link unless we have one that describes events/tags more precisely.

plugins/inputs/statsd/running_stats.go
```diff
@@ -49,7 +49,7 @@ func (rs *RunningStats) AddValue(v float64) {
 	}
 
 	// These are used for the running mean and variance
-	rs.n += 1
+	rs.n++
```
danielnelson: avoid this type of code churn

plugins/inputs/statsd/datadog_test.go
```diff
@@ -0,0 +1,472 @@
...
+			require.Equal(t, tests[i].expected.aggregationKey, m.Tags["aggregation_key"])
+			require.Equal(t, tests[i].expected.sourceTypeName, m.Fields["source_type_name"])
+			if tests[i].expected.checkTags != nil {
+				require.Equal(t, tests[i].expected.checkTags, m.Tags)
+			}
+		})
+	}
+}
+
+func TestEventError(t *testing.T) {
```
danielnelson: tabletests

plugins/inputs/statsd/datadog_test.go
```diff
@@ -0,0 +1,472 @@
...
+			err := s.parseEventMessage(tests[i].now, tests[i].message, tests[i].hostname)
+			if tests[i].err {
+				require.NotNil(t, err)
+			} else {
+				require.Nil(t, err)
+			}
+			require.Equal(t, uint64(i+1), acc.NMetrics())
+
+			require.Nil(t, err)
+			require.Equal(t, tests[i].expected.title, acc.Metrics[i].Measurement)
```
danielnelson: These are fine, no need to update, but in the future use this method:
https://github.com/influxdata/telegraf/blob/master/plugins/inputs/minecraft/minecraft_test.go#L89
https://github.com/influxdata/telegraf/blob/master/plugins/inputs/minecraft/minecraft_test.go#L121

plugins/inputs/statsd/datadog.go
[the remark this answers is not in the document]
danielnelson: Let's do this

plugins/inputs/statsd/datadog.go
```diff
@@ -0,0 +1,187 @@
...
+	message = message[titleLen+1+textLen:]
+
+	if len(rawTitle) == 0 || len(rawText) == 0 {
+		return fmt.Errorf("Invalid event message format: empty 'title' or 'text' field")
+	}
+
+	// Handle hostname, with a priority to the h: field, then the host:
+	// tag and finally the defaultHostname value
+	// Metadata
+	m := event{
```
danielnelson: I don't think this type is useful anymore.

plugins/inputs/statsd/datadog.go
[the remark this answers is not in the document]
danielnelson: check this

plugins/inputs/statsd/datadog.go
```diff
@@ -0,0 +1,187 @@
...
+			m.fields["source_type_name"] = rawMetadataFields[i][2:]
+		default:
+			if rawMetadataFields[i][0] == '#' {
+				parseDataDogTags(m.tags, rawMetadataFields[i][1:])
+			} else {
+				return fmt.Errorf("unknown metadata type: '%s'", rawMetadataFields[i])
+			}
+		}
+	}
+	if host, ok := m.tags["host"]; ok {
```
danielnelson: How would m.tags have a host key?

plugins/inputs/statsd/datadog.go
```diff
@@ -0,0 +1,187 @@
...
+	// Metadata
+	m := event{
+		name: rawTitle,
+	}
+	m.tags = make(map[string]string, strings.Count(message, ",")+2) // allocate for the approximate number of tags
+	m.fields = make(map[string]interface{}, 9)
+	m.fields["alert_type"] = eventInfo // default event type
+	m.fields["text"] = uncommenter.Replace(string(rawText))
+	// host is a magic tag in the system, and it expects it to replace the result of h: if it is present
+	// telegraf will add a"host" tag anyway with different meaning than dogstatsd, so we need to use source instead of host.
```
danielnelson: These comments are true but need to be cleaned up to be more readable.  Also they seem out of place here.  Somewhere else, maybe on line 119, you could put:
```
// Use source tag because host is reserved tag key in Telegraf.
```

plugins/inputs/statsd/datadog.go
```diff
@@ -0,0 +1,187 @@
...
+	rawText := message[titleLen+1 : titleLen+1+textLen]
+	message = message[titleLen+1+textLen:]
+
+	if len(rawTitle) == 0 || len(rawText) == 0 {
+		return fmt.Errorf("Invalid event message format: empty 'title' or 'text' field")
+	}
+
+	// Handle hostname, with a priority to the h: field, then the host:
+	// tag and finally the defaultHostname value
+	// Metadata
```
danielnelson: Remove these comments

plugins/inputs/statsd/datadog.go
```diff
@@ -0,0 +1,187 @@
+package statsd
+
+import (
+	"errors"
+	"fmt"
+	"strconv"
+	"strings"
+	"time"
+
+	tmetric "github.com/influxdata/telegraf/metric"
```
danielnelson: Is this alias needed?

plugins/inputs/statsd/README.md
```diff
@@ -185,6 +193,7 @@ the accuracy of percentiles but also increases the memory usage and cpu time.
 - **templates** []string: Templates for transforming statsd buckets into influx
 measurements and tags.
 - **parse_data_dog_tags** boolean: Enable parsing of tags in DataDog's dogstatsd format (http://docs.datadoghq.com/guides/dogstatsd/)
+- **parse_data_dog_events** boolean: Enable parsing of events in DataDog's dogstatsd format (http://docs.datadoghq.com/guides/dogstatsd/)
```
danielnelson: remove/update

plugins/inputs/statsd/datadog.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+	if host, ok := m.tags["host"]; ok {
+		delete(m.tags, "host")
+		m.tags["source"] = host
+	}
+	newM, err := tmetric.New(m.name, m.tags, m.fields, m.ts)
+	if err != nil {
+		return err
+	}
+	s.acc.AddMetric(newM)
```
danielnelson: Since we are not returning the telegraf.Metric, just use acc.AddFields

plugins/inputs/statsd/statsd.go
```diff
@@ -144,6 +157,7 @@ type metric struct {
 	additive   bool
 	samplerate float64
 	tags       map[string]string
+	ts         time.Time // for events
```
danielnelson: I don't think this is needed anymore.

plugins/inputs/statsd/statsd.go
```diff
@@ -131,6 +138,12 @@ type Statsd struct {
 	bufPool sync.Pool
 }
 
+type input struct {
+	*bytes.Buffer
+	time.Time
```
danielnelson: Do we still need a time field?

plugins/inputs/statsd/datadog.go
danielnelson: Remove this comment

plugins/inputs/statsd/datadog.go
danielnelson: This comment, while true, needs cleaned up/improved.  Also it doesn't seem relevant in its location.  Perhaps move down line 118 and changing to something more like:
```
// Use the source to contain the event host since host is a reserved tag name in Telegraf.
```

plugins/inputs/statsd/running_stats.go
docmerlin: Was being flagged by linter.

plugins/inputs/statsd/datadog_test.go
docmerlin: This is a lot longer and less readable as a table test, imo.

plugins/inputs/statsd/datadog.go
docmerlin: because in datadog's parser one way of setting the host (what we call source) is explicitly though a tag.  They treat that as equivalent to the h: value.

plugins/inputs/statsd/datadog.go
docmerlin: Yes, because there is a `metric` type used everywhere in this package, that collides with it.

plugins/inputs/statsd/statsd.go
docmerlin: It had collision with a type.

plugins/inputs/statsd/running_stats.go
danielnelson: The only static checkers we are using are `go vet` and `go fmt`, You will need to disable any other linters or ignore their errors, since we don't want to update unrelated items like this.

plugins/inputs/statsd/datadog_test.go
danielnelson: Change it anyway, I would rather have all failing tests report instead of one big test.  I'm also fine with individual tests.

plugins/inputs/statsd/datadog.go
danielnelson: I see, over in statsd.go

plugins/inputs/statsd/statsd.go
danielnelson: Okay but in the future don't worry about it.

plugins/inputs/statsd/statsd.go
docmerlin: In the udp case, there isn't a possibility of it not parsing, because the address it gives us is *net.UDPAddr, so we can pull out the IP directly.
I'll modify the TCP case. though

plugins/inputs/statsd/running_stats.go
docmerlin: ok.

plugins/inputs/statsd/statsd.go
```diff
@@ -807,7 +813,14 @@ func (s *Statsd) handler(conn *net.TCPConn, id string) {
 		s.forget(id)
 		s.CurrentConnections.Incr(-1)
 	}()
-
+	addr := conn.RemoteAddr()
+	parsedURL, err := url.Parse(addr.String())
+	if err != nil {
+		// this should never happen because the conn handler should give us parsable addresses,
+		// but if it does we will know
+		log.Printf("E! failed to parse %s\n", addr)
```
glinton: Add the plugin to the log output `E! [inputs.statsd] ...`

plugins/inputs/statsd/statsd.go
```diff
@@ -170,6 +183,8 @@ type cachedtimings struct {
 	tags   map[string]string
 }
 
+//this is used internally for building out an event
+
```
danielnelson: Remove hunk

plugins/inputs/statsd/statsd.go
```diff
@@ -439,17 +460,16 @@ func (s *Statsd) udpListen(conn *net.UDPConn) error {
...
+				log.Printf("E! [inputs.statsd] Error READ: %s\n", err.Error())
 				continue
 			}
 			b := s.bufPool.Get().(*bytes.Buffer)
 			b.Reset()
 			b.Write(buf[:n])
-
 			select {
-			case s.in <- b:
+			case s.in <- input{Buffer: b, Time: time.Now(), Addr: addr.IP.String()}:
```
danielnelson: Noticed this when checking into the toml option name.  This happens when stopping the plugin.

```
2019-05-10T20:23:23Z I! [inputs.statsd] Statsd UDP listener listening on:  0.0.0.0:8125
2019-05-10T20:23:23Z I! [inputs.statsd] Started the statsd service on :8125
^C2019-05-10T20:23:27Z I! [inputs.statsd] Stopping the statsd service
2019-05-10T20:23:27Z I! Stopped Statsd listener service on  :8125
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x16f7a20]

goroutine 8 [running]:
github.com/influxdata/telegraf/plugins/inputs/statsd.(*Statsd).udpListen(0xc0004d4380, 0xc0001e2008, 0xc0004d4408, 0x0)
        /home/dbn/src/influx/src/github.com/influxdata/telegraf/plugins/inputs/statsd/statsd.go:472 +0x250
github.com/influxdata/telegraf/plugins/inputs/statsd.(*Statsd).Start.func2(0xc0004d4380, 0xc0001e2008)
        /home/dbn/src/influx/src/github.com/influxdata/telegraf/plugins/inputs/statsd/statsd.go:378 +0x64
created by github.com/influxdata/telegraf/plugins/inputs/statsd.(*Statsd).Start
        /home/dbn/src/influx/src/github.com/influxdata/telegraf/plugins/inputs/statsd/statsd.go:376 +0x961

File: plugins/inputs/statsd/README.md [hunk 47]
```diff
@@ -176,6 +183,7 @@ to allow. Used when protocol is set to tcp.
 - **delete_counters** boolean: Delete counters on every collection interval
 - **delete_sets** boolean: Delete set counters on every collection interval
 - **delete_timings** boolean: Delete timings on every collection interval
+
```
danielnelson: remove

plugins/inputs/statsd/README.md
```diff
@@ -34,6 +34,7 @@
   ## Reset timings & histograms every interval (default=true)
   delete_timings = true
 
+
```
danielnelson: remove

plugins/inputs/statsd/statsd.go
```diff
@@ -67,7 +69,12 @@ type Statsd struct {
 	MetricSeparator string
 	// This flag enables parsing of tags in the dogstatsd extension to the
 	// statsd protocol (http://docs.datadoghq.com/guides/dogstatsd/)
-	ParseDataDogTags bool
+	ParseDataDogTags bool // depreciated in 1.10; use datadog_extensions
+
+	// Parses extensions to statsd in the datadog statsd format
+	// currently supports metrics and datadog tags.
+	// http://docs.datadoghq.com/guides/dogstatsd/
+	DataDogExtensions bool
```
danielnelson: Can you add a toml struct tag?  I wasn't sure if this would work with this spelling or only with "data_dog_extensions", turns out it works with both but I think it would be best to be explicit and anyway we probably don't want both forms.
```
`toml:"datadog_extensions"`
````

PR #5802, opened 2019-05-04
Javicrespo syslog output

javicrespo:
### Required for all PRs:


plugins/outputs/syslog/syslog_mapper.go
```diff
@@ -0,0 +1,201 @@
...
+	} else if value, err := os.Hostname(); err == nil {
+		msg.SetHostname(value)
+	}
+}
+
+func mapTimestamp(metric telegraf.Metric, msg *rfc5424.SyslogMessage) {
+	timestamp := metric.Time()
+	if value, ok := metric.GetField("timestamp"); ok {
+		if v, err := parseIntValue(value); err == nil {
+			timestamp = time.Unix(v, 0).UTC()
```
danielnelson: Does this need to be `time.Unix(0, v)`?

javicrespo: I forgot to address this remark from the other PR, sorry. Will do that soon

javicrespo: If I remove the UTC() call, then I the timestamp is written with time offset.

`-<26>2 2010-11-10T23:30:00Z testhost testapp 25 555 - Test message`
`+<26>2 2010-11-11T00:30:00+01:00 testhost testapp 25 555 - Test message`

That's why I'd added the UTC() call in the first place.

danielnelson: I'm actually referring to the order of the arguments to Unix(s, ns).  I looked at the syslog input and it appears that it writes the timestamp in nanoseconds so don't we need to pass it back in as nanoseconds here in order to roundtrip correctly?

javicrespo: Yep, that's already fixed in the last commit:
https://github.com/influxdata/telegraf/blob/004fd1f924d50c15843d5b72d78afb32e3709b89/plugins/outputs/syslog/syslog_mapper.go#L148

but I kept the trailing UTC() function, otherwise the time offset is added into the syslog message.

Issue #4256, opened 2018-06-08
Add cpu_speed (in mhz) to cpu or system measurement


randallt:
## Add cpu_speed to cpu or system measurement

We are currently in the process of converting from Ganglia to Telegraf. (yeah!) Unfortunately, we have some existing dependence on the Ganglia cpu_speed metric. This is not found in Telegraf.

### Proposal:

Add a cpu_speed field or equivalent to the cpu or system measurement. This would be in MHz.

### Use case: [Why is this important (helps with prioritizing requests)]

This helps mostly in the capacity management area, when mapping cpu mhz of an application group that is targeted for migration to new hosts. We can get the cpu speed other ways, of course, but having it directly and natively in Telegraf would be optimal.

danielnelson:
A quick look suggests that we could use the `cpu.Info()` function from gopsutil to pull in some additional cpu fields:
```go
type InfoStat struct {
	CPU        int32    `json:"cpu"`
	VendorID   string   `json:"vendorId"`
	Family     string   `json:"family"`
	Model      string   `json:"model"`
	Stepping   int32    `json:"stepping"`
	PhysicalID string   `json:"physicalId"`
	CoreID     string   `json:"coreId"`
	Cores      int32    `json:"cores"`
	ModelName  string   `json:"modelName"`
	Mhz        float64  `json:"mhz"`
	CacheSize  int32    `json:"cacheSize"`
	Flags      []string `json:"flags"`
	Microcode  string   `json:"microcode"`
}
```

This reads and parses `/proc/cpuinfo` on Linux.

phemmer:
Well it depends on what we're really looking for here. Are we wanting maximum speed, or current speed? What about the max or min limits?

randallt:
I was looking for just the "CPU MHz" field from the linux command 'lspcu', which appears to be the same as the "cpu MHz" field of each CPU core from 'cat /proc/cpuinfo'. This doesn't change for me and matches the CPU description, like "Intel(R) Xeon(R) CPU E5-2697 v2 @ 2.70GHz". I'm on a VMware infrastruture, though.

I was not looking for instantaneous frequency or even max boost frequency, just the frequency corresponds to the CPU description--which when combined with the CPU count can help give some comparative sense of capacity among VMs and environments.

danielnelson:
This field is the instantaneous frequency of the processor, but there is also min and max, here it is on my laptop:
```
CPU MHz:             499.877
CPU max MHz:         3400.0000
[...]

PR #5806, opened 2019-05-06
CPU Arch Support as tag in CPU plugin

### LINKED ISSUE #5805: CPU Arch Support in Telegraf
## Feature Request

### Proposal: 
It would be good to have the architecture of the CPU as part of the tags in CPU plugin. 

### Current behavior:
CPU Arch metric isn't being sent by any of the telegraf plugins
### Desired behavior:
Send CPU Arch value via the CPU plugin
### Use case: 
The CPU architecture will help in various filters and queries to differentiate hosts. Just like how total RAM,total disk,number of cores etc describes the host , the CPU arch value will further help describe the host . Examples of CPU Arch : amd64,arm etc

prashanthjbabu:
Resolves #5805
### Required for all PRs:

### DIFF
--- a/plugins/inputs/cpu/README.md
+++ b/plugins/inputs/cpu/README.md
@@ -64,7 +64,9 @@ cpu  3357 0 4313 1362393
 
 Meta:
 - units: CPU Time
-- tags: `cpu=<cpuN> or <cpu-total>`
+- tags:
+  - `cpu=<cpuN> or <cpu-total>`
+  - `cpu_arch=<cpu arch>`
 
 Measurement names:
 - cpu_time_user
--- a/plugins/inputs/cpu/cpu.go
+++ b/plugins/inputs/cpu/cpu.go
@@ -2,6 +2,7 @@ package cpu
 
 import (
 	"fmt"
+	"runtime"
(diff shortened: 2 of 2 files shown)


danielnelson:
This won't be entirely accurate since it really reports which build of Telegraf you have, for example this would report 386 if you used the 386 build, even if the host system was amd64.

I think it doesn't fit well with the `cpu` input, maybe we could include this information with the `internal_agent` metric?

prashanthjbabu:
@glinton  I was hoping there would be a more automated way to fetch this info.
@danielnelson  You're right, this approach seems to be inaccurate . Is there a more deterministic way in golang to get the CPU Arch value? I really think it would be useful as a tag in the CPU plugin.

danielnelson:
On Linux, most of this info is in `/proc/cpuinfo` or in `/sys/devices/system/cpu`.  The `lscpu` command is helpful here as well.  With more thought, maybe this should somehow become part of the proposed cpu info plugin in #4256?  I think adding cpu metadata could be a slippery slope and since there is so much potential information that might be collected, we should have a well thought out solution.

PR #5809, opened 2019-05-06
A builder for custom, smaller telegraf builds

docmerlin:
### Required for all PRs:


danielnelson:
I'm sure this will be very popular, especially if we can add deb/rpm support, which I think accounts for 95% of installs.

I have a few ideas about how to organize the pieces:  In the influxdata/telegraf repo I think we should have a command for configuring the all.go files.  The rest of this build server should move into to a new repo, this will give it more room to grow with styling, documentation, assets, and other features, and help keep keep a clear separation of responsibilities between it and the telegraf build system.

rawkode:
So I've been working on something similar:

https://github.com/rawkode/bring-your-own-telegraf


This currently uses a basic Toml file to generate the all.go files for inputs and outputs. Ideally this would take an actual `telegraf.conf` and build.

My approach only builds a Docker image

---

I wonder if we can somehow combine our efforts to provide a static binary and a Docker image, based on flags?

docmerlin:
That would be pretty neat.  Do you feel comfortable working on a parser for the conf?

srebhan:
Related to #8519 using a similar approach to the one of @rawkode... I think we should concentrate on one approach to push this forward. @docmerlin, @rawkode what do you think?

srebhan:
Let's bundle effort and continue in #11524.

Issue #5755, opened 2019-04-23
Plugin: ipmi_sensor: fails to detect psu failure


vvershkov:
### Relevant telegraf.conf:
```[[inputs.ipmi_sensor]]
interval = "60s"
metric_version = 1
timeout = "10s"
```

### System info:
Any OS or telegraf version: this bug is caused due to ipmi itself

### Steps to reproduce:
We have a server with failed PSU: we know it because we saw it: it has amber light instead of green and sound alarm too. But ipmi and telegraf detect it status as OK:
```# ipmitool sdr
...
PS1 Status       | 0x03              | ok
PS2 Status       | 0x01              | ok
```
However 0x03 flag is "failure":
https://support.hpe.com/hpsc/doc/public/display?docId=emr_na-sg8039en_us&docLocale=en_US

### Expected behavior:
PS1 Status is bad (0)

### Actual behavior:
PS1 Status is OK (1)

### Additional info:
AFAIK there is no readings with 0x03 and "OK" status. for old motherboards - 0x03 for CPU for example means overheating. But maybe flag check needed only for PSU.

danielnelson:
It looks like we don't handle hexadecimal values currently.  Could you also run this command and add the output for these sensors so I can be sure to fix it for `metric_version = 2` as well:
```
ipmitool sdr elist
```

vvershkov:
Wow, that is a worst timing ever :) I just replaced it with new ones.
But I can try to reproduce this bug with dead PSU and some server in office tomorrow.

vvershkov:
Ok, ipmitool sdr and sdr elist outputs looks like this:
Old version of ipmi failed PSU:
```
PS Status        | 0x02              | ok
PS Status        | 55h | ok  | 10.1 | Failure detected
```
(both PSU in one status, this is horrible)

after bmc update, turning off one power line:
```
PS1 Status       | 0x0b              | ok
PS2 Status       | 0x01              | ok
PS1 Status       | C8h | ok  | 10.1 | Presence detected, Failure detected, Power Supply AC lost
PS2 Status       | C9h | ok  | 10.2 | Presence detected
```
switching to failed psu:
```
PS1 Status       | 0x03              | ok
PS2 Status       | 0x01              | ok
PS1 Status       | C8h | ok  | 10.1 | Presence detected, Failure detected
PS2 Status       | C9h | ok  | 10.2 | Presence detected
```

PS funny thing:
```
PS1 Status       | 0x00              | ok
[...]

PR #5816, opened 2019-05-07
Fix/ipmi status codes

AlirieGray:
Closes: #5755

This PR adds support for hex values in IPMI messages. It also adds unit tests for both version 1 & 2 parsing of IPMI messages.

### DIFF
--- a/plugins/inputs/ipmi_sensor/ipmi.go
+++ b/plugins/inputs/ipmi_sensor/ipmi.go
@@ -153,7 +153,7 @@ func parseV1(acc telegraf.Accumulator, hostname string, cmdOut []byte, measured_
 		description := ipmiFields["description"]
 
 		// handle hex description field
-		if len(description) > 1 && description[:2] == "0x" {
+		if strings.HasPrefix(description, "0x") {
 			descriptionInt, err := strconv.ParseInt(description, 0, 64)
 			if err != nil {
 				continue
--- a/plugins/inputs/ipmi_sensor/ipmi.go
+++ b/plugins/inputs/ipmi_sensor/ipmi.go
@@ -150,9 +150,19 @@ func parseV1(acc telegraf.Accumulator, hostname string, cmdOut []byte, measured_
 			fields["status"] = 0
 		}
 
-		if strings.Index(ipmiFields["description"], " ") > 0 {
+		description := ipmiFields["description"]
+
+		// handle hex description field
+		if len(description) > 1 && description[:2] == "0x" {
+			descriptionInt, err := strconv.ParseInt(description, 0, 64)
+			if err != nil {
+				continue
+			}
+
+			fields["value"] = float64(descriptionInt)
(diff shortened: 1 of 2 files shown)


glinton:
Also, there's no need to force push on telegraf, we squash and merge so it ends up being one commit in the end.

plugins/inputs/ipmi_sensor/ipmi.go
```diff
@@ -150,9 +150,19 @@ func parseV1(acc telegraf.Accumulator, hostname string, cmdOut []byte, measured_
 			fields["status"] = 0
 		}
 
-		if strings.Index(ipmiFields["description"], " ") > 0 {
+		description := ipmiFields["description"]
+
+		// handle hex description field
+		if description[:2] == "0x" {
```
danielnelson: Check length of description first to avoid panic

plugins/inputs/ipmi_sensor/ipmi.go
```diff
@@ -150,9 +150,19 @@ func parseV1(acc telegraf.Accumulator, hostname string, cmdOut []byte, measured_
...
+			descriptionInt, err := strconv.ParseInt(description, 0, 64)
```
danielnelson: Check error before using descriptionInt

safadig:
can you please update documentation for where to add hex value? Thanks

danielnelson:
@safadig This PR should allow hex values to be parsed from the output of `ipmitool`, but it should be transparent and there is no configuration, values are reported in decimal.  Let me know if I am misunderstanding or you can open a new issue.

safadig:
> @safadig This PR should allow hex values to be parsed from the output of `ipmitool`, but it should be transparent and there is no configuration, values are reported in decimal. Let me know if I am misunderstanding or you can open a new issue.

@danielnelson  I am trying to use the ipmi from a Dell iDRAC that has user; password, as well as a hex based encryption key to allow access. 
I can get access using ipmitool:
sudo ipmitool -I lanplus -H 192.168.1.1 -U user -P password -y 0000000000000000000000000000000000000000 sdr -v
However, this is how the documentation lists the variables in telegraf.conf. I tried inputing the encrypted hex key before root and other areas but was not successful:
`
[[inputs.ipmi_sensor]]
#   ## optionally specify one or more servers via a url matching
#   ##  [username[:password]@][protocol[(address)]]
#   ##  e.g.
#   ##    root:passwd@lan(127.0.0.1)
   servers =["root:Password@lan(192.168.1.1)"]
   interval = "30s"
   timeout = "20s"
#
#   ## Schema Version: (Optional, defaults to version 1)
#   metric_version = 2
`

danielnelson:
Got it.  It won't be possible to pass this option right now, but I see your feature request #6858.

Raynok:
@danielnelson, also need this feature. I believe it can be a set variable that can be commented or uncommented in the config file (similar to the privilege level variable)

encryption Key = "String"

PR #5826, opened 2019-05-09
Fix parse of unix timestamp with more than ns precision

### LINKED ISSUE #5824: Problem parsing timestamps in CSV parser
### Relevant telegraf.conf:
```toml
[[inputs.file]]
  files = ["data2s.csv"]
  data_format = "csv"
  csv_header_row_count = 0
  csv_column_names = ["pid","time","msg"]
  csv_timestamp_column = "time"
  csv_timestamp_format = "unix"
```
### System info:
Telegraf 1.10.3
Ubuntu 16.04 LTS

### Steps to reproduce:
CSV contains Unix timestamps with fractional seconds with a large numver of decimal places.

Input csv file:
```csv
2108,1551129661.95456123352050781250,0105000008000000cb2e392df52c9c2d
2108,1551129662.14315605163574218750,0405000008000000527bec7ab47a20bf
2108,1551129662.14344882965087890625,02050000080000006a2f312ed62ca02d
```

### Expected behavior:
Timestamps would be parsed as Unix epoch seconds.fractional_seconds at down to nanosecond precision.

### Actual behavior:
The timestamps were being parsed without giving an error message, but were incorrect when imported to the database.

### Additional info:
It turned out that the ParseTimestampWithLocation function in internal/internal.go was receiving a timestamp string with an exponent e.g. 1.5515752371385763e+09 . This was confusing the part of the function which splits the string into timeInt and timeFractional.

### Fix:
The fix was to change one line in plugins/parsers/csv/parser.go inside the function parseTimestamp.  I changed the line:
`tStr := fmt.Sprintf("%v", recordFields[timestampColumn])`
to
`tStr := fmt.Sprintf("%.9f", recordFields[timestampColumn])`

This retains (up to) nanosecond precision, but forces it not to give an exponent.

EDIT:
In my haste, I forgot to test this with integer timestamps.  Of course forcing Sprintf to try to parse as a float causes integers to be parsed incorrectly. The final solution is slightly longer:

```
import "reflect"

...

var tStr string
k := reflect.ValueOf(recordFields[timestampColumn]).Kind()
if k == reflect.Float32 || k == reflect.Float64 {
    tStr = fmt.Sprintf("%.9f", recordFields[timestampColumn])
} else {
    tStr = fmt.Sprintf("%d", recordFields[timestampColumn])
}
```

danielnelson:
closes #5824

### Required for all PRs:

### DIFF
--- a/plugins/parsers/csv/parser_test.go
+++ b/plugins/parsers/csv/parser_test.go
@@ -5,6 +5,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/influxdata/telegraf"
 	"github.com/influxdata/telegraf/metric"
 	"github.com/influxdata/telegraf/testutil"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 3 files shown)


plugins/parsers/csv/parser_test.go
```diff
@@ -322,3 +323,30 @@ func TestParseStream(t *testing.T) {
...
+	}
+	data := `1551129661.95456123352050781250,42`
+
+	expected := []telegraf.Metric{
+		testutil.MustMetric(
+			"csv",
+			map[string]string{},
+			map[string]interface{}{
+				"value": 42,
+				"time":  1551129661.954561233,
```
danielnelson: I think this is a bug, we should be removing the time field when it is used as a timestamp column.  I will fix this in a separate pr for 1.11.

PR #5830, opened 2019-05-10
Add support for gzip compression to amqp input and output

### LINKED ISSUE #3557: AMQP Body Compression
### Proposal:
When using input plugins, like the logparser, that can put a large amount of text in the body of an AMQ message (over 300K), it can take much longer for Telegraf to pull messages from the queue or for Rabbit to send them across federated links.  It would be great if the amqp output plugin could check the size of the message body and if it's over a configurable amount, gzip the body.  The amqp_consumer input plugin would then need to auto-detect the content type and unzip as necessary.

### Current behavior:
All messages are plain text without any compression.

### Desired behavior:
Add a compression option to the amqp_consumer input plugin and amqp output plugin to allow gzip body compression over a configurable amount.

When Telegraf is configured with the logparser plugin and from_beginning set to true, it can dump a huge amount of very large messages onto the queue.  It then takes Telegraf a long time to pull those messages off of the queue.  With several thousand 300K+ messages load balanced across 3 queues on 3 different brokers I've only been able to hit 30 msg/s with 4 Telegraf consumers.  Once the backlog of logparser messages have been processed the throughput skyrockets to 1000 msg/s.

The logparser messages clog the queues and we lose access to real time stats for other metrics.

danielnelson:
closes #3557

### Required for all PRs:

### DIFF
--- a/internal/content_coding.go
+++ b/internal/content_coding.go
@@ -3,8 +3,34 @@ package internal
 import (
 	"bytes"
 	"compress/gzip"
+	"errors"
 	"io"
 )
+
+// NewContentEncoder returns a ContentEncoder for the encoding type.
+func NewContentEncoder(encoding string) (ContentEncoder, error) {
+	switch encoding {
+	case "gzip":
+		return NewGzipEncoder()
+
+	case "identity", "":
+		return NewIdentityEncoder(), nil
+	default:
+		return nil, errors.New("invalid value for content_encoding")
+	}
+}
+
+// NewContentDecoder returns a ContentDecoder for the encoding type.
+func NewContentDecoder(encoding string) (ContentDecoder, error) {
+	switch encoding {
+	case "gzip":
+		return NewGzipDecoder()
+	case "identity", "":
+		return NewIdentityDecoder(), nil
+	default:
+		return nil, errors.New("invalid value for content_encoding")
+	}
+}
 
 // ContentEncoder applies a wrapper encoding to byte buffers.
 type ContentEncoder interface {
--- a/internal/content_coding.go
+++ b/internal/content_coding.go
@@ -1,0 +1,96 @@
+package internal
+
+import (
+	"bytes"
+	"compress/gzip"
+	"io"
+)
+
+// ContentEncoder applies a wrapper encoding to byte buffers.
+type ContentEncoder interface {
+	Encode([]byte) ([]byte, error)
+}
+
+// GzipEncoder compresses the buffer using gzip at the default level.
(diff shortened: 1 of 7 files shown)


danielnelson:
- [telegraf-1.11.0~97f47699-0.x86_64.rpm](https://32577-33258973-gh.circle-artifacts.com/0/build/linux/amd64/telegraf-1.11.0~97f47699-0.x86_64.rpm)
- [telegraf_1.11.0~97f47699-0_amd64.deb](https://32577-33258973-gh.circle-artifacts.com/0/build/linux/amd64/telegraf_1.11.0~97f47699-0_amd64.deb)
- [telegraf-1.11.0~97f47699_windows_amd64.zip](https://32577-33258973-gh.circle-artifacts.com/0/build/windows/amd64/telegraf-1.11.0~97f47699_windows_amd64.zip)

Esity:
We should add in the compression message property so we know in RMQ if it is compressed. This should be set based on if compression is enabled or not. 

Content encoding, e.g. "gzip". Used by applications, not core RabbitMQ
While we are at it, we should also add one more

Content type, e.g. "application/json". Used by applications, not core RabbitMQ
Grabbed from this page https://www.rabbitmq.com/consumers.html

Esity:
Also tested with and without gzip compression
I added 10k messages to RMQ without 10k metric points per message for a total of 10,000,000 metrics.

Without compression the size on disk was over 4gb. With compression turned on it was 7kB. I would consider that a success @danielnelson 

It might be worth testing to see what the additional overhead against Telegraf is for enabling compression but I think it is worth it

danielnelson:
I added the Content-Encoding property, but held off on the Content-Type.  I think we should first make this part of the serializer interface separately so that we can pick the right value in a nice way.  Another idea that this brings up is doing automatic decoding on the consumer side based on the property, I think it could be nice but I'll have to leave that for later.

- [telegraf-1.11.0~543178e3-0.x86_64.rpm](https://33079-33258973-gh.circle-artifacts.com/0/build/linux/amd64/telegraf-1.11.0~543178e3-0.x86_64.rpm)
- [telegraf_1.11.0~543178e3-0_amd64.deb](https://33079-33258973-gh.circle-artifacts.com/0/build/linux/amd64/telegraf_1.11.0~543178e3-0_amd64.deb)
- [telegraf-1.11.0~543178e3_windows_amd64.zip](https://33079-33258973-gh.circle-artifacts.com/0/build/windows/amd64/telegraf-1.11.0~543178e3_windows_amd64.zip)

PR #5840, opened 2019-05-11
templating: fix templateSpecs.Less()

charlievieth:
Using the length of the slice returned by strings.Split() to count the number of separators in a string is incredibly wasteful and makes for expensive sort operations.  This commit fixes that.

```sh
$ benchcmp base.txt new.txt
benchmark                    old ns/op     new ns/op     delta
BenchmarkTemplateLess-16     180           12.5          -93.06%

benchmark                    old allocs     new allocs     delta
BenchmarkTemplateLess-16     2              0              -100.00%

benchmark                    old bytes     new bytes     delta
BenchmarkTemplateLess-16     192           0             -100.00%
```

### Required for all PRs:

N/A
A benchmark was added to identify regressions.

### DIFF
--- a/internal/templating/template.go
+++ b/internal/templating/template.go
@@ -124,21 +124,16 @@ type templateSpecs []templateSpec
 // Less reports whether the element with
 // index j should sort before the element with index k.
 func (e templateSpecs) Less(j, k int) bool {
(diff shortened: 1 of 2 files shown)


charlievieth:
Thanks for approving.

PR #5841, opened 2019-05-12
graphite: reduce the memory used by GraphiteParser.Parse() by ~80%

charlievieth:
The commit reduces the amount of memory used by `GraphiteParser.Parse()` by 80% and increases performance by 30%.  Honestly, this would be more efficient if we stepped through `buf` line-by-line but this improvement is good enough to justify a PR.

```
benchmark             old ns/op     new ns/op     delta
BenchmarkParse-16     1825          1235          -32.33%

benchmark             old allocs     new allocs     delta
BenchmarkParse-16     24             20             -16.67%

benchmark             old bytes     new bytes     delta
BenchmarkParse-16     5200          896           -82.77%
```

### Required for all PRs:

Also, this codebase and in particular anything on the statsd path spends quite a bit of time and memory splitting strings.  Fixing this would lead to some pretty easy perf wins.

### DIFF
--- a/plugins/parsers/graphite/parser.go
+++ b/plugins/parsers/graphite/parser.go
@@ -1,6 +1,7 @@
 package graphite
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"math"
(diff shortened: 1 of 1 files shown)


charlievieth:
Will investigate going line by line and will update the PR if it improves perf or note why it didn't (and will remove my comment if thats the case).

charlievieth:
@glinton updated to step through the buffer line-by-line the improves perf by roughly ~5%.

plugins/parsers/graphite/parser.go
```diff
@@ -68,21 +69,26 @@ func (p *GraphiteParser) Parse(buf []byte) ([]telegraf.Metric, error) {
...
+			metric, err := p.ParseLine(string(line))
+			if err == nil {
+				metrics = append(metrics, metric)
+			} else {
 				errs = append(errs, err.Error())
-				continue
 			}
-			metrics = append(metrics, metric)
 		}
+		if n == -1 {
```
glinton: Kind of a nitpick, but I'd feel safer if this was `n < 0`. Your call

charlievieth: doesn't hurt, changed both.

PR #5842, opened 2019-05-12
wavefront/serializer: improve performance by ~30%

charlievieth:
This PR improves the performance and memory consumption of the Wavefront serializer by roughly 30%.

```
benchmark                 old ns/op     new ns/op     delta
BenchmarkSerialize-16     1168          778           -33.39%

benchmark                 old allocs     new allocs     delta
BenchmarkSerialize-16     16             8              -50.00%

benchmark                 old bytes     new bytes     delta
BenchmarkSerialize-16     1116          800           -28.32%
```

### Required for all PRs:

### DIFF
--- a/plugins/serializers/wavefront/wavefront.go
+++ b/plugins/serializers/wavefront/wavefront.go
@@ -159,7 +159,7 @@ func formatMetricPoint(b *buffer, metricPoint *wavefront.MetricPoint, s *Wavefro
 	b.WriteString(`" `)
 	b.WriteFloat64(metricPoint.Value)
 	b.WriteChar(' ')
-	b.WriteUnit64(uint64(metricPoint.Timestamp))
+	b.WriteUint64(uint64(metricPoint.Timestamp))
 	b.WriteString(` source="`)
 	b.WriteString(metricPoint.Source)
 	b.WriteChar('"')
@@ -204,7 +204,7 @@ func (b *buffer) WriteChar(c byte) {
 	*b = append(*b, c)
 }
 
-func (b *buffer) WriteUnit64(val uint64) {
+func (b *buffer) WriteUint64(val uint64) {
 	*b = strconv.AppendUint(*b, val, 10)
 }
 
--- a/plugins/serializers/wavefront/wavefront.go
+++ b/plugins/serializers/wavefront/wavefront.go
@@ -192,10 +192,6 @@ func (b *buffer) Copy() []byte {
 	return p
 }
 
-func (b *buffer) Write(p []byte) {
-	*b = append(*b, p...)
-}
-
 func (b *buffer) WriteString(s string) {
 	*b = append(*b, s...)
 }
--- a/plugins/serializers/wavefront/wavefront.go
+++ b/plugins/serializers/wavefront/wavefront.go
@@ -1,8 +1,6 @@
 package wavefront
 
 import (
-	"errors"
-	"fmt"
 	"log"
 	"strconv"
 	"strings"
@@ -71,8 +69,8 @@ func (s *WavefrontSerializer) serialize(buf *buffer, m telegraf.Metric) {
 
 		name = pathReplacer.Replace(name)
 
-		metricValue, buildError := buildValue(value, name)
-		if buildError != nil {
+		metricValue, valid := buildValue(value, name)
+		if !valid {
 			// bad value continue to next metric
 			continue
 		}
@@ -136,28 +134,27 @@ func buildTags(mTags map[string]string, s *WavefrontSerializer) (string, map[str
 	return tagValueReplacer.Replace(source), mTags
 }
 
-func buildValue(v interface{}, name string) (float64, error) {
+func buildValue(v interface{}, name string) (val float64, valid bool) {
 	switch p := v.(type) {
 	case bool:
 		if p {
-			return 1, nil
-		} else {
-			return 0, nil
+			return 1, true
 		}
+		return 0, true
 	case int64:
-		return float64(p), nil
+		return float64(p), true
 	case uint64:
-		return float64(p), nil
+		return float64(p), true
 	case float64:
-		return p, nil
+		return p, true
 	case string:
-		// return an error but don't log
-		return 0, errors.New("string type not supported")
+		// return false but don't log
+		return 0, false
 	default:
-		// return an error and log a debug message
-		err := fmt.Errorf("unexpected type: %T, with value: %v, for :%s", v, v, name)
-		log.Printf("D! Serializer [wavefront] %s\n", err.Error())
-		return 0, err
+		// log a debug message
+		log.Printf("D! Serializer [wavefront] unexpected type: %T, with value: %v, for :%s\n",
+			v, v, name)
+		return 0, false
 	}
 }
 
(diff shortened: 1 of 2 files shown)


charlievieth:
Thanks for approving.

plugins/serializers/wavefront/wavefront.go
```diff
@@ -50,16 +58,17 @@ func NewSerializer(prefix string, useStrict bool, sourceOverride []string) (*Wav
 
 // Serialize : Serialize based on Wavefront format
 func (s *WavefrontSerializer) Serialize(m telegraf.Metric) ([]byte, error) {
-	out := []byte{}
-	metricSeparator := "."
+	const metricSeparator = "."
+	var out []byte
+	buf := pbFree.Get().(*buffer)
```
danielnelson: It probably makes sense to just create a single bytes.Buffer and then pass this into the various functions, the `out` slice would be removed and then there is no reason to append.  I don't see a reason we need both of these.

danielnelson:
@puckpuck Can you review?  I assume any changes would also need to go into the wavefront-sdk.

plugins/serializers/wavefront/wavefront.go
charlievieth: Agreed, will change.

plugins/serializers/wavefront/wavefront.go
```diff
@@ -39,6 +41,12 @@ var tagValueReplacer = strings.NewReplacer("\"", "\\\"", "*", "-")
 
 var pathReplacer = strings.NewReplacer("_", ".")
 
+var bufPool = sync.Pool{
```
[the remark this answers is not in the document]
charlievieth: I still think there is value to using a `sync.Pool` that is global to the package.  The `sync.Pool` will [pin](https://golang.org/src/sync/pool.go?#L178) the current goroutine to a pool, which should improve locality of reference.  Additionally, it does not make the `WavefrontSerializer` unsafe to use concurrently and is dynamic in size.

plugins/serializers/wavefront/wavefront.go
charlievieth: Expect something from me tomorrow or Friday as I'm pretty short on bandwidth right now.

plugins/serializers/wavefront/wavefront.go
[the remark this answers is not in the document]
danielnelson: There isn't any reason for it to be dynamic though because it will never need more than a single buffer.  Another issue with the pool is that you can't return the buffer to it unless you copy the data out of the buffer, I believe the current use in this pull request is unsafe even in a single threaded environment.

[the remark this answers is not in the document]
charlievieth: > Another issue with the pool is that you can't return the buffer to it unless you copy the data out of the buffer

We do that.  The contents of pooled buffers are copied to the [`out`](https://github.com/influxdata/telegraf/blob/ec4b147bebacece7e6198d3ae783b3c9c8cea638/plugins/serializers/wavefront/wavefront.go#L95) buffer (note: `out` will never alias `buf`, [example](https://play.golang.org/p/lCmOVPwqrme)).  This code is thread-safe.

Personally, I would prefer to not add a mutable field to the `WavefrontSerializer`, but I understand the argument for simplicity and will make the change.

What are your thoughts on adding a mutex to protect the `WavefrontSerializer`'s scratch buffer?  From you comment and a quick scan of the code it appears there is no need for it to be thread-safe, but if that were to change it would be beneficial to have it as a guard against this code being overlooked during a refactor.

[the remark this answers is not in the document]
charlievieth: I missed this, but this pool was added by accident the one we actually use is [`pbFree`](https://github.com/influxdata/telegraf/pull/5842/files#diff-cab4bf7f8e4b237a6576ccf08eea4f1cR197).  That said, I'm still removing the pools in favor of a shared buffer - as requested.

plugins/serializers/wavefront/wavefront.go
```diff
@@ -70,80 +69,62 @@ func (s *WavefrontSerializer) Serialize(m telegraf.Metric) ([]byte, error) {
 
 		name = pathReplacer.Replace(name)
 
-		metric := &wavefront.MetricPoint{
-			Metric:    name,
-			Timestamp: m.Time().Unix(),
-		}
-
-		metricValue, buildError := buildValue(value, metric.Metric)
+		metricValue, buildError := buildValue(value, name)
```
charlievieth: We always ignore this error - is there any point returning it from `buildValue()`?

plugins/serializers/wavefront/wavefront.go
```diff
@@ -172,31 +153,62 @@ func buildValue(v interface{}, name string) (float64, error) {
...
+// This is named WriteChar instead of WriteByte because the 'stdmethods' check
+// of 'go vet' wants WriteByte to have the signature:
+//
+// 	func (b *buffer) WriteByte(c byte) error { ... }
+//
+func (b *buffer) WriteChar(c byte) {
+	*b = append(*b, c)
+}
+
+func (b *buffer) WriteUnit64(val uint64) {
```
puckpuck: function name has a typo in it, which can be misleading (Unit vs Uint)

plugins/serializers/wavefront/wavefront.go
```diff
@@ -172,31 +153,62 @@ func buildValue(v interface{}, name string) (float64, error) {
...
+		b.WriteChar('"')
 	}
 
-	buffer.WriteString("\n")
+	b.WriteChar('\n')
+
+	return *b
+}
+
+// Use a fast and simple buffer for constructing statsd messages
```
puckpuck: I'm guessing this comment is a left over from something else?

plugins/serializers/wavefront/wavefront.go
glinton: Seems to be used to determine if the value should be skipped or not.

puckpuck: Typically you log on error from the caller, but we log inside the buildValue function instead to avoid noisy log outputs when the value is of type string (which is expected to not work).

plugins/serializers/wavefront/wavefront.go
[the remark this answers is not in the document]
danielnelson: Don't add anything to make it thread safe.  I agree it there is a risk of code sneaking in that use these wrong, especially with the way the SerializerOutput interface works (sending in a single instance), but I'd like to move the other way with parsers/serializers.

plugins/serializers/wavefront/wavefront.go
charlievieth: kinda, and good catch - will remove.

plugins/serializers/wavefront/wavefront.go
charlievieth: Thanks for the insight, I just changed this to return a `bool` instead of an `error` since we used the `error` as a bool (nil / not-nil) and this saves an alloc.

plugins/serializers/wavefront/wavefront.go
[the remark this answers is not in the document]
charlievieth: Cool, I made `WavefrontSerializer` thread-safe.

plugins/serializers/wavefront/wavefront.go
charlievieth: 🤦‍♂ thanks - fixed

PR #5848, opened 2019-05-13
Stop claiming plugin contributions are eagerly accepted

mjiderhamn:
### Required for all PRs:

- [-] Has appropriate unit tests.

The `README.md` claims that plugin contributions via pull requests are "eagerly" accepted. As subjective as "eagerly" may be interpreted, the claim seems less than honest considering the age of many of the open pull requests. (While I subjectively think mine is way over due for "eagerly", it isn't even old in comparison.)

To avoid leading companies to make tooling decision based on less than accurate information, I suggest this claim is removed.


danielnelson:
I'll reword this paragraph.

PR #5850, opened 2019-05-13
Revert field name case in interrupts input

### LINKED ISSUE #5836: (#5024: Add option to store cpu as a tag in interrupts input) breaks compatibility
### Relevant telegraf.conf:
```
[[inputs.interrupts]]
        cpu_as_tag = false
```

### System info:
`Linux 3.10.0-957.12.1.el7.x86_64 #1 SMP Mon Apr 29 14:59:59 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux`
`CentOS Linux release 7.6.1810 (Core)`

`telegraf --version`
`Telegraf 1.10.3 (git: HEAD 294bb666)`

### Steps to reproduce:

`> show field keys;`
`name: soft_interrupts`
`fieldKey fieldType`
`-------- ---------`
`CPU0     integer`
`CPU1     integer`
`CPU2     integer`
`CPU3     integer`
`count    integer`
`cpu0     integer`
`cpu1     integer`
`cpu2     integer`
`cpu3     integer`
`total    integer `

### Expected behavior:
Fieldkeys format is expected to be preserved between versions,  as well as any other filed, previously used. Not doing so, leaves the data, collected so far, "behind".

### Actual behavior:
Data collection is imported in the lower letters keys _cpu[0-9]_ and _count_. The behavior has changed since v1.10.

### Additional info:

glinton:
Resolves #5836

Should we also change the case of the newly added cpu as tag to uppercase to match?

### DIFF
--- a/plugins/inputs/interrupts/interrupts.go
+++ b/plugins/inputs/interrupts/interrupts.go
@@ -102,7 +102,7 @@ func gatherTagsFields(irq IRQ) (map[string]string, map[string]interface{}) {
 	tags := map[string]string{"irq": irq.ID, "type": irq.Type, "device": irq.Device}
 	fields := map[string]interface{}{"total": irq.Total}
 	for i := 0; i < len(irq.Cpus); i++ {
-		cpu := fmt.Sprintf("cpu%d", i)
+		cpu := fmt.Sprintf("CPU%d", i)
 		fields[cpu] = irq.Cpus[i]
 	}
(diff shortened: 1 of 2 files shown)


danielnelson:
Don't forget to update the tests.  Let's stay with lowercase cpu in the newer (`cpu_as_tag=true`) format, since it matches the cpu plugin.

PR #5851, opened 2019-05-13
Allow multiple tag replaces in the string processor

### LINKED ISSUE #5843: multuple instance of processors.strings.replace not working as expected
### Relevant telegraf.conf:

```toml
[[processors.strings]]  
order = 102 
   [[processors.strings.replace]]
   tag = "SenderId"
   old = "IPC_123456_01"
   new = "SITE01"
   dest = "site_id"  
    
   [[processors.strings.replace]]
   tag = "SenderId"
   old = "IPC_812345_01"
   new = "SITE02"
   dest = "site_id" 
```

### System info:

1.10.1
### Steps to reproduce:

The telgraf uses infux output to store data.
In above situation, only the second tag value is replaced, not the first one.

### Expected behavior:
Both tag values should be converted to SITE01 and SITE02 respectively.
### Actual behavior:
Only IPC_123456_01 is converter to SITE02 is stored

glinton:
Resolves #5843

With the following config file, it is easy to reproduce the initial reported bug:
```toml
[[outputs.file]]

[[inputs.exec]]
  command = "echo -en 'testmessage,SenderId=IPC_123456_01 message=\"sweet nothings\" 1530654676316265790\ntestmessage,SenderId=IPC_812345_01 message=\"sweet nothings\" 1530654676316265790'"
  data_format = "influx"

  [[processors.strings]]  
    [[processors.strings.replace]]
      tag = "SenderId"
      old = "IPC_123456_01"
      new = "SITE01"
      dest = "site_id"  
      
    [[processors.strings.replace]]
      tag = "SenderId"
      old = "IPC_812345_01"
      new = "SITE02"
      dest = "site_id" 
```

Difference between 1.10.3 and this commit
```diff
testmessage,SenderId=IPC_812345_01,site_id=SITE02 message="sweet nothings" 1530654676000000000
< testmessage,SenderId=IPC_123456_01,site_id=IPC_123456_01 message="sweet nothings" 1530654676000000000
> testmessage,SenderId=IPC_123456_01,site_id=SITE01 message="sweet nothings" 1530654676000000000
```

### DIFF
--- a/plugins/processors/strings/strings.go
+++ b/plugins/processors/strings/strings.go
@@ -91,13 +91,15 @@ func (c *converter) convertTag(metric telegraf.Metric) {
 	var tags map[string]string
 	if c.Tag == "*" {
 		tags = metric.Tags()
-	} else {
-		tags = make(map[string]string)
-		tv, ok := metric.GetTag(c.Tag)
+	}
+
+	tags = make(map[string]string)
+	for _, tag := range []string{c.Tag, c.Dest} {
+		tv, ok := metric.GetTag(tag)
 		if !ok {
-			return
+			continue
 		}
-		tags[c.Tag] = tv
+		tags[tag] = tv
 	}
 
 	for key, value := range tags {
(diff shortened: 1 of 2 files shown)


glinton:
This same behavior is seen in `convertField`, where this fix could be applied as well.

glinton:
It seems another solution would be to add support tags in the enum processor.

danielnelson:
Closing, as we discussed offline the replacement is not conditional and changing it break some other use cases such as replacement of characters that may or may not be in the source key.

PR #5855, opened 2019-05-14
Support tags in enum processor

### LINKED ISSUE #5033: enum processor should also support tags
## Feature Request
As documented in #4616 the enum processor should support both fields and tags.

### Proposal:
Support for tags should be added.

### Current behavior:
In its current v1.9.0 form it only supports fields, not tags.

### Desired behavior:
It should also support tags.

Here is an example where the SNMP object `hrStorageType`, which has a value that is an OID, can be made more user-friendly.

```
[[processors.enum]]
  [[processors.enum.mapping]]
    tag = "hrStorageType"
    [processors.enum.mapping.value_mappings]
      ".1.3.6.1.2.1.25.2.1.1" = "Other"
      ".1.3.6.1.2.1.25.2.1.2" = "RAM"
      ".1.3.6.1.2.1.25.2.1.3" = "Virtual Memory"
      ".1.3.6.1.2.1.25.2.1.4" = "Fixed Disk"
      ".1.3.6.1.2.1.25.2.1.5" = "Removable Disk"
      ".1.3.6.1.2.1.25.2.1.6" = "Floppy Disk"
      ".1.3.6.1.2.1.25.2.1.7" = "Compact Disc"
      ".1.3.6.1.2.1.25.2.1.8" = "RAM Disk"
      ".1.3.6.1.2.1.25.2.1.9" = "Flash Memory"
      ".1.3.6.1.2.1.25.2.1.10" = "Network Disk"
```

glinton:
Resolves #5033

### DIFF
--- a/plugins/processors/enum/enum.go
+++ b/plugins/processors/enum/enum.go
@@ -1,6 +1,7 @@
 package enum
 
 import (
+	"fmt"
 	"strconv"
 
 	"github.com/influxdata/telegraf"
@@ -33,6 +34,7 @@ type EnumMapper struct {
 }
 
 type Mapping struct {
+	Tag           string
 	Field         string
 	Dest          string
 	Default       interface{}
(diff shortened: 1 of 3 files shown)


plugins/processors/enum/enum.go
```diff
@@ -56,10 +58,26 @@ func (mapper *EnumMapper) Apply(in ...telegraf.Metric) []telegraf.Metric {
...
+				if adjustedValue, isString := adjustBoolValue(originalValue).(string); isString == true {
+					if mappedValue, isMappedValuePresent := mapping.mapValue(adjustedValue); isMappedValuePresent == true {
+						writeField(metric, mapping.getDestination(), mappedValue)
+					}
+				}
+			}
+		}
+		if mapping.Tag != "" {
+			if originalValue, isPresent := metric.GetTag(mapping.Tag); isPresent == true {
+				if adjustedValue, isString := adjustBoolValue(originalValue).(string); isString == true {
```
danielnelson: Tags are always strings, so I think this area can be simplified.

danielnelson:
Just give the docs/sampleconfig a pass and I think it's good.

PR #5870, opened 2019-05-17
Support for SNMP over TCP

skyfmmf:
While SNMP is usually done over UDP, it is possible to transport it using TCP. This is specified in RFC 3430 and should be used when transferring lots of data.

This PR adds support for SNMP over TCP to telegraf. The heavy lifting is done by the gosnmp library. Therefore it was only necessary to set the transport protocol for the agents, that should be gathered via TCP. To use it, you have to configure an agent like this:

```toml
[[inputs.snmp]]
  agents = [ "tcp://1.2.3.4" ]
```

~The notation of prefixing with `tcp:` has been taken from net-snmp which uses it for their command line tools like snmpbulkwalk.~

### Required for all PRs:

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -489,6 +489,14 @@
   version = "v1.1.1"
 
 [[projects]]
+  digest = "1:530233672f656641b365f8efb38ed9fba80e420baff2ce87633813ab3755ed6d"
+  name = "github.com/golang/mock"
+  packages = ["gomock"]
+  pruneopts = ""
+  revision = "51421b967af1f557f93a59e0057aaf15ca02e29c"
+  version = "v1.2.0"
+
+[[projects]]
   digest = "1:f958a1c137db276e52f0b50efee41a1a389dcdded59a69711f3e872757dab34b"
   name = "github.com/golang/protobuf"
   packages = [
@@ -1042,12 +1050,12 @@
   version = "v1.0.5"
 
 [[projects]]
-  branch = "master"
-  digest = "1:4b0cabe65ca903a7b2a3e6272c5304eb788ce196d35ecb901c6563e5e7582443"
+  digest = "1:a1cb5e999ad98b9838147e11ed1bdb000e750ee8872e2e21c74d9464cc9110c0"
   name = "github.com/soniah/gosnmp"
   packages = ["."]
   pruneopts = ""
-  revision = "96b86229e9b3ffb4b954144cdc7f98fe3ee1003f"
+  revision = "40eae407a1f8cbbe3f3f14c57bde0b16db1cfe85"
+  version = "v1.22.0"
 
 [[projects]]
   branch = "master"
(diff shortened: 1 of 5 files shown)


Gopkg.lock
```diff
@@ -1043,11 +1051,11 @@
 
 [[projects]]
   branch = "master"
```
glinton: I wonder if we shouldn't use a release for this..

plugins/inputs/snmp/snmp.go
```diff
@@ -623,6 +623,11 @@ func (s *Snmp) getConnection(idx int) (snmpConnection, error) {
 	gs := gosnmpWrapper{&gosnmp.GoSNMP{}}
 	s.connectionCache[idx] = gs
 
+	if strings.HasPrefix(agent, "tcp:") {
+		parts := strings.SplitN(agent, ":", 2)
+		gs.Transport = parts[0]
+		agent = parts[1]
```
glinton: Not checking the length here made me nervous, but I guess at the worst, parts[1] would only be blank.

Gopkg.lock
```diff
@@ -488,6 +488,14 @@
   revision = "636bf0302bc95575d69441b25a2603156ffdddf1"
   version = "v1.1.1"
 
+[[projects]]
+  digest = "1:530233672f656641b365f8efb38ed9fba80e420baff2ce87633813ab3755ed6d"
+  name = "github.com/golang/mock"
```
danielnelson: I see this is a new dependency of gosnmp, maybe if it were generated into the gosnmp_test package we wouldn't need this testing dep?  I'd like to have someone look into if this is possible before updating gosnmp.

plugins/inputs/snmp/snmp.go
```diff
@@ -623,6 +623,11 @@ func (s *Snmp) getConnection(idx int) (snmpConnection, error) {
 	gs := gosnmpWrapper{&gosnmp.GoSNMP{}}
 	s.connectionCache[idx] = gs
 
+	if strings.HasPrefix(agent, "tcp:") {
```
danielnelson: Let's use `tcp://` format instead of adopting the net-snmp style.  I think it will provide more consistency with other plugins.

plugins/inputs/snmp/snmp.go
danielnelson: It is safe, because the HasPrefix check, but I'd probably use strings.TrimPrefix to remove it instead.

Gopkg.lock
danielnelson: Let's try to remember to adjust this to a release when one is available.

glinton: Is [1.21.0](https://github.com/soniah/gosnmp/releases/tag/v1.21.0) sufficient (latest release from april)

danielnelson: No, the upstream tcp support was just added a few weeks ago.

glinton: Ah, right on

skyfmmf: Now there is a release so I used this

Gopkg.lock
skyfmmf: I understand that you do not want to include this testing dep, but I am not sure how to fix this.

danielnelson:
Sorry about stranding this PR for so long, while I'd still love the gomock dependency removed it's not enough to hold up this PR.

PR #5874, opened 2019-05-17
Collect info commandstats from redis

### LINKED ISSUE #5774: gatherCommandstatsLine
## Feature Request

### Desired behavior:

```
func gatherCommandstatsLine(
	name string,
	line string,
	acc telegraf.Accumulator,
	global_tags map[string]string,
) {
	if strings.Contains(line, "calls=") {
		tags := make(map[string]string)
		for k, v := range global_tags {
			tags[k] = v
		}
		tags["key_name"] = name
		dbparts := strings.Split(line, ",")
		for _, dbp := range dbparts {
			kv := strings.Split(dbp, "=")
			ival, err := strconv.ParseInt(kv[1], 10, 64)
			if err == nil {
				fields[kv[0]] = ival
			}
		}
		acc.AddFields("commandstats", fields, tags)
	}
}
```

wingyplus:
Redis has an info type "commandstats" that can be collect each command
statistics inside redis. To make this work, I call info with commandstats
argument after called info (default information), append it and then parse
metrics.

Closes #5774

### Required for all PRs:

### DIFF
--- a/plugins/inputs/redis/redis.go
+++ b/plugins/inputs/redis/redis.go
@@ -27,7 +27,7 @@ type Redis struct {
 }
 
 type Client interface {
-	Info() *redis.StringCmd
+	Info(section ...string) *redis.StringCmd
 	BaseTags() map[string]string
 }
 
@@ -36,8 +36,8 @@ type RedisClient struct {
 	tags   map[string]string
 }
 
-func (r *RedisClient) Info() *redis.StringCmd {
-	return r.client.Info()
+func (r *RedisClient) Info(section ...string) *redis.StringCmd {
+	return r.client.Info(section...)
 }
 
 func (r *RedisClient) BaseTags() map[string]string {
@@ -188,7 +188,10 @@ func (r *Redis) gatherServer(client Client, acc telegraf.Accumulator) error {
 		return err
 	}
 
-	rdr := strings.NewReader(info)
+	// skip for commandstats if we can't retrieve it.
+	cmdstats, _ := client.Info("commandstats").Result()
+
+	rdr := strings.NewReader(info + "\n\n" + cmdstats)
 	return gatherInfoOutput(rdr, acc, client.BaseTags())
 }
 
@@ -246,6 +249,10 @@ func gatherInfoOutput(
 			if section == "Keyspace" {
 				kline := strings.TrimSpace(string(parts[1]))
 				gatherKeyspaceLine(name, kline, acc, tags)
+				continue
+			} else if section == "Commandstats" {
+				kline := strings.TrimSpace(parts[1])
+				gatherCommandstatsLine(name, kline, acc, tags)
 				continue
 			}
 			metric = name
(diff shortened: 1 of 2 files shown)


plugins/inputs/redis/redis.go
```diff
@@ -321,6 +328,48 @@ func gatherKeyspaceLine(
 	}
 }
 
+// Parse the special Keyspace line at end of redis stats
+// This is a special line that looks something like:
+//     db0:keys=2,expires=0,avg_ttl=0
+// And there is one for each db on the redis instance
+//
+// TODO: write code comments
```
glinton: Remove or do `TODO`

plugins/inputs/redis/redis.go
```diff
@@ -321,6 +328,48 @@ func gatherKeyspaceLine(
...
+
+	fields := make(map[string]interface{})
+	tags := make(map[string]string)
+	for k, v := range global_tags {
+		tags[k] = v
+	}
+	tags["cmdstat"] = strings.TrimPrefix(name, "cmdstat_")
+	stats := strings.Split(line, ",")
+	for _, stat := range stats {
+		kv := strings.Split(stat, "=")
```
glinton: Check length to avoid a panic later

plugins/inputs/redis/redis.go
```diff
@@ -321,6 +328,48 @@ func gatherKeyspaceLine(
...
+		var (
+			ival interface{}
+			err  error
+		)
+		// only calls that always be int
```
glinton: nitpick: rephrase this comment

plugins/inputs/redis/redis.go
```diff
@@ -321,6 +328,48 @@ func gatherKeyspaceLine(
...
+		if kv[0] == "calls" {
+			ival, err = strconv.ParseInt(kv[1], 10, 64)
+		} else {
+			ival, err = strconv.ParseFloat(kv[1], 64)
+		}
+		if err == nil {
+			fields[kv[0]] = ival
+		}
+	}
+	acc.AddFields("redis_commandstats", fields, tags)
```
glinton: What if `fields` is empty?

plugins/inputs/redis/redis.go
wingyplus: Fixed

plugins/inputs/redis/redis.go
wingyplus: Fixed

plugins/inputs/redis/redis.go
```diff
@@ -188,7 +188,10 @@ func (r *Redis) gatherServer(client Client, acc telegraf.Accumulator) error {
 		return err
 	}
 
-	rdr := strings.NewReader(info)
+	// skip for commandstats if we can't retrieve it.
+	cmdstats, _ := client.Info("commandstats").Result()
```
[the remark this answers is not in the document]
wingyplus: I'm not sure that which ways is better betwen log an error and let info return default information if we cannot collect command stats and returns an error.

plugins/inputs/redis/redis.go
wingyplus: It’s hard to be empty because redis are not returns any line if it has no any command calls. Except that commandstats return the result incorrectly.

danielnelson:
This functionality was merged in #5926

PR #5882, opened 2019-05-20
Add health output plugin

### LINKED ISSUE #5707: Add healthcheck output
### Proposal:
add the ability for load balancers to determine if a telegraf service is healthy via an http(s) endpoint that returns 200 for OK or not 200 for NOT_OK. 

### Current behavior:
no plugin exists

### Desired behavior:
this plugin would have the following capabilities:
 - the ability to configure a threshold for output buffers to be considered healthy. For example, if the buffer is >=90% full, the service is unhealthy
 - the plugin should expose a url on a configurable port and path
 - the plugin should get its information from the `internal` plugin
 - a threshold should be applied to each output buffer for each plugin, with the ability to specify a default if none exists

this would enable telegraf to be used effectively behind a load balancer. today, there are workarounds, but this would allow health to be determined via buffer size, and reduce the probability that telegraf might drop metrics when the buffer size limit is reached.

danielnelson:
closes #5707

### Required for all PRs:

### DIFF
--- a/plugins/outputs/health/health.go
+++ b/plugins/outputs/health/health.go
@@ -167,10 +167,8 @@ func (h *Health) listen(tlsConf *tls.Config) (net.Listener, error) {
 }
 
 func (h *Health) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
-	var code int
-	if h.isHealthy() {
-		code = http.StatusOK
-	} else {
+	var code = http.StatusOK
+	if !h.isHealthy() {
 		code = http.StatusServiceUnavailable
 	}
 
--- a/plugins/outputs/health/health.go
+++ b/plugins/outputs/health/health.go
@@ -17,15 +17,18 @@ import (
 )
 
 const (
-	defaultReadTimeout  = 5 * time.Second
-	defaultWriteTimeout = 5 * time.Second
+	defaultServiceAddress = "tcp://:8080"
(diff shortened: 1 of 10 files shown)


plugins/outputs/health/health.go
```diff
@@ -0,0 +1,215 @@
...
+const (
+	defaultReadTimeout  = 5 * time.Second
+	defaultWriteTimeout = 5 * time.Second
+)
+
+var sampleConfig = `
+  ## Address and port to listen on
+  # service_address = ":8080"
+
+  ## The maximum duration for reading the entire ## request.
```
goller: Extra ##

plugins/outputs/health/health.go
```diff
@@ -0,0 +1,215 @@
...
+
+	return nil
+}
+
+func onAuthError(rw http.ResponseWriter, code int) {
+	http.Error(rw, http.StatusText(code), code)
+}
+
+func (h *Health) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
+	var code int
```
goller: What do you think of

```go
code := http.StatusOK
if !h.isHealthy() {
code = http.StatusServiceUnavailable
}
```

danielnelson:
Changes made.  Got carried away and did some of the other stuff we observed as well as added support for listening on a unix socket.

PR #5921, opened 2019-05-29
Parse the Redis optional replication section redis_replication measuremet

adamflott:
Parse the Redis replication nodes into individual pieces (ip, port, offset, lag) instead of a single string

### Required for all PRs:

- [ N/A] Associated README.md updated.


adamflott:
@glinton done

plugins/inputs/redis/redis.go
```diff
@@ -321,6 +327,48 @@ func gatherKeyspaceLine(
...
+) {
+	fields := make(map[string]interface{})
+	tags := make(map[string]string)
+	for k, v := range global_tags {
+		tags[k] = v
+	}
+
+	tags["replica_id"] = strings.TrimLeft(name, "slave")
+	// change our role for just the replicas (slaves)
+	tags["replication_role"] = "replica"
```
danielnelson: I think we should set this to `slave`, since connecting the plugin directly to the slave would currently set `role=slave`.

plugins/inputs/redis/redis.go
```diff
@@ -248,6 +248,12 @@ func gatherInfoOutput(
 				gatherKeyspaceLine(name, kline, acc, tags)
 				continue
 			}
+			if section == "Replication" && strings.HasPrefix(name, "slave") {
```
danielnelson: This pulls in some other lines that begin with `slave`, and so they no longer get processed.  We should add a unit test for slave redis.

plugins/inputs/redis/redis.go
```diff
@@ -321,6 +327,48 @@ func gatherKeyspaceLine(
...
+	for _, rp := range parts {
+		kv := strings.Split(rp, "=")
+		if len(kv) == 2 {
+			switch kv[0] {
+			case "ip":
+				tags["replica_ip"] = kv[1]
+			case "port":
+				tags["replica_port"] = kv[1]
+			case "state":
+				fields[kv[0]] = kv[1]
```
danielnelson: Consider this as a tag.

adamflott:
@danielnelson @glinton I believe I addressed all the feedback, please review.

PR #5941, opened 2019-06-01
Fix for #5692

### LINKED ISSUE #5692: Temperature input plugin stops working when WiFi is turned off
### Relevant telegraf.conf:

[[inputs.temp]]

### System info:

- Telegraf 1.10.1-1 installed from .deb package / Ubuntu 16.04 + Kernel 5.0
- Intel Wireless LAN hardware, iwlwifi module

### Steps to reproduce:

1. Enable WiFi, connect to a network
2. Enable the `inputs.temp` plugin
3. Observe that the plugin is working correctly: various `temp` metrics should exist, including one called `temp,sensor=iwlwifi_input`
4. Disconnect from the network, turn off WiFi

### Expected behavior:

No new metrics should be produced for `temp,sensor=iwlwifi_input`. New metrics should still be produced for everything else in `temp`.

### Actual behavior:

**No** new metric is produced for anything in `temp`.

### Additional info:

    Apr 08 14:36:50 telegraf[6305]: 2019-04-08T06:36:50Z E! [inputs.temp]: Error in plugin: error getting temperatures info: read /sys/class/hwmon/hwmon4/temp1_input: no data available

marcv81:
In case of non-critical temperature warnings: silently ignore them instead of dropping all the temperature metrics. I would appreciate friendly advice from the maintainers as to how to log the warnings instead please. I'm not sure whether to log them directly from Temperature() (probably not!) or to pass them anywhere else for logging.

closes #5692

### Required for all PRs:


marcv81:
Would appreciate some kind help with the CI tests please.
- To update `shirou/gopsutil` I ran `dep ensure -update github.com/shirou/gopsutil` and committed the resulting changes. How should I have approached this?
- I do not believe I modified `plugins/inputs/ping` or any of its dependencies. Am I wrong?
- I was able to run `make package` locally (Ubuntu Bionic), install the package, and no smoke came out.

danielnelson:
I think you did the `dep ensure` step correctly, but it looks maybe like your IDE reformatted some files for you automatically.  You will need to reset these files back to the unmodified versions (including Go.lock) so that it is only the temperature changes.

marcv81:
The issue was on outdated version of dep. Thanks for your comment which pointed me in the right direction @danielnelson. Also FYI I had to push the same change twice because the first build failed AppVeyor due to network issues.

danielnelson:
Can you tell me about the cpu stolen changes?

marcv81:
@danielnelson, gopsutil removed cpu.Stolen because it was declared but never set. The proposed change is required to update gopsutil as they introduced a breaking API change. The commit links to a PR in gopsutil, which in turns links to an issue with more information.

plugins/inputs/elasticsearch/README.md
```diff
@@ -219,7 +219,6 @@ Operating system stats, load average, cpu, mem, swap measurement names:
   - cpu_user value=0
   - cpu_idle value=99
   - cpu_usage value=0
-  - cpu_stolen value=0
```
danielnelson: I think this shouldn't be changed, these OS stats come from the Elasticsearch server and not gopsutil.

marcv81: Well spotted, thanks for the careful review. Updated accordingly.

PR #5767, opened 2019-04-25
prometheus metric_version = 2 & make scrape url as a configurable tag

vishiy:
### Required for all PRs:

- [*] Signed [CLA](https://influxdata.com/community/cla/).
- [*] Associated README.md updated.
- [* ] Has appropriate unit tests.

https://github.com/influxdata/telegraf/issues/4415
https://github.com/influxdata/telegraf/issues/5689
https://github.com/influxdata/telegraf/issues/5798

* Ensured the following -
- make check
- make test
- all prometheus input tests work (existing & new tests)

PR #5942, opened 2019-06-01
Add Prometheus remote write output plugin.

achilles42:
Add an output plugin that allows Telegraf to send metrics to systems which support Prometheus remote write format (such as Cortex and list of all remote writes are defined [here](https://prometheus.io/docs/operating/integrations/#remote-endpoints-and-storage)).

This is PR contains the following changes:
1. Contains [this PR ](https://github.com/influxdata/telegraf/pull/5003) changes. (Seems like no one is actively working on this PR. I have added missing features and bug fixes as part of this PR)
2. Add bearer Token support for authentication apart from username and password based authentication.
3. Sanitize the metrics to follow Prometheus format i.e having an underscore in metrics.
4. Added sort functionality in metrics to handle timestamp out of order issue causing issues in remote write.
5. Add feature toggle in the config file for retry logic if data ingestion failed in remote write because of various reasons, like - rate limiting, metrics out of order due to partial write of metrics.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,9 @@
 - [#6241](https://github.com/influxdata/telegraf/pull/6241): Add configurable timeout setting to smart input.
 - [#6249](https://github.com/influxdata/telegraf/pull/6249): Add memory_usage field to procstat input plugin.
 - [#5971](https://github.com/influxdata/telegraf/pull/5971): Add support for custom attributes to vsphere input.
+- [#5926](https://github.com/influxdata/telegraf/pull/5926): Add cmdstat metrics to redis input.
+- [#6261](https://github.com/influxdata/telegraf/pull/6261): Add content_length metric to http_response input plugin.
+- [#6257](https://github.com/influxdata/telegraf/pull/6257): Add database_tag option to influxdb_listener to add database from query string.
 
 #### Bugfixes
 
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -501,7 +501,13 @@
 [[projects]]
   digest = "1:6e73003ecd35f4487a5e88270d3ca0a81bc80dc88053ac7e4dcfec5fba30d918"
   name = "github.com/gogo/protobuf"
-  packages = ["proto"]
+  packages = [
+    "gogoproto",
+    "proto",
+    "protoc-gen-gogo/descriptor",
+    "sortkeys",
+    "types",
+  ]
   pruneopts = ""
   revision = "636bf0302bc95575d69441b25a2603156ffdddf1"
   version = "v1.1.1"
@@ -510,8 +516,12 @@
   digest = "1:f958a1c137db276e52f0b50efee41a1a389dcdded59a69711f3e872757dab34b"
   name = "github.com/golang/protobuf"
   packages = [
+    "jsonpb",
     "proto",
     "protoc-gen-go/descriptor",
+    "protoc-gen-go/generator",
+    "protoc-gen-go/generator/internal/remap",
+    "protoc-gen-go/plugin",
     "ptypes",
     "ptypes/any",
     "ptypes/duration",
@@ -593,6 +603,18 @@
   pruneopts = ""
   revision = "e3702bed27f0d39777b0b37b664b6280e8ef8fbf"
   version = "v1.6.2"
+
+[[projects]]
+  digest = "1:dee8ec16fa714522c6cad579dfeeba3caf9644d93b8b452cd7138584402c81f7"
+  name = "github.com/grpc-ecosystem/grpc-gateway"
+  packages = [
+    "internal",
+    "runtime",
+    "utilities",
+  ]
+  pruneopts = ""
+  revision = "8fd5fd9d19ce68183a6b0934519dfe7fe6269612"
+  version = "v1.9.0"
 
 [[projects]]
   branch = "master"
@@ -1033,6 +1055,14 @@
   revision = "ae68e2d4c00fed4943b5f6698d504a5fe083da8a"
 
 [[projects]]
+  digest = "1:b9aa2275e8c025c1b7514b6bc3fb04c3fbfc922891979472eb7227b7110c40ae"
+  name = "github.com/prometheus/prometheus"
+  packages = ["prompb"]
+  pruneopts = ""
+  revision = "d20e84d0fb64aff2f62a977adc8cfb656da4e286"
+  version = "v2.10.0"
+
+[[projects]]
   branch = "master"
   digest = "1:15bcdc717654ef21128e8af3a63eec39a6d08a830e297f93d65163f87c8eb523"
   name = "github.com/rcrowley/go-metrics"
@@ -1446,6 +1476,7 @@
   packages = [
     "googleapis/api/annotations",
     "googleapis/api/distribution",
+    "googleapis/api/httpbody",
     "googleapis/api/label",
     "googleapis/api/metric",
     "googleapis/api/monitoredres",
@@ -1708,10 +1739,12 @@
     "github.com/go-redis/redis",
     "github.com/go-sql-driver/mysql",
     "github.com/gobwas/glob",
+    "github.com/gogo/protobuf/proto",
     "github.com/golang/protobuf/proto",
     "github.com/golang/protobuf/ptypes/duration",
     "github.com/golang/protobuf/ptypes/empty",
     "github.com/golang/protobuf/ptypes/timestamp",
+    "github.com/golang/snappy",
     "github.com/google/go-cmp/cmp",
     "github.com/google/go-cmp/cmp/cmpopts",
     "github.com/google/go-github/github",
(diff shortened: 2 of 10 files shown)


achilles42:
Hi @danielnelson,

Currently above PR only supports counter and gauge. I am planning to add support for Histogram.

We are converting Influxdb metrics tags to Prometheus labels and field to new metrics with the appended field name in metrics with all the labels.

Ex- Influxdb metrics - 
`foobar tag1=value1 tag2=value2 field1=fvalue1 x`
Prometheus metrics - 
```
foobar{tag1=value1, tag2=value2} x
foobar_field1{tag1=value1, tag2=value2} fvalue1
```

For adding histogram support, I am planning to move fields to labels while converting metrics, so that we can query based on buckets, which won't be possible in case of field appended to metrics.

Please let me know WDYT about this approach.

danielnelson:
We should make sure we can round trip metrics from the prometheus input, there are a few things in motion right now, in particular #4415.  This will mean we have two ways of converting from prometheus metrics to influxdb/telegraf metrics.  On the way out, we should at least support the new transformation, and it could be a nice touch to support both.

achilles42:
Agree, we need a proper mapping between Influxdb/Telegraf metrics and Prometheus metrics.

>We should make sure we can round trip metrics from the Prometheus input

By this you mean, Telegraf input plugin scrape metrics and send to buffer, then prometheus_client output plugin read from the buffer and convert it into Prometheus format. After that Prometheus remote write plugin reads the converted metrics and push to remote write config?

Also this #4415 talks about Prometheus to Influxdb conversion. But in our case, we are converting from Influxdb to Prometheus format. Please correct me If I misunderstood anything.

danielnelson:
Sorry, what I meant by round tripping is that the input and the output transformations should mirror each other.  Ideally the total transformation would be equivalent to what Prometheus would do, and then Telegraf could be swapped in/out with Prometheus for this task.

achilles42:
no problem. So you are suggesting building central metrics format converter, which will convert from Influxdb format to Prometheus format and vice versa? as we have this logic spread out in 3 (Prometheus input, prometheus_client and prometheus_output plugins) places.

Should we build that as part of this PR? or we can go ahead without Histogram support for now and create separate PR for central metrics converter of Prometheus format?

danielnelson:
We would still have separate converters, but we want to make sure that input side and the output side are mirrored, as well as some level of consistency between the various prometheus output plugins.

As an example of output consistency, if you write to the prometheus_client output and then use prometheus to read and send using remote write, you should get the same results as if you were to use this plugin with remote write.

I think it is likely that some code can be shared between all these plugins and if we identify this we can always place it under the `internal` package.  However, the code sharing is also something we can work on over time, and will be easier to identify once everything has fully landed, so my number one concern is the behavioral consistency.

So that all is pretty abstract, and please keep in mind I haven't looked at the code closely yet, but I think we have two options here:
1. We support the current prometheus input and output format as well as the new format.  We use an option in the plugin config to switch between modes if needed.
2. We support only the future prometheus format, I think it is much better and will be much simpler for you to convert.  I'm planning to complete the pr for #4415 before I do this plugin.

jtyr:
@achilles42 Please amend and force-push the code again in order to trigger the CI pipeline. Hopefully the test will pass this time :crossed_fingers:

ihard:
I build telegraf with this PR and test for internal metric from grafana as example.

config:
```
[[inputs.prometheus]]
  interval = "30s"
  urls = ["http://127.0.0.1:3000/metrics"]
  [inputs.prometheus.tags]
   scrape_interval = "30s"
   job = "grafana"

[[outputs.prometheus_remote_write]]
  flush_interval = "30s"
  url = "https://localhost:9090"
```

Get metric
```
curl -s 127.1:3000/metrics
```

There are some problems with metric names:

summary
```
# HELP http_request_duration_milliseconds http request summary
# TYPE http_request_duration_milliseconds summary
http_request_duration_milliseconds{handler="/",method="get",statuscode="200",quantile="0.5"} 4
http_request_duration_milliseconds{handler="/",method="get",statuscode="200",quantile="0.9"} 4
http_request_duration_milliseconds{handler="/",method="get",statuscode="200",quantile="0.99"} 4
http_request_duration_milliseconds_sum{handler="/",method="get",statuscode="200"} 280
http_request_duration_milliseconds_count{handler="/",method="get",statuscode="200"} 71
```

transformation to:
```
# metric name good
http_request_duration_milliseconds{dc="dc1",handler="/",host="metric",job="grafana",method="get",quantile="0.5",scrape_interval="30s",statuscode="200",url="http://127.0.0.1:3000/metrics"}
http_request_duration_milliseconds{dc="dc1",handler="/",host="metric",job="grafana",method="get",quantile="0.9",scrape_interval="30s",statuscode="200",url="http://127.0.0.1:3000/metrics"}
http_request_duration_milliseconds{dc="dc1",handler="/",host="metric",job="grafana",method="get",quantile="0.99",scrape_interval="30s",statuscode="200",url="http://127.0.0.1:3000/metrics"}
# metric name mismatch
http_request_duration_milliseconds{dc="dc1",handler="/",host="metric",job="grafana",method="get",quantile="count",scrape_interval="30s",statuscode="200",url="http://127.0.0.1:3000/metrics"}
http_request_duration_milliseconds{dc="dc1",handler="/",host="metric",job="grafana",method="get",quantile="sum",scrape_interval="30s",statuscode="200",url="http://127.0.0.1:3000/metrics"}
```


counter
```
# HELP http_request_total http request counter
# TYPE http_request_total counter
http_request_total{handler="/",method="get",statuscode="200"} 67
```

transformation to:
```
# metric name mismatch
http_request_total_counter{dc="dc1",handler="/",host="metric",job="grafana",method="get",scrape_interval="30s",statuscode="200",url="http://127.0.0.1:3000/metrics"}
```

gauge
```
# HELP go_memstats_heap_objects Number of allocated objects.
# TYPE go_memstats_heap_objects gauge
go_memstats_heap_objects 92864
```

transformation to:
```
# metric name mismatch
go_memstats_heap_objects_gauge{dc="dc1",host="metric",job="grafana",scrape_interval="30s",url="http://127.0.0.1:3000/metrics"}
```

achilles42:
Hey @danielnelson We have also added support of Histogram and summary metrics and used most of the Promtheus output plugin code in order to avoid duplication in metrics conversion.
Please advise on the next step to take it forward and merge.

achilles42:
Hey, @danielnelson can you please help us here?

danielnelson:
@achilles42 Hoping to review soon, thanks for your patience.  We are also planning to merge #5767 soon which adds an new format for prometheus data, would be good if this plugin could "round-trip" this style of metric too, most likely with an option to switch modes.

danielnelson:
@achilles42 As mentioned in my last comment, we have merged #5767 with the newer format for 1.13, and I have opened a pull request for a serializer plugin in #6703.

The serializer supports round tripping the `metric_version=2` setting in the prometheus input and, I hope, could be used by this pull request to implement the telegraf.Metric -> prometheus style metric conversion.  In particular, the `Collection` type can be used to convert to `[]*dto.MetricFamily`, which I know isn't exactly the same as the required `prompb.WriteRequest` type used in this pull request, but I believe it would be much easier to convert to as the differences are much less than compared directly to the telegraf.Metric.

The #6703 pull request was just opened, so it's still subject to change, but can you take a look at it and let me know if you think we could use that here?

ihard:
@achilles42,  is there a chance that this request will be updated for compatibility with metric_version=2 ?

valyala:
FYI, there is [vmagent](https://github.com/VictoriaMetrics/VictoriaMetrics/blob/master/app/vmagent/README.md) tool, which can accept Influx line protocol data and replicate it to any number of Prometheus-compatible remote storage systems via 'remote_write API`.

joakimberglund:
Any progress on this?

johntdyer:
@danielnelson - So based on commend https://github.com/influxdata/telegraf/pull/5942#issuecomment-557352904 it seems like there was some changes pending on #6703.  I noticed that it is now merged so is there any plans to circle back around and get this merged ?

aarnaud:
>  In particular, the `Collection` type can be used to convert to `[]*dto.MetricFamily`, which I know isn't exactly the same as the required `prompb.WriteRequest` type used in this pull request, but I believe it would be much easier to convert to as the differences are much less than compared directly to the telegraf.Metric.

I tried to update or create a new PR to use HTTP output and prometheus serializer to have compatibility like remote_write, for use with @cortexproject 

I not success to converte `[]*dto.MetricFamily` to `prompb.WriteRequest` that need to add `[]TimeSeries` that contains `[]Label` and `[]Samples`

aarnaud:
Hi, 
I create a PR https://github.com/influxdata/telegraf/pull/8346 that use the prometheus serializer and http output for remote_write
I tested with Cortex

sjwang90:
Thanks @achilles42 so much for the PR. 
Closing this now that the `prometheusremotewrite` serializer #8360 is merged. If this does not fully fulfill your use case please feel free to re-open.

PR #5945, opened 2019-06-02
Upgrade aws-go-sdk to 1.19.41

gtaylor:
We are unable to pull some of the new ELB metrics (like SpilloverCount) due to being pretty far behind in AWS SDK versions. This metric in particular has been helpful in debugging some recent issues, so we'd love to get it flowing through telegraf.

I've ran a build, the tests, and smoke tested this a bit with the plugin I'm working on. _Seems_ to be OK.

As an added bonus: this also sheds the go-ini/ini dependency, since aws-go-sdk no longer needs it (and was the sole user).

### Required for all PRs:

- [X] (Not applicable) Associated README.md updated.
- [X] (Not applicable) Has appropriate unit tests.


CHANGELOG.md
```diff
@@ -56,6 +56,7 @@
 - [#5901](https://github.com/influxdata/telegraf/pull/5901): Set user agent in stackdriver output.
 - [#5885](https://github.com/influxdata/telegraf/pull/5885): Extend metrics collected from Nvidia GPUs.
 - [#5547](https://github.com/influxdata/telegraf/pull/5547): Add file rotation support to the file output.
+- [#5945](https://github.com/influxdata/telegraf/pull/5945): Upgrade from aws-sdk-go 1.15.54 to 1.19.41.
```
danielnelson: Can you remove this, we would only add it to the changelog if it addressed a particular bug or feature.  Also it's just a pain to avoid conflicts in this file so we usually update it post merge.

Gopkg.toml
```diff
@@ -16,7 +16,7 @@
 
 [[constraint]]
   name = "github.com/aws/aws-sdk-go"
-  version = "1.15.54"
+  version = "1.19.41"
```
danielnelson: Undo this change, since we don't really require a newer version than this to build.  When you open a pull request that requires the newer version you can then set it to, ideally, the first version that would support the feature you need.

The idea is to allow builders flexibility in the version they use in case they decide not to use our lock values, which is nice for system integrators.

gtaylor: Aside from the new input we're working on, the more immediate need for the new version of the aws SDK is to pull some new ELB metrics for the existing CloudWatch input. This does not require a change to the CloudWatch telegraf input, does require a newer version of the AWS SDK.

CHANGELOG.md
gtaylor: Dropped this.

Gopkg.toml
gtaylor: Updated the PR description to focus more on this particular issue.

danielnelson: Can you tell me a bit more about this?  What is you want to do and what feature of the current sdk is missing?

danielnelson: Which version is required for the SpilloverCount metric to work?

PR #5951, opened 2019-06-04
Add disk host to fields

zombielinux:
Allow for easier monitoring of large disk deployments by allowing tag filtering based on originating host.

### Required for all PRs:


danielnelson:
I think we should skip modifying the Disk type and handle this all in the top hddtemp.go, we should also use the `source` tag to hold only the hostname portion of the address as this is something we are attempting to standardize on.  It was a big easier for me to describe with a pull request, so I just put it in #5955, can you take a look?

PR #5971, opened 2019-06-07
Added support for custom attributes

prydin:
### Required for all PRs:

Custom attributes is an important construct that allows administrators to tag resources workloads according to custom schemes. This can be used to group assets, tie containers to VMs, classify workload security requirements among many other things. Being able to use these in queries in target solutions such as InfluxDB has been desired by several users of the vSphere plugin.

Custom attribute collection is controlled by the ```custom_attribute_include``` and ```custom_attribute_exclude``` configuration settings.


plugins/parsers/wavefront/parser.go
```diff
@@ -98,6 +98,8 @@ func (p *PointParser) Parse(buf []byte) ([]telegraf.Metric, error) {
 		buf = append(buf, []byte("\n")...)
 	}
 
+	//log.Printf("D! [parsers.wavefront] Received data: %s", string(buf))
```
danielnelson: Delete me

plugins/inputs/vsphere/README.md
```diff
@@ -118,9 +118,13 @@ vm_metric_exclude = [ "*" ]
     "storageAdapter.write.average",
     "sys.uptime.latest",
   ]
+  ## Only consider IPv4 addresses when trying to map NICs to IP addresses.
+  # ipv4_only = true
```
danielnelson: Why would you only want to do IPv4?  Should this be something along the lines of `nic_address = ["ipv4", "ipv6"]` so you can have both/neither/ipv6 only?  If we are excluding due to cardinality maybe we could have a mask `nic_cidr_mask = "/24"` that converts 192.168.0.56 to xxx.xxx.xxx.56?

prydin: This is a bit of a hack to reduce the number of possible IP addresses reported for a VM. I think I will remove this code from this PR, since I need to think a little deeper on the design.

prydin: @danielnelson I'm working on implementing this the "right" way. What's your convention on lists of things appearing in tags? For example, a NIC can sometimes have more than one ipv4 address. The way I go about it now is to emit tags like this: ipv4_0 = 10.10.10.10, ipv4_1 = 10.10.10.11 and so on. I.e. I'm just appending a sequence number to the tag key. Is that the correct way of dealing with this?

danielnelson: Unfortunately, there isn't really a great way.  We had something similar come up recently on the Elasticsearch input where a node can have one or more roles, and we settled on comma separated values: `roles=data,ingest,master` taking care to sort the values to avoid unneeded cardinality.

It is possible to use a regex query in InfluxDB to lookup these values, there are corner cases to it but in general it works okay.  Not sure if other TSDB's have similar functionality.
```
select * from elasticsearch where roles =~ /master/
```

I think it would still make sense to have an option that limits the addresses.  Having a CIDR mask could still work well for this and Go has the functionality to check if an address matches.

prydin: The TSDB I'm working with can handle that, so as far as I'm concerned it's OK. :) I just want to make sure we're handling it in a way that's consistent with the rest of Telegraf. Also, the ipv6_n notation is pretty ugly and can potentially lead to exploding cardinalities.

prydin: Cleaned this up as follows: 
1) IP addresses are now collected into two tags names ```ipv4``` and ```ipv6``` and reported per network interface. In other words, metrics on ```vsphere.vm.net.*``` may have these tags. Notice that they are only populated on NICs visible through the ```guest.net.ipConfig``` property on a VM in vSphere. 
2) Added a config attribute called ```ip_addresses```. This allows you to specify which type of IP addresses to collect. This is a list of strings and valid values are ```ipv4``` and ```ipv6```. For example, to collect both ipv6 and ipv4 addresses, one would add the following line to the config file:

```ip_addresses = ["ipv4", "ipv6"]```

3) While I was messing with this, I also added support for the guest hostname through a tag called ```guesthostname```. This is the hostname as reported from inside the guest OS and may be different from the VM name used as the primary identifier for a VM.

I consider this resolved.

plugins/parsers/wavefront/parser.go
prydin: Done.

prydin:
@danielnelson This PR will most likely generate a nasty merge with the vSAN stuff that @YuxinZhou is working on. How do you want me to handle that? Just wait for her merge to clear and then clean up the merge on my fork? Or you want us to merge our two branches into one?

danielnelson:
Unfortunately, I'm not going to have time to help finish the vSAN code for the upcoming 1.12 release.  Let's merge this in and have @YuxinZhou base his code on it.

Issue #5300, opened 2019-01-16
procstat_lookup always shows 1 when monitoring systemd_unit


Frakenz:
### System info:
[Include Telegraf version, operating system name, and other relevant details]
Telegraf 1.9.1 (git: HEAD 20636091)
CentOS 7
InfluxDB shell version: 1.7.2

### Steps to reproduce:

1. `sudo systemctl stop puppet.service`
2. Wait for telegraf interval to update, in my case 1 minute
3. Enter into influxdb CLI with the corresponding database and run this query:
`SELECT last("pid_count") FROM "autogen"."procstat_lookup" WHERE ("systemd_unit" = 'puppet.service') AND time > now()-2m GROUP BY "myTag"`

### Expected behavior:

> name: procstat_lookup
> tags: myTag1
> time                last
> ----                ----
> 1547670360000000000 0

### Actual behavior:

> name: procstat_lookup
> tags: myTag1
> time                last
> ----                ----
> 1547670360000000000 1

### Additional info:
I only learned a bit of GO to read the source code, but from what I understood, the problem is that [here](https://github.com/influxdata/telegraf/blob/e95b88e01b2c664e43ea45faf77569d8db44a656/plugins/inputs/procstat/procstat.go#L132) len(pids) is not enough, since running systemctl show puppet.service when the service is not running will give MainPID=0 (see also #3612) and `systemdUnitPIDs() ` does not check for zero values.

[Include gist of relevant config, logs, etc.]
### Relevant telegraf.d/puppet.conf:
[[inputs.procstat]]
  systemd_unit = "puppet.service"

glinton:
There are cases where `MainPID=0` when the service is running, which seems to be caused by using the incorrect systemd unit, or incorrectly constructed file. Maybe `systemdUnitPIDs()` should check for zero values to encourage the proper use of the units/files. ...or maybe it should check for the pids in a different way as many of the units/files are created by default when installing a service.

Frakenz:
A different way to check if the service is running could be by [checking](https://unix.stackexchange.com/a/396633) the `SubState` or the `ActiveState` in the same way that the `MainPID` is being currently obtained.
In any case it would be good if someone could confirm if this is a bug or if it has something to do with my setup, so I can know how to proceed :)

glinton:
[...]

PR #5972, opened 2019-06-08
Fix pid_count with systemd_unit in procstat_lookup

Frakenz:
Attempt at fixing #5300. Not sure if using an OR statement or writing a separate if was a better practice. Since both check for valid PIDs I used an OR statement.
Also I tested this on golang and know almost nothing of GO, so there might be better ways to write this.

### Required for all PRs:

### DIFF
--- a/plugins/inputs/procstat/procstat.go
+++ b/plugins/inputs/procstat/procstat.go
@@ -398,7 +398,7 @@ func (p *Procstat) systemdUnitPIDs() ([]PID, error) {
 		if !bytes.Equal(kv[0], []byte("MainPID")) {
 			continue
 		}
-		if len(kv[1]) == 0 {
+		if len(kv[1]) == 0 || bytes.Equal(kv[1], []byte("0")){
 			return nil, nil
 		}
 		pid, err := strconv.Atoi(string(kv[1]))
--- a/plugins/inputs/procstat/procstat.go
+++ b/plugins/inputs/procstat/procstat.go
@@ -398,7 +398,7 @@ func (p *Procstat) systemdUnitPIDs() ([]PID, error) {
 		if !bytes.Equal(kv[0], []byte("MainPID")) {
 			continue
 		}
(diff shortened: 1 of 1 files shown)


Frakenz:
Is there a chance this failed because I made the edit directly on github's website?
What am I doing wrong? 

```[ERROR] gofmt has found errors in the following files:
plugins/inputs/procstat/procstat.go
Run make fmt to fix them.
make: *** [fmtcheck] Error 1
Command exited with code 2

**Comment by danielnelson** [comment]
I think you just need a space before the `{`:
```diff
- if len(kv[1]) == 0 || bytes.Equal(kv[1], []byte("0")){
+ if len(kv[1]) == 0 || bytes.Equal(kv[1], []byte("0")) {
```

xczdenis:
It seems to me that it doesn't work as expected. When I use influx db as output client, it works fine. However, I also use `prometheus_client`:
```
[[outputs.prometheus_client]]
    listen = ":9193"
    path = "/metrics"
    expiration_interval = "0s"
```

I want to control the status of `nexus.service`. When the nexus is running, I get a response on url `my-server-ip:9193/metrics` like this:
```
# HELP systemd_units_active_code Telegraf collected metric
# TYPE systemd_units_active_code untyped
systemd_units_active_code{active="active",load="loaded",name="nexus.service",sub="running"} 0
```

But if I shut down the service, I get this:
```
# HELP systemd_units_active_code Telegraf collected metric
# TYPE systemd_units_active_code untyped
systemd_units_active_code{active="active",load="loaded",name="nexus.service",sub="running"} 0
systemd_units_active_code{active="inactive",load="loaded",name="nexus.service",sub="dead"} 2
```

So when I use this query: `systemd_units_active_code{name="nexus.service"}` I get 2 rows every time period - `active` and `inactive`. Therefore, I can't define the correct current state. Maybe i'm doing smth wrong.

PR #5975, opened 2019-06-10
Support array of addressses in http_response input

### LINKED ISSUE #5828: Make the http_response plugin support arrays of adresses
## Feature Request

It would be nice if the http_response plugin would have a array of addresses instead of one single. With an array of addresses you could build a simple pingdom like dashboard for a range of sites.

### Proposal:
Make address become an array field instead of string.

### Current behavior:
Addresss field is a string.

### Desired behavior:
Address field shoud be an array - and also support arrays in the code.

### Use case:
Build a pingdom like dashboard for a range of sites. Simpler handling of alerts if sites are down.

glinton:
Resolves #5828

### DIFF
--- a/plugins/inputs/http_response/http_response.go
+++ b/plugins/inputs/http_response/http_response.go
@@ -294,7 +294,7 @@ func (h *HTTPResponse) Gather(acc telegraf.Accumulator) error {
 		if h.Address == "" {
 			h.URLs = []string{"http://localhost"}
 		} else {
-			log.Printf("E! [inputs.http_response] 'address' deprecated in telegraf 1.12, please use 'urls'")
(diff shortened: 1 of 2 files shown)


plugins/inputs/http_response/http_response.go
```diff
@@ -284,20 +289,15 @@ func (h *HTTPResponse) Gather(acc telegraf.Accumulator) error {
...
-	}
 
-	// Prepare data
-	var fields map[string]interface{}
-	var tags map[string]string
+	if len(h.URLs) == 0 {
+		if h.Address == "" {
+			h.URLs = []string{"http://localhost"}
+		} else {
+			log.Printf("E! [inputs.http_response] 'address' deprecated in telegraf 1.12, please use 'urls'")
```
danielnelson: Use `W!`

PR #5986, opened 2019-06-13
Omit keys when creating measurement names for GNMI telemetry

### LINKED ISSUE #5983: inputs.cisco_telemetry_gnmi and wildcard in path - measurement name always empty
### Relevant telegraf.conf:
```
# Send telegraf metrics to file(s)
[[outputs.file]]
  ## Files to write to, "stdout" is a specially handled file.
  files = ["/tmp/metrics.out"]
  data_format = "influx"


<!-- [Place config betweetn triple backticks '```'] -->
[[inputs.cisco_telemetry_gnmi]]
  ## Address and port of the GNMI GRPC server
  addresses = ["xxxx:57400"]

  ## define credentials
  username = "auto"
  password = "T3l3M!%eT5y!"

  ## GNMI encoding requested (one of: "proto", "json", "json_ietf")
  encoding = "json"

  ## redial in case of failures after
  redial = "10s"

  [[inputs.cisco_telemetry_gnmi.subscription]]
    ## Name of the measurement that will be emitted
    name = "trafficstats"

    #origin = "openconfig-interfaces"
    path = "/state/port[port-id=*]/ethernet/statistics"

    # Subscription mode (one of: "target_defined", "sample", "on_change") and interval
    subscription_mode = "sample"
    sample_interval = "30s"

    ## Suppress redundant transmissions when measured values are unchanged
    # suppress_redundant = false

    ## If suppression is enabled, send updates at least every X seconds anyway
    # heartbeat_interval = "60s"
```

### Steps to reproduce:

I tried to use the inputs.cisco_telemetry_gnmi to collect telemetry data on Nokia SROS gNMI compliant router. But on Nokia we need to provide for some paths wildcard key for some elem :

"/state/port[port-id=*]/ethernet/statistics"

and therefore the measurement name is always empty because in line 346 :
 ``` 
		if _, exists := c.aliases[name]; exists {
			aliasPath = name
		}
```
name is never match /state/port[port-id=*]/ethernet/statistics because here name is just an instance for each port and finally aliasPath is empty and measurement name is empty which triggers issue when output encoding is influx. 

if might be interesting in line 113: 
```
		if len(name) > 0 {
>>> here remove the key filter information from the provided path. 
>>> here path = "/state/port[port-id=*]/ethernet/statistics"
>>> before provisioning c.aliases remove all key filter from the path : 
>>> fix path to become : "/state/port/ethernet/statistics"
			c.aliases[path] = name
                 }
```

### PR DESCRIPTION
**Pull request by sbyx**
Fixes #5983.
Passing configured subscriptions paths through the same parser as incoming paths when creating aliases. This should make sure they match up.

### DIFF
--- a/plugins/inputs/cisco_telemetry_gnmi/cisco_telemetry_gnmi.go
+++ b/plugins/inputs/cisco_telemetry_gnmi/cisco_telemetry_gnmi.go
@@ -101,14 +101,16 @@ func (c *CiscoTelemetryGNMI) Start(acc telegraf.Accumulator) error {
 	// Invert explicit alias list and prefill subscription names
 	c.aliases = make(map[string]string, len(c.Subscriptions)+len(c.Aliases))
 	for _, subscription := range c.Subscriptions {
-		var path string
 		// Build the subscription path without keys
-		if parsed, _ := parsePath(subscription.Origin, subscription.Path, ""); parsed != nil {
-			path, _ = c.handlePath(parsed, nil, "")
+		gnmiPath, err := parsePath(subscription.Origin, subscription.Path, "")
+		if err != nil {
+			return err
 		}
 
-		// If the user didn't provide a measurement name, use last path element
+		path, _ := c.handlePath(gnmiPath, nil, "")
 		name := subscription.Name
+
+		// If the user didn't provide a measurement name, use last path element
 		if len(name) == 0 {
 			name = path[strings.LastIndexByte(path, '/')+1:]
 		}
--- a/plugins/inputs/cisco_telemetry_gnmi/cisco_telemetry_gnmi.go
+++ b/plugins/inputs/cisco_telemetry_gnmi/cisco_telemetry_gnmi.go
@@ -101,11 +101,13 @@ func (c *CiscoTelemetryGNMI) Start(acc telegraf.Accumulator) error {
 	// Invert explicit alias list and prefill subscription names
 	c.aliases = make(map[string]string, len(c.Subscriptions)+len(c.Aliases))
 	for _, subscription := range c.Subscriptions {
-		path := subscription.Path
-		if len(subscription.Origin) > 0 {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: plugins/inputs/cisco_telemetry_gnmi/cisco_telemetry_gnmi.go [hunk 1]
```diff
@@ -101,11 +101,13 @@ func (c *CiscoTelemetryGNMI) Start(acc telegraf.Accumulator) error {
 	// Invert explicit alias list and prefill subscription names
 	c.aliases = make(map[string]string, len(c.Subscriptions)+len(c.Aliases))
 	for _, subscription := range c.Subscriptions {
-		path := subscription.Path
-		if len(subscription.Origin) > 0 {
-			path = subscription.Origin + ":" + path
+		var path string
+		// Build the subscription path without keys
+		if parsed, _ := parsePath(subscription.Origin, subscription.Path, ""); parsed != nil {
```
danielnelson: If an error occurs here, it seems like we return it because that means the path is invalid in the configuration and an invalid config value should be fatal.

sbyx:
In practice this should not be able to happen since we have are checking this in:

https://github.com/cisco-ie/telegraf/blob/ab3b634cb40528ccafd6fc682439df097a145f6e/plugins/inputs/cisco_telemetry_gnmi/cisco_telemetry_gnmi.go#L84

and

https://github.com/cisco-ie/telegraf/blob/ab3b634cb40528ccafd6fc682439df097a145f6e/plugins/inputs/cisco_telemetry_gnmi/cisco_telemetry_gnmi.go#L148

So technically we run parsePath twice and check for the error and abort in the first run.

danielnelson:
I would check the error again anyway, it is less likely to cause a bug if the code is refactored and avoids questions about why the error is not being handled.

sbyx:
Alright, done. I guess it doesn't hurt anyway.

PR #6004, opened 2019-06-18
Elasticsearch Input changes

### LINKED ISSUE #2711: [elastic plugin] cluster health index-level stats + index node stats are merged into one measurement
This was somewhat surprising. I noticed it by seeing clusterhealth-like fields in elasticsearch_indices and then after a bit of digging (essentially checking what is tagged with 'index') it turned out (and plugin code confirmed it) that elasticsearch_indices is essentially a "merge" of 2 queries:

- /_cluster/health?level=indices (tagged per index)
- /_nodes/stats/indices (tagged per node's host/id/name)

But the former is on cluster level, the latter is on node level. So mixing them is somewhat unusual choice. Both parts are also completely disjoint (no common fields) 

**Weren't they supposed to go (tagged per index) to elasticsearch_cluster_health ?**

P.S.
A bit of a side not - but detailed per index stats can be obtained (and tagged with index name) by doing e.g. /_stats query on a node. It's also possible (though level option is documented only for index api) to do e.g. /_nodes/_local/stats/indices?level=indices . This would likely be more fitting for elasticsearch_indices - giving per node + per index stats.

dupondje:
closes #2711
closes #2650

### Required for all PRs:

Fixing some long-standing issues in the elasticsearch plugin.
Still a lot of work todo, as it seems like ES6 isn't working correctly anymore?

Any chance to review (and eventrually merge) this first?
Cause I don't want to spend more time on fixing next issues if the base isn't going to be accepted :)


plugins/inputs/elasticsearch/elasticsearch.go
```diff
@@ -186,25 +193,37 @@ func (e *Elasticsearch) Gather(acc telegraf.Accumulator) error {
...
+			go func(s string, acc telegraf.Accumulator) {
+				info := serverInfo{}
 
-	for _, serv := range e.Servers {
-		go func(s string, acc telegraf.Accumulator) {
-			defer wg.Done()
-			url := e.nodeStatsUrl(s)
-			e.isMaster = false
+				// Gather node ID
+				if err := e.gatherNodeID(&info, s+"/_nodes/_local/name"); err != nil {
```
danielnelson: Is this information also in gatherNodeStats?   If so, could we reorder this to go over the node stats first and then go over the cluster stats, skipping one call?

plugins/inputs/elasticsearch/elasticsearch.go
```diff
@@ -267,6 +286,22 @@ func (e *Elasticsearch) nodeStatsUrl(baseUrl string) string {
 	return fmt.Sprintf("%s/%s", url, strings.Join(e.NodeStats, ","))
 }
 
+func (e *Elasticsearch) gatherNodeID(info *serverInfo, url string) error {
```
danielnelson: Can you have this and also `setCatMaster` return just the nodeID and the masterID, we can then put them together in the calling function and avoid using out parameters.

When updating the struct field, we will need a mutex to avoid a race error since these are being called in a goroutine.

plugins/inputs/elasticsearch/elasticsearch.go
```diff
@@ -221,7 +241,7 @@ func (e *Elasticsearch) Gather(acc telegraf.Accumulator) error {
 				}
 			}
 
-			if e.ClusterStats && (e.isMaster || !e.ClusterStatsOnlyFromMaster || !e.Local) {
+			if e.ClusterStats && (e.serverInfo[s].isMaster() || !e.ClusterStatsOnlyFromMaster || !e.Local) {
```
danielnelson: Check this out, there is a way to do this in the request itself in >=6.3.  Unfortunately, I think we still want support for older versions.

https://www.elastic.co/guide/en/elasticsearch/reference/6.8/cluster.html#cluster-nodes

plugins/inputs/elasticsearch/elasticsearch.go
dupondje: It is, but then it should have been called by /_nodes/_local/stats. Cause that seems to be the only way to gather the local nodeID instead of all.
So we could only skip it if we use _local, otherwise this additional call is always needed.
Might be cleaner to keep it just this way?

danielnelson: Makes sense, flow is already hard enough to follow so let's keep it how you have it.

plugins/inputs/elasticsearch/elasticsearch.go
dupondje: I would prefer getting this merged first (as this will work for all versions).
My next pull req would be adding 6+ support, and will check to implement this then :)

danielnelson: Yeah, unfortunately I don't think we can really use this at all until we drop ES5 support.

plugins/inputs/elasticsearch/elasticsearch.go
```diff
@@ -186,25 +195,44 @@ func (e *Elasticsearch) Gather(acc telegraf.Accumulator) error {
...
+				if info.masterID, err = e.getCatMaster(s + "/_cat/master"); err != nil {
 					acc.AddError(fmt.Errorf(mask.ReplaceAllString(err.Error(), "http(s)://XXX:XXX@")))
 					return
 				}
-			}
 
-			// Always gather node states
+				e.serverInfoMutex.Lock()
+				e.serverInfo[s] = info
+				e.serverInfoMutex.Unlock()
```
danielnelson: I notice now that we don't wait for these goroutines to complete before moving on to the next section of the code, this means that we still might be writing the serverInfo by the time we try to read from it later on.  It is possible that we still haven't figured out who the master is on line 252.

It looks like we will need to use a sync.WaitGroup and then wait for all goroutines in this loop before moving into the next for loop.

dupondje: Correct! I missed that indeed. Fixed now.

Issue #5856, opened 2019-05-15
Support binding to a specific local address for `http_response`


zoonage:
## Feature Request

Opening a feature request kicks off a discussion.

### Proposal:
Add a bind interface/IP address to the `http_response` plugin to force the request to egress via a specific interface

```
[[inputs.http_response]]
  address = "http://google.com"
  bind = "eth0"
```

### Current behavior:
The request uses the default `net.Dialer` which selects an interface for the request.

### Desired behavior:
To be able to specify an interface or IP address as the address to bind the `http.Client` to.

### Use case: 
For hosts that have VPN tunnels for a specific interface's traffic it is useful to check that you can send a request via this interface

danielnelson:
Sounds good, I think we should name the option `interface` since bind is something the server usually does.  It seems like this is implemented with the `LocalAddr` field on the dialer?  I guess we would use the first address assigned to the interface?

Is this something you could put together a pull request for?

zoonage:
@danielnelson I'm unsure due to work commitments, but I'll give it a go

PR #6006, opened 2019-06-18
Add support for interface field in http_response input plugin

GeorgeMac:
This is an attempt to implement the `interface` field in the `http_response` input plugin.

The `interface` field can be populated with an interface name such as `eth0`. A local address associated with the network interface will be derived and used for dialing.

Resolves: https://github.com/influxdata/telegraf/issues/5856

### Required for all PRs:

### DIFF
--- a/plugins/inputs/http_response/http_response_test.go
+++ b/plugins/inputs/http_response/http_response_test.go
@@ -217,7 +217,7 @@ func findInterface() (net.Interface, error) {
 
 	for _, i := range potential {
 		// ignore interfaces which are down or not used for broadcast
-		if (i.Flags&net.FlagUp == 0) || (i.Flags&net.FlagBroadcast == 0) {
+		if (i.Flags&net.FlagUp == 0) || (i.Flags&net.FlagLoopback == 0) {
 			continue
 		}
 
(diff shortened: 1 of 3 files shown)


plugins/inputs/http_response/http_response_test.go
```diff
@@ -210,6 +212,69 @@ func TestFields(t *testing.T) {
...
+	var (
+		mux = setUpTestMux()
+		ts  = httptest.NewServer(mux)
+	)
+
+	defer ts.Close()
+
+	intf, err := findInterface()
+	if err != nil {
+		t.Fatal(err)
```
glinton: I likely would have just `t.Skip(err)`, but then again I believe `httptest.NewServer` requires at least the loopback interface, so this is fine.

GeorgeMac: I can change that if skip is preferred. I didn’t realize skip existed in that form.

GeorgeMac: I can change that if skip is preferred. I didn’t realize skip existed in that form.

plugins/inputs/http_response/http_response_test.go
```diff
@@ -210,6 +212,69 @@ func TestFields(t *testing.T) {
...
+
+func TestInterface(t *testing.T) {
+	var (
+		mux = setUpTestMux()
+		ts  = httptest.NewServer(mux)
+	)
+
+	defer ts.Close()
+
+	intf, err := findInterface()
```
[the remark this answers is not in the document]
GeorgeMac: Absolutely. That makes more sense. I was trying to test a non-loopback interface, which in hindsight makes no sense. I will make it get the name of the first loopback interface it finds.

PR #6007, opened 2019-06-18
 Add separate ping_native input plugin

glinton:
This plugin provides ping capability without requiring a `ping` binary installed and should work with multiple languages as it isn't parsing output. It should also alleviate open-file issues when pinging multiple hosts.

This has been tested on Windows 10, macOS Sierra, and Ubuntu. IPv6 testing was on Ubuntu 18.04.

### DIFF
--- a/plugins/inputs/ping_native/README.md
+++ b/plugins/inputs/ping_native/README.md
@@ -2,7 +2,7 @@
 
 Sends a ping message and reports the results. This is done in pure go, eliminating the need to execute the system `ping` command.
 
-There is currently no support for TTL on windows, track progress https://github.com/golang/go/issues/7175 and https://github.com/golang/go/issues/7174
+There is currently no support for TTL on windows, track progress at https://github.com/golang/go/issues/7175 and https://github.com/golang/go/issues/7174
 
 
 ### Configuration:
@@ -33,7 +33,7 @@ There is currently no support for TTL on windows, track progress https://github.
 
 #### Permission Caveat (non Windows)
 
-Since this plugin listens on unprivileged raw sockets on linux and darwin, the system group of the user running telegraf must be allowed to create ICMP Echo sockets. [See man pages icmp(7) for `ping_group_range`](http://man7.org/linux/man-pages/man7/icmp.7.html) for linux and [man pages icmp(4) for ``](https://www.freebsd.org/cgi/man.cgi?query=icmp&apropos=0&sektion=0&manpath=Darwin+8.0.1%2Fppc&format=html) for darwin.
+Since this plugin listens on unprivileged raw sockets on Linux, the system group of the user running telegraf must be allowed to create ICMP Echo sockets. [See man pages icmp(7) for `ping_group_range`](http://man7.org/linux/man-pages/man7/icmp.7.html). On Linux hosts, run the following to give a group the proper permissions:
 
 ```
 sudo sysctl -w net.ipv4.ping_group_range="GROUPID   GROUPID"
@@ -43,7 +43,7 @@ sudo sysctl -w net.ipv4.ping_group_range="GROUPID   GROUPID"
 
 - ping
   - tags:
-    - ip
+    - source
   - fields:
     - packets_transmitted (integer)
     - packets_received (integer)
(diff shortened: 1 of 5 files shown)


plugins/inputs/ping_native/ping.go
```diff
@@ -0,0 +1,260 @@
...
+		p.listenAddr = getAddr(p.Interface)
+	}
+
+	p.buildHostCache(acc)
+	if len(p.hostCache) == 0 {
+		return errors.New("no valid hosts to ping")
+	}
+
+	conn, err := ping.Listen(p.network, p.listenAddr)
+	defer conn.Close()
```
danielnelson: Add defer after checking error.

plugins/inputs/ping_native/ping.go
```diff
@@ -0,0 +1,260 @@
...
+	}
+
+	if p.Count < 0 {
+		p.Count = 0
+	}
+
+	var size uint = 56
+	if p.IPV6 {
+		size = 128
+	}
```
danielnelson: The setting of size is quite mysterious, I assume this is:
>       -s packetsize
>           Specifies the number of data bytes to be sent. The default is 56,
>           which translates into 64 ICMP data bytes when combined with the 8
>           bytes of ICMP header data.

Why is it increased for IPv6?  It seems like these should be the defaults for the ping library that we don't need to deal with here.

plugins/inputs/ping_native/ping.go
```diff
@@ -0,0 +1,260 @@
...
+				}
+				return ip.String()
+			}
+		}
+	}
+
+	return ""
+}
+
+func (p *Ping) buildHostCache(acc telegraf.Accumulator) {
```
danielnelson: This function is problematic because it means we won't notice address changes.  I think we should resolve the address on every call to Gather.

plugins/inputs/ping_native/ping.go
```diff
@@ -0,0 +1,260 @@
...
+		size = 128
+	}
+
+	pinger, err := ping.NewPinger(
+		conn,
+		ping.WithOnRecieve(p.onRecv),
+		ping.WithOnFinish(p.onFinish(acc)),
+		ping.WithSize(size),
+		ping.WithCount(uint(p.Count)),
+		ping.WithInterval(time.Nanosecond*time.Duration(p.PingInterval*1000000000)),
```
danielnelson: Why `time.Nanosecond *`

plugins/inputs/ping_native/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+Sends a ping message and reports the results. This is done in pure go, eliminating the need to execute the system `ping` command.
+
+There is currently no support for TTL on windows, track progress at https://github.com/golang/go/issues/7175 and https://github.com/golang/go/issues/7174
+
+
+### Configuration:
+
+```toml
+[[inputs.ping_native]]
```
danielnelson: Merge this with the ping plugin, since we keep the same API: config and metric format.  Easiest way I think will be to pull the whole old implementation out and either call it or this one from a top plugin, ala:
```
func (p *Ping) Gather(acc telegraf.Accumulator) error {
	if p.Method == "native" {
		return p.native.Gather(acc)
	} else {
		return p.exec.Gather(acc)
	}
}

File: plugins/inputs/ping_native/ping.go [hunk 6]
```diff
@@ -0,0 +1,260 @@
...
+	return ""
+}
+
+func (p *Ping) buildHostCache(acc telegraf.Accumulator) {
+	if p.hostCache != nil {
+		return
+	}
+
+	p.network = "ip4:icmp"
+	if runtime.GOOS == "darwin" || runtime.GOOS == "linux" {
```
danielnelson: I think it could be nice to allow the selection of the type of socket you want to create or to implement an algorithm like iputils does with fallback.  I'm also 90% sure modern BSD will support non-privileged ping sockets, though not sure if this is a limitation of icmp or maybe improperly documented.

What if on non-Windows platform we try ping sockets, and then try raw sockets, and then error?

https://github.com/iputils/iputils/blob/master/ping.c#L109

plugins/inputs/ping_native/ping.go
```diff
@@ -0,0 +1,260 @@
...
+	## Whether to ping ipv6 addresses.
+  # ipv6 = false
+`
+
+func (*Ping) SampleConfig() string {
+	return sampleConfig
+}
+
+func getAddr(iface string) string {
+	if addr := net.ParseIP(iface); addr != nil {
```
danielnelson: Can this ever work, isn't this an interface name?

plugins/inputs/ping_native/ping.go
```diff
@@ -0,0 +1,260 @@
...
+
+	ifaces, err := net.Interfaces()
+	if err != nil {
+		return ""
+	}
+
+	var ip net.IP
+	for i := range ifaces {
+		if ifaces[i].Name == iface {
+			addrs, err := ifaces[i].Addrs()
```
danielnelson: I'm not exactly sure about this, but I wonder how this should work with multiple addresses, do we need to pick an address that will be able to route to the target IP?  Might be worth digging into the Go code to see how this is normally done.

plugins/inputs/ping_native/ping.go
```diff
@@ -0,0 +1,260 @@
...
+		}
+		if addr == nil {
+			continue
+		}
+		p.hostCache = append(p.hostCache, addr)
+	}
+}
+
+func (p *Ping) onRecv(pkt *ping.Packet) {
+	p.rcvdCache[pkt.IPAddr] = pingResults{ttl: pkt.TTL}
```
danielnelson: I think you would need to lock this, maybe it can be done without a member variable.

glinton:
Superseded by #6050

PR #6018, opened 2019-06-20
increase the judgment of judging the result of the query

fangjianfeng:
### Required for all PRs:


plugins/inputs/dns_query/dns_query.go
```diff
@@ -152,7 +152,7 @@ func (d *DnsQuery) getDnsQueryTime(domain string, server string) (float64, int,
 	if err != nil {
 		return dnsQueryTime, -1, err
 	}
-	if r.Rcode != dns.RcodeSuccess {
+	if r.Rcode != dns.RcodeSuccess || len(r.Answer) == 0{
```
danielnelson: This doesn't seem right to me and will result in a metric that with an error result code and a rcode_value of "no error".  What is the issue you are attempting to address?

PR #6027, opened 2019-06-21
Adding TLS mutual auth supoort to jti_openconfig_telemetry plugin

littlespace:
### Required for all PRs:

### DIFF
--- a/plugins/inputs/jti_openconfig_telemetry/openconfig_telemetry.go
+++ b/plugins/inputs/jti_openconfig_telemetry/openconfig_telemetry.go
@@ -80,13 +80,12 @@ var (
    "/interfaces",
   ]
 
-  ## enable client-side TLS and define CA to authenticate the device
-  # enable_tls = true
+  ## Optional TLS Config
   # tls_ca = "/etc/telegraf/ca.pem"
-  # insecure_skip_verify = true
-  ## define client-side TLS certificate & key to authenticate to the device
   # tls_cert = "/etc/telegraf/cert.pem"
   # tls_key = "/etc/telegraf/key.pem"
+  ## Use TLS but skip chain & host verification
+  # insecure_skip_verify = false
   
   ## Delay between retry attempts of failed RPC calls or streams. Defaults to 1000ms.
   ## Failed streams/calls will not be retried if 0 is provided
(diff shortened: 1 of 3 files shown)


plugins/inputs/jti_openconfig_telemetry/openconfig_telemetry.go
```diff
@@ -74,10 +80,14 @@ var (
...
-  ssl_cert = "/etc/telegraf/cert.pem"
-
+  ## enable client-side TLS and define CA to authenticate the device
+  # enable_tls = true
+  # tls_ca = "/etc/telegraf/ca.pem"
+  # insecure_skip_verify = true
+  ## define client-side TLS certificate & key to authenticate to the device
+  # tls_cert = "/etc/telegraf/cert.pem"
+  # tls_key = "/etc/telegraf/key.pem"
+
```
danielnelson: Let's use the typical comment block here for consistency with other plugins:
```
  ## Optional TLS Config
```

plugins/inputs/jti_openconfig_telemetry/openconfig_telemetry.go
```diff
@@ -374,8 +383,8 @@ func (m *OpenConfigTelemetry) Start(acc telegraf.Accumulator) error {
 		}
 
 		// If a certificate is provided, open a secure channel. Else open insecure one
-		if transportCredentials != nil {
-			grpcClientConn, err = grpc.Dial(server, grpc.WithTransportCredentials(transportCredentials))
+		if tlscfg != nil {
+			grpcClientConn, err = grpc.Dial(server, grpc.WithTransportCredentials(credentials.NewTLS(tlscfg)))
```
danielnelson: All the code is right, but if you are interested in going further I can see a way we could have done this nicer originally. We should have created a `grpc.DialOption` above based on EnableTLS, and then here we don't need the if/else.  Optional.

plugins/inputs/jti_openconfig_telemetry/openconfig_telemetry.go
littlespace: done

plugins/inputs/jti_openconfig_telemetry/openconfig_telemetry.go
littlespace: done

PR #6038, opened 2019-06-24
fix filecount plugin size tests #6011

### LINKED ISSUE #6011: filecount plugin size test fails on some platforms
### Relevant telegraf.conf:

```toml

```

### Steps to reproduce:

1. clone telegraf (>1.9) repo to Mac / Docker during CI tests
2. run go test

### Expected behavior:
tests pass regardless of which OS they are run on
### Actual behavior:
filecount plugin tests fail
### Additional info:
I see this locally on Mac OS 10.14.5, as well as in containers running Golang 1.1x.x-stretch:  
the OS reports **directory** sizes in Bytes many times smaller than the assumed "4096". This means that the hardcoded "expectedSize" values are wrong, and the tests then fail.

here is the same directory on Ubuntu:
```
root@drone1:~/github.com/influxdata/telegraf/plugins/inputs/filecount# ls -ltr testdata/
total 8
-rwxr-xr-x 1 root root    0 Dec 14  1979 baz
drwxr-xr-x 3 root root 4096 Jun 19 08:15 subdir
-rwxr-xr-x 1 root root  446 Jun 19 08:15 qux
-rwxr-xr-x 1 root root    0 Jun 19 08:15 foo
-rwxr-xr-x 1 root root    0 Jun 19 08:15 bar
```
and on MacOS:
```
bash-3.2$ cd testdata/
bash-3.2$ ls -ltr
total 8
-rwxr-xr-x  1 nicgrobler  staff    0 Dec 14  1979 baz
-rwxr-xr-x  1 nicgrobler  staff    0 Jun 19 10:30 bar
-rwxr-xr-x  1 nicgrobler  staff    0 Jun 19 10:30 foo
-rwxr-xr-x  1 nicgrobler  staff  446 Jun 19 10:30 qux
drwxr-xr-x  5 nicgrobler  staff  160 Jun 19 10:30 subdir
```
Note the sizes of the subdir directory...

I have extended the existing tests to set the expectedSize value at runtime, allowing the tests to function across all of the platforms. If you confirm this behaviour (I have tried it on multiple platforms, and can reproduce...but seems like no one else experiences this, so might be that my issue is between my chair and my keyboard), I will submit a PR for the fix.

so, for example, this:  
```
func TestNoFilters(t *testing.T) {
	fc := getNoFilterFileCount()
	matches := []string{"foo", "bar", "baz", "qux",
		"subdir/", "subdir/quux", "subdir/quuz",
		"subdir/nested2", "subdir/nested2/qux"}

	fileCountEquals(t, fc, len(matches), 9084)
}
```
becomes:
```
func TestNoFilters(t *testing.T) {
	fc := getNoFilterFileCount()
	matches := []string{"foo", "bar", "baz", "qux",
		"subdir/", "subdir/quux", "subdir/quuz",
		"subdir/nested2", "subdir/nested2/qux"}

	totalSize := getTotalFileSizesFromOs(t, getTestdataDir(), matches)

	fileCountEquals(t, fc, len(matches), totalSize)
}
```

nicgrobler:
closes #6011 

### Required for all PRs:

### DIFF
--- a/plugins/inputs/filecount/filecount.go
+++ b/plugins/inputs/filecount/filecount.go
@@ -58,6 +58,7 @@ type FileCount struct {
 	MTime       internal.Duration `toml:"mtime"`
 	fileFilters []fileFilterFunc
 	globPaths   []globpath.GlobPath
+	Fs          fileSystem
 }
 
 func (_ *FileCount) Description() string {
@@ -159,7 +160,7 @@ func (fc *FileCount) count(acc telegraf.Accumulator, basedir string, glob globpa
 		if err == nil && rel == "." {
 			return nil
 		}
-		file, err := os.Stat(path)
+		file, err := fc.Fs.Stat(path)
 		if err != nil {
 			if os.IsNotExist(err) {
 				return nil
@@ -244,7 +245,7 @@ func (fc *FileCount) Gather(acc telegraf.Accumulator) error {
 	}
 
 	for _, glob := range fc.globPaths {
-		for _, dir := range onlyDirectories(glob.GetRoots()) {
+		for _, dir := range fc.onlyDirectories(glob.GetRoots()) {
 			fc.count(acc, dir, glob)
 		}
 	}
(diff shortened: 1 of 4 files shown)


danielnelson:
Thank you!

PR #6049, opened 2019-06-26
Grace period for metrics late for aggregation

pberlowski:
This addresses a use-case when it's expected that some metrics will be late for aggregation (i.e. outside of the aggregation period) and `delay` doesn't allow to account for these. For example when timestamped metrics are submitted by an application using the _http_listener_ input and they're more than a full period late.

Setting the `grace` parameter allows these metrics to still be passed to the aggregator plugin for a certain duration. 

### Required for all PRs:


danielnelson:
Thanks!

PR #6052, opened 2019-06-27
Fireboard Input Plugin

### LINKED ISSUE #6051: Input plugin for fireboard connected thermometers.
## Feature Request

Import real time temperatures from Fireboard connected thermometers

### Proposal:

Connect to the Fireboard REST API servers and import temps as metrics.

### Current behavior:
N/A
### Desired behavior:
Import temps from fireboard servers
### Use case: 
Another source of information to be ingested by telegraf.

ronnocol:
### Required for all PRs:

New plugin to import data from fireboard servers to gather real time temps from their WiFi connected thermometers.

Resolves #6051

### DIFF
--- a/plugins/inputs/fireboard/README.md
+++ b/plugins/inputs/fireboard/README.md
@@ -1,0 +1,56 @@
+# Fireboard Input Plugin
+
+The fireboard plugin gathers the real time temperature data from fireboard
+thermometers.  In order to use this input plugin, you'll need to sign up
+to use their REST API, you can find more information on their website
+here [https://docs.fireboard.io/reference/restapi.html]
+
+### Configuration
+
+This section contains the default TOML to configure the plugin.  You can
+generate it using `telegraf --usage <plugin-name>`.
+
+```toml
+[[inputs.fireboard]]
+  ## Specify auth token for your account
+  ## https://docs.fireboard.io/reference/restapi.html#Authentication
+  # authToken = "b4bb6e6a7b6231acb9f71b304edb2274693d8849"
+  ## You can override the fireboard server URL if necessary
+  # URL = https://fireboard.io/api/v1/devices.json
+```
+
+#### authToken
+
+In lieu of requiring a username and password, this plugin requires the
+authToken that you can generate using the Fireboard REST API as described
+in their docs [https://docs.fireboard.io/reference/restapi.html#Authentication]
+
+#### URL
+
+While there should be no reason to override the URL, the option is available
+in case Fireboard changes their site, etc.
+
+### Metrics
+
+The Fireboard REST API docs have good examples of the data that is available,
+currently this input only returns the real time temperatures. Temperature 
+values are included if they are less than a minute old.
+
+- fireboard
+  - tags:
+    - channel
+    - scale (1=celcius; 2=farenheit)
+    - title (name of the Firebaord)
+    - uuid (UUID of the Firebaord)
+  - fields:
+    - temperature (float, unit)
+
+### Example Output
+
+This section shows example output in Line Protocol format.  You can often use
+`telegraf --input-filter <plugin-name> --test` or use the `file` output to get
+this information.
+
+```
+fireboard,channel=2,host=patas-mbp,scale=2,title=telegraf-FireBoard,uuid=b55e766c-b308-49b5-93a4-df89fe31efd0 temperature=78.2 1561690040000000000
+```
--- a/plugins/inputs/fireboard/README.md
+++ b/plugins/inputs/fireboard/README.md
@@ -13,22 +13,28 @@ generate it using `telegraf --usage <plugin-name>`.
 ```toml
 [[inputs.fireboard]]
   ## Specify auth token for your account
-  ## https://docs.fireboard.io/reference/restapi.html#Authentication
-  # authToken = "b4bb6e6a7b6231acb9f71b304edb2274693d8849"
+  auth_token = "invalidAuthToken"
   ## You can override the fireboard server URL if necessary
-  # URL = https://fireboard.io/api/v1/devices.json
+  # url = https://fireboard.io/api/v1/devices.json
+  ## You can set a different http_timeout if you need to
+  # http_timeout = 4
 ```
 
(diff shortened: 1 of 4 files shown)


plugins/inputs/fireboard/fireboard.go
```diff
@@ -0,0 +1,137 @@
...
+		return fmt.Errorf("fireboard responded with unexepcted status code %d", resp.StatusCode)
+	}
+	// Decode the response JSON into a new stats struct
+	var stats []fireboardStats
+	if err := json.NewDecoder(resp.Body).Decode(&stats); err != nil {
+		return fmt.Errorf("unable to decode fireboard response: %s", err)
+	}
+	// Range over all devices, gathering stats. Returns early in case of any error.
+	for _, s := range stats {
+		acc.AddError(r.gatherTemps(s, acc))
```
glinton: Since there is never an error returned from `gatherTemps`, remove the call to `AddError`. I would also suggest updating `gatherTemps` to not return anything.

plugins/inputs/fireboard/fireboard.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+	return nil
+}
+
+// Gathers stats from a single device, adding them to the accumulator
+func (r *Fireboard) gatherTemps(s fireboardStats, acc telegraf.Accumulator) error {
+
```
glinton: Nitpick: remove this blank line

ronnocol:
@glinton I forget how this workflow works. I've made what I believe to be the requested changes, including adding the all.go file that I forgot to add to the original check-in and a README.md. Let me know if you need anything else.

plugins/inputs/fireboard/README.md
```diff
@@ -0,0 +1,56 @@
...
+This section contains the default TOML to configure the plugin.  You can
+generate it using `telegraf --usage <plugin-name>`.
+
+```toml
+[[inputs.fireboard]]
+  ## Specify auth token for your account
+  ## https://docs.fireboard.io/reference/restapi.html#Authentication
+  # authToken = "b4bb6e6a7b6231acb9f71b304edb2274693d8849"
+  ## You can override the fireboard server URL if necessary
+  # URL = https://fireboard.io/api/v1/devices.json
```
danielnelson: Use the `snake_case` version of the options, it would be best to add struct tags to the enforce this naming, I'll add a note on the struct.

plugins/inputs/fireboard/README.md
```diff
@@ -0,0 +1,56 @@
...
+### Metrics
+
+The Fireboard REST API docs have good examples of the data that is available,
+currently this input only returns the real time temperatures. Temperature 
+values are included if they are less than a minute old.
+
+- fireboard
+  - tags:
+    - channel
+    - scale (1=celcius; 2=farenheit)
```
danielnelson: Let's use the scale name instead of the integer: `scale=celcius`

plugins/inputs/fireboard/README.md
```diff
@@ -0,0 +1,56 @@
...
+    - title (name of the Firebaord)
+    - uuid (UUID of the Firebaord)
```
danielnelson: typo on line 43/44: ao -> oa

plugins/inputs/fireboard/fireboard.go
```diff
@@ -0,0 +1,126 @@
...
+	"time"
+
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/plugins/inputs"
+)
+
+// Fireboard gathers statistics from the fireboard.io servers
+type Fireboard struct {
+	AuthToken string
+	URL       string
```
danielnelson: Adding struct tags would be good, as it enforces the correct case and sometimes avoids parsing abiguity (there is none here but they are nice for consistency):
```go
AuthToken string `toml:"auth_token"`
URL       string `toml:"url"`
```

plugins/inputs/fireboard/fireboard.go
```diff
@@ -0,0 +1,126 @@
...
+
+	client *http.Client
+}
+
+// NewFireboard return a new instance of Fireboard with a default http client
+func NewFireboard() *Fireboard {
+	tr := &http.Transport{ResponseHeaderTimeout: time.Duration(3 * time.Second)}
+	client := &http.Client{
+		Transport: tr,
+		Timeout:   time.Duration(4 * time.Second),
```
danielnelson: It could be nice to expose the timeout as a configurable option.  However, this will make it difficult to initialize the http.Client here.  We recently added an optional Init() function that could be used for this and the config validation steps in Gather.
```
func (r *Fireboard) Init() error {
```

plugins/inputs/fireboard/README.md
```diff
@@ -0,0 +1,56 @@
...
+### Configuration
+
+This section contains the default TOML to configure the plugin.  You can
+generate it using `telegraf --usage <plugin-name>`.
+
+```toml
+[[inputs.fireboard]]
+  ## Specify auth token for your account
+  ## https://docs.fireboard.io/reference/restapi.html#Authentication
+  # authToken = "b4bb6e6a7b6231acb9f71b304edb2274693d8849"
```
danielnelson: Uncomment this option since we don't have a good default and it is required.  Don't add the URL in the config file, but lets make sure we have a link in the README.

plugins/inputs/fireboard/README.md
ronnocol: I do this now in fireboard.go GatherTemps() line 123/130. Would appreciate feedback on if this is the best way to do it.

danielnelson: The worry with this method is that if somehow the data is not 0-2, then it will cause a panic.  I'd probably make a new function and use switch:
```go
func scale(n int) string {
	switch n {
	case 1:
		return "celcius"
	case 2:
		return "farenheit"
	default:
		return ""
	}
}
```

ronnocol: Thanks. Implemented as a function.

PR #6058, opened 2019-07-01
Add filtering to stackdriver labels

memory:
# Add filtering to stackdriver labels

Stackdriver has a hard quota of no more than 10 labels per external metric.

Previously, if an input metric contained over 10 tags, we would remove them sequentially to stay under the limit, regardless of which tags were present.

This PR adds the ability to specify a list of tag names (as a comma-separated string) which the stackdriver plugin will preferentially and consistently add to the list of stackdriver labels for the metric.

### Required for all PRs:

### DIFF
--- a/plugins/outputs/stackdriver/README.md
+++ b/plugins/outputs/stackdriver/README.md
@@ -21,6 +21,13 @@ Additional resource labels can be configured by `resource_labels`. By default th
   ## The namespace for the metric descriptor
   namespace = "telegraf"
 
+  ## Preferred labels:
+  ## Stackdriver custom metrics are limited to 10 labels total;
+  ## if there are >10 tags associated with the input data, we
+  ## will attempt to preserve these tags while staying under
+	## the stackdriver quota.
+  # preferred_labels = "foo,bar,baz"
+
   ## Custom resource type
   # resource_type = "generic_node"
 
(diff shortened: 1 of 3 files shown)


danielnelson:
This is an interesting PR, actually we could use this type of functionality for several other plugins such as OpenTSDB and Azure Monitor.

I think it would make sense to move it out to a processor, something like this:
```toml
[[processors.tag_limit]]
  limit = 10
  keep = ["host", "source", "foo"]
```

memory:
Okay, it may take a few days but I'll take a crack at it later on this week or early next.

danielnelson:
@memory Awesome, thank you!

memory:
Actually, looking at the code a bit: you already have the taginclude/tagexclude functionality in models/filter.go.  Would it make more sense to stick this functionality in there rather than as a separate processor?  Something like:

```
[[inputs.cpu]]
  percpu = true
  totalcpu = true
  tagtrimlimit = 5
  tagtrimprefer = ["cpu"]
```

danielnelson:
I'd like to keep those filters more basic for now, and I think tag limiting is a bit too specialized.

memory:
Okay, closing in favor of https://github.com/influxdata/telegraf/pull/6086

PR #6068, opened 2019-07-04
testutil: print diff if fields are not equal

dupondje:
During tests it was sometimes annoying I couldn't see what was the exact difference.
Adding a compare in the testutil that prints a diff.

### Required for all PRs:

### DIFF
--- a/testutil/accumulator.go
+++ b/testutil/accumulator.go
@@ -9,6 +9,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/google/go-cmp/cmp"
 	"github.com/influxdata/telegraf"
 	"github.com/stretchr/testify/assert"
 )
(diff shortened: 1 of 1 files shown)


danielnelson:
I'm planning to remove most of the test code at some point; since its pretty messy, could you try using this method from in testutil/metric.go:
```
	expected := []telegraf.Metric{
		testutil.MustMetric("smart_device",
			map[string]string{
				"device":    ".",
				"model":     "TS128GMTE850",
				"serial_no": "D704940282?",
			},
			map[string]interface{}{
				"exit_status": 0,
				"health_ok":   true,
				"temp_c":      38,
			},
			time.Now(),
		),
	}

	// Generally you would want SortMetrics() to avoid ordering errors and
	// IgnoreTime unless you are testing time specific code.
	testutil.RequireMetricsEqual(t, expected, acc.GetTelegrafMetrics(),
		testutil.SortMetrics(), testutil.IgnoreTime())
```

Issue #6041, opened 2019-06-25
Add more smart attributes written from NVME drives


coccobill1:
## Feature Request

More smart attributes should be logged when using the inputs.smart plugin with NVME devices.
Currently the plugin pulls smart_device data from NVME drives, along with the following smart_attributes: Temperature and Power Cycle Count. Temperature is already included in the smart_device data, so the benefit of adding attributes: true is limited. It would be helpful if additional smart attribute fields were logged.

### Proposal:

Add at the minimum the following smart attributes (from Samsung 970 EVO):
- Power On Hours
- Media and Data Integrity Errors
- Error Information Log Entries
- Critical Warning
- Available Spare

Other key attributes possibly logged by other NVME devices:
- Reallocated Sector Count
- Wear Leveling Count
- Uncorrectable Errors Count

### Use case: 

The data can be used to detect or predict a drive failure.

PR #6079, opened 2019-07-06
Smart Input Plugin Extra Attributes

GeorgeMac:
This PR attempts to deliver most of https://github.com/influxdata/telegraf/issues/6041

Currently this PR supports:
- [x] Media and Data Integrity Errors (count)
- [x] Error Information Log Entries (count)
- [x] Available Spare (percentage)
- [x] Critical Warning (integer bit set)

Outstanding Questions:

1. Documentation for smart attributes seem to be a little elusive to me. Can anyone point me in the direction of the attributes id's for all the above attributes. Perhaps you know @coccobill1 ?

2. `Critical Warning` appears to be a bit flag printed as a hex value. I have struggled to find decent documentation around the meaning of all these attributes. Maybe someone can point me in the best direction of better docs. I have this https://www.intel.com/content/dam/support/us/en/documents/solid-state-drives/Intel_SSD_Smart_Attrib_for_PCIe.pdf for example.

Given we know the definitive meaning (and that there is one) for what each flag means, do we have any suggestions on how to communicate that best to telegraf? cc @danielnelson 

What I mean is, for example, would we want to communicate each flags as individual distinct fields. Each field with a boolean value (on or off)?

e.g. The reading `0x01` given the first bit in the byte represents `available spare is below threshold` (as I have currently taken from the shared intel doc above) would lead to a field like `Critical_Warning_Available_Spare_Below_Threshold` with a `raw_value=1`?

3. I couldn't find any decent examples of the extra fields to build a sensible test from or have a clear idea on the format of the value in the output. So I haven't taken them on yet. These are:

- Reallocated Sector Count
- Wear Leveling Count
- Uncorrectable Errors Count

I could implement them all as just counts and fabricate a test and hope they exist as described here if people are happy with this.

I could be completely missing the point here 😂 so any advice welcomed.

### Required for all PRs:

- [x] Associated README.md updated. (there currently doesn't seem to exist any documentation at this fidelity)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,5 +3,3 @@
 /telegraf.exe
 /telegraf.gz
 /vendor
-*.swp
-*.swo
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@
 /telegraf.exe
 /telegraf.gz
 /vendor
+*.swp
+*.swo
--- a/plugins/inputs/smart/smart.go
+++ b/plugins/inputs/smart/smart.go
@@ -48,6 +48,8 @@ var (
 	nvmePowerCycleAttr = regexp.MustCompile("^Power Cycles:\\s+(.*)$")
 	// Power On Hours: 6,038
 	nvmePowerOnAttr = regexp.MustCompile("^Power On Hours:\\s+(.*)$")
+	// Media and Data Integrity Errors: 0
+	nvmeIntegrityErrAttr = regexp.MustCompile("^Media and Data Integrity Errors:\\s+(.*)$")
 
 	// ID# ATTRIBUTE_NAME          FLAGS    VALUE WORST THRESH FAIL RAW_VALUE
 	//   1 Raw_Read_Error_Rate     -O-RC-   200   200   000    -    0
@@ -330,6 +332,18 @@ func gatherDisk(acc telegraf.Accumulator, usesudo, collectAttributes bool, smart
 					tags["id"] = "9"
 					tags["name"] = "Power_On_Hours"
 					i, err := strconv.ParseInt(strings.Replace(powerOn[1], ",", "", -1), 10, 64)
+					if err != nil {
+						continue
+					}
+					fields["raw_value"] = i
+
+					acc.AddFields("smart_attribute", fields, tags)
+					continue
+				}
+
+				if errors := nvmeIntegrityErrAttr.FindStringSubmatch(line); len(errors) > 1 {
+					tags["name"] = "Media_and_Data_Integrity_Errors"
+					i, err := strconv.ParseInt(strings.Replace(errors[1], ",", "", -1), 10, 64)
 					if err != nil {
 						continue
 					}
--- a/plugins/inputs/smart/smart.go
+++ b/plugins/inputs/smart/smart.go
@@ -18,6 +18,20 @@ import (
 	"github.com/influxdata/telegraf/plugins/inputs"
 )
 
+const (
+	// NVME Critical Warnings as per https://nvmexpress.org
+	// critical warning spare threshold
+	criticalWarningST uint8 = 1 << iota
+	// critical warning temperature above or under threshold
+	criticalWarningTAUT
+	// critical warning reliability degraded
+	criticalWarningRD
+	// critical warning read only
+	criticalWarningRO
+	// critical warning volatile backup memory failed
+	criticalWarningVMBF
+)
+
 var (
 	// Device Model:     APPLE SSD SM256E
 	// Product:              HUH721212AL5204
@@ -51,6 +65,14 @@ var (
 		"190": "temp_c",
 		"194": "temp_c",
 		"199": "udma_crc_errors",
+	}
+
+	nvmeCriticalWarnings = map[uint8]string{
+		criticalWarningST:   "Critical_Warning_Spare_Treshold",
+		criticalWarningTAUT: "Critical_Warning_Temperature_Above_or_Under_Threshold",
+		criticalWarningRD:   "Critical_Warning_Reliability_Degraded",
+		criticalWarningRO:   "Critical_Warning_Read_Only",
+		criticalWarningVMBF: "Critical_Warning_Volative_Memory_Backup_Failed",
 	}
 
 	sasNvmeAttributes = map[string]struct {
@@ -336,6 +358,23 @@ func gatherDisk(acc telegraf.Accumulator, usesudo, collectAttributes bool, smart
 		} else {
 			if collectAttributes {
 				if matches := sasNvmeAttr.FindStringSubmatch(line); len(matches) > 2 {
+					if matches[1] == "Critical Warning" {
+						var flags uint8
+						if _, err := fmt.Sscanf(matches[2], "0x%x", &flags); err != nil {
+							continue
+						}
+
+						for flag, name := range nvmeCriticalWarnings {
+							tags["name"] = name
+
+							acc.AddFields("smart_attribute", map[string]interface{}{
+								"raw_value": flags&flag > 0,
+							}, tags)
+						}
+
+						continue
+					}
+
 					if attr, ok := sasNvmeAttributes[matches[1]]; ok {
 						tags["name"] = attr.Name
 						if attr.ID != "" {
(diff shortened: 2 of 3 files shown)


GeorgeMac:
Small update: https://nvmexpress.org/wp-content/uploads/NVM-Express-Management-Interface-1.1-Ratified.pdf

This document seems to correlate well r.e. the meanings of the critical warning flags for nvme.

coccobill1:
First off I'd like to say that a lot of paid software don't have support like this even when showing the money. Greatly appreciated guys!

I'm certainly no expert on s.m.a.r.t. data, but from my experience they are quite a wild west, a lot of manufacturer specific stuff and seems impossible to find up to date information on them. On Samsung's page the only document I can find is this, and it's clearly outdated for the NVME drives:

http://downloadcenter.samsung.com/content/UM/201711/20171115115128872/SSD_Application_Note_SMART_final.pdf

That NVME specification is probably as good as it gets for a standardized model, but assuming all vendors stick to it might be optimistic. Smartctl doesn't display the ID's for the attributes on my NVME drive.

Regarding 3. I have no idea whether those attributes are used anymore with NVME drives, but they used to be key attributes to monitor with sata/ide drives, so if they are still valid they would be a great addition.

GeorgeMac:
Hey @coccobill1 👋 Thank you for the kind words. That was my first week at Influx and first thing contributed as an employee. So it meant a lot 🙇 

That document you linked it super helpful as well. Cross-referencing all these different docs helps to find some sensible answer.
I had the same issue with my macbook nvme drive. No attribute IDs. Actually looking at the smart input plugin implementation the IDs for nvme are hard-coded rather than gathered from the output of smartctl. So just need to make sure I am hard-coded it correctly.
I will get another opinion from some influxers on what to do with those extra attributes as well and get back to you.

GeorgeMac:
Update: refactored the collection in order to make it more extensible. New matches for SAS and NVME attributes can be quickly added via the `sasNvmeAttributes` map. This also reduces N regexp's per line down to one for smart attribute extraction.

plugins/inputs/smart/smart.go
```diff
@@ -62,6 +52,55 @@ var (
 		"194": "temp_c",
 		"199": "udma_crc_errors",
 	}
+
+	sasNvmeAttributes = map[string]struct {
+		ID    string
+		Name  string
+		Parse func(fields, deviceFields map[string]interface{}, str string) error
+	}{
```
docmerlin: Why is this a map instead of a struct?  It seems it doesn't need any dynamic items?

GeorgeMac: @docmerlin so it goes:
1. Generically match anything in the form `(Some Attribute): (1234)`
2. Lookup "Some Attribute" in `sasNvmeAttributes`
3. If match use Name field of attribute (as this differs from match in smartctl output) as name in metric point.
4. Use ID in point if set
5. Parse value as integer (which may contain a ","). If override parse function is supplied, use that instead (useful for temperatures which also populate deviceFields and percentages).

Perhaps I am missing something. But I wanted to avoid having to search through every possible attribute name that could be in the line. In favor of generically pulling out anything `key: value` and looking up the key in a set of known attribute keys.

plugins/inputs/smart/smart.go
```diff
@@ -424,6 +405,29 @@ func parseInt(str string) int64 {
 	return 0
 }
 
+func parseCommaSeperatedInt(fields, deviceFields map[string]interface{}, str string) error {
```
docmerlin: replace `deviceFields` name with an `_`, since it isn't being used in the function?

.gitignore
```diff
@@ -3,3 +3,5 @@
 /telegraf.exe
 /telegraf.gz
 /vendor
+*.swp
```
glinton: Revert this file, editor and IDE files belong in your global `.gitignore`. If you don't have one, you can add it with:

```
git config --global core.excludesfile $HOME/.gitignore
```

Then add these entries to `$HOME/.gitignore`

plugins/inputs/smart/smart.go
```diff
@@ -62,6 +52,55 @@ var (
...
+			Name: "Media_and_Data_Integrity_Errors",
+		},
+		"Error Information Log Entries": {
+			Name: "Error_Information_Log_Entries",
+		},
+		"Available Spare": {
+			Name: "Available_Spare",
+			Parse: func(fields, deviceFields map[string]interface{}, str string) error {
+				if str[len(str)-1] == '%' {
+					str = str[:len(str)-1]
```
danielnelson: Bad things happen if str has length 0, I'd use strings.TrimSuffix here to make it a bit more readable.

plugins/inputs/smart/smart.go
```diff
@@ -36,18 +36,8 @@ var (
...
-	// Current Drive Temperature:     34 C
-	sasTempAttr = regexp.MustCompile("^Current Drive Temperature:\\s+(.*)\\s+C(.*)$")
-	// Temperature: 38 Celsius
-	nvmeTempAttr = regexp.MustCompile("^Temperature:\\s+(.*)\\s+(.*)$")
-	// Power Cycles: 472
-	nvmePowerCycleAttr = regexp.MustCompile("^Power Cycles:\\s+(.*)$")
-	// Power On Hours: 6,038
-	nvmePowerOnAttr = regexp.MustCompile("^Power On Hours:\\s+(.*)$")
+	// sasNvmeAttr is a SAS or NVME SMART attribute
+	sasNvmeAttr = regexp.MustCompile(`^([^:]*):\s+(.*)$`)
```
danielnelson: May be useful to require submatch strings have a length > 0: `^([^:]+):\s+(.+)$`

GeorgeMac:
Update: I added the following critical warning fields with boolean values:
- Critical_Warning_Spare_Treshold
- Critical_Warning_Temperature_Above_or_Under_Threshold
- Critical_Warning_Reliability_Degraded
- Critical_Warning_Read_Only
- Critical_Warning_Volative_Memory_Backup_Failed

plugins/inputs/smart/smart.go
```diff
@@ -300,82 +357,41 @@ func gatherDisk(acc telegraf.Accumulator, usesudo, collectAttributes bool, smart
...
-					if err != nil {
-						continue
-					}
-					fields["raw_value"] = i
-
-					acc.AddFields("smart_attribute", fields, tags)
-					continue
-				}
+							acc.AddFields("smart_attribute", map[string]interface{}{
+								"raw_value": flags&flag > 0,
```
danielnelson: We won't be add this as a bool since we already have it as an integer, and InfluxDB won't allow us to have mixed types.  Since we are attempting to model the NVMe log as `smart_attributes`, I would just do it like this and document or link to the values in the README (see net_response for an example):
```
smart_attribute,name=critical_warning raw_value=0i
```

Often with enum style data we do something like this with a tag + integer field, but I don't think we should do it here since it doesn't fit into the existing model very well:
```
smart_attribute,name=critical\ warning,warning=reliability\ degraded warning_code=0i
```

GeorgeMac: @danielnelson ahh I see what you mean! Do you think having it as an integer is a little hard to interpret? Since it represents so many states / combinations of flags? i.e. 8 means just `read only`, but 9 means `read only` and `available spare threshold exceeded`. I suppose no one is screaming for this warning fidelity at the moment, so I am probably just splitting hairs 😂 . They can atleast do something like alert when any critical warning occurs with a simple >0 check.

danielnelson: That's a good point, this isn't an enum it's a bit set, so it wouldn't really work to pair it with a human readable name=foo tag.  I also agree, it's pretty difficult to deal with a large list of integers.  Still because this is a smart_attribute, I think we should start with just the single field, but read on for more discussion of booleans and bit sets.

One reason we have often favored integer values over booleans is that Grafana/Chronograf aren't very good at displaying booleans outside of tables.  UI support is always changing though, so it may be worth looking into what support there is now.  Many of the outputs also do not support booleans, and require users to use the `enum` processor.

I think there are 2 things that a user would be interested in with a bitset: is the item in an error state (non-zero), and what flags are set.  Checking for an error state is not always something that makes sense depending on the semantics of the bit set, but I think it often does.  If we were going to encode a bit set for InfluxDB, we could do it as multiple fields, one for the bit set encoded an integer, and one for each flag in the bit set:
```
foo critical_warning=3i,critical_warning_spare_threshold=true,critical_warning_temperature_above_or_under_threshold=true
```

However, this might be tricky to query, since you would need to grab all bit set fields and check each one to see if it is set.  Probably only something you want to do if you are checking for a particular flag.  It occurs to me that the set type idea for tags we discussed could help here, imagine:
```
foo,critical_warning=spare_threshold,critical_warning=temperature_above_or_under_threshold critical_warning=3i
```

GeorgeMac: 👍 I like it. I will simplify critical warning to just an int for now then.

PR #5691, opened 2019-04-06
Add plugin for 389 Directory server

gnulux:
### Required for all PRs:

- [x] Signed [CLA](https://influxdata.com/community/cla/).
- [x] Associated README.md updated.
- [x] Has appropriate unit tests.

PR #6083, opened 2019-07-08
New plugins for LDAP

falon:
### Required for all PRs:

This work is inspired by Openldap plugin. Unit tests from that plugin can apply too.


danielnelson:
@falon I think there is probably some duplication with #5691, can you check it out?

falon:
> @falon I think there is probably some duplication with #5691, can you check it out?
@danielnelson I didn't notice someone was already working on that! :)

The ldap_org plugin is not duplicated, instead.

glinton:
Could these enhancements be part of the openldap plugin itself now? Is it another call to the server to gather these metrics?

sjwang90:
Are there differences between #3800

falon:
> Could these enhancements be part of the openldap plugin itself now? Is it another call to the server to gather these metrics?

This plugin collects different metrics. In my opinion could act as a separate plugin.

sjwang90:
>This plugin collects different metrics. In my opinion could act as a separate plugin.

Could metrics be added to the existing `inputs.openldap`? We are hesitant to add a completely separate plugin that doesn't have a broad use case (with over 200+ plugins we are already increasing Telegraf's binary size with every release).

falon:
> > This plugin collects different metrics. In my opinion could act as a separate plugin.
> 
> Could metrics be added to the existing `inputs.openldap`? We are hesitant to add a completely separate plugin that doesn't have a broad use case (with over 200+ plugins we are already increasing Telegraf's binary size with every release).

I understand. This is not easy. 389ds and openldap plugins interface with cn=monitor tree. This is a completely different work.

sjwang90:
Hey @falon, right now the plugin in its current state is not enough to cover a broad number of use cases to implement into Telegraf as a standalone plugin. If there are additional metrics that could make the plugin more useful or anyway to combine it with #3800 or `inputs.openldap` we will consider it. 

You also have the option on having this plugin as an external plugin. You can find the docs on how how to convert it [here](https://github.com/influxdata/telegraf/tree/master/plugins/common/shim#telegraf-execd-go-shim), and once that's done please open a pull request to get it added to the [list of external plugins](https://github.com/influxdata/telegraf/blob/master/EXTERNAL_PLUGINS.md).

falon:
I opened #9217.

I don't think that adding this to _ldap_response_ is good.
The _ldap_org_ plugin could need an high `interval` to run in order to avoid high load on LDAP server. Moreover it's useless to run frequently it.

I think you can close this PR. Thank you.

Issue #5758, opened 2019-04-23
Get Docker Thin Pool Minimum Free Space Metric Gathered


akhon:
## Feature Request

Opening a feature request kicks off a discussion.

### Proposal:
Get stats of `Thin Pool Minimum Free Space` which is available for devicemapper storage driver:
```
Storage Driver: devicemapper
 Pool Name: docker-docker--pool
 Pool Blocksize: 524.3kB
 Base Device Size: 107.4GB
 Backing Filesystem: ext4
 Udev Sync Supported: true
 Data Space Used: 4.149GB
 Data Space Total: 53.04GB
 Data Space Available: 48.89GB
 Metadata Space Used: 6.468MB
 Metadata Space Total: 54.53MB
 Metadata Space Available: 48.06MB
 Thin Pool Minimum Free Space: 5.304GB
 Deferred Removal Enabled: true
 Deferred Deletion Enabled: true
 Deferred Deleted Device Count: 0
 Library Version: 1.02.135-RHEL7 (2016-11-16)
```

### Current behavior:
Data is not collected at all

### Desired behavior:
Get data collected

### Use case:
Need that bad boy to be monitored, looks like it missed!

PR #6084, opened 2019-07-08
Docker Input Plugin Measure Thin Pool Minimum Free Space

GeorgeMac:
Closes https://github.com/influxdata/telegraf/issues/5758

This adds a new measurement `docker_thin_pool` with the field `minimum_free_space` measured in bytes.

### Required for all PRs:

### DIFF
--- a/plugins/inputs/docker/README.md
+++ b/plugins/inputs/docker/README.md
@@ -159,7 +159,7 @@ The above measurements for the devicemapper storage driver can now be found in t
     - metadata_space_available_bytes
     - thin_pool_minimum_free_space_bytes
 
-- docker_container_mem
++ docker_container_mem
   - tags:
     - engine_host
     - server_version
@@ -167,7 +167,7 @@ The above measurements for the devicemapper storage driver can now be found in t
     - container_name
     - container_status
     - container_version
-  - fields:
+  + fields:
     - total_pgmafault
     - cache
     - mapped_file
@@ -212,7 +212,7 @@ The above measurements for the devicemapper storage driver can now be found in t
     - container_status
     - container_version
     - cpu
-  - fields:
+  + fields:
     - throttling_periods
     - throttling_throttled_periods
     - throttling_throttled_time
@@ -223,7 +223,7 @@ The above measurements for the devicemapper storage driver can now be found in t
     - usage_percent
     - container_id
 
-- docker_container_net
++ docker_container_net
   - tags:
     - engine_host
     - server_version
@@ -232,7 +232,7 @@ The above measurements for the devicemapper storage driver can now be found in t
     - container_status
     - container_version
     - network
-  - fields:
+  + fields:
     - rx_dropped
     - rx_bytes
     - rx_errors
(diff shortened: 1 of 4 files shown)


danielnelson:
I think we may want to reorganize the storage driver metrics.  Right now the devicemapper metrics are spread out over several measurements:
```
- docker
  - fields:
  	- pool_blocksize

- docker_data
  - tags:
    - unit
    - engine_host
    - server_version
  - fields:
    - available
    - total
    - used

- docker_metadata
  - tags:
    - unit
    - engine_host
    - server_version
  - fields:
    - available
    - total
    - used
```

We can't get rid of these, since we don't want to break any queries, dashboards, alerts, etc.  However, we can deprecate them and try again.  I think this layout would be preferable:
```
- docker_devicemapper
  - tags:
    - engine_host
    - server_version
    - pool_name
  - fields:
    - data_space_available_bytes
    - data_space_total_bytes
    - data_space_used_bytes
    - metadata_space_available_bytes
    - metadata_space_total_bytes
    - metadata_space_used_bytes
    - thin_pool_minimum_free_space_bytes
    - (other stuff)
```

I like this layout more because it gives us room to grow, without a measurement explosion, and it groups all the devicemapper related values into a single measurement.

GeorgeMac:
I like it @danielnelson

plugins/inputs/docker/docker.go
```diff
@@ -322,21 +322,49 @@ func (d *Docker) gatherInfo(acc telegraf.Accumulator) error {
...
+		switch name {
+		case "pool_blocksize",
+			"data_space_used",
+			"data_space_total",
+			"data_space_available",
+			"metadata_space_used",
+			"metadata_space_total",
+			"metadata_space_available",
+			"thin_pool_minimum_free_space":
+			deviceMapperFields[fmt.Sprintf("%s_bytes", name)] = value
```
danielnelson: Very nitpicky, I would just go with `name + "_bytes"` here just to skip the cost of Sprintf.

plugins/inputs/docker/README.md
```diff
@@ -142,6 +142,23 @@ some storage drivers such as devicemapper.
     - total
     - used
 
+The above measurements for the devicemapper storage driver can now be found in the new `docker_devicemapper` measurement
+
+- docker_devicemapper
```
danielnelson: You don't need to fix this, but this is good opportunity to show an obscure README issue.  In Markdown, or maybe it's only Github Markdown, if you have multiple nested lists you will need to alter the first list character when repeating them, otherwise you get a weird bit of inconsistent vertical whitespace.  To see this, in the current plugin README compare the vertical whitespace difference between the docker measurement and the docker_data.  We usually use `-` and `+`.

Bad:
```
- foo
  - bar
- foo2
  - bar2
- foo3
  - bar3
- foo4
  - bar4
```
Good:
```
- foo
  - bar
+ foo2
  - bar2
- foo3
  - bar3
+ foo4
  - bar4
```

GeorgeMac: Oh yeah that is weird. Nice I did not know this!

GeorgeMac: What have you done. Now I can't unsee the whitespace 😂 :trollface:

PR #6086, opened 2019-07-08
Tag limit processor

memory:
Add a processor that will clamp the number of exported tags to a specified limit, and take a list of tag names to preferentially preserve when trimming to the limit.

### Required for all PRs:

### DIFF
--- a/plugins/processors/tag_limit/README.md
+++ b/plugins/processors/tag_limit/README.md
@@ -5,7 +5,8 @@ preserved for any given metric, and to choose the tags to preserve when the
 number of tags appended by the data source is over the limit.
 
 This can be useful when dealing with output systems (e.g. Stackdriver) that
-impose hard limits on the number of tags/labels per metric.
+impose hard limits on the number of tags/labels per metric or where high
+levels of cardinality are computationally and/or financially expensive.
 
 ### Configuration
 
@@ -21,6 +22,6 @@ impose hard limits on the number of tags/labels per metric.
 ### Example
 
 ```diff
-- throughput month=Jun,environment=qa,region=us-east1,lower=10i,upper=1000i,mean=500i 1560540094000000000
-- throughput environment=qa,region=us-east1,lower=10i 1560540094000000000
++ throughput month=Jun,environment=qa,region=us-east1,lower=10i,upper=1000i,mean=500i 1560540094000000000
++ throughput environment=qa,region=us-east1,lower=10i 1560540094000000000
 ```
--- a/plugins/processors/tag_limit/README.md
+++ b/plugins/processors/tag_limit/README.md
@@ -1,0 +1,26 @@
+# Tag Limit Processor Plugin
+
+Use the `tag_limit` processor to ensure that only a certain number of tags are
+preserved for any given metric, and to choose the tags to preserve when the
+number of tags appended by the data source is over the limit.
+
+This can be useful when dealing with output systems (e.g. Stackdriver) that
+impose hard limits on the number of tags/labels per metric.
+
+### Configuration
+
+```toml
+[[processors.tag_limit]]
+  ## Maximum number of tags to preserve
+  limit = 3
+
+  ## List of tags to preferentially preserve
+  keep = ["environment", "region"]
+```
+
+### Example
+
+```diff
+- throughput month=Jun,environment=qa,region=us-east1,lower=10i,upper=1000i,mean=500i 1560540094000000000
+- throughput environment=qa,region=us-east1,lower=10i 1560540094000000000
+```
--- a/plugins/processors/tag_limit/tag_limit.go
+++ b/plugins/processors/tag_limit/tag_limit.go
@@ -1,9 +1,14 @@
 package taglimit
 
 import (
+	"fmt"
 	"github.com/influxdata/telegraf"
 	"github.com/influxdata/telegraf/plugins/processors"
 	"log"
+)
+
+var (
+	keepTags map[string]string
 )
 
 const sampleConfig = `
@@ -17,6 +22,7 @@ const sampleConfig = `
 type TagLimit struct {
 	Limit int      `toml:"limit"`
 	Keep  []string `toml:"keep"`
+	init  bool
 }
 
 func (d *TagLimit) SampleConfig() string {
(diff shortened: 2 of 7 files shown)


plugins/processors/tag_limit/tag_limit.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+func (d *TagLimit) Apply(in ...telegraf.Metric) []telegraf.Metric {
+	// convert list of tags to a map so we can do constant-time lookups
+	keepTags := make(map[string]string)
+	for _, tag := range d.Keep {
+		keepTags[tag] = ""
+	}
+	for _, point := range in {
+		pointTags := point.Tags()
```
danielnelson: This function is pretty slow since it makes a copy of all the tags into a map.  Let's iterate over the TagList() and build up an array of tags we want to remove, then call RemoveTag on all of them.  We can initialize the list to the length of limit, and optionally, reuse the array each call to Apply.

memory:
all reasonable requests; hoping to get to this next week.

plugins/processors/tag_limit/tag_limit.go
memory: I don't think we can assume that the number of tags to drop is <= the limit number?  Or that the list of tags to drop would be consistent from metric to metric?

Anyway, sorry, I hadn't quite realized that point.Tags() was so expensive but you're quite right -- converting over to using TagList().

plugins/processors/tag_limit/tag_limit.go
```diff
@@ -0,0 +1,71 @@
...
+		for k := range pointTags {
+			if _, ok := keepTags[k]; !ok {
+				delete(pointTags, k)
+			}
+			if len(pointTags) <= d.Limit {
+				break
+			}
+		}
+		// we've trimmed off all the non-keep tags but we're still
+		// over the limit, so start trimming
```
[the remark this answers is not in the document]
memory: Actually, yeah, that's probably a straight up error.  Added a check in the init function.

plugins/processors/tag_limit/tag_limit.go
```diff
@@ -0,0 +1,87 @@
...
+
+import (
+	"fmt"
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/plugins/processors"
+	"log"
+)
+
+var (
+	keepTags map[string]string
```
glinton: nitpick: put this var in the `TagLimit` struct

plugins/processors/tag_limit/tag_limit.go
```diff
@@ -0,0 +1,87 @@
...
+	err := d.initOnce()
+	if err != nil {
+		log.Printf("E! [processors.tag_limit] could not create tag_limit processor: %v", err)
+		return in
+	}
+	for _, point := range in {
+		pointOriginalTags := point.TagList()
+		lenPointTags := len(pointOriginalTags)
+		tagsToRemove := make([]string, lenPointTags)
+		if len(pointOriginalTags) <= d.Limit {
```
glinton: Use `lenPointTags` here

plugins/processors/tag_limit/tag_limit.go
```diff
@@ -0,0 +1,87 @@
...
+func (d *TagLimit) Apply(in ...telegraf.Metric) []telegraf.Metric {
+	err := d.initOnce()
+	if err != nil {
+		log.Printf("E! [processors.tag_limit] could not create tag_limit processor: %v", err)
+		return in
+	}
+	for _, point := range in {
+		pointOriginalTags := point.TagList()
+		lenPointTags := len(pointOriginalTags)
+		tagsToRemove := make([]string, lenPointTags)
```
glinton: Should this be `make([]string, lenPointTags - d.Limit)` to avoid overallocating?

plugins/processors/tag_limit/tag_limit.go
```diff
@@ -0,0 +1,87 @@
...
+		if len(pointOriginalTags) <= d.Limit {
+			continue
+		}
+		// remove extraneous tags, stop once we're at the limit
+		for i, t := range pointOriginalTags {
+			if _, ok := keepTags[t.Key]; !ok {
+				tagsToRemove[i] = t.Key
+				lenPointTags = lenPointTags - 1
```
glinton: Nitpick: Use the simplified `lenPointTags--`

plugins/processors/tag_limit/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+### Configuration
+
+```toml
+[[processors.tag_limit]]
+  ## Maximum number of tags to preserve
+  limit = 3
+
+  ## List of tags to preferentially preserve
+  keep = ["environment", "region"]
```
glinton: What do you think about calling this `include`? An argument for doing so would be that it would simplify adding in an `exclude` in the future if desired.

plugins/processors/tag_limit/tag_limit.go
memory: done

plugins/processors/tag_limit/tag_limit.go
memory: oops, yes. done.

plugins/processors/tag_limit/tag_limit.go
memory: yes; also moving it to below the first length check, since if we hit the `continue` there's no point in allocating any list at all.

plugins/processors/tag_limit/tag_limit.go
memory: done

plugins/processors/tag_limit/README.md
memory: I went back and forth on that and eventually decided that `include` generally indicates an exclusive inclusion, whereas this code will allow unlisted tags through as along as we're still under the limit.

memory: (feel free to unresolve if you disagree)

memory:
I've been staring at the circleci failure on go-1.11 for a while now and am largely convinced it's irrelevant to my change; does anyone believe differently?

@danielnelson I think we're good to go?

PR #6092, opened 2019-07-09
Fix https support in activemq input.

### LINKED ISSUE #6076: ActiveMQ input plugin fails with HTTPS targets
### Relevant telegraf.conf:

```toml
[[inputs.activemq]]
  server = "https://somebroker.mq.eu-west-1.amazonaws.com"
  port = 8162
  username = "admin"
  password = "admin"
  webadmin = "admin"
  response_timeout = "5s"
```

### System info:
Telegraf 1.11.1, MacOS Mojave, ActiveMQ 5.15.0 (AmazonMQ)

### Steps to reproduce:

1. Set target URL to an HTTPS broker admin page
2. Start Telegraf
3. Plugin errors: `E! [inputs.activemq]: Error in plugin: EOF`

### Expected behavior:
Telegraf plugin connects to activemq admin page and gets metrics
### Actual behavior:
Telegraf fails to connect and get metrics
### Additional info:
SSL certificate is valid, but testing with `insecure_skip_verify = true` (this is actually missing from the README example for the activemq plugin) and tested without uri protocol in target.

Tested with non https activemq locally and works correctly.

danielnelson:
Needed to provide a way to select between `http` and `https`.  While I was tempted to go with an `enable_tls = true` option, I instead went for a new `url` option that combines the scheme, host and port.  This is more similar to how we handle other HTTP inputs.

I will have to do some minor changes to apply this to the release branch, since it uses the new Init function.

closes #6076

### Required for all PRs:

### DIFF
--- a/plugins/inputs/activemq/README.md
+++ b/plugins/inputs/activemq/README.md
@@ -1,4 +1,4 @@
-# Telegraf Input Plugin: ActiveMQ
+# ActiveMQ Input Plugin
 
 This plugin gather queues, topics & subscribers metrics using ActiveMQ Console API.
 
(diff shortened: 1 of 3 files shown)


plugins/inputs/activemq/README.md
```diff
@@ -7,12 +7,14 @@ This plugin gather queues, topics & subscribers metrics using ActiveMQ Console A
 ```toml
 # Description
 [[inputs.activemq]]
+  ## ActiveMQ WebConsole URL
+  url = "http://127.0.0.1:8161"
+
   ## Required ActiveMQ Endpoint
+  ##   deprecated in 1.11; use the url option
```
glinton: :+1:

PR #6104, opened 2019-07-10
Enable agent to send Version string

nicgrobler:
This PR (addresses influxdata#5958) introduces a monitor for the agent itself - this is a small, specific “mini agent” that sits alongside the main agent (rather than within the lower-levels, where the plugins live). Due to this, it is able to listen for signals (from telegraph), and also see and report on any meta-data concerning the agent itself - for example, the version string, number of inputs etc (which no plugin can see because it has no visibility back up into the agent). This agent monitor is autonomous once started, and does not interact with the main agent.

It brings the following functionality:
- able to parse version string at runtime from agent itself.

- split string into numeric fields allowing sorting / filtering etc.

- validates that it is semantic in format.

- via config, allows for ignoring version format, which will send the string to DB, but not numerics (as these are no longer useful if format is invalid). Can be useful in testing.

- listen and report on signals received by telegraf.

- all configurable via telegraf agent configuration (frequency, what it collects etc).

The version allows for queries like this (Chronograf as example):  
Screenshot 2019-07-09 at 07 43 02

And the state changes look like this:  
Screenshot 2019-07-09 at 14 47 37

### Required for all PRs:

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"os"
 	"runtime"
 	"sync"
 	"time"
@@ -29,7 +30,7 @@ func NewAgent(config *config.Config) (*Agent, error) {
 }
 
 // Run starts and runs the Agent until the context is done.
-func (a *Agent) Run(ctx context.Context) error {
+func (a *Agent) Run(ctx context.Context, signalsMonitorAgent chan os.Signal) error {
 	log.Printf("I! [agent] Config: Interval:%s, Quiet:%#v, Hostname:%#v, "+
 		"Flush Interval:%s",
 		a.Config.Agent.Interval.Duration, a.Config.Agent.Quiet,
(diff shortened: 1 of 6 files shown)


danielnelson:
Nice, thanks for sharing.  This gives me some ideas, but I do think we will probably want to implement this as part of the internal plugin using the selfstats system.  Still I'd like keep this open for anyone who wants something now and also as a reference.

nicgrobler:
@danielnelson yep, I initially went that way, but needed state changes (signals), and the code smell got pretty bad. This represented the ‘least bad’ solution I came up with without rewriting large parts of the ‘internal’ code to do it...I am happy if you get anything from my efforts, and look forward to learning how you guys solve it

srebhan:
Can you please explain why you don't send the metric in the existing `agent` at the end of the startup? It should have all information I think...

srebhan:
@nicgrobler are you still interested in this PR?

powersj:
Hi,

Thanks!

Issue #6107, opened 2019-07-10
lustre2 input plugin not customizable due to TOML parsing issue


shawnahall71:
As requested by @danielnelson in #5771.

### Relevant telegraf.conf:

```toml
# Read metrics from local Lustre service on OST, MDS
[[inputs.lustre2]]
  # An array of /proc globs to search for Lustre stats
  # If not specified, the default will work on Lustre 2.5.x
  #
   ost_procfiles = [
     "/proc/fs/lustre/obdfilter/*/stats",
     "/proc/fs/lustre/osd-ldiskfs/*/stats",
   #  "/proc/fs/lustre/obdfilter/*/job_stats",
   ]
   mds_procfiles = [
     "/proc/fs/lustre/mdt/*/md_stats",
   #  "/proc/fs/lustre/mdt/*/job_stats",
   ]
```

### System info:

Telegraf 1.11.2 on CentOS 7.4

### Steps to reproduce:

Use the above Telegraf config on different versions of Telegraf.

### Expected behavior:

When I use telegraf 1.10.4, it works as expected:

```
# telegraf -config /etc/telegraf/telegraf.conf.test -test                                                                                                                                                                            
2019-07-10T13:01:29Z I! Starting Telegraf 1.10.4
> cpu,cpu=cpu0,host=hpclbo00 usage_guest=0,usage_guest_nice=0,usage_idle=98.00000004470348,usage_iowait=0,usage_irq=0,usage_nice=0,usage_softirq=0,usage_steal=0,usage_system=1.999999996041879,usage_user=0 1562763690000000000
> cpu,cpu=cpu1,host=hpclbo00 usage_guest=0,usage_guest_nice=0,usage_idle=98.0392157292421,usage_iowait=0,usage_irq=0,usage_nice=0,usage_softirq=0,usage_steal=0,usage_system=1.9607843164109648,usage_user=0 1562763690000000000
> cpu,cpu=cpu2,host=hpclbo00 usage_guest=0,usage_guest_nice=0,usage_idle=100,usage_iowait=0,usage_irq=0,usage_nice=0,usage_softirq=0,usage_steal=0,usage_system=0,usage_user=0 1562763690000000000
[...]

PR #5771, opened 2019-04-26
Fixed an issue in the lustre input polgin with the jobstats only being reporeted for one job per storage target

frroberts:
The current Lustre input plugin will only report jobstats for one job per storage target. This can be tested by running the included tests with the current version of the input plugin. 

With this fix all jobs active on each storage target are recorded.

### Required for all PRs:

- [x] Signed [CLA](https://influxdata.com/community/cla/).
- [x] Associated README.md updated.
- [x] Has appropriate unit tests.

PR #6114, opened 2019-07-12
Fix lustre2 input plugin config parse regression

GeorgeMac:
This closes https://github.com/influxdata/telegraf/issues/6107

PR contains a test definition to capture the regression scenario and an associated fix.

Oustanding question:

Was there a desire to parse these new `ost_jobstat` and `mds_jobstat` fields also? Do we need to support those field names as aliases?

### Required for all PRs:

- [x] ~Associated README.md updated.~

### DIFF
--- a/plugins/inputs/lustre2/lustre2.go
+++ b/plugins/inputs/lustre2/lustre2.go
@@ -25,8 +25,8 @@ type tags struct {
 // Lustre proc files can change between versions, so we want to future-proof
 // by letting people choose what to look at.
 type Lustre2 struct {
-	Ost_procfiles []string `toml:"ost_jobstat"`
-	Mds_procfiles []string `toml:"mds_jobstat"`
+	Ost_procfiles []string
+	Mds_procfiles []string
 
 	// allFields maps and OST name to the metric fields associated with that OST
 	allFields map[tags]map[string]interface{}
--- a/plugins/inputs/lustre2/lustre2.go
+++ b/plugins/inputs/lustre2/lustre2.go
@@ -25,8 +25,8 @@ type tags struct {
 // Lustre proc files can change between versions, so we want to future-proof
 // by letting people choose what to look at.
 type Lustre2 struct {
-	Ost_procfiles []string
-	Mds_procfiles []string
+	Ost_procfiles []string `toml:"ost_procfiles"`
+	Mds_procfiles []string `toml:"mds_procfiles"`
 
 	// allFields maps and OST name to the metric fields associated with that OST
 	allFields map[tags]map[string]interface{}
--- a/plugins/inputs/lustre2/lustre2_test.go
+++ b/plugins/inputs/lustre2/lustre2_test.go
@@ -360,7 +360,7 @@ func TestLustre2CanParseConfiguration(t *testing.T) {
 
 	assert.Equal(t, Lustre2{
 		Ost_procfiles: []string{
-			"/proc/fs/lustre2/obdfilter/*/stats",
+			"/proc/fs/lustre/obdfilter/*/stats",
 			"/proc/fs/lustre/osd-ldiskfs/*/stats",
 		},
 		Mds_procfiles: []string{
--- a/plugins/inputs/lustre2/lustre2_test.go
+++ b/plugins/inputs/lustre2/lustre2_test.go
@@ -6,6 +6,9 @@ import (
 	"testing"
 
 	"github.com/influxdata/telegraf/testutil"
+	"github.com/influxdata/toml"
+	"github.com/influxdata/toml/ast"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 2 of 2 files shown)


GeorgeMac:
@frroberts will reverting this name change cause issues with your previous change https://github.com/influxdata/telegraf/pull/5771 ?

frroberts:
I don't see why it would cause any issues. (Unfortunately I do not have access to a real system to test on right now.)

plugins/inputs/lustre2/lustre2.go
```diff
@@ -25,8 +25,8 @@ type tags struct {
 // Lustre proc files can change between versions, so we want to future-proof
 // by letting people choose what to look at.
 type Lustre2 struct {
-	Ost_procfiles []string `toml:"ost_jobstat"`
-	Mds_procfiles []string `toml:"mds_jobstat"`
+	Ost_procfiles []string
+	Mds_procfiles []string
```
danielnelson: We should keep the struct tags, they just need corrected.  This is mostly style but I think it also blocks alternate capitalization like `Ost_procfiles = ["foo"]`

plugins/inputs/lustre2/lustre2_test.go
```diff
@@ -330,3 +333,38 @@ func TestLustre2GeneratesJobstatsMetrics(t *testing.T) {
...
+   ost_procfiles = [
+     "/proc/fs/lustre/obdfilter/*/stats",
+     "/proc/fs/lustre/osd-ldiskfs/*/stats",
+   ]
+   mds_procfiles = [
+     "/proc/fs/lustre/mdt/*/md_stats",
+   ]`)
+
+	table, err := toml.Parse([]byte(config))
+	assert.NoError(t, err)
```
danielnelson: Use `require.NoError(`, since this will stop the test if it fails and other follow on errors which will just be confusing.  I almost always use `require` instead of `assert`.

plugins/inputs/lustre2/lustre2_test.go
```diff
@@ -330,3 +333,38 @@ func TestLustre2GeneratesJobstatsMetrics(t *testing.T) {
 	err = os.RemoveAll(os.TempDir() + "/telegraf")
 	require.NoError(t, err)
 }
+
+func TestLustre2CanParseConfiguration(t *testing.T) {
```
danielnelson: Just an observation.  I have mixed feelings on this test, you could imagine having tests like this for all plugins and overall I think if we did this it would be too burdensome in it's current form.  On the other hand obviously it does check for a valid bug.

plugins/inputs/lustre2/lustre2_test.go
GeorgeMac: Happy to 👍 My (rough) rule of thumb has always been `require` when I need to fail fast and its not something I am directly testing. i.e. my test is focussed on behavior A, but there are some error conditions which need to be nil before I can assert that. Saying that, I totally disregarded that rule here 😂

plugins/inputs/lustre2/lustre2_test.go
GeorgeMac: I agree that in this form it would be burdensome. I actually had mentioned about having a test like this everywhere with @goller.

Would a kind of test harness for all plugins be worthwhile?

e.g.

```go
import harness "github.com/influxdata/telegraf/internal/harness/input"

func TestHarness(t *testing.T) {
    harness.TestPlugin(t, "some_plugin", &SomePlugin{})
}
```

Which would validate the plugin has been registered under `some_plugin`. Parse the result of `plugin.SamplePlugin()` and assert it against the provided `telegraf.Input`.

Could have one per plugin type and capture some high-level criteria like this.

plugins/inputs/lustre2/lustre2_test.go
danielnelson: I can see that from a language point of view, require the pre-conditions and then make the assertion.  But the semantic difference between the two forms is "continue or not", and I've found the answer is virtually always no.  In practice its probably moot since a good test checks one thing, and the assert would just be the last check.

plugins/inputs/lustre2/lustre2_test.go
danielnelson: I guess I'm ambivalent, it's not something that has historically been a huge source of bugs, so it might just not be worth the effort or a situation where we have bigger fish to fry.  We could definitely provide a better interface to the config package to ease toml string -> plugin, but I'm not entirely sure how the harness would work.  It seems like you need to provide at a minimum the input TOML with all options set, and a filled plugin struct to compare against.

GeorgeMac: Yeah my illustration is lacking a little. For this PR shall I just leave it as is or remove the test?

The harness would ideally test all configurable fields, so the sample would need uncommenting in part if it were used. This could be done with a simple regexp, but that may be brittle and difficult to explain.

Issue #5153, opened 2018-12-18
Convert to go 1.11 modules


smyrman:
## Feature Request

I searched and didn't find an issue for this. I realize telegraf have _recently_ (within the last 6 months) moved to Go `dep`. But is there interest for moving it yet again to `go modules`?

### Proposal:

Use go modules instead of `dep` (and Godeps?) for managing/installing dependencies. For maintenance reasons, remove the old dependency files and update Makefiles to utilize them.

Requires the minimal supported Go version to be 1.11

### Current behavior:

Uses `dep` (and Godeps?) for managing dependencies.

### Use case: [Why is this important (helps with prioritizing requests)]

First of all, it's a lot faster for the number of dependencies required by telegraf to utilize `go modules`, especially while working with several check-outs:

- Go modules appear to be faster at downloading dependencies.
- Go modules holds a user-scoped cache of all dependencies, making it faster and more convinient to work with

Secondly, it doesn't require telegraf to be checked out into GOPATH, making it easeir to checkout and work with telegraf using multiple checkouts.

As a final small benefit, if working with a project that uses go modules, checking out a copy of telegraf within your application module (.e.g. as part of a Docker  build-process), the main module won't dive into telegraf to look for dependencies, and won't run it's tests on `go teset ./...`.

prydin:
+1

danielnelson:
We will consider this and likely make the change in the future but for now we still need support for building with Go 1.9 and we don't want to support two systems.

caelifer:
> We will consider this and likely make the change in the future but for now we still need support for building with Go 1.9 and we don't want to support two systems.

As of Go v1.9.7, support for go.mod was backported - [cmd/go: add minimal support for vgo transition](https://github.com/golang/go/issues?q=milestone%3AGo1.9.7). Would this speed up adoption of official Go modules?

danielnelson:
It might, could you try to update the appveyor and circleci configurations and see if we can get it building on all versions?

PR #6118, opened 2019-07-15
Use go modules for dependency management

GeorgeMac:
This will eventually close https://github.com/influxdata/telegraf/issues/5153

This is a work in progress attempt to move over to go modules from dep.

### Required for all PRs:


danielnelson:
Sorry, I broke it in #6101

GeorgeMac:
s'all good @danielnelson This is bound to conflict with most PRs.

GeorgeMac:
@danielnelson I bumped the gjson version to v1.3.0 and added the pretty package as per that PR to this one.

glinton:
I seem to remember a discussion somewhere about holding off on go modules for telegraf until after the release of go 1.13 as there could/would be disruptions to modules functionality or something.

appveyor.yml
```diff
@@ -2,39 +2,33 @@ version: "{build}"
...
-  - IF NOT EXIST "C:\Cache\gnuwin32-dep.zip" curl -o "C:\Cache\gnuwin32-dep.zip" https://dl.influxdata.com/telegraf/ci/make-3.81-dep.zip
   - IF EXIST "C:\Go" rmdir /S /Q C:\Go
   - msiexec.exe /i "C:\Cache\go1.12.5.msi" /quiet
-  - 7z x "C:\Cache\gnuwin32-bin.zip" -oC:\GnuWin32 -y
-  - 7z x "C:\Cache\gnuwin32-dep.zip" -oC:\GnuWin32 -y
-  - go get -d github.com/golang/dep
-  - cd "%GOPATH%\src\github.com\golang\dep"
-  - git checkout -q v0.5.0
-  - go install -ldflags="-X main.version=v0.5.0" ./cmd/dep
+  - choco install make
```
danielnelson: Nice, but do you know if this gets cached?  If so, and especially if we can install specific versions, maybe we should use it for Go too?

README.md
```diff
@@ -42,8 +42,7 @@ Ansible role: https://github.com/rossmcdonald/telegraf
 
 Telegraf requires golang version 1.10 or newer, the Makefile requires GNU make.
 
-1. [Install Go](https://golang.org/doc/install) >=1.10 (1.12 recommended)
-2. [Install dep](https://golang.github.io/dep/docs/installation.html) ==v0.5.0
+1. [Install Go](https://golang.org/doc/install) >=1.11 (1.12 recommended)
```
danielnelson: We should just remove the 1.10 tests if we bump this, I assume that it's just not possible to `go mod download` in 1.10?

I think this looks fine, but one thing I always check when updating Go is the operating systems they removed support for:

> As announced in the Go 1.10 release notes, Go 1.11 now requires OpenBSD 6.2 or later, macOS 10.10 Yosemite or later, or Windows 7 or later; support for previous versions of these operating systems has been removed.

go.mod
```diff
@@ -0,0 +1,119 @@
+module github.com/influxdata/telegraf
+
+go 1.12
```
danielnelson: What is the significance of this line?  Should it be the oldest supported version `go 1.11`?

Makefile
```diff
@@ -33,7 +33,7 @@ all:
 
 .PHONY: deps
 deps:
-	dep ensure -vendor-only
+	go mod download
```
danielnelson: Related to the previous `-vendor-only` option, we need to ensure that the command here does not modify the go.mod/go.sum files.

Also, we need to ensure that the commands in other targets never download or modify the go.mod/go.sum.

go.mod
```diff
@@ -0,0 +1,119 @@
...
+	github.com/googleapis/gax-go v2.0.0+incompatible // indirect
+	github.com/gorilla/mux v1.6.2
+	github.com/hailocab/go-hostpool v0.0.0-20160125115350-e80d13ce29ed // indirect
+	github.com/harlow/kinesis-consumer v0.0.0-20181230152818-2f58b136fee0
+	github.com/hashicorp/consul v1.2.1
+	github.com/hashicorp/go-cleanhttp v0.0.0-20171218145408-d5fe4b57a186 // indirect
+	github.com/hashicorp/go-rootcerts v0.0.0-20160503143440-6bb64b370b90 // indirect
+	github.com/hashicorp/serf v0.8.1 // indirect
+	github.com/influxdata/go-syslog v1.0.1 // indirect
+	github.com/influxdata/go-syslog/v2 v2.0.0
```
danielnelson: Why is go-syslog listed twice?  Is this just one of the `go mod tidy` things?

.circleci/config.yml
```diff
@@ -19,54 +25,55 @@ jobs:
     steps:
       - checkout
       - restore_cache:
-          key: vendor-{{ checksum "Gopkg.lock" }}
+          key: go-mod-v1-{{ checksum "go.sum" }}
```
danielnelson: I reran the workflow on CircleCI, and it said:
> No cache is found for key: go-mod-v1-5m8iTjU8woo3ixYf_mEKZ4D_98q5g9IaxeuniYq+mTU=

Is the cache working right?

README.md
GeorgeMac: Yeah I had to add the explicit `go mod vendor` to the `deps` task of the circle ci to get around this. Because by the time the workspace is moved into the go 1.10 image all the `go mod` commands don't exist and the go command is oblivious to the go mod build cache in `pkg`. So the vendor directly needs to be populated. Actually removing this step would mean we can drop the extra vendor step and make the packing and migrating of the workspace quicker as it should take up nearly half the space.

go.mod
GeorgeMac: I believe there were some module feature transitions from `go 1.11` to `go 1.12`. Though most have been back ported to 1.11 somewhere around `>= go 1.11.4`. At which point `go 1.11` can parse `go 1.12` safely. So having it at 1.12 should be fine and backwards compatible as far as 1.11.4.

https://github.com/golang/go/issues/30446

go.mod
GeorgeMac: 🤔 I will investigate. Looks like it is keeping both versions around. Maybe we have references to both.

.circleci/config.yml
GeorgeMac: It should be. This is actualy how circle recommend configuring their cache for go modules.

https://circleci.com/blog/go-v1.11-modules-and-circleci/#caching

I can poke and prod a bit more. Perhaps our cache got invalidated for some reason.

appveyor.yml
GeorgeMac: Im not sure how to get appveyors cache to speed up chocolatey installs. Though I watched a fair few appveyor builds while getting this to work and it took no time at all. Also I had issues with the old gnu make installation. It just kept failing with a non-descript error code. As soon as I installed it via choco it just worked :/

GeorgeMac: Just had a look at there is a `choco install golang` 👍

Makefile
GeorgeMac: Looks like `go mod download` adhears to the current `go.sum` contents. Plus we do a `go mod verify`:
```
The 'go mod verify' command checks that
the cached copies of module downloads still match both their recorded
checksums and the entries in go.sum.
```

GeorgeMac: We do this in circleci after the module download.

go.mod
GeorgeMac: Turns out I needed to do a `go mod tidy`!

appveyor.yml
GeorgeMac: Looks like we could cache the directory chocolately uses to download packages into https://help.appveyor.com/discussions/suggestions/592-speeding-up-chocolatey-downloads-and-installs and then invalidate it on appveyor change. We could then explicitly set golang version?
Thoughts?

danielnelson: Yeah let's try it

Makefile
danielnelson: If someone added a plugin with a new dependency but didn't add it to go.sum would this still pass?  The cached copies would still match the go.sum.

GeorgeMac: Good though. Will experiment and check back in!

GeorgeMac: I cannot find an equivalent to dep ensure --vendor-only. All the go mod commands appear to manipulate the `go.sum` file so I am adding a manual check using git diff.

danielnelson: What about `-mod=readonly`?

https://github.com/golang/go/wiki/Modules#can-i-control-when-gomod-gets-updated-and-when-the-go-tools-use-the-network-to-satisfy-dependencies

GeorgeMac: 🙌 ahh great find

GeorgeMac: Just had a play and basically you can put `-mod=readonly` on any non `mod` command. i.e. `go build` or `go install`. We would need to add that to everyone. The result of which is that command will fail iff any dependencies in the mod file are not in the build cache. However, if you do a `go mod download` they will already be there. At which point it still update the go.sum file with the one it has in its build cache. So the result is still an updated go.sum file that is not checked in.

affo: As far as I know  we use `go mod tidy` in order to assert that the `go.mod` and `go.sum` are in sync with the dependencies that the codebase really uses.

We have a script that checks that `go mod tidy` didn't produce any change in mod/sum files: https://github.com/influxdata/flux/blob/master/etc/checktidy.sh. That script is used  in CI in the step `make checktidy`: https://github.com/influxdata/flux/blob/master/Makefile#L58.

@danielnelson do you think that approach could solve your problem?

pgray:
This PR is awesome. Any idea when it'll get merged? Looks like it's mostly there minus a rebase?

Issue #1992, opened 2016-11-04
Aggregator plugins: Support historical data


sparrc:
Currently aggregator plugins only support aggregating data with "current" timestamps.

We should try to support a single stream of historical data in the aggregator plugins, as this likely could be a common use-case when parsing a logfile.

see https://github.com/influxdata/telegraf/issues/1989

from that issue:

> One "easy" workaround might be to create a single "historical window" which would have it's period started on the timestamp of the first "historical" metric that arrives. This would cover your use-case where you have a single stream of historical and ordered metrics arriving (ordered is the crucial point). I imagine that when dealing with historical data, this would be the most common use-case, especially when dealing with logfile parsing.

mujahideen555:
Is there [[aggregator.mean]] plugin available in telegraf ? I am searching through but couldn't find if there is any aggregator plugin other than [[aggregator.minmax]].

danielnelson:
@ghost Not currently but there is one in the works for 1.3.0 #2167

bolek2000:
we are testing log parsing with telegraf and plan to import older logfiles into the DB. It will be handy if we can use the aggregator for historical data too.

kevin-cantwell:
Can this issue be closed now with the advent of https://github.com/influxdata/telegraf/pull/6049 ?

ssoroka:
I think so. If the grace period doesn't meet your needs, you also have the ability to use the state variable in a [custom starlark script](https://github.com/influxdata/telegraf/blob/master/plugins/processors/starlark/testdata/compare_metrics.star) to write any sort of aggregation rules you would like, at least as far as available memory will allow.

PR #6119, opened 2019-07-15
aggregators: allow historical data in aggregators

kevin-cantwell:
Addresses https://github.com/influxdata/telegraf/issues/1992

This PR adds a boolean config to aggregator plugins, `allow_historical`, which overrides the default behavior of ignoring metrics outside the period of the plugin. This override is useful when aggregating old logs or metrics that are delayed beyond the period window.

### Required for all PRs:

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -1065,6 +1065,18 @@ func buildAggregator(name string, tbl *ast.Table) (*models.AggregatorConfig, err
 		}
 	}
 
+	if node, ok := tbl.Fields["allow_historical"]; ok {
+		if kv, ok := node.(*ast.KeyValue); ok {
+			if b, ok := kv.Value.(*ast.Boolean); ok {
+				var err error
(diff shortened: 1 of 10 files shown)


internal/config/config.go
```diff
@@ -1065,6 +1065,18 @@ func buildAggregator(name string, tbl *ast.Table) (*models.AggregatorConfig, err
 		}
 	}
 
+	if node, ok := tbl.Fields["allow_historical"]; ok {
```
cagollnick: am I correctly inferring that this is internal to telegraf? would this work for all plugins?

kevin-cantwell: It would work for all aggregator type plugins. Which means it extends all existing aggregator plugins with added functionality, but because the default value is false, it does not have adverse affects to existing deployments.

danielnelson:
I just merged #6049, it should be possible to use it to get this type of behavior just by setting the new `grace` period option very high.

kevin-cantwell:
Thanks @danielnelson your solution works for us 👍

PR #6132, opened 2019-07-17
Add per user metrics to mysql input

### LINKED ISSUE #6131: Add metrics from performance_schema.events_statements_summary_by_account_by_event_name
## Feature Request

Currently it's not possible to gather metrics per username (number of queries and so on) for mysql plugin

### Proposal:

There is the table which contains desired information: performance_schema.events_statements_summary_by_account_by_event_name need to add this functionality to mysql plugin.
For example: add flag into config:
`gather_perf_sum_per_acc_per_event = true`
and we can filter by only desired events:
`perf_summary_events = ["statement/sql/select", "statement/sql/update", "statement/sql/delete", "statement/sql/insert", "statement/sql/commit", "statement/sql/rollback", "statement/sql/savepoint"]`

### Current behavior:
absent
### Desired behavior:
metrics from performance_schema.events_statements_summary_by_account_by_event_name is available on influxdb

### Use case: 
In system where lots of component it's hard to understand which component overuses resources. Will be very helpful for monitoring multi-component setup

AlexKapustin:
closes #6131

### Required for all PRs:

- [ N/A] Has appropriate unit tests.


AlexKapustin:
Which version of go do you guys use ?
I checked with go1.9.4 & go1.10.4 - both are fine, and gofmt does not change file

AlexKapustin:
Pull request was opened one year ago, and no any single comment :(
It highly "encourage and motivate" people to contribute

plugins/inputs/mysql/mysql.go
```diff
@@ -115,6 +117,12 @@ var sampleConfig = `
   ## Some queries we may want to run less often (such as SHOW GLOBAL VARIABLES)
   interval_slow                   = "30m"
 
+  ## gather metrics from PERFORMANCE_SCHEMA.EVENTS_STATEMENTS_SUMMARY_BY_ACCOUNT_BY_EVENT_NAME
+  gather_perf_sum_per_acc_per_event         = false
+  ## list of events to be gathered for gather_perf_sum_per_acc_per_event
```
srebhan, line 122: Add newline here for readability.

plugins/inputs/mysql/mysql.go
```diff
@@ -1227,6 +1274,142 @@ func (m *Mysql) gatherInnoDBMetrics(db *sql.DB, serv string, acc telegraf.Accumu
 	return nil
 }
 
+// gatherPerfSummaryPerAccountPerEvent can be used to fetch enabled metrics from
+// performance_schema.events_statements_summary_by_account_by_event_name
+func (m *Mysql) gatherPerfSummaryPerAccountPerEvent(db *sql.DB, serv string, acc telegraf.Accumulator) error {
+
```
srebhan, line 1280: Remove newline to match style of other gather functions.

plugins/inputs/mysql/mysql.go
```diff
@@ -1227,6 +1274,142 @@ func (m *Mysql) gatherInnoDBMetrics(db *sql.DB, serv string, acc telegraf.Accumu
...
+			&sumSelectFullRangeJoin,
+			&sumSelectRange,
+			&sumSelectRangeCheck,
+			&sumSelectScan,
+			&sumSortMergePasses,
+			&sumSortRange,
+			&sumSortRows,
+			&sumSortScan,
+			&sumNoIndexUsed,
+			&sumNoGoodIndexUsed,
```
srebhan, line 1369: Hmm I think this won't work if you specify only a subset of events using the `perf_summary_events` option. So either you change the logic here or remove the option to select the events to query...

srebhan (changes requested):
Hey @AlexKapustin! I totally agree that this is killing motivation and I want to apologize for this taking so long. InfluxData is trying to improve and here I am.. ;-P

I have some comments in the code. The most striking one is that I think the code breaks when limiting the number of collected events using the `perf_summary_events` option. Could you please take a look!? I promise it will not take another year to get this in, given we can work out the issues. :-/

AlexKapustin:
Hello, @srebhan !

Thank you very much for your response :) 
No need to be sorry, thank you for reply at all :)
And sorry for my late reply. I will have a look at your comments shortly

plugins/inputs/mysql/mysql.go
AlexKapustin: I think here is a bit misunderstanding.
Main source of data is table: performance_schema.events_statements_summary_by_account_by_event_name

table structure is:
event_name | metrics | which | are | provided | above.

Basically saying, it's key-value structure. Table has many "keys". Providing  `perf_summary_events` just limit number of rows to be collected, basically saying we filter rows, but colunms will be always the same.

srebhan: Oh I see, so it limits what comes out in the third column and thus the rows, but the row structure stays the same!? 

If so, then please mark this part as resolved. But please consider putting this explanation as a comment next to the code above! Just for the next one not asking the same question... :-)

AlexKapustin: > but the row structure stays the same!?
Yes, exactly!

I will add comment. Thank you.

plugins/inputs/mysql/mysql.go
```diff
@@ -121,6 +123,13 @@ const sampleConfig = `
   # perf_events_statements_limit = 250
   # perf_events_statements_time_limit = 86400
 
+  ## gather metrics from PERFORMANCE_SCHEMA.EVENTS_STATEMENTS_SUMMARY_BY_ACCOUNT_BY_EVENT_NAME
+  gather_perf_sum_per_acc_per_event         = false
```
srebhan, line 127: Sorry for missing this: Please comment the lines containing default values!

plugins/inputs/mysql/mysql.go
```diff
@@ -121,6 +123,13 @@ const sampleConfig = `
...
+
+  ## list of events to be gathered for gather_perf_sum_per_acc_per_event
+  ## in case of empty list all events will be gathered
+  perf_summary_events                       = []
```
srebhan, line 131: Same here, comment that line as it contains the default.

AlexKapustin:
It's done! Thank you for your review.

srebhan (approved):
Looks good to me.

reimda (commented):
This change could also use some tests.  The mysql plugin in general doesn't have much test coverage.  We could mock the result of a query and make sure the output metric is as expected.  Even an integration test that requires mysql to be running would be useful.

srebhan:
@AlexKapustin could you look into the unit tests @reimda  suggested!?

AlexKapustin:
> @AlexKapustin could you look into the unit tests @reimda suggested!?

Yes, I will have a look today. Thank you for remind me :)

AlexKapustin:
I had have a look, and basically saying, mysql plugin doesn't have tests at all, only connections and one integration test!
Do you, guys, think it would be good idea to separate that from this scope, create new ticket to cover all measurements by tests?
It's quite big task to do so, hence, I would prefer to create separate ticket for that.
Thoughts?

reimda:
Hi @AlexKapustin, thanks for being so responsive even though it took us so long to get to this PR.

It would be great if you could write tests in this PR that cover the new functionality in this PR.  If this will require setting up a lot of common infrastructure and you would like to do it in a separate PR that's ok too.  Whatever you prefer.

I went ahead and made a new issue about the mysql missing tests (#8561).  If you would like to write the missing tests for the existing functionality it would be very appreciated!

srebhan:
Thanks @AlexKapustin for clarification. I think we can merge this one as the code looks fine... As @reimda said, tests are appreciated and I promise it won't take another year to review them. ;-)

AlexKapustin:
Hello, @reimda  and @srebhan !
Thank you for your responses.
I would prefer to do tests in new ticket. I will try to come up with some tests in #8561 
Thanks again!

reimda (approved):

PR #6134, opened 2019-07-17
feat: add customjson plugin serializers

FortunatCE:
### Required for all PRs:

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -1918,6 +1918,22 @@ func buildSerializer(name string, tbl *ast.Table) (serializers.Serializer, error
 		}
 	}
 
+	if node, ok := tbl.Fields["jmespath_expression"]; ok {
+		if kv, ok := node.(*ast.KeyValue); ok {
+			if str, ok := kv.Value.(*ast.String); ok {
+				c.JmespathExpression = str.Value
+			}
+		}
+	}
+
+	if node, ok := tbl.Fields["tags_prefix"]; ok {
+		if kv, ok := node.(*ast.KeyValue); ok {
+			if str, ok := kv.Value.(*ast.String); ok {
+				c.TagsPrefix = str.Value
+			}
+		}
+	}
+
 	delete(tbl.Fields, "influx_max_line_bytes")
 	delete(tbl.Fields, "influx_sort_fields")
 	delete(tbl.Fields, "influx_uint_support")
@@ -1929,6 +1945,8 @@ func buildSerializer(name string, tbl *ast.Table) (serializers.Serializer, error
 	delete(tbl.Fields, "splunkmetric_hec_routing")
 	delete(tbl.Fields, "wavefront_source_override")
 	delete(tbl.Fields, "wavefront_use_strict")
+	delete(tbl.Fields, "jmespath_expression")
+	delete(tbl.Fields, "tags_prefix")
 	return serializers.NewSerializer(c)
 }
 
--- a/plugins/serializers/customjson/customjson.go
+++ b/plugins/serializers/customjson/customjson.go
@@ -1,4 +1,4 @@
-package flattenjson
+package customjson
 
 import (
 	"bytes"
@@ -27,7 +27,7 @@ func (s *serializer) Serialize(metric telegraf.Metric) ([]byte, error) {
 
 	m, err := s.createObject(metric)
 	if err != nil {
-		return nil, fmt.Errorf("D! [serializer.flattenjson] Dropping invalid metric: %s", metric.Name())
+		return nil, fmt.Errorf("D! [serializer.customjson] Dropping invalid metric: %s", metric.Name())
 	}
 
 	return m, nil
@@ -40,7 +40,7 @@ func (s *serializer) SerializeBatch(metrics []telegraf.Metric) ([]byte, error) {
 	for _, metric := range metrics {
 		m, err := s.createObject(metric)
 		if err != nil {
-			return nil, fmt.Errorf("D! [serializer.flattenjson] Dropping invalid metric: %s", metric.Name())
+			return nil, fmt.Errorf("D! [serializer.customjson] Dropping invalid metric: %s", metric.Name())
 		} else if m != nil {
 			serialized = append(serialized, m...)
 		}
@@ -52,7 +52,7 @@ func (s *serializer) SerializeBatch(metrics []telegraf.Metric) ([]byte, error) {
 func (s *serializer) createObject(metric telegraf.Metric) (metricGroup []byte, err error) {
 
 	/*  All fields index become dimensions and all tags index can be prefixed by tags_prefix config input and located on json root.
-	    ** Default flattenjson format contains the following fields:
+	    ** Default customjson format contains the following fields:
 		** metric_family: The name of the metric
 		** metric_name:   The name of the fields dimension
 		** metric_value:  The value of the fields dimension
--- a/plugins/serializers/customjson/customjson.go
+++ b/plugins/serializers/customjson/customjson.go
@@ -139,8 +139,8 @@ func buildDefaultTagsParameter(metric telegraf.Metric, dataGroup map[string]inte
 			tagsBuffer.WriteString("_")
 		}
 		tagsBuffer.WriteString(n)
-		tags_name := tagsBuffer.String()
-		dataGroup[tags_name] = t
+		tagsName := tagsBuffer.String()
+		dataGroup[tagsName] = t
 	}
 	return json.Marshal(dataGroup)
 }
--- a/plugins/serializers/registry.go
+++ b/plugins/serializers/registry.go
@@ -6,7 +6,7 @@ import (
 
 	"github.com/influxdata/telegraf"
 	"github.com/influxdata/telegraf/plugins/serializers/carbon2"
-	"github.com/influxdata/telegraf/plugins/serializers/flattenjson"
+	"github.com/influxdata/telegraf/plugins/serializers/customjson"
 	"github.com/influxdata/telegraf/plugins/serializers/graphite"
 	"github.com/influxdata/telegraf/plugins/serializers/influx"
 	"github.com/influxdata/telegraf/plugins/serializers/json"
@@ -81,10 +81,10 @@ type Config struct {
 	// When enabled forward slash (/) and comma (,) will be accepted
 	WavefrontUseStrict bool
 
-	// Include JMESPath expression for flattenjson output
+	// Include JMESPath expression for customjson output
 	JmespathExpression string
 
-	// Include tags prefix for flattenjson output
+	// Include tags prefix for customjson output
 	TagsPrefix string
 }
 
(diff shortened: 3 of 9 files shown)


danielnelson:
There are so many potential formats for outputting JSON, I wonder if we could have a JMESPath expression that could transform the output JSON from a known base layout into a custom layout.  Would it be possible for you to check if JMESPath could do this transformation from the normal JSON serializer output?

FortunatCE:
> There are so many potential formats for outputting JSON, I wonder if we could have a JMESPath expression that could transform the output JSON from a known base layout into a custom layout. Would it be possible for you to check if JMESPath could do this transformation from the normal JSON serializer output?

Thanks for your quick response.
I think it will be more appropriate to use Jolt expression in order to transform the default json format into a custom one.
The main idea behind this PR is to have a unitary metric record, in which you will have a metric name and its value.
Since JMESPath provides only querying expressions on json documents, I'm not sure that we could use it to generate unitary metric records.

danielnelson:
Makes sense, appreciate the advice.  Jolt might be hard to integrate, since it is written in Java, but I did find the [qntfy/kazaam](https://github.com/qntfy/kazaam) project that could be an option.

FortunatCE:
> Makes sense, appreciate the advice. Jolt might be hard to integrate, since it is written in Java, but I did find the [qntfy/kazaam](https://github.com/qntfy/kazaam) project that could be an option.

Thanks for your proposition.
We tried JMESPath and qntfy/kazaam, but we couldn't managed to achieve the json transformation we need. After several tests, it seems that both libraries are limited and not suited for this tranformation use case.

What we wanted to achieve with this PR is to transform this document:
{
  "fields": {
    "reads": 973,
    "write_bytes": 2097152,
	"foobar": 555
  },
  "name": "diskio",
  "tags": {
    "host": "myhost",
    "name": "vda1"
  },
  "timestamp": 1556936700000
}

into the following expected output:
{"metric_family": "diskio", "timestamp": 1556936700000, "tags_host": "myhost", "tags_name": "vda1", "metric_name": "diskio_reads", "metric_value": 973}
{"metric_family": "diskio", "timestamp": 1556936700000, "tags_host": "myhost", "tags_name": "vda1", "metric_name": "diskio_write_bytes", "metric_value": 2097152}
{"metric_family": "diskio", "timestamp": 1556936700000, "tags_host": "myhost", "tags_name": "vda1", "metric_name": "diskio_foobar", "metric_value": 2097152}

The idea is to have basic flattened unitary json metrics data points, so it can be pushed to multiple datastores (like Druid, Elasticsearch, OpenTSDB, ... etc). We can then imagine another output that takes in a kasaam or JMESPath expression and transform this unitary metric data point to another format.

Do you think that if we change the name of the output format from flattenjson to jsonunitarymetric, it would be more appropriate ?

danielnelson:
> We tried JMESPath and qntfy/kazaam, but we couldn't managed to achieve the json transformation we need.

Too bad.  I think we need to keep looking for a good generic solution for building out JSON from Telegraf metrics.  However, I don't think we would accept another json serializer that doesn't allow for user control in some way, any format we pick would just be too specific to a single users need.

FortunatCE:
Based on your response, we have implemented a new serializer that allow user to customize the output format with JMESPath grammar expression based on default unitary metric output.
Commons parameters can be customized with JMESPath grammar by specifying "jmespath_expression" parameter in the Telegraf configuration file and tags parameter can be prefixed by specifying "tags_prefix" parameter in the Telegraf configuration file.
There are some event example on the README.md file that explain how the serializer works.

Do you have any recommendation about this new update please ?
Thanks

danielnelson:
I think this shows our prior concerns were founded, that JMESPath isn't really be suitable for your use case.  The evidence being that you needed to include options specifically for prefixing the tags with a string since this doesn't seem possible.

srebhan:
@FortunatCE are you still interested in getting this merged?

FortunatCE:
> @FortunatCE are you still interested in getting this merged?

Hello @srebhan, 
Yes I am still interested in getting this pull request merged
Regards,

internal/config/config.go
```diff
@@ -1918,6 +1918,22 @@ func buildSerializer(name string, tbl *ast.Table) (serializers.Serializer, error
...
+		}
+	}
+
+	if node, ok := tbl.Fields["tags_prefix"]; ok {
+		if kv, ok := node.(*ast.KeyValue); ok {
+			if str, ok := kv.Value.(*ast.String); ok {
+				c.TagsPrefix = str.Value
+			}
+		}
+	}
```
srebhan, line 1935: How about adding a fields-prefix the same way? Or alternatively define this as a "prefix for elements not handled otherwise"... Does that make sense?

plugins/serializers/customjson/customjson.go
```diff
@@ -0,0 +1,168 @@
...
+func NewSerializer(jmespath_expression string, tags_prefix string) (*serializer, error) {
+	s := &serializer{
+		JmespathExpression: jmespath_expression,
+		TagsPrefix:         tags_prefix,
+	}
+	return s, nil
+}
+
+func (s *serializer) Serialize(metric telegraf.Metric) ([]byte, error) {
+
```
srebhan, line 27: Please remove this newline and all others not helping to structure the code.

plugins/serializers/customjson/customjson.go
```diff
@@ -0,0 +1,168 @@
...
+	m, err := s.createObject(metric)
+	if err != nil {
+		return nil, fmt.Errorf("D! [serializer.customjson] Dropping invalid metric: %s", metric.Name())
```
srebhan, line 30: Please use only the error message for errors e.g.
```suggestion
		return nil, fmt.Errorf("dropping invalid metric: %s", metric.Name())
```

plugins/serializers/customjson/customjson.go
```diff
@@ -0,0 +1,168 @@
...
+}
+
+func (s *serializer) SerializeBatch(metrics []telegraf.Metric) ([]byte, error) {
+
+	var serialized []byte
+
+	for _, metric := range metrics {
+		m, err := s.createObject(metric)
+		if err != nil {
+			return nil, fmt.Errorf("D! [serializer.customjson] Dropping invalid metric: %s", metric.Name())
```
srebhan, line 43: Same as above. Please rethink if you want to stop serialization if one metric is invalid or not. Both is fine for me...

plugins/serializers/customjson/customjson.go
```diff
@@ -0,0 +1,168 @@
...
+	// Build output result
+	dataGroup := map[string]interface{}{}
+	var metricJson []byte
+
+	if s.JmespathExpression != "" {
+		jmespath.MustCompile(s.JmespathExpression)
+	}
+
+	for _, field := range metric.FieldList() {
+
```
srebhan, line 72: Remove newline as it does not help structuring the code but rather clutters it.

plugins/serializers/customjson/customjson.go
```diff
@@ -0,0 +1,168 @@
...
+		if !valid {
+			log.Printf("D! Can not parse value: %v for key: %v", field.Value, field.Key)
+			continue
+		}
+
+		// Build root parameter
+		dataGroup["metric_family"] = metric.Name()
+		// Convert ns to float milliseconds since epoch.
+		dataGroup["timestamp"] = float64(metric.Time().UnixNano()) / float64(1000000)
```
srebhan, line 83: Why do you convert the time here?

plugins/serializers/customjson/customjson.go
```diff
@@ -0,0 +1,168 @@
...
+
+		// Build fields parameter
+		dataGroup["metric_name"] = field.Key
+		dataGroup["metric_value"] = fieldValue
```
srebhan, line 87: Why do you enforce "one-field per output"?

plugins/serializers/customjson/customjson.go
```diff
@@ -0,0 +1,168 @@
...
+
+		// Build tags parameter
+		if s.JmespathExpression != "" {
+			metricJson, err = buildJmespathTagsParameter(metric, dataGroup, s)
+		} else {
+			metricJson, err = buildDefaultTagsParameter(metric, dataGroup, s)
+		}
```
srebhan, line 94: You might want to do this decision in an `Init()` function and store the function pointer in the serializer struct...

plugins/serializers/customjson/customjson.go
```diff
@@ -0,0 +1,168 @@
...
+
+		// Output the data as a fields array.
+		metricJson = append(metricJson, '\n')
```
srebhan, line 97: Is this really necessary?

plugins/serializers/customjson/customjson.go
```diff
@@ -0,0 +1,168 @@
...
+
+		metricGroup = append(metricGroup, metricJson...)
+
+		if err != nil {
+			return nil, err
+		}
```
srebhan, line 103: Which error are you checking here?

plugins/serializers/customjson/customjson.go
```diff
@@ -0,0 +1,168 @@
...
+			// Otherwise store 0
+			valid = true
+			value = 0
+		}
+	default:
+		valid = true
+		value = v
+	}
+	return value, valid
+}
```
srebhan, line 168: Why is a string value invalid? Why do you convert boolean values to numbers? I don't get the idea behind this function... Both string and boolean values are fine for encoding in json as value...

plugins/serializers/registry.go
```diff
@@ -100,6 +107,8 @@ func NewSerializer(config *Config) (Serializer, error) {
 		serializer, err = NewCarbon2Serializer()
 	case "wavefront":
 		serializer, err = NewWavefrontSerializer(config.Prefix, config.WavefrontUseStrict, config.WavefrontSourceOverride)
+	case "customjson":
```
srebhan, line 110: Should we maybe use "jmetransformer" or similar?

srebhan (changes requested):
I left some comments in the code, nothing too big. Please rebase your code on the latest master and convert to go-modules for handling the dependencies.

FortunatCE:
> I left some comments in the code, nothing too big. Please rebase your code on the latest master and convert to go-modules for handling the dependencies.



> I left some comments in the code, nothing too big. Please rebase your code on the latest master and convert to go-modules for handling the dependencies.

Thanks for your review !
I will work on it based on your recommendations. 
Regards,

FortunatCE:
> I left some comments in the code, nothing too big. Please rebase your code on the latest master and convert to go-modules for handling the dependencies.

Thanks for your review @srebhan !
I will work on it based on your recommendations.
Regards,

srebhan:
@FortunatCE any update on this?

srebhan:
@FortunatCE do you think you can work on this PR?

srebhan:
Any volunteer for taking over this PR?

powersj:
Hi,

Thanks!

Issue #5894, opened 2019-05-22
Arbitrary rule can cause all Graphite templates to be ignored


Crack:
### Relevant telegraf.conf:
```toml
[agent]
  omit_hostname = true

[[inputs.socket_listener]]
  service_address = "tcp://:2000"
  data_format = "graphite"
  templates = [
    "taskmanagerTask.alarm-detector.Assign.alarmDefinitionId metricsType.process.nodeId.x.alarmDefinitionId.measurement.field rule=1",
    "taskmanagerTask.*.*.*.*                                 metricsType.process.nodeId.measurement rule=2"
  ]
[[outputs.file]]
  files = ["stdout"]
  data_format = "influx"
  flush_interval = "1s"
```

### System info:

Telegraf 1.10.4

### Steps to reproduce:

1. `telegraf --config conf-from-bug-report.conf`
2. ```
    echo "taskmanagerTask.alarm-detector.Assign.alarmDefinitionId.timeout_errors.duration.p75 100 $(date +%s)" | nc 127.0.0.1 2000
    echo "taskmanagerTask.alarm-detector.Assign.numRecordsInPerSecond.m5_rate 100 $(date +%s)" | nc 127.0.0.1 2000
    ````

### Expected behavior:

Telegraf output:
```
duration,alarmDefinitionId=timeout_errors,metricsType=taskmanagerTask,nodeId=Assign,process=alarm-detector,rule=1,x=alarmDefinitionId p75=100 1558538449000000000
numRecordsInPerSecond,metricsType=taskmanagerTask,nodeId=Assign,process=alarm-detector,rule=2 value=100 1558537634000000000
```

### Actual behavior:

Telegraf output:
```
duration,alarmDefinitionId=timeout_errors,metricsType=taskmanagerTask,nodeId=Assign,process=alarm-detector,rule=1,x=alarmDefinitionId p75=100 1558538449000000000
taskmanagerTask.alarm-detector.Assign.numRecordsInPerSecond.m5_rate value=100 1558537531000000000
```

### Additional info:

I get expected behavior when I remove first template, but I need both of them to work correctly.

GeorgeMac:
After doing some digging into this problem I have found some behavior I need to clarify further.

What I find is hard to express in a PR, so bare with me and I am happy to attempt to attempt to clarify more.

The parse tree for these filters looks like this:
```
[taskmanagerTask]
       |            \
[alarm-detector]    [*]
       |             |
   [Assign]         [*]
       |             |
[...]

PR #6135, opened 2019-07-17
Fix templating patterns wildcard matching

GeorgeMac:
Closes: https://github.com/influxdata/telegraf/issues/5894

This fixes an issue within templating rules used by line formats like the graphite format. The templating rules do not exhaustively explore all matches. The moment a match is found at any level, if the search fails to match the filters children it will bail out and return a nil template (which gets replace by teh caller with a default template).

Instead the template should return to the previous successful match and check to see if a wildcard exists which can also be explored.

### Required for all PRs:

### DIFF
--- a/internal/templating/engine_test.go
+++ b/internal/templating/engine_test.go
@@ -3,6 +3,7 @@ package templating
 import (
 	"testing"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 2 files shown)


internal/templating/engine_test.go
```diff
@@ -20,3 +21,46 @@ func TestEngineAlternateSeparator(t *testing.T) {
...
+	assert.Equal(t, map[string]string{
+		"metricsType":       "taskmanagerTask",
+		"process":           "alarm-detector",
+		"nodeId":            "Assign",
+		"x":                 "alarmDefinitionId",
+		"alarmDefinitionId": "timeout_errors",
+		"rule":              "1",
+	}, tags)
+
+	measurement, tags, field, err = engine.Apply(lineTwo)
```
danielnelson: Do a table test with two subtests, or multiple tests.

GeorgeMac:
@danielnelson which milestone should be put this in?

Issue #5496, opened 2019-02-27
MongoDB input plugin: Adding per Collection stats


carlo4002:
## Feature Request

Since https://github.com/influxdata/telegraf/pull/1466 we got the option to gather metrics for every database. Isn't it useful to have metrics per collection? 
### Proposal:
add the option `gather_percollection_stats`
### Current behavior:
The metrics are collected at database level.
### Desired behavior:
Have a better granularity to collect metrics by database and collection 
### Use case: [Why is this important (helps with prioritizing requests)]
Different collections are dedicated to different services therefore it is important to know which collection grows faster for example

scottprichard:
We are using the mongo_db stats and would love this per collection metric!

danielnelson:
Closed in #6137

PR #6137, opened 2019-07-18
Gather per collections stats in mongodb input plugin

seuf:
This PR add Mongodb input plugin per collection stats gathering.

It will discover collection in each database.
You can specify the databases you want to gather collection stats by setting the `col_stats_db` option.

```
  ## When true, collect per collection stats
  # gather_col_stats = false
  ## List of db where collections stats are collected
  ## If empty, all db are concerned
  # col_stats_dbs = ["local"]
```

New metrics per collections are :
mongodb_col_stats
  - tags:
    - hostname
    - collection
    - db_name
  - fields:
    - size (integer)
    - avg_obj_size (integer)
    - storage_size (integer)
    - total_index_size (integer)
    - ok (integer)
    - count (integer)
    - type (tring)

Related Issue : https://github.com/influxdata/telegraf/issues/5496

### Required for all PRs:

- [x] Signed [CLA](https://influxdata.com/community/cla/). (signed a long time ago. Still valid ?)
- [x] Has appropriate unit tests. (no dedicated unit tests for custom gathering)

### DIFF
--- a/plugins/inputs/mongodb/mongodb_server.go
+++ b/plugins/inputs/mongodb/mongodb_server.go
@@ -70,7 +70,7 @@ func (s *Server) gatherOplogStats() *OplogStats {
 	return stats
 }
 
-func (s *Server) gatherData(acc telegraf.Accumulator, gatherDbStats bool) error {
+func (s *Server) gatherData(acc telegraf.Accumulator, gatherDbStats bool, gatherColStats bool, colStatsDbs []string) error {
 	s.Session.SetMode(mgo.Eventual, true)
 	s.Session.SetSocketTimeout(0)
 	result_server := &ServerStatus{}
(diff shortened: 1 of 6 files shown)


plugins/inputs/mongodb/mongodb_server.go
```diff
@@ -147,11 +147,48 @@ func (s *Server) gatherData(acc telegraf.Accumulator, gatherDbStats bool) error
...
+				for _, col_name := range colls {
+					col_stat_line := &ColStatsData{}
+					err = s.Session.DB(db_name).Run(bson.D{
+						{
+							Name:  "collStats",
+							Value: col_name,
+						},
+					}, col_stat_line)
+					if err != nil {
+						log.Println("E! Error getting col stats from " + col_name + "(" + err.Error() + ")")
```
oboukili: Should fmt.Errorf be used here to return an error and prevent further execution?

glinton: I think maybe a `continue` so it continues adding other (likely valid) colls

danielnelson: 17465b0aaf2489eac8ec1a5a9b72ed32f7d2ebf1

danielnelson: Yeah, also we need to do something on 155.  Probably should move this into a function.

PR #6139, opened 2019-07-18
Add documentation regarding collecting invalid certs

### LINKED ISSUE #6066: x509_cert: Error when site SSL is not valid
Relevant telegraf.conf:
[[inputs.x509_cert]]
sources = ["https://www.testsite.com:443""]

Additional info: 
The SSL of the site is currently expired, and I am trying to set up monitoring for that URL.

Expected behavior:
Telegraf exposes SSL metrics for www.google.com, and indicated that it is expired (i.e. expiry as a negative value). That way alerts can be set up for certificates that are already expired, and no monitored sites are lost out of sight when the dashboard time frame moves.  

Actual behavior:
Because www.testsite.com is expired, the pluging throws an error "[inputs.x509_cert]: Error in plugin: cannot get SSL cert 'https://testsite.com:443': x509: certificate has expired or is not yet valid"

glinton:
Resolves #6066

I wonder if a better solution would be to always enable `insecure_skip_verify`. The tests set it by default.

### DIFF
--- a/plugins/inputs/x509_cert/README.md
+++ b/plugins/inputs/x509_cert/README.md
@@ -3,6 +3,8 @@
 This plugin provides information about X509 certificate accessible via local
 file or network connection.
 
+In order to always fetch cert information, it is suggested that you use `insecure_skip_verify = true` as telegraf fails to collect information on invalid certs without it.
+
 
 ### Configuration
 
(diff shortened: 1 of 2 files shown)


danielnelson:
The one thing having `insecure_skip_verify = false` does for us now is give some sort of indication if the certificate is invalid.  Unfortunately it's not a positive claim, the metric just stops being emitted.

Originally we wanted to report if the cert was valid but we removed it because our check could report valid for some invalid certs.  If we research the right arguments to `x509.Verify(` I think we could potentially add validity information and remove this option.

One more thing, I notice that even with `insecure_skip_verify = true` you can't get a self signed certificate over HTTP (assuming you don't have the CA cert for it).

glinton:
Preferring 6143 over this

PR #6143, opened 2019-07-20
Add certificate verification status to x509_cert input

### LINKED ISSUE #4877: Add certificate verification status to the x509_cert input
### Proposal:

Add certificate verification status to the x509_cert input.

### Current behavior:

Certificate validation is not reported, some aspects can be computed at query time but some checks are not performed.

### Desired behavior:

Validate the certificate and report the status, potentially using https://golang.org/pkg/crypto/x509/#Certificate.Verify.

This could be reported as a tag `valid=true` or `validation=success` and an integer field `validation_code=0i`.

### Use case:

Needed to detect client issues that could occur using the certificate.

glinton:
Resolves #4877 

This always sets `insecure_skip_verify` as it verifies the cert once it's been collected/parsed.

To verify a self-signed cert (remote endpoint or file), set the `tls_ca` to the self-signing ca, or the self-signed cert itself for validation to be successful.

Alternate to #6139

### DIFF
--- a/plugins/inputs/x509_cert/README.md
+++ b/plugins/inputs/x509_cert/README.md
@@ -19,9 +19,6 @@ file or network connection.
   # tls_ca = "/etc/telegraf/ca.pem"
   # tls_cert = "/etc/telegraf/cert.pem"
   # tls_key = "/etc/telegraf/key.pem"
-
-  ## Use TLS but skip chain & host verification
-  # insecure_skip_verify = false
 ```
 
 
@@ -35,7 +32,10 @@ file or network connection.
     - country
     - province
     - locality
+    - validation
   - fields:
+    - validation (int)
+    - validation_error (string)
     - expiry (int, seconds)
     - age (int, seconds)
     - startdate (int, seconds)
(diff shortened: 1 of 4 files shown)


plugins/inputs/x509_cert/x509_cert.go
```diff
@@ -164,15 +166,38 @@ func (c *X509Cert) Gather(acc telegraf.Accumulator) error {
...
+				opts.DNSName = u.Hostname()
+			}
+			if tlsCfg.RootCAs != nil {
+				opts.Roots = tlsCfg.RootCAs
+			}
+
+			_, err = cert.Verify(opts)
+			if err == nil {
+				tags["validation"] = "success"
+				fields["validation"] = 0
```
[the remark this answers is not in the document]
glinton: I do like `verification` and `verification_code`

plugins/inputs/x509_cert/x509_cert.go
```diff
@@ -164,22 +159,62 @@ func (c *X509Cert) Gather(acc telegraf.Accumulator) error {
...
+			}
+
+			_, err = cert.Verify(opts)
+			if err == nil {
+				tags["verification"] = "valid"
+				fields["verification"] = 0
+			} else {
+				tags["verification"] = "invalid"
+				fields["verification"] = 1
+				fields["validation_error"] = err.Error()
```
danielnelson: `verification_error`

plugins/inputs/x509_cert/README.md
```diff
@@ -35,7 +32,10 @@ file or network connection.
     - country
     - province
     - locality
+    - validation
   - fields:
+    - validation (int)
+    - validation_error (string)
```
danielnelson: Needs updated

Issue #6138, opened 2019-07-18
json_query not parsing through all entries in the JSON array


harnitsignalfx:
Hi, I think there is a bug in how JSON array entries are being parsed by JSONQuery. Would really appreciate guidance on whether I've made a configuration error or this is indeed reproducible by other folks.

### Relevant telegraf.conf:

```
[[inputs.tail]]
  files = ["/test.log"]
  watch_method = "poll"
  data_format = "json"
  name_override = "test-key"
  tag_keys = ["first"]
  json_query = "friends"
```

### System info:

Telegraf v 1.11.2
OS: Debian 9

### Steps to reproduce:

1. Follow the example for JSON Query - 
https://github.com/influxdata/telegraf/tree/master/plugins/parsers/json#query
So a sample entry in the file would be - 
```
{"friends": [{"first": "Dale", "last": "Murphy", "age": 44},{"first": "Roger", "last": "Craig", "age": 68},{"first": "Jane", "last": "Murphy", "age": 47}]}
```

### Expected behavior:
All entries in the array are parsed (very similar to the posted example)
```
{"fields":{"age":44},"name":"test-key","tags":{"first":"Dale","host":"test-us1-instance","path":"/test.log"},"timestamp":1563466236}
{"fields":{"age":68},"name":"test-key","tags":{"first":"Roger","host":"test-us1-instance","path":"/test.log"},"timestamp":1563466236}
{"fields":{"age":47},"name":"test-key","tags":{"first":"Jane","host":"test-us1-instance","path":"/test.log"},"timestamp":1563466236}
```

### Actual behavior:
Only the first array entry is being parsed.
```{"fields":{"age":44},"name":"test-key","tags":{"first":"Dale","host":"test-us1-instance","path":"/test.log"},"timestamp":1563466236}```

danielnelson:
This is a bug in the tail plugin caused by some special logic for the csv parser when reading the first line of the file.  I'll put together a fix for the next release.

harnitsignalfx:
Appreciate your looking into it. Thanks Daniel!

GeorgeMac:
@danielnelson been playing about with tests on this one and come to realise that there is a little interface design discrepancy going on between the tail plugin, the parser interface and the json implementation.

The parser interface assumers 1 metric per line:
https://github.com/influxdata/telegraf/blob/877c42362d9d3c8264add86a727af7b10f01c571/plugins/parsers/registry.go#L38-L58

[...]

PR #6160, opened 2019-07-23
Allow stateful parsers

GeorgeMac:
closes: https://github.com/influxdata/telegraf/issues/6138

This is a work in progress.

It updates the `tail` plugin to only use the `parser.Parse` function instead of `parser.ParseLine`. This is a standard we would like to obvserve in all plugins going forward. Deprecating the ParseLine method. As per discussion in the above issue.

This brings with it other changes in the form of:

- `plugin.SetParserFunc` to be standard in place of `plugin.SetParser` allowing plugins to acknowledge when new parses are required. In the event that a parse may be stateful (e.g. csv).
- Update to `file` input plugin to call parser func per new file to be read. Rather than use same parser for each file.

TODO:

- [ ] better understand implications of stateful parsers for other parser consumers (other than just `file` and `tail`).

### Required for all PRs:

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -980,17 +980,6 @@ func (c *Config) addInput(name string, table *ast.Table) error {
 	}
 	input := creator()
 
-	// If the input has a SetParser function, then this means it can accept
-	// arbitrary types of input, so build the parser and set it.
-	switch t := input.(type) {
-	case parsers.ParserInput:
-		parser, err := buildParser(name, table)
-		if err != nil {
-			return err
-		}
-		t.SetParser(parser)
-	}
-
 	switch t := input.(type) {
 	case parsers.ParserFuncInput:
 		config, err := getParserConfig(name, table)
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -151,7 +151,7 @@ func TestConfig_LoadDirectory(t *testing.T) {
 		DataFormat: "json",
 	})
 	assert.NoError(t, err)
-	ex.SetParser(p)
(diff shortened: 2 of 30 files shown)


danielnelson:
Looking good, however I am starting to become nervous about how expensive creating a new parser can be.  To preserve behavior, in all plugins other than tail we would need to recreate the parser on each call to Parse(), and in ServiceInput plugins especially the `*_consumer` plugins we could be calling Parse thousands of times per second.

We ought to look closer into the performance difference and if it's going to be noticeable try to improve it.  I think we can rework the factory code in `plugins/parsers/registry.go` and get some large performance gains.

GeorgeMac:
@danielnelson I see the concern. I am just wondering if we need to address all of it in one go.

This PR does move all plugins to `SetParserFunc`. However, the majority of them call it immediately and set the resulting parser forever. Giving most existing plugins exactly the same behavior.

The one difference (for now) being the `tail` + `file` plugins. Which now produce one parser per file they consume.

I am conerned that existing things that interact with the csv parser may break due to the slight adjustment in behavior.

GeorgeMac:
Ahh I see one problem. SetParserFunc needs to take a `parsers.ParseFunc` which can return an error. 
Makes sense. I missed that. Will re-assess.

srebhan:
@GeorgeMac: Any progress on this?
To solve the performance issue, we might add a `NewInstance()` function to the parser, letting the parser itself decide if it is stateful (while CSV is, JSON is not) and either return a new instance of the parser or create a singleton and return that.

srebhan:
@GeorgeMac any news on this PR?

powersj:
Hi,

Thanks!

Issue #2280, opened 2017-01-17
POST request to http_listener fails with EOF 


salarali:
## Bug report
Intermittently the post request to the http_listener fails and I get the following error message:
`Post http://localhost:8186/write: EOF`

### Relevant telegraf.conf:

Vanilla conf. with http_listener enabled.

sparrc:
what version are you using? do you have any steps to reproduce?

salarali:
Sorry for the lack of information.

Telegraf version: 1.1.2

I am making the http post request from a golang program.
```
resp, err := http.Post(API_URL, "", bytes.NewBuffer([]byte(metricsString)))

if err != nil || resp.StatusCode != 204 {
	logger.Error("Telegraf push failed: ", err)
} else {
	defer resp.Body.Close()
}
```

metricsString looks like this:
```
test num_error_no_category=122 1484680136892545372
test num_error_http_setup=0 1484680136892545372
test num_active_go_routines=2045 1484680136892545372
test num_error=1803 1484680136892545372
test num_duplicate=760 1484680136892545372

```

And this is all the log that go provides to me. Is there a way I can get more information?

I have turned debug logging on for telegraf but I do not see any errors there.

danielnelson:
@salarali Any chance you have found a way to reproduce this?

ncode:
I am receiving the same error, but it only happens with 21+ Goroutines pushing metrics to telegraf. Also It never happens if you use the discard as output.

danielnelson:
Possibly related to #2919?

jpereira:
I am experiencing the same issue in my setup.

```
2017-09-29T17:53:01Z E! InfluxDB Output Error: Post https://metrics.internal.lan:443/write?db=telegraf&rp=default: EOF
```

danielnelson:
@jpereira Just to clarify, is the error when you are sending to the http_listener in Telegraf, or when you are sending to InfluxDB?

jpereira:
@danielegozzi  Sending to the InfluxDB Cluster.

danielnelson:
Okay, this is a different issue.  It indicates that InfluxDB closed the connection, I would check the InfluxDB logs to see if anything is reported.

jpereira:
@danielegozzi I am using the latest version 1.4.1

ILoGM:
We figured out that we exceeded max_line_size. 
I set it as twice as default value and disconnects disappeared.
[...]

Issue #6124, opened 2019-07-16
Using influxdb proxy (http_listener, influxdb_listener, http_listener_v2) should write the same points as direct write to InfluxDB


theodiefenthal:
### System info:
Official telegraf 1.11 docker image, relevant to all telegraf versions in all OS.

### Steps to reproduce:

1. Write a batch of points in one request to Telegraf containing some invalid points. 

### Expected behavior:

If I send a batch of points to InfluxDB with some invalid points, I receive a partial write error. Nonetheless, the valid points in that batch are stored.

### Actual behavior:

If I send the same batch of points to http_listener, influxdb_listener or http_listener_v2, letting it serve as a proxy to my influxdb instance, none of the batch points are stored in InfluxDB. Telegraf will respond with a "bad request" error and discard the entire batch before even forwarding it to InfluxDB. 

### Additional info:

This issue is slightly related to https://github.com/influxdata/telegraf/issues/4742 but with much more impact. It's not just about output / error messages. We observe a totally different behavior in terms of what is actually stored into InfluxDB when just adding a proxy into the connection. 

### Our scenario:

We run Apache Flink and Apache Spark applications on a cluster which can communicate to our InfluxDB backend storage only via a dedicated proxy node. The applications produce tons of metrics, some containing the value +-infinity, which can't be handled neither by InfluxDB nor by telegraf. When sending metrics directly to InfluxDB in the test environment, all works fine (only the invalid metrics are just not written), when using telegraf as a proxy, we don't get any metrics reported.

danielnelson:
Thanks for the bug report, I took a quick look at the code the current behavior of `influxdb_listener` is to parse lines until the first error and then stop, whereas InfluxDB parses all lines skipping the unparseable lines in full.  The `http_listener_v2` plugin is not meant to behave the same as InfluxDB, so let's ignore it here.

To summarize, given the input:
```
cpu,host=a value1=1
cpu,host=b value1=1,value2=+Inf,value3=3
cpu,host=c value1=1
```

Actual output
```
cpu,host=a value1=1
```

Expected output
```
cpu,host=a value1=1
cpu,host=c value1=1
```

PR #6185, opened 2019-07-30
Mirror influxdb api for large payloads in influx_listener input

glinton:
Addresses comments in #2280

An attempt was made to add a `content-length` header to the influxdb output, but it was decided that would cost more than it was worth (read the body to count bytes, then allocate a new reader to use in the request)


danielnelson:
The old code is too tricky, I don't feel comfortable that it works.  We also do a bunch of work so that we can apply a line length limit, a feature that InfluxDB doesn't even appear to have.

Let's remove the ability to limit lines, accepting the option but warning when it is set, and only retain max body limit.  This will allow us to pass the full body into the parser and we can use the much simpler processing.

This will be needed anyway to solve #6124.

glinton:
Closing in favor of [http_listener rewrite referred to here](https://github.com/influxdata/telegraf/issues/2280#issuecomment-518332139)

PR #6186, opened 2019-07-30
Update influxdb_listener to allow partial writes

### LINKED ISSUE #6124: Using influxdb proxy (http_listener, influxdb_listener, http_listener_v2) should write the same points as direct write to InfluxDB

glinton:
Resolves #6124 

Previously, if a batch contained an invalid metric, the batch failed. This change allows for valid metrics to be processed and a "partial write" error to be sent to the client.

### DIFF
--- a/plugins/inputs/influxdb_listener/http_listener.go
+++ b/plugins/inputs/influxdb_listener/http_listener.go
@@ -11,6 +11,7 @@ import (
 	"log"
 	"net"
 	"net/http"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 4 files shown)


reimda:
Continued on #6974

Issue #6162, opened 2019-07-23
Allow Github Enterprise repos


Esity:
## Feature Request
Allow Telegraf to poll Github Enterprise repos. Currently it would appear from the documentation you can only enter repos from github.com

### Proposal:
Add URL variable for the github input plugin so you grab metrics from on prem github

Input Plugin [github](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/github)

shanempope:
@Esity something like this: https://github.com/influxdata/telegraf/pull/6194

danielnelson:
Closed in #6194

PR #6194, opened 2019-08-02
Add support for Enterprise repos in Github plugin

shanempope:
### Required for all PRs:

https://github.com/influxdata/telegraf/issues/6162

### DIFF
--- a/plugins/inputs/github/github.go
+++ b/plugins/inputs/github/github.go
@@ -18,12 +18,11 @@ import (
 
 // GitHub - plugin main structure
 type GitHub struct {
-	Repositories        []string          `toml:"repositories"`
-	AccessToken         string            `toml:"access_token"`
-	EnterpriseBaseURL   string            `toml:"enterprise_base_url"`
-	EnterpriseUploadURL string            `toml:"enterprise_upload_url"`
-	HTTPTimeout         internal.Duration `toml:"http_timeout"`
-	githubClient        *github.Client
+	Repositories      []string          `toml:"repositories"`
+	AccessToken       string            `toml:"access_token"`
+	EnterpriseBaseURL string            `toml:"enterprise_base_url"`
+	HTTPTimeout       internal.Duration `toml:"http_timeout"`
+	githubClient      *github.Client
 
 	obfusticatedToken string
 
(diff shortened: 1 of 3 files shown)


plugins/inputs/github/github.go
```diff
@@ -71,9 +77,16 @@ func (g *GitHub) createGitHubClient(ctx context.Context) (*github.Client, error)
...
 		g.obfusticatedToken = g.AccessToken[0:4] + "..." + g.AccessToken[len(g.AccessToken)-3:]
 
-		return github.NewClient(oauthClient), nil
+		return g.newGithubClient(oauthClient)
 	}
 
+	return g.newGithubClient(httpClient)
+}
+
+func (g *GitHub) newGithubClient(httpClient *http.Client) (*github.Client, error) {
```
shanempope: I do wonder if there's a better function name for this...

plugins/inputs/github/github.go
```diff
@@ -71,9 +77,16 @@ func (g *GitHub) createGitHubClient(ctx context.Context) (*github.Client, error)
...
+	if g.EnterpriseBaseURL != "" && g.EnterpriseUploadURL != "" {
```
shanempope:
```suggestion
	if g.EnterpriseBaseURL != "" {
```
I think it's actually ok for me to allow EnterpriseUploadURL to be `""`. It works fine (tested it and by looking at the code). I don't think the UploadURL will be necessary for these stats anyway. It's basically only used for `NewUploadRequest` as far as I can tell.

plugins/inputs/github/github.go
```diff
@@ -71,9 +75,16 @@ func (g *GitHub) createGitHubClient(ctx context.Context) (*github.Client, error)
...
-		return github.NewClient(oauthClient), nil
+		return g.newGithubClient(oauthClient)
 	}
 
+	return g.newGithubClient(httpClient)
+}
+
+func (g *GitHub) newGithubClient(httpClient *http.Client) (*github.Client, error) {
+	if g.EnterpriseBaseURL != "" {
+		return github.NewEnterpriseClient(g.EnterpriseBaseURL, "", httpClient)
```
shanempope: the `""` in the 2nd arg is the Enterprise Upload URL that isn't necessary for the plugin. You can compare this commit to my first commit, but I checked `go-github` and as far as I can tell it isn't necessary for this plugin.

PR #6195, opened 2019-08-02
Added TLS & credentials configuration for NATS_consumer input plugin

mmelnyk:
### Required for all PRs:

Basic changes to add missing TLS & credentials options for NATS. Mutual TLS is supported too.

### DIFF
--- a/plugins/inputs/nats_consumer/README.md
+++ b/plugins/inputs/nats_consumer/README.md
@@ -12,12 +12,21 @@ instances of telegraf can read from a NATS cluster in parallel.
 [[inputs.nats_consumer]]
   ## urls of NATS servers
   servers = ["nats://localhost:4222"]
-  ## Use Transport Layer Security
-  secure = false
   ## subject(s) to consume
   subjects = ["telegraf"]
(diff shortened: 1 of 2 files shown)


danielnelson:
I've got thinking about this a bit more and realize that we still need the `secure` option, otherwise it isn't possible to connect with a default tls.Config.  Normally it would be controlled by the URL in a HTTP input (http vs https), or by using a `enable_tls` option, but since we already have `secure` I added it back in ffe9494663d0b88838c6a88131d70294b27c259d.

mmelnyk:
There are some inconsistencies in options for other plugins - e.g. nats output (btw, readme for this plugin needs updating) does not have "secure" option, but usually used in a pair with nats_consumer.

danielnelson:
I added the same option to the output too now, technically this is a breaking change as you now must set the `secure` option to use TLS, it's not enough to just use non default settings.

When it comes to plugin inconsistency, yeah it's quite difficult to keep up, especially considering the number of plugins these days.  Definitely relying on the community to open PRs.

PR #6200, opened 2019-08-03
logsfs available bytes fix

robcowart:
Fixed misspelled field name `logsfs_available_bytes`


danielnelson:
Previous spelling in this case was bug like enough that I feel comfortable making the change.

PR #6202, opened 2019-08-05
fix broken link at readme

alexus1024:
### Required for all PRs:

- [ ] Signed [CLA](https://influxdata.com/community/cla/). - hmm...
- [ ] Associated README.md updated. -  %)
- [ ] Has appropriate unit tests.  - It`s just readme, guys...


danielnelson:
This we could accept without signing the CLA, but I think we should use this url instead, since this is the link provided by the project and I'm not sure if Mozilla was planning to host those docs for 3rd parties:
```
https://raw.githack.com/rabbitmq/rabbitmq-management/rabbitmq_v3_6_9/priv/www/api/index.html
```

I will make the change, thanks for letting us know.

PR #6203, opened 2019-08-05
Added a timestamp processor. This adds the metric time as a unix nano…

mmriis:
This processor adds a unix ns timestamp to metrics.

I needed this to create custom syslog entries and have them presented in Chronograf.

It's very, very simple. But I cannot figure out another way to do this - ideally better log support in the stack would be ideal.

Alternatively a more generic plugin that can add timestamps based on a date field could be written. The "date" processor only allows output in formatted dates which do not include unix time.

### Required for all PRs:

### DIFF
--- a/plugins/processors/timestamp/README.md
+++ b/plugins/processors/timestamp/README.md
@@ -1,0 +1,20 @@
+# Timestamp Processor Plugin
+
+Use the `timestamp` processor to add a unix nano timestamp to the metric.
+
+This can be used to mimic logs from syslog that you'd want to display in Chronograf.
+
+### Configuration
+
+```toml
+[[processors.timestamp]]
+  ## New tag to create
+  field_key = "timestamp"
+```
+
+### Example
+
+```diff
+- syslog,appname=myapp,facility=user,hostname=test,severity=notice message="notice msg",severity_code=5i,version=1i,facility_code=1i 1564997347582799644
++ syslog,appname=myapp,facility=user,hostname=test,severity=notice message="notice msg",severity_code=5i,version=1i,facility_code=1i,timestamp=1564997347582799644i 1564997347582799644
+```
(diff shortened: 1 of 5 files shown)


plugins/processors/timestamp/README.md
```diff
@@ -0,0 +1,20 @@
...
+[[processors.timestamp]]
+  ## New tag to create
+  field_key = "timestamp"
+```
+
+### Example
+
+```diff
+- syslog,appname=myapp,facility=user,hostname=test,severity=notice message="notice msg",severity_code=5i,version=1i,facility_code=1i 1564997347582799644
++ syslog,appname=myapp,facility=user,hostname=test,severity=notice message="notice msg",severity_code=5i,version=1i,facility_code=1i,timestamp=1564997347582799644i 1564997347582799644
```
glinton: I'm not sure I understand the use case or value of adding a tag with the timestamp of the metric. Were you able to use a parser with the syslog input?

danielnelson: I believe this is for compatibility with the Chronograf Log Viewer, I just checked and without the timestamp field it is reporting the log at the unix epoch.  @mmriis Is this what you are trying to work around?

mmriis: @danielnelson exactly...

Maybe the better approach would be a PR to Chronograf which defaults to using the metric time if a timestamp tag isn't present. But TS isn't my forte...

glinton: >Maybe the better approach would be a PR to Chronograf which defaults to using the metric time if a timestamp tag isn't present.

I would agree as that does seem to be where the issue really lies. If not a PR, at least a bug report linking to this maybe?

danielnelson:
Yeah, this is a good workaround for now but I don't think we will merge it.  Can you open a new issue against Chronograf though, I think it should fallback to the metric timestamp if there isn't a field named timestamp.

PR #6207, opened 2019-08-05
Add ability to label inputs for logging

### LINKED ISSUE #1815: Naming plugins & improve log verbosity
The agent currently throws an error log with the plugin's name in case of a collection interval timeout.

When using multiple input plugins of the same type (e.g. multiple inputs.snmp instances which is a usual case) more verbosity is required to determine which one times out.

https://github.com/influxdata/telegraf/blob/master/agent/agent.go#L174

Since each plugin has different configuration structure they can't all handled the same way for verbosity.

A workaround would be to support a `user_comment` configuration option for each plugin.

The user_comment message would be thrown at timeout (or by other plugin-level errors as well) to help the user identify the problematic plugin.

If there is no user_comment defined, this section would be omitted.

E.g.

```
[[inputs.snmp]]
  user_comment = "localhost_polling"
  agents = [ "127.0.0.1:161" ]
  version = 2
  community = "public"

  name = "system"
  [[inputs.snmp.field]]
    name = "hostname"
    oid = ".1.0.0.1.1"
    is_tag = true
```

`ERROR: input snmp: localhost_polling: took longer to collect than collection interval 10s`

glinton:
Resolves #1815 

### Required for all PRs:


internal/models/log.go
```diff
@@ -0,0 +1,86 @@
...
+import (
+	"log"
+	"strings"
+
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/selfstat"
+)
+
+// PluginConfig contains individualized plugin configuration.
+type PluginConfig struct {
```
danielnelson: Won't need this anymore if you switch the telegraf.PluginConfig to a type struct.

internal/models/log.go
```diff
@@ -0,0 +1,86 @@
...
+		iErrors.Incr(1)
+	case strings.HasPrefix(l.Name, "output"):
+		oErrors.Incr(1)
+	case strings.HasPrefix(l.Name, "processor"):
+		pErrors.Incr(1)
+	}
+}
+
+var (
+	aErrors = selfstat.Register("agent", "aggregator_errors", map[string]string{})
```
danielnelson: Have the selfstat passed in when the Logger is created, then no need to switch on the name.

Change these over to be like:
```
internal_gather,input=exec,alias=foo errors=1i
internal_write,output=foo,alias=bar errors=1i

# new measurements
internal_process,processor=foo,alias=bar errors=1i
internal_aggregate,aggregator=foo,alias=bar errors=1i
```

Also go ahead and add the alias to the existing selfstats.

Probably should switch the functions to pointer receiver, since we have a bit of state and it's not really a value type.

agent/agent.go
```diff
@@ -659,7 +659,7 @@ func (a *Agent) connectOutputs(ctx context.Context) error {
 				return err
 			}
 		}
-		log.Printf("D! [agent] Successfully connected to output: %s\n", output.Name)
+		log.Printf("D! [agent] Successfully connected to %s", output.LogName())
```
danielnelson: We are a little inconsistent with `[%s]` vs `%s` and LogName, it's a bigger issue to really fixup the log messages but for now lets go with `%s` with no brackets.  I feel like `[agent] [inputs.exec]` is too brackety for my taste.  I also want to get people away from thinking of the plugin as `[inputs.foo]` as it causes some confusion with the toml.

internal/models/running_output.go
```diff
@@ -195,15 +206,15 @@ func (ro *RunningOutput) WriteBatch() error {
...
+		log.Printf("E! [%s] Error closing output: %v", ro.LogName(), err)
 	}
 }
 
 func (ro *RunningOutput) write(metrics []telegraf.Metric) error {
 	dropped := atomic.LoadInt64(&ro.droppedMetrics)
 	if dropped > 0 {
-		log.Printf("W! [outputs.%s] Metric buffer overflow; %d metrics have been dropped",
-			ro.Name, dropped)
+		log.Printf("W! [%s] Metric buffer overflow; %d metrics have been dropped",
```
danielnelson: Use the plugin's logger in the RunningFoo's

plugins/inputs/exec/exec_test.go
```diff
@@ -97,6 +98,7 @@ func TestExec(t *testing.T) {
 		MetricName: "exec",
 	})
 	e := &Exec{
+		log:      models.Logger{},
```
danielnelson: Let's make a testutil.Logger, I don't want the models package used directly in any plugins or tests.  We could even make model.Logger unexported, but I care less about this so long as there are no examples to be copy/pasted.

plugins/outputs/influxdb/http.go
```diff
@@ -173,6 +175,7 @@ func NewHTTPClient(config HTTPConfig) (*httpClient, error) {
 		},
 		createdDatabases: make(map[string]bool),
 		config:           config,
+		log:              models.Logger{Name: userAgent},
```
danielnelson: Don't instantiate Logger here.  Maybe some leftover code from development?

plugins/outputs/influxdb/http.go
```diff
@@ -182,6 +185,10 @@ func (c *httpClient) URL() string {
 	return c.config.URL.String()
 }
 
+func (c *httpClient) SetLogger(log telegraf.Logger) {
```
danielnelson: remove

plugins/outputs/influxdb/influxdb.go
```diff
@@ -167,13 +169,17 @@ func (i *InfluxDB) Connect() error {
 				return err
 			}
 
+			c.SetLogger(i.log)
```
danielnelson: Just pass it over with the rest of the config?

plugins/outputs/influxdb/influxdb.go
```diff
@@ -294,6 +306,7 @@ func init() {
 			CreateUDPClientF: func(config *UDPConfig) (Client, error) {
 				return NewUDPClient(*config)
 			},
+			log: models.Logger{Name: "outputs.influxdb"},
```
danielnelson: remove

plugins/inputs/exec/exec.go
```diff
@@ -145,6 +147,7 @@ func (e *Exec) ProcessCommand(command string, acc telegraf.Accumulator, wg *sync
 	defer wg.Done()
 	_, isNagios := e.parser.(*nagios.NagiosParser)
 
+	e.log.Debugf("command %#v", command)
```
danielnelson: remove

agent/accumulator.go
```diff
@@ -111,7 +112,7 @@ func (ac *accumulator) AddError(err error) {
 		return
 	}
 	NErrors.Incr(1)
-	log.Printf("E! [%s]: Error in plugin: %v", ac.maker.Name(), err)
+	log.Printf("E! [%s]: Error in plugin: %v", ac.maker.LogName(), err)
```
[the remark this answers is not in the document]
glinton: Perhaps I'm missing something, but I'm not seeing how this can be pulled out of metricmaker as several plugins call `acc.AddError` directly, if we remove `LogName` from metricmaker, those logs won't use the plugin alias.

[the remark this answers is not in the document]
danielnelson: Yeah, how about instead of RunningInput having a `LogName()` function it has a `Logger() telegraf.Logger` function.  We then pass in a Logger to the Accumulator in agent/agent.go.

agent/agent.go
glinton: Not sure what the recommendation is here. To me, the bracketed thing is the part of telegraf doing the logging (allowing to later filter on the source of the logs (agent, inputs, outputs, etc...)) which is why when one part, in this case agent, logs something about a plugin, that plugin name/alias is not bracketed.

internal/models/running_output.go
glinton: :grey_question:

plugins/outputs/influxdb/http.go
glinton: Not leftover, the httpclient logs things, so this was intentionally added so it could.

plugins/outputs/influxdb/influxdb.go
glinton: You mean in the call to `i.udpClient(parts, i.log)` and `i.httpClient(ctx, parts, proxy, i.log)`?

agent/accumulator.go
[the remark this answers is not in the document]
danielnelson: Still need to do this, also need to try what we talked about with the reflection method for injecting the Logger.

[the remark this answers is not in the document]
danielnelson: Oh wait I see the reflection stuff, so we just need to pull PluginConfig out.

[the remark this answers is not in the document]
glinton: Oops, looks like I did forget to pull out PluginConfig from the plugin readme's

PR #6212, opened 2019-08-06
Update gopsutil

### LINKED ISSUE #6103: [linux] inputs.mem: SlabReclaimable and SlabUnreclaim
## Feature Request

On linux slab is subdivided into two: SReclaimable and SUnreclaim.
It is interesting to know these two values.

### Proposal:

Add SReclaimable and SUnreclaim values in addition to Slab to inputs.mem

### Use case:
SReclaimable can be freed with /proc/sys/vm/drop_caches.

danielnelson:
closes #6103

### Required for all PRs:

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -1348,7 +1348,7 @@
 
 [[projects]]
   branch = "master"
-  digest = "1:6a6eed3727d0e15703d9e930d8dbe333bea09eda309d75a015d3c6dc4e5c92a6"
+  digest = "1:0b5c2207c72f2d13995040f176feb6e3f453d6b01af2b9d57df76b05ded2e926"
   name = "golang.org/x/sys"
   packages = [
     "unix",
@@ -1360,7 +1360,7 @@
     "windows/svc/mgr",
   ]
   pruneopts = ""
-  revision = "7c4c994c65f702f41ed7d6620a2cb34107576a77"
+  revision = "51ab0e2deafac1f46c46ad59cf0921be2f180c3d"
   source = "https://github.com/golang/sys.git"
 
 [[projects]]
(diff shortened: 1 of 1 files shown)


adrianlzt:
This PR updates ``Gopkg.lock`` but does not update ``Gopkg.toml``.
Is that an error?

Currently in master:

Gopkg.toml
```
[[constraint]]
  name = "github.com/shirou/gopsutil"
  version = "2.18.12"
```

Gopkg.lock
```
[[projects]]
  digest = "1:55dcddb2ba6ab25098ee6b96f176f39305f1fde7ea3d138e7e10bb64a5bf45be"
  name = "github.com/shirou/gopsutil"
  packages = [
    "cpu",
    "disk",
    "host",
    "internal/common",
    "load",
    "mem",
    "net",
    "process",
  ]
  pruneopts = ""
  revision = "e4ec7b275ada47ca32799106c2dba142d96aaf93"
  version = "v2.19.8"
```

adrianlzt:
An important change also coming with this update is a change in disk device names in [this commit](https://github.com/shirou/gopsutil/commit/d7405fd87376efc3079c865d3b9977a12ee193f6).

Previously:
```
> disk,device=mapper/vg00-lvusr,path=/usr ...
```

After the new gopsutil version:
```
> disk,device=dm-2,path=/usr
```

danielnelson:
> This PR updates Gopkg.lock but does not update Gopkg.toml.
Is that an error?

In this case yes, we need to update the toml to the version of gopsutil that has the SUnreclaim field.  I'm fixing this in #6716.

> An important change also coming with this update is a change in disk device names in this commit.

Thanks for the heads up, I'm still looking into this but we may want to use `func (s *DiskIO) diskName(devName string)` like code to deal with this.

danielnelson:
I think since the disk device name change was included in 1.12.0 and we are getting ready to release 1.13, it will be best at this point to not attempt to work around the change anymore.

PR #6214, opened 2019-08-06
Get the number of logcal CPUs dynamically

### LINKED ISSUE #5451: restart needed for system.ncpus to update

adrianlzt:
runtime.NumCPU() return the number of logical CPUs at the startup of the
process, https://golang.org/pkg/runtime/#NumCPU.
gopsutil return the current number of CPUs, giving the correct number
when CPU hot un/plug are being done.

closes #5451

### Required for all PRs:

### DIFF
--- a/plugins/inputs/system/README.md
+++ b/plugins/inputs/system/README.md
@@ -3,6 +3,8 @@
 The system plugin gathers general stats on system load, uptime,
 and number of users logged in. It is similar to the unix `uptime` command.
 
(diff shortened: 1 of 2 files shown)


alvarolmedo:
+1

PR #6225, opened 2019-08-07
Add note about force-pushing after a review has begun

glinton:
Address a personal pet peeve

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -61,6 +61,9 @@ make test-all
 
 Use `make docker-kill` to stop the containers.
 
+**Addressing Feedback:**
+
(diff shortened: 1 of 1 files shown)


danielnelson:
I don't feel comfortable asking this until we can be more timely with our reviews.  Currently a contributor may need to keep their branch open for quite some time and it may be easier to do this with a rebase.

glinton:
ok, makes sense

danielnelson:
> Maybe we shouldn't review until they update their branch if there are conflicts to avoid the above scenario.

Conflicts can occur anytime we make another change, its not possible to wait and avoid them unless we do everything one at a time.  I can think of many reasons why force pushes are the right thing to do.

> Re-reviewing entire prs when only one change was requested takes time we don't have.

Them's the breaks

Issue #6240, opened 2019-08-12
SMART plugin: measurement dropped when HDD under heavy use.


marcv81:
### Relevant telegraf.conf:

```toml
[[inputs.smart]]
  use_sudo = true
  attributes = true
```

### System info:

Ubuntu 18.04 LTS

    $ uname -a
    Linux desktop 4.15.0-55-generic #60-Ubuntu SMP Tue Jul 2 18:22:20 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux

Telegraf version: latest master.

### Steps to reproduce:

1. Enable the SMART input plugin.
2. Collect metrics.
3. Load the HDD: `time sh -c "dd if=/dev/zero of=test bs=1M count=3k && sync"` (should take around 30s on a modern HDD).

### Expected behavior:
Telegraf captures SMART measurements for the supported block devices.

### Actual behavior:
Telegraf does not capture any SMART measurements when the HDD is under load.

### Additional info:
This is due to a hardcoded 5s timeout for `smartctl`. It is possible to verify that under those circumstances `smartctl` takes more than 5s even outside of Telegraf. Verified on 2 different computers with different HDDs.

PR #6241, opened 2019-08-12
Made the SMART timeout a parameter, and set the default to 30s

marcv81:
Fix for https://github.com/influxdata/telegraf/issues/6240.

This change makes the SMART timeout a configuration parameter, and bumps the default value from 5s to 30s.

- Previous behavior: measurements are dropped when the HDD is under heavy load.
- Current behavior: measurements are delayed (5-10s on my system) but not dropped.

### Required for all PRs:

### DIFF
--- a/plugins/inputs/smart/smart.go
+++ b/plugins/inputs/smart/smart.go
@@ -446,14 +446,13 @@ func parseTemperature(fields, deviceFields map[string]interface{}, str string) e
 }
 
 func init() {
-	m := NewSmart()
-	path, _ := exec.LookPath("smartctl")
-	if len(path) > 0 {
-		m.Path = path
-	}
-	m.Nocheck = "standby"
-
 	inputs.Add("smart", func() telegraf.Input {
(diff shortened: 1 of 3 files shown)


plugins/inputs/smart/smart.go
```diff
@@ -436,14 +446,14 @@ func parseTemperature(fields, deviceFields map[string]interface{}, str string) e
...
+	m := NewSmart()
 	path, _ := exec.LookPath("smartctl")
 	if len(path) > 0 {
 		m.Path = path
 	}
 	m.Nocheck = "standby"
 
 	inputs.Add("smart", func() telegraf.Input {
-		return &m
+		return m
```
danielnelson: This was probably a bug before, but we need this function to always return a new instance for it to work correctly if the plugin is specified multiple times.

marcv81: Thanks Daniel. How does it look now?

PR #6246, opened 2019-08-12
Add possibility to setting up tls security levels

### LINKED ISSUE #5914: Make the TLS listening version configurable.
## Feature Request

Configure the TLS listening at the moment its not possible to listing only on TLS 1.2 or TLS 1.3.
Also it would be great to configure the cipher suites.

### Proposal:

Make the TLS listening version configurable.

### Current behavior:

Not possible to disable unwanted TLS versions.

### Desired behavior:

Default as current, but possible to enable only wanted TLS versions.

### Use case: 

Realy needed for security concerns, we want to disable old and maybe broken TLS versions.

svagner:
closes #5914

- tls_cipher_suites support was added for http server configuration
- tls_min_version/tls_max_version support was added for http server configuration

### Required for all PRs:


internal/tls/utils.go
```diff
@@ -0,0 +1,59 @@
...
+		"TLS_RSA_WITH_3DES_EDE_CBC_SHA":           tls.TLS_RSA_WITH_3DES_EDE_CBC_SHA,
+		"TLS_RSA_WITH_RC4_128_SHA":                tls.TLS_RSA_WITH_RC4_128_SHA,
+		"TLS_ECDHE_RSA_WITH_RC4_128_SHA":          tls.TLS_ECDHE_RSA_WITH_RC4_128_SHA,
+		"TLS_ECDHE_ECDSA_WITH_RC4_128_SHA":        tls.TLS_ECDHE_ECDSA_WITH_RC4_128_SHA,
+	}
+	for _, cipher := range ciphers {
+		if v, ok := cipherMap[cipher]; ok {
+			suites = append(suites, v)
+		} else {
+			return suites, fmt.Errorf("unsupported cipher %q", cipher)
```
danielnelson: `return nil, fmt.Errorf("unsupported cipher %q", cipher)`

plugins/outputs/prometheus_client/README.md
```diff
@@ -40,6 +40,42 @@ This plugin starts a [Prometheus](https://prometheus.io/) Client, it exposes all
   ## enable mutually authenticated TLS connections
   # tls_allowed_cacerts = ["/etc/telegraf/clientca.pem"]
 
+  ## Use only tls ciphers defined in this list
```
danielnelson: As the new options apply to many plugins, let's move this documentation here to a new document `docs/TLS.md` with a description of all the options in `internal/tls/config.go` and I'll link it in after we merge.

plugins/outputs/prometheus_client/README.md
```diff
@@ -40,6 +40,42 @@ This plugin starts a [Prometheus](https://prometheus.io/) Client, it exposes all
...
+  ## TLS_ECDHE_ECDSA_WITH_RC4_128_SHA
+  ## If value wasn't defined default will be used
+  # tls_cipher_suites = ["TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305", "TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305", "TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256", "TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256", "TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384", "TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384", "TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256", "TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA", "TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256", "TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA", "TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA", "TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA", "TLS_RSA_WITH_AES_128_GCM_SHA256", "TLS_RSA_WITH_AES_256_GCM_SHA384", "TLS_RSA_WITH_AES_128_CBC_SHA256", "TLS_RSA_WITH_AES_128_CBC_SHA", "TLS_RSA_WITH_AES_256_CBC_SHA"]
+
+  ## contains the minimum SSL/TLS version that is acceptable.
+  ## If not set, then TLS 1.0 is taken as the minimum.
+  # tls_min_version = "TLS11"
+
+  ## contains the maximum SSL/TLS version that is acceptable.
+  ## If not set, then the maximum version supported by this package is used,
```
danielnelson: The term "this package" is a bit confusing outside of the Go documentation.  It doesn't help that TLS1.3 requires an environment variable which we don't have support for here.  Let's write that the maximum version depends on version of Go that Telegraf is built with and leave it at that.

internal/tls/utils.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+func ParseTLSVersion(version string) (uint16, error) {
+	versionMap := map[string]uint16{
+		"SSL30": tls.VersionSSL30,
+		"TLS10": tls.VersionTLS10,
+		"TLS11": tls.VersionTLS11,
+		"TLS12": tls.VersionTLS12,
+		// for go >= 1.12
+		//"TLS13": tls.VersionTLS13,
```
danielnelson: This change will actually make it impossible to enable TLS1.3 anymore, not that I'm aware of anyone using it.  However, I think we should split this file into several files `common_go110.go`, `common_go120.go` so we have multiple version of the values and it is still able to set TLS1.3.

Let's do this for not only the TLS version but also the ciphers.

internal/tls/utils.go
```diff
@@ -0,0 +1,59 @@
...
+			return suites, fmt.Errorf("unsupported cipher %q", cipher)
+		}
+	}
+
+	return suites, nil
+}
+
+func ParseTLSVersion(version string) (uint16, error) {
+	versionMap := map[string]uint16{
+		"SSL30": tls.VersionSSL30,
```
danielnelson: Let's not even present SSL3.0 as an option, as it has known flaws and it's currently impossible to set and no reason to extend support to cover it.

svagner: SSL30 was removed from list

plugins/outputs/prometheus_client/README.md
svagner: I've added separate doc. Could you please review? I'm bad writer of docs:(

internal/tls/utils.go
svagner: Fixed

plugins/outputs/prometheus_client/README.md
svagner: It was changed:)

internal/tls/utils.go
svagner: I'm not sure that I've changed it with right way. I've create files _go112 and _notgo112 (for build on all other version without tls1.3). Basically as I understand we can create separate source for each version but in that way there will be error while move to 1.13...

danielnelson: Sorry, I told you wrong.  We want a `common.go` which will always be compiled, then we also should have a common_go112.go, these version tags work a bit different from the platform tags.  The file gets built when using any version of Go later than 1.12 (so it will get built in 1.13+ without a rename).

So in common.go we should have:
```
cipherMap = map[string]uint16{
	"TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305": tls.TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305,
}
```

Then over in common_go112.go we add in the new values:
```
func init() {
    cipherMap["blah"] = tls.BLAH
}
```

svagner: Oh, I see. But may be it would be better to have common_go110 and common_go111 that will remove unsupported versions/ciphers? I think it would be simpler to support that in future. How do you think?

danielnelson: Sounds good in theory but I'm not sure it's possible, the primary file needs to work in all versions.  Try it out and see how it goes.

svagner: So my idea isn't working you right because in that case I  should use undeclared constants from crypto/tls... I wanted to have this map as constnt. It it why I've covered it with getter.  
I also double checked, basically with current implementation it should be compiled with any versions, because in notgo112 there is `+build !go1.12` (< go.1.12) and in go112 - `+build go1.12` (>- go.1.12). But if you really want to have common.go it will work as well. It can just confuse that we add something in runtime to constant list...

svagner: Should I commit changes with this common.go as you offered?:) Or current implementation will work for you?

kmurph48:
Does anyone know how to get to the mother board of a Ubee router and reconfigure the data plan and subdue any unwanted traffic ?

plugins/outputs/prometheus_client/README.md
svagner: @danielnelson , is it looks ok?

internal/tls/utils.go
svagner: @danielnelson I've pushed changes that was recommended. Could you please review?:)

PR #6249, opened 2019-08-13
Add memory_usage (percent) field to procstat input plugin

sgtsquiggs:
I needed it and [gopsutils](https://github.com/shirou/gopsutil/blob/master/process/process.go#L222-L241) already provides it.

### Required for all PRs:


danielnelson:
Normally I don't like to add computed values, sticking to raw values, but I think in this case cpu/memory percent are okay due to how often they are needed.

danielnelson:
Thanks @rsaffi, I added it in 2754a484f980386176a3fbaccfb5d38dd9b8c4a7

Issue #6253, opened 2019-08-14
influxdb_listener doesn't respect the db in the query string


morfien101:
When sending metrics via the influxdb_listener allows you to send the database in the query string, however the plugin does not respect it.

Use case:
We would like to proxy our requests that our developers send directly to Influx to allow us to add and black list tags or fields.
We don't really want to have the development teams change how they send metrics. We control the location via environment variables.

Expected behaviour:
When a database is sent in the query it should be added as a database tag unless there is one already. This allows downstream outputs to send the metrics to the correct databases. Given this is a InfluxDB proxy, I would expect this to be the correct way or handling requests.

I'm happy to submit a PR for this feature/bug if it is accepted as a problem. If not can I be told what I should rather be doing to achieve this behaviour.

PR #6257, opened 2019-08-14
Allow influxdb_listener to keep the database name from the query string if supplied

morfien101:
Updating to allow the users to keep the database that gets sent in the query string. See issue #6253

Adding a test for the code change
Adding new configuration to the README.
Adding toml tags to make it like the rest of the plugins.

### Required for all PRs:

### DIFF
--- a/plugins/inputs/influxdb_listener/README.md
+++ b/plugins/inputs/influxdb_listener/README.md
@@ -46,6 +46,16 @@ submits data to InfluxDB determines the destination database.
   tls_cert = "/etc/telegraf/cert.pem"
   tls_key = "/etc/telegraf/key.pem"
 
+  ## If the write has a database on it then it should be kept
+  ## for metrics further on. The database will be added as a tag.
+  ## This tag can be used in downstream outputs.
+  keep_database = true
+
+  ## Optional tag name used to store the database if you want to change it to something custom. 
+  ## If not set it will be "database"
+  ## Only used if keep_database is set to true.
+  # database_tag = database
+
   ## Optional username and password to accept for HTTP basic authentication.
   ## You probably want to make sure you have TLS configured above for this.
   # basic_username = "foobar"
--- a/plugins/inputs/influxdb_listener/README.md
+++ b/plugins/inputs/influxdb_listener/README.md
@@ -50,6 +50,8 @@ submits data to InfluxDB determines the destination database.
   ## If the write has a database in the query string then it will be kept in this tag name.
   ## This tag can be used in downstream outputs.
   ## The default value of nothing means it will be off and the database will not be recorded.
+  ## If you have a tag that is the same as the one specified below, and supply a database,
+  ## the tag will be overwritten with the database supplied.
   # database_tag = ""
 
   ## Optional username and password to accept for HTTP basic authentication.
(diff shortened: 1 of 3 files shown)


plugins/inputs/influxdb_listener/README.md
```diff
@@ -46,6 +46,16 @@ submits data to InfluxDB determines the destination database.
   tls_cert = "/etc/telegraf/cert.pem"
   tls_key = "/etc/telegraf/key.pem"
 
+  ## If the write has a database on it then it should be kept
+  ## for metrics further on. The database will be added as a tag.
+  ## This tag can be used in downstream outputs.
+  keep_database = true
```
danielnelson: Skip this option and set the database tag based on if `database_tag` is set.

plugins/inputs/influxdb_listener/README.md
```diff
@@ -46,6 +46,16 @@ submits data to InfluxDB determines the destination database.
...
+
+  ## Optional tag name used to store the database if you want to change it to something custom. 
+  ## If not set it will be "database"
+  ## Only used if keep_database is set to true.
+  # database_tag = database
```
danielnelson: Have the default be "" which will mean do not set.

plugins/inputs/influxdb_listener/http_listener.go
```diff
@@ -371,6 +388,16 @@ func (h *HTTPListener) parse(b []byte, t time.Time, precision string) error {
 	}
 
 	for _, m := range metrics {
+		// Do we need to keep the database name in the query string
+		if h.KeepDatabase {
+			// Did we get a database argument. If we didn't get it. We can't set it.
+			if db != "" {
+				// Is there already a database set. If not use the database in the query string.
+				if _, ok := m.Tags()[h.DatabaseTag]; !ok {
```
danielnelson: Since the default will be off, it will be best to clobber the tag if it is already set.

plugins/inputs/influxdb_listener/README.md
morfien101: done

plugins/inputs/influxdb_listener/README.md
morfien101: done

plugins/inputs/influxdb_listener/http_listener.go
morfien101: I'm not sure what you mean by this one, so I'm assuming you mean if the db is set but the tag is not we should just ignore it. That's what I did, so if you want to change that you can comment on the line again.

plugins/inputs/influxdb_listener/http_listener.go
```diff
@@ -371,6 +380,16 @@ func (h *HTTPListener) parse(b []byte, t time.Time, precision string) error {
 	}
 
 	for _, m := range metrics {
+		// Do we need to keep the database name in the query string
+		if h.DatabaseTag != "" {
+			// Did we get a database argument. If we didn't get it. We can't set it.
+			if db != "" {
+				// Is there already a database set. If not use the database in the query string.
+				if _, ok := m.Tags()[h.DatabaseTag]; !ok {
```
danielnelson: I think we should set the tag unconditionally, overwriting the tag from the line protocol:
```
if h.DatabaseTag != "" && db != "" {
    m.AddTag(h.DatabaseTag, db)
}
```

morfien101: Right, I guess this would make it act more like influxdb's write endpoint which is the point... I'll make this change and push it up.

PR #1571, opened 2016-08-03
Disable darwin builds

sparrc:
### Required for all PRs:

- [ ] CHANGELOG.md updated

telegraf can't be cross-compiled for darwin, it has C dependencies and
thus many of the system plugins won't work.

cc @rossmcdonald

PR #6259, opened 2019-08-14
feat(scripts): add darwin (macOS) builds to the release

### LINKED ISSUE #4801: Provide a tar.gz release package for Darwin / MacOS with CGO
### Proposal:

It would be very helpful to obtain the "tar.gz" binary packages for Darwin under https://github.com/influxdata/telegraf/releases/latest .

### Current behavior:

At a particular release area, such as ["latest",](https://github.com/influxdata/telegraf/releases/latest), there are many platforms listed, but Darwin is not one of them.

### Desired behavior:

See a tar.gz binary package for Darwin-amd64 as a platform choice in the releases.

### Use case:

We are building a tool that automates installation of Telegraf on user devices and having a unified way of downloading and extracting the pre-built binary bundle on Linux, Windows, and MacOS would avoid one-off conditional code for MacOS devices. Right now the one-off code has to access the binaries via [Homebrew's repo](https://homebrew.bintray.com/bottles/telegraf-1.8.0.high_sierra.bottle.tar.gz) where the structure of that package is inconsistent with the official packages.

goller:
Closes #4801

This will build a darwin tar.gz package for macOS.  This allows people an easy way to download
releases from github directly.

### Required for all PRs:

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,7 @@ version.
 ### Nightly Builds
 
 These builds are generated from the master branch:
+- [telegraf-nightly_darwin_amd64.tar.gz](https://dl.influxdata.com/telegraf/nightlies/telegraf-nightly_darwin_amd64.tar.gz)
 - [telegraf_nightly_amd64.deb](https://dl.influxdata.com/telegraf/nightlies/telegraf_nightly_amd64.deb)
 - [telegraf_nightly_arm64.deb](https://dl.influxdata.com/telegraf/nightlies/telegraf_nightly_arm64.deb)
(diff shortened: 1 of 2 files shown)


vldgeorgiev:
This will be very useful for easier deployment of the executable as part of another package.  
However, I saw that some time ago those darwin builds were disabled because there was an issue with cross-compiling depencencies https://github.com/influxdata/telegraf/pull/1571.  
Is that solved now, so plugins will work in the new darwin tar.gz releases?

Issue #6262, opened 2019-08-15
Including the MQTT topic as tag or not


keomabrun:
## Feature Request

Hey! Thank you for the awesome work. I am using this plugin and I like it very much.

Could it be possible to decide whether the topic is saved as a tag or not ?

### Current behavior:

Currently, the topic is saved as a tag in Influxdb and it produces a lot of different series.
https://github.com/influxdata/telegraf/tree/master/plugins/inputs/mqtt_consumer#metrics

As stated in the Influx document, creating too many series will increase memory usage:
https://docs.influxdata.com/influxdb/v1.7/concepts/schema_and_data_layout/#don-t-have-too-many-series

### Desired behavior:

Could it be possible to let the user decide whether they want the topic to be saved as a tag or not ? 

### Use case:

I don't use the information contained in the topic after the data is saved in InfluxDB so I would like to remove the topics from the tags list.

keomabrun:
Wow ! That was fast. Thanks !

PR #6266, opened 2019-08-15
Add topic_tag option to mqtt_consumer

danielnelson:
Add a new option `topic_tag` to the mqtt_consumer input.  When unset the default is `topic` in order to preserve backwards compatibility, when set to `""` no topic tag is added, otherwise the value is used for the topic tag.  This is the basic pattern I would like to use for all inputs with parsers and want to add metadata.

In 2.0 we should change the default to not add the topic tag.  We don't have a mechanism for changing the defaults yet, this is something that needs a plan.

closes: #6262

### Required for all PRs:

### DIFF
--- a/plugins/inputs/mqtt_consumer/README.md
+++ b/plugins/inputs/mqtt_consumer/README.md
@@ -18,6 +18,10 @@ and creates metrics using one of the supported [input data formats][].
     "sensors/#",
   ]
 
+  ## The message topic will be stored in a tag specified by this value.  If set
+  ## to the empty string no topic tag will be created.
+  # topic_tag = "topic"
+
   ## QoS policy for messages
   ##   0 = at most once
   ##   1 = at least once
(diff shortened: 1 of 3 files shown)


plugins/inputs/mqtt_consumer/README.md
```diff
@@ -18,6 +18,10 @@ and creates metrics using one of the supported [input data formats][].
     "sensors/#",
   ]
 
+  ## The message topic will be stored in a tag specified by this value.  If set
+  ## to the empty string no topic tag will be created.
```
glinton: s/to the empty string/to an empty string,/ reads better

danielnelson: There is only one empty string, generally referred to as the empty string.  Check out the usage here: https://en.wikipedia.org/wiki/Empty_string

PR #6267, opened 2019-08-15
Add exec output plugin

glinton:
Continues and closes #4776


plugins/outputs/exec/exec.go
```diff
@@ -0,0 +1,113 @@
...
+
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/internal"
+	"github.com/influxdata/telegraf/plugins/outputs"
+	"github.com/influxdata/telegraf/plugins/serializers"
+)
+
+var sampleConfig = `
+  ## Command
+  command = "/usr/bin/mycollector --foo=bar"
```
goller: probably a nice command taht could be used with the sample config like, "echo"

Should this be an array?

plugins/outputs/exec/README.md
```diff
@@ -0,0 +1,24 @@
...
+This plugin executes external programs with serialized metrics data passed on stdin, one metric per line.  
+The external program is free to do what it wants with the data.
+
+### Configuration:
+
+```toml
+[[outputs.exec]]
+  # Shell/command
+  # Full command line to executable with parameters.
+  command = "/usr/local/bin/telegraf-output --config /etc/telegraf-output.conf"
```
goller: Probably echo here or some other command that could show multiple arguments

goller: would be nice to document this README and the sample config similarly to Docker's entrypoint:

https://docs.docker.com/engine/reference/builder/#entrypoint

plugins/outputs/exec/README.md
```diff
@@ -0,0 +1,24 @@
...
+  
+  # Timeout for each command to complete.
+  timeout = "30s"
+  
+  # Data format to output.
+  ## Each data format has its own unique set of configuration options, read
+  ## more about them here:
+  ## https://github.com/influxdata/telegraf/blob/master/docs/DATA_FORMATS_OUTPUT.md
+  data_format = "json"
```
goller: Is it typical for the sample config and the readme to have different configurations?  In the sample it is influx

plugins/outputs/exec/exec.go
```diff
@@ -0,0 +1,113 @@
...
+
+		status, _ := internal.ExitStatus(err)
+		return fmt.Errorf("[outputs.exec] %q exited %d with %s", command, status, err.Error())
+	}
+
+	return nil
+}
+
+func (e *Exec) Description() string {
+	return "Send Telegraf metrics to commands that can input from stdin"
```
goller: return "Send metrics to command as input over stdin"

plugins/outputs/exec/exec.go
```diff
@@ -0,0 +1,113 @@
...
+
+	var buffer bytes.Buffer
+	for _, metric := range metrics {
+		value, err := e.serializer.Serialize(metric)
+		if err != nil {
+			return err
+		}
+		buffer.Write(value)
+	}
+
```
goller: Might as well remove the len(metrics) and just check if buffer.Len() > 0

plugins/outputs/exec/exec.go
```diff
@@ -0,0 +1,113 @@
...
+	return nil
+}
+
+func init() {
+	outputs.Add("exec", func() telegraf.Output {
+		return &Exec{
+			runner:  CommandRunner{},
+			Timeout: internal.Duration{Duration: time.Second * 5},
```
goller: Shoudl this Timeout be 30 * time.Second ?  That's what is in the config.

plugins/outputs/exec/exec.go
```diff
@@ -0,0 +1,139 @@
...
+	return nil
+}
+
+// Close kills the process if it is still running.
+func (c *CommandRunner) Close() error {
+	if c.cmd == nil {
+		return nil
+	}
+
+	return c.cmd.Process.Kill()
```
danielnelson: After internal.RunTimeout returns the process is no longer running, so I don't think this Close function would ever do anything.  Telegraf will not call Close until Write returns.

plugins/outputs/exec/exec_test.go
```diff
@@ -0,0 +1,57 @@
...
+		{
+			command: []string{"sleep", "5s"},
+			err:     true,
+		},
+		{
+			command: []string{"/no/exist", "-h"},
+			err:     true,
+		},
+	}
+
```
danielnelson: This test appears to run real commands, so it needs to excluded as part of `go test -short`  It also likely doesn't work on Windows or anywhere that tee/sleep are not in the PATH.

internal/internal.go
```diff
@@ -236,11 +236,8 @@ func WaitTimeout(c *exec.Cmd, timeout time.Duration) error {
 	})
 
 	err := c.Wait()
-	isTimeout := timer.Stop()
 
-	if err != nil {
-		return err
-	} else if isTimeout == false {
+	if !timer.Stop() {
```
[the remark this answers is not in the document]
glinton: I never could get it to return a timeout error when the command was killed without this.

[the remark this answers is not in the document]
glinton: The error returned when we kill it (after timeout) is always `-1 killed` so returning `TimeoutErr` after killing by timeout i feel like is the intended behavior of this function anyways.

PR #6269, opened 2019-08-16
Update README.md

timhallinflux:
added example for multiple repos.  Doc only change.

### Required for all PRs:


danielnelson:
Normally we don't show multiple configurations examples, the hope is that the configuration will be documented well enough that there won't be any confusion about the alterations you can make.  Perhaps we should just have the default config list multiple repos?

danielnelson:
See edb05b58a0d8258a6be54df89086108f2e66956b

Issue #6278, opened 2019-08-16
--test does not return a correct status


fhriley:
I was attempting to add simple `telegraf` testing to my CI using the `--test` option. However, [as you can see here](https://github.com/influxdata/telegraf/blob/73914ac920e6e95763fcbc0ccf378a60046c7d04/agent/agent.go#L238), `telegraf` will not return an error if any of the input plugins fails. Since the `Test()` method doesn't return an error there, [the `telegraf` command will also not return an error](https://github.com/influxdata/telegraf/blob/73914ac920e6e95763fcbc0ccf378a60046c7d04/cmd/telegraf/telegraf.go#L208) and will have a status code of 0, which will pass in CI.

PR #6279, opened 2019-08-16
Fix issue #6278

fhriley:
A couple notes:

1. I added a separate `--test-status` option to `telegraf` to implement this. I did this to maintain backwards compatibility in case anyone is depending on the current behavior of always returning 0.

2. The exec plugin was always returning no error. I fixed this plugin since we are using it. It is possible other plugins are also not correctly returning any errors.

closes: #6278

### Required for all PRs:

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -235,6 +235,9 @@ func (a *Agent) Test(ctx context.Context, waitDuration time.Duration) error {
 		a.stopServiceInputs()
 	}
 
+	if NErrors.Get() > 0 {
+		return fmt.Errorf("One or more input plugins had an error")
+	}
 	return nil
 }
 


danielnelson:
Thanks for opening a pull request, I have a few high level ideas:

I don't think we need to add a new cli flag, if there are any errors during test we can return 1.

For plugins like exec with multiple commands, or plugins that connect to multiple remote services, we don't actually want to fish the error out and instead use the AddError function on the accumulator.  This prevents multiple logging and keeps us from needing lists of errors.  What we should do in Test() is call `NErrors.Get()` and if it is greater than 0 return an error.

fhriley:
Updated based on your comments.

PR #6293, opened 2019-08-21
Add selective-build script to build Telegraf with selected plugins

farzadghanei:
By default Telegraf is built with all the plugins. This PR adds a script to modify the source code in place, by excluding unnecessary plugins to build a Telegraf with only selected plugins, resulting to a binary with smaller footprint.

The script accepts list of plugins from command line arguments, environment variables or an `.ini` configuration file (with same keys as cmd args). By default builds `telegraf` and reverts the local changes, but both steps can be skipped (`--no-build` and `--no-git`).

The script doesn't take care of building Docker image nor packaging scripts, but maybe used as tool for those processes.

### Examples
Build Telegraf with all aggregator plugins excluded except `minmax`,`basicstats`, and [disables `pprof` endpoint](https://github.com/influxdata/telegraf/blob/ac66838f9fa58cc5b9632c1614db6f49c2fa4515/cmd/telegraf/telegraf.go#L10).

```
python3 scripts/selective-build.py --verbose --config build.ini
```

where `build.ini` is:

```
[telegraf-selective-build]
aggregators=minmax,basicstats
no-pprof=true
```

Exclude all the aggregators, all outputs but `influxdb` and skip building and don't revert local changes
```
python3 ./scripts/selective-build.py --aggregators '_' --outputs=influxdb --no-build --no-git
```

### Required for all PRs:

### DIFF
--- a/scripts/selective-build.py
+++ b/scripts/selective-build.py
@@ -1,0 +1,173 @@
+#!/usr/bin/env python3
+
+import sys
+from os import EX_OK, EX_TEMPFAIL, environ
+from subprocess import check_output, call
+import os.path
+from logging import DEBUG, INFO, StreamHandler, getLogger
+from configparser import ConfigParser
+from argparse import ArgumentParser
+
+CONFIG_SECTION = 'telegraf-selective-build'
+PLUGIN_TYPES = ('inputs', 'outputs', 'aggregators', 'processors')
+logger = getLogger(__name__)
+
+
+def parse_args(args=None):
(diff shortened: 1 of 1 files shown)


danielnelson:
This looks cool, thanks for sharing.  We do want to add something like this, but right now I don't want to bring in any new python code (actually hoping to retire build.py and fpm(ruby) one of these days).

related https://github.com/influxdata/telegraf/pull/5809

PR #6304, opened 2019-08-22
Split out -w argument in iptables input

### LINKED ISSUE #6234: iptables input plugin use_lock error
Hi,
Versions: Telegraf 1.11.4 / CentOS 7.6

This is how telegraf calls iptables when use_lock is true:

`execve("/usr/sbin/iptables", ["/usr/sbin/iptables", "-wnvL", "OUTPUT", "-t", "filter", "-x"], [/* 10 vars */] <unfinished ...>`

Telegraf errors with exit code 2 in the log.

This is why:
```
# /usr/sbin/iptables -wnvL FORWARD -t filter -x
iptables v1.4.21: wait seconds not numeric
```

Solution: separate the -w flag from the rest of the flags:

```
# /usr/sbin/iptables -w -nvL FORWARD -t filter -x
Chain FORWARD (policy ACCEPT 0 packets, 0 bytes)
    pkts      bytes target     prot opt in     out     source               destination
```

danielnelson:
closes #6234

### Required for all PRs:

### DIFF
--- a/plugins/inputs/iptables/iptables.go
+++ b/plugins/inputs/iptables/iptables.go
@@ -37,7 +37,7 @@ func (ipt *Iptables) SampleConfig() string {
   ## iptables can be restricted to only list command "iptables -nvL".
   use_sudo = false
   ## Setting 'use_lock' to true runs iptables with the "-w" option.
-  ## Adjust your sudo settings appropriately if using this option ("iptables -w -nvl")
+  ## Adjust your sudo settings appropriately if using this option ("iptables -w 5 -nvl")
   use_lock = false
   ## Define an alternate executable, such as "ip6tables". Default is "iptables".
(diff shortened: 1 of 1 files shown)


plugins/inputs/iptables/iptables.go
```diff
@@ -91,7 +91,7 @@ func (ipt *Iptables) chainList(table, chain string) (string, error) {
 	}
 	iptablesBaseArgs := "-nvL"
 	if ipt.UseLock {
-		iptablesBaseArgs = "-wnvL"
+		iptablesBaseArgs = "-w -nvL"
```
glinton: Shouldn't this be a slice in the command args?

danielnelson: Yeah

PR #6324, opened 2019-08-28
Set per file csv parser to allow individual headers per file

### LINKED ISSUE #6318: input.file plugin behavior with different input metrics files
### Relevant telegraf.conf:

```
[[inputs.file]]
  files = ["/data/siem/input/*.csv"]
  data_format = "csv"
  csv_header_row_count = 1
```

### Steps to reproduce:

1. use two or more csv files for input but set different headers rows there

### Expected behavior:
Read header row from every file in the input list and use it for ongoing metrics.

### Actual behavior:
Telegraf when started reads header row only from the first file of “files = [”/data/siem/input/*.csv"]" and uses it for the next file which has its own different header row.

### Additional info:
I just removed the column names check from the source. But ideally this should be configurable behavior.
Below are diffs for
~/go/src/github.com/influxdata/telegraf/plugins/parsers/csv/parser.go
```diff
> 64c64
-       if len(p.ColumnNames) == 0 {
> ---
+ //    if len(p.ColumnNames) == 0 {
> 84,89c84,89
-       } else {
-               // if columns are named, just skip header rows
-               for i := 0; i < p.HeaderRowCount; i++ {
-                       csvReader.Read()
-               }
-       }
> ---
+ //    } else {
+ //            // if columns are named, just skip header rows
+ //            for i := 0; i < p.HeaderRowCount; i++ {
+ //                    csvReader.Read()
+ //            }
+ //    }
```

glinton:
One method to resolve #6318

### DIFF
--- a/plugins/inputs/file/file.go
+++ b/plugins/inputs/file/file.go
@@ -15,6 +15,8 @@ type File struct {
 	Files  []string `toml:"files"`
 	parser parsers.Parser
 
+	csvParser csv.Parser
+	csv       bool
 	filenames []string
 }
 
@@ -51,11 +53,9 @@ func (f *File) Gather(acc telegraf.Accumulator) error {
 	for _, k := range f.filenames {
 		p := f.parser
 
-		switch pp := f.parser.(type) {
-		case *csv.Parser:
-			// get and use a unique parser per file (allows headers to be re-read)
-			p2 := *pp
-			p = &p2
+		if f.csv {
+			pp := f.csvParser
+			p = &pp
 		}
 
 		metrics, err := f.readMetric(p, k)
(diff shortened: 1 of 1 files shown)


rkleivel:
Is there any info as to if/when this fix will reach a release? I need to process files with varying numbers and orders of columns, and #6318 is a showstopper.
Thanks!

srebhan:
Let's get this rolling again. @glinton: I can reproduce the problem but it seems (after rebasing) that your approach does not fix the problem. Instead the field names are still wrong... I noticed that the pointers to the parsers in `readMetric()` are all the same. So can you please first rebase your patches and resolve the merge conflict. And second check that your approach solves #6318!? Thanks for your effort!

ssoroka:
merged in #8400

PR #6332, opened 2019-08-30
Fix filecount for paths with trailing slash

### LINKED ISSUE #6329: Filecount: Trailing slash in directories does not work
**Relevant telegraf configuration:**
```
[[inputs.filecount]]
	directories = ["/tmp/test/"]
```
**System info:**
- Debian 9.9
- Telegraf 1.11.5

**Steps to reproduce:**
1. Create a test directory:
```
mkdir /tmp/test
touch /tmp/test/file{1,2,3,4}
```
2. Create an input file (/tmp/input.conf) 
```
[[inputs.filecount]]
	directories = ["/tmp/test/"]
```
3. Execute it: 
```
test@example:/tmp# telegraf --config input.conf --test
2019-08-29T12:04:33Z I! Starting Telegraf 1.11.5
test@example:/tmp#
```
**Expected output:**
Command should return:
```
test@example:/tmp# telegraf --config input.conf --test
2019-08-29T12:05:41Z I! Starting Telegraf 1.11.5
> filecount,directory=/tmp/test/,host=backup count=4i,size_bytes=0i 1567080341000000000
test@example:/tmp# 
```

If we remove the trailing / in directories, it works fine.
It could be an expected behavior to process only directories without trailing slashes, but in this case, it would be great to emphasize it in documentation.

By the way, thank you for your work on telegraf, it is a great piece of software :)

danielnelson:
Paths returned during the walk have had filepath.Clean applied to them, so we need to create a glob that will match these cleaned paths.

closes #6329 

### Required for all PRs:

### DIFF
--- a/plugins/inputs/filecount/filecount.go
+++ b/plugins/inputs/filecount/filecount.go
@@ -1,6 +1,7 @@
 package filecount
 
 import (
+	"fmt"
 	"log"
 	"os"
 	"path/filepath"
@@ -156,6 +157,7 @@ func (fc *FileCount) count(acc telegraf.Accumulator, basedir string, glob globpa
 	childSize := make(map[string]int64)
 
 	walkFn := func(path string, de *godirwalk.Dirent) error {
+		fmt.Println(path)
 		rel, err := filepath.Rel(basedir, path)
 		if err == nil && rel == "." {
 			return nil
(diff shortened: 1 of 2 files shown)


plugins/inputs/filecount/filecount.go
```diff
@@ -156,6 +157,7 @@ func (fc *FileCount) count(acc telegraf.Accumulator, basedir string, glob globpa
 	childSize := make(map[string]int64)
 
 	walkFn := func(path string, de *godirwalk.Dirent) error {
+		fmt.Println(path)
```
glinton: Remove this leftover print.

Issue #6347, opened 2019-09-04
Allow the agent to not apply selected global tags on some inputs.


pberlowski:
## Feature Request

When setting up multi-layer telegraf system, we often find ourselves in a situation where we expect global tags to be applied on all but one plugin. An example of such situation is a proxy-telegraf where we need to apply the "hostname" for all the internal and hardware plugins but we need it excluded from the http_listener input. If the measurement comes via HTTP without the "host" tag, we want to honor that. On top of that the capability to "tagexclude" doesn't work in this case as we expect some metrics to come in with the "host" present and we want to retain it.

### Proposal:

Create a configuration on the running_input to exclude some global tags.

### Current behavior:

All global tags are applied on every input.

### Desired behavior:

It is possible to configure an input to not apply specific global tags.

### Use case: 

Telegraf agent accepts measurements submitted via http (used as influxdb proxy). It doesn't apply the hostname tag to the measurements that are proxied and don't have it, but it applies hostname tag to its internal telemetry.

danielnelson:
The `host` tag set by `hostname = ""` in the `agent` section of the configuration should work this way.  It is only set if there is no existing `host` tag.

Another possible option is to use the `override` processor instead of the global_tags section to set the tag.  This processor runs after the inputs and could use an attribute of the metrics to decide if it should set the tag.  The processor only runs on the metrics that are allowed in via namepass/namedrop and tagpass/tagdrop.  For example this sets the tag `hostname=howdy` on all metrics with the measurement `cpu`:
```
[[processors.override]]
  namepass = ["cpu"]
  [processors.override.tags]
    hostname = "howdy"
```

pberlowski:
Thanks for looking at this @danielnelson. We've investigated the solutions you posted and found that they don't work or add some undesired overhead in the config.

Also, I might not have been detailed enough on the use-cases. Here's more information on what we're trying to achieve.

## Baseline:

### Use case 1: Metric is received via HTTP without "host"

1. The agent receives a `foo` metric via HTTP.
1. The agent does not apply `host` to the metric even though it's missing.
1. The agent emits the metric without `host` tag.

### Use case 2: Metric is received via HTTP with "host"

1. The agent receives a `bar` metric via HTTP.
1. The agent does not change the value of the `host` tag.
1. The agent emits the metric with the original `host` tag.

### Use case 3: Metric is collected via a cpu input without "host"

1. The agent collects a `baz` metric via a cpu input.
[...]

PR #6348, opened 2019-09-04
Issue #6347: Allow ignoring global tags on an input

pberlowski:
This PR addresses issue #6347

When setting up multi-layer telegraf system, we often find ourselves in a situation where we expect global tags to be applied on all but one plugin. An example of such situation is a proxy-telegraf where we need to apply the "hostname" for all the internal and hardware plugins but we need it excluded from the http_listener input. If the measurement comes via HTTP without the "host" tag, we want to honor that. On top of that the capability to "tagexclude" doesn't work in this case as we expect some metrics to come in with the "host" present and we want to retain it.

### Required for all PRs:


danielnelson:
Closing the PR based on the discussion in #6347, I don't think we should add this.  We can continue to discuss solutions on that issue.

Issue #6342, opened 2019-09-04
Cannot connect to Azure EventHub using inputs.kafka_consumer


OnurbHartmann:
The Microsoft Azure EventHub exposes a Kafka 1.0 compatible protocol. With that we tried to connect using the inputs.kafka_consumer and SASL Authentication. However it won't work.
As described in the sarama issue below even though Microsoft implemented 1.0 protocol the SASL authentication works only with V0. This cannot be configured via telegraf.conf yet.

### Relevant telegraf.conf:
```
[[inputs.kafka_consumer]]
#   ## kafka servers
   brokers = ["<namespace>.servicebus.windows.net:9092"]
#   ## topic(s) to consume
   topics = ["<Event Hub>"]
#   ## Add topic as tag if topic_tag is not empty
#   # topic_tag = ""
#
#   ## Optional Client id
    client_id = "AzEvHub"
#
#   ## Set the minimal supported Kafka version.  Setting this enables the use of new
#   ## Kafka features and APIs.  Of particular interest, lz4 compression
#   ## requires at least version 0.10.0.0.
#   ##   ex: version = "1.1.0"
    version = "1.1.0"
#
#   ## Optional TLS Config
#   # tls_ca = "/etc/telegraf/ca.pem"
#   # tls_cert = "/etc/telegraf/cert.pem"
#   # tls_key = "/etc/telegraf/key.pem"
#   ## Use TLS but skip chain & host verification
#   # insecure_skip_verify = false
#
#   ## Optional SASL Config
    sasl_username = "<username>"
    sasl_password = "<password>"
#
#   ## the name of the consumer group
   consumer_group = "telegraf"
#   ## Offset (must be either "oldest" or "newest")
   offset = "oldest"
#   ## Maximum length of a message to consume, in bytes (default 0/unlimited);
#   ## larger messages are dropped
   max_message_len = 1000000
#
#   ## Maximum messages to read from the broker that have not been written by an
#   ## output.  For best throughput set based on the number of metrics within
#   ## each message and the size of the output's metric_batch_size.
#   ##
#   ## For example, if each message from the queue contains 10 metrics and the
#   ## output metric_batch_size is 1000, setting this to 100 will ensure that a
#   ## full batch is collected and the write is triggered immediately without
#   ## waiting until the next flush_interval.
#   # max_undelivered_messages = 1000
#
#   ## Data format to consume.
#   ## Each data format has its own unique set of configuration options, read
#   ## more about them here:
[...]

PR #6350, opened 2019-09-05
Add kafka SASL version control to kafka_consumer

danielnelson:
Add manual control of SASL version for testing.  Before merging this support should also be added to the output.

related #6342

### Required for all PRs:

### DIFF
--- a/plugins/inputs/kafka_consumer/kafka_consumer.go
+++ b/plugins/inputs/kafka_consumer/kafka_consumer.go
@@ -2,6 +2,7 @@ package kafka_consumer
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log"
 	"strings"
@@ -95,6 +96,7 @@ type KafkaConsumer struct {
 	Version                string   `toml:"version"`
 	SASLPassword           string   `toml:"sasl_password"`
 	SASLUsername           string   `toml:"sasl_username"`
+	SASLVersion            *int     `toml:"sasl_version"`
 
 	tls.ClientConfig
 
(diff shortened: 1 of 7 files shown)


danielnelson:
To use with Azure EventHub:
- Set `enable_tls = true`
- Set `sasl_version = 0`
- Patch Shopify/sarama with https://github.com/Shopify/sarama/pull/1542, this will hopefully not be required soon.

plugins/inputs/kafka_consumer/kafka_consumer.go
```diff
@@ -158,20 +166,32 @@ func (k *KafkaConsumer) Init() error {
...
 	tlsConfig, err := k.ClientConfig.TLSConfig()
 	if err != nil {
 		return err
 	}
 
 	if tlsConfig != nil {
 		config.Net.TLS.Config = tlsConfig
+
+		// TLS is forced on if a custom tls.Config is set.
 		config.Net.TLS.Enable = true
```
reimda: If my config had custom TLS settings and also "enable_tls = false" I would expect the TLS config to be ignored and TLS not to be used.  This line would enable tls even though I explicitly said not to.

plugins/outputs/kafka/kafka.go
```diff
@@ -258,20 +263,32 @@ func (k *Kafka) Connect() error {
...
 	tlsConfig, err := k.ClientConfig.TLSConfig()
 	if err != nil {
 		return err
 	}
 
 	if tlsConfig != nil {
 		config.Net.TLS.Config = tlsConfig
+
+		// TLS is forced on if a custom tls.Config is set.
 		config.Net.TLS.Enable = true
```
reimda: same here

plugins/inputs/kafka_consumer/kafka_consumer.go
danielnelson: I'll update the option to be a `*bool` so we can have the distinction between unset and `enable_tls = false`.

danielnelson:
I added the EnableTLS option as a `*bool`, so that we have 3 states.  When set to false, TLS will not be used.  For backwards compatibility, if unset TLS will be enabled when the tlsconfig is modified.

I also added a warning message that will be displayed encouraging the use of `enable_tls` to control this setting, in 2.0 this option should be required to turn on TLS.  I'd like to start adding more warnings like this so that it will be easier to update a configuration before a major release.

PR #6357, opened 2019-09-05
Document and add support to input plugins for logging alias

glinton:


plugins/aggregators/basicstats/basicstats.go
```diff
@@ -222,45 +220,38 @@ func parseStats(names []string) *configuredStats {
...
-	defaults.min = true
-	defaults.max = true
-	defaults.mean = true
-	defaults.variance = true
-	defaults.stdev = true
-	defaults.sum = false
-	defaults.non_negative_diff = false
-
-	return defaults
+var defaultStats = &configuredStats{
```
danielnelson: I wouldn't do this, as it means all instances share the same configuredStats instance.  If any plugin is modified in the future to modify this struct it will affect all plugin instances.  So I feel this is playing with fire.

If you want to simplify this, I'd call getConfiguredStats() from the Init() function and have it return/set a non-pointer struct.  If we initialize it in Init we don't need to check it for nil on Push anymore.  The `parseStats(` is probably better as a non-member function.

docs/CONFIGURATION.md
```diff
@@ -186,6 +186,7 @@ driven operation.
 
 Parameters that can be used with any input plugin:
 
+- **alias**: Alias that will get logged along with the plugin name. Allows to differentiate between multiple definitions of a plugin.
```
danielnelson: Split this definition from the rest of the PR, this can go into 1.12.1 and the rest into 1.13.

The documentation should describe `alias` as a way to name the instance of the plugin, you can mention logging and internal metrics as examples of where this may appear, but don't describe it as just a logging option.

Also line wrap at 78-80 chars.

glinton: Why would we document it before it's supported though? Seems like a package deal.

plugins/aggregators/basicstats/basicstats.go
glinton: D'oh! good catch

plugins/aggregators/basicstats/basicstats.go
```diff
@@ -222,45 +220,38 @@ func parseStats(names []string) *configuredStats {
 			parsed.non_negative_diff = true
 
 		default:
-			log.Printf("W! Unrecognized basic stat '%s', ignoring", name)
+			b.Log.Warnf("unrecognized basic stat '%s', ignoring", name)
```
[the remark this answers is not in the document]
glinton: Agreed. About a third of the way through with this, I wondered if whatever recently merged plugin that logs in lowercase was a mistake to use as an example of how we prefer to log.

PR #6361, opened 2019-09-06
Added `script` option to be able set a sql file as a query source

### LINKED ISSUE #6242: Feature Request: Add ability to read sql file
## Feature Request

### Proposal:
Add ability to pass in parameter to sql file location

### Current behavior:
sqlquery="SELECT * FROM pg_stat_database where datname"

### Desired behavior:
sqlquery=/var/tmp/sqlfile.sql

### Use case: 
This would help greatly on servers that have multiple long queries instead of putting it into sqlquery which could get really large

Zyqsempai:
### Required for all PRs:

Fixes #6242

### DIFF
--- a/plugins/inputs/postgresql_extensible/postgresql_extensible.go
+++ b/plugins/inputs/postgresql_extensible/postgresql_extensible.go
@@ -3,7 +3,9 @@ package postgresql_extensible
 import (
 	"bytes"
 	"fmt"
+	"io/ioutil"
 	"log"
+	"os"
 	"strings"
 
 	// register in driver.
@@ -25,6 +27,7 @@ type Postgresql struct {
 
 type query []struct {
 	Sqlquery    string
+	Script      string
 	Version     int
 	Withdbname  bool
 	Tagvalue    string
@@ -108,6 +111,20 @@ func (p *Postgresql) IgnoredColumns() map[string]bool {
 	return ignoredColumns
 }
 
+func (p *Postgresql) ReadQueryFromFile(filePath string) (string, error) {
+	file, err := os.Open(filePath)
+	if err != nil {
+		return "", err
+	}
+	defer file.Close()
+
+	qyery, err := ioutil.ReadAll(file)
+	if err != nil {
+		return "", err
+	}
+	return string(qyery), err
+}
+
 func (p *Postgresql) Gather(acc telegraf.Accumulator) error {
 	var (
 		err         error
@@ -129,8 +146,18 @@ func (p *Postgresql) Gather(acc telegraf.Accumulator) error {
 	// We loop in order to process each query
 	// Query is not run if Database version does not match the query version.
 	for i := range p.Query {
+		if p.Query[i].Sqlquery != "" && p.Query[i].Script != "" {
+			return fmt.Errorf("both 'sqlquery' and 'script' specified, please choose one option")
+		}
 		sql_query = p.Query[i].Sqlquery
+		if sql_query == "" {
+			sql_query, err = p.ReadQueryFromFile(p.Query[i].Script)
+			if err != nil {
+				return err
+			}
+		}
 		tag_value = p.Query[i].Tagvalue
+
 		if p.Query[i].Measurement != "" {
 			meas_name = p.Query[i].Measurement
 		} else {
--- a/plugins/inputs/postgresql_extensible/postgresql_extensible_test.go
+++ b/plugins/inputs/postgresql_extensible/postgresql_extensible_test.go
@@ -145,7 +145,7 @@ func TestPostgresqlFieldOutput(t *testing.T) {
 	}
 
 	acc := queryRunner(t, query{{
-		Sqlquery:   "select * from pg_stat_database",
+		Script:     "test.sql",
 		Version:    901,
 		Withdbname: false,
 		Tagvalue:   "",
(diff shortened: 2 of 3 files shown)


plugins/inputs/postgresql_extensible/postgresql_extensible.go
```diff
@@ -129,8 +146,18 @@ func (p *Postgresql) Gather(acc telegraf.Accumulator) error {
 	// We loop in order to process each query
 	// Query is not run if Database version does not match the query version.
 	for i := range p.Query {
+		if p.Query[i].Sqlquery != "" && p.Query[i].Script != "" {
+			return fmt.Errorf("both 'sqlquery' and 'script' specified, please choose one option")
```
glinton: I would remove this and just document that `sqlquery` takes preference if both are defined.

plugins/inputs/postgresql_extensible/README.md
```diff
@@ -43,6 +43,9 @@ The example below has two queries are specified, with the following parameters:
   # withdbname was true.
   # Be careful that if the withdbname is set to false you don't have to define
   # the where clause (aka with the dbname)
+  # Also script option can be used to speciafy the .sql file path,
```
glinton: Revise this. Add a space before the block. Plainly document what `script` is for.

Update `SampleConfig()` as well

plugins/inputs/postgresql_extensible/postgresql_extensible.go
```diff
@@ -108,6 +111,20 @@ func (p *Postgresql) IgnoredColumns() map[string]bool {
...
 }
 
+func (p *Postgresql) ReadQueryFromFile(filePath string) (string, error) {
+	file, err := os.Open(filePath)
+	if err != nil {
+		return "", err
+	}
+	defer file.Close()
+
+	qyery, err := ioutil.ReadAll(file)
```
glinton: `s/qyery/query/`

plugins/inputs/postgresql_extensible/test.sql
```diff
@@ -0,0 +1 @@
+select * from pg_stat_database
```
glinton: Put this into a `dev` or `testfiles` directory and reference it as such

plugins/inputs/postgresql_extensible/postgresql_extensible_test.go
```diff
@@ -145,7 +145,7 @@ func TestPostgresqlFieldOutput(t *testing.T) {
 	}
 
 	acc := queryRunner(t, query{{
-		Sqlquery:   "select * from pg_stat_database",
+		Script:     "test.sql",
```
glinton: Add a new test for this so query testing isn't removed.

plugins/inputs/postgresql_extensible/test.sql
danielnelson: Name the directory `testdata`, since this is treated specially and is idiomatic:
```
$ go help packages | grep testdata -C1
Directory and file names that begin with "." or "_" are ignored
by the go tool, as are directories named "testdata".
```

plugins/inputs/postgresql_extensible/postgresql_extensible.go
danielnelson: I don't have a strong opinion on returning an error vs precedence, though if I was asked to choose I think an error is slightly less prone to mistakes.  However, I would like the script loading to be done only once.  The easiest way to do this is to move it into a new Init() function:
```
func (p *Postgresql) Init() error {
    // load the file and save onto the Postgresql struct.
    // return an error if the file is unreadable or configuration is invalid
}
```

plugins/inputs/postgresql_extensible/postgresql_extensible.go
```diff
@@ -108,6 +111,20 @@ func (p *Postgresql) IgnoredColumns() map[string]bool {
 	return ignoredColumns
 }
 
+func (p *Postgresql) ReadQueryFromFile(filePath string) (string, error) {
```
danielnelson: Make this a non-member function.

Zyqsempai:
@glinton PTAL

plugins/inputs/postgresql_extensible/README.md
```diff
@@ -44,6 +44,9 @@ The example below has two queries are specified, with the following parameters:
   # Be careful that if the withdbname is set to false you don't have to define
   # the where clause (aka with the dbname)
   #
+  # The script option can be used to speciafy the .sql file path,
```
glinton: `s/speciafy/specify` then end the sentence on this line, remove `in case` on the next line and just start it with `If script and...`

Same goes for the sampleConfig in the source file (make sure they match)

plugins/inputs/postgresql_extensible/postgresql_extensible.go
glinton: Spelling still needs fixed.

plugins/inputs/postgresql_extensible/postgresql_extensible_test.go
```diff
@@ -145,7 +145,7 @@ func TestPostgresqlFieldOutput(t *testing.T) {
 	}
 
 	acc := queryRunner(t, query{{
-		Sqlquery:   "select * from pg_stat_database",
+		Script:     "select * from pg_stat_database",
```
glinton: Change this back to `Sqlquery`

plugins/inputs/postgresql_extensible/postgresql_extensible.go
```diff
@@ -130,7 +150,14 @@ func (p *Postgresql) Gather(acc telegraf.Accumulator) error {
 	// Query is not run if Database version does not match the query version.
 	for i := range p.Query {
 		sql_query = p.Query[i].Sqlquery
+		if sql_query == "" {
+			sql_query, err = ReadQueryFromFile(p.Query[i].Script)
```
glinton: I agree with Daniel, move this into `Init()` and save it on the struct so the file isn't read each call to `Gather`

```go
func (p *Postgresql) Init() error {
    // load the file and save onto the Postgresql struct.
    // return an error if the file is unreadable or configuration is invalid
}
```

PR #6365, opened 2019-09-07
New Feature Juniper Telemetry Interface native parser

AndrewGambell:
New parser for Juniper Networks Native Telemetry interface. Parses telemetry protobuf messages emitted from Juniper devices over UDP and formats into Telegraf/Influx metrics.

This differs from the OpenConfig JTI plugin which is a subscribe model, the JTI Native sensor implements a fire and forget, push based telemetry model.

An overview of the two models can be found here:
https://www.juniper.net/documentation/en_US/junos/topics/concept/junos-telemetry-interface-oveview.html

### Required for all PRs:


daemonkeeper:
How is the code size the determining factor for a plugin's reasoning to reject? Looking at it, most of it is indeed just protobuf message format definitions, which are not introducing any sort of dependency into the base architecture that would be a burden to maintain. 

On the other hand, this plugin leverages the support for telemetry data streaming to Telegraf, which in turn makes it the primary product to look at for monitoring solutions of the 21st century. I for once are very happy @AndrewGambell took the effort to write it, since I neither feel like using fluentd (which has support for this type of data), or ELK.

shaygun:
This is awesome solution for juniper telemetry, the code look promising aside from lots of proto buffer. But I am trying to give it a try! 
Thanks @AndrewGambell

makzdot:
When will this be merged? I could really use this.

mrambossek:
i could also very much use this and dont really see the reasoning behind the "too large to merge" argument? but i am not a developer. as a sysadmin i would appreciate taking this into serious consideration again. thanks :)

caohunan:
@mrambossek I came across this thread while searching for a plugin to handle juniper native telemetry. I was a bit frustrated as well, but I got what it means by 'too big'. 
After merging the code to the latest master branch, the telegraf binary was increased from 95MB to 105MB, so it is literally 15% larger. I think 'too big' is reasonable to me.

sjwang90:
Hey @AndrewGambell, thanks for your contribution but based off the comments above this parser at the current state is too big to merge into Telegraf. 

However the `processors.execd` that was released will allow you to do what daniel mentioned above to collect data via an external process. You can find more information about [external plugins](https://github.com/influxdata/telegraf/blob/master/docs/EXTERNAL_PLUGINS.md) that can be used with `execd` to run seamlessly with Telegraf.

sjwang90:
Due to inactivity I'm going to close the issue for now. If you'd like to submit this as an external plugin or look into re-writing the parser to reduce the code size we'd gladly take another look.

PR #6369, opened 2019-09-09
New ouputs for WhaTap server 

### LINKED ISSUE #6228: Add WhaTap output for application and server monitoring
### Proposal:
I want to add output plugin.
WhaTap that provides application and server monitoring.
Output plugin for delivering telegraf's collected data to WhaTap server.

https://www.whatap.io

API documentation
https://docs.whatap.io/kr/agent_telegraf/api.html

### Current behavior:
I am developing and testing WhaTap oupt plugin.
Similar to socket_writer. it's pass information to WhaTap over TCP.

### Desired behavior:
pull request. 

### Use case: 
We want to provide people using WhaTap service with various information that can be collected by Telegraf.

whatapdev:
### Required for all PRs:

Add WhaTap output for application and server monitoring. Closes #6228


plugins/outputs/whatap/DataOutputX.go
```diff
@@ -0,0 +1,225 @@
...
+
+func ToBytesBool(b bool) []byte {
+	if b {
+		return []byte{1}
+	} else {
+		return []byte{0}
+	}
+}
+
+func ToBytesShort(v int16) []byte {
```
danielnelson: There are function for serializing these types of values in https://golang.org/pkg/encoding/binary/.  For larger operations it may make sense to use https://golang.org/pkg/encoding/gob/, which essentially allows you to write a struct to the socket.  See if you can make use of either of these packages.

plugins/outputs/whatap/whatap.go
```diff
@@ -0,0 +1,372 @@
...
+		}
+		nbyteleft -= nbytethistime
+		pos += nbytethistime
+	}
+	return err
+
+}
+
+var table = [...]int64{
+	0x00000000, 0x77073096, 0xee0e612c, 0x990951ba, 0x076dc419,
```
danielnelson: What is this?

plugins/outputs/whatap/whatap.go
```diff
@@ -0,0 +1,372 @@
...
+
+		if err := w.send(0, dout.ToByteArray()); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+func (w *Whatap) send(code byte, b []byte) (err error) {
+	defer func() {
+		if r := recover(); r != nil {
```
danielnelson: Don't recover, allow Telegraf to close with error.  Make sure in your code you don't do anything that would cause a panic during normal operation.  We treat panics more similar to a C assert, programming errors only.

plugins/outputs/whatap/DataOutputX.go
jhlim-whatap: It is a library to support multi languages ​​(java, python, etc.). It is different from the standard serialize library specification and cannot be applied

plugins/outputs/whatap/whatap.go
jhlim-whatap: TCP 서버와 공용으로 사용하는 hash table 입니다. 캐싱또는 ID처럼 사용하는 값으로 서버와 동일한 로직으로 되어 있어야 합니다.

plugins/outputs/whatap/whatap.go
jhlim-whatap: recover has been modified to be disabled.

whatapdev:
I wrote about the question. Modifications are complete. 
Please tell us what needs to be done further.

whatapdev:
I'm curious how the review progresses.
Can you see the progress of the review?

plugins/outputs/whatap/DataOutputX.go
danielnelson: So this style is used so that it is more similar to the implementation you are using for other languages?  I do think we must switch to using the `binary` package here as much as possible, documenting when we cannot.  The `binary` package is well tested and uses more efficient options to serialize values.

plugins/outputs/whatap/whatap.go
danielnelson: So this is a random value shared with the server implementation?  Nobody would be able to figure this out, so if we are to include it then at a minimum it would need to be well documented in a code comment as well as in https://docs.whatap.io/kr/agent_telegraf/api.html

danielnelson:
Sorry we have a large backlog on pull requests, but I commented on some issues from before which needs further discussion.

whatapdev:
Thank you for answer.
We'll fix some of the things mentioned.

plugins/outputs/whatap/DataOutputX.go
whatapdev: Modified to use binary package.
```
func ToBytesShort(v int16) []byte {
	buf := make([]byte, 2)
	binary.BigEndian.PutUint16(buf[0:], uint16(v))
	return buf
}
```

plugins/outputs/whatap/whatap.go
whatapdev: I added a comment to the code for hash table.
Added a description of the hash format and a code description to the api documentation.
```

/**
 * The following table array is a hash lookup table.
 * Generate int32, int64 hash of string or byte array based on the table.
 *
 * The generated hash must be the same as the WhaTap server side.
 * (The same hash lookup table and function are on the WhaTap server.)
 * This is because the WhaTap server caches or indexes data
 * with that hash value.
 *
 * The HashBytes, HashStr function converts a string or byte array
 * to an int32 hash value.
 *
 * The Hash64Bytes, Hash64Str function converts a string or byte array
 * to an int64 hash value.
 */
var table = [...]int64{

```

whatapdev:
Reworked the two remaining issues. May I ask for a review?

whatapdev:
if removed request for review then How do I request a review again?

danielnelson:
I'll do a follow up review soon, thank you for your patience.

whatapdev:
Is there anything new about the review?

sjwang90:
Sorry for the delay, but we're currently pretty busy reviewing and building some other things for our next minor release in mid-March.  If we get a chance before then we'll definitely take a look.

whatapdev:
Please let me know if there are any other progress. 
I want to know when the review is complete and the source can be merged.

whatapdev:
I have no other progress, so I want to see how it goes.

sjwang90:
Would you be willing to submit this WhaTap Output plugin as an [external plugin](https://github.com/influxdata/telegraf/blob/master/EXTERNAL_PLUGINS.md) that can be used with `execd` to run seamlessly with Telegraf? Telegraf users would be able to use the plugin sooner by having it as an external plugin since it wouldn't have to go through the typical review process. 

Here is some [information on external plugins](https://github.com/influxdata/telegraf/blob/master/docs/EXTERNAL_PLUGINS.md) and using them with `execd`.

whatapdev:
I want to be included in the Telegraf release.
The reason is to use only configuration settings without additional installation like external plugins.

plugins/outputs/whatap/DataOutputX.go
```diff
@@ -0,0 +1,204 @@
+package whatap
```
srebhan, line 1: Can we please rename this file to more match "usual" names e.g. to `whatap_dataoutput.go` or similar?

plugins/outputs/whatap/DataOutputX.go
```diff
@@ -0,0 +1,204 @@
...
+	"bytes"
+	"encoding/binary"
+	"math"
+)
+
+const (
+	INT3_MIN_VALUE  = -8388608 /*0xff800000*/
+	INT3_MAX_VALUE  = 0x007fffff
+	LONG5_MIN_VALUE = -549755813888 /*0xffffff8000000000*/
+	LONG5_MAX_VALUE = 0x0000007fffffffff
```
srebhan, line 13: Do we need to export those?

plugins/outputs/whatap/DataOutputX.go
```diff
@@ -0,0 +1,204 @@
...
+)
+
+type DataOutputX struct {
```
srebhan, line 16: Do we need to export his structure?

plugins/outputs/whatap/DataOutputX.go
```diff
@@ -0,0 +1,204 @@
...
+	buffer  bytes.Buffer
+	written int
+}
+
+func NewDataOutputX() *DataOutputX {
+	in := new(DataOutputX)
+	return in
+}
```
srebhan, line 24: I think you should rather place that one-line body on the caller site.

plugins/outputs/whatap/DataOutputX.go
```diff
@@ -0,0 +1,204 @@
...
+		} else {
+			buff := []byte{254, 0, 0, 0, 0}
+			out.WriteBytes(SetBytesInt(buff, 1, int32(sz)))
+			out.WriteBytes(value)
+		}
+	}
+	return out
+}
+func (out *DataOutputX) WriteDecimal(v int64) *DataOutputX {
+
```
srebhan, line 61: Please remove those newlines at the begin of the function body. Only use newlines to structure the code.

plugins/outputs/whatap/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+  ## WhaTap license. Required
+  #license = "xxxx-xxxx-xxxx"
+
+  ## WhaTap project code. Required
+  #pcode = 1111
+
+  ## WhaTap server IP. Required
+  ## Put multiple IPs. ["tcp://1.1.1.1:6600","tcp://2.2.2.2:6600"]
+  #servers = ["tcp://1.1.1.1:6600","tcp://2.2.2.2:6600"]
```
srebhan, line 17: Please don't comment required options.
```suggestion
  ## WhaTap license. Required
  license = "xxxx-xxxx-xxxx"

  ## WhaTap project code. Required
  pcode = 1111

  ## WhaTap server IP. Required
  ## Put multiple IPs. ["tcp://1.1.1.1:6600","tcp://2.2.2.2:6600"]
  servers = ["tcp://1.1.1.1:6600","tcp://2.2.2.2:6600"]

```
and maybe you should use `project` instead of `pcode`?

plugins/outputs/whatap/whatap.go
```diff
@@ -0,0 +1,382 @@
...
+	//"github.com/influxdata/telegraf/internal"
+	"github.com/influxdata/telegraf/plugins/outputs"
+)
+
+const (
+	NETSRC_AGENT_ONEWAY = 10
+	VALUE_DECIMAL       = 20
+	VALUE_DOUBLE        = 40
+	VALUE_TEXT          = 50
+	VALUE_MAP           = 80
```
srebhan, line 23: I think this should be moved to the `DataOutputX` location and the write functions should directly write that type.

plugins/outputs/whatap/whatap.go
```diff
@@ -0,0 +1,382 @@
...
+)
+
+type Whatap struct {
+	License string        `toml:"license"`
+	Servers []string      `toml:"servers"`
+	Pcode   int64         `toml:"pcode"`
+	Timeout time.Duration `toml:"timeout"`
+	Oname   string
+	Oid     int32
+	Session TcpSession
```
srebhan, line 35: Please do not export this struct members.

plugins/outputs/whatap/whatap.go
```diff
@@ -0,0 +1,382 @@
...
+	"github.com/influxdata/telegraf/plugins/outputs"
+)
+
+const (
+	NETSRC_AGENT_ONEWAY = 10
+	VALUE_DECIMAL       = 20
+	VALUE_DOUBLE        = 40
+	VALUE_TEXT          = 50
+	VALUE_MAP           = 80
+	// 5633
```
srebhan, line 24: What does that comment tells us?

plugins/outputs/whatap/whatap.go
```diff
@@ -0,0 +1,382 @@
+package whatap
+
+import (
+	"fmt"
+	"log"
```
srebhan, line 5: Please do not use `log` but rather use the `telegraf.Logger` facility. If you add 
```golang
```
to your struct, it will automatically injected.

plugins/outputs/whatap/whatap.go
```diff
@@ -0,0 +1,382 @@
...
+	"net"
+	"os"
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/influxdata/telegraf"
+
+	//"github.com/influxdata/telegraf/internal"
```
srebhan, line 14: Please remove this if not needed.

plugins/outputs/whatap/whatap.go
```diff
@@ -0,0 +1,382 @@
...
+	VALUE_MAP           = 80
+	// 5633
+	PACK_TAG_COUNT = 0x1601
+)
+
+type Whatap struct {
+	License string        `toml:"license"`
+	Servers []string      `toml:"servers"`
+	Pcode   int64         `toml:"pcode"`
+	Timeout time.Duration `toml:"timeout"`
```
srebhan, line 32: Please use
```suggestion
	Timeout config.Duration `toml:"timeout"`
```
instead to allow suffices like `ms`, `s`, `m` or `h`.

plugins/outputs/whatap/whatap.go
```diff
@@ -0,0 +1,382 @@
...
+	Oname   string
+	Oid     int32
+	Session TcpSession
+}
+
+type TcpSession struct {
+	Client net.Conn
+	Dest   int
+}
```
srebhan, line 41: Why do you need this struct? Can't you simply place those two fields in the `WhatTap` struct?

plugins/outputs/whatap/whatap.go
```diff
@@ -0,0 +1,382 @@
...
+		w.Session.Dest = 0
+	}
+
+	addr := strings.SplitN(w.Servers[w.Session.Dest], "://", 2)
+	if len(addr) != 2 {
+		return fmt.Errorf("invalid address: %s", w.Servers[w.Session.Dest])
+	}
+	if addr[0] != "tcp" {
+		return fmt.Errorf("only tcp is supported: %s", w.Servers[w.Session.Dest])
+	}
```
srebhan, line 74: How  about using `net/url` here to do the parsing?

plugins/outputs/whatap/whatap.go
```diff
@@ -0,0 +1,382 @@
...
+
+	t := w.Timeout * time.Millisecond
+	if client, err := net.DialTimeout(addr[0], addr[1], t); err != nil {
+		return err
+	} else {
+		w.Session.Client = client.(*net.TCPConn)
+		log.Println("I! [outputs.whatap] Connected tcp to ", addr)
+	}
```
srebhan, line 82: You don't need an `else` clause here as you return in the `if`
```suggestion
	}
	w.Session.Client = client.(*net.TCPConn)
	w.Log.Infof("Connected tcp to %s", addr)
```

plugins/outputs/whatap/whatap.go
```diff
@@ -0,0 +1,382 @@
...
+	if w.Session.Client == nil {
+		return nil
+	}
+	err := w.Session.Client.Close()
+	w.Session.Client = nil
+	return err
+}
+
+func (w *Whatap) Description() string {
+	return "Configuration for WhaTap"
```
srebhan, line 98:
```suggestion
	return "Plugin to send metrics to a WhaTap server"
```

plugins/outputs/whatap/whatap.go
```diff
@@ -0,0 +1,382 @@
...
+	return sampleConfig
+}
+
+func (w *Whatap) Write(metrics []telegraf.Metric) error {
+	if len(metrics) == 0 {
+		return nil
+	}
+	if w.Session.Client == nil {
+		if err := w.Connect(); err != nil {
+			w.Close()
```
srebhan, line 111: Please tell the linter to not report unchecked error here
```suggestion
			//nolint:revive // We cannot handle close errors anyway.
			w.Close()
```

plugins/outputs/whatap/whatap.go
```diff
@@ -0,0 +1,382 @@
...
+			return err
+		}
+	}
+	for _, m := range metrics {
+		p := NewTagCountPack()
+		p.Pcode = w.Pcode
+		p.Oid = w.Oid
+		p.Category = fmt.Sprintf("%s%s", "telegraf_", m.Name())
```
srebhan, line 119: You could easily setup the struct directly here
```suggestion
		p := TagCountPack {
			Pcode: w.Pcode,
			Oid: w.Oid,
			Category: "telegraf_" + m.Name(),
			Tags: make(map[string]string),
			Data: make(map[string]interface{}),
		}
```

plugins/outputs/whatap/whatap.go
```diff
@@ -0,0 +1,382 @@
...
+		for k, v := range m.Fields() {
+			p.Put(k, v)
+		}
+		for k, v := range m.Tags() {
+			p.PutTag(k, v)
+		}
+		// Add Oname
```
srebhan, line 126: This comment is not adding any information.

plugins/outputs/whatap/whatap.go
```diff
@@ -0,0 +1,382 @@
...
+		p.PutTag("oname", w.Oname)
+		p.Time = m.Time().UnixNano() / int64(time.Millisecond)
```
srebhan, line 128: A comment here could help
```suggestion
		// Convert time to microseconds.
		p.Time = m.Time().UnixNano() / int64(time.Millisecond)
```

plugins/outputs/whatap/whatap.go
```diff
@@ -0,0 +1,382 @@
...
+
+		dout := NewDataOutputX()
+		// pack type
```
srebhan, line 131: Please tell _why_ you do something or give an overall description on what the intention of the code is if it isn't obvious.

plugins/outputs/whatap/whatap.go
```diff
@@ -0,0 +1,382 @@
...
+		dout.WriteShort(p.GetPackType())
+		p.Write(dout)
+
+		if err := w.send(0, dout.ToByteArray()); err != nil {
```
srebhan, line 135: The first parameter of `send` is always zero, so why not remove it and add it as a constant to the called function?

plugins/outputs/whatap/whatap.go
```diff
@@ -0,0 +1,382 @@
...
+	pos := 0
+	for 0 < nbyteleft {
+		nbytethistime := 0
+		// Set Deadline
+		err = w.Session.Client.SetWriteDeadline(time.Now().Add(
+			w.Timeout * time.Millisecond))
+		if err != nil {
+			log.Println("W! [outputs.whatap] cannot set tcp write deadline:",
+				err)
+		}
```
srebhan, line 162: I think this counts for a non-recoverable error
```suggestion
		deadline := time.Now().Add(w.Timeout)
		if err := w.Session.Client.SetWriteDeadline(deadline);  err != nil {
			return fmt.Errorf("cannot set write deadline: %v", err)
		}
```

plugins/outputs/whatap/whatap.go
```diff
@@ -0,0 +1,382 @@
...
+	out.WriteLong(w.Pcode)
+	out.WriteLong(Hash64Str(w.License))
+	out.WriteIntBytes(b)
+	// pack data
+	sendbuf := out.ToByteArray()
+
+	nbyteleft := len(sendbuf)
+	pos := 0
+	for 0 < nbyteleft {
+		nbytethistime := 0
```
srebhan, line 155:
```suggestion
	var pos int
	var nbytethistime int
	for 0 < nbyteleft {
```

plugins/outputs/whatap/whatap.go
```diff
@@ -0,0 +1,382 @@
...
+func Hash64Bytes(bytes []byte) int64 {
+	crc := uint64(0xffffffffffffffff)
+	sz := len(bytes)
+	for i := 0; i < sz; i++ {
+		b := bytes[i]
+		crc = (crc)>>8 ^ uint64(int32(table[uint8(crc^uint64(b))]))
+	}
+	crc = crc ^ 0xffffffffffffffff
+	return int64(crc)
+}
```
srebhan, line 270: Please move this to an own file (e.g. `whatap_hash.go`). What hash-algorithm is this? Is it a standard algorithm?

plugins/outputs/whatap/whatap.go
```diff
@@ -0,0 +1,382 @@
...
+		case float32, float64:
+			dout.WriteByte1(VALUE_DOUBLE)
+			dout.WriteDouble(v.(float64))
+		case string:
+			dout.WriteByte1(VALUE_TEXT)
+			dout.WriteText(v.(string))
+		default:
+		}
+	}
+}
```
srebhan, line 366: How about also moving this to an own file?

plugins/outputs/whatap/whatap.go
```diff
@@ -0,0 +1,382 @@
...
+		Oname:   hostname,
+		Oid:     HashStr(hostname),
+	}
+}
+func init() {
+	outputs.Add("whatap", func() telegraf.Output {
+		return newWhatap()
+	})
+
+}
```
srebhan, line 382: Please merge the two
```suggestion
}
func init() {
	hostname, _ := os.Hostname()
	outputs.Add("whatap", func() telegraf.Output {
		return &Whatap{
			Timeout: 60 * time.Second,
			Session: TcpSession{},
			Oname:   hostname,
			Oid:     HashStr(hostname),
		}
	})

}
```

plugins/outputs/whatap/whatap_test.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/testutil"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+var (
+	License = "x2tggtnopk2t9-z39dt59pe1pmjc-xipbnkb0ph6bn"
+	Server  = "121.166.140.134"
```
srebhan, line 19: Why do you need this?

plugins/outputs/whatap/whatap_test.go
```diff
@@ -0,0 +1,66 @@
...
+
+	metrics := []telegraf.Metric{testutil.TestMetric(1, "testerr")}
+
+	// close the socket to generate an error
+	lconn.Close()
+	w.Session.Client.Close()
+	err = w.Write(metrics)
+	require.Error(t, err)
+	assert.Nil(t, w.Session.Client)
+}
```
srebhan, line 66: The tests (as-is) require internet connection to reach the server above. This is not guaranteed in short-test mode, so you need to guard this here (grep for `short` in test files).

srebhan (changes requested):
Hey @whatapdev, thanks for contributing and sorry for taking so long to give your plugin a review!

Besides the comments in the code, I have a more general question: How about moving the data-conversions and hashing (`DataOutputX`, `TagCountPack` and `Hash`) to a library outpside of telegraf? This would reduce the code-size of this plugin tremendously and ensures that the code can co-exist with the server code as it seems both are closely coupled (e.g. due to the hash-table)... What do you think?

whatapdev:
All right. Some data transformation and hashing (DataOutputX, TagCountPack and Hash) will be handled by external libraries.

plugins/outputs/whatap/whatap.go
whatapdev: Edited and applied. We ask for confirmation.

plugins/outputs/whatap/whatap.go
whatapdev: Add and change necessary comments

plugins/outputs/whatap/whatap.go
whatapdev: Add and change necessary comments

plugins/outputs/whatap/DataOutputX.go
whatapdev: Moved to a library outpside of telegraf

plugins/outputs/whatap/DataOutputX.go
whatapdev: Moved to a library outpside of telegraf

plugins/outputs/whatap/DataOutputX.go
whatapdev: Moved to a library outpside of telegraf

plugins/outputs/whatap/DataOutputX.go
whatapdev: Moved to a library outpside of telegraf

plugins/outputs/whatap/DataOutputX.go
whatapdev: Moved to a library outpside of telegraf

plugins/outputs/whatap/README.md
whatapdev: Edited and applied. We ask for confirmation.

plugins/outputs/whatap/whatap.go
whatapdev: Moved to a library outpside of telegraf

plugins/outputs/whatap/whatap.go
whatapdev: Edited and applied. We ask for confirmation.

plugins/outputs/whatap/whatap.go
whatapdev: Edited and applied. We ask for confirmation.

plugins/outputs/whatap/whatap.go
whatapdev: Edited and applied. We ask for confirmation.

plugins/outputs/whatap/whatap.go
whatapdev: Edited and applied. We ask for confirmation.

plugins/outputs/whatap/whatap.go
whatapdev: Edited and applied. We ask for confirmation.

plugins/outputs/whatap/whatap.go
whatapdev: Edited and applied. We ask for confirmation.

plugins/outputs/whatap/whatap.go
whatapdev: Edited and applied. We ask for confirmation.

plugins/outputs/whatap/whatap.go
whatapdev: There is an initialization process inside the external library NewTagCountPack() function, so it is used without modification.

plugins/outputs/whatap/whatap.go
whatapdev: Edited and applied. We ask for confirmation.

plugins/outputs/whatap/whatap.go
whatapdev: Edited and applied. We ask for confirmation.

plugins/outputs/whatap/whatap.go
whatapdev: Edited and applied. We ask for confirmation.

plugins/outputs/whatap/whatap.go
whatapdev: Edited and applied. We ask for confirmation.

plugins/outputs/whatap/whatap.go
whatapdev: Moved to a library outpside of telegraf

plugins/outputs/whatap/whatap.go
whatapdev: Moved to a library outpside of telegraf

plugins/outputs/whatap/whatap_test.go
whatapdev: Edited and applied. We ask for confirmation.

plugins/outputs/whatap/whatap.go
whatapdev: Edited and applied. We ask for confirmation.

plugins/outputs/whatap/whatap_test.go
whatapdev: That code has been removed.

plugins/outputs/whatap/whatap.go
whatapdev: Moved to a library outpside of telegraf

plugins/outputs/whatap/whatap.go
whatapdev: Edited and applied. We ask for confirmation.

plugins/outputs/whatap/README.md
```diff
@@ -0,0 +1,21 @@
...
```
srebhan, line 17: As all options are mandatory you should not comment those option to directly indicate to the user that they need to fill those in
```suggestion
  ## WhaTap license. Required
  license = "xxxx-xxxx-xxxx"

  ## WhaTap project code. Required
  pcode = 1111

  ## WhaTap server IP. Required
  ## Put multiple IPs. ["tcp://1.1.1.1:6600","tcp://2.2.2.2:6600"]
  servers = ["tcp://1.1.1.1:6600","tcp://2.2.2.2:6600"]

```

plugins/outputs/whatap/whatap.go
```diff
@@ -0,0 +1,168 @@
...
+	Pcode   int64           `toml:"pcode"`
+	Timeout config.Duration `toml:"timeout"`
+	Log     telegraf.Logger `toml:"-"`
+	oname   string
+	oid     int32
+	conn    net.Conn
+	dest    int
+}
+
+var sampleConfig = `
```
srebhan, line 35: Please make this a constant.
```suggestion
const sampleConfig = `
```

plugins/outputs/whatap/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+```toml
+  ## You can create a project on the WhaTap site(https://www.whatap.io) 
+  ## to get license, project code and server IP information.
+
+  ## WhaTap license. Required
+  #license = "xxxx-xxxx-xxxx"
+
+  ## WhaTap project code. Required
+  #pcode = 1111
```
srebhan, line 13: Can we please use a more speaking name for the option 
```suggestion
  project_code = 1111

```

plugins/outputs/whatap/whatap.go
```diff
@@ -0,0 +1,168 @@
...
+)
+
+const (
+	NetSrcAgentOneway = 10
+)
+
+type Whatap struct {
+	License string          `toml:"license"`
+	Servers []string        `toml:"servers"`
+	Pcode   int64           `toml:"pcode"`
```
srebhan, line 26: You should use a more speaking name at least on the user-facing side
```suggestion
	Pcode   int64           `toml:"project_code"`
```

plugins/outputs/whatap/whatap.go
```diff
@@ -0,0 +1,168 @@
...
+		return fmt.Errorf("only tcp is supported: %s", w.Servers[w.dest])
+	}
+	client, err := net.DialTimeout(u.Scheme, u.Host, time.Duration(w.Timeout))
+	if err != nil {
+		return err
+	}
+
+	w.conn = client.(*net.TCPConn)
+	w.Log.Info("Connected ", u.String())
+	return nil
```
srebhan, line 73: Can you elaborate on what you are doing here? It seems like a type of iterating through the servers to find one to connect to? This might be easier and more clear when doing
```suggestion
	for _, server := range w.Servers {
		u, err := url.Parse(server)
		if err != nil {
			w.Log.Errorf("invalid address: %s", server)
			continue
		}
		if u.Scheme != "tcp" {
			w.Log.Errorf("only tcp is supported: %s", server)
			continue
		}
		client, err := net.DialTimeout(u.Scheme, u.Host, time.Duration(w.Timeout))
		if err != nil {
			w.Log.Errorf("connecting to %s failed: %v", server, err)
			continue
		}
  
		w.conn = client.(*net.TCPConn)
		w.Log.Info("Connected to %v", server)
		return nil
	}
	return fmt.Errorf("could not connect to any of the %d servers", len(w.Servers))
```
You might also want to check for correct schemes (i.e. the first two checks) in `Init()` and remove the invalid entries from the list to not trying to connect in the first place.

plugins/outputs/whatap/whatap.go
```diff
@@ -0,0 +1,168 @@
...
+	return sampleConfig
+}
+
+func (w *Whatap) Write(metrics []telegraf.Metric) error {
+	if len(metrics) == 0 {
+		return nil
+	}
+	if w.conn == nil {
+		if err := w.Connect(); err != nil {
+			_ = w.Close()
```
srebhan, line 99: Does this really make sense to close the connection if connecting failed?

plugins/outputs/whatap/whatap.go
```diff
@@ -0,0 +1,168 @@
...
+	"os"
+	"time"
+
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/config"
+	"github.com/influxdata/telegraf/plugins/outputs"
+
+	whatap_io "github.com/whatap/go-api/common/io"
+	whatap_pack "github.com/whatap/go-api/common/lang/pack"
+	whatap_hash "github.com/whatap/go-api/common/util/hash"
```
srebhan, line 16: Please do not use dashes as this seems alien in Golang. Either use `whatapIo` etc or shorten to `wio`, `wpack` and `whash`...

plugins/outputs/whatap/whatap.go
```diff
@@ -0,0 +1,168 @@
...
+			_ = w.Close()
+			return err
+		}
+	}
+	// Transform telegraf metrics to whatap protocol.
+	for _, m := range metrics {
+		p := whatap_pack.NewTagCountPack()
+		p.Pcode = w.Pcode
+		p.Oid = w.oid
+		p.Category = fmt.Sprintf("%s%s", "telegraf_", m.Name())
```
srebhan, line 108:
```suggestion
		p.Category = "telegraf_" + m.Name()
```

plugins/outputs/whatap/whatap.go
```diff
@@ -0,0 +1,168 @@
...
+		// Convert time to microseconds.
+		p.Time = m.Time().UnixNano() / int64(time.Millisecond)
+
+		dout := whatap_io.NewDataOutputX()
+		dout.WriteShort(p.GetPackType())
+		p.Write(dout)
+
+		if err := w.send(dout.ToByteArray()); err != nil {
+			return err
+		}
```
srebhan, line 126: Can we be sure that the returned error is recoverable? If not, the same data will be pushed in the next _flush_ iteration and we fail in exactly the same place leading to an output buffer that fills until it overflows. A non-recoverable error might e.g. be an invalid value or type in the metrics, while connection error might be recoverable.

plugins/outputs/whatap/whatap.go
```diff
@@ -0,0 +1,168 @@
...
+}
+func (w *Whatap) send(b []byte) (err error) {
+	// Transmits data in compliance with whatap protocol.
+	dout := whatap_io.NewDataOutputX()
+	dout.WriteByte(NetSrcAgentOneway)
+	dout.WriteByte(0)
+	dout.WriteLong(w.Pcode)
+	dout.WriteLong(whatap_hash.Hash64Str(w.License))
+	dout.WriteIntBytes(b)
+	sendbuf := dout.ToByteArray()
```
srebhan, line 138: I'm a bit puzzles. Isn't this already done in the `Write()` function? I thought you here get a byte-array that you just need to ship over the line...

If this is not the case, please add a corresponding function to the external library to abstract the protocol internals and keep it off-telegraf.

plugins/outputs/whatap/whatap.go
```diff
@@ -0,0 +1,168 @@
...
+
+		nbytethistime, err = w.conn.Write(sendbuf[pos : pos+nbyteleft])
+		if err != nil {
+			_ = w.Close()
+			return err
+		}
+		nbyteleft -= nbytethistime
+		pos += nbytethistime
+	}
+	return err
```
srebhan, line 157:
```suggestion
	return nil
```

plugins/outputs/whatap/whatap.go
```diff
@@ -0,0 +1,168 @@
...
+}
+func init() {
+	hostname, _ := os.Hostname()
+	outputs.Add("whatap", func() telegraf.Output {
+		return &Whatap{
+			Timeout: config.Duration(60 * time.Second),
+			oname:   hostname,
+			oid:     whatap_hash.HashStr(hostname),
```
srebhan, line 165: Can you please do this logic in an `Init()` function. This function is automatically called by telegraf after initializing all user-configured fields.

plugins/outputs/whatap/whatap_test.go
```diff
@@ -0,0 +1,80 @@
...
+func TestWhatapConnect(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping test in short mode.")
+	}
+	listener, err := net.Listen("tcp", "127.0.0.1:0")
+	require.NoError(t, err)
+
+	w := newWhatap()
+	addr := listener.Addr().String()
+	fmt.Println(addr)
```
srebhan, line 38: Please avoid using `fmt.Print*`.

plugins/outputs/whatap/whatap_test.go
```diff
@@ -0,0 +1,80 @@
...
+		Timeout: config.Duration(60 * time.Second),
+		Log:     testutil.Logger{},
+		oname:   hostname,
+		oid:     whatap_hash.HashStr(hostname),
+	}
+}
+func TestWhatapConnect(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping test in short mode.")
+	}
```
srebhan, line 32: Why can't this be executed in short-mode? As long as you do not leave localhost this should be ok.

srebhan (commented):
Very nice update @whatapdev! This looks much better. I've some comments left, but nothing outstanding. Looking forward to the next round.

plugins/outputs/whatap/README.md
whatapdev: Changed pcode to project_code.

plugins/outputs/whatap/whatap_test.go
whatapdev: Deleted that phrase

plugins/outputs/whatap/whatap_test.go
whatapdev: Deleted that phrase

plugins/outputs/whatap/whatap.go
whatapdev: Moved to Init() function

plugins/outputs/whatap/whatap.go
whatapdev: Moved to Init() function

plugins/outputs/whatap/whatap.go
whatapdev: Deleted that phrase.

plugins/outputs/whatap/whatap.go
whatapdev: Changed  to wio, wpack, whash

plugins/outputs/whatap/whatap.go
whatapdev: It has been modified to terminate the connection after outputting the log without returning err.

plugins/outputs/whatap/whatap.go
whatapdev: Updated.

plugins/outputs/whatap/whatap.go
whatapdev: It's not duplicate code. The write function creates transfer data.
In the send function, the transmitted data is composed of the body, and the header is added to complete the actually sent data.
Added comments.

plugins/outputs/whatap/whatap.go
```diff
@@ -0,0 +1,183 @@
...
+func (w *Whatap) send(b []byte) (err error) {
+	dout := wio.NewDataOutputX()
+	// Header : add the header before sending
+	dout.WriteByte(NetSrcAgentOneway)
+	dout.WriteByte(NetSrcAgentVersion)
+	dout.WriteLong(w.Pcode)
+	dout.WriteLong(whash.Hash64Str(w.License))
+	// Body : Set the converted whatap data(b []bytes) as the body
+	dout.WriteIntBytes(b)
+	sendbuf := dout.ToByteArray()
```
srebhan, line 131: In line 111-113 you are creating a `dout` only containing `p` as far as I do understand the code, then in `send` you create a new `dout` where you add a header and the the `dout` you created before. Why don't you create the **complete** `dout` (i.e. the second one) _before passing it into `send`? This would have been more intuitive and would group all `Write...` functions in one place...

plugins/outputs/whatap/whatap.go
```diff
@@ -0,0 +1,183 @@
...
+
+	wio "github.com/whatap/go-api/common/io"
+	wpack "github.com/whatap/go-api/common/lang/pack"
+	whash "github.com/whatap/go-api/common/util/hash"
+)
+
+const (
+	NetSrcAgentOneway  = 10
+	NetSrcAgentVersion = 0
+	Scheme             = "tcp"
```
srebhan, line 22: I don't think you need to export those constants, do you?

plugins/outputs/whatap/whatap.go
```diff
@@ -0,0 +1,183 @@
...
+func (w *Whatap) nextHost() string {
+	w.dest++
+	if w.dest >= len(w.Servers) {
+		w.dest = 0
+	}
+	return w.hosts[w.dest]
+}
+func (w *Whatap) Init() error {
+	if len(w.Servers) == 0 {
+		return fmt.Errorf("WhaTap server IP is Required")
```
srebhan, line 160: In Golang, errors should start with a lowercase letter. Maybe
```suggestion
		return fmt.Errorf("no WhaTap server IP configured")
```

srebhan: Maybe you should do this check _after_ validating the servers to make sure you do have at least _one_ valid host?

srebhan (commented):
We are close @whatapdev. Some comments left...

plugins/outputs/whatap/whatap.go
```diff
@@ -0,0 +1,183 @@
...
+		u, err := url.Parse(server)
+		if err != nil {
+			return fmt.Errorf("invalid address: %s", server)
+		}
+		if u.Scheme != "tcp" {
+			return fmt.Errorf("only tcp is supported: %s", server)
+		}
+		w.hosts = append(w.hosts, u.Host)
+	}
+	w.oname, _ = os.Hostname()
```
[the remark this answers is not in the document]
whatapdev: Edited and applied. We ask for confirmation. Checked error.

plugins/outputs/whatap/whatap.go
```diff
@@ -0,0 +1,183 @@
...
+	// Change and connect multiple servers sequentially.
+	host := w.nextHost()
+	client, err := net.DialTimeout(Scheme, host, time.Duration(w.Timeout))
+	if err != nil {
+		return err
+	}
+	w.conn = client.(*net.TCPConn)
+	w.Log.Info("Connected ", host)
+	return nil
+}
```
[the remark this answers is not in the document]
whatapdev: Edited and applied. We ask for confirmation.

plugins/outputs/whatap/whatap.go
whatapdev: Edited and applied. We ask for confirmation.

plugins/outputs/whatap/whatap.go
whatapdev: Edited and applied. We ask for confirmation.

plugins/outputs/whatap/whatap.go
whatapdev: Edited and applied. We ask for confirmation. I modified the WriteHeader function by creating it in an external library.

plugins/outputs/whatap/whatap.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+	if len(w.hosts) == 0 {
+		return fmt.Errorf("no WhaTap server IP configured")
+	}
+	if hn, err := os.Hostname(); err != nil {
+		w.oname = "unknown_host"
+	} else {
+		w.oname = hn
+	}
+	w.oid = whash.HashStr(w.oname)
```
srebhan, line 167: Wouldn't it be better to fail here? It's not a must, I'm just asking...

plugins/outputs/whatap/whatap.go
```diff
@@ -0,0 +1,176 @@
...
+			return fmt.Errorf("cannot set write deadline: %v", err)
+		}
+
+		nbytethistime, err = w.conn.Write(sendbuf[pos : pos+nbyteleft])
+		if err != nil {
+			return err
+		}
+		nbyteleft -= nbytethistime
+		pos += nbytethistime
+	}
```
srebhan, line 142: Wouldn't this be clearer
```suggestion
	for pos := 0; pos < len(sendbuf); {
		deadline := time.Now().Add(time.Duration(w.Timeout))
		if err := w.conn.SetWriteDeadline(deadline); err != nil {
			return fmt.Errorf("cannot set write deadline: %v", err)
		}

		nbytethistime, err := w.conn.Write(sendbuf[pos : len(sendbuf)])
		if err != nil {
			return err
		}
		pos += nbytethistime
	}
```

plugins/outputs/whatap/whatap_test.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+func newWhatap(addr string) *Whatap {
+	servers := make([]string, 0)
+	servers = append(servers, "tcp://"+addr)
+	w := &Whatap{
+		Timeout: config.Duration(60 * time.Second),
+		Log:     testutil.Logger{},
+		Servers: servers,
+	}
```
srebhan, line 22: I think this is too complicated, use
```suggestion
	w := &Whatap{
		Timeout: config.Duration(60 * time.Second),
		Log:     testutil.Logger{},
		Servers: []string{"tcp://"+addr},
	}
```
instead. I would also suggest to put this to the respective locations instead of making it a function...

plugins/outputs/whatap/whatap_test.go
```diff
@@ -0,0 +1,64 @@
...
+	require.NoError(t, err)
+	err = lconn.(*net.TCPConn).SetWriteBuffer(256)
+	require.NoError(t, err)
+
+	metrics := []telegraf.Metric{testutil.TestMetric(1, "testerr")}
+
+	err = lconn.Close()
+	require.NoError(t, err)
+
+	_ = w.Close()
```
srebhan, line 61: Please check the error here or put a comment on why you do not check it...

srebhan (commented):
Very nice @whatapdev! Only some minor things I'd like to see. The code becomes really beautiful IMO.

plugins/outputs/whatap/whatap.go
whatapdev: I have modified it according to your suggestion.

plugins/outputs/whatap/whatap.go
whatapdev: I have modified it according to your suggestion.

srebhan (approved):
Thanks @whatapdev for the changes! This looks really good now IMO!

whatapdev:
Thanks for review

sjwang90:
Hi @whatapdev team - we appreciate all your hard work on this PR over the years and addressing the recent reviews from @srebhan but we have decided to not include this plugin into Telegraf. Telegraf is getting more restrictive on the number of plugins that get included in the internal binary as we are growing many MBs every release. We are focusing on including plugins, especially output plugins, that cater to a very large user base. 

If you would like to turn this into an [external plugin](https://github.com/influxdata/telegraf/blob/master/EXTERNAL_PLUGINS.md), our team would gladly assist.

PR #6393, opened 2019-09-13
Require goplugin build flag to enable go plugin support

danielnelson:
It has come to my attention that the Go plugin support added in #6024 had an unexpected and hefty cost in terms of binary size, to the tune of about 19MB.

Despite this feature been released in 1.12, I propose we add a build flag and disable it on official Telegraf builds.  Based on my testing of this feature, it is quite difficult to get plugins working without building the plugin and Telegraf together on the same host, and the extra binary size is just too much for a feature that will be rarely used.

To build with Go plugin support add `-tags goplugin` to the build command:
```
go build -tags goplugin -ldflags "-w -s -X main.commit=9e3f918d -X main.branch=HEAD" ./cmd/telegraf
```

### Required for all PRs:

### DIFF
--- a/cmd/telegraf/telegraf.go
+++ b/cmd/telegraf/telegraf.go
@@ -10,9 +10,6 @@ import (
 	_ "net/http/pprof" // Comment this line to disable pprof endpoint.
 	"os"
 	"os/signal"
-	"path"
-	"path/filepath"
-	"plugin"
 	"runtime"
 	"strings"
 	"syscall"
@@ -21,6 +18,7 @@ import (
 	"github.com/influxdata/telegraf/agent"
 	"github.com/influxdata/telegraf/internal"
 	"github.com/influxdata/telegraf/internal/config"
+	"github.com/influxdata/telegraf/internal/goplugin"
 	"github.com/influxdata/telegraf/logger"
 	_ "github.com/influxdata/telegraf/plugins/aggregators/all"
 	"github.com/influxdata/telegraf/plugins/inputs"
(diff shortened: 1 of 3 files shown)


danielnelson:
cc @Timidger @adrianlzt @mtp401

adrianlzt:
Thanks for the update.
I was trying to use the plugin feature but it does not work with Windows so I return to using a normal plugin with a custom build.

mildebrandt:
@danielnelson I'm working on a way to build our internal plugins and package them into the telegraf docker container. Correct me if I'm wrong, but it sounds like the docker container won't include a build of telegraf that supports plugins. I was hoping to build on top of the existing container, insert my plugins/conf files, and repackage. With this PR, it sounds like I need to separately build/host/support a telegraf binary/container if I want plugin support.

Is that correct? If so, any chance of Influx making a separate docker container with a telegraf binary that supports plugins as part of the release cycle?

danielnelson:
That is true, you will need to make an entirely new docker container.

I'm not sure we would want to package a docker container with this feature, since there are a bunch of requirements that have to line up between the Telegraf build and the plugin build: Go compiler version, GOPATH, CGO_ENABLED, and potentially other details.  I wonder if it can be made compatible enough where it can be taken and reused without issue.

That said, I think we would be willing to add an AMD64 .deb/.rpm package with the tag enabled and you could also open a new issue for the docker image and we can see what sort of demand there is.

mildebrandt:
Thanks, a package that enables plugin support would also be welcome. I've opened this to request it: https://github.com/influxdata/telegraf/issues/7003

mildebrandt:
And I've opened this for the docker container: https://github.com/influxdata/influxdata-docker/issues/369

PR #6413, opened 2019-09-18
Go module support

zhangjianweibj:
use go module better than gopkg


glinton:
Thanks for this PR, but there is already a pull request in progress (#6118), feel free to add suggestions to that.

PR #6427, opened 2019-09-20
SFlow input plugin and associated parser

deanasheehan:
### Required for all PRs:

As an employee of Influx is this required? In any case, of course I will


deanasheehan:
I resolved the dependency issue for honnef.co/go/netdb by removing the need for that package.

Now noticing a few of my tests have dependency on port service name mappings of actual machine the test is executed on. I'll resolve.

Also likely I'm going to close this draft as I've decided I want to rebuild the parsing approach significantly - now that I have a full grasp on SFlow.

deanasheehan:
Not sure what the process is to put this on hold but I now have "Netflow" plugin in development and expect to bring sflow and netflow together into a single PR. There will be a little refactor of this sflow parser and input plugin to support the reuse and then the addition of the small addition of the netflow format decoder.

deanasheehan:
The Netflow additions are still in development but I have determined that 99% of this sFlow PR is still valid. I would appreciate an early review of this piece while I complete the Netflow additions.

Netflow is a reuse of most of the decoding machinery I built for sFlow with a very thin pair of 'parser' and 'input' plugin veneers. When it comes to update this PR with the Netflow additions that it is some file/package renames (to help separate the namespaces) and then some simple additions.

deanasheehan:
@danielnelson hoping you get a chance to review this for inclusion into next release. Both sFlow and Netflow are now there.

There is a "go generate" issue that I have avoided by moving the offending .go to .notgo files - there is a dependency issue that I need a little assistance to resolve but not pertinent to the functionality committed at this point.

danielnelson:
@deanasheehan This pull request is too large as is, we going to need to bring it in as smaller chunks.  Can you create a pull request with just the sFlow input but without any of the resolvers?  I don't think it will make sense to have a the parser as a data_format, so you can move the parser code into the input plugin.  After we add the input plugin without the resolvers, we can discuss how we want to go about adding the resolvers.

We'll take a look for the 1.14, but the upcoming 1.13 release window is already over.

deanasheehan:
@danielnelson - how odd, GitHub didn't have me 'watch' this automatically. I assumed it would, given it was my PR and you actually @me as well! Anyway, thanks to Russ I got a nudge to come look.

Yes I can break up into just sFlow and then send in Netflow after.
Combining sFlow plugin and parser into 1 - can do, keeping it separate does allow for someone to put sFlow parser behind Kafka ingestion (for example) but that is a slightly academic example.

Re hostname/iface resolution - what are your thoughts on an alternative approach?

danielnelson:
> Re hostname/iface resolution - what are your thoughts on an alternative approach?

It may depend on the type of lookup, we may want a mechanism for reverse dns lookups that can be used across more input plugins.  It may also make sense to have it and other, more complicated, types of lookups might be processors, although this raises issues about handling async processors.  I am currently on the fence about embedding the large tables directly into Telegraf.

deanasheehan:
@danielnelson re DNS resolution, as well as the SNMP iface name resolution, what I have currently works. Should we come up with a better and more general approach (cross plugin for example) in the future then great, we can refactor. For now I would say it is what it is.

deanasheehan:
@danielnelson per our discussion last week. I have taken out the Netflow parser and input plugin, leaving just the SFlow parser and input plugin. I have also removed IP and Interface asynchronous name resolution - as a first pass.

I have left in port->service name mapping as this is simpler synchronous mapping and removal would cause some painful test packet rework.

If we can design an asynchronous processor interface then we can add the IP and interface resolution back in as generic processors that can be applied in many situations. I would likely also extract the port->svc name into a processor as well at that time.

The dashboards that have been created as part of the associated template rely on human readable names to make the data useful/accessible so we need to repackage this behaviour before we can release.

Once we have that design, I can repackage the resolvers I have as async processors and then after that add contribute Netflow.

Happy to discuss an async processor contract but think the requirements are pretty clear and you'll have strong opinions on the design I'm sure.

danielnelson:
I'd like to have the port->service name mapping removed before getting started on a review.

I think it is best to report ports numerically by default.
- Any port->service mapping is lossy when multiple ports can be used for a service.
- If there is no service name mapping you would need to fallback to ports, resulting is a mixture of data types that will be difficult to work with.
- The mapping is not time series data (port 80 is always http) so there is a strong argument for leaving this as a display issue.
- Everywhere else we report as port numbers, to be consistent the sflow plugin should do the same.
- Ultimate plan is to move it to a processor anyway.

We can create a processor for this, but I'd like to have it use `net.LookupPort` instead of embedding the table into Telegraf.  This function calls the systems native lookup function, such as`getservbyport`, which can be customized and will avoid an unnecessary increase in binary size.  This won't require any new processor type, the current interface will work fine for this.

deanasheehan:
@danielnelson will do.

_Frustrated that I've only just seen your comment - once again, I'm not clear why GitHub isn't sending my notifications - I get notifications from Telegraf repo all the time - except for when I actually need it._

deanasheehan:
@danielnelson - have removed the protodb portno->svcname enrichment.

Please review and if you could @deanasheehan in any comments (to resolve whatever issue I'm having with notifications against this PR) I would be grateful.

Thanks
-Dean

dapryor:
@deanasheehan 
Were you planning on making a new PR to add in the netflow support you removed from the original PR?

deanasheehan:
@sjwang90 - are the team taking ownership of my original work here? Please see @dapryor ask above.

sjwang90:
Hi @dapryor, yes we will get the NetFlow portion added in a separate PR. I'll tag you when it's up and feel free to provide any feedback

dapryor:
Thank you both.  Looking forward to the Netflow plugin! I have been enjoying the product and will provide feedback as I see issues! I have not seen any yet :)

Issue #4046, opened 2018-04-19
Input: CloudWatch - Partial Wildcard or Regexp support?


mooyoul:
## Feature Request

Opening a feature request kicks off a discussion.

### Proposal: 

Support Partial Wildcard (e.g. `*prod*`, like glob) or RegExp support (e.g. `/prod/`) to filter out unwanted metrics

For example, below configuration will pull production metrics only:

Using RegExp:

```toml
[[inputs.cloudwatch]]
  namespace = "AWS/Lambda"
  [[inputs.cloudwatch.metrics]]
    names = ["Invocations", "Errors", "Duration"]
    [[inputs.cloudwatch.metrics.dimensions]]
      name = "FunctionName"
      value = "/prod/"
```

... or partial wildcard (globbing): 

```toml
[[inputs.cloudwatch]]
  namespace = "AWS/Lambda"
  [[inputs.cloudwatch.metrics]]
    names = ["Invocations", "Errors", "Duration"]
    [[inputs.cloudwatch.metrics.dimensions]]
      name = "FunctionName"
      value = "*prod*"
```

### Current behavior:

There's no easy way to select production-only metrics without specifying them.

### Desired behavior:

Users can filter metrics using Glob or RegExp

### Use case:

Since our architecture of our application is microservices, We have 300+ lambda functions and we have 3 stages (production, staging, development) and number of functions are growing.
We replicate cloudwatch metrics to our InfluxDB using Telegraf with Cloudwatch Input Plugin to detect anomaly metrics (e.g. Latency Spikes). When something goes wrong, we receive alarms from Bosun.

Anyway currently filtering production resources is so hard. everytime we release new "service" to  production, we need to change bunch line in the telegraf configuration to filter non-production metrics.

danielnelson:
We currently have support only for `*` in the dimension value, but I think this can be extended to support all globs.  This method would require following the code path that runs `ListMetrics`.

jivanpuri:
We are also waiting for this feature. Is there any plan to include this in the near future?

danielnelson:
> in the near future

There is no work being done on this by us right now, but we have some other fairly large improvemnts coming to this plugin in #5544.

glinton:
Can you show an `aws list metrics` command with input that produces your desired output?

CpuID:
> Can you show an `aws list metrics` command with input that produces your desired output?

[...]

PR #6433, opened 2019-09-23
Release 1.12 cloudwatch filter

hflamboauto1:
### Required for all PRs:

### DIFF
--- a/plugins/inputs/cloudwatch/README.md
+++ b/plugins/inputs/cloudwatch/README.md
@@ -139,6 +139,36 @@ Then 2 metrics will be output:
 If the `AvailabilityZone` wildcard dimension was omitted, then a single metric (name: `p-example`)
 would be exported containing the aggregate values of the ELB across availability zones.
 
+
+Example2:
+```
+[[inputs.cloudwatch]]
+  period = "1m"
+  interval = "5m"
+
+  [[inputs.cloudwatch.metrics]]
+    names = ["Latency"]
+
+    ## Dimension filters for Metric (optional)
+    [[inputs.cloudwatch.metrics.dimensions]]
+      name = "LoadBalancerName"
+      value = "*"
+      value_excludes = ["example"]
+
+    [[inputs.cloudwatch.metrics.dimensions]]
+      name = "AvailabilityZone"
+      value = "*"
+```
+
+If the following ELBs are available:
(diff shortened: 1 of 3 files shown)


hflamboauto1:
relates to https://github.com/influxdata/telegraf/issues/4046

danielnelson:
You should be able to use the `tagdrop` option to achieve the same result:
```toml
[[inputs.cloudwatch]]
  period = "1m"
  interval = "5m"

  [[inputs.cloudwatch.metrics]]
    names = ["Latency"]

    [[inputs.cloudwatch.metrics.dimensions]]
      name = "LoadBalancerName"
      value = "*"

    [[inputs.cloudwatch.metrics.dimensions]]
      name = "AvailabilityZone"
      value = "*"

  [[inputs.cloudwatch.tagdrop]]
    LoadBalancerName = ["example"]
```

Give this a shot and let me know how it goes.

hflamboauto1:
@danielnelson 

Thanks for your thoughts and working example. 

Yes that's right, it could be done like that if the number of LoadBalancerName(s) to drop is small.  It becomes impractical (on my particular case at least) as I have a "huge" AWS Account with a considerable amount of LoadBalancers to be excluded. 

Imagine having to specify 20 or 30 or even more LoadBalancerName(s) tags to drop explicitly by specifying exact names! That becomes not very handy specially in dynamic environments.

With the way you propose  then I guess that would no need for tag drop at all and just register to the exactly LoadBalancerName(s) of interest in similar fashion, but again this becomes impractical for same previous reasons as I have lets say 30 that can be of interested and another 20 of non interest for a particular domain ... like  qa | prod | staging | test ...  or whatever.

Maybe there's another option or I misunderstood something that would allow me to achieve that with ease but did not found and some kind of filtering options based on LoadBalancerName match help a lot and it solves issue in my case.

danielnelson:
You can use globs/wildcards in all of the filtering options, we don't have regular expression support right now, but in my experience globs are usually enough to get the job done for hostnames.  We use [github.com/gobwas/glob](https://github.com/gobwas/glob) syntax.
```
  [[inputs.cloudwatch.tagdrop]]
    LoadBalancerName = ["prod.*", "staging.*"]
```

hflamboauto1:
Lovely .. I think I can go with that .. shameless completely missed that. A big thanks.

Issue #6445, opened 2019-09-25
Output exec plugin does not use Serialise batch function


morfien101:
### Relevant telegraf.conf:

```toml
[[output.exec]]
  command = ["cat"]
  data_format = "json"
```

### System info:

Any system
Telegraf 1.12.2 (git: HEAD 8b4c9a0e)

### Steps to reproduce:

Send more than a single metric into the output stream

### Expected behavior:

Data format should be parsable using the format specified.
 
### Actual behavior:

Data is given as if it was a single metric over many lines.

PR #6446, opened 2019-09-25
Change the Exec output to use BatchSerialize()

morfien101:
### Required for all PRs:

Solves #6445

When the exec output serialises the metrics it does it one by one and writes it to the buffer. Error checking is done on the serialising but not the writing into the buffer. So there is no reason to not use the batch serialiser.

When developing against the output exec I expected the output to follow the rules of the data format that I specified. For my use case I am using JSON and expected a list of metrics but was unfortunately greeted with a multiple lines of single metrics, making it extremely difficult to render the metrics again.

Sample of bad output:

```json
{"fields":{"bytes_recv":0,"bytes_sent":0,"drop_in":0,"drop_out":0,"err_in":0,"err_out":0,"packets_recv":0,"packets_sent":0},"name":"net","tags":{"host":"pickle","interface":"br-0ac7079a5349"},"timestamp":1569418417000}
{"fields":{"bytes_recv":526400,"bytes_sent":1924156,"drop_in":0,"drop_out":0,"err_in":0,"err_out":0,"packets_recv":5289,"packets_sent":13564},"name":"net","tags":{"host":"pickle","interface":"docker0"},"timestamp":1569418417000}
{"fields":{"bytes_recv":0,"bytes_sent":0,"drop_in":0,"drop_out":0,"err_in":0,"err_out":0,"packets_recv":0,"packets_sent":0},"name":"net","tags":{"host":"pickle","interface":"br-349b964a2f38"},"timestamp":1569418417000}
```

This PR removes the looping over the metrics and rather just use the batch serialise function.

### DIFF
--- a/plugins/outputs/exec/exec.go
+++ b/plugins/outputs/exec/exec.go
@@ -67,13 +67,11 @@ func (e *Exec) SampleConfig() string {
 // Write writes the metrics to the configured command.
 func (e *Exec) Write(metrics []telegraf.Metric) error {
 	var buffer bytes.Buffer
-	for _, metric := range metrics {
-		value, err := e.serializer.Serialize(metric)
-		if err != nil {
-			return err
-		}
-		buffer.Write(value)
+	serializedMetrics, err := e.serializer.SerializeBatch(metrics)
+	if err != nil {
+		return err
 	}
+	buffer.Write(serializedMetrics)
 
 	if buffer.Len() <= 0 {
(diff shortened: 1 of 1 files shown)


danielnelson:
This is a change of behavior of this newly introduced feature, but I think we are still within the grace period.  We should try to always use SerializeBatch on any new code, and I'll add a comment officially deprecating the use of Serialize.

PR #6462, opened 2019-09-28
Build official packages with Go 1.13.1

danielnelson:
Drop CI testing for Go 1.10 and 1.11.

### Required for all PRs:


plugins/outputs/prometheus_client/prometheus_client_tls_test.go
```diff
@@ -1,114 +0,0 @@
-package prometheus_client_test
```
glinton: Was this intentional?

danielnelson: Yes, these tests do not work with Go 1.13 and essentially test that the Go TLS library works.

PR #6468, opened 2019-09-30
Remove shared package level vars from sqlserver and mysql input plugins

GeorgeJahad:
### Required for all PRs:

The current implementations of mysql and sqlserver input plugins don't support running multiple times with different configurations because they use shared package level vars

I've just revised the package level var's to either declare them as const's or move them into the instance struct.  Also added tests, both integration and unit, that correspond to those changes.


GeorgeJahad:
looks like i need to run go fmt.  will do that soon.

GeorgeJahad:
thank you sir!

PR #6470, opened 2019-10-01
Stackdriver output: Set startTime to timestamp - 1

Legogris:
Our cumulative Stackdriver metrics complete broke overnight this weekend. Instead of as previously resulting in one-minute intervals, we now have single timeSeries values extending from `T - 2*52 weeks` to `T`.

My guess is that Google has changed the supported timewindow for writing cumulative time series from very short to 104 weeks (~2 years).

I am not 100% certain if the approach here (startTime = timestamp - 1 second), but it looks like it's resolving the problem for us after running it for a bit.

### Required for all PRs:

### DIFF
--- a/plugins/outputs/stackdriver/README.md
+++ b/plugins/outputs/stackdriver/README.md
@@ -11,6 +11,8 @@ Metrics are grouped by the `namespace` variable and metric key - eg: `custom.goo
 
 Additional resource labels can be configured by `resource_labels`. By default the required `project_id` label is always set to the `project` variable.
 
+For cumulative metric [intervals](https://cloud.google.com/monitoring/api/ref_v3/rest/v3/TimeInterval), the start time is by default set to the start time of the telegraf process. To get periodic intervals, the `cumulative_interval_seconds` parameter can be set to define the number of seconds between each interval.
+
 ### Configuration
 
 ```toml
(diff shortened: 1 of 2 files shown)


Legogris:
This does not fix the issue, but semantics for startTime for cumulative values seem to indeed have been changed/broken in stackdriver now.

danielnelson:
What if we switched to using the Telegraf start time for all metrics?
```
StartTime = time.Now().Unix()
```

Legogris:
> What if we switched to using the Telegraf start time for all metrics?
> 
> ```
> StartTime = time.Now().Unix()
> ```

This looks closer to defined behavior, but still doesn't produce the same (from my side intended) behavior as previously.

I made an updated approach, what do you think @danielnelson? Basically the default is what you suggested, with the option of setting a newly introduced `cumulative_interval_seconds` parameter on the stackdriver output plugin, defining the length of intervals for cumulative metrics.

Legogris:
Cleaned up a bit, should be ready for review. Been running for a while with `cumulative_interval_seconds = 60` (with `agent_interval = 1m`)

danielnelson:
I'd prefer if we can avoid adding a new option, what issue are you seeing with the static timestamp?

Legogris:
> I'd prefer if we can avoid adding a new option, what issue are you seeing with the static timestamp?

My understanding is that the start time should be reset according to the intended interval of ones metric, which would make it equal to the `interval`. An alternative to introducing an option would be to just make it equal to the agent interval (which is what we're doing anyway so I have no problem with that), but I am also thinking others may have different requirements which would warrant the addition of the new option. I was also thinking it might be a bit too magic to inherit this form the agent configuration, but maybe it isn't?

Have been trying different configurations, and resetting the timestamp once a minute is the only configuration I found so far that produces the intended behavior (equal to running with the static timestamp  of `1` up until ~10 days ago).

kirecek:
Hello,

is there any update/suggestion on this? Looks like comulative metrics are still broken in the latest version due to StartTime value.

srebhan:
@Legogris any news here?

powersj:
Hi,

Thanks!

PR #6490, opened 2019-10-07
feat(processors): flux processor

affo:
This PR requires George's one to be merged: https://github.com/influxdata/telegraf/pull/6118

This adds a Flux processor to Telegraf.

### Required for all PRs:


affo:
@danielnelson I rebased this PR off master, but there where conflicts that I couldn't solve. I kept master and I have the diffs with me.

srebhan:
@affo this is really cool. Could you please restart this from the current master!? Like you save your core code somewhere and start a new branch from master with the backup-ed core-code!? Please force-push to this remote branch to update the PR.

srebhan:
@affo ping...

srebhan:
@affo Are you still interested in pushing this forward?

affo:
Hey, @srebhan !

Sorry for the latency. Yeah, I also think this is really cool.

I created this PR when I was working @ Influxdata and I left it as it is.

Anyways, I think I can get back to it.  
Just, I need to find some time to make it through.

Would this sound good to you, or is there any urgency?

srebhan:
Sounds good. No urgency here, it's only cool stuff that deserves to be taken on. :-)

srebhan:
@affo can you give a rough estimate on when you will be able to work on this!?

affo:
@srebhan

I removed @GeorgeMac commit as it was no longer needed, and fixed the code not to use `MetricBuilder`.

This is the current situation:

 - upgrading flux to 0.106.0 (latest) is not that easy, because I think that building flux requires some extra step that should be included in telegraf's CI;
 - leaving flux to 0.46 does not work because of a clash with some k8s lib and protobuf: `panic: mismatching message name: got k8s.io.kubernetes.pkg.watch.versioned.Event, want github.com/ericchiang.k8s.watch.versioned.Event`
 - downgrading to 0.44.1 brings back to protobuf 1.3.x and removes the above error, but the `replace ...uuid...` in Telegraf's `go.mod` makes Flux build fail.

Which do yo think is the next step? @srebhan

plugins/processors/flux/README.md
```diff
@@ -0,0 +1,49 @@
...
+```
+
+When the debug mode is active, the script does not run, and the output is saved to the specified path.
+
+### Configuration:
+```toml
+# # Runs a flux script to process inputs and produce outputs
+# [[processors.flux]]
+#   ## Path to a Flux script.
+#   path = "path/to/script.flux"
```
srebhan, line 25: You might also want to allow to specify the script using a multi-line option!?

plugins/processors/flux/flux.go
```diff
@@ -0,0 +1,342 @@
...
+
+func (p *Flux) SampleConfig() string {
+	return sampleConfig
+}
+
+func (p *Flux) Description() string {
+	return description
+}
+
+func (p *Flux) init() error {
```
srebhan, line 62: Please make this
```suggestion
func (p *Flux) Init() error {
```
which is called automatically by telegraf after parsing and initializing the config options.

plugins/processors/flux/flux.go
```diff
@@ -0,0 +1,342 @@
...
+		p.Log.Errorf("Error on flux processor init: %v", err)
+		return nil
+	}
+
+	tbls, err := getTables(metrics...)
+	if err != nil {
+		p.Log.Errorf("Could not extract tables from metrics: %v", err)
+		return nil
+	}
+	if len(p.Debug) > 0 {
```
srebhan, line 90: Let's play save and change this to
```suggestion
	if p.debugFile != nil {
```

plugins/processors/flux/flux.go
```diff
@@ -0,0 +1,342 @@
...
+		encoded, err := getCSV(tbls)
+		if err != nil {
+			p.Log.Errorf("Error in encoding CSV: %v", err)
```
srebhan, line 93: Please leave out "Error" here
```suggestion
			p.Log.Errorf("Encoding CSV failed: %v", err)
```

plugins/processors/flux/flux.go
```diff
@@ -0,0 +1,342 @@
...
+			p.Log.Errorf("Error while writing to debug file: %v", err)
+		}
+		return nil
+	}
+	max := time.Unix(0, 0)
+	for _, m := range metrics {
+		if m.Time().After(max) {
+			max = m.Time()
+		}
+	}
```
srebhan, line 106: Why do you need this? Metrics should always have a time, shouldn't they?

plugins/processors/flux/flux.go
```diff
@@ -0,0 +1,342 @@
...
+	provider, err := tables.NewProvider(tbls)
+	if err != nil {
+		p.Log.Errorf("Internal error while creating table provider: %v", err)
+		return nil
+	}
+	deps := Deps{
+		Tables: provider,
+		Max:    max,
+	}
```
srebhan, line 115: Do you have to do it in every `Apply()`? I guess we can do it once in `Init()` and reuse it...

srebhan: Oh I see you need the tables that depend on the metrics right? Can you somehow inject the metrics later?

plugins/processors/flux/flux.go
```diff
@@ -0,0 +1,342 @@
...
+	tbls := make([]flux.Table, 0)
+	builders := execute.NewGroupLookup()
+	getBuilder := func(gk flux.GroupKey) (execute.TableBuilder, bool) {
+		var tb execute.TableBuilder
+		v, found := builders.Lookup(gk)
+		if !found {
+			tb = execute.NewColListTableBuilder(gk, new(memory.Allocator))
+			builders.Set(gk, tb)
+		} else {
+			tb = v.(execute.TableBuilder)
```
srebhan, line 158: Please check the type-assertion as we panic otherwise if it's not true.

plugins/processors/flux/flux.go
```diff
@@ -0,0 +1,342 @@
...
+		return err
+	}
+	p.program = lang.CompileAST(astPkg, time.Now())
+	return nil
+}
+
+func getTables(metrics ...telegraf.Metric) ([]flux.Table, error) {
+	tbls := make([]flux.Table, 0)
+	builders := execute.NewGroupLookup()
+	getBuilder := func(gk flux.GroupKey) (execute.TableBuilder, bool) {
```
srebhan, line 151: So this function groups the metric by the given key. Is my understanding correct? Could you please spend a small comment for all non-flux devs. :-)

docs/LICENSE_OF_DEPENDENCIES.md
```diff
@@ -23,6 +23,7 @@ following works:
 - github.com/amir/raidman [The Unlicense](https://github.com/amir/raidman/blob/master/UNLICENSE)
 - github.com/antchfx/xmlquery [MIT License](https://github.com/antchfx/xmlquery/blob/master/LICENSE)
 - github.com/antchfx/xpath [MIT License](https://github.com/antchfx/xpath/blob/master/LICENSE)
+- github.com/apache/arrow/go/arrow [Apache License 2.0](https://github.com/apache/arrow/blob/master/LICENSE.txt)
```
srebhan, line 26: Most of the entries you make are not even remotely used by your code. Why did you add them?

srebhan (commented):
@affo I've skimmed through the code. See my comments. For the version question I've asked Influx-guys as I cannot decide this... :-) However, to me it looks like moving the flux-code to the flux repository might relax the situation wouldn't it?

srebhan:
@affo can you please check if your version problem benefits from #8937!?

srebhan:
@affo first of all let me thank you again for spending your time looking at stuff that you are no longer involved with!

Regarding the version/dependency question. I think it doesn't make sense to use an older version of _flux_ as the charm would be that people can use _flux_ in both telegraf and the database. However, looking at the complexity (build-process, generation, dependencies etc) that comes with adding the current version of _flux_ makes me skeptical if this is maintainable at all. Additionally, we already have a scripting language (starlark) for processing the metrics, so IMO the effort is huge compare to the gain. I thus propose that we do not proceed further with flux __for now__. If there is need for having it in we can still look at the issue later.
Does this sound ok for you @affo?

affo:
@srebhan Yes, it sounds good to me!

No problem at all, it was fun to have a look again at this codebase :)

I should also look into startlark then!

If you need any further change in the future, do not hesitate to ping me!

Cheers

srebhan:
Closing this as agreed on due to the vast amount of dependencies this would pull in. Please use the starlark processor instead. If you find use-cases where _flux_ is a _must_ please open an issue and reference this PR.

Issue #5019, opened 2018-11-22
Add Processor(s) that combine multiple tags


noodle-rubrik:
## Feature Request

There are many ways to rewrite tags, convert fields to tags, or to extract tag(s) from a tag value, but as far as I can tell, there is no way to combine the values of two tags into one.

### Proposal:

A new processor could be added that simply combines N tags with a given separator. The Regex operator is flexible enough to process tags before they're combined, or to do something else with the resulting tag value.

```
[[processors.combine]]
  tags = ["tagA", "tagB"]
  delimiter = ":"
  dest = "AB"
```

### Use case: [Why is this important (helps with prioritizing requests)]

This is a pretty niche use case I imagine, but we have just excluded the "container_name" tag from our 
docker_container_cpu measurement, because it contained irrelevant information that was leading to high cardinality. It would have been a perfect replacement for our existing dashboards to have the option to replace container_name with a combination of io.kubernetes.container.name and io.kubernetes.pod.name, but that doesn't seem to be possible using the current set of processors.

danielnelson:
I understand removing a tag if it is causing high cardinality and it is not required for series uniqueness, but if you already have the 2 tags separately why not leave them separate?  I don't think you will gain much from combining them into a single tag since you will still have the same number of series.  For example, here you will have 2 series after combining them:


```
foo,first=a,second=x value=42
foo,first=b,second=y value=42
```

```
foo,combined=a:x value=42
foo,combined=b:y value=42
```

noodle-rubrik:
In our case, it would have just been simpler to replace the data inside the container_name with something roughly equivalent, without having to go change it in all of our dashboards and metrics that reference container_name. We were also rewriting pod names matching a certain regex to be non-unique at the same time, so constructing the container_name out of the container and pod tags would have let us do the regex just once instead of twice.

More generally, you can already extract two tags from one tag value using regex with result_key, so it's just surprising to me that you can't do the opposite.

vadella:
Combining 2 fields would allow me to combine a separate date and time field to a datetime. My source looks like this:
```
<ROWDATA>
<ROW Nr="1" Date="2022/12/22" Time="00:06:59" Aad="0" Agr="0" AFct="Value"/>
<ROW Nr="2" Date="2022/12/22" Time="00:06:59" Aad="33" Agr="1" AFct="System"/>
<ROW Nr="3" Date="2022/12/22" Time="00:16:48" Aad="0" Agr="0" AFct="Value"/>
[...]

PR #6494, opened 2019-10-08
Support multiple tags for RabbitMQ routing tag topic

RobMalvern:
### Required for all PRs:

I have added this functionality for my company. If it is useful to the community I will update the README and add unit tests as appropriate

Now the routing_tag can take the following form: host.objectname
Which will be translated into TESTPC1.Processor (for example).


glinton:
I'm not sure I understand the benefit of stripping out a tag, can you elaborate and provide a clear use case please?

RobMalvern:
I'll try and explain better :-)

Currently a single tag from a measurement can be used as an ampq topic (this is still possible with this modification).

If using a topic exchange in ampq, using more tags to create multiple topics allows more routing options.

I'll explain my use case to clarify:
I have a number of machines that I'm monitoring. So I'm using two tags to create my ampq topics.
The first tag is the host name, the second tag is a custom "Level" tag which is either Support, or Development.
In the config file this takes the form: Host.Level

So in ampq, using a variety of routing keys I can:
- Subscribe to all the metrics from a single host, e.g. PC-1.#
- Or subscribe to all support metrics, e.g *.Support.# (I use this send support specific data to a custom support application, that only shows information the support team can act on immediately)
- Or subscribe to all metrics (support and development): *

Thanks for your time!

plugins/outputs/amqp/amqp.go
```diff
@@ -210,9 +210,23 @@ func (q *AMQP) Close() error {
...
+		var routingTags = strings.Split(q.RoutingTag, ".")
+		var validTags = make([]string, 0, len(routingTags))
+		k := 0
+		for _, tag := range routingTags {
+			if tag != "" {
+				key, ok := metric.GetTag(tag)
+				if ok {
+					// extend the slice for the new key
+					validTags = validTags[0 : len(validTags)+1]
+					validTags[k] = key
```
glinton: Change this to `validTags = append(validTags, key)`, don't change line 214, but remove `k`s initialization.

danielnelson:
This looks interesting and useful but I think we should implement the feature a bit differently.

We could have a processor that can create new tags/fields based on the metric, I think it would make sense to use Go's built-in template package for this.  Output plugins would only support routing via a single tag, we wouldn't need to include this type of logic directly in each output, and it would be more easy to introduce new or custom styles.


Here is how I imagine it might look:

```toml
[[processors.template]]
  tag = "topic"
  template = "{{.Tags.host}}.{{.Tags.level}}"

[[outputs.amqp]]
  topic_tag = "topic"
```

The processor would make this type of transformation:
```diff
- cpu,host=localhost,level=prod time_idle=42
+ cpu,topic=localhost.prod,host=localhost,level=prod time_idle=42
```

RobMalvern:
Hi @danielnelson, an approach similar to this was suggested in #5019, and rejected. 
If this now looks like the best way forward then I'm happy to have a bash at implementing it (I'm new to Go so it might take me a little while)

danielnelson:
Thanks for linking the previous issue.  With this use case I feel more comfortable adding a processor that can combine values.  cc @noodle-rubrik

Would be great if you could work on it.  I've thought about writing a template serializer which would be similar, and I think we will need an adapter type to pass into [Template.Execute](https://golang.org/pkg/text/template/#Template.Execute).  This is because the `telegraf.Metric` isn't template friendly with the multi-valued functions, and also to restrict access to some of the modifying functions.  It expect it would look something like this:
```
type TemplateMetric struct {
	metric telegraf.Metric
}

func (m *TemplateMetric) Measurement() string {
	panic("not implemented")
}

func (m *TemplateMetric) Tag(key string) string {
	panic("not implemented")
}

func (m *TemplateMetric) Field(key string) interface{} {
	panic("not implemented")
}

func (m *TemplateMetric) Time() time.Time {
	panic("not implemented")
}
```

RobMalvern:
FYI I am looking at this again. Last month was a crunch month at work, so I didn't have time to look at this.

RobMalvern:
So I've made a bit of progress - I actually have a unit test which is passing!

The shape of the configuration is currently not correct, it looks like this: 
```
[[processors.template]]
  tag = "topic"
  template = "{{ index .Tags \"hostname\" }}.{{ index .Tags \"level\" }}"
```

My understanding is that the TemplateMetric would be an intermediate struct to prevent this and give a cleaner configuration. However I'm not quite sure how that would work, so I would appreciate some guidance there.
Apart from that work still to do, I've put this early commit in to make sure I'm on the right track.

Thanks very much for your time!

plugins/processors/template/template.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+type TemplateProcessor struct {
+	Tag      string `toml:"tag"`
+	Template string `toml:"template"`
+}
+
+func (r *TemplateProcessor) Apply(in ...telegraf.Metric) []telegraf.Metric {
+	// create template
+	tmpl, err := template.New("test").Parse(r.Template)
```
danielnelson: Since we only need to do this once, let's move it into a Init() function.  This gets called by Telegraf on startup, in your tests you will need to do it manually:
```
func (r *TemplateProcessor) Init() error {
}
```

plugins/processors/template/template.go
```diff
@@ -0,0 +1,38 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	// for each metric in "in" array
+	for _, metric := range in {
+		var b strings.Builder
+		//newM := TemplateMetric{metric}
```
danielnelson: When you pass this into Execute, make sure to pass a pointer.  Either `&TemplateMetric{metric}` or `&newM` below.

plugins/processors/template/template.go
```diff
@@ -0,0 +1,38 @@
...
+		err := tmpl.Execute(&b, metric)
+		if err != nil {
+			panic(err)
+		}
+
+		metric.AddTag(r.Tag, b.String())
+	}
+
+	// convert/wrap metric in TemplateMetric
+	// convert TemplateMetric back to metric?
```
danielnelson: You won't need to do this, `return in` will be fine.

plugins/processors/template/template_metric.go
```diff
@@ -0,0 +1,26 @@
...
+type TemplateMetric struct {
+	metric telegraf.Metric
+}
+
+func (m *TemplateMetric) Measurement() string {
+	return m.Measurement()
+}
+
+func (m *TemplateMetric) Tag(key string) string {
+	return m.Tag(key)
```
danielnelson:
```
str, _ := m.metric.GetTag(key)
return str
```

plugins/processors/template/template_test.go
```diff
@@ -0,0 +1,51 @@
...
+	result := tmp.Apply(m1)
+
+	w.Close()
+	out, _ := ioutil.ReadAll(r)
+	os.Stdout = rescueStdout
+
+	t.Logf("Captured: %s", out) // prints: Captured: Hello, playground
+	resultTaglist := result[0].TagList()
+
+	assert.True(t, contains(resultTaglist, "topic", "localhost.debug"))
```
danielnelson: Easiest way to setup the tests is using `testutil.MustMetric` and `testutil.RequireMetricsEqual`:

https://github.com/influxdata/telegraf/blob/master/plugins/processors/pivot/pivot_test.go

plugins/processors/template/template.go
RobMalvern: I tried removing the AddTag call and the unit test now fails. Presumably the template Execute has no way of adding a Tag.
Or is that something I should add to TemplateMetric?

RobMalvern:
I believe I have corrected all your review points.
With the exception of removing the AddTag usage in Apply - this didn't seem to work.

Should we keep the direct usage of the Template in the configuration? It seems a bit clunky having to escape the quotes: "{{.Tag \"hostname\"}}.{{ .Tag \"level\" }}"

danielnelson:
Agreed it is a bit of a pain with the quotes, though with a bit of TOML knowledge a user can use single quotes for the toml string, we could use that in the sample config:
```toml
template = '{{.Tag "hostname"}}.{{ .Tag "level" }}'
```

I'm not sure how we could avoid direct use of the template in the configuration though, without inventing our own format.  If you have an idea let me know but switching the quotes automatically I think will open a can of worms.

Don't forget to fill in the functions I left unimplemented in `template_metric.go`.

RobMalvern:
I've added a few more tests - please let me know if there needs to be more. And if the documentation is clear enough.

Thanks for your time!

RobMalvern:
FYI the test failure is for an input module I've not modified: 

> FAIL	github.com/influxdata/telegraf/plugins/inputs/cisco_telemetry_gnmi

danielnelson:
Looking good, can you add a README.md file?  You can base it off the [pivot processor README](https://github.com/influxdata/telegraf/tree/master/plugins/processors/pivot/README.md).

RobMalvern:
Done - sorry I missed that!

danielnelson:
Thanks @RobMalvern, I've merged this and it will be part of 1.14.0 as well as the [nightly builds](https://github.com/influxdata/telegraf#nightly-builds) soon.

I also added a few more examples to the README, I think this plugin will need more than the usual processor since templates can be used in so many ways.

RobMalvern:
Thanks Daniel!

And thanks for your patience, this was my first contribution to “proper” open source 😊

________________________________

PR #6496, opened 2019-10-08
Report collection time, not gather time in statsd input

### LINKED ISSUE #6495: Statsd plugin reports metrics with the incorrect time stamp
### Relevant telegraf.conf:

```toml
[[inputs.statsd]]
  protocol = "udp"
  service_address = ":8125"
  percentiles = [50.0, 90.0, 99.0, 99.9, 99.95, 100.0]
  metric_separator = "_"
  percentile_limit = 1000
```

### System info:
```
> telegraf --version
Telegraf 1.12.1 (git: HEAD 8de41160)

> uname -a
Linux raspberrypi 4.19.57-v7l+ #1244 SMP Thu Jul 4 18:48:07 BST 2019 armv7l GNU/Linux
```
### Steps to reproduce:

1. enable statsd telegraf plugin
2. send constantly increasing value
3. calculate the rate with `NON_NEGATIVE_DERIVATIVE`

### Expected behavior:
get a constant flat line as the value changes by the same amount.

### Actual behavior:
the value is lumpy as the reporting time doesn't line up with the collection time.
Screen Shot 2019-10-08 at 10 32 02 AM
The collection interval can be adjusted to make the issue a little less of a problem. This is what it looks like when the interval is adjusted to `2s` vs `5s`:
Screen Shot 2019-10-08 at 10 49 10 AM

### Additional info:

I will have a PR shortly, but from my initial investigation it looks like the statsd doesn't report the metrics with the time they are collected, but instead they are reported with the time that they are aggregated (not sure of the terminology). This leads to periods of time that can hold more or fewer sets of metrics, which causes the rate to be incorrect.

For example, if a stated client reports every 5 seconds, and the telegraf instance is set up with a data collection interval of `5s`, there will exist intervals that contain no new metrics, and intervals that contain more then 1 metric. so instead of the rate being constant, the rate can vary between 0, 2x the constant, and the constant.

DBarney:
### Required for all PRs:

- [√] Signed [CLA](https://influxdata.com/community/cla/).
- [√] Associated README.md updated. (not applicable?)
- [√] Has appropriate unit tests.

Resolves #6495.
Fixed the statsd plugin to store when metrics are collected, so that they can be reported with the correct time stamp.

### DIFF
--- a/plugins/inputs/statsd/statsd.go
+++ b/plugins/inputs/statsd/statsd.go
@@ -150,30 +150,35 @@ type metric struct {
 	additive   bool
 	samplerate float64
 	tags       map[string]string
+	tm         time.Time
 }
 
 type cachedset struct {
 	name   string
 	fields map[string]map[string]bool
 	tags   map[string]string
+	tm     time.Time
 }
 
 type cachedgauge struct {
 	name   string
 	fields map[string]interface{}
 	tags   map[string]string
+	tm     time.Time
 }
 
 type cachedcounter struct {
 	name   string
 	fields map[string]interface{}
 	tags   map[string]string
+	tm     time.Time
 }
 
 type cachedtimings struct {
 	name   string
 	fields map[string]RunningStats
 	tags   map[string]string
+	tm     time.Time
 }
 
 func (_ *Statsd) Description() string {
@@ -246,7 +251,6 @@ func (_ *Statsd) SampleConfig() string {
 func (s *Statsd) Gather(acc telegraf.Accumulator) error {
 	s.Lock()
 	defer s.Unlock()
-	now := time.Now()
 
 	for _, m := range s.timings {
 		// Defining a template to parse field names for timers allows us to split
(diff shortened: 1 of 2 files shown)


plugins/inputs/statsd/statsd.go
```diff
@@ -746,14 +751,16 @@ func (s *Statsd) aggregate(m metric) {
...
+		counter, ok := s.counters[m.hash]
 		if !ok {
-			s.counters[m.hash] = cachedcounter{
+			counter = cachedcounter{
 				name:   m.name,
 				fields: make(map[string]interface{}),
 				tags:   m.tags,
 			}
 		}
+		counter.tm = tm
```
glinton: Nitpick, but set `tm` inline for these:
```go
s.counters[m.hash] = cachedcounter{
  name:   m.name,
  fields: make(map[string]interface{}),
  tags:   m.tags,
  tm:     tm,      
}
```

DBarney: I originally did that, however I realized that would only set the time when the `cached*` structs were created. We need the time to be updated even if we don't have a new structure.

glinton: ooh yes, good call

danielnelson:
@DBarney Thanks for the pull request, if I'm reading it correctly then this would adjust the timestamp to the time of the last value within the aggregation window.

Since statsd reports only aggregates, not raw event data, I think the current behavior of setting the timestamp to the aggregation time makes more sense than reporting the time of the last sample.  Most importantly we are really attempting to match the https://github.com/statsd/statsd server, and I believe they are using a similar method.

If you want to store raw events in a similar way, it might make more sense to send to the `socket_listener` plugin configured to receive UDP using InfluxDB line protocol, this way there is no aggregation applied before storage.

PR #6504, opened 2019-10-09
Add lang parameter to OpenWeathermap input plugin

### LINKED ISSUE #6042: lang parameter for openweathermap
Hi,

That would be great if you could add a "lang" parameter to the openweathermap plugin in order to grab a localized description field as explained here: https://openweathermap.org/current#multi

Thanks

reimda:
closes #6042 

New fields: condition_description, condition_icon
New tags: condition_id, condition_main

### Required for all PRs:


plugins/inputs/openweathermap/openweathermap.go
```diff
@@ -91,26 +102,16 @@ func (n *OpenWeatherMap) Gather(acc telegraf.Accumulator) error {
...
 	}
 
-	units := n.Units
 	switch n.Units {
-	case "imperial", "standard":
-		break
+	case "imperial", "standard", "metric":
 	default:
-		units = defaultUnits
+		return fmt.Errorf("unknown units: %s", n.Units)
```
glinton: No change required, but this may break someone's dependency on this feature, in this case we'll prefer less automagic.

plugins/inputs/openweathermap/openweathermap.go
```diff
@@ -91,26 +102,16 @@ func (n *OpenWeatherMap) Gather(acc telegraf.Accumulator) error {
 		n.client = client
 	}
 
-	units := n.Units
 	switch n.Units {
-	case "imperial", "standard":
-		break
+	case "imperial", "standard", "metric":
 	default:
-		units = defaultUnits
```
glinton: Remove the definition for `defaultUnits`

plugins/inputs/openweathermap/openweathermap.go
```diff
@@ -253,53 +253,63 @@ func gatherWeatherUrl(r io.Reader) (*Status, error) {
...
+		fields := map[string]interface{}{
+			"cloudiness":   e.Clouds.All,
+			"humidity":     e.Main.Humidity,
+			"pressure":     e.Main.Pressure,
+			"rain":         e.Rain.Rain3,
+			"temperature":  e.Main.Temp,
+			"wind_degrees": e.Wind.Deg,
+			"wind_speed":   e.Wind.Speed,
+		}
+		tags := map[string]string{
```
glinton: Move this back out of the loop so it doesn't get allocated each iteration.

plugins/inputs/openweathermap/openweathermap.go
```diff
@@ -311,7 +321,23 @@ func init() {
 		return &OpenWeatherMap{
```
glinton: Don't forget to set `Lang` to `"en"` to match the config example.

plugins/inputs/openweathermap/openweathermap_test.go
```diff
@@ -293,38 +294,46 @@ func TestForecastGeneratesMetrics(t *testing.T) {
 		testutil.MustMetric(
 			"weather",
 			map[string]string{
-				"city_id":  "2988507",
-				"forecast": "3h",
-				"city":     "Paris",
-				"country":  "FR",
+				"city_id":        "2988507",
```
glinton: No change required. I'm sorta surprised we let `city_id` be a tag and not a field. Must have added `city` later.

plugins/inputs/openweathermap/README.md
```diff
@@ -53,16 +60,20 @@ https://openweathermap.org/city/2643743
...
+    - condition_icon
 
 
 ### Example Output
 
 ```
-> weather,city=San\ Francisco,city_id=5391959,country=US,forecast=* cloudiness=40i,humidity=72i,pressure=1013,rain=0,sunrise=1559220629000000000i,sunset=1559273058000000000i,temperature=13.31,visibility=16093i,wind_degrees=280,wind_speed=4.6 1559268695000000000
-> weather,city=San\ Francisco,city_id=5391959,country=US,forecast=3h cloudiness=0i,humidity=86i,pressure=1012.03,rain=0,temperature=10.69,wind_degrees=222.855,wind_speed=2.76 1559271600000000000
-> weather,city=San\ Francisco,city_id=5391959,country=US,forecast=6h cloudiness=11i,humidity=93i,pressure=1012.79,rain=0,temperature=9.34,wind_degrees=212.685,wind_speed=1.85 1559282400000000000
+> weather,city=San\ Francisco,city_id=5391959,condition_id=800,condition_main=Clear,country=US,forecast=*,host=debian cloudiness=1i,condition_description="clear sky",condition_icon="01d",humidity=35i,pressure=1012,rain=0,sunrise=1570630329000000000i,sunset=1570671689000000000i,temperature=21.52,visibility=16093i,wind_degrees=280,wind_speed=5.7 1570659256000000000
```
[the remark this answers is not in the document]
reimda: ok.  Thanks for the omit_hostname tip

plugins/inputs/openweathermap/openweathermap.go
reimda: Yes.  If you had a typo in the units setting in your conf it would fall back to metric without telling the user.

plugins/inputs/openweathermap/openweathermap.go
reimda: ok

plugins/inputs/openweathermap/openweathermap.go
reimda: oops, thanks

plugins/inputs/openweathermap/openweathermap.go
reimda: I'll change it to leave out the lang query param if there's no lang setting in the config.  The OWM api already defaults to English if there's no query param.

glinton: Either way. Seems easier to just set it here in the struct you're returning, vs if not empty logic.

plugins/inputs/openweathermap/README.md
```diff
@@ -42,6 +47,8 @@ https://openweathermap.org/city/2643743
   - tags:
     - city_id
     - forecast
+    - condition_id
```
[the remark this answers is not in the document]
glinton: > It's part of OWM's API and someone familiar with it would think it's missing if we don't include it

It's quite common for telegraf to not create a metric from everything provided by an api. Some things just don't make sense as time series data.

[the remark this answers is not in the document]
glinton: Regarding the human readable vs id: telegraf has standardized enum "codes" to be stored as fields and the string representations to be stored as tags. One example is the `kube_inventory` plugin (`state` and `state_code`):
https://github.com/influxdata/telegraf/blob/817c9a69a94060aece50889c27e7c821a36a6530/plugins/inputs/kube_inventory/pod.go#L55-L63

>I would rather select ID=802

Very much agree. I'm not sure where/why you would do the other option. Both are still possible in either configuration from what I understand, but in influx, you can only `group by` tags.

plugins/inputs/openweathermap/openweathermap.go
```diff
@@ -73,12 +77,17 @@ func (n *OpenWeatherMap) Description() string {
...
-	if err != nil {
-		return err
+	var (
+		wg   sync.WaitGroup
+		strs []string
+		err  error
+	)
+
+	if n.baseURL == nil {
+		n.baseURL, err = url.Parse(n.SiteURL)
```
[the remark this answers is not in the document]
reimda: done.  I had to add a bunch of n.Init() in the tests because it's not called automatically called there.

plugins/inputs/openweathermap/openweathermap.go
```diff
@@ -73,12 +77,17 @@ func (n *OpenWeatherMap) Description() string {
...
-	var wg sync.WaitGroup
-	var strs []string
-
-	base, err := url.Parse(n.BaseUrl)
-	if err != nil {
-		return err
+	var (
+		wg   sync.WaitGroup
+		strs []string
+		err  error
```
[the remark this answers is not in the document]
reimda: Moving url formatting to a new function helped shorten it.  Moving initialization to a new Init() function will help too.

plugins/inputs/openweathermap/openweathermap.go
```diff
@@ -23,20 +23,21 @@ const (
...
-	defaultUnits           string        = "metric"
 )
 
 type OpenWeatherMap struct {
 	AppId           string            `toml:"app_id"`
 	CityId          []string          `toml:"city_id"`
+	Lang            string            `toml:"lang"`
 	Fetch           []string          `toml:"fetch"`
-	BaseUrl         string            `toml:"base_url"`
+	SiteURL         string            `toml:"base_url"`
```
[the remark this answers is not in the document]
reimda: I undid the name change from BaseUrl to SiteURL.  I'll name the *url.URL parsed version lowercase baseUrl.

plugins/inputs/openweathermap/openweathermap_test.go
```diff
@@ -492,3 +517,22 @@ func TestBatchWeatherGeneratesMetrics(t *testing.T) {
...
+func TestFormatURL(t *testing.T) {
+	b, _ := url.Parse("http://foo.com")
+	n := &OpenWeatherMap{
+		AppId:   "appid",
+		Units:   "units",
+		Lang:    "lang",
+		baseURL: b,
+	}
+
+	t.Run("path 1", func(t *testing.T) {
```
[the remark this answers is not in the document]
reimda: I think this was just left over from when Chris was walking me through various ways to unit test in Go.  I've removed the t.Run()

plugins/inputs/openweathermap/openweathermap_test.go
```diff
@@ -492,3 +517,22 @@ func TestBatchWeatherGeneratesMetrics(t *testing.T) {
...
+		s := n.formatURL("/data/2.5/forecast", "12345")
+		if want := "http://foo.com/data/2.5/forecast?APPID=appid&id=12345&lang=lang&units=units"; s != want {
+			t.Errorf("FormatOwnUrl: got %s, want %s", s, want)
```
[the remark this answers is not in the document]
reimda: done.  That is simpler

plugins/inputs/openweathermap/openweathermap_test.go
```diff
@@ -492,3 +517,22 @@ func TestBatchWeatherGeneratesMetrics(t *testing.T) {
...
+		}
+	})
+
+	//t.Errorf("bad things happened")
```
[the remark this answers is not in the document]
reimda: oops

plugins/inputs/openweathermap/README.md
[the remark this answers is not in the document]
reimda: The influxdb schema design docs have some advice on choosing tags vs fields (https://docs.influxdata.com/influxdb/v1.7/concepts/schema_and_data_layout/). It says to make it a tag if it's commonly queried or used in a group by.  Only tags are indexed, so querying on fields is slow.

After reading those docs I think I'd like to leave id and main as tags and description as a field.  I imagine querying on and grouping by conditions will be common, so main and id should be tags.  If you want the general condition like snowy days, use main="snow".  If you want a specific condition like all days with freezing rain, use id="511".  I think description isn't going to be commonly queried because it's localized and too long, so it should be a field.

I think the difference between this and kube_inventory is that kube_inventory's state is short (one word), only has three values, and isn't localized.  That makes state easier to use than state_code, so it fits the tags criteria better.

I'd be fine leaving out icon but I don't think it's hurting anything.  It doesn't give you much over id and main for grouping or querying, so if it stays it should be a field.  It's only really useful if you build custom ui and need a weather conditions icon.  I'm leaning towards leaving it in and letting people filter it out if they don't want it.

reimda:
> I take it OWM's default `unit` is "metric" if none is specified? My only hesitation doing it this way is that it gives them the control, if they change their default settings, telegraf user's are affected..
> 
> No change needed unless "metric" isn't their default `unit`

I checked on OWM's default unit and actually it isn't metric.  To avoid changing behavior for telegraf people upgrading, I added back the telegraf metric default.  There's still a check for expected units so you won't accidentally get metric if you misspelled standard or imperial like you did before.

plugins/inputs/openweathermap/openweathermap.go
```diff
@@ -310,8 +293,54 @@ func init() {
...
+	v := url.Values{
+		"id":    []string{city},
+		"APPID": []string{n.AppId},
+	}
+
+	if n.Units != "" {
+		v["units"] = []string{n.Units}
+	}
+
+	if n.Lang != "" {
```
glinton: Do this in one of your init functions (set a default in `init`, or check and set in `Init`)

plugins/inputs/openweathermap/openweathermap.go
```diff
@@ -310,8 +293,54 @@ func init() {
...
+	return nil
+}
+
+func (n *OpenWeatherMap) formatURL(path string, city string) string {
+	v := url.Values{
+		"id":    []string{city},
+		"APPID": []string{n.AppId},
+	}
+
+	if n.Units != "" {
```
glinton: This check was already done in `Init` and set to a value, so this would always be included.
Including it is preferred as it allows us keep the control over the defaults and not rely on their API to keep the same default setting telegraf did/does.

plugins/inputs/openweathermap/openweathermap.go
reimda: done

plugins/inputs/openweathermap/openweathermap.go
reimda: done, thanks

PR #6520, opened 2019-10-13
Add input plugin for QSC Q-SYS products

zjstraus:
Q-SYS is an audio and control processor used in many large-scale audiovisual installations. This PR adds support for its JSONRPC API to gather system status and any data exposed by the programmer.

### Required for all PRs:


sjwang90:
Due to the specific use case for this plugin - this plugin would be a good candidate for an [external plugin](https://github.com/influxdata/telegraf/blob/master/docs/EXTERNAL_PLUGINS.md). Would you be willing to submit this plugin as an [external plugin](https://github.com/influxdata/telegraf/blob/master/EXTERNAL_PLUGINS.md) that can be used with `execd` to run seamlessly with Telegraf? Telegraf users would be able to use the plugin sooner by having it as an external plugin since it wouldn't have to go through the typical review process.

PR #6522, opened 2019-10-14
feat(tail): add quiet flag to squash errors of non-existent files

### LINKED ISSUE #6518: [inputs.tail] Add flag to ignore files if not found and not give error
## Feature Request
a flag to ignore if file not present, and not throw error in telegraf logs

### Current behavior:
Throws error for file not present

### Desired behavior:
Does not throw error and fail silently

### Use case: 
I don't want to create different configuration for each server because we are using ansible to deploy everything to servers
And I don't want to clutter Elasticsearch with error logs that are not needed

rawkode:
Closes #6518

### Required for all PRs:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## v1.13 [unreleased]
 
+#### Release Notes
+
+- Official packages are built with Go 1.13.1.
+
 #### New Inputs
 
 - [azure_storage_queue](/plugins/inputs/azure_storage_queue/README.md) - Contributed by @mjiderhamn


danielnelson:
Let's make this error a debug level log message instead.  This will make exact paths and glob paths consistent in how they log errors.

rawkode:
@danielnelson Updated

plugins/inputs/tail/tail.go
```diff
@@ -158,8 +158,9 @@ func (t *Tail) tailNewFiles(fromBeginning bool) error {
 					Pipe:      t.Pipe,
 					Logger:    tail.DiscardingLogger,
 				})
+
 			if err != nil {
-				t.acc.AddError(err)
+				t.Log.Errorf("Failed to tail file (%s): %s", filepath, err.Error())
```
danielnelson: I think we want `t.Log.Debugf`, also minor style nit, for the error use "%v" and err:
```
t.Log.Debugf("Failed to tail file (%s): %v", filepath, err)
```

rawkode: I'm worried that many people won't have debug logging enabled and if there's a genuine mistake (typo?) in their config and they're running Telegraf, hiding this message behind debug would mean they lose a lot of time trying to work out what is wrong.

danielnelson: This is fine then by me, as we were essentially already logging at error level, but of course you will have to set the log level for all of Telegraf to `quiet` as we don't support per module log levels yet.

Issue #2546, opened 2017-03-18
Support Telegraf being run manually just once and then exit (e.g. in a cron job or over ssh).


Robpol86:
## Feature Request

Allow Telegraf to run once and exit. Like the `-test` option but have it write to output plugins vs just stdout. Maybe a `-once` command line option or setting `agent.interval` to "0" or "0s". Currently setting it to "0s" causes a crash:

```
2017-03-17T23:53:57Z I! Agent Config: Interval:0s, Quiet:false, Hostname:"raspberrypi", Flush Interval:10s 
2017-03-17T23:53:57Z E! FATAL: Input [inputs.disk] panicked: non-positive interval for NewTicker, Stack:
goroutine 21 [running]:
github.com/influxdata/telegraf/agent.panicRecover(0x117dba40)
	/home/ubuntu/telegraf-build/src/github.com/influxdata/telegraf/agent/agent.go:89 +0x98
panic(0xabe760, 0x1140a5a8)
	/usr/local/go/src/runtime/panic.go:458 +0x454
time.NewTicker(0x0, 0x0, 0x11505f40)
	/usr/local/go/src/time/tick.go:23 +0xc8
github.com/influxdata/telegraf/agent.(*Agent).gatherer(0x1140be40, 0x11542040, 0x117dba40, 0x0, 0x0, 0x11542300)
	/home/ubuntu/telegraf-build/src/github.com/influxdata/telegraf/agent/agent.go:117 +0x214
github.com/influxdata/telegraf/agent.(*Agent).Run.func3(0x114da190, 0x1140be40, 0x11542040, 0x11542300, 0x117dba40, 0x0, 0x0)
	/home/ubuntu/telegraf-build/src/github.com/influxdata/telegraf/agent/agent.go:379 +0x80
created by github.com/influxdata/telegraf/agent.(*Agent).Run
	/home/ubuntu/telegraf-build/src/github.com/influxdata/telegraf/agent/agent.go:380 +0xa70
...
```

### Current behavior:

Telegraf only supports "run once" with the `-test` argument, but that doesn't write to output plugins (e.g. InfluxDB), just stdout. Omiting the `[agent]` section in the config or `agent.interval` option (or setting it to "0" or "0s") either results in the default agent interval or a crash.

### Desired behavior:

Ability to run once Telegraf manually on demand and have it exit as soon as it's done writing to the database.

### Use case:

[...]

PR #6537, opened 2019-10-17
feat: add -run-once option to exit after a single collection

rawkode:
### Required for all PRs:

Decided to have a pop at #2546

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -17,13 +17,15 @@ import (
 
 // Agent runs a set of plugins.
 type Agent struct {
-	Config *config.Config
+	Config  *config.Config
+	RunOnce bool
 }
 
 // NewAgent returns an Agent for the given Config.
-func NewAgent(config *config.Config) (*Agent, error) {
+func NewAgent(config *config.Config, runOnce bool) (*Agent, error) {
 	a := &Agent{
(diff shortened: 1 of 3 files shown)


rawkode:
Managed to clean this up a little more :+1:

agent/agent.go
```diff
@@ -17,13 +17,15 @@ import (
...
 // Agent runs a set of plugins.
 type Agent struct {
-	Config *config.Config
+	Config  *config.Config
+	RunOnce bool
 }
 
 // NewAgent returns an Agent for the given Config.
-func NewAgent(config *config.Config) (*Agent, error) {
+func NewAgent(config *config.Config, runOnce bool) (*Agent, error) {
```
danielnelson: The implementation of this isn't something I would accept, adding booleans as trailing parameters is not allowed and it pushes a lot of new complexity into the agent.

The interface to this should be a new function, similar to what was done with the Test function.

Implementation might be shared, but it ought to guarantee a single Gather, single Aggregation, and as many writes are are required to send it all.  So, I'm not sure if sharing much would work out.

danielnelson:
Let's reopen the pull request when we have something closer to what we would merge.

PR #6563, opened 2019-10-22
Ensure metrics generated are correct in ping plugin using "native"

### LINKED ISSUE #6521: Result code is not set correctly with ping and method = "native"
Hi,
telegraf version: telegraf-1.12.2

I am monitoring 3000 IPs from a single telegraf config file, using method "native". PFB telegraf configuration for ping:
[agent]
  interval = "120s"

[[inputs.ping]]
urls = ["**********","*****", "10.10.10.10"]
method = "native"
count = 3
timeout = 10.0

below is the output of this configuration:
[root@N1VL-PA-INFL01 ~]# telegraf --config /etc/telegraf/telegraf.conf --test | grep -i 10.10.10.10
2019-10-14T08:00:23Z I! Starting Telegraf 1.12.2
> ping,IP=10.14.124.23,host=N1VL-PA-INFL01,url=10.10.10.10 average_response_ms=1.437718,maximum_response_ms=2.466695,minimum_response_ms=0.660697,packets_received=3i,packets_transmitted=3i,percent_packet_loss=0,result_code=0i,standard_deviation_ms=0.758515,ttl=63i 1571040027000000000

where this IP does not exist, and result_code should not be 0:

[root@N1VL-PA-INFL01 ~]# ping -c 3 -W 10 10.10.10.10
PING 10.10.10.10 (10.10.10.10) 56(84) bytes of data.

--- 10.10.10.10 ping statistics ---
3 packets transmitted, 0 received, 100% packet loss, time 1999ms

[root@N1VL-PA-INFL01 ~]#

there is few IPs which are individually getting pinged from source server but in telegraf result_code is showing '2'.
Can you please help to rectify this error?

glinton:
There was a bug when using raw socket mode with `method = "native"` on the ping input plugin where responses would be processed for the wrong host. This has been rectified upstream and the library updated in telegraf.
Better matching in `result_code` has also been implemented in telegraf between `exec` and `native` methods.

Resolves #6521

### DIFF
--- a/plugins/inputs/ping/ping.go
+++ b/plugins/inputs/ping/ping.go
@@ -252,6 +252,25 @@ func (p *Ping) pingToURLNative(destination string, acc telegraf.Accumulator) {
 	var doErr error
 	var packetsSent int
 
+	type sentReq struct {
+		err  error
+		sent bool
+	}
+	sents := make(chan sentReq)
+
+	r.Add(1)
+	go func() {
+		for sent := range sents {
+			if sent.err != nil {
+				doErr = sent.err
+			}
+			if sent.sent {
+				packetsSent++
+			}
+		}
+		r.Done()
+	}()
+
 	for i := 0; i < p.Count; i++ {
 		select {
 		case <-ctx.Done():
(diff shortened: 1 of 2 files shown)


plugins/inputs/ping/ping.go
```diff
@@ -261,11 +269,14 @@ func (p *Ping) pingToURLNative(destination string, acc telegraf.Accumulator) {
 					Seq: seq,
 				})
 				if err != nil {
-					acc.AddFields("ping", map[string]interface{}{"result_code": 2}, map[string]string{"url": destination})
-					acc.AddError(err)
+					if !strings.Contains(err.Error(), "not permitted") {
+						packetsSent++
+					}
+					doErr = err
```
danielnelson: Data race with doErr and packetsSent, likely want to package these up and send them into the channel.

PR #5122, opened 2018-12-07
Adding "javascript" processor plugin

JesusIslam:
### Required for all PRs:

- [x] Signed [CLA](https://influxdata.com/community/cla/).
- [x] Associated README.md updated.
- [x] Has appropriate unit tests.

Adding "javascript" processor plugin, you can now using JavaScript to process your metrics.
The new dependency will be [Otto](https://github.com/robertkrimen/otto).

PR #6579, opened 2019-10-25
Scripting processor, aggregator, output and input

adrianlzt:
These processor/aggregator/output/input allows to define which code
should be run in the config file.
It uses github.com/containous/yaegi interpreter to parse Go code at
runtime.
In the config file an script should be defined with the same methods
(Apply/Add/Push/etc).
To simplify plain methods are used (instead of methods on struct types).
To store state and configs global vars are used.

Some workarounds are needed to overcome yaegi limitations, especially
with "interface{}" types.

Binary size is increased by 13MB.

### Required for all PRs:


adrianlzt:
I don't know why it's failing the ``test-go-1.13-386``. I'm unable to reproduce the error using ``quay.io/influxdb/telegraf-ci:1.13.3`` and ``GOARCH=386 make test``

danielnelson:
Thanks for the very interesting PR, might take a bit before I have time to review as I think this is something I need to look very closing at.  I'm also interested in comparing against https://github.com/google/starlark-go and #5122, I want to include a few options similar to this but for size reasons probably not all of them.

ssoroka:
closing this as there's other options now, with the starlark processor, and the execd input, execd processor, and execd output.

PR #6581, opened 2019-10-25
Fix incorrect results in ping plugin

### LINKED ISSUE #6576: Native Ping not working after 1.12.4
### Relevant telegraf.conf:

```toml
[[inputs.ping]]
  ## Hosts to send ping packets to.
  urls = ["1.1.1.1",
          "1.0.0.1",
          "dash.cloudflare.com",
          "www.google.com",
          "facebook.com",
          "twitter.com",
          "www.youtube.com"
         ]

  ## Method used for sending pings, can be either "exec" or "native".  When set
  ## to "exec" the systems ping command will be executed.  When set to "native"
  ## the plugin will send pings directly.
  ##
  ## While the default is "exec" for backwards compatibility, new deployments
  ## are encouraged to use the "native" method for improved compatibility and
  ## performance.
  method = "native"

  ## Number of ping packets to send per interval.  Corresponds to the "-c"
  ## option of the ping command.
  count = 5

  ## Time to wait between sending ping packets in seconds.  Operates like the
  ## "-i" option of the ping command.
  ping_interval = 0.5

  ## If set, the time to wait for a ping response in seconds.  Operates like
  ## the "-W" option of the ping command.
  timeout = 2.0

  ## If set, the total ping deadline, in seconds.  Operates like the -w option
  ## of the ping command.
  # deadline = 10

  ## Interface or source address to send ping from.  Operates like the -I or -S
  ## option of the ping command.
  interface = "ens3"

  ## Specify the ping executable binary.
  # binary = "ping"

  ## Arguments for ping command. When arguments is not empty, the command from
  ## the binary option will be used and other options (ping_interval, timeout,
  ## etc) will be ignored.
  # arguments = ["-c", "3"]

  ## Use only IPv6 addresses when resolving a hostname.
  ipv6 = false
  interval = "20s"
```

### System info:
Debian 10, Telegraf 1.12.4 (git: HEAD 9a4f08e9)

### Steps to reproduce:

1. Use the above config file in 1.12.3 and observe it working.
2. Update telegraf to 1.12.4 release.

### Expected behavior:

Nine metrics are available:
average_response_ms
maximum_response_ms
minimum_response_ms
packet_received
packet_transmitted
percent_packet_loss
result_code
standard_deviation_ms
ttl

### Actual behavior:

Before updating, there are nine metrics available. But after update, only result_code, packet_received and packet_transmitted are available.

### Additional info:

It doesn’t output any log even in debug log level.

Maybe related to this: https://github.com/influxdata/telegraf/pull/6563

glinton:
In the last change, sending to a channel was overlooked, causing all
ping errors and packet sent count to be ignored.

Resolves #6576

### DIFF
--- a/plugins/inputs/ping/ping.go
+++ b/plugins/inputs/ping/ping.go
@@ -293,10 +293,12 @@ func (p *Ping) pingToURLNative(destination string, acc telegraf.Accumulator) {
 					if strings.Contains(err.Error(), "not permitted") {
 						sent.sent = false
 					}
+					sents <- sent
 					return
 				}
 
 				resps <- resp
+				sents <- sent
 			}(i + 1)
 		}
 	}
--- a/plugins/inputs/ping/ping_test.go
+++ b/plugins/inputs/ping/ping_test.go
@@ -355,4 +355,5 @@ func TestPingGatherNative(t *testing.T) {
 
 	assert.NoError(t, acc.GatherError(p.Gather))
 	assert.True(t, acc.HasPoint("ping", map[string]string{"url": "localhost"}, "packets_transmitted", 5))
+	assert.True(t, acc.HasPoint("ping", map[string]string{"url": "localhost"}, "packets_received", 5))
 }


superjeng1:
I can confirm the debian amd64 version is working again. Thanks a lot!

Issue #6461, opened 2019-09-27
input.smart - error counter log


gregorybrzeski:
## Feature Request

Opening a feature request kicks off a discussion.

### Proposal:

Additional measurement `smart_error_counter_log` which would include data parsed from smartctl "Error Counter Log".

* smart_error_counter_log
    * tags
        * capacity
        * device
        * enabled
        * model
        * serial_no
        * wwn
        * page_name - read, write, verify

    * fields
        * errors_corrected_by_eccfast
        * errors_corrected_by_eccdelayed
        * errors_corrected_by_rereads_rewrites
        * total_errors_corrected
        * correction_algorithm_invocations
        * gigabytes_processed
        * total_uncorrected_errors

For devices attached using scsi protocol (including SAS drives) smartctl prints error counter log if it is supported.

HGST ssd sas drives do not return much information in attributes:

```
$ smartctl --attributes --format=brief /dev/sdi 
smartctl 6.5 2016-05-07 r4318 [x86_64-w64-mingw32-2016] (sf-6.5-1)
Copyright (C) 2002-16, Bruce Allen, Christian Franke, www.smartmontools.org

=== START OF READ SMART DATA SECTION ===
Current Drive Temperature:     26 C
Drive Trip Temperature:        70 C

Manufactured in week 32 of year 2016
Specified cycle count over device lifetime:  0
Accumulated start-stop cycles:  0
Specified load-unload count over device lifetime:  0
Accumulated load-unload cycles:  0
defect list format 6 unknown
Elements in grown defect list: 0

Vendor (Seagate) cache information
  Blocks sent to initiator = 4200776633156356
``` 

Most of the important information is included in the Error Counter Log:

```
$ smartctl --log error --format=brief /dev/sdi
smartctl 6.5 2016-05-07 r4318 [x86_64-w64-mingw32-2016] (sf-6.5-1)
Copyright (C) 2002-16, Bruce Allen, Christian Franke, www.smartmontools.org

=== START OF READ SMART DATA SECTION ===
Error counter log:
           Errors Corrected by           Total   Correction     Gigabytes    Total
               ECC          rereads/    errors   algorithm      processed    uncorrected
           fast | delayed   rewrites  corrected  invocations   [10^9 bytes]  errors
read:          0        0         0         0          0       4797.541           0
write:         0        0         0         0          0       8932.303           0
verify:        0        0         0         0          0          5.713           0

Non-medium error count:        0
```

Smartctl print code for the Error Counter Log:
[...]

PR #6585, opened 2019-10-27
Smart Input Plugin: support for metrics from Error Counter Log

gregorybrzeski:
Additional measurement `smart_error_counter_log` which includes data parsed from smartctl "Error Counter Log".

* smart_error_counter_log
    * tags
        * capacity
        * device
        * enabled
        * model
        * serial_no
        * wwn
        * page_name - read, write, verify

    * fields
        * errors_corrected_by_eccfast
        * errors_corrected_by_eccdelayed
        * errors_corrected_by_rereads_rewrites
        * total_errors_corrected
        * correction_algorithm_invocations
        * gigabytes_processed
        * total_uncorrected_errors

For devices attached using scsi protocol (including SAS drives) smartctl prints error counter log if it is supported.

Discussed in #6461
 
### Required for all PRs:

### DIFF
--- a/plugins/inputs/smart/smart.go
+++ b/plugins/inputs/smart/smart.go
@@ -391,6 +391,37 @@ func (m *Smart) gatherDisk(acc telegraf.Accumulator, device string, wg *sync.Wai
 				}
 			}
 		}
+
+		if m.ErrorCounterLog {
+			ecl := errorCounterLog.FindStringSubmatch(line)
+			if len(ecl) > 1 {
+				eclTags := map[string]string{}
+				eclFields := make(map[string]interface{})
+
+				eclTags["page_name"] = ecl[1]
+
+				keys := [...]string{"device", "model", "serial_no", "wwn", "capacity", "enabled"}
+				for _, key := range keys {
+					if value, ok := deviceTags[key]; ok {
+						eclTags[key] = value
+					}
+				}
+
+				eclFields["errors_corrected_by_eccfast"] = ecl[2]
+				eclFields["errors_corrected_by_eccdelayed"] = ecl[3]
+				eclFields["errors_corrected_by_rereads_rewrites"] = ecl[4]
+				eclFields["total_errors_corrected"] = ecl[5]
+				eclFields["correction_algorithm_invocations"] = ecl[6]
+				gigabytes_processed, err := strconv.ParseFloat(ecl[7], 64)
+				if err != nil {
+					gigabytes_processed = 0
+				}
+				eclFields["gigabytes_processed"] = gigabytes_processed * 1000 * 1000 * 1000
+				eclFields["total_uncorrected_errors"] = ecl[8]
+
+				acc.AddFields("smart_error_counter_log", eclFields, eclTags)
+			}
+		}
 	}
 	acc.AddFields("smart_device", deviceFields, deviceTags)
 }
--- a/plugins/inputs/smart/smart.go
+++ b/plugins/inputs/smart/smart.go
@@ -394,7 +394,7 @@ func (m *Smart) gatherDisk(acc telegraf.Accumulator, device string, wg *sync.Wai
 
 		if m.ErrorCounterLog {
 			ecl := errorCounterLog.FindStringSubmatch(line)
-			if len(ecl) > 1 {
+			if len(ecl) == 9 {
 				eclTags := map[string]string{}
 				eclFields := make(map[string]interface{})
 
--- a/plugins/inputs/smart/smart.go
+++ b/plugins/inputs/smart/smart.go
@@ -112,13 +112,15 @@ var (
 )
 
 type Smart struct {
-	Path       string
-	Nocheck    string
-	Attributes bool
-	Excludes   []string
-	Devices    []string
-	UseSudo    bool
-	Timeout    internal.Duration
+	Path            string            `toml:"path"`
+	Nocheck         string            `toml:"nocheck"`
+	Attributes      bool              `toml:"attributes"`
+	ErrorCounterLog bool              `toml:"error_counter_log"`
+	Excludes        []string          `toml:"excludes"`
+	Devices         []string          `toml:"devices"`
+	UseSudo         bool              `toml:"use_sudo"`
+	Timeout         internal.Duration `toml:"timeout"`
+	Log             telegraf.Logger
 }
 
 var sampleConfig = `
@@ -142,6 +144,10 @@ var sampleConfig = `
   ## Gather all returned S.M.A.R.T. attribute metrics and the detailed
   ## information from each drive into the 'smart_attribute' measurement.
   # attributes = false
+
+  ## Gather all returned S.M.A.R.T. error counter log metrics
+  ## for each drive into the 'smart_error_counter_log' measurement.
+  # error_counter_log = false
 
   ## Optionally specify devices to exclude from reporting.
   # excludes = [ "/dev/pass6" ]
--- a/plugins/inputs/smart/smart.go
+++ b/plugins/inputs/smart/smart.go
@@ -268,6 +268,9 @@ func (m *Smart) gatherDisk(acc telegraf.Accumulator, device string, wg *sync.Wai
 	defer wg.Done()
 	// smartctl 5.41 & 5.42 have are broken regarding handling of --nocheck/-n
 	args := []string{"--info", "--health", "--attributes", "--tolerance=verypermissive", "-n", m.Nocheck, "--format=brief"}
+	if m.ErrorCounterLog {
+		args = append(args, "--log", "error")
+	}
 	args = append(args, strings.Split(device, " ")...)
 	out, e := runCmd(m.Timeout, m.UseSudo, m.Path, args...)
 	outStr := string(out)
--- a/plugins/inputs/smart/smart.go
+++ b/plugins/inputs/smart/smart.go
@@ -232,7 +232,7 @@ func (m *Smart) getAttributes(acc telegraf.Accumulator, devices []string) {
 	wg.Add(len(devices))
 
 	for _, device := range devices {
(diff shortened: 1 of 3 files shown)


plugins/inputs/smart/smart_test.go
```diff
@@ -379,6 +379,13 @@ func TestExcludedDev(t *testing.T) {
 }
 
 func TestGatherSATAInfo(t *testing.T) {
+	s := NewSmart()
```
glinton: Nitpick: on these, just define inline:
```go
s := &Smart{
  Timeout: internal.Duration{Duration: time.Second * 30},
  UseSudo: true,
  etc...
}
```

plugins/inputs/smart/smart.go
```diff
@@ -373,6 +391,37 @@ func gatherDisk(acc telegraf.Accumulator, timeout internal.Duration, usesudo, co
 				}
 			}
 		}
+
+		if m.ErrorCounterLog {
+			ecl := errorCounterLog.FindStringSubmatch(line)
+			if len(ecl) > 1 {
```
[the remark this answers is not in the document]
gregorybrzeski: I am not sure if I follow, do you mean to check for number of match groups like code below ?

```
if len(ecl) == 8 {
```

plugins/inputs/smart/smart_test.go
gregorybrzeski: NewSmart() is setting default Timeout, maybe just remove setting Timeout in these tests and allow NewSmart() to do it ? It's going to be in line with other tests like: TestGatherAttributes, TestGatherNoAttributes.

plugins/inputs/smart/smart.go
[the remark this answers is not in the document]
glinton: Correct

plugins/inputs/smart/smart_test.go
glinton: Makes sense, don't set Timeout so match those other tests, or update the other tests to define inline. I'm fine with either way.

gregorybrzeski: removed Timeout

plugins/inputs/smart/smart.go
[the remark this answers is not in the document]
gregorybrzeski: I agree that it's more explicit and readable, changed. However it's just one simple regex and it would match only as a whole with 8 match groups or not match at all.

gregorybrzeski:
@danielnelson @glinton - is there anything else what I can do to get it merged ?

gregorybrzeski:
@danielnelson @glinton any news on merging this ? Can I help in any way ?

ad-m (approved):

plugins/inputs/smart/smart.go
```diff
@@ -112,13 +121,15 @@ var (
...
-	Timeout    internal.Duration
+	Path            string            `toml:"path"`
+	Nocheck         string            `toml:"nocheck"`
+	Attributes      bool              `toml:"attributes"`
+	ErrorCounterLog bool              `toml:"error_counter_log"`
+	Excludes        []string          `toml:"excludes"`
+	Devices         []string          `toml:"devices"`
+	UseSudo         bool              `toml:"use_sudo"`
+	Timeout         internal.Duration `toml:"timeout"`
+	Log             telegraf.Logger
```
srebhan, line 132: Could you please change to 
```suggestion
	Log             telegraf.Logger `toml:"-"`
```
so that the toml parser won't try to fill this field.

plugins/inputs/smart/smart.go
```diff
@@ -373,6 +391,37 @@ func gatherDisk(acc telegraf.Accumulator, timeout internal.Duration, usesudo, co
...
+					if value, ok := deviceTags[key]; ok {
+						eclTags[key] = value
+					}
+				}
+
+				eclFields["errors_corrected_by_eccfast"] = ecl[2]
+				eclFields["errors_corrected_by_eccdelayed"] = ecl[3]
+				eclFields["errors_corrected_by_rereads_rewrites"] = ecl[4]
+				eclFields["total_errors_corrected"] = ecl[5]
+				eclFields["correction_algorithm_invocations"] = ecl[6]
```
srebhan, line 414: As the metric already contains an `error` label, should we maybe remove the `errors` part from the field name?

plugins/inputs/smart/smart.go
```diff
@@ -373,6 +391,37 @@ func gatherDisk(acc telegraf.Accumulator, timeout internal.Duration, usesudo, co
...
+				gigabytes_processed, err := strconv.ParseFloat(ecl[7], 64)
+				if err != nil {
+					gigabytes_processed = 0
+				}
+				eclFields["gigabytes_processed"] = gigabytes_processed * 1000 * 1000 * 1000
```
srebhan, line 419: Isn't this now the `bytes_processed`? And are we talking _GiB_ (base 2) or _GB_ (base 10)?

plugins/inputs/smart/smart_test.go
```diff
@@ -389,12 +395,18 @@ func TestGatherSATAInfo(t *testing.T) {
 	)
 
 	wg.Add(1)
-	gatherDisk(acc, internal.Duration{Duration: time.Second * 30}, true, true, "", "", "", wg)
+	s.gatherDisk(acc, "", wg)
```
srebhan, line 398: You probably should wait for the accumulator to collect the metrics using `acc.Wait()` otherwise this test might be racy.

srebhan (commented):
This looks nice already. I have some minor comments in the code...

srebhan:
@gregorybrzeski any update?

ad-m:
@gregorybrzeski, could you help anyway?

plugins/inputs/smart/smart.go
gregorybrzeski: @srebhan idea was to keep the names same as returned by smartctl so it's easier to correlate to the output of the tool. However I can see the issues raised by you in the other comment next comment regarding "gigabytes_processed".

plugins/inputs/smart/smart.go
gregorybrzeski: You are right this values is no longer expressed as gigabytes therefore maybe it's better to keep it as gigabytes. However it's provided by smartctl output with precision down to three decimal places. Need to think how to handle this:

* send it as float
* turn it into bytes or megabytes
* round it to neareset gigabyte (loosing precision)

Looking at the output of smartctl command it states that this field is `10^9 bytes` therefore I assume it's GB (base 10).

plugins/inputs/smart/smart_test.go
gregorybrzeski: You are right, it's not my test I was just changing where the gatherDisk() comes from.

gregorybrzeski:
@srebhan most probably I will look at it second weekend of May. Few months ago I had a look how complex the rebase is going to be after new code was added to that plugin and AFAIR there wasn't anything complex to be expected. 

I'm glad that you have looked at it and would like to merge it, as I have few other enhancements for this plugin which require PR however I was waiting for this PR to be merged first. Thank you!

gregorybrzeski:
@ad-m thanks for offer, I will manage to handle it.

plugins/inputs/smart/smart.go
srebhan: I would either go for the bytes (option 2) or the float option. Loosing precision is kind of the last resort...

plugins/inputs/smart/smart.go
srebhan: Understood. Not sure though if it is really a good idea to track the smartctl names. What if they change the output name? Will you change the field? ;-)

srebhan:
@gregorybrzeski just drop me a note whenever you are ready to go. :-)

sspaink:
@gregorybrzeski thank you for your work here! When you've addressed @srebhan comments, the @influxdata/telegraf-team will loop back to it and help get it merged.

srebhan:
Any volunteers for taking over this PR?

gregorybrzeski:
@srebhan if someone is willing I'm happy for them to take over, however I have nearly everything in place after rebase. Just need to clean things up as my development of this module was going way beyond the PR as for well over a year after positive review I wasn't able to get anyone interested to get it merged.

srebhan:
I'm happy to give it another review round. Please just let me know and I'll remove the help-wanted tag immediately.

gregorybrzeski:
@srebhan this or next weekend I should take care of it unless someone will take over before.

As I have mentioned earlier I have done considerable changes to the module for our use. Usually I would have done small PRs with each iteration, but it didn't happened while I was waiting for merge of this one. However in present state can you advise what would be the best approach to do a major overhaul of this plugin? Do you know of another plugin which could be a good example of smoothly transitioning from version 1 into version 2?

srebhan:
> However in present state can you advise what would be the best approach to do a major overhaul of this plugin? Do you know of another plugin which could be a good example of smoothly transitioning from version 1 into version 2?

Well there are basically two options

1. If your changes are (or can be made) backward compatible, do the overhaul in place. If very large, try to cut down into digestible PRs.
2. If changes are not or cannot be made backward compatible, add a new plugin (e.g. `smart_v2`).

As we are already carrying tons of plugins, option 1 would be preferred. Backward compatibility means, old configs still work (i.e. no removal of config options etc) and produce the same metric outputs. The later one means that adding new fields/tags is guarded by an `include_...`/`enable_...` and/or `exclude_...`/`disable_...` list for enabling/disabling features. These 'new' lists should be set such that the old set of metrics/tags/fields is produced by default. The `github` input plugin for example has an `additional_fields` setting which explicitly enables additional data to be produced... 

Does that help?

srebhan:
@gregorybrzeski any time to work on this?

PR #6587, opened 2019-10-28
WIP: added wireguard input plugin

gitirabassi:
### Required for all PRs:

### DIFF
--- a/plugins/inputs/wireguard/README.md
+++ b/plugins/inputs/wireguard/README.md
@@ -1,0 +1,42 @@
+# Wireguard Input Plugin
+
+This plugin provides information about Wireguard interfaces
+
+### Configuration
+
+Make sure that running `wg show` on the host works correctly before enabling this plugin
+
+```toml
+
+[[inputs.wireguard]]
+  # if none are provided, all of the available ones will be scraped
+  interfaces = ["wg0"]
+
+```
+
+### Metrics
+
+- wireguard
+  - tags:
+    - name
+    - type
+    - serverpublikey
+    - listenport
+    - firewallmark
+    - peerpublickey
+    - endpoint
+  - fiels:
+    - received_bytes
+    - transmit_bytes
+    - protocol
+    - last_hanshake_time
+    - persisten_keepalive_interval
+
+### Example Output:
+
+```
+% telegraf --config=plugins/inputs/wireguard/dev/telegraf.conf --test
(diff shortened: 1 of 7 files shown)


plugins/inputs/wireguard/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+```
+
+### Metrics
+
+- wireguard
+  - tags:
+    - name
+    - type
+    - serverpublikey
```
BondAnthony: Spelling, `serverpublickey`

plugins/inputs/wireguard/README.md
```diff
@@ -0,0 +1,42 @@
...
+    - listenport
+    - firewallmark
+    - peerpublickey
+    - endpoint
+  - fiels:
```
BondAnthony: `fields`

plugins/inputs/wireguard/README.md
```diff
@@ -0,0 +1,42 @@
...
+    - received_bytes
+    - transmit_bytes
+    - protocol
+    - last_hanshake_time
```
BondAnthony: `last_handshake_time`

plugins/inputs/wireguard/README.md
```diff
@@ -0,0 +1,42 @@
...
+    - persisten_keepalive_interval
```
BondAnthony: `last_handshake_time` and `persistent_keepalive_interval`

danielnelson:
Related #6367

plugins/inputs/wireguard/README.md
glinton: We prefer snake cased tag/field names as well

gitirabassi:
@danielnelson Before attempting any more work on this I want to make sure, since `golang.zx2c4.com/wireguard/wgctrl ` is the official wireguard client in Go now, that I will be ok and acceptable to import `github.com/mdlayher/genetlink` in the telegraf repository. This was the issue blocking #6367 . Personally I don't think neither wgctrl nor telegraf will be migrating netlink library. I believe it comes down to how bad to we want wireguard input plugin and I think it's quite important since it will be the next big vpn and we might just be a little bit behind given https://github.com/MindFlavor/prometheus_wireguard_exporter

danielnelson:
Overall I agree with your assessment on the netlink library, though I'm perhaps less concerned about using the official client library and more optimistic that docker will do the switch for us someday. 

I'll look over these libraries soon, just my normal pass for really bad things, and get back with you.  Usually I don't find anything thats a problem.  Then let's work to update #6367 with any features you think need to be changed or added instead of continuing with this PR.

gitirabassi:
I agree #6367 is better. Thanks and i’ll wait for your feedback

PR #6590, opened 2019-10-29
feat(inputs.win_perf_counters): Add remote computers support

### LINKED ISSUE #1899: win_perf_counters and remote Windows host

vlastahajek:
Resolves #1899 

Added support for reading performance counters from remote computers. User, under who is Telegraf running, must be already authenticated to the remote computer(s).

The support is added via additional optional global parameter `Sources`. This parameter can be also set for performance object to configure per host-specific performance counter:
``` toml
[[inputs.win_perf_counters]]
  PreVistaSupport=false
  UseWildcardsExpansion=false
  UsePerfCounterTime=true
  CountersRefreshInterval="0"
  PrintValid=true
  # Names or ip addresses of remote computers to gather counters from, including local computer.
  # Telegraf's user must be already authenticated to the remote computers.
  # It can be overridden at the object level
  Sources=["localhost", "192.168.10.12"]

   # Overrides the Sources global parameter for current performance object.
    Sources = ["192.168.1.68"]
    # Processor usage, alternative to native, reports on a per core.
    ObjectName = "Processor"
    Instances = ["*"]
    Counters = [
      "% Processor Time",
    ]
    Measurement = "win_cpu"
```
It allows to define the same configuration for multiple computers and also different configurations per computer.

Computer name is then added as the `source` tag. Telegraf host compuper has the `source` set to the computer name.

Readme has been also enhanced with plugins's schema description.

### Required for all PRs:

### DIFF
--- a/plugins/inputs/win_perf_counters/README.md
+++ b/plugins/inputs/win_perf_counters/README.md
@@ -24,6 +24,18 @@ For more information on concepts and terminology including object,
 counter, and instance names, see the help in the Windows Performance
 Monitor app.
 
+### Schema
+
+*Measurement name* is specified per performance object or `win_perf_counters` by default.
+
+*Tags:*
+
+- source - computer name, as specified in the `Sources` parameter. Name `localhost` is translated into the host name
+- objectname - normalized name of the performance object
+- instance - instance name, if performance object supports multiple instances, otherwise omitted
+
+*Fields* are counters of the performance object. The field name is normalized counter name.
+
 ### Plugin wide
 
 Plugin wide entries are underneath `[[inputs.win_perf_counters]]`.
@@ -114,6 +126,21 @@ You can find the list of possible errors here: [PDH errors](https://github.com/i
 Example:
 `IgnoredErrors=["PDH_NO_DATA"]`
 
+#### Sources
+
+(Optional)
+
+Host names or ip addresses of computers to gather all performance counters from.  User, under which Telegraf runs, must be already authenticated to the remote computer(s).
+E.g. via Windows sharing `net use \\SQL-SERVER-01`.
+Use either localhost (`"localhost"`) or real local computer name to gather counters also from localhost among other computers.
+Skip, if gather only from localhost.
+
+If a performance object is present only on specific hosts set `Source` param on the object level configuration to override global Sources.
+
+Example:  `Sources = ["localhost", "SQL-SERVER-01", "SQL-SERVER-02", "SQL-SERVER-03"]`
+
+Default: `Sources = ["localhost"]`
+
 ### Object
 
 See Entry below.
@@ -169,6 +196,12 @@ This must be specified for every counter you want the results of, or use
 `["*"]` for all the counters of the object, if the `UseWildcardsExpansion` param
 is set to `true`.
 
+#### Sources (Object)
+
+(Optional)
+
+Overrides the [Sources](#sources) global parameter for current performance object. See [Sources](#sources) description for more details.
+
 #### Measurement
 
 (Optional)
@@ -189,7 +222,7 @@ This key is optional. It is a simple bool.
 If set to `true`, counter values will be provided in the raw, integer, form. This is in contrast with the default behavior, where values are returned in a formatted, displayable, form
 as seen in the Windows Performance Monitor.  
 A field representing raw counter value has the `_Raw` suffix. Raw values should be further used in a calculation, e.g. `100-(non_negative_derivative("Percent_Processor_Time_Raw",1s)/100000`
-Note: Time based counters (i.e. _% Processor Time_) are reported in hundredths of nanoseconds.
+Note: Time based counters (i.e. `% Processor Time`) are reported in hundredths of nanoseconds.
 
 Example: `UseRawValues = true`
 
--- a/plugins/inputs/win_perf_counters/win_perf_counters.go
+++ b/plugins/inputs/win_perf_counters/win_perf_counters.go
@@ -6,7 +6,9 @@ package win_perf_counters
 import (
 	"errors"
 	"fmt"
+	"os"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/influxdata/telegraf"
@@ -41,6 +43,11 @@ var sampleConfig = `
   ## You can find the list here: https://github.com/influxdata/telegraf/blob/master/plugins/inputs/win_perf_counters/pdh.go
   ## e.g.: IgnoredErrors = ["PDH_NO_DATA"]
   # IgnoredErrors = []
+  # Names or ip addresses of remote computers to gather counters from, including local computer.
+  # Telegraf's user must be already authenticated to the remote computers.
+  # It can be overridden at the object level
+  # Sources = ["localhost"]
+
 
   [[inputs.win_perf_counters.object]]
     # Processor usage, alternative to native, reports on a per core.
@@ -61,6 +68,8 @@ var sampleConfig = `
     # WarnOnMissing = false
     # Gather raw values instead of formatted. Raw value is stored in the field name with the "_Raw" suffix, e.g. "Disk_Read_Bytes_sec_Raw".
     # UseRawValues = true
+    # Overrides the Sources global parameter for current performance object.
+    # Sources = ["localhost", "SQL-server1"]
 
   [[inputs.win_perf_counters.object]]
     # Disk times and queues
@@ -151,7 +160,7 @@ var sampleConfig = `
     Measurement = "win_swap"
 `
 
-type Win_PerfCounters struct {
+type WinPerfCounters struct {
 	PrintValid                 bool `toml:"PrintValid"`
 	PreVistaSupport            bool `toml:"PreVistaSupport" deprecated:"1.7.0;determined dynamically"`
 	UsePerfCounterTime         bool
@@ -160,15 +169,29 @@ type Win_PerfCounters struct {
 	UseWildcardsExpansion      bool
 	LocalizeWildcardsExpansion bool
 	IgnoredErrors              []string `toml:"IgnoredErrors"`
+	Sources                    []string
 
 	Log telegraf.Logger
 
 	lastRefreshed time.Time
-	counters      []*counter
-	query         PerformanceQuery
+	queryCreator  PerformanceQueryCreator
+	hostCounters  map[string]*hostCountersInfo
+	// cached os.Hostname()
+	cachedHostname string
+}
+
+type hostCountersInfo struct {
+	// computer name used as key and for printing
+	computer string
+	// computer name used in tag
+	tag       string
+	counters  []*counter
+	query     PerformanceQuery
+	timestamp time.Time
 }
 
 type perfobject struct {
+	Sources       []string
 	ObjectName    string
 	Counters      []string
 	Instances     []string
@@ -181,6 +204,7 @@ type perfobject struct {
 
 type counter struct {
 	counterPath   string
+	computer      string
 	objectName    string
 	counter       string
 	instance      string
@@ -196,14 +220,17 @@ type instanceGrouping struct {
 	objectname string
 }
 
+type fieldGrouping map[instanceGrouping]map[string]interface{}
+
 var sanitizedChars = strings.NewReplacer("/sec", "_persec", "/Sec", "_persec",
 	" ", "_", "%", "Percent", `\`, "")
 
 // extractCounterInfoFromCounterPath gets object name, instance name (if available) and counter name from counter path
 // General Counter path pattern is: \\computer\object(parent/instance#index)\counter
 // parent/instance#index part is skipped in single instance objects (e.g. Memory): \\computer\object\counter
-func extractCounterInfoFromCounterPath(counterPath string) (object string, instance string, counter string, err error) {
-
+func extractCounterInfoFromCounterPath(counterPath string) (computer, object, instance, counter string, err error) {
+	computer = ""
+	leftComputerBorderIndex := -1
 	rightObjectBorderIndex := -1
 	leftObjectBorderIndex := -1
 	leftCounterBorderIndex := -1
@@ -219,6 +246,8 @@ func extractCounterInfoFromCounterPath(counterPath string) (object string, insta
 					leftCounterBorderIndex = i
 				} else if leftObjectBorderIndex == -1 {
 					leftObjectBorderIndex = i
+				} else if leftComputerBorderIndex == -1 {
+					leftComputerBorderIndex = i
 				}
 			}
 		case '(':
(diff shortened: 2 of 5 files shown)


plugins/inputs/win_perf_counters/README.md
```diff
@@ -19,7 +19,19 @@ as counters used when performance monitoring
...
  This file is likely to be updated in the future with more examples for
  useful configurations for separate scenarios.
-
+ 
+ ### Schema
+ 
+ *Measurement name* is specified per performance object or `win_perf_counters` by default.
+ 
+ *Tags:*
+  - computer - computer name, if specified in the `Computers` parameter, otherwise omitted
```
glinton: I wonder if this should alwasys be set and be renamed to `source` as discussed in #4413

plugins/inputs/win_perf_counters/README.md
```diff
@@ -19,7 +19,19 @@ as counters used when performance monitoring
...
+  - objectname - normalized name of the performance object
```
glinton: Snake case this - `object_name`

plugins/inputs/win_perf_counters/README.md
```diff
@@ -92,6 +104,14 @@ This must follow before other plugin configurations,
 beneath the main win_perf_counters entry, `[[inputs.win_perf_counters]]`.
 
 Following this are 3 required key/value pairs and three optional parameters and their usage.
+#### Computers
+*Optional*
+
+Host names or IPs of computers to gather counters from. User, under which Telegraf runs, must be already authenticated to the remote computer(s). E.g. via Windows sharing `net use \\SQL-SERVER-01`.
+Skip, if gather only from localhost.
```
glinton: For config consistency, is it valid to show that `computers = ["localhost"]` is the default?
Also, are IPs valid?

plugins/inputs/win_perf_counters/README.md
```diff
@@ -92,6 +104,14 @@ This must follow before other plugin configurations,
...
+Use either empty string `""`, localhost `"localhost"` or real computer name to gather counters also from localhost among other computers.
+
+Example:  `Computers = ["localhost", "SQL-SERVER-01", "SQL-SERVER-02", "SQL-SERVER-03"]`
```
glinton: Change case of `Computers` to lower

plugins/inputs/win_perf_counters/win_perf_counters.go
```diff
@@ -43,6 +44,9 @@ var sampleConfig = `
       "% DPC Time",
     ]
     Measurement = "win_cpu"
+    # Names or IPs of remote computers to gather counters from, including local computer.
+    # Telegraf's user must be already authenticated to the remote computers.
+    # Computers = ["localhost","SQL-SERVER-01","SQL-SERVER-02"]
```
glinton: case - `computers`

plugins/inputs/win_perf_counters/win_perf_counters.go
```diff
@@ -149,11 +153,22 @@ type Win_PerfCounters struct {
...
-	query         PerformanceQuery
+	queryCreator  PerformanceQueryCreator
+	hostCounters  map[string]*hostCountersInfo
+}
+
+type hostCountersInfo struct {
+	// computer name used as key and for printing
+	computer string
+	// computer name used in tag
+	computerTag string
```
glinton: Can't `computer` be re-used for this? It seems always setting the tag (even if it is "localhost") is better than selectively doing so.

plugins/inputs/win_perf_counters/win_perf_counters.go
```diff
@@ -305,29 +355,44 @@ func (m *Win_PerfCounters) AddItem(counterPath string, objectName string, instan
...
-					objectname := PerfObject.ObjectName
-
-					if instance == "------" {
-						counterPath = "\\" + objectname + "\\" + counter
-					} else {
-						counterPath = "\\" + objectname + "(" + instance + ")\\" + counter
-					}
+			computers := PerfObject.Computers
+			if len(computers) == 0 {
+				computers = append(computers, "")
```
glinton: Using `append` seems odd here, just do `computers = []string{""}` instead

plugins/inputs/win_perf_counters/win_perf_counters.go
```diff
@@ -341,61 +406,79 @@ func (m *Win_PerfCounters) ParseConfig() error {
...
+			m.Log.Debugf("gathering from %s", hostInfo.computer)
+			start := time.Now()
+			err = m.gatherComputerCounters(hostInfo, acc)
+			m.Log.Debugf("gathering from %s finished in %.3fs", hostInfo.computer, time.Now().Sub(start).Seconds())
+			if err != nil {
+				acc.AddError(fmt.Errorf("error during collecting data on host '%s': %s", hostInfo.computer, err.Error()))
+			}
+			wg.Done()
+		}(hostCounterInfo)
+		wg.Add(1)
```
glinton: `Add` before calling the goroutine.

plugins/inputs/win_perf_counters/win_perf_counters.go
```diff
@@ -429,21 +512,33 @@ func (m *Win_PerfCounters) Gather(acc telegraf.Accumulator) error {
...
-	for instance, fields := range collectFields {
+	for instance, fields := range collectedFields {
 		var tags = map[string]string{
 			"objectname": instance.objectname,
 		}
 		if len(instance.instance) > 0 {
 			tags["instance"] = instance.instance
 		}
-		acc.AddFields(instance.name, fields, tags, timestamp)
+		if len(hostCounterInfo.computerTag) > 0 {
```
glinton: Instead, check if `.computer` is empty and set to localhost, otherwise set to the `source` (computer->source) tag to the value of `.computer`

plugins/inputs/win_perf_counters/win_perf_counters.go
```diff
@@ -341,61 +406,79 @@ func (m *Win_PerfCounters) ParseConfig() error {
...
 	// Parse the config once
 	var err error
 
 	if m.lastRefreshed.IsZero() || (m.CountersRefreshInterval.Duration.Nanoseconds() > 0 && m.lastRefreshed.Add(m.CountersRefreshInterval.Duration).Before(time.Now())) {
-		if m.counters != nil {
-			m.counters = m.counters[:0]
-		}
-
-		if err = m.query.Open(); err != nil {
+		if err = m.cleanQueries(); err != nil {
```
glinton: This seems to call a `.Close()` on the query, is that intended?

plugins/inputs/win_perf_counters/README.md
```diff
@@ -19,7 +19,19 @@ as counters used when performance monitoring
...
-
+ 
+ ### Schema
+ 
+ *Measurement name* is specified per performance object or `win_perf_counters` by default.
+ 
+ *Tags:*
+  - computer - computer name, if specified in the `Computers` parameter, otherwise omitted
+  - objectname - normalized name of the performance object
+  - instance - instance name, if performance object supports multiple instances, otherwise omitted
```
glinton: What is the drawback/limitation to always setting `instance`?

plugins/inputs/win_perf_counters/README.md
vlastahajek: `source` is very overloaded name. In this case, `computer` is more suitable as it is officially part of complete counter path, see [https://docs.microsoft.com/en-us/windows/win32/perfctrs/specifying-a-counter-path](https://docs.microsoft.com/en-us/windows/win32/perfctrs/specifying-a-counter-path)

plugins/inputs/win_perf_counters/README.md
vlastahajek: In this part of readme, I just added schema description, which was missing. 
`objectname` is legacy tag name. Even it is not according to best practises, I  think we cannot change it now.

plugins/inputs/win_perf_counters/README.md
vlastahajek: It is for consistency with other params, which are also capitalized.

plugins/inputs/win_perf_counters/win_perf_counters.go
vlastahajek: Same as above

plugins/inputs/win_perf_counters/win_perf_counters.go
vlastahajek: It depends if plugin should add tag computer=localhost, when it gathers from localhost.
Until now it wasn't there, and many users will, probably, not gather from multiple hosts.

Maybe, it should add this tag only if it gathers from multiple hosts?

plugins/inputs/win_perf_counters/win_perf_counters.go
vlastahajek: ok

plugins/inputs/win_perf_counters/win_perf_counters.go
vlastahajek: ok

plugins/inputs/win_perf_counters/win_perf_counters.go
vlastahajek: In this branch, all previous counter infos must be cleaned and fresh infos are added.
There were only one reused query object previously, where `open` function called internally `close` if necessary.

plugins/inputs/win_perf_counters/README.md
vlastahajek: Instance part of performance object that has only single instance is empty.

plugins/inputs/win_perf_counters/win_perf_counters.go
glinton: I prefer always setting the tag, but it wouldn't kill me if it wasn't

vlastahajek: I'm really not sure what is the best way to go with. I decided to go with method with no change for users with normal usage. However, when I imagine seeing a bunch of data points gathered from multiple hosts and couple of lines show `localhost` among the real names, it doesn't look good neither. 
Maybe using real computer name, as reported in the host tag, would be the good solution, but it doesn't clearly indicate it is gathered on localhost.

Maybe we could ask Daniel for an advice, as he has a great sense for regression and good  overview of all plugins, so he may recall plugin with similar problem.

plugins/inputs/win_perf_counters/README.md
danielnelson: I don't disagree with your point but all the same this should be changed to `source` for forward consistency and special handling with some outputs.

plugins/inputs/win_perf_counters/README.md
```diff
@@ -92,6 +104,14 @@ This must follow before other plugin configurations,
 beneath the main win_perf_counters entry, `[[inputs.win_perf_counters]]`.
 
 Following this are 3 required key/value pairs and three optional parameters and their usage.
+#### Computers
```
danielnelson: Set this option only at the top level of the plugin:
```
[[inputs.win_perf_counters]]
  computers = ["127.0.0.1", "SQL-SERVER-01"]
  [[inputs.win_perf_counters.objects]]
    object_name = "LogicalDisk"
```

This will be a bit less flexible but this is a reasonable level of control and more in line with expectations from other plugins.

plugins/inputs/win_perf_counters/win_perf_counters.go
danielnelson: I think we should always add the tag, for localhost though we should use `os.Hostname()` though.  Adding new tags won't break any queries so long as there isn't an existing field with the same name.

plugins/inputs/win_perf_counters/README.md
danielnelson: I think the default should be shown with `localhost`, this gives a nice progression when removing the item.  Internally it should be treated the same.  This will work better and be more important when we update the plugin to match the current Telegraf styling.

This looks like we are only adding a server:
```diff
- computer = []
+ computer = ["SQL-SERVER-01"]
```

This style appears more accurate, we are removing localhost and adding another computer:
```diff
- computer = ["localhost"]
+ computer = ["SQL-SERVER-01"]
```

DmitryMaletin:
Is there any ETA that PR will be merged?

plugins/inputs/win_perf_counters/README.md
vlastahajek: Ok

plugins/inputs/win_perf_counters/README.md
vlastahajek: Make sense to have it global for all objects
But as some instances of the same performance object can have different name on different computers (e.g. network interfaces) as would allow to override the source at object level.

plugins/inputs/win_perf_counters/README.md
vlastahajek: Ok.

plugins/inputs/win_perf_counters/win_perf_counters.go
vlastahajek: Ok

danielnelson: Yes, I think this will be okay.

plugins/inputs/win_perf_counters/README.md
danielnelson: This will be okay as the user can have multiple `[[inputs.win_perf_counters]]`.

vlastahajek: And what about possibility to override this at performance object level  `[[inputs.win_perf_counters.object]]` (currently left in the implementation)? Does it seems confusing?
It allows a bit more compact configuration.
When using  multiple `[[inputs.win_perf_counters]]`, user will have to copy all other global parameters and sync them every time a change is needed.

powersj:
@vlastahajek are you interested in continuing this? Looks like a rebase on master to resolve some conflicts, and then a few outstanding points are what is left.

vlastahajek:
@powersj, OK. I will look at this later this week or at the beginning of the next week.

vlastahajek:
It's ready.

sjwang90:
@vlastahajek Do you know where this original request came from? Determining the priority of getting this reviewed and included into Telegraf

sspaink:
I moved these changes to a local branch and created a new pull request. At the moment we are waiting for customer/community participation to help ensure this change meets their needs (configuration makes sense, etc.). Therefore we will need to be able to re-base it and get new build artifacts on request, moving this pull request to be under a local branch seems like the best approach.

Thank you @vlastahajek for all your work on this and quick responses, I will be closing this pull request so you aren't responsible anymore for helping us re-base it while we wait.

PR #6598, opened 2019-10-30
Ensure user is stripped out of logs in influx input

### LINKED ISSUE #6588: InfluxDB 1.7.9 authentication setting for /debug/vars access
## Feature Request

Need new, optional settings for InfluxDB v1 plugin to allow for authenticated access to /debug/vars.
username/password must be supplied if the the InfluxDB config `pprof-auth-enabled = true` is set.

### Proposal:
Please add this support to the InfluxDB v1 plugin.

glinton:
The returned error may contain sensitive information if older versions
of go are used to build. (1.12+ are ok)

Resolves #6588

### DIFF
--- a/plugins/inputs/influxdb/influxdb.go
+++ b/plugins/inputs/influxdb/influxdb.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"net/url"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


danielnelson:
Let's not do this and add the `username` + `password` option. https://github.com/influxdata/telegraf/issues/6588#issuecomment-550074295

PR #6603, opened 2019-10-31
Handle per-plugin flush jitter, fixes #6569

### LINKED ISSUE #6569: Allow unique `flush_jitter` values for different output plugins
## Feature Request

### Proposal:
Allow unique `flush_jitter` values for different output plugins

### Current behavior:
Every output plugin is forced to use the same `flush_jitter`

### Desired behavior:
Specify `flush_jitter` per output plugin in the config file. If no custom `flush_jitter` is defined for a plugin, use the default `flush_jitter ` (similar to how `flush_interval` works now)

### Use case: 
This would be useful for forwarding collections to multiple output plugins that flush at different `flush_interval`s. It would be nice to be able change `flush_jitter` per plugin, otherwise each output plugin is forced to use a `flush_jitter` that fits within the smallest `flush_interval`

dbutler-starry:
This supports setting unique flush intervals per output plugin.

See [#6569](https://github.com/influxdata/telegraf/issues/6569)

### Required for all PRs:

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -503,6 +503,12 @@ func (a *Agent) runOutputs(
 			interval = output.Config.FlushInterval
 		}
 
+		jitter := jitter
+		// Overwrite agent flush_jitter if this plugin has its own.
+		if output.Config.FlushJitter != nil {
+			jitter = *output.Config.FlushJitter
+		}
+
 		wg.Add(1)
 		go func(output *models.RunningOutput) {
 			defer wg.Done()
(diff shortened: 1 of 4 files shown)


dbutler-starry:
I had some internal debate about the data type of `FlushJitter` in the `OutputConfig`.

The `Agent` relies on default values for other output configs (e.g. `FlushInterval`) to indicate that the user has not specified an override value, but the default value for `FlushJitter` is actually a valid value (i.e. the user could specify a jitter of 0).

For this reason I decided to implement `FlushInterval` as an `interface` to imitate "optional" behaviour. I'm open to other options/opinions about different ways to do this.

danielnelson:
Would it work/does it make sense for the value to be a pointer? `*internal.Duration`

dbutler-starry:
It could be done with a pointer, do you like the look of this better?

danielnelson:
Yeah, looks good, can you clear up the conflict and then I can merge.

dbutler-starry:
:+1: Fixed

Issue #6546, opened 2019-10-19
Fibaro plugin: Get battery level from devices


theovassiliou:
## Feature Request

Opening a feature request kicks off a discussion.

### Proposal:
Implement functionality to gather the battery level of devices.

### Current behavior:
Only collects values, power and enery. 

### Desired behavior:
Read (and collect) data from the "batteryLevel": field/variable for the devices in the Fibaro API.

### Use case: 
To build graphs based on the actual battery level per devices. This would also be very helpful to create some predictions on required maintenance of fibaro devices.

theovassiliou:
I have already implemented this feature, defined tests and tested it against a real world fibaro HC2 system. Works like a charm. Would open a pull request if agreed.

danielnelson:
Thanks, yes please open a PR and we can take a look.

sjwang90:
Closing due to inactivity.

PR #6612, opened 2019-11-04
implementation for #6546 fibaro-plugin battery level

theovassiliou:
### Required for all PRs:

This commit provides an extension to the fibaro plugin to record in addition also the battery level of the devices as outlined in feature request #6546

### DIFF
--- a/plugins/inputs/fibaro/README.md
+++ b/plugins/inputs/fibaro/README.md
@@ -34,6 +34,7 @@ Those values could be true (1) or false (0) for switches, percentage for dimmers
     - power (float, when available from device)
     - value (float)
     - value2 (float, when available from device)
+    - battery (float, when available from device)
 
 
 ### Example Output:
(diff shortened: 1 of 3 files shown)


plugins/inputs/fibaro/fibaro.go
```diff
@@ -207,6 +208,12 @@ func (f *Fibaro) Gather(acc telegraf.Accumulator) error {
 			}
 		}
 
+		if device.Properties.Battery != nil {
+			if fValue, err := strconv.ParseFloat(device.Properties.Battery.(string), 64); err == nil {
```
danielnelson: I see above that this is an existing issue, but would you be able to clean it up?  If we declare the struct type as a string we don't need to do this operation, and we avoid a panic if it is not a string.

Up on line 77:
```
Battery string `json:"batteryLevel"`
```

If you could do this for the other `interface{}` types it would be appreciated.

plugins/inputs/fibaro/README.md
```diff
@@ -44,12 +45,12 @@ fibaro,deviceId=10,host=vm1,name=Escaliers,room=Dégagement,section=Pièces\ com
 fibaro,deviceId=13,host=vm1,name=Porte\ fenêtre,room=Salon,section=Pièces\ communes,type=com.fibaro.FGRM222 energy=4.33,power=0.7,value=99,value2=99 1529996807000000000
 fibaro,deviceId=21,host=vm1,name=LED\ îlot\ central,room=Cuisine,section=Cuisine,type=com.fibaro.binarySwitch value=0 1529996807000000000
 fibaro,deviceId=90,host=vm1,name=Détérioration,room=Entrée,section=Pièces\ communes,type=com.fibaro.heatDetector value=0 1529996807000000000
-fibaro,deviceId=163,host=vm1,name=Température,room=Cave,section=Cave,type=com.fibaro.temperatureSensor value=21.62 1529996807000000000
+fibaro,deviceId=163,host=vm1,name=Température,room=Cave,section=Cave,type=com.fibaro.temperatureSensor battery=88 value=21.62 1529996807000000000
```
danielnelson: Looks like you need a comma between the fields to make this valid line protocol.

plugins/inputs/fibaro/fibaro.go
theovassiliou: Yes, I can do this. But I will need a little bit time, as I have to check out on how to test this on my fork. Unfortunately I am not so familliar with all the github features. Would I have to commit a new pull request, or can I work on the exisiting one, somehow? What is the normal way to do this? Sorry for bothering you with this.

danielnelson: You can just make a new commit on your repository to this same branch, then when you push the branch to your remote it will automatically update this pull request.

sjwang90:
@theovassiliou Checking in as it's been awhile. Can you please make the changes above?

plugins/inputs/fibaro/fibaro.go
```diff
@@ -69,11 +69,12 @@ type Devices struct {
...
-		Energy interface{} `json:"energy"`
-		Power  interface{} `json:"power"`
-		Value  interface{} `json:"value"`
-		Value2 interface{} `json:"value2"`
+		Dead    interface{} `json:"dead"`
+		Energy  interface{} `json:"energy"`
+		Power   interface{} `json:"power"`
+		Value   interface{} `json:"value"`
+		Value2  interface{} `json:"value2"`
+		Battery interface{} `json:"batteryLevel"`
```
ssoroka, line 77: if someone wants to pick this up later, this should probably be something more like this, and should check it against the published api. Be considerate of different api version out there in the wild.
```suggestion
		Dead    string `json:"dead"`
		Energy  string `json:"energy"`
		Power   string `json:"power"`
		Value   string `json:"value"`
		Value2  string `json:"value2"`
		Battery float64 `json:"batteryLevel"`
```

ssoroka (changes requested):

ssoroka:
Closing as this looks abandoned and probably doesn't work as is. (battery type differs from official api and will panic)

PR #6629, opened 2019-11-06
Add snmp_trap input plugin

### LINKED ISSUE #4377: SNMP trap collection Telegraf
## Feature Request

### Proposal:
Have telegraf receive passive snmp traps and send them to influxdb.
### Current behavior:
Currently telegraf can get active snmp, via queries to a host, but host send snmp are not collectable with telegraf.  Example would be an snmp trap sent because of a failed power supply, telegraf does not collect the trap and therefore is not able to forward it to influxdb.
### Desired behavior:
Telegraf to receive and forward snmp traps from hosts to influxdb.
ie.
1.) host send traps to telegraf because of a power supply failure.
2.) telegraf receives trap, translates it to influxdb line language using vendor MIB
3.) sends proper influxdb input to influxdb.
4.) to make it perfect, Kapacitor shows alert.
We are monitoring lots of systems, and we want to centralize all monitoring to one monitoring system.  To make this complete with influxdb, grafana and kapacitor, we are missing the snmp trap from hosts, like PSU failover, disk failure, node reboot, etc..

reimda:
### Required for all PRs:

closes #4377


reimda:
to do:
- Mock OID name resolution so unit tests pass even when snmptranslate isn't available
- Add v1 traps.  Version 2c traps seem to work now.
- Format values better, possibly using the same mechanism as the snmp plugin or by adding support for textual convention in the MIB.
- Check on upstream support for inform.

plugins/inputs/snmp_trap/snmp_trap.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	s.wg.Add(1)
+	go func() {
+		defer s.wg.Done()
+
+		//no ip means listen on all interfaces, ipv4 and ipv6
+		err := s.listener.Listen(":" + strconv.FormatUint(uint64(s.Port), 10))
+		if err != nil {
+			s.Errch <- err
```
glinton: This channel isn't read anywhere, nor initialized, probably want to address.

plugins/inputs/snmp_trap/snmp_trap.go
```diff
@@ -0,0 +1,144 @@
...
+			log.Panicf("error in listen: %s", err)
```
glinton: I don't believe `log.Panic` is a preferred method of logging in telegraf

plugins/inputs/snmp_trap/snmp_trap.go
```diff
@@ -0,0 +1,144 @@
...
+
+	return nil
+}
+
+func (s *SnmpTrap) Stop() {
+	s.listener.Close()
+	s.wg.Wait()
+}
+
+func makeTrapHandler(s *SnmpTrap) func(packet *gosnmp.SnmpPacket, addr *net.UDPAddr) {
```
glinton: This feels odd. Was there a reason not to simply pass in the timefunc and accumulator?

plugins/inputs/snmp_trap/snmp_trap.go
```diff
@@ -0,0 +1,144 @@
...
+		tags["trap_version"] = packet.Version.String()
+
+		for _, v := range packet.Variables {
+			//build a name and value for each variable to use as tags
+			//and fields.  defaults are the uninterpreted values
+			name := v.Name
+			value := v.Value
+
+			//use system mibs to resolve the name if possible
+			_, _, oidText, _, err := snmp.SnmpTranslate(v.Name)
```
glinton: Did you mean to use `name` here and on line 128 instead of v.Name?

plugins/inputs/snmp_trap/snmp_trap.go
```diff
@@ -0,0 +1,144 @@
...
+		tm := s.timeFunc()
+		fields := map[string]interface{}{}
+		tags := map[string]string{}
+
+		tags["trap_version"] = packet.Version.String()
+
+		for _, v := range packet.Variables {
+			//build a name and value for each variable to use as tags
+			//and fields.  defaults are the uninterpreted values
+			name := v.Name
```
glinton: I'm not sure what is returned as a name here, but would it be preferred to snake case them to fit better as field/tag keys?

Gopkg.lock
```diff
@@ -513,6 +513,14 @@
   revision = "636bf0302bc95575d69441b25a2603156ffdddf1"
   version = "v1.1.1"
 
+[[projects]]
+  digest = "1:68c64bb61d55dcd17c82ca0b871ddddb5ae18b30cfe26f6bfd4b6df6287dc2e0"
+  name = "github.com/golang/mock"
```
glinton: Looks like this is unused, is this still in the plans to use?

plugins/inputs/snmp_trap/snmp_trap_test.go
```diff
@@ -0,0 +1,149 @@
...
+
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/testutil"
+
+	"github.com/influxdata/telegraf/plugins/inputs/snmp"
+	"github.com/stretchr/testify/require"
+)
+
+func TestTranslate(t *testing.T) {
+	mibName, oidNum, oidText, conversion, err := snmp.SnmpTranslate(".1.3.6.1.6.3.1.1.5.1")
```
glinton: Is there a reason this test isn't in `inputs/snmp/snmp_test.go`, as it only tests a function from that package?

plugins/inputs/snmp_trap/snmp_trap_test.go
```diff
@@ -0,0 +1,149 @@
...
+	if err != nil {
+		t.Errorf("SendTrap() err: %v", err)
+	}
+
+	return now
+}
+
+// TestReceiveTrap
+func TestReceiveTrap(t *testing.T) {
+	const port = 12399 //todo: find unused port
```
danielnelson: Use port 0 to autoselect an unused port.

glinton: Looks like he's trying to send to the that socket though, and it's currently not plumbed to get the port back that go selects.

reimda: Listening on port 0 would be ideal but I couldn't see a way to do it without changing gosnmp.  Since gosnmp uses net.ListenUDP() listening on port 0 should work, but (*TrapListener) Listen() doesn't expose the connection you need to get which port was autoselected.

plugins/inputs/snmp_trap/snmp_trap_test.go
reimda: I put it here so these tests fail quickly if somebody changes snmp.SnmpTranslate() without knowing that it's used by snmp_trap.

plugins/inputs/snmp_trap/snmp_trap.go
reimda: I meant to use v.Name in both places but the logic here could be clearer.  I think I'll rearrange things a little.

plugins/inputs/snmp_trap/snmp_trap.go
reimda: v.Name is a numeric OID, something like ".1.3.6.1.6.3.1.1.5.1".  If MIB lookup is set up we'll get something like "coldStart" that we use instead.  SNMP convention is camel case and well defined in many MIBs over many decades.  I was going to leave field names as camel case.  If there's enough demand we could consider a configuration option to automatically convert to snake case.

plugins/inputs/snmp_trap/snmp_trap.go
reimda: Passing in timefunc and accumulator would work for the code as it is.  I expect to need more info from the SnmpTrap struct in the future, like settings on how to interpret or convert data from the trap.  I think it's going to be enough variables that it will make sense to have the whole SnmpTrap struct available.

Would it be less odd to have makeTrapHandler be a method of SnmpTrap?

plugins/inputs/snmp_trap/snmp_trap.go
reimda: It's meant to be used in the test but it's clearly not working like I wanted if it's not initialized :)

plugins/inputs/snmp_trap/snmp_trap.go
reimda: Switching to log.Errorf(), thanks

plugins/inputs/snmp_trap/snmp_trap.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+func (s *SnmpTrap) Start(acc telegraf.Accumulator) error {
+	s.acc = acc
+	s.listener = gosnmp.NewTrapListener()
+	s.listener.OnNewTrap = makeTrapHandler(s)
+	s.listener.Params = gosnmp.Default
+	//s.listener.Params.Logger = log.New(os.Stdout, "", 0)
+
+	//wrap the handler, used in unit tests
```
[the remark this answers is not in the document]
reimda: Thanks.  It looks like I've been doing it wrong all this time :)

plugins/inputs/snmp/snmp.go
```diff
@@ -948,7 +948,7 @@ var snmpTranslateCachesLock sync.Mutex
 var snmpTranslateCaches map[string]snmpTranslateCache
 
 // snmpTranslate resolves the given OID.
-func snmpTranslate(oid string) (mibName string, oidNum string, oidText string, conversion string, err error) {
+func SnmpTranslate(oid string) (mibName string, oidNum string, oidText string, conversion string, err error) {
```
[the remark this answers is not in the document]
reimda: It seems natural that snmp_trap would depend on snmp.  I don't see any benefit to moving the code to a common dir.

I generally don't like the idea of common dirs. It seems like they always hold stuff that would fit better elsewhere.

plugins/inputs/snmp_trap/README.md
```diff
@@ -0,0 +1,33 @@
...
+configurable.
+
+OIDs can be resolved to strings using system MIB files. This is done
+in same way as the SNMP input plugin. See the section "MIB Lookups" in
+the SNMP [README.md](../snmp/README.md) for details.
+
+### Configuration
+```toml
+  ## Port to listen on.  Default 162
+  #port = 162
```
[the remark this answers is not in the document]
reimda: good idea, I'll do this

plugins/inputs/snmp_trap/README.md
```diff
@@ -0,0 +1,33 @@
...
+  - tags:
+	- trap_name (string, value from SNMPv2-MIB::snmpTrapOID.0 PDU)
+	- trap_version (string, "1" or "2c" or "3")
+  - fields:
+	- $NAME (the type is variable and depends on the PDU)
+	- $NAME_type (string, description of the Asn1BER type of the PDU.  Examples: "Integer", "TimeTicks", "IPAddress")
+
+### Example Output
+```
+> snmp_trap,host=debian,trap_name=coldStart,trap_version=2c sysUpTimeInstance_type="TimeTicks",snmpTrapEnterprise.0="linux",snmpTrapEnterprise.0_type="ObjectIdentifier",sysUpTimeInstance=1i 1573078928344595213
```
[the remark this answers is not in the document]
reimda: ok, thanks

plugins/inputs/snmp_trap/snmp_trap.go
```diff
@@ -0,0 +1,144 @@
...
+	wg       sync.WaitGroup
+	timeFunc func() time.Time
+
+	makeHandlerWrapper func(func(packet *gosnmp.SnmpPacket, addr *net.UDPAddr)) func(packet *gosnmp.SnmpPacket, addr *net.UDPAddr)
+	Errch              chan error
+}
+
+var sampleConfig = `
+  ## Port to listen on.  Default 162
+  #port = 162
```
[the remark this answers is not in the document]
reimda: This would require a change to gosnmp.  "udp" is hard coded in (t *TrapListener) Listen().

https://github.com/soniah/gosnmp/blob/master/trap.go#L154

plugins/inputs/snmp_trap/snmp_trap.go
```diff
@@ -0,0 +1,144 @@
...
+func (s *SnmpTrap) Init() error {
+	return nil
+}
+
+func (s *SnmpTrap) Start(acc telegraf.Accumulator) error {
+	s.acc = acc
+	s.listener = gosnmp.NewTrapListener()
+	s.listener.OnNewTrap = makeTrapHandler(s)
+	s.listener.Params = gosnmp.Default
+	//s.listener.Params.Logger = log.New(os.Stdout, "", 0)
```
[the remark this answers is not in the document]
reimda: Yep, I'll clear them out before the real PR

plugins/inputs/snmp_trap/snmp_trap.go
```diff
@@ -0,0 +1,144 @@
...
+	if nil != s.makeHandlerWrapper {
+		s.listener.OnNewTrap = s.makeHandlerWrapper(s.listener.OnNewTrap)
+	}
+
+	s.wg.Add(1)
+	go func() {
+		defer s.wg.Done()
+
+		//no ip means listen on all interfaces, ipv4 and ipv6
+		err := s.listener.Listen(":" + strconv.FormatUint(uint64(s.Port), 10))
```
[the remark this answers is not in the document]
reimda: I'll do this.  The test uses Listening() to make sure things are ready before sending a trap.  It will be nicer to move that code from the test into Start().  It will get rid of Errch too.

plugins/inputs/snmp_trap/snmp_trap.go
danielnelson: Normally we use snake_case in Telegraf, even when the source system uses another style, but for this plugin I think we should mirror the snmp input behavior.

plugins/inputs/snmp/snmp.go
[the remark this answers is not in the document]
danielnelson: Perhaps, we have some things that are either shared by many plugins, mostly these are in `internal` right now, which ought to be moved outside of this special directory so that they may be used by plugins outside of the repo.  I would be inclined to agree it doesn't make sense to have these in a common directory, but I don't want too many top level directories.  We also have some functions that are used from both, for example the kafka input and the kafka output where it's less clear which plugin should own the logic.

Gopkg.lock
reimda: It's a new dependency of gosnmp.  I updated to latest gosnmp to get a recent change related to printing the type of a trap variable.

tesibelda:
Great job reimda. Some suggestions:
 - it would be useful to have a snmp_mib tag added. It would be helpfull for post processing filtering of these points (say with kapacitor). When using a large amount of MIBs, some of them might use the same string for snmp_trap.
  - it would be useful to have a snmp_host tag added. It would include the DNS resolved name of the IP address of device sending the trap. This would help identify the device having the issue reported by the trap.

reimda:
Thanks for the feedback @tesibelda 

>     * it would be useful to have a snmp_mib tag added ...

I added this as a tag called trap_mib to match the existing tag trap_name.

>     * it would be useful to have a snmp_host tag added. It would include the DNS resolved name of the IP address of device sending the trap ...

Since we already have the numeric IP address, I added it as a tag called source.  Other inputs use the name "source" for this so it makes sense to follow convention.  I agree a DNS name could be more useful than the IP but I'd like to wrap up this PR.  Adding a setting to do reverse DNS sounds like a great follow-on feature.

plugins/inputs/snmp_trap/snmp_trap.go
```diff
@@ -0,0 +1,147 @@
...
+
+		for _, v := range packet.Variables {
+			// build a name and value for each variable to use as tags
+			// and fields.  defaults are the uninterpreted values
+			name := v.Name
+			value := v.Value
+
+			// use system mibs to resolve the name if possible
+			_, _, oidText, _, err := snmp.SnmpTranslate(v.Name)
+			if nil == err {
```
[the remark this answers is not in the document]
reimda: Wouldn't dropping it entirely make it harder to set up and debug?  How about always adding the oid in one tag and optionally adding other tags for mib/name if they're available?  Then even if someone doesn't have snmptranslate and net-snmp set up right, at least won't have missing data.

plugins/inputs/snmp_trap/snmp_trap.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	s.wg.Add(1)
+	go func() {
+		defer s.wg.Done()
+
+		// no ip means listen on all interfaces, ipv4 and ipv6
+		err := s.listener.Listen(s.ServiceAddress)
+		if err != nil {
+			s.Log.Errorf("error in listen: %s", err)
```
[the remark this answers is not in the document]
reimda: Would it be better to have Start() return the error and not log anything?

plugins/inputs/snmp_trap/snmp_trap.go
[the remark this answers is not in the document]
danielnelson: It would be a bit harder to debug, you would need to go look at the logfile of the Telegraf instance.  We do have a internal metric for errors, so we could increment this (currently done by logging an error or using acc.AddError).

I still think we should skip the metrics in spite of this, because there isn't any way to clean up bad data in the database.  Once it has been added, it is essentially permanent until and unless the retention policy removes it.

plugins/inputs/snmp_trap/snmp_trap.go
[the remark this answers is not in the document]
danielnelson: Yeah

plugins/inputs/snmp_trap/snmp_trap.go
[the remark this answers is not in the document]
danielnelson: > Once it has been added, it is essentially permanent until and unless the retention policy removes it.

I guess this isn't completely true, you could always use:
```
drop series from snmp_tap where 'oid' == '0.1.2.3'
```

plugins/inputs/snmp_trap/snmp_trap.go
```diff
@@ -0,0 +1,147 @@
...
+		}
+	}()
+
+	<-s.listener.Listening()
```
danielnelson: Add a info log message with the listening address and port, keeping style with **most** of the other listening plugins.  For example:
```
2019-11-15T00:04:11Z I! [inputs.socket_listener] Listening on udp://0.0.0.0:8094
```

plugins/inputs/snmp_trap/snmp_trap.go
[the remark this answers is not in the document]
reimda: I'll change it to drop the trap and log a warning.  I agree putting data into the database with numeric

plugins/inputs/snmp_trap/README.md
```diff
@@ -0,0 +1,40 @@
...
+  ##   example: "udp://127.0.0.1:1234"
+  # service_address = udp://:162
+```
+
+### Metrics
+
+- snmp_trap
+  - tags:
+	- source (string, IP address of trap source)
+	- trap_name (string, value from SNMPv2-MIB::snmpTrapOID.0 PDU)
```
danielnelson: Update names here

plugins/inputs/snmp_trap/snmp_trap.go
```diff
@@ -0,0 +1,251 @@
...
+	inputs.Add("snmp_trap", func() telegraf.Input {
+		return &SnmpTrap{
+			timeFunc:       time.Now,
+			ServiceAddress: "udp://:162",
+		}
+	})
+}
+
+func realExecCmd(arg0 string, args ...string) ([]byte, error) {
+	return exec.Command(arg0, args...).Output()
```
danielnelson: Use the functions in `internal/exec.go` with a timeout.

plugins/inputs/snmp_trap/snmp_trap.go
```diff
@@ -0,0 +1,251 @@
...
+	s.listener = gosnmp.NewTrapListener()
+	s.listener.OnNewTrap = makeTrapHandler(s)
+	s.listener.Params = gosnmp.Default
+
+	// wrap the handler, used in unit tests
+	if nil != s.makeHandlerWrapper {
+		s.listener.OnNewTrap = s.makeHandlerWrapper(s.listener.OnNewTrap)
+	}
+
+	split := strings.SplitN(s.ServiceAddress, "://", 2)
```
danielnelson: I'd have probably used `url.Parse`

plugins/inputs/snmp_trap/snmp_trap.go
```diff
@@ -0,0 +1,251 @@
...
+	case err := <-s.errCh:
+		return err
+	}
+
+	return nil
+}
+
+func (s *SnmpTrap) Stop() {
+	s.listener.Close()
+	_ = <-s.errCh
```
danielnelson: Not required, will be garbage collected, this would allow you to make the channel a local variable.

plugins/inputs/snmp_trap/snmp_trap.go
```diff
@@ -0,0 +1,251 @@
...
+	s.cacheLock.Lock()
+	defer s.cacheLock.Unlock()
+	s.cache[oid] = e
+}
+
+func (s *SnmpTrap) snmptranslate(oid string) (e mibEntry, err error) {
+	var out []byte
+	out, err = s.execCmd("snmptranslate", "-Td", "-Ob", "-m", "all", oid)
+
+	//var e mibEntry
```
danielnelson: delete, also line 239

plugins/inputs/snmp_trap/snmp_trap.go
```diff
@@ -0,0 +1,251 @@
...
+		return e, err
+	}
+
+	scanner := bufio.NewScanner(bytes.NewBuffer(out))
+	ok := scanner.Scan()
+	if err = scanner.Err(); !ok && err != nil {
+		return e, err //Errorf(scanner.Err(), "getting OID text")
+	}
+
+	e.oidText = scanner.Text()
```
danielnelson: Optional, I wonder if it would be good to use fmt.Scanf for this text?

plugins/inputs/snmp_trap/snmp_trap.go
```diff
@@ -0,0 +1,251 @@
...
+
+			e, err := s.lookup(v.Name)
+			if nil != err {
+				s.Log.Errorf("Error resolving OID: %v", err)
+				return
+			}
+
+			name := e.oidText
+
+			fields[name] = value
```
danielnelson: What are the possible types?  Do we need to do any conversions?

plugins/inputs/snmp_trap/snmp_trap.go
```diff
@@ -0,0 +1,251 @@
...
+			// only handles textual convention for ip and mac
+			// addresses
+
+			switch v.Type {
+			case gosnmp.ObjectIdentifier:
+				val, ok := v.Value.(string)
+				var e mibEntry
+				var err error
+				if !ok {
+					s.Log.Errorf("Error getting value OID: %v", err)
```
danielnelson: `err` is going to be nil here...

plugins/inputs/snmp_trap/snmp_trap.go
reimda: Instead of making s.errCh local to Start() I left it in the SnmpTrap struct and changed Stop() to log if there's an error.

plugins/inputs/snmp_trap/snmp_trap.go
reimda: Thanks for the suggestion.  I'll leave it because it's about as much work to use url.Parse and then build a new string with the host and port.  If we ever support other transports we should switch to url.Parse

plugins/inputs/snmp_trap/snmp_trap.go
reimda: I suspect we'll need to do more work converting and formatting.  To do that I need some real world traps to test with.  I was hoping we could get this out and ask people watching the feature request for test cases.

PR #6640, opened 2019-11-08
Add Splunk MultiMetric support

ronnocol:
### Required for all PRs:

Adds support for Splunk 8.0's multi metric format.

### DIFF
--- a/plugins/serializers/splunkmetric/splunkmetric.go
+++ b/plugins/serializers/splunkmetric/splunkmetric.go
@@ -13,6 +13,23 @@ type serializer struct {
 	SplunkmetricMultiMetric bool
 }
 
+type CommonTags struct {
+	Time   float64
+	Host   string
+	Index  string
+	Source string
+	Fields map[string]interface{}
+}
+
+type HECTimeSeries struct {
+	Time   float64                `json:"time"`
+	Event  string                 `json:"event"`
+	Host   string                 `json:"host,omitempty"`
+	Index  string                 `json:"index,omitempty"`
+	Source string                 `json:"source,omitempty"`
+	Fields map[string]interface{} `json:"fields"`
+}
+
 // NewSerializer Setup our new serializer
 func NewSerializer(splunkmetric_hec_routing bool, splunkmetric_multimetric bool) (*serializer, error) {
 	/*	Define output params */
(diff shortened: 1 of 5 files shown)


plugins/serializers/splunkmetric/splunkmetric.go
```diff
@@ -9,12 +9,16 @@ import (
...
 
 type serializer struct {
-	HecRouting bool
+	HecRouting              bool
+	SplunkmetricMultiMetric bool
 }
 
-func NewSerializer(splunkmetric_hec_routing bool) (*serializer, error) {
+// NewSerializer Setup our new serializer
+func NewSerializer(splunkmetric_hec_routing bool, splunkmetric_multimetric bool) (*serializer, error) {
```
danielnelson: We should probably switch to passing in enum types, similar to what was done in the `influx` serializer with UintSupport.  This is just much more readable at the call point.

plugins/serializers/splunkmetric/splunkmetric.go
```diff
@@ -106,14 +140,54 @@ func (s *serializer) createObject(metric telegraf.Metric) (metricGroup []byte, e
...
+			case true:
+				// Output the data as a fields array and host,index,time,source overrides for the HEC.
+				metricJson, err = json.Marshal(dataGroup)
+			default:
+				// Just output the data and the time, useful for file based outuputs
+				dataGroup.Fields["time"] = dataGroup.Time
+				metricJson, err = json.Marshal(dataGroup.Fields)
+			}
+
+			metricGroup = append(metricGroup, metricJson...)
```
danielnelson: Check the error before appending.

plugins/serializers/splunkmetric/splunkmetric.go
```diff
@@ -66,37 +70,67 @@ func (s *serializer) createObject(metric telegraf.Metric) (metricGroup []byte, e
...
+		} else if n == "index" {
+			commonTags.Index = t
+		} else if n == "source" {
+			commonTags.Source = t
+		} else {
+			commonTags.Fields[n] = t
+		}
+	}
+	commonTags.Time = float64(metric.Time().UnixNano()) / float64(1000000000)
+	switch s.SplunkmetricMultiMetric {
```
danielnelson: I suggest having two functions, one for multi metric and one for single.  The idea being to keep these tasks as separate as possible and reduce the function length so that it can be more easily understood.

plugins/serializers/splunkmetric/splunkmetric.go
ronnocol: done

plugins/serializers/splunkmetric/splunkmetric.go
ronnocol: done

plugins/serializers/splunkmetric/splunkmetric.go
ronnocol: I don't understand what your ask here is. Happy to make changes with a little more direction.

danielnelson: This isn't super important, so we can do it later.  What I've found is that having multiple boolean parameters to a function is somewhat hard to read.  Compare these calls of the functions:
```
splunkmetric.NewSerializer(true, false)
splunkmetric.NewSerializer(HecRouting, NoMultiMetric)
```

Another advantage of this is that it is harder to mix up the argument order without getting a compile error.

Here is an example of defining a boolean type in the influx serializer:
https://github.com/influxdata/telegraf/blob/7ff6ec19630d4c0e9da61ba3e6a048797b37bf99/plugins/serializers/influx/influx.go#L17-L22

Again though, this isn't urgent, we can add it in a future change.

PR #6641, opened 2019-11-09
Add source and port tags to jenkins plugin

### LINKED ISSUE #6359: Add node name prefix to Jenkins input plugin so that we can separate multiple master on the same Docker instance.
Add node name prefix to Jenkins input plugin so that we can separate multiple masters on the same Docker instance. 

### Proposal:
```
[[inputs.jenkins]]
url = "http://localhost:8080"
username = "username"
password = "pass"
nodeprefix = XX

[[inputs.jenkins]]
url = "http://localhost:8081"
username = "username"
password = "pass"
nodeprefix = YY
```

The nodeprefix string should the be concatenated to the node name if it exist in the 

### Current behavior:
All Jenkins masters will have node name "master" when telegraf is run outside the container which makes it impossible to separate the node metrics in influx.   

### Desired behavior:
Separation between node metrics in influx when multiple Jenkins masters are present on the same host. 

### Use case: 
Running multiple Jenkins masters in docker on the same host I would guess is a common use case. The Telegraf Jenkins plugin would be mor effictive in this scenario if we could have this feature.

nwneisen:
### Required for all PRs:

Resolves #6359.

### DIFF
--- a/plugins/inputs/jenkins/jenkins.go
+++ b/plugins/inputs/jenkins/jenkins.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"net/url"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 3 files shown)


plugins/inputs/jenkins/jenkins.go
```diff
@@ -193,6 +193,10 @@ func (j *Jenkins) gatherNodeData(n node, acc telegraf.Accumulator) error {
 	tags["disk_path"] = monitorData.HudsonNodeMonitorsDiskSpaceMonitor.Path
 	tags["temp_path"] = monitorData.HudsonNodeMonitorsTemporarySpaceMonitor.Path
 
+	urlTokens := strings.Split(j.URL, ":")
+	tags["source"] = urlTokens[1][2:]
+	tags["port"] = urlTokens[2]
```
danielnelson: Before doing any indexing make sure the expected number of items are in the array, otherwise this could panic.  However, for URL parsing, its best to just use the `url.Parse` function.

nwneisen: I'll switch this over.  I saw `net.SplitHostPort()` which didn't like this url format  but not `url.parse`.

PR #6645, opened 2019-11-11
Feature/iisproc

rezarms:
### Required for all PRs:

- [ +] Signed [CLA](https://influxdata.com/community/cla/).
- [ +] Associated README.md updated.
- [+ ] Has appropriate unit tests.


danielnelson:
Thanks for sharing your plugin, however I don't think we ought to include this as part of this Telegraf repository.  We are moving away from including plugins which execute another command, as these are often a lot of work to maintain and can be ran as scripts to the `exec` plugin.  It's not a hard rule but in general this is the direction we are going when including new plugins.

PR #6647, opened 2019-11-11
Fix known mysql type conversion issues

### LINKED ISSUE #5529: Critical/Major Bug related to MySQL Input -- "field type conflict" on ssl_verify_depth / ssl_ctx_verify_depth
Over the weekend, without any user changing either the telegraf version, Influxdb version, or any configuration, our active thread mysql data started not reporting to influxdb. After a bit of work tracking down the issue, we tracked it back to two fields being created within out influxdb database with the same name as fields that already existed, but having the wrong fieldType. This causes Telegraf to be unable to write any metrics related to this to Influxdb, thus making our monitoring one legged, and not reporting all useful data needed to drive monitoring. 

After searching and searching, there is no method to fix this as far as I can tell, either. If InfluxDB allowed us to either alter table, or delete fields, then this would be fixed with no issue, but it appears neither of these are supported within influxdb. The only method I have seemed to run across is to delete the entire mysql measurement (losing all data, for all hosts, for all time for MySQL), and then allowing the telegraf agent to recreate it when writing new metrics to the DB. This seems crazy, and there MUST be a better way. Please tell me there is a better way to recover from such a simple error (at least in my mind, but I am used to MySQL..)

This user seems to have the same issue as we have, yet there seems to be no fix presented to them: https://github.com/influxdata/telegraf/issues/5055 

ANY HELP WOULD BE USEFUL!
Thank you!

### LINKED ISSUE #6646: inputs.mysql: field type conflict: input field "gtid_mode" - string vs integer
# Relevant telegraf.conf:

```toml
[[inputs.mysql]]
  servers = ["monitoring:xxx@tcp(hostname:3306)/?tls=false"]
  metric_version = 2
  interval_slow = "1m"
```

### System info:

Telegraf 1.12.4

Remote mysql version: 5.7.22-log, 5.7.26-log

The two mysql servers are Amazon RDS in master-slave setup. It should be reproducible outside AWS as well.

### Steps to reproduce:

Monitor two mysql servers with different `gtid_mode` variable. One with boolean-like value and the other without.

```
mysql> show global variables like '%gtid_mode%';
+---------------+----------------+
| Variable_name | Value          |
+---------------+----------------+
| gtid_mode     | OFF_PERMISSIVE |
+---------------+----------------+
```

```
mysql> show global variables like '%gtid_mode%';
+---------------+-------+
| Variable_name | Value |
+---------------+-------+
| gtid_mode     | OFF   |
+---------------+-------+
```

### Expected behavior:

No errors.

### Actual behavior:

```
Nov 11 18:12:09 monitoring telegraf[8487]: 2019-11-11T18:12:09Z E! [outputs.influxdb] when writing to [https://influxdb.foo.com:8086]: received error partial write: field type conflict: input field "gtid_mode" on measurement "mysql_variables" is type string, already exists as type integer dropped=1; discarding points
```

### Additional info:

From the MySQL doc, it seems that `gtid_mode` could contain four different values and should never be treated as boolean. I guess, that's where this integer is coming from.

danielnelson:
Add the ability to have per field conversion functions for the global status and global variables queries.  These queries both return tables of the form:
```
+----------------+---------------+------+-----+---------+-------+
| Field          | Type          | Null | Key | Default | Extra |
+----------------+---------------+------+-----+---------+-------+
| VARIABLE_NAME  | varchar(64)   | NO   |     |         |       |
| VARIABLE_VALUE | varchar(2048) | NO   |     |         |       |
+----------------+---------------+------+-----+---------+-------+
```

closes #5529
closes #6646

### Required for all PRs:

### DIFF
--- a/plugins/inputs/mysql/v2/convert.go
+++ b/plugins/inputs/mysql/v2/convert.go
@@ -79,6 +79,10 @@ var GlobalVariableConversions = map[string]ConversionFunc{
 }
 
 func ConvertGlobalStatus(key string, value sql.RawBytes) (interface{}, error) {
+	if bytes.Equal(value, []byte("")) {
+		return nil, nil
+	}
+
 	if conv, ok := GlobalStatusConversions[key]; ok {
 		return conv(value)
 	}
@@ -87,6 +91,10 @@ func ConvertGlobalStatus(key string, value sql.RawBytes) (interface{}, error) {
 }
 
 func ConvertGlobalVariables(key string, value sql.RawBytes) (interface{}, error) {
+	if bytes.Equal(value, []byte("")) {
+		return nil, nil
+	}
+
 	if conv, ok := GlobalVariableConversions[key]; ok {
 		return conv(value)
 	}
--- a/plugins/inputs/mysql/v2/convert.go
+++ b/plugins/inputs/mysql/v2/convert.go
@@ -46,14 +46,6 @@ func ParseGTIDMode(value sql.RawBytes) (interface{}, error) {
 	}
 }
 
-var GlobalStatusConversions = map[string]ConversionFunc{
-	"ssl_ctx_verify_depth": ParseInt,
-}
-
-var GlobalVariableConversions = map[string]ConversionFunc{
-	"gtid_mode": ParseGTIDMode,
-}
-
 func ParseValue(value sql.RawBytes) (interface{}, error) {
 	if bytes.EqualFold(value, []byte("YES")) || bytes.Compare(value, []byte("ON")) == 0 {
 		return 1, nil
@@ -75,6 +67,15 @@ func ParseValue(value sql.RawBytes) (interface{}, error) {
 	}
 
 	return nil, fmt.Errorf("unconvertible value: %q", string(value))
+}
+
+var GlobalStatusConversions = map[string]ConversionFunc{
+	"ssl_ctx_verify_depth": ParseInt,
+	"ssl_verify_depth":     ParseInt,
+}
+
+var GlobalVariableConversions = map[string]ConversionFunc{
+	"gtid_mode": ParseGTIDMode,
 }
 
 func ConvertGlobalStatus(key string, value sql.RawBytes) (interface{}, error) {
--- a/plugins/inputs/mysql/v2/convert.go
+++ b/plugins/inputs/mysql/v2/convert.go
@@ -1,0 +1,94 @@
+package v2
+
+import (
+	"bytes"
+	"database/sql"
+	"fmt"
+	"strconv"
+)
+
+type ConversionFunc func(value sql.RawBytes) (interface{}, error)
+
+func ParseInt(value sql.RawBytes) (interface{}, error) {
+	v, err := strconv.ParseInt(string(value), 10, 64)
+
+	// Ignore ErrRange.  When this error is set the returned value is "the
+	// maximum magnitude integer of the appropriate bitSize and sign."
+	if err, ok := err.(*strconv.NumError); ok && err.Err == strconv.ErrRange {
+		return v, nil
+	}
+
+	return v, err
+}
+
+func ParseBoolAsInteger(value sql.RawBytes) (interface{}, error) {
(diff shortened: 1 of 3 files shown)


plugins/inputs/mysql/v2/convert.go
```diff
@@ -0,0 +1,94 @@
...
+
+type ConversionFunc func(value sql.RawBytes) (interface{}, error)
+
+func ParseInt(value sql.RawBytes) (interface{}, error) {
+	v, err := strconv.ParseInt(string(value), 10, 64)
+
+	// Ignore ErrRange.  When this error is set the returned value is "the
+	// maximum magnitude integer of the appropriate bitSize and sign."
+	if err, ok := err.(*strconv.NumError); ok && err.Err == strconv.ErrRange {
+		return v, nil
```
reimda: Are we expecting ints bigger than int64 or is this here to be defensive?

plugins/inputs/mysql/v2/convert.go
```diff
@@ -0,0 +1,94 @@
...
+	// https://dev.mysql.com/doc/refman/8.0/en/replication-mode-change-online-concepts.html
+	v := string(value)
+	switch v {
+	case "OFF":
+		return int64(0), nil
+	case "ON":
+		return int64(1), nil
+	case "OFF_PERMISSIVE":
+		return int64(0), nil
+	case "ON_PERMISSIVE":
```
reimda: Is the permissive aspect of this mode not important?  I'm a little concerned that we're throwing away information.  Is that just for backwards compatibility?

plugins/inputs/mysql/v2/convert.go
```diff
@@ -0,0 +1,94 @@
...
+		return int64(1), nil
+	default:
+		return nil, fmt.Errorf("unrecognized gtid_mode: %q", v)
+	}
+}
+
+var GlobalStatusConversions = map[string]ConversionFunc{
+	"ssl_ctx_verify_depth": ParseInt,
+}
```
reimda: Issue #5529 also mentions type problems with ssl_verify_depth.  Shouldn't that also have an entry here?

plugins/inputs/mysql/v2/convert.go
```diff
@@ -0,0 +1,94 @@
...
+
+var GlobalVariableConversions = map[string]ConversionFunc{
+	"gtid_mode": ParseGTIDMode,
+}
+
```
reimda: gtid_mode, ssl_ctx_verify_depth, ssl_verify_depth are the fields that have caused issues for us but I wonder if there are others that will be a problem in the future.  Were you able to find a list of these changes in mysql API docs for instance?  I wonder if adding them one at a time will be a whack-a-mole style headache that we could prevent by finding all the changes before they turn into issues.

plugins/inputs/mysql/v2/convert.go
danielnelson: Expecting larger integers, I believe this is actually the cause of the 5529 bug.  In the old strategy we tried to parse as an integer and then fell back to a float64, causing the conflict.  In theory we also have a unsigned integer type that we could work with, but there isn't a clean way to migrate to it.

plugins/inputs/mysql/v2/convert.go
danielnelson: I did do this for backwards compatibility, but since we are converting to an integer we could add a 2/3 value.  This would mean we need to document the meaning of these values as they would be somewhat arbitrary.

plugins/inputs/mysql/v2/convert.go
danielnelson: Yes, I'll fix this.

plugins/inputs/mysql/v2/convert.go
danielnelson: I agree this is going to be whack-a-mole, but I just checked and I don't see a document describing changes in this table.  I believe the table is designed more for interactive use than how we are using it.

Perhaps it would make sense to whitelist in the known variables instead of trying to parse them dynamically?  I also have some concerns about breaking compatibility if I get too aggressive changing the parse behavior of the fields though.

PR #6650, opened 2019-11-12
Add a nameable file tag to file input plugin

### LINKED ISSUE #6644: Add the ability to use filename as a tag in files plugin
## Feature Request

I'm new to influxdb and I was using telegraf plugin to get data stored into several csv files. I tried to include the name of the file as another field of data, but din't find the way to do it through the configuration, and I ended up including the filename as an additional column when generating the file.

### Proposal:
Allow using current filename where the data is readed from to be used as a tag (or a field) when importing data to influxdb with the telegraf plugin files

### Current behavior:
It's not possible 

### Desired behavior:
Define whether to use the filename as tag or not in the configuration of the plugin
```
[[inputs.file]]
  ## Files to parse each interval.
  ## These accept standard unix glob matching rules, but with the addition of
  ##   /var/log/**.log     -> recursively find all .log files in /var/log
  ##   /var/log/*/*.log    -> find all .log files with a parent dir in /var/log
  ##   /var/log/apache.log -> only read the apache log file
  files = ["/var/log/apache/access.log"]

  ## Data format to consume.
  ## more about them here:
  data_format = "csv"

  ## Include filename as custom tag.
  ## Define a tag name to be filled with the source filename, or leave blank to not include it.
  input_filename_custom_tag = "session"

```

### Use case: 
Collecting data from several csv files in a directory, where the csv filename denotes a particular characteristic of the data, like a session name or a particular datasource. Allowing including filename as a tag will allow filtering out data coming from every single file.

nwneisen:
This PR adds the ability to create and name a tag containing the filename of the file parsed using the file input plugin.  Leaving this option blank will cause the tag not to be generated.

Resolves #6644.

### Required for all PRs:

### DIFF
--- a/plugins/inputs/file/file.go
+++ b/plugins/inputs/file/file.go
@@ -3,6 +3,7 @@ package file
 import (
 	"fmt"
 	"io/ioutil"
+	"path/filepath"
 
 	"github.com/influxdata/telegraf"
 	"github.com/influxdata/telegraf/internal/globpath"
@@ -12,6 +13,7 @@ import (
 
 type File struct {
 	Files  []string `toml:"files"`
+	File   string   `toml:"file_tag"`
 	parser parsers.Parser
 
 	filenames []string
@@ -31,6 +33,10 @@ const sampleConfig = `
   ## more about them here:
   ## https://github.com/influxdata/telegraf/blob/master/docs/DATA_FORMATS_INPUT.md
   data_format = "influx"
+  
+  ## Name a tag containing the name of the file the data was parsed from.  Leave empty 
+  ## to disable.
+  # file_tag = "filename"
 `
 
 // SampleConfig returns the default configuration of the Input
@@ -54,6 +60,9 @@ func (f *File) Gather(acc telegraf.Accumulator) error {
 		}
 
 		for _, m := range metrics {
+			if f.File != "" {
+				m.AddTag(f.File, filepath.Base(k))
+			}
 			acc.AddFields(m.Name(), m.Fields(), m.Tags(), m.Time())
 		}
 	}
--- a/plugins/inputs/file/file.go
+++ b/plugins/inputs/file/file.go
@@ -12,9 +12,9 @@ import (
 )
 
 type File struct {
-	Files  []string `toml:"files"`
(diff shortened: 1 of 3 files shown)


plugins/inputs/file/file.go
```diff
@@ -12,6 +13,7 @@ import (
 
 type File struct {
 	Files  []string `toml:"files"`
+	File   string   `toml:"file_tag"`
```
danielnelson: Call this field `FileTag`, this just reduces the cognitive load keeping track of what option a variable maps to.  Also, I feel like we need the tag part to properly describe what the option does.

plugins/inputs/file/file.go
```diff
@@ -31,6 +33,10 @@ const sampleConfig = `
   ## more about them here:
   ## https://github.com/influxdata/telegraf/blob/master/docs/DATA_FORMATS_INPUT.md
   data_format = "influx"
+  
+  ## Name a tag containing the name of the file the data was parsed from.  Leave empty 
+  ## to disable.
+  # file_tag = "filename"
```
danielnelson: Show the default value of this option, so:
```
## Name a tag containing the name of the file the data was parsed from.  Leave empty 
## to disable.
# file_tag = ""
```

plugins/inputs/file/file.go
nwneisen: I was thinking something would be better for the name.  Maybe `input_file_tag_name`?  Anything without tag sounds like it's overriding the file name in some way.

danielnelson: I think we should refer to it as the `file` instead of `input_file` because we already named the `files` option without the input prefix.  The `_name` suffix would be reasonable, though we do have quite a few pre-existing variables that are similar, and I'd like to keep them in the same style unless we the new suffix is enough of an improvement to justify a switch.

In `kafka_consumer` and `mqtt_consumer` we have `topic_tag = ""`, which works the same, we also have `database_tag` in `http_listener` There is also, very soon to be merged, a pull request for the `prometheus` input with a `url_tag = ""`.

danielnelson: So I would stick with `file_tag` as the option name for those reasons.

PR #4911, opened 2018-10-23
Add Filebeat input plugin

dmitryilyin:
Add a new Filebeat 6 Monitoring API input plugin.

- [X] Signed [CLA](https://influxdata.com/community/cla/).
- [X] Associated README.md updated.
- [X] Has appropriate unit tests.

PR #6653, opened 2019-11-13
Add Beat input plugin

nferch:
### Required for all PRs:

As discussed in https://github.com/influxdata/telegraf/pull/4911, combine @dmitryilyin 's filebeat and kafkabeat plugins.

### DIFF
--- a/plugins/inputs/beat/README.md
+++ b/plugins/inputs/beat/README.md
@@ -1,0 +1,143 @@
+# Beat Plugin
+The Beat plugin will collect metrics from the given Beat instances. It is
+known to work with Filebeat and Kafkabeat.
+### Configuration:
+```toml
+  ## An URL from which to read beat-formatted JSON
+  ## Default is "http://127.0.0.1:5066".
+  url = "http://127.0.0.1:5066"
+
+  ## Enable collection of the listed stats
+  ## An empty list means collect all. Available options are currently
+  ## "beat", "libbeat", "system" and "filebeat".
+  # include = ["beat", "libbeat", "filebeat"]
+
+  ## HTTP method
+  # method = "GET"
+
+  ## Optional HTTP headers
+  # headers = {"X-Special-Header" = "Special-Value"}
+
+  ## Override HTTP "Host" header
+  # host_header = "logstash.example.com"
+
+  ## Timeout for HTTP requests
+  # timeout = "5s"
+
+  ## Optional HTTP Basic Auth credentials
+  # username = "username"
+  # password = "pa$$word"
+
+  ## Optional TLS Config
+  # tls_ca = "/etc/telegraf/ca.pem"
+  # tls_cert = "/etc/telegraf/cert.pem"
+  # tls_key = "/etc/telegraf/key.pem"
+  ## Use TLS but skip chain & host verification
+  # insecure_skip_verify = false
+```
+### Measurements & Fields
+- **beat**
+  * Fields:
+    - cpu_system_ticks
+    - cpu_system_time_ms
+    - cpu_total_ticks
+    - cpu_total_time_ms
+    - cpu_total_value
+    - cpu_user_ticks
+    - cpu_user_time_ms
+    - info_uptime_ms
+    - memstats_gc_next
+    - memstats_memory_alloc
+    - memstats_memory_total
+    - memstats_rss
+  * Tags:
+    - beat_beat
+    - beat_host
+    - beat_id
+    - beat_name
+    - beat_version
+
+- **beat_filebeat**
+  * Fields:
+    - events_active
+    - events_added
+    - events_done
+    - harvester_closed
+    - harvester_open_files
+    - harvester_running
+    - harvester_skipped
+    - harvester_started
+    - input_log_files_renamed
+    - input_log_files_truncated
+  * Tags:
+    - beat_beat
+    - beat_host
+    - beat_id
+    - beat_name
+    - beat_version
+
+- **beat_libbeat**
+  * Fields:
+    - config_module_running
+    - config_module_starts
+    - config_module_stops
+    - config_reloads
+    - output_events_acked
+    - output_events_active
+    - output_events_batches
+    - output_events_dropped
+    - output_events_duplicates
+    - output_events_failed
+    - output_events_total
+    - output_type
+    - output_read_bytes
+    - output_read_errors
+    - output_write_bytes
+    - output_write_errors
+    - outputs_kafka_bytes_read
+    - outputs_kafka_bytes_write
+    - pipeline_clients
+    - pipeline_events_active
+    - pipeline_events_dropped
+    - pipeline_events_failed
+    - pipeline_events_filtered
+    - pipeline_events_published
+    - pipeline_events_retry
+    - pipeline_events_total
+    - pipeline_queue_acked
+  * Tags:
+    - beat_beat
+    - beat_host
+    - beat_id
+    - beat_name
+    - beat_version
+
+- **beat_system**
+  * Field:
+    - cpu_cores
+    - load_1
+    - load_15
+    - load_5
+    - load_norm_1
+    - load_norm_15
+    - load_norm_5
+  * Tags:
+    - beat_beat
+    - beat_host
+    - beat_id
+    - beat_name
+    - beat_version
+
+### Example Output:
+```
+$ telegraf --input-filter beat --test
+
+> beat,beat_beat=filebeat,beat_host=node-6,beat_id=9c1c8697-acb4-4df0-987d-28197814f788,beat_name=node-6-test,beat_version=6.4.2,host=node-6
+  cpu_system_ticks=656750,cpu_system_time_ms=656750,cpu_total_ticks=5461190,cpu_total_time_ms=5461198,cpu_total_value=5461190,cpu_user_ticks=4804440,cpu_user_time_ms=4804448,info_uptime_ms=342634196,memstats_gc_next=20199584,memstats_memory_alloc=12547424,memstats_memory_total=486296424792,memstats_rss=72552448 1540316047000000000
+> beat_libbeat,beat_beat=filebeat,beat_host=node-6,beat_id=9c1c8697-acb4-4df0-987d-28197814f788,beat_name=node-6-test,beat_version=6.4.2,host=node-6
+  config_module_running=0,config_module_starts=0,config_module_stops=0,config_reloads=0,output_events_acked=192404,output_events_active=0,output_events_batches=1607,output_events_dropped=0,output_events_duplicates=0,output_events_failed=0,output_events_total=192404,output_read_bytes=0,output_read_errors=0,output_write_bytes=0,output_write_errors=0,outputs_kafka_bytes_read=1118528,outputs_kafka_bytes_write=48002014,pipeline_clients=1,pipeline_events_active=0,pipeline_events_dropped=0,pipeline_events_failed=0,pipeline_events_filtered=11496,pipeline_events_published=192404,pipeline_events_retry=14,pipeline_events_total=203900,pipeline_queue_acked=192404 1540316047000000000
+> beat_system,beat_beat=filebeat,beat_host=node-6,beat_id=9c1c8697-acb4-4df0-987d-28197814f788,beat_name=node-6-test,beat_version=6.4.2,host=node-6
+  cpu_cores=32,load_1=46.08,load_15=49.82,load_5=47.88,load_norm_1=1.44,load_norm_15=1.5569,load_norm_5=1.4963 1540316047000000000
+> beat_filebeat,beat_beat=filebeat,beat_host=node-6,beat_id=9c1c8697-acb4-4df0-987d-28197814f788,beat_name=node-6-test,beat_version=6.4.2,host=node-6
+  events_active=0,events_added=3223,events_done=3223,harvester_closed=0,harvester_open_files=0,harvester_running=0,harvester_skipped=0,harvester_started=0,input_log_files_renamed=0,input_log_files_truncated=0 1540320286000000000
+```
--- a/plugins/inputs/beat/beat.go
+++ b/plugins/inputs/beat/beat.go
@@ -1,0 +1,234 @@
+package beat
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/url"
+	"time"
+
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/internal"
+	"github.com/influxdata/telegraf/internal/choice"
+	"github.com/influxdata/telegraf/plugins/common/tls"
+	"github.com/influxdata/telegraf/plugins/inputs"
+
+	jsonparser "github.com/influxdata/telegraf/plugins/parsers/json"
+)
+
+const sampleConfig = `
+  ## An URL from which to read Beat-formatted JSON
+  ## Default is "http://127.0.0.1:5066".
+  url = "http://127.0.0.1:5066"
+
+  ## Enable collection of the listed stats
+  ## An empty list means collect all. Available options are currently
+  ## "beat", "libbeat", "system" and "filebeat".
+  # include = ["beat", "libbeat", "filebeat"]
+
+  ## HTTP method
+  # method = "GET"
+
+  ## Optional HTTP headers
+  # headers = {"X-Special-Header" = "Special-Value"}
+
+  ## Override HTTP "Host" header
+  # host_header = "logstash.example.com"
+
+  ## Timeout for HTTP requests
+  # timeout = "5s"
+
+  ## Optional HTTP Basic Auth credentials
+  # username = "username"
+  # password = "pa$$word"
+
+  ## Optional TLS Config
+  # tls_ca = "/etc/telegraf/ca.pem"
+  # tls_cert = "/etc/telegraf/cert.pem"
+  # tls_key = "/etc/telegraf/key.pem"
+  ## Use TLS but skip chain & host verification
+  # insecure_skip_verify = false
+`
+
+const description = "Read metrics exposed by Beat"
+
+const suffixInfo = "/"
+const suffixStats = "/stats"
+
+type BeatInfo struct {
+	Beat     string `json:"beat"`
+	Hostname string `json:"hostname"`
+	Name     string `json:"name"`
+	UUID     string `json:"uuid"`
+	Version  string `json:"version"`
+}
+
+type BeatStats struct {
+	Beat     map[string]interface{} `json:"beat"`
+	FileBeat interface{}            `json:"filebeat"`
+	Libbeat  interface{}            `json:"libbeat"`
+	System   interface{}            `json:"system"`
+}
+
+type Beat struct {
+	URL string `toml:"url"`
+
+	Includes []string `toml:"include"`
+
+	Username   string            `toml:"username"`
+	Password   string            `toml:"password"`
+	Method     string            `toml:"method"`
+	Headers    map[string]string `toml:"headers"`
+	HostHeader string            `toml:"host_header"`
+	Timeout    internal.Duration `toml:"timeout"`
+
+	tls.ClientConfig
+	client *http.Client
+}
+
+func NewBeat() *Beat {
+	return &Beat{
+		URL:      "http://127.0.0.1:5066",
+		Includes: []string{"beat", "libbeat", "filebeat"},
+		Method:   "GET",
+		Headers:  make(map[string]string),
+		Timeout:  internal.Duration{Duration: time.Second * 5},
+	}
+}
+
+func (beat *Beat) Init() error {
+	availableStats := []string{"beat", "libbeat", "system", "filebeat"}
+
+	var err error
+	beat.client, err = beat.createHTTPClient()
+
+	if err != nil {
+		return err
+	}
+
+	err = choice.CheckSlice(beat.Includes, availableStats)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (beat *Beat) Description() string {
+	return description
+}
+
+func (beat *Beat) SampleConfig() string {
+	return sampleConfig
+}
+
+// createHTTPClient create a clients to access API
+func (beat *Beat) createHTTPClient() (*http.Client, error) {
+	tlsConfig, err := beat.ClientConfig.TLSConfig()
+	if err != nil {
+		return nil, err
+	}
+
+	client := &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: tlsConfig,
+		},
+		Timeout: beat.Timeout.Duration,
+	}
+
+	return client, nil
+}
+
+// gatherJSONData query the data source and parse the response JSON
+func (beat *Beat) gatherJSONData(url string, value interface{}) error {
+	request, err := http.NewRequest(beat.Method, url, nil)
+	if err != nil {
+		return err
+	}
+
+	if beat.Username != "" {
+		request.SetBasicAuth(beat.Username, beat.Password)
+	}
+	for k, v := range beat.Headers {
+		request.Header.Add(k, v)
+	}
+	if beat.HostHeader != "" {
+		request.Host = beat.HostHeader
+	}
+
+	response, err := beat.client.Do(request)
+	if err != nil {
+		return err
+	}
+
+	defer response.Body.Close()
+
+	return json.NewDecoder(response.Body).Decode(value)
+}
+
+func (beat *Beat) Gather(accumulator telegraf.Accumulator) error {
+	beatStats := &BeatStats{}
+	beatInfo := &BeatInfo{}
+
+	infoUrl, err := url.Parse(beat.URL + suffixInfo)
+	if err != nil {
+		return err
+	}
+	statsUrl, err := url.Parse(beat.URL + suffixStats)
+	if err != nil {
+		return err
+	}
+
+	err = beat.gatherJSONData(infoUrl.String(), beatInfo)
+	if err != nil {
+		return err
+	}
+	tags := map[string]string{
+		"beat_beat":    beatInfo.Beat,
+		"beat_id":      beatInfo.UUID,
+		"beat_name":    beatInfo.Name,
+		"beat_host":    beatInfo.Hostname,
+		"beat_version": beatInfo.Version,
+	}
+
+	err = beat.gatherJSONData(statsUrl.String(), beatStats)
+	if err != nil {
+		return err
+	}
+
+	for _, name := range beat.Includes {
+		var stats interface{}
+		var metric string
+
+		switch name {
+		case "beat":
+			stats = beatStats.Beat
+			metric = "beat"
+		case "filebeat":
+			stats = beatStats.FileBeat
+			metric = "beat_filebeat"
+		case "system":
+			stats = beatStats.System
+			metric = "beat_system"
+		case "libbeat":
+			stats = beatStats.Libbeat
+			metric = "beat_libbeat"
+		default:
+			return fmt.Errorf("unknown stats-type %q", name)
+		}
+		flattener := jsonparser.JSONFlattener{}
+		err := flattener.FullFlattenJSON("", stats, true, true)
+		if err != nil {
+			return err
+		}
+		accumulator.AddFields(metric, flattener.Fields, tags)
+	}
+
+	return nil
+}
+
+func init() {
+	inputs.Add("beat", func() telegraf.Input {
+		return NewBeat()
+	})
+}
--- a/plugins/inputs/beat/beat_test.go
+++ b/plugins/inputs/beat/beat_test.go
@@ -1,0 +1,228 @@
+package beat
+
+import (
+	"fmt"
+	"io/ioutil"
+	"net"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"testing"
+
+	"github.com/influxdata/telegraf/testutil"
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_BeatStats(test *testing.T) {
+	var beat6StatsAccumulator testutil.Accumulator
+	var beatTest = NewBeat()
+	// System stats are disabled by default
+	beatTest.Includes = []string{"beat", "libbeat", "system", "filebeat"}
+	err := beatTest.Init()
+	if err != nil {
+		panic(fmt.Sprintf("could not init beat: %s", err))
+	}
+	fakeServer := httptest.NewUnstartedServer(http.HandlerFunc(func(w http.ResponseWriter, request *http.Request) {
+		var jsonFilePath string
+
+		switch request.URL.Path {
+		case suffixInfo:
+			jsonFilePath = "beat6_info.json"
+		case suffixStats:
+			jsonFilePath = "beat6_stats.json"
+		default:
+			panic("Cannot handle request")
+		}
+
+		data, err := ioutil.ReadFile(jsonFilePath)
+
+		if err != nil {
+			panic(fmt.Sprintf("could not read from data file %s", jsonFilePath))
+		}
+		w.Write(data)
+	}))
+	requestURL, err := url.Parse(beatTest.URL)
+	if err != nil {
+		test.Logf("Can't parse URL %s", beatTest.URL)
+	}
+	fakeServer.Listener, err = net.Listen("tcp", fmt.Sprintf("%s:%s", requestURL.Hostname(), requestURL.Port()))
+	if err != nil {
+		test.Logf("Can't listen for %s: %v", requestURL, err)
+	}
+
+	fakeServer.Start()
+	defer fakeServer.Close()
+
+	err = beatTest.Gather(&beat6StatsAccumulator)
+	if err != nil {
+		test.Logf("Can't gather stats")
+	}
+
+	beat6StatsAccumulator.AssertContainsTaggedFields(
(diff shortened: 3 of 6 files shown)


nferch:
@danielnelson any thoughts or feedback?

danielnelson:
I'll have to get back to you on a full review.  Quick look though overall the code is simple and straightforward, though I'd rather avoid FlattenJSON and just pick the values out manually.  I'm also unsure of the utility of collecting the system metrics via this method.  Since all the stats are in one response I also suggest we remove the options about what sections to collect.

nferch:
Filebeat exposes two different URIs, `/` and `/stats`. `/stats` is the only one I really care about, inclusion of the other metrics was based on @esity's comment in https://github.com/influxdata/telegraf/pull/4911 to make a generalized Beats plugin.

Can't speak to much to the implementation as this PR is largely @dmitryilyin's code, but can change from FlattenJSON at the cost of tying the implementation to the API response du jour.

nferch:
I have been using this PR in limited production and it satisfies my use case, detecting queues/backpressure in Filebeat using the `beat_filebeat.events_active` measurement. @dmitryilyin what was your use case? Do you have a Kafka instance to test against?

axelrtgs:
I have been eagerly watching this and waiting for the feature to come to telegraf. I wanted to leave a comment because i believe the current implementation would also work for my use case

Esity:
@danielnelson Just wondering where we are at with this one. Looks pretty good to me

nferch:
@danielnelson Happy New Year! :) Any further thoughts or feedback?

axelrtgs:
@danielnelson Hope all is well, Been eagerly waiting for this to be implemented. Was wondering when you would have a minute to look at it or if there was anything those of us waiting could do to help speed the process along?

axelrtgs:
@danielnelson @sjwang90 I know you guys have a lot on your plate but this seems to be ready since November. Every few months it keeps getting bumped to the next version  Is there anything I can do to get this looked at and make it easier for you to approve and maybe actually get it into the 1.15 release?

alsargent:
Glad we have this on our radar. Just looking at Elastic Beats plugins (aka "community beats"), there are a number of technologies that Telegraf doesn't yet cover: https://www.elastic.co/guide/en/beats/libbeat/current/community-beats.html

nferch:
@sjwang90 re https://github.com/influxdata/telegraf/pull/4911#issuecomment-730862173, any update or end in sight on this PR? It's been open for over a year with several missed milestone inclusions.

Happy to take a look at the code and ensure it's up to date with the current master branch if there's some indication that'd be time worth spending.

plugins/inputs/beat/beat.go
```diff
@@ -0,0 +1,278 @@
...
+	"github.com/influxdata/telegraf/internal/tls"
+	"github.com/influxdata/telegraf/plugins/inputs"
+
+	jsonparser "github.com/influxdata/telegraf/plugins/parsers/json"
+)
+
+const sampleConfig = `
+  ## An URL from which to read Beat-formatted JSON
+  ## Default is "http://127.0.0.1:5066".
+  url = "http://127.0.0.1:5066"
```
srebhan, line 20: Would it make sense to collect from multiple hosts? Then this should be a list...

plugins/inputs/beat/beat.go
```diff
@@ -0,0 +1,278 @@
...
+  collect_beat_stats = true
+
+  ## Enable the collection if Libbeat stats
+  collect_libbeat_stats = true
+
+  ## Enable the collection of OS level stats
+  collect_system_stats = false
+
+  ## Enable the collection of Filebeat stats
+  collect_filebeat_stats = true
```
srebhan, line 32: Please choose a good default e.g. all is collected and keep the number of uncommented sample config options low. The idea is that one can start by just passing the host and might later customize with the options.

plugins/inputs/beat/beat.go
```diff
@@ -0,0 +1,278 @@
...
+  # method = "GET"
+
+  ## Optional HTTP headers
+  # headers = {"X-Special-Header" = "Special-Value"}
+
+  ## Override HTTP "Host" header
+  # host_header = "logstash.example.com"
+
+  ## Timeout for HTTP requests
+  timeout = "5s"
```
srebhan, line 44: Same. Please comment this option with the (good) default to not confuse new users.

plugins/inputs/beat/beat.go
```diff
@@ -0,0 +1,278 @@
...
+		CollectBeatStats:     true,
+		CollectLibbeatStats:  true,
+		CollectSystemStats:   true,
+		CollectFilebeatStats: true,
+		Method:               "GET",
+		Headers:              make(map[string]string),
+		HostHeader:           "",
+		Timeout:              internal.Duration{Duration: time.Second * 5},
+	}
+}
```
srebhan, line 109: Make this an `Init()` function and later (in `init()`) pass an empty config.

plugins/inputs/beat/beat.go
```diff
@@ -0,0 +1,278 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	client := &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: tlsConfig,
+		},
+		Timeout: beat.Timeout.Duration,
+	}
```
srebhan, line 131: How about e.g. proxy settings?

plugins/inputs/beat/beat.go
```diff
@@ -0,0 +1,278 @@
...
+
+// gatherJsonData query the data source and parse the response JSON
+func (beat *Beat) gatherJsonData(url string, value interface{}) error {
+
+	var method string
+	if beat.Method != "" {
+		method = beat.Method
+	} else {
+		method = "GET"
+	}
```
srebhan, line 144: This default should be set in `Init()`.

plugins/inputs/beat/beat.go
```diff
@@ -0,0 +1,278 @@
...
+
+	request, err := http.NewRequest(method, url, nil)
+	if err != nil {
+		return err
+	}
+
+	if (beat.Username != "") || (beat.Password != "") {
```
srebhan, line 151: Does it make sense to have a password with an empty username? Does this work at all?

plugins/inputs/beat/beat.go
```diff
@@ -0,0 +1,278 @@
...
+		request.SetBasicAuth(beat.Username, beat.Password)
+	}
+	for header, value := range beat.Headers {
```
srebhan, line 154: The local variable `value` might shadow the function argument `value`. To be honest I'm also not sure that I like the passing the result `value` by reference instead of constructing it here and return it. Is there any reason for doing this?

plugins/inputs/beat/beat.go
```diff
@@ -0,0 +1,278 @@
...
+
+import (
+	"encoding/json"
+	"net/http"
+	"net/url"
+	"time"
+
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/internal"
+	"github.com/influxdata/telegraf/internal/tls"
```
srebhan, line 11: This is 
```suggestion
	"github.com/influxdata/telegraf/plugins/common/tls"
```
now I think.

plugins/inputs/beat/beat.go
```diff
@@ -0,0 +1,278 @@
...
+	}
+
+	defer response.Body.Close()
+
+	err = json.NewDecoder(response.Body).Decode(value)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
srebhan, line 173: You might directly return `err` here.

plugins/inputs/beat/beat.go
```diff
@@ -0,0 +1,278 @@
...
+}
+
+func (beat *Beat) gatherInfoTags(url string) (map[string]string, error) {
+	beatInfo := &BeatInfo{}
+
+	err := beat.gatherJsonData(url, beatInfo)
+	if err != nil {
+		return nil, err
+	}
+
```
srebhan, line 183: Couldn't you move this out of this function so that we have the same structure below?

plugins/inputs/beat/beat.go
```diff
@@ -0,0 +1,278 @@
...
+	infoUrl, err := url.Parse(beat.URL + suffixInfo)
+	if err != nil {
+		return err
+	}
+	statsUrl, err := url.Parse(beat.URL + suffixStats)
+	if err != nil {
+		return err
+	}
+
```
srebhan, line 206: Maybe call `beat.gatherJsonData(infoUrl.String(),...)` here and let `beat.gatherInfoTags(...)` only scrape the tags from the return. This then has all the "shoot a query and parse the return" functions on the same level.

plugins/inputs/beat/beat.go
```diff
@@ -0,0 +1,278 @@
...
+	err = beat.gatherJsonData(statsUrl.String(), beatStats)
+	if err != nil {
+		return err
+	}
+
+	if beat.CollectBeatStats {
+		flattenerBeat := jsonparser.JSONFlattener{}
+		err := flattenerBeat.FlattenJSON("", beatStats.Beat)
```
srebhan, line 219: According to the comment, `FlattenJSON()` does *NOT* handle booleans and strings. Do you maybe want `FullFlattenJSON()` here?

plugins/inputs/beat/beat.go
```diff
@@ -0,0 +1,278 @@
...
+		if err != nil {
+			return err
+		}
+		accumulator.AddFields("beat", flattenerBeat.Fields, tags)
+	}
+
+	if beat.CollectFilebeatStats {
+		flattenerBeat := jsonparser.JSONFlattener{}
```
srebhan, line 227: Maybe you can have one instance of the `flattenerBeat` and clear the fields between the calls...

plugins/inputs/beat/beat.go
```diff
@@ -0,0 +1,278 @@
...
+		err := flattenerBeat.FlattenJSON("", beatStats.FileBeat)
+		if err != nil {
+			return err
+		}
+		accumulator.AddFields("beat_filebeat", flattenerBeat.Fields, tags)
```
srebhan, line 232: Your metric name `beat_filebeat` does not follow the convention (default: plugin name). This is especially bad if people use `name_override` to customize the output and suddenly don't get the three distinct streams but an undistinguishible mixture.
How about adding the "information type" (`filebeat` in this case) as a tag (e.g. `source="filebeat"`)?

plugins/inputs/beat/beat.go
```diff
@@ -0,0 +1,278 @@
...
+	}
+
+	if beat.CollectSystemStats {
+		flattenerSystem := jsonparser.JSONFlattener{}
+		err := flattenerSystem.FlattenJSON("", beatStats.System)
+		if err != nil {
+			return err
+		}
+		accumulator.AddFields("beat_system", flattenerSystem.Fields, tags)
+	}
```
srebhan, line 251: Same comments as for the `beat.CollectBeatStats` section.

plugins/inputs/beat/beat.go
```diff
@@ -0,0 +1,278 @@
...
+
+func (beat *Beat) Gather(accumulator telegraf.Accumulator) error {
+	if beat.client == nil {
+		client, err := beat.createHttpClient()
+
+		if err != nil {
+			return err
+		}
+		beat.client = client
+	}
```
srebhan, line 264: Could be done in `Init()`.

plugins/inputs/beat/beat.go
```diff
@@ -0,0 +1,278 @@
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func init() {
+	inputs.Add("beat", func() telegraf.Input {
+		return NewBeat()
```
srebhan, line 276:
```suggestion
		&Beat{}
```
and setting the defaults in an `Init()` function.

plugins/inputs/beat/beat_test.go
```diff
@@ -0,0 +1,225 @@
...
+
+	"github.com/influxdata/telegraf/testutil"
+	"github.com/stretchr/testify/assert"
+)
+
+var beatTest = NewBeat()
+
+var (
+	beat6StatsAccumulator testutil.Accumulator
+)
```
srebhan, line 19: Please move this to the individual tests.

plugins/inputs/beat/beat_test.go
```diff
@@ -0,0 +1,225 @@
...
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"testing"
+
+	"github.com/influxdata/telegraf/testutil"
+	"github.com/stretchr/testify/assert"
+)
+
+var beatTest = NewBeat()
```
srebhan, line 15: Please move this to the individual tests.

plugins/inputs/beat/beat_test.go
```diff
@@ -0,0 +1,225 @@
...
+	defer fakeServer.Close()
+
+	if beatTest.client == nil {
+		client, err := beatTest.createHttpClient()
+
+		if err != nil {
+			test.Logf("Can't createHttpClient")
+		}
+		beatTest.client = client
+	}
```
srebhan, line 56: Why is this necessary? The beat-plugin should connect on its own no?

plugins/inputs/beat/beat.go
```diff
@@ -0,0 +1,278 @@
...
+
+	if (beat.Username != "") || (beat.Password != "") {
+		request.SetBasicAuth(beat.Username, beat.Password)
+	}
+	for header, value := range beat.Headers {
+		request.Header.Add(header, value)
+	}
+	if beat.HostHeader != "" {
+		request.Host = beat.HostHeader
+	}
```
srebhan, line 159: Can also be moved to `Init()`.

plugins/inputs/beat/beat_test.go
```diff
@@ -0,0 +1,225 @@
...
+			func(writer http.ResponseWriter, request *http.Request) {
+				writer.Header().Set("Content-Type", "application/json")
+				if request.URL.String() == suffixInfo {
+					fmt.Fprintf(writer, "%s", string(beat6Info))
+				} else if request.URL.String() == suffixStats {
+					fmt.Fprintf(writer, "%s", string(beat6Stats))
+				} else {
+					test.Logf("Unkown URL: " + request.URL.String())
+				}
+			},
```
srebhan, line 34: The test-data should move to a JSON file you load and dump over HTTP. In the current schema you rely on the compiler to detect that the consts can be compiled out which is... not so nice. :-)

Look at other current test schemes (e.g. into the starlark processor or the `inputs/rabbitmq` tests) to get an idea how this could look like.

plugins/inputs/beat/sample_version6.go
```diff
@@ -0,0 +1,151 @@
+package beat
+
+const beat6Info = `
```
srebhan, line 3: Should be a JSON file. Really. :-)

plugins/inputs/beat/sample_version6.go
```diff
@@ -0,0 +1,151 @@
...
+{
+  "beat": "filebeat",
+  "hostname": "node-6",
+  "name": "node-6-test",
+  "uuid": "9c1c8697-acb4-4df0-987d-28197814f785",
+  "version": "6.4.2"
+}
+`
+
+const beat6Stats = `
```
srebhan, line 13: Same here.

srebhan (changes requested):
Sorry for this taking soooo long!

I have some comments and like you to consider introducing an `Init()` function to set the defaults and one-time OPs there. This could simplify the code.
Furthermore, please rework the tests. Check e.g. `rabbitmq` on how to keep and use the testdata as JSON files.
Beside this there are some more comments in the code...

Let me know if you have questions.

srebhan:
@nferch any news on this?

nferch:
@srebhan from my cursory review the comments make sense, should have some time this weekend to work on it.

plugins/inputs/beat/beat.go
nferch: Think this would be of limited use, I imagine the most common deployments would be a local Telegraf agent and Beat agent per host. So I'd rather address this in a future PR if there's a need that isn't satisfied by a separate configuration per host.

plugins/inputs/beat/beat.go
nferch: Will fix

plugins/inputs/beat/beat.go
nferch: Will fix

plugins/inputs/beat/beat.go
nferch: Will fix

plugins/inputs/beat/beat.go
nferch: Couldn't find many other plugins that support proxies, was there something about this one in particular that you thought necessitated it?

plugins/inputs/beat/beat.go
nferch: will fix

plugins/inputs/beat/beat.go
nferch: will fix

plugins/inputs/beat/beat.go
srebhan: Fair enough, please make as resolved.

plugins/inputs/beat/beat.go
nferch: not sure if I understand the latter part of your comment, this bit is just applying the headers configuration. will rename the variables from the range to avoid them clashing with the value function parameter, LMK if there's other concerns.

plugins/inputs/beat/beat.go
nferch: will fix

plugins/inputs/beat/beat.go
nferch: will fix

plugins/inputs/beat/beat.go
nferch: will fix

plugins/inputs/beat/beat.go
srebhan: The http plugin supports proxies for examples. The reason is that I had quite some software _not_ accepting proxies in the past and that might be a nightmare in enterprise environments. I don't think adding those is super hard either.

plugins/inputs/beat/beat.go
nferch: completely flattened `gatherInfoTags` since it doesn't actually do that much. think this reads much clearer, but open to other ideas.

plugins/inputs/beat/beat.go
nferch: will fix

plugins/inputs/beat/beat.go
nferch: only 4 instances so I'll just leave it as-is, but open to changing it if there's considerable efficiency to be gained...

plugins/inputs/beat/beat.go
srebhan: What I mean is instead of passing `value` to `gatherJsonData()` you could do `value := &BeatInfo{}` *in* that function and then return `value` to the caller.

plugins/inputs/beat/beat.go
nferch: my thinking here was that it made more sense to have separate measurements as there's not much overlap in the responses, it's the interface that is consistent between the endpoints.

don't really have a strong opinion on it though.

plugins/inputs/beat/beat.go
nferch: will fix

plugins/inputs/beat/beat.go
nferch: will fix

plugins/inputs/beat/beat_test.go
nferch: will fix

plugins/inputs/beat/beat_test.go
nferch: will fix

plugins/inputs/beat/beat.go
srebhan: Hmm. I don't see the reason not to declare `flattener := jsonparser.JSONFlattener{}` and use it in the `if`s. The flattener is even stateless, so you save the construction lines... :-)

plugins/inputs/beat/beat.go
nferch: I don't understand, this is setting `Host` on the HTTP request object, which is instantiated per `gatherJsonData` call.

plugins/inputs/beat/beat.go
srebhan: Go ahead as-is for now. I have to discuss this more with Influx-people... Many plugins do this and if you don't be careful you easily screw up the metrics when using `name_override` as it doesn't care about the suffix...
Mark resolved please.

plugins/inputs/beat/beat.go
srebhan: You are absolutely right. Sorry missed this one. Mark as resolved.

plugins/inputs/beat/beat.go
nferch: From what I can tell, the http plugin is the only other plugin that supports proxies. I could see adding it if there's enough prevailing need or consistency with the other plugins that support it. Seems like something that could be easily added if necessary.

plugins/inputs/beat/beat.go
nferch: ahh yeah, that is a bit weird. I'm not the original author so I don't have the full context as far as history or intentions, but I'll fix this one.

nferch: actually, I understand now. :) It's passed a `BeatInfo` and later a `BeatStats`, so it needs to accept an `interface{}`

plugins/inputs/beat/beat.go
nferch: I'm not too familiar with the API, but that doesn't seem to be the case. Simply making the additional calls use the same instance without clearing it (is there a convenient way to do that?) causes the tests to fail.

plugins/inputs/beat/beat_test.go
nferch: will fix

plugins/inputs/beat/beat_test.go
nferch: will fix

plugins/inputs/beat/sample_version6.go
nferch: will fix

plugins/inputs/beat/sample_version6.go
nferch: will fix

nferch:
@srebhan ok, I think I addressed all of your comments. LMK if there's anything I missed.

Would welcome any other patches or real-world testing, as I'm not the original author of the code and only have Filebeat running to test it against.

plugins/inputs/beat/beat.go
```diff
@@ -78,10 +78,10 @@ type BeatStats struct {
...
 	URL string `toml:"url"`
 
-	CollectBeatStats     bool `toml:"collect_beat_stats"`
-	CollectLibbeatStats  bool `toml:"collect_libbeat_stats"`
-	CollectSystemStats   bool `toml:"collect_system_stats"`
-	CollectFilebeatStats bool `toml:"collect_filebeat_stats"`
+	CollectBeatStats     *bool `toml:"collect_beat_stats"`
+	CollectLibbeatStats  *bool `toml:"collect_libbeat_stats"`
+	CollectSystemStats   *bool `toml:"collect_system_stats"`
+	CollectFilebeatStats *bool `toml:"collect_filebeat_stats"`
```
srebhan, line 84: Why do you need those to be pointers?

plugins/inputs/beat/beat.go
```diff
@@ -94,18 +94,49 @@ type Beat struct {
...
+const (
+	defaultURL    = "http://127.0.0.1:5066"
+	defaultMethod = "GET"
+)
+
+var defaultCollectBeatStats = true
+var defaultCollectLibbeatStats = true
+var defaultCollectSystemStats = false
+var defaultCollectFilebeatStats = true
+var defaultTimeout = internal.Duration{Duration: time.Second * 5}
```
srebhan, line 106: Please don't declare them here. Either you declare those in `Init()` or you directly set the defaults there.

plugins/inputs/beat/beat.go
nferch: Since we are setting defaults in `Init()` after the config has been read, it would not be possible to discern a false setting from an unset setting. Open to other ways to achieve this, this is how I've seen it done in other projects.

plugins/inputs/beat/beat.go
nferch: Was necessary to make them vars in order to take the address of them. Thought it was cleaner to separate the default values from the code of setting the defaults.

plugins/inputs/beat/README.md
```diff
@@ -0,0 +1,149 @@
...
+  # method = "GET"
+
+  ## Optional HTTP headers
+  # headers = {"X-Special-Header" = "Special-Value"}
+
+  ## Override HTTP "Host" header
+  # host_header = "logstash.example.com"
+
+  ## Timeout for HTTP requests
+  timeout = "5s"
```
srebhan, line 32: You might also want to comment the default to match `sampleConfig`.

plugins/inputs/beat/beat.go
```diff
@@ -0,0 +1,268 @@
...
+
+func (beat *Beat) Init() error {
+	var err error
+	beat.client, err = beat.createHTTPClient()
```
srebhan, line 110:
```suggestion
	beat.client, err := beat.createHTTPClient()
```

plugins/inputs/beat/beat.go
srebhan: I take your word on this! ;-) If you really don't want to do it now, mark this as resolved...

plugins/inputs/beat/beat.go
```diff
@@ -0,0 +1,268 @@
...
+
+	response, err := beat.client.Do(request)
+	if err != nil {
+		return err
+	}
+
+	defer response.Body.Close()
+
+	err = json.NewDecoder(response.Body).Decode(value)
+	return err
```
srebhan, line 192: You can also do
```suggestion
	return json.NewDecoder(response.Body).Decode(value)
```
Both ok for me.

plugins/inputs/beat/README.md
nferch: good catch, will fix.

plugins/inputs/beat/beat.go
srebhan: Please mark as resolved.

plugins/inputs/beat/beat.go
srebhan: Still open...

plugins/inputs/beat/beat.go
nferch: unless I'm mistaken, it's necessary to declare separately since `:=` will only work for new variables or variables defined in the current block. The suggested change does not compile.

plugins/inputs/beat/beat.go
```diff
@@ -0,0 +1,268 @@
...
+  # collect_beat_stats = true
+
+  ## Enable the collection if Libbeat stats
+  # collect_libbeat_stats = true
+
+  ## Enable the collection of OS level stats
+  # collect_system_stats = false
+
+  ## Enable the collection of Filebeat stats
+  # collect_filebeat_stats = true
```
srebhan, line 32: Concerning our "default value" discussion below, it might be easier to allow the user to specify a list, i.e.
```
collect =["beat", "libbeat", "filebeat"]
```
This way you know if an option was selected (it's in the list) or not. Maybe you can define an empty list to "collect all"...

plugins/inputs/beat/beat.go
srebhan: Hmmm what is the error message? It should work I think...

plugins/inputs/beat/beat.go
srebhan: This is also more extensible in the future if you want to e.g. increase granularity...

plugins/inputs/beat/beat.go
srebhan: See my comment to change to a collection list in the form `collect = ["beat", "libbeat",...]`. You can use `internal/choice` to check the options an see what is in the user-setting.

plugins/inputs/beat/beat.go
nferch: `plugins/inputs/beat/beat.go:109:6: non-name beat.client on left side of :=`

plugins/inputs/beat/beat.go
nferch: I don't have a strong opinion on this, but don't feel much motivation to change it either. The design decision to use booleans predated my commits which sought to integrate the plugins in the simplest way possible. The use of booleans to toggle functionality seemed similar to other plugins so it didn't seem unusual to use them in this way. I don't see how it would make a difference to the end user either, but I obviously have a limited perspective on this.

plugins/inputs/beat/beat.go
srebhan: Sorry, missed the structure part... Should sleep more. :-( Keep as is and mark my this one as resolved.

plugins/inputs/beat/beat.go
srebhan: Yeah it's similar to the plugins where we currently have a lot of trouble extending them without breaking configs... Many plugins therefore move to `include`, `exclude` lists to allow sophisticated filtering of metrics. This is nice for the end-user for cases where you want all but one.

nferch: So is this just a "nice to have" or a blocker for moving forward? I can see how it'd be useful for some more complicated use cases or extensibility, but I don't have much reason to believe that how it is currently implemented doesn't solve the problems of the present. And I'm reluctant to spend much time solving the problems of the future, when they can be easily addressed via a PR once they materialize.

I'd assume that the prospective users of this plugin would prefer having a potentially less-perfect solution sooner than a more future proof solution later, but I defer to anyone else who is following this thread to chime in with their use cases and preferences, as all I have to go by on is my personal one.

axelrtgs: I would vote for nice to have vs blocker. i know i would love to see this out so i can monitor filebeats status even if its not perfect. Although if its not too much work it may be reasonable to do it now since i know fixing things later rarely happens and although i would love to contribute i have 0 knowledge of GO to even pick this up myself in a second PR.

Perhaps merging this functional PR as is with the non perfect config but then opening an issue or second PR to enhance the config may be a compromise between both options.

plugins/inputs/beat/beat.go
ssoroka: set the defaults in init(), and then you don't have to check them at all, and the fields don't have to be pointers.

nferch: @ssoroka This is how it was prior to the comments on https://github.com/influxdata/telegraf/pull/6653#discussion_r530877623

ssoroka: Sorry about that. the correct way to set defaults is when you create the plugin, (around the lowercase `init()` call), By the time `Init()` is called, the config has been loaded over top of the defaults, and has the correct settings. Trying to set them in `Init()` leads to some weirdness, as you've experienced. I'd also like to go ahead with the change to change the bools to a `[]string` slice, because once we merge it the config is locked and we can't break backwards compatibility.

srebhan:
Here is my [implementation](https://github.com/influxdata/telegraf/files/5705290/beat-plugin-include-list.patch.txt)  of converting the booleans to a list of includes... Feel free to use it.

plugins/inputs/beat/beat.go
srebhan: The problem is that we **cannot** fix this *later* as we need to keep backward compatible with the config. Instead of arguing over this again I implemented what is necessary in less time than the discussion took so far...
Please see my comment below with the attached patch (`git apply ...`). Feel free to use it!

plugins/inputs/beat/beat.go
srebhan: @ssoroka this is my fault. I thought it is cleaner to initialize the defaults in `Init()` but completely missed that you cannot know if the boolean was specified as false by the user or it is false due to not being initialized. Again, my fault. Sorry @nferch for the confusion I caused!

plugins/inputs/beat/beat.go
nferch: Thanks, applied.

plugins/inputs/beat/beat.go
nferch: Ok, I've reverted the flow to set defaults in `init()`.

nferch:
@srebhan @ssoroka I believe I've addressed all the comments and feedback. LMK if there's anything else, otherwise I'll squash and rebase so it's ready for merge.

srebhan (approved):
Looks good to me now! Also thx for further cleanup!

srebhan:
@ssoroka can you maybe add a second opinion as I provided the include-list-part before flagging this ready-to-merge!?

ssoroka (approved):
lgtm :)

ssoroka:
thanks for the epic effort here. much appreciated

PR #6655, opened 2019-11-13
Feature/nf conntrack parse stats

gabrielperezs:
If you need to monitor the NAT layer in a Linux server this flags will enable it to export the metrics with telegraf. Our experience was over Kubernetes cluster that was facing port exhaustion, using these changes we were able to monitor the conn status across all the nodes.

### Required for all PRs:


danielnelson:
Thanks for the pull request, are we sure this information is not exposed in `/proc`?  Can you check where the conntrack binary ultimately gets the info from?

gabrielperezs:
Hi, my first approach was to read this information from /proc/net/nf_conntrack but, the existence of /proc/net/nf_conntrack depends on the Linux kernel config CONFIG_NF_CONNTRACK_PROCFS  that is not enabled on all the distributions. I tried over a production server with Ubuntu 16.04 and don't have this location. On the other hand the command conntrack works across all the distributions. This is the easiest way I found to expose this metrics without include a lib with cgo to the same as the conntrack command.

I have the code to read from a file (like /proc/.., /sys/..), I can update this PR is someone knows another path to reach that information.

danielnelson:
From `man conntrack`:

> conntrack provides a full featured userspace interface to the netfilter connection tracking system that is intended to replace the old /proc/net/ip_conntrack interface.

So this does appear to be the recommended method, and this command appears to use netlink, I doubt we would want to duplicate that.  For systems with `CONFIG_NF_CONNTRACK_PROCFS` which files have the equivalent information?

jdef:
> Is in /proc/net/nf_conntrack. Do you want to check if the file exists and read from there and if don't exists use the command, maybe?

I like this idea

powersj:
Hi,

Thanks!

PR #6669, opened 2019-11-17
Add MongoDB uptime_ms

pertu:
MongoDB doc: [serverstatus.uptimeMillis](https://docs.mongodb.com/manual/reference/command/serverStatus/#serverstatus.uptimeMillis)

### DIFF
--- a/plugins/inputs/mongodb/mongostat.go
+++ b/plugins/inputs/mongodb/mongostat.go
@@ -477,7 +477,7 @@ type StatLine struct {
 	IsMongos bool
 	Host     string
 
-	UptimeMillis int64
+	UptimeNanos int64
 
 	// The time at which this StatLine was generated.
 	Time time.Time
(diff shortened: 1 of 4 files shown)


plugins/inputs/mongodb/mongostat.go
```diff
@@ -659,6 +661,8 @@ func NewStatLine(oldMongo, newMongo MongoStatus, key string, all bool, sampleSec
 		Faults:    -1,
 	}
 
+	returnVal.UptimeMillis = newStat.UptimeMillis
```
danielnelson: Looks good, but can you update the field to be in nanoseconds and rename the field to `uptime_ns` to match up with `flushes_total_time_ns`?  See line 742

pertu: I get it why it's better to name things consistently, but I have doubts:
1. It may be misleading to have `_ns` suffix on something that has millisecond resolution.
2. It's clearer for user when metric name matches with MongoDB's command output.

danielnelson: These are good points but I'd still like to make this change for consistency.  We are trying to do this across the board for all duration and time values and nanosecond precision gives us the most flexibility and is consistend with our timestamp format.

pertu: Replaced `uptime_ms` with `uptime_ns`.

PR #6677, opened 2019-11-19
Add timeout option to Sql Server input plugin

### LINKED ISSUE #2824: Add configurable timeout to SQL Server input
Per the conversations in #2051, we need to be able to have a timeout variable for this.  

There are situations which can cause this to take longer to complete than 1m, and having a hard-code timeout, does not seem to make sense.

@sparrc sorry it took so long to get this suggestion in per that thread.

nwneisen:
### Required for all PRs:

Closes #2824 

Tests work with a SQL server instance running locally.  I'm not sure if there's a better way for testing timeouts.

### DIFF
--- a/plugins/inputs/sqlserver/sqlserver.go
+++ b/plugins/inputs/sqlserver/sqlserver.go
@@ -8,16 +8,17 @@ import (
 
 	_ "github.com/denisenkom/go-mssqldb" // go-mssqldb initialization
 	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/internal"
 	"github.com/influxdata/telegraf/plugins/inputs"
 )
 
 // SQLServer struct
 type SQLServer struct {
-	Servers       []string `toml:"servers"`
-	QueryVersion  int      `toml:"query_version"`
-	AzureDB       bool     `toml:"azuredb"`
-	ExcludeQuery  []string `toml:"exclude_query"`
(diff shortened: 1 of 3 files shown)


plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -16,6 +17,7 @@ type SQLServer struct {
 	QueryVersion  int      `toml:"query_version"`
 	AzureDB       bool     `toml:"azuredb"`
 	ExcludeQuery  []string `toml:"exclude_query"`
+	QueryTimeout  int      `toml:"query_timeout"`
```
danielnelson: Use `internal.Duration` for this option, so that the timeout can be specified using the `"5s"` format.

plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -159,8 +164,15 @@ func (s *SQLServer) gatherServer(server string, query Query, acc telegraf.Accumu
 	}
 	defer conn.Close()
 
+	ctx := context.Background()
```
danielnelson: Put this as the first line of the Gather function, then pass the single context down into this function.  This way we only need to setup (and cancel) the timeout once for all servers.

nwneisen:
I have just a basic docker image running that I'm testing against.  I'd be happy to take a look at the other issues and configure the docker image more as needed.

PR #6695, opened 2019-11-21
Allow multiple certificates on same file for x509 input

jonathannegrin:
Allow to load multiple certificates from the same file with the x509 input

### Required for all PRs:

- [] Associated README.md updated. (Not needed)

### DIFF
--- a/plugins/inputs/x509_cert/x509_cert.go
+++ b/plugins/inputs/x509_cert/x509_cert.go
@@ -2,6 +2,7 @@
 package x509_cert
 
 import (
+	"bytes"
 	"crypto/tls"
 	"crypto/x509"
 	"crypto/x509/pkix"
@@ -98,7 +99,7 @@ func (c *X509Cert) getCert(u *url.URL, timeout time.Duration) ([]*x509.Certifica
 		}
 		var certs []*x509.Certificate
 		for {
-			block, rest := pem.Decode(content)
+			block, rest := pem.Decode(bytes.TrimSpace(content))
 			if block == nil {
 				return nil, fmt.Errorf("failed to parse certificate PEM")
 			}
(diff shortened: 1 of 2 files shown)


plugins/inputs/x509_cert/x509_cert.go
```diff
@@ -96,18 +96,24 @@ func (c *X509Cert) getCert(u *url.URL, timeout time.Duration) ([]*x509.Certifica
 		if err != nil {
 			return nil, err
 		}
+		var certs []*x509.Certificate
+		for {
+			block, rest := pem.Decode(content)
+			if block == nil {
+				return nil, fmt.Errorf("failed to parse certificate PEM")
```
danielnelson: I'm concerned this could get hit if there is trailing whitespace after the final block.  Can you add a test for this?

plugins/inputs/x509_cert/x509_cert_test.go
```diff
@@ -140,7 +140,7 @@ func TestGatherLocal(t *testing.T) {
 		{name: "permission denied", mode: 0001, error: true},
 		{name: "not a certificate", mode: 0640, content: "test", error: true},
 		{name: "wrong certificate", mode: 0640, content: wrongCert, error: true},
-		{name: "correct certificate", mode: 0640, content: pki.ReadServerCert()},
+		{name: "correct certificate", mode: 0640, content: pki.ReadServerCert() + pki.ReadCACert()},
```
danielnelson: Minor tweak here, could you add the multiple certificate case as a separate case?

plugins/inputs/x509_cert/x509_cert.go
jonathannegrin: Added trim for whitespaces because it resulted in error and added some test cases

plugins/inputs/x509_cert/x509_cert_test.go
jonathannegrin: Added a few more test cases

Issue #6701, opened 2019-11-22
Set types for stats produced by BasicStats aggregator


kadaan:
### Proposal:
Configure the types for the stats produced by the BasicStats aggregator

### Current behavior:
Currently all stats produced by the BasicStats aggregator are Untyped.

### Desired behavior:
The stats produced by the BasicStats aggregator should be typed as Counter or Gauge where applicable.  Initially, we could go with Counter for the `count` stat and Gauge for the rest.

### Use case: 
It is helpful with a prometheus output to indicate the type of the metrics.

powersj:
Hi,

Looks like you had some back and forth on the PR in the past. I know there was a concern about memory usage, but I want to backup. What is the use-case for adding this information to metrics? It sort of sounds like you want to add Prometheus-based metric types to the stats that are generated by BasicStats?

Thanks

PR #6702, opened 2019-11-22
Configure types for stats produced by BasicStats aggregator

kadaan:
### Required for all PRs:

### Change
Currently all stats produces by BasicStats aggregator have a `telegraf.ValueType` of `Untyped`.  

This PR will change that, making:
1. count stat have a `telegraf.ValueType` of `Counter`
2. other stats have a `telegraf.ValueType` of `Gauge`

`testUtil/Accumulator` has been improved to allow recording and asserting against the expected `telegraf.ValueType`.

**Fix for #6701**


danielnelson:
Before we can merge a change like this we need to improve the way we store type info, right now this is a breaking change for the InfluxDB output as it splits the metrics.  We have an issue open #3670 but no solution that is reasonable with respect to memory usage.

kadaan:
@danielnelson Can you summarize the current memory concern?

kadaan:
@danielnelson Looking at the code quickly, I don't see anything in the InfluxDb output that looks at ValueType.  What is using it that will cause the split?  If the InfluxDb output currently has an unsolved issue, can't we just specifically have that output ignore the ValueType?

danielnelson:
Right now it is only possible for the `telegraf.Metric` to have a single type, so to have a gauge and counter requires the metric to be split.  This results in multiple lines when converted to line protocol, and this can break certain workloads where the data is required to be together.

The memory use I am concerned about is when metrics are queued in the metric buffer, more memory usage means this buffer must be set smaller, we need is a per field value type in a way that won't penalize every output.

kadaan:
Got it. One option would be to group fields by type in the metric.  Seems like it would work and wouldn’t expand the data structure much. One could even flatten the structure in cases where only one field type exists on the metric. 

That said I assume this is a bigger change than to just Telegraf. Is there a way to move forward with supporting types for outputs that can handle this, and allow InfluxDB to tackle this on its own time?

kadaan:
Really what I'm saying is that there a a ton of use cases for Telegraf that don't involve InfluxDb.  If it is possible to enable those to be richer, we can tackle the InfluxDb ones when field level type are figured out.

danielnelson:
Yes, it is possible for sure, and of course this issue goes beyond simple metric types.  The grouping fields idea is interesting, do you mean something like:
```diff
- type Metric struct {
-     Fields []Field
- }
+ type Metric struct {
+     Counters []Field
+     Gauge []Field
+     Untyped []Field
+ }
```

kadaan:
@danielnelson Yes, that is exactly what I am thinking about.  Basically it is columnar storage.

kadaan:
@danielnelson 
> This results in multiple lines when converted to line protocol, and this can break certain workloads where the data is required to be together.

Can't the types be ignored when serializing to the line protocol?  Seems like a little indirection would go a long way.  Current line protocol can just ignore types and produce the current output.  Other protocols could choose to persist type information.

Assume type information was added like mentioned above: 
```
 type Metric struct {
     Counters []Field
     Gauges []Field
     Untyped []Field
}
```

Then, when `addField` it checks for the key across all three field lists: `Counters`, `Gauges`, `Untyped`.  The performance for this should be comparable to the current code.  

And then in serializers/influx we could change `writeMetric` to write all fields as if they were all Untyped.

srebhan:
The suggested changes require a deeper rework on (at least) the output level to handle the newly introduced types. Such a fundamental change is currently out of scope, but we are considering the matter discussed here for future work on Telegraf.
Closing this for the time being.

Issue #4382, opened 2018-07-05
SQL Server input plugin: Add maximum log file size


matthenning:
### Proposal:

To monitor SQL Server log file usage you require the file size, the maximum file size and the file usage.
The query could look like this:
```
WITH fs
AS
(
    SELECT database_id, type, size * 8 / 1024 size, max_size * 8 / 1024 max_size
    FROM sys.master_files
    WHERE database_id NOT IN (3,1,4)
)
SELECT
    name,
    (SELECT SUM(size) FROM fs WHERE type = 0 AND fs.database_id = db.database_id) DataFileSizeMB,
    (SELECT SUM(max_size) FROM fs WHERE type = 1 AND fs.database_id = db.database_id) LogFileSizeMB
FROM sys.databases db
WHERE database_id NOT IN (3,1,4) 
```

### Current behavior:

Currently the SQL Server Telegraf plugin only collects the file size and the file usage.

### Desired behavior:

The maximum file size is also collected

### Use case:

I want to monitor SQL Server log file usage with Telegraf, InfluxDB and Kapacitor.
If the current file usage is 100% of the current file size there is no reason to trigger an alert as long as the maximum file size is not yet reached. The maximum size isn't collected by the SQL Server plugin though, so I have the source this metric from somewhere else.
It would be great to have this included in the Telegraf plugin.

glinton:
What alert triggers have you configured in kapacitor? It seems like you just need to set a size[s] that you'd like to be alerted for (by kapacitor) when telegraf reports that the file has reached that size. I'm thinking this isn't a thing for telegraf.

danielnelson:
I think this would just make the Kapacitor alerting based on the values configured in the Database.  @matthenning Are you using `query_version = 2`?

matthenning:
@danielnelson Yes, `query_version = 2` ist set.

@glinton I agree that it's not a performance metric per se but defining a static value in the Kapacitor task would require the DBA to open a ticket with the monitoring team every time the max file size is adjusted. And because the value can be adjusted frequently or even by some kind of automated process I think collecting it would be the way to go.

danielnelson:
@m82labs Is is another issue I would appreciate it if you could review, does it seem like a good idea and does it fit well with the existing philosophy of the new format?  Things that you don't feel like they are good fits could go into a input similar to `postgresql_extensible`, we already have several submissions for something like this which are currently just waiting on us.

m82labs:
[...]

PR #6711, opened 2019-11-23
Add SQL server log file size metrics

nwneisen:
### Required for all PRs:

Relates to #4382

### DIFF
--- a/plugins/inputs/sqlserver/README.md
+++ b/plugins/inputs/sqlserver/README.md
@@ -70,6 +70,7 @@ GO
   ## - AzureDBResourceGovernance
   ## - SqlRequests
   ## - ServerProperties
+  ## - LogFileSizes
   exclude_query = [ 'Schedulers' , 'SqlRequests']
 ```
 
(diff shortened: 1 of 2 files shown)


danielnelson:
@m82labs @denzilribeiro Can you take a look?

denzilribeiro:
Went through the pull request effectively we are adding a query to get 2 things but not sure we are achieving much.
a. sum space for all files on disk ( not used) - This should already be there in performance countner format.
b. max space for log

This PR doesn't achieve what you want theoretically - log can be 99% empty so just looking at size on disk isn't helpful - Even if size on disk = max size but 99% is free why would it matter?

Besides you have Data File and Log file size form the performance counters already per database  in the sqlserver_performance measurement.I can add size on disk to existing collector already, aka where we get the IO numbers from virtual filestats if the perfmon number isn't enough. Ideally would be great to get used space from all files but that is much tricker given it requires DB context to do something FILEPROPERTY(file_name,'UsedSpace') or have to write more complex query to loop through databases - not sure worth effort or if worth it has to be at a different frequency totally not every 10 seconds by default.

Issue #2616, opened 2017-04-03
[Feature request] ZFS Zpool properties on Linux


tomtastic:
## Feature Request
Currently, only FreeBSD has a zfs input plugin which reports zpool list properties. Linux should have this too ideally.

### Proposal:
Using the existing inputs.zfs 'poolMetrics = true||false' option, to decide whether or not to try executing : '/bin/zpool list -Hp' and parsing the output like the FreeBSD code does.

### Current behavior:
Only kstats are queried on Linux platforms, whereas FreeBSD queries both kstats and the userspace
'zpool list -Hp' command.

### Desired behavior:
telegraf allowed to run 'zpool list' and collection of stats enabled :
```
Apr 03 20:55:40 nibbler telegraf : TTY=unknown ; PWD=/ ; USER=root ; COMMAND=/sbin/zpool list -Hp
> zfs_pool,host=nibbler,pool=rpool allocated=3695163326464i,capacity=92i,dedupratio=1,fragmentation=51i,free=290566324224i,health="ONLINE",size=3985729650688i 1527000570000000000
```

### Use case:
Monitoring zpool health and fragmentation is a really useful metric for ZFS users,
as it can alert administrators to problems requiring urgent attention.

tomtastic:
I think this only requires a minor re-write of the zfs_linux.go plugin code, and an explaination in the telegraf.conf inputs.zfs stanza to describe the prior sudoers configuration work required.

I have some working code currently which I will submit as a PR soon, does **influxdata** prefer single (merged) commits for PRs ?

danielnelson:
For commits you can leave them separate, when we merge PRs we generally squash them.

tomtastic:
Thanks @danielnelson.
I'm having a problem with another PR, circle-test.sh timing out whilst pulling docker.io/stealthly/docker-riemann (https://github.com/influxdata/telegraf/pull/2623)
Something I've done?

danielnelson:
I started a rebuild for it, we need to improve our test script to not pull the docker image all the time.  If something like this happens again just leave a comment on the PR.

rwarren:
Any update on your work, @tomtastic?

I've found this issue while trying to figure out if any of the detailed parameters that are collected by the existing `zfs` plugin actually yield (even if indirectly) any of the significantly more useful parameters like:

* `health` -- ie: do I need to replace a hard drive?
  * not sure how you would report this out via telegraf, anyway?
* `size` -- current size of pool
[...]

PR #6724, opened 2019-11-28
Issue #2616: ZFS Zpool properties on Linux

yvasiyarov:
Updated Linux ZFS agent to collect information exposed by zpool list 
It was requested in https://github.com/influxdata/telegraf/issues/2616 but for some reason was not implemented.

Since that information was already collected by ZFS agent on FreeBSD is just reused that implementation and updated unit tests accordingly. Shared part of Linux and FreeBSD agents has been moved to zfs.go file.
It was tested on Debian + FreeBSD 11

### Required for all PRs:


yvasiyarov:
@danielnelson Is there anything I can do to get this PR merged ?

antifuchs:
While this is waiting to be merged (I really hope it gets merged soon!), I hacked around the absence of useful zpool health checks with this mildly ridiculous python script. Feel free to use it in the meantime:

```py
#!/usr/bin/python

import subprocess
import re
import sys

class ZpoolCheck:
    def check_pool_health(self, pool):
        line=subprocess.check_output(['/sbin/zpool', 'status', '-x', pool]).splitlines()[0]
        if re.match("pool '%s' is healthy" % pool, line):
            # healthy, yay: status 0.
            print("zpool_health,pool=%s health_status=0" % pool)
        else:
            # unhealthy, boo: status 1.
            print("zpool_health,pool=%s health_status=1" % pool)

if __name__ == '__main__':
    c = ZpoolCheck()
    for pool in sys.argv[1:]:
        c.check_pool_health(pool)
```

Install it in a place where it can be found, as `check_zpool_health.py` and use the following config:
```toml
[[inputs.exec]]
  commands = ["/usr/local/lib/telegraf/exec/check_zpool_health.py data"]
  data_format = "influx"
```

Skaronator:
@danielnelson would it be possible to get this merged for the next version? Seems to be ready for review since december. Thanks!

danielnelson:
As the current plugin works using only procfs, I'd prefer not to add a dependency on the zpool command on the Linux side.  The performance isn't very good and it introduces a lot of potential issues including differences in the output, complex permission errors, timeouts.

Is it possible to get the critical information we need from `/proc`?  I read that pool health was recently added https://github.com/openzfs/zfs/pull/7563, though I'm not sure which version it is available in.

yvasiyarov:
Hello @danielnelson, 
unfortunately that information is not available at /proc and openzfs/zfs#7563 PR will add only one metric (zfs pool status itself, without any metrics).
 
There are only two options to get zpool properties:
 - use ioctl syscall directly
 - parse output of zpool command

None of this options is perfect,  but I've chosen the second one. Because:
- Its a way easier to support plugin which parse text output than one operating with binary bindings.
- Its much easier to install zfs utilities then carry with you and compile binary bindings to syscall

To address your concerns about adding dependencies on  zpool command in Linux:
- First of all - this dependency is already here. Current implementation of zfs plugin for FreeBSD already depends on zpool command
- Second: even if zpool command is not exists, or its output format has been changed - nothing will be broken. Plugin will just do not report zpool metrics. 
- Thirdly: There is no performance impact on the telegraf service itself. Zpool runs in separate process. Even if its hangs out - it will not affect telegraf itself. 

So in worst case - we will report only metrics available at /proc, but normally all plugin users can benefit from new metrics available out of the box. Without using ad-hoc solutions, like one described at https://github.com/influxdata/telegraf/pull/6724#issuecomment-612528047

tomtastic:
I agree with @yvasiyarov, zpool command dependancy has been there for FreeBSD since the start. We could / should have had this available in Linux many years ago had the previous attempt not been derailed.
It's about time this got resolved once and for all!

danielnelson:
@richardelling I'd like to get your input on this issue and how we should move forward.  Should we proceed with shelling out to zpool on Linux, would it make more sense to use your [zpool_influxdb](https://github.com/richardelling/zpool_influxdb) utility, or is there another recommended solution.  Perhaps OpenZFS consider adding additional information to the procfs interface?

richardelling:
@danielnelson I believe the containment of a blocked ioctl() is best done outside of telegraf itself. It is quite common for us to prototype new collectors and special-purpose using the `socket_listener` interface, so I'll propose the following:

1. I'll contrib zpool_influxdb to the OpenZFS community, so it can be built alongside ZFS, which is the only viable way to keep up with the unstable (as in interface stability) internal ZFS data structures
2. as part of that, I'll modify to optionally run as a daemon and output to a socket, allowing telegraf folks to simply use the existing socket_listener input plugin
3. existing stdout remains, so if someone really wants to run it via the exec input plugin, it should work, though perhaps less robustly
4. there is also histogram data available inside OpenZFS. I'd like to know the best way to send that to influxdb for heatmaps. In prometheus the buckets are labels (tags) because they have only one field. So is it better to have more fields or higher cardinality?

danielnelson:
Socket_listener works for me, this improves on the zpool method as it doesn't need launched each interval and has machine readable output.

One more option that we now have is the execd input, which would allow Telegraf to run zpool_influxdb as a long-lived child process and read the standard out.  This would act in a similar way to the socket_listener method, in that it would be zpool_influxdb's job to push updates.

Our current recommendation for doing histograms is essentially the Prometheus model, using tags for the bucket bounds and the same field key for all buckets.  Cardinality actually ends up the same since each unique field key is a new series.
```
cpu,cpu=cpu1,le=0.0 usage_idle_bucket=0i
cpu,cpu=cpu1,le=10.0 usage_idle_bucket=1i
cpu,cpu=cpu1,le=50.0 usage_idle_bucket=2i
cpu,cpu=cpu1,le=100.0 usage_idle_bucket=4i
cpu,cpu=cpu1,le=+Inf usage_idle_bucket=4i
```

richardelling:
got it. 
We definitely want the ZFS ioctl consumer in a separate process than telegraf and it is better kept in the OpenZFS source tree.

richardelling:
I have added histograms and setup `zpool_influxdb` to work well with the (awesome) `execd` input plugin, as an option. I'll push this into my master branch after a little bit of soak time and then work to get it added to the OpenZFS contrib directory. Meanwhile, more eyeballs are welcome.
https://github.com/richardelling/zpool_influxdb/tree/histograms

danielnelson:
@richardelling This is really great, thanks so much for doing this.

While it is super easy to setup with `execd` (below), it still might make sense to integrate this directly into the zfs input to give a more polished interface.  If someone is willing to do this, I'd be happy to merge it.

```toml
[[inputs.execd]]
  command = ["zpool_influxdb", "--execd"]
  signal = "STDIN"
  restart_delay = "10s"
  data_format = "influx"
```

danielnelson:
> Would it also make sense to install into /etc/telegraf/telegraf.d if it exists?

This is an interesting idea, I like it because you automatically start receiving the stats if you are a Telegraf user.  I can however think about a few cases where this could cause problems:
- If you are running Telegraf < 1.14 the execd plugin won't exist and you would get an error starting Telegraf.
- Since it's not a pattern we generally use it may conflict with manual setup, unlikely at first but more likely if we integrate this with the zfs plugin.

Maybe it would be safer to include the configuration as part of the package documentation?

Mic92:
This is a pure awk solution to get data errors in the influx line protocol:

```awk
#!/usr/bin/env awk
BEGIN {
    while ("zpool status" | getline) {
        if ($1 ~ /pool:/) { printf "zpool_status,name=%s ", $2 }
        if ($1 ~ /state:/) { printf " state=%s", $2 }
        if ($1 ~ /errors:/) {
            if (index($2, "No")) printf " errors=0\n"; else printf " errors=%d\n", $2
        }
    }
}
```

```
sudo awk -f ./test.sh
zpool_status,name=zroot  state=ONLINE errors=0
zpool_status,name=ztest  state=ONLINE errors=1
```

PR #6726, opened 2019-11-29
Add twitter input plugin

ykzts:
### Required for all PRs:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 - [snmp_trap](/plugins/inputs/snmp_trap/README.md) - Contributed by @influxdata
 - [suricata](/plugins/inputs/suricata/README.md) - Contributed by @satta
 - [synproxy](/plugins/inputs/synproxy/README.md) - Contributed by @rfrenayworldstream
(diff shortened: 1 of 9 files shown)


danielnelson:
@ykzts Thanks for opening the pull request.  Plugin looks good, but I don't think we would include it in Telegraf due to the amount of dependencies it requires, weighed against the number of users who I anticipate would be interested in running it.

We are looking into more ways we can allow plugins like this to be loaded at runtime, and I think that would be a better fit.

ykzts:
> We are looking into more ways we can allow plugins like this to be loaded at runtime, and I think that would be a better fit.

👍

ykzts:
move to https://github.com/inabagumi/twitter-telegraf-plugin

danielnelson:
Very nice, would it be okay if we add your repo to the new list of [external plugins](https://github.com/influxdata/telegraf/blob/master/EXTERNAL_PLUGINS.md)?

ykzts:
This product is still WIP. I'm honored if it's still good.

danielnelson:
That's okay, you are added

Issue #6730, opened 2019-11-29
Report page_faults if present under wired tiger storage engine in mongodb input plugin


rlodge:
## Feature Request

The wired tiger storage engine still reports page_faults, and this is still a useful piece of data to receive when monitoring mongo system, so it should be reported if present.

### Proposal:

Report page_faults under the wired tiger system.

### Current behavior:

Page faults are not reported.

### Desired behavior:

Page faults should be reported.

### Use case:

Monitoring times when performance is slow because of inadequate memory size under wired tiger.

PR #6732, opened 2019-11-29
#6730 - report page_faults for wired tiger system

rlodge:
Implementation for #6730.

### Required for all PRs:

### DIFF
--- a/plugins/inputs/mongodb/mongodb_data.go
+++ b/plugins/inputs/mongodb/mongodb_data.go
@@ -246,6 +246,9 @@ func (d *MongodbData) AddDefaultStats() {
 			d.add(key, floatVal)
 		}
 		d.addStat(statLine, WiredTigerExtStats)
+		if d.StatLine.FaultsCnt > 0 {
+			d.add("page_faults", d.StatLine.FaultsCnt)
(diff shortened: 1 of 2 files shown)


plugins/inputs/mongodb/mongodb_data.go
```diff
@@ -246,6 +246,9 @@ func (d *MongodbData) AddDefaultStats() {
 			d.add(key, floatVal)
 		}
 		d.addStat(statLine, WiredTigerExtStats)
+		if d.StatLine.FaultsCnt > 0 {
```
danielnelson: Would it be wrong to report 0 here?

rlodge: I don't think there's any _harm_ in sending 0.  I was trying with this stat to follow the existing patterns where the values were sent only if they were non-null, or non-empty; I'm not sure how valuable it is in this case.  I'm not even sure, in this instance, if the serverStatus output always includes the value or not.  If you prefer, I can just send it every time.

danielnelson: Yeah, let's change it to always send.

Issue #6725, opened 2019-11-28
Support opLatencies in mongodb input plugin


rlodge:
## Feature Request

The mongodb input plugin should read the opLatencies field of the serverStatus output and report those metrics.

### Proposal:

Add DefaultLatencyStats to mongodb_data.go, and add LatencyStats and OpLatenciesStats structures go mongostat.go, and populate them if the data read from serverStatus contains the appropriate fields.

### Current behavior:

No latency stats are reported.

### Desired behavior:

Latency stats should be reported so that the plugin can more fully support monitoring mongo.

### Use case:

We are replacing a proprietary monitoring system with one based on InfluxDB, and the one piece of data we use a lot in the existing system that the plugin doesn't support is the opLatencies values

powersj:
Fixed in #6725. Closing.

PR #6733, opened 2019-11-29
#6725 - add latency stats to mongo input

rlodge:
Implementation for #6725.

### Required for all PRs:


plugins/inputs/mongodb/mongostat.go
```diff
@@ -680,6 +711,12 @@ func NewStatLine(oldMongo, newMongo MongoStatus, key string, all bool, sampleSec
 		returnVal.Command, returnVal.CommandCnt = diff(newStat.Opcounters.Command, oldStat.Opcounters.Command, sampleSecs)
 	}
 
+	if newStat.OpLatencies != nil && oldStat.OpLatencies != nil {
+		returnVal.ReadLatency = computeLatencyDiff(newStat.OpLatencies.Reads, oldStat.OpLatencies.Reads)
```
danielnelson: Instead of calculating the diffs, let's report these as raw values.  On the query side a user can use `non_negative_difference` or `non_negative_derivative` to get the deltas or rate.

rlodge: I decided to send the diffs in order to mimic what 10gen's cloud tooling does, displaying the average latency per op over the selected period, which I think is the only real interesting value.  I find it a little bit of a pain to have to use `non_negative_derivative(latency, period) / non_negative_derivative(ops, period)` every time I want to display the periodic average.  I also run into problems if there's ever a gap in the time-series; e.g. if there's a short network outage I get giant spikes in the graph because of the gap, and so I find it more reliable to have the value I want (the average), rather than have to rely on the not-very-reliable derivative functions.

danielnelson: I know it is sometimes convienent to have precomputed diffs but it has drawbacks too.  It requires the plugin to run twice in order to emit data, plus the additional calculation logic.  When adopted as a style this is a large burden on writing and maintaining the plugins.  Precomputed aggregated values are often less flexible and accurate to work with so we prefer to leave the aggregation and calculation to the query side.  This plugin has a fair bit of rate calculation in it, due to it being one of the original plugins, and eventually we are going to need to strip all of it out.

As an alternative you can try using the `basicstats` aggregator to calculate a difference.  In the future we are planning some other methods for doing calculations before insertion into the database.

danielnelson:
Looks good, can you take care of the readme conflict though?

rlodge:
Sure, conflicts resolved

PR #6746, opened 2019-12-03
Set message timestamp to the metric time in kafka output

### LINKED ISSUE #6696: Kafka output with Timestamp
## Feature Request

Telegraf should add a timestamp to the Kafka Message. I suggest to add in the next order: "Metric's timestamp" if empty than "Current timestamp". This behavior is the same as we configure Kafka to automatically set Message Timestamp when it stored by a Kafka Worker.

Initially, that feature was implemented here: https://github.com/influxdata/telegraf/commit/74b330922527510c209645474b7ac664c1b2b9d9 on Sep 23, 2015. But then after refactoring was removed https://github.com/influxdata/telegraf/commit/c26ce9c4fe4a1ece56473e2122e6e4157319f187 on Oct 17, 2015. I guess by a mistake

### Current behavior:

Telegraf skip adding a timestamp.

### Desired behavior:

Telegraf should add a timestamp.

### Use case:

It is a good thing to know what is the timestamp of the tail or the head message on the topic.

danielnelson:
closes #6696

### Required for all PRs:

### DIFF
--- a/plugins/outputs/kafka/kafka.go
+++ b/plugins/outputs/kafka/kafka.go
@@ -50,6 +50,8 @@ type (
 		// SASL Password
 		SASLPassword string `toml:"sasl_password"`
 
+		Log telegraf.Logger `toml:"-"`
+
 		tlsConfig tls.Config
 		producer  sarama.SyncProducer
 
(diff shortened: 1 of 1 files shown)


plugins/outputs/kafka/kafka.go
```diff
@@ -316,13 +318,14 @@ func (k *Kafka) Write(metrics []telegraf.Metric) error {
 	for _, metric := range metrics {
 		buf, err := k.serializer.Serialize(metric)
 		if err != nil {
-			log.Printf("D! [outputs.kafka] Could not serialize metric: %v", err)
+			k.Log.Debugf("Could not serialize metric: %v", err)
```
johnatannvmd: I think it's more like error than debug information. Because it's about data loss, and not about "how and where?". It's my opinion.

danielnelson: No change is made here to the level, so I think we can handle this in a separate issue.  I think it may have been added at some point due to metrics without fields, need to do some research on the history.

PR #6759, opened 2019-12-05
Breaker processor, drop metrics based on a flag

adrianlzt:
New processor to be able to drop metrics if a dynamic flag is not meet.
This flag is setted by another metric from telegraf.

### Required for all PRs:


plugins/processors/breaker/README.md
```diff
@@ -0,0 +1,74 @@
+# Breaker processor Plugin
+
+This processor could drop metrics based on the value of another selected (flag) metric.
+
+The parameters ``name`` y ``field`` are used to select with metric will be used to define the status of the breaker.
```
reimda: y should be and

plugins/processors/breaker/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+### Configuration:
+
+```toml
+[[processors.breaker]]
+  # By default is false, meaning it will block
+  enabled = false
+
+  # Select which metric should be used to define the state of the breaker
+  name = "flag_metric"
```
reimda: Could we use namepass and namedrop instead of adding another way to select the metric?  See docs/CONFIGURATION.md

plugins/processors/breaker/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+The metric used se the "flag" will pass also.
+
+Normally this processor will be used with some filter to select which metrics should be affected.
+
+### Configuration:
+
+```toml
+[[processors.breaker]]
+  # By default is false, meaning it will block
```
reimda: This comment says enabled=false means block.  The comment on line 64 say the opposite.

plugins/processors/breaker/README.md
```diff
@@ -0,0 +1,74 @@
...
+[[inputs.exec]]
+  commands = [ "hostname" ]
+  timeout = "5s"
+  data_format = "value"
+  data_type = "string"
+  name_override = "active_node"
+  [inputs.exec.tags]
+    host = "virtualHost"
+
+# B
```
reimda: stray comment

plugins/processors/breaker/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+  # Which values of the selected metric will be used to enable or disable the breaker
+  # These values will be transformed to string to be compared with metrics values (also converted to strings)
+  value_enable = "foo"
+  value_disable = "bar"
+```
+
+Example configuration dropping metrics if host does not have configured the IP resolved by a domain.
+
+The idea is two servers in a active-pasive cluster configuration, where only the active one should send metrics.
```
reimda: pasive should be passive here and in a few places later on

plugins/processors/breaker/test.conf
```diff
@@ -0,0 +1,33 @@
+[agent]
```
reimda: It doesn't look like this file is used by the code or tests.  Please delete this and input.file

sjwang90:
@adrianlzt Would this breaker processor be able to be done with the [Starlark processor](https://github.com/influxdata/telegraf/blob/master/plugins/processors/starlark/README.md)? There is logic capabilities that can be done with Starlark, including some [drop examples](https://github.com/influxdata/telegraf/blob/master/plugins/processors/starlark/README.md#examples) we have listed.

Jagularr:
This functionality can be handled using our `starlark` processor. See the comment above for examples on how the logic can be written, essentially basic checks can be done on fields and field contents, and if criteria is met to skip, one can do `return None` in the script's `apply` function, which gets called for each metric being passed through. If I've missed a use case that starlark might not be able to handle, feel free to re-open and let me know. Thank you!

PR #6770, opened 2019-12-08
unbound: expose unbound-control config file option

elohmeier:
This PR exposes the [-c cfgfile option of unbound-control](https://linux.die.net/man/8/unbound-control) allowing Telegraf users to monitor non-default unbound setups.
Unbounds default location `/etc/unbound/unbound.conf` is set as Telegraf default.

### Required for all PRs:

### DIFF
--- a/plugins/inputs/unbound/unbound.go
+++ b/plugins/inputs/unbound/unbound.go
@@ -17,7 +17,7 @@ import (
 	"github.com/influxdata/telegraf/plugins/inputs"
 )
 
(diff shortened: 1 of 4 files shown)


plugins/inputs/unbound/unbound.go
```diff
@@ -207,6 +216,7 @@ func init() {
 			UseSudo:     false,
 			Server:      "",
 			ThreadAsTag: false,
+			ConfigFile:  defaultConfigFile,
```
danielnelson: Let's default this to "", so that `unbound` picks the default file.

elohmeier: Thanks for your feedback. I've changed that.

PR #6773, opened 2019-12-09
csv: TrimLeadingSpace when TrimSpace is on

DSpeichert:
When TrimSpace is used and Delimeter is a space, it trips up the CSV reader otherwise.

Visual representation of now-supported format:
```
   first   second   third   fourth
abcdefgh        0       2    false
  abcdef      3.3       4     true
       f        0       2    false
```

### Required for all PRs:


danielnelson:
Thanks, this seems right, can you add a unit test with an example too?

DSpeichert:
@danielnelson done!

Issue #6789, opened 2019-12-11
MySQL Input Plugin should have config option to disable collecting global vars


benhymans:
## Feature Request

Opening a feature request kicks off a discussion.

### Proposal: 
There should be a config value that allows for enabling/disabling collection of global variables.

### Current behavior: 
Global Variables are always collected if the input plugin is active

### Desired behavior: 
Setting a config value to false would prevent global variables from being collected

### Use case: 

Global Variables could contain sensitive information, and depending on the permissions policy in influx, and the downstream systems (grafana, etc), this could pose a security risk. The data is useful though, so I propose allowing the engineer configuring the input plugin the option to enable/disable collecting this data.

I intend to submit a PR to enable  this functionality.

Esity:
@danielnelson @glinton This can be closed with the successful merge of #6790

PR #6790, opened 2019-12-11
MySQL - Option for collecting global variables

benhymans:
This PR addresses Issue #6789 

There should be a config value that allows for enabling/disabling collection of MySQL Global Variables.

- Readme: Added line for boolean config for collecting global vars
- mysql.go: added line to retrieve  config value
- mysql.go: added logic to either collect or not collect vars based on config

### Required for all PRs:


Esity:
@danielnelson & @glinton 
Looking for review on simple feature flag

plugins/inputs/mysql/mysql.go
```diff
@@ -94,6 +95,9 @@ const sampleConfig = `
   ## gather metrics from SHOW BINARY LOGS command output
   # gather_binary_logs = false
 
+  ## gather metrics from PERFORMANCE_SCHEMA.GLOBAL_VARIABLES
+  # gather_global_variables = false
```
danielnelson: This was previously always enabled, so we should display this as true here and initialize it to true in the plugins `init()` function.

PR #6792, opened 2019-12-11
Add dedup processor

igomura:
### Required for all PRs:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 /telegraf.exe
 /telegraf.gz
 /vendor
+.idea
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,3 @@
 /telegraf.exe
 /telegraf.gz
 /vendor
-.idea
--- a/plugins/processors/dedup/README.md
+++ b/plugins/processors/dedup/README.md
@@ -13,8 +13,5 @@ time while maintain proper precision when value gets changed rapidly
 [[processors.dedup]]
   ## Maximum time to suppress output
   dedup_interval = "600s"
-
-  ## Maximum time to keep cached metric without update
-  evict_interval = "1h"
 ```
 
--- a/plugins/processors/dedup/README.md
+++ b/plugins/processors/dedup/README.md
@@ -1,0 +1,20 @@
+# Dedup Processor Plugin
+
+If a metric sends the same value over successive intervals, suppress sending
+the same value to the TSD until this many seconds have elapsed.  This helps
+graphs over narrow time ranges still see timeseries with suppressed datapoints.
+
+This feature can be used to reduce traffic when metric's value does not change over
+time while maintain proper precision when value gets changed rapidly
+
+### Configuration
+
+```toml
+[[processors.dedup]]
+  ## Maximum time to suppress output
+  dedup_interval = "600s"
+
+  ## Maximum time to keep cached metric without update
+  evict_interval = "1h"
+```
+
--- a/plugins/processors/dedup/dedup.go
+++ b/plugins/processors/dedup/dedup.go
@@ -82,15 +82,15 @@ func (d *Dedup) Apply(metrics ...telegraf.Metric) []telegraf.Metric {
 		}
 
 		// For each filed compare value with the cached one
-		same := true
+		changed := false
 		for _, f := range metric.FieldList() {
 			if value, ok := m.GetField(f.Key); ok && value != f.Value {
-				same = false
+				changed = true
 				continue
 			}
 		}
 		// If any field value has changed then refresh the cache
-		if !same {
+		if changed {
 			d.save(metric, id)
 			continue
 		}
--- a/plugins/processors/dedup/dedup.go
+++ b/plugins/processors/dedup/dedup.go
@@ -1,7 +1,6 @@
 package dedup
 
 import (
-	"log"
 	"time"
 
 	"github.com/influxdata/telegraf"
@@ -12,20 +11,18 @@ import (
 var sampleConfig = `
   ## Maximum time to suppress output
   dedup_interval = "600s"
-  ## Maximum time to keep cached metric without update
-  evict_interval = "1h"
 `
 
 type Dedup struct {
 	DedupInterval internal.Duration `toml:"dedup_interval"`
-	EvictInterval internal.Duration `toml:"evict_interval"`
+	FlushTime     time.Time
 	Cache         map[uint64]telegraf.Metric
 }
 
 func NewDedup() *Dedup {
 	return &Dedup{
 		DedupInterval: internal.Duration{Duration: 10 * time.Minute},
-		EvictInterval: internal.Duration{Duration: 1 * time.Hour},
+		FlushTime:     time.Now(),
 		Cache:         make(map[uint64]telegraf.Metric),
 	}
 }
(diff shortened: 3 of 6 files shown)


igomura:
Hi, my I have review for my PR? Is that something you would consider for 1.13.1?

danielnelson:
> Is that something you would consider for 1.13.1?

We are only doing bug fixes for 1.13.x releases, but we will try for 1.14.

plugins/processors/dedup/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+### Configuration
+
+```toml
+[[processors.dedup]]
+  ## Maximum time to suppress output
+  dedup_interval = "600s"
+
+  ## Maximum time to keep cached metric without update
+  evict_interval = "1h"
```
danielnelson: I don't understand why we need two durations, could these be combined?

plugins/processors/dedup/dedup.go
```diff
@@ -0,0 +1,89 @@
...
+	d.Cache = keep
+}
+
+func (d *Dedup) Apply(metrics ...telegraf.Metric) []telegraf.Metric {
+	for idx, metric := range metrics {
+		id := metric.HashID()
+		if m, ok := d.Cache[id]; ok {
+			// compare all fields values
+			sameValue := true
+			for k, v := range metric.Fields() {
```
danielnelson: Use `metric.FieldList()` here since it is much faster.

plugins/processors/dedup/dedup.go
```diff
@@ -0,0 +1,89 @@
...
+				if m.Fields()[k] != v {
+					log.Printf("D! [processors.dedup]: metric value has changed: %s", metric.Name())
+					sameValue = false
+				}
+			}
+			// value has not changed and cache is not expired
+			if sameValue && time.Since(m.Time()) < d.DedupInterval.Duration {
+				// Deduplicate this metric
+				log.Printf("D! [processors.dedup]: suppress metric: %s", metric.Name())
```
danielnelson: Can you remove all the logging messages, they are too detailed for our current logging system.

plugins/processors/dedup/dedup.go
```diff
@@ -0,0 +1,89 @@
...
+				metrics = remove(metrics, idx)
+				continue
+			}
+		}
+		log.Printf("D! [processors.dedup]: update cached metric: %s", metric.Name())
+		d.Cache[id] = metric.Copy()
```
danielnelson: Call metric.Accept() on the copy, this is important in case the cached metric is a "tracking metrics"

plugins/processors/dedup/dedup.go
```diff
@@ -0,0 +1,89 @@
...
+
+// Remove single item from slice
+func remove(slice []telegraf.Metric, i int) []telegraf.Metric {
+	slice[len(slice)-1], slice[i] = slice[i], slice[len(slice)-1]
+	return slice[:len(slice)-1]
+}
+
+// Remove expired items from cache
+func (d *Dedup) cleanup() {
+	keep := make(map[uint64]telegraf.Metric, 0)
```
danielnelson: Since the processors get called for every metric, I think we should limit how often this is called and reuse the cache map.

scripts/stretch.docker
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.11.0 as builder
+FROM golang:1.12.0 as builder
```
danielnelson: Revert unrelated change

.gitignore
```diff
@@ -3,3 +3,4 @@
 /telegraf.exe
 /telegraf.gz
 /vendor
+.idea
```
danielnelson: Please revert this change, you can add it to your global Git `core.excludesfile`

plugins/processors/dedup/README.md
igomura: You are right. I will combine them

igomura:
Thanks for review. This is my "Hello World" in go. I appreciate any suggestion to make this module better.

plugins/processors/dedup/dedup.go
```diff
@@ -0,0 +1,98 @@
...
+func (d *Dedup) Apply(metrics ...telegraf.Metric) []telegraf.Metric {
+	for idx, metric := range metrics {
+		id := metric.HashID()
+		// check if metric is already in cache. Otherwise save it in cache
+		if m, ok := d.Cache[id]; ok {
+			// if cache is not expired then check values. Otherwise save it in cache
+			if time.Since(m.Time()) < d.DedupInterval.Duration {
+				for _, field := range metric.FieldList() {
+					// if same value then drop it. Otherwise save it in cache
+					if m.Fields()[field.Key] == field.Value {
```
danielnelson: Use `m.GetField(field.Key)` here instead of `m.Fields()`.  The Fields() method is very slow, hopefully we can remove it soon to avoid confusion.

plugins/processors/dedup/dedup.go
```diff
@@ -0,0 +1,98 @@
...
+		if time.Since(metric.Time()) < d.DedupInterval.Duration {
+			keep[id] = metric
+		}
+	}
+	d.Cache = keep
+}
+
+// Save item to cache
+func (d *Dedup) save(metric telegraf.Metric) {
+	id := metric.HashID()
```
danielnelson: Let's pass in the `id` we obtained from line 69, to avoid recalculating it.

plugins/processors/dedup/dedup.go
```diff
@@ -0,0 +1,98 @@
...
+			// if cache is not expired then check values. Otherwise save it in cache
+			if time.Since(m.Time()) < d.DedupInterval.Duration {
+				for _, field := range metric.FieldList() {
+					// if same value then drop it. Otherwise save it in cache
+					if m.Fields()[field.Key] == field.Value {
+						metrics = remove(metrics, idx)
+					} else {
+						d.save(metric)
+					}
+					continue
```
danielnelson: Since this is the last line of a for loop `continue` won't have any effect, we should remove it.

plugins/processors/dedup/dedup.go
```diff
@@ -0,0 +1,98 @@
...
+		// check if metric is already in cache. Otherwise save it in cache
+		if m, ok := d.Cache[id]; ok {
+			// if cache is not expired then check values. Otherwise save it in cache
+			if time.Since(m.Time()) < d.DedupInterval.Duration {
+				for _, field := range metric.FieldList() {
+					// if same value then drop it. Otherwise save it in cache
+					if m.Fields()[field.Key] == field.Value {
+						metrics = remove(metrics, idx)
+					} else {
+						d.save(metric)
```
danielnelson: If we are saving the metric in the cache, I think we can skip looking at the other fields.

plugins/processors/dedup/dedup.go
```diff
@@ -0,0 +1,98 @@
...
+					}
+					continue
+				}
+			} else {
+				d.save(metric)
+			}
+		} else {
+			d.save(metric)
+		}
```
danielnelson: It is a little hard to match these else statements up with the conditions above.  I'd flip the cases around like this:
```
if m, ok := d.Cache[id]; !ok {
    d.save(metric)
    continue
}

// put the "ok" case here
```

plugins/processors/dedup/dedup_test.go
```diff
@@ -0,0 +1,157 @@
...
+	assertCacheRefresh(t, &deduplicate, source)
+	assertMetricPassed(t, target, source)
+}
+
+func TestSuppressRepeatedValue(t *testing.T) {
+	deduplicate := createProcessor()
+	source := createMetric("m1", 1)
+	target := deduplicate.Apply(source)
+	// wait less than deduplication interval
+	time.Sleep(1 * time.Second)
```
danielnelson: We shouldn't actually sleep in the tests, since it will result in a race condition and makes the tests slow.  How we usually handle this is to allow the current time to be set in the test.  Here are the basic steps to set this up:
- define a function type for getting the time: `type TimeFunc func() time.Time`
- add a TimeFunc field on the processor struct
- initialize the new field to time.Now
- in the tests, you can swap the function out to change the time.

plugins/processors/dedup/dedup_test.go
```diff
@@ -0,0 +1,157 @@
...
+		FlushTime:     time.Now(),
+		Cache:         make(map[uint64]telegraf.Metric),
+	}
+}
+
+func assertCacheRefresh(t *testing.T, proc *Dedup, item telegraf.Metric) {
+	id := item.HashID()
+	name := item.Name()
+	// cache is not empty
+	assert.NotEqual(t, 0, len(proc.Cache))
```
danielnelson: Use `require` instead of `assert` everywhere you can, the difference is that require will halt the test and with assert it will keep going:
```
require.NotEqual(t, 0, len(proc.Cache))
```

plugins/processors/dedup/dedup.go
igomura: This is attempt to skip looking other fields metric may have. So, it may have affect.

plugins/processors/dedup/dedup.go
igomura: Yep. this is what line 81 for

igomura:
Let me know if you would like me to squash commits at the end

PR #6793, opened 2019-12-11
Interpret SNMP v1 traps as described in RFC 2576 3.1

### LINKED ISSUE #6786: Add support for SNMP version 1 traps
gosnmp delivers SNMP v1 traps but they aren't collected in telegraf because there are no fields.  We should convert the v1 fields as described in [rfc2576](https://tools.ietf.org/html/rfc2576#page-12).  This will guarantee fields and improve the compatibility between v1 and v2 traps.

See discussion in #4377.

reimda:
### Required for all PRs:

closes #6786

### DIFF
--- a/plugins/inputs/snmp_trap/snmp_trap.go
+++ b/plugins/inputs/snmp_trap/snmp_trap.go
@@ -181,7 +181,10 @@ func makeTrapHandler(s *SnmpTrap) handler {
 				}
 				setTrapOid(tags, trapOid, e)
 			}
-			tags["agent_address"] = packet.AgentAddress
+
+			if packet.AgentAddress != "" {
+				tags["agent_address"] = packet.AgentAddress
+			}
 
 			fields["sysUpTimeInstance"] = packet.Timestamp
 		}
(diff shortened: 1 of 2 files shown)


plugins/inputs/snmp_trap/snmp_trap.go
```diff
@@ -155,6 +162,30 @@ func makeTrapHandler(s *SnmpTrap) handler {
...
+
+			if trapOid != "" {
+				e, err := s.lookup(trapOid)
+				if err != nil {
+					s.Log.Errorf("Error resolving V1 OID: %v", err)
+					return
+				}
+				setTrapOid(tags, trapOid, e)
+			}
+			tags["agent_address"] = packet.AgentAddress
```
danielnelson: Let's set this as the "source" tag.  Possibly in the future this could be the `snmpTrapAddress.0` if we decide we should act as a proxy but I don't feel confident enough now that we will want to do this.

reimda: I'm not sure it's always valid.  After chatting with you we decided to continue using addr.IP as the "source" tag and validate agent address before adding it as a tag.

PR #6805, opened 2019-12-16
Update merge aggregator config file in README.md

sjwang90:
Update merge aggregator config file in README.md with `drop_original = true`. This type of aggregator would want to default to true.

### Required for all PRs:


danielnelson:
Also add this to `merge.go`, this will make it so the output of `telegraf config` will have this option as well.

PR #6810, opened 2019-12-17
Sort alphabetically plugin lists

maurorappa:
With tens of ouputs of plugins, and hundreds of input plugins, I struggled to easily read the needed plugin, so I though about sorting them alphabetically.
It can be useful to every user, instead of using grep to look for the needed plugin.

### Required for all PRs:

- [ ] Signed [CLA](https://influxdata.com/community/cla/) - signed
- [ ] Associated README.md updated - no relavant readme ( --help maybe?) 
- [ ] Has appropriate unit tests. - no test possible

### DIFF
--- a/cmd/telegraf/telegraf.go
+++ b/cmd/telegraf/telegraf.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"os/signal"
 	"runtime"
+	"sort"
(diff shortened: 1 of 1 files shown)


danielnelson:
Thank you

PR #6815, opened 2019-12-20
Improve suricata input unit test debugging

danielnelson:
This pull request is meant to improve the reliability and output of the suricata inputs unit tests to help track down some intermittent failures in the test suite.

### Required for all PRs:

### DIFF
--- a/plugins/inputs/suricata/suricata_test.go
+++ b/plugins/inputs/suricata/suricata_test.go
@@ -1,7 +1,6 @@
 package suricata
 
 import (
-	"bytes"
 	"fmt"
 	"io/ioutil"
 	"log"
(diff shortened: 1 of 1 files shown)


plugins/inputs/suricata/suricata_test.go
```diff
@@ -297,85 +237,23 @@ func TestSuricataEmptyJSON(t *testing.T) {
...
 	}
 	c.Write([]byte("\n"))
 	c.Close()
 
 	acc.WaitError(1)
-
-	s.Stop()
-}
-
-func TestSuricataInvalidInputs(t *testing.T) {
```
reimda: Was this test removed because it's not useful?  If so, we also need to remove the unused brokenType* and brokenStruct* vars at the top

danielnelson: These tests were removed to avoid testing log output.  The test was inherently racy; there is no way to know when the logging has completed as there is no feedback to the sender.  Time permitting, I'd like to come back to this plugin since I believe some of these warnings ought to be errors that could be tested.

danielnelson: I have removed the now unused variables.

PR #6816, opened 2019-12-20
Openweathermap doesnt use /data/2.5/group, /data/2.5/weather is used …

gdunstone:
Here is a link to the docs where the correct api endpoint is specfied:

https://openweathermap.org/current

### Required for all PRs:

- [y] Signed [CLA](https://influxdata.com/community/cla/).
- [na] Associated README.md updated.
- [na] Has appropriate unit tests.


tmapes:
`/group` is used for requesting multiple `city_ids` which `/weather` does not support.

gdunstone:
`/group` didn't work at all before for me, I was getting a 401 for `/group` but not for `/weather`.
I think openweathermap was gaslighting me for some reason.

Thanks for the explanation, I'm able to get data from openweathermap now.

Issue #6584, opened 2019-10-27
Enable debug logging on a per-plugin basis


lhanson:
## Feature Request

I'd like to be able to see debug logging for specific plugins I'm using rather than simply for the top-level agent.

### Proposal:

Allow the `[agent]` level "debug" configuration value to be applied on a per-plugin basis.

### Current behavior:
Enabling `debug = true` at the `[agent]` level only emits high-level debug output.

### Desired behavior:

Allow `debug = true` to be set for individual plugins I've defined, which then log specific actions they are or are not taking as they run.

### Use case: 

I can't figure out why I can't seem to use `[[processors.rename.replace]]` to replace fields based on a specific tag; all subsequent series are apparently affected regardless of `tagpass` values I'm trying to use. It would be helpful to have verbose logging to ascertain how inputs are being processed by my current configuration so I can determine what I'm doing wrong.

glinton:
> Enabling `debug = true` at the `[agent]` level only emits high-level debug output.

Setting `debug = true` in the `agent section actually turns on debugging for everything. The problem is that many plugins don't implement any debug logging.

Since this sounds like a config issue, maybe posting your config and asking over at the [InfluxData Community](https://community.influxdata.com) site will yield some results.

lhanson:
@glinton  That's helpful, thanks. My immediate issue has been mostly resolved through a combination of trial-and-error and code spelunking (you're correct, it was a config issue; the TOML bracket syntax caught me out and I was using `[[processors.rename.tagpass]]` instead of `[processors.rename.tagpass]`).

I think that the lack of debugging output I was hoping to see (but didn't) is, as you say, the result of plugins not printing any debug information. Would PRs adding additional debugging output to plugins be welcome if I find them useful to fix any future issues?

Thinking more broadly, would it be worth adding a note in [CONTRIBUTING.md](https://github.com/influxdata/telegraf/blob/master/CONTRIBUTING.md) explaining how the debug setting applies to plugins?

glinton:
> Would PRs adding additional debugging output to plugins be welcome if I find them useful to fix any future issues?

It would be helpful to add the least amount of debug logs required to debug an issue, but I'm not sure what our "official" stance is, @danielnelson can say better when he's back in.

danielnelson:
[...]

PR #6817, opened 2019-12-20
feat(uwsgi): Allow plugin to continue when stats server(s) are unavailable

### LINKED ISSUE #6795: Telegraf is unable start when using uwsgi in on-demand mode
I was not sure if this should be a bug or feature request, if this should be a feature request my apologies.

The problem is that after starting uWSGI in on-demand mode not all stat ports are actually available. There needs to be a request first to activate the vassal after which you can see something actually listening on port 9191.

Telegraf however will not start if any stat socket is unreachable.

### Relevant telegraf.conf:
```toml
[[inputs.uwsgi]]
  servers = ["tcp://localhost:9191"]
```

### System info:
* Telegraf 1.12.6 (git: HEAD 6c7f2d62)
* Debian GNU/Linux 9.11 (stretch)
* uWSGI 2.0.18

### Steps to reproduce:

In any directory of your choosing;
1. Create the needed directories
`mkdir vassals sockets`

2. Start uWSGI in emperor mode with ondemand
```
uwsgi --master --thunder-lock --emperor vassals/ --emperor-on-demand-directory sockets/ --emperor-on-demand-extension socket
```

3. Create an example vassal
```
cat <<EOF > vassals/example.ini
[uwsgi]
socket = 127.0.0.1:3031
chdir = /tmp/uwsgi
wsgi-file = myproject/wsgi.py
processes = 4
threads = 2
stats = 127.0.0.1:9191

EOF
```

4. Try to start Telegraf

### Expected behavior:

I would hope to see telegraf actually starting, and gathering statistics when the vassal is activated. 

### Actual behavior:

Telegraf is unable to start;
```
2019-12-12T13:59:21Z E! [telegraf] Error running agent: One or more input plugins had an error
```

### Additional info:

The stats socket (tcp or unix) is being created when the first requests comes in when running uwsgi in emperor mode with on-demand. Existing stats sockets could also disappear after being initially activated when `--idle` and `--die-on-idle` are set.

Not much unlike with phpfpm althought the same socket is used for requests and stats alike.

hkraal:
Fixes #6795

- Added: flag `skip_errors` to allow for skipping any errors
- Changed: `s.source` in case of unix sockets to use the socket path instead of the hostname to keep the statistics identifiable

### Required for all PRs:


danielnelson:
@hkraal Thanks for sharing your solution, but I don't think we would incorporate this patch due to the new option.  My thinking currently is that we need to design a system where the logging level can be set on a per plugin basis.  https://github.com/influxdata/telegraf/issues/6584#issuecomment-550072462

hkraal:
@danielnelson I'm willing to modify the PR to suit both our needs, would removing the option and skipping unreachable stats servers have more chance to make it?

My main problem was that the plugin stops when it encounters an inactive stats socket. Depending on the used uWSGI deployment this behaviour should be expected and should not stop the plugin (and Telegraf) dead in its tracks.

I figured the main issue was in the fact an error was logged. Just ignoring (and logging if desirable) unreachable sockets crossed my mind as well but it less explicit than adding `skip_errors: true` to the configuration and so I choose for adding the option.

danielnelson:
> I'm willing to modify the PR to suit both our needs, would removing the option and skipping unreachable stats servers have more chance to make it?

My issue with this is that when the socket cannot be reached it is usually an error, and without the error the plugin will be harder to debug.

Almost any plugin could produce an error that isn't considered an error by the operator.  For example maybe the operator doesn't consider it an error if a site connected to by the `http` server is down.  In order to avoid adding options like this piecemeal to Telegraf, I think we need a more generic solution.

PR #6819, opened 2019-12-23
Report rabbitmq_node measurement and return on gather error

### LINKED ISSUE #6814: RabbitMQ input reports forever that it did not complete in time after a single timeout
### Relevant telegraf.conf:
```toml
[agent]
  interval = "10s"
  logfile = "Z:/tick/telegraf/telegraf.log"
[[outputs.influxdb]]
[[inputs.rabbitmq]]
  username = "guest"
  password = "guest"
```

### System info:
Windows Server 2016 64-bit, Telegraf 1.13.0

### Steps to reproduce:
1. Run telegraf and collect RabbitMQ node stats.
2. Somehow get the `requestJSON` call to fail when making a request to `/api/healthchecks/node/<nodename>`
3. Watch the log file

I achieved step 2 by modifying the RabbitMQ source file such that on the 3rd gather only it changes the username, and then made `requestJSON` return an error when it saw that username (to mimic a network failure).

### Expected behavior:
A single gather times out but future gathers succeed without error (as soon as the broker has recovered from whatever caused the original failure).

### Actual behavior:
After a single failure collecting node stats, the gather call to the RabbitMQ plugin times out every time.

### Additional info:
Here is an example of what my log file looks like:
```
2019-12-18T05:48:38Z I! Loaded inputs: win_perf_counters rabbitmq internal
2019-12-18T05:48:38Z I! Loaded aggregators: 
2019-12-18T05:48:38Z I! Loaded processors: 
2019-12-18T05:48:38Z I! Loaded outputs: influxdb
2019-12-18T05:48:38Z I! Tags enabled: dc=London host=sms-main-1
2019-12-18T05:48:38Z I! [agent] Config: Interval:10s, Quiet:false, Hostname:"sms-main-1", Flush Interval:10s
2019-12-19T00:03:03Z E! [inputs.rabbitmq] Error in plugin: Get http://localhost:15672/api/healthchecks/node/rabbit@sms-main-1: net/http: timeout awaiting response headers
2019-12-19T00:03:10Z W! [agent] [inputs.rabbitmq] did not complete within its interval
2019-12-19T00:03:20Z W! [agent] [inputs.rabbitmq] did not complete within its interval
2019-12-19T00:03:30Z W! [agent] [inputs.rabbitmq] did not complete within its interval
2019-12-19T00:03:40Z W! [agent] [inputs.rabbitmq] did not complete within its interval
2019-12-19T00:03:50Z W! [agent] [inputs.rabbitmq] did not complete within its interval
2019-12-19T00:04:00Z W! [agent] [inputs.rabbitmq] did not complete within its interval
2019-12-19T00:04:10Z W! [agent] [inputs.rabbitmq] did not complete within its interval
2019-12-19T00:04:20Z W! [agent] [inputs.rabbitmq] did not complete within its interval
2019-12-19T00:04:30Z W! [agent] [inputs.rabbitmq] did not complete within its interval
2019-12-19T00:04:40Z W! [agent] [inputs.rabbitmq] did not complete within its interval
2019-12-19T00:04:50Z W! [agent] [inputs.rabbitmq] did not complete within its interval
<etc>
```

Looking at the source for the Rabbit input, [the `gatherNodes` function starts a goroutine](https://github.com/influxdata/telegraf/blob/2f943e97b77b9e7599103169ba540a94204ea381/plugins/inputs/rabbitmq/rabbitmq.go#L480) for each node that it needs to run health checks against and then waits on a channel for the same number of results. If that health check call fails, it returns an error from inside the goroutine but never sends anything to the channel, so I expect the problem is that the initial failing call to `gatherNodes()` is stuck blocking forever waiting on that channel for a health check that it will never receive.

danielnelson:
closes #6814 

### Required for all PRs:

### DIFF
--- a/plugins/inputs/rabbitmq/rabbitmq.go
+++ b/plugins/inputs/rabbitmq/rabbitmq.go
@@ -510,6 +510,7 @@ func gatherNodes(r *RabbitMQ, acc telegraf.Accumulator) {
 			err := r.requestJSON("/api/healthchecks/node/"+node.Name, &health)
 			if err != nil {
 				acc.AddError(err)
+				return
 			}
 
 			if health.Status == "ok" {
@@ -522,6 +523,7 @@ func gatherNodes(r *RabbitMQ, acc telegraf.Accumulator) {
 			err = r.requestJSON("/api/nodes/"+node.Name+"/memory", &memory)
 			if err != nil {
 				acc.AddError(err)
+				return
 			}
 
 			if memory.Memory != nil {
--- a/plugins/inputs/rabbitmq/rabbitmq.go
+++ b/plugins/inputs/rabbitmq/rabbitmq.go
@@ -448,134 +448,110 @@ func gatherOverview(r *RabbitMQ, acc telegraf.Accumulator) {
 }
 
 func gatherNodes(r *RabbitMQ, acc telegraf.Accumulator) {
-	allNodes := make([]Node, 0)
-	// Gather information about nodes
+	allNodes := make([]*Node, 0)
+
 	err := r.requestJSON("/api/nodes", &allNodes)
 	if err != nil {
 		acc.AddError(err)
 		return
 	}
 
-	nodes := make(map[string]Node)
+	nodes := allNodes[:0]
 	for _, node := range allNodes {
 		if r.shouldGatherNode(node) {
-			nodes[node.Name] = node
+			nodes = append(nodes, node)
 		}
 	}
 
-	numberNodes := len(nodes)
-	if numberNodes == 0 {
-		return
-	}
-
-	type NodeCheck struct {
-		NodeName    string
-		HealthCheck HealthCheck
-		Memory      *Memory
-	}
-
-	nodeChecksChannel := make(chan NodeCheck, numberNodes)
-
+	var wg sync.WaitGroup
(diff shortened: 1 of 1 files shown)


plugins/inputs/rabbitmq/rabbitmq.go
```diff
@@ -448,134 +448,110 @@ func gatherOverview(r *RabbitMQ, acc telegraf.Accumulator) {
...
-			err = r.requestJSON("/api/nodes/"+nodeName+"/memory", &memoryresponse)
-			nodeCheck.Memory = memoryresponse.Memory
+			if health.Status == "ok" {
+				fields["health_check_status"] = int64(1)
+			} else {
+				fields["health_check_status"] = int64(0)
+			}
+
+			var memory MemoryResponse
+			err = r.requestJSON("/api/nodes/"+node.Name+"/memory", &memory)
```
reimda: Is it ok to make another request when the health check failed?  Now that we have "defer wg.done()" shouldn't we return early when health check fails?

danielnelson: Second request is safe to do, but is somewhat likely to fail in the case of an earlier error from requestJSON.  It is probably best to not create a partial metric, so I'll change this to return early.

PR #6820, opened 2019-12-23
Fixes for two pgbouncer issues. 

wingedpig:
The pgbouncer input plugin was not checking for errors from the 'SHOW POOLS' query. For at least newer versions of pgbouncer, this query was failing, silently, causing none of those values to be recorded. The second issue was the cause of that query failing. For whatever reason, we need to register pgtype.NumericOID because pgbouncer is now sending data with that. See https://github.com/jackc/pgx/issues/649 for a discussion about this.

### Required for all PRs:

### DIFF
--- a/plugins/inputs/pgbouncer/pgbouncer.go
+++ b/plugins/inputs/pgbouncer/pgbouncer.go
@@ -2,6 +2,7 @@ package pgbouncer
 
 import (
 	"bytes"
+	"strconv"
 
 	"github.com/influxdata/telegraf"
 	"github.com/influxdata/telegraf/internal"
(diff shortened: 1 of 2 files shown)


plugins/inputs/pgbouncer/pgbouncer.go
```diff
@@ -70,12 +71,18 @@ func (p *PgBouncer) Gather(acc telegraf.Accumulator) error {
 		for col, val := range columnMap {
 			_, ignore := ignoredColumns[col]
 			if !ignore {
-				fields[col] = *val
+				// these values are returned by pgbouncer as strings, which we need to convert.
+				fields[col], _ = strconv.ParseUint((*val).(string), 10, 64)
```
danielnelson: This will cause a panic with older pgbouncer versions, since then it came in as an int64, easy fix though I'll just take care of it post merge.

danielnelson: cb2c26a3324964d432bf83f535eca7b4d7879db3

PR #6832, opened 2019-12-29
WIP: [Varnish] Set tag with vcl active, for VBE.* only retry metrics from vcl active

### LINKED ISSUE #5622: varnish plugin should cleanup backend metric name
## Feature Request
Varnish plugin dumps the backend list as metric names, making this fields mostly useless for query. We need a common name for the metric and the backend names and profiles should be pushed to labels

### Proposal:
Either parse the varnish VBE metrics and extract the labels for the backends instead of dumping then in the metric name OR rename varnish VBE metric to a more friendly name like varnish_vbe_* , so we can try to fix this outside by renaming the metrics (at least prometheus can do this, do not know about other tools)

### Current behavior:
right now if we request the varnish plugin to gather backend metrics, we get this:

image

I can't even try to rename/relabel this in prometheus, as it's directly a varnish_*, so i can't create a rule to parse this as it would also parse all other metrics. There is no common anchor for all backend metrics names, in my example, the boot is the profile name, loading more profiles i can get any name (eg: profile named cache), so it can collide with existent metrics

### Desired behavior:
The current setup is good for all metrics except VBE, with it, its mostly useless, i can't use those names in anything as it would require me to know the profile, backend name and ip. instead it should take the VBE metric as a special metric and extract the labels from the name and create the varnish_vbe_{comm,happy,req,etc}

Getting backend metrics and health status in varnish is important information, but the plugin takes the VBE metric as a normal metric and it should be special as it can have random info. Current setup can overload any system with useless metric names, making everything slow

ismaelpuerto:
### Required for all PRs:

Fixed #5622. Let me keep as WIP until I have Unit Tests

### DIFF
--- a/plugins/inputs/varnish/varnish.go
+++ b/plugins/inputs/varnish/varnish.go
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"fmt"
 	"os/exec"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
@@ -22,17 +23,20 @@ type runner func(cmdName string, UseSudo bool, InstanceName string, Timeout inte
 // Varnish is used to store configuration values
 type Varnish struct {
 	Stats        []string
-	Binary       string
+	StatBinary   string
+	AdmBinary    string
 	UseSudo      bool
 	InstanceName string
 	Timeout      internal.Duration
 
-	filter filter.Filter
-	run    runner
+	filter  filter.Filter
+	statrun runner
+	admrun  runner
 }
 
 var defaultStats = []string{"MAIN.cache_hit", "MAIN.cache_miss", "MAIN.uptime"}
-var defaultBinary = "/usr/bin/varnishstat"
+var defaultStatBinary = "/usr/bin/varnishstat"
+var defaultAdmBinary = "/usr/bin/varnishadm"
 var defaultTimeout = internal.Duration{Duration: time.Second}
 
 var sampleConfig = `
@@ -40,7 +44,10 @@ var sampleConfig = `
   #use_sudo = false
 
   ## The default location of the varnishstat binary can be overridden with:
-  binary = "/usr/bin/varnishstat"
+  varnishstat_binary = "/usr/bin/varnishstat"
+
+  ## The default location of the varnishadm binary can be overridden with:
+  varnishsadm_binary = "/usr/bin/varnishstat"
 
   ## By default, telegraf gather stats for 3 metric points.
   ## Setting stats will override the defaults shown below.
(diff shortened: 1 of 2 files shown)


kfkawalec:
when we can expect that this change will be accepted?

anthosz:
+1

ismaelpuerto:
Hello @helenosheaa give me some days to adapt code and resolve conflicts. Thanks.

plugins/inputs/varnish/varnish.go
```diff
@@ -157,6 +232,7 @@ func (s *Varnish) Gather(acc telegraf.Accumulator) error {
 	for section, fields := range sectionMap {
 		tags := map[string]string{
 			"section": section,
+			"vcl":     vclactive,
```
helenosheaa, line 235: what are your thoughts on adding it as a field named vcl instead? Due to the timestamp being included in the name of the vcl there are concerns around cardinality with it as a tag.

plugins/inputs/varnish/varnish_test.go
```diff
@@ -114,6 +125,12 @@ MEMPOOL.vbc.pool                   10          .   In Pool
 MEMPOOL.vbc.sz_wanted              88          .   Size requested
 `
 
+var admOutput = `
+available   auto/warm          0 boot
```
helenosheaa, line 129: What version of varnish-cache are you using? For me the output from varnishadm looks like this:

`active      auto warm          0 reload_20200301_181101_4812`

plugins/inputs/varnish/varnish.go
```diff
@@ -115,13 +149,35 @@ func (s *Varnish) Gather(acc telegraf.Accumulator) error {
...
+	admscanner := bufio.NewScanner(admout)
+	vclactive := ""
+	for admscanner.Scan() {
+		vcllist := strings.Fields(admscanner.Text())
+		if len(vcllist) == 0 {
+			continue
+		}
+
+		status := vcllist[0]
+		vclname := vcllist[3]
```
helenosheaa, line 166: For me the name would be in another one down `vclname := vcllist[4]`. What version are you using?

plugins/inputs/varnish/varnish.go
```diff
@@ -22,25 +23,31 @@ type runner func(cmdName string, UseSudo bool, InstanceName string, Timeout inte
 // Varnish is used to store configuration values
 type Varnish struct {
```
helenosheaa, line 24: Add an option of metric_version which defaults to 1. Which `metric_version = 2` adding this new behavior. See the [prometheus input](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/prometheus#configuration) for an example.

helenosheaa (commented):
Thanks for looking back at this PR. I've added a couple of review comments. 

Have you found any other fieldnames which are problematic other than the ones containing the vclname (which includes a timestamp)?

plugins/inputs/varnish/varnish_test.go
anthosz: Hi @helenosheaa ,

If it can help you, on Varnish 6.5:
- After a restart:
active   auto    warm         0    boot
- After a reload:
active      auto    warm         2    reload8

anthosz: But can be interesting to extract only the backend name from VBE instead to take into account the reload name
Example via varnishstat:
Via boot:
Replace the name "VBE.boot.node1.happy" by "VBE.node1.happy" 
Via reload:
Replace the name "VBE.reload8.node1.happy" by "VBE.node1.happy"

plugins/inputs/varnish/varnish.go
ismaelpuerto: You will have as many tags as reload you run. Maybe could make sense send this tag if user configure it, right?

plugins/inputs/varnish/varnish.go
ismaelpuerto: good catch, in the moment of PR, vclname lived in position 3, now with last varnish the position change it. This is exit from varnishadm vcl.list between different versions:
varnish 6.0:
```active      auto/warm          0 boot```
varnish 6.6:
```active   auto    warm         0    boot```

Maybe make more sense execute ```varnishadm vcl.list -j```, this command give same between versions. But this option only are supported with varnish >= 6.

helenosheaa: That makes sense. It looks like [all versions prior to 6.0.7 unsupported](https://varnish-cache.org/releases/index.html) by varnish-cache now anyway

plugins/inputs/varnish/varnish.go
helenosheaa: It could be a user configured tag option or if it's a single field it should avoid cardinality issues.

sjwang90:
@ismaelpuerto Is this PR still required now that #9432 is in Telegraf?

PR #6835, opened 2019-12-30
[Input: Diskio] Invalid diskinfo cache if the device Mtim has changed

### LINKED ISSUE #6345: [Input: DiskIO] Deleted devices are not deleted from the diskInfoCache
### Relevant telegraf.conf:
```toml
[[inputs.diskio]]
  name_templates = ["$ID_FS_LABEL","$DM_VG_NAME/$DM_LV_NAME"]
```

### System info:

```
$ telegraf --version
Telegraf 1.10.2 (git: HEAD 3303f5c3)
 uname -ar
Linux [REDACTED] 4.4.0-151-generic #178-Ubuntu SMP Tue Jun 11 08:30:22 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
```

### Steps to reproduce:

1. Create a Logival Volume (using LVM) with the name `lv-a`
2. Let telegraf gather some points
3. Delete the LV `lv_a`
4. Create a LV with the name `lv-b`
5. Let telegraf gather some points

### Expected behavior:

The new points should have the tag: `vg-1/lv-b`

### Actual behavior:

The new points will have the tag: `vg-1/lv-a`

### Additional info:

This is because linux reuse DM ID. Let's say that the first LV had the DM ID 39 its path would be `/dev/dm-39`.

When deleting the LV, this dm ID will be freed. Thus linux can (and will) reuse it when we'll create another LV.

However since telegraf use a cache for the device infos (https://github.com/influxdata/telegraf/blob/master/plugins/inputs/diskio/diskio_linux.go#L33). It will always return the `lv-a` informations.

A solution would be to track DM deletions (see if something is in the cache but not in the device list) and remove it from the cache.

johnsudaar:
closes #6345

### Required for all PRs:

### DIFF
--- a/plugins/inputs/diskio/diskio_linux.go
+++ b/plugins/inputs/diskio/diskio_linux.go
@@ -11,6 +11,7 @@ import (
 )
 
 type diskInfoCache struct {
+	modifiedAt   int64 // Unix Nano timestamp of the last modification of the device. This value is used to invalidate the cache
 	udevDataPath string
 	values       map[string]string
 }
@@ -31,7 +32,8 @@ func (s *DiskIO) diskInfo(devName string) (map[string]string, error) {
 		s.infoCache = map[string]diskInfoCache{}
 	}
 	ic, ok := s.infoCache[devName]
-	if ok {
+
+	if ok && stat.Mtim.Nano() == ic.modifiedAt {
 		return ic.values, nil
 	}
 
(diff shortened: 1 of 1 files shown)


johnsudaar:
Fix #6345

johnsudaar:
I'm unsure on how to write the tests since the specs uses a dummy device on `/dev/null` however i can't modify the mtim of this device.

danielnelson:
I think it is okay to skip unit tests for now, we would need to make much larger changes to test this particular aspect.

Issue #6842, opened 2020-01-01
Source and port tags are missing in jenkins job metrics


vikkyomkar:
### Relevant telegraf.conf:
```
[[inputs.jenkins]]
  ## The Jenkins URL in the format "schema://host:port"
  url = "http://jenkins.prod.test.com:80"
  username = "admin"
  password = "admin"

  ## Set response_timeout
  response_timeout = "5s"

```

### System info:
Telegraf version: Latest [Build from source code]
operating system: Linux Box [Centos7.X]

### Steps to reproduce:

1. Use Jenkins input plugin.
2. Run it to see the tags of jenkins_node and jenkins_jobs metrics

### Expected behavior:

jenkins_jobs metrics should have source and port tags similar to jenkins_node metrics

### Actual behavior:

source and port tags are missing in jenkins_jobs metrics 

### Additional info:

Plugin code needs to be optimised

vikkyomkar:
I would like to look into this issue

vikkyomkar:
#6844

PR #6844, opened 2020-01-01
Added source and port tags in jenkins_job metrics

vikkyomkar:
### Required for all PRs:

- [x] Signed [CLA](https://influxdata.com/community/cla/). - Done
- [x] Associated README.md updated. -  Done
- [x] Has appropriate unit tests. - Done

-> Earlier "url.Parse" function was called "N Nodes" and "J Jobs" times which is now reduced to just  2 calls by creating separate function for URL parsing and source, port split.
-> Added missing source and port tags in jenkins_job metrics

### DIFF
--- a/plugins/inputs/jenkins/jenkins.go
+++ b/plugins/inputs/jenkins/jenkins.go
@@ -167,9 +167,20 @@ func (j *Jenkins) initialize(client *http.Client) error {
 	return j.client.init()
 }
 
-func (j *Jenkins) gatherNodeData(n node, acc telegraf.Accumulator) error {
-
+// Function to get source and port tags
+func (j *Jenkins) getInitialTags() (map[string]string, error) {
 	tags := map[string]string{}
+	u, err := url.Parse(j.URL)	
+	if err != nil {
+		return tags, err
+		}
+	tags["source"] = u.Hostname()
+	tags["port"] = u.Port()
+	return tags, nil
+}
+
+func (j *Jenkins) gatherNodeData(tags map[string]string, n node, acc telegraf.Accumulator) error {
+
 	if n.DisplayName == "" {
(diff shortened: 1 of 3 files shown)


vikkyomkar:
#6842

plugins/inputs/jenkins/jenkins.go
```diff
@@ -432,8 +448,10 @@ func (jr jobRequest) parentsString() string {
 	return strings.Join(jr.parents, "/")
 }
 
-func gatherJobBuild(jr jobRequest, b *buildResponse, acc telegraf.Accumulator) {
-	tags := map[string]string{"name": jr.name, "parents": jr.parentsString(), "result": b.Result}
+func gatherJobBuild(tags map[string]string, jr jobRequest, b *buildResponse, acc telegraf.Accumulator) {
+	tags["name"] = jr.name
+	tags["parents"] = jr.parentsString()
+	tags["result"] = b.Result
```
danielnelson: I like the idea but I think this introduces a race condition with this tag map, since the jobs are collected concurrently and there is only one map being written here.

I'd suggest going about this a slightly different way.  If you define a `func (j *Jenkins) Init() error` function it will be called when the plugin is created, you can parse the URL and add the `source` and `port` tags to the Jenkins struct.  Then we can add them to the new map here and we won't need to pass the map down.

vikkyomkar: @danielnelson Code is updated as per the suggestion. Now
> Source and port tags are part of Jenkins struct
> Tags are not passed down to the functions as argument.

vikkyomkar:
@danielnelson Code is updated as per the suggestion. Now

* Source and port tags are part of Jenkins struct
* Tags are not passed down to the functions as argument.

Can you please review the changes now ?

Issue #6834, opened 2019-12-30
SQL Server Input Plugin - Add availability group configuration and state metrics


mikedavem:
## Feature Request

To monitor the health / configuration of SQL Server availability group. It would be good to add TSQL some queries where results would be collected by the SQL Server Input Plugin.

#### Underlying Windows Failover Cluster info

`SELECT 
	member_name,
	member_type_desc AS member_type,
	member_state_desc AS member_state,
	number_of_quorum_votes
FROM sys.dm_hadr_cluster_members;`

#### AG level info => Recovery health + synchronization state

`SELECT 
	g.name as ag_name,
	rgs.primary_replica,
	rgs.primary_recovery_health_desc AS [primary_recovery_health],
	rgs.synchronization_health_desc AS [synchronization_health]
FROM sys.dm_hadr_availability_group_states as rgs
JOIN sys.availability_groups AS g ON rgs.group_id = g.group_id;`

#### AG replica level info => synchronization state + operational state + recovery health

`SELECT 
	g.name as ag_name,
	LEFT(r.replica_server_name, CHARINDEX('\', r.replica_server_name) - 1) AS replica_cluster_name,
	r.replica_server_name,
	rs.is_local,
	role_desc AS [role],
	rs.operational_state_desc AS [operational_state],
	rs.connected_state_desc AS [connected_state],
	rs.recovery_health_desc AS [recovery_health],
	rs.synchronization_health_desc AS [synchronization_health]
FROM sys.dm_hadr_availability_replica_states AS rs
JOIN sys.availability_replicas AS r
	ON rs.replica_id = r.replica_id
JOIN sys.availability_groups AS g
	ON g.group_id = r.group_id`

#### AG database level info => synchronization state + operational state + database state recovery health

`SELECT 
	g.name as ag_name,
	LEFT(r.replica_server_name, CHARINDEX('\', r.replica_server_name) - 1) AS replica_cluster_name,
	r.replica_server_name,
	DB_NAME(drs.database_id) AS [database_name],
	drs.is_local,
	drs.is_primary_replica,
	drs.synchronization_health_desc AS [synchronization_health],
	drs.synchronization_state_desc AS [synchronization_state],
	drs.database_state_desc AS [database_state],
	drs.is_suspended,
	drs.suspend_reason_desc AS [suspend_reason],
	drs.secondary_lag_seconds
FROM sys.dm_hadr_database_replica_states AS drs
		 JOIN sys.availability_replicas AS r
[...]

PR #6854, opened 2020-01-04
SQL Server plugin: add availabity group feature

mikedavem:
### Required for all PRs:

Issue #6834


danielnelson:
@denzilribeiro Can you review?

mikedavem:
> @denzilribeiro Can you review?

Hi @denzilribeiro  / @danielnelson 

Did you get a chance to review it ? :)

Regards

denzilribeiro:
Sorry for the delay  - some of these DMV's are only in SQL 14 later ( which is fine ) but you may be better off doing a version check given am not sure if telegraf being used against older versions. For example sys.dm_hadr_cluster. Also should exclude collecting if it Azure versions as it won't be relevant and all the SQL Server queries are still run. Same with the other HADR DMV's..

mikedavem:
@denzilribeiro 

Thanks for considering my PR :) Yes, it does make sense to check the version in this case so I added it to the AG queries. I just committed the new changes. If I'm not wrong after checking on documentation, all AG DMVs used here are available since SQL 12.

mikedavem:
@denzilribeiro . Do you think we can consider this PR in a future version? :) Many thanks for your feedback

denzilribeiro:
@mikedavem  I looked at some of this briefly
- Why would you need things like log send queue size etc  - Aka they are already being collected in the perfmon collector so why collect them again?

https://docs.microsoft.com/en-us/sql/database-engine/availability-groups/windows/monitor-performance-for-always-on-availability-groups?view=sql-server-ver15

Also I am still not seeing a check for SERVERPROPERTY('EngineEdition') to exclude Azure.
The other concern I have is some of the information being collected is really AG config and State config. If the idea is to put alerts on say a status change of AG, Telegraf is not a right place, alerts should happen on a state change/error, not via a polling mechanism every n seconds which is what the telegraf collector is.

What is end goal of this The config collection - How would you use this info?

mikedavem:
@denzilribeiro 

Thanks for the feedback and I can fix different issues accordingly. 

**Why would you need things like log send queue size etc - Aka they are already being collected in the perfmon collector so why collect them again?**

You're right and it is a duplicate from existing counters in Telegraf. I can fix it.

**Also I am still not seeing a check for SERVERPROPERTY('EngineEdition') to exclude Azure.**

I can fix it as well. Thanks

**What is end goal of this The config collection - How would you use this info?**

My intention is not to trigger alerts for AG state changes. There is already another place for it in my case :). In fact, I would like to get the same kind of information as we can have with sqlServerPropertiesV2 as the SQL instance version, edition, environment type (physical, virtual etc ...), last restart time, nb of databases and their states (online, in recovery etc ...).

With existing metrics and "state information" provided by sqlServerPropertiesV2, we already created this kind of dashboard (summary dashboard for all environments and detailed dashboard per instance)

image

Currently there exists metrics for AGs we already extract and consume through our dashboards like but there is no real information about underlying AG infrastructure like nb of replicas (and corresponding states),  nb of AG (and corresponding states), nb of high available DBs (and corresponding states), which instance is primary or secondary which may be useful to have for extend our system observability. 

Probably I may rewrite the queries to fit better with the telegraf "mindset" like you did with sqlServerPropertiesV2 ? Let me know what do you think about it?

denzilribeiro:
@mikedavem  my apologies never got back to this slipped through the cracks. There have been many changes done, can you check https://github.com/influxdata/telegraf/pull/8379 and see if it addresses your needs too?

mikedavem:
No worries @danielnelson . 
I just made a proposal to add info of the top-level AG state in the replica level query.
Of course we can define / compute the state of the AG itself with sub-level states (AG replicas or AG databases) but having directly the info could make the life easier with dash boarding stuff. 
Anyway, I'm glad to see queries from #8379 including to a next release of this plugin :)

sjwang90:
Should we close this now that #8379 is merged or is there additional information in this PR?

srebhan:
@mikedavem any answer to @sjwang90's question? Can we close this PR?

dba-leshop:
That's ok for me. We can lose this PR

PR #6857, opened 2020-01-04
Docker container logs input added

i-prudnikov:
__The primary motivation for this input plugin is to provide the following
features that ain't supported by current plugins (docker_log):__
1. Allow to set limits to how often and how many log data is read from the
each docker container. This primary use case here is to be protected from the situation 
lots of log entries (stack-tracing for example). In case wy simply stream this log entries as it is, via telegraf plugin, we will get high CPU utilisation of telegraf, and in case there are several such containers on the host we can hit the CPU limit.
3. Optimisation for running under k8s for streaming logs from the containers in a POD.
Moreover, using telegraf provides great flexibility for tagging and filtering, that is of great help. The main feature here is to shutdown telegraf when all containers from which we are streaming logs are exited/terminated. This allows to terminate POD in a consistent way, after all logs are delivered.
---
### Required for all PRs:

### DIFF
--- a/plugins/inputs/docker_cnt_logs/README.md
+++ b/plugins/inputs/docker_cnt_logs/README.md
@@ -1,0 +1,123 @@
+# Docker container logs Input plugin
+
+###The docker_cnt_logs plugin uses docker API to stream logs from container.
+
+---
+__The primary motivation for this input plugin is to provide the following
+features that are not support by current plugins:__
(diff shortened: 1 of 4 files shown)


i-prudnikov:
Will reopen win more meaningful branch name

i-prudnikov:
will reopen with more meaningful branch name

Issue #5639, opened 2019-03-26
vSphere: omit objects by name


kkruzich:
RE: vsphere plugin 

I've followed some of the discussions stemming from #5165 and read through the docs on [Metric Filtering](https://github.com/influxdata/telegraf/blob/master/docs/CONFIGURATION.md#metric-filtering) yet I'm still not clear if there's any means of *excluding* objects by name. For example, in listing our datastores, I see the conventions below. I want to *exclude* all datastores that begin with `Local*` 

Have I overlooked something or is it not currently possible to do this? 

```
/datastore/e612af807be232cebab3cad7d1a0b568
/datastore/LocalStorage-USPDXESX363
/datastore/LocalStorage-USPDXESX017
/datastore/845bc80751cb34cf927700f7f616dad4
/datastore/Local-LGA-ESX038
/datastore/8e7804a1a5363272b001cd74f3223d1c
/datastore/d5573007bed737dd827991b7fa1d6d76
```

danielnelson:
We do have the `datastore_include = []` option but there is no converse option so it would be difficult to exclude items, especially since this option only supports `*` and `**` wildcards.  I believe you would need:
```
[[inputs.vsphere]]
  # ... snip ...
  [inputs.vsphere.tagdrop]
    dsname = ["Local*"]
```

Prydin, do you think it might be a good idea to add a `datastore_exclude = []`?

As an aside, I see there are some copy/paste typo's in the README and sample config around this option, so I'll fix those.

prydin:
Unfortunately, an exclude function is a bit tricky to implement due to the way the APIs work. I will put that on my list of things to do, but it may take a while before I get to it. 

To elaborate a bit: Excluding using full paths is a major change, but we could do a simple exclude on names only. The only issue I would have with that is that include and exclude would have different semantics.

prydin:
I'm working on a solution to this one. Expect a PR within the next couple of weeks or so.

PR #6859, opened 2020-01-06
[inputs.vsphere] Exclude resources by inventory path

prydin:
### Required for all PRs:

This PR addresses issue #5639 (omit metrics by name).

## Description
In addition to host_include, datastore_include, datacenter_include, cluster_include and vm_include configuration parameter, we are adding their *_exclude counterparts. They allow you to exclude any resource using an inventory path. For example ```vm_exclude = [/host/**/host1/**]``` would exclude any VM on the host ```host1```.

## Caveat
If the exclude inventory path contains a reference to an inventory node to which the service account does nto have access, the inventory discovery will fail. Rewrite such paths using wildcards instead so that you avoid referencing any nodes you don't have access to.


danielnelson:
@prydin Could you take a look at the conflict in endpoint.go?

prydin:
@danielnelson Done!

PR #6863, opened 2020-01-06
Docker container logs input added

i-prudnikov:
###Refactor pull request #6857 with more meaningful source branch name.
__The primary motivation for this input plugin is to provide the following
features that ain't supported by current plugins (docker_log):__
1. Allow to set limits to how often and how many log data is read from the
each docker container. This primary use case here is to be protected from the situation 
lots of log entries (stack-tracing for example). In case wy simply stream this log entries as it is, via telegraf plugin, we will get high CPU utilisation of telegraf, and in case there are several such containers on the host we can hit the CPU limit.
3. Optimisation for running under k8s for streaming logs from the containers in a POD.
Moreover, using telegraf provides great flexibility for tagging and filtering, that is of great help. The main feature here is to shutdown telegraf when all containers from which we are streaming logs are exited/terminated. This allows to terminate POD in a consistent way, after all logs are delivered.
---
### Required for all PRs:

### DIFF
--- a/plugins/inputs/docker_cnt_logs/README.md
+++ b/plugins/inputs/docker_cnt_logs/README.md
@@ -1,0 +1,134 @@
+# Docker container logs Input plugin
+
+###The docker_cnt_logs plugin uses docker API to stream logs from container.
+
+---
+__The primary motivation for this input plugin is to provide the following
+features that ain't supported by current plugins:__
+1. Allow to set limits to how often and how many log data is read from the
+each docker container. This primary use case here is to be protected from the situation 
+when the container (from which the plugin reading logs) fall in to unlimited loop producing 
+lots of log entries (stack-tracing for example). In case wy simply stream this log entries as it is, via telegraf plugin, we will get high CPU utilisation of telegraf, and in case there are several such containers on the host we can hit the CPU limit.
+2. Allows to stream logs from the particular point in time even if telegraf crashed. This is achieved by storing offset (unix time stamp in nanoseconds of the last read log entry) for every container. When plugin is started it checks for the offset existence, and in case it found it, the logs will be streamed since the offset, so no entries would be lost. Entries will come with the original time-stamp.
+3. Optimisation for running under k8s for streaming logs from the containers in a POD.
+Although there are some already available solutions to deliver logs from k8s containers,
+this feature is of interest, because single telegraf binary can be used at the same time to
+deliver metrics from the running applications and logs. No other solution needed.
+Moreover, using telegraf provides great flexibility for tagging and filtering, that is of great help. The main feature here is to shutdown telegraf when all containers from which we are streaming logs are exited/terminated. This allows to terminate POD in a consistent way, after all logs are delivered.
+features that are not support by current plugins:__
+1. Allow to set limits to how often and how many log data is read from the
+each docker container. This primary use case here is to be protected from the situation 
+when the container (from which the plugin reading logs) fall in to unlimited loop producing 
+lots of log entries (stack-tracing for example). In case wy simply stream this log entries as it is, via telegraf plugin,
+we will get high CPU utilization of telegraf, and in case there are several such containers on the host
+we can hit the CPU limit.
+2. Allows to stream logs from the particular point in time even if telegraf crashed. This is 
+achieved by storing offset (unix time stamp in nanoseconds of the last read log entry)
+for every container. When plugin is started it checks for the offset existence, and in case it found
+it, the logs will be streamed since the offset, so no entries would be lost. Entries will come with the
+original time-stamp.
+3. Optimization for running under k8s for streaming logs from the containers in a POD.
+Although there are some already available solutions to deliver logs from k8s containers,
+this feature is of interest, because single telegraf binary can be used at the same time to
+deliver metrics from the running applications and logs. No other solution needed.
+Moreover, using telegraf provides great flexibility for tagging and filtering, that is
+of great help. The main feature here is to shutdown telegraf when all containers from which we are streaming logs
+are exited/terminated. This allows to terminate POD in a consistent way, after all logs are delivered.
+---
+
+To be able to use it, docker socket should be provided for runtime,
+and docker container logging driver should be set to `json-file` or `journald`.
+
+To query API, the possible oldest version used - 1.21 (https://docs-stage.docker.com/engine/api/v1.21/), 
+to support as much of variety of docker versions as possible. 
+To stream logs the following API endpoint is used `GET /containers/(id or name)/logs`
+
+API version vs Docker version compatibility matrix: https://docs.docker.com/develop/sdk/
+(see `API version matrix` chapter)
+
+### Configuration:
+
+```toml
+[[inputs.docker_cnt_logs]]  
+  ## Interval to gather data from docker sock.
+  ## the longer the interval the fewer request is made towards docker API (less CPU utilization on dockerd).
+  ## On the other hand, this increase the delay between producing logs and delivering it. Reasonable trade off
+  ## should be chosen
+  interval = "2000ms"
+  
+  ## Docker Endpoint
+  ##  To use unix, set endpoint = "unix:///var/run/docker.sock" (/var/run/docker.sock is default mount path)
+  ##  To use TCP, set endpoint = "tcp://[ip]:[port]"
+  ##  To use environment variables (ie, docker-machine), set endpoint = "ENV"
+  endpoint = "unix:///var/run/docker.sock"
+
+  ## Optional TLS Config
+  # tls_ca = "/etc/telegraf/ca.pem"
+  # tls_cert = "/etc/telegraf/cert.pem"
+  # tls_key = "/etc/telegraf/key.pem"
+
+  ## Use TLS but skip chain & host verification
+  # insecure_skip_verify = false
+
+  ## Log streaming settings
+  ## Set initial chunk size (length of []byte buffer to read from docker socket)
+  ## If not set, default value of 'defaultInitialChunkSize = 1000' will be used
+  # initial_chunk_size = 1000 # 1K symbols (half of 80x25 screen)
+
+  ## Max chunk size (length of []byte buffer to read from docker socket)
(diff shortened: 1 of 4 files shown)


danielnelson:
Any improvements will need to go into the main docker log plugin, we wouldn't take a second docker logging plugin.  However, let's go over the 3 features at a high level:

> Allow to set limits to how often and how many log data is read from the
each docker container.

I couldn't understand from the README how this works, can you go into more detail?

> Allows to stream logs from the particular point in time even if telegraf crashed. This is achieved by storing offset (unix time stamp in nanoseconds of the last read log entry) for every container.

We have something similar in the tail plugin which allow Telegraf to resume at the same point when sent a SIGHUP, it doesn't save offsets to file currently though.  We could add this to the docker_log plugin if you can split it out as a separate pull request.  We need to make sure it will work in Windows/Linux/Mac/etc and that it will clean up after itself.

> The main feature here is to shutdown telegraf when all containers from which we are streaming logs are exited/terminated. This allows to terminate POD in a consistent way, after all logs are delivered.

Plugins stopping the Telegraf process isn't a feature we would accept at this time.

i-prudnikov:
Hello @danielnelson,
Let me provide addition info.
> Any improvements will need to go into the main docker log plugin, we wouldn't take a second docker logging plugin.

Well, I understand your rationale here, but let me highlight that the use cases for `docker_log` plugin and the current plugin is different. The current plugin is optimized to deliver logs when running inside a pod in a side-car container. Also, the plugins architecture is totally different (see below in details).

__Overall, I also see no point to have 2 plugins for nearly the same purpose, and I can propose to add some missing features of `docker_log` plugin (actually there is only one – see below) into current plugin, so to have one plugin that supports all features and optimizations. What do you think?__

_Missing feature:_ periodically scan docker API for the new containers or stopped containers and begin/stop logs’ streaming automatically. Current plugin simply reads the static config about what containers should be streamed from.


>>  Allow to set limits to how often and how many log data is read from the each docker container.

> I couldn't understand from the README how this works, can you go into more detail?

Within the interval, plugin reads limited amount of logs data from the docker API. The amount of data and interval is configurable. Current `docker_log` plugin simply copies the stream. At first I followed the same approach (the current plugin was developed half a year earlier than current `docker_log` plugin, we gave it a good testing  and run it in production before we submit this pull request) but it appears that it is not acceptable in case container start producing lots of logs (for example, "stack-tracing" in endless cycle - quite often use-case), the plugin consumes 100% CPU, as you copying non-stop the streamed logs, moreover, `dockerd` is also is of high CPU utilisation, obviously. So, it is no go for real production usage.
This is the reason why I don't like the idea to incorporate the changes from my plugin into the current one. The overall approach to read logs from container in the current plugin is totally different from the `docker_log` plugin. Also, the primary use-case of the current plugin is to deliver logs from the pod's containers in k8s, however other use-cases also supported.

>> Allows to stream logs from the particular point in time even if telegraf crashed. This is achieved by storing offset (unix time stamp in nanoseconds of the last read log entry) for every container.

>We have something similar in the tail plugin which allow Telegraf to resume at the same point when sent a SIGHUP, it doesn't save offsets to file currently though. We could add this to the docker_log plugin if you can split it out as a separate pull request. We need to make sure it will work in Windows/Linux/Mac/etc and that it will clean up after itself.

This sounds interesting, I support idea to reuse something that already works and tested. Regarding implementing changes into `docker_log` I see no problem doing that once I go through the whole idea (maybe it is good to involve the `docker_log` plugin maintainer into this discussion).

> Plugins stopping the Telegraf process isn't a feature we would accept at this time.
1. First of all it is configurable, you can either enable or disable this feature.
2. I would like to ask you to review how it was implemented, maybe you can point me out potentially dangerous code or explain the reason of your point of view.
3. This feature servers the primary use case for this plugin (that's why it is crucial to implement it… maybe there are some other ways to reach this) - to finish the primary container process (i.e. telegraf) when it is nothing to deliver anymore, let me explain in details: Imagine that the container with telegraf is running inside the k8s POD, the telegraf is configured to deliver logs and metrics from the containers in the POD. Once the containers in POD stopped (for example they execute some job, and job is done), we need to be able to terminate the POD. In case the telegraf container will be running it can be achieved, but if the telegraf container also stopped, the everything is in order. The logs and metrics from POD's container delivered and POD is terminated. The shut-down telegraf feature is realized as graceful shutdown, when all metrics is flushed and send to outputs, and so on, i.e. we can guarantee that the data is sent out from the telegraf.

danielnelson:
> Missing feature: periodically scan docker API for the new containers or stopped containers and begin/stop logs’ streaming automatically. Current plugin simply reads the static config about what containers should be streamed from.

The current `docker_log` plugin checks for new containers every interval (default 10s) and will being following them.  When containers stop and the end of log is reached they are also no longer tailed.

> At first I followed the same approach (the current plugin was developed half a year earlier than current docker_log plugin, we gave it a good testing and run it in production before we submit this pull request) but it appears that it is not acceptable in case container start producing lots of logs (for example, "stack-tracing" in endless cycle - quite often use-case), the plugin consumes 100% CPU, as you copying non-stop the streamed logs, moreover, dockerd is also is of high CPU utilisation, obviously. So, it is no go for real production usage.

What if we added rate limiting where if the log rate is exceeded we close the response from the docker engine?  I don't think this needs to be a large change from the current docker_log code.

> I would like to ask you to review how it was implemented, maybe you can point me out potentially dangerous code or explain the reason of your point of view.

I think requirements for shutdown would be too use-case specific.  Adding this feature would set a bad precedence and I don't want a new type of functionality just for this single use which I see as a bit of edge case.  Overall it's not something I would want to add today, maybe later but I'm not confident that it is a good addition at this point.

> In case the telegraf container will be running it can be achieved, but if the telegraf container also stopped, the everything is in order. The logs and metrics from POD's container delivered and POD is terminated. 

I feel like the signal to stop should come from outside of Telegraf.  How about something like [process namespace sharing](https://kubernetes.io/docs/tasks/configure-pod-container/share-process-namespace/#understanding-process-namespace-sharing) and the main container could signal Telegraf when the job is complete.

i-prudnikov:
Overall, I would like to emphasize, that I see no point refactoring current docker_log plugin, instead I can adjust my plugin to have as similar as possible configuration interface and output (metric names, tags, etc.)  + add missing features to get single plugin with all of the features. Do you agree with this approach? The reason why I'm proposing this approach is in my comments below.

>> Missing feature: periodically scan docker API for the new containers or stopped containers and begin/stop logs’ streaming automatically. Current plugin simply reads the static config about what containers should be streamed from.

> The current docker_log plugin checks for new containers every interval (default 10s) and will being following them. When containers stop and the end of log is reached they are also no longer tailed.

Yes, I'm talking exactly about this, it is not supported by my plugin, but I can add this.

>> At first I followed the same approach (the current plugin was developed half a year earlier than current docker_log plugin, we gave it a good testing and run it in production before we submit this pull request) but it appears that it is not acceptable in case container start producing lots of logs (for example, "stack-tracing" in endless cycle - quite often use-case), the plugin consumes 100% CPU, as you copying non-stop the streamed logs, moreover, dockerd is also is of high CPU utilisation, obviously. So, it is no go for real production usage.

> What if we added rate limiting where if the log rate is exceeded we close the response from the docker engine? I don't think this needs to be a large change from the current docker_log code.

Yes we can do smth. like this, but I see no point doing this as I already did that in my plugin, it reads only limited amount of data, with some pauses. 
And it is nothing you can do on the docker side for reate-limiting as docker API provides simple io.ReadCloser interface for accessing logs. And in case you ask to incorporate these changes into docker_log plugin, it will cause lots of changes (you need to address the following: concatenating message chunks, respect edge cases when message chunk ended somewhere inside docker log header, parse every message to detect original timestamp and so on...). So I see not point refactoring docker_log plugin for that, as I have already tested and working code in my plugin.

> I think requirements for shutdown would be too use-case specific. Adding this feature would set a bad precedence and I don't want a new type of functionality just for this single use which I see as a bit of edge case. Overall it's not something I would want to add today, maybe later but I'm not confident that it is a good addition at this point.

Indeed, this is use-case specific, for this particular plugin it makes LOTS of sence and really simplify operations. I bet there can be some other plugins/cases where it can be useful.

>> In case the telegraf container will be running it can be achieved, but if the telegraf container also stopped, the everything is in order. The logs and metrics from POD's container delivered and POD is terminated.

> I feel like the signal to stop should come from outside of Telegraf. How about something like process namespace sharing and the main container could signal Telegraf when the job is complete.

I spent some time looking around for solution before I came to signaling from inside. Sharing process namespaces is not a good option from security perspective. 
As far as I understand, you are talking about sending an event from the plugin to the some kind of process manager, that can signal telegraf process, yes it is possible, but imply some operational limitations and looks really synthetic:
- you need to have process supervisor in your telegraf container
- you need to implement some functionality that can handle event that is sent from plugin

Anyway, could you please point me out, what part of the following code is not good from your perspective?
```go
func (dl *DockerCNTLogs) shutdownTelegraf() {
	var err error
	var p *os.Process
	p, err = os.FindProcess(os.Getpid())
	if err != nil {
		log.Printf("E! [inputs.docker_cnt_logs] Can't get current process PID "+
			"to initiate graceful shutdown: %v.\nHave to panic for shutdown...", err)
	} else {
		if runtime.GOOS == "windows" {
			err = p.Signal(os.Kill) //Interrupt is not supported on windows
		} else {
			err = p.Signal(os.Interrupt)
		}
		if err != nil {
			log.Printf("W! [inputs.docker_cnt_logs] Can't send signal to main process "+
				"for initiating Telegraf shutdown, reason: %v\nHave to panic for shutdown...", err)
		} else {
			return
		}
	}

	panic(errors.New("Graceful shutdown is not possible, force panic."))
}
```
__If the abovementioned question is out of discussion, I will try to invent workaround.__

i-prudnikov:
Hello @srebhan, this one is indeed outdated. We can close it. At the end it was split to the smaller PRs:

- https://github.com/influxdata/telegraf/pull/7434 - Merged
- https://github.com/influxdata/telegraf/pull/7749 - Still actual and on a hold, as we are waiting for implementation of state interface for plugins (https://github.com/influxdata/telegraf/issues/8352)

PR #6865, opened 2020-01-06
Diskio (merge operations and average time for i/o)

victornet:
This PR extends the diskio plugin.

For analyzing storage specific issues, i tought it would be helpful to gather the average time for an i/o took to complete. 

Furthermore this request provides counters for merge-operations.

### Required for all PRs:


danielnelson:
Thanks for the pull request @victornet, but we don't normally add aggregate metrics like these that are calculated from the existing metrics.  With most of our outputs these are best done in the query language when displaying the data.  This is more flexible since you can calculate them over any time window and perform any aggregation, is more storage space friendly, and helps with maintenance since it keeps the plugins simple.

If you are interested in adding an example InfluxQL or Flux query to the README for this calculation that would be helpful.

PR #6875, opened 2020-01-08
update consul address with default port

like-inspur:


danielnelson:
Can you add this to consul.go as well: https://github.com/influxdata/telegraf/blob/master/plugins/inputs/consul/consul.go#L31

danielnelson:
Oh whoops, I got confused and push this up already.  Sorry about that.

0693748c35ba33dee3a7154c355de5477a16ac14

Issue #6877, opened 2020-01-08
Updated code for processors.date with offsets


Crashtein:
## Feature Request

I'm proposing a feature for processors.date. I needed grouping by time(1d,6h) (6 hours offset because of fact that shifts in factory starts at 6:00AM) and by new tag "month" - created by this plugin. All in factory counts between 6:00AM to 6:00AM.

### Proposal:

I've made some changes in /plugins/processors/date/date.go. Now there is a new parameter in .conf file called time_shift with working as described in code section toml. Here is complete new code of date plugin:

```
package date

import (
	"time"
	"github.com/influxdata/telegraf"
	"github.com/influxdata/telegraf/plugins/processors"
)

const sampleConfig = `
  ## New tag to create
  tag_key = "month"

  ## Date format string, must be a representation of the Go "reference time"
  ## which is "Mon Jan 2 15:04:05 -0700 MST 2006".
  date_format = "Jan"
  ## Time shift in hours. allows to configure tag value changing at diffierent time of day. 
  ## Input as string. Working example: if time_shift = "-6.0h", month will change 6 hours later.
  time_shift = "0.0h"
`

type Date struct {
	TagKey     string `toml:"tag_key"`
	DateFormat string `toml:"date_format"`
	TimeShift  string `toml:"time_shift"`
}

func (d *Date) SampleConfig() string {
	return sampleConfig
}

func (d *Date) Description() string {
	return "Dates measurements, tags, and fields that pass through this filter."
}

func (d *Date) Apply(in ...telegraf.Metric) []telegraf.Metric {
	for _, point := range in {
		timeshiftParsed, _ := time.ParseDuration(d.TimeShift)
		timeref := point.Time().Add(timeshiftParsed)
		point.AddTag(d.TagKey, timeref.Format(d.DateFormat))
	}

	return in
}

func init() {
	processors.Add("date", func() telegraf.Processor {
		return &Date{}
	})
}
```

### Current behavior:

In described case, when i group by tag month and time(1d,6h) at the last day of the month I get 2 values: for last month and new one which is a partial values of 6:00-6:00 work-day.

### Desired behavior:

With new code adding a tag month is based on time+/-time_shift so I can group by month without having two values at the end of the month, everything works as I wanted.

### Use case: 

Maybe in other factories it will be useful too. Of course code implements various tag offsets in standard time format duration. In this case you would be able to make new fields with values grouping by work-days/months/days easily grouping by tag added with new offset parameter.

In attachment the new code:

[...]

PR #6886, opened 2020-01-09
Add date offset option to date processor

danielnelson:
closes: #6877

### Required for all PRs:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@
   renamed to `sqlserver_azure_db_resource_stats` due to an issue where numeric
   metrics were previously being reported incorrectly as strings.
 
+- The `date` processor now uses the UTC timezone when creating its tag.  In
+  previous versions the local time was used.
+
 #### New Outputs
 
 - [warp10](/plugins/outputs/warp10/README.md) - Contributed by @aurrelhebert
--- a/plugins/processors/date/README.md
+++ b/plugins/processors/date/README.md
@@ -19,6 +19,9 @@ A few example usecases include:
   ## Date format string, must be a representation of the Go "reference time"
   ## which is "Mon Jan 2 15:04:05 -0700 MST 2006".
   date_format = "Jan"
+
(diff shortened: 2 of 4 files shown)


danielnelson:
Uh oh, it appears we have been setting the tag based on the local system time instead of using UTC to match the metric timestamps.  I will fix this but it will not be fully backwards compatible.

Crashtein:
> Uh oh, it appears we have been setting the tag based on the local system time instead of using UTC to match the metric timestamps. I will fix this but it will not be fully backwards compatible.

Are you sure it is correct? Most people use this plugin to put tag offset appropiate to their timezone. 
My machines are time synced and timezone has been set. Also there is a seasonal time change +/-1h in my country. When I put my point into DB through telegraf processors plugin it reads the same time stample as it has in origin machine.
So assume I want an offset -6.00h to change tag at 6:00AM local time. You want to base tag on UTC time, so finally the offset is UTC time + offset. In described case it will be -5.00h. Am I correct? Could you explain your point of view?

danielnelson:
I believe your calculation is correct.  My line of thinking here is that metrics are always reported in UTC, and adding a tag with a 0 offset should match the timestamp in the database, no matter the timezone of the system who collected it.

> Most people use this plugin to put tag offset appropiate to their timezone.

I don't have a good way to judge this.  I suspect the primary use of this plugin is to insert day of the week or month of the year, and it is possible that many users haven't noticed the discrepancy.  This would be a change of behavior though, so at the very least we should add this to the release notes.

Crashtein:
Ok, now I understand it. One more thing I can add to this discussion is a RTC time - reference time which may be used on linux with i.e. timezonectl. It lets you set hardware time to UTC or local. You may also consider to put an information about that and let the user choose the best solution for use.

danielnelson:
Rebased on master due to CHANGLOG.md conflict.

Added new `timezone` option to allow control over the time zone used.  Added release note about breaking change.

Issue #6906, opened 2020-01-15
Error running agent: json_strict is not defined


VitaliyBreslavskyi:
Hi @rawkode,

Submitting this bug report as requested in #6536
Could you please have a look?

Best regards,
Vitaliy

```
telegraf_1       | 2020-01-15T13:28:05Z I! Starting Telegraf 1.13.0
telegraf_1       | 2020-01-15T13:28:05Z I! Using config file: /etc/telegraf/telegraf.conf
telegraf_1       | 2020-01-15T13:28:05Z E! [telegraf] Error running agent: Error parsing /etc/telegraf/telegraf.d/input_json.conf, line 37: field corresponding to `json_strict' is not defined in file.File
tick-sandbox_telegraf_1 exited with code 1
```

## Relevant telegraf.conf:

```toml
[[inputs.file]]
  files = ["example"]

  ## Data format to consume.
  ## Each data format has its own unique set of configuration options, read
  ## more about them here:
  ## https://github.com/influxdata/telegraf/blob/master/docs/DATA_FORMATS_INPUT.md
  data_format = "json"

  ## When strict is true and a JSON array is being parsed, all objects within the
  ## array must be valid
  json_strict = false

  ## Query is a GJSON path that specifies a specific chunk of JSON to be
  ## parsed, if not specified the whole document will be parsed.
  ##
  ## GJSON query paths are described here:
  ##   https://github.com/tidwall/gjson/tree/v1.3.0#path-syntax
  json_query = ""

  ## Tag keys is an array of keys that should be added as tags.  Matching keys
  ## are no longer saved as fields.
  tag_keys = [
    "my_tag_1",
    "my_tag_2"
  ]

  ## Array of glob pattern strings keys that should be added as string fields.
  json_string_fields = []

  ## Name key is the key to use as the measurement name.
  json_name_key = ""

  ## Time key is the key containing the time that should be used to create the
  ## metric.
  json_time_key = ""

  ## Time format is the time layout that should be used to interprete the json_time_key.
  ## The time must be `unix`, `unix_ms`, `unix_us`, `unix_ns`, or a time in the
  ## "reference time".  To define a different format, arrange the values from
  ## the "reference time" in the example to match the format you will be
  ## using.  For more information on the "reference time", visit
  ## https://golang.org/pkg/time/#Time.Format
  ##   ex: json_time_format = "Mon Jan 2 15:04:05 -0700 MST 2006"
[...]

Issue #6532, opened 2019-10-16
JSON parser: how to handle data with missing time keys


russorat:
## Feature Request

Opening a feature request kicks off a discussion.

### Proposal:
Add a way to handle json data with missing time keys

### Current behavior:
json parsing fails the entire document

### Desired behavior:
1. the json parser should have a more helpful error message than "JSON time key could not be found". Ideally it would print the record it failed on
1. there could be a config option to ignore records with missing time keys, or an option to use a default time for missing time keys

### Use case: 
extends the capability of the json parser. enables more json files to be parsed without the need for custom application logic.

``` toml
[[inputs.http]]
  interval = "10s"
  urls = [
    "https://data.nasa.gov/resource/y77d-th95.json"
  ]
  name_override = "meteorevent"
  method = "GET"
  timeout = "5s"
  data_format = "json"
  json_query = ""
  tag_keys = [
    "nametype",
    "recclass",
    "fall",
    "geolocation_latitude",
    "geolocation_longitude",
    "year"
  ]
  json_string_fields = [
    "name",
    "id"
  ]
  json_name_key = ""
  json_time_key = "year"
  json_time_format = "2006-01-02T15:04:05.000"
```

rawkode:
I don't think using the current time, in the event of a missing key, would ever be the intended behaviour (IMO)

I think by default the parser should continue on error, rather than rejecting the whole file. Perhaps change the default behaviour to that, but as a strict flag to revert to the current behaviour?

PR #6909, opened 2020-01-16
Hook up json_strict option with default of true

danielnelson:
closes: #6906

### Required for all PRs:

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -1411,7 +1411,9 @@ func buildParser(name string, tbl *ast.Table) (parsers.Parser, error) {
 }
 
 func getParserConfig(name string, tbl *ast.Table) (*parsers.Config, error) {
-	c := &parsers.Config{}
+	c := &parsers.Config{
+		JSONStrict: true,
+	}
 
 	if node, ok := tbl.Fields["data_format"]; ok {
 		if kv, ok := node.(*ast.KeyValue); ok {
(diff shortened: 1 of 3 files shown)


plugins/parsers/json/README.md
```diff
@@ -20,7 +20,7 @@ ignored unless specified in the `tag_key` or `json_string_fields` options.
 
   ## When strict is true and a JSON array is being parsed, all objects within the
   ## array must be valid
-  strict = false
+  strict = true
```
reimda: Doesn't this need to be named json_strict?

internal/config/config.go
```diff
@@ -1411,7 +1411,9 @@ func buildParser(name string, tbl *ast.Table) (parsers.Parser, error) {
 }
 
 func getParserConfig(name string, tbl *ast.Table) (*parsers.Config, error) {
-	c := &parsers.Config{}
+	c := &parsers.Config{
+		JSONStrict: true,
```
reimda: I think @rawkode intended to have it default false to fix #6532, see https://github.com/influxdata/telegraf/issues/6532#issuecomment-542928948  Won't default true cause a regression there?

plugins/parsers/json/README.md
danielnelson: Yes, thanks.

internal/config/config.go
danielnelson: The regression is in 1.13.0 because we changed the parse behavior, I disagree that this should parse be accepted by default because it has the non-obvious behavior of discarding the object.

Issue #6894, opened 2020-01-14
[[inputs.kibana]] cannot get the correct metrics


congwensheng:
hi, i am use the kibana inputs plugin,this is my configuration:
[[inputs.kibana]]
  servers = ["http://localhost:5601"]
  timeout = "5s"

then,i get the metrics like this:
HELP kibana_concurrent_connections Telegraf collected metric
TYPE kibana_concurrent_connections untyped
kibana_concurrent_connections{host="mgt01",name="kibana",source="10.150.83.158:5601",status="green",version="6.5.4"} 0
 HELP kibana_heap_max_bytes Telegraf collected metric
 TYPE kibana_heap_max_bytes untyped
kibana_heap_max_bytes{host="mgt01",name="kibana",source="10.150.83.158:5601",status="green",version="6.5.4"} 0
HELP kibana_heap_used_bytes Telegraf collected metric
TYPE kibana_heap_used_bytes untyped
kibana_heap_used_bytes{host="mgt01",name="kibana",source="10.150.83.158:5601",status="green",version="6.5.4"} 0
HELP kibana_requests_per_sec Telegraf collected metric
TYPE kibana_requests_per_sec untyped
kibana_requests_per_sec{host="mgt01",name="kibana",source="10.150.83.158:5601",status="green",version="6.5.4"} 0.2
HELP kibana_response_time_avg_ms Telegraf collected metric
 TYPE kibana_response_time_avg_ms untyped
kibana_response_time_avg_ms{host="mgt01",name="kibana",source="10.150.83.158:5601",status="green",version="6.5.4"} 2
HELP kibana_response_time_max_ms Telegraf collected metric
TYPE kibana_response_time_max_ms untyped
kibana_response_time_max_ms{host="mgt01",name="kibana",source="10.150.83.158:5601",status="green",version="6.5.4"} 2
HELP kibana_status_code Telegraf collected metric
TYPE kibana_status_code untyped
kibana_status_code{host="mgt01",name="kibana",source="10.150.83.158:5601",status="green",version="6.5.4"} 1
HELP kibana_uptime_ms Telegraf collected metric
TYPE kibana_uptime_ms untyped
kibana_uptime_ms{host="mgt01",name="kibana",source="10.150.83.158:5601",status="green",version="6.5.4"} 0

[...]

Issue #5744, opened 2019-04-18
Update Kibana input plugin with support Kibana 6.6.x, 6.7.x and 7.0.x


yuriy-slobodchikov:
## Feature Request
Update Kibana input plugin with support Kibana 6.6.x, 6.7.x and 7.0.x

### Proposal:
Add configuration to point version of kibana to use - like string constants for example
kibana_version = "6.3.x"  "6.6.x" "6.7.x" "7.0.x"
for each version implement its own json mapping and parsing response.

### Current behavior:
Looks like in Kibana 6.6.x version json of /api/status was changed
and now kibana plugin cannot parse response for Kibana 6.6.x, 6.7.x and 7.0.x

### Desired behavior:
In configurstion of plugin there should be some field pointing version.
Kibana plugin should support Kibana 6.6.x, 6.7.x and 7.0.x

### Use case: 
After upgrading from 6.3 to 6.7.1 kibana plugin fail to parse new format of response.

yuriy-slobodchikov:
Hello @lpic10 
Please look into this ticket, may be you have some time to implement it.

vikkyomkar:
@yuriy-slobodchikov / @danielnelson I have raised a PR to address this issue

PR #6923, opened 2020-01-18
Kibana Input Plugin: added compatibility for all 6.X.X versions

vikkyomkar:
### Required for all PRs:

### DIFF
--- a/plugins/inputs/kibana/README.md
+++ b/plugins/inputs/kibana/README.md
@@ -42,7 +42,8 @@ with following rules:
 
 - kibana
   - status_code: integer (1, 2, 3, 0)
-  - heap_max_bytes/heap_total_bytes: integer
+  - heap_total_bytes: integer
+  - heap_max_bytes: integer
   - heap_used_bytes: integer
   - uptime_ms: integer
   - response_time_avg_ms: float
--- a/plugins/inputs/kibana/README.md
+++ b/plugins/inputs/kibana/README.md
@@ -42,7 +42,7 @@ with following rules:
 
 - kibana
   - status_code: integer (1, 2, 3, 0)
-  - heap_max_bytes: integer
+  - heap_max_bytes/heap_total_bytes: integer
   - heap_used_bytes: integer
   - uptime_ms: integer
   - response_time_avg_ms: float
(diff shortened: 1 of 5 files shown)


vikkyomkar:
this PR fixes #6894 and #5744

vikkyomkar:
@danielnelson can we get some eyes on this PR ?

lpic10:
With these changes the plugin seems to work up to version 7.5 (there are no changes to these metrics even in master branch: https://github.com/elastic/kibana/blob/master/src/legacy/server/status/lib/metrics.js)

@vikkyomkar The change on kibana that needs the fix happened on version 6.4, not 6.5, so it would be better to change the PR to reflect that. 

Also, there is a metric name change, but the underlying metric seems to be the same:
https://github.com/elastic/kibana/blob/cb16be83a47c5bc8a5214c2461dddb8c2299f6f5/src/server/status/metrics.js#L32
https://github.com/elastic/kibana/blob/564b25f2ab7b33545e38980443c24162f6ad6d4e/src/server/status/lib/metrics.js#L74

I don't know if it is a good practice to change the metric name on telegraf. OTOH I think the metric name in the updated kibana makes more sense.

@danielnelson what do you recommend?

plugins/inputs/kibana/kibana.go
```diff
@@ -187,15 +199,27 @@ func (k *Kibana) gatherKibanaStatus(baseUrl string, acc telegraf.Accumulator) er
...
-
-	fields["uptime_ms"] = kibanaStatus.Metrics.UptimeInMillis
 	fields["concurrent_connections"] = kibanaStatus.Metrics.ConcurrentConnections
-	fields["heap_max_bytes"] = kibanaStatus.Metrics.Process.Mem.HeapMaxInBytes
-	fields["heap_used_bytes"] = kibanaStatus.Metrics.Process.Mem.HeapUsedInBytes
 	fields["response_time_avg_ms"] = kibanaStatus.Metrics.ResponseTimes.AvgInMillis
 	fields["response_time_max_ms"] = kibanaStatus.Metrics.ResponseTimes.MaxInMillis
 	fields["requests_per_sec"] = float64(kibanaStatus.Metrics.Requests.Total) / float64(kibanaStatus.Metrics.CollectionIntervalInMilles) * 1000
 
+	Version, err := strconv.ParseFloat(strings.Join(strings.Split(kibanaStatus.Version.Number, ".")[:2], "."), 64)
```
danielnelson: This line could panic with unexpected version info, in particular when indexing `[:2]`.  Check the length of the split item to ensure it is long enough.  To match Go naming convention use a lowercase variable `version :=`.

plugins/inputs/kibana/kibana.go
```diff
@@ -187,15 +199,27 @@ func (k *Kibana) gatherKibanaStatus(baseUrl string, acc telegraf.Accumulator) er
...
+	if err != nil {
+		return err
+	}
+
+	if Version >= 6.5 {
+		fields["uptime_ms"] = kibanaStatus.Metrics.Process.UptimeInMillis
+		fields["heap_total_bytes"] = kibanaStatus.Metrics.Process.Memory.Heap.TotalInBytes
```
danielnelson: For backwards compatibility, either record this only as only `heap_max_bytes` or as both `heap_total_bytes` and `heap_max_bytes`.  Make sure to add a comment so we can remember why we did this.

plugins/inputs/kibana/README.md
```diff
@@ -52,12 +52,16 @@ with following rules:
...
 
 ### Example Output
+```
+Kibana Version < 6.5
+kibana,host=myhost,name=my-kibana,source=localhost:5601,status=green,version=6.3.2 concurrent_connections=0i,heap_max_bytes=136478720i,heap_used_bytes=119231088i,requests_per_sec=1,response_time_avg_ms=59i,response_time_max_ms=300i,status_code=1i,uptime_ms=2187428019i 1534864502000000000
 
-kibana,host=myhost,name=my-kibana,source=localhost:5601,version=6.3.2 concurrent_connections=0i,heap_max_bytes=136478720i,heap_used_bytes=119231088i,response_time_avg_ms=0i,response_time_max_ms=0i,status="green",status_code=1i,uptime_ms=2187428019i 1534864502000000000
+Kibana Version > 6.5
+kibana,host=myhost,name=my-kibana,source=localhost:5601,status=green,version=6.5.4 concurrent_connections=8i,heap_total_bytes=447778816i,heap_used_bytes=380603352i,requests_per_sec=1,response_time_avg_ms=57.6,response_time_max_ms=220i,status_code=1i,uptime_ms=6717489805i 1534864502000000000
+```
```
danielnelson: Only show the >6.5 version

plugins/inputs/kibana/kibana.go
vikkyomkar: code updated along with the comments as per the suggestions

plugins/inputs/kibana/kibana.go
vikkyomkar: @danielnelson code is updated to handle any panic situation

plugins/inputs/kibana/README.md
vikkyomkar: Plugin output is made generic for all the versions

PR #5370, opened 2019-02-03
Azure IoT Hub / Azure Event Hub input plugin

R290:
As discussed in issue #5231.
Feedback on code quality or anything else is greatly appreciated.

### Required for all PRs:

- [x] Signed [CLA](https://influxdata.com/community/cla/).
- [x] Associated README.md updated.
- [x] Has appropriate unit tests.

Issue #5231, opened 2019-01-04
Azure IoT Hub / Azure Event Hub input plugin 


R290:
### Proposal:

The proposal is to develop a plugin, which could be as simple as building a wrapper for [azure-event-hubs-go](https://github.com/Azure/azure-event-hubs-go). I currently have a proof of principle plugin conforming to the telegraf.ServiceInput interface. In the future I'm planning to add metric tracking and a configurable persistent receiver offset (so each message is received once).

### Current behavior:

No Azure IoT Hub or Azure Event Hub input plugin is currently present in Telegraf.
Development on the [AMQP 1.0 plugin](https://github.com/influxdata/telegraf/pull/4327) seems to have stagnated.

### Desired behavior:

Configurable plugin for Azure IoT Hub and Azure Event Hub.

### Use case: [Why is this important (helps with prioritizing requests)]

This plugin would allow telegraf and the rest of the TICK-stack to be used as backend for IoT applications and other Azure services. Azure itself does offer a similar service ([Time Series Insights](https://azure.microsoft.com/en-us/services/time-series-insights/)), but in my specific use case there is a strong preference for using the TICK-stack.

### Implementation:

I'd be willing to put in the time and effort to mature the proof of principle. Some feedback might be needed since I'm pretty to new to Go and programming a backend in general.

R290:
This is my Gopkg.lock after `dep ensure -vendor-only` (including #5259 and my changes):

[Gopkg.lock.txt](https://github.com/influxdata/telegraf/files/2735853/Gopkg.lock.txt)

As far as I can see the new dependencies are:
- github.com/Azure/azure-amqp-common-go 
- github.com/Azure/azure-event-hubs-go
- github.com/Azure/azure-sdk-for-go

`dep version` output:

> dep:
 version     : devel
 build date  :
 git hash    :
 go version  : go1.8.3
 go compiler : gc
 platform    : linux/amd64

R290:
PR is merged, so closing this issue. In the future a new issue might be opened for a generic AMQP 1.0 consumer with Azure IoT Hub / Azure Event Hubs configuration options.

PR #6928, opened 2020-01-20
Azure IoT Hub / Azure Event Hub input plugin (with go modules)

R290:
As discussed in issue #5231 and pull request #5370.
Feedback on code quality or anything else is greatly appreciated.

### Required for all PRs:


sjwang90:
Hi @R290, we will review this PR this week.

plugins/inputs/eventhub/eventhub.go
```diff
@@ -0,0 +1,382 @@
...
+  # max_undelivered_messages = 1000
+
+  ## Prefix to use for the system properties of Event Hub and IoT Hub messages
+  # system_properties_prefix = "SystemProperties."
+
+  ## Set either option below to true to use a system property as timestamp.
+  ## You have the choice between EnqueuedTime and IoTHubEnqueuedTime.
+  ## It is recommended to use this setting when the data itself has no timestamp.
+  # enq_time_ts = true
+  # iot_hub_enq_time_ts = true
```
danielnelson: Are both EnqueuedTime and IoTHubEnqueuedTime set with IoTHub?

plugins/inputs/eventhub/eventhub.go
```diff
@@ -0,0 +1,382 @@
+package eventhub
+
+// TODO: (optional) Test authentication with AAD TokenProvider environment variables?
```
danielnelson: If you can, that would be great, otherwise we can work through it when we have a user who needs this authentication method.

plugins/inputs/eventhub/eventhub.go
```diff
@@ -0,0 +1,382 @@
...
+// TODO: (optional) Event Processor Host, only applicable for multiple Telegraf instances?
```
danielnelson: This is an interesting idea, it seems like it would be a separate program above Telegraf, generating new telegraf.conf files and reloading Telegraf as needed.  Telegraf would communicate the offsets back to it.

We can think about this later of course, for now can you go ahead and remove any TODO messages in the pr.

plugins/inputs/eventhub/eventhub.go
```diff
@@ -0,0 +1,382 @@
...
+
+  ## By default the event hub receives all messages present on the broker.
+  ## Alternative modes can be set below. The timestamp should be in RFC3339 format.
+  ## The 3 options below only apply if no valid persister is read from memory or file (e.g. first run).
+  # from_timestamp = ""
+  # starting_offset = ""
+  # latest = true
+
+  ## Set a custom prefetch count for the receiver(s)
+  # prefetch_count = 1000
```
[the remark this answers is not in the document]
R290: https://github.com/Azure/azure-event-hubs-go/blob/7a1b2e2089e284bcec716b3bdc64ee2e5e626871/receiver.go#L46

plugins/inputs/eventhub/eventhub.go
R290: Yes they are, although one would expect that only `EnqueuedTime` is set when connecting to an Azure Event Hub. I should check if `EnqueuedTime` changes when the IoT Hub message is routed through one or more event hubs.

plugins/inputs/eventhub/eventhub.go
R290: Moved to README

plugins/inputs/eventhub/eventhub.go
R290: Moved to README

plugins/inputs/eventhub/eventhub.go
```diff
@@ -0,0 +1,382 @@
...
+	"github.com/influxdata/telegraf/plugins/inputs"
+	"github.com/influxdata/telegraf/plugins/parsers"
+
+	eventhub "github.com/Azure/azure-event-hubs-go/v3"
+	"github.com/Azure/azure-event-hubs-go/v3/persist"
+)
+
+const (
+	defaultMaxUndeliveredMessages = 1000
+	defaultUserAgent              = "telegraf"
```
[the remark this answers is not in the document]
R290: Changing here raises compiler error: `internal.ProductToken() (value of type string) is not constant`.
Dropped the `defaultUserAgent` and inserted `internal.ProductToken()` directly at the hub option.

plugins/inputs/eventhub/eventhub.go
```diff
@@ -0,0 +1,382 @@
...
+	for {
+		select {
+		case <-ctx.Done():
+			if len(e.tracker.messages) == 0 { // everything has been delivered
+				return
+			}
+		case DeliveryInfo := <-e.acc.Delivered():
+			log.Printf("D! [inputs.eventhub] tracking:: ID %d delivered: %t", DeliveryInfo.ID(), DeliveryInfo.Delivered())
+
+			if DeliveryInfo.Delivered() {
```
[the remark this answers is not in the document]
R290: I don't think consumption rate will be limited when an in memory persister is used? It should be possible to add a third type of persister to azure-event-hubs-go which is something like a batched file persister (write every N events instead of every single event). If I find the time I can create an issue. 

I think your other points on out of order processing and concurrency are perfectly valid, but do we need to include this functionality in the plugin? Or is expanding the azure-event-hubs-go repository the preferred way to do this? The latter option might be more beneficial to the broad public.

plugins/inputs/eventhub/eventhub.go
```diff
@@ -0,0 +1,382 @@
...
+
+	if err != nil {
+		log.Printf("E! [inputs.eventhub] error %s", err)
+		return err
+	}
+
+	log.Printf("D! [inputs.eventhub] %d metrics found after parsing", len(metrics))
+
+	for i := range metrics {
+		metrics[i].AddField(e.SystemPropertiesPrefix+"SequenceNumber", *event.SystemProperties.SequenceNumber)
```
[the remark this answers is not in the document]
R290: See e29fbda804656539e513aed9aeedd495dd578405

plugins/inputs/eventhub/eventhub.go
danielnelson: What do you think about removing these options and using iothub time if available, and if not using eventhub time?

plugins/inputs/eventhub/eventhub.go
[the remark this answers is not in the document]
danielnelson: There are a few things we generally want to be able to do in a plugin like this that consumes from a message queue:
- Take at least enough metrics to trigger an immediate flush in the output plugin, by filling the metric batch.
- Wait for these metrics to be written by the output plugin before taking more messages, we don't want to consume to far ahead.
- Depending on the queue, acknowledge or update offsets when a message is successfully written, and reject messages that are lost.  In Telegraf messages are lost if the metric buffer fills.

I looked into the azure-event-hubs-go code a bit, and it appears that the handler is called in the same goroutine for each message.  Once the handler returns, the message is immediately marked as acknowledged if there is no error and the offsets are persisted. If the handler returns an error the message is marked as delivery failed.

This means there is no way to take multiple messages and report back to Event Hub later on if they were processed successfully or not.

The only way it can work is for us to report that the message are handled successfully and hope it can be written by the output.

Since you are re-adding metrics if they weren't able to be delivered, the only way we could still lose metrics is on shutdown/restart/crash and only if the output is down.  It isn't a full durability guarantee but I think it is the best we can do with this library.

The other thing we need to do is to stop consuming messages once we have `max_undelivered_messages` in flight.  Otherwise we will just keep grabbing new messages even though nothing is being written.

plugins/inputs/eventhub/eventhub.go
danielnelson: Nevermind this suggestion, I see now that it won't work well since we need at least one option turn this on/off.

go.mod
```diff
@@ -137,3 +137,9 @@ require (
 	gotest.tools v2.2.0+incompatible // indirect
 	k8s.io/apimachinery v0.17.1 // indirect
 )
+
+// replaced due to https://github.com/satori/go.uuid/issues/73
+replace github.com/satori/go.uuid => github.com/gofrs/uuid v3.2.0+incompatible
+
+// Set the Azure Event Hub log abstraction to the Telegraf format:
+replace github.com/devigned/tab => github.com/R290/tab v0.1.2
```
danielnelson: I'd like to skip this change, generally I don't like to allow libraries to access the Logger, in the case we did want to log the library we would hook it all up to debug level messages.  However, if we want to do that instead of forking the library we should create a custom tab.Tracer, tab.Spanner, tab.Logger, and then call tab.Register(tracer).

plugins/inputs/eventhub/eventhub.go
```diff
@@ -0,0 +1,381 @@
...
+				return
+			}
+		case DeliveryInfo := <-e.acc.Delivered():
+			if DeliveryInfo.Delivered() {
+				e.tracker.mux.Lock()
+				delete(e.tracker.messages, DeliveryInfo.ID())
+				e.tracker.mux.Unlock()
+			} else {
+				e.tracker.mux.Lock()
+				id := e.acc.AddTrackingMetricGroup(e.tracker.messages[DeliveryInfo.ID()])
```
danielnelson: In order to safely re-add the metrics we would need to save a copy of them.  This way if the metric is modified downstream we can re-add the original metric here.

plugins/inputs/eventhub/README.md
```diff
@@ -0,0 +1,99 @@
+# Azure Event Hubs input plugin
+
+This plugin provides a consumer for use with Azure Event Hubs and Azure IoT Hub. The implementation is in essence a wrapper for [Microsoft Azure Event Hubs Client for Golang](https://github.com/Azure/azure-event-hubs-go).
+
+## Configuration
+
+```toml
+[[inputs.eventhub]]
```
danielnelson: I'd like to rename this plugin `eventhub_consumer` to match the other queuing plugins.

danielnelson:
I opened the PR against your fork, https://github.com/R290/telegraf/pull/1, which should take care of the metric handling issues.  There are a few more things on top of that PR that we need to take care of:
1. I noticed during testing that if you are consuming more than one partition, it isn't safe to use the same offset to restart the consumer.  This is because one partition could have a new offset that would be invalid on the other partition.  I believe for the `starting_offset` option we will need to either allow it to be set per offset.  Perhaps we should drop this option for now?
2. Still need to add some tests.
3. Rename to `eventhub_consumer`, we can do this after merge though.

plugins/inputs/eventhub/eventhub.go
R290: Indeed, there could be timestamps within the payload which can be parsed by for example the json parser.

go.mod
R290: Merged the pull request for removing the log abstraction replacement.

plugins/inputs/eventhub/eventhub.go
R290: Merged pull request

R290:
Thank you for your pull request! 

1. You are right and the from starting offset option is dropped
2. What could we test without actually connecting to an eventhub? In an earlier comment you mentioned testing the `Init()` function, how would one go about writing tests for this? (I'm still very green when it comes to writing tests)
3. Fine by me

At some point we could consider dropping the `Azure/azure-event-hubs-go` library altogether and creating a generic AMQP 1.0 consumer based on `Azure/go-amqp`. Some elements from `Azure/azure-amqp-common-go` could be needed for the Event Hub and IoT Hub specific stuff. I think this would allow for a plugin closer to the Telegraf design philosophy?

plugins/inputs/eventhub/README.md
R290: After merge

plugins/inputs/eventhub/eventhub.go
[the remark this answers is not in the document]
R290: See the bottom of https://github.com/influxdata/telegraf/pull/6928#issuecomment-599225047.

danielnelson:
I rebased your branch since I couldn't figure out why GitHub still reported a conflict.

> At some point we could consider dropping the Azure/azure-event-hubs-go library altogether and creating a generic AMQP 1.0 consumer based on Azure/go-amqp.

I'd definitely like to do have a generic amqp1 input.  I'm imagining we would have an eventhub_consumer + amqp1_consumer inputs and an eventhub + amqp1 output.  I'm not sure if we could get away with eventually having only the amqp1 plugins though, we might need the eventhub plugin for some of the authentication options?

PR #6940, opened 2020-01-25
plugins/filestat: Create option to disable logging of missing files

nroach44:
Create configuration option NoLogFileMissing to prevent log spam when
the file is missing. For example, when using telegraf to check for
/var/run/reboot-needed (which is only present when the reboot is needed)
the agent will log the following each time the file is checked, and missing:

2020-01-25T15:23:20Z E! [inputs.filestat] Unable to get info for file "/var/run/reboot-required", possible permissions issue

This will allow us to disable this log alert.

### Required for all PRs:

### DIFF
--- a/plugins/inputs/filestat/README.md
+++ b/plugins/inputs/filestat/README.md
@@ -14,6 +14,9 @@ The filestat plugin gathers metrics about file existence, size, and other stats.
 
   ## If true, read the entire file and calculate an md5 checksum.
   md5 = false
+
+  ## If true, don't log an error if the file is missing.
+  NoLogFileMissing = false
 ```
 
 ### Measurements & Fields:
--- a/plugins/inputs/filestat/filestat.go
+++ b/plugins/inputs/filestat/filestat.go
@@ -25,11 +25,15 @@ const sampleConfig = `
 
   ## If true, read the entire file and calculate an md5 checksum.
   md5 = false
+  
+  ## If true, don't log an error if the file is missing.
+  NoLogFileMissing = false
 `
 
 type FileStat struct {
-	Md5   bool
-	Files []string
+	Md5              bool
+	NoLogFileMissing bool
+	Files            []string
 
 	Log telegraf.Logger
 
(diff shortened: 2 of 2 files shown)


srebhan (changes requested):
I have two comments.
1. The option you added is _CamelCase_ while options usually are in _snake_case_. Please also change to _snake_case_.
1. IMO you should output the information _once_ to prevent unexpected behavior when specifying the option, the file is there but there is indeed a permission issue.

Speaking of this, maybe you could make this the new behavior without any option. If there is a problem stat-ing the file, output the error once and set a flag. Then do not output as long as the flag is set. Once you successfully accessed the file, unset the flag and next time you fail output once again a.s.o...

srebhan:
@nroach44 any news? Are you still willing to carry on with this PR?

nroach44:
@srebhan I can adjust the option casing but I don't have the go experience to perform the "output only once" change

srebhan:
@nroach44 would appreciate the casing-changes. If you like we can work on the "output only once" together!?
The idea is that additional to adding `NoLogFileMissing` you also add a map (like a hashmap or maybe you know dict in python) that keeps the warning state for each file. That is add something like `warnedFiles map[string]bool` to the `FileStat` struct. Then in `NewFileStat()`  you need to init that map by `warnedFiles: make(map[string]bool)`. Now you do have the data-structure for doing the book-keeping. When you now warn you check if the file is already in `warnedFiles` and if so if the state is true e.g. `if warned, ok := f.warnedFiles[fileName]; !ok || !warned { ... }` then after warning you set the warning flag `f.warnedFiles[fileName] = true`. You also might want to reset that flag for files that were successfully queries to warn again...

srebhan:
@nroach44 any chance you can adopt my suggestion? We can work this out together if you like!?

srebhan:
@nroach44 any news on this one? Could you please at least fix the CamelCase issue!?

sjwang90:
Should we close this PR since we have this ready to be merged? https://github.com/influxdata/telegraf/pull/7316

srebhan:
I say yes on closing this. The PR you referenced does the same, is reviewed and doesn't add a config option.

PR #6952, opened 2020-01-29
Wildcard support for x509_cert files

### LINKED ISSUE #6951: wildcard (glob) support for files sources in plugins/inputs/x509_cert (PR included)
## Feature Request

wildcard (glob) support for files sources in plugins/inputs/x509_cert

### Proposal:

Support usage of '*' in config file:
```
[[inputs.x509_cert]]
  ## List certificate sources, support wildcard expands for files
  sources = [ "/etc/mycerts/sub*.mydomain.org.pem"]
```

### Current behavior:
That just doesn't work

### Desired behavior:
expand the files list relevant to the wildcard

### Use case:
Not having to write down a hundred of certificate

jaroug:
### Required for all PRs:

- [x] Has appropriate unit tests. (as much as there is no existing ones)

Closes #6951

### DIFF
--- a/plugins/inputs/x509_cert/x509_cert.go
+++ b/plugins/inputs/x509_cert/x509_cert.go
@@ -16,15 +16,15 @@ import (
 
 	"github.com/influxdata/telegraf"
 	"github.com/influxdata/telegraf/internal"
-	_tls "github.com/influxdata/telegraf/plugins/common/tls"
 	"github.com/influxdata/telegraf/internal/globpath"
+	_tls "github.com/influxdata/telegraf/plugins/common/tls"
 	"github.com/influxdata/telegraf/plugins/inputs"
 )
 
 const sampleConfig = `
   ## List certificate sources
   sources = ["/etc/ssl/certs/ssl-cert-snakeoil.pem", "tcp://example.org:443",
-            "/etc/mycerts/*.mydomain.org.pem"]
+            "/etc/mycerts/*.mydomain.org.pem", "file:///path/to/*.pem"]
 
   ## Timeout for SSL connection
   # timeout = "5s"
@@ -60,17 +60,21 @@ func (c *X509Cert) SampleConfig() string {
 	return sampleConfig
 }
 
-func (c *X509Cert) locationToURL(location string) (*url.URL, error) {
-	if strings.HasPrefix(location, "/") {
-		location = "file://" + location
-	}
-	if strings.Index(location, ":\\") == 1 {
-		location = "file://" + filepath.ToSlash(location)
-	}
+func (c *X509Cert) sourcesToURLs() error {
+	for _, source := range c.Sources {
+		if strings.HasPrefix(source, "/") {
+			source = "file://" + source
+		}
+		if strings.Index(source, ":\\") == 1 {
+			source = "file://" + filepath.ToSlash(source)
+		}
 
-	u, err := url.Parse(location)
-	if err != nil {
-		return nil, fmt.Errorf("failed to parse cert location - %s", err.Error())
+		u, err := url.Parse(source)
+		if err != nil {
+			return fmt.Errorf("failed to parse cert location - %s", err.Error())
+		}
+
+		c.urls = append(c.urls, u)
 	}
 
 	return nil
@@ -207,12 +211,16 @@ func getTags(cert *x509.Certificate, location string) map[string]string {
 	return tags
 }
 
-// copied from plugins/inputs/file/file.go
-func (c *X509Cert) refreshFilePaths() error {
+// copied from refreshFilePaths() in plugins/inputs/file/file.go
+func (c *X509Cert) expandFilePaths() error {
 	var allFiles []string
 
 	for _, source := range c.Sources {
-		if strings.HasPrefix(source, "/") {
+		if strings.HasPrefix(source, "file://") ||
+			strings.HasPrefix(source, "/") {
+			if strings.HasPrefix(source, "file://") {
+				source = strings.TrimPrefix(source, "file://")
+			}
 			g, err := globpath.Compile(source)
 			if err != nil {
 				return fmt.Errorf("could not compile glob %v: %v", source, err)
@@ -235,14 +243,8 @@ func (c *X509Cert) refreshFilePaths() error {
 func (c *X509Cert) Gather(acc telegraf.Accumulator) error {
 	now := time.Now()
 
-	for _, location := range c.Sources {
-		u, err := c.locationToURL(location)
-		if err != nil {
-			acc.AddError(err)
-			return nil
-		}
-
-		certs, err := c.getCert(u, c.Timeout.Duration*time.Second)
+	for _, url := range c.urls {
+		certs, err := c.getCert(url, c.Timeout.Duration*time.Second)
 		if err != nil {
 			acc.AddError(fmt.Errorf("cannot get SSL cert '%s': %s", url, err.Error()))
 		}
@@ -258,7 +260,7 @@ func (c *X509Cert) Gather(acc telegraf.Accumulator) error {
 				KeyUsages:     []x509.ExtKeyUsage{x509.ExtKeyUsageAny},
 			}
 			if i == 0 {
-				opts.DNSName, err = c.serverName(u)
+				opts.DNSName, err = c.serverName(url)
 				if err != nil {
 					return err
 				}
@@ -290,12 +292,12 @@ func (c *X509Cert) Gather(acc telegraf.Accumulator) error {
 }
 
 func (c *X509Cert) Init() error {
-	err := c.refreshFilePaths()
+	err := c.expandFilePaths()
 	if err != nil {
 		return err
 	}
 
-	err = c.locationToURL()
+	err = c.sourcesToURLs()
 	if err != nil {
 		return err
 	}
--- a/plugins/inputs/x509_cert/x509_cert.go
+++ b/plugins/inputs/x509_cert/x509_cert.go
@@ -23,6 +23,7 @@ import (
 
 const sampleConfig = `
   ## List certificate sources
+  ## Prefix your entry with 'file://' if you intend to use relative paths
   sources = ["/etc/ssl/certs/ssl-cert-snakeoil.pem", "tcp://example.org:443",
             "/etc/mycerts/*.mydomain.org.pem", "file:///path/to/*.pem"]
 
@@ -50,6 +51,7 @@ type X509Cert struct {
 	urls                []*url.URL
 	globFilePathsToUrls []*url.URL
 	globpaths           []*globpath.GlobPath
+	Log                 telegraf.Logger
 }
 
 // Description returns description of the plugin.
@@ -64,16 +66,17 @@ func (c *X509Cert) SampleConfig() string {
 
 func (c *X509Cert) sourcesToURLs() error {
 	for _, source := range c.Sources {
-		if strings.HasPrefix(source, "/") {
+		if strings.HasPrefix(source, "file://") ||
+			strings.HasPrefix(source, "/") ||
+			strings.Index(source, ":\\") == 1 {
+
+			source = filepath.ToSlash(strings.TrimPrefix(source, "file://"))
 			g, err := globpath.Compile(source)
 			if err != nil {
 				return fmt.Errorf("could not compile glob %v: %v", source, err)
 			}
 			c.globpaths = append(c.globpaths, g)
 		} else {
-			if strings.Index(source, ":\\") == 1 {
-				source = "file://" + filepath.ToSlash(source)
-			}
 
 			u, err := url.Parse(source)
 			if err != nil {
@@ -225,7 +228,8 @@ func (c *X509Cert) expandFilePathsToUrls() error {
 	for _, globpath := range c.globpaths {
 		files := globpath.Match()
 		if len(files) <= 0 {
-			return fmt.Errorf("could not find file: %v", globpath)
+			c.Log.Errorf("could not find file: %v", globpath)
+			continue
 		}
 		for _, file := range files {
 			file = "file://" + file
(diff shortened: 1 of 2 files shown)


jaroug:
related to #6951

plugins/inputs/x509_cert/x509_cert.go
```diff
@@ -187,10 +188,39 @@ func getTags(cert *x509.Certificate, location string) map[string]string {
 	return tags
 }
 
+// copied from plugins/inputs/file/file.go
+func (c *X509Cert) refreshFilePaths() error {
```
danielnelson: Let's add this logic to an `func (c *X509Cert) Init() error` function, it will only be ran once on plugin setup.  We can setup the globpath once, and then in `Gather()` we will call `g.Match()` each time.

plugins/inputs/x509_cert/x509_cert.go
```diff
@@ -187,10 +188,39 @@ func getTags(cert *x509.Certificate, location string) map[string]string {
...
+			if err != nil {
+				return fmt.Errorf("could not compile glob %v: %v", source, err)
+			}
+			files := g.Match()
+			if len(files) <= 0 {
+				return fmt.Errorf("could not find file: %v", source)
+			}
+			allFiles = append(allFiles, files...)
+		} else {
+			allFiles = append(allFiles, source)
```
danielnelson: Why do we skip glob compilation if the path isn't absolute?  It seems that we can just pass any string into globpath.Compile.

jaroug: Well that seems to be the original logic of this plugin, lines 60 to 63:
```func (c *X509Cert) locationToURL(location string) (*url.URL, error) {
	if strings.HasPrefix(location, "/") {
		location = "file://" + location
	}

Comment (danielnelson) [reply plugins/inputs/x509_cert/x509_cert.go]: Ahh, strings not starting with `/` are URLs, I forgot we are handling both here.

Would you be able to move the matching into Init along with the URL parsing?  We could have separate fields on the struct for each type of path, file paths as a []GlobPath and URLs as a []*url.URL.  Then during gather we can call Match() on the globs and use the http URLs as they are.

File: plugins/inputs/x509_cert/x509_cert.go (hunk 1 above)
Comment (jaroug) [reply plugins/inputs/x509_cert/x509_cert.go:196]: Hi, sorry I've been busy.

I've tried to match your request with 19441a29738b85f7bbd9f9e741636b6aba938c78

File: plugins/inputs/x509_cert/x509_cert.go (hunk 2 above)
Comment (jaroug) [reply plugins/inputs/x509_cert/x509_cert.go:211]: We could get rid of those `strings.HasPrefix` by forcing users to use `file://` url format in configuration, but that would break configuration's backward compatibility, I don't know if this is acceptable.
Also matching a leading `/` will probably not work for windows users right?

**Comment by ikkemaniac** [comment]
Would this also allow to add a complete dir?
```
["/etc/certs/*"]
```

**Comment by jaroug** [comment]
> Would this also allow to add a complete dir?
> 
> ```
> ["/etc/certs/*"]
> ```

Yes :)

**Comment by ikkemaniac** [comment]
> > Would this also allow to add a complete dir?
> > ```
> > ["/etc/certs/*"]
> > ```
> 
> Yes :)

AWEsome, let's hope this get's merged quickly.

I should be able to build it from 9309072 right?

**Comment by jaroug** [comment]
> > > Would this also allow to add a complete dir?
> > > ```
> > > ["/etc/certs/*"]
> > > ```
> > 
> > 
> > Yes :)
> 
> AWEsome, let's hope this get's merged quickly.
> 
> I should be able to build it from [9309072](https://github.com/influxdata/telegraf/commit/930907296fe0116ba6f36c0833578d9cbfc0ac7c) right?

I just rebased master, resolved the merge conflict, it works :)

**Comment by ikkemaniac** [comment]
Took a little bit to x-compile but i got r running. No strange things yet so thumbs up from me!

Thx bud!

**Comment by MatthewGow** [comment]
Wow, this is exactly what we were looking to do - great work @jaroug !

@danielnelson - is there anything holding up the review of this other than time?

**Comment by VorobevPavel-dev** [comment]
It’s sad that it’s been all year since PR was created. Maybe there is something new or you just forgot about it? Very useful functions and features cannot be added for a whole year. Please, take a look, @danielnelson, @jaroug

File: plugins/inputs/x509_cert/x509_cert.go [hunk 3]
```diff
@@ -187,25 +192,43 @@ func getTags(cert *x509.Certificate, location string) map[string]string {
 	return tags
 }
 
+// copied from plugins/inputs/file/file.go
+func (c *X509Cert) refreshFilePaths() error {
+	var allFiles []string
+
+	for _, source := range c.Sources {
+		if strings.HasPrefix(source, "/") {
```
srebhan, line 200: I think you should swap the order of processing and _first_ call `sourcesToURLs()` and afterwards this function. This way you can be sure that _file_ always have `file://` prepended. Otherwise you might miss files where the user added `file://` already for you.
```suggestion
		if strings.HasPrefix(source, "file://") {
```

plugins/inputs/x509_cert/x509_cert.go
```diff
@@ -187,25 +192,43 @@ func getTags(cert *x509.Certificate, location string) map[string]string {
...
+			g, err := globpath.Compile(source)
+			if err != nil {
+				return fmt.Errorf("could not compile glob %v: %v", source, err)
+			}
+			files := g.Match()
+			if len(files) <= 0 {
+				return fmt.Errorf("could not find file: %v", source)
+			}
+			allFiles = append(allFiles, files...)
```
srebhan, line 209: Please remember to prepend `file://` again here after swapping functions.

plugins/inputs/x509_cert/x509_cert.go
```diff
@@ -247,6 +270,16 @@ func (c *X509Cert) Gather(acc telegraf.Accumulator) error {
...
+	err := c.refreshFilePaths()
+	if err != nil {
+		return err
+	}
+
+	err = c.sourcesToURLs()
+	if err != nil {
+		return err
+	}
+
```
srebhan, line 282: I think it's better to swap these functions so `refreshFilePaths()` (which might be called `expandFilePaths()`)  can work on normalized inputs as otherwise you miss entries where the user specified files in the form `file:///etc/certs/*`.

srebhan (changes requested):
I suggest swapping the two functions in `Init()` and work on normalized URLs instead of relying on the user not specifying `file://` prefixed with wildcards. See my comments in the code.

ssoroka (approved):
looks like it needs some merge conflicts resolved. Any additional test coverage would be great, but not a blocker.

sjwang90:
@jaroug If you can address the comments from @srebhan and @ssoroka we can do another review and look to get this into our next feature release. Thanks!

jaroug:
@sjwang90 gonna have a look asap

plugins/inputs/x509_cert/x509_cert.go
jaroug: the globing tools don't work with url styled path, so I'll have to strip the leading file://

srebhan (commented):
Looks good so far. Only one more question: You now moved the globbing to `Init()` which means you won't catch new certs and maybe will poke old (already deleted) certs. Is this by intention? Maybe it would make sense to take the extra cycles and update the cert-list in every gather-cycle?!?

jaroug:
@srebhan it was requested by the previous person holding the PR: https://github.com/influxdata/telegraf/pull/6952#discussion_r376097920

srebhan:
@jaroug almost, if you read the comment you cited carefully the idea was to _prepare_ the glob-statement in `Init()` but call the `Match()` method in _each_ gather cycle. Do you think this is doable to prepare the glob statement in `Init()` then call the `Match()` for the glob entries in `Gather()` and join the result with the static entries? This would mean to split the `refreshFilePaths()` function into two parts...

jaroug:
@srebhan oh yeah ok, I totally misunderstood, my bad, I'll have a looks asap

sjwang90:
Hi @jaroug, just checking on the status. We're hoping for a Telegraf 1.18 release candidate this week and would love to get this in.

jaroug:
:wave: I made those changes a bit in a hurry so it might be far from perfect but it seems to work. I haven't considered searching for globpath on "file://" inputs since I'm not sure it's supposed to be supported by the URL package, should I?

plugins/inputs/x509_cert/x509_cert.go
```diff
@@ -57,20 +62,29 @@ func (c *X509Cert) SampleConfig() string {
...
+		if strings.HasPrefix(source, "/") {
+			g, err := globpath.Compile(source)
+			if err != nil {
+				return fmt.Errorf("could not compile glob %v: %v", source, err)
+			}
+			c.globpaths = append(c.globpaths, g)
+		} else {
+			if strings.Index(source, ":\\") == 1 {
+				source = "file://" + filepath.ToSlash(source)
+			}
```
srebhan, line 76: I think this will prevent globs prefixed with `file://` from being compiled. How about
```suggestion
		if strings.HasPrefix(source, "file://") || strings.HasPrefix(source, "/") || strings.Index(source, ":\\") == 1 {
			// We have a file path, so glob it
			path := filepath.ToSlash(string.TrimPrefix(source, "file://"))
			g, err := globpath.Compile(path)
			if err != nil {
				return fmt.Errorf("could not compile glob %v: %v", source, err)
			}
			c.globpaths = append(c.globpaths, g)
		} else {
			// We don't have a file path, so simply store it as a URL
```
This leaves the necessity to always prefix relative paths with `file://` as otherwise we cannot distinguish them from other URLs.

plugins/inputs/x509_cert/x509_cert.go
```diff
@@ -204,25 +218,47 @@ func getTags(cert *x509.Certificate, location string) map[string]string {
...
 
+// copied from refreshFilePaths() in plugins/inputs/file/file.go
+func (c *X509Cert) expandFilePathsToUrls() error {
+	var urls []*url.URL
+
+	for _, globpath := range c.globpaths {
+		files := globpath.Match()
+		if len(files) <= 0 {
+			return fmt.Errorf("could not find file: %v", globpath)
+		}
```
srebhan, line 229: Isn't this a bit harsh? Do we always expect a matching file for the glob paths? In not we could just `continue` here and maybe issue a debug-log-entry.

plugins/inputs/x509_cert/x509_cert.go
jaroug: As I asked in my previous comment should we consider parsing "files://" since it's not supported by RFC 3986?
If yes, then we should also search for prefix "file://../" and "../" (that I forgot)
Also I dont have windows system on my hands, but I think there is no globbing support on windows system right?

plugins/inputs/x509_cert/x509_cert.go
jaroug: you are right totally not suitable

plugins/inputs/x509_cert/x509_cert.go
srebhan: I guess Windows has support for globbing, but not sure... @ssoroka do you know if globbing works on Windows? Regarding the `file://` handling, I think we should support globbing there as you even have this as an example for the config option. :-)
Why do we need to check for `file://../`? It's covered already because we enter the "file" branch when prefixed with `file://`. For `../` it's a different story. As I said you _always_ need to prefix relative paths with `file://` otherwise you run into the trouble of checking `./something`, `somepath/` etc.

plugins/inputs/x509_cert/x509_cert.go
```diff
@@ -204,25 +221,48 @@ func getTags(cert *x509.Certificate, location string) map[string]string {
 	return tags
 }
 
+// copied from refreshFilePaths() in plugins/inputs/file/file.go
+func (c *X509Cert) expandFilePathsToUrls() error {
```
srebhan, line 225: Can we change this function to return the overall list of URLs instead of passing it by the structure? While you are at it you can do the `append(c.urls, c.globFilePathsToUrls...)` here by initializing the overall list with `c.urls`...

srebhan (commented):
Hey just two more points/ideas. If we change `expandFilePathsToUrls` to something like `collectCertURLs` we can save one more struct field. Furthermore, please update the `README.md` file to match the config-option comments. :-)

srebhan (approved):
Looks good to me.

ikkemaniac:
*sees 'ready to merge' tag

is this really happening ? :crossed_fingers:

jaroug:
> *sees 'ready to merge' tag
> 
> is this really happening ?

Well hopefully, poke @sjwang90

srebhan:
@jaroug and @ikkemaniac please stay calm and let the Influx guys do their job. It might take some more time due to the amount of PRs they have to handle and the way PRs are merged during the release cycle. So poking at this every second day takes away review resources (e.g. mine ;-))...

plugins/inputs/x509_cert/x509_cert.go
```diff
@@ -204,25 +220,45 @@ func getTags(cert *x509.Certificate, location string) map[string]string {
...
+
+	for _, globpath := range c.globpaths {
+		files := globpath.Match()
+		if len(files) <= 0 {
+			c.Log.Errorf("could not find file: %v", globpath)
+			continue
+		}
+		for _, file := range files {
+			file = "file://" + file
+			u, err := url.Parse(file)
```
Jagularr, line 234: Looks like this causes windows tests to fail, as it assumes the path is a linux-style path. On windows the path will start with `C:` or something and it will recognize the colon as a port. A workaround for windows is adding an extra slash to the `file://` part but it looks like there might be other issues (see: https://github.com/golang/go/issues/13276). I'm thinking maybe we'd just need to manually handle the windows case here, not 100% sure but we will need it to work on windows.

reimda:
Hi @jaroug, are you able/willing to fix this windows path problem? If not, let us know and we'll try to find someone who can.  This PR is useful and I'd hate to see it stagnate again after getting so close to being merged!  Thanks!

aslgithub:
This Pull seems to be the only change I can obviously find between v1.18.1 and v1.19.0 for x509_cert and I'm concerned it may have broken URL testing on Windows ?
https://github.com/influxdata/telegraf/issues/9384

PR #6960, opened 2020-01-31
Fix duplicate TrackingIDs returned

asgaut:
There is a small chance the newTrackingID() function in tracking.go
will return the same id to multiple simultaneous callers.
The function must return the value returned by atomic.AddUint64()
to be safe.

### Required for all PRs:

### DIFF
--- a/metric/tracking.go
+++ b/metric/tracking.go
@@ -34,8 +34,7 @@ var (
 )
 
 func newTrackingID() telegraf.TrackingID {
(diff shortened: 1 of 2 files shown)


asgaut:
I see the same bug exists in the newTrackingID() function in accumulator.go.

danielnelson:
Nice catch, I'll get the other instance of the bug and also do a quick audit over our other uses of atomic.

PR #6964, opened 2020-01-31
Fix atomic usage in tests

danielnelson:
related: #6960

### Required for all PRs:

### DIFF
--- a/internal/models/running_aggregator_test.go
+++ b/internal/models/running_aggregator_test.go
@@ -1,7 +1,6 @@
 package models
 
 import (
-	"sync/atomic"
 	"testing"
 	"time"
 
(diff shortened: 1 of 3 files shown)


internal/models/running_aggregator_test.go
```diff
@@ -246,7 +246,7 @@ type TestAggregator struct {
 func (t *TestAggregator) Description() string  { return "" }
 func (t *TestAggregator) SampleConfig() string { return "" }
 func (t *TestAggregator) Reset() {
-	atomic.StoreInt64(&t.sum, 0)
+	t.sum := 0
```
danielnelson: Atomic not be required, the RunningAggregator should prevent concurrent calls.

plugins/inputs/vsphere/vsphere_test.go
```diff
@@ -230,36 +227,6 @@ func TestParseConfig(t *testing.T) {
 	require.NotNil(t, tab)
 }
 
-func TestThrottledExecutor(t *testing.T) {
```
danielnelson: This test is racy and somewhat slow to run, I don't think it provides enough value to keep.

PR #6969, opened 2020-02-01
add functionality to scrape pods per kubernetes node

### LINKED ISSUE #6950: Prometheus scrape metrics from pods on same node
## Feature Request

Currently, when we deploy telegraf as a daemonset in kubernetes, each telegraf pod scrapes metrics from all target pods on all nodes. 
Issue/s arising with this:
1) it creates duplicate metrics 

### Proposal:
Create a config parameter called 'kubernetes_node', the existence of which will trigger a filter based on the pods Spec.NodeName. If the pods Spec.NodeName matched the config parameter, then registerPod for scraping else disregard the pod.

### Current behavior:
Telegraf pod scrapes all pods and creates duplicate metrics if more than one telegraf pod id deployed.
### Desired behavior:
When the proposed/new config parameter (kubernetes_node) is provided, Telegraf pod should only concern itself with scraping metrics from pods on the same node that it is situated on.
I have this working and can submit a PR, let me know.

### Use case: 
Currently, when we deploy telegraf as a daemonset in kubernetes, each telegraf pod scrapes metrics from all target pods on all nodes. If we can go ahead with this proposal then we can de-duplicate metrics at source and reduce burden on infrastructure.

mg03:
### Required for all PRs:

- [ y] Signed [CLA](https://influxdata.com/community/cla/).
- [ y] Associated README.md updated.
- [ y] Has appropriate unit tests.

Fixes #6950

### DIFF
--- a/plugins/inputs/prometheus/kubernetes.go
+++ b/plugins/inputs/prometheus/kubernetes.go
@@ -15,6 +15,8 @@ import (
 	"github.com/ericchiang/k8s"
 	corev1 "github.com/ericchiang/k8s/apis/core/v1"
 	"github.com/ghodss/yaml"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/fields"
 )
 
 type payload struct {
@@ -82,8 +84,20 @@ func (p *Prometheus) start(ctx context.Context) error {
 // pod, causing errors in the logs. This is only true if the pod going offline is not
 // directed to do so by K8s.
 func (p *Prometheus) watch(ctx context.Context, client *k8s.Client) error {
+
+	selectors, err := podSelector(p)
+
+	if err != nil {
+		return err
+	}
+
+	//filter := k8s.QueryParam("fieldSelector", "spec.nodeName="+p.KubernetesNode)
+	//l := new(k8s.LabelSelector)
+	//l.Eq("blah", "booyah")
+	//options = append(options, filter)
+	//options = append(options, k8s.QueryParam("labelSelector", "app=api"))
 	pod := &corev1.Pod{}
-	watcher, err := client.Watch(ctx, p.PodNamespace, &corev1.Pod{})
+	watcher, err := client.Watch(ctx, p.PodNamespace, &corev1.Pod{}, selectors...)
 	if err != nil {
 		return err
 	}
@@ -103,7 +117,7 @@ func (p *Prometheus) watch(ctx context.Context, client *k8s.Client) error {
 
 			// If the pod is not "ready", there will be no ip associated with it.
 			if pod.GetMetadata().GetAnnotations()["prometheus.io/scrape"] != "true" ||
-				!podReady(pod.Status.GetContainerStatuses()) || !podOnNode(pod, p.KubernetesNode) {
+				!podReady(pod.Status.GetContainerStatuses()) {
 				continue
 			}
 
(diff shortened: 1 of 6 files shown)


mg03:
@danielnelson : please review my PR, thank you

mg03:
@jackzampolin : Hi, Im tagging you since I see you are the primary moderator on community.influxdata.com and I thought you might have some suggestion/s . Would you know who could possibly help with PR review ?

mg03:
@danielnelson : hi would it be possible for you to take a look at this PR and let me know your thoughts on the same.

danielnelson:
Code looks good but I do feel that this method is very specific to your case and that we should try to generalize it more.

I'm not a Kubernetes expert, but perhaps we could provide access to the labelSelector option when starting the watch.  With this we could watch any custom subset of pods and as a bonus it should perform better since it doesn't need to skip over unmatched pods.  However, this would require having the right labels set, maybe this technique can work: https://stackoverflow.com/questions/40131682/how-to-set-label-to-kubernetes-node-at-creation-time

mg03:
@danielnelson : thank you for the feedback.
1) Agree the code is to specific
2) Labels on nodes do not help filter pods during a watch. Only pod attributes can be used to filter during a watch. 
3) Assuming we go with filtering on pod label selector, still we cannot hone in on filtering by pod running on a specific node, the reason is that the nodename attribute is not part of pod labels. Labels are under metadata parent field. the codename is under the spec parent field. 
4) Due to above may I suggest if we can have both label selectors and field selectors, so that users can filter on both or any or none of the filters .
Let me know if you are ok with my suggestion in point 4) above and I can work towards it ....

danielnelson:
That sounds great, thanks.

mg03:
@danielnelson : As per our discussion above, Ive updated the code to support Kubernetes label and field selectors. Please review and let me know your thoughts.

mg03:
@danielnelson : would it be possible for you to take a look at the PR, which is updated as per our conversation about using Kubernetes label and field selectors?

mg03:
@danielnelson : The functionality implemented here is important for adoption within our org, as such please provide guidance on the path forward on this PR.

plugins/inputs/prometheus/kubernetes.go
```diff
@@ -135,6 +144,31 @@ func podReady(statuss []*corev1.ContainerStatus) bool {
 	return true
 }
 
+func podSelector(p *Prometheus) ([]k8s.Option, error) {
+	options := []k8s.Option{}
+
+	var err error
+
+	if len(p.KubernetesLabelSelector) > 0 {
+		_, err1 := labels.Parse(p.KubernetesLabelSelector)
```
danielnelson: This is nice, I didn't think we would be able check the syntax ahead of time.  What could be nice is to check this in an Init() function, so that we can report syntax errors on startup.  Over in `prometheus.go`, add a new function:
```go
func (p *Prometheus) Init() error {
    // check selectors
}
```

This function gets called on plugin startup once, if you return an error, then Telegraf will refuse to start.

plugins/inputs/prometheus/kubernetes.go
```diff
@@ -135,6 +144,31 @@ func podReady(statuss []*corev1.ContainerStatus) bool {
...
+		if err1 != nil {
+			return nil, fmt.Errorf("Label Selector %s validation fail. Skipping it. Error %v", p.KubernetesLabelSelector, err)
```
danielnelson: Keeping with Go [error string style](https://github.com/golang/go/wiki/CodeReviewComments#error-strings), I'd change the string to:
```
fmt.Errorf("label selector validation failed %q: %v",  p.KubernetesLabelSelector, err)
```

plugins/inputs/prometheus/prometheus.go
```diff
@@ -90,6 +96,7 @@ var sampleConfig = `
   ## Restricts Kubernetes monitoring to a single namespace
   ##   ex: monitor_kubernetes_pods_namespace = "default"
   # monitor_kubernetes_pods_namespace = ""
+  # kubernetes_node = ""
```
danielnelson: Update to match the README

plugins/inputs/prometheus/kubernetes.go
mg03: @danielnelson : thank you, will do so

plugins/inputs/prometheus/kubernetes.go
mg03: will do so

plugins/inputs/prometheus/prometheus.go
mg03: will do

mg03:
@danielnelson : I have made the changes as requested, please review.

plugins/inputs/prometheus/prometheus.go
```diff
@@ -124,6 +137,23 @@ func (p *Prometheus) Init() error {
 	if p.MetricVersion != 2 {
 		p.Log.Warnf("Use of deprecated configuration: 'metric_version = 1'; please update to 'metric_version = 2'")
 	}
+
+	if len(p.KubernetesLabelSelector) > 0 {
+		_, err := labels.Parse(p.KubernetesLabelSelector)
+		if err != nil {
+			p.Log.Errorf("label selector validation failed %q: %v", p.KubernetesLabelSelector, err)
+			return err
```
danielnelson: To avoid double logging, Telegraf logs any errors returned from this function, only return the error here without logging.  Include the extra context with the error:
```
return fmt.Errorf("label selector validation failed %q: %v", p.KubernetesLabelSelector, err)
```

plugins/inputs/prometheus/prometheus.go
```diff
@@ -124,6 +137,23 @@ func (p *Prometheus) Init() error {
...
+		}
+	}
+
+	if len(p.KubernetesFieldSelector) > 0 {
+		_, err := fields.ParseSelector(p.KubernetesFieldSelector)
+		if err != nil {
+			p.Log.Errorf("field selector validation failed %s: %v", p.KubernetesFieldSelector, err)
+			return err
```
danielnelson: Only return error

mg03:
@danielnelson : updated the PR with changes as requested, please review.

bamvenkatesh:
@mg03 - Hi, I was just curious why choose DaemonSet instead of a deployment for Telegraf. 

Like @russorat asked, Is there a specific reason you don't want to run a single telegraf to handle the scraping?

Issue #4908, opened 2018-10-23
influxdb_listener plugin: not parsing string fields


danievanzyl:
### Relevant telegraf.conf:
```
[global_tags]
  enduser = "a-dev-server"

[agent]
  interval = "30s"
  round_interval = false
  metric_batch_size = 1000
  metric_buffer_limit = 50000
  collection_jitter = "5s"
  flush_interval = "30s"
  flush_jitter = "0s"
  precision = ""

  debug = false
  quiet = false
  logfile = "/var/test/log/telegraf.log"
  hostname = "dev_hostname"
  omit_hostname = false

[[outputs.influxdb]]

  urls = ["http://localhost:8086"]
  database = "devdb"
  retention_policy = ""
  write_consistency = "any" # Not used
  timeout = "10s"
  username = "" # Not used
  password = "" # Not used
  # udp_payload = 512
  insecure_skip_verify = false
  content_encoding = ""

 # Influx HTTP write listener / ONLY ENABLED
[[inputs.http_listener]]
#   ## Address and port to host HTTP listener on
#   service_address = ":8186"
#
#   ## maximum duration before timing out read of the request
#   read_timeout = "10s"
#   ## maximum duration before timing out write of the response
#   write_timeout = "10s"
#
#   ## Maximum allowed http request body size in bytes.
#   ## 0 means to use the default of 536,870,912 bytes (500 mebibytes)
#   max_body_size = 0
#
#   ## Maximum line size allowed to be sent in bytes.
#   ## 0 means to use the default of 65536 bytes (64 kibibytes)
#   max_line_size = 0
#
#   ## Set one or more allowed client CA certificate file names to
#   ## enable mutually authenticated TLS connections
#   tls_allowed_cacerts = ["/etc/telegraf/clientca.pem"]
#
#   ## Add service certificate and key
#   tls_cert = "/etc/telegraf/cert.pem"
#   tls_key = "/etc/telegraf/key.pem"
#
#   ## Optional username and password to accept for HTTP basic authentication.
#   ## You probably want to make sure you have TLS configured above for this.
#   # basic_username = "foobar"
#   # basic_password = "barfoo"


```
### System info:

Telegraf 1.8.2 (git: HEAD 996a91f2)
Linux dev-vm1 3.10.0-693.21.1.el7.x86_64 #1 SMP Wed Mar 7 19:03:37 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux

### Steps to reproduce:

[...]

PR #6974, opened 2020-02-03
Influxdb listener refactor

### LINKED ISSUE #2280: POST request to http_listener fails with EOF

reimda:
Change parser to stream data instead of reading the entire request before parsing.
Remove parsed line length restriction.
Update HTTP server handlers and url routing.
Gracefully shut down HTTP server.
Improve logging and error messages.

Resolves #2280, #4908, #6124, #6186


plugins/inputs/influxdb_listener/influxdb_listener.go
```diff
@@ -0,0 +1,398 @@
...
+		parser := influx.NewStreamParser(body)
+		parser.SetTimeFunc(h.TimeFunc)
+
+		var err error
+		precisionStr := req.URL.Query().Get("precision")
+		if precisionStr != "" {
+			var precision time.Duration
+			precision, err = getPrecisionMultiplier(precisionStr)
+			if err != nil {
+				h.Log.Debug(err.Error())
```
danielnelson: add context to log message

plugins/inputs/influxdb_listener/influxdb_listener.go
```diff
@@ -0,0 +1,398 @@
...
+		d = time.Microsecond
+	case "ms":
+		d = time.Millisecond
+	case "s":
+		d = time.Second
+	case "m":
+		d = time.Minute
+	case "h":
+		d = time.Hour
+	default:
```
danielnelson: Add case for ns

plugins/inputs/influxdb_listener/influxdb_listener.go
```diff
@@ -0,0 +1,398 @@
...
+				badRequest(res, err.Error())
+			}
+			parser.SetTimePrecision(precision)
+		}
+
+		var m telegraf.Metric
+		var partial bool
+		for {
+			m, err = parser.Next()
+			h.bytesRecv.Incr(int64(parser.Position())) //todo incr by position difference
```
danielnelson: do the todo

plugins/inputs/influxdb_listener/influxdb_listener.go
```diff
@@ -0,0 +1,398 @@
...
+
+			//continue parsing metrics even if some are malformed
+			if _, ok := err.(*influx.ParseError); ok {
+				//todo save error for X-Influxdb-Error header
```
danielnelson: todo and clean up other comments

plugins/inputs/influxdb_listener/influxdb_listener.go
```diff
@@ -0,0 +1,398 @@
...
+			} else if err != nil {
+				break
+			}
+
+			if h.DatabaseTag != "" && db != "" {
+				m.AddTag(h.DatabaseTag, db)
+			}
+			h.acc.AddMetric(m)
+		}
+		if err.Error() != "EOF" {
```
danielnelson: if err != influx.EOF

plugins/inputs/influxdb_listener/influxdb_listener.go
```diff
@@ -0,0 +1,398 @@
...
+
+				h.authFailures.Incr(1)
+				http.Error(res, "Unauthorized", http.StatusUnauthorized)
+				return
+			}
+		}
+
+		f(res, req)
+	}
+}
```
danielnelson: Use internal.AuthHandler

plugins/inputs/influxdb_listener/influxdb_listener.go
```diff
@@ -0,0 +1,398 @@
...
+	"github.com/influxdata/telegraf/plugins/inputs"
+	"github.com/influxdata/telegraf/plugins/parsers/influx"
+	"github.com/influxdata/telegraf/selfstat"
+)
+
+const (
+	// DEFAULT_MAX_BODY_SIZE is the default maximum request body size, in bytes.
+	// if the request body is over this size, we will return an HTTP 413 error.
+	// 500 MB
+	DEFAULT_MAX_BODY_SIZE = 500 * 1024 * 1024
```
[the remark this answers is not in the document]
reimda: done

plugins/inputs/influxdb_listener/influxdb_listener.go
```diff
@@ -0,0 +1,398 @@
...
+)
+
+type InfluxDBListener struct {
+	ServiceAddress string `toml:"service_address"`
+	// Port gets pulled out of ServiceAddress
+	Port int
```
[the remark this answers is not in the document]
reimda: The port is used in unit tests.  I unexported it to prevent it from being changed by toml config.

plugins/inputs/influxdb_listener/influxdb_listener.go
```diff
@@ -0,0 +1,398 @@
...
+	tlsint.ServerConfig
+
+	ReadTimeout   internal.Duration `toml:"read_timeout"`
+	WriteTimeout  internal.Duration `toml:"write_timeout"`
+	MaxBodySize   internal.Size     `toml:"max_body_size"`
+	BasicUsername string            `toml:"basic_username"`
+	BasicPassword string            `toml:"basic_password"`
+	DatabaseTag   string            `toml:"database_tag"`
+
+	TimeFunc influx.TimeFunc
```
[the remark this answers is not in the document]
reimda: unexported

plugins/inputs/influxdb_listener/influxdb_listener.go
```diff
@@ -0,0 +1,398 @@
...
+	requestsServed  selfstat.Stat
+	writesServed    selfstat.Stat
+	queriesServed   selfstat.Stat
+	pingsServed     selfstat.Stat
+	requestsRecv    selfstat.Stat
+	notFoundsServed selfstat.Stat
+	buffersCreated  selfstat.Stat
+	authFailures    selfstat.Stat
+
+	Log telegraf.Logger
```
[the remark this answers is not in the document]
reimda: unexported

plugins/inputs/influxdb_listener/influxdb_listener.go
reimda: Done.  I changed internal.AuthHandler to call http.Error so all users wouldn't have to do it separately.  I also added a realm parameter.

plugins/inputs/influxdb_listener/influxdb_listener.go
```diff
@@ -0,0 +1,395 @@
...
+	h.log.Infof("Started HTTP listener service on %s", h.ServiceAddress)
+
+	return nil
+}
+
+// Stop cleans up all resources
+func (h *InfluxDBListener) Stop() {
+	deadline := time.Now().Add(10 * time.Second)
+	ctx, cancel := context.WithDeadline(context.Background(), deadline)
+	defer cancel() // go vet requires us to call cancel even though we're waiting for the deadline
```
ssoroka: what happens if the deadline passes, will wait hang?

plugins/inputs/influxdb_listener/influxdb_listener.go
```diff
@@ -0,0 +1,395 @@
...
+func (h *InfluxDBListener) handlePing() http.HandlerFunc {
+	return func(res http.ResponseWriter, req *http.Request) {
+		defer h.pingsServed.Incr(1)
+		verbose := req.URL.Query().Get("verbose")
+
+		// respond to ping requests
+		if verbose != "" && verbose != "0" && verbose != "false" {
+			res.WriteHeader(http.StatusOK)
+			b, _ := json.Marshal(map[string]string{"version": "1.0"}) // based on header set above
+			res.Write(b)
```
ssoroka: I'd probably just replace these with:
```
res.Write([]byte(`{"version": "1.0"}`))
```

plugins/inputs/influxdb_listener/influxdb_listener.go
```diff
@@ -0,0 +1,395 @@
...
+		// Check that the content length is not too large for us to handle.
+		if req.ContentLength > h.MaxBodySize.Size {
+			tooLarge(res)
+			return
+		}
+
+		db := req.URL.Query().Get("db")
+
+		body := req.Body
+		body = http.MaxBytesReader(res, body, h.MaxBodySize.Size)
```
ssoroka: 💯 👍

plugins/inputs/influxdb_listener/influxdb_listener.go
```diff
@@ -0,0 +1,395 @@
...
+		// Handle gzip request bodies
+		if req.Header.Get("Content-Encoding") == "gzip" {
+			var err error
+			body, err = gzip.NewReader(req.Body)
```
ssoroka: should this be `body` instead of `req.Body` ? seems to be bypassing the MaxBytesReader

plugins/inputs/influxdb_listener/influxdb_listener.go
```diff
@@ -0,0 +1,395 @@
...
+		if precisionStr != "" {
+			var precision time.Duration
+			precision = getPrecisionMultiplier(precisionStr)
+			parser.SetTimePrecision(precision)
+		}
+
+		var m telegraf.Metric
+		var parseErrorCount int
+		var lastPos int
+		lastPos = 0
```
ssoroka:
```suggestion
		lastPos := 0
```
and then you can remove the var.

plugins/inputs/influxdb_listener/influxdb_listener.go
```diff
@@ -0,0 +1,395 @@
...
+			h.bytesRecv.Incr(int64(pos - lastPos))
+			lastPos = pos
+
+			//continue parsing metrics even if some are malformed
+			if parseErr, ok := err.(*influx.ParseError); ok {
+				parseErrorCount += 1
+				errStr := parseErr.Error()
+				if firstParseErrorStr == "" {
+					firstParseErrorStr = errStr
+				} else {
```
ssoroka: don't need else

plugins/inputs/influxdb_listener/influxdb_listener.go
```diff
@@ -0,0 +1,395 @@
...
+				}
+				continue
+			} else if err != nil {
```
ssoroka: don't need else, can be a new `if`, or removed if you do the `for err == nil`

plugins/inputs/influxdb_listener/influxdb_listener.go
```diff
@@ -0,0 +1,395 @@
...
+			precision = getPrecisionMultiplier(precisionStr)
+			parser.SetTimePrecision(precision)
+		}
+
+		var m telegraf.Metric
+		var parseErrorCount int
+		var lastPos int
+		lastPos = 0
+		var firstParseErrorStr string
+		for {
```
ssoroka: `for err == nil {` ?

plugins/inputs/influxdb_listener/influxdb_listener.go
reimda: yep, good catch

plugins/serializers/prometheus/collection.go
```diff
@@ -328,9 +329,11 @@ func (c *Collection) Add(metric telegraf.Metric) {
 }
 
 func (c *Collection) Expire(now time.Time, age time.Duration) {
+	fmt.Println(0, now, age)
 	expireTime := now.Add(-age)
 	for _, entry := range c.Entries {
 		for key, metric := range entry.Metrics {
+			fmt.Println(1, metric.Time, expireTime)
```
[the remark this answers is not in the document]
reimda: I agree.  It looks like this came in by accident in one of Daniel's commits.

internal/http.go
```diff
@@ -9,35 +9,40 @@ import (
 // ErrorFunc is a callback for writing an error response.
 type ErrorFunc func(rw http.ResponseWriter, code int)
```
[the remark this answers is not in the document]
reimda: It's still used by ipRangeHandler below in this file.  I'll move it in the file to be closer to ipRangeHandler.

internal/http.go
```diff
@@ -9,35 +9,40 @@ import (
...
 	password string
-	onError  ErrorFunc
+	realm    string
+	onError  BasicAuthErrorFunc
 	next     http.Handler
 }
 
 func (h *basicAuthHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
-	if h.username != "" || h.password != "" {
+	if h.username != "" && h.password != "" {
```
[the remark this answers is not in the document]
reimda: I assumed this was coding error but I checked with Daniel and he had a couple use cases for either being empty and we checked the RFC which says neither is required.  I'll revert this change.

plugins/inputs/influxdb_listener/influxdb_listener.go
reimda: This change is what we think will fix the panics mentioned in influxdata/EAR#1379.  They seem to be caused by http server connections receiving data after the server is shutdown.  Before this change we would close the listeners directly but not shut down existing connections.

We think using Server.Shutdown will fix the problem because in addition to closing listeners it also shuts down existing connections.  We didn't do that before because Server.Shutdown didn't exist at the time.  The last change to this file related to server shutdown was Oct 2016 and Server.Shutdown was added in go 1.8 in Feb 2017.  

To answer your question, I don't think wait will hang if the deadline passes because Server.Serve should return as soon as the listeners are closed.  But now that I review it more closely I think it's possible that this won't completely prevent the condition that causes the panics.  If the connections don't close in the 10 seconds the panics could still happen.

I don't think we want to wait indefinitely but the deadline may just set up a race condition.  What do you think we should do?  @danielnelson?

plugins/inputs/influxdb_listener/influxdb_listener.go
reimda: I wanted it to be clear that it lastPos is of type int because that's what parser.Position() returns later on.  I could leave out the assignment altogether because it defaults to zero anyway, but I wanted it to look like zeroing it initially was intentional.

To get rid of the extra line I'll change it to

```suggestion
		var lastPos int = 0
```
which does the same thing but looks like it's on purpose.

plugins/inputs/influxdb_listener/influxdb_listener.go
danielnelson: Wait will wait until Server returns (above in Connect).  I assumed that the Server waited for all handler goroutines before returning from Serve, but I guess I'm not 100% sure.  You could probably check for this by temporarily adding a Sleep call into the handler and seeing how long it takes to complete.

Nitpick: Since a deadline is absolute but a timeout is relative, I'd use a timeout so you don't need to add it to time.Now.

plugins/inputs/influxdb_listener/influxdb_listener.go
reimda: It does look a little funny, but I think it's more correct to do the error check in the middle of the loop here.

plugins/inputs/influxdb_listener/influxdb_listener.go
reimda: Thanks.  I'm surprised go vet didn't raise a stink about this.

plugins/inputs/influxdb_listener/influxdb_listener.go
reimda: If I remove the else, it will break out too early when err is a parse error that we've already handled.  If I do ```for err == nil```  and remove the check here, it will add a metric after a non-parse error when the metric isn't valid.  I think the logic is right as is.

plugins/inputs/influxdb_listener/influxdb_listener.go
reimda: I would too, but this was already there and it didn't seem worth changing.

plugins/inputs/influxdb_listener/influxdb_listener.go
```diff
@@ -0,0 +1,395 @@
...
+	return nil
+}
+
+// Stop cleans up all resources
+func (h *InfluxDBListener) Stop() {
+	deadline := time.Now().Add(10 * time.Second)
+	ctx, cancel := context.WithDeadline(context.Background(), deadline)
+	defer cancel() // go vet requires us to call cancel even though we're waiting for the deadline
+	h.server.Shutdown(ctx)
+	h.wg.Wait()
```
ssoroka: So I've read over the docs and I think I've convinced myself that the `wg.Wait` will not hang, however I also thing it's not necessary, since `server.Shutdown(ctx)` is a blocking call, and this can be simplified by removing the waitgroup and it will have the same functionality.

plugins/inputs/influxdb_listener/influxdb_listener.go
ssoroka: 👍 WithTimeout.

> it also shuts down existing connections.

To be pedantic, it stops listening and lets the connections finish what they're doing. It's still possible for them to write a response before the deadline, and it's possible that they don't finish at all before the server exits. A good practice for long-running http connections would be to pass a main context down through the app stack so that requests can "notice" if the app is trying to shut down (their contexts are cancelled), and gracefully stop doing work.
details here https://golang.org/src/net/http/server.go?s=83923:83977#L2643

> the deadline may just set up a race condition

I don't think that's the case, other than the aforementioned requests-not-finishing-in-time, which is fine.

reimda: I reworked this to have the connection pay attention to the context.  With connections closing quickly we can let shutdown happen gracefully and remove the 10 second timeout.  Shutdown will block which prevents the accumulator from going away while the connection needs it, so the panic should still be prevented.

plugins/inputs/influxdb_listener/influxdb_listener.go
reimda: I checked this and agree.  I removed the wg

plugins/inputs/influxdb_listener/influxdb_listener.go
```diff
@@ -0,0 +1,404 @@
...
+  # basic_username = "foobar"
+  # basic_password = "barfoo"
+`
+
+func (h *InfluxDBListener) SampleConfig() string {
+	return sampleConfig
+}
+
+func (h *InfluxDBListener) Description() string {
+	return "InfluxDB listener"
```
danielnelson: Let's come up with a better description.

plugins/inputs/influxdb_listener/influxdb_listener.go
```diff
@@ -0,0 +1,404 @@
...
+}
+
+// Stop cleans up all resources
+func (h *InfluxDBListener) Stop() {
+	err := h.server.Shutdown(context.Background())
+	if err != nil {
+		h.log.Infof("Error shutting down HTTP server: %v", err.Error())
+	}
+
+	h.log.Infof("Stopped HTTP listener service on %s", h.ServiceAddress)
```
danielnelson: Let's remove this log message.  Usually we don't have plugins announce they are starting/stopping with the exception of plugins that listen on a port, but for these we usually only announce the listening.

plugins/inputs/influxdb_listener/influxdb_listener.go
```diff
@@ -0,0 +1,404 @@
...
+		}
+
+		parser := influx.NewStreamParser(body)
+		parser.SetTimeFunc(h.timeFunc)
+
+		precisionStr := req.URL.Query().Get("precision")
+		if precisionStr != "" {
+			var precision time.Duration
+			precision = getPrecisionMultiplier(precisionStr)
+			parser.SetTimePrecision(precision)
```
danielnelson: I still want to remove this, and here in this plugin multiple the metric as needed.

plugins/inputs/influxdb_listener/README.md
```diff
@@ -30,14 +30,10 @@ submits data to InfluxDB determines the destination database.
...
 
-  ## Maximum allowed http request body size in bytes.
-  ## 0 means to use the default of 536,870,912 bytes (500 mebibytes)
+  ## Maximum allowed HTTP request body size in bytes.
+  ## 0 means to use the default of 32MiB.
   max_body_size = 0
 
-  ## Maximum line size allowed to be sent in bytes.
-  ## 0 means to use the default of 65536 bytes (64 kibibytes)
-  max_line_size = 0
```
danielnelson: Even though we are removing this option, let's keep it in the configuration so that it won't cause an error for those who already have it set.  I wrote up how I've been handling this: https://github.com/influxdata/telegraf/wiki/Deprecation#deprecate-options

plugins/inputs/influxdb_listener/influxdb_listener.go
reimda: What's the preferred format?  I chose this after looking at the other *_listener plugin descriptions which are similar.

I looked over the other plugin descriptions and couldn't find a clear pattern.  For service inputs they generally describe what the plugin is, but for non-service inputs they generally describe what the plugin does.  For example, socket listener says "Generic socket listener capable of handling multiple socket types" and memory says "Read metrics about memory usage".  Some of the "does" style ones are first person "read" and some are third person "reads".  There are a few that are long complete sentences "This plugin implements x to gather y data to troubleshoot z".

Anyway, do either of these sound better or do you have suggestions?
"HTTP server implementing InfluxDB 1.x API to accept metrics"
"Accept metrics over InfluxDB 1.x HTTP API"

danielnelson: Either of those sound great, and I love if we could decide on and write down some guidelines to ease writing descriptions in the future.

plugins/inputs/influxdb_listener/influxdb_listener.go
```diff
@@ -0,0 +1,404 @@
...
+	requestsServed  selfstat.Stat
+	writesServed    selfstat.Stat
+	queriesServed   selfstat.Stat
+	pingsServed     selfstat.Stat
+	requestsRecv    selfstat.Stat
+	notFoundsServed selfstat.Stat
+	buffersCreated  selfstat.Stat
+	authFailures    selfstat.Stat
+
+	log telegraf.Logger
```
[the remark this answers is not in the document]
reimda: It worked for me in the unit tests because I changed the tests to set lower case log.  I see now that it doesn't work like that.  :)  I've changed it like you suggested.

plugins/inputs/influxdb_listener/influxdb_listener.go
reimda: I've changed it to "Accept metrics over InfluxDB 1.x HTTP API".  I agree we should get docs to review all the descriptions.

plugins/inputs/influxdb_listener/influxdb_listener.go
reimda: done

plugins/inputs/influxdb_listener/README.md
reimda: done

plugins/inputs/influxdb_listener/influxdb_listener.go
reimda: We paired on this after your last comment and we realized that since the precision logic needs to know whether or not the timestamp is present in the line protocol, it makes to have that logic in MetricHandler.  Moving all the precision logic to the input plugin would require us to add a way to have the parser tell the plugin whether there is a timestamp which didn't seem worth it.

We were able to clean things up a little by removing metric.Builder and moving the timestamp truncation into MetricHandler.

plugins/inputs/influxdb_listener/influxdb_listener.go
```diff
@@ -180,26 +182,24 @@ func (h *InfluxDBListener) Start(acc telegraf.Accumulator) error {
...
 // Stop cleans up all resources
 func (h *InfluxDBListener) Stop() {
-	deadline := time.Now().Add(10 * time.Second)
-	ctx, cancel := context.WithDeadline(context.Background(), deadline)
-	defer cancel() // go vet requires us to call cancel even though we're waiting for the deadline
-	h.server.Shutdown(ctx)
-	h.wg.Wait()
-
-	h.log.Infof("Stopped HTTP listener service on %s", h.ServiceAddress)
+	err := h.server.Shutdown(context.Background())
```
ssoroka: I'm a bit concerned about having no deadline at all. If something doesn't close cleanly, it means the service could just hang on close?

plugins/inputs/influxdb_listener/influxdb_listener.go
```diff
@@ -271,45 +271,56 @@ func (h *InfluxDBListener) handleWrite() http.HandlerFunc {
...
 				parseErrorCount += 1
 				errStr := parseErr.Error()
 				if firstParseErrorStr == "" {
 					firstParseErrorStr = errStr
-				} else {
 				}
 				continue
 			} else if err != nil {
+				// Either we're exiting cleanly (err ==
+				// influx.EOF) or there's an unexpected error
```
ssoroka: do we want to log it if it's not influx.EOF?

plugins/inputs/influxdb_listener/influxdb_listener.go
danielnelson: I think the agent could give up on waiting for the plugin after a period of time instead of handling that in plugin.  We could consider adding this in a different PR.

The ReadTimeout option should be the maximum amount of time a request can run, and I think that is the max time we can spend here.

reimda: When we call Shutdown it closes its internal context and I modified handleWrite to check that context before parsing each metric.  With that check I don't know of anything that would cause graceful shutdown to block.

plugins/inputs/influxdb_listener/influxdb_listener.go
reimda: that's on line 323

PR #6982, opened 2020-02-04
 processor/strings: add support for `title` convertor

themontem:
- Title (https://golang.org/pkg/strings/#Title) function support all unicode letters.
- BUG: The rule Title uses for word boundaries does not handle Unicode punctuation properly, that should not effect much with use-case of Title rule in telegraf.

### Required for all PRs:


danielnelson:
Thanks, nice addition.

PR #6985, opened 2020-02-05
Adding routing by measurement name to kafka output plugin

### LINKED ISSUE #4170: Kafka Producer & Consumer - Topic based routing based on criteria specified in an expression
## Feature Request

Proposal:

Wayfair is requesting this capability to enable smarter routing of Kafka traffic to various Topics.
Telegraf Kafka Consumer could then direct the data to different InfluxDB instances based on the Topic.

Below is an example of what this would look like:
We are looking to have  flexible Kafka routing engine for both the Telegraf Kafka consumer and producer.
 
For Kafka Producer…
 
We are looking for something that works sort of like a map data structure (or something like a case statement).  Imagine the following…
 
TOPIC_ROUTES:
    CRITERIA_SET_1:
        TOPIC(S)
    CRITERIA_SET_2:
        TOPIC(S)
    CRITERIA_SET_3:
        TOPIC(S)
    DEFAULT:
         TOPIC(S)
 
NOTE: we want the option of using measurement name or tag based criteria sets.  Note we may want the flexibility of replicating data to more than one topic.
 
For the Kafka Consumer we are looking for a routing scheme that will enable us to specify a specific cluster and DB destination per topic set (not necessarily a v1 feature)
 
                NOTE: HOW CAN WE COMBINE TOPIC AND MEASUREMENT_TAG based filtering?
 
                      DB_ROUTES:
                          TOPIC_AND_MEASUREMENT_TAG_SET_1:
                              DB(s)
                          TOPIC_AND_MEASUREMENT_TAG_SET_1:
                              DB(s)
                           DEFAULT:
                              DB(s)
 
Note we may want the flexibility of replicating data to more than one db (not necessarily a v1 feature)

It would allow Telegraf to direct data to appropriate InfluxDB cluster or instance.  A very powerful feature which will greatly simplify scaling up or segmenting workloads.

scottgrover:
This change adds functionality to route measurements by name to different kafka topics. You can do a substring match or an exact match. 

Things to note: 
- Rules are evaluated sequentially, as defined by the configuration.
- If a topic suffix configuration exists, and no configured routing rules match, the topic suffix configuration will be applied. 
- If no topic suffix configuration exists, and no rules match, the topic from the base output configuration ([[outputs.kafka]]) is applied. 

Let me know your thoughts / if anything is unclear.  This aims to close #4170.

## Config example:
```
[agent]
  interval = "1s"
  flush_interval = "5s"

[[inputs.exec]]
   commands = ["/tmp/create_line_protocol.sh"]
   timeout = "5s"
   data_format = "influx"

[[outputs.kafka]]
  # URLs of kafka brokers
  brokers = ["kafka.com:9092"]
  topic = "sgrover_testing_3"
  routing_key = "random"
  compression_codec = 2
  required_acks = 1
  max_retry = 1
  data_format = "influx"
  version = "1.1.0"

[[outputs.kafka.topic_routing]]
  method = "measurement"
  match_type = "exact"
  match_values = ["measurement_name_1", "bar", "other"]
  topic = "sgrover_testing_1"

[[outputs.kafka.topic_routing]]
  method = "measurement"
  match_type = "substring"
  match_values = ["name_2", "foo"]
  topic = "sgrover_testing_2"
```

`/tmp/create_line_protocol`:
```
#!/bin/bash
#
# This script creates line protocol and sends it to stdout for the telegraf inputs.exec plugin to execute.

TIMESTAMP=$(date +%s%N)
echo "measurement_name_1,host=devnode.com,tag_set=true temperature=82 $TIMESTAMP"
echo "measurement_name_2,host=devnode.com,tag_set=false temperature=80 $TIMESTAMP"
echo "measurement_name_3,host=devnode.com,tag_set=false temperature=90 $TIMESTAMP"
```

With the above configuration, we expect `measurement_name_1` measurements to end up in `sgrover_testing_1` topic, `measurement_name_2` measurements end up in `sgrover_testing_2` topic, and `measurement_name_3` measurements don't have an associated rule, so they ends up in the default topic: `sgrover_testing_3`

### Required for all PRs:


danielnelson:
@scottgrover Thanks for sharing this pull request, but I'd like to handle metric routing within plugins using a slightly different approach that will be unified across outputs and reduce the amount of code we need to mirror between outputs.

The idea is that processors will set tags and then outputs can use these for routing.  I know you have already seen the method that involves using multiple output plugins and [metric filtering][], but for situations where the routing is more dynamic or in the case where we want to avoid multiple connections we will add two options to the plugin to handle this, for this example it would be `topic_tag` and `exclude_topic_tag`.  These simple allow the topic to read from a tag value and the tag to be removed if desired.

So for your above example:
```
[[processors.override]]
  namepass = ["measurement_name_1", "bar", "other"]
  [[processors.override.tag]]
    topic = "sgrover_testing_1"

[[processors.override]]
  namepass = ["name_2", "foo"]
  [[processors.override.tag]]
    topic = "sgrover_testing_1"

[[outputs.kafka]]
  brokers = ["kafka.com:9092"]
  routing_key = "random"
  compression_codec = 2
  required_acks = 1
  max_retry = 1
  data_format = "influx"
  version = "1.1.0"
  topic_tag = "topic"
  exclude_topic_tag = true
```

Would be happy to accept a pull request that adds `topic_tag` and `exclude_topic_tag`.  I also have an idea for improving the performance of `exclude_topic_tag` over the implementation we currently have in other plugins such as the `influxdb` output, I can share if you are interested in working on this.

[metric filtering]: https://github.com/influxdata/telegraf/blob/master/docs/CONFIGURATION.md#metric-filtering

scottgrover:
Hi Daniel,

As discussed in #4170, adding processor plugins into our Telegraf configuration is not something that will be feasible at Wayfair scale. We handle close to 80B metrics per day, so our config is, and needs to remain, pretty lean. We cannot add a processor plugins for each of our topics. We are looking for this logic in the kafka output plugin since we already use this plugin.

Let us know if this is an issue/how to proceed. 

Scott

danielnelson:
We should be able to make the processor method comparable in terms of configuration file size.  In my above example I showed the `override` processor only because it was enough to handle your earlier example, but for more complicated mappings we can use or add different processors.  I think it might be useful to see more examples of how you are mapping metrics to topics.

scottgrover:
What I meant by a lean configuration isn't necessarily the size but instead the complexity in terms of plugin usage.

From your override processor plugin example, am I right in my understanding that all instantiated plugins will be evaluated, regardless if the first matched? This is not scalable, as for each processor plugin we add, we add another evaluation on every metric passing through our receivers. As a pilot, we would like to take our 30 noisiest metrics and send them to their own topic. We will not be adding 30 processor plugins to our configuration. The impact on performance would be too significant.

With this Kafka output change, it returns the topic name after a rule matches. Further rules are not evaluated after the first match. This makes it more efficient, as you're not wasting cycles on further rule evaluations after a match has already occurred.

danielnelson:
Each case wouldn't necessarily need to be its own processor instance, a single processor could handle similar cases.  For the exact and substring match, you could do it all with a single processor.   We also can introduce new processors as needed.  I'm confident we can structure it in a way that the processors will be extremely minor compared to the cost of sending to Kafka. 

Give me a better idea of your real world routing rules and your performance requirements, either here publicly or through support, and I can help complete or improve any missing parts.

PR #6991, opened 2020-02-06
Support obtaining all builds within the max_build_age in jenkins input

### LINKED ISSUE #6966: Jenkins plugin doesn't collect all builds for a job
The plugin only collects the metrics of builds that have finished. That's fine, but it also only collects the metrics of the last build to run within a job. So if there is a new build that starts before the previous has finished, the metrics of the previous build will not be collected.

### System info:
telegraf 1.13.2 Docker image

### Steps to reproduce:
1. Configure Jenkins plugin
2. Run two concurrent builds in Jenkins

### Expected behavior:
Both builds will be recorded. Even better, I want all builds within the `max_build_age` setting of the conf file to be recorded.

### Actual behavior:
Only the last build to finish will be recorded.

mildebrandt:
This fixes #6966 

This changes the job query to include a depth parameter so the build
information for the job is obtained in the same result as the job
details. I also limit the size of the body to only the fields we're
interested in using the tree parameter.

Before the change, we would get metrics of 20 nodes and 100 builds in
23.56 seconds on average. After this change, we get metrics of 20 nodes
and 2006 builds in 7.76 seconds on average.

We were able to get a 3x decrease in time while increasing the number
of metrics by over 16x.

### Required for all PRs:

- [NA] Associated README.md updated.

### DIFF
--- a/plugins/inputs/jenkins/client.go
+++ b/plugins/inputs/jenkins/client.go
@@ -48,7 +48,7 @@ func (c *client) init() error {
 		}
 	}
 	// first api fetch
-	if err := c.doGet(context.Background(), jobPath, new(jobResponse)); err != nil {
+	if err := c.doGet(context.Background(), jsonSuffix, new(jobResponse)); err != nil {
 		return err
 	}
 	return nil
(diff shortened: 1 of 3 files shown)


danielnelson:
I'm concerned that changing the depth may break the sub job handling, but honestly I'm not sure how to setup subjobs nowadays.  It might be that the plugins it used are no longer current.

@kelwang Do you remember how to setup a job with subjobs?

kelwang:
> I'm concerned that changing the depth may break the sub job handling, but honestly I'm not sure how to setup subjobs nowadays. It might be that the plugins it used are no longer current.
> 
> @kelwang Do you remember how to setup a job with subjobs?

I haven't touch that for a while. @mark-rushakoff do you know?

mark-rushakoff:
It's been too long for me either. I can't provide any meaningful help without going in and re-learning everything about Jenkins jobs.

kelwang:
> It's been too long for me either. I can't provide any meaningful help without going in and re-learning everything about Jenkins jobs.

@mark-rushakoff  Can you check the url like
https://jenkins.a.influxcloud.net/api/json/job/k8s-cloud/job/apps/job/chore%252Frevert-urm?depth=1&tree=jobs[name,jobs],name,builds[building,duration,result,timestamp,number]

I can't see any builds info on my end, not sure whether it's my permission, or it's actually the depth related.

mark-rushakoff:
@kelwang I do see a JSON response at that endpoint, but there doesn't appear to be any numeric data for it.

If you need to debug further, I would suggest using a publicly accessible Jenkins instance so we can rule out permissions issues. https://wiki.jenkins.io/pages/viewpage.action?pageId=58001258 contains a list of public-facing Jenkins servers.

srebhan:
@mildebrandt and others, is there still interest in getting this merged?

mildebrandt:
We're no longer using telegraf or influx, so I have no need for this anymore.

srebhan:
Ok I will close the PR then. If anybody wants to take over this, please feel free to speak up and reopen/resubmit!

Pauss:
Hello, 
It is possible to merge this PR? I'm having the same issue that not all builds of the same job are collected by telegraf, if they were started in parallel.

PR #7002, opened 2020-02-09
Fix float conversion and Connection issue

garciaolais:
### Required for all PRs:

this solve the next 2 issues that [@serrj-sv](https://github.com/serrj-sv) reported

- [improve float conversion](https://github.com/influxdata/telegraf/pull/6154#issuecomment-583509623)
- [if host associated with input plugin is down it prevents telegraf service from starting](https://github.com/influxdata/telegraf/pull/6154#issuecomment-583538364)

### DIFF
--- a/plugins/inputs/modbus/modbus.go
+++ b/plugins/inputs/modbus/modbus.go
@@ -154,13 +154,7 @@ func (m *Modbus) Init() error {
 		return fmt.Errorf("device name is empty")
 	}
 
-	err := connect(m)
-	if err != nil {
-		m.isConnected = false
-		return err
-	}
-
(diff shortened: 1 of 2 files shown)


plugins/inputs/modbus/modbus.go
```diff
@@ -562,6 +580,7 @@ func (m *Modbus) Gather(acc telegraf.Accumulator) error {
 
 	err := m.getFields()
 	if err != nil {
+		disconnect(m)
```
danielnelson: Why do we need to disconnect?

garciaolais: Because if something wrong happened we need to close the connection, 

for example, if I turn off my serial controller and then turn on, I get multiple errors about the connection.

Issue #5922, opened 2019-05-29
NewRelic output plugin


hsinghkalsi:
## Feature Request
      Submission of telegraf output plugin for NewRelic
### Proposal:
      Create telegraf output plugin to push data to NewRelic 
### Current behavior:
      No plugin exists 
### Desired behavior:
      Ability to configure output plugin to 
### Use case: 
      Integrate telegraf output with NewRelic Insights

PR #7019, opened 2020-02-12
 NewRelic output plugin 

hsinghkalsi:
### Required for all PRs:

Issue: NewRelic output plugin #5922


danielnelson:
Thank you for the pull request, we will take a look as soon as we can.

hsinghkalsi:
@danielnelson  is there any timeline associated with this? we are eagerly awaiting a formal build for this

sjwang90:
Hi @hsinghkalsi, we're currently working on features for our upcoming release next month.  We will get this in our list of output plugins for a future release.

hsinghkalsi:
@sjwang90 Do we have any insights into the timelines of the future releases, where we can expect this PR. Thanks

JimHagan:
HI @danielnelson This is Jim Hagan, formerly of Wayfair and now with New Relic.  I've been asked to check on the status of this particular output plugin as there are a number of use cases we'd like to solve with it.

Feel free to let me know if there are are open critique's of the code that need to be addressed.  


Thanks!

.gitignore
```diff
@@ -3,3 +3,4 @@
 /telegraf.exe
 /telegraf.gz
 /vendor
+telegraf.conf
```
danielnelson, line 6: Can you revert this file, I like to keep this trimmed down to only the files the build system creates.

plugins/outputs/newrelic/newrelic.go
```diff
@@ -0,0 +1,115 @@
...
+
+// Write takes in group of points to be written to the Output
+func (nr *NewRelic) Write(metrics []telegraf.Metric) error {
+	for _, metric := range metrics {
+		// create tag map
+		tags := make(map[string]interface{})
+		for k, v := range metric.Tags() {
+			tags[k] = v
+		}
+		for k, v := range metric.Fields() {
```
danielnelson, line 67: Update the Tags() and Fields() loops to use TagList() and FieldList(), these will perform much better.

plugins/outputs/newrelic/newrelic.go
```diff
@@ -0,0 +1,115 @@
...
+			case uint64:
+				mvalue = float64(n)
+			case float64:
+				mvalue = float64(n)
+			case bool:
+				mvalue = float64(0)
+				if n {
+					mvalue = float64(1)
+				}
+			default:
```
danielnelson, line 87: The string type is pretty common, so I we won't want to warn each time it is encountered.  Add case for strings that skips them.

plugins/outputs/newrelic/newrelic.go
```diff
@@ -0,0 +1,115 @@
...
+	return "Send metrics to New Relic metrics endpoint"
+}
+
+// SampleConfig : return  default configuration of the Output
+func (nr *NewRelic) SampleConfig() string {
+	return `
+	## New Relic Insights API key (required)
+	insights_key = "insights api key"
+	#metric_prefix if defined, prefix's metrics name for easy identification (optional)
+	# metric_prefix = "Telegraf_"
```
danielnelson, line 33: We should show the default value for this item in the comment, so right now without a default set:
```
# metric_prefix = ""
```

plugins/outputs/newrelic/newrelic_test.go
```diff
@@ -0,0 +1,100 @@
...
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			nr := &NewRelic{
+				harvestor:    tt.newrelic.harvestor,
+				InsightsKey:  tt.newrelic.InsightsKey,
+				MetricPrefix: tt.newrelic.MetricPrefix,
+			}
+			if err := nr.Connect(); (err != nil) != tt.wantErr {
```
danielnelson, line 95: If this actually connects to a remote service, we should make the test be skipped in `-short` mode.

hsinghkalsi: @danielnelson This test is more of a negative test and only validates that we fail if we provide dummy values. Should we still skip it?

.gitignore
hsinghkalsi: Done

plugins/outputs/newrelic/newrelic.go
hsinghkalsi: updated to use TagList() & FieldList()

plugins/outputs/newrelic/newrelic.go
hsinghkalsi: Updated code to identify and skip string case

plugins/outputs/newrelic/newrelic.go
hsinghkalsi: updated the comment to default set.

plugins/outputs/newrelic/newrelic_test.go
JimHagan: @hsinghkalsi Is there a way to make this test seem more deterministic.  If the connection will always fail when run as a unit test, it may make sense to omit the Connect() since it will always fail or comment somewhere that fail is expected in this scenario.

@danielnelson as a side note, we are currently working on a blog post to highlight a really nice use case of this new plugin.   If we can get this merged in the next couple of days, when would we possible see this plugin in a released version of Telegraf?

hsinghkalsi: @danielnelson @jimhagan skipped the test in short mode.

JimHagan:
@danielnelson,  @hsinghkalsi and I will meet up this morning to go over the concern about go routines and buffering -- I know we are using a standardized New Relic SDK for the output (which I think in general is a good thing, but I think we should be relying on Telegraf's buffering and flush configurations where possible).

hsinghkalsi:
@danielnelson, @JimHagan   I changed the code to set the HarvestPeriod to zero. HarvestPeriod controls how frequently data will be sent to New Relic. If HarvestPeriod is zero , then NewHarvester will not spawn a goroutine to send data and it is incumbent on the plugin to call  Harvester.HarvestNow when data should be sent.    
Invoking Harvester. HarvestNow() method along with a zero Config.HarvestPeriod control exactly when data is sent to New Relic servers. As we are invoking  HarvestNow() every time Write() function returns, we will not buffering any metrics.

Please let me know your thoughts on this. In the meantime I am working on adding more test cases.

plugins/outputs/newrelic/newrelic.go
```diff
@@ -40,7 +40,12 @@ func (nr *NewRelic) Connect() error {
 		return fmt.Errorf("InsightKey is a required for newrelic")
 	}
 	var err error
-	nr.harvestor, err = telemetry.NewHarvester(telemetry.ConfigAPIKey(nr.InsightsKey))
```
JimHagan: the benefit of using a `ConfigHarvestPeriod` of `0` is that the agent will not do any of its own buffer management of any kind.  This was an issue brought up by @danielnelson in the most recent review.

plugins/outputs/newrelic/newrelic.go
```diff
@@ -0,0 +1,123 @@
...
+					Value:      mvalue,
+					Name:       mname,
+					Attributes: tags})
+			}
+		}
+	}
+	// By default, the Harvester sends metrics and spans to the New Relic
+	// backend every 5 seconds.  You can force data to be sent at any time
+	// using HarvestNow.
+	nr.harvestor.HarvestNow(context.Background())
```
danielnelson, line 116: One thing we will want to ensure is that if the metrics are not sent successfully we return an error.  This will let Telegraf know that it should retry the batch.

plugins/outputs/newrelic/README.md
```diff
@@ -0,0 +1,17 @@
+#New Relic output plugin
+  
+This plugins writes to New Relic insights.
+
+```
+[[outputs.newrelic]]
+## New Relic Insights API key
+insights_key = "insights api key"
+##metric_prefix if defined, prefix's metrics name for easy identification
+# metric_prefix = "Telegraf_"
```
JimHagan, line 10: @hsinghkalsi I'd recommend making this "" by default.  At least that's what we decided for the use case we are trialing this form.

plugins/outputs/newrelic/newrelic.go
JimHagan: @hsinghkalsi can we make sure that we can return an error if the SDK code fails?

hsinghkalsi: @danielnelson  updated the code to return error in case any error is encountered while sending metrics

JimHagan: Thanks @hsingkalsi we are close.

JimHagan: @danielnelson I think we may be there

plugins/outputs/newrelic/newrelic.go
```diff
@@ -15,8 +17,9 @@ import (
 type NewRelic struct {
 	harvestor    *telemetry.Harvester
 	dc           *cumulative.DeltaCalculator
-	InsightsKey  string `toml:"insights_key"`
-	MetricPrefix string `toml:"metric_prefix"`
+	InsightsKey  string            `toml:"insights_key"`
```
JimHagan, line 20: @hsinghkalsi Were these whitespace changes intended?

hsinghkalsi: @JimHagan yes we added new configurable parameter for timeout

plugins/outputs/newrelic/newrelic.go
```diff
@@ -0,0 +1,147 @@
...
+	// backend every 5 seconds.  You can force data to be sent at any time
+	// using HarvestNow.
+	nr.harvestor.HarvestNow(context.Background())
+
+	//Check if we encountered errors
+	if len(nr.savedErrors) != 0 {
+		// we have errors, build error string
+		er := fmt.Sprintf("%#v", nr.savedErrors)
+		nr.savedErrors = nil
+		return fmt.Errorf("unable to harvest metrics  %s", er)
```
danielnelson, line 136: This isn't going to produce very nice output, there could be many errors and log message could be very long, and the map[string]interface{} may not be very readable.

I think it would be best to output only the first error, and to unpack the map into a more readable format, perhaps taking into account the possible fields that could be set.  This will help with troubleshooting as there won't be as many follow on errors.

plugins/outputs/newrelic/newrelic.go
```diff
@@ -0,0 +1,147 @@
...
+// Connect to the Output
+func (nr *NewRelic) Connect() error {
+	if nr.InsightsKey == "" {
+		return fmt.Errorf("InsightKey is a required for newrelic")
+	}
+	var err error
+	nr.harvestor, err = telemetry.NewHarvester(telemetry.ConfigAPIKey(nr.InsightsKey),
+		telemetry.ConfigHarvestPeriod(0),
+		func(cfg *telemetry.Config) {
+			cfg.Product = "NewRelic-Telgraf-Plugin"
```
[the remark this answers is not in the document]
hsinghkalsi: @danielnelson  thanks for pointing this out, fixed

plugins/outputs/newrelic/newrelic.go
hsinghkalsi: @danielnelson  updated the error formats to return the latest error. The errors are much precise. Also I have update test cases to run successfully in short mode.

JimHagan: @hsinghkalsi can you confirm that the short tests will work without internet connection per Dans comment `Make sure the short tests can pass without network access.`

hsinghkalsi: @JimHagan before submitting I tested use cases successfully,  with and without network

JimHagan: @danielnelson It seems like we are in good shape here.  Is there a chance we can merge this week?

plugins/outputs/newrelic/newrelic.go
```diff
@@ -0,0 +1,155 @@
...
+		return fmt.Errorf("unable to connect to newrelic %v", err)
+	}
+
+	nr.dc = cumulative.NewDeltaCalculator()
+	return nil
+}
+
+// Close any connections to the Output
+func (nr *NewRelic) Close() error {
+	nr.harvestor = nil
```
danielnelson, line 77: No need to nil the struct fields, after Close is called the instance will not be called again.  It would be nice if we could close the transport at this point.  With the http.Client you would call https://golang.org/pkg/net/http/#Client.CloseIdleConnections

plugins/outputs/newrelic/newrelic_test.go
```diff
@@ -0,0 +1,134 @@
...
+	nr := &NewRelic{
+		MetricPrefix: "Test",
+		InsightsKey:  "12345",
+		Timeout:      internal.Duration{Duration: time.Second * 5},
+	}
+
+	err := nr.Connect()
+	require.NoError(t, err)
+
+	err = nr.Write(testutil.MockMetrics())
```
danielnelson, line 35: This test makes a connection to newrelics servers, so it would need to be skipped on -short.  However the big issue here is that this test doesn't test anything specifically.

plugins/outputs/newrelic/newrelic_test.go
```diff
@@ -0,0 +1,134 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			nr := &NewRelic{
+				harvestor:    tt.fields.harvestor,
+				InsightsKey:  tt.fields.InsightsKey,
+				MetricPrefix: tt.fields.MetricPrefix,
+			}
+			if err := nr.Write(tt.metrics); (err != nil) != tt.wantErr {
+				t.Errorf("NewRelic.Write() error = %v, wantErr %v", err, tt.wantErr)
+			}
```
danielnelson, line 77: It seems that this test doesn't have any proper assertions, the only way it can fail is if Write panics.  Each test should exercise a specific error condition or verify a particular behavior.

plugins/outputs/newrelic/newrelic.go
```diff
@@ -0,0 +1,155 @@
...
+			case float64:
+				mvalue = float64(n)
+			case bool:
+				mvalue = float64(0)
+				if n {
+					mvalue = float64(1)
+				}
+			case string:
+				// Do not log everytime we encounter string
+				// we just skip
```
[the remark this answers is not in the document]
hsinghkalsi: fixed

plugins/outputs/newrelic/newrelic_test.go
```diff
@@ -0,0 +1,134 @@
...
+		})
+	}
+}
+
+func TestNewRelic_Connect(t *testing.T) {
+	//if testing.Short() {
+	////	t.Skip("skipping test in short mode.")
+	//}
```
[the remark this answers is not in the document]
hsinghkalsi: removed the code

plugins/outputs/newrelic/newrelic.go
hsinghkalsi: Explicitly added the http client so we can control the connection closure

plugins/outputs/newrelic/newrelic_test.go
hsinghkalsi: Skipped the test case, asserted the error to contain a particular string, this is more of a negative test case

plugins/outputs/newrelic/newrelic_test.go
hsinghkalsi: Rewrote this test case to validate the metrics string that is posted to New relic backend. The test now test various datatypes that can be present in the telegraf metrics type

hsinghkalsi:
@danielnelson Thanks for the code review once again. I did make the changes as suggested to the core code and added new test cases that verify the metrics string that is posted to New relic backend. The test now test various datatypes that we expect in the telegraf metrics type 
cc @JimHagan

JimHagan:
Hi @danielnelson please let us know if we are on the right track w/ the latest test updates.

plugins/outputs/newrelic/newrelic.go
```diff
@@ -0,0 +1,159 @@
...
```
ssoroka, line 120: why discard strings? New Relic seems to support them just fine. If it's about saving space, this might be better off as a configuration option.

hsinghkalsi: @ssoroka  newrelic-telemetry-sdk supports float64 as value for the metric, so we are converting each value to float64.  We do add tag values in the attributes for the metric. Since strings are not supported as metric values we can document it in the README and revisit once it is supported by new relic
https://godoc.org/github.com/newrelic/newrelic-telemetry-sdk-go/telemetry#Gauge.

ssoroka: Why not just store them as part of the attributes?

hsinghkalsi: Attributes in New relic are similar to Tags in the Telegraf. If you think we must, We can add them as attributes, the just would not have functionality related to metric visualization and querying.

JimHagan, line 120: @hsinghkalsi Could you append it to tags here if it's a string?

ssoroka: No, just curious. 👍

JimHagan: Consider this for a future refactor after we get feedback from the initial users.

hsinghkalsi: 👍

PR #7024, opened 2020-02-14
 net_response plugin: Read up to 1024 bytes in tcp response 

jjbuchan:
Fixes https://jira.rax.io/browse/SALUS-803

# What

Rather than reading the first line returned from the tcp server, read up to 1024 bytes.

# How

`Read` was reading a partial response most of the time which caused tests to fail intermittently.

When adding a 100ms wait before that step everything would be fine.  To avoid wait times I chose to do a `ReadAll` but to also avoid reading too much data I'm using a `LimitedReader`.  For any standard tcp response it'll read the whole thing and for any large response it'll only read up to 1024 bytes (which is the value i set).

# Why

This allows us to create monitors that match of the string response that comes after the `send` string.

e.g.
For this connection we could set the `expect` to `blocked_clients:0` and it would work.
```
$ telnet mycustomerservice.com 3125
Trying mycustomerservice.com...
Connected to mycustomerservice.com.
Escape character is '^]'.

> info
connected_clients:18
client_longest_output_list:0
client_biggest_input_buf:0
blocked_clients:0
```

It could potentially be used for a basic smtp monitor too by matching on `250`.

```
$ telnet mysmtptest.com 25

Trying mysmtptest.com...
Connected to mysmtptest.com.
Escape character is '^]'.
220 mysmtptest.com ESMTP Postfix

> ehlo me.com
250-mysmtptest.com
250-PIPELINING
250-SIZE 10240000
250-VRFY
250-ETRN
250-AUTH PLAIN LOGIN
250-ENHANCEDSTATUSCODES
250-8BITMIME
250 DSN

> quit
221 2.0.0 Bye
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -124,7 +124,6 @@ require (
 	golang.org/x/net v0.0.0-20191004110552-13f9640d40b9
 	golang.org/x/oauth2 v0.0.0-20190226205417-e64efc72b421
 	golang.org/x/sys v0.0.0-20190826190057-c7b8b68b1456
-	golang.org/x/tools v0.0.0-20190425150028-36563e24a262
 	gonum.org/v1/gonum v0.6.2 // indirect
 	google.golang.org/api v0.3.1
 	google.golang.org/genproto v0.0.0-20190404172233-64821d5d2107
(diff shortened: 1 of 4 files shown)


jjbuchan:
This was supposed to be on a local fork for now.

PR #7030, opened 2020-02-16
feat: add new binary data parser

### LINKED ISSUE #6804: Binary data parser feature request
## Feature Request

Binary Parser similar to the existing Value Parser with a possibility of parsing binary-encoded data (records) with multiple fields.

### Proposal:

Implement a Binary Parser able to parse binary-encoded data (records) containing multiple fields.

A working prototype for raw binary records is here: https://github.com/yolkhovyy/telegraf/tree/develop/plugins/parsers/bindata. At the end the Binary Parser will also support other binary data encoding protocols such as Protobuf or CBOR.

The binary data (record)  parser configuration is shown below, where record's fields are specified by name, type, offset and optional size. Also common parameters such as endianess, binary protocol and time field format are specified:
```toml
[[inputs.mqtt_consumer]]
  name_override = "drone_status"

  ...

  data_format = "bindata"
  bindata_protocol = "raw"
  bindata_endiannes = "be"
  bindata_time_format = "unix"
  bindata_fields = [
    {name="version",type="uint16",offset=0},
    {name="time",type="int32",offset=2},
    {name="location_latitude",type="float64",offset=6},
    {name="location_longitude",type="float64",offset=14},
    {name="location_altitude",type="float32",offset=22},
    {name="orientation_heading",type="float32",offset=26},
    {name="orientation_elevation",type="float32",offset=30},
    {name="orientation_bank",type="float32",offset=34},
    {name="speed_ground",type="float32",offset=38},
    {name="speed_air",type="float32",offset=42,size=4},
    {name="status",type="string",offset=46,size=10},
  ]
```
### Current behavior:

Only a single scalar value parsing is supported - by the Value Parser.

### Desired behavior:

Be able to parse binary records, raw and in the future - Protobuf, CBOR. 

### Use case: 

Parsing binary-encoded data from IoT and other domains.

yolkhovyy:
### Required for all PRs:

This is an implementation of feature request #6804 - binary data parser.

closes #6804

+ [x] Signed [CLA](https://influxdata.com/community/cla/).
+ [x] Associated README.md updated.
+ [x] Has appropriate unit tests.


plugins/parsers/bindata/README.md
```diff
@@ -0,0 +1,50 @@
...
+  ## Timestamp format - "unix", "unix_ms", "unix_us", "unix_ns"
+  bindata_time_format = "unix"
+
+  ## String encoding - "UTF-8" is the default
+  bindata_string_encoding = "UTF-8"
+
+  ## Binary data descriptor
+  ## Fields are described by:
+  ## - name - field name, use Golang naming conventions
+  ## - type - field type, use Golang type names
```
srebhan, line 34: Does this also support `bool`? If so, do you expect one byte?

plugins/parsers/bindata/parser.go
```diff
@@ -0,0 +1,205 @@
...
+
+const timeKey = "time"
+const timezone = "UTC"
+const defaultStringEncoding = "UTF-8"
+
+// Field is a binary data field descriptor
+type Field struct {
+	Name string
+	Type string
+	Size uint
```
srebhan, line 25: Please make them private.

plugins/parsers/bindata/parser.go
```diff
@@ -0,0 +1,205 @@
...
+}
+
+// BinData is a binary data parser
+type BinData struct {
+	MetricName     string
+	TimeFormat     string
+	Endiannes      string
+	byteOrder      binary.ByteOrder
+	StringEncoding string
+	Fields         []Field
```
srebhan, line 35: Please make them private.

plugins/parsers/bindata/README.md
```diff
@@ -0,0 +1,50 @@
...
+  data_format = "bindata"
+
+  ## Numeric fields endiannes, "be" or "le"
+  bindata_endiannes = "be"
+
+  ## Timestamp format - "unix", "unix_ms", "unix_us", "unix_ns"
+  bindata_time_format = "unix"
+
+  ## String encoding - "UTF-8" is the default
+  bindata_string_encoding = "UTF-8"
```
srebhan, line 29: Please comment default values.

plugins/parsers/bindata/README.md
```diff
@@ -0,0 +1,50 @@
...
+  ## Each data format has its own unique set of configuration options, read
+  ## more about them here:
+  ## https://github.com/influxdata/telegraf/blob/master/docs/DATA_FORMATS_INPUT.md
+  data_format = "bindata"
+
+  ## Numeric fields endiannes, "be" or "le"
+  bindata_endiannes = "be"
+
+  ## Timestamp format - "unix", "unix_ms", "unix_us", "unix_ns"
+  bindata_time_format = "unix"
```
srebhan, line 26: Should we make `unix` the default or do we need to force the user to specify one?

plugins/parsers/bindata/README.md
```diff
@@ -0,0 +1,50 @@
...
+  name_override = "drone_status"
+
+  ## Data format to consume.
+  ## Each data format has its own unique set of configuration options, read
+  ## more about them here:
+  ## https://github.com/influxdata/telegraf/blob/master/docs/DATA_FORMATS_INPUT.md
+  data_format = "bindata"
+
+  ## Numeric fields endiannes, "be" or "le"
+  bindata_endiannes = "be"
```
srebhan, line 23: How about adding `machine` and make it the default? That is, by default use the endianess of the machine that we have to detect.

plugins/parsers/bindata/parser.go
```diff
@@ -0,0 +1,205 @@
...
+		}
+
+		if fields[i].Type != "padding" {
+			// Check for duplicate field names
+			fieldName := fields[i].Name
+			for j := i + 1; j < len(fields); j++ {
+				if fieldName == fields[j].Name {
+					return nil, fmt.Errorf(`duplicate field name %s`, fieldName)
+				}
+			}
```
srebhan, line 91: How about you add
```
knownFields := make(map[string]bool)
```
before the above for loop (line 77) and use
```suggestion
			fieldName := fields[i].Name
			if _, ok := knownFields[fieldName]; ok {
				return nil, fmt.Errorf(`duplicate field name %s`, fieldName)
			}
```

plugins/parsers/bindata/parser.go
```diff
@@ -0,0 +1,205 @@
...
+	reader := io.NewSectionReader(bytes.NewReader(data), 0, int64(len(data)))
+	var offset int64 = 0
+
+	for _, field := range binData.Fields {
+		if field.Type != "padding" {
+			fieldBuffer := make([]byte, field.Size)
+
+			if _, err := reader.ReadAt(fieldBuffer, offset); err != nil {
+				return nil, err
+			}
```
srebhan, line 129: Not sure we need the reader. As we need sequentially we could also extract slices from data directly like
```suggestion
			fieldBuffer := data[offset:offset+field.Size]
```

plugins/parsers/bindata/parser.go
```diff
@@ -0,0 +1,205 @@
...
+
+			fieldType, ok := fieldTypes[strings.ToLower(field.Type)]
```
srebhan, line 131: Please do the `strings.ToLower()` part in the `NewBinDataParser()` function to avoid doing it in the "hot-path".

plugins/parsers/bindata/parser.go
```diff
@@ -0,0 +1,205 @@
...
+			if !ok {
+				return nil, fmt.Errorf(`invalid field type %s`, field.Type)
+			}
+
+			switch fieldType.Name() {
+			case "string":
+				fields[field.Name] = string(fieldBuffer)
```
srebhan, line 138: What about the string encoding?

plugins/parsers/bindata/parser.go
```diff
@@ -0,0 +1,205 @@
...
+			if _, err := reader.ReadAt(fieldBuffer, offset); err != nil {
+				return nil, err
+			}
+
+			fieldType, ok := fieldTypes[strings.ToLower(field.Type)]
+			if !ok {
+				return nil, fmt.Errorf(`invalid field type %s`, field.Type)
+			}
+
+			switch fieldType.Name() {
```
srebhan, line 136: Isn't this equivalent to `field.Type`?

plugins/parsers/bindata/parser.go
srebhan: Furthermore, we did the test already in the `NewBinDataParser()` function, but if you feel better doing a double-check I'm fine with this.

plugins/parsers/bindata/parser.go
```diff
@@ -0,0 +1,205 @@
...
+func (binData *BinData) SetDefaultTags(tags map[string]string) {
+	binData.DefaultTags = tags
+}
+
+// Parse implements Parser.Parse()
+func (binData *BinData) Parse(data []byte) ([]telegraf.Metric, error) {
+
+	fields := make(map[string]interface{})
+	reader := io.NewSectionReader(bytes.NewReader(data), 0, int64(len(data)))
+	var offset int64 = 0
```
srebhan, line 121: Can we make this an unsigned!?

plugins/parsers/bindata/parser.go
srebhan: You also might want to check if the start (`offset`) and end (`offset+field.Size`) is within the delivered data.

plugins/parsers/bindata/parser.go
```diff
@@ -0,0 +1,205 @@
...
+		offset += int64(field.Size)
+	}
+
+	metricTime, err := binData.getTime(fields)
+	if err != nil {
+		return nil, err
+	}
+
+	metric, err := metric.New(binData.MetricName, binData.DefaultTags,
+		fields, metricTime)
```
srebhan, line 155: Please don't do a line-break here... :-)

plugins/parsers/bindata/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+  ## Timestamp format - "unix", "unix_ms", "unix_us", "unix_ns"
+  bindata_time_format = "unix"
+
+  ## String encoding - "UTF-8" is the default
+  bindata_string_encoding = "UTF-8"
+
+  ## Binary data descriptor
+  ## Fields are described by:
+  ## - name - field name, use Golang naming conventions
```
srebhan, line 33: In the code it seems you assign a special meaning to the `time` field name (i.e. this is the time for the metric). Either put a big comment here that this is the case **or** better introduce a option called `bindata_time_field = "time"` allowing you to specify the field containing the metric time. If you want to be especially nice, add `bindata_drop_time_field = true` to allow the user to drop the used time field.

plugins/parsers/bindata/parser.go
```diff
@@ -0,0 +1,205 @@
...
+	"int16":   reflect.TypeOf((*int16)(nil)).Elem(),
+	"uint32":  reflect.TypeOf((*uint32)(nil)).Elem(),
+	"int32":   reflect.TypeOf((*int32)(nil)).Elem(),
+	"uint64":  reflect.TypeOf((*uint64)(nil)).Elem(),
+	"int64":   reflect.TypeOf((*int64)(nil)).Elem(),
+	"float32": reflect.TypeOf((*float32)(nil)).Elem(),
+	"float64": reflect.TypeOf((*float64)(nil)).Elem(),
+	"string":  reflect.TypeOf((*string)(nil)).Elem(),
+	"padding": reflect.TypeOf((*[]byte)(nil)).Elem(),
+}
```
srebhan, line 183: Put this to the top of the file.

plugins/parsers/bindata/parser.go
srebhan: Furthermore, could we also support null-terminated strings i.e. variable length?

plugins/parsers/bindata/parser.go
```diff
@@ -0,0 +1,205 @@
...
+	timeValue := fields[timeKey]
+	if timeValue != nil {
+		var err error
+		switch binData.TimeFormat {
+		case "unix":
+			metricTime, err = internal.ParseTimestamp(binData.TimeFormat, int64(timeValue.(int32)), timezone)
+		case "unix_ms", "unix_us", "unix_ns":
+			metricTime, err = internal.ParseTimestamp(binData.TimeFormat, int64(timeValue.(int64)), timezone)
+		default:
+			return *nilTime, fmt.Errorf("invalid time format %s", binData.TimeFormat)
```
srebhan, line 197:
```suggestion
			return time.Time{}, fmt.Errorf("invalid time format %s", binData.TimeFormat)
```

plugins/parsers/bindata/parser.go
```diff
@@ -0,0 +1,205 @@
...
+
+func (binData *BinData) getTime(fields map[string]interface{}) (time.Time, error) {
+	nilTime := new(time.Time)
+	metricTime := time.Now()
+	timeValue := fields[timeKey]
+	if timeValue != nil {
+		var err error
+		switch binData.TimeFormat {
+		case "unix":
+			metricTime, err = internal.ParseTimestamp(binData.TimeFormat, int64(timeValue.(int32)), timezone)
```
srebhan, line 193: This will panic if the datatype of the `timeValue` field does not match what you expect here. The least line of defense would be IMO to check this in the `NewBinaryParser()` function by comparing the `type` of the `timeField` against the time-format.
But we here really want to check the type-assertion I think.

srebhan (changes requested):
@yolkhovyy this is very cool. I have some comments in the code. The main ones are the specification of the time field. IMO you should make this explicit with a sensible default e.g. "time" or empty == `time.Now()`.
As a (later) extension please think about bitfields which are very common for boolean flags in embedded devices.

srebhan:
@yolkhovyy any chance you work on this?

srebhan:
@yolkhovyy any news on this?

srebhan:

sjwang90:

plugins/parsers/bindata/README.md
yolkhovyy: Sure, can be done that way.

plugins/parsers/bindata/README.md
yolkhovyy: unix should be the default indeed

plugins/parsers/bindata/README.md
yolkhovyy: yes, it does - will add it in the example

plugins/parsers/bindata/README.md
yolkhovyy: But, what if the measurement arrives from a  different machine which has different than the endianness of the machine where Telegraf runs?

plugins/parsers/bindata/parser.go
yolkhovyy: good point about the reader, and the range check too

plugins/parsers/bindata/parser.go
yolkhovyy: good point

plugins/parsers/bindata/parser.go
yolkhovyy: will move ToLower to the NewBinDataParser and remove the type check here

plugins/parsers/bindata/parser.go
yolkhovyy: you're right, it will be more clear this way

plugins/parsers/bindata/README.md
yolkhovyy: done

plugins/parsers/bindata/README.md
yolkhovyy: I suggest to use `padding` for the case when the user wants to drop the time filed, and put that in the big comment.

plugins/parsers/bindata/parser.go
yolkhovyy: yes, it's better

plugins/parsers/bindata/parser.go
yolkhovyy: String encoding is specified in bindata_string_encoding.

yolkhovyy: null-terminated strings requires a bit of thinking.

plugins/parsers/bindata/parser.go
yolkhovyy: sure

plugins/parsers/bindata/parser.go
yolkhovyy: agree

plugins/parsers/bindata/parser.go
yolkhovyy: Added a check in the NewBinDataParser

yolkhovyy:
I have just implemented requested review changes, is it possible to re-open this pull request?

yolkhovyy:
@sjwang90 Could you please re-open this one? I wanted to complete it.

plugins/parsers/bindata/README.md
srebhan: Then you have to specify the endianness (which can be done). However, it is not uncommon to have machines with the same endianness, so a sensible default should be chosen IMO.

plugins/parsers/bindata/README.md
srebhan: Please comment all options that have a sensible default
```suggestion
  # bindata_time_format = "unix"
```

plugins/parsers/bindata/README.md
srebhan: If so, please also comment the present line in the sample-config...
```suggestion
  # bindata_string_encoding = "UTF-8"
```

plugins/parsers/bindata/README.md
srebhan: Also fine with me.

plugins/parsers/bindata/README.md
```diff
@@ -0,0 +1,55 @@
...
+    {name="Longitude",type="float64"},
+    {name="Altitude",type="float32"},
+    {name="Heading",type="float32"},
+    {name="Elevation",type="float32"},
+    {name="Bank",type="float32"},
+    {name="GroundSpeed",type="float32"},
+    {name="AirSpeed",type="float32"},
+    {name="None",type="padding", size=16},
+    {name="Status",type="string",size=7},
+    {name="StatusOK",type="bool"},
```
srebhan, line 53: The more I think about it, the more I wish you introduced a `omit` boolean flag to allow the user to drop the field. The background is the following: Often you want all fields in here for documentation and easy on/off switching of data e.g. if your use-case requirements change etc...
```suggestion
  ## Use padding when auto-generated metric time is preferred.
  bindata_fields = [
    {name="Version",type="uint16"},
    {name="Time",type="int32"},
    {name="Latitude",type="float64"},
    {name="Longitude",type="float64"},
    {name="Altitude",type="float32"},
    {name="Heading",type="float32"},
    {name="Elevation",type="float32"},
    {name="Bank",type="float32", omit=true},
    {name="GroundSpeed",type="float32"},
    {name="AirSpeed",type="float32"},
    {name="",  type="padding", size=16, omit=true},
    {name="Status",type="string",size=7, omit=true},
    {name="OK",type="bool"},
```
 What do you think?

plugins/parsers/bindata/parser.go
```diff
@@ -0,0 +1,216 @@
...
+	}
+	stringEncoding = strings.ToUpper(stringEncoding)
+	if stringEncoding != defaultStringEncoding {
+		return nil, fmt.Errorf(`invalid string encoding %s`, stringEncoding)
+	}
+
+	// Field types, names and sizes
+	knownFields := make(map[string]bool)
+	for i := 0; i < len(fields); i++ {
+		fieldType, ok := fieldTypes[strings.ToLower(fields[i].Type)]
```
srebhan, line 99:
```suggestion
	for i, field := range fields {
		fieldType, ok := fieldTypes[strings.ToLower(field.Type)]
```

plugins/parsers/bindata/parser.go
```diff
@@ -0,0 +1,216 @@
...
+}
+
+// SetDefaultTags implements Parser.SetDefaultTags()
+func (binData *BinData) SetDefaultTags(tags map[string]string) {
+	binData.DefaultTags = tags
+}
+
+// Parse implements Parser.Parse()
+func (binData *BinData) Parse(data []byte) ([]telegraf.Metric, error) {
+
```
srebhan, line 149: Please remove this newline as it will not help in structuring the code.

plugins/parsers/bindata/parser.go
srebhan: Ok so I'll mark this as resolved, but we should keep this for later.

plugins/parsers/bindata/parser.go
```diff
@@ -0,0 +1,216 @@
...
+	fields := make(map[string]interface{})
+	var offset uint = 0
+
+	for _, field := range binData.fields {
```
srebhan, line 153:
```suggestion
	
	offset := uint(0)
	for _, field := range binData.fields {
```

plugins/parsers/bindata/parser.go
```diff
@@ -0,0 +1,216 @@
...
+			metricTime, err = internal.ParseTimestamp(binData.timeFormat, int64(timeValue.(int64)), timezone)
+		default:
+			return *nilTime, fmt.Errorf("invalid time format %s", binData.timeFormat)
+		}
+		if err != nil {
+			return *nilTime, err
+		}
+		delete(fields, timeKey)
+	}
+	return metricTime, nil
```
srebhan, line 215: This should be simplified a bit
```suggestion
	t, found := fields[timeKey]
	if !found {
		return time.Now()
	}
	delete(fields, timeKey)
	
	switch binData.timeFormat {
	case "unix":
		tval, ok := t.(int32)
		if !ok {
			return nil, fmt.Errorf("invalid time type %T, must be int32", t)
		}
		return internal.ParseTimestamp(binData.timeFormat, int64(tvalue), timezone)
	case "unix_ms", "unix_us", "unix_ns":
		tval, ok := t.(int64)
		if !ok {
			return nil, fmt.Errorf("invalid time type %T, must be int64", t)
		}
		return internal.ParseTimestamp(binData.timeFormat, int64(tval), timezone)
	}
	return nil, fmt.Errorf("invalid time format %q", binData.timeFormat)
```

plugins/parsers/bindata/parser_test.go
```diff
@@ -0,0 +1,574 @@
+package bindata
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
```
srebhan, line 6: Please get rid of `assert` and use `require` only.

plugins/parsers/bindata/parser_test.go
```diff
@@ -0,0 +1,574 @@
...
+var binaryDataBigEndianTimeS = []byte{
+	0x01, 0x6D, 0xD9, 0xE7,
+}
+
+var binaryDataBigEndianTimeXs = []byte{
+	0x00, 0x00, 0x01, 0x6D, 0xD9, 0xE7, 0x0B, 0x17,
+}
+
+func TestBigEndian(t *testing.T) {
+
```
srebhan, line 196: Please remove the newline(s) also for other tests!

plugins/parsers/bindata/parser_test.go
```diff
@@ -0,0 +1,574 @@
...
+	var parser, err = NewBinDataParser(
```
srebhan, line 197:
```suggestion
	parser, err := NewBinDataParser(
```

plugins/parsers/bindata/parser_test.go
```diff
@@ -0,0 +1,574 @@
...
+		"big_endian",
+		"unix",
+		"be",
+		"utf-8",
+		fields,
+		nil,
+	)
+	require.NoError(t, err)
+	assert.NotNil(t, parser)
```
srebhan, line 206:
```suggestion
	require.NotNil(t, parser)
```

plugins/parsers/bindata/parser_test.go
```diff
@@ -0,0 +1,574 @@
...
+
+	metrics, err := parser.Parse(binaryDataBigEndian)
+	require.NoError(t, err)
+	assert.Len(t, metrics, 1)
+	require.Equal(t, parser.metricName, metrics[0].Name())
+	assert.Equal(t, int64(0x5DA86C4C), metrics[0].Time().Unix())
+	assert.Equal(t, expectedParseResult, metrics[0].Fields())
```
srebhan, line 213:
```suggestion
	require.Len(t, metrics, 1)
	require.Equal(t, parser.metricName, metrics[0].Name())
	require.Equal(t, int64(0x5DA86C4C), metrics[0].Time().Unix())
	require.Equal(t, expectedParseResult, metrics[0].Fields())
```

plugins/parsers/bindata/parser_test.go
```diff
@@ -0,0 +1,574 @@
...
+}
+
+func TestLittleEndian(t *testing.T) {
```
srebhan, line 216: Please also apply the examples above to all following tests.

srebhan (commented):
Thanks @yolkhovyy for the nice update. There are some more comments but the code looks much better now. One thing I want you to think about is to use a `switch` statement instead of using reflection for doing the type-conversions. I guess the former is easier to read...

Please also rebase this PR on latest master to resolve the merge conflicts and get the latest CI testing.

plugins/parsers/bindata/README.md
yolkhovyy: I think the initial idea was to use padding to drop fields, omit would add more flexibility and good for documenting indeed.

plugins/parsers/bindata/parser.go
yolkhovyy: I think this away each `field` will be copied in a new var, I am fine with that - code reads better then, but at the end it will be slightly worse performance.

plugins/parsers/bindata/README.md
yolkhovyy: "be" is the default, and I commented the line out in README.md but can't see the update here.

plugins/parsers/bindata/parser_test.go
yolkhovyy: Done

plugins/parsers/bindata/README.md
yolkhovyy: @srebhan There was a comment from you to support null-terminated strings - I will look at how to do that next week.

yolkhovyy:
@srebhan I am also not sure what this mean:
Semantic Pull Request — add a semantic commit or PR title 
Could you please elaborate on this? Thanks

powersj:
> @srebhan I am also not sure what this mean: Semantic Pull Request — add a semantic commit or PR title Could you please elaborate on this? Thanks

It is looking for a commit message like this:

`feat: add new binary data parser`

plugins/parsers/bindata/README.md
```diff
@@ -0,0 +1,55 @@
...
+  name_override = "drone_status"
+
+  ## Data format to consume.
+  ## Each data format has its own unique set of configuration options, read
+  ## more about them here:
+  ## https://github.com/influxdata/telegraf/blob/master/docs/DATA_FORMATS_INPUT.md
+  data_format = "bindata"
+
+  ## Numeric fields endiannes, "be" or "le", default "be"
+  # bindata_endiannes = "be"
```
srebhan, line 23: I think the default should be `host`, i.e. do not mess with endianess?

plugins/parsers/bindata/README.md
srebhan: I think it's save to assume UTF-8 as a default encoding isn't it? If so, please comment this line as shown above.

plugins/parsers/bindata/README.md
srebhan: > I think the initial idea was to use padding to drop fields, omit would add more flexibility and good for documenting indeed. Shall we do that this time?

We need to add it now as the config format needs to be kept backward compatible, so once it's out you carry that for some time (at least until the major release)...

plugins/parsers/bindata/parser.go
```diff
@@ -0,0 +1,216 @@
...
+}
+
+// BinData is a binary data parser
+type BinData struct {
+	metricName     string
+	timeFormat     string
+	endiannes      string
+	byteOrder      binary.ByteOrder
+	stringEncoding string
+	fields         []Field
```
srebhan, line 35: Please export these fields. Please also add toml-tags to those options to allow for the new parser format (see PR #8791 and the CSV parser as an example).

plugins/parsers/bindata/parser.go
```diff
@@ -0,0 +1,216 @@
...
+) (*BinData, error) {
+
+	// Time format
+	switch timeFormat {
+	case "":
+		timeFormat = defaultTimeFormat
+	case "unix", "unix_ms", "unix_us", "unix_ns":
+	default:
+		return nil, fmt.Errorf("invalid time format %s", timeFormat)
+	}
```
srebhan, line 73: Please fuse this check with the time-field check below.

plugins/parsers/bindata/parser.go
```diff
@@ -0,0 +1,216 @@
...
+			switch field.Type {
+			case "string":
+				fields[field.Name] = string(fieldBuffer)
+			default:
+				fieldValue := reflect.New(fieldTypes[field.Type])
+				byteReader := bytes.NewReader(fieldBuffer)
+				binary.Read(byteReader, binData.byteOrder, fieldValue.Interface())
+				fields[field.Name] = fieldValue.Elem().Interface()
+			}
+		}
```
srebhan, line 172: To be honest I'd like to see something like
```suggestion
		switch field.Type {
		case "padding":
			continue
		case "bool":
			var v bool
			r := bytes.NewReader(data[offset : offset+1])
			if err := binary.Read(r, binData.byteOrder, &v); err != nil {
				return nil, err
			}
			fields[field.Name] = v
		case "uint8":
			var v uint8
			r := bytes.NewReader(data[offset : offset+1])
			if err := binary.Read(r, binData.byteOrder, &v); err != nil {
				return nil, err
			}
			fields[field.Name] = v
		case "int8":
			...
		case "uint16":
			...
		case "int16":
			...
		case "uint32":
			...
		case "int32":
			...
		case "uint64":
			...
		case "int64":
			...
		case "float32":
			...
		case "float64":
			...
		case "string":
			fields[field.Name] = string(data[offset:offset+field.Size)
		}
```

plugins/parsers/bindata/parser.go
```diff
@@ -0,0 +1,216 @@
...
+		return nil, err
+	}
+
+	return []telegraf.Metric{metric}, err
+}
+
+// ParseLine implements Parser.ParseLine()
+func (binData *BinData) ParseLine(line string) (telegraf.Metric, error) {
+	return nil, fmt.Errorf("BinData.ParseLine() not supported")
+}
```
srebhan, line 192: Why, you could just call `Parse()` with `[]byte(line)`?

srebhan (commented):
Thanks @yolkhovyy for the nice update! I have some more comments, but we are definitively coming closer...

yolkhovyy:
> @yolkhovyy this is very cool. I have some comments in the code. The main ones are the specification of the time field. IMO you should make this explicit with a sensible default e.g. "time" or empty == `time.Now()`. As a (later) extension please think about bitfields which are very common for boolean flags in embedded devices.

@srebhan I am not sure about your comment about the "time" field - if it is not present in the spec, it will be initialized to time.Now() at line 194:

```golang
func (binData *BinData) getTime(fields map[string]interface{}) (time.Time, error) {
	t, found := fields[timeKey]
	if !found {
		return time.Now(), nil
	}
	delete(fields, timeKey)
...
}

plugins/parsers/bindata/parser.go
yolkhovyy: Totally agree with padding in switch. Regarding reflection - I though it was cool :) is reflection against IndluxData style/rules?

plugins/parsers/bindata/README.md
yolkhovyy: I see

plugins/parsers/bindata/parser.go
srebhan: Well it's probably a matter of taste, but I think the switch/case above is much more readable/understandable compared to reflection. There also might be a performance impact, but that's not my primary concern... So please switch to the switch-statement. :-)

srebhan: Regarding the null-terminated strings. How about, if a length is given, we respect this length, otherwise we go for null-termination. This would allow to read non-null-terminated strings (i.e. fixed length fields) which you sometimes see in embedded devices.

srebhan:
@yolkhovyy not sure anymore about my comment regarding time. I think my idea was to explicitly add a user setting saying in which field the time is to be expected instead of **implicitly** using a field named `"Time"`. One reason is case-sensitivity and the fact that the user "needs to know" this. If there is an option `# time = ""` in TOML, i.e. an option to specify the time field **explicitly** I would feel better. The logic is then, if this option is empty (default) we are using `time.Now` otherwise we use the given field and error out if that field does not exist.
What do you think?

plugins/parsers/bindata/README.md
yolkhovyy: @srebhan Could you please elaborate on this? Not sure whether I understand this comment.

plugins/parsers/bindata/README.md
yolkhovyy: @srebhan agree

plugins/parsers/bindata/README.md
srebhan: What I mean is that people tend to not knowing the endianess of data and that's fine as long as the do not cross endianess boarders. If you for example generate the data in a machine (with its endianess) and the parse it on that same machine, you probably do not need to know if that host has big- or little-endian. Therefore I suggest a shortcut for these use-cases like using `native` or `host` as endianess instead of explicitly defining it. Does that make things clearer?

plugins/parsers/bindata/parser.go
yolkhovyy: null-terminated strings - sounds good, good point

srebhan:
As discussed, we should take over this one...

srebhan:
Close in preference of #11552.

PR #7040, opened 2020-02-18
azure_iothub_consumer - Input plugin

chrishayles:
…crosoft fix.

### Required for all PRs:

To follow once code adjustments have been made following cert fix.


haylesnortal:
Known issues/required changes:

- Include tests for plugins. [Telegraf plugin code]
- More detailed README. [Telegraf plugin code]
- PR required for iothub module to introduce "iotmodule" type. [iothub module code]
- Move output plugin from "iotdevice" to "iotmodule." [Telegraf plugin code]
- Move input plugin from "iotservice" to "iotmodule." [Telegraf plugin code]
- Add other transport protocols to "iothub" module and create second pull request. [iothub module code]
- input plugin here needs to work with Telegraf parser and accumulator instead of current log output. [Telegraf plugin code]

danielnelson:
Can you also add some information about when this plugin would need to be used instead of #6928?

haylesnortal:
@danielnelson Hey Daniel, forwarded on an email I sent to your colleagues last month. Lookout for something from chris.hayles@nortal.com. 

Short version: the IoT Hubs has an "Event Hubs Compatible" endpoint, but that endpoint doesn't have anywhere near the functionality of establishing a device connection with the IoT Hub and won't make use of an edge setup.

plugins/outputs/azure_iothub/iothub/iotdevice/transport/transport.go
```diff
@@ -0,0 +1,52 @@
...
+	"context"
+	"crypto/tls"
+	"github.com/amenzhinsky/iothub/logger"
+	"time"
+
+	"github.com/amenzhinsky/iothub/common"
+)
+
+// Transport interface.
+type Transport interface {
```
haylesnortal: This interface is implemented by the ./mqtt/mqtt.go with module-specific client and functions in ./mqtt/module_mqtt.go and this establishes a connection with the edgeHub/iotedge runtime. The edgeHub then connects to the Azure IoT Hub service via AMQP 1.0 in the scenario tested.

Next step would be to include an AMQP connection to the edgeHub. It would need to conform to the interface here. 

Note: Some of these functions can't be implemented for a module connection. For example, direct methods don't currently work for edge devices/modules. Any errors when calling this on an edge client are expected.

plugins/outputs/azure_iothub/iothub/iotdevice/credentials.go
```diff
@@ -0,0 +1,113 @@
+package iotdevice
```
haylesnortal: In this file the two existing credentials types are implemented. There is another authentication workflow for modules that this code does not implement, but would be a valuable addition. 

Currently the SharedAccessKey credential requires a connection string, or at least a shared access key, in order to authenticate. It does this by creating a token based on the endpoint being accessed (audience), expiry time, and signing it all with the shared access key provided. 

Modules running on the edge can authenticate without the shared access key by creating the token using the same properties but getting it all signed via the workload API /sign endpoint. 

In order to get the token response back from the sign endpoint, the edge runtime is validating the module and generation ID of the running module against what it expects to be running (controlled by the edgeAgent) and so security is provided that way. 

In short, this would allow for telegraf to be deployed without providing a connection string in the configuration. However, some configuration would still be necessary for input/output options.

haylesnortal:
Fixed dependency issue, removed iothub library codebase to reflect this. Cleaned up the metric creation.

srebhan:
@chrishayles is this still a thing?

Issue #6970, opened 2020-02-02
Set the measurement name to a tag value


JollyJosy:
## Feature Request

telegraf to set the measurement name equal to the filename if file input is used

### Proposal:

it would be helpful if the input parameter "name_override" have a "filename" argument.
Similar to the "file_tag"parameter of the file input plugin.

### Current behavior:

The "name_override" allows to override the measurement name with a static name.

### Desired behavior:

The "name_override" allows to override the measurement name with the filename of the file used as the input.

### Use case: 

I have multiple servers producing measurements. Each type of measurement (including different fields) is having it's own filename.
When the filename can be used as measurement name it would allow for a single input section per server using the wildcard for the filenames.

danielnelson:
Usually the data is more flexibly queried if you use fewer measurement name, so I do recommend using the tag to store the file name, and setting the measurement to a value that describes the particular shape of data.  Generally this means that each plugin definition has a single name which works out with `name_override`.

However, we could add the ability for the `converter` processor to convert a tag/field value to the measurement name.

mvgoylen:
Having this possibility in the converter processor would be also a good solution for the problem . 
thanks already for this fast feedback.

DiddleDaddle:
Any idea when this could be added to the converter processor? This would really be a big help and avoid using external scripting to treat the files.

danielnelson:
@JollyJosy Will be in the upcoming 1.14 release and the next nightly build.

PR #7049, opened 2020-02-19
Add support for converting tag or field to measurement in converter processor

danielnelson:
closes: #6970

### Required for all PRs:

### DIFF
--- a/plugins/processors/converter/README.md
+++ b/plugins/processors/converter/README.md
@@ -9,7 +9,7 @@ Values that cannot be converted are dropped.
 uniquely identifiable.  Fields with the same series key (measurement + tags)
 will overwrite one another.
 
-### Configuration:
+### Configuration
 ```toml
 # Convert values to another metric value type
 [[processors.converter]]
@@ -19,6 +19,7 @@ will overwrite one another.
   ## select the keys to convert.  The array may contain globs.
   ##   <target-type> = [<tag-key>...]
   [processors.converter.tags]
+    measurement = []
     string = []
     integer = []
     unsigned = []
@@ -31,6 +32,7 @@ will overwrite one another.
   ## select the keys to convert.  The array may contain globs.
   ##   <target-type> = [<field-key>...]
   [processors.converter.fields]
+    measurement = []
     tag = []
     string = []
     integer = []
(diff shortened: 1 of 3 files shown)


plugins/processors/converter/README.md
```diff
@@ -39,19 +41,40 @@ will overwrite one another.
...
+Convert `port` tag to a string field:
 ```toml
 [[processors.converter]]
   [processors.converter.tags]
     string = ["port"]
+```
+
+```diff
+- apache,port=80,server=debian-stretch-apache BusyWorkers=1,BytesPerReq=0
++ apache,server=debian-stretch-apache port="80",BusyWorkers=1,BytesPerReq=0
```
ssoroka: should these be in the same order to be more clear?

plugins/processors/converter/converter.go
```diff
@@ -39,30 +40,32 @@ var sampleConfig = `
...
+	Boolean     []string `toml:"boolean"`
+	Float       []string `toml:"float"`
 }
 
 type Converter struct {
-	Tags   *Conversion `toml:"tags"`
-	Fields *Conversion `toml:"fields"`
+	Tags   *Conversion     `toml:"tags"`
+	Fields *Conversion     `toml:"fields"`
+	Log    telegraf.Logger `toml:"-"`
```
ssoroka: should log be private?

plugins/processors/converter/converter.go
```diff
@@ -73,15 +76,11 @@ func (p *Converter) Description() string {
...
 
-func (p *Converter) Apply(metrics ...telegraf.Metric) []telegraf.Metric {
-	if !p.initialized {
-		err := p.compile()
-		if err != nil {
-			logPrintf("initialization error: %v\n", err)
-			return metrics
-		}
-	}
+func (p *Converter) Init() error {
```
ssoroka: is this a contract change that we need to worry about? previously Apply would init for you.

plugins/processors/converter/converter_test.go
```diff
@@ -451,41 +402,123 @@ func TestConverter(t *testing.T) {
...
+					"msg": "howdy",
+				},
+				time.Unix(0, 0),
 			),
-			expected: Metric(
-				metric.New(
-					"cpu",
+			expected: []telegraf.Metric{
+				testutil.MustMetric(
+					"/var/log/syslog",
```
reimda: #6970 talks about using filenames as the measurement name.  I don't think they'll want to use the full path.  Is there a way to get just the basename of the file input plugin's file_tag?

plugins/processors/converter/README.md
danielnelson: Yeah I'll update this, it should read better if I move the port to the end of the fields:
```
+ apache,server=debian-stretch-apache BusyWorkers=1,BytesPerReq=0,port="80"
```

plugins/processors/converter/converter.go
danielnelson: It does needs to be exported, which is why we mark it unsettable from the toml.  This is due to how we inject the Logger from outside the package (which we have never been super happy with).  When adding this we also toyed around with injecting the logger on init: `Init(Logger) error` but ultimately decided not to inject it there.

plugins/processors/converter/converter.go
danielnelson: The Init() function is optional on any plugin type, but will be called if defined.  It makes it a little bit easier to initialize a plugin without needing the manage it and improves the plugins ability to report errors with the configuration was setup improperly.

plugins/processors/converter/converter_test.go
danielnelson: You would need to use another processor to modify the tag value first, it ought to be possible to do this with the regex processor though.  We could potentially have a `filepath` processor that does common transformations, or perhaps it would fit in with the `strings` processor.

PR #7060, opened 2020-02-21
Missing perf counters fix

### LINKED ISSUE #7059: SqlServer Input - Missing "Batch Resp Statistics" for named instances
Soon I will open a PR with the fix

### Relevant telegraf.conf:

```toml
[[inputs.sqlserver]]
  servers = [
    "Server=__AnyNamedInstance__;User Id=<user>;Password=<pw>;app name=telegraf;log=1;",
  ]

  query_version = 2
  ## Just Don't exclude 'PerformanceCounters'
  exclude_query = [ 'Schedulers' , 'SqlRequests']
```

### System info:
Not relevant, it happens on every (on-prem) named instance, regardless of Windows server or SQL Server version (from SQL Server 2012 and above)

### Steps to reproduce:

Compare the data gathered with the plugin, between a default instance and a named instance, i.e:
- QDSRVMONITOR → This is a default instance, its hidden name is "MSSQLSERVER" but does not need to be specified
- QDSRVMONITOR\SQL2017 → This is a named instance ("SQL2017" in this case)

In the measurement "sqlserver_performance", you will miss a series of tags for the SQL Instance, their name pattern is "Batches >= __value__ms & < __value__ms 

### Expected behavior:

Those counters are gathered also for named instances

### Actual behavior:
Those counters are not gathered

### Additional info:

This is due to a filter in the query, which excludes this counter for named instances
```sql
  {etc}
  ) OR (
    object_name LIKE 'SQLServer:Batch Resp Statistics%'
  ) OR {etc}
```

For default instances, the "object_name" will match, but named instances have a different value for "object_name", which is like this "MSSQL$<InstanceName>"

Trovalo:
### Required for all PRs:

Resolves #7059 
Fixed some filter conditions in the Performance Counters queries (both v1 and v2).
The column "object_name" has different values between default and named instance
- **Default Instance** - object_name = SQLServer:Batch Resp Statistics
- **Named Instance** - object_name = <Instancename>:Batch Resp Statistics

The fix (in its own commit) has been made also for the v1 query (the logic is exactly the same), but I have no way to test it since none of my databases gathers the affected "object_name", which are "%FileTable" and"%Backup Device".

### DIFF
--- a/plugins/inputs/sqlserver/sqlserver.go
+++ b/plugins/inputs/sqlserver/sqlserver.go
@@ -632,7 +632,7 @@ WHERE	(
 			object_name LIKE '%User Settable%'
 			OR object_name LIKE '%SQL Errors%'
 		) OR (
-			object_name LIKE 'SQLServer:Batch Resp Statistics%'
+			object_name LIKE '%Batch Resp Statistics%'
 		) OR (
 			instance_name IN ('_Total')
 			AND counter_name IN (
--- a/plugins/inputs/sqlserver/sqlserver.go
+++ b/plugins/inputs/sqlserver/sqlserver.go
@@ -2302,7 +2302,7 @@ SELECT DISTINCT RTrim(spi.object_name) object_name
 , spi.cntr_value
 , spi.cntr_type
 FROM sys.dm_os_performance_counters spi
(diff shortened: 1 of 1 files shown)


danielnelson:
@denzilribeiro Can you take a look?

PR #7068, opened 2020-02-23
Refactored docker_log input

i-prudnikov:
### Required for all PRs:

In continue for https://github.com/influxdata/telegraf/pull/6863, there were 3 major points in discussion. All of them are addressed:
1. Missing features from proposed separate plugin (in https://github.com/influxdata/telegraf/pull/6863) merged into docker_log plugin. To have single plugin that deliver docker logs.
2. Approach to store offset not in a file (as implemented in tail plugin) is not working here, since the point is to make offset persistent even if telegraf was shut down. Still implemented via saving offset in a file.
3. Functionality to stop telegraf wasn't implemented, instead we can execute command specified in config. This is enough to communicate with environment.

__Summary of features added:__
1. Throttling for gathering container logs by limiting amount of log data to be read from the
each docker container and interval of log's reading. This primary use case here is to be protected from the situation 
lots of log entries (stack-tracing for example). This can cause high CPU utilisation of telegraf, and in case there are several such containers on the host we can hit the CPU limit.
3. Use original docker timestamps - to get the precise picture.
4. Optimisation for running under k8s for streaming logs from the containers in a POD.
The main feature here is to execute command when all containers from which we are streaming logs are exited/terminated. This allows to terminate POD in a consistent way, after all logs are delivered.


danielnelson:
Thanks @i-prudnikov, but I'll have to get back to this one in April after 1.13 is finished, can you send me reminder then?

i-prudnikov:
Hello Daniel, sure will ping you again. What is the best way . -to open a new ticket or leave  comment in this . PR?

danielnelson:
Just comment here and I should get the notification.

i-prudnikov:
Hello Daniel, just a reminder about this PR, as discussed above.

i-prudnikov:
Hello again @danielnelson, a kind reminder.

danielnelson:
I think what we should do is open a several new issues/pull requests, each that address one of these specific functionality changes:
- Use docker timestamps as metric timestamp.
- Save docker timestamp as offset and resume from offset.
- Add rate limiting to tailStream, using https://godoc.org/golang.org/x/time/rate

I don't think we should integrate the custom read logic.  It is too much complex and error prone code when we can use Dockers stdcopy.StdCopy function, and the options to configure it expose too much internal implementation details for my preference.

i-prudnikov:
Hello Daniel, let's proceed with your suggestion. I will split the changes into small PRs.
Regarding the custom read logic, so far the reason for it was to:
1. Be able to parse time-stamps as they go as a part of header of every log message
2. Rate limiting not the frequency of the API call but the amount of data.

I would be happy to simplify the code and utilise Dockers stdcopy.StdCopy function as it really much easier and less error prone, but so far I see no possibility to implement the improvements without custom read logic. 
Let me dig around, and I will come back with the first PR (Use docker timestamps as metric timestamp), hopefully without custom read logic.

sjwang90:
Going to close this PR - split into #7434 and #7749

PR #7069, opened 2020-02-23
Make input/docker work with windows

hoerup:
Fix: #4666
### Required for all PRs:

### DIFF
--- a/etc/telegraf.conf
+++ b/etc/telegraf.conf
@@ -2262,6 +2262,7 @@
 #   ## Docker Endpoint
 #   ##   To use TCP, set endpoint = "tcp://[ip]:[port]"
 #   ##   To use environment variables (ie, docker-machine), set endpoint = "ENV"
+#   ##   To Use Windows npipe, set endpoint = "npipe:////./pipe/docker_engine"
 #   endpoint = "unix:///var/run/docker.sock"
 #
 #   ## Set to true to collect Swarm metrics(desired_replicas, running_replicas)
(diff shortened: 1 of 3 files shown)


danielnelson:
At one point we moved the version forward to 1.24, but it was pulled back to version 1.21 since it broke connections to the version of Docker included with Ubuntu 14. (#4434)  This version of Ubuntu, while somewhat old, is still on extended support until midway though 2022, so I'd still like to keep it working.

What would be ideal is if we can find a way to set the variable to work with both versions, perhaps using the api version negotation in the docker library.  If that is impossible, we could make this user configurable but I'd first like an attempt to be made to find a way where that is not required.

hoerup:
I haven't tested it yet, but would it be possible to, even though non-versioned requests are deprecated, first do a version-less client lookup to dockers "/version" api endpoint to discover min/max supported version- and then use that to determine the right api version for the rest of the session ?

danielnelson:
I think this would probably work, if not I imagine there is another way we can make this work with one of the version negotiation functions.  We should test any change against the version included with Ubuntu 14 before merging to be sure.

hoerup:
@danielnelson what can I do get this moving along ? update PR to remove version pinning, so we rely on https://github.com/moby/moby/blob/master/client/client.go#L200

danielnelson:
Could you research a solution that will work for both Windows and Ubuntu 14?  Remember it is important that it works even when querying across platforms: Windows -> Ubuntu 14

erichiller:
Is there anything holding this PR up?

hoerup:
Do this really still need Ubuntu 14 compability ??

powersj:
Hi,

We merged the change for the client version already in a different PR. I am going to close this PR as it is conflicting and only adds some additional README changes. I would be happy to see those changes get added to the README, but no need to keep this open in the current state.

Thanks!

PR #7071, opened 2020-02-24
Histogram aggregator: Enable non-cumulative stats

### LINKED ISSUE #7034: Enable non-cumulative values in histogram aggregator
## Feature Request

*Histogram plugin currently only creates cumulative histograms; proposal is to add an option for ordinary histograms.*

### Proposal:

The [histogram aggregator plugin](https://github.com/influxdata/telegraf/tree/master/plugins/aggregators/histogram) should be able put values in buckets without cumulating smaller values.

### Current behavior:

> The histogram aggregator plugin creates histograms containing the counts of field values within a range.
>
> Values added to a bucket are also added to the larger buckets in the distribution. This creates a cumulative histogram.

For example with buckets `[0.0, 10.0, 100.0]` and values `0, 1, 15, 20`, following metrics are generated:

```
measurement,le=0.0 field_bucket=1i
measurement,le=10.0 field_bucket=3i
measurement,le=100.0 field_bucket=4i
measurement,le=+Inf field_bucket=4i
```

### Desired behavior:

There should be a configuration option `cumulative` (boolean, defaults to true) that defines whether values added to a bucket are also added to the larger buckets.

With `cumulative = false` and the same buckets `[0.0, 10.0, 100.0]` and values `0, 1, 15, 20`, following metrics should be generated:

```
measurement,le=0.0 field_bucket=1i
measurement,le=10.0 field_bucket=1i
measurement,le=100.0 field_bucket=2i
measurement,le=+Inf field_bucket=0i
```

It may also be considered adding a `gt` (greater than) tag:

```
measurement,gt=-Inf,le=0.0 field_bucket=1i
measurement,gt=0.0,le=10.0 field_bucket=1i
measurement,gt=10.0,le=100.0 field_bucket=2i
measurement,gt=100.0,le=+Inf field_bucket=0i
```

### Use case:

Grafana's [Heatmap panel](https://grafana.com/docs/grafana/latest/features/panels/heatmap/) works with [pre-bucketed data](https://grafana.com/docs/grafana/latest/features/panels/heatmap/#pre-bucketed-data), but it requires the data to be non-cumulative.

Example with buckets `[0.0, 10.0, 100.0, 1000.0]` and non-cumulative bucketing (note that I've used the `le` tag and renamed `+Inf` to `>1000` as Grafana reads the tag values as numbers. A `gt` tag would be much better here):

```
time                 le    response_time_ms_bucket
----                 --    -----------------------
2020-02-17T08:40:20Z 0     1
2020-02-17T08:40:20Z 10    2
2020-02-17T08:40:20Z 100   1
2020-02-17T08:40:20Z 1000  2
2020-02-17T08:40:20Z >1000 0
...
2020-02-17T08:41:00Z 0     0
2020-02-17T08:41:00Z 10    2
2020-02-17T08:41:00Z 100   1
2020-02-17T08:41:00Z 1000  2
2020-02-17T08:41:00Z >1000 0
...
2020-02-17T08:41:30Z 0     0
2020-02-17T08:41:30Z 10    1
2020-02-17T08:41:30Z 100   0
2020-02-17T08:41:30Z 1000  4
2020-02-17T08:41:30Z >1000 0
```

grafik

grafik

fxedel:
Closes #7034.

### Required for all PRs:

### DIFF
--- a/plugins/aggregators/histogram/README.md
+++ b/plugins/aggregators/histogram/README.md
@@ -3,8 +3,9 @@
 The histogram aggregator plugin creates histograms containing the counts of
 field values within a range.
 
-Values added to a bucket are also added to the larger buckets in the
-distribution.  This creates a [cumulative histogram](https://en.wikipedia.org/wiki/Histogram#/media/File:Cumulative_vs_normal_histogram.svg).
+If `cumulative` is set to true, values added to a bucket are also added to the
+larger buckets in the distribution. This creates a [cumulative histogram](https://en.wikipedia.org/wiki/Histogram#/media/File:Cumulative_vs_normal_histogram.svg).
+Otherwise, values are added to only one bucket, which creates an [ordinary histogram](https://en.wikipedia.org/wiki/Histogram#/media/File:Cumulative_vs_normal_histogram.svg)
 
 Like other Telegraf aggregators, the metric is emitted every `period` seconds.
 By default bucket counts are not reset between periods and will be non-strictly
@@ -16,7 +17,7 @@ increasing while Telegraf is running. This behavior can be changed by setting th
 Each metric is passed to the aggregator and this aggregator searches
 histogram buckets for those fields, which have been specified in the
 config. If buckets are found, the aggregator will increment +1 to the appropriate
-bucket otherwise it will be added to the `+Inf` bucket.  Every `period`
+bucket. Otherwise, it will be added to the `+Inf` bucket.  Every `period`
 seconds this data will be forwarded to the outputs.
 
 The algorithm of hit counting to buckets was implemented on the base
@@ -39,16 +40,20 @@ of the algorithm which is implemented in the Prometheus
   ## of accumulating the results.
   reset = false
 
+  ## Whether bucket values should be accumulated. If set to false, "gt" tag will be added.
+  ## Defaults to true.
+  cumulative = true
+
   ## Example config that aggregates all fields of the metric.
   # [[aggregators.histogram.config]]
-  #   ## The set of buckets.
+  #   ## Right borders of buckets (with +Inf implicitly added).
   #   buckets = [0.0, 15.6, 34.5, 49.1, 71.5, 80.5, 94.5, 100.0]
   #   ## The name of metric.
   #   measurement_name = "cpu"
 
   ## Example config that aggregates only specific fields of the metric.
   # [[aggregators.histogram.config]]
-  #   ## The set of buckets.
+  #   ## Right borders of buckets (with +Inf implicitly added).
   #   buckets = [0.0, 10.0, 20.0, 30.0, 40.0, 50.0, 60.0, 70.0, 80.0, 90.0, 100.0]
   #   ## The name of metric.
   #   measurement_name = "diskio"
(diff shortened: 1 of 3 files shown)


plugins/aggregators/histogram/README.md
```diff
@@ -39,16 +40,20 @@ of the algorithm which is implemented in the Prometheus
   ## of accumulating the results.
   reset = false
 
+  ## Whether bucket values should be accumulated. If set to false, "gt" tag will be added.
+  ## Defaults to true.
+  cumulative = true
```
ssoroka: What's going to be the value if this isn't set in a user's config?

fxedel: If `cumulative` isn't set in config, it defaults to true:

https://github.com/influxdata/telegraf/blob/55ad4fb3ebdb1e106f7c46ee3ff131920086fbfb/plugins/aggregators/histogram/histogram.go#L67-L75

plugins/aggregators/histogram/histogram.go
```diff
@@ -167,18 +180,35 @@ func (h *HistogramAggregator) groupFieldsByBuckets(
...
 
-	h.groupField(metricsWithGroupedFields, name, field, count, tags)
+		h.groupField(metricsWithGroupedFields, name, field, count, tags)
+	} else {
+		buckets := h.getBuckets(name, field)
+		for index, count := range counts {
+			tags[bucketLeftTag] = bucketNegInf
+			if index > 0 {
+				tags[bucketLeftTag] = strconv.FormatFloat(buckets[index-1], 'f', -1, 64)
+			}
```
danielnelson: I think all we should add is non-cumulative buckets, can't we just stick the raw counts into the bucket instead of incrementing?  The bucket specifiers in the configuration are defined as the right borders, so by definition nothing should go into the -Inf bucket, negative values should be included in the smallest bucket.

fxedel: Hello @danielnelson, what exactly do you mean by "stick the raw counts into the bucket instead of incrementing"?

> The bucket specifiers in the configuration are defined as the right borders, so by definition nothing should go into the -Inf bucket, negative values should be included in the smallest bucket.

When you say they should go "into the XY bucket", you're meaning the bucket with right border XY, right? Because that's just how my implementation works, I only use the specified bucket borders and +Inf as right bucket borders - the `le` tag (which is the right bucket border) has the same values as in the cumulative case. However, in the non-cumulative case, buckets not only have an upper bound but also a lower bound. I wanted to also include that information with the `gt` tag as it can be quite difficult in tools like grafana to get information from the next-lower bucket to receive the left border of the current bucket. I also find this more concrete, as a metric with tags `gt=0.0, le=1.0` simply defines the mathematical interval `]0, 10]`.  (In the cumulative case, intervals are always `]-Inf, 0]`, `]-Inf, 10]`, ...)

danielnelson: I see now, was mixed up by the change in what is being iterated over, buckets vs counts, and I didn't notice that -Inf is for a new tag.

What I'm wondering now is if it makes sense to merge these loops, they are mostly doing the same thing only the non-cumulative doesn't add to count and adds the extra tag.  Something like this (untested):

```diff
+ lowerBound := "-Inf"
count := int64(0)
for index, bucket := range h.getBuckets(name, field) {
+	if !h.Cumulative {
+		count = 0
+		tags[bucketLeftTag] = lowerBound
+		lowerBound = strconv.FormatFloat(bucket, 'f', -1, 64)
+	}

	count += counts[index]
	tags[bucketRightTag] = strconv.FormatFloat(bucket, 'f', -1, 64)
	h.groupField(metricsWithGroupedFields, name, field, count, copyTags(tags))
}
```

PR #7089, opened 2020-02-27
Allow globs in FPM unix socket paths

andrenth:
This PR allows glob patterns to be used in the `unixsocket` configuration. This is useful in environments where FPM pools are managed dynamically, so that the configuration file doesn't have to be edited every time a pool is created or deleted.

### Required for all PRs:

### DIFF
--- a/plugins/inputs/phpfpm/phpfpm.go
+++ b/plugins/inputs/phpfpm/phpfpm.go
@@ -17,6 +17,7 @@ import (
 	"github.com/influxdata/telegraf/internal"
 	"github.com/influxdata/telegraf/internal/tls"
 	"github.com/influxdata/telegraf/plugins/inputs"
+	"github.com/pkg/errors"
 )
 
 const (
--- a/plugins/inputs/phpfpm/phpfpm.go
+++ b/plugins/inputs/phpfpm/phpfpm.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"sync"
@@ -95,7 +96,12 @@ func (g *phpfpm) Gather(acc telegraf.Accumulator) error {
 
 	var wg sync.WaitGroup
 
-	for _, serv := range g.Urls {
+	urls, err := expandUrls(g.Urls)
+	if err != nil {
+		return err
+	}
+
+	for _, serv := range urls {
 		wg.Add(1)
 		go func(serv string) {
 			defer wg.Done()
@@ -153,15 +159,10 @@ func (g *phpfpm) gatherServer(addr string, acc telegraf.Accumulator) error {
 			statusPath = "status"
 		}
 	} else {
-		socketAddr := strings.Split(addr, ":")
-		if len(socketAddr) >= 2 {
-			socketPath = socketAddr[0]
-			statusPath = socketAddr[1]
-		} else {
-			socketPath = socketAddr[0]
+		socketPath, statusPath = unixSocketPaths(addr)
+		if statusPath == "" {
 			statusPath = "status"
 		}
-
 		if _, err := os.Stat(socketPath); os.IsNotExist(err) {
 			return fmt.Errorf("Socket doesn't exist  '%s': %s", socketPath, err)
 		}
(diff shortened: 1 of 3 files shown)


plugins/inputs/phpfpm/phpfpm.go
```diff
@@ -277,6 +278,68 @@ func importMetric(r io.Reader, acc telegraf.Accumulator, addr string) (poolStat,
...
+		addrs = append(addrs, paths...)
+	}
+	return addrs, nil
+}
+
+func globUnixSocket(url string) ([]string, error) {
+	pattern, status := unixSocketPaths(url)
+	paths, err := filepath.Glob(pattern)
+	if err != nil {
+		return nil, err
```
ssoroka: might be good to add some context here. 
```suggestion
		return nil, errors.Wrapf(err, "couldn't read the file path pattern %q", pattern)
```

andrenth: I didn't add any context because `filepath.Glob` can only fail due to invalid syntax in the pattern, but I guess it's useful to add the bad pattern in the error message, especially if multiple patterns are used.

andrenth:
> Thanks for the PR!
> Do we want to consider processing the list of file sockets once on startup instead of every gather request?

Doing it on every gather would simplify things for me, as no further action would be needed to monitor/stop monitoring dynamically managed pools. Doing it on startup wouldn't be a deal breaker:  I'd simply have to reload [restart?] Telegraf every time a pool is added or removed, along with the reload of the FPM daemon that I already perform.

One advantage of doing it on every gather is that it won't cause pauses in the data collection of unrelated metrics because Telegraf had to be restarted to rebuild the FPM socket list.

plugins/inputs/phpfpm/phpfpm.go
```diff
@@ -277,6 +279,68 @@ func importMetric(r io.Reader, acc telegraf.Accumulator, addr string) (poolStat,
...
+	}
+
+	addrs := make([]string, 0, len(paths))
+
+	if len(paths) == 0 {
+		_, err := os.Stat(pattern)
+		if os.IsNotExist(err) {
+			return nil, fmt.Errorf("Socket doesn't exist  '%s': %s", pattern, err)
+		}
+		return nil, err
```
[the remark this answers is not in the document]
andrenth: I did it this way to keep the previous behavior, which is to error out on nonexistent socket paths. Since `filepath.Glob` ignores any errors other than syntax ones, I've used `os.Stat` to get the appropriate error. The special handling of `os.ErrNotExist` is there because it's also done elsewhere in the plugin.

If you think we consider that not matching any sockets is not an error, this code would indeed be a bit cleaner.

plugins/inputs/phpfpm/phpfpm.go
```diff
@@ -277,6 +279,68 @@ func importMetric(r io.Reader, acc telegraf.Accumulator, addr string) (poolStat,
...
+			return nil, err
+		}
+		addrs = append(addrs, paths...)
+	}
+	return addrs, nil
+}
+
+func globUnixSocket(url string) ([]string, error) {
+	pattern, status := unixSocketPaths(url)
+	paths, err := filepath.Glob(pattern)
```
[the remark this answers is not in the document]
andrenth: Done.

plugins/inputs/phpfpm/phpfpm.go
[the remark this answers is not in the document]
andrenth: One example of where this handling is useful is when the unix socket path is in fact *not* a glob, but the file does not exist. In that case, we'd get an empty slice of strings as a result from the glob, with no error, so this extra checking restores the error message one would get before glob support.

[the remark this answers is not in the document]
danielnelson: Okay, let's continue to show an error when the socket doesn't exist.  I think we should then remove the Stat call on L166, which doesn't seem useful anymore and anyway we will report an error afterwards connecting to the socket if it doesn't exist.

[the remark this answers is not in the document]
andrenth: Looking at the `globpath` module I noticed `Match` actually returns the given parameter when it's a static path, so the latest commit handles this case correctly now. I've also removed the redundant `os.Stat` call.

If the PR is OK now, do you want me to rebase before merging?

plugins/inputs/phpfpm/phpfpm.go
```diff
@@ -277,6 +275,75 @@ func importMetric(r io.Reader, acc telegraf.Accumulator, addr string) (poolStat,
...
+func globUnixSocket(url string) ([]string, error) {
+	pattern, status := unixSocketPaths(url)
+	glob, err := globpath.Compile(pattern)
+	if err != nil {
+		return nil, fmt.Errorf("could not compile glob %q: %v", pattern, err)
+	}
+	paths := glob.Match()
+	if len(paths) == 0 {
+		return nil, nil
+	}
```
danielnelson: If we are planning to show an error if a glob doesn't have any matches, then I think we need to do it here.

plugins/inputs/phpfpm/phpfpm.go
```diff
@@ -277,6 +275,75 @@ func importMetric(r io.Reader, acc telegraf.Accumulator, addr string) (poolStat,
...
+	// globpath.Match() returns the given argument when it's a static path,
+	// i.e., not a glob pattern. In that case, ensure it exists.
+	if len(paths) == 1 && paths[0] == pattern {
+		if _, err := os.Stat(paths[0]); err != nil {
+			if os.IsNotExist(err) {
+				return nil, fmt.Errorf("Socket doesn't exist  '%s': %s", pattern, err)
+			}
+			return nil, err
+		}
+	}
```
danielnelson: Since, as you mentioned, Match() will return the path if there are no special glob characters, I think we can remove this too.  We will still display an error when we dial the socket.

plugins/inputs/phpfpm/phpfpm.go
andrenth: Hmm ok, my thinking was that since globpath can diferentiate static paths from patterns, to allow non-matching globs, but to error on static paths.

plugins/inputs/phpfpm/phpfpm.go
andrenth: Done (this required a change to the unit tests due to the error now coming from Dial).

danielnelson:
Merged for 1.14, thanks!

andrenth:
Thank you!

Issue #7070, opened 2020-02-24
statsd processing slow after 1.5.3


stevenh:
### Relevant telegraf.conf:

```toml
[agent]
  interval = "1s"
  round_interval = true
  metric_batch_size = 10000
  metric_buffer_limit = 100000000
  flush_buffer_when_full = true
  collection_jitter = "0s"
  flush_interval = "20s"
  flush_jitter = "0s"
  debug = false
  quiet = false

[[outputs.influxdb]]
  database = "XXX"
  password = "XXX"
  urls = ["http://influx"]
  username = "XXX"

[[inputs.statsd]]
  allowed_pending_messages = 100000
  delete_counters = true
  delete_gauges = true
  delete_sets = true
  delete_timings = true
  metric_separator = "."
  percentile_limit = 1000
  percentiles = [90, 95, 99]
  service_address = ":8125"
```

### System info:

FreeBSD 11
Telegraf later 1.5.3 have tried 1.6.0 1.8.0 1.10.0 and 1.13.7 (latest)

### Steps to reproduce:
1. Send significant amounts of statsd to telegraf on anything later than 1.5.3
2. tail the log file or look at the stats in influx

### Expected behavior:
Stats should be processed

### Actual behavior:
Most stats are dropped due to slow processing and the following is in the logs
```
2020-02-23T23:40:18Z E! Error in plugin [inputs.statsd]: took longer to collect than collection interval (1s)
2020-02-23T23:40:20Z E! Error in plugin [inputs.statsd]: took longer to collect than collection interval (1s)
2020-02-23T23:40:21Z E! Error in plugin [inputs.statsd]: took longer to collect than collection interval (1s)
```

### Additional info:
After bisecting the commits all the way back to 1.5.3 the following commit is the cause:
https://github.com/influxdata/telegraf/commit/1c0f63a90d60da228ff32944b4078e57d826dc83

stevenh:
[profiles.zip](https://github.com/influxdata/telegraf/files/4242803/profiles.zip)
Attached is a profile of before and after this patch for investigation

stevenh:
Enabling debugging shows a grim picture:
**1.5.3**
```
2020-02-24T15:02:22Z D! Output [influxdb] wrote batch of 202 metrics in 4.797833ms
2020-02-24T15:02:22Z D! Output [influxdb] wrote batch of 7647 metrics in 181.80504ms
[...]

PR #7094, opened 2020-02-28
feat(inputs/statsd): Improve gather performance

stevenh:
Improve the gather performance to the statsd input processor by processing metric types in parallel.

Use Agent.MetricBatchSize to configure the processing channel buffer, which allows high throughput inputs to process entire batches.

Together these reduces the processing time of 50k metrics from over 1 second to ~100ms when not limited by the current influx metric serializer with a good selection of metric types.

Also:
* Add debug information about gather timings.

Helps: #7070

### Required for all PRs:

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -51,9 +51,9 @@ func (a *Agent) Run(ctx context.Context) error {
 		return err
 	}
 
-	inputC := make(chan telegraf.Metric, 100)
-	procC := make(chan telegraf.Metric, 100)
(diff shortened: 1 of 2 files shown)


danielnelson:
Thanks for the pr, let's keep this open while we continue to work through #7070.

danielnelson:
Closing, I believe this was addressed through InfluxData support.

stevenh:
While we did address this core issue this is still a valid improvement so should be considered on its own merits

PR #7104, opened 2020-03-03
Set headers correctly on influxdb_listener

nhojjohn:
The ping endpoint in influx returns the version in a HTTP header (eg X-Influxdb-Version: 1.0).  The presence of this header is checked by Springboot influxdb ping() implementation (at quite possibly other clients).  This PR adds this header to ping responses and additionally sets the Content-Type correctly when used with the verbose parameter.

### Required for all PRs:


reimda:
We just merged #6974 which renamed http_listener.go to influxdb_listener.go and changed the url routing.  This caused a conflict in your PR.  Could you resolve the conflict so we can merge this? Thanks!

PR #7112, opened 2020-03-04
Huawei_Routers_Telemetry input Plugin

DamRCorba:
### Required for all PRs:


DamRCorba:
Huawei Routers Telemetry Udp input Plugin.

BSpendlove:
Hey DamRCorba,

Nice work, are you going to be implementing grpc any time soon? 😂

DamRCorba:
> Hey DamRCorba,
> 
> Nice work, are you going to be implementing grpc any time soon? joy

yes, I am working on it. But Dial-Out model only, for now.

plugins/inputs/huawei_routers_telemetry/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
```
srebhan, line 1: Please remove the license file.

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
```diff
@@ -0,0 +1,230 @@
...
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/internal"
+	"github.com/influxdata/telegraf/metric"
+	"github.com/influxdata/telegraf/plugins/inputs"
+	"io"
+	"net"
+	"os"
+	"strings"
+	"sync"
+	"time"
```
srebhan, line 17: Please group this for standard libraries, telegraf and other...
```suggestion
	"fmt"
	"io"
	"net"
	"os"
	"strings"
	"sync"
	"time"

	"github.com/DamRCorba/huawei_telemetry_sensors"
	"github.com/DamRCorba/huawei_telemetry_sensors/sensors/huawei-telemetry"

	"github.com/golang/protobuf/proto"

	"github.com/influxdata/telegraf"
	"github.com/influxdata/telegraf/internal"
	"github.com/influxdata/telegraf/metric"
	"github.com/influxdata/telegraf/plugins/inputs"
```

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
```diff
@@ -0,0 +1,230 @@
...
+  Listen UDP packets and call the telemetryDecoder.
+*/
+func (psl *packetSocketListener) listen() {
+	buf := make([]byte, 64*1024) // 64kb - maximum size of IP packet
+	for {
+		n, _, err := psl.ReadFrom(buf)
+		if err != nil {
+			if !strings.HasSuffix(err.Error(), ": use of closed network connection") {
+				psl.Log.Error(err.Error())
+			}
```
srebhan, line 90: Why do you only log this error type?

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
```diff
@@ -0,0 +1,230 @@
...
+			break
+		}
+
+		body, err := psl.decoder.Decode(buf[:n])
+		if err != nil {
+			psl.Log.Errorf("Unable to decode incoming packet: %s", err.Error())
+		}
```
srebhan, line 97: I think you should continue here. Furthermore you can shorten the log by
```suggestion
		if err != nil {
			psl.Log.Errorf("Unable to decode incoming packet: %v", err)
			continue
		}
```

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
```diff
@@ -0,0 +1,230 @@
...
+		// Inicia el manejo de la telemetria
```
srebhan, line 98: Can you please use English.

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
```diff
@@ -0,0 +1,230 @@
...
+		grouper := HuaweiTelemetryDecoder(body)
+		for _, metric := range grouper.Metrics() {
+			psl.AddMetric(metric)
+		}
+
+		if err != nil {
+			psl.Log.Errorf("Unable to parse incoming packet: %s", err.Error())
+
+			continue
```
srebhan, line 107: No need to `continue` here, we are already at the end of the loop.

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
```diff
@@ -0,0 +1,230 @@
...
+}
+
+func (sl *HuaweiRoutersTelemetry) Description() string {
+	return "Huawei Telemetry UDP model input Plugin"
+}
+
+func (sl *HuaweiRoutersTelemetry) SampleConfig() string {
+	return `
+  ## UDP Service Port to capture Telemetry
+  service_port = "8080"
```
srebhan, line 132: This is probably a sane default, isn't it? So comment this line
```suggestion
  # service_port = "8080"
```

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
```diff
@@ -0,0 +1,230 @@
...
+
+`
+}
+
+func (sl *HuaweiRoutersTelemetry) Gather(_ telegraf.Accumulator) error {
+	return nil
+}
+
+func (hrt *HuaweiRoutersTelemetry) Start(acc telegraf.Accumulator) error {
```
srebhan, line 141: Please always use the same variable name `(h *HuaweiRoutersTelemetry)`.

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
```diff
@@ -0,0 +1,230 @@
...
+	hrt.Log.Infof("Listening on %s://%s", "udp", pc.LocalAddr())
+
+	psl := &packetSocketListener{
+		PacketConn:             pc,
+		HuaweiRoutersTelemetry: hrt,
+	}
+
+	hrt.Closer = psl
+	hrt.wg = sync.WaitGroup{}
+	hrt.wg.Add(1)
```
srebhan, line 172: What do you need the `WaitGroup` for?

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
```diff
@@ -0,0 +1,230 @@
...
+*/
+func (psl *packetSocketListener) listen() {
+	buf := make([]byte, 64*1024) // 64kb - maximum size of IP packet
+	for {
+		n, _, err := psl.ReadFrom(buf)
+		if err != nil {
+			if !strings.HasSuffix(err.Error(), ": use of closed network connection") {
+				psl.Log.Error(err.Error())
+			}
+			break
```
srebhan, line 91: I'm not sure this is a good idea as this means the plugin will go defunct after the first read error.

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
```diff
@@ -0,0 +1,230 @@
...
+	if hrt.Closer != nil {
+		hrt.Close()
+		hrt.Closer = nil
+	}
+	hrt.wg.Wait()
+}
+
+func newHuaweiRoutersTelemetry() *HuaweiRoutersTelemetry {
+	return &HuaweiRoutersTelemetry{}
+}
```
srebhan, line 215: Please construct the struct directly in `init()`.

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
```diff
@@ -0,0 +1,230 @@
...
+	path   string
+	closer io.Closer
+}
+
+func (uc unixCloser) Close() error {
+	err := uc.closer.Close()
+	os.Remove(uc.path) // ignore error
+	return err
+}
+
```
srebhan, line 227: I think this is not used.

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+}
+
+type HuaweiRoutersTelemetry struct {
+	ServicePort     string        `toml:"service_port"`
+	ReadBufferSize  internal.Size `toml:"read_buffer_size"`
+	ContentEncoding string        `toml:"content_encoding"`
+	wg              sync.WaitGroup
+
+	Log telegraf.Logger
```
srebhan, line 118: Please use
```suggestion
	Log telegraf.Logger `toml:"-"`
```
to tell the config to ignore this as option.

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
```diff
@@ -0,0 +1,230 @@
...
+/*
+  Telemetry Decoder.
+
+*/
+func HuaweiTelemetryDecoder(body []byte) *metric.SeriesGrouper {
+	msg := &telemetry.Telemetry{}
+	err := proto.Unmarshal(body[12:], msg)
+	if err != nil {
+		fmt.Println("Error", err)
+		panic(err)
```
srebhan, line 48: Please never panic as this will tear-down telegraf as a whole.

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+	grouper := metric.NewSeriesGrouper()
+	for _, gpbkv := range msg.GetDataGpb().GetRow() {
```
srebhan, line 51: Are you sure `msg.GetDataGpb()` can never be `nil`?

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
```diff
@@ -0,0 +1,230 @@
...
+		dataTime := gpbkv.Timestamp
+		if dataTime == 0 {
+			dataTime = msg.MsgTimestamp
+		}
+		timestamp := time.Unix(int64(dataTime/1000), int64(dataTime%1000)*1000000)
```
srebhan, line 56:
```suggestion
		timestamp := time.Unix(0, int64(dataTime)*1000000)
```

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
```diff
@@ -0,0 +1,230 @@
...
+		sensorMsg := huawei_sensorPath.GetMessageType(msg.GetSensorPath())
+		err = proto.Unmarshal(gpbkv.Content, sensorMsg)
+		if err != nil {
+			fmt.Println("Sensor Error", err)
```
srebhan, line 60: Please make this function a member of your struct and use the logger.

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
```diff
@@ -0,0 +1,230 @@
...
+		} else {
```
srebhan, line 61: Please return an error here and remove the `else` statement to save an indention level.

srebhan (changes requested):
This is a first round of review. Please see my comments in the code.

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
DamRCorba: Comment added

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
DamRCorba: you are right

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
DamRCorba: yes, of course, my bad

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
DamRCorba: yes , continue added

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
DamRCorba: In the start of the project i only wanted to know if the message could be read. 
I have chage it to log all errors.

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
DamRCorba: i made the change. I pushing in when i finish te review the other points

plugins/inputs/huawei_routers_telemetry/LICENSE
DamRCorba: removed

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
DamRCorba: maybe i misunderstood, if the packet could not be read the loop must not continue. If i use break the plugin stop? I am using like this in a "production enviroment" and its work fine.

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
DamRCorba: done

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
DamRCorba: removed, I dont remeber why i need it. Sure in my initial testing

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
DamRCorba: Ok

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
DamRCorba: done

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
DamRCorba: changed to return error

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
DamRCorba: but, i want the seconds.

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
DamRCorba: because of the conection handling.

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
DamRCorba: If the router send a message, it has info always.

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
DamRCorba: ok

DamRCorba:
i have this error with __make tidy__

```bash
$ make tidy
go mod verify
all modules verified
go mod tidy
github.com/influxdata/telegraf/metric imports
        hash/maphash: malformed module path "hash/maphash": missing dot in first path element
make: *** [Makefile:146: tidy] Error 1
```
Any clue?

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
srebhan: Well you get the seconds, it's just that you specify the time in nanoseconds and `time.Unix` does the magic. Please not that the modulo operator for nanoseconds is removed in my suggestion. Give it a try on go playground... :-)

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
```diff
@@ -0,0 +1,219 @@
...
+		var addr *net.UDPAddr
+		var err error
+		var ifi *net.Interface
+		if spl := strings.SplitN(address, "%", 2); len(spl) == 2 {
+			address = spl[0]
+			ifi, err = net.InterfaceByName(spl[1])
+			if err != nil {
+				return nil, err
+			}
+		}
```
srebhan, line 196: If I read your code correctly this should never be executed as you always only have the port as config option and pass `:<configured port>` in here.

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
```diff
@@ -0,0 +1,219 @@
...
+	go func() {
+		defer h.wg.Done()
+		psl.listen()
+	}()
+	return nil
+}
+
+func udpListen(network string, address string) (net.PacketConn, error) {
+	switch network {
+	case "udp", "udp4", "udp6":
```
srebhan, line 186: With the current config options you will _always_ end up in this path as you pass `"upd"` as a fixed string for `network`. Can we please remove this and you add the code back once there are other options you can pass through the config!?

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
```diff
@@ -0,0 +1,219 @@
...
+func (h *HuaweiRoutersTelemetry) Stop() {
+	if h.Closer != nil {
+		h.Close()
+		h.Closer = nil
+	}
+	h.wg.Wait()
+}
+
+func init() {
+	inputs.Add("huawei_routers_telemetry", func() telegraf.Input { return &HuaweiRoutersTelemetry{} })
```
srebhan, line 218: Usually plugins that open a "server" port waiting for clients to connect and pass metrics are names `..._listener`. How about `huawei_router_listener` as plugin name?

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
```diff
@@ -0,0 +1,219 @@
...
+
+type streamSocketListener struct {
+	net.Listener
+	*HuaweiRoutersTelemetry
+
+	sockType string
+
+	connections    map[string]net.Conn
+	connectionsMtx sync.Mutex
+}
```
srebhan, line 34: I cannot find a reference to this type. Please remove.

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
```diff
@@ -0,0 +1,219 @@
...
+
+	Log telegraf.Logger `toml:"-"`
+
+	telegraf.Accumulator
+	io.Closer
+	decoder internal.ContentDecoder
+}
+
+func (h *HuaweiRoutersTelemetry) Description() string {
+	return "Huawei Telemetry UDP model input Plugin"
```
srebhan, line 130: How about
```suggestion
	return "Input plugin for receiving Huawei Router Telemetry data via UDP"
```

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
```diff
@@ -0,0 +1,219 @@
...
+}
+
+func (h *HuaweiRoutersTelemetry) SampleConfig() string {
+	return `
+  ## UDP Service Port to capture Telemetry
+  # service_port = "8080"
+
```
srebhan, line 137: Documentation for at least `read_buffer_size` and `content_encoding` is missing.

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
```diff
@@ -0,0 +1,219 @@
...
+	ServicePort     string        `toml:"service_port"`
+	ReadBufferSize  internal.Size `toml:"read_buffer_size"`
+	ContentEncoding string        `toml:"content_encoding"`
+	wg              sync.WaitGroup
+
+	Log telegraf.Logger `toml:"-"`
+
+	telegraf.Accumulator
+	io.Closer
+	decoder internal.ContentDecoder
```
srebhan, line 126: Please group by external/internal/anonymous:
```suggestion
	Log telegraf.Logger `toml:"-"`

	decoder internal.ContentDecoder
	wg         sync.WaitGroup

		telegraf.Accumulator
	io.Closer
```

srebhan: Oh and add a variable for `telegraf.Accumulator` such as `acc telegraf.Accumulator`.

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
```diff
@@ -0,0 +1,219 @@
...
+
+`
+}
+
+func (h *HuaweiRoutersTelemetry) Gather(_ telegraf.Accumulator) error {
+	return nil
+}
+
+func (h *HuaweiRoutersTelemetry) Start(acc telegraf.Accumulator) error {
+	h.Accumulator = acc
```
srebhan, line 146:
```suggestion
	h.acc = acc
```

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
```diff
@@ -0,0 +1,219 @@
...
+
+	if h.ReadBufferSize.Size > 0 {
+		if srb, ok := pc.(setReadBufferer); ok {
+			srb.SetReadBuffer(int(h.ReadBufferSize.Size))
+		} else {
+			h.Log.Warnf("Unable to set read buffer on a %s socket", "udp")
+		}
+	}
+
+	h.Log.Infof("Listening on %s://%s", "udp", pc.LocalAddr())
```
srebhan, line 167: I see that you prepare for other non-udp connections, but for me it is unclear if this ever happens, so please use `udp` only now to simplify code. If we ever want to integrate something else I think the structure is in the code to do it easily at a later time.

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
```diff
@@ -0,0 +1,219 @@
...
+
+	psl := &packetSocketListener{
+		PacketConn:             pc,
+		HuaweiRoutersTelemetry: h,
+	}
```
srebhan, line 172: Can we please invert this and simply add the connection as a field to the `HuaweiRoutersTelemetry` structure. This is the way all other plugins go, so it is of least surprise.

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
```diff
@@ -0,0 +1,219 @@
...
+		return net.ListenUDP(network, addr)
+	}
+	return net.ListenPacket(network, address)
+}
+
+func (h *HuaweiRoutersTelemetry) Stop() {
+	if h.Closer != nil {
+		h.Close()
+		h.Closer = nil
+	}
```
srebhan, line 213: Please add the connection as a field to the `HuaweiRoutersTelemetry` struct and do
```suggestion
	if h.connection != nil {
		h.connection.Close()
	}
```

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
srebhan: What I mean is: Once you encounter a read error, you will leave the eternal `for` loop and `listen` will end. Now nothing is listening for packets anymore, so if the router side recovers or network is back or whatever, you will not receive new data anymore until you restart the plugin (and/or telegraf). That's bad if you rely on metrics coming in.

To solve this, the plugin should close the connection and try to listen again (like do an internal restart) in order to be prepared for the router-side to reconnect.

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
```diff
@@ -0,0 +1,219 @@
...
+	for {
+		n, _, err := h.ReadFrom(buf)
+		if err != nil {
+			h.Log.Error("Unable to read buffer: %s", err.Error())
+			break
+		}
+
+		body, err := h.decoder.Decode(buf[:n])
+		if err != nil {
+			h.Log.Errorf("Unable to decode incoming packet: %s", err.Error())
```
srebhan, line 97: You can shorten to
```suggestion
			h.Log.Errorf("Unable to decode incoming packet: %v", err)
```

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
```diff
@@ -0,0 +1,219 @@
...
+			h.Log.Errorf("Unable to decode telemetry information: %s", err.Error())
+			break
+		}
+		for _, metric := range grouper.Metrics() {
+			h.AddMetric(metric)
+		}
+
+		if err != nil {
+			h.Log.Errorf("Unable to parse incoming packet: %s", err.Error())
+		}
```
srebhan, line 112: I think you can never end up here!?

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
srebhan: However, if something else connects to your UDP port and sends you garbage you will rip-down telegraf completely with a panic! Please check for errors and don't rely on data coming over a wire not fully under your control!

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
```diff
@@ -0,0 +1,219 @@
...
+  Telemetry Decoder.
+
+*/
+func HuaweiTelemetryDecoder(body []byte) (*metric.SeriesGrouper, error) {
+	msg := &telemetry.Telemetry{}
+	err := proto.Unmarshal(body[12:], msg)
+	if err != nil {
+		fmt.Println("Unable to decode incoming packet: ", err.Error())
+		return nil, err
+		//panic(err)
```
srebhan, line 51: Remove dead code.

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
```diff
@@ -0,0 +1,219 @@
...
+	for _, gpbkv := range msg.GetDataGpb().GetRow() {
+		dataTime := gpbkv.Timestamp
+		if dataTime == 0 {
+			dataTime = msg.MsgTimestamp
+		}
+		timestamp := time.Unix(int64(dataTime/1000), int64(dataTime%1000)*1000000)
+		sensorMsg := huawei_sensorPath.GetMessageType(msg.GetSensorPath())
+		err = proto.Unmarshal(gpbkv.Content, sensorMsg)
+		if err != nil {
+			fmt.Println("Sensor Error: ", err.Error())
```
srebhan, line 63: Please use proper logging!

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry_test.go
```diff
@@ -0,0 +1,27 @@
...
... 0x15, 0x2a, 0x13, 0x0a, 0x11, 0x22, 0x02, 0x31, 0x34, 0x08, 0x81, 0x80, 0xb8, 0x08, 0x28, 0x03, 0x18, 0x5a, 0x30, 0x4b, 0x10, 0x08, 0x0a, 0x1d, 0x08, 0x87, 0xab, 0xa2, 0xd9, 0x83, 0x2e, 0x5a, 0x14, 0x2a, 0x12, 0x0a, 0x10, 0x22, 0x01, 0x35, 0x08, 0x81, 0x80, 0x94, 0x08, 0x28, 0x12, 0x18, 0x5a, 0x30, 0x4b, 0x10, 0x08, 0x0a, 0x1d, 0x08, 0x87, 0xab, 0xa2, 0xd9, 0x83, 0x2e, 0x5a, 0x14, 0x2a, 0x12, 0x0a, 0x10, 0x22, 0x01, 0x34, 0x08, 0x81, 0x80, 0x90, 0x08, 0x28, 0x10, 0x18, 0x5a, 0x30, 0x4b, 0x10, 0x08, 0x0a, 0x1d, 0x08, 0x87, 0xab, 0xa2, 0xd9, 0x83, 0x2e, 0x5a, 0x14, 0x2a, 0x12, 0x0a, 0x10, 0x22, 0x01, 0x31, 0x08, 0x81, 0x80, 0x84, 0x08, 0x28, 0x2a, 0x18, 0x5a, 0x30, 0x4b, 0x10, 0x08, 0x0a, 0x1d, 0x08, 0x87, 0xab, 0xa2, 0xd9, 0x83, 0x2e, 0x5a, 0x14, 0x2a, 0x12, 0x0a, 0x10, 0x22, 0x01, 0x37, 0x08, 0x81, 0x80, 0x9c, 0x08, 0x28, 0x29, 0x18, 0x5a, 0x30, 0x4b, 0x10, 0x08, 0x0a, 0x1d, 0x08, 0x87, 0xab, 0xa2, 0xd9, 0x83, 0x2e, 0x5a, 0x14, 0x2a, 0x12, 0x0a, 0x10, 0x22, 0x01, 0x38, 0x08, 0x81, 0x80, 0xa0, 0x08, 0x28, 0x0e, 0x18, 0x5a, 0x30, 0x4b, 0x10, 0x08, 0x40, 0x87, 0xab, 0xa2, 0xd9, 0x83, 0x2e, 0x48, 0x90, 0x4e, 0x52, 0x02, 0x4f, 0x4b, 0x5a, 0x05, 0x4e, 0x45, 0x34, 0x30, 0x45, 0x60, 0x00}
```
srebhan, line 17: In case you could add more test messages, it might be an option to put this data to files in a `testdata` subdirectory and read them as test-cases.

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry_test.go
```diff
@@ -0,0 +1,27 @@
...
+
+	grouper, _ := HuaweiTelemetryDecoder(packet)
+	acc.AddMetric(grouper.Metrics()[0])
+
+	require.Empty(t, acc.Errors)
```
srebhan, line 22:
```suggestion
	require.Empty(t, acc.Errors)
```

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry_test.go
```diff
@@ -0,0 +1,27 @@
...
+
+	grouper, _ := HuaweiTelemetryDecoder(packet)
+	acc.AddMetric(grouper.Metrics()[0])
```
srebhan, line 20:
```suggestion
	metrics := grouper.Metrics()
	assert.Len(t, metrics, 1)
	acc.AddMetric(metrics[0])
```

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry_test.go
```diff
@@ -0,0 +1,27 @@
...
+
+	require.Empty(t, acc.Errors)
+
+	tags := map[string]string{"path": "huawei-devm:devm/cpuInfos/cpuInfo", "position": "\"9\"", "source": "HuaweiTest1", "subscription": "ProbandoTelemetriaSubscripcion"}
+	fields := map[string]interface{}{"entIndex": uint32(17367041), "interval": uint32(8), "ovloadThreshold": uint32(90), "systemCpuUsage": uint32(10), "unovloadThreshold": uint32(75)}
+	acc.AssertDoesNotContainsTaggedFields(t, "huawei-devm:devm/cpuInfos/cpuInfo", fields, tags)
```
srebhan, line 26: Shouldn't you test for the actual metric instead of asserting on something is not contained? How about
```suggestion
	expected := []telegraf.Metric{
		testutil.MustMetric(name, tags, fields),
	}
	testutil.RequireMetricsEqual(t, metrics, expected)
```

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
```diff
@@ -0,0 +1,219 @@
...
+import (
+	"fmt"
+	"io"
+	"net"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/DamRCorba/huawei_telemetry_sensors"
+	"github.com/DamRCorba/huawei_telemetry_sensors/sensors/huawei-telemetry"
```
srebhan, line 12: Oh wow this was now costing me really a lot of time just to find out that the package you are importing here is internally names `package huawei_sensorPath`. This is totally unexpected! First you are thinking _"Where the heck is this used?"_ only to wonder _"Where does this 'huawei_sensorPath' thing come from?"_ in the code.

Please make this consistent by either renaming that imported package or move the repo.

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
```diff
@@ -0,0 +1,219 @@
...
+		tags["source"] = msg.GetNodeIdStr()
+		tags["subscription"] = msg.GetSubscriptionIdStr()
+		tags["path"] = msg.GetSensorPath()
+		// Search for Tags
+		for i := 0; i < len(fields); i++ {
+			tags = huawei_sensorPath.AppendTags(fields[i], vals[i], tags, msg.GetSensorPath())
+		}
+		// Create Metrics
+		for i := 0; i < len(fields); i++ {
+			huawei_sensorPath.CreateMetrics(grouper, tags, timestamp, msg.GetSensorPath(), fields[i], vals[i])
```
srebhan, line 77: No. Please do not do this in the external package! First of all you add a dependency to _telegraf_ in your package! While this is not the problem of telegraf, you create a big ball of mud with intertwined code totally killing modularization! Furthermore, nobody will be able to debug problems without modifying your packages as you cannot get any idea of what data is involved.

Please modify the `huawei_sensorPath.CreateMetrics()` to return the data in a sensible way _without_ relying on the grouper and then group the data here.

srebhan (changes requested):
Hey @DamRCorba, sorry but we have to work more on this plugin!

First of all you should change the way you structure the data-structures in here. Please use your plugin structure `HuaweiRoutersTelemetry` as the central point and add information you need as fields instead of anonymous structs!

Second, please get the usage of your external package straight! It is totally unexpected to import a package that is named different internally! While valid Golang code, it makes reviews a mess. Please use consistent naming of what you import and how the packages are named. Furthermore, define clear interfaces between your package and telegraf! Cut the two in self-contained pieces and do not rely on telegraf structures in your package!

Third, remove all "future-preparation" code from the plugin and boil down the implementation to what you support _NOW_! You only support UDP well then use UDP-connections directly! You might structure the code such that adding features later is easy, but please do not complicate the current code with stuff that might never happen.

Last but not least, your test is the opposite of what is expected. You are just testing that something is _NOT_ in the result but do not check if you actually get the fields/tags that should be there! This renders the test not very useful as even an empty metric would pass that test! Please think of tests! Test what you get against what you expect! Implement a test for the full code path, i.e. setup the plugin, connect via UDP and send a packet, test what you got in the accumulator. This is some work, but it will save everybody else in cases where this plugin needs to be touched.

DamRCorba:
@srebhan, no problem. I wrote this plugin to be used by the community. I will make all the necesary changes. i would like some reference for the tests. Could you give me some docs, or guides?

srebhan:
@DamRCorba glad you are so motivated! Let's first work out the decoupling between your package and this plugin. Let me know if you have fixed this. You can use the re-request review button on the top-right or just drop me a message.
Afterwards we should work out the structure, etc. How does that sound to you?

DamRCorba:
> 
> 
> @DamRCorba glad you are so motivated! Let's first work out the decoupling between your package and this plugin. Let me know if you have fixed this. You can use the re-request review button on the top-right or just drop me a message.
> Afterwards we should work out the structure, etc. How does that sound to you?

Ok, step by step.

srebhan:
@DamRCorba any update?

DamRCorba:
> 
> 
> @DamRCorba any update?

@srebhan Sorry, i have a hard month in this hard times.... I will be coding next week.

srebhan:
@DamRCorba no worries, I just want to make sure people don't loose interest. If you don't have time to work on the PR it's also fine, I'll just try to ping you from time-to-time to see you are still caring. :-)

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
DamRCorba: ok

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
DamRCorba: you are right. deleted

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
DamRCorba: I understand but i dont want to lose the Huawei_Router_Telemetry,  huawei_router_telemetry_listener its awfuly long. It so particular use case.

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
DamRCorba: Done

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
DamRCorba: changed

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
DamRCorba: done

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
DamRCorba: Indeed, i modify it to:
```go
h.Log.Infof("Listening Routers on port %s", pc.LocalAddr())
```

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
DamRCorba: done

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
DamRCorba: done

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
DamRCorba: done

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
DamRCorba: done

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
DamRCorba: In progress

DamRCorba:
@srebhan I am having trouble to build telegraf
```bash
build github.com/influxdata/telegraf/cmd/telegraf: cannot load github.com/hashicorp/consul/api: ambiguous import: found github.com/hashicorp/consul/api in multiple modules:
github.com/hashicorp/consul v1.2.1 (/home/damian/go/pkg/mod/github.com/hashicorp/consul@v1.2.1/api)
github.com/hashicorp/consul/api v1.4.0 (/home/damian/go/pkg/mod/github.com/hashicorp/consul/api@v1.4.0)
```

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
DamRCorba: > @srebhan I am having trouble to build telegraf
> 
> ```shell
> build github.com/influxdata/telegraf/cmd/telegraf: cannot load github.com/hashicorp/consul/api: ambiguous import: found github.com/hashicorp/consul/api in multiple modules:
> github.com/hashicorp/consul v1.2.1 (/home/damian/go/pkg/mod/github.com/hashicorp/consul@v1.2.1/api)
> github.com/hashicorp/consul/api v1.4.0 (/home/damian/go/pkg/mod/github.com/hashicorp/consul/api@v1.4.0)
> ```

I clean all go mod as root and update golang to 1.16, i had go 1.13 so i updated. Its running now. I will be running the plugin with real metrics and fix some issues

srebhan:
@DamRCorba version 1.2.1 seems really old, current telegraf uses 1.8.1 (which is old either). Your problem comes from importing multiple versions of the module if I'm not mistaken. You should fix that.

DamRCorba:
@srebhan how are you? I am making the modifications to de plugin in order to have all the code in the plugin and only the sensors in my github. 

I try to build telegraf but it had some errors and not in my plugin.

```bash
# get https://sum.golang.org/lookup/k8s.io/kubernetes@v1.13.0
# get https://sum.golang.org/lookup/k8s.io/kubernetes@v1.13.0: 200 OK (0.005s)
go build -ldflags " -X main.commit=9e155ea5 -X main.branch=huawei_routers_telemetry -X main.goos=linux -X main.goarch=amd64" ./cmd/telegraf
# github.com/docker/libnetwork/ipvs
../../.go/pkg/mod/github.com/docker/libnetwork@v0.8.0-dev.2.0.20181012153825-d7b61745d166/ipvs/ipvs.go:100:32: cannot use &tv (type *syscall.Timeval) as type *unix.Timeval in argument to sock.SetSendTimeout
../../.go/pkg/mod/github.com/docker/libnetwork@v0.8.0-dev.2.0.20181012153825-d7b61745d166/ipvs/ipvs.go:104:35: cannot use &tv (type *syscall.Timeval) as type *unix.Timeval in argument to sock.SetReceiveTimeout
../../.go/pkg/mod/github.com/docker/libnetwork@v0.8.0-dev.2.0.20181012153825-d7b61745d166/ipvs/netlink.go:219:13: assignment mismatch: 2 variables but s.Receive returns 3 values
make[1]: *** [Makefile:91: telegraf] Error 2
make: *** [Makefile:67: all] Error 2
```

Any suggestion?

srebhan:
Hey @DamRCorba, I'm fine, thanks! How about you?
Regarding your build error, I don't know TBH. It seems like you using another API than what is expected. However, please note that the lib (`github.com/docker/libnetwork`) is _very_ old (@v0.8.0-dev.2.0.**2018**1012153825). Maybe that's the source of your issue...

DamRCorba:
@srebhan I make a fetch upstream and all this odd stuff is gone. But now i realise that "internal.Size" was removed, can be?

```bash
make
go mod download -x
go build -ldflags " -X main.commit=856d9be2 -X main.branch=master -X main.goos=linux -X main.goarch=amd64" ./cmd/telegraf
# github.com/influxdata/telegraf/plugins/inputs/huawei_routers_telemetry
plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go:45:18: undefined: internal.Size
make[1]: *** [Makefile:93: telegraf] Error 2
make: *** [Makefile:67: all] Error 2
```
It was removed or moved?

srebhan:
> @srebhan I make a fetch upstream and all this odd stuff is gone. But now i realise that "internal.Size" was removed, can be?

Yes, this was removed and replaced by `config.Size` you should use instead.

DamRCorba:
> 
> 
> > @srebhan I make a fetch upstream and all this odd stuff is gone. But now i realise that "internal.Size" was removed, can be?
> 
> Yes, this was removed and replaced by `config.Size` you should use instead.

With config.Size build ok in my local repository. I can capture de telemetry again. But the CircleCI verifications fails in make check.
I try to run it localy and i recive the same error in go vet.

```bash
make check
go vet $(go list ./... | grep -v ./plugins/parsers/influx)
# github.com/influxdata/telegraf/plugins/inputs/huawei_routers_telemetry
vet: plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry_test.go:19:45: too few arguments in call to HuaweiTelemetryDecoder

go vet has found suspicious constructs. Please remediate any reported errors
to fix them before submitting code for review.
make: *** [Makefile:129: vet] Error 1
```

I dont have any clue. What should i change?

DamRCorba:
> 
> 
> > > @srebhan I make a fetch upstream and all this odd stuff is gone. But now i realise that "internal.Size" was removed, can be?
> > 
> > 
> > Yes, this was removed and replaced by `config.Size` you should use instead.
> 
> With config.Size build ok in my local repository. I can capture de telemetry again. But the CircleCI verifications fails in make check.
> I try to run it localy and i recive the same error in go vet.
> 
> ```shell
> make check
> go vet $(go list ./... | grep -v ./plugins/parsers/influx)
> # github.com/influxdata/telegraf/plugins/inputs/huawei_routers_telemetry
> vet: plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry_test.go:19:45: too few arguments in call to HuaweiTelemetryDecoder
> 
> go vet has found suspicious constructs. Please remediate any reported errors
> to fix them before submitting code for review.
> make: *** [Makefile:129: vet] Error 1
> ```
> 
> I dont have any clue. What should i change?

i have open my eyes and notice that the issue is in the test file. I was looking in the main file. 
Sorry

DamRCorba:
@srebhan  the test-go-1_16 fails in another plugin test.

```bash
ok      github.com/influxdata/telegraf/plugins/inputs/huawei_routers_telemetry  (cached)
....
....
...
--- FAIL: TestSimpleReverseLookup (0.89s)
    reversedns_test.go:47: 
                Error Trace:    reversedns_test.go:47
                Error:          Not equal: 
                                expected: "localhost"
                                actual  : "localhost.localdomain."
                            
                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1 +1 @@
                                -localhost
                                +localhost.localdomain.
                Test:           TestSimpleReverseLookup
FAIL
FAIL    github.com/influxdata/telegraf/plugins/processors/reverse_dns   1.059s
```

srebhan:
> @srebhan the test-go-1_16 fails in another plugin test.

It reports an unrelated issue in an unrelated (`reversedns`) plugin. I have no idea on why this happens. You can ignore it for now if you did not modify that plugin.

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
srebhan: I'm not sure core-devs will accept this...

srebhan:
@DamRCorba please let me know if when you are ready for another review round.

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
DamRCorba: A middle name option: huawei_telemetry_listener?

DamRCorba:
> 
> 
> @DamRCorba please let me know if when you are ready for another review round.

Ok Srebhan. I making some tests in a production environment, including the sensors of the newest version of routers. I put all the translations in the plugin and i left only the sensor descriptions in my github. I have to clean the code and erase or translate some comments.

I have to modify a test.go, i remember that you told me that the tests are not correct. Where can i read some docs about plugin tests?

plugins/inputs/huawei_routers_telemetry/huawei_routers_telemetry.go
srebhan: Fine with me.

srebhan:
You can take a look at PR #9409 where I tried to show some nice examples on how to do things. You might also want to take a look in other plugins (e.g. `plugins/processors/starlark`).

Just drop me a note once you are ready. ;-)

DamRCorba:
> 
> 
> @DamRCorba any update?
hi  @srebhan , sorry for the delay. I have somre troubles with the interpretation of new sensors in the newest realeses. Huawei chaged them all. I gonna clean the config for that sensors and limit the plugin to V8R10 router version.

srebhan:
@DamRCorba did you make any progress on the changed data?

Issue #7105, opened 2020-03-03
Ability to set http header values in the http_listener_v2 as tags


dneray:
## Feature Request

Opening a feature request kicks off a discussion.

### Proposal:
I would like to be able to use http headers in the http_listener_v2 as tags in the metric data.

### Desired behavior:
Something like this would work for me:

http_header_tags = {"SOME_HTTP_HEADER" = "MEASUREMENT_TAG_NAME"}

### Use case: 
I have a system where edge devices are communicating with a reverse proxy that handles authentication and the traffic is being forwarded to telegraf. 

Device ----> Reverse Proxy ----> Telegraf ----> Influxdb

For my use case, I would like to pass an identifier for the device from the reverse proxy. It's difficult to perform this kind of transformation in the reverse proxy through modifying the payload but it's easy to pass an http header so it would be convenient if I could do this transformation in telegraf.

danielnelson:
I think this would be a good addition, we are about to make a fairly large update to this plugin in #6974 but after that pr is merged would you be able to work on it?

dneray:
Yes, I have implemented it locally and will push after I finish testing but I will wait to open a PR until after #6974 is finished so that I can handle the conflicts myself.

Thanks!

dneray:
Have implemented my change here https://github.com/dneray/telegraf now just waiting for #6974 to open the pr

PR #7113, opened 2020-03-04
Add ability to specify HTTP headers in influxdb_listener which will b…

dneray:
…e transformed into tags and added to each metric (#7105)

### Required for all PRs:

### DIFF
--- a/plugins/inputs/influxdb_listener/influxdb_listener_test.go
+++ b/plugins/inputs/influxdb_listener/influxdb_listener_test.go
@@ -155,6 +155,75 @@ func TestWriteBasicAuth(t *testing.T) {
 	require.EqualValues(t, http.StatusNoContent, resp.StatusCode)
 }
 
+func TestWriteHTTPTransformHeaderValuesToTagsSingleWrite(t *testing.T) {
+	listener := newTestListener()
+	listener.HTTPHeaderTags = map[string]string{"Present_http_header_1": "presentMeasurementKey1", "Present_http_header_2": "presentMeasurementKey2", "NOT_PRESENT_HEADER": "notPresentMeasurementKey"}
+
+	acc := &testutil.Accumulator{}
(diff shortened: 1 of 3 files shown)


plugins/inputs/influxdb_listener/influxdb_listener_test.go
```diff
@@ -155,6 +155,68 @@ func TestWriteBasicAuth(t *testing.T) {
...
+	resp.Body.Close()
+	require.EqualValues(t, 204, resp.StatusCode)
+
+	acc.Wait(1)
+	acc.AssertContainsTaggedFields(t, "cpu_load_short",
+		map[string]interface{}{"value": float64(12)},
+		map[string]string{"host": "server01", "presentMeasurementKey1": "PRESENT_HTTP_VALUE_1", "presentMeasurementKey2": "PRESENT_HTTP_VALUE_2"},
+	)
+
+	// post single message to listener with a database tag in it already. It should be clobbered.
```
danielnelson: Could you split each test case into a separate tests or use a table test?

https://github.com/golang/go/wiki/TableDrivenTests
https://github.com/influxdata/telegraf/blob/7b03a6f4abbf96cbb0c5fc7497071b34c19c9ee8/plugins/inputs/influxdb_listener/influxdb_listener_test.go#L549-L589

dneray: I split the single write and bulk write into separate tests, I think that makes more sense here than using a table test

danielnelson:
I'm having some second thoughts about adding this to the `influxdb_listener` since this plugin is meant to operate as closely as possible to the the InfluxDB API.  What do you think about adding this instead to the `http_listener_v2` input?  When this plugin is setup with `data_format = "influx"` it behaves quite similarly to `influxdb_listener`.

dneray:
I agree, closing this one

PR #7140, opened 2020-03-09
Apply ping deadline to dns lookup

### LINKED ISSUE #5059: Ping timetouts not respected on DNS lookups.
### Relevant telegraf.conf:
```
[[inputs.ping]]
  interval = "2s"
  urls = ["bing.com","google.com"]
  count = 1
  timeout = 1.0
  deadline = 1
```

### System info:

Ubuntu 18.04
Telegraf 1.8.3

### Steps to reproduce:

1. Setup ping telegraf and ping plugin with short timeouts (e.g., 1s)
2. Ping a host that is not resolvable.

### Expected behavior:
An error is collected in under 1 second.

### Actual behavior:
An error is eventually collected, but long after the 1s timeout, resulting in missed batches due to collection taking too long.

### Additional info:
This issue was brought up in https://github.com/influxdata/telegraf/issues/3333, but it seems the issue got distracted with how to present DNS errors and the eventual "fix" for the issue didn't actually include addressing the ping timeout problem, though it did address the DNS lookup error problem.

Perhaps it is possible to have a global timeout for the ping operation in Go (both for the DNS lookup and the ping combined) and if the timeout is reached the goroutine is killed and an error is collected.  At the moment, there are a number of things that can timeout, but in many cases if it takes longer than some well known amount of time (e.g., 1 second) then from the user's point of view it is an error and should be reported as such, regardless of what specifically went wrong.  This would resolve the collection sometimes taking longer than interval, thus you would always get _something_ each collection period, even if it was just an error.

dbutler-starry:
This includes dns lookup in ping deadline. The deadline context is used for performing dns lookups, thus a ping collection will fail if the dealine times out during a lookup

Fixes #5059

### Required for all PRs:

### DIFF
--- a/plugins/inputs/ping/ping.go
+++ b/plugins/inputs/ping/ping.go
@@ -123,13 +123,6 @@ func (p *Ping) Gather(acc telegraf.Accumulator) error {
 	}
 
 	for _, host := range p.Urls {
-		_, err := net.LookupHost(host)
-		if err != nil {
-			acc.AddFields("ping", map[string]interface{}{"result_code": 1}, map[string]string{"url": host})
-			acc.AddError(err)
-			continue
-		}
-
 		p.wg.Add(1)
 		go func(host string) {
 			defer p.wg.Done()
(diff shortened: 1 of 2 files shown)


dbutler-starry:
Had trouble coming up with a good way to test the timeout but open to suggestions

dbutler-starry:
Modified to use `Resolver`. Opted for a method (similar to `pingHost`) instead of an interface. Let me know if this works, I'll squash the commits

dbutler-starry:
@danielnelson Is this what you had in mind?

PR #7141, opened 2020-03-09
Add support for setting retention policy using tag

### LINKED ISSUE #7053: Allow RP to be set from a tag
Currently we have the ability to specify `database_tag = ""` in the influxdb output plugin. It would be nice if we could specify `rp_tag = ""` so we could dynamically send data to a specific RP based on a tag

Reference #397

danielnelson:
closes #7053

### Required for all PRs:

### DIFF
--- a/plugins/outputs/influxdb/http.go
+++ b/plugins/outputs/influxdb/http.go
@@ -83,21 +83,23 @@ func (r WriteResponse) Error() string {
 }
 
 type HTTPConfig struct {
-	URL                  *url.URL
-	UserAgent            string
-	Timeout              time.Duration
-	Username             string
-	Password             string
-	TLSConfig            *tls.Config
-	Proxy                *url.URL
-	Headers              map[string]string
-	ContentEncoding      string
-	Database             string
-	DatabaseTag          string
-	ExcludeDatabaseTag   bool
-	RetentionPolicy      string
-	Consistency          string
-	SkipDatabaseCreation bool
+	URL                       *url.URL
+	UserAgent                 string
+	Timeout                   time.Duration
+	Username                  string
+	Password                  string
+	TLSConfig                 *tls.Config
+	Proxy                     *url.URL
+	Headers                   map[string]string
+	ContentEncoding           string
+	Database                  string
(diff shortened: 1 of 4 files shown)


plugins/outputs/influxdb/http.go
```diff
@@ -236,55 +238,66 @@ func (c *httpClient) CreateDatabase(ctx context.Context, database string) error
 	}
 }
 
+type dbrp struct {
+	Database        string
+	RetentionPolicy string
+}
```
ssoroka: can you have the same database with different retention policies? The map below would seem to imply so

plugins/outputs/influxdb/http.go
```diff
@@ -236,55 +238,66 @@ func (c *httpClient) CreateDatabase(ctx context.Context, database string) error
...
+		if err != nil {
+			return err
+		}
 	}
 	return nil
 }
 
-func (c *httpClient) writeBatch(ctx context.Context, db string, metrics []telegraf.Metric) error {
-	url, err := makeWriteURL(c.config.URL, db, c.config.RetentionPolicy, c.config.Consistency)
+func (c *httpClient) writeBatch(ctx context.Context, db, rp string, metrics []telegraf.Metric) error {
```
ssoroka: rp might not be super clear as retentionPolicy

plugins/outputs/influxdb/http.go
danielnelson: Yes, this is the most common case actually.  I conceptualize this by thinking about all points being inside a database+retention_policy (not just the database).  If no rp is set, values will go into the default `autogen` rp.

This means you could have 2 exactly the same values at the same timestamp in the same database, and only by querying with the full "dbrp" will you be able to tell them apart.  It can be confusing to see because normally the retention policy is not seen.

plugins/outputs/influxdb/http.go
danielnelson: Agreed, but I think it should be okay as it matches InfluxDB lingo https://docs.influxdata.com/influxdb/v1.7/concepts/glossary#retention-policy-rp

PR #7161, opened 2020-03-12
Bugfix - SqlServer DatabaseIO

Trovalo:
### Required for all PRs:

- [x] Signed [CLA](https://influxdata.com/community/cla/).
- [ ] Associated README.md updated.
- [ ] Has appropriate unit tests.

As noticed yesterday by @denzilribeiro, the last PR created an issue for Azure SQL DB.
It was caused by multiple columns with the same name (one was there by accident).

More details [here](https://github.com/influxdata/telegraf/pull/7103#issuecomment-597844550)

This PR removes the "duplicated" column and another detail noticed by @denzilribeiro. (he wanted to make the PR, but the next release is imminent and I had time to do it today)

It has been tested with a combo of SqlServer 2012,  SqlServer 2019 and Azure SQL DB (all in the same config)

PR #7163, opened 2020-03-12
Adding Database IO Tempdb per Azure DB collection, Fixing PerfmonV2 collection for on-prem

denzilribeiro:
### Required for all PRs:

Datababase IO changes
Today on Azure SQL DB there was a bug where left outer join from virtual file stats and database_files was only on file_id. Tempdb was exposed on virtual file stats, but not in Database files. On Azure SQL DB have to use tempdb.sys.database_files. In addition each Azure SQL DB has its own tempdb so needed to add Database_id/File ID to be able to distinguish so that reporting can be done by Logical Server + Azure DB + database_id.  If you just looked at tempdb for a logical server would get multiple tempdb's belonging to different Azure SQL DB's as the logical server construct is just that a placeholder logical server not actual physical server.
Also corrected bug introduced earlier

PerfmonV2 Fix for [#7164](https://github.com/influxdata/telegraf/issues/7164)

### DIFF
--- a/plugins/inputs/sqlserver/sqlserver.go
+++ b/plugins/inputs/sqlserver/sqlserver.go
@@ -359,24 +359,40 @@ BEGIN
 	SELECT
 		 ''sqlserver_database_io'' As [measurement]
 		,REPLACE(@@SERVERNAME,''\'','':'') AS [sql_instance]
-		,DB_NAME([vfs].[database_id]) AS [database_name]
+		,DB_NAME() as database_name
+		,vfs.database_id   -- /*needed as tempdb is different for each Azure SQL DB as grouping has to be by logical server + db_name + database_id*/
+		,vfs.file_id
 		,vfs.io_stall_read_ms AS read_latency_ms
 		,vfs.num_of_reads AS reads
 		,vfs.num_of_bytes_read AS read_bytes
 		,vfs.io_stall_write_ms AS write_latency_ms
 		,vfs.num_of_writes AS writes
 		,vfs.num_of_bytes_written AS write_bytes
-		,vfs.io_stall_queued_read_ms as rg_read_stall_ms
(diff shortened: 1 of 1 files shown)


denzilribeiro:
@Trovalo  overlapped with your PR [#7161](https://github.com/influxdata/telegraf/pull/7161) just submitted as I was in process of submitting didn't realize - can you take a look at this one and see if addresses issues as I also added additional logic to pull tempdb IO for Azure SQL so is a bit more than the bug fix

Trovalo:
I will have a look at it in a few minutes.

But I found a bigger problem while testing the whole plugin on multiple versions earlier.
While testing on SQL 2012. SQL 2019 and Azure SQL DB, I get an error
>I! Invalid column name 'physical_database_name'.
E! [inputs.sqlserver] Error in plugin: mssql: Invalid column name 'physical_database_name'.

It is caused by the latest changes (January) made to the query and not yet published (in fact those are not in telegraf 1.13.4).
The field [physical_database_name] exists only in azure SQL DB and SQL Server 2019. Any other version will fail to gather the metrics due to a compile error.

I'm opening an issue now, the best solution probably is to add dynamic SQL, so it is not validated by SQL Server until it has to execute it

denzilribeiro:
I will add the dynamic SQL and resubmit shortly

denzilribeiro:
@Trovalo  - actually go ahead and open an issue for that as is in a different part not IO :) misunderstood

Trovalo:
I did it few minutes ago
https://github.com/influxdata/telegraf/issues/7164

denzilribeiro:
@Trovalo - one last validation - pushed commit for the Perfmon counters too

Trovalo:
I will run the tests tomorrow morning and let you know

PR #7173, opened 2020-03-13
Fix GNMI TypedValue to scalar value conversion and update supported platforms

### LINKED ISSUE #7167: Handling of Decimal64 for gnmi
### Relevant telegraf.conf:

```toml

In Yang we have datatype of decimal64 and this data type sent to Influxdb or ES is null instead of proper value. This will cause issue for openconfig models with data type like below


grouping avg-min-max-instant-stats-precision1 {
--
  | description
  | "Common grouping for recording an instantaneous statistic value
  | in addition to avg-min-max stats";
  |  
  | leaf instant {
  | type decimal64 {
  | fraction-digits 1;
  | }
  | description
  | "The instantaneous value of the statistic.";
  | }



```

### Steps to reproduce:

1. gnmi subscribe openconfig model with output plugin as ES or influxdb for any container having property type of decimal64
2. ...

### Fix
```toml

@@ -318,6 +319,11 @@ func (c *CiscoTelemetryGNMI) handleTelemetryField(update *gnmi.Update, tags map[   
                value = val.BytesVal                                                                    
        case *gnmi.TypedValue_DecimalVal:                                                               
                value = val.DecimalVal                                                                   
+                dig := float64(val.DecimalVal.Digits)                                                  
+                prec := float64(val.DecimalVal.Precision)                                               
+                fprec := math.Pow(10, prec)                                                            
+                fv := dig / fprec                                                                      
+                value = fv                                                                              
        case *gnmi.TypedValue_FloatVal:     
```

sbyx:
This uses the GNMI library native ToScalar function to convert GNMI TypedValues to scalar Go values.
Also adds an updated note on tested device platforms.

Fixes #7167

### DIFF
--- a/plugins/inputs/cisco_telemetry_gnmi/cisco_telemetry_gnmi.go
+++ b/plugins/inputs/cisco_telemetry_gnmi/cisco_telemetry_gnmi.go
@@ -7,6 +7,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"math"
 	"net"
 	"path"
 	"strings"
@@ -332,7 +333,7 @@ func (c *CiscoTelemetryGNMI) handleTelemetryField(update *gnmi.Update, tags map[
 	case *gnmi.TypedValue_BytesVal:
 		value = val.BytesVal
 	case *gnmi.TypedValue_DecimalVal:
-		value = val.DecimalVal
+		value = float64(val.DecimalVal.Digits) / math.Pow(10, float64(val.DecimalVal.Precision))
 	case *gnmi.TypedValue_FloatVal:
 		value = val.FloatVal
 	case *gnmi.TypedValue_IntVal:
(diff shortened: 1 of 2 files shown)


plugins/inputs/cisco_telemetry_gnmi/cisco_telemetry_gnmi.go
```diff
@@ -324,23 +323,9 @@ func (c *CiscoTelemetryGNMI) handleTelemetryField(update *gnmi.Update, tags map[
 		return aliasPath, nil
 	}
 
+	value, _ := gnmivalue.ToScalar(update.Val)
```
danielnelson: Check the error and also there are some value types we should probably skip, such as []byte.  The field needs to be one of int64, float64, uint64, string, bool, there is some massaging that is done when acc.AddField is called but that was done mostly to avoid errors with older plugins, we try do it in the plugin now.

Is this going to work with the JsonVal/JsonIetfVal types?  Maybe we should do the ToScaler as the default case of the switch?

sbyx: I see your points, I therefore reverted those changes and just changed the handling for the decimalval type. I think in hindsight it would be more trouble switching to the native casting function since we would then need to deal with the error cases as it also does not deal with the Json types which we would need to handle manually anyway.

PR #7177, opened 2020-03-16
Omit retained messages for mqtt_consumer plugin

dersimn:
Solves an issue when dealing with retained messages.
Without this option being set to true, retained messages are stored twice into a database, when Telegraf is restarted. If you don't want this behaviour, set the new option to true. Messages received by the broker as retained are omitted (and thus not stored twice).

### Required for all PRs:

### DIFF
--- a/plugins/inputs/mqtt_consumer/README.md
+++ b/plugins/inputs/mqtt_consumer/README.md
@@ -51,6 +51,15 @@ and creates metrics using one of the supported [input data formats][].
   ## publishing.
   # persistent_session = false
 
+  ## Omit retained messages.
+  ## If this value is set to true, retained messages are not sent to the output
+  ## plug-ins. This prevents duplicate messages from being saved when Telegraf is
+  ## restarted. However, this also prevents the last state from being received when
+  ## Telegraf is first started. This only affects messages that were sent by the
+  ## broker as retained messages. Messages received during an active subscription
+  ## are still processed as usual, regardless if the retained flag is set or not.
+  # omit_retained = false
+
   ## If unset, a random client ID will be generated.
   # client_id = ""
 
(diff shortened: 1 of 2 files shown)


danielnelson:
Can you add some more information about your use case?  Why do you need to use retained messages when you don't want them proccessed, why not just clear the retained message?

dersimn:
First you need to know that you can _send_ a message as a retained message and _receive_ a retained message (meaning that the retain flag is set to true/false).
If you have subscribed to a topic and someone _sends_ a message with `retain = true` on that topic, the broker will send a regular message to you (`retain = false`).  You _receive_ a message with `retain = true` only if you have subscribed to a topic and the Broker sends you an older message (from the time before you subscribed), no matter why you didn't subscribe before (one reason might be, that the subscriber needed to restart).

Given this, it is always useful to omit retained messages as a recipient if you do not want duplicates stored in your database, but other clients depend on them. I use retained messages for battery powered sensors that only send an update every hour or so.  
Other clients who subscribe to the sensor after a recent measurement should still receive the latest sensor value, but I don't want to store duplicates in a database when I need to restart Telegraf or the MQTT broker for maintenance purposes.  I think it is less likely that a value will be missed in a rather short restart compared to the "damage" caused by duplicate values in further processing (because every data point indicates e.g. that the sensor was still alive at a certain point in time).

Alternatively, I could save the retained flag as a tag and omit such data points when processing 🤔.

sjwang90:
@dersimn We still need the appropriate unit tests before we can review. Thanks.

srebhan:
@dersimn any updates on this PR?

sjwang90:
@dersimn Please let us know if you'd like continue working on this PR or else we'll close your draft.

PR #7179, opened 2020-03-16
update docker files for go modules. Closes #7170

### LINKED ISSUE #7170: Docker build doesn't work, since master branch doesn't use dep anymore dependencies.
Docker file needs to be fixed, so we can compile our own plugins and build docker image.
Compiling custom built plugins using the docker file https://github.com/influxdata/telegraf/blob/master/scripts/stretch.docker does not work, fails trying to copy Gopkg.toml and Gopkg.lock.

Learnt that, Master branch does not use dep any more and has moved to go modules.

MacOS High Sierra
Docker 2.1

ssoroka:
### Required for all PRs:

### DIFF
--- a/scripts/stretch.docker
+++ b/scripts/stretch.docker
@@ -1,9 +1,6 @@
-FROM golang:1.11.0 as builder
-ENV DEP_VERSION 0.5.0
-RUN curl -fsSL -o /usr/local/bin/dep https://github.com/golang/dep/releases/download/v${DEP_VERSION}/dep-linux-amd64 && chmod +x /usr/local/bin/dep
+FROM golang:1.14.0 as builder
 WORKDIR /go/src/github.com/influxdata/telegraf
-COPY Gopkg.toml Gopkg.lock ./
-RUN dep ensure -vendor-only
+
 COPY . /go/src/github.com/influxdata/telegraf
 RUN make go-install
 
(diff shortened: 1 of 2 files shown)


scripts/stretch.docker
```diff
@@ -1,9 +1,6 @@
-FROM golang:1.11.0 as builder
-ENV DEP_VERSION 0.5.0
-RUN curl -fsSL -o /usr/local/bin/dep https://github.com/golang/dep/releases/download/v${DEP_VERSION}/dep-linux-amd64 && chmod +x /usr/local/bin/dep
+FROM golang:1.13.8 as builder
 WORKDIR /go/src/github.com/influxdata/telegraf
-COPY Gopkg.toml Gopkg.lock ./
-RUN dep ensure -vendor-only
+
```
danielnelson: I wonder why we had the dependency downloading as a separate step.  Perhaps this helps with layer caching when you rebuild the image?

ssoroka: I assumed it was required with dep. I guess the change is that I'm no longer venturing it, so there's no ability to reuse the vendored step. We could bring this back with a call to `RUN go mod vendor` right before.

ssoroka: I think the intermediate builds get removed anyway, so I'm not sure this would have any impact on release size.

PR #7198, opened 2020-03-18
Remove init checks for valid label and field selectors

danielnelson:
Importing these libraries starts a goroutine at init time.  The removal of the checks will push a startup time error to a runtime error.

```
goroutine 8 [chan receive]:
k8s.io/klog.(*loggingT).flushDaemon(0x454f3c0)
        /home/dbn/src/influx/pkg/mod/k8s.io/klog@v1.0.0/klog.go:1010 +0x8b
created by k8s.io/klog.init.0
        /home/dbn/src/influx/pkg/mod/k8s.io/klog@v1.0.0/klog.go:411 +0xd6
```

related #6969

### Required for all PRs:


danielnelson:
cc @mg03

mg03:
@danielnelson : thank you

qs: I'm assuming the goroutine launched by klog is not an acceptable. can you please help me understand why ?

danielnelson:
That's right, it appears that this goroutine is launched during package init so even if you are not using the prometheus input it is started.  You can imagine that if every plugin did this it would be quite a mess.  I haven't looked into why this goroutine is launched, if it were optional or if we could control when it starts and stops we could add these checks back.

mg03:
@danielnelson : I agree and have reached out to the creator of the packages on GitHub and asked them the qs of either not importing the dependency or control the start/stop of the goroutine. In the meantime, runtime failure should be acceptable since the code snippets were testing input format and should not have any detrimental effect on usability of the PR I submitted initially.

PR #7201, opened 2020-03-19
CloudLinux/LVE (/proc/lve/list) input plugin

DSpeichert:
Input plugin to scrape CloudLinux LVE status format from `/proc/lve/list` (https://docs.cloudlinux.com/command-line_tools/#lvetop)

It works very much like CSV input but also resolved username from Linux user ID, which CSV input plugin wouldn't do.

### Required for all PRs:

### DIFF
--- a/plugins/inputs/all/all.go
+++ b/plugins/inputs/all/all.go
@@ -82,6 +82,7 @@ import (
 	_ "github.com/influxdata/telegraf/plugins/inputs/logparser"
 	_ "github.com/influxdata/telegraf/plugins/inputs/logstash"
 	_ "github.com/influxdata/telegraf/plugins/inputs/lustre2"
+	_ "github.com/influxdata/telegraf/plugins/inputs/lve"
 	_ "github.com/influxdata/telegraf/plugins/inputs/mailchimp"
 	_ "github.com/influxdata/telegraf/plugins/inputs/marklogic"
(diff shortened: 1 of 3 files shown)


danielnelson:
How about doing this as a `user_id` processor, for example we could have:
```
[[processors.uid]]
  tag = "uid"
  dest = "user"
```

It would transform:
```
foo,uid=1000 tno=42i
```

into:
```
foo,uid=1000,user=DSpeichert tno=42i
```

This would allow you to use the `file` input with `csv` parser for the initial collection, and optionally resolve the user_id.

If this ends up being very simple we could add it to Telegraf, otherwise we might recommend either the `exec` or `execd` inputs, or in the future we hope to have a `exec/execd` processor plugin as well.

sjwang90:
@DSpeichert Just checking in if you're still working on this PR

DSpeichert:
@sjwang90 Not actively. I'm not using it and I haven't caught up on current processing options (like mentioned above) that might render this plugin unnecessary. I assume a `uid` processor does not actually exist yet?

sjwang90:
@DSpeichert Not sure specifically what you're looking from a UID processor but there isn't a standalone processor for that but we do have a [Starlark processor](https://github.com/influxdata/telegraf/blob/master/plugins/processors/starlark/README.md) where you could script something that could be useful.

PR #7210, opened 2020-03-21
Add limit to number of undelivered lines to read ahead in tail

danielnelson:
This avoids the input reading too far ahead and filling the metric buffer.  Particularly useful when doing a batch import with `from_beginning = true`.

### Required for all PRs:

### DIFF
--- a/plugins/inputs/tail/tail.go
+++ b/plugins/inputs/tail/tail.go
@@ -270,9 +270,10 @@ func (t *Tail) receiver(parser parsers.Parser, tailer *tail.Tail) {
 			metric.AddTag("path", tailer.Filename)
 		}
 
+		// Block until plugin is stopping or room is available to add metrics.
 		select {
 		case <-t.ctx.Done():
-			break
+			return
 		case t.sem <- empty{}:
 			t.acc.AddTrackingMetricGroup(metrics)
 		}
--- a/plugins/inputs/tail/tail.go
+++ b/plugins/inputs/tail/tail.go
@@ -3,6 +3,8 @@
 package tail
 
 import (
+	"context"
+	"errors"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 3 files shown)


plugins/inputs/tail/tail.go
```diff
@@ -229,21 +268,18 @@ func (t *Tail) receiver(parser parsers.Parser, tailer *tail.Tail) {
...
-			t.acc.AddMetric(metric)
 		}
-	}
-
-	t.Log.Debugf("Tail removed for %q", tailer.Filename)
 
-	if err := tailer.Err(); err != nil {
-		t.Log.Errorf("Tailing %q: %s", tailer.Filename, err.Error())
+		select {
+		case <-t.ctx.Done():
```
ssoroka: I think it's possible for this plugin to have the context signal that it's cancelled and then the plugin doesn't exit because it's tailing a file with very infrequent updates. Probably don't need to do anything about this, but I figured I'd point it out.

plugins/inputs/tail/tail.go
```diff
@@ -229,21 +268,18 @@ func (t *Tail) receiver(parser parsers.Parser, tailer *tail.Tail) {
...
+			break
+		case t.sem <- empty{}:
```
ssoroka: in a select, this won't block, it'll drop the metric. I assume that's what you wanted here.

ssoroka: maybe a comment to that effect will make this obvious for anyone in the future less familiar with channel/select semantics.

plugins/inputs/tail/tail.go
danielnelson: In Stop() we close the Tailer, this should make tailer.Lines return immediately.

plugins/inputs/tail/tail.go
danielnelson: This should block because there is no default case.

danielnelson: On 275 I should return, not break.

PR #7233, opened 2020-03-25
Fixed vSphere 6.7 data loss issues

prydin:
### Required for all PRs:

vSphere 6.7 introduced some potential data loss issues due to the way it sometimes delays metrics. The main issue is that different resources can be delayed by different amounts of time and we need to keep track of the latest good sample per resource and metric.


prydin:
@danielnelson Can you have a look at why this is failing the ```make tidy``` step on the CI checks? It works when I run a ```make tidy``` on my machine.

danielnelson:
I think maybe you are using an older version of Go, does it help if you use Go 1.13.8 to run `go mod tidy`?

prydin:
That seemed to be the issue. Thanks!

PR #7239, opened 2020-03-27
set GOMAXPROCS in container env.

jxiewei:
set GOMAXPROCS in container env. 
In my test setup with 24core machine, telegraf runing  in docker with --cpus 1,  a simple pipeline of   logparser input and file output, I see 30% performance gain.


danielnelson:
Thanks for the pull request, but I don't think we would merge.  You can check and set this outside the process before startup.

PR #7242, opened 2020-03-27
FIX: Perfmon collector adding OPTION RECOMPILE for perf reasons

### LINKED ISSUE #7236: SqlServer: Massive performance degradation in PerformanceCounters Query
### Relevant telegraf.conf:

```toml
[[inputs.sqlserver]]
  ## Specify instances to monitor with a list of connection strings.
  ## All connection parameters are optional.
  ## By default, the host is localhost, listening on default port, TCP 1433.
  ##   for Windows, the user is the currently running AD user (SSO).
  ##   See https://github.com/denisenkom/go-mssqldb for detailed connection
  ##   parameters.
  servers = [
    "Server=REDACTED"
  ]

  ## Optional parameter, setting this to 2 will use a new version
  ## of the collection queries that break compatibility with the original
  ## dashboards.
  query_version = 2
```

### System info:

- Telegraf v1.14.0-rc1
- SQL Server 2014/2016 on Windows Server

### Steps to reproduce:

1. have a relatively busy SQL instance (e.g. 500 databases and 40 resource pools/groups)
2. Use SQL profiler observe Execution times of PerformanceCounters queries

### Expected behavior:

Similar performance to the telegraf version we used previously (from around mid 2019): in the ballpark on 500-800ms query duration for a busy server

### Actual behavior:

Query durations between 60-90 seconds on the same busy server.

### Additional info:

Running the old/new versions of the PerformanceCounters query manually shows the same behaviour: old query sub-second duration, new query 60+ seconds.

I've done some troubleshooting and it seems that the issue is due to a removed `OPTION (RECOMPILE)` query hint in the following commit: https://github.com/influxdata/telegraf/commit/1df88dd22bf372327f24341b4043bb383e33529c#diff-5359fe28fcc15fe8b504f2d8033ce59dL662

Adding it back leads to similar execution times than before. Some quick research leads me to believe that the execution plan is optimized very poorly without that option as the row estimates are vastly off.

I wonder why it was removed and if there is any reason to not add it back.

denzilribeiro:
Peformance Fix for #7236 
### Required for all PRs:

Resolves #7236

### DIFF
--- a/plugins/inputs/sqlserver/sqlserver.go
+++ b/plugins/inputs/sqlserver/sqlserver.go
@@ -790,7 +790,8 @@ FROM	@PCounters AS pc
 			AND pc.object_name = pc1.object_name
 			AND pc.instance_name = pc1.instance_name
 			AND pc1.counter_name LIKE '%base'
-WHERE	pc.counter_name NOT LIKE '% base';
+WHERE	pc.counter_name NOT LIKE '% base'
+OPTION(RECOMPILE);
 `
 
 // Conditional check based on Azure SQL DB v/s the rest aka (Azure SQL Managed instance OR On-prem SQL Server)


danielnelson:
Can we make this a more patch release friendly by only including the performance change without the new query type?

denzilribeiro:
Only pushing perf fix

Issue #5766, opened 2019-04-25
[New Plugin] Nightscout


mattyjones:
## Feature Request

This is for a new plugin called *nightscout* that I would like to submit.

### Proposal:

There is an open source project called [Nightscout](http://www.nightscout.info/) that helps with monitoring the blood glucose level of Type 1 Diabetics. I would like to contribute my plugin that reads the data from their api and drops it into InfluxDB. This would be hugely helpful for families and groups that want to track this data from a single place such as a grafana dashboard rather than a separate site for each person.

### Current behavior:

Nothing like this is currently written or available

### Desired behavior:

1. Pull json data from a remote or local REST endpoint
1. Normalize the data by adding a field to describe the source (person)
1. Push the data to InfluxDB

### Use case: 

I have two young kids that are T1D and looking at multiple browser windows for each child is a pia. By pulling the data from N people into a single place using a common format I can then use other parts of the TICK stack to generate dashboards, alerts, or further enhance the functionality of monitoring and alerting.

Currently I have hardware devices that are reading streaming data via micropython and I would like to further work on these projects as well, but would like a consistent interface and endpoint with which to pull multiple streams from.

![58373106_10156800417081634_8519009340319858688_o](https://user-images.githubusercontent.com/672940/56701338-91fab900-66cc-11e9-866a-590a64b7e015.jpg)

danielnelson:
Sounds like it would be a really nice addition, go ahead an open up a PR.  Do you have links to the API documentation?

mattyjones:
@danielnelson The api docs are locally built so I would have to provide a link to a live site that I am using which I am kinda not cool with ;) I will see what I can do though about creating a copy of at least the initial endpoints I am going to use for this plugin.

danielnelson:
Okay, test cases will be enough to show how it works I think.

mattyjones:
![D503YbkWAAUW4Cy](https://user-images.githubusercontent.com/672940/57199241-ced77300-6f4a-11e9-82d7-c4cb8cf1b001.jpg)

So I finally have it fully hooked up and working. I will be opening a pr at this point later today as a WIP as I think there are several more things to be done on this, tests and docs included but it is stable and functioning.

twentythousandphantoms:
How it's going?

mattyjones:
@twentythousandphantoms I dropped the ball on this, I will poke at my code and dig it up. I think it was fairly close and I just needed tests.

sspaink:
[...]

PR #7249, opened 2020-03-28
[WIP] Feature/nightscout plugin

mattyjones:
### Required for all PRs:

This is in reference to #5766


mattyjones:
The only thing left on this is tests and review. I have no idea how to write tests for this at the moment. I am working through trying to figure that out. Once I have completed that I will drop the *[WIP]* tag from the title and we can hopefully start the review process at some point.

plugins/inputs/nightscout/README.md
```diff
@@ -0,0 +1,1142 @@
...
+```toml
+  ## This is a list of nightscout sites to pull data from.
+  ##  example: servers = ["https://my.nightscout.site/api/v1/entries"]
+  servers = []
+  ## The person is the owner of the nightscout site. The order of the list of owners must match the order of the
+  ## list of servers, for example the first server should match the first owner.
+  # owners = ""
+
+  ## The secret is the SHA-1 hash of the api secret used to access the site.
+  secret = ""
```
viq: I'd probably leave that commented out, to encourage people to use tokens.

plugins/inputs/nightscout/README.md
```diff
@@ -0,0 +1,1142 @@
...
+
+  ## The count is the number of results to pull at a time. Unless you have a specific need, this should remain as 1.
+  # count = "1"
+
+  ## The token is the unique token that matches a role within your nightscout site(s).
+  token = ""
```
viq: How does that work with multiple servers?

plugins/inputs/nightscout/nightscout.go
```diff
@@ -0,0 +1,198 @@
...
+	SGV        int    `json:"sgv"`
+	Direction  string `json:"direction"`
+	Noise      int    `json:"noise"`
+	Filtered   int    `json:"filtered"`
+	Unfiltered int    `json:"unfiltered"`
+	RSSI       int    `json:"rssi"`
+}
+
+// Nightscout holds configuration and query parameters.
+type Nightscout struct {
```
viq: Maybe instead of a flat structure with some lists in it, it would make more sense to have a list of dictionaries? So something like (YAML representation, can try to put it into something else if you prefer):
```yaml
nightscout:
  count: 1
  servers:
    - address: https://server1
      owner: owner1
      token: token1
    - address: https://server2
      owner: owner2
      secret: secret2
```

danielnelson: What I'd suggest for this is to use multiple plugin instances:
```toml
[[inputs.nightscout]]
  servers = ["https://server1"]
  token: "token1"

[[inputs.nightscout]]
  servers = ["https://server2"]
  token: "token2"
```

It may even be worth considering having only a single server url plugin instead of a list.  We should think about the typical pattern for connecting, does one usually connect to a single server or to a cluster?

danielnelson:
Looks like you need to rebase your branch to remove the unrelated changes.

For tests, I suggest using the httptest package included in the Go standard library.  Check out `plugins/outputs/http/http_test.go` for some examples.

plugins/inputs/nightscout/nightscout.go
mattyjones: @danielnelson I was unaware you could do multiple plugin instances. I am 100% for that and will refactor with that in mind. Thanks for the tip!

plugins/inputs/nightscout/README.md
mattyjones: Now that I am looking at it, it won't thanks for the tip. This is going to go away though in favor of simply using multiple plugin instances.

plugins/inputs/nightscout/README.md
mattyjones: It needs both the token and the secret. I will better document this. The secret is the password for the site, everyone needs it for any access at all. The token is specific to a user and provides a small measure of RBAC.
@danielnelson

plugins/inputs/nightscout/nightscout.go
mattyjones: I will also add more documentation in the readme to discuss use cases for both single person and multiple person uses.

For a single person, though you connect to a single server, there is no real concept of clustering for theses site. You could do something with a load balancer and share the MongoDB backend but I have not heard of anyone needing that kind of up, it may well exist though.

plugins/inputs/nightscout/README.md
viq: You can get data from your site via curl with only token, same with OpenAPS scripts. The API secret is a master-key-root-level-access to everything, without any further ACLs, as far as I'm aware.

mattyjones: Thanks for this, I made the changes and confirmed they work.

plugins/inputs/nightscout/README.md
mattyjones: Confirmed multiple plugin instances work as expected.

plugins/inputs/nightscout/nightscout.go
mattyjones: Made the changes, to include multiple plugins. I left in the slice for ranging through several servers per person in case they have something like that implemented.

mattyjones:
Created a new PR to keep the history clean. #7286

PR #7253, opened 2020-03-30
Sflow rework

ssoroka:
Reworked sflow plugin to separate the binary protocol parsing and the metric creation, and take some of the magic out of it so that anyone can read and extend it easily.


plugins/inputs/sflow/packetdecoder/packetdecoder.go
```diff
@@ -0,0 +1,506 @@
...
+	case "IPv4":
+		h.IPHeader, err = d.decodeIPv4Header(r)
+	case "IPv6":
+		h.IPHeader, err = d.decodeIPv6Header(r)
+	default:
+	}
+	if err != nil {
+		return h, err
+	}
+	// missing reading crc check bytes?
```
danielnelson: Let's figure this out and improve the comment.

plugins/inputs/sflow/types/decoder.go
```diff
@@ -0,0 +1,285 @@
...
+		"tcp_urgent_pointer": h.TCPUrgentPointer,
+		"tcp_window_size":    h.TCPWindowSize,
+	}
+}
+
+type UDPHeader struct {
+	SourcePort      uint16
+	DestinationPort uint16
+	UDPLength       uint16
+	Checksum        uint16
```
danielnelson: I noticed Checksum isn't used, should we remove it?

plugins/inputs/sflow/decoder_test.go
```diff
@@ -159,7 +202,7 @@ func TestExpandFlow(t *testing.T) {
 				"ip_total_length":    uint64(0x05a6),
 				"ip_ttl":             uint64(0x7e),
 				"sampling_rate":      uint64(0x0400),
-				"tcp_header_length":  uint64(0x40),
+				"tcp_header_length":  uint64(0x14),
```
danielnelson: Can we try to cross referenced this change with host-sflow, sflowtool/wireshark?

plugins/inputs/sflow/decoder_test.go
```diff
@@ -479,6 +524,8 @@ func TestIPv6SW(t *testing.T) {
 				"ether_type":       "IPv6",
 				"header_protocol":  "ETHERNET-ISO88023",
 				"input_ifindex":    "257",
+				"ip_dscp":          "0",
+				"ip_ecn":           "0",
```
danielnelson: These should remain fields, possibly in the future we could add a string representation as tags.

plugins/inputs/sflow/packetdecoder/packetdecoder.go
```diff
@@ -0,0 +1,506 @@
...
+	// 	 reserved 3 bits
+	// 	 flags 9 bits
+	var stuff uint16
+	if err := read(r, &stuff, "TCP Header Octet offset 12"); err != nil {
+		return h, err
+	}
+	h.TCPHeaderLength = uint8((stuff >> 12) * 4)
+	// I don't have a use for the flags, so just stuff them into one variable.
+	h.Flags = stuff & 0x1FF
+	// bit reading done
```
danielnelson: Do a little variable naming and comment cleanup here

plugins/inputs/sflow/metricencoder.go
```diff
@@ -0,0 +1,46 @@
...
+			if flowRecord.FlowData != nil {
+				tags2 := flowRecord.FlowData.GetTags()
+				fields2 := flowRecord.FlowData.GetFields()
+				for k, v := range tags {
+					tags2[k] = v
+				}
+				for k, v := range fields {
+					fields2[k] = v
+				}
+				m, err := metric.New("sflow", tags2, fields2, time.Now())
```
danielnelson: Call time.Now() once at the beginning of the function, and pass in here.  This way the whole packet has the same timestamp.

plugins/inputs/sflow/metricencoder.go
```diff
@@ -0,0 +1,46 @@
...
+	"strconv"
+	"time"
+
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/metric"
+	"github.com/influxdata/telegraf/plugins/inputs/sflow/types"
+)
+
+func makeMetrics(p *types.V5Format) ([]telegraf.Metric, error) {
+	metrics := []telegraf.Metric{}
```
danielnelson: Initialize to len(FlowRecords)?

plugins/inputs/sflow/types/decoder.go
```diff
@@ -0,0 +1,285 @@
...
+	return t
+}
+func (h IPV6Header) GetFields() map[string]interface{} {
+	var f map[string]interface{}
+	if h.ProtocolHeader != nil {
+		f = h.ProtocolHeader.GetFields()
+	} else {
+		f = map[string]interface{}{}
+	}
+	// f["ip_total_length"] = h.PayloadLength // is this the same?
```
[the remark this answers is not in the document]
ssoroka: wasn't sure if we wanted to stick this in the same field as the other header type

plugins/inputs/sflow/packetdecoder/packetdecoder.go
```diff
@@ -0,0 +1,506 @@
...
+	if err := read(r, &sam.SampleType, "SampleType"); err != nil {
+		return sam, err
+	}
+	sampleDataLen := uint32(0)
+	if err := read(r, &sampleDataLen, "Sample data length"); err != nil {
+		return sam, err
+	}
+	if sampleDataLen > MaxSampleDataLen {
+		return sam, fmt.Errorf("sample data length too large")
+	}
```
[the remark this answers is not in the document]
ssoroka: This was just to avoid reading too much with malicious packets, but Go won't let you do a buffer overflow vulnerability here anyway, even if the data length is absurdly large, so maybe it's not needed.

plugins/inputs/sflow/packetdecoder/packetdecoder.go
```diff
@@ -0,0 +1,506 @@
...
+	if err := read(r, &t.SequenceNumber, "SequenceNumber"); err != nil {
+		return t, err
+	}
+	var sourceID uint32
+	if err := read(r, &sourceID, "SourceID"); err != nil { // source_id line: 1622
+		return t, err
+	}
+	// split source id to source id type and source id index
+	t.SourceIDIndex = sourceID & 0x00ffffff // line: 1468
+	t.SourceIDType = sourceID >> 24         // source_id_type Line 1465
```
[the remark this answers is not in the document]
ssoroka: sensible idea. Will do.

plugins/inputs/sflow/packetdecoder/packetdecoder.go
```diff
@@ -0,0 +1,506 @@
...
+
+	if err := read(r, &t.OutputIfIndex, "OutputIfIndex"); err != nil {
+		return t, err
+	}
+	t.OutputIfFormat = t.OutputIfIndex >> 30
+	t.OutputIfIndex = t.OutputIfIndex & 0x3FFFFFFF
+
+	switch t.SourceIDIndex {
+	case t.OutputIfIndex:
+		t.SampleDirection = "egress"
```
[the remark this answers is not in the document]
ssoroka: Hmm. This field relationship might be a little obscure to expect another package to do it.

plugins/inputs/sflow/packetdecoder/packetdecoder.go
```diff
@@ -0,0 +1,506 @@
...
+	}
+	if err := read(r, &t.InputIfIndex, "InputIfIndex"); err != nil {
+		return t, err
+	}
+	if err := read(r, &t.OutputIfFormat, "OutputIfFormat"); err != nil { // line: 1728
+		return t, err
+	}
+	if err := read(r, &t.OutputIfIndex, "OutputIfIndex"); err != nil {
+		return t, err
+	}
```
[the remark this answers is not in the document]
ssoroka: Ok, will consider that. A binary unmarshaller here would be helpful, you could customize how things are decoded, but we don't have one at the moment. 

Will think on this, I would like to remove some of the unnecessarily explicit calls here.

plugins/inputs/sflow/packetdecoder/packetdecoder.go
```diff
@@ -0,0 +1,506 @@
...
+		return h, err
+	}
+
+	if h.HeaderLength > MaxPacketHeaderLen {
+		return h, fmt.Errorf("packet header length too large")
+	}
+
+	b := make([]byte, h.HeaderLength)
+	read(r, &b, "header data")
+	buf := bytes.NewBuffer(b)
```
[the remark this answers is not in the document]
ssoroka: I actually need the data to be explicitly read even if it's not used. It'd be more like a MinReader. I'd be happy to add that in.

plugins/inputs/sflow/packetdecoder/packetdecoder.go
ssoroka: Pretty sure we're missing more than a couple things and we just don't care about it. The MinReader behavior covers this.

plugins/inputs/sflow/types/decoder.go
ssoroka: Not sure if it might be useful to someone.

plugins/inputs/sflow/decoder_test.go
ssoroka: yeah. The relevant documentation is: 
> Data offset (4 bits)
> Specifies the size of the TCP header in 32-bit words. The minimum size header is 5 words and the maximum is 15 words thus giving the minimum size of 20 bytes and maximum of 60 bytes, allowing for up to 40 bytes of options in the header. This field gets its name from the fact that it is also the offset from the start of the TCP segment to the actual data.

from https://en.wikipedia.org/wiki/Transmission_Control_Protocol

It's only 4 bits. thus the max is 1111(binary) = 16(dec). 15 (the max) * 4-byte-words = 60 bytes max value for this field. given that the test was expecting 0x40, which is 64(dec), there's no way the test was correct. The original code misread the 4 byte mask 1111 as 11110000, making the value 0101 (0x05) appear as if it was 01010000 (0x50). After multiplying by 4, the correct answer is 00010100 (0x14), but they got 0x0140, which overflowed the one byte value, dropping the first bit and resulting in 0x40 (again, larger than the max value allowed for this field). So it was definitely an implementation error.

plugins/inputs/sflow/metricencoder.go
ssoroka: You'd have to loop samples first and count it up, then do the allocation. I'm not entirely sure it's worth it.

plugins/inputs/sflow/packetdecoder/packetdecoder.go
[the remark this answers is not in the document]
ssoroka: After some thought, I think I'd like to leave this up to a new, smarter, binary-aware packet decoder.

plugins/inputs/sflow/types/packets.go
```diff
@@ -0,0 +1,285 @@
+package types
```
reimda, line 1: Is there a reason "types" and "packetdecoder" are packages?  Now that we've switched to a straightforward coding style, there's not enough complexity to need separate dirs and packages.

It seems like packetdecoder/* should be moved to the main dir and types/packets.go should be moved to types.go in the main dir.  The name types.go is clearer to me than packets.go.

plugins/inputs/sflow/packetdecoder/packetdecoder.go
```diff
@@ -0,0 +1,484 @@
+package packetdecoder
+
+import (
+	"encoding/binary"
+	"fmt"
+	"io"
+
+	"github.com/influxdata/binaryio"
```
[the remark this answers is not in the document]
ssoroka: Fair enough. The plan was to add a binary io reader to that package and move this code over to it. Not sure if that changes your stance at all.

plugins/inputs/sflow/packetdecoder/packetdecoder.go
```diff
@@ -0,0 +1,484 @@
...
+	}
+	if err := read(r, &h.Checksum, "Checksum"); err != nil {
+		return h, err
+	}
+	return h, err
+}
+
+func read(r io.Reader, data interface{}, name string) error {
+	err := binary.Read(r, binary.BigEndian, data)
+	return errors.Wrapf(err, "failed to read packet %s", name)
```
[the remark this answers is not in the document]
ssoroka: k. I agree there's room for improvement here.

plugins/inputs/sflow/types/packets.go
ssoroka: valid point. It's mostly for flexibility in how the types are used with other decoders/encoders, but I don't really think we have that scenario here, so tighter coupling would be fine.

plugins/inputs/sflow/sflow_version_5.txt
```diff
@@ -0,0 +1,2803 @@
...
+                                                               July 2004
+
+
+                            sFlow Version 5
+
+
+
+Copyright Notice
+
+Copyright (C) sFlow.org (2004). All Rights Reserved.
```
danielnelson, line 23: It doesn't look like we are allowed to copy this, let's link to this in the documentation but not include it.

Issue #3709, opened 2018-01-22
Create a math processor plugin


rakshith-b91:
## Bug report
As of now conversion  supports only float, can we do the same for divide and multiply the data before output.

### Steps to reproduce:

1. Run telegraf conf with snmp input plugin, the output of timestamp is given in timeticks, we need to divide it by 6000 to get in minutes, we are not able do any conversion other than float in snmp plugin.

danielnelson:
I think the solution to this would be a processor that can do mathematical operations, this is lightly discussed in https://github.com/influxdata/telegraf/issues/2667#issuecomment-301538731.  In the meantime you should be able to do this conversion on query.

cruscio:
Additional use case - Calculating % of total CPU utilization of a Windows process.

Windows Perfmon provides process-level CPU utilization in `Process\% Processor Time`: The sum of processor time on each processor. This needs to be multiplied by the number of logical CPU cores to get a % of total CPU used by a process. `([Process(x)\% Processor Time] * ([Logical CPU Cores] * 100)) * 100`. It'd be ideal to perform this calculation locally at time of collection.

sslupsky:
Another use case:  Merging two fields/tags into one.  For instance, the high byte and low byte of a 16 bit integer.  ie:  result = high_byte * 256 + low_byte.

simoelmou:
This type of mathematical processing would be very useful in the case of the use of some types of time series databases such as OpenTSDB which doesn't support a lot of aggregation expressions, and it'll help to calculate SLAs an option.

TimurDela:
Unit conversions can be done in queries because they always use the exact same operation but mathematical operations that may change in time cannot be done after the data has been logged.
For instance, I have a device that provides analogue signals (always between 4 and 20mA).
Initially, I set it so that 4mA means 0kW and 20mA means 1000kW.
After a while, I realise that my power never drops below 10kW nor exceeds 20kW so in order to increase the precision of the measurement, I set my device so that 4mA means 10kW and 20mA means 20kW.
Logging the analogue signal in mA is meaningless, I want Telegraf to directly log the quantity I am interested, without having to change all my queries.

Issue #7194, opened 2020-03-18
Add a general purpose processor based on the Starlark language


danielnelson:
## Feature Request

The [Starlark][] language is a dialect of python designed to be embedded into a larger application.  The Go implementation that allows embedding the language using pure Go code and with few dependencies.  It is used primarily in the Bazel build system and is well maintained.

[Starlark]: https://github.com/google/starlark-go/blob/master/doc/spec.md

### Proposal:

Create a processor plugin that allows applying a starlark function to Telegraf metrics.  Expose and document an interface for reading and modifying the metric name, tags, fields, time.

### Desired behavior:

Example configuration, this script does not modify the metric but prints the values.  Calling print in the processor will be connected to the Telegraf log:
```toml
[[processors.starlark]]
  source = """
def apply(metric):
    print(metric.name)
    for tag in metric.tags:
        key, value = tag
        print('tag: %s=%s' % (key, value))
    for field in metric.fields:
        key, value = field
        print('field: %s=%s' % (key, value))
    print('time:', metric.time)
    return metric
"""
```

### Use case: 

This would be a general purpose plugin which will be helpful with specialized processing and extending Telegraf without requiring new plugins.  I'll be adding some additional use cases as they come up and are collected from the backlog:

- Performing simple mathematical operations such as the division of two fields.

danielnelson:
Would also implement #3709

PR #7254, opened 2020-03-30
[processor] Compute field value(s) by a given formula

srebhan:
### Required for all PRs:

This PR adds a processor plugin that can compute the value of a given field using the given formula. To achieve this, the formula is parsed into an abstract-syntax-tree and is then interpreted by walking the tree in a depth-first way. When reaching a leaf-node the value is propagated upwards until the value for the root node is known.

Currently integer and float computation like simple arithmetic, few functions and usage of field values as variables are supported. However, these functions can be extended easily in code and there is a way to provide tag-variables and more sophisticated default values if there is demand.

This PR might also be better named "math"... You decide. Anyway it probably solves the problem described in #3709.

### DIFF
--- a/README.md
+++ b/README.md
@@ -360,6 +360,7 @@ For documentation on the latest development code see the [documentation index][d
 
 * [clone](/plugins/processors/clone)
 * [converter](/plugins/processors/converter)
+* [compute](/plugins/processors/compute)
 * [date](/plugins/processors/date)
 * [dedup](/plugins/processors/dedup)
 * [enum](/plugins/processors/enum)
--- a/plugins/processors/all/all.go
+++ b/plugins/processors/all/all.go
@@ -2,6 +2,7 @@ package all
 
 import (
 	_ "github.com/influxdata/telegraf/plugins/processors/clone"
+	_ "github.com/influxdata/telegraf/plugins/processors/compute"
 	_ "github.com/influxdata/telegraf/plugins/processors/converter"
 	_ "github.com/influxdata/telegraf/plugins/processors/date"
 	_ "github.com/influxdata/telegraf/plugins/processors/dedup"
--- a/plugins/processors/compute/README.md
+++ b/plugins/processors/compute/README.md
@@ -1,0 +1,105 @@
+# Compute Processor
+
+The compute processor is used to calculate field values of metrics by
+specifying formulas for those target fields. These formulas can contain
+constants, arithmetic operation as well as variables referencing to existing
+field values.
+
+When referencing fields in the formulas that do not exist, a strategy can be
+provided to ignore or fill the missing values.
+
+### Configuration
+```toml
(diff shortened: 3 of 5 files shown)


danielnelson:
Thanks for sharing this plugin.  I'm hoping to have an initial version of a Starlark based processor soon #7194 that would fill a similar role, would love to get your take on that approach.

srebhan:
@danielnelson, while the Starlark-processor will for sure make this plugin superfluous, I'm in general not a big fan of putting"_program code_" into config files.

I'll leave security issues aside as, from a short skimming over Starlark, it seems they are putting a lot of thoughts into this. However, with a general purpose (interpreter) language you run into the risk that functionality will slowly  trickle into the config file. Think about some custom protocol that you can fetch via http, why should I write an input plugin if I can solve my immediate problem with some tenth lines of Starlark? Well the next guy will have the same problem, so we share things (in the best case) via forum or whatever. In the worst case everyone is coding her own small Starlark script to deal with their custom stuff...

So in my opinion, while it seems sweet in the short run, you **might** end up in script-snipplet-sharing hell after a while. Again, this is just my personal opinion and I'm perfectly fine with replacing this processor with the Starlark one, just make sure we don't need to discuss on how to share protocol parsing scripts soon... ;-)

danielnelson:
Thanks for your analysis, I have some of the same concerns too.  We actually have this issue of sharing configuration snippets today, but the more powerful we make processors the more it will become a need.

We also have this issue in other areas of Telegraf.  For example we have the exec input and in 1.14 we added the execd input too, these run arbitrary programs.  We also have the postgresql_extensible plugin today and soon I hope to add a set of plugins that will allow ad-hoc queries against mysql and sqlserver.

These are all nice because you can customize inputs without writing code that needs to be included in Telegraf, but with more custom setup comes the need to share.  We also know that it would be helpful to share other items such as dashboards and alerting rules.  This is something we have been thinking about and there are plans to work towards having some facilities for doing this. 
An early effort has been started over on the [community-templates](https://github.com/influxdata/community-templates).

Starlark is pretty heavily sandboxed, I don't expect it will be possible to write input plugins since you can't open sockets or files out of the box.  I think providing wrapped functions to do this would be too much work even if we wanted to do it.  I'm hoping to use it only in a limited role.

srebhan:
@danielnelson can you please outline your strategy on this and Starlark's PR, so I can adjust my plan accordingly. Are you planning to merge Starlark and close this PR or are you willing to merge both? I'm asking because I'm currently planning a first draft PoC for our system and want to use a future-proof foundation. ;-)

danielnelson:
I think we should close this PR.  I might be convinced of something very similar but using starlark to evaluate expressions.  I'm a bit on the fence with this, as I'm not sure it is different enough to be worth it.

PR #7256, opened 2020-03-30
Improve documentation for the Metric interface

### LINKED ISSUE #7251: Question: are metric tags, as consumed by output plugins, guaranteed to be sorted?
It looks like the code in the metric/ package make an effort to keep tags sorted. I'm wondering if I can depend on this invariant in my output plugin - that if I iterate over TagList the keys will be in sort-order.

danielnelson:
closes #7251

### Required for all PRs:

### DIFF
--- a/metric.go
+++ b/metric.go
@@ -17,43 +17,93 @@ const (
 	Histogram
 )
 
+// Tag represents a single tag key and value.
 type Tag struct {
 	Key   string
 	Value string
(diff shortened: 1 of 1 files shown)


danielnelson:
The real holdup on Tags() and Fields() is removing it's use within Telegraf.  I like the composed interface idea though, may be worth doing.

PR #7267, opened 2020-04-02
Add selector tags for service (kube_inventory)

jimmyseto:
### Required for all PRs:


jimmyseto:
This is a small change to the kube_inventory plugin to extract selector information for a given service.

jimmyseto:
@danielnelson - this is a relatively small change.  any chance we can get this low-hanging fruit reviewed and merged?

danielnelson:
I'm concerned about this change producing a lot of new tags depending on how selectors are being used.  Especially if certain bits of tooling are creating selectors automatically (perhaps Helm or Istio).

I think we should add a selector whitelist so that only the wanted selectors are added, we could use a style similar to what is done in the docker input plugin:
```
selectors_include = []
selectors_exclude = ["*"]
```

Another question is if this should be part of only the service type, or if we need similar behavior for all types?

jimmyseto:
thanks for the feedback, @danielnelson 

the potential influx of tags is a fair concern (similar to what we have with labels elsewhere).  i think users can always configure their output plugin(s) to filter out undesired tags/selectors/labels, if needed.  but, the ability to control this earlier at the input plugin level has its benefits.  i've gone ahead, and implemented the filtering capability as you suggested.

i think selector information for the other types supported by kube_inventory is more of a nice-to-have.  with the consolidated bits of information we can gather via the kubernetes plugin and prometheus plugin (kube-state-metrics), we can connect the dots between the various kubernetes object types today. i don't think we currently have this with kubernetes services (i.e. the ability to connect services to/from pods).  either way, i've gone ahead, and added this for the other applicable object types as well.

jimmyseto:
hi @danielnelson - what do you think of the latest update?  do they address your concerns?  thanks.

jimmyseto:
@danielnelson - just wanted to make sure this was still on your radar.  thanks in advance.

plugins/inputs/kube_inventory/kube_state.go
```diff
@@ -114,6 +125,15 @@ func (ki *KubernetesInventory) Gather(acc telegraf.Accumulator) (err error) {
 		return err
 	}
 
+	//err = ki.createSelectorFilters()
+	//if err != nil {
+	//	return err
+	//}
```
danielnelson, line 131: Can you remove this

jimmyseto: done, and done.

plugins/inputs/kube_inventory/kube_state.go
```diff
@@ -65,6 +70,12 @@ var sampleConfig = `
   ## Overrides resource_exclude if both set.
   # resource_include = [ "deployments", "nodes", "statefulsets" ]
 
+  ## selectors to include and exclude as tags.  Globs accepted.
+  ## Note that an empty array for both will include all selectors as tags
+  ## selector_exclude overrides selector_include if both set.
+  selector_include = []
+  selector_exclude = ["*"]
```
danielnelson, line 77: Sorry, missed this. Let's initialize these options down in `func init() {` so that upgrades doesn't accidentally get set to include all.  We can then list these commented out here in the sample configuration.

jimmyseto: done

danielnelson:
Thanks!

jimmyseto:
> Thanks!

thanks, @danielnelson.  is there a roadmap/timeline for 1.15.0?

danielnelson:
Hoping to do RC1 tomorrow, then a release in about a week depending on what issues come up.

PR #7271, opened 2020-04-02
[inputs/modbus] Implement retry when slave is busy

srebhan:
### Required for all PRs:

Currently, when the slave device is busy, e.g. because it already serves a request of another client, we get an error and the measurement is lost for the gathering interval. This potentially leads to many lost measurements for very busy devices.

With this PR you can specify the number of retry attempts (and a wait time) to be performed if the device is busy. This achives a higher probability to get a measurement on busy devices for each query interval.

### DIFF
--- a/plugins/inputs/modbus/README.md
+++ b/plugins/inputs/modbus/README.md
@@ -20,6 +20,11 @@ The Modbus plugin collects Discrete Inputs, Coils, Input Registers and Holding R
  ## Timeout for each request
  timeout = "1s"
 
+ ## Maximum number of retries and the time to wait between retries
+ ## when a slave-device is busy
+ #busy_retries = 0
+ #busy_retries_wait = "100ms"
+
  # TCP - connect via Modbus/TCP
  controller = "tcp://localhost:502"
 
@@ -53,7 +58,7 @@ The Modbus plugin collects Discrete Inputs, Coils, Input Registers and Holding R
 
  ## Analog Variables, Input Registers and Holding Registers
  ## measurement - the (optional) measurement name, defaults to "modbus"
- ## name       - the variable name 
+ ## name       - the variable name
  ## byte_order - the ordering of bytes
  ##  |---AB, ABCD   - Big Endian
  ##  |---BA, DCBA   - Little Endian
--- a/plugins/inputs/modbus/README.md
+++ b/plugins/inputs/modbus/README.md
@@ -21,7 +21,10 @@ The Modbus plugin collects Discrete Inputs, Coils, Input Registers and Holding R
  timeout = "1s"
 
  ## Maximum number of retries and the time to wait between retries
- ## when a slave-device is busy
+ ## when a slave-device is busy.
+ ## NOTE: Please make sure that the overall retry time (#retries * wait time)
+ ##       is always smaller than the query interval as otherwise you will get
+ ##       an "did not complete within its interval" warning.
  #busy_retries = 0
  #busy_retries_wait = "100ms"
 
--- a/plugins/inputs/modbus/modbus.go
+++ b/plugins/inputs/modbus/modbus.go
@@ -174,10 +174,6 @@ func (m *Modbus) Init() error {
 		return fmt.Errorf("retries cannot be negative")
 	}
 
-	if m.Retries > 0 && m.RetriesWaitTime.Duration < 0 {
-		return fmt.Errorf("wait time between retries cannot be negative")
-	}
-
 	err := m.InitRegister(m.DiscreteInputs, cDiscreteInputs)
 	if err != nil {
 		return err
--- a/plugins/inputs/modbus/modbus.go
+++ b/plugins/inputs/modbus/modbus.go
@@ -661,19 +661,18 @@ func (m *Modbus) Gather(acc telegraf.Accumulator) error {
 	for retry := 0; retry <= m.Retries; retry += 1 {
 		timestamp = time.Now()
 		err := m.getFields()
-		if err == nil {
-			// Reading was successful, leave the retry loop
-			break
-		} else {
+		if err != nil {
 			mberr, ok := err.(*mb.ModbusError)
 			if ok && mberr.ExceptionCode == mb.ExceptionCodeServerDeviceBusy && retry < m.Retries {
 				log.Printf("I! [inputs.modbus] device busy! Retrying %d more time(s)...", m.Retries-retry)
-			} else {
-				disconnect(m)
-				m.isConnected = false
-				return err
+				continue
 			}
+			disconnect(m)
+			m.isConnected = false
+			return err
 		}
+		// Reading was successful, leave the retry loop
+		break
 		time.Sleep(m.RetriesWaitTime.Duration)
 	}
 
--- a/plugins/inputs/modbus/modbus.go
+++ b/plugins/inputs/modbus/modbus.go
@@ -674,7 +674,6 @@ func (m *Modbus) Gather(acc telegraf.Accumulator) error {
 				return err
 			}
 		}
-		log.Printf("D! [inputs.modbus] sleeping %d millisecond(s)...", m.RetriesWaitTime.Duration.Milliseconds())
 		time.Sleep(m.RetriesWaitTime.Duration)
 	}
 
--- a/plugins/inputs/modbus/modbus.go
+++ b/plugins/inputs/modbus/modbus.go
@@ -88,7 +88,10 @@ const sampleConfig = `
  timeout = "1s"
 
  ## Maximum number of retries and the time to wait between retries
- ## when a slave-device is busy
+ ## when a slave-device is busy.
+ ## NOTE: Please make sure that the overall retry time (#retries * wait time)
+ ##       is always smaller than the query interval as otherwise you will get
+ ##       an "did not complete within its interval" warning.
  #busy_retries = 0
  #busy_retries_wait = "100ms"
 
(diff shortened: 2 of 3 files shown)


plugins/inputs/modbus/modbus.go
```diff
@@ -159,6 +167,14 @@ func (m *Modbus) Init() error {
 		return fmt.Errorf("device name is empty")
 	}
 
+	if m.Retries < 0 {
+		return fmt.Errorf("retries cannot be negative")
+	}
+
+	if m.Retries > 0 && m.RetriesWaitTime.Duration < 0 {
+		return fmt.Errorf("wait time between retries cannot be negative")
+	}
```
ssoroka, line 176: we might want to check that RetriesWaitTime * Retries is not greater than the time the input has to collect the metrics, 1s, I think.

plugins/inputs/modbus/modbus.go
```diff
@@ -642,11 +658,23 @@ func (m *Modbus) Gather(acc telegraf.Accumulator) error {
...
+		} else {
+			mberr, ok := err.(*mb.ModbusError)
+			if ok && mberr.ExceptionCode == mb.ExceptionCodeServerDeviceBusy && retry < m.Retries {
+				log.Printf("I! [inputs.modbus] device busy! Retrying %d more time(s)...", m.Retries-retry)
+			} else {
+				disconnect(m)
+				m.isConnected = false
+				return err
+			}
+		}
```
ssoroka, line 676: I found this a bit confusing. I'd clean it up like so:
```suggestion
		if err != nil {
			mberr, ok := err.(*mb.ModbusError)
			if ok && mberr.ExceptionCode == mb.ExceptionCodeServerDeviceBusy && retry < m.Retries {
				log.Printf("I! [inputs.modbus] device busy! Retrying %d more time(s)...", m.Retries-retry)
				continue
			}
			disconnect(m)
			m.isConnected = false
			return err
		}
		// Reading was successful, leave the retry loop
		break
```

plugins/inputs/modbus/modbus.go
srebhan: Hmm this might be challenging if the time-frame is really 1s. There might be cases with a slow but busy device where come close to this wait time... However it is worth a trial (at least for all other cases). Is the 1 second a fixed value or can I get it from somewhere in the code?

ssoroka: Ok I looked into this a bit more, it's not actually 1s per se; it logs "did not complete within its interval" if the runtime is longer than the agent interval setting. 
We could warn if the `RetriesWaitTime * Retries > interval`, where interval is the input's configured interval.

ssoroka: Agh. we might not be exposing that setting to the plugin. May not be possible.. Ok maybe just a note in the settings not to pick a value that would end up retrying longer than your interval.

plugins/inputs/modbus/modbus.go
```diff
@@ -642,11 +658,22 @@ func (m *Modbus) Gather(acc telegraf.Accumulator) error {
...
+				log.Printf("I! [inputs.modbus] device busy! Retrying %d more time(s)...", m.Retries-retry)
+				continue
+			}
+			disconnect(m)
+			m.isConnected = false
+			return err
+		}
+		// Reading was successful, leave the retry loop
+		break
+		time.Sleep(m.RetriesWaitTime.Duration)
```
ssoroka, line 676: This sleep is now unreachable (and it's breaking the build), I think you want to move it to line 667 after the log.Printf and before the continue?

srebhan: Done with the commit(s) below.

plugins/inputs/modbus/modbus.go
srebhan: See commit(s) below.

plugins/inputs/modbus/modbus.go
```diff
@@ -159,6 +170,14 @@ func (m *Modbus) Init() error {
 		return fmt.Errorf("device name is empty")
 	}
 
+	if m.Retries < 0 {
+		return fmt.Errorf("retries cannot be negative")
+	}
+
+	if m.Retries > 0 && m.RetriesWaitTime.Duration < 0 {
```
reimda, line 177: Duration needs to be positive.  0 wait time makes no sense

ssoroka: I would agree this check is not adding any value. While it may technically be possible to set a negative duration, Go will not blow up and will interpret it as a zero wait. This check is too defensive and should probably be removed.

srebhan: @reimda:
> Duration needs to be positive. 0 wait time makes no sense
Well zero-wait time actually makes sense as it means to directly try again. Some devices take quite some time to respond and you can basically try to poll them in a busy-waiting fashion.

@ssoroka:
Well I just thought that a negative value is with 99.99999% probability a configuration error and instead of silently using another value (as Go will map it basically to 0) might hide some issues. Imagine you wanted to give e.g. 100ms but due to some cat running over your keyboard (ok this is made up) you get -100ms. You will wonder forever why this device will fail even though you thought you allowed 5 retries with 100ms wait time.

But anyway, I will follow the final decision as I interpret the above as @reimda wants a more strict check while @ssoroka  wants the check removed. So which way I should follow?

srebhan: Removed check as requested in commit below.

reimda: @srebhan 
I initially thought the zero wait time was a simple off by one error.  Thanks for explaining how quick retries can be useful for devices that take a long time to reply that they're busy.

I'm new to modbus so maybe you can help me understand what's going on here.  If a device replies quickly that it's busy, won't this code immediately retry?  Without any delay it will be unlikely that the retries get anything other than more quick busy responses.

It seems to me that we need to handle both cases - retry quickly on slow busy responses and retry not too quickly on quick busy responses.

Maybe change this:
`time.Sleep(m.RetriesWaitTime.Duration)`
to this:
`elapsed = now() - timestamp
time.Sleep(m.RetriesWaitTime.Duration - elapsed)`

Then also have a small lower limit on duration, maybe 20ms

What do you think?

danielnelson:
We have similar logic in the SNMP plugin, and it has been somewhat problematic.  The big issues we have there are:
- Total request time can be very hard to predict, since its not clear how many requests will be made, and so it is hard to set the correct retry settings.
- May cause intervals to be skipped which can be worse than skipping especially if polling multiple devices (less of a concern due to this plugin connecting to a single controller).
- Can make it so that Telegraf becomes unresponsive to shutdown signals.  As Telegraf will wait for the plugin to complete Gather.
- Results often get spread out over the interval more than expected, not necessarily an issue as the timestamp is set correctly, but usually not desired and can make aggregation difficult.
- Causes increased load on the device.

Some of these issues could be addressed by having a total Gather timeout.

We should consider if it would be enough to ignore `ExceptionCodeServerDeviceBusy` errors, continuing to request the remaining data instead of the current stop behavior, and leaving "retry" to be handled by Telegraf at the next interval.

plugins/inputs/modbus/modbus.go
srebhan: I think you are over engineering here. Normally you know how long your device will likely be busy (which is usually only the case if some other process also querying the device) and how quickly it responds. Take for example my use-case, I have another process (that I cannot influence) querying the device and it takes that process about 500-700ms to finish. So I will probably set the retries to 5 and the wait time to 200ms to achieve a trade-off between the delay and the "guarantee" that one of the trials will succeed. Anyway in some other use-case the delay/latency of getting the information might be crucial, so in this case you probably do not set a wait interval but set the number of retries high.
This being said, it more depends on the use-case, the individual setup and last but not least on the individual preferences to choose between long-wait, less retries vs. busy-waiting with many retries... So I think the extra effort is not worth it. :-)

srebhan:
@danielnelson while I do understand your concerns things here are a bit more complicated. First of all the "stop behavior" only applies for one device and one plugin-process, so influences on other devices are not to be expected. Furthermore, we use the information gathered to do some control-decisions and in this case a missing datum is one of the worst things that can happen. So while a jitter of 1 or 2 seconds is perfectly fine, having a whole missing interval is bad and we cannot reduce the interval too much as otherwise the concurring process will not get access to the device any longer.

So while I agree that the implementation is sub-optimal I think there are two possible solutions:
 
1. The plugin gets to know the interval and can take care to finish before the interval passed. While this is also not optimal as you need to trust the plugin to do the right thing, you at least allow the plugin to do the right thing. ;-)
2. The perfect solution is that you allow Gather() to return the wish to retry and handle the things in the agent. This way you can fix the shutdown as well as the missed interval problems...

plugins/inputs/modbus/README.md
```diff
@@ -20,6 +20,14 @@ The Modbus plugin collects Discrete Inputs, Coils, Input Registers and Holding R
  ## Timeout for each request
  timeout = "1s"
 
+ ## Maximum number of retries and the time to wait between retries
+ ## when a slave-device is busy.
+ ## NOTE: Please make sure that the overall retry time (#retries * wait time)
+ ##       is always smaller than the query interval as otherwise you will get
+ ##       an "did not complete within its interval" warning.
```
danielnelson, line 27: Let's remove this note, I think this isn't the right advice.  It may actually be desired to configure the plugin to exceed the interval.  This way you are more likely to get a response, and not get into a situation where the device never fully reply in time.

srebhan:
> I do suggest then that you don't take control decisions if all data is not present, just to be extra safe.
That's exactly what we do and that's why we need the retries. If another process (that might not be under our control) is reading by chance at the same time, we will never get any data without retrying.

plugins/inputs/modbus/README.md
srebhan: Hmm, this note was requested earlier by @ssoroka:
>Agh. we might not be exposing that setting to the plugin. May not be possible.. Ok maybe just a note in the settings not to pick a value that would end up retrying longer than your interval.

Could you guy please resolve this conflict and tell me if I should leave that comment in or not... ;-)

danielnelson: We can resolve this post merge.

PR #7278, opened 2020-04-02
kapacitor_alert and kapacitor_cluster measurements

kelseiv:
- Update ReadMe to include `kapacitor_alert` and `kapacitor_cluster` measurements. 

@danielnelson, I've incorporated all edits from @nathanielc.

Thanks guys!

### DIFF
--- a/plugins/inputs/kapacitor/README.md
+++ b/plugins/inputs/kapacitor/README.md
@@ -33,7 +33,7 @@ The Kapacitor plugin collects metrics from the given Kapacitor instances.
 	- [notification_dropped](#notification_dropped) _(integer)_
 	- [primary-handle-count](#primary-handle-count) _(integer)_
 	- [secondary-handle-count](#secondary-handle-count) _(integer)_
-- [kapacitor_cluster](#kapacitor_cluster) _(integer)_ (Kapacitor Enterprise only)
+- (Kapacitor Enterprise only) [kapacitor_cluster](#kapacitor_cluster) _(integer)_ 
 	- [dropped_member_events](#dropped_member_events) _(integer)_
 	- [dropped_user_events](#dropped_user_events) _(integer)_
 	- [query_handler_errors](#query_handler_errors) _(integer)_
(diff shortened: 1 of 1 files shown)


plugins/inputs/kapacitor/README.md
```diff
@@ -23,12 +23,20 @@ The Kapacitor plugin will collect metrics from the given Kapacitor instances.
...
     - [num_subscriptions](#num_subscriptions) _(integer)_
     - [num_tasks](#num_tasks) _(integer)_
+- [kapacitor_alert](#kapacitor_alert)
+	- [notification_dropped](#notification_dropped) _(integer)_
+	- [primary-handle-count](#primary-handle-count) _(integer)_
+	- [secondary-handle-count](#secondary-handle-count) _(integer)_
+- [kapacitor_cluster](#kapacitor_cluster) _(integer)_ (Kapacitor Enterprise only)
+	- [dropped_member_events](#dropped_member_events) _(integer)_
+	- [dropped_user_events](#dropped_user_events) _(integer)_
+	- [query_handler_errors](#query_handler_errors) _(integer)_
```
nathanielc, line 39: All of these metrics only appear in the Enterprise edition of Kapacitor so we should call that out somewhere.

plugins/inputs/kapacitor/README.md
```diff
@@ -96,6 +104,36 @@ The total number of Kapacitor tasks.
 
 ---
 
+### kapacitor_alert
+The `kapacitor_alert` measurement stores fields with information related to
+[Kapacitor alerts](https://docs.influxdata.com/kapacitor/v1.5/working/alerts/).
+
+#### notification-dropped
+The number of alert or event handlers that failed.
```
nathanielc, line 112: This is a bit misleading  as the `notification` in this context is not the alert handler but an internal notification used to deduplicate alerts. 

I would recommend something along these lines:

```
The number of internal notifications that were dropped because they arrived too late from another Kapacitor node. If this count is increasing it indicates that Kapacitor Enterprise nodes are not able to communicate fast enough compared to the volume of alerts.
```

plugins/inputs/kapacitor/README.md
```diff
@@ -96,6 +104,36 @@ The total number of Kapacitor tasks.
...
+
+#### primary-handle-count
+The number of alerts counted for the primary event handler.
```
nathanielc, line 115: Suggest:

```
The number of times this node handled an alert as the primary. This count should increase under normal conditions.
```

plugins/inputs/kapacitor/README.md
```diff
@@ -96,6 +104,36 @@ The total number of Kapacitor tasks.
...
+
+#### secondary-handle-count
+The number of alerts counted for the secondary event handler.
```
nathanielc, line 118: Suggest:

```
The number of times this node handled an alert as the secondary. An increase in this counter indicates that the primary is failing to handle alerts in a timely manner.
```

plugins/inputs/kapacitor/README.md
```diff
@@ -96,6 +104,36 @@ The total number of Kapacitor tasks.
...
+
+---
+
+### kapacitor_cluster
+The `kapacitor_cluster` measurement stores fields with information related to
+[Kapacitor clusters](https://docs.influxdata.com/enterprise_kapacitor/v1.5/cluster-management/). Kapacitor Enterprise deduplicates alert data to prevent duplicate alert notifications from being sent.
```
nathanielc, line 124: The cluster metrics are not specific to alert deduplication rather the reflect the ability of Kapacitor nodes to communicate with one another. Specifically these metrics track the gossip communication between the Kapacitor nodes.  https://docs.influxdata.com/enterprise_kapacitor/v1.5/administration/configuration/#cluster-communications

plugins/inputs/kapacitor/README.md
```diff
@@ -96,6 +104,36 @@ The total number of Kapacitor tasks.
...
+
+#### dropped_member_events
+The number of member events dropped for alert deduplication.
```
nathanielc, line 127: Suggest:

```
The number of gossip member events that were dropped.
```

plugins/inputs/kapacitor/README.md
```diff
@@ -96,6 +104,36 @@ The total number of Kapacitor tasks.
...
+
+#### dropped_user_events
+The number of user events drooped for alert deduplication.
```
nathanielc, line 130:
```
The number of gossip user events that were dropped.
````

File: plugins/inputs/kapacitor/README.md [hunk 8]
```diff
@@ -96,6 +104,36 @@ The total number of Kapacitor tasks.
...
+
+#### query_handler_errors
+The number of query handler errors for alert deduplication.
```
nathanielc, line 133: This one is no longer used and does not need to be added.

PR #7283, opened 2020-04-03
Support Go execd plugins with shim

ssoroka:
Plugins written for Telegraf can easily be converted to external execd based plugins with this shim.


plugins/inputs/execd/shim/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Steps to build and run your plugin
+
+1. Build the cmd/main.go. For my rand project this looks like `go build -o rand cmd/main.go`
+1. Test out the binary if you haven't done this yet. eg `./rand -config plugin.conf`. Depending on your polling settings and whether you implemented a service plugin or an input gathering plugin, you may see data right away, or you may have to hit enter first, or wait for your poll duration to elapse, but the metrics will be written to STDOUT. Ctrl-C to end your test.
+1. Configure Telegraf to call your new plugin binary. eg: 
+```
+[[inputs.execd]]
+  command = ["/path/to/rand", "-config", "/path/to/plugin.conf"]
+  signal = "none"
```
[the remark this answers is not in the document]
ssoroka: It'll be out of sync regardless with the other execd plugins, which are all paying attention to stdin/stdout.
Do we have precedent for using signals to trigger metric collection elsewhere?

plugins/inputs/execd/shim/example/cmd/main.go
```diff
@@ -0,0 +1,35 @@
...
+
+func main() {
+	flag.Parse()
+	if *configFile == "" {
+		flag.Usage()
+		os.Exit(2)
+	}
+
+	cfg = config.NewConfig()
+	if err := cfg.LoadConfig(*configFile); err != nil {
```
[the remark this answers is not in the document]
ssoroka: Maybe a good alternative is just load normally if there's no config file specified

plugins/inputs/execd/shim/goshim.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+}
+
+func stdinCollectMetricsPrompt(ctx context.Context, collectMetricsPrompt chan<- os.Signal) {
+	s := bufio.NewScanner(os.Stdin)
+	for s.Scan() {
+		select { // don't block.
+		case <-ctx.Done():
+			return
+		case collectMetricsPrompt <- nil:
```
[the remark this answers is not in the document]
ssoroka: non-blocking write to collectMetricsPrompt.
This reads: "for every line read from stdin, make sure we're not supposed to quit, then push a message on to the collectMetricsPrompt"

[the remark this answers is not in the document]
ssoroka: but I just realized select order is non-deterministic, so sometimes it checks for quit first, and sometimes second. We might want to fix that.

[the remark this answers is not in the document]
ssoroka: and... it would only do one, not both.. so maybe this was busted. good question!

plugins/inputs/execd/shim/goshim.go
```diff
@@ -0,0 +1,137 @@
...
+		default:
+		}
+	}
+}
+
+func collectMetrics() {
+	for i := 0; i < len(gatherPromptChans); i++ {
+		// don't block
+		select {
+		case gatherPromptChans[i] <- empty{}:
```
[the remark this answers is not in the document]
ssoroka: I'm updating the comment to 
```
		// push a message out to each channel to collect metrics. don't block.
```

plugins/inputs/execd/shim/goshim.go
```diff
@@ -0,0 +1,137 @@
...
+	t := time.NewTicker(pollInterval)
+	defer t.Stop()
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-gatherPromptCh:
+			if err := runningInput.Gather(acc); err != nil {
+				handleErr(err)
+			}
```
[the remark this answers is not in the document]
ssoroka: will give priority to stopping. otherwise this is probably okay.

plugins/inputs/execd/shim/goshim.go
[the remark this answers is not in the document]
ssoroka: should be resolved now

plugins/inputs/execd/execd_test.go
```diff
@@ -0,0 +1,90 @@
...
+
+	"github.com/influxdata/telegraf/plugins/parsers"
+
+	"github.com/influxdata/telegraf"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestExternalInputWorks(t *testing.T) {
+	jsonParser, err := parsers.NewInfluxParser()
+	assert.NoError(t, err)
```
danielnelson, line 24: Recommend using `require.NoError(t, err)` form almost always.  Otherwise the test will continue and that usually just makes debugging the test more confusing.

plugins/inputs/execd/execd_unix.go
```diff
@@ -14,6 +14,11 @@ func (e *Execd) Gather(acc telegraf.Accumulator) error {
 	if e.cmd == nil || e.cmd.Process == nil {
```
danielnelson, line 14: Optional, should we rename this file `execd_posix.go`?

plugins/inputs/execd/execd_unix.go
```diff
@@ -14,6 +14,11 @@ func (e *Execd) Gather(acc telegraf.Accumulator) error {
...
 		return nil
 	}
+	select {
+	case <-e.ctx.Done():
+		return nil
+	default:
+	}
```
danielnelson, line 21: This can never happen as a type invariant, Gather won't be called after Stop() completes.

plugins/inputs/execd/execd_test.go
```diff
@@ -0,0 +1,90 @@
...
+	// test that a later gather will not panic
+	e.Gather(acc)
+}
+
+func fileShellScriptPath() string {
+	return filepath.Join(telegrafPath(), "plugins/inputs/execd/examples/count.sh")
+}
+
+func telegrafPath() string {
+	wd, err := os.Getwd()
```
danielnelson, line 69: `go test` always run from the test directory (plugins/inputs/execd), you can take advantage of this as there are many tests using this property now.  This was recommended by Dave Cheney: https://dave.cheney.net/2016/05/10/test-fixtures-in-go

plugins/inputs/execd/shim/README.md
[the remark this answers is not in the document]
danielnelson: Let's leave it as recommending none then, with stdin if you don't read it then I believe it would lock up the plugin.

plugins/inputs/execd/shim/goshim.go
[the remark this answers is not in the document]
danielnelson: I don't think this is sufficient, it still seems racy. Gather/Stop really need to be called from the same goroutine.

plugins/inputs/execd/shim/goshim.go
```diff
@@ -0,0 +1,178 @@
...
+		case <-t.C:
+			if err := runningInput.Gather(acc); err != nil {
+				handleErr(err)
```
danielnelson, line 169: I think this would be a lot nicer if we could return/print the error and then do an orderly shutdown.

agent/agent_test.go
```diff
@@ -22,35 +22,35 @@ func TestAgent_OmitHostname(t *testing.T) {
 func TestAgent_LoadPlugin(t *testing.T) {
 	c := config.NewConfig()
 	c.InputFilters = []string{"mysql"}
-	err := c.LoadConfig("../internal/config/testdata/telegraf-agent.toml")
+	err := c.LoadConfig("../config/testdata/telegraf-agent.toml")
```
danielnelson, line 25: Back on the topic of LoadConfig (since I can't reply to my previous comment anymore), what I'm really worried about is that this function has a lot of behavior.  I'm also sure we will want to modify it in the near future.  If we use this in the shim it may be difficult to guarantee the same handling of options.  It will also be quite permissive on what can be set, including things which won't work such as the interval, namepass/namedrop, global_tags, even the old single plugin per type Telegraf style, etc.

I think we would be better served with a simple toml parse to struct function.

plugins/inputs/execd/examples/count.sh
```diff
@@ -4,9 +4,9 @@
...
 
-while read; do
+while read LINE; do
     echo "counter_bash count=${counter}"
-    let counter=counter+1
+    counter=`expr $counter + 1`
 done
 
-(>&2 echo "terminate")
+(>&2 echo "terminate")
```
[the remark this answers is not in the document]
ssoroka: honestly don't remember why this is here. I think just so something goes to stderr

plugins/inputs/execd/examples/count.sh
```diff
@@ -4,9 +4,9 @@
 
 counter=0
 
-while read; do
+while read LINE; do
     echo "counter_bash count=${counter}"
-    let counter=counter+1
+    counter=`expr $counter + 1`
```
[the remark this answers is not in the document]
ssoroka: lol but now I need to retest it. :D

plugins/inputs/execd/execd.go
```diff
@@ -67,21 +70,28 @@ func (e *Execd) SetParser(parser parsers.Parser) {
...
 
 	if len(e.Command) == 0 {
 		return fmt.Errorf("E! [inputs.execd] FATAL no command specified")
 	}
 
 	e.wg.Add(1)
 
-	var ctx context.Context
-	ctx, e.cancel = context.WithCancel(context.Background())
+	e.ctx, e.cancel = context.WithCancel(context.Background())
```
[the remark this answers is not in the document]
ssoroka: how do you avoid storing it when you can't pass it to Gather without breaking the interface? Can I count on the agent not calling Gather after Stop has been called?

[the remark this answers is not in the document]
danielnelson: > Can I count on the agent not calling Gather after Stop has been called?

Yes, additionally they will not be called concurrently.

plugins/inputs/execd/execd.go
```diff
@@ -90,59 +97,62 @@ func (e *Execd) Stop() {
...
 
 		select {
 		case <-ctx.Done():
-			return
-		case <-time.After(e.RestartDelay.Duration):
+			return nil
+		case <-time.After(time.Duration(e.RestartDelay)):
 			// Continue the loop and restart the process
+			if err := e.cmdStart(); err != nil {
+				return err
```
danielnelson, line 132: If we return here the process will never be restarted, I think we need to log, wait another RestartDelay, and try to start it again.

plugins/inputs/execd/execd.go
```diff
@@ -90,59 +97,62 @@ func (e *Execd) Stop() {
...
 	if len(e.Command) > 1 {
 		e.cmd = exec.Command(e.Command[0], e.Command[1:]...)
 	} else {
 		e.cmd = exec.Command(e.Command[0])
 	}
 
-	stdin, err := e.cmd.StdinPipe()
+	e.stdin, err = e.cmd.StdinPipe()
 	if err != nil {
 		return fmt.Errorf("E! [inputs.execd] Error opening stdin pipe: %s", err)
```
danielnelson, line 147: Can you fix this preexisting issue with the error strings in this function, they should be like: "error opening stdin pipe: %v", without the logging prefix.

plugins/inputs/execd/execd.go
ssoroka: So then... what do we care if start returns the start error or not? Maybe I should go back to how I had this in the cmdLoop

danielnelson: Anytime the subprocess exits the execd input should restarts it, we want to retry forever just like we would if an output plugin was unable to connect.  Exception being the error from Exec on startup and that is to catch and report configure time errors.

ssoroka: if it fails on this line it's probably a configuration error and it'll never start. I guess the question is do you want it to keep spamming the logs with retries to what is likely a config error?

ssoroka:
build is passing; just stuck.

plugins/inputs/execd/shim/goshim.go
```diff
@@ -0,0 +1,281 @@
...
+func (s *Shim) Run(pollInterval time.Duration) error {
+	wg := sync.WaitGroup{}
+	quit := make(chan os.Signal, 1)
+	signal.Notify(quit, syscall.SIGINT, syscall.SIGTERM)
+
+	collectMetricsPrompt := make(chan os.Signal, 1)
+	listenForCollectMetricsSignals(collectMetricsPrompt)
+
+	wg.Add(1) // wait for the metric channel to close
+	metricCh := make(chan telegraf.Metric, 1)
```
danielnelson, line 81: Is this different from `Shim.metricCh`?

ssoroka: Ah, I started to move it but didn't end up needing it. Removing the struct copy.

plugins/inputs/execd/shim/goshim_notwindows.go
```diff
@@ -0,0 +1,14 @@
...
+
+import (
+	"os"
+	"os/signal"
+	"syscall"
+)
+
+func listenForCollectMetricsSignals(collectMetricsPrompt chan os.Signal) {
+	// just listen to all the signals.
+	signal.Notify(collectMetricsPrompt, syscall.SIGHUP, syscall.SIGUSR1, syscall.SIGUSR2)
```
reimda, line 13: It's a convention for daemon processes to reload config on HUP.  Even if the shim doesn't do that yet I think we ought to avoid using HUP as a collection signal in case we want to use it to reload config in the future.

plugins/inputs/execd/shim/goshim_notwindows.go
```diff
@@ -0,0 +1,14 @@
...
+}
```
reimda, line 14: I don't think I like that execd in general (not just the shim) can signal collection through unix signals, whether it's HUP or USR1 or USR2.  They're tricky to use in a simple program in most languages because they make read() and write() return EINTR.  If you're writing your data too slowly and you're in the middle of a write when there's another collection signal, you may lose data.  I'm not sure how go handles this, it might insulate the coder from EINTR but execd is meant to be used with any language.

If "newline on stdin" works to signal collection, and it works on posix and windows, and it's safer/easier to implement than unix async signals, why isn't "newline on stdin" the only way to signal collection?

plugins/inputs/execd/shim/goshim_notwindows.go
ssoroka: yeah I don't like it either. Unfortunately I inherited this from whoever did the first execd version where they decided they should let the user pick how to signal metric handling, providing SIGHUP as an option. So if we remove it we have to agree we're okay with breaking the last published version of execd input.

plugins/inputs/execd/shim/goshim_notwindows.go
ssoroka: This is news to me. Looks like Go handles it for you: [example](https://go-review.googlesource.com/c/go/+/14484/), so definitely the shim would as well. I've used signals pretty extensively in Ruby as well and never run into EINTR, so hopefully it's common for languages to handle this.

Personally I'd prefer to recommend use of STDIN, but this prevents the user from also triggering their own data collection, as they can't write to Telegraf's STDIN pipe that shared with the process, so there may be some cases where signals are preferred because they can be sent from another process.

reimda: > this prevents the user from also triggering their own data collection, as they can't write to Telegraf's STDIN pipe that shared with the process

I don't follow.  The subprocess can write a new metric to its stdout whenever it wants to.  Just because telegraf is signaling a new data collection interval with a newline on stdin doesn't mean the subprocess can't also listen for unix signals from other sources or do any other IPC to trigger collection.

Anyway, I understand the shim needs to follow the execd plugin design and this PR isn't a good place to rehash that design.  I missed that review back in Feb.  It would have been nice if I had brought this up then :)

PR #7289, opened 2020-04-05
Fix export timestamp not working for prometheus on v2

kir4h:

### Required for all PRs:

### Summary

This PR fixes the behavior for `export_timestamp` configuration in prometheus output plugin when using metrics version 2, so that the timestamp is actually issued along with the metric.

No update to the `README.md` is required I think, I take it this was the expected behaviour according to the documentation.


plugins/outputs/prometheus_client/v2/collector.go
```diff
@@ -43,11 +43,16 @@ type Collector struct {
 	coll           *serializer.Collection
 }
 
-func NewCollector(expire time.Duration, stringsAsLabel bool) *Collector {
+func NewCollector(expire time.Duration, stringsAsLabel bool, exportTimestamp bool) *Collector {
```
ssoroka, line 46: Maybe the collector should take the entire `p *PrometheusClient` as an argument?

kir4h: We would reduce the number of arguments to 1, so prettier from that side (specially if in the future we add more arguments there).

On the other hand, it feels kind of weird since the result is stored on `p` itself, and if we don't want to make a copy of `p` it should be passed through a pointer (and then why not updating the object itself in the method instead of returning it?)

```diff
- p.collector = v2.NewCollector(p.ExpirationInterval.Duration, p.StringAsLabel, p.ExportTimestamp)
+ v2.NewCollector(&p) // NewCollector modified not to return a Collector object
```
We could of course also pass by reference for memory saving but still return a collector, but not sure if its confusing semantic wise or its a common practice

```diff
- p.collector = v2.NewCollector(p.ExpirationInterval.Duration, p.StringAsLabel, p.ExportTimestamp)
+ p.collector = v2.NewCollector(&p) // NewCollector returning Collector object
``` 
Not sure if I'm making myself clear.

Issue #7293, opened 2020-04-08
Feature Request: Allow collecting and sending metrics immediately on start


concreted:
### Proposal:

Add a configuration option to perform one round of metrics collection + flush to InfluxDB immediately on start. 

### Current behavior:

Current observed behavior is that on startup, Telegraf will wait for the full flush interval before sending metrics. 

### Desired behavior:

If configured to do so, Telegraf will collect metrics from all enabled plugins and flush metrics immediately on start.

### Use case: 
This is useful when wanting to ensure that hosts will quickly send metrics when they come online and have a longer flush interval set. For our use case we are working with a network provider that will not certify devices for use on the network that communicate more than once every ~30m while idle, but is OK with an initial burst of communication on startup. With the current behavior we will not see metrics for those devices until ~30m after they are first powered on.

ssoroka:
I just wanted to confirm that you're using a low `interval` and a high `flush_interval`? With this setup you're expecting only the additional flush, and that the metrics will already have been collected.

ssoroka:
I'm also curious if you use the round_interval setting, as that has some effect on when the first flush is.

concreted:
@ssoroka In my current setup I'm using an `interval` of 10m and `flush_interval` of 28m. I don't mind lowering the `interval`, but for my use case it would be best if it actually collected metrics immediately on start also, and then flushed. That way we can get a fresh set of metrics immediately on boot.

concreted:
Also we're not using the `round_interval` setting for this use case. To clarify though, if we had `round_interval` turned on and `interval` set to `12m` for example, would it do the colleciton at `X:00, X:12, X:24, X:36, X:48`? What would happen if 60 is not divisible by `interval`?

ssoroka:
Correct, if you used it with 12m you'd get metrics every 12m on minute 0, 12, 24, 36, 48. If it's not divisible by 60 it would still collect on the round_interval requested, it just wouldn't be cyclic for one hour. eg 24m would be:
```00:00
00:24
00:48
01:12
01:36
02:00```

ssoroka:
Ok so I found out that metrics are _collected_ right away on start (as long as you're not using the round_interval, then it's on the round interval exactly). It's the _flushing_ that doesn't happen until the flush interval (30m in your case). There's another issue open that–if resolved–might solve this for you. [Option to flush buffer on request #5120](https://github.com/influxdata/telegraf/issues/5120). What do you think?

concreted:
[...]

Issue #5120, opened 2018-12-06
Option to flush buffer on request


r0nald:
## Feature Request

To have an option to flush telegraf buffer on request. For example, telegraf could (attempt to) flush the buffer on, e.g. USR1 IPC signal.

### Use case:

There are monitoring scenarios where the desirable behavior would be to flush the buffered data based on time or size most of the time, but on rare occasions, on some event, flush it ASAP.

glinton:
By flush, you mean to flush to the endpoint and not drop, correct?

r0nald:
Yes. On request telegraf could try to flush buffered data to the endpoint (e.g. InfluxDB).

voiprodrigo:
This already happens automatically with a graceful shutdown , right?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

danielnelson:
Yes, during a normal shutdown where the output is able to write, the full metric buffer is written.

PR #7302, opened 2020-04-09
add support for flush_after_first_interval_collection

ssoroka:
Support agent setting of flush_after_first_interval_collection, useful for really long collection flush durations.

Closes https://github.com/influxdata/telegraf/issues/7293

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -494,20 +494,20 @@ func (a *Agent) runOutputs(
 	startTime time.Time,
 	src <-chan telegraf.Metric,
 ) error {
-	interval := a.Config.Agent.FlushInterval.Duration
-	jitter := a.Config.Agent.FlushJitter.Duration
-
 	ctx, cancel := context.WithCancel(context.Background())
 
 	var wg sync.WaitGroup
 	for _, output := range a.Config.Outputs {
-		interval := interval
+		firstIntervalCollectionTime := a.Config.Agent.Interval.Duration + a.Config.Agent.CollectionJitter.Duration
+		flushImmediately := a.Config.Agent.FlushAfterFirstIntervalCollection
+
+		interval := a.Config.Agent.FlushInterval.Duration
 		// Overwrite agent flush_interval if this plugin has its own.
 		if output.Config.FlushInterval != 0 {
 			interval = output.Config.FlushInterval
 		}
 
-		jitter := jitter
+		jitter := a.Config.Agent.FlushJitter.Duration
 		// Overwrite agent flush_jitter if this plugin has its own.
 		if output.Config.FlushJitter != nil {
 			jitter = *output.Config.FlushJitter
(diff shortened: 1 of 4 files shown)


agent/agent.go
```diff
@@ -517,6 +517,15 @@ func (a *Agent) runOutputs(
 		go func(output *models.RunningOutput) {
 			defer wg.Done()
 
+			if flushImmediately {
+				// wait long enough for one input collection, then flush.
+				if err := internal.SleepContext(ctx, firstIntervalCollectionTime+1*time.Second); err != nil {
```
reimda, line 522: Sleeping for interval+jitter+1 feels too racy.  If we're going to add this, would it make sense to gather from inputs without timeout?

etc/telegraf.conf
```diff
@@ -54,6 +54,13 @@
   ## ie, a jitter of 5s and interval 10s means flushes will happen every 10-15s
   flush_jitter = "0s"
 
+	## FlushAfterFirstIntervalCollection instructs Telegraf to flush metrics after
+	## first starting and collecting the first round of metrics. This is
+	## specifically used for cases where your FlushInterval is really large, but
+	## you don't want to wait the full FlushInterval to see your first metrics
+	## come through. It's typically not required to set this.
+	# flush_after_first_interval_collection = false
```
reimda, line 62: We generally avoid adding true/false settings and settings that are essentially "make telegraf work for a specific user".  If we accepted changes of this type for every user request there would be too many settings.  Too many settings make it more confusing for everyone, therefore less useful for everyone.  The code also becomes full of tiny special case branches and harder to maintain and test.

I think we should put some thought into meeting this user request without a setting.  I suspect most users wouldn't care if there was an initial quick gather before normal intervals start.

If we can't find a way to make this work without a true/false setting we may need to acknowledge that telegraf isn't the right tool for the job and suggesting that the user fork and build a custom telegraf with this change.

agent/agent.go
ssoroka: It's absolutely race-y. Especially if you want to take interval out of the calculation and replace it with the_longest_avg_runtime_i_really_want_to_reasonably_wait_for_all_inputs_to_complete (worst or best variable name ever?).
The outputs have no context for how long to wait for a single round of all inputs to complete. As far as i know there's no good answer to that, and if you're going to use this feature you're going to have to either: 
1. accept a disclaimer that you might miss metrics
2. set your own time you feel is reasonable to wait until the first metric collection (but #1 is also true)

etc/telegraf.conf
ssoroka: I totally agree that using booleans to customize code paths is the road to madness, but I don't think there's anything inherently bad about booleans, as long as they couldn't be better represented by another type.

I think the above boolean doesn't convey intent very well. What I really want are two things:
```
collect_metrics_immediately_on_start = true
must_flush_metrics_on_start_after = "2m"
```
and `must_flush_metrics_on_start_after` is a symptom of the outputs not knowing how long to wait to make sure every input has reported all their metrics (and as far as i know has no way to tell).

This should give the right level of detail to configure it to any scenario, and not any one specific user.

If you look at what https://github.com/influxdata/telegraf/issues/7293 is asking, the issue is that `interval` and `flush_interval`s are not enough dials to tune for reasonable behavior when using high values for these fields for infrequently-collected and infrequently-uploaded metrics.

ssoroka:
Ok, in light of https://github.com/influxdata/telegraf/issues/5120 being a valid option and merged, and the uncertainty around when exactly is a good time to flush (because detecting completion of all collections is complicated), I'm going to close this.

Issue #7303, opened 2020-04-09
 Add the ability to specify a file with token for the HTTP input plugin.


M0rdecay:
### Proposal:
Add the ability to specify a file with token for the HTTP input plugin like in kubernetes plugins.

### Current behavior:
At the moment, it is not possible to set the authorization header without specifying the token explicitly in the configuration file. The use of environment variables for this task is also not suitable, since for this we must already have a token.

### Desired behavior:
The bearer_token parameter for the plugin, indicating the file from which to read the authorization token, which will then be added to the headers. This will allow delivering the agent to a protected environment when we should not see the token, but only indicate where to look for it.

### Use case:
This request is related to issue #7184. We put an agent that collects data from the Kubernetes API directly into the cluster, but we do not have the ability to access the token file.

M0rdecay:
@ssoroka hello! Have you made any decision? This functionality is now very lacking for us.

ssoroka:
I think we're good with this. I'm just waiting on another review to merge. I'll poke one of the team. 👍

Issue #7184, opened 2020-03-17
Allow using globs for enum processor


M0rdecay:
### Proposal:
Allow using globs in the enum plugin for field and tag keys

### Current behavior:
Currently, when using the Enum processor, you can select a field or tag only by their name.

### Desired behavior:
It will be possible to select a list of the necessary keys or fields according to the specified mask. Mapping rules apply to the entire list.
It is expected that you can use this configuration:
```toml
[[processors.enum]]
  [[processors.enum.mapping]]
    field = "*"
    [processors.enum.mapping.value_mappings]
      False = 0
      True = 1
      Unknown = 2
```

### Use case:
In our case, such a need arose when analyzing data from kubernetes nodes api (/api/v1/nodes).
A node condition is described by a key "status" in an array of types:
```
        "conditions": [
          {
            "type": "NetworkUnavailable",
            "status": "False",
            "lastHeartbeatTime": "2020-03-16T07:44:09Z",
            "lastTransitionTime": "2020-03-16T07:44:09Z",
            "reason": "CalicoIsUp",
            "message": "Calico is running on this node"
          },
          {
            "type": "MemoryPressure",
            "status": "False",
            "lastHeartbeatTime": "2020-03-17T08:15:39Z",
            "lastTransitionTime": "2020-03-16T07:42:00Z",
            "reason": "KubeletHasSufficientMemory",
            "message": "kubelet has sufficient memory available"
          },
          {
            "type": "DiskPressure",
            "status": "False",
            "lastHeartbeatTime": "2020-03-17T08:15:39Z",
            "lastTransitionTime": "2020-03-16T07:42:00Z",
            "reason": "KubeletHasNoDiskPressure",
            "message": "kubelet has no disk pressure"
          },
          {
            "type": "PIDPressure",
            "status": "False",
            "lastHeartbeatTime": "2020-03-17T08:15:39Z",
            "lastTransitionTime": "2020-03-16T07:42:00Z",
            "reason": "KubeletHasSufficientPID",
            "message": "kubelet has sufficient PID available"
          },
          {
            "type": "Ready",
            "status": "True",
            "lastHeartbeatTime": "2020-03-17T08:15:39Z",
            "lastTransitionTime": "2020-03-16T07:43:51Z",
            "reason": "KubeletReady",
            "message": "kubelet is posting ready status"
          }
        ]
```
[...]

PR #7304, opened 2020-04-09
Adds the ability to read an authorization token from a file

M0rdecay:
### Required for all PRs:

Associated Feature Request #7303

### DIFF
--- a/plugins/inputs/http/README.md
+++ b/plugins/inputs/http/README.md
@@ -26,6 +26,10 @@ The HTTP input plugin collects metrics from one or more HTTP(S) endpoints.  The 
   ## compress body or "identity" to apply no encoding.
   # content_encoding = "identity"
 
+  ## Optional file with Bearer token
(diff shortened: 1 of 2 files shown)


ssoroka:
Thanks for the PR, @M0rdecay! 
The only part that strikes me as a bit unexpected is the reading from file. Usually bearer tokens aren't that long, and given that we're already expecting the password to be put in the file, it seems like it's not necessary. Was that to primarily to avoid escaping issues? or did you have something else in mind?

M0rdecay:
@ssoroka reading a token from a file solves this problem for us - we need to deliver telegraf to a number of different protected environments when we do not need to know what the token looks like.

A special case is the use of an agent deployment in a k8s cluster to access its API via HTTPS when the token is located in the /run/secrets/kubernetes.io/serviceaccount/token file.

M0rdecay:
This particular case was described in more detail in the #7184.

PR #7311, opened 2020-04-10
Fix error in docs about exclude_retention_policy_tag

### LINKED ISSUE #7308: exclude_retention_policy_tag comment behaves the opposite of commented explanation
### Relevant telegraf.conf:

```toml
  ## If true, the 'retention_policy_tag' will not be removed from the metric.
  # exclude_retention_policy_tag = false
```

### System info:

1.14

### Steps to reproduce:

The name of the tag `exclude_retention_policy_tag` means you want the tag excluded from the metric if set to true, which is the opposite of the commented explanation:

> If true, the 'retention_policy_tag' will not be removed from the metric.

### Expected behavior:

```toml
# If true, the 'retention_policy_tag' will not be added to the metric.
```
It should match the comment for `exclude_database_tag`

```toml
# If true, the database tag will not be added to the metric.
```

### LINKED ISSUE #7803: MongoDB, add "pages written from cache"
### Proposal:
Currently the mongodb input supports already a lot of different statistics about the performance of an instance. However, the attribute "pages written from cache" is not implemented. Even mongodb says in their documentation (see: https://docs.mongodb.com/manual/reference/command/serverStatus/#serverstatus.wiredTiger.cache.pages-written-from-cache):     Number of pages written from the cache. wiredTiger.cache.pages written from cache with the wiredTiger.cache.pages read into cache can provide an overview of the I/O activity.

### Current behavior:
"wiredTiger.cache.pages written from cache" is not gathered at all.

### Desired behavior:
"wiredTiger.cache.pages written from cache" would be available for further database performance analysis.

### Use case: 
As already stated in the documentation it would be great to have that attribute to get a better overview of I/O activity of a mongodb instance.

Thank you! :-)

reimda:
resolves #7308

### DIFF
--- a/plugins/outputs/influxdb/README.md
+++ b/plugins/outputs/influxdb/README.md
@@ -23,7 +23,7 @@ The InfluxDB output plugin writes metrics to the [InfluxDB v1.x] HTTP or UDP ser
   ## tag is not set the 'database' option is used as the default.
   # database_tag = ""
 
-  ## If true, the database tag will not be added to the metric.
+  ## If true, the 'database_tag' will not be included in the written metric.
   # exclude_database_tag = false
 
   ## If true, no CREATE DATABASE queries will be sent.  Set to true when using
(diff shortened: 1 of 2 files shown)


plugins/outputs/influxdb/README.md
```diff
@@ -39,7 +39,7 @@ The InfluxDB output plugin writes metrics to the [InfluxDB v1.x] HTTP or UDP ser
   ## tag is not set the 'retention_policy' option is used as the default.
   # retention_policy_tag = ""
 
-  ## If true, the 'retention_policy_tag' will not be removed from the metric.
+  ## If true, the 'retention_policy_tag' will not be added to the metric.
```
danielnelson, line 42: Since the retention policy tag already exists on the metric, I think we should change this to "If true, ... will be removed from"

PR #7312, opened 2020-04-10
[plugins/input/docker] Make perdevice affect also cpu and add class granularity through perdevice_include/total_include 

### LINKED ISSUE #6609: plugins/inputus/docker CPU stats should respect perdevice
## Feature Request

docker has a config setting called perdevice:

```
## Whether to report for each container per-device blkio (8:0, 8:1...) and
## network (eth0, eth1, ...) stats or not
perdevice = true
```

While it applies to blkio and network, it does not apply to CPU. So I get 5 CPU values, one for each CPU.

The spirit of `perdevice` is clearly "Don't overload my network and my influxdb database with per-device metrics that I don't care about. 

### Proposal:

The same sentiment should apply to CPU metrics. Either `perdevice` value should be used, or a new `percpu` config param should be added.

### Use case:

Important to cut down bytes transferred for IOT devices on 3G networks.

kir4h:

### Required for all PRs:

### Description

This PR resolves #6609 by making the `perdevice` setting for docker input plugin affect also the `cpu` resource

### DIFF
--- a/plugins/inputs/docker/README.md
+++ b/plugins/inputs/docker/README.md
@@ -44,11 +44,26 @@ to gather stats from the [Engine API](https://docs.docker.com/engine/api/v1.24/)
   timeout = "5s"
 
   ## Whether to report for each container per-device blkio (8:0, 8:1...),
-  ## network (eth0, eth1, ...) and cpu (cpu0, cpu1, ...) stats or not
+  ## network (eth0, eth1, ...) and cpu (cpu0, cpu1, ...) stats or not.
+  ## Usage of this setting is discouraged since it will be deprecated in favor of 'perdevice_include'.
+  ## Default value is 'true' for backwards compatibility, please set it to 'false' so that 'perdevice_include' setting 
+  ## is honored.
   perdevice = true
-
-  ## Whether to report for each container total blkio and network stats or not
+  
+  ## Specifies for which classes a per-device metric should be issued
+  ## Possible values are 'cpu' (cpu0, cpu1, ...), 'blkio' (8:0, 8:1, ...) and 'network' (eth0, eth1, ...)
+  perdevice_include = ["cpu"]
+  
+  ## Whether to report for each container total blkio and network stats or not.
+  ## Usage of this setting is discouraged since it will be deprecated in favor of 'total_include'.
+  ## Default value is 'false' for backwards compatibility, please set it to 'true' so that 'total_include' setting 
+  ## is honored.
   total = false
+  
+  ## Specifies for which classes a total metric should be issued. Total is an aggregated of the 'perdevice' values.
+  ## Possible values are 'cpu', 'blkio' and 'network'  
+  ## Total 'cpu' is reported directly by Docker daemon, and 'network' and 'blkio' totals are aggregated by this plugin.
+  total_include = ["cpu", "blkio", "network"]
 
   ## docker labels to include and exclude as tags.  Globs accepted.
   ## Note that an empty array for both will include all labels as tags
--- a/plugins/inputs/docker/README.md
+++ b/plugins/inputs/docker/README.md
@@ -53,7 +53,7 @@ to gather stats from the [Engine API](https://docs.docker.com/engine/api/v1.24/)
   ## Specifies for which classes a per-device metric should be issued
   ## Possible values are 'cpu' (cpu0, cpu1, ...), 'blkio' (8:0, 8:1, ...) and 'network' (eth0, eth1, ...)
   ## Please note that this setting has no effect if 'perdevice' is set to 'true'
-  perdevice_include = ["cpu"]
+  # perdevice_include = ["cpu"]
   
   ## Whether to report for each container total blkio and network stats or not.
   ## Usage of this setting is discouraged since it will be deprecated in favor of 'total_include'.
@@ -65,7 +65,7 @@ to gather stats from the [Engine API](https://docs.docker.com/engine/api/v1.24/)
   ## Possible values are 'cpu', 'blkio' and 'network'  
   ## Total 'cpu' is reported directly by Docker daemon, and 'network' and 'blkio' totals are aggregated by this plugin.
   ## Please note that this setting has no effect if 'total' is set to 'false'
-  total_include = ["cpu", "blkio", "network"]
+  # total_include = ["cpu", "blkio", "network"]
 
   ## docker labels to include and exclude as tags.  Globs accepted.
   ## Note that an empty array for both will include all labels as tags
--- a/plugins/inputs/docker/README.md
+++ b/plugins/inputs/docker/README.md
@@ -52,6 +52,7 @@ to gather stats from the [Engine API](https://docs.docker.com/engine/api/v1.24/)
   
   ## Specifies for which classes a per-device metric should be issued
   ## Possible values are 'cpu' (cpu0, cpu1, ...), 'blkio' (8:0, 8:1, ...) and 'network' (eth0, eth1, ...)
+  ## Please note that this setting has no effect if 'perdevice' is set to 'true'
   perdevice_include = ["cpu"]
   
   ## Whether to report for each container total blkio and network stats or not.
@@ -63,6 +64,7 @@ to gather stats from the [Engine API](https://docs.docker.com/engine/api/v1.24/)
   ## Specifies for which classes a total metric should be issued. Total is an aggregated of the 'perdevice' values.
   ## Possible values are 'cpu', 'blkio' and 'network'  
   ## Total 'cpu' is reported directly by Docker daemon, and 'network' and 'blkio' totals are aggregated by this plugin.
+  ## Please note that this setting has no effect if 'total' is set to 'false'
   total_include = ["cpu", "blkio", "network"]
 
   ## docker labels to include and exclude as tags.  Globs accepted.
--- a/plugins/inputs/docker/README.md
+++ b/plugins/inputs/docker/README.md
@@ -43,8 +43,8 @@ to gather stats from the [Engine API](https://docs.docker.com/engine/api/v1.24/)
   ## Timeout for docker list, info, and stats commands
   timeout = "5s"
 
-  ## Whether to report for each container per-device blkio (8:0, 8:1...) and
-  ## network (eth0, eth1, ...) stats or not
+  ## Whether to report for each container per-device blkio (8:0, 8:1...),
+  ## network (eth0, eth1, ...) and cpu (cpu0, cpu1, ...) stats or not
   perdevice = true
 
   ## Whether to report for each container total blkio and network stats or not
--- a/plugins/inputs/docker/docker.go
+++ b/plugins/inputs/docker/docker.go
@@ -19,6 +19,7 @@ import (
 	"github.com/influxdata/telegraf"
 	"github.com/influxdata/telegraf/filter"
 	"github.com/influxdata/telegraf/internal"
+	"github.com/influxdata/telegraf/internal/choice"
 	"github.com/influxdata/telegraf/internal/docker"
 	tlsint "github.com/influxdata/telegraf/internal/tls"
 	"github.com/influxdata/telegraf/plugins/inputs"
@@ -31,12 +32,14 @@ type Docker struct {
 
 	GatherServices bool `toml:"gather_services"`
 
-	Timeout        internal.Duration
-	PerDevice      bool     `toml:"perdevice"`
-	Total          bool     `toml:"total"`
-	TagEnvironment []string `toml:"tag_env"`
-	LabelInclude   []string `toml:"docker_label_include"`
-	LabelExclude   []string `toml:"docker_label_exclude"`
+	Timeout          internal.Duration
+	PerDevice        bool     `toml:"perdevice"`
+	PerDeviceInclude []string `toml:"perdevice_include"`
+	Total            bool     `toml:"total"`
+	TotalInclude     []string `toml:"total_include"`
+	TagEnvironment   []string `toml:"tag_env"`
+	LabelInclude     []string `toml:"docker_label_include"`
+	LabelExclude     []string `toml:"docker_label_exclude"`
 
 	ContainerInclude []string `toml:"container_name_include"`
 	ContainerExclude []string `toml:"container_name_exclude"`
@@ -75,9 +78,10 @@ const (
 )
 
 var (
-	sizeRegex       = regexp.MustCompile(`^(\d+(\.\d+)*) ?([kKmMgGtTpP])?[bB]?$`)
-	containerStates = []string{"created", "restarting", "running", "removing", "paused", "exited", "dead"}
-	now             = time.Now
+	sizeRegex              = regexp.MustCompile(`^(\d+(\.\d+)*) ?([kKmMgGtTpP])?[bB]?$`)
+	containerStates        = []string{"created", "restarting", "running", "removing", "paused", "exited", "dead"}
+	containerMetricClasses = []string{"cpu", "network", "blkio"}
+	now                    = time.Now
 )
 
 var sampleConfig = `
@@ -111,11 +115,26 @@ var sampleConfig = `
   timeout = "5s"
 
   ## Whether to report for each container per-device blkio (8:0, 8:1...),
-  ## network (eth0, eth1, ...) and cpu (cpu0, cpu1, ...) stats or not
+  ## network (eth0, eth1, ...) and cpu (cpu0, cpu1, ...) stats or not.
+  ## Usage of this setting is discouraged since it will be deprecated in favor of 'perdevice_include'.
+  ## Default value is 'true' for backwards compatibility, please set it to 'false' so that 'perdevice_include' setting 
+  ## is honored.
   perdevice = true
 
-  ## Whether to report for each container total blkio and network stats or not
+  ## Specifies for which classes a per-device metric should be issued
+  ## Possible values are 'cpu' (cpu0, cpu1, ...), 'blkio' (8:0, 8:1, ...) and 'network' (eth0, eth1, ...)
+  perdevice_include = ["cpu"]
+
+  ## Whether to report for each container total blkio and network stats or not.
+  ## Usage of this setting is discouraged since it will be deprecated in favor of 'total_include'.
+  ## Default value is 'false' for backwards compatibility, please set it to 'true' so that 'total_include' setting 
+  ## is honored.
   total = false
+
+  ## Specifies for which classes a total metric should be issued. Total is an aggregated of the 'perdevice' values.
+  ## Possible values are 'cpu', 'blkio' and 'network'  
+  ## Total 'cpu' is reported directly by Docker daemon, and 'network' and 'blkio' totals are aggregated by this plugin.
+  total_include = ["cpu", "blkio", "network"]
 
   ## Which environment variables should we use as a tag
   ##tag_env = ["JAVA_HOME", "HEAP_SIZE"]
@@ -139,6 +158,21 @@ func (d *Docker) SampleConfig() string { return sampleConfig }
 // Description the metrics returned.
 func (d *Docker) Description() string {
 	return "Read metrics about docker containers"
+}
+
+// Config Validation
+func (d *Docker) Init() error {
+	err := choice.CheckSlice(d.PerDeviceInclude, containerMetricClasses)
+	if err != nil {
+		return fmt.Errorf("error validating 'perdevice_include' setting : %v", err)
+	}
+
+	err = choice.CheckSlice(d.TotalInclude, containerMetricClasses)
+	if err != nil {
+		return fmt.Errorf("error validating 'total_include' setting : %v", err)
+	}
+
+	return nil
 }
 
 // Gather metrics from the docker server.
@@ -518,7 +552,7 @@ func (d *Docker) gatherContainerInspect(
 		for _, envvar := range info.Config.Env {
 			for _, configvar := range d.TagEnvironment {
 				dockEnv := strings.SplitN(envvar, "=", 2)
-				//check for presence of tag in whitelist
+				// check for presence of tag in whitelist
 				if len(dockEnv) == 2 && len(strings.TrimSpace(dockEnv[1])) != 0 && configvar == dockEnv[0] {
 					tags[dockEnv[0]] = dockEnv[1]
 				}
@@ -565,9 +599,51 @@ func (d *Docker) gatherContainerInspect(
 		}
 	}
 
-	parseContainerStats(v, acc, tags, container.ID, d.PerDevice, d.Total, daemonOSType)
+	// Temporary logic needed for backwards compatibility until 'perdevice' setting is removed.
+	perDeviceInclude := getEffectivePerDeviceInclude(d.PerDevice, d.PerDeviceInclude, d.Log)
+	// Temporary logic needed for backwards compatibility until 'total' setting is removed.
+	totalInclude := getEffectiveTotalInclude(d.Total, d.TotalInclude, d.Log)
+
+	parseContainerStats(v, acc, tags, container.ID, perDeviceInclude, totalInclude, daemonOSType)
 
 	return nil
+}
+
+// getEffectivePerdeviceInclude returns the effective 'perdevice_include' setting to be applied.
+// This is a temporary logic to keep backwards compatibility and should be removed once 'perdevice' is deprecated.
+// If 'perdevice' is 'true', 'network' and 'blkio' are unconditionally added.
+func getEffectivePerDeviceInclude(perDevice bool, perDeviceInclude []string,
+	log telegraf.Logger) (effectivePerDeviceInclude []string) {
+	if !perDevice {
+		effectivePerDeviceInclude = perDeviceInclude
+	} else {
+		log.Warn("'perdevice' setting is set to 'true' so 'blkio' and 'network' metrics will be collected. " +
+			"Please set it to 'false' and use 'perdevice_include' instead to control this behaviour as 'perdevice' " +
+			"will be deprecated")
+		if choice.Contains("cpu", perDeviceInclude) {
+			effectivePerDeviceInclude = append(effectivePerDeviceInclude, "cpu")
+		}
+
+		effectivePerDeviceInclude = append(effectivePerDeviceInclude, "network", "blkio")
+	}
+	return
+}
+
+// getEffectiveTotalInclude returns the effective 'total_include' setting to be applied.
+// This is a temporary logic to keep backwards compatibility and should be removed once 'total' is deprecated.
+// If 'total' is false, 'network' and 'blkio' are unconditionally removed.
+func getEffectiveTotalInclude(total bool, totalInclude []string, log telegraf.Logger) (effectiveTotalInclude []string) {
+	if total {
+		effectiveTotalInclude = totalInclude
+	} else {
+		log.Warn("'total' setting is set to 'false' so 'blkio' and 'network' metrics will not be collected. " +
+			"Please set it to 'true' and use 'total_include' instead to control this behaviour as 'total' will be " +
+			"deprecated")
+		if choice.Contains("cpu", totalInclude) {
+			effectiveTotalInclude = append(effectiveTotalInclude, "cpu")
+		}
+	}
+	return
 }
 
 func parseContainerStats(
@@ -575,8 +651,8 @@ func parseContainerStats(
 	acc telegraf.Accumulator,
 	tags map[string]string,
 	id string,
-	perDevice bool,
-	total bool,
+	perDeviceInclude []string,
+	totalInclude []string,
 	daemonOSType string,
 ) {
 	tm := stat.Read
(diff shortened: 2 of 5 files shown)


kir4h:
@danielnelson @ssoroka any feedback or objection on this PR?

plugins/inputs/docker/docker_test.go
```diff
@@ -1117,3 +1119,87 @@ func TestHostnameFromID(t *testing.T) {
...
+
+				// blkio
+				for device := range deviceStatMap {
+					acc.AssertTagValuesInMeasurement(t, "docker_container_blkio", map[string]string{"device": device})
+				}
+
+			} else {
+				// cpu
+				for i := 0; i < numCpus; i++ {
+					acc.AssertTagValuesNotInMeasurement(t, "docker_container_cpu", map[string]string{"cpu": fmt.Sprintf("cpu%d", i)})
```
ssoroka, line 1190: shouldn't we see some sort of aggregate metric for total cpu time instead of nothing?

kir4h: Yes, aggregated metrics are always present regardless of `perDevice`. In the case of cpu, there is a tag with the name `cpu_total` containing the aggregated value (same goes for disk and network).

I was just focusing on asserting the differences between `perDevice` being `true` or `false`

ssoroka: thanks for clarifying!

testutil/accumulator.go
```diff
@@ -388,6 +388,65 @@ func (a *Accumulator) AssertDoesNotContainsTaggedFields(
 	return
 }
 
+// Asserts that at least one of the metrics contains all the required tags
+func (a *Accumulator) AssertTagValuesInMeasurement(
```
danielnelson, line 392: Instead of adding these new functions, use `acc.GetTelegrafMetrics()` and check for the expected metrics:
```
actual := acc.GetTelegrafMetrics()
// filter out unrelated metrics
testutil.RequireMetricsEqual(t, expected, actual, testutil.IgnoreTime())
```

kir4h:
> We need to make this more backwards compatible. One way we can do this is by adding a new option like `instances = ["cpu", "blkio", "network"]` that would control each item separately, and deprecate `perdevice`. If `perdevice` is set to true it would add the blkio and network options to the new option.

Yup, hadn't realized the change was not backwards compatible (same configuration will stop reporting per cpu metrics) so I will work on adding a new option as suggested.

testutil/accumulator.go
kir4h: The reason why I was choosing this approach is being more specific in what I'm testing (in this case whether those tags are present of not), rather than comparing the whole metric and having to define the expected metric result in my test (filtering out unrelated metrics as you suggested will help for sure reducing the noise).

Another reason is that in general, I prefer having specific test messages (in this case we would get a failure stating if it was related to `docker_container_cpu`, `docker_container_net` or `docker_container_blkio`; if asserting the whole metric we would get to compare the whole metric to see what is wrong).

Anyway it's your call, if you think it's better to assert the whole metric (or your guidelines follow this approach and you want an homogeneous approach) I can remove this new function and change the test to assert the whole metric.

kir4h: Thanks for the feedback, I will refactor to follow the guidelines you provided :)
> I'm trying to phase out all of the other acc.Assert* functions as well
....
Sorry about all the inconsistent styling in the tests, I realize this makes it impossible to follow the style we prefer.

I was thinking that maybe we should include some comments regarding this on the contributing guidelines (state `acc.Assert*` functions will be deprecated and encourage the use of the whitelisted `testutil` functions that should be used )

kir4h:
@danielnelson while starting to implement the changes I realized that there is an extra setting controlling whether to report the aggregated metrics for blkio and network.

```toml
  ## Whether to report for each container total blkio and network stats or not
  total = false
```

This doesn't apply to `cpu`, whose total is always reported.

Should we also make `cpu_total` optional? In any case I guess following the same philosophy we should at least make it configurable per network or per blkio.

```toml
  ## Whether to report per instance statistics for these resources.
  [inputs.docker.instances]
    ## Per core statistics (cpu0, cpu1, ...).
    cpu = true
    ## Per blkio statistics (8:0, 8:1, ...).
    blkio = true
    ## Per network interface statistics (eth0, eth1, ...).
    network = true
 ## Whether to report aggregated statistics for these resources 
 [inputs.docker.total]
    ## Aggregated stats for cpu
    cpu = true
    ## Aggregated stats for blkio
    blkio = true
    ## Aggregated stats for network interfaces
    network = true
``` 

Does this approach makes sense? We could also include all these booleans under the `inputs.docker.instances` setting with different names. I'm not 100% sure about the `instances` term, but I haven't found a better name not including the `device` word.

danielnelson:
Actually, if we were starting from a blank slate I wouldn't want to add the total stats at all and would always add per device.  The total stats don't provide any new information and make it more difficult to aggregate the values since they must be excluded.  Totals can always be created from the per device values and can be computed in the basic stats aggregator before insertion to the database if wanted.  If a user wants neither per device or totals, the namedrop option could be used to remove all data.

This makes me think we should drop this pull request and consider deprecating `perdevice` and `total`, updating the available options to make it possible to turn off the cpu-totals.

Presumably this wasn't what you were originally going for, did you originally want to disable the per cpu stats to reduce the amount of metrics?

kir4h:
> Actually, if we were starting from a blank slate I wouldn't want to add the total stats at all and would always add per device. The total stats don't provide any new information and make it more difficult to aggregate the values since they must be excluded. 

Indeed, and I think that should be also the general way of looking at metrics. In my case for cpu I use the `_percent` which is not per cpu so I hadn´t thought about that, but having to exclude the `total` sounds kind of ugly.

> Totals can always be created from the per device values and can be computed in the basic stats aggregator before insertion to the database if wanted. If a user wants neither per device or totals, the namedrop option could be used to remove all data.

Yes, and I think this would be the more "telegrafic" way, leveraging aggregator plugins to achieve this and using built-in mechanisms like namedrop to provide such funcionality.

> This makes me think we should drop this pull request and consider deprecating `perdevice` and `total`, updating the available options to make it possible to turn off the cpu-totals.

I´m ok in working on a new ticket to do so. If so, we should maybe explain this scenario in https://github.com/influxdata/telegraf/issues/6609 for future readers and then open a new one.
I guess we want to maintain backwards compatibility while doing so right?

> Presumably this wasn't what you were originally going for, did you originally want to disable the per cpu stats to reduce the amount of metrics?

The issue was actually not created by me but that was the goal, quoting the requestor: `Don't overload my network and my influxdb database with per-device metrics that I don't care about.`.

And this is a totally legitimate use case, but could be accomplished by the basic stats aggregator.
On the other hand, it is true that by doing this action on the input plugin we reduce the load (we are actually not collecting metrics that will be aggregated and dropped later on but rather using an aggregated one that is already provided by docker, so for IoT devices this might actually be a point to consider)

danielnelson:
Here is a summary of the tasks that need completed then:

1. Add ability to enable/disable cpu totals with a new option `aggregate_totals = []` (open to other naming suggestions).
   - Possibly values are "blkio", "network", "cpu".
   - Default value is [].
   - If `total` is set we add "blkio" and "network" to the set.
   - Deprecate this new option, recommend basicstats aggregator.

2. Deprecate the `total` option, warn if set and recommend basicstats. [reference these notes](https://github.com/influxdata/telegraf/wiki/Deprecation#deprecate-options)

3. Deprecate the `perdevice` option, warn if it is set to false, but keep it working.

4. Create a sample configuration for aggregating blkio/network/cpu with basicstats.

> On the other hand, it is true that by doing this action on the input plugin we reduce the load

This is true, but I think we should provide an option like this only when there is a way to have Docker report the aggregated values directly.   Adding aggregations, even common ones, in each plugin just isn't very scalable for maintenance.

kir4h:
>Add ability to enable/disable cpu totals with a new option aggregate_totals = [] (open to other >naming suggestions).
>
>Possibly values are "blkio", "network", "cpu".

Wondering: Free text list vs. a struct with booleans?

```toml
[agggregate_totals]
  blkio = true
  network = true
  cpu = true
```

Not that I prefer booleands, just thinking that the booleans would have `toml` validation out of the box and the list requires writing the validation? (maybe there is a way of validating it directly with toml, I haven´t lot of experience with it)

 > > On the other hand, it is true that by doing this action on the input plugin we reduce the load
> 
> This is true, but I think we should provide an option like this only when there is a way to have Docker report the aggregated values directly. Adding aggregations, even common ones, in each plugin just isn't very scalable for maintenance.

Docker is actually reporting the aggregated values directly for the cpu, according to https://docs.docker.com/engine/api/v1.21/#21-containers (`GET /containers/(id or name)/stats` endpoint)

```json
"cpu_stats" : {
        "cpu_usage" : {
           "percpu_usage" : [
              8646879,
              24472255,
              36438778,
              30657443
           ],
           "usage_in_usermode" : 50000000,
           "total_usage" : 100215355,
           "usage_in_kernelmode" : 30000000
        },
        "system_cpu_usage" : 739306590000000,
        "throttling_data" : {"periods":0,"throttled_periods":0,"throttled_time":0}
     },
```

`total_usage` is the aggregation of `percpu_usage` .

This is not the case for `blkio` and `network` stats (that makes me thing that this was actually the reason why the original implementation had the perdevice setting only for network and blkio while cpu was always included).

Should this change the proposed approach? Providing different choices for cpu ("total" and "per cpu" can be enabled/disabled independently) and network/blkio (in this case since docker is not providing aggregates, aggregation should be done through the basicstats aggregator and we should deprecate current `total` option)

danielnelson:
> Free text list vs. a struct with booleans?

I'd rather stick to a list of strings.  Tables are a more compilicated to use if you are less familiar with TOML, on top of that we have some issues reporting certain errors due to limitations of our TOML parser.  I do wish we could validate the list of strings without code, but you can use `choice.Check()` in the Init() function.

> Docker is actually reporting the aggregated values directly for the cpu... Should this change the proposed approach?

Well that does make things more interesting, I don't have a hard and fast rule for this.  I think one mistake we made originally was to name the per cpu usage `usage_total`, because this can cause it to be aggregated along with the all cpu `usage_total` if care is not taken.  I wouldn't mind renaming the per cpu field to `usage` to avoid that but I think it's not worth the disruption.  Let's keep the cpu data as-is and only work on network/blkio?

kir4h:
> but you can use `choice.Check()` in the Init() function.

Nice! If there is an already internal function it doesn't seem so weird

> > Docker is actually reporting the aggregated values directly for the cpu... Should this change the proposed approach?
> 
> Let's keep the cpu data as-is and only work on network/blkio?

CPU
---
If we keep cpu as it is, it will always report both per cpu and total metrics. In this case since total is precalculated by docker it makes sense to have it available, so maybe just working on documentation side to:
* Add an example of how to drop per cpu metrics for someone wanting only the aggregated (no need of aggregator for cpu). I guess this should be a tagdrop selector, but I think we cannot use regular expressions so I guess the best might be something like 

```toml
 [inputs.docker.tagdrop]
    cpu = [ "cpu[0-9]*" ]
```
* Add an example of how to drop total cpu metric for someone wanting to use the basic aggregator (not sure if there is a use case, maybe for other stats other than sum).

```toml
 [inputs.docker.tagdrop]
    cpu = [ "total" ]
```

Network/Blkio
---

Follow what you suggested 

1. Add ability to enable/disable cpu totals with a new option aggregate_totals = [] (open to other naming suggestions).
    * Possibly values are "blkio", "network".
    * Default value is `[]`.
    * If total is set we add "blkio" and "network" to the set.
    * Deprecate this new option, recommend basicstats aggregator.

1. Deprecate the total option, warn if set and recommend basicstats. (Keep it working)

1. Deprecate the perdevice option, warn if it is set to false, but keep it working.

1. Create a sample configuration for aggregating blkio/network with basicstats.

srebhan:
Hey @kir4h any update on this one?

kir4h:
Had completely lost track of this one. Rereading a little, it seems my last comment was a proposal for a way forward for CPU/Network/Blkio and I was waiting for confirmation from @danielnelson (but I just noticed I didn't ask it directly but rather just write my proposed way forward).

I can work on it if my comment sounds good, or you can take it if you wish @srebhan (either reusing this branch or starting from clean if you prefer)

srebhan:
Hey @kir4h sorry for coming back to this so late. :-/
I think your suggestions are fine and if I understand it correctly you suggest to filter the *individual* values for cpu using `tagdrop` and filtering the *totals* with an option (alternatively using basicstats). Is this correct?
If so, my preferred solution would be to filter the individual cpu values as you suggest (adding a short readme section) and allow to filter the totals with the option. Please also include a `cpu` choice there to remove the total cpu metric.
The reason to not use basicstats in my view is that here you can be sure the requested values are in sync. That is you can be sure that you aggregate values from the (almost) same timestep, while basicstats relies on a nicely chosen `period`.

srebhan:
@kir4h any comment or update on this one?

kir4h:
> @kir4h any comment or update on this one?

Sorry, been kind of busy and forgot about answering

> I think your suggestions are fine and if I understand it correctly you suggest to filter the _individual_ values for cpu using `tagdrop`

Correct

>...and filtering the _totals_ with an option (alternatively using basicstats). Is this correct?

No, I was rather suggesting filtering out the totals the same way, using `tagdrop`. My reference to `basicstats` was just trying to find a use case for dropping the totals, since total cpu is already provided directly by docker.

> If so, my preferred solution would be to filter the individual cpu values as you suggest (adding a short readme section) and allow to filter the totals with the option. Please also include a `cpu` choice there to remove the total cpu metric.

At the moment both `perdevice` and `total` are shared accross `blkio` and `network` (changing one of them will affect both of them). You propose to create a `cpu_total` boolean config value that models whether to include or not total values for cpu, right?  (As mentioned above, filtering with tagdrop and documenting it is another option)

> The reason to not use basicstats in my view is that here you can be sure the requested values are in sync. That is you can be sure that you aggregate values from the (almost) same timestep, while basicstats relies on a nicely chosen `period`.

Agree; haven't worked with basicstats in a while but given that the `total` metric is provided directly by Docker it should be the most reliable source of truth.

srebhan:
Still not sure if I get you. Sorry! There are three classes `cpu`, `blkio` and `network` with two granularities of `perdevice` and `total`. Is this correct? Now you want to add a "class" tag and a "granularity" tag to be able to filter them with `tagdrop` right?
Computing the totals for `blkio` and `network` in the plugin is ok IMO.

If this is the case I can support your approach, otherwise you have to explain again what you want to do... :-S :-)

kir4h:
> Still not sure if I get you. Sorry! 

No worries, I had to reread the whole conversation and check the code before replying since I have wiped this from my mind since creating the PR :)

> There are three classes `cpu`, `blkio` and `network` with two granularities of `perdevice` and `total`. Is this correct? 

No, that's the thing. The behaviour you describe happens only for `blkio` and `network`. For `cpu`, `total` and `perdevice` are always reported.  The initial scope of this PR was to address #6609 by changing the behaviour of `perdevice`  boolean setting so that is also applies to `cpu` metrics.

After the initial PR, @danielnelson noticed this would break backwards compatibility (people having `perdevice=false` in their config would stop receiving `perdevice` metrics for `cpu`), and that's when things started to get more complicated and we thought of possible approaches (for instance being able to control this in a granular way where we could enable `perdevice` for `network` and not for `blkio` for instance)

Another difference between `cpu` and `network` / `blkio` is that in the case of `cpu` docker is already reporting the total value (for `network` and `blkio` the plugins is aggregating them)

> Now you want to add a "class" tag and a "granularity" tag to be able to filter them with `tagdrop` right?

Initially I just wanted to extend the `perdevice` setting so that it also affects `cpu`. 

If we want to maintain this backwards compatibility, I was suggesting just giving instructions on how one can just drop those `cpu` `perdevice` metrics by using `tagdrop` (no need to add class or granularity for this).

> If this is the case I can support your approach, otherwise you have to explain again what you want to do... :-S :-)

Basically, some of the options are:

* Make `perdevice` setting affect also `cpu` (we could extend this to `total` setting if we wanted, not sure there is a use case). 
  * Pros: Already in this PR, simpler. 
  * Cons: Breaking background compatibility as same config would produce different output
* Don't touch the code but just add an example on how to achieve the desired behaviour by configuration using `tagdrop`. 
  * Pros: Nothing to do, just document.
  * Cons: We keep some inconsistency as having `perdevice` affect only `blkio` and `network` might be confusing
* Deeper transformation where we add granularity (ability to select `perdevice` and `total` for the three resources independently), but in the meantime keep background compatibility. 
  * Pros: Granularity, more intuitive config
  * Cons: Complexity, Maintaining background compatibility might be tricky)

 There are of course other options. Daniel was in favor of deprecating `total` and `perdevice` settings to rely on basicstats for aggregation (except for `cpu` since Docker provides the metric)

Hope it's clearer, feel free to ask any clarification as this topic is kind of convoluted now with so many comments :)

srebhan:
Ah ok now I see. So for me the easiest would be to define two new options `perdevice_include` and `total_include`. Both are a list that can contain `cpu`, `blkio` and/or `network`. If the option is in the list we output the metric otherwise we skip (we might even want to skip the query...). The default is `perdevice_include=["cpu"]`, `total_include=["cpu", "blkio", "network"]`. An empty list means output no metric for the granularity. If `perdevice=true` we internally add `"blkio"` and `"network"` to `perdevice_include` no matter if they are specified explicitly. This way we can copy the old behavior while allowing a more fine-grained control.

Then we might note in the `sampleConfig` and readme that `perdevice` should not be used anymore. We can fade out support if/when a new major release of telegraf appears...

What do you think? Using `internal.choice` eases handling of those lists, so it's not a big deal code-wise.

kir4h:
Seems like a good approach to me! 

Summarizing the four scenarios

| Setting  | Value | Behaviour |
| ------------- | ------ | ------------|
| `perdevice`  | `true`  | `blkio` and `network` unconditionally added to `perdevice_include` | 
| `perdevice`  | `false`  | Nothing to do, `perdevice_include` is honored| 
| `total`  |  `true` | Nothing to do, `total_include` is honored |
| `total` | `false` |  `blkio` and `network` unconditionally removed from `perdevice_include` |

If `perdevice` is `true` or `total` is `false` we might want to log some message to warn about deprecation and that the new variables are partially overriden to keep backwards compatibility?

srebhan:
Seems reasonable to me. The important thing is that the plugin should behave as if the old options are still valid. So yes the old variables will override the new ones to keep backward compatible just as you outlined in your table.
You might also want to use `filter` as it offers an include/exclude pattern... But maybe that's too much here.

kir4h:
@srebhan I have implemented the approach. Feel free to add any comment if something needs to be changed/addressed. CI tests fail because of timeout on `tail` input plugin, so I don't think this PR has anything to do with it.

plugins/inputs/docker/README.md
```diff
@@ -43,12 +43,27 @@ to gather stats from the [Engine API](https://docs.docker.com/engine/api/v1.24/)
...
+  ## Usage of this setting is discouraged since it will be deprecated in favor of 'perdevice_include'.
+  ## Default value is 'true' for backwards compatibility, please set it to 'false' so that 'perdevice_include' setting 
+  ## is honored.
   perdevice = true
-
-  ## Whether to report for each container total blkio and network stats or not
+  
+  ## Specifies for which classes a per-device metric should be issued
+  ## Possible values are 'cpu' (cpu0, cpu1, ...), 'blkio' (8:0, 8:1, ...) and 'network' (eth0, eth1, ...)
+  perdevice_include = ["cpu"]
```
srebhan, line 55: You should comment that line (as it is the default) and maybe add
`This setting has no effect if perdevice is set to true.`

plugins/inputs/docker/README.md
```diff
@@ -43,12 +43,27 @@ to gather stats from the [Engine API](https://docs.docker.com/engine/api/v1.24/)
...
+  ## Whether to report for each container total blkio and network stats or not.
+  ## Usage of this setting is discouraged since it will be deprecated in favor of 'total_include'.
+  ## Default value is 'false' for backwards compatibility, please set it to 'true' so that 'total_include' setting 
+  ## is honored.
   total = false
+  
+  ## Specifies for which classes a total metric should be issued. Total is an aggregated of the 'perdevice' values.
+  ## Possible values are 'cpu', 'blkio' and 'network'  
+  ## Total 'cpu' is reported directly by Docker daemon, and 'network' and 'blkio' totals are aggregated by this plugin.
+  total_include = ["cpu", "blkio", "network"]
```
srebhan, line 66: Please also comment this line (as it is the default). IIRC this option is only taken into account if total is not false? If so please add the respective comment again here.

plugins/inputs/docker/docker.go
```diff
@@ -110,13 +114,28 @@ var sampleConfig = `
...
+  ## Usage of this setting is discouraged since it will be deprecated in favor of 'total_include'.
+  ## Default value is 'false' for backwards compatibility, please set it to 'true' so that 'total_include' setting 
+  ## is honored.
   total = false
 
+  ## Specifies for which classes a total metric should be issued. Total is an aggregated of the 'perdevice' values.
+  ## Possible values are 'cpu', 'blkio' and 'network'  
+  ## Total 'cpu' is reported directly by Docker daemon, and 'network' and 'blkio' totals are aggregated by this plugin.
+  total_include = ["cpu", "blkio", "network"]
+
```
srebhan, line 138: Comments from README also apply here.

plugins/inputs/docker/docker.go
```diff
@@ -141,6 +160,21 @@ func (d *Docker) Description() string {
 	return "Read metrics about docker containers"
 }
 
+// Config Validation
+func (d *Docker) Init() error {
+	err := choice.CheckSlice(d.PerDeviceInclude, containerMetricClasses)
```
srebhan, line 165: What happens if `p.PerDeviceInclude` is empty? The check will not trigger. However, other plugins assume that all options are set if the include list is empty. You might also want to do this. In any case, please document the behavior in `sampleConfig` and the README.

plugins/inputs/docker/docker.go
```diff
@@ -141,6 +160,21 @@ func (d *Docker) Description() string {
...
+	if err != nil {
+		return fmt.Errorf("error validating 'perdevice_include' setting : %v", err)
+	}
+
+	err = choice.CheckSlice(d.TotalInclude, containerMetricClasses)
+	if err != nil {
+		return fmt.Errorf("error validating 'total_include' setting : %v", err)
+	}
+
```
srebhan, line 174: Is having both lists empty a valid option? If not, please check here and return an error.

plugins/inputs/docker/docker.go
```diff
@@ -565,18 +599,60 @@ func (d *Docker) gatherContainerInspect(
 		}
 	}
 
-	parseContainerStats(v, acc, tags, container.ID, d.PerDevice, d.Total, daemonOSType)
+	// Temporary logic needed for backwards compatibility until 'perdevice' setting is removed.
+	perDeviceInclude := getEffectivePerDeviceInclude(d.PerDevice, d.PerDeviceInclude, d.Log)
+	// Temporary logic needed for backwards compatibility until 'total' setting is removed.
+	totalInclude := getEffectiveTotalInclude(d.Total, d.TotalInclude, d.Log)
```
srebhan, line 605: Please do it once in the `Init()` function. This way `Gather()` is clean wrt backward-case handling.

plugins/inputs/docker/docker.go
```diff
@@ -565,18 +599,60 @@ func (d *Docker) gatherContainerInspect(
...
+			"Please set it to 'false' and use 'perdevice_include' instead to control this behaviour as 'perdevice' " +
+			"will be deprecated")
+		if choice.Contains("cpu", perDeviceInclude) {
+			effectivePerDeviceInclude = append(effectivePerDeviceInclude, "cpu")
+		}
+
+		effectivePerDeviceInclude = append(effectivePerDeviceInclude, "network", "blkio")
+	}
+	return
+}
```
srebhan, line 630: You might ease your life drastically if you make this a `(d *Docker)` member function. This way you could write:
```suggestion
func (d *Docker) getEffectivePerDeviceInclude() {
	if perDevice {
		d.Log.Warn("'perdevice' setting is set to 'true' so 'blkio' and 'network' metrics will be collected. " +
			"Please set it to 'false' and use 'perdevice_include' instead to control this behaviour as 'perdevice' " +
			"will be deprecated")
		d.PerDeviceInclude = []string{"network", "blkio")  // Or whatever the current default is
	}
}
```

These few line could also directly be added to the `Init()` function... Furthermore, you could define the deprecation warning string as constant somewhere at the top of the file so we can shorten the code.

plugins/inputs/docker/docker.go
```diff
@@ -565,18 +599,60 @@ func (d *Docker) gatherContainerInspect(
...
+		log.Warn("'total' setting is set to 'false' so 'blkio' and 'network' metrics will not be collected. " +
+			"Please set it to 'true' and use 'total_include' instead to control this behaviour as 'total' will be " +
+			"deprecated")
+		if choice.Contains("cpu", totalInclude) {
+			effectiveTotalInclude = append(effectiveTotalInclude, "cpu")
+		}
+	}
+	return
+}
+
```
srebhan, line 648: Same as for `getEffectivePerDeviceInclude()`.

plugins/inputs/docker/docker_test.go
```diff
@@ -1117,3 +1125,191 @@ func TestHostnameFromID(t *testing.T) {
...
+			}
+
+			numNetworks := len(stats.Networks)
+
+			deviceStatMap := getDeviceStatMap(stats.BlkioStats)
+
+			// per device
+			for i := 0; i < numCpus; i++ {
+				if choice.Contains("cpu", tt.args.perDeviceInclude) {
+					acc.AssertTagValuesInMeasurement(t, "docker_container_cpu",
```
srebhan, line 1198: I think you could achieve the same with [a cmp option](https://pkg.go.dev/github.com/google/go-cmp/cmp#FilterPath) added to `RequireMetricEqual()` instead of inventing a new function. Maybe we could add
```golang
func OnlyTags() cmp.Option {
	f := func(p cmp.Path) bool {return p.String() != "Tags"}
	return cmp.FilterPath(f, cmp.Ignore())
}
````
to testutils instead...

File: plugins/inputs/docker/docker_test.go [hunk 12]
```diff
@@ -1117,3 +1125,191 @@ func TestHostnameFromID(t *testing.T) {
...
+						map[string]string{"cpu": fmt.Sprintf("cpu%d", i)})
+				} else {
+					acc.AssertTagValuesNotInMeasurement(t, "docker_container_cpu",
+						map[string]string{"cpu": fmt.Sprintf("cpu%d", i)})
```
srebhan, line 1202: This is probably very specific to your plugin. I suggest to include this test function here, using the `choice` option for doing the work.

srebhan (commented):
I suggest to do the backward-compatible `include` handling in the `Init()` function. This way we have the processing itself future-proof already.
Furthermore, please think about your additions to testutils. Here I suggest to rather add a new `cmp` option instead of hand-coding the comparison. See my comments in the code.

kir4h:
Thanks for your feedback @srebhan! I will work on addressing the PR. I might be short on time during the week so probably on the week-end.

srebhan:
@kir4h thank you very much! Please close my comments as resolved, so I know what is still open (if anything). This eases my life a bit... ;-)

plugins/inputs/docker/README.md
kir4h: > You should comment that line (as it is the default)

It is the default only because I've set this value in the default config, but I am not defaulting to this value in code. 

If I default to it in code (i.e: on `Init()`), I'm thinking I should set `omitempty` on its toml annotation so that I am able to differentiate an intentional empty value (`perdevice_include=[]`, user does not want any perdevice metric) from a missing configuration item (this could happen if the user is using an old config with a new Telegraf version and doesn't have this new property in it; in this case, I should default to the backwards compatible value)

Does this make sense?


> and maybe add `This setting has no effect if perdevice is set to true.`
:+1:

plugins/inputs/docker/docker.go
kir4h: I would say it is; by having both lists empty there are still a couple of lines written by the plugin (memory, container stats...)

```
> docker,engine_host=myhost,host=myhost,server_version=19.03.12 n_containers=3i,n_containers_paused=0i,n_containers_running=0i,n_containers_stopped=3i,n_cpus=8i,n_goroutines=34i,n_images=68i,n_listener_events=0i,n_used_file_descriptors=21i 1612686961000000000
> docker,engine_host=myhost,host=myhost,server_version=19.03.12 memory_total=33616916480i 1612686961000000000
```

plugins/inputs/docker/docker.go
kir4h: Indeed! Didn't realize I was polluting `Gather()` which is not suited for a one time logic

plugins/inputs/docker/docker.go
kir4h: Yup, resultant code is definitely shorter, thx!

plugins/inputs/docker/docker_test.go
kir4h: Implemented this approach.

First time using options, had to add an extra condition to get it working


```
f := func(p cmp.Path) bool { return p.String() != "Tags" && p.String() != "" }
```
Does it make sense?

kir4h:
@srebhan Implemented some of your suggestions. 

Pending your feedback [here](https://github.com/influxdata/telegraf/pull/7312/files/2eb9fd4d8fe872d04e44014ce434922e6ceb0612#r571582197) (changes pending on my side but confirming just in case) and [here](https://github.com/influxdata/telegraf/pull/7312/files/2eb9fd4d8fe872d04e44014ce434922e6ceb0612#r571647747) (not 100% sure if the extra condition I added is correct)

Thx!

plugins/inputs/docker/README.md
srebhan: Well what should the default be to keep the previous behavior? That's up to you to decide. :-) Anyway, the option should be commented with the default used in code.

However, I guess 'omitemty' is only available for _encoding_ but not decoding toml. The easiest way is to specify the default when constructing the `Docker` instance (i.e. usually in `init()`). If you set `["cpu"]` there, you can be sure that the user was overriding the option if the field is empty in `Init()`. I would not do any magic here and set the default in `init()` and use what we get in `Init()`...

plugins/inputs/docker/docker_test.go
srebhan: Huh? If the path entry is empty we don't want to ignore it? Whatever, do whatever is necessary. I'll test the result on my side anyway... :-)

plugins/inputs/docker/README.md
srebhan: Please make this the default in code and comment this line. IIRC this is what you need to reproduce the past behavior even if `perdevice` is `true`.

plugins/inputs/docker/README.md
srebhan: Still open.

plugins/inputs/docker/docker_test.go
srebhan: Your code is absolutely correct. It seems that the root of a structure has an empty path. Who could have expected this. :-) Keep as is and mark as resolved.

srebhan:
@kir4h your testing function is correct (see my comment), but you need to change the default thing. To set the defaults and let the user override them do the following in `init()`:
```patch
func init() {
	inputs.Add("docker", func() telegraf.Input {
		return &Docker{
			PerDevice:      true,
			Timeout:        internal.Duration{Duration: time.Second * 5},
			Endpoint:       defaultEndpoint,
+			PerDeviceInclude: []string{"cpu"},
+			TotalInclude: []string{"cpu", "blkio", "network"},
			newEnvClient:   NewEnvClient,
			newClient:      NewClient,
			filtersCreated: false,
		}
	})
}

```
and afterwards comment the options in `sampleConfig` and `README.md`. You probably have to fix the formatting...

plugins/inputs/docker/README.md
kir4h: You're right! didn't realize `omitempty` was only for encoding purposes. I have implemented the `init()` approach as suggested

plugins/inputs/docker/README.md
kir4h: Yup, hadn't make this change until I had a way forward for the default in code. Resolving it now.

kir4h:
@srebhan Done (except for the formatting fix,  not sure what it's related to; I've run `go fmt` and for the text part I have formatted the same way other variables are formatted)

srebhan (approved):
Looks fine thanks for working on this for such a long time!

kir4h:
Thanks for your valuable feedback to improve the PR!

Last thing: Topic for the branch was `Make perdevice setting for docker input plugin affect also cpu` , but finally we have also implemented the ability to choose `perdevice` and `total` on a per class basis with the new options `perdevice_include` and `total_include`. We should probably reflect this on the commit message when sqashing into master so that it´s included in the Changelog.

srebhan:
Right, so feel free to change the PR title to reflect the change in scope.

PR #7316, opened 2020-04-11
plugins/filestat: Skip missing files

### LINKED ISSUE #7315: plugins/filestat: probing files that don't exist results in `Unable to get info for file "/does/not/exist", possible permissions issue`
### Relevant telegraf.conf:

```toml
[[inputs.filestat]]
  files = ["/does/not/exist"]
```

### System info:

Telegraf 1.14.0-1 from the `bionic` repo `https://repos.influxdata.com/ubuntu/`.

### Steps to reproduce:

1. Ensure the file `/does/not/exist` does not exist
2. Setup an input config with the configuration snippet above & restart telegraf
### Expected behavior:

If the file doesn't exist: No log message.

If the file does exist but can't be probed (e.g. on an unreadable directory): A log message giving the actual error.

### Actual behavior:

1. In system logs: telegraf logs that it can't probe the file, but guesses incorrectly that the permissions are wrong: `[inputs.filestat] Unable to get info for file "/does/not/exist", possible permissions issue`
2. The filestat measurement about the file is correctly reported with `exists: 0`.

### Additional info:

A PR exists for this issue, at #6940: it attempts to fix this with an additional configuration flag. I think that is unnecessary & will submit another PR with a less config-heavy fix.

antifuchs:
This PR attempts to fix #7315, the same problem as #6940, but without adding a config flag.

Previously, the filestat plugin would check for a file's existence, and then try to compute all sorts of things for the missing file, which failed and resulted in a log message about possible permission errors.

Now, the plugin just skips the file if it exists. If an error is encountered probing an existing file, the plugin now logs that error, too, instead of hazarding a guess at permission problems.

### Required for all PRs:

- [x] Associated README.md updated. (no update needed)
- [x] Has appropriate unit tests: Added a test.

### DIFF
--- a/plugins/inputs/filestat/filestat.go
+++ b/plugins/inputs/filestat/filestat.go
@@ -35,11 +35,18 @@ type FileStat struct {
 
 	// maps full file paths to globmatch obj
 	globs map[string]*globpath.GlobPath
+
+	// files that were missing - we only log the first time it's not found.
+	missingFiles map[string]bool
+	// files that had an error in Stat - we only log the first error.
+	filesWithErrors map[string]bool
 }
 
 func NewFileStat() *FileStat {
 	return &FileStat{
-		globs: make(map[string]*globpath.GlobPath),
+		globs:           make(map[string]*globpath.GlobPath),
+		missingFiles:    make(map[string]bool),
+		filesWithErrors: make(map[string]bool),
 	}
 }
 
@@ -86,13 +93,22 @@ func (f *FileStat) Gather(acc telegraf.Accumulator) error {
 			if os.IsNotExist(err) {
 				fields["exists"] = int64(0)
 				acc.AddFields("filestat", fields, tags)
+				if !f.missingFiles[fileName] {
+					f.Log.Warnf("File %q not found", fileName)
+					f.missingFiles[fileName] = true
+				}
 				continue
 			}
+			f.missingFiles[fileName] = false
 
 			if fileInfo == nil {
-				f.Log.Errorf("Unable to get info for file %q: %v",
-					fileName)
+				if !f.filesWithErrors[fileName] {
+					f.filesWithErrors[fileName] = true
+					f.Log.Errorf("Unable to get info for file %q: %v",
+						fileName, err)
+				}
 			} else {
+				f.filesWithErrors[fileName] = false
 				fields["size_bytes"] = fileInfo.Size()
 				fields["modification_time"] = fileInfo.ModTime().UnixNano()
 			}
--- a/plugins/inputs/filestat/filestat.go
+++ b/plugins/inputs/filestat/filestat.go
@@ -85,10 +85,12 @@ func (f *FileStat) Gather(acc telegraf.Accumulator) error {
 			fileInfo, err := os.Stat(fileName)
 			if os.IsNotExist(err) {
 				fields["exists"] = int64(0)
+				acc.AddFields("filestat", fields, tags)
+				continue
 			}
 
 			if fileInfo == nil {
-				f.Log.Errorf("Unable to get info for file %q, possible permissions issue",
+				f.Log.Errorf("Unable to get info for file %q: %v",
 					fileName)
 			} else {
 				fields["size_bytes"] = fileInfo.Size()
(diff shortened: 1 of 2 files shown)


srebhan (commented):
I'm not sure we want to ignore the files silently. What if the user misspelled the filename? He probably will notice quite late (when analyzing the data). So my guts-feeling is we should log at least once.
How about adding a `map[string]bool` to the struct for remembering which filename was logged already and output a warning on the first encounter of the missing file. Then set the map entry for the file to `true` and skip any further warning-outputs...

How does that sound to you?

srebhan:
@antifuchs any update?

antifuchs:
Ah, sorry I missed this! I will look into getting this implemented.

antifuchs:
@srebhan Updated - I hope this reflects what you had in mind.

plugins/inputs/filestat/filestat.go
```diff
@@ -35,11 +35,18 @@ type FileStat struct {
 
 	// maps full file paths to globmatch obj
 	globs map[string]*globpath.GlobPath
+
+	// files that were missing - we only log the first time it's not found.
+	missingFiles map[string]struct{}
+	// files that had an error in Stat - we only log the first error.
+	filesWithErrors map[string]struct{}
```
srebhan, line 42: How about
```suggestion
	missingFiles map[string]bool
	// files that had an error in Stat - we only log the first error.
	filesWithErrors map[string]bool
```

plugins/inputs/filestat/filestat.go
```diff
@@ -85,12 +92,24 @@ func (f *FileStat) Gather(acc telegraf.Accumulator) error {
...
 				fields["exists"] = int64(0)
+				acc.AddFields("filestat", fields, tags)
+				if _, ok := f.missingFiles[fileName]; !ok {
+					f.Log.Warnf("File %q not found", fileName)
+					f.missingFiles[fileName] = struct{}{}
+				}
+				continue
+			} else {
+				delete(f.missingFiles, fileName)
 			}
```
srebhan, line 103: I think you can simplify it to
```suggestion
				if ! f.missingFiles[fileName] {
					f.Log.Warnf("File %q not found", fileName)
					f.missingFiles[fileName] = true
				}
				continue
			}
			f.missingFiles[fileName] = false
```

plugins/inputs/filestat/filestat.go
```diff
@@ -85,12 +92,24 @@ func (f *FileStat) Gather(acc telegraf.Accumulator) error {
...
 			if fileInfo == nil {
-				f.Log.Errorf("Unable to get info for file %q, possible permissions issue",
-					fileName)
+				if _, ok := f.filesWithErrors[fileName]; !ok {
+					f.filesWithErrors[fileName] = struct{}{}
+					f.Log.Errorf("Unable to get info for file %q: %v",
+						fileName, err)
+				}
 			} else {
+				delete(f.filesWithErrors, fileName)
```
srebhan, line 112: See above. Use a string/bool map.

srebhan (changes requested):
Please use a `map[string]bool` instead. See the comments in the code. Otherwise looks good!

antifuchs:
Restructured to use string/bool maps, hope that works for you

srebhan (approved):
Looks good to me.

plugins/inputs/filestat/filestat.go
```diff
@@ -35,11 +35,18 @@ type FileStat struct {
 
 	// maps full file paths to globmatch obj
 	globs map[string]*globpath.GlobPath
+
+	// files that were missing - we only log the first time it's not found.
+	missingFiles map[string]bool
+	// files that had an error in Stat - we only log the first error.
+	filesWithErrors map[string]bool
```
ssoroka, line 42: fyi, struct{} uses no memory.
```suggestion
	// files that were missing - we only log the first time it's not found.
	missingFiles map[string]struct{}
	// files that had an error in Stat - we only log the first error.
	filesWithErrors map[string]struct{}
```

ssoroka (approved):
I'm a bit concerned that the memory use of this grows infinitely on directories with a lot of files going in/out. I guess I'm okay merging it, but I think we should come back and fix that up.

plugins/inputs/filestat/filestat.go
antifuchs: Agreed that a `struct{}` map would use less memory, but then again it'll use about as much memory as the number of files configured to be checked, whose config structure already use memory ... it comes out a wash, I feel like.

PR #7319, opened 2020-04-13
Fibaro input: for battery operated devices, add battery level scraping

dynek:
### Required for all PRs:

This simply adds battery level scraping for battery operated devices, in order to be able to alert when a battery runs out of juice.

### DIFF
--- a/plugins/inputs/fibaro/fibaro.go
+++ b/plugins/inputs/fibaro/fibaro.go
@@ -69,11 +69,12 @@ type Devices struct {
 	Type       string `json:"type"`
 	Enabled    bool   `json:"enabled"`
 	Properties struct {
-		Dead   interface{} `json:"dead"`
-		Energy interface{} `json:"energy"`
-		Power  interface{} `json:"power"`
-		Value  interface{} `json:"value"`
-		Value2 interface{} `json:"value2"`
+		BatteryLevel interface{} `json:"batteryLevel"`
+		Dead         interface{} `json:"dead"`
+		Energy       interface{} `json:"energy"`
+		Power        interface{} `json:"power"`
+		Value        interface{} `json:"value"`
+		Value2       interface{} `json:"value2"`
 	} `json:"properties"`
 }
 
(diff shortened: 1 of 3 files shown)


plugins/inputs/fibaro/fibaro.go
```diff
@@ -69,11 +69,12 @@ type Devices struct {
 	Type       string `json:"type"`
 	Enabled    bool   `json:"enabled"`
 	Properties struct {
-		Dead   interface{} `json:"dead"`
-		Energy interface{} `json:"energy"`
-		Power  interface{} `json:"power"`
-		Value  interface{} `json:"value"`
-		Value2 interface{} `json:"value2"`
+		BatteryLevel interface{} `json:"batteryLevel"`
```
ssoroka, line 72: I realize you're just continuing the pattern here, but it's unclear to me why these are interface{} and not floats, bools, etc. Have any insight?

ssoroka:
Looks good. Thanks for the PR!

plugins/inputs/fibaro/fibaro.go
```diff
@@ -174,6 +175,12 @@ func (f *Fibaro) Gather(acc telegraf.Accumulator) error {
 		}
 		fields := make(map[string]interface{})
 
+		if device.Properties.BatteryLevel != nil {
+			if fValue, err := strconv.ParseFloat(device.Properties.BatteryLevel.(string), 64); err == nil {
```
ssoroka, line 179: Some post-merge review here: are we sure that `device.Properties.BatteryLevel` is always a string? this will panic if not. Also: if it's always a string, why not just type it as a string? We may be making more problems for ourselves by using interface{}

dynek: The value is always quoted in the payload.
As said in my other comment, the JSON payload is dynamic and batteryLevel doesn't exist as a property for all devices but only for those which are battery operated.
How would you recommend to manage this case?

ssoroka: You should still be able to decode a json object with missing keys into the struct. it'll just default to the zero value. eg https://play.golang.org/p/me0tDzHlQc_N

dynek: OK but then if it defaults to zero, how do I differentiate zero from non existing in json payload? Because a not operated battery device doesn't have this info, it should not make it into the DB.

ssoroka: In this case the "zero" value for a string is just an empty string, `""`, which you could still check for... but I see what you're getting at. If you don't want the fields reported when they're not there, you need some way to distinguish a nil from a zero value. 👍

danielnelson: One way to handle this is to have the struct field be a pointer: `BatteryLevel *string`.  The value will be nil if not set in the document.

dynek: Got it, thank you: https://github.com/influxdata/telegraf/pull/7411

PR #7333, opened 2020-04-14
Run create database query once per database

### LINKED ISSUE #7235: InfluxDB Output repeats create database indefinitely if it fails to create it
```toml
[outputs.influxdb]
urls = ["https://influxdburl:8086"]
database = "metrics"
username = "DevUser"
password = "SuperSecret"
database_tag = "influx_database"
exclude_database_tag = true
precision = "s"
skip_database_creation = true
```

### Expected behaviour:

Telegraf should check if it needs to create the database that it is about to write to, rather than blindly creating the database.

### Actual behaviour:

When running telegraf with the influxdb output and the "database_tag" option, telegraf will try to create the database on every flush interval. If it fails to create that database it will keep trying forever and log the failure excessively. The write itself works but there is no context in the code between the two processes.

### Additional info:

This is an example of the log that we get

```text
telegraf 2020-03-26T10:57:26Z W! [outputs.influxdb] When writing to [https://influxdburl:8086]: database "metrics" creation failed: 403 Forbidden
```

danielnelson:
Additional create database queries only ran if a database not found
error is received.

closes #7235

### Required for all PRs:

### DIFF
--- a/plugins/outputs/influxdb/http.go
+++ b/plugins/outputs/influxdb/http.go
@@ -107,9 +107,9 @@ type HTTPConfig struct {
 }
 
 type httpClient struct {
-	client           *http.Client
-	config           HTTPConfig
-	createdDatabases map[string]bool
+	client                 *http.Client
+	config                 HTTPConfig
+	createDatabaseExecuted map[string]bool
 
 	log telegraf.Logger
 }
@@ -177,9 +177,9 @@ func NewHTTPClient(config HTTPConfig) (*httpClient, error) {
 			Timeout:   config.Timeout,
 			Transport: transport,
 		},
-		createdDatabases: make(map[string]bool),
-		config:           config,
-		log:              config.Log,
+		createDatabaseExecuted: make(map[string]bool),
+		config:                 config,
+		log:                    config.Log,
 	}
 	return client, nil
 }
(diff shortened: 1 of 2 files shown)


plugins/outputs/influxdb/http.go
```diff
@@ -209,13 +209,19 @@ func (c *httpClient) CreateDatabase(ctx context.Context, database string) error
 	}
 	defer resp.Body.Close()
 
+	// Track that a 'create database` statement was executed for the database
+	// even if the statement fails.  One initial attempt to create the database
+	// is made each time a new database is encountered in the database_tag.
+	// After this initial attempt it will only be executed if a "database not
+	// found" error occurs.
+	c.createDatabaseExecuted[database] = true
```
ssoroka, line 217: seems like we should only do this on success or permission errors. There's some errors communicating with the database that shouldn't trigger it to stop trying.

plugins/outputs/influxdb/http.go
```diff
@@ -231,11 +228,19 @@ func (c *httpClient) CreateDatabase(ctx context.Context, database string) error
...
+	// Even with a 200 status code there can be an error in the response body.
+	// If there is also no error string then the operation was successful.
 	if resp.StatusCode == http.StatusOK && queryResp.Error() == "" {
+		c.createDatabaseExecuted[database] = true
 		return nil
 	}
 
+	// Don't attempt to recreate the database after a 403 Forbidden error.
+	// This behavior exists only to maintain backwards compatiblity.
+	if resp.StatusCode == http.StatusForbidden {
```
ssoroka, line 240: shouldn't this be the other way around, we only set true if it's the specific result code that means we already have a database? Do we know what the actual code is? `400 Bad Request` or something else?

danielnelson: I think it is correct.  It might help if I mention the current behavior:
- When writing to a single statically configured database, `create database` was called once and regardless of the error returned never again unless a `database not found` error is received during a write.
- When writing to dynamically configured databases via `database_tag`, `create database` was called during every write until it succeeded at least once.  If a `database not found` error is received it is also retried.

With this PR both paths use the same method:
- `Create database` is called once per database, if it fails with any error other than forbidden it will be retried the next write.  If a `database not found` error is returned it will also be retried.

danielnelson: We only know the database exists if we receive a 200 status.

ssoroka: Okay, I would have probably not tried creating the database unless I specifically received a `database not found`, but it sounds like this is closer to the original behavior. Will resolve and approve.

PR #7335, opened 2020-04-15
Use new higher limit for cloudwatch GetMetricData

### LINKED ISSUE #7322: Use new AWS Cloudwatch GetMetricData metric limit of 500 metrics per call (was 100)
## Feature Request

Please change the Cloudwatch API plugin to batch requests using the new 500 metrics per call limit (https://aws.amazon.com/about-aws/whats-new/2020/03/amazon-cloudwatch-now-provides-more-metric-data-faster-with-getmetricdata-quota-increase/). This should significantly reduce the load and the runtimes (and potentially cost?) when polling the Cloudwatch API.

We are struggling with scaling Telegraf to our needs due to latency and API request volume and heavily rely on this improvement.

FYI we have also filed a PFR with AWS to allow the ListMetrics API call to restrict the time period for which it returns resources & their metrics which will reduce the amount of API calls required for this single threaded (pagination) call. 

### Proposal:

Adopt GetMetricData to use new limit of 500 metrics per call instead of 100.

### Current behavior:

Batches of 100 metrics per call are used.

### Desired behavior:

Each GetMetricData call should request a batch of 500 metrics instead of the current 100.

### Use case: 

We are trying to scale Telegraf to collect metrics from various regions around the world and for  hundreds of thousands of instances and metrics. We run into constraints of the API calls taking too long due to latency & volume for some accounts. When rate limiting GetMetricData calls we run the risk of not finishing the metric collection within our period we use to provide the business the granularity they require. When increasing the rate limit we hit account's API call limits due to multiple namespaces we collect from. This is a very undesirable situation and won't work as our accounts grow.

danielnelson:
closes #7322

### Required for all PRs:

### DIFF
--- a/plugins/inputs/cloudwatch/cloudwatch.go
+++ b/plugins/inputs/cloudwatch/cloudwatch.go
@@ -212,7 +212,7 @@ func (c *CloudWatch) Gather(acc telegraf.Accumulator) error {
 	results := []*cloudwatch.MetricDataResult{}
 
 	// 100 is the maximum number of metric data queries a `GetMetricData` request can contain.
-	batchSize := 100
+	batchSize := 500
 	var batches [][]*cloudwatch.MetricDataQuery
 
 	for batchSize < len(queries) {


danielnelson:
- [telegraf-1.15.0~b19fa8a5-0.x86_64.rpm](https://52461-33258973-gh.circle-artifacts.com/0/build/linux/amd64/telegraf-1.15.0~b19fa8a5-0.x86_64.rpm)
- [telegraf_1.15.0~b19fa8a5-0_amd64.deb](https://52461-33258973-gh.circle-artifacts.com/0/build/linux/amd64/telegraf_1.15.0~b19fa8a5-0_amd64.deb)
- [telegraf-1.15.0~b19fa8a5_windows_amd64.zip](https://52461-33258973-gh.circle-artifacts.com/0/build/windows/amd64/telegraf-1.15.0~b19fa8a5_windows_amd64.zip)

Issue #7338, opened 2020-04-15
Agent won't start if remote HTTP config endpoint is down


schmorgs:
## Feature Request
A useful feature of Telegraf is to pull config from a remote HTTP server, thus centralising agent configuration.  However, if the HTTP endpoint is unavailable when the agent is started, the agent throws an error and dies.  

### Proposal:
Have the agent retry the endpoint periodically if unavailable, and when online, pull the config and start as normal.

### Current behavior:
Agent will not start.

### Desired behavior:
Agent re-attempts to get the configuration, ideally driven by startup switches to control retry-interval and max retry-attempts

### Use case: 
In a large environment, it is entirely possible that the HTTP endpoint could go offline (e.g. planned or unplanned outage).  Any agents that get restarted during this interval (e.g. reboot, agent installation) would simply die, meaning some recovery process is needed to bring them back online.  
Having this retry ability in the agent would provide a more reliable solution and self-recovery.

schmorgs:
Pull request https://github.com/influxdata/telegraf/pull/7349 submitted to support this with a new CLI flag to control how ofter to retry the polling interval

sjwang90:
Adding some notes on if we implement this some behavior we would want to go about it:
- By default, set the retries to 0 - to preserve the current behavior
- Log attempts and the specific failure response codes.

sjwang90:
Related #3723

schmorgs:
Hi, I see this has been closed with a fixed retry of 3 times, but this doesn't really help much for my use case.  Basically, the agent has 30 seconds to get the config from the remote server or it dies and we need some level of recovery to get it back online which isn't great across 1000's servers.

So unfortunately, not a fix for me.

ssoroka:
Hey Schmorgs. I'd like to hear more about your use case. I am a bit concerned about Telegraf failing quietly where everything looks okay, because it's not loudly failing, but in reality it might never come up (say url was wrong or dns issue).
How do you manage your servers? doesn't Telegraf restart if it's not available? Why isn't the config mostly available? This seems like it should be a really rare problem. Is this an IoT situation where the devices are frequently offline?

I'm not against reopening the issue, though we may want to rename it to be more specific about your needs.

sjwang90:
Opened #8854 to cover the configuring behavior settings portion of this issue.

schmorgs:
Hi, thanks @sjwang90 for opening the new issue.

[...]

Issue #3723, opened 2018-01-26
Allow agent to start when input or output cannot be connected to


kelein:
I have installed telegraf **v1.4.4** via rpm and configured a input for kafka_consumer as follows:
```
## Read metrics from Kafka topic(s)
[[inputs.kafka_consumer]]
  brokers = ["localhost:9092"]
  topics = ["telegraf"]
```
It works well for gathering kafka metrics. Unfortunately, when the kafka broker is down abnormally,  It is failed to restart telegraf. The telegraf log hints this: 

```
$ service telegraf status
 
Redirecting to /bin/systemctl status telegraf.service
● telegraf.service - The plugin-driven server agent for reporting metrics into InfluxDB
   Loaded: loaded (/usr/lib/systemd/system/telegraf.service; enabled; vendor preset: disabled)
   Active: inactive (dead) since Fri 2018-01-26 20:18:16 CST; 45min ago
     Docs: https://github.com/influxdata/telegraf
  Process: 8954 ExecStart=/usr/bin/telegraf -config /etc/telegraf/telegraf.conf -config-directory /etc/telegraf/telegraf.d $TELEGRAF_OPTS (code=exited, status=0/SUCCESS)
 Main PID: 8954 (code=exited, status=0/SUCCESS)
```
> /var/log/telegraf/telegraf.log
```
2018-01-26T10:33:56Z I! Starting Telegraf v1.4.4
2018-01-26T10:33:56Z I! Loaded outputs: influxdb opentsdb prometheus_client
2018-01-26T10:33:56Z I! Loaded inputs: inputs.kernel inputs.system inputs.filestat inputs.mongodb inputs.kafka_consumer inputs.diskio inputs.swap inputs.disk inputs.docker inputs.internal inputs.kernel_vmstat inputs.cpu inputs.mem inputs.processes inputs.net inputs.zookeeper inputs.logparser
2018-01-26T10:33:56Z I! Tags enabled: host=127.0.0.1 user=telegraf
2018-01-26T10:33:56Z I! Agent Config: Interval:10s, Quiet:false, Flush Interval:10s 
2018-01-26T10:33:57Z E! Error when creating Kafka Consumer, brokers: [localhost:9092], topics: [telegraf]
2018-01-26T10:33:57Z E! Service for input inputs.kafka_consumer failed to start, exiting
kafka: client has run out of available brokers to talk to (Is your cluster reachable?)
```

### Expected behavior:

[...]

PR #7349, opened 2020-04-16
feature-7338 Retry failed remote HTTP Config on startup

schmorgs:
### Required for all PRs:

- [Y] Signed [CLA](https://influxdata.com/community/cla/).
- [Y] Associated README.md updated.
- [Y] Has appropriate unit tests.

Retry remote HTTP config pull on agent startup to prevent agent stopping if HTTP endpoint is down
https://github.com/influxdata/telegraf/issues/7338

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -3,6 +3,7 @@ package config
 import (
 	"bytes"
 	"errors"
+	"flag"
 	"fmt"
 	"io/ioutil"
 	"log"
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -851,26 +851,34 @@ func loadConfig(config string) ([]byte, error) {
 }
 
 func fetchConfig(u *url.URL) ([]byte, error) {
-	req, err := http.NewRequest("GET", u.String(), nil)
+	// Read the startup flag for config-http-retry and try to convert to integer
+	configHttpRetry, err := strconv.Atoi(flag.Lookup("config-http-retry").Value.String())
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to process config-http-retry ")
 	}
+	// Loop to retry http config request to prevent agent dying on startup if endpoint is unavailable
+	for {
+		req, err := http.NewRequest("GET", u.String(), nil)
+		if err != nil {
+			return nil, err
+		}
 
-	if v, exists := os.LookupEnv("INFLUX_TOKEN"); exists {
-		req.Header.Add("Authorization", "Token "+v)
+		if v, exists := os.LookupEnv("INFLUX_TOKEN"); exists {
+			req.Header.Add("Authorization", "Token "+v)
+		}
+		req.Header.Add("Accept", "application/toml")
+		resp, err := http.DefaultClient.Do(req)
+		if err != nil {
+			log.Printf("Unable to connect to HTTP config server.  Retry in %d seconds", configHttpRetry)
+		} else if resp.StatusCode != http.StatusOK {
+			log.Printf("Error getting HTTP config, retry in %d seconds.  Status=%s", configHttpRetry, resp.Status)
+		} else {
+			log.Printf("Successful retrieval of config from HTTP endpoint.")
+			defer resp.Body.Close()
+			return ioutil.ReadAll(resp.Body)
+		}
+		time.Sleep(time.Duration(configHttpRetry) * time.Second)
 	}
-	req.Header.Add("Accept", "application/toml")
-	resp, err := http.DefaultClient.Do(req)
-	if err != nil {
-		return nil, err
-	}
-
-	if resp.StatusCode != http.StatusOK {
-		return nil, fmt.Errorf("failed to retrieve remote config: %s", resp.Status)
-	}
-
-	defer resp.Body.Close()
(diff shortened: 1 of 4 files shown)


danielnelson:
I'd prefer to avoid adding a new cli option specifically for this, but perhaps this could be part of a solution to #3723.  @ssoroka I know you working in this area of the code, do you think it would make sense to tackle this at the same time?

schmorgs:
Hi, 

Thanks for the update.

 

I took a look at 3723 and I can’t tell if it’s the same area, but it does look like in that case, the agent has already received/parsed it’s config phase, whereas here, the agent still hasn’t started.

 

Either way, I’m also happy to not have the dedicated CLI and use a hard-coded (say) 10 minute retry if the HTTP server is down if that makes it easier.

 

 

Thanks, Trevor

ssoroka:
I think the hope is to move the retry logic out of the plugins so that you can decide if they should retry at the agent level, and then get consistent behavior across all of them. I guess the one thing that would get a bit weird here is that the output plugin's `Connect` method would get called more than once, which it currently does not. That may expose some new bugs with plugins that don't expect this. That's probably still the way to go, though.

schmorgs:
As an interim and quick solution, could we put the retry logic around the HTTP config request so that it loops / sleeps until the web server is up?  

Can use a fixed retry period as it's better than the agent failing completely.

internal/config/config.go
```diff
@@ -851,26 +852,34 @@ func loadConfig(config string) ([]byte, error) {
...
+		if v, exists := os.LookupEnv("INFLUX_TOKEN"); exists {
+			req.Header.Add("Authorization", "Token "+v)
+		}
+		req.Header.Add("Accept", "application/toml")
+		resp, err := http.DefaultClient.Do(req)
+		if err != nil {
+			log.Printf("Unable to connect to HTTP config server.  Retry in %d seconds", configHttpRetry)
+		} else if resp.StatusCode != http.StatusOK {
+			log.Printf("Error getting HTTP config, retry in %d seconds.  Status=%s", configHttpRetry, resp.Status)
+		} else {
```
ssoroka, line 876: remove the use of else blocks. It indents the main-line code path and makes it harder to read. Use `continue` instead to restart the for loop.

internal/config/config.go
```diff
@@ -851,26 +852,34 @@ func loadConfig(config string) ([]byte, error) {
...
-		req.Header.Add("Authorization", "Token "+v)
-	}
-	req.Header.Add("Accept", "application/toml")
-	resp, err := http.DefaultClient.Do(req)
-	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to process config-http-retry ")
 	}
+	// Loop to retry http config request to prevent agent dying on startup if endpoint is unavailable
+	for {
```
ssoroka, line 861: should probably support max-retries before permanently failing. Not a big fan of looping forever.

ssoroka:
I think I misunderstood what you were trying to accomplish. I read it over again and left some feedback. I'd be in support of getting this in with a couple small changes.

internal/config/config.go
schmorgs: Sure, I did propose a CLI flag to set max retries further up the chain, but understand it adds to the mass of options.

Personally, I would like it to return forever, otherwise someone it means some manual intervention is required, but others may not so think an option would be useful to drive desired behaviour.

Would using an environment variable setting be an alternative approach to a CLI switch here?
e.g. export TELEGRAF_HTTP_RETRIES=50
It not set, then we just use a default of (say) 10 iterations every 1 minute.

If so, then I'll take a look at building this in an resubmitting for review.

schmorgs:
Thanks for the update Daniel, the future state sounds a much better approach.

Just to reiterate the need for retries.
If (as will happen) the HTTP server is down for some unplanned and prolonged reason, we don't want any Telegraf agents which start up during that time to fail as it means someone or something needs to intervene and take action,
So whilst a fixed x retries @ y seconds would help, in my environment, I'd want it to keep retrying and auto-recover itself when the endpoint is back online.

If other users of Telegraf want a shorter and fixed retry period, the pull request would support that also.

Could you consider merging my code (or a form of it) with the option to specify retry intervals and attempts through the Env variables as it provides the flexibility to meet my own use case and others too.

danielnelson:
Not with the configurable options since these cannot be removed, and we wouldn't need them as part of the future design.  Setting up the init system to restart should be a decent workaround for now, or you can always build a custom version with whatever you need.

schmorgs:
Ok, understand the approach and how this could build in some legacy which is difficult to remove. 
Will look into the systemd restart approach.  On Windows however, I guess SCM would be the workaround there but there are no options to set Recovery action when installed Telegraf as a service.  I realise we could hack the registry after installing as a service, but wanted to check if there are any options in the built-in Telegraf service install option to set recovery actions?

danielnelson:
One tool that could be helpful is https://nssm.cc/, I know manyWindows  Telegraf users prefer to setup the service with this tool.

sjwang90:
Hey @schmorgs - based off our conversation I'm going to close this PR for our tracking purposes. I've kept https://github.com/influxdata/telegraf/issues/8854 open to keep track of this feature as we build out our larger configuration management capabilities.

PR #7350, opened 2020-04-16
Sql server - sql 2008 SP4 and 2008 R2 compatibility

### LINKED ISSUE #7292: Sql Server - Support for SQL Server 2008
## Feature Request

Edit the v2 queries to make them compatible with older editions of SQL server, like SQL 2008 and 2008 R2.
Some queries are already compatible with SQL Server 2008 R2.

Of course, it's not possible to achieve full compatibility and some fields will be missing but lots of information can be gathered.

**Edit:**
The initial idea was to make retro compatibility up to SQL Server 2005, which is doable for the queries themselves, but it's not for the plugin, in fact the "go-mssqldb" library has some limitations.
It supports:
- SQL Server 2008 SP3 with CU pack 3 (and later, latest version is SQL 2008 SP4)
- SQL Server 2008 R2 SP2
Any previous version will return an error when trying to connect.

### Proposal:
Make the queries more compatible
### Current behavior:
Most query won't run on version pre SQL 2008 R2
### Desired behavior:
being able to monitor SQL Server 2008 and 2008 R2
### Use case: 
Monitor the older version of SQL Server, because you can't just throw old instances in the garbage.
Even if you won't have the same data a SQL 2012+ can offer, something is better than nothing.

Trovalo:
### Required for all PRs:

closes #7292

This PR adds compatibility for all the queries form SQL Server 2008 SP4 and later version.
Main changes:
- Most of the SQL code has been formatted
- The switch logic used to compose the queries is now based on several variables (MajorVersion, MinorVersion, BuildVersion, RevisionVers, EngineEdition) instead of being calculated every time 
- Several switches have been added to manage SQL 2008 and 2008 R2
- Minor changes to the query to make them compatible (ie: database_id is taken form table "a" instead of table "b")

It has been tested on:
- SQL 2008 SP4
- SQL 2008 R2
- SQL 2012
- SQL 2019
(tests on SQL 2014, 2016, 2016 SP2 and 2017 will be made soon)

Additional Fix:
In the current version of telegraf, in the "performance counters" query, the column [total_cpu_usage_preemptive_ms] is never considered, it is added as follows:
```SQL
CASE WHEN SERVERPROPERTY('ProductMajorVersion') > 10 THEN 'rgwg.total_cpu_usage_preemptive_ms

-- Sadly that switch does not work unless you also cast SERVERPROPERTY to int, run this to check
SELECT IIF( SERVERPROPERTY('ProductMajorVersion') > 10 ,'ok','not ok')
```
Now that the switches are based on variables (with proper data type) the metric will be loaded for SQL Server 2016 and later, before it does not exists


Trovalo:
@denzilribeiro I know you don't use/monitor the on-prem version of SQL Server but can you have a quick look at this?
Azure SQL Db and Managed instance have not been touched in this PR.

PR #7352, opened 2020-04-16
Deprecate logparser input and recommend tail input as replacement

danielnelson:
The tail plugin provides all the functionality of the `logparser` plugin along with additional features.  This change deprecates logparser to avoid confusion about which to use and save maintenance time.

### Required for all PRs:


FramnkRulez:
How does 'tail' support grok patterns?  I don't see anything about it in the documentation for it...

PR #7353, opened 2020-04-16
Document distinction between file and tail inputs

danielnelson:
### Required for all PRs:

### DIFF
--- a/plugins/inputs/file/README.md
+++ b/plugins/inputs/file/README.md
@@ -1,10 +1,10 @@
 # File Input Plugin
 
 The file plugin updates a list of files every interval and parses the contents
-using the selected [input data format](/docs/DATA_FORMATS_INPUT.md).
+using the selected [input data format][].
 
-Files will always be read in their entirety, if you wish to tail/follow a file
-use the [tail input plugin](/plugins/inputs/tail) instead.
+Files will always be read in their entirety, if you wish to parse only newly
+appended lines use the [tail input plugin][] instead.
 
 ### Configuration:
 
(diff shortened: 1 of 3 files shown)


plugins/inputs/tail/tail.go
```diff
@@ -100,7 +100,7 @@ func (t *Tail) SampleConfig() string {
 }
 
 func (t *Tail) Description() string {
-	return "Stream a log file, like the tail -f command"
+	return "Parse the new lines appended to a log file"
```
reimda, line 103: Let's mention the interval like we do for the file input.  Also I wonder if leaving out "log file" would help too.  When I hear that I think /var/log/messages and I don't think of a growing csv or line protocol or json file.

How about "Parse the lines appended to a file since the last interval".

If you agree avoiding "log file" is right, we need to remove it from the readme too.

danielnelson: Agree about the "log file", we don't want people to think of this plugin is only for traditional logs.  I have some reservations about mentioning interval as this plugin doesn't use the interval.  How about: "Parse the newly appended lines to a file"?

plugins/inputs/file/README.md
```diff
@@ -1,16 +1,17 @@
 # File Input Plugin
 
-The file plugin updates a list of files every interval and parses the contents
-using the selected [input data format](/docs/DATA_FORMATS_INPUT.md).
+The file plugin parses the complete contents of a file each interval using the
+selected [input data format][].
 
-Files will always be read in their entirety, if you wish to tail/follow a file
-use the [tail input plugin](/plugins/inputs/tail) instead.
+**Note:** If you wish to parse only newly appended lines use the [tail][] input
```
ssoroka, line 6: > only newly appended lines
Not just newly appended lines, but if you only want to parse the file _once_, you should use tail.

ssoroka: said another way, if you don't want the file to be parsed from the beginning every interval, you should use tail

PR #7014, opened 2020-02-12
Collect sys.dm_db_resource_stats with Azure Managed Instance

Eivissa:
### Required for all PRs:

- [X] Signed [CLA](https://influxdata.com/community/cla/).
- [X] Associated README.md updated.
- [X] Has appropriate unit tests.

PR #7359, opened 2020-04-17
Sql server - ring buffer cpu v2

### LINKED ISSUE #7269: SQL Server - Missing CPU data for Standard Edition
## Feature Request

Find the best way to get CPU usage information and add them to an existing measurement or on a new one. 
In the v1 queries, this has a measurement on his own,  called 'CPU (%)', code [here](https://github.com/influxdata/telegraf/blob/c083c1f64d05d72bdacfe03847f2831d224a4ee0/plugins/inputs/sqlserver/sqlserver.go#L2349).

As of now, CPU data are present only for enterprise and developer edition of SQL Server, the idea is to get it also for Standard and Express editions.
### Proposal:
For what I've found so far the best way is to get the data from the ring buffer (table sys.os_dm_ring_buffer) as is done in the queries v1.

### Current behavior:
As of now, in the measurement 'sqlserver_performance' you can find the counter 'CPU usage %', with a record for each application pool and workload group. If no UD (user-defined) resource pool has been configured you will find "internal" and "default"
Resource groups:
- **internal** - The internal pool represents the resources consumed by the SQL Server itself. (meaning resource used by SQL Server itself to work)
- **default** and UD - gathers data about workloads (ie: consumption of queries sent by an application)

**The Problem**
In order to track default and UD resource pools, _Resource Governor_ must be enabled on the instance, but it is available only on the enterprise edition of SQL Server (and also Developer edition).

This means that CPU data can be tracked only on Enterprise and Developer edition of SQL Server (and not for Standard and Express).
### Desired behavior:

Having CPU data for regardless of the SQL Server edition.

### Use case:
This is crucial information for monitoring, missing it for some edition of SQL Server is not nice

Trovalo:
### Required for all PRs:

Closes #7269

This PR adds a new measurement called "sqlserver_cpu" with the following structure:
| measurement | sql_instance | sqlserver_process_cpu | system_idle_cpu | other_process_cpu |
| ------ | ------ | ------ | ------ | ------ |
| sqlserver_cpu | QDLP03:SQL2019 | 0 | 95 | 5 |

It reads the CPU data from the ring buffer, therefore it allows to get CPU data from any version of SQL Server (including express and standard).
In the source, the data are updated once every minute.

As of now, it works only on the On-prem version of SQL Server (from SQL 2008 to 2019) but I'm looking to make it work also on Azure SQL DB and managed Instance. (maybe using sys.dm_db_resource_stats or sys.server_resource_stats)

### DIFF
--- a/plugins/inputs/sqlserver/README.md
+++ b/plugins/inputs/sqlserver/README.md
@@ -65,6 +65,7 @@ GO
   ## - Schedulers
   ## - SqlRequests
   ## - VolumeSpace
+  ## - Cpu
   ## Version 1:
   ## - PerformanceCounters
   ## - WaitStatsCategorized
@@ -81,13 +82,13 @@ GO
   # include_query = []
 
   ## A list of queries to explicitly ignore.
-  exclude_query = [ 'Schedulers' , 'SqlRequests']
+  exclude_query = [ 'Schedulers' , 'SqlRequests' ]
 ```
 
 ### Metrics:
 To provide backwards compatibility, this plugin support two versions of metrics queries.
 
-**Note**: Version 2 queries are not backwards compatible with the old queries. Any dashboards or queries based on the old query format will not work with the new format. The version 2 queries are written in such a way as to only gather SQL specific metrics (no overall CPU related metrics) and they only report raw metrics, no math has been done to calculate deltas. To graph this data you must calculate deltas in your dashboarding software.
+**Note**: Version 2 queries are not backwards compatible with the old queries. Any dashboards or queries based on the old query format will not work with the new format. The version 2 queries only report raw metrics, no math has been done to calculate deltas. To graph this data you must calculate deltas in your dashboarding software.
 
 #### Version 1 (deprecated in 1.6):
 The original metrics queries provide:
@@ -124,6 +125,7 @@ The new (version 2) metrics provide:
   dm_exec_sessions that gives you running requests as well as wait types and
   blocking sessions.
 - *VolumeSpace* - uses sys.dm_os_volume_stats to get total, used and occupied space on every disk that contains a data or log file. (Note that even if enabled it won't get any data from Azure SQL Database or SQL Managed Instance). It is pointless to run this with high frequency (ie: every 10s), but it won't cause any problem.
+- *Cpu* - uses the buffer ring (sys.dm_os_ring_buffers) to get CPU data, the table is updated once per minute. (Note that even if enabled it won't get any data from Azure SQL Database or SQL Managed Instance).
 
   In order to allow tracking on a per statement basis this query produces a
   unique tag for each query.  Depending on the database workload, this may
--- a/plugins/inputs/sqlserver/sqlserver.go
+++ b/plugins/inputs/sqlserver/sqlserver.go
@@ -1596,17 +1596,7 @@ IF SERVERPROPERTY('EngineEdition') NOT IN (5,8)
 
 const sqlServerCpuV2 string = `
 /*The ring buffer has a new value every minute*/
-DECLARE
-	 @EngineEdition AS int
-
-SELECT 
-	 @EngineEdition = y.[EngineEdition]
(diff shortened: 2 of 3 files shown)


Trovalo:
@denzilribeiro I would like to hear your opinion about it, especially for making it compatible with Azure SQL DB and Managed Instance. (I have a SQL DB, but not a Managed Instance).
Is sys.dm_db_resource_stats a good way to get those data? do you have any suggestions?

denzilribeiro:
I would not use this for the cloud - there are better mechanisms. Even for VM's the better approach is to collect OS level metrics right?

For managed instance need to add sys.resource_stats, for SQL DB sys.dm_db_resource_stats is already being collected. I need to spend some time to make couple changes in this area add this as well.
https://docs.microsoft.com/en-us/sql/relational-databases/system-catalog-views/sys-server-resource-stats-azure-sql-database?view=azuresqldb-current . I meant to submit a PR and forgot about it had commented here -  https://github.com/influxdata/telegraf/pull/7014

Trovalo:
I know that getting the data directly from the VM OS is probably the best option, but if you are monitoring a remote instance of SQL Server,  it is not possible to use the windows plugin to gather data. In this kind of scenario, the ring buffer data are a good alternative.

Also I'm not sure about what the windows plugin can get "out of the box", can it get Cpu data at process level? (I will check it myself)

denzilribeiro:
Note you can also get CPU from perfmon ( workload group counters have CPU) but won't get instance level CPU from there. At some point have to make sure we are not spawning too many connections for collection :) the balance.

Trovalo:
I already get the "CPU %" from the perfcounters measurement (in Resource Pool and Workload Groups), but sadly the Resource governor is available only in Enterprise and developer edition. Therefore, as of now, if you have a Standard or Express edition of SQL Server you can't get Cpu data using the SQL server plugin alone.

I also thought about adding this as a row in the perfcounters measurement, something like Counter = "CPU %", Instance = "_Total". But I didn't like that since it is not an actual perfcounter.

Trovalo:
In general, I think that the SQL Server plugin alone should allow you to fully monitor your SQL Server instances, in this specific case the CPU. 
Having this data in its own measurement/query allows you to be flexible, if you get the CPU from Windows perfmon then that's perfect, just disable the 'Cpu' query in order to not fetch the same info twice.

That said, I agree with you about keeping the plugin as light as possible.
A way to achieve this is the usage of custom errors in case the query is not supported on the target environment. (this can be extended to every query)

i.e. if the 'Cpu' query is used on Azure SQL DB or Managed instance we can execute
> THROW 50001, 'Telegraf "Cpu" query is not supported on Azure SQL DB and Managed Instance', 1;

This might be a somehow breaking change, but it will ensure that people keep their config clean.
Let me know what you think about it

denzilribeiro:
Agree on having only targeted queries for the platform in question.When I find time will find a way to actually segregate sections, had a discussion with Daniel as well.. will write up a note before I do that. Effectively can have 3 sections in conf one for each platform ( SQL Server, Azure SQL DB, Azure SQL Managed instance) and a bit different flag rather than just AzureDB = true or false... Will work on that hopefully soon.

Trovalo:
There is no problem in simplifying the loop by removing the variable declaration, cast and assignment. It will work as long as the comparison is "IN" or "=", but it won't if we use ">" or "<" (in this case no problem).

I did that to match and somehow standardize some changes I have in another open PR about compatibility up to SQL Server 2008 SP4, in which I used the variables to avoid repeating the SERVERPROPERTY(...) and CAST(...) combo.

About the segregation of the different platforms, I will also think about some nice ways to achieve it and then share it with you.
If you already have some ideas can you open a new feature request in order to keep the conversation there?

Trovalo:
I've updated the readme, now everything should be ok. @denzilribeiro can you check?

Trovalo:
Hi @denzilribeiro, any update on this?

PR #7363, opened 2020-04-17
systemd-timings: add systemd boot and unit timing input plugin

### LINKED ISSUE #7010: systemd-analyze like input plugin
## Feature Request

Hello, you might be aware of systemd-analyze (https://www.freedesktop.org/software/systemd/man/systemd-analyze.html).  This is a useful systemd tool which uses dbus to return system startup timing related metrics, these metrics are then pretty printed on the terminal.  We would like to be able to capture such metrics (read via dbus) using telegraf.

### Proposal:

A new input plugin could be written to gather metrics via dbus.  I spent a few hours hacking on this yesterday and was able to at least gather basic metrics using telegraf with the --test argument.  I would like to work further on this implementation and contribute it upstream if possible.

Note the gathering of these metrics only need to be collected once after the system is booted, I'm not sure if there is support for such a "one shot" style metric collection currently in telegraf.  Thoughts on this (and infact the applicability of telegraf to this problem) would be much appreciated.

### Current behavior:

I'm not aware of a way to gather such metrics at present.

### Desired behavior:

It would be nice to be able to gather all systemd startup timing related metrics and have them displayed using our of the output plugins.

### Use case: 

We have a large fleet of nodes where system start up time and restart time via kexec is very important.  We'd like to be able to visualize/analyze this using telegraf so we can further drill down on startup related performance issues.

Thanks!

pdmorrow:
### Required for all PRs:

This input plugin collects systemd boot and unit related timestamps. Timestamps are accessed via systemd exposed APIs on D-Bus.  There are 2 classes of timestamps collected:

- System wide: These are timestamps pertaining to the overall boot process.
- Per unit: These timestamps indicate the time taken to start or stop a unit.

System wide timestamps are collected and sent once after telegraf is started.  Per unit timestamps are sent whenever telegraf is started and also later if a unit is started or stopped.  All timestamps are at a microsecond resolution.

For more information see README.md.

closes #7010

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -31,6 +31,7 @@ require (
 	github.com/cenkalti/backoff v2.0.0+incompatible // indirect
 	github.com/cisco-ie/nx-telemetry-proto v0.0.0-20190531143454-82441e232cf6
 	github.com/cockroachdb/apd v1.1.0 // indirect
+	github.com/coreos/go-systemd/v22 v22.0.0
 	github.com/couchbase/go-couchbase v0.0.0-20180501122049-16db1f1fe037
 	github.com/couchbase/gomemcached v0.0.0-20180502221210-0da75df14530 // indirect
 	github.com/couchbase/goutils v0.0.0-20180530154633-e865a1461c8a // indirect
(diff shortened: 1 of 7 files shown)


sjwang90:
Would you be willing to submit this plugin as an [external plugin](https://github.com/influxdata/telegraf/blob/master/EXTERNAL_PLUGINS.md) that can be used with `execd` to run seamlessly with Telegraf? Telegraf users would be able to use the plugin sooner by having it as an external plugin since it wouldn't have to go through the typical review process. 

Here is some [information on external plugins](https://github.com/influxdata/telegraf/blob/master/docs/EXTERNAL_PLUGINS.md) and using them with `execd`. Feel free to respond with any questions you may have.

pdmorrow:
Sure, I can take a look at doing this.

pdmorrow:
Hi, I opened a PR with this as an execd plugin instead:

https://github.com/influxdata/telegraf/pull/8263

Thanks,
Peter.

PR #7366, opened 2020-04-17
add support for SIGUSR1 to trigger flush

ssoroka:
Closes https://github.com/influxdata/telegraf/issues/5120

### Required for all PRs:

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -8,7 +8,6 @@ import (
 	"os/signal"
 	"runtime"
 	"sync"
-	"syscall"
 	"time"
 
 	"github.com/influxdata/telegraf"
@@ -17,8 +16,6 @@ import (
 	"github.com/influxdata/telegraf/internal/models"
 	"github.com/influxdata/telegraf/plugins/serializers/influx"
 )
-
-const flushSignal = syscall.SIGUSR1
 
 // Agent runs a set of plugins.
 type Agent struct {
@@ -558,9 +555,11 @@ func (a *Agent) flushLoop(
 	}
 
 	// watch for flush requests
-	flushRequested := make(chan os.Signal, 1)
-	signal.Notify(flushRequested, flushSignal)
-	defer signal.Stop(flushRequested)
+	if runtime.GOOS != "windows" {
+		flushRequested := make(chan os.Signal, 1)
+		signal.Notify(flushRequested, flushSignal)
+		defer signal.Stop(flushRequested)
+	}
 
 	// align to round interval
 	if a.Config.Agent.RoundInterval {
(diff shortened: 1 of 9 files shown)


docs/OUTPUTS.md
```diff
@@ -94,6 +94,15 @@ You should also add the following to your `SampleConfig()`:
...
 ```
 
+## Flushing Metrics to Outputs
+
+Metrics are flushed to outputs when any of the following events happen:
+- `flush_interval + rand(flush_jitter)` has elapsed since start or the last flush interval
+- At least `metric_batch_size` count of metrics are waiting in the buffer
+- The telegraf process has received a SIGUSR1 signal
+
+Note that if the flush takes longer than the `agent.interval` to write the metrics to the output, you'll see a message saying the output `did not complete within its flush interval`. This may mean your output is not keeping up with the flow of metrics, and you may want to look into enabling compression, reducing the size of your metrics, or investigate other reasons why the writes might be taking longer than expected.
```
danielnelson, line 104: Please word wrap around 80 chars

agent/agent.go
```diff
@@ -575,8 +586,14 @@ func (a *Agent) flush(
 		}
 
 		select {
+		case <-ctx.Done():
+			logError(a.flushOnce(output, interval, output.Write))
+			return
 		case <-ticker.C:
 			logError(a.flushOnce(output, interval, output.Write))
+		case <-flushRequested:
+			log.Printf("I! [agent] User-requested flush received")
```
danielnelson, line 595: I wish this was logged only once across all outputs, but I'm not sure it would be worth it.

cmd/telegraf/telegraf.go
```diff
@@ -36,7 +34,7 @@ var pprofAddr = flag.String("pprof-addr", "",
 	"pprof address to listen on, not activate pprof if empty")
 var fQuiet = flag.Bool("quiet", false,
 	"run in quiet mode")
-var fTest = flag.Bool("test", false, "enable test mode: gather metrics, print them out, and exit")
+var fTest = flag.Bool("test", false, "enable test mode: gather metrics, print them out, and exit. Note: Test mode only runs inputs, not processors, aggregators, or outputs")
```
danielnelson, line 37: BTW, I don't think these strings are ever output.  Check `internal/usage.go`, maybe we should merge these usage files with the newly added files.

agent/agent.go
ssoroka: Oh, good point. Alternatively we could not log anything.

cmd/telegraf/telegraf.go
ssoroka: they're supposed to output on --help.

PR #7367, opened 2020-04-17
add another grok example for custom timestamps

ssoroka:


plugins/parsers/grok/README.md
```diff
@@ -160,6 +160,21 @@ Wed Apr 12 13:10:34 PST 2017 value=42
...
+```
+21/02/2017 13:10:34 value=42
+```
+
+```toml
+[[inputs.file]]
+  grok_patterns = ['%{MY_TIMESTAMP:timestamp:ts-"02/01/2006 15:04:05"} value=%{NUMBER:value:int}']
+
+  grok_custom_patterns = '''
+    MY_TIMESTAMP (?:\d{2}.\d{2}.\d{4} \d{2}:\d{2}:\d{2})
```
reimda, line 174: I thought if you provide an example using the reference timestamp you didn't have to specify a custom pattern.  In other words, wouldn't this work?
```toml
[[inputs.file]]
    grok_patterns = ['%{TIMESTAMP:timestamp:ts-"02/01/2006 15:04:05"} value=%{NUMBER:value:int}']
```
(and no "grok_custom_patterns = xyz" line)

ssoroka: Only if the reference TIMESTAMP pattern matches your pattern. I definitely had a case where it did not.

danielnelson: There needs to be two patterns, the first one is a regular expression that grok uses to match the string.  The second pattern is what Go uses to create a time.Time from the matched string.

You always need both, but you can use a built-in pattern if it is defined.  I don't think there is a built-in pattern named TIMESTAMP.

You could also use some intermediate built-in patterns here, you would get a little tighter control over the day/month range:
```toml
grok_custom_patterns = '''
  MY_TIMESTAMP %{MONTHDAY}/%{MONTH}/%{YEAR}:%{TIME}
'''
```

It's also nice to have an example from first principles though.

ssoroka: think those slashes need to be escaped?

danielnelson: In your original pattern no, because you are using a multiline literal string (`'''`).  In my pattern no because they are forward slashes.

ssoroka: I'm surprised \d worked then. :D

ssoroka: good to merge?

ssoroka:
GitHub thinks CI is pending, but it's actually passed and Github hasn't noticed.

plugins/parsers/grok/README.md
reimda: Yep, thanks

Issue #2564, opened 2017-03-24
Create static fields parsing log files


phemmer:
### Proposal:
A way when using the logparser to create static fields that are returned whenever a line matches.

### Use case: [Why is this important (helps with prioritizing requests)]
In our use case, we poll haproxy for statistics every few seconds. One of the things we alert on is when all the servers in a pool are down. However if there is a blip, where we lose all the servers in a pool, but they come back up a few seconds later, our polling interval might not catch the issue. However in this situation, haproxy logs the following message:

    backend foo has no server available!

We would instead like to use the log parser to match this pattern, and emit a point with the field for number of servers in the pool set to `0`.

The simplest implementation I can think of would be to allow a pattern definition such as:

    HAPROXY_NO_SERVERS backend %{NOTSPACE:pxname:tag} has no server available!%{'0':act:int}

In this, the `%{'':x:y}` would be special syntax where the value isn't parsed from the line, but instead from between the quotes. The value would then be parsed as an int, and put in the field `act`.

&nbsp;

This feels like it might be useful for lots of other use cases, but it also feels a little like an edge case.
However I cannot think of any other way of accomplishing our goal without writing a script that tails the log looking for the pattern, and then sends a point to telegraf. Which would be rather unpleasant.

danielnelson:
I'd rather not add special syntax if it can be avoided, do you think it is possible to do this with a processor?

phemmer:
Maybe. The difficulty i think is filtering. Making sure the processor only handles points from that specific logparser. You could add a tag on the input.logparser definition, and then have the processor filter on that tag (and remove the tag so it doesn't go to influx). But this feels a little messy.

danielnelson:
What about if we allow setting fields explicitly in the config?  The field/value pair would simply be set on all metrics.

phemmer:
That could work. As long as in this specific use case we say this logparser will only ever emit this one point. Which wouldn't be hard as if we want to parse other stuff, we can just define another logparser.

However one thought I just had, what about allowing per-plugin-instance parsers?

```
[[inputs.logparser]]
...
[inputs.logparser.processor]
...
```

We already have several global properties which apply to all inputs (`interval`,`name_override`,`tags`, etc). So why not add processors?
We could then create a processor that can do things like adding, removing, and transforming fields.
This would be a very flexible solution that could solve a very large number of use cases.

danielnelson:
[...]

PR #7370, opened 2020-04-19
Set default value for fields: add defaulter processor

jregistr:
Hellos! 👋 

So I've been experimenting w/ this amazing application y'all have here. 
Now, for certain fields coming out of the grok parser Input, I wanted a processor to set a default value when they're either nil, empty or zero. 

I didn't see a way to do this with the current set of processors so I got to some experimenting and made this very rough implementation.

I made this draft PR to get some thoughts on if a processor is the way to approach this, as well as thoughts on the config and implementation.

All comments are welcomed. 

Thank you all very much for taking a look.

### Required for all PRs:

### DIFF
--- a/plugins/processors/defaulter/README.md
+++ b/plugins/processors/defaulter/README.md
@@ -1,0 +1,43 @@
+# Defaulter Processor
+
+The *Defaulter* processor allows you to ensure certain fields will always exist with a specified default value on your metric(s).
+
+There are two cases where this processor will insert a configured default field.
+
+1. The field is nil on the incoming metric
+1. The field is present, but its value is empty.
+    
+    This processor considers the field empty if its value is the empty string or a single space character.
+
+### Configuration
+```toml
+## Set default fields on your metric(s) when they are nil or empty
+[[processors.defaulter]]
+
+## This table determines what fields will be inserted in your metric(s)
+  [processors.defaulter.fields]
+    field_1 = "bar"
+    time_idle = 0
+    is_error = true
+```
+
+### Example
+Ensure a _status\_code_ field with _N/A_ is inserted in the metric when one it's not set in the metric be default:
+
+```toml
+[[processors.defaulter]]
+  [processors.defaulter.fields]
+    status_code = "N/A"
+```
+
+```diff
+- lb,http_method=GET cache_status=HIT latency=230
++ lb,http_method=GET cache_status latency=230 status_code="N/A"
+```
+
+Ensure an empty string gets replaced by a default:
+
+```diff
+- lb,http_method=GET cache_status=HIT latency=230 status_code=""
++ lb,http_method=GET cache_status latency=230 status_code="N/A"
+```
--- a/plugins/processors/defaulter/defaulter.go
+++ b/plugins/processors/defaulter/defaulter.go
@@ -6,68 +6,52 @@ import (
 )
 
 const sampleConfig = `
-	[[processors.defaulter.values]]
-		fields = ["field_1", "field_2", "field_3"]
-		value = "NONE"
-		metric_name = "CPU"
-
-
-	[[processors.defaulter.values]]
-		field = ["field_4", "field_5"]
-		value = "TEST"
-		metric_name = "Disk"
-
-	# If the same field shows up in multiple of these value objects,
-	#then the last one will win out.
+	## Ensure's a set of fields always exist on your metric(s) with at the specified default value.
+	## For any given field pair (key = default) under *[processors.defaulter.fields]*, if it's not set, or its value
+	## is 'empty' on the incoming metric, a new or updated metric is set on the metric with the specified default.
+	## 
+	#	[processors.defaulter.fields]
+	#	 field_1 = "bar"
+	#    time_idle = 0
+	#    is_error = true
 `
 
-type DefaultFieldsSet struct {
-	Fields []string `toml:"fields"`
-	Metric string   `toml:"metric_name"`
-	Value  string   `toml:"value"`
-}
-
+// Defaulter is a processor for ensuring certain fields always exist
+// on your Metrics with at least a default value.
 type Defaulter struct {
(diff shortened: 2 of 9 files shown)


danielnelson:
Thanks for the pull request, I'm sure this will come in handy.

How about this as an alternate configuration style, I think it would fit more closely with existing plugins:
```toml
[[processors.default]]
  [processors.default.fields]
    foo = "bar"
    time_idle = 0
```

Operating on metric's with specific names and tags can be done with the [metric filtering](https://github.com/influxdata/telegraf/blob/master/docs/CONFIGURATION.md#metric-filtering) options that are available automatically for all plugins:

```toml
[[processors.default]]
  namepass = ["cpu"]
  [processors.default.fields]
    time_idle = 0
```

We should make sure to support floats/ints/bool/strings for the field type.

related #2564

jregistr:
I'm still pretty new to golang... I'm not sure why the tidy step is failing. I ran it locally and checked the changes.

jregistr:
But besides that, this is ready for a look. I updated the config according to your suggestion @danielnelson . I went ahead and updated the impl and tests to match.

plugins/processors/defaulter/README.md
```diff
@@ -0,0 +1,43 @@
+# Defaulter Processor
```
danielnelson, line 1: Let's call the plugin either `defaults` or `default`, this fits the noun style of most of our processors.

plugins/processors/defaulter/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+```toml
+[[processors.defaulter]]
+  [processors.defaulter.fields]
+    status_code = "N/A"
+```
+
+```diff
+- lb,http_method=GET cache_status=HIT latency=230
++ lb,http_method=GET cache_status latency=230 status_code="N/A"
```
danielnelson, line 35: In order for these examples to be valid line protocol we must separate fields with a comma:
```
lb,http_method=GET cache_status,latency=230,status_code="N/A"
```

plugins/processors/defaulter/defaulter.go
```diff
@@ -0,0 +1,62 @@
...
+
+const sampleConfig = `
+	## Ensure's a set of fields always exist on your metric(s) with at the specified default value.
+	## For any given field pair (key = default) under *[processors.defaulter.fields]*, if it's not set, or its value
+	## is 'empty' on the incoming metric, a new or updated metric is set on the metric with the specified default.
+	## 
+	#	[processors.defaulter.fields]
+	#	 field_1 = "bar"
+	#    time_idle = 0
+	#    is_error = true
```
danielnelson, line 16: Follow this guide to match style with other plugins: https://github.com/influxdata/telegraf/wiki/SampleConfig

2 space indent, wrap line length, etc.

plugins/processors/defaulter/defaulter.go
```diff
@@ -0,0 +1,62 @@
...
+// Apply contains the main implementation of this processor.
+// For each metric in 'inputMetrics', it goes over each default pair.
+// If the field in the pair does not exist on the metric, the associated default is added.
+// If the field was found, then, if its value is the empty string or a single space, it is replaced
+// by the associated default.
+func (def *Defaulter) Apply(inputMetrics ...telegraf.Metric) []telegraf.Metric {
+	for _, metric := range inputMetrics {
+		for defField, defValue := range def.DefaultFieldsSets {
+			if maybeCurrent, isSet := metric.GetField(defField); !isSet {
+				def.Log.Debugf("Field with name: %v, was not set on metric: %v.", defField, metric.Name())
```
danielnelson, line 46: Remove these debug log messages, they are a bit too verbose for the final code, also on L49

plugins/processors/defaulter/defaulter.go
```diff
@@ -0,0 +1,62 @@
...
+				metric.AddField(defField, defValue)
+			} else if maybeCurrent == "" || maybeCurrent == " " {
```
danielnelson, line 48: Call `strings.TrimSpace` and compare it against "" when deciding if the value needs replaced.  Undate the note in the README about what is considered unset.

plugins/processors/defaulter/defaulter.go
```diff
@@ -0,0 +1,62 @@
...
+				def.Log.Debugf("Field with name: %v was set, but the value (%v) is considered empty. Setting new value to %v", defField, maybeCurrent, defValue)
+				metric.RemoveField(defField)
+				metric.AddField(defField, defValue)
```
danielnelson, line 51: You can just call AddField, it will replace the existing value.

plugins/processors/defaulter/README.md
jregistr: Gotcha, `defaults` is a great replacement!

plugins/processors/defaulter/defaulter.go
jregistr: aaah, interesting! Will update!

jregistr:
@danielnelson  Thank you for the feedback, I'll update and push the changes.

jregistr:
Hello again @danielnelson, I believe this is ready for another look.

jregistr:
>Error: No such image: quay.io/influxdb/telegraf-ci:1.13.8

Looks like ci failed to download an image?

plugins/processors/defaults/README.md
```diff
@@ -0,0 +1,42 @@
+# Defaults Processor
+
+The *Defaults* processor allows you to ensure certain fields will always exist with a specified default value on your metric(s).
+
+There are three cases where this processor will insert a configured default field.
+
+1. The field is nil on the incoming metric
+1. The field is not nil, but its value is an empty string.
+1. The field is not nil, but it's value is a string of one or more empty spaces.
```
reimda, line 9: typo: it's should be its

jregistr: Good catch! Fixed!

jregistr:
Amazing!

Issue #7369, opened 2020-04-18
Apache Pulsar Output Plugin


skyrocknroll:
## Feature Request

Apache Pulsar Output Plugin 

### Current behavior:
Right now pulsar output is not supported 

### Desired behavior:

Apache Pulsar is growing as next-generation streaming and queuing system. Pulsar output plugin would be very useful.

PR #5279, opened 2019-01-11
add pulsar output plugin

JesusIslam:
Add output plugin for [Apache Pulsar](https://pulsar.apache.org).

### Required for all PRs:

- [x] Signed [CLA](https://influxdata.com/community/cla/).
- [x] Associated README.md updated.
- [x] Has appropriate unit tests.

PR #7372, opened 2020-04-19
Apache Pulsar output support 

skyrocknroll:
### Required for all PRs:


skyrocknroll:
@danielnelson Please Take a look. Thanks

skyrocknroll:
Closes https://github.com/influxdata/telegraf/issues/7369

etc/telegraf.conf
```diff
@@ -1123,6 +1123,123 @@
...
+# #    separator = "-"
+#
+# ## Suffix equals to "_" + measurement's "foo" and "bar"
+# ##   tag values, separated by "_". If there is no such tags,
+# ##   their values treated as empty strings.
+# #  [outputs.pulsar.topic_suffix]
+# #    method = "tags"
+# #    keys = ["foo","bar"]
+# #    separator = "-"
+
```
danielnelson, line 1242: Let's remove the topic suffic configuration, this was an attempt to give more flexibility to the topic selection, but it didn't work out as well as I had hoped.  Instead we can use the `topic_tag` option and for custom jobs use it along with processors.

etc/telegraf.conf
```diff
@@ -1123,6 +1123,123 @@
...
+#
+# ## BatchingMaxPublishDelay set the time period within which the messages sent will be batched (default: 10ms)
+# ## if batch messages are enabled. If set to a non zero value, messages will be queued until this time
+# ## interval or until
+# # batching_max_publish_delay = "10ms"
+#
+# ## BatchingMaxMessages set the maximum number of messages permitted in a batch. (default: 1000)
+# ## If set to a value greater than 1, messages will be queued until this threshold is reached or
+# ## batch interval has elapsed.
+# # batching_max_messages = 1000
```
danielnelson, line 1196: Can we set this to the metric_batch_size?

etc/telegraf.conf
```diff
@@ -1123,6 +1123,123 @@
...
+# # compression_type = 0
+#
+# ## MaxPendingMessages set the max size of the queue holding the messages pending to receive an
+# ## acknowledgment from the broker.
+# # max_pending_messages = 1000
+#
+# ## BatchingMaxPublishDelay set the time period within which the messages sent will be batched (default: 10ms)
+# ## if batch messages are enabled. If set to a non zero value, messages will be queued until this time
+# ## interval or until
+# # batching_max_publish_delay = "10ms"
```
danielnelson, line 1191: We should write all messages for the batch immediately, before returning from Write.  This is important for durability and retries since returning from Write without error signifies to Telegraf that all metrics were written.  I know this will be slower but it is important to be reliable.

etc/telegraf.conf
```diff
@@ -1123,6 +1123,123 @@
...
+# # auth_provider = ""
+#
+## For token auth provider
+# # auth_token = ""
+# # Set the following values for tls auth provider
+# # tls_allow_insecure_connection = false
+# # tls_trust_certs_file_path = ""
+# # tls_validate_host_name = true
+# # tls_certificate_path = ""
+# # tls_private_key_path = ""
```
danielnelson, line 1168: Can we use the regular TLS options for some of these?

etc/telegraf.conf
skyrocknroll: @danielnelson I am using the length of the []metrics to determine the batch size. Is it possible to directly get metric_batch_size defined in the config? This will help me to completely discard this option.   Because length may not tell the batch size. and batch size is set at producer creation times in pulsar unlike kafka where it is set and every sendMessages.

skyrocknroll: @danielnelson is it possible to get the metric_batch_size config? If so how can I get it?  Right now I am using length of []metrics to determine the batch size. In pulsar, batch size needs to be set at Producer creation time unlike in kafka where batch size can be adjusted everytime  we send messages.

skyrocknroll:
@danielnelson Please I have made couple changes Please review it. I would like to onboard pulsar output in the next release.

liangyuanpeng:
When can we have pulsar output?

sjwang90:
Is there any preference for this to #5279?

skyrocknroll:
@sjwang90 This PR uses pure go pulsar client so that we don't need to install C++ pulsar client

garysu:
Is there an ETA on merging this PR given all checks have passed?

plugins/outputs/pulsar/README.md
```diff
@@ -0,0 +1,89 @@
...
+This plugin writes to a [Apache Pulsar](https://pulsar.apache.org/en/) acting as Pulsar Producer.
+
+### Configuration:
+```toml
+## Configuration for apache pulsar  to send metrics to 
+[[outputs.pulsar]]
+ ## URLs of pulsar url
+  url = "pulsar://localhost:6650"
+ ## Pulsar topic for producer messages
+  topic = "persistent://public/default/telegraf"
```
srebhan, line 12: There seems to be a formatting issue.

plugins/outputs/pulsar/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+ ## The value of this tag will be used as the topic.  If not set the 'topic'
+ ## option is used.
+ # topic_tag = "foo"
+
+ ## If true, the 'topic_tag' will be removed from to the metric.
+ # exclude_topic_tag = false
+
+  routing_tag = "host"
```
srebhan, line 21: Can we please have a description for this parameter.

plugins/outputs/pulsar/README.md
```diff
@@ -0,0 +1,89 @@
...
+ # Set the following values for tls auth provider
+ # insecure_skip_verify = false
+ # tls_ca = ""
+ # tls_validate_host_name = true
+ # tls_cert = ""
+ # tls_key = ""
+
+ ## Optional timeout Config
+ # connection_timeout = "30s"
+ # operation_timeout = "30s"
```
srebhan, line 56: Do you really need this fine-grained timeout handling?

plugins/outputs/pulsar/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+ ## Optional Producer Config
+
+ ## CompressionType represents the various compression codecs recognized by
+ ## Pulsar in messages.
+ ##  0 : No compression
+ ##  1 : LZ4 compression
+ ##  2 : ZLib compression
+ ##  3 : ZSTD compression
+ # compression_type = 0
```
srebhan, line 66: Please use a string here and map it internally to the number.

plugins/outputs/pulsar/pulsar.go
```diff
@@ -0,0 +1,378 @@
...
+	"strings"
+	"time"
+
+	"github.com/apache/pulsar-client-go/pulsar"
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/plugins/outputs"
+	"github.com/influxdata/telegraf/plugins/serializers"
+)
+
+var ValidTopicSuffixMethods = []string{
```
srebhan, line 17: Please make this a private global variable or even better move it to the struct.

plugins/outputs/pulsar/pulsar.go
```diff
@@ -0,0 +1,378 @@
...
+	"",
+	"measurement",
+	"tags",
+}
+
+var zeroTime = time.Unix(0, 0)
```
srebhan, line 23: You use this only once, so please move this directly to that location. You can also use `time.Time{}`.

plugins/outputs/pulsar/pulsar.go
```diff
@@ -0,0 +1,378 @@
...
+
+type (
+	Pulsar struct {
```
srebhan, line 26: Please use individual declarations
```suggestion
type Pulsar struct {
```
to ease searching.

plugins/outputs/pulsar/pulsar.go
```diff
@@ -0,0 +1,378 @@
...
+		RoutingTag      string `toml:"routing_tag"`
+		RoutingKey      string `toml:"routing_key"`
+
+		//Client Options
+		AuthProvider               string `toml:"auth_provider"`
+		TLSAllowInsecureConnection bool   `toml:"insecure_skip_verify"`
+		TLSTrustCertsFilePath      string `toml:"tls_ca"`
+		TLSValidateHostname        bool   `toml:"tls_validate_host_name"`
+		TLSCertificatePath         string `toml:"tls_cert"`
+		TLSPrivateKeyPath          string `toml:"tls_key"`
```
srebhan, line 40: Please use `tls.TLSConfig` for this set of parameters. See the `influxdb` or `mqtt` output plugin for example.

plugins/outputs/pulsar/pulsar.go
```diff
@@ -0,0 +1,378 @@
...
+		AuthToken                  string `toml:"auth_token"`
+		ConnectionTimeout          string `toml:"connection_timeout"`
+		OperationTimeout           string `toml:"operation_timeout"`
+
+		//Producer Options
+		CompressionType         int    `toml:"compression_type"`
```
srebhan, line 46: Please make this a `string` containing the compression name and map it internally to the int. This way it's easier for the user to understand what is going on.

plugins/outputs/pulsar/pulsar.go
```diff
@@ -0,0 +1,378 @@
...
+
+		client pulsar.Client
+
+		serializer serializers.Serializer
+	}
+)
+
+var sampleConfig = `
+  [[outputs.pulsar]]
+ ## URLs of pulsar url
```
srebhan, line 65: Please fix the formatting.

plugins/outputs/pulsar/pulsar.go
```diff
@@ -0,0 +1,378 @@
...
+  url = "pulsar://localhost:6650"
```
srebhan, line 66: Is this a sensible default? If so, please comment this line as follows
```suggestion
  # url = "pulsar://localhost:6650"
```
to indicate the default value of the parameter.

plugins/outputs/pulsar/pulsar.go
```diff
@@ -0,0 +1,378 @@
...
+ ## Pulsar topic for producer messages
+  topic = "persistent://public/default/telegraf"
```
srebhan, line 68: Is this mandatory to be set by the user or can we find a sensible default?

plugins/outputs/pulsar/pulsar.go
```diff
@@ -0,0 +1,378 @@
...
+ ## option is used.
+ # topic_tag = "foo"
+
+ ## If true, the 'topic_tag' will be removed from to the metric.
+ # exclude_topic_tag = false
+
+  routing_tag = "host"
+
+ ## The routing key is set as the message key and used to determine which
+ ## partition to send the message to.  This value is only used when no
```
srebhan, line 80:
```suggestion
 ## partition to send the message to. This value is only used when no
```

plugins/outputs/pulsar/pulsar.go
```diff
@@ -0,0 +1,378 @@
...
+  topic = "persistent://public/default/telegraf"
+
+ ## The value of this tag will be used as the topic.  If not set the 'topic'
+ ## option is used.
+ # topic_tag = "foo"
+
+ ## If true, the 'topic_tag' will be removed from to the metric.
+ # exclude_topic_tag = false
+
+  routing_tag = "host"
```
srebhan, line 77: Please provide a separate description or move to the explanation below.

plugins/outputs/pulsar/pulsar.go
```diff
@@ -0,0 +1,378 @@
...
+ ## acknowledgment from the broker.
+ # max_pending_messages = 1000
+
+ ## BatchingMaxMessages set the maximum number of messages permitted in a batch. (default: metric_batch_size)
+ ## If set to a value greater than 1, messages will be queued until this threshold is reached or
+ ## batch interval has elapsed. By Default it is set to "metric_batch_size"
+ # batching_max_messages = 1000
+
+ ## HashingScheme change the "HashingScheme" used to chose the partition on where to publish a particular message.
+ ##  JavaStringHash : JavaStringHash Hshing
```
srebhan, line 134:
```suggestion
 ##  JavaStringHash : JavaStringHash Hashing
```

plugins/outputs/pulsar/pulsar.go
```diff
@@ -0,0 +1,378 @@
...
+ ##  Murmur3_32Hash : Murmur3_32Hash Hashing
+ # hashing_scheme = "JavaStringHash"
```
srebhan, line 136: Can we shorten the options to `java` and `murmur3`?

plugins/outputs/pulsar/pulsar.go
```diff
@@ -0,0 +1,378 @@
...
+`
+
+func ValidateTopicSuffixMethod(method string) error {
+	for _, validMethod := range ValidTopicSuffixMethods {
+		if method == validMethod {
+			return nil
+		}
+	}
+	return fmt.Errorf("Unknown topic suffix method provided: %s", method)
+}
```
srebhan, line 154: You can replace this using `internal.choice` as `choice.Contains(method, []string{"", "measurements", "tags"})`.

plugins/outputs/pulsar/pulsar.go
```diff
@@ -0,0 +1,378 @@
...
+
+func (p *Pulsar) GetTopicName(metric telegraf.Metric) (telegraf.Metric, string) {
```
srebhan, line 156: Please make this a private function.

plugins/outputs/pulsar/pulsar.go
```diff
@@ -0,0 +1,378 @@
...
+		} else {
+			p.Log.Infof("Created producer for topic %s", topic)
+			p.producerCache[topic] = producer
+			return producer, nil
+		}
+
+	} else {
+		p.Log.Debugf("Returning cached producer for topic %s", topic)
+		return producer, nil
+	}
```
srebhan, line 219: Please invert the logic to save some indention level(s):
```suggestion
	producer, ok := p.producerCache[topic]
	if ok {
		p.Log.Debugf("Returning cached producer for topic %s", topic)
		return producer, nil
	}
	producerOptions := pulsar.ProducerOptions{
		Topic:               topic,
		BatchingMaxMessages: metricBatchSize,
	}
	...	
```

plugins/outputs/pulsar/pulsar.go
```diff
@@ -0,0 +1,378 @@
...
+			producerOptions.MaxPendingMessages = p.MaxPendingMessages
+		}
+		duration, err := time.ParseDuration(p.BatchingMaxPublishDelay)
+		if err != nil {
+			producerOptions.BatchingMaxPublishDelay = duration
+		}
+		producer, err := p.client.CreateProducer(producerOptions)
+		if err != nil {
+			return nil, err
+		} else {
```
srebhan, line 210: No need for the `else` as you return in the `if`. This saves another indention level.

plugins/outputs/pulsar/pulsar.go
```diff
@@ -0,0 +1,378 @@
...
+
+}
+func (p *Pulsar) SetSerializer(serializer serializers.Serializer) {
+	p.serializer = serializer
+}
+
+func (p *Pulsar) Connect() error {
+	p.producerCache = make(map[string]pulsar.Producer)
+	splits := strings.Split(p.Topic, "/")
+	p.tenantNameSpace = strings.Join(splits[0:4], "/")
```
srebhan, line 229: This will panic if `splits` has less than four elements. Please check the length of `splits` before doing this operation. Furthermore, why do you strip the rest of the elements? Why four? If this is arbitrary, please let the user decide, otherwise, please add a comment explaining why you need to do this.

plugins/outputs/pulsar/pulsar.go
```diff
@@ -0,0 +1,378 @@
...
+		//Client Options
+		AuthProvider               string `toml:"auth_provider"`
+		TLSAllowInsecureConnection bool   `toml:"insecure_skip_verify"`
+		TLSTrustCertsFilePath      string `toml:"tls_ca"`
+		TLSValidateHostname        bool   `toml:"tls_validate_host_name"`
+		TLSCertificatePath         string `toml:"tls_cert"`
+		TLSPrivateKeyPath          string `toml:"tls_key"`
+		AuthToken                  string `toml:"auth_token"`
+		ConnectionTimeout          string `toml:"connection_timeout"`
+		OperationTimeout           string `toml:"operation_timeout"`
```
srebhan, line 43: Please using this
```suggestion
		ConnectionTimeout          internal.Duration `toml:"connection_timeout"`
		OperationTimeout           internal.Duration `toml:"operation_timeout"`
```
to allow the `toml` parser to parse the duration string for you.

plugins/outputs/pulsar/pulsar.go
```diff
@@ -0,0 +1,378 @@
...
+	splits := strings.Split(p.Topic, "/")
+	p.tenantNameSpace = strings.Join(splits[0:4], "/")
+	clientOptions := pulsar.ClientOptions{
+		URL:                        p.URL,
+		TLSAllowInsecureConnection: p.TLSAllowInsecureConnection,
+		TLSTrustCertsFilePath:      p.TLSTrustCertsFilePath,
+		TLSValidateHostname:        p.TLSValidateHostname,
+	}
+	if p.ConnectionTimeout != "" {
+		connectionTimeout, err := time.ParseDuration(p.ConnectionTimeout)
```
srebhan, line 237: Please use `internal.Duration` for the field as commented above so that you don't need to do this manually here.

plugins/outputs/pulsar/pulsar.go
```diff
@@ -0,0 +1,378 @@
...
+		if err != nil {
+			p.Log.Error("Invalid ConnectionTimeout. Setting it to 30s ", err.Error())
```
srebhan, line 239: If this is an error "die" here and return the error instead of trying to circumvent it.

plugins/outputs/pulsar/pulsar.go
```diff
@@ -0,0 +1,378 @@
...
+	p.producerCache = make(map[string]pulsar.Producer)
+	splits := strings.Split(p.Topic, "/")
+	p.tenantNameSpace = strings.Join(splits[0:4], "/")
+	clientOptions := pulsar.ClientOptions{
+		URL:                        p.URL,
+		TLSAllowInsecureConnection: p.TLSAllowInsecureConnection,
+		TLSTrustCertsFilePath:      p.TLSTrustCertsFilePath,
+		TLSValidateHostname:        p.TLSValidateHostname,
+	}
+	if p.ConnectionTimeout != "" {
```
srebhan, line 236: `ConnectionTimeout` will be `""` in case we comment the option and use the default. Therefore, the user expects that, in case the option is commented you use the default. The best way is to initialize `ConnectionTimeout` when creating the struct instance in `init()`.

plugins/outputs/pulsar/pulsar.go
```diff
@@ -0,0 +1,378 @@
...
+	if p.OperationTimeout != "" {
+		operationTimeout, err := time.ParseDuration(p.OperationTimeout)
+		if err != nil {
+			p.Log.Error("Invalid OperationTimeout. Setting it to 30s", err.Error())
+			clientOptions.OperationTimeout = 30 * time.Second
+		} else {
+			clientOptions.OperationTimeout = operationTimeout
+		}
+
+	}
```
srebhan, line 254: See `ConnectionTimeout` comment.

plugins/outputs/pulsar/pulsar.go
```diff
@@ -0,0 +1,378 @@
...
+
+	if p.AuthProvider == "tls" {
+		if p.TLSCertificatePath != "" && p.TLSPrivateKeyPath != "" {
+			clientOptions.Authentication = pulsar.NewAuthenticationTLS(p.TLSCertificatePath, p.TLSPrivateKeyPath)
+		} else {
+			return fmt.Errorf("tls auth specified but tls_certificate_path or tls_private_key_path is miising")
+		}
+
```
srebhan, line 262: Please invert the logic
```suggestion
		if p.TLSCertificatePath == "" || p.TLSPrivateKeyPath == "" {
			return fmt.Errorf("tls auth specified but tls_certificate_path or tls_private_key_path is miising")
		}
		clientOptions.Authentication = pulsar.NewAuthenticationTLS(p.TLSCertificatePath, p.TLSPrivateKeyPath)

```

plugins/outputs/pulsar/pulsar.go
```diff
@@ -0,0 +1,378 @@
...
+	} else if p.AuthProvider == "token" {
+		if p.AuthToken != "" {
+			clientOptions.Authentication = pulsar.NewAuthenticationToken(p.AuthToken)
+		} else {
+			return fmt.Errorf("auth provider is token but auth_token is empty")
+		}
+
```
srebhan, line 269:
```suggestion
		if p.AuthToken == "" {
			return fmt.Errorf("auth provider is token but auth_token is empty")
		}
		clientOptions.Authentication = pulsar.NewAuthenticationToken(p.AuthToken)
```

plugins/outputs/pulsar/pulsar.go
```diff
@@ -0,0 +1,378 @@
...
+	}
+
+	if p.RoutingKey == "measurement" {
+		return metric.Name()
+	}
+
+	if p.RoutingKey == "random" {
+		u, err := uuid.NewV4()
+		if err != nil {
+			p.Log.Errorf("Unable to generate uudid %s", err.Error())
```
srebhan, line 310: Error messages should start lowercase.

plugins/outputs/pulsar/pulsar.go
```diff
@@ -0,0 +1,378 @@
...
+			return ""
+		}
+		return u.String()
+	}
+	return p.RoutingKey
+}
+func sendAsyncCallback(id pulsar.MessageID, message *pulsar.ProducerMessage, err error) {
+	if err != nil {
+		log.Printf("Failed to publish metrics : %s", err.Error())
```
srebhan, line 319: No usage of `log` please. You might want to use a closure to achieve your goal.

plugins/outputs/pulsar/pulsar.go
```diff
@@ -0,0 +1,378 @@
...
+}
+
+func (p *Pulsar) Write(metrics []telegraf.Metric) error {
+	var batchSize uint
+	// Setting the batch size to metric_batch_size is unset
+	if p.BatchingMaxMessages == 0 {
+		batchSize = uint(len(metrics))
+	} else {
+		batchSize = p.BatchingMaxMessages
+	}
```
srebhan, line 331:
```suggestion
	// Setting the batch size to metric_batch_size is unset
	batchSize := p.BatchingMaxMessages
	if p.BatchingMaxMessages == 0 {
		batchSize = uint(len(metrics))
	}
```

plugins/outputs/pulsar/pulsar.go
```diff
@@ -0,0 +1,378 @@
...
+
+			// Negative timestamps are not allowed by the Pulsar protocol.
+			if !metric.Time().Before(zeroTime) {
+				m.EventTime = metric.Time()
+			}
+			producer.SendAsync(context.Background(), &m, sendAsyncCallback)
+		} else {
+			p.Log.Errorf("Unable to create producer for topic %s : %s", topic, err.Error())
+			return err
+		}
```
srebhan, line 355: Please again invert the logic. Furthermore you might not want to `return` here as this way you will drop all following metrics even if only this single `metric` is faulty. A `continue` would probably be better.

plugins/outputs/pulsar/pulsar.go
```diff
@@ -0,0 +1,378 @@
...
+			m := pulsar.ProducerMessage{
+				Payload:   buf,
+				EventTime: metric.Time(),
+				Key:       p.routingKey(metric),
+			}
+
+			// Negative timestamps are not allowed by the Pulsar protocol.
+			if !metric.Time().Before(zeroTime) {
+				m.EventTime = metric.Time()
+			}
```
srebhan, line 350: Well even though it is not allowed you still use that time? Did you mean
```suggestion
			if !metric.Time().Before(zeroTime) {
				m.EventTime = time.Now()
			}
```

plugins/outputs/pulsar/pulsar.go
```diff
@@ -0,0 +1,378 @@
...
+			producer.SendAsync(context.Background(), &m, sendAsyncCallback)
+		} else {
+			p.Log.Errorf("Unable to create producer for topic %s : %s", topic, err.Error())
+			return err
+		}
+
+	}
+	// We flush and make sure all the metrics are published and acknowledged by the broker
+	// so that in case of failure the metrics are retried
+	for topic, producer := range p.producerCache {
```
srebhan, line 360: Wouldn't it be easier, more efficient and more safe to collect all producers used in the loop above and flush those instead of this indirect way?
Furthermore, would it make sense to flush all producers in parallel?

plugins/outputs/pulsar/pulsar.go
```diff
@@ -0,0 +1,378 @@
...
+		err := producer.Flush()
+		if err != nil {
+			p.Log.Errorf("Failed to publish the metrics in topic %s : %s ", topic, err.Error())
```
srebhan, line 363: Error messages should start lowercase.

srebhan (changes requested):
Thank you very much @skyrocknroll for the cool plugin. I do have mostly minor comments but it's quite a number. It would be nice if you could check them in the code.
The most severe issue I see is the completely missing test-suite! You probably can mock a `Producer` and "poison" the `producerCache` for a full fledged test!

srebhan:
@skyrocknroll any chance you can take a look at my comments to push this forward?

srebhan:

sjwang90:

PR #7374, opened 2020-04-20
add User-Agent header on fetchConfig

ronopitz:
include OS and ARCH

needed on server side for:
identify the operating system to generate certain configurations
identify the telegraf version to check plugin is available

### Required for all PRs:

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -860,7 +860,7 @@ func fetchConfig(u *url.URL) ([]byte, error) {
 		req.Header.Add("Authorization", "Token "+v)
 	}
 	req.Header.Add("Accept", "application/toml")
-	req.Header.Set("User-Agent", "Telegraf/"+internal.Version()+" ("+runtime.GOOS+" "+runtime.GOARCH+")")
+	req.Header.Set("User-Agent", internal.ProductToken()+" ("+runtime.GOOS+" "+runtime.GOARCH+")")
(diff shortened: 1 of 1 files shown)


internal/config/config.go
```diff
@@ -860,6 +860,7 @@ func fetchConfig(u *url.URL) ([]byte, error) {
 		req.Header.Add("Authorization", "Token "+v)
 	}
 	req.Header.Add("Accept", "application/toml")
+	req.Header.Set("User-Agent", "Telegraf/"+internal.Version()+" ("+runtime.GOOS+" "+runtime.GOARCH+")")
```
danielnelson, line 863: Use `internal.ProductToken()` for the value, it will look something like: `Telegraf/1.14.0 Go/1.13.8` which is a rfc2616 [product token](https://tools.ietf.org/html/rfc2616#section-3.8).

ronopitz: My commit based on the User-Agent Header from the http output plugin.
You mean like this.
`req.Header.Set("User-Agent", internal.ProductToken()+" ("+runtime.GOOS+" "+runtime.GOARCH+")")`

danielnelson: We'll just want the product token part, I'll update the http output too
```
req.Header.Set("User-Agent", internal.ProductToken())
```

danielnelson:
see #7752

PR #7390, opened 2020-04-23
Rework plugin tickers to prevent drift and spread write ticks

### LINKED ISSUE #4232: Allow full flush interval to be used for flushing with flush_jitter.
## Feature Request

When using a `flush_interval` and `flush_jitter` of the same duration and the
random sleep jitter is close to the length of the interval, the output may
have no time available before the "Skipping a scheduled flush" warning is
emitted.

This forces the flush_jitter to be smaller than the interval making it
impossible to take advantage of the full interval period.

### Current behavior:

If jitter is the same as interval and the flush takes "2s" you can expect
warnings if there is a high jitter followed by a low jitter:
```
[agent]
  flush_interval = "10s"
  flush_jitter = "10s"
```

If the jitter is half the interval then no warnings, but all writes take place
in the first half of the flush interval:
```
[agent]
  flush_interval = "10s"
  flush_jitter = "5s"
```

### Desired behavior:

If a flush starts at the end of the period, there shouldn't be a warning
unless it exceeds a certain period of time.

Balancing connections to InfluxDB over the full flush interval without uneven
writes.

### LINKED ISSUE #4652: Windows : telegraf doesn't respect round_interval and collect data too soon
Might be related (or not) to #4487

### System info:
Telegraf-1.7.3 on Windows Server 2012

### Steps to reproduce:
use   `interval = "5m"` with `round_interval = true` in telegraf.conf
1. run telegraf and wait up to ~48h
2. query the windows measurements in Influxdb 

### Expected behavior:
query influxdb, returning the metrics with a timestamp at [0 or 5]:00 +jitter sec

### Actual behavior:
the metrics are returned with a timestamp that is not rounded, resulting in a real interval lower than the telegraf setting.

It's usually at [0 or 5]m00s +jitter sec, but it will also be at [9 or 4]m5?s +jitter sec

Influxdb query : 
(I used uptime, but it's the same for ALL windows metrics) 
````
Precision rfc3339
SELECT "System_Up_Time" FROM "win_system" WHERE "host" =~ /^XXXXX$/ AND time >= now() - 2d
2018-09-07T14:20:00Z 464354.03125       <= OK
2018-09-07T14:24:58Z 464652.0625        <= not rounded, should be at T14:25:00 +?s
2018-09-07T14:30:00Z 464954.03125       <= OK
2018-09-07T14:34:58Z 465252.5           <= not rounded, should be at T14:35:00 +?s
2018-09-07T14:39:58Z 465552.65625       <= ...
2018-09-07T14:44:59Z 465853.34375
2018-09-07T14:49:58Z 466152.8125
````
All data should be at [0 or 5]m00s +? sec, those at [4 or 9]m59s  are the anomaly. I had some up to :57s.

By the way, the influxdb log is consistent (mind the timezone), confirming it really comes from telegraf  : 
````
[xWINDOWS SERVER IPx] - telegraf [07/Sep/2018:16:34:59 +0200] "POST /write?consistency=any&db=telegraf HTTP/1.1" 204 0 "-" "telegraf-node" 32f440cc-b2ab-11e8-bdf4-000000000000 57296
[xWINDOWS SERVER IPx] - telegraf [07/Sep/2018:16:39:59 +0200] "POST /write?consistency=any&db=telegraf HTTP/1.1" 204 0 "-" "telegraf-node" e5a85994-b2ab-11e8-be12-000000000000 54763
````

### Additional info:
This doesn't happen immediately, the agent need to work for at least 24h to start exhibiting this behavior. Restarting the agent fixes the issue, but for some hours only.

I'm running the same agent configuration on Linux and Windows, with a global influxdb server, collecting both systems. 
The Linux servers don't and never have show this behavior, only the single existing Windows server does. All are VMware virtual machines, plugged on the same NTP server, and the Windows server shows the exact same time up to the second as the influxdb server.

The log is empty on the Windows server, no error is present.

It's worth noting that IIRC, this issue didn't appear before Telegraf v1.4 

### Relevant telegraf.conf:
````
# Telegraf configuration

# Telegraf is entirely plugin driven. All metrics are gathered from the
# declared inputs, and sent to the declared outputs.

# Plugins must be declared in here to be active.
# To deactivate a plugin, comment out the name and any variables.

# Use 'telegraf -config telegraf.conf -test' to see what metrics a config
# file would generate.

# Global tags can be specified here in key="value" format.
[global_tags]
  # dc = "us-east-1" # will tag all metrics with dc=us-east-1
  # rack = "1a"
[ ... ]
  
# Configuration for telegraf agent
[agent]
  ## Default data collection interval for all inputs
  interval = "5m"
  ## Rounds collection interval to 'interval'
  ## ie, if interval="10s" then always collect on :00, :10, :20, etc.
  round_interval = true

  ## Telegraf will send metrics to outputs in batches of at most
  ## metric_batch_size metrics.
  ## This controls the size of writes that Telegraf sends to output plugins.
  metric_batch_size = 1000

  ## For failed writes, telegraf will cache metric_buffer_limit metrics for each
  ## output, and will flush this buffer on a successful write. Oldest metrics
  ## are dropped first when this buffer fills.
  ## This buffer only fills when writes fail to output plugin(s).
  metric_buffer_limit = 10000

  ## Flush the buffer whenever full, regardless of flush_interval.
  flush_buffer_when_full = true

  ## Collection jitter is used to jitter the collection by a random amount.
  ## Each plugin will sleep for a random time within jitter before collecting.
  ## This can be used to avoid many plugins querying things like sysfs at the
  ## same time, which can have a measurable effect on the system.
  collection_jitter = "2s"

  ## Default flushing interval for all outputs. You shouldn't set this below
  ## interval. Maximum flush_interval will be flush_interval + flush_jitter
  flush_interval = "10s"
  ## Jitter the flush interval by a random amount. This is primarily to avoid
  ## large write spikes for users running a large number of telegraf instances.
  flush_jitter = "5s"

  ## By default, precision will be set to the same timestamp order as the
  ## collection interval, with the maximum being 1s.
  ## Precision will NOT be used for service inputs, such as logparser and statsd.
  ## Valid values are "ns", "us" (or "µs"), "ms", "s".
  precision = "s"
  
  ## Logging configuration:
  ## Run telegraf in debug mode
  debug = false
  ## Run telegraf in quiet mode
  quiet = true
  
  ## Specify the log file name. The empty string means to log to stdout.
  logfile = "C:/Server/Telegraf/log/telegraf.log"
  
  ## Override default hostname, if empty use os.Hostname()
  hostname = "XXXXXX"
  ## If set to true, do no set the "host" tag in the telegraf agent.
  omit_hostname = false

#                                  OUTPUTS                                    #
````

### LINKED ISSUE #5335: Changing system time while Telegraf is running produces wrong round_interval timestamps
### Relevant telegraf.conf:
 interval = "1m"
 round_interval = true

 flush_interval = "1m"

### System info:
CentOS 7
Telegraf 1.9.3 (git: HEAD 6ad8c8b0)
Telegraf 1.9.1 (git: HEAD 20636091) (on a second VM)
InfluxDB shell version: 1.7.3

### Steps to reproduce:

1. Make sure Telegraf is running and feeding data to InfluxDB.
2. Check current system time with `date`
3. Change the time to a time with a different `second` mark. For example `sudo date +%T -s "17:55:32"` (make sure current time is not close to being 32 to avoid not changing the second mark)
4. Wait for your interval and flush interval to cycle once.
5. Enter InfluxDB through the CLI and use the appropriate database.
6. `precision rfc3339`
7. Make a simple query like `select pid from procstat where time > now() - 5m`

### Expected behavior:
```
2019-01-23T16:28:00Z 4944
2019-01-23T16:28:00Z 2796
2019-01-23T16:28:00Z 2115
2019-01-23T16:28:00Z 2169
2019-01-23T17:56:00Z 2115
```
### Actual behavior:
```
2019-01-23T16:28:00Z 4944
2019-01-23T16:28:00Z 2796
2019-01-23T16:28:00Z 2115
2019-01-23T16:28:00Z 2169
2019-01-23T17:56:16Z 2115
```
### Additional info:
Restarting Telegraf fixes this issue. I found this because I was running some VMs on a computer which was put on power saving mode during lunchtime, and the VMs clocks desynchronized. Using `chronyc makestep` produced this unexpected behavior.
This is probably the root cause of other similar issues, because if a system clock desynchronizes itself from official time and is later updated by chronyd, Telegraf will start recording metrics with the wrong timestamps.
One problem this causes is making a query which uses group by time between `2019-01-23T17:50:00Z` and `2019-01-23T17:56:00Z` with `fill(0)` can give you false nulls if there is other data with correct timestamps.

### LINKED ISSUE #5937: reporting interval drift over time
### Relevant telegraf.conf:

```
# Configuration for telegraf agent
[agent]
  ## Default data collection interval for all inputs
  interval = "60s"
  ## Rounds collection interval to 'interval'
  ## ie, if interval="10s" then always collect on :00, :10, :20, etc.
  round_interval = true

  ## Telegraf will send metrics to outputs in batches of at most
  ## metric_batch_size metrics.
  ## This controls the size of writes that Telegraf sends to output plugins.
  metric_batch_size = 1000

  ## For failed writes, telegraf will cache metric_buffer_limit metrics for each
  ## output, and will flush this buffer on a successful write. Oldest metrics
  ## are dropped first when this buffer fills.
  ## This buffer only fills when writes fail to output plugin(s).
  metric_buffer_limit = 10000

  ## Collection jitter is used to jitter the collection by a random amount.
  ## Each plugin will sleep for a random time within jitter before collecting.
  ## This can be used to avoid many plugins querying things like sysfs at the
  ## same time, which can have a measurable effect on the system.
  collection_jitter = "0s"

  ## Default flushing interval for all outputs. You shouldn't set this below
  ## interval. Maximum flush_interval will be flush_interval + flush_jitter
  flush_interval = "10s"
  ## Jitter the flush interval by a random amount. This is primarily to avoid
  ## large write spikes for users running a large number of telegraf instances.
  flush_jitter = "0s"

  ## By default or when set to "0s", precision will be set to the same
  ## timestamp order as the collection interval, with the maximum being 1s.
  ##   ie, when interval = "10s", precision will be "1s"
  ##       when interval = "250ms", precision will be "1ms"
  ## Precision will NOT be used for service inputs. It is up to each individual
  ## service input to set the timestamp at the appropriate precision.
  ## Valid time units are "ns", "us" (or "µs"), "ms", "s".
  precision = ""

  ## Logging configuration:
  ## Run telegraf with debug log messages.
  debug = false
  ## Run telegraf in quiet mode (error log messages only).
  quiet = false
  ## Specify the log file name. The empty string means to log to stderr.
  logfile = "/var/log/telegraf/telegraf.log"

  ## Override default hostname, if empty use os.Hostname()
  hostname = ""
  ## If set to true, do no set the "host" tag in the telegraf agent.
  omit_hostname = false

[[outputs.wavefront]]
  host = "CHEF_REPLACE_WAVEFRONT_HOST"
  port = 5879
  metric_separator = "."
  #source_override = ["source", "hostname", "snmp_host", "node_host"]
  convert_paths = true

#                            INPUT PLUGINS                                    #

[[inputs.cpu]]
  percpu = false
  totalcpu = true
  collect_cpu_time = false
  ## If true, compute and report the sum of all non-idle CPU states.
  report_active = true
  fieldpass = ["usage_active","usage_user","usage_system","usage_idle","usage_iowait"]

# Read metrics about disk usage by mount point
[[inputs.disk]]
  ## By default stats will be gathered for all mount points.
  ## Set mount_points will restrict the stats to only the specified mount points.
  # mount_points = ["/"]
  interval = "30m"
  ## Ignore mount points by filesystem type.
  ignore_fs = ["tmpfs", "devtmpfs", "devfs", "rootfs"]
  fieldpass = [ "total", "free" ]

# Read metrics about disk IO by device
[[inputs.diskio]]
  ## By default, telegraf will gather stats for all devices including
  ## disk partitions.
  ## Setting devices will restrict the stats to the specified devices.
  interval = "2m"
  devices = ["sd?", "hd?"]
  fieldpass = [ "read_time", "write_time", "read_bytes", "write_bytes" ]

# manually get cpu_speed
[[inputs.exec]]
  interval = "30m"
  commands = ['''sh -c "echo \"cpu speed=$(cat /proc/cpuinfo | grep 'cpu MHz' | head -n 1 | awk '{print $4;}')\""''']
  timeout = "5s"
  data_format = "influx"

# Get kernel statistics from /proc/stat
[[inputs.kernel]]
  fieldpass = [ "context_switches", "entropy_avail" ]

# Read metrics about memory usage
[[inputs.mem]]
  fieldpass = [ "available" ]  

# Read metrics about memory usage
[[inputs.mem]]
  interval = "30m"
  fieldpass = [ "total" ]  

# # Read metrics about network interface usage
[[inputs.net]]
  ## By default, telegraf gathers stats from any up interface (excluding loopback)
  ## Setting interfaces will tell it to gather these explicit interfaces,
  ## regardless of status.
  ##
  # interfaces = ["eth0"]
  ##
  ## On linux systems telegraf also collects protocol stats.
  ## Setting ignore_protocol_stats to true will skip reporting of protocol metrics.
  ##
  ignore_protocol_stats = true
  fieldpass = [ "bytes_sent", "bytes_recv", "packets_sent", "packets_recv" ]

# # Read TCP metrics such as established, time wait and sockets counts.
[[inputs.netstat]]
  fieldpass = [ "tcp_established", "tcp_time_wait", "tcp_close_wait", "tcp_listen" ]

# Get the number of processes and group them by status
[[inputs.processes]]
  fieldpass = [ "running", "total" ]

# Read metrics about swap memory usage
[[inputs.swap]]
  interval = "30m"
  fieldpass = [ "total" ]

# Read metrics about swap memory usage
[[inputs.swap]]
  fieldpass = [ "free", "in" ]
  
[[inputs.system]]
  fieldpass = [ "load1","load5","load15" ]

[[inputs.system]]
  interval = "30m"
  fieldpass = [ "uptime" ]

# manually get cpu count due to https://github.com/influxdata/telegraf/issues/5451
[[inputs.exec]]
  interval = "30m"
  commands = ['''sh -c "echo \"system n_cpus=$(cat /proc/cpuinfo | grep 'physical id' | wc -l)\""''']
  timeout = "5s"
  data_format = "influx"

#                            SERVICE INPUT PLUGINS                            #


# # Generic socket listener capable of handling multiple socket types.
[[inputs.socket_listener]]
  ## run more often to empty the read buffer
  interval = "5s"

  ## URL to listen on
  service_address = "tcp://localhost:5883"

  ## Maximum number of concurrent connections.
  ## Only applies to stream sockets (e.g. TCP).
  ## 0 (default) is unlimited.
  max_connections = 1024

  ## Read timeout.
  ## Only applies to stream sockets (e.g. TCP).
  ## 0 (default) is unlimited.
  read_timeout = "30s"

  ## Maximum socket buffer size in bytes.
  ## For stream sockets, once the buffer fills up, the sender will start backing up.
  ## For datagram sockets, once the buffer fills up, metrics will start dropping.
  ## Defaults to the OS default.
  read_buffer_size = 8388608

  ## Period between keep alive probes.
  ## Only applies to TCP sockets.
  ## 0 disables keep alive probes.
  ## Defaults to the OS configuration.
  keep_alive_period = "5m"

  ## Data format to consume.
  ## more about them here:
  data_format = "influx"

```

### System info:

Telegraf version 1-6.1 through 1.10.4-1.
CentOS 6 and 7.

### Steps to reproduce:

1. Start telegraf with round_interval=true
2. Confirm metrics received at desired interval and timestamps
3. Wait a few days to a few weeks
4. Notice drift in the timestamps of reported metrics, e.g. 54 seconds over a few weeks in my case.

### Expected behavior:
Even after several weeks, metrics continue to be reported at the rounded interval timestamps, e.g. at the top of each minute.

### Actual behavior:
After several weeks, the timestamps of the metrics drift, meaning that slightly more time has elapsed between metric points than the indicated interval.
### Additional info:
This seems to affect all the plugins I am using, regardless of their interval, except the metrics reported by the vsphere plugin, although the internal vsphere metrics are also affected (enabled by the 'internal' plugin). The vsphere plugin is defined in a separate file on only a couple of hosts, and I don't see any change in the agent settings as defined above.

### PR DESCRIPTION
**Pull request by danielnelson**
Fix several drift issues when using `round_interval` and use a rolling ticker for doing the full write flush.  In the future I would like to rework the output to only do full flushes after `flush_interval` seconds has elapsed without a batch write.  This will reduce unneeded calls to write but can be done separately.

It is possible that we will want to use the "RollingTicker" for inputs as well, since it would give you the best collection spread.  However, making the switch when using jitter requires changing the configuration in order to maintain the same collection pace, so I haven't switched in this pull request.

closes #5937
closes #5335
closes #4232
closes #4652

### Required for all PRs:

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -265,57 +265,45 @@ func (a *Agent) runInputs(
 			interval = input.Config.Interval
 		}
 
+		var ticker Ticker
+		if a.Config.Agent.RoundInterval {
+			ticker = NewAlignedTicker(startTime, interval, jitter)
+		} else {
+			ticker = NewUnalignedTicker(interval, jitter)
+		}
+		defer ticker.Stop()
+
 		acc := NewAccumulator(input, dst)
 		acc.SetPrecision(a.Precision())
 
 		wg.Add(1)
 		go func(input *models.RunningInput) {
 			defer wg.Done()
-
-			if a.Config.Agent.RoundInterval {
-				err := internal.SleepContext(
-					ctx, internal.AlignDuration(startTime, interval))
-				if err != nil {
-					return
-				}
-			}
-
-			a.gatherOnInterval(ctx, acc, input, interval, jitter)
+			a.gatherLoop(ctx, acc, input, ticker)
 		}(input)
 	}
-	wg.Wait()
 
+	wg.Wait()
 	return nil
 }
 
 // gather runs an input's gather function periodically until the context is
 // done.
-func (a *Agent) gatherOnInterval(
+func (a *Agent) gatherLoop(
 	ctx context.Context,
 	acc telegraf.Accumulator,
 	input *models.RunningInput,
-	interval time.Duration,
-	jitter time.Duration,
+	ticker Ticker,
 ) {
 	defer panicRecover(input)
 
-	ticker := time.NewTicker(interval)
-	defer ticker.Stop()
-
 	for {
-		err := internal.SleepContext(ctx, internal.RandomDuration(jitter))
-		if err != nil {
-			return
-		}
-
-		err = a.gatherOnce(acc, input, interval)
-		if err != nil {
-			acc.AddError(err)
-		}
-
 		select {
-		case <-ticker.C:
-			continue
+		case <-ticker.Elapsed():
+			err := a.gatherOnce(acc, input, ticker)
+			if err != nil {
+				acc.AddError(err)
+			}
 		case <-ctx.Done():
 			return
 		}
@@ -327,11 +315,8 @@ func (a *Agent) gatherOnInterval(
 func (a *Agent) gatherOnce(
 	acc telegraf.Accumulator,
 	input *models.RunningInput,
-	timeout time.Duration,
+	ticker Ticker,
 ) error {
-	ticker := time.NewTicker(timeout)
-	defer ticker.Stop()
-
 	done := make(chan error)
 	go func() {
 		done <- input.Gather(acc)
@@ -341,7 +326,7 @@ func (a *Agent) gatherOnce(
 		select {
 		case err := <-done:
 			return err
-		case <-ticker.C:
+		case <-ticker.Elapsed():
 			log.Printf("W! [agent] [%s] did not complete within its interval",
 				input.LogName())
 		}
@@ -514,10 +499,13 @@ func (a *Agent) runOutputs(
 			jitter = *output.Config.FlushJitter
 		}
 
+		ticker := NewRollingTicker(interval, jitter)
+		defer ticker.Stop()
+
 		wg.Add(1)
 		go func(output *models.RunningOutput) {
 			defer wg.Done()
-			a.flushLoop(ctx, startTime, output, interval, jitter)
+			a.flushLoop(ctx, output, ticker)
 		}(output)
 	}
 
@@ -542,10 +530,8 @@ func (a *Agent) runOutputs(
 // done.
 func (a *Agent) flushLoop(
 	ctx context.Context,
-	startTime time.Time,
 	output *models.RunningOutput,
-	interval time.Duration,
-	jitter time.Duration,
+	ticker Ticker,
 ) {
 	logError := func(err error) {
 		if err != nil {
@@ -558,44 +544,30 @@ func (a *Agent) flushLoop(
 	watchForFlushSignal(flushRequested)
 	defer stopListeningForFlushSignal(flushRequested)
 
-	// align to round interval
-	if a.Config.Agent.RoundInterval {
-		err := internal.SleepContext(
-			ctx, internal.AlignDuration(startTime, interval))
-		if err != nil {
-			return
-		}
-	}
-
-	// since we are watching two channels we need a ticker with the jitter
-	// integrated.
-	ticker := NewTicker(interval, jitter)
-	defer ticker.Stop()
-
 	for {
 		// Favor shutdown over other methods.
 		select {
 		case <-ctx.Done():
-			logError(a.flushOnce(output, interval, output.Write))
+			logError(a.flushOnce(output, ticker, output.Write))
 			return
 		default:
 		}
 
 		select {
 		case <-ctx.Done():
-			logError(a.flushOnce(output, interval, output.Write))
+			logError(a.flushOnce(output, ticker, output.Write))
 			return
-		case <-ticker.C:
-			logError(a.flushOnce(output, interval, output.Write))
+		case <-ticker.Elapsed():
+			logError(a.flushOnce(output, ticker, output.Write))
 		case <-flushRequested:
-			logError(a.flushOnce(output, interval, output.Write))
+			logError(a.flushOnce(output, ticker, output.Write))
 		case <-output.BatchReady:
 			// Favor the ticker over batch ready
 			select {
-			case <-ticker.C:
-				logError(a.flushOnce(output, interval, output.Write))
+			case <-ticker.Elapsed():
+				logError(a.flushOnce(output, ticker, output.Write))
 			default:
-				logError(a.flushOnce(output, interval, output.WriteBatch))
+				logError(a.flushOnce(output, ticker, output.WriteBatch))
 			}
 		}
 	}
@@ -605,12 +577,9 @@ func (a *Agent) flushLoop(
 // interval it fails to complete before.
 func (a *Agent) flushOnce(
 	output *models.RunningOutput,
-	timeout time.Duration,
+	ticker Ticker,
 	writeFunc func() error,
 ) error {
-	ticker := time.NewTicker(timeout)
-	defer ticker.Stop()
-
 	done := make(chan error)
 	go func() {
 		done <- writeFunc()
@@ -621,7 +590,7 @@ func (a *Agent) flushOnce(
 		case err := <-done:
 			output.LogBufferStatus()
 			return err
-		case <-ticker.C:
+		case <-ticker.Elapsed():
 			log.Printf("W! [agent] [%q] did not complete within its flush interval",
 				output.LogName())
 			output.LogBufferStatus()
--- a/agent/tick.go
+++ b/agent/tick.go
@@ -5,53 +5,260 @@ import (
 	"sync"
 	"time"
 
+	"github.com/benbjohnson/clock"
 	"github.com/influxdata/telegraf/internal"
 )
 
-type Ticker struct {
-	C          chan time.Time
-	ticker     *time.Ticker
-	jitter     time.Duration
-	wg         sync.WaitGroup
-	cancelFunc context.CancelFunc
+type empty struct{}
+
+type Ticker interface {
+	Elapsed() <-chan time.Time
+	Stop()
 }
 
-func NewTicker(
-	interval time.Duration,
-	jitter time.Duration,
-) *Ticker {
-	ctx, cancel := context.WithCancel(context.Background())
+// AlignedTicker delivers ticks at aligned times plus an optional jitter.  Each
+// tick is realigned to avoid drift and handle changes to the system clock.
+//
+// The ticks may have an jitter duration applied to them as an random offset to
+// the interval.  However the overall pace of is that of the interval, so on
+// average you will have one collection each interval.
+//
+// The first tick is emitted at the next alignment.
+//
+// Ticks are dropped for slow consumers.
+//
+// The implementation currently does not recalculate until the next tick with
+// no maximum sleep, when using large intervals alignment is not corrected
+// until the next tick.
+type AlignedTicker struct {
+	interval time.Duration
+	jitter   time.Duration
+	ch       chan time.Time
+	cancel   context.CancelFunc
+	wg       sync.WaitGroup
+}
 
-	t := &Ticker{
-		C:          make(chan time.Time, 1),
-		ticker:     time.NewTicker(interval),
-		jitter:     jitter,
-		cancelFunc: cancel,
+func NewAlignedTicker(now time.Time, interval, jitter time.Duration) *AlignedTicker {
+	return newAlignedTicker(now, interval, jitter, clock.New())
+}
+
+func newAlignedTicker(now time.Time, interval, jitter time.Duration, clock clock.Clock) *AlignedTicker {
+	ctx, cancel := context.WithCancel(context.Background())
+	t := &AlignedTicker{
+		interval: interval,
+		jitter:   jitter,
+		ch:       make(chan time.Time, 1),
+		cancel:   cancel,
 	}
 
+	d := t.next(now)
+	timer := clock.Timer(d)
+
 	t.wg.Add(1)
-	go t.relayTime(ctx)
+	go func() {
+		defer t.wg.Done()
+		t.run(ctx, timer)
+	}()
 
 	return t
 }
 
-func (t *Ticker) Stop() {
-	t.cancelFunc()
+func (t *AlignedTicker) next(now time.Time) time.Duration {
+	next := internal.AlignTime(now, t.interval)
+	d := next.Sub(now)
+	if d == 0 {
+		d = t.interval
+	}
+	d += internal.RandomDuration(t.jitter)
+	return d
+}
+
+func (t *AlignedTicker) run(ctx context.Context, timer *clock.Timer) {
+	for {
+		select {
+		case <-ctx.Done():
+			timer.Stop()
+			return
+		case now := <-timer.C:
+			select {
+			case t.ch <- now:
+			default:
+			}
+
+			d := t.next(now)
+			timer.Reset(d)
+		}
+	}
+}
+
+func (t *AlignedTicker) Elapsed() <-chan time.Time {
+	return t.ch
+}
+
+func (t *AlignedTicker) Stop() {
+	t.cancel()
 	t.wg.Wait()
 }
 
-func (t *Ticker) relayTime(ctx context.Context) {
-	defer t.wg.Done()
+// UnalignedTicker delivers ticks at regular but unaligned intervals.  No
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

File: agent/tick_test.go [hunk 1]
```diff
@@ -0,0 +1,229 @@
...
+	"testing"
+	"time"
+
+	"github.com/benbjohnson/clock"
+	"github.com/stretchr/testify/require"
+)
+
+var format = "2006-01-02T15:04:05.999Z07:00"
+
+func TestAlignedTicker(t *testing.T) {
```
ssoroka, line 15: Do you need to specifically test the system's clock changing? or did I miss that?

agent/tick_test.go
```diff
@@ -0,0 +1,229 @@
...
+		}
+		clock.Add(10 * time.Second)
+	}
+
+	require.Equal(t, expected, actual)
+}
+
+func TestAlignedTickerJitter(t *testing.T) {
+	interval := 10 * time.Second
+	jitter := 5 * time.Second
```
ssoroka, line 49: Go should use the same rand seed every time, so you should be able to test the specific jitter if desired

agent/tick_test.go
```diff
@@ -0,0 +1,229 @@
...
+	since := clock.Now()
+	until := since.Add(60 * time.Second)
+
+	ticker := newAlignedTicker(since, interval, jitter, clock)
+
+	last := since
+	for !clock.Now().After(until) {
+		select {
+		case tm := <-ticker.Elapsed():
+			require.True(t, tm.Sub(last) <= 15*time.Second)
```
ssoroka, line 61: test it's at least 5?

agent/tick_test.go
```diff
@@ -0,0 +1,229 @@
...
+	interval := 10 * time.Second
+	jitter := 5 * time.Second
+
+	clock := clock.NewMock()
+
+	ticker := newRollingTicker(interval, jitter, clock)
+	printDist(ticker, clock)
+}
+
+func printDist(ticker Ticker, clock *clock.Mock) {
```
ssoroka, line 203: I think we check a that the avg wait time is within expected bounds, and that we got the number of ticks expected.

agent/agent.go
```diff
@@ -265,57 +265,45 @@ func (a *Agent) runInputs(
 			interval = input.Config.Interval
 		}
 
+		var ticker Ticker
+		if a.Config.Agent.RoundInterval {
+			ticker = NewAlignedTicker(startTime, interval, jitter)
+		} else {
+			ticker = NewUnalignedTicker(interval, jitter)
+		}
+		defer ticker.Stop()
```
ssoroka, line 274: much cleaner, nice!

agent/tick.go
```diff
@@ -5,53 +5,260 @@ import (
...
 	for {
 		select {
-		case tm := <-t.ticker.C:
-			internal.SleepContext(ctx, internal.RandomDuration(t.jitter))
+		case <-ctx.Done():
+			ticker.Stop()
+			return
+		case <-ticker.C:
+			jitter := internal.RandomDuration(t.jitter)
+			sleep(ctx, jitter, clock)
```
ssoroka, line 169: doesn't handle returned error. next line pushes to chan even if context done

agent/tick_test.go
danielnelson: You didn't miss it.  I tried to do a test for this, but there doesn't appear to be a good way to do it.  Open to any ideas.

agent/tick_test.go
danielnelson: I think I need to redo `internal.RandomDuration` to support a custom rand instance first, I'm getting fluctuations as-is even if I reset the Seed when the test starts.  Anyway it would be nice to have the tests work if ran concurrently.

agent/tick_test.go
ssoroka: so basically, you need a time object where the [monotonic clock](https://golang.org/pkg/time/#hdr-Monotonic_Clocks) doesn't change but the wall clock does. I think you'd have to implement the clock.Clock interface, but replace the time.Time with your own time object where you can change the wall clock.

agent/tick_test.go
ssoroka: I think the problem is that internal.RandomDuration is depending on `crypto/rand`, which doesn't make a lot of sense. We really don't need cryptographic randomness for setting sleep timers. might want to change this to math/rand and it should start respecting the seed.

agent/agent.go
```diff
@@ -265,57 +265,45 @@ func (a *Agent) runInputs(
 			interval = input.Config.Interval
 		}
 
+		var ticker Ticker
+		if a.Config.Agent.RoundInterval {
+			ticker = NewAlignedTicker(startTime, interval, jitter)
+		} else {
+			ticker = NewUnalignedTicker(interval, jitter)
```
reimda, line 272: Just curious, when is unaligned useful?  I'm not sure I understand the issues we're fixing with this PR.

danielnelson: For inputs this PR is more about the aligned ticker. The unaligned ticker is what is used when the agent has `round_interval = false` set and is essentially how the current tickers work.

The main property today that makes them useful is that they run immediately on startup.  With the aligned ticker you need to wait for the next aligned interval.  This is mostly an issue with large intervals when you don't want to wait for the initial collection.

Another way it can be used is with many systems running Telegraf to get similar results to what you would get with jitter.  Since each Telegraf will collect at different intervals they will be spread out in their requests to the outputs.  In some respects it is improved over jitter because you will always have the full `interval` between gathers, as opposed to with jitter when you have `interval - jitter`.

The other part of this PR is the rolling ticker as the only type for the outputs.  This is an ticker that isn't aligned and always has at least `interval` time even when using jitter.  Since outputs are less sensitive to the write time we are able to switch to this style with a uniform distribution over time.

reimda: Ok, thanks for the context

PR #7394, opened 2020-04-23
Use nanosecond precision in prometheus input

praveenkumar-reddy:
### Required for all PRs:

### DIFF
--- a/plugins/inputs/prometheus/prometheus.go
+++ b/plugins/inputs/prometheus/prometheus.go
@@ -211,6 +211,7 @@ func (p *Prometheus) GetAllURLs() (map[string]URLAndAddress, error) {
 // Reads stats from all configured servers accumulates stats.
 // Returns one of the errors encountered while gather stats (if any).
 func (p *Prometheus) Gather(acc telegraf.Accumulator) error {
+	acc.SetPrecision(time.Nanosecond)
 	if p.client == nil {
 		client, err := p.createHTTPClient()
 		if err != nil {


danielnelson:
Thanks for the pull request, but we can't change the precision across the board here as it will override the level set in the agent.  I have a guess of what you might be trying to do, but I'm not sure, what would be best is to open a new issue describing what you are trying to solve.

praveenkumar-reddy:
> Thanks for the pull request, but we can't change the precision across the board here as it will override the level set in the agent. I have a guess of what you might be trying to do, but I'm not sure, what would be best is to open a new issue describing what you are trying to solve.


@danielnelson Thanks for taking time in reviewing the pull request. Here are few points.

1. Making change here will not override the agent precision , this change only reflects for prometheus input plugin.

2. I see you are already doing the same for docker_log inputs plugin. (https://github.com/influxdata/telegraf/blob/master/plugins/inputs/docker_log/docker_log.go#L206)

3. As we can't set the precision for inputs from the conf file, this is the only way to override the precision. If the interval is 5s , precision is being set to seconds without honouring the precision configuration in the agent section.
Example:
[agent]
  interval = "5s"
  precision = "ms" 

Though the precision is set to "ms", it's not honoured.

4. If it's problem like this , I can pass on a precision configuration from prometheus input and use that to set precision(If configuration is not set then leave to default).

danielnelson:
I'm sure we can come up with a solution here, but first can you describe the issue you are trying to address?  Is this related to collecting prometheus metrics that already have a timestamp set?

praveenkumar-reddy:
@danielnelson Yes, If metrics has already a timestamp and agent interval is set to "5s" , the timestamp precision is rounding off to seconds. We are loosing the milli seconds here. 

Example metric: ruby_shoryuken_jobs_total{job_name="DefaultWorker",queue_name="test"} 1 1587479736979

Loosing milli seconds out of timestamp is not a problem ,but if we have two metrics with the same tags in the same second, eventually we are going to lose one metric as timestamp is rounding off to seconds.

PR #7399, opened 2020-04-23
Update nvidia-smi README for Windows users

timhallinflux:
Updated for Windows users. Clarified toml edits and location of nvidia-smi.exe on Win10

### Required for all PRs:

### DIFF
--- a/plugins/inputs/nvidia_smi/README.md
+++ b/plugins/inputs/nvidia_smi/README.md
@@ -19,7 +19,7 @@ This plugin uses a query on the [`nvidia-smi`](https://developer.nvidia.com/nvid
 On Windows, `nvidia-smi` is generally located at `C:\Program Files\NVIDIA Corporation\NVSMI\nvidia-smi.exe`
 On Windows 10, you may also find this located here `C:\Windows\System32\nvidia-smi.exe`
 
-You'll need to escape the `\` within the `telegraf.config` like this: `C:\\Program Files\\NVIDIA Corporation\\NVSMI\\nvidia-smi.exe`
+You'll need to escape the `\` within the `telegraf.conf` like this: `C:\\Program Files\\NVIDIA Corporation\\NVSMI\\nvidia-smi.exe`
 
 ### Metrics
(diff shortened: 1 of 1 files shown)


plugins/inputs/nvidia_smi/README.md
```diff
@@ -17,6 +17,9 @@ This plugin uses a query on the [`nvidia-smi`](https://developer.nvidia.com/nvid
 #### Windows
 
 On Windows, `nvidia-smi` is generally located at `C:\Program Files\NVIDIA Corporation\NVSMI\nvidia-smi.exe`
+On Windows 10, you may also find located at `"C:\Windows\System32\nvidia-smi.exe`
+
+You'll need to escape the `\` within the `telegraf.config` like this: `C:\\Program Files\\NVIDIA Corporation\\NVSMI\\nvidia-smi.exe`
```
danielnelson, line 22: `telegraf.config` -> `telegraf.conf`

plugins/inputs/nvidia_smi/README.md
```diff
@@ -17,6 +17,9 @@ This plugin uses a query on the [`nvidia-smi`](https://developer.nvidia.com/nvid
 #### Windows
 
 On Windows, `nvidia-smi` is generally located at `C:\Program Files\NVIDIA Corporation\NVSMI\nvidia-smi.exe`
+On Windows 10, you may also find located at `"C:\Windows\System32\nvidia-smi.exe`
```
danielnelson, line 20: Double quote here is mismatched, we should probably remove the leading one to match the above line.

I see some information that this binary could be in `C:\Windows\System32\DriverStore\FileRepository\nvdm*\nvidia-smi.exe`.

https://stackoverflow.com/questions/57100015/how-do-i-run-nvidia-smi-on-windows

timhallinflux: removed the "

yeah, when Win10 pulls the binary down and lands it, they put it into the DriverStore\FileRepo.... but the install process lands it into System32 (which is in the default path for Windows).

Issue #7403, opened 2020-04-24
Allow to collect HTTP Headers thanks to http_response


essobedo:
### Proposal:

The idea would be to have the ability to collect also HTTP Headers from the response thanks to the plugin http_response.

### Current behavior:

So far, we only have access to very few headers which could be insufficient in particular use cases to properly treat the response since part of the data could be stored in the HTTP Headers of the response. 

### Desired behavior:

Provide the ability to configure a list of HTTP Headers to collect from the response. The value of those headers could be then stored as fields.

### Use case:

I use the plugin http_response to access to different http endpoints which works pretty well but there are some endpoints that provide important information using custom HTTP headers that I cannot collect due to the current limitation, because of that, unless there is a workaround, I could have to use something else which I would like to avoid of course since I like Telegraf.

essobedo:
A potential PR https://github.com/influxdata/telegraf/pull/7405

essobedo:
@danielnelson Hi, could you please at least tell me if this feature request and the corresponding PR make sense to you or not? Thanks in advance

PR #7405, opened 2020-04-24
7403 - Allow to collect HTTP Headers thanks to http_response

essobedo:
fixes https://github.com/influxdata/telegraf/issues/7403

### Required for all PRs:

### DIFF
--- a/plugins/inputs/http_response/http_response.go
+++ b/plugins/inputs/http_response/http_response.go
@@ -27,6 +27,7 @@ type HTTPResponse struct {
 	Body            string
 	Method          string
 	ResponseTimeout internal.Duration
+	HTTPHeaderTags  map[string]string `toml:"http_header_tags"`
 	Headers         map[string]string
 	FollowRedirects bool
 	// Absolute path to file with Bearer token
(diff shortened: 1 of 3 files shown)


danielnelson:
What do you think about making this look more similar to the new `http_header_tags` option we added to the `http_listener_v2` plugin?  We could name it `response_header_tags`, instead of fields it would produce a tag, and it would require the user to give a name for the tag.

essobedo:
@danielnelson thx for the feedback, here is a new version base on what has been done in `http_listener_v2` plugin

essobedo:
@danielnelson could you please tell me if this new version of the PR is what you had in mind?

plugins/inputs/http_response/http_response.go
```diff
@@ -239,6 +245,14 @@ func (h *HTTPResponse) httpGather(u string) (map[string]interface{}, map[string]
 	resp, err := h.client.Do(request)
 	response_time := time.Since(start).Seconds()
 
+	// Add the response headers
+	for headerName, measurementName := range h.HTTPHeaderTags {
```
danielnelson, line 249: Rename `measurementName` -> `tagKey` or just `tag`.  This will reduce confusion against the metric measurement/name.

essobedo: @danielnelson done, please check again

essobedo:
@danielnelson thanks :-)

PR #7408, opened 2020-04-24
feat: RunOnce / Serverless Telegraf

rawkode:
Follow-up to #6537

I can reduce some of the duplication between `Run` and `RunOnce`, but I don't want to put too much time into this unless it's closer to what was expected.

### Required for all PRs:

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -18,15 +18,13 @@ import (
 
 // Agent runs a set of plugins.
 type Agent struct {
-	Config  *config.Config
-	RunOnce bool
+	Config *config.Config
 }
 
 // NewAgent returns an Agent for the given Config.
-func NewAgent(config *config.Config, runOnce bool) (*Agent, error) {
+func NewAgent(config *config.Config) (*Agent, error) {
 	a := &Agent{
-		Config:  config,
-		RunOnce: runOnce,
+		Config: config,
 	}
 	return a, nil
 }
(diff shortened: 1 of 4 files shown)


agent/agent.go
```diff
@@ -272,6 +361,14 @@ func (a *Agent) runInputs(
 		go func(input *models.RunningInput) {
 			defer wg.Done()
 
+			if a.Config.Agent.RunOnce {
+				err := a.gatherOnce(acc, input, interval)
+				if err != nil {
+					acc.AddError(err)
```
ssoroka, line 367: I might move error handling inside of gatherOnce. It won't affect existing use of it and it'll clean up both callers.

agent/agent.go
```diff
@@ -142,6 +142,95 @@ func (a *Agent) Run(ctx context.Context) error {
...
+		return err
+	}
+
+	log.Printf("D! [agent] Connecting outputs")
+	err = a.connectOutputs(ctx)
+	if err != nil {
+		return err
+	}
+
+	inputC := make(chan telegraf.Metric, 100)
```
ssoroka, line 168: You'll want massive buffers here, and you're going to have a hard limit on how many metrics you can process this way. eg: a file input with 2 million lines is not going to work; it's going to fill this channel up before the rest of the chain can run. Since you're running the pieces one at a time. the `n`th + 1 write over `n` buffer size is going to block forever.

danielnelson:
@rawkode There are a couple other minor things to do in addition to what @ssoroka found and I have a PR open that is going to cause further work.  I'll sort out the remaining issues once #7390 is merged.

rawkode:
Happy to make changes, just keep me updated @danielnelson .

Thank you :)

danielnelson:
See #7474

PR #7412, opened 2020-04-26
plugins/parsers/json: Use same timestamp for all objects in arrays

presslab-us:
I am new to this, so please let me know if this is not appropriate.

My data is sent as a single JSON packet but the resulting measurements generated from the JSON array all had different timestamps.  This made it difficult to work with the data.  Because all objects in the JSON array arrive at the same time they should have the same timestamp.  With the same timestamp the data has the correlation that it should.

### Required for all PRs:

- [X] Associated README.md updated. (no changes needed)

### DIFF
--- a/plugins/parsers/json/parser.go
+++ b/plugins/parsers/json/parser.go
@@ -67,13 +67,13 @@ func New(config *Config) (*Parser, error) {
 	}, nil
 }
 
-func (p *Parser) parseArray(data []interface{}) ([]telegraf.Metric, error) {
+func (p *Parser) parseArray(data []interface{}, timestamp time.Time) ([]telegraf.Metric, error) {
 	results := make([]telegraf.Metric, 0)
 
 	for _, item := range data {
 		switch v := item.(type) {
 		case map[string]interface{}:
-			metrics, err := p.parseObject(v)
+			metrics, err := p.parseObject(v, timestamp)
 			if err != nil {
 				if p.strict {
 					return nil, err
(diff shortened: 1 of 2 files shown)


ssoroka:
Looks great, thanks so much for the contribution!

presslab-us:
> Looks great, do you think you can add a unit test too?

I did run `make test` and it passed that plugin:
```
ok  	github.com/influxdata/telegraf/plugins/parsers/json	(cached)
```
I figured the code coverage would be unchanged due to the nature of the change.  Was there some specific test you would like to see?

danielnelson:
Just something to check and document the behavior: parse a document that creates two metrics and ensure that the timestamp is the same between them.  The document could be something very simple like:
```json
[{"value": 1}, {"value":2}]
```

presslab-us:
Okay, I see what you are saying.  It looks like the unit test is in `config.go`?  The structure is not clear to me.  Is there some other test I can copy from to make this?  I have not worked with Golang before this.

danielnelson:
You would want to add the test to `plugins/parsers/json/parser_test.go`.  Some of the tests in this file are on the older side and aren't great examples of our current style, but something like this might be a good starting point.
```go
func TestShareTimestamp(t *testing.T) {
	parser, err := New(
		&Config{
			MetricName: "json",
		},
	)
	require.NoError(t, err)

	actual, err := parser.Parse([]byte(`[{"value": 1}, {"value":2}]`))
	require.NoError(t, err)

	// check that all metrics have the same timestamp
}
```

presslab-us:
Thanks a lot for your help.  I have added the test.

PR #7418, opened 2020-04-27
Add filepath processor plugin

### LINKED ISSUE #7342: [processor] New processor: filepath
## Feature Request

This issue proposes a new `filepath` processor plugin, that will handle transformations to file paths strings by exposing some of the functions implemented in https://golang.org/pkg/path/filepath/

### Proposal:

This proposal has its roots in https://github.com/influxdata/telegraf/issues/7290, while looking for an easy way to manipulat the `path` created by `logparser` input plugin.  In its related PR (https://github.com/influxdata/telegraf/pull/7301), @danielnelson suggested to migrate this functionality to a dedicated processor where more alike functions could be added (and also so that other plugins, for instance `tail` plugin, could also benefit from it.

The proposal is described by Daniel in https://github.com/influxdata/telegraf/pull/7301#issuecomment-614216516 : 

>  If we use the Go filepath package we can support these operations in a way that should work on Windows as well. I think it would be good to use some of the typical unix terminology for these operations:
```yaml
[[processors.filepath]]
  [[processors.filepath.basename]]
    tag = "path"
  [[processors.filepath.dirname]]
    tag = "path"
  [[processors.filepath.stem]] # name taken from python, removes one extension `foo.txt` -> `foo`
    tag = "path"
```
>There are other functions in the filepath package that might be helpful, but don't need to be added initially, such as Clean, Rel, ToSlash, VolumneName.

I am happy to work on this issue, will use this thread to discuss/agree on details and namings (for instance I guess we should apply this to either `tags` or `fields`; or maybe we should accept a list of tags, but I don´t see a strong use case for it so I wouldn´t go for it)

kir4h:

### Required for all PRs:

### Summary

This PR resolves #7342, adding a new processor plugin to include some path operations that are available through the `filepath` golang package.

First time dealing with a processor and did some refactors on the way, so any feedback is more than welcome.

closes #7342

### DIFF
--- a/plugins/processors/filepath/README.md
+++ b/plugins/processors/filepath/README.md
@@ -1,0 +1,207 @@
+# Filepath Processor Plugin
+
+The `filepath` processor plugin maps certain go functions from [path/filepath](https://golang.org/pkg/path/filepath/)
+onto tag and field values. Values can be modified in place or stored in another key.
+
+Implemented functions are:
+
+* [Base](https://golang.org/pkg/path/filepath/#Base) (accessible through `[[processors.filepath.basename]]`)
+* [Rel](https://golang.org/pkg/path/filepath/#Rel) (accessible through `[[processors.filepath.rel]]`)
+* [Dir](https://golang.org/pkg/path/filepath/#Dir) (accessible through `[[processors.filepath.dir]]`)
+* [Clean](https://golang.org/pkg/path/filepath/#Clean) (accessible through `[[processors.filepath.clean]]`)
+* [ToSlash](https://golang.org/pkg/path/filepath/#ToSlash) (accessible through `[[processors.filepath.toslash]]`)
+
+ On top of that, the plugin provides an extra function to retrieve the final path component without its extension. This
+ function is accessible through the `[[processors.filepath.stem]]` configuration item.
+
+Please note that, in this implementation, these functions are processed in the order that they appear above( except for
+`stem` that is applied in the first place).
+
+Specify the `tag` and/or `field` that you want processed in each section and optionally a `dest` if you want the result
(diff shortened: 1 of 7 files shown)


plugins/processors/filepath/README.md
```diff
@@ -0,0 +1,207 @@
...
+
+ ```toml
+[[processors.filepath]]
+   [[processors.filepath.dir]]
+     tag = "path"
+ ```
+
+### ToSlash
+
+The effects of this function are only noticeable on Windows platforms, because of the underlying golang implementation.
```
danielnelson, line 86: It occurs to me that this processor's behavior will depend on which arch is is running on, at least for some of these functions.  I suppose a user can always finish up with ToSlash in order to get consistent results?

kir4h: Yes, the behavior depends on the platform. This can be observed for instance in the `ToSlash` function (https://github.com/golang/go/blob/fffe6229637d1cf37c1400623bdeba57b4a787cb/src/path/filepath/path.go#L165-L170), where the OS separator is used as reference and nothing will be done in case of the separator being `/`. 

Regarding the consistent results yes, one could always use `ToSlash` in case of having different path styles when running on Windows (there is also a `FromSlash` one doing the reverse operation that could also be included in case the user prefers consistent results but with backslash notation).

kir4h:
> Looks great, the only thing that I think we ought to do is try to use the standard test functions instead of the new helpers. Just create the input and expected metrics with `testutil.MustMetric`, and compare using `testutil.RequireMetricsEqual`.

Perfect, I will  do the changes to use assertions with `testutil.RequireMetricsEqual`

PR #7426, opened 2020-04-28
Ignore fields with NaN or Inf floats in the JSON serializer

### LINKED ISSUE #7388: Prometheus plugin - series with +Inf , -Inf, NaN values fail for json output as its not a valid json for value
### Relevant telegraf.conf:

```toml




# Generic socket writer capable of handling multiple socket types.
[[outputs.socket_writer]]
  ## URL to connect to
  address = "tcp://0.0.0.0:25226"
  

  ## Optional TLS Config

  ## Period between keep alive probes.
  ## Only applies to TCP sockets.
  ## 0 disables keep alive probes.
  ## Defaults to the OS configuration.
  # keep_alive_period = "5m"

  ## Data format to generate.
  ## more about them here:
  data_format = "json"

[[inputs.prometheus]]
  interval = "1m"
  
  ## Scrape Kubernetes pods for the following prometheus annotations:
  ## - prometheus.io/scrape: Enable scraping for this pod
  ## - prometheus.io/scheme: If the metrics endpoint is secured then you will need to
  ##     set this to `https` & most likely set the tls config.
  monitor_kubernetes_pods = true
  metric_version = 2
  ## Kubernetes config file to create client from.
  # kube_config = "/path/to/kubernetes.config"
  ## Use bearer token for authorization. ('bearer_token' takes priority)
  bearer_token = "/var/run/secrets/kubernetes.io/serviceaccount/token"
  ## OR
  # bearer_token_string = "abc_123"
  ## Specify timeout duration for slower prometheus clients (default is 3s)
  response_timeout = "15s"

  ## Optional TLS Config
  tls_ca = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
  #tls_cert = /path/to/certfile
  # tls_key = /path/to/keyfile
  insecure_skip_verify = true


```

### Steps to reproduce:

1. Here is a gauge metric with -ve infinity as value (-Inf) - This is a correct value as per prometheus line format (+Inf , -Inf & NaN are valid values)

#### TYPE flink_taskmanager_job_task_operator_KafkaProducer_record_size_max gauge
flink_taskmanager_job_task_operator_KafkaProducer_record_size_max{job_id="40c668ac8435480261a4c2afb8f88deb",task_id="c8e26e1de46a579b823e994ab7ebd862",task_attempt_id="f099f1bb3403a8762dfc12af3a580c29",host="flink_tm_0",operator_id="fd06000fa5e590dc6d2474670a0b4c2c",operator_name="Sink:_Unnamed",task_name="KeyedProcess_____Sink:_Unnamed__Sink:_Unnamed_",task_attempt_num="0",job_name="pulse_TictracService",tm_id="7f41cff8bb3199807ca9b9bb164eed23",subtask_index="2",} -Inf

2. Stitch the prometheus input with json output format, and you will see the below error -
2020-04-21T01:01:05Z E! [agent] Error writing to output [socket_writer]: json: unsupported value: -Inf

### Expected behavior:
No error when serializing as JSON
Looks to me we are parsing values using float64() rather than parseFloat(), which is the root cause.

danielnelson:
With this change it is possible to have a metric with no encodeable fields.  I debated having the serializer skip the metric and potentially return an empty buffer, this would be similar to what the influx serializer does, but decided against it in this case.  With line protocol, it isn't possible to write a line without fields, but with JSON it is still possible to write the remaining data and guarantees that something will be written.  Since this data format is often used for debugging I think this will be the best behavior.

closes #7388

### Required for all PRs:

### DIFF
--- a/plugins/serializers/json/json.go
+++ b/plugins/serializers/json/json.go
@@ -2,6 +2,7 @@ package json
 
 import (
 	"encoding/json"
+	"math"
 	"time"
 
 	"github.com/influxdata/telegraf"
@@ -49,8 +50,26 @@ func (s *serializer) SerializeBatch(metrics []telegraf.Metric) ([]byte, error) {
 
 func (s *serializer) createObject(metric telegraf.Metric) map[string]interface{} {
 	m := make(map[string]interface{}, 4)
-	m["tags"] = metric.Tags()
-	m["fields"] = metric.Fields()
+
+	tags := make(map[string]string, len(metric.TagList()))
+	for _, tag := range metric.TagList() {
+		tags[tag.Key] = tag.Value
+	}
+	m["tags"] = tags
+
+	fields := make(map[string]interface{}, len(metric.FieldList()))
+	for _, field := range metric.FieldList() {
+		switch fv := field.Value.(type) {
+		case float64:
+			// JSON does not support these special values
+			if math.IsNaN(fv) || math.IsInf(fv, 0) {
+				continue
+			}
+		}
+		fields[field.Key] = field.Value
+	}
+	m["fields"] = fields
+
 	m["name"] = metric.Name()
 	m["timestamp"] = metric.Time().UnixNano() / int64(s.TimestampUnits)
 	return m
--- a/plugins/serializers/json/json_test.go
+++ b/plugins/serializers/json/json_test.go
@@ -2,14 +2,15 @@ package json
 
 import (
 	"fmt"
+	"math"
 	"testing"
 	"time"
 
-	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
-
 	"github.com/influxdata/telegraf"
 	"github.com/influxdata/telegraf/metric"
+	"github.com/influxdata/telegraf/testutil"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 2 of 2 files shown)


plugins/serializers/json/json.go
```diff
@@ -49,8 +50,26 @@ func (s *serializer) SerializeBatch(metrics []telegraf.Metric) ([]byte, error) {
...
 func (s *serializer) createObject(metric telegraf.Metric) map[string]interface{} {
 	m := make(map[string]interface{}, 4)
-	m["tags"] = metric.Tags()
-	m["fields"] = metric.Fields()
+
+	tags := make(map[string]string, len(metric.TagList()))
+	for _, tag := range metric.TagList() {
+		tags[tag.Key] = tag.Value
+	}
+	m["tags"] = tags
```
reimda, line 58: Just curious why this is better than before.  Is it better not to refer to the same map?  Kind of a deep copy but what's the benefit?

danielnelson: I'd like to remove the `metric.Tags()` and `Fields()` method the interface, they are marked as deprecated now, so I just took the liberty to do stop using them here.  We don't use a map in the metric implementation anymore, since a list is more memory and GC friendly in most cases.  Both of these functions make a new map and are basically equivalent to what I've done here.

reimda: makes sense, thanks

PR #7430, opened 2020-04-29
Add RestartSec to service file to delay restarts

### LINKED ISSUE #7429: systemd service telegraf has issues when system is low on memory.
### Proposal:

in telegraf.service set something like
```
RestartSec=5
```

### Current behavior:

if system is low on memory Telegraf fails because the system is near out of memory. Service crashes and immediately seems to restart and crash again and again ~5 times. Before the OOM killer can kill the bad process.
```
Apr 28 03:20:02 cpu147 telegraf[64455]:         /usr/local/go/src/runtime/proc.go:543 +0x72 fp=0x7ffd5c39fe60 sp=0x7ffd5c39fe08 pc=0x433e12
Apr 28 03:20:02 cpu147 telegraf[64455]: runtime.rt0_go(0x7ffd5c39fe98, 0x5, 0x7ffd5c39fe98, 0x0, 0x7fa77cee70ca, 0x5, 0x7ffd5c3a1ea5, 0x7ffd5c3a1eb
Apr 28 03:20:02 cpu147 telegraf[64455]:         /usr/local/go/src/runtime/asm_amd64.s:214 +0x125 fp=0x7ffd5c39fe68 sp=0x7ffd5c39fe60 pc=0x45f3f5
Apr 28 03:20:02 cpu147 systemd[1]: telegraf.service: Main process exited, code=exited, status=2/INVALIDARGUMENT
Apr 28 03:20:02 cpu147 systemd[1]: telegraf.service: Failed with result 'exit-code'.
Apr 28 03:20:02 cpu147 systemd[1]: telegraf.service: Service hold-off time over, scheduling restart.
Apr 28 03:20:02 cpu147 systemd[1]: telegraf.service: Scheduled restart job, restart counter is at 6.
Apr 28 03:20:02 cpu147 systemd[1]: Stopped The plugin-driven server agent for reporting metrics into InfluxDB.
Apr 28 03:20:02 cpu147 systemd[1]: telegraf.service: Start request repeated too quickly.
Apr 28 03:20:02 cpu147 systemd[1]: telegraf.service: Failed with result 'exit-code'.
```

### Desired behavior:

After telegraf service crash wait some time so the OOM killer can do it's thing so when telegraf restarts there is free memory.

### Use case:

Work for an university that has many big servers. Some servers have multiple users and others allow researches to have exclusive access. Sadly the odd user when working with large data sets or just testing bad code hits the OOM killer. I see this too often :(

Like was already said when the system is low on memory things start to crash... Telegraf is one of them. Sadly the Telegraf service restarts too aggressively and ends up giving up after too many restarts.

danielnelson:
closes #7429

### Required for all PRs:

### DIFF
--- a/scripts/telegraf.service
+++ b/scripts/telegraf.service
@@ -10,6 +10,7 @@ ExecStart=/usr/bin/telegraf -config /etc/telegraf/telegraf.conf -config-director
 ExecReload=/bin/kill -HUP $MAINPID
 Restart=on-failure
 RestartForceExitStatus=SIGPIPE
+RestartSec=5
 KillMode=control-group
 
 [Install]


scripts/telegraf.service
```diff
@@ -10,6 +10,7 @@ ExecStart=/usr/bin/telegraf -config /etc/telegraf/telegraf.conf -config-director
 ExecReload=/bin/kill -HUP $MAINPID
 Restart=on-failure
 RestartForceExitStatus=SIGPIPE
+RestartSec=5
```
reimda, line 13: I have some concerns about this.

Restarting under normal conditions happens much more frequently than restarting under low memory conditions.  Delaying the restart will mean lost data in the common case for no reason.  It's not easy to weigh a potential benefit in an edge case vs a drawback in a much more common case, but I think this edge case is so small that it's not worth it.

Also, the choice of five seconds for the delay is completely arbitrary and not a general fix for restarting under low memory conditions for all machines.  It may be effective for the machines in the issue but it could be too short in other OOM conditions, such as when a lot of swap is involved.

Since this is a drawback under normal conditions and it's also not a general solution for dealing with out of memory state, I think it's better to leave it out.

We might suggest to the issue opener to use cgroup to limit memory usage of the user processes that might take up too much memory.  It's typically more effective to prevent the out of memory condition than try to clean up afterward.  If they don't want to do that, they can still add RestartSec=5 to telegraf.service after telegraf is installed.

danielnelson:
@steverweber I think I'm in agreement with @reimda's review, since this is the general Telegraf service file we should stick to the systemd default unless there are recommendations from Debian/Redhat that this be set differently.

If you want to change this locally I recommend adding the option to an override file to ease package updates, this should open the file in the right spot for your system:
```
sudo systemctl edit telegraf
```

PR #7434, opened 2020-04-29
Docker time stamps as metric time stamps

i-prudnikov:
### Required for all PRs:

- [] Associated README.md updated.

This PR comes from #7068, the first part - Use docker timestamps as metric timestamp.
Logs are streamed with timestamps, time stamps parsed from log message and added as metric timestamp. No custom read logic implemented, as stdcopy.StdCopy appears to remove headers from messages while demultiplexing that is a perfect fit.

Readme is not updated as nothing was changed in the plugin configuration interface.

### DIFF
--- a/plugins/inputs/docker_log/docker_log.go
+++ b/plugins/inputs/docker_log/docker_log.go
@@ -2,8 +2,10 @@ package docker_log
 
 import (
 	"bufio"
+	"bytes"
 	"context"
 	"crypto/tls"
+	"fmt"
 	"io"
 	"strings"
 	"sync"
@@ -67,6 +69,9 @@ const (
 	// docker code:
 	// https://github.com/moby/moby/blob/master/daemon/logger/copier.go#L21
 	maxLineBytes = 16 * 1024
+
+	//Length of docker time stamp in every log message
+	dockerTimeStampLength = 30
 )
 
 var (
@@ -287,7 +292,7 @@ func (d *DockerLogs) tailContainerLogs(
 	logOptions := types.ContainerLogsOptions{
 		ShowStdout: true,
 		ShowStderr: true,
-		Timestamps: false,
+		Timestamps: true,
 		Details:    false,
 		Follow:     true,
 		Tail:       tail,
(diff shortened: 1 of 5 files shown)


plugins/inputs/docker_log/docker_log.go
```diff
@@ -311,6 +316,37 @@ func (d *DockerLogs) tailContainerLogs(
 	}
 }
 
+func parseLogEntry(rawMessage []byte, acc *telegraf.Accumulator, containerID *string) (time.Time, []byte) {
+	var message []byte
+	switch length := len(rawMessage); {
```
danielnelson, line 321: In this function use `bytes.SplitN(line, []byte(" "), 2)`. if there aren't 2 splits return an error, otherwise you are good to parse and return (time, message, err)

plugins/inputs/docker_log/docker_log.go
```diff
@@ -329,21 +365,25 @@ func tailStream(
...
 	var err error
-	var message string
+	var message []byte
+	var messageTs time.Time
+
 	for {
-		message, err = r.ReadString('\n')
+		message, err = r.ReadBytes('\n')
+		//Parsing timestamp and message
+		messageTs, message = parseLogEntry(message, &acc, &containerID)
```
danielnelson, line 374: Change this function to support returning an error, if an error is returned you can `acc.AddError(err)` and skip processing the line.

plugins/inputs/docker_log/docker_log_test.go
```diff
@@ -52,7 +52,12 @@ type Response struct {
 func (r *Response) Close() error {
 	return nil
 }
-
+func getFirst(p ...interface{}) interface{} {
```
danielnelson, line 55: Define this like:
```go
func MustParse(layout, value string) time.Time {
    tm, err := time.Parse(layout, value)
    if err != nil {
        panic(err)
    }
    return tm
}
```

plugins/inputs/docker_log/docker_log.go
i-prudnikov: Looks pretty good,  but I think theres is one edge case here,  when the message is empty, I mean that it is only <timestamp> without any trailing whitespace, it is not an error. So in this case we need to return ts, empty message, and nil error. The caller function will make a decision either emit empty message as a telegraf metric or not.
Moreover,  I think that we should emit empty messages, as that is what container produced, but right now,  this is thrown away. What is your opinion?

danielnelson: Let's emit the empty string as the message.

i-prudnikov: Updated as discussed

danielnelson:
Opened a PR against your fork, can you take a look: https://github.com/devopsext/telegraf/pull/1

plugins/inputs/docker_log/docker_log.go
```diff
@@ -316,42 +313,23 @@ func (d *DockerLogs) tailContainerLogs(
...
 
-	timeStamp, err = time.Parse(time.RFC3339Nano, string(ts))
+	ts, err := time.Parse(time.RFC3339Nano, tsString)
 	if err != nil {
-		return time.Now(), message, fmt.Errorf("can't parse timestamp from string %q, reason: %v", ts, err)
+		return time.Time{}, "", fmt.Errorf("error parsing timestamp %q: %v", tsString, err)
 	}
 
-	return timeStamp, message, nil
+	return ts, message, nil
```
i-prudnikov, line 332: @danielnelson , This cause copying the content of `message`, initially the function was returning `[]byte` that is cheaper from this perspective (especially respecting the possible long lines of output from multiple containers). Moreover, later, the message is trimmed with  `TrimRightFunc(message, unicode.IsSpace)`, that again involve building new string. Should we revert back this to using `[]bytes` until calling `acc.AddFields`? Do you agree?

danielnelson: Let's move TrimRightFunc into this function and apply the bytes version of it before converting the message to a string.  The conversion/copy must happen at some point, its still best to do it here, but we just want to do it once.

i-prudnikov: Adjusted as discussed

PR #7435, opened 2020-04-29
Set RestartSec to keep systemd service running after consecutive crash

steverweber:
RestartSec defaults to an aggressive 100ms. When Telegraf fails due to an out of memory issue (caused by another process on a system) We need systemd to give the OOM killer time to handle the issue before restarting Telegraf.

Without waiting the Telegraf service will crash multiple times in a short span of ~1second. This puts the service into a failed state that will not try restarting.

https://github.com/influxdata/telegraf/issues/7429

### Required for all PRs:


steverweber:
opps this was already done in: https://github.com/influxdata/telegraf/pull/7430

PR #7439, opened 2020-04-29
Handle multiple metrics with the same timestamp in dedup plugin

### LINKED ISSUE #7343: uptime (and uptime_format) from system-plugin always suppressed when dedup active
### Relevant telegraf.conf: 

```
[[inputs.system]]
  ## Uncomment to remove deprecated metrics.
  # fielddrop = ["uptime_format"]
[[processors.dedup]]
  ## Maximum time to suppress output
  dedup_interval = "600s"
```

### Steps to reproduce:

1.  Activate/inactivate dedup-section (new since 1.14.0)

### Expected behavior:

Also with active dedup-section uptime-metric from system-plugin should be send to influx-db (can't imagine duplicates on uptime ;-)

### Actual behavior:

With active dedup-section uptime-metric is not send to influx-db

### Additional info:

Other metrics from system (load, cpu, users) are ok.

reimda:
resolves #7343

### DIFF
--- a/plugins/processors/dedup/dedup_test.go
+++ b/plugins/processors/dedup/dedup_test.go
@@ -1,9 +1,10 @@
 package dedup
 
 import (
-	"github.com/stretchr/testify/require"
 	"testing"
 	"time"
+
+	"github.com/stretchr/testify/require"
 
 	"github.com/influxdata/telegraf"
 	"github.com/influxdata/telegraf/internal"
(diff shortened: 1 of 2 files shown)


plugins/processors/dedup/dedup_test.go
```diff
@@ -152,3 +153,42 @@ func TestCacheShrink(t *testing.T) {
...
+	var in telegraf.Metric
+	var out []telegraf.Metric
+
+	in, _ = metric.New("metric",
+		map[string]string{"tag": "value"},
+		map[string]interface{}{"foo": 1}, // field
+		now,
+	)
+	out = dedup.Apply(in)
+	require.Equal(t, []telegraf.Metric{in}, out) // pass
```
danielnelson, line 169: May want to use `testutil.RequireMetricsEqual` here for better error display.  (I'm not sure if this is testing the pointer or the value)

plugins/processors/dedup/dedup_test.go
```diff
@@ -152,3 +153,42 @@ func TestCacheShrink(t *testing.T) {
...
+
+	in, _ = metric.New("metric",
+		map[string]string{"tag": "value"},
+		map[string]interface{}{"bar": 1}, // different field
+		now,
+	)
+	out = dedup.Apply(in)
+	require.Equal(t, []telegraf.Metric{in}, out) // pass
+
+	in, _ = metric.New("metric",
```
danielnelson, line 179: Should we split this into two tests?

PR #7446, opened 2020-05-01
Fix gzip support in socket_listener with tcp sockets

### LINKED ISSUE #7377: `socket_listener` unable to decode compressed TCP payloads
### Relevant telegraf.conf:

```toml
[[inputs.socket_listener]]
  service_address = "tcp://:8096"
  data_format = "influx"
  content_encoding = "gzip"
```

### System info:
Telegraf: Telegraf 1.14.0 (git: HEAD fefd7ff1)
OS: Windows Server 2016 Standard - Version 1607 build 14393.3630

Telegraf installed as a Windows Service by `telegraf.exe`

### Steps to reproduce:

1. Start a Telegraf instance listening with `socket_listener` with the provided configuration to listen for TCP connections with `gzip` compressed payloads
2. Write compressed line protocol messages to the TCP socket that Telegraf is listening on
    1. This can be done with the build artifact from [abandoned PR 7355](https://github.com/influxdata/telegraf/pull/7355)

### Expected behavior:

Telegraf can decode and consume line protocol messages compressed with gzip on a TCP socket

### Actual behavior:

Telegraf unable to decode and consume line protocol messages compressed with gzip on a TCP socket

### Additional info:

Log file messages:
```
2020-04-20T20:13:46Z I! [inputs.socket_listener] Listening on tcp://[::]:8096
2020-04-20T20:14:50Z E! [inputs.socket_listener] Unable to decode incoming line: unexpected EOF
2020-04-20T20:14:50Z E! [inputs.socket_listener] Unable to decode incoming line: gzip: invalid header
...
```
[Telegraf_CompressedTCP.pcapng.zip](https://github.com/influxdata/telegraf/files/4505991/Telegraf_CompressedTCP.pcapng.zip)

danielnelson:
closes #7377

### Required for all PRs:

### DIFF
--- a/internal/content_coding.go
+++ b/internal/content_coding.go
@@ -8,19 +8,6 @@ import (
 	"io"
 )
 
-// NewStreamContentEncoder returns a writer that will encode the stream
-// according to the encoding type.
-func NewStreamContentEncoder(encoding string, w io.Writer) (io.Writer, error) {
-	switch encoding {
-	case "gzip":
-		return NewGzipWriter(w)
-	case "identity", "":
-		return w, nil
-	default:
-		return nil, errors.New("invalid value for content_encoding")
-	}
-}
-
 // NewStreamContentDecoder returns a reader that will decode the stream
 // according to the encoding type.
 func NewStreamContentDecoder(encoding string, r io.Reader) (io.Reader, error) {
@@ -32,30 +19,6 @@ func NewStreamContentDecoder(encoding string, r io.Reader) (io.Reader, error) {
 	default:
 		return nil, errors.New("invalid value for content_encoding")
 	}
-}
-
-// GzipWriter is similar to the gzip.Writer but flushes after each call to
-// write, this ensure that the written data can be decompressed on the read
-// size immediately.
-type GzipWriter struct {
-	w io.Writer
-	z *gzip.Writer
-}
-
-func NewGzipWriter(w io.Writer) (io.Writer, error) {
-	return &GzipWriter{w: w, z: new(gzip.Writer)}, nil
-}
-
-func (w *GzipWriter) Write(p []byte) (int, error) {
-	w.z.Reset(w.w)
-
-	n, err := w.z.Write(p)
-	if err != nil {
-		return n, err
-	}
-
-	err = w.z.Close()
-	return n, err
 }
 
 // GzipReader is similar to gzip.Reader but reads only a single gzip stream per read.
--- a/internal/content_coding.go
+++ b/internal/content_coding.go
@@ -1,18 +1,115 @@
 package internal
 
 import (
+	"bufio"
 	"bytes"
 	"compress/gzip"
 	"errors"
 	"io"
 )
 
+// NewStreamContentEncoder returns a writer that will encode the stream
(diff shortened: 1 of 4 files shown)


internal/content_coding.go
```diff
@@ -1,18 +1,115 @@
...
+	"bufio"
 	"bytes"
 	"compress/gzip"
 	"errors"
 	"io"
 )
 
+// NewStreamContentEncoder returns a writer that will encode the stream
+// according to the encoding type.
+func NewStreamContentEncoder(encoding string, w io.Writer) (io.Writer, error) {
```
reimda, line 13: This isn't used outside of tests.  Are you planning to use it later?

internal/content_coding.go
```diff
@@ -1,18 +1,115 @@
...
+	w io.Writer
+	z *gzip.Writer
+}
+
+func NewGzipWriter(w io.Writer) (io.Writer, error) {
+	return &GzipWriter{w: w, z: new(gzip.Writer)}, nil
+}
+
+func (w *GzipWriter) Write(p []byte) (int, error) {
+	w.z.Reset(w.w)
```
reimda, line 50: I'm not sure why we start a new gzip stream for each write.  Doesn't this prevent the gzip algorithm from getting enough state to compress well?

plugins/inputs/socket_listener/socket_listener.go
```diff
@@ -111,7 +111,12 @@ func (ssl *streamSocketListener) read(c net.Conn) {
 	defer ssl.removeConnection(c)
 	defer c.Close()
 
-	scnr := bufio.NewScanner(c)
+	decoder, err := internal.NewStreamContentDecoder(ssl.ContentEncoding, c)
+	if err != nil {
+		ssl.Log.Error("Read error: %v", err)
+	}
+
+	scnr := bufio.NewScanner(decoder)
```
reimda, line 119: We always wrap with decoder here.  Wrapping makes sense to me for gzip, but not for identity.  If identity, why not pass in c and not wrap at all?

internal/content_coding.go
danielnelson: I'll remove it for now, originally was planning to use it in the socket_writer but ended up leaving it with the batching version.  Potentially we could use something like this to reduce memory usage.

internal/content_coding.go
danielnelson: It would if the write is small, my idea here was that it would be nice to be able to read each batch.  I'm going to remove the Writer completely for now since it isn't used and we can sort this out later.

plugins/inputs/socket_listener/socket_listener.go
danielnelson: In the 'identity' case, internal.NewStreamContentDecoder is just returning the reader unwrapped.  I thought it would be a bit nicer to push the switch logic into this factory function.

danielnelson:
I removed the GzipWriter related code, didn't end up using it for now so we can revive and complete it when we have the need.

m4ce:
@danielnelson - I tested this PR and I am seeing issues. I see random TCP disconnects when sending data.

PR #7455, opened 2020-05-04
Remove debug fields from spunkmetric serializer

### LINKED ISSUE #7448: [Splunkmetric Serializer Plugin] Unnecessary fields generated: "config:hecRouting" and "config:multiMetric"
### Actual behavior:
Splunkmetic serializer plugin generates two unnecessary fields for each event: "config:hecRouting" and "config:multiMetric". 
These are indexed in Splunk but not really needed.

Interestingly in the output example the fields are output with underscore prefix:
https://github.com/influxdata/telegraf/tree/master/plugins/serializers/splunkmetric

```
{
  "time": 1572469920,
  "event": "metric",
  "host": "mono.local",
  "fields": {
    "_config_hecRouting": false,
    "_config_multiMetric": true,
    "class": "osx",
    "cpu": "cpu0",
    "metric_name:telegraf.cpu.usage_guest": 0,
    "metric_name:telegraf.cpu.usage_guest_nice": 0,
  ...
  }
}
```

That would be ok, because fields with underscore prefix are not indexed in Splunk. 
https://docs.splunk.com/Documentation/Splunk/8.0.3/Metrics/BestPractices
"...start with an underscore ( _ ). Such dimensions will not be indexed."
Still, the data has to be written and trasferred to Splunk, only to be removed by the indexer engine. 

Unfortunately the fields are being written to output without underscore (tested with file output, _config_hecRouting=false and _config_multiMetric=true).

In splunkmetric.go I simply commented out the following two lines and now it works correctly:

```
//commonObj["config:hecRouting"] = s.HecRouting
//commonObj["config:multiMetric"] = s.SplunkmetricMultiMetric
```

### Expected behavior:
The fields are omitted in the events or are written with underscore prefix.

danielnelson:
closes #7448

### Required for all PRs:

### DIFF
--- a/plugins/serializers/splunkmetric/README.md
+++ b/plugins/serializers/splunkmetric/README.md
@@ -38,8 +38,6 @@ you can send all of your CPU stats in one JSON struct, an example event looks li
   "event": "metric",
   "host": "mono.local",
   "fields": {
-    "_config_hecRouting": false,
-    "_config_multiMetric": true,
     "class": "osx",
     "cpu": "cpu0",
     "metric_name:telegraf.cpu.usage_guest": 0,
--- a/plugins/serializers/splunkmetric/splunkmetric.go
+++ b/plugins/serializers/splunkmetric/splunkmetric.go
@@ -179,12 +179,7 @@ func (s *serializer) createObject(metric telegraf.Metric) (metricGroup []byte, e
 	// The tags are common to all events in this timeseries
 	commonTags := CommonTags{}
 
-	commonObj := map[string]interface{}{}
-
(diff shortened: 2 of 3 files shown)


ronnocol:
Thanks. Looks good to me. Since Splunk doesn't charge for indexing dimensions/tags, I didn't see a problem with leaving config breadcrumbs in the data.

danielnelson:
Good point about being able to add them back, hopefully this won't be a disruptive change for anyone but there is a workaround if it is.  Technically this is a breaking change, so I'll add it as a release note as well.

Issue #6968, opened 2020-02-01
XML parser


M0rdecay:
## Feature Request
I would like to contribute to the Telegraf as an xml parser for input plugins.
Please rate the [proposed solution](https://github.com/M0rdecay/telegraf/tree/xml_parser/plugins/parsers/xml) and let it know if it suits you.
If you make a positive decision, I will prepare unit tests for this plugin.

This should close issue [1758](https://github.com/influxdata/telegraf/issues/1758)

Waiting for your feedback!

M0rdecay:
Friends, do I expect an answer from you?

danielnelson:
@M0rdecay Thanks for the link, unfortunately due to time constraints I don't have time to evaluate the parser right now, I'll have to take a look later.

M0rdecay:
Thanks for the answer! Of course, I will wait for your decision. If you find the code inappropriate, please indicate the shortcomings.

M0rdecay:
@danielnelson, hello! Have you already seen the proposed solution? We waiting for your decision.

M0rdecay:
@danielnelson? :(

If in the current state the parser looks poorly written or poorly functional, please let us know what you would like to see in the solution.

I will explain why this is important for us - in the current form, the parser covers our needs, but we are not ready to use it in a productive environment so as not to branch off from you. We would prefer to improve the proposed solution to meet the required level.

sjwang90:
Hi @M0rdecay, sorry for the delay but we're  pretty backlogged with the new features for the upcoming Telegraf release.  We'll definitely take a look at your proposed solution soon.

cmakaloclifford:
All respect for you guys being backed up with lots of other stuff to do.
Just wanted to add a vote for this feature. We would really like to have a plugin for parsing XML outputs from our Wowza servers, so I hope we could see that implemented some time soon. :)

M0rdecay:
Comrades, we are waiting for your assessment.
It seems that this functionality can be useful not only to me, but also to other users.

yurividal:
Any updates on this? I really need this feature.

M0rdecay:
@yurividal hello!

I need your vote here - https://github.com/influxdata/telegraf/pull/8121/ and here - https://influxcommunity.slack.com/archives/CH99HUH8V/p1601645481073800

We want to show that people need an XML parser.

sjwang90:
[...]

PR #7460, opened 2020-05-05
Adds a new parser for XML data

M0rdecay:
### Required for all PRs:

This is a draft of XML parser (related issues #6968, #1758). Please review the proposed solution and let us know if improvements are required.


M0rdecay:
@danielnelson, hello!
I am sorry for disturbing, is there a chance this PR will end up in 1.16? Should i fix merge conflicts?

danielnelson:
I think what I would like to see on this pull request is more evidence that it will be useful across different documents.  It would be useful if we could get more community feedback on if this plugin was or was not sufficient for XML processing.

It is always helpful to resolve any merge conflicts, this way we know the build works and is able to be merged.  If we see conflicts or test failures, we sometimes assume it is a work in progress.

I'm going to be stepping away from being the Telegraf maintainer; @reimda and @ssoroka will be taking over and will help you further.

M0rdecay:
@ssoroka, hello!
Merge conflicts resolved. What should I do next?

ssoroka:
I think this runs into the same problems as the json parser. It's fine for pulling out a specific value, but it's difficult to iterate over parts of the document to build one or more metrics for any document that's non-trivial.

I'm not really sure what the answer to this is. I don't think xpath really helps here. I'd like to do a bit more research to find a flexbile approach for both.

M0rdecay:
You are right, I thought about it too. 

Maybe we should try something like this:
```toml
[[inputs.file]]
...
  data_format = "xml"

  [[inputs.file.xml]]
    query = ...

  [[inputs.file.xml]]
    query = ...
```

This way we could analyze the same document several times. 
At first glance, this shouldn't be difficult to implement.

ssoroka:
that definitely helps, but it's still not clear how to go to a specific path and output one metric per element in the array, for example, or what to do if the fields are spread across the document, which they often are, or if part of the xpath needs to be extracted as a field/tag name or value.

I wonder if there's some kind of object/structure mappers that do this well already?

M0rdecay:
> but it's still not clear how to go to a specific path and output one metric per element in the array  


This task can be solved by explicitly specifying the name of the node, which will be the key to the element in the array.
For example:
```xml
<WowzaStreamingEngine>
    <ConnectionsCurrent>1</ConnectionsCurrent>
    <ConnectionsTotal>1</ConnectionsTotal>
    <ConnectionsTotalAccepted>1</ConnectionsTotalAccepted>
    <ConnectionsTotalRejected>1</ConnectionsTotalRejected>
    <MessagesInBytesRate>1.0</MessagesInBytesRate>
    <MessagesOutBytesRate>1.0</MessagesOutBytesRate>
    <VHost>
        <Name>host_1</Name>
        <TimeRunning>64679.885</TimeRunning>
        <ConnectionsLimit>0</ConnectionsLimit>
        <ConnectionsCurrent>0</ConnectionsCurrent>
        <ConnectionsTotal>0</ConnectionsTotal>
        <ConnectionsTotalAccepted>0</ConnectionsTotalAccepted>
        <ConnectionsTotalRejected>0</ConnectionsTotalRejected>
        <MessagesInBytesRate>0.0</MessagesInBytesRate>
        <MessagesOutBytesRate>0.0</MessagesOutBytesRate>
    </VHost>
    <VHost>
        <Name>host_2</Name>
        <TimeRunning>67777.885</TimeRunning>
        <ConnectionsLimit>2</ConnectionsLimit>
        <ConnectionsCurrent>2</ConnectionsCurrent>
        <ConnectionsTotal>2</ConnectionsTotal>
        <ConnectionsTotalAccepted>2</ConnectionsTotalAccepted>
        <ConnectionsTotalRejected>2</ConnectionsTotalRejected>
        <MessagesInBytesRate>2.2</MessagesInBytesRate>
        <MessagesOutBytesRate>2.2</MessagesOutBytesRate>
    </VHost>
</WowzaStreamingEngine>
```

```toml
[[inputs.file]]
  files = [ "data.xml" ]
  data_format = "xml"
  xml_query = "//VHost/"
  xml_merge_nodes = true
  tag_keys = [ "Name" ]
  xml_array_key = "VHost" # <- This functionality needs to be added
```
After minor improvements, it will be possible to achieve the following result:
```
file,host=host.local,Name=host_1 TimeRunning=64679.885,ConnectionsTotalAccepted=0i,MessagesInBytesRate=0,MessagesOutBytesRate=0,ConnectionsCurrent=0i,ConnectionsTotal=0i,ConnectionsTotalRejected=0i,ConnectionsLimit=0i 1598357710000000000
file,host=host.local,Name=host_2 ConnectionsTotal=2i,MessagesInBytesRate=2.2,ConnectionsTotalAccepted=2i,ConnectionsTotalRejected=2i,MessagesOutBytesRate=2.2,TimeRunning=67777.885,ConnectionsLimit=2i,ConnectionsCurrent=2i 1598357710000000000
```


For the rest - unfortunately, it seems to me that it will not work to achieve such flexibility by providing only a declarative method for solving the problem.  
I would also like to avoid using xslt to keep the configuration simple.

M0rdecay:
@ssoroka hello!  
What do you think of the information above?
If this sounds good to you, I would gladly take on the implementation.

ssoroka:
So we're going to go ahead with your suggestion, and this will be about equal to the approach with the json parsing plugin, so seems reasonable. Both approaches don't solve the problem well for non-trivial documents. If possible, could you add more usage examples to the readme, and add a note that if your XML document is complex you may need to move over to an execd plugin? One way to do this is to use the _value_ parser, then use a processors.execd script to parse that into the proper metric.

M0rdecay:
Thanks for the answer!
I will be updating the branch shortly and will write when everything is ready.

M0rdecay:
@ssoroka, hello!

I have prepared the changes that we discussed with you.
Looking forward to your review.

M0rdecay:
Hmm...
I don't fully understand why the Windows test failed. The last commit has nothing to do with win_perf_counters

M0rdecay:
still failing...

@ssoroka, hello!
Could you review the proposal, please? We would like to be in time before the 1.16 release.

M0rdecay:
@HRI-EU  
I also looked at the solution suggested in the request.

I like the functionality to get a custom time stamp from the document and the ability to set the measurement name based on the value from the document.

But the need to enumerate fields and specify their types manually seems inconvenient. It also makes the configuration heavy.

If needed, I can add measurement name and timestamp extraction to this implementation.

Your opinion?

srebhan:
Please rebase on master to fix the AppVeyor problem.

M0rdecay:
@srebhan great, thanks!

M0rdecay:
@reimda, hello!

Unfortunately, Steven hasn't been in touch yet.
Could you please consider the offered parsers and suggest which one you think is more suitable?
We would like to see any of them in upstream - the main thing is to get a tool for processing xml.

M0rdecay:
Okay, I guess I'm missing the ability to get tags or fields from arbitrary place in the document and add them to each generated metric.  
Something like this should work:
```toml
xml_tags = [ "../node", "//data" ]
```

This feature should replace the current functionality of getting the name of the measurement.  
To turn a tag or field into a metric name, we can use `converter` processor

M0rdecay:
I think it's done

M0rdecay:
@ssoroka, hello!

I apologize for the frequent mention, we are really looking forward to the XML parser in telegraf.

Can we wait for the soon inclusion of this or an alternative solution in the master? 
Maybe something needs to be improved?

Right now we use the parser through a wrapper as an external processor, but it's not very convenient ...

srebhan:
Just two comments from my side where I see problems with the approach you are taking @M0rdecay:

1. **Implicit type conversions**:
You are trying to "automagically" figure out the type of a value and there you have _int_ (taking precedence) and _float_. Imagine a weird service that outputs floating-point values, but whenever the float ends with .0 (e.g. `42.0`) it will give you `42`. In the next iteration you get `23.5`. Now you parse the first value as int, the later one as float...
2. **"Reimplementation" of XPath functionality**:
While your code reads nice and clean I've the feeling that you partially reimplement what XPath was meant for.

However, I see the advantage of your approach for some use-cases and really would also like to merge this into my PR #8047. But we have to agree on the underlying XML-library first I guess...

M0rdecay:
@srebhan good morning!

On the first point - of course, such a situation is possible, but it would violate the XML schema. I consider the probability of such a situation to be low, but, of course, not zero. This is a rather rare case, such situations can be handled through a `converter` processor.

About function reimplementation- indeed, the library I am using is not as functional in the `XPath` part as the one you are using, so I was forced to add attribute extraction. In general, the libraries you are using seem to be more suitable, so we can use them.

srebhan:
> On the first point - of course, such a situation is possible, but it would violate the XML schema. I consider the probability of such a situation to be low, but, of course, not zero. 

And I have exactly that kind of service running here... :-) Could you explain how it _violates the XML schema_? Do you have any reference? To my knowledge, XML does not make any assumptions on the node content, so in general it is text. Otherwise you need a DTD i think.

> This is a rather rare case, such situations can be handled through a `converter` processor.

Wouldn't it be better to leave out the automatic conversion and use the `converter` processor by default? Alternatively you should be able to explicitly specify the output type.

> Here I will also add that if `42.0` is in XML, the conversion will be correct - https://play.golang.org/p/IdI6XxI7Aqn

Sure, but my point was the possible confusion with an int.

> Upd. We can continue the discussion in the Slack InfluxDB Community, but I would prefer to do it here and further so that others can see the correspondence.

As you wish. :-)

M0rdecay:
XSD unambiguously declares the type [xsd:double](http://books.xmlschemata.org/relaxng/ch19-77065.html) - are you using something different for pre-validation before XML posting?

srebhan:
True XSD **can** strictly define the type! The important thing is that this is completely optional. It is ok to send out an XML without _any_ validation. The service I'm forced to use obviously does that... :-(

M0rdecay:
Well, I see you are right. In this case, your solution becomes more versatile at the expense of the desired ease of configuration.

I would like to know what the upstream owners think about this.
For me, the main goal here is to see any of the solutions in the master branch.

M0rdecay:
We discussed about data typing with @srebhan and came to the following conclusion - the flag that indicates whether the parser will dynamically type the values ​​of the nodes may be a suitable solution.

The last commit adds this flag, so far without changing the library for working with XML.
I would like to change the library only if using the parser shows that it is necessary - this is mentioned in the README.

M0rdecay:
Let's continue at https://github.com/influxdata/telegraf/pull/8121

It seems that the second attempt came out more successful.

PR #7479, opened 2020-05-07
[outputs.influxdb_v2] Include some more context as to why write failed

ssoroka:


plugins/outputs/influxdb_v2/http.go
```diff
@@ -245,10 +245,10 @@ func (c *httpClient) writeBatch(ctx context.Context, bucket string, metrics []te
 
 	switch resp.StatusCode {
 	case http.StatusBadRequest, http.StatusRequestEntityTooLarge:
-		log.Printf("E! [outputs.influxdb_v2] Failed to write metric: %s\n", desc)
+		log.Printf("E! [outputs.influxdb_v2] Failed to write metric (%s): %s\n", resp.Status, desc)
```
danielnelson, line 248: I'm wondering why the status code isn't already in the description at this point?

ssoroka: good question. Desc was empty, meaning we're not getting a body.

danielnelson: It seems that the json decode was successful, else we would have the desc on L243.  Maybe we need to be using Unmarshal instead of Decode.

ssoroka: I suspect the server just returned an empty body. we don't handle that case. We should be using the server's resp.Status, not parsing it out of the rest.Body

danielnelson: Yes, this is the intended behavior already. However, L239-244 has a bug where desc is not set.

ssoroka:
```
	if err != nil || len(desc) == 0 {
		desc = resp.Status
	}
```
?

ssoroka:
@danielnelson is this resolved elsewhere now?

danielnelson:
Will have this in my upcoming PR, lets close.

Issue #7364, opened 2020-04-17
Add integer support to processors.enum


tackittj:
### Proposal:
Add integer->string mapping support.

### Current behavior:
The [Enum Processor Plugin](https://github.com/influxdata/telegraf/tree/master/plugins/processors/enum) currently only maps a string to an integer.  When asked to map an integer to a string it simply ignores the request.  There is no error or other output.

`green = 1`

### Desired behavior:
The `[processors.enum.mapping.value_mappings]` configuration should allow an integer to e mapped to a string.
`1 = green`

Perhaps it should even allow either string or integer on either or both sides of the mapping, though this is beyond my own immediate need.
`green = Passed`

Or perhaps an error could be generated when the enum processor is misconfigured to process a non-string.

### Use case:

This would prevent the need to first use the `[[processors.converter]]` to convert the int to a string.

HarshitOnGitHub:
I am willing to pick this up if it is available. :)

danielnelson:
Yes it is all yours, thanks!

HarshitOnGitHub:
Thanks @danielnelson! Have opened a PR here: https://github.com/influxdata/telegraf/pull/7483

PR #7483, opened 2020-05-08
[processors] Add integer support to `processors.enum`

HarshitOnGitHub:
Fixes: #7364 

### Required for all PRs:

### DIFF
--- a/plugins/processors/enum/enum_test.go
+++ b/plugins/processors/enum/enum_test.go
@@ -14,7 +14,9 @@ func createTestMetric() telegraf.Metric {
 		map[string]string{"tag": "tag_value"},
 		map[string]interface{}{
 			"string_value": "test",
-			"int_value":    int(13),
+			"int_value":    int(200),
+			"uint_value":   uint(500),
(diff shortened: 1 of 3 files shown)


plugins/processors/enum/enum_test.go
```diff
@@ -91,6 +93,22 @@ func TestMapSingleBoolValue(t *testing.T) {
...
+}
+
+func TestMapSingleUintValue(t *testing.T) {
+	mapper := EnumMapper{Mappings: []Mapping{{Field: "uint_value", ValueMappings: map[string]interface{}{"50": int64(5)}}}}
+
+	fields := calculateProcessedValues(mapper, createTestMetric())
+
+	assertFieldValue(t, int64(5), "uint_value", fields)
+}
+
```
ssoroka, line 111: can we test mapping this to a string? seems like a common use case.

HarshitOnGitHub: Makes sense, I removed the existing tests and added table tests to cover it more exhaustively. :)

PR #7484, opened 2020-05-08
feat: Add dynamic tagging to gnmi plugin

### LINKED ISSUE #7462: gNMI: Add support for static and dynamic tag mappings
## Feature Request

Currently, the SNMP input plugin supports adding tags to measurements based on other data - it is common to add the interface description to interface counter readings with the following config
```
    [[inputs.snmp.table.field]]
      oid = "IF-MIB::ifDescr"
      name = "ifDescr"
      is_tag = true
```

Ideally, we should be able to do similar with cisco_telemetry_gnmi, via subscribing to `openconfig-interfaces:/interfaces/interface/state/description` to maintain an up-to-date map, and adding the tag when transmitting measurements

### Proposal:
Be able to configure gNMI paths as tag-only, with additional configuration to identify the search key.  We should be able to read the current value of those paths, and subscribe to updates to maintain an in-memory map to apply to measurements
### Current behavior:
Only global tags are supported?
### Desired behavior:
Dynamic tags based on measurement values
### Use case:
When using InfluxDB as a backend, tags cannot be added to a measurement post-processing, so they need to be present in the initial reading.  Being able to apply dynamic device metadata to each measurement improves the searchability of measurements (show me interface stats for all interfaces whose descriptions start with "TRANSIT", etc)

bewing:
### Required for all PRs:

Closes #7462 

Add the ability to specify gNMI subscriptions to be tag-only, storing them for application to other subscriptions with the exact same tagged "name".  The primary application of the author is to apply interface description tags to all other interface metrics, to ease search in Influx.

I am extremely open to suggestion on improvement for either the memory storage of the tags, or alternatives for additional aliasing of the dynamic tag naming.

### DIFF
--- a/plugins/inputs/gnmi/README.md
+++ b/plugins/inputs/gnmi/README.md
@@ -65,17 +65,17 @@ It has been optimized to support gNMI telemetry as produced by Cisco IOS XR (64-
     ## If suppression is enabled, send updates at least every X seconds anyway
     # heartbeat_interval = "60s"
 
-  # [[inputs.gnmi.subscription]]
-  #    name = "descr"
-  #    origin = "openconfig-interfaces"
-  #    path = "/interfaces/interface/state/description"
-  #    subscription_mode = "on_change"
+  #[[inputs.gnmi.subscription]]
+    # name = "descr"
+    # origin = "openconfig-interfaces"
+    # path = "/interfaces/interface/state/description"
+    # subscription_mode = "on_change"
 
-  #    ## If tag_only is set, the subscription in question will be utilized to maintain a map of
-  #    ## tags to apply to other measurements emitted by the plugin, by matching path keys
-  #    ## All fields from the tag-only subscription will be applied as tags to other readings,
-  #    ## in the format <name>_<fieldBase>.
-  #    tag_only = true
+    ## If tag_only is set, the subscription in question will be utilized to maintain a map of
+    ## tags to apply to other measurements emitted by the plugin, by matching path keys
+    ## All fields from the tag-only subscription will be applied as tags to other readings,
+    ## in the format <name>_<fieldBase>.
+    # tag_only = true
 ```
 
 ## Example Output
--- a/plugins/inputs/gnmi/README.md
+++ b/plugins/inputs/gnmi/README.md
@@ -64,11 +64,23 @@ It has been optimized to support gNMI telemetry as produced by Cisco IOS XR (64-
 
     ## If suppression is enabled, send updates at least every X seconds anyway
     # heartbeat_interval = "60s"
+
+  # [[inputs.gnmi.subscription]]
+  #    name = "descr"
+  #    origin = "openconfig-interfaces"
+  #    path = "/interfaces/interface/state/description"
+  #    subscription_mode = "on_change"
+
+  #    ## If tag_only is set, the subscription in question will be utilized to maintain a map of
+  #    ## tags to apply to other measurements emitted by the plugin, by matching path keys
+  #    ## All fields from the tag-only subscription will be applied as tags to other readings,
+  #    ## in the format <name>_<fieldBase>.
+  #    tag_only = true
 ```
 
 ## Example Output
 
 ```shell
-ifcounters,path=openconfig-interfaces:/interfaces/interface/state/counters,host=linux,name=MgmtEth0/RP0/CPU0/0,source=10.49.234.115 in-multicast-pkts=0i,out-multicast-pkts=0i,out-errors=0i,out-discards=0i,in-broadcast-pkts=0i,out-broadcast-pkts=0i,in-discards=0i,in-unknown-protos=0i,in-errors=0i,out-unicast-pkts=0i,in-octets=0i,out-octets=0i,last-clear="2019-05-22T16:53:21Z",in-unicast-pkts=0i 1559145777425000000
-ifcounters,path=openconfig-interfaces:/interfaces/interface/state/counters,host=linux,name=GigabitEthernet0/0/0/0,source=10.49.234.115 out-multicast-pkts=0i,out-broadcast-pkts=0i,in-errors=0i,out-errors=0i,in-discards=0i,out-octets=0i,in-unknown-protos=0i,in-unicast-pkts=0i,in-octets=0i,in-multicast-pkts=0i,in-broadcast-pkts=0i,last-clear="2019-05-22T16:54:50Z",out-unicast-pkts=0i,out-discards=0i 1559145777425000000
+ifcounters,path=openconfig-interfaces:/interfaces/interface/state/counters,host=linux,name=MgmtEth0/RP0/CPU0/0,source=10.49.234.115,descr/description=Foo in-multicast-pkts=0i,out-multicast-pkts=0i,out-errors=0i,out-discards=0i,in-broadcast-pkts=0i,out-broadcast-pkts=0i,in-discards=0i,in-unknown-protos=0i,in-errors=0i,out-unicast-pkts=0i,in-octets=0i,out-octets=0i,last-clear="2019-05-22T16:53:21Z",in-unicast-pkts=0i 1559145777425000000
+ifcounters,path=openconfig-interfaces:/interfaces/interface/state/counters,host=linux,name=GigabitEthernet0/0/0/0,source=10.49.234.115,descr/description=Bar out-multicast-pkts=0i,out-broadcast-pkts=0i,in-errors=0i,out-errors=0i,in-discards=0i,out-octets=0i,in-unknown-protos=0i,in-unicast-pkts=0i,in-octets=0i,in-multicast-pkts=0i,in-broadcast-pkts=0i,last-clear="2019-05-22T16:54:50Z",out-unicast-pkts=0i,out-discards=0i 1559145777425000000
 ```
--- a/plugins/inputs/gnmi/gnmi.go
+++ b/plugins/inputs/gnmi/gnmi.go
@@ -594,17 +594,17 @@ const sampleConfig = `
   ## If suppression is enabled, send updates at least every X seconds anyway
   # heartbeat_interval = "60s"
 
-  # [[inputs.gnmi.subscription]]
-  #   name = "tag"  # tag name will be name + "_" + shortPath ie "tag_description"
-  #   origin = "openconfig-interfaces"
-  #   path = "/interfaces/interface/state/description"
-  #   subscription_mode = "on_change"
+  #[[inputs.gnmi.subscription]]
+    # name = "descr"
+    # origin = "openconfig-interfaces"
+    # path = "/interfaces/interface/state/description"
+    # subscription_mode = "on_change"
 
-  #   ## If tag_only is set, the subscription in question will be utilized to maintain a map of
-  #   ## tags to apply to other measurements emitted by the plugin, by matching path keys
-  #   ## All fields from the tag-only subscription will be applied as tags to other readings,
-  #   ## in the format <name>_<fieldBase>.
-  #   tag_only = true
+    ## If tag_only is set, the subscription in question will be utilized to maintain a map of
+    ## tags to apply to other measurements emitted by the plugin, by matching path keys
+    ## All fields from the tag-only subscription will be applied as tags to other readings,
+    ## in the format <name>_<fieldBase>.
+    # tag_only = true
 `
 
 // SampleConfig of plugin
--- a/plugins/inputs/gnmi/gnmi.go
+++ b/plugins/inputs/gnmi/gnmi.go
@@ -56,6 +56,8 @@ type GNMI struct {
 	acc             telegraf.Accumulator
 	cancel          context.CancelFunc
 	wg              sync.WaitGroup
+	// Lookup/device+name/key/value
+	lookup map[string]map[string]map[string]interface{}
 
 	Log telegraf.Logger
 }
@@ -73,6 +75,9 @@ type Subscription struct {
 	// Duplicate suppression
 	SuppressRedundant bool            `toml:"suppress_redundant"`
 	HeartbeatInterval config.Duration `toml:"heartbeat_interval"`
+
+	// Mark this subscription as a tag-only lookup source, not emitting any metric
+	TagOnly bool `toml:"tag_only"`
 }
 
 // Start the http listener service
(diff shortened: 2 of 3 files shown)


bewing:
If anyone wants to do integration testing against non-Arista hardware, it would be appreciated.

bewing:
Switched from using `/` as the tag separator to `_`, as grafana was having a heck of a time when tags contained slashes per https://github.com/grafana/grafana/issues/6012

kb70:
Hi,

any chance to see that in a release any time soon?
That's exactly the feature I'm missing for the telemetry stuff of our Nokia SRs. Often, ie. for ports or BGP neighbors, the description can only be found in the /configure tree, not in the /state tree. So having a possibility to dynamically lookup tags in some other path would help a lot.

Thanks for your work,
Kurt

bewing:
Updated for new plugin name

bewing:
Am I missing any data/information needed to merge this feature?  What are the next steps from my side?

plugins/inputs/gnmi/README.md
```diff
@@ -64,10 +64,22 @@ It has been optimized to support gNMI telemetry as produced by Cisco IOS XR (64-
 
     ## If suppression is enabled, send updates at least every X seconds anyway
     # heartbeat_interval = "60s"
+
+  [[inputs.gnmi.subscription]]
+   name = "descr"
+   origin = "openconfig-interfaces"
+   path = "/interfaces/interface/state/description"
+   subscription_mode = "on_change"
+
```
ssoroka, line 73: Usually we leave optional configuration commented out. Make sure the indentation matches what's already here

plugins/inputs/gnmi/gnmi.go
```diff
@@ -55,6 +55,8 @@ type GNMI struct {
 	acc     telegraf.Accumulator
 	cancel  context.CancelFunc
 	wg      sync.WaitGroup
+	// Lookup/device/name/key/value
+	lookup map[string]map[string]map[string]map[string]interface{}
```
ssoroka, line 59: this seems like an odd structure. 
Do we know nothing about the data structure? 
Does it have always exactly 4 segments? 
can we store it as a single string key (and parse it into pieces on demand) instead?

plugins/inputs/gnmi/gnmi.go
```diff
@@ -72,6 +74,9 @@ type Subscription struct {
 	// Duplicate suppression
 	SuppressRedundant bool              `toml:"suppress_redundant"`
 	HeartbeatInterval internal.Duration `toml:"heartbeat_interval"`
+
+	// Tag-only identification
+	TagOnly bool `toml:"tag_only"`
```
ssoroka, line 79: not sure I understand the utility of this. It's checked only in one place.

plugins/inputs/gnmi/gnmi.go
bewing: The premise of the plugin is to treat some of the different update messages we get from the gNMI server as "data to be tagged on other, non-tag-specific updates".  IE, if there is an openconfig endpoint that can map `interface name` -> `interface description`, when we are receiving updates giving us counter information keyed by `interface name`, go ahead and tag that update with the value from the known `interface description`.  

In order to do that, we need to track and store these key/value pairs, and index them by lookup (the gNMI subscription/endpoint), which device we receive them from, and the key value for this specific lookup (eg, `interface name` in our previous example, with the bottom layer being key/value pairs for the actual tags to apply to the measurement.

We need the top-level map (string ->map) for the specific lookup, so we can perform the retrieve test in [L304](https://github.com/influxdata/telegraf/pull/7484/files#diff-8f015c9f22f0f8f4f8ee17b671e6d8f3ffa1871b86fc765a96e00d9d45ba0f39R304) when processing an update from the server, to see if it should be used to update our tag lookup map, or process it as a normal telemetry update.

The bottom layer map of tag name -> tag value is needed when enumerating them to apply to other updates in [L314](https://github.com/influxdata/telegraf/pull/7484/files#diff-8f015c9f22f0f8f4f8ee17b671e6d8f3ffa1871b86fc765a96e00d9d45ba0f39R314)

Would it make sense to collapse the middle two `map[string]map[string]` for `device` and `interface name` into an anonymous struct?

`lookup map[string]map[{device, name string}]map[string]interface{}`

Really do appreciate the feedback on this -- it was the section of the code I was least confident in.

plugins/inputs/gnmi/gnmi.go
bewing: Serves the same function as [isTag](https://github.com/influxdata/telegraf/blob/d369003912e6568aeb43c8d56207e407550afa01/plugins/inputs/snmp/snmp.go#L231-L232) in SNMP plugin -- to identify that this specific reading shouldn't be emitted to outputs, but stored to apply to other telemetry flows

bewing:
> Switched from using `/` as the tag separator to `_`, as grafana was having a heck of a time when tags contained slashes per [grafana/grafana#6012](https://github.com/grafana/grafana/issues/6012)

Given that Granfana has resolved their issue with slashes, does it make sense to switch back to using slashes as separators in this plugin?

slampt:
Hola, just curious why this wasn't merged in?

cprecup:
Hello @bewing,

Tested this with IOS-XR and the OpenConfig model `openconfig-interfaces` but I can't confirm it works on my end. 

Here is the configuration:
```toml
[...]
 [[inputs.gnmi.subscription]]
   origin = "openconfig-interfaces"
   path = "/interfaces/interface/state/counters"
   subscription_mode = "sample"
   sample_interval = "60s"

 [[inputs.gnmi.subscription]]
   name = "descr"
   origin = "openconfig-interfaces"
   path = "/interfaces/interface/state/description"
   subscription_mode = "on_change"
   tag_only = true
[...]
```

The output is as follows:
```bash
telegraf git:(7484) ✗ ./telegraf --config telegraf-7484-gnmi-embedded-tags.conf 
2021-03-18T17:58:17Z I! Starting Telegraf 
[...]
2021-03-18T17:58:19Z E! [inputs.gnmi] Subscribe error (3), "openconfig-interfaces:interfaces/interface/state/description:ERR:No sysdb paths found for yang path openconfig-interfaces:interfaces/interface/state/description\x00"
counters,host=telegraf,name=GigabitEthernet0/0/0/0,path=openconfig-interfaces:/interfaces/interface/state/counters,source=x.f.g.h in_octets=0i,out_octets=0i,in_multicast_pkts=0i,in_broadcast_pkts=0i,out_multicast_pkts=0i,out_broadcast_pkts=0i,in_unknown_protos=0i,in_errors=0i,out_errors=0i,last_clear="2021-03-18T15:10:02Z",in_unicast_pkts=0i,in_discards=0i,out_unicast_pkts=0i,out_discards=0i 1616089682978000000
counters,host=telegraf,name=GigabitEthernet0/0/0/1,path=openconfig-interfaces:/interfaces/interface/state/counters,source=x.f.g.h in_octets=0i,out_octets=0i,in_multicast_pkts=0i,in_broadcast_pkts=0i,out_multicast_pkts=0i,out_broadcast_pkts=0i,in_unknown_protos=0i,in_errors=0i,out_errors=0i,last_clear="2021-03-18T15:10:02Z",in_unicast_pkts=0i,in_discards=0i,out_unicast_pkts=0i,out_discards=0i 1616089682978000000
[...]
```

bewing:
Rebased for merge conflict

plugins/inputs/gnmi/gnmi.go
```diff
@@ -296,6 +312,22 @@ func (c *GNMI) handleSubscribeResponseUpdate(address string, response *gnmi.Subs
...
+
+		// Apply lookups if present
+		for k, v := range c.lookup {
+			if t, ok := v[tags["source"]][tags["name"]]; ok {
+				for name, val := range t {
+					tagName := fmt.Sprintf("%s_%s", k, name)
+					tags[tagName] = val.(string)
+				}
+			}
+		}
```
srebhan, line 329: Similar to the previous comment of Steven, I'd like to see this lookup simplified. So how about using a `map[string]map[string]map[string]string` for `c.lookup`. This would save one `map`... The code here then looks like the following
```suggestion
		subscriptionKey := tags["source"] + "/" + tags["name"]
		if _, ok := c.lookup[name]; ok {
			// We are subscribed to this, so add the fields to the lookup-table
			if _, ok := c.lookup[name][subscriptionKey]; !ok {
				c.lookup[name][subscriptionKey] = make(map[string]string)
			}
			for k, v := range fields {
				c.lookup[name][subscriptionKey][path.Base(k)] = v
			}
			// Do not process the data further as we only subscribed here to update the lookup-table
			continue
		}

		// Apply lookups if present
		for subscriptionName, values := range c.lookup {
			annotations, ok := values[subscriptionKey]
			if !ok {
				continue
			}
			for k, v := range annotations {
				tags[subscriptionName + "_" + k] = v
			}
		}
```

plugins/inputs/gnmi/gnmi.go
```diff
@@ -559,6 +603,18 @@ const sampleConfig = `
...
+  # name = "tag"  # tag name will be name + "_" + shortPath ie "tag_description"
+  # origin = "openconfig-interfaces"
+  # path = "/interfaces/interface/state/description"
+  # subscription_mode = "on_change"
+
+  ## If tag_only is set, the subscription in question will be utilized to maintain a map of
+  ## tags to apply to other measurements emitted by the plugin, by matching path keys
+  ## All fields from the tag-only subscription will be applied as tags to other readings,
+  ## in the format <name>_<fieldBase>.
+  # tag_only = true
```
srebhan, line 617: Please fix the indentation and/or the whitespaces to match the style above.

srebhan (commented):
@bewing thanks for submitting this PR. I have some comments similar to the previous comments by Steven. Beside some smaller comments in the code I propose a way to save at least one level in the map. While I'm not 100% satisfied with three nested maps, it's a beginning. Maybe we can boil it further down?

plugins/inputs/gnmi/README.md
```diff
@@ -64,10 +64,22 @@ It has been optimized to support gNMI telemetry as produced by Cisco IOS XR (64-
...
+  # name = "descr"
+  # origin = "openconfig-interfaces"
+  # path = "/interfaces/interface/state/description"
+  # subscription_mode = "on_change"
+
+  ## If tag_only is set, the subscription in question will be utilized to maintain a map of
+  ## tags to apply to other measurements emitted by the plugin, by matching path keys
+  ## All fields from the tag-only subscription will be applied as tags to other readings,
+  ## in the format <name>_<fieldBase>.
+  # tag_only = true
```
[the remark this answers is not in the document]
bewing: I think this is fixed in the latest revision -- if there are still issues, please provide more detail or a specific example, as I can't find too many clear differences with other plugins at this point.

plugins/inputs/gnmi/gnmi.go
```diff
@@ -72,6 +74,9 @@ type Subscription struct {
 	// Duplicate suppression
 	SuppressRedundant bool            `toml:"suppress_redundant"`
 	HeartbeatInterval config.Duration `toml:"heartbeat_interval"`
+
+	// Tag-only identification
+	TagOnly bool `toml:"tag_only"`
```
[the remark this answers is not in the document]
bewing: Updated comment as requested

plugins/inputs/gnmi/gnmi.go
bewing: Thanks for the suggestion!  I did collapse the two values into a single subscription key, but kept the bottom level of the map as an interface, in case there's a use case where someone wants to use a non-string metric as a tag.

plugins/inputs/gnmi/gnmi.go
bewing: Per other comment, I believe I have addressed this, but let me know if additional work is necessary

bewing:
> @bewing thanks for submitting this PR. I have some comments similar to the previous comments by Steven. Beside some smaller comments in the code I propose a way to save at least one level in the map. While I'm not 100% satisfied with three nested maps, it's a beginning. Maybe we can boil it further down?

Thanks for taking the time to look at this PR!  Apologies for the delay in the response, it took some time for me to circle back and get back into the same headspace I was when I first wrote it.  I force pushed a new change, please let me know your thoughts on the latest revision.

srebhan (approved):
@bewing that's exactly what I had in mind. :-) Thanks for taking the time to implement it!

Looks good to me.

plugins/inputs/gnmi/README.md
```diff
@@ -64,11 +64,23 @@ It has been optimized to support gNMI telemetry as produced by Cisco IOS XR (64-
...
+  #    name = "descr"
+  #    origin = "openconfig-interfaces"
+  #    path = "/interfaces/interface/state/description"
+  #    subscription_mode = "on_change"
+
+  #    ## If tag_only is set, the subscription in question will be utilized to maintain a map of
+  #    ## tags to apply to other measurements emitted by the plugin, by matching path keys
+  #    ## All fields from the tag-only subscription will be applied as tags to other readings,
+  #    ## in the format <name>_<fieldBase>.
+  #    tag_only = true
```
powersj, line 78: This is good to go, but I would really like the config to be formatted correctly in the README and in the example config. Removing extra white space, and only double, not triple pound sign for comments.

I believe this is more in-line with what we the config currently uses:

```
    # [[inputs.gnmi.subscription]]
    # name = "descr"
    # origin = "openconfig-interfaces"
    # path = "/interfaces/interface/state/description"
    # subscription_mode = "on_change"

    ## If tag_only is set, the subscription in question will be utilized to maintain a map of
    ## tags to apply to other measurements emitted by the plugin, by matching path keys
    ## All fields from the tag-only subscription will be applied as tags to other readings,
    ## in the format <name>_<fieldBase>.
    # tag_only = true
```

powersj: @bewing is this something you could do and then we could land this?

powersj: @sspaink has resolved this, thank you!

powersj:
@bewing have you signed the CLA? @sspaink has fixed up the readme/config docs, so this is ready to merge, minus that.

bewing:
> @bewing have you signed the CLA? @sspaink has fixed up the readme/config docs, so this is ready to merge, minus that.

I have not.  I have an inquiry with my legal department pending regarding that.  I will follow up with them this week.

powersj:
!signed-cla

powersj (approved):

powersj:
Thank you @bewing!

PR #7495, opened 2020-05-11
Sql server old version compatibility

### LINKED ISSUE #7292: Sql Server - Support for SQL Server 2008

Trovalo:
### Required for all PRs:

closes #7292

This PR adds compatibility for all the queries from SQL Server 2008 SP4 and later version, I made a very similar PR some weeks ago, the difference is that this one contains only the minimum to obtain compatibility.

Main changes:
- The switch logic used to compose the queries is now based on variables when appropriate (MajorVersion, MinorVersion, EngineEdition) instead of being fetched every time, this also solved some existing (but not reported) issues, see bottom of the PR
- Several CASE statement have been added to manage SQL 2008 and 2008 R2

It has been tested on:
- SQL 2008 SP4
- SQL 2008 R2
- SQL 2012
- SQL 2014
- SQL 2016 SP2
- SQL 2017
- SQL 2019
- Azure SQL DB

Additional Fix:
In the current version of telegraf, in the "performance counters" query, the column [total_cpu_usage_preemptive_ms] is never considered, it is added as follows:
```SQL
CASE WHEN SERVERPROPERTY('ProductMajorVersion') > 10 THEN 'rgwg.total_cpu_usage_preemptive_ms

-- Sadly that switch does not work unless you also cast SERVERPROPERTY to int, run this to check
SELECT IIF( SERVERPROPERTY('ProductMajorVersion') > 10 ,'ok','not ok')
```
Now that the switches are based on variables (with proper data type) the metric will be loaded for SQL Server 2016 and later, in previous versions it does not exists


Trovalo:
TLDR:
- complexity - the queries can be improved and standardized to be more readable and easier to maintain, but won't be less complex
- avoid compatibility breaks - who writes the code must check from which version the referenced object is available and possibly test its query on previous and later versions of SQL Server

**About complexity**
I'm sure that the code can be improved to be more readable and easier to maintain, I'm reading [Erland Sommarskog The Curse and Blessings of Dynamic SQL](http://www.sommarskog.se/dynamic_sql.html) to achieve that. Defining a sort of standard about how to write SQL code for the plugin should help, especially for the dynamic SQL parts.
_I will update this PR or make a new one when I come up with something better (and I will)_

In any case, I doubt that the queries will be less complex, in my opinion, the dynamic SQL is the best option to manage additional/different columns/objects by checking edition and version of SQL Server.

**About avoiding compatibility breaks**
there are few things to take into consideration
- Before adding/editing something in a query is to have a look at msdocs or other for info about the column/view, in order to add it only in the appropriate edition and version. (easy for recent version, not always easy for old versions)
- Test the query on previous and later version of SQL server to ensure it works correctly

other things not strictly related to telegraf are:
- go-mssqldb (the package used to connect) - it supports the connection to SQL 2008 SP3 and SQL 2008R2 SP2 and later versions. [source](https://github.com/denisenkom/go-mssqldb#known-issues).
There are lots of SQL 2008/2008R2 still around (sadly), therefore I doubt they will remove compatibility for those versions (I see no reason to do so anyway)
- SQL Server itself - Sometimes a CU adds a new useful view, but as of now, (almost) all the views used in the plugin have been there since SQL 2008 and have not been marked as "deprecated". Even if something will be marked as "deprecated" there will be plenty of time to develop a workaround, in fact the deprecated object is not usually removed before at least 2 versions of SQL server (if it is removed at all, in SQL 2019 you can still find objects deprecated since SQL 2005)

danielnelson:
@denzilribeiro Can you take a look when you have time?

denzilribeiro:
My thought is you can support older versions forever and you pay for them in some way :) both with maintaining and a huge text matrix. [End of life support for 2008/R2](https://support.microsoft.com/en-gb/help/4456242/end-of-support-for-sql-server-2008-and-sql-server-2008-r2) has been reached last year  I don't even have any 2008 lying around to test.

I don't agree with changing everything to dynamic SQL - debuggability of dynamic SQL is painful as well - makes sense when you are already using dynamic SQL to adhere to a standard.
Ideally each connection string first detects it version once and then picks a relevant set of simplified not over complicated scripts - would make it easier to maintain. Right now single set of scripts actually. I will start some discussion around feature request next week over long weekend to start splitting some of this to make it more manageable, this has grown quick and # versions increased drastically :) and not yet added elastic pool support .

Trovalo:
> Note I haven't tested these against all versions, semantics have by and large not changed, these ae more standardization. There is no need to change sqlServerPropertiesV2 to dynamic SQL it does not achieve a whole lot I would not make this change - what we should do for this is split into on-prem v/s cloud.
> 
> Formulating the approach you have for dynamic SQL is fine IF you need dynamic SQL, I wouldn't convert those queries that don't need dynamic SQL to dynamic SQL.
> 
> Maybe a bit of over complication in variable section don't need a derived table - this is in multiple places - example below - you could even have that assigned as part of declare statement itself.
> 
> SELECT
> @EngineEdition = CAST(SERVERPROPERTY('EngineEdition') AS int)
> ,@majorversion = CAST(PARSENAME(CAST(SERVERPROPERTY('ProductVersion') as nvarchar),4) AS int)
> 
> Secondly not sure of the reason for this change
> 
> * Session_db_name comes form the session DMV not the request DMV?
> * DB_NAME(r.database_id) as session_db_name

- You are right about the dynamic SQL for sqlServerPropertiesV2, it's not needed in every query but only in the on-prem editions one. (the column _virtual_machine_type_desc_ is missing in SQL 2008)
- You are also right about the variable's value assignment, I can set it in the declare statement
- about the _database_id_ column, that's for compatibility, in fact, the field is available in [sys.dm_exec_sessions](https://docs.microsoft.com/en-us/sql/relational-databases/system-dynamic-management-views/sys-dm-exec-sessions-transact-sql?view=sql-server-ver15) since SQL 2012, but is available in [sys.dm_exec_requests](https://docs.microsoft.com/en-us/sql/relational-databases/system-dynamic-management-views/sys-dm-exec-requests-transact-sql?view=sql-server-ver15) for any version (and since it was already in the query I took it from there)

Trovalo:
> My thought is you can support older versions forever and you pay for them in some way :) both with maintaining and a huge text matrix. [End of life support for 2008/R2](https://support.microsoft.com/en-gb/help/4456242/end-of-support-for-sql-server-2008-and-sql-server-2008-r2) has been reached last year I don't even have any 2008 lying around to test.
> 
> I don't agree with changing everything to dynamic SQL - debuggability of dynamic SQL is painful as well - makes sense when you are already using dynamic SQL to adhere to a standard.
> Ideally each connection string first detects it version once and then picks a relevant set of simplified not over complicated scripts - would make it easier to maintain. Right now single set of scripts actually. I will start some discussion around feature request next week over long weekend to start splitting some of this to make it more manageable, this has grown quick and # versions increased drastically :) and not yet added elastic pool support .

About supporting older versions (even out of extended support), I think it's worth it since those versions are still around (I have an instance of SQL 2000...) and as a dba being able to monitor them is really helpful. There are already several switches to manage different versions and adding at worst two more (for SQL 2008 and 2008R2) won't make a difference

Personally, if I have to choose between maintaining a set of scripts (in the worst case one for each version) and a more complex dynamic SQL script I would choose the latest.

Currently, my idea to make it more readable is to create the "standard" script with some placeholders/variables (ie: for the columns), at the beginning the placeholder variable will be populated with the columns appropriate for the version and then injected in the SQL string, this should improve the query readability by a lot since it moves all the CASE statements to a different area leaving the query cleaner.

Trovalo:
In the last version of this branch:
- I've improved/edited the queries as suggested by @denzilribeiro 
- the dynamic SQL has been rewritten to be way more readable and manageable
- I've already managed the merge with the latest master branch (there were only 2 PRs on this plugin, #7526 and #7524)

Tests:
- I've run the tests on SQL Server from 2008 to 2019 and also on Azure SQL DB
- A long-running test is now in progress (on SQL 2017 but with a real workload)

Trovalo:
My long running test is raising several warnings
> W! [agent] ["outputs.influxdb"] did not complete within its flush interval

The same error sometimes occurs for the sqlserver input plugin, but the same error occurs also on the official version of telegraf even if with less frequency. 
I have 2 telegraf instances (the official and the test one) monitoring the same database and writing on different retention policies
I'm not sure about the cause of the error though, @danielnelson have you got any idea about how can I get more info about it?

ssoroka:
@Trovalo That warning is likely unrelated to this plugin. It happens for lossy output connections, and can happen with the old v1 InfluxDB output when gzip is not turned on and there is a lot of data to send. By itself it just means that the write to InfluxDB took longer than whatever you have set for "flush_interval" in your settings, and is not necessarily indicative of an immediate problem.

Trovalo:
@denzilribeiro, @danielnelson any update on this?

plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -422,9 +452,30 @@ BEGIN
...
 BEGIN
 
+	DECLARE @MajorMinorVersion AS int = CAST(PARSENAME(CAST(SERVERPROPERTY('ProductVersion') as nvarchar),4) AS int) * 100 + CAST(PARSENAME(CAST(SERVERPROPERTY('ProductVersion') as nvarchar),3) AS int)
+
+	DECLARE @Columns as nvarchar(max) = ''
+	DECLARE @Tables as nvarchar(max) = ''
+
+	IF @MajorMinorVersion >= 1050 BEGIN 
+		/* [volume_mount_point] TRIMS trailing "\" which are not allowed in InfluxDB */
+		SET @Columns += N',LEFT(vs.[volume_mount_point], LEN(vs.[volume_mount_point])-(PATINDEX(''%[^\]%'',REVERSE([volume_mount_point]))-1)) AS [volume_mount_point]'
```
danielnelson, line 465: Let's do something for #7558 first so we can avoid it in the query.

plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -422,9 +452,30 @@ BEGIN
...
+		
+	IF @MajorMinorVersion > 1100 BEGIN
+		SET @Columns += N'
+,vfs.io_stall_queued_read_ms AS [rg_read_stall_ms] 
+,vfs.io_stall_queued_write_ms AS [rg_write_stall_ms]'
+	END
+	
+	/*Debug only*/
+	--PRINT @Columns
+	--PRINT @Tables
```
danielnelson, line 477: Remove any debug lines, even commented, from from the final code.

plugins/inputs/sqlserver/sqlserver.go
Trovalo: This code removes the trailing ```\``` from the string. It looks complex but it's the safest way to remove all the trailing chars in case the strings ends with multiple ```\```. (which does not make any sense but it's better to be safe)

plugins/inputs/sqlserver/sqlserver.go
Trovalo: The idea was to keep those there to make debugging easier for anyone who wanted to check what was going, I would like to keep at least the last of them, which prints out the query that will be sent to SQL Server. (if that's against the standard/rules I will just remove them all)

plugins/inputs/sqlserver/sqlserver.go
danielnelson: I hate for you to have to do this in query, and for outputs that do support trailing slashes it imposes the line protocol limitations on them too.  I opened #7652, let's see what others think of having the solution there.

plugins/inputs/sqlserver/sqlserver.go
danielnelson: Normally, I wouldn't allow this in Go code.  In this case I'll leave it to @denzilribeiro

denzilribeiro: These are never going to be triggerred unless uncommented so why leave it in the code? I am with @danielnelson  on this... if the intent is to run the query and debug elsewhere add it later. There is nothing from a config perspective that you can change to have this part triggered so I would remove.

plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -433,40 +484,36 @@ BEGIN
...
-	N'FROM sys.dm_io_virtual_file_stats(NULL, NULL) AS vfs
+		,vfs.[num_of_bytes_written] AS [write_bytes]'
+		+ @Columns + N'
+	FROM sys.dm_io_virtual_file_stats(NULL, NULL) AS vfs
 	INNER JOIN sys.master_files AS mf WITH (NOLOCK)
 		ON vfs.[database_id] = mf.[database_id] AND vfs.[file_id] = mf.[file_id]
-	CROSS APPLY sys.dm_os_volume_stats(vfs.[database_id], vfs.[file_id]) AS vs
 	'
+	+ @Tables;
+
```
denzilribeiro, line 499: Remove Debug print statement?

plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -433,40 +484,36 @@ BEGIN
...
+	--PRINT @SqlStatement
+	
 	EXEC sp_executesql @SqlStatement
 
 END
-
 `
 
 // Conditional check based on Azure SQL DB, Azure SQL Managed instance OR On-prem SQL Server
 // EngineEdition=5 is Azure SQL DB, EngineEdition=8 is Managed instance
```
denzilribeiro, line 510: This complexity isn't worth it as we have taken a query that you could copy/paste and run to one that has dynamic SQL and to see the SQL generated you have to print it.  If there was already dynamic SQL I get changing it to tidy up but if not I wouldn't make a change to dynamic SQL unless is providing value.

plugins/inputs/sqlserver/sqlserver.go
Trovalo: All the debug prints have been removed, a pair of them also in the v1 plugin

plugins/inputs/sqlserver/sqlserver.go
Trovalo: This is part of the query sqlDatabaseIOV2 and is an actual part of the SQL statement, it adds the following when appropriate 
> CROSS APPLY sys.dm_os_volume_stats(vfs.[database_id], vfs.[file_id]) AS vs

Trovalo:
> Would revert the sqlServerPropertiesV2 change as it changes a non dynmic SQL to dynamic SQL and even in that there are server property calls in the query anyways.. Also agree with Daniel in removing the debug prints.

About reverting sqlServerPropertiesV2. 
- the dynamic SQL is created and execute only if necessary, so only on SQL Server on prem
- the dynamic SQL is needed to manage the column [hardware_type] (based on [virtual_machine_type_desc]), which does not exists in older versions.
- I kept the call to ```SERVERPROPERTY(''EngineEdition'')``` in the query itself because it was more readable than a parametrized statement, but I will change it.

I will also change the ```@EngineEdition``` parameter from int to tinyint since its values range goes from 1 to 8

plugins/inputs/sqlserver/sqlserver.go
Trovalo: Since this has been fixed on the telegraf side I will try it out and then make a PR to avoid this logic in the query itself

ssoroka:
@Trovalo congrats on getting this merged 😃. That was an impressive effort; Thank you! ❤️

Issue #7447, opened 2020-05-01
[Wavefront Output Plugin] Allow option to truncate tag value if it is too long


jhua-vmware:
## Feature Request
Allow option to truncate tag value if it is too long. So that points won't be blocked by wavefront proxy.

### Proposal:
introduce a new configurable allow_truncate_tag_value (true/false)

### Current behavior:
We are using [[inputs.procstat]] with cmdline_tag = true

Currently, the points with very long cmdline tag will be blocked by wavefront proxy. According to https://docs.wavefront.com/wavefront_data_format.html on pointTags, the max len(key) + len(value) is 254.

### Desired behavior:
Introduce a configurable plugin parameter allow_truncate_tag_value (false by default), when it is set to true, the warefront output plugin can truncate value as:
                 if w.AllowTruncateTagValue && len(key)+len(val) > 254 {
			val = val[:254-len(key)]
		}

### Use case: 
We need to use cmdline tag to identify the application processes for procstat meters, and many meters are blocked due to long cmdline without this feature.

danielnelson:
You can do this with the `strings` processor and it's "left" operation as a workaround, but we should update the wavefront output to do this automatically.

cc @prydin

jhua-vmware:
@danielnelson Thanks for the workaround. We are able to move forward, and at mean time, we can wait for the plugin fix in the future release.

prydin:
@danielnelson Can you assign this to me? I'll have a look at it.

PR #7503, opened 2020-05-13
[output.wavefront] Added truncate_tags setting

prydin:
### Required for all PRs:

Fixes issue #7447 

The total length of tag key + tag value isn't allowed to exceed 254 characters when sending metrics to Wavefront. By default, these metrics are discarded by wavefront. The new ```truncate_tags``` flag in the configuration file causes the Wavefront output to truncate the tag value so that the length of the key + value never exceeds 254. If the tag key is longer than 254 characters, the metric point is dropped and a warning is logged.

The ```truncate_tags``` flag is turned off by default to ensure backwards compatibility.


danielnelson:
One thing to consider is if we think the new option should remain long term, or if we want to push users towards truncate being true and eventually remove the setting entirely.  Depending on what you think it may be worth adding a log message to suggest setting the option.

PR #7513, opened 2020-05-14
add --no-config option to start without a main cfg file

ssoroka:
if you're loading all of your config from the config directory, you don't necessarily need a config file

### DIFF
--- a/cmd/telegraf/telegraf.go
+++ b/cmd/telegraf/telegraf.go
@@ -40,6 +40,7 @@ var fTestWait = flag.Int("test-wait", 0, "wait up to this many seconds for servi
 var fConfig = flag.String("config", "", "configuration file to load")
 var fConfigDirectory = flag.String("config-directory", "",
 	"directory containing additional *.conf files")
+var fNoConfig = flag.Bool("no-config", false, "Run Telegraf with no main config file. Useful with --config-directory")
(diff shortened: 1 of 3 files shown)


danielnelson:
The way I was thinking to have this done is that if no `--config` option is specified we check the default location, and if there is no file there it's not an error.  The same is done for the config-directory.  Finally we run checks to see if we have the right amount of inputs/outputs (this is already done and probably not part of this).

ssoroka:
That's what I set out to do when I first started, but if we do that we sort of lose the discoverability of this path functionality, which I do like. I think there's some value in just running Telegraf to see what happens, and then you learn it looks for a config in these places. Cool. Ignoring that, if we remove it anyway, I figured you may not realize Telegraf was running without the config you intended, and thought it might be better to be explicit here.

srebhan:
@ssoroka how about specifying `--config ""` to tell telegraf to not use any config file? You can either set the default to something invalid (e.g. "-"?) to determine if the option was set or use the more robust `flag.Visit()` function.

ssoroka:
@srebhan or maybe just `--no-config`

srebhan:
Ok. But please fail for cases where the user specifies `--config` **and** `--no-config` as otherwise to the user it is unclear what will happen...

srebhan:
@ssoroka any news on this PR?

PR #7514, opened 2020-05-14
fix randomly failing CI test

ssoroka:
Okay, it's a sleep, but it's a pretty reasonable one. :D

### DIFF
--- a/internal/exec_unix.go
+++ b/internal/exec_unix.go
@@ -56,3 +56,11 @@ func WaitTimeout(c *exec.Cmd, timeout time.Duration) error {
 	// Otherwise there was an error unrelated to termination.
 	return err
 }
+
+func GracefulStop(cmd *exec.Cmd, timeout time.Duration) {
+	cmd.Process.Signal(syscall.SIGTERM)
+	go func() {
+		<-time.NewTimer(timeout).C
+		cmd.Process.Kill()
+	}()
+}
--- a/internal/exec_unix.go
+++ b/internal/exec_unix.go
@@ -56,11 +56,3 @@ func WaitTimeout(c *exec.Cmd, timeout time.Duration) error {
 	// Otherwise there was an error unrelated to termination.
 	return err
 }
-
-func GracefulStop(cmd *exec.Cmd, timeout time.Duration) {
-	cmd.Process.Signal(syscall.SIGTERM)
-	go func() {
-		<-time.NewTimer(timeout).C
-		cmd.Process.Kill()
-	}()
-}
--- a/internal/exec_windows.go
+++ b/internal/exec_windows.go
@@ -39,8 +39,3 @@ func WaitTimeout(c *exec.Cmd, timeout time.Duration) error {
 	// Otherwise there was an error unrelated to termination.
 	return err
 }
-
-func GracefulStop(cmd *exec.Cmd, timeout time.Duration) {
-	// TODO: Maybe we can send a WM_ENDSESSION message here first.
-	cmd.Process.Kill()
-}
--- a/internal/exec_windows.go
+++ b/internal/exec_windows.go
@@ -39,3 +39,8 @@ func WaitTimeout(c *exec.Cmd, timeout time.Duration) error {
 	// Otherwise there was an error unrelated to termination.
 	return err
 }
+
+func GracefulStop(cmd *exec.Cmd, timeout time.Duration) {
+	// TODO: Maybe we can send a WM_ENDSESSION message here first.
+	cmd.Process.Kill()
+}
--- a/plugins/inputs/execd/execd.go
+++ b/plugins/inputs/execd/execd.go
@@ -47,14 +47,15 @@ type Execd struct {
 	Signal       string
 	RestartDelay config.Duration
 
-	acc    telegraf.Accumulator
-	cmd    *exec.Cmd
-	parser parsers.Parser
-	stdin  io.WriteCloser
-	stdout io.ReadCloser
-	stderr io.ReadCloser
-	cancel context.CancelFunc
-	wg     sync.WaitGroup
+	acc         telegraf.Accumulator
+	cmd         *exec.Cmd
+	parser      parsers.Parser
+	stdin       io.WriteCloser
+	stdout      io.ReadCloser
+	stderr      io.ReadCloser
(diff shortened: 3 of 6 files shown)


danielnelson:
Try using the testutil.Accumulator and the Wait method

danielnelson:
Can you link to the failure this addresses?

ssoroka:
[Here you go](https://circleci.com/gh/influxdata/telegraf/54926#tests/containers/0). 
```
2020/05/14 22:44:04 Starting process: [sh ./examples/count.sh]
2020/05/14 22:44:04 E! [TestPlugin.test] Error in plugin: Error reading stdout: read |0: file already closed
2020/05/14 22:44:04 E! [TestPlugin.test] Error in plugin: Error reading stderr: read |0: file already closed
--- FAIL: TestExternalInputWorks (10.00s)
    execd_test.go:56: 
        	Error Trace:	execd_test.go:56
        	            				execd_test.go:39
        	Error:      	timeout waiting for metric
        	Test:       	TestExternalInputWorks
FAIL
FAIL	github.com/influxdata/telegraf/plugins/inputs/execd	10.009s
```

danielnelson:
What I believe is happening is that `e.cmdReadOut` and `e.cmdReadErr` must finish before cmd.Wait() is called.  If they are still running after Wait() returns, the Pipe's will be closed and they will produce the `file already closed` error if they are read from.

To be safe, I think we need to only call `(c *Cmd) Wait()` once and from a single location.

I can still produce that error, and also the race detector shows:
```
WARNING: DATA RACE
Read at 0x00c0000e6910 by goroutine 9:
  os/exec.(*Cmd).Wait()
      /usr/lib/go/src/os/exec/exec.go:502 +0x8c
  github.com/influxdata/telegraf/plugins/inputs/execd.(*Execd).cmdWait()
      /home/dbn/src/influx/src/github.com/influxdata/telegraf/plugins/inputs/execd/execd.go:202 +0x10c
  github.com/influxdata/telegraf/plugins/inputs/execd.(*Execd).cmdLoop.func1()
      /home/dbn/src/influx/src/github.com/influxdata/telegraf/plugins/inputs/execd/execd.go:113 +0x38

Previous write at 0x00c0000e6910 by goroutine 8:
  os/exec.(*Cmd).Wait()
      /usr/lib/go/src/os/exec/exec.go:505 +0xaf
  github.com/influxdata/telegraf/internal.WaitTimeout()
      /home/dbn/src/influx/src/github.com/influxdata/telegraf/internal/exec_unix.go:37 +0x124
  github.com/influxdata/telegraf/plugins/inputs/execd.(*Execd).cmdLoop()
      /home/dbn/src/influx/src/github.com/influxdata/telegraf/plugins/inputs/execd/execd.go:122 +0x5b6
  github.com/influxdata/telegraf/plugins/inputs/execd.(*Execd).Start.func1()
      /home/dbn/src/influx/src/github.com/influxdata/telegraf/plugins/inputs/execd/execd.go:92 +0x4c
```

internal/exec_unix.go
```diff
@@ -56,3 +56,11 @@ func WaitTimeout(c *exec.Cmd, timeout time.Duration) error {
 	// Otherwise there was an error unrelated to termination.
 	return err
 }
+
+func GracefulStop(cmd *exec.Cmd, timeout time.Duration) {
```
danielnelson, line 60: Move these to execd plugin, also suggest calling this something like `SignalStop`.

plugins/inputs/execd/execd.go
```diff
@@ -85,17 +86,23 @@ func (e *Execd) Start(acc telegraf.Accumulator) error {
 		return err
 	}
 
+	e.startDoneWg.Add(2) // stream reader synchronization
```
danielnelson, line 89: When last I remember looking, I couldn't figure out why we needed this waitgroup, can you explain?

plugins/inputs/execd/execd.go
```diff
@@ -112,9 +119,7 @@ func (e *Execd) cmdLoop(ctx context.Context) error {
 		case <-ctx.Done():
 			if e.stdin != nil {
 				e.stdin.Close()
-				// Immediately exit process but with a graceful shutdown
-				// period before killing
-				internal.WaitTimeout(e.cmd, 200*time.Millisecond)
+				internal.GracefulStop(e.cmd, 200*time.Millisecond)
```
danielnelson, line 122: Now I'm off topic, but 200ms seems too short.  I think we should give 5s to shutdown before kill.

internal/exec_windows.go
```diff
@@ -39,3 +39,8 @@ func WaitTimeout(c *exec.Cmd, timeout time.Duration) error {
 	// Otherwise there was an error unrelated to termination.
 	return err
 }
+
+func GracefulStop(cmd *exec.Cmd, timeout time.Duration) {
+	// TODO: Maybe we can send a WM_ENDSESSION message here first.
```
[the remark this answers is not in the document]
ssoroka: Yeah, I figure if we send this to the app it'll likely think windows is shutting down and quit gracefully.. Maybe. Will make an issue.

plugins/inputs/execd/execd.go
ssoroka: it gives the stdin/stdout stream readers, which are in separate goroutines, time to start before Stop is allowed to kill the context and start shutting everything down. ... buuut.. I think we reorganized things and it's no longer needed because we aren't trying to stop the plugin before reading the value... Tests seem to suggest I can remove it 👍

internal/exec_unix.go
ssoroka: Moved it, but I kind of like the name; signalStop sounds like ... stop sending or listening for signals or something.

plugins/inputs/execd/execd.go
ssoroka: I'd be open to increasing it, but I wonder what effect this will have on the restart delay or other things; if they'll wait in line properly or launch another process without waiting.

danielnelson: It should only slow down restart if the process takes longer than this to exit when sent a SIGTERM.  FWIW, Debian has a 30s delay in start_stop_daemon while systemd defaults to 90s.

PR #7537, opened 2020-05-18
feat: initial spike for state hooks

rawkode:
This is a rather rough idea, seeking feedback before cleaning up the code.

This PR aims to provide the means for input plugins to become stateful. By adding 2 functions, they can provide a means for Telegraf to save and load their state. As a PoC, I've implemented this for `tail` with rather minimum effort.

This implementation contains a `JSON` state plugin that stores the state on the filesystem. Though I've kept it generic enough for this to be modified for other stores. 

The `syncStateLoop` reuses the flush configuration, as I assume that the state should be inline with writes out of Telegraf. Currently the state sync doesn't care if a write fails, but this could be worked in.

Also, the state system fails when using a plugin twice; as it currently uses `logName` from `models.RunningInput` to identify the plugin. This would need thought through also.

### Required for all PRs:

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -14,6 +14,7 @@ import (
 	"github.com/influxdata/telegraf/internal"
 	"github.com/influxdata/telegraf/models"
 	"github.com/influxdata/telegraf/plugins/serializers/influx"
+	jsonStore "github.com/influxdata/telegraf/plugins/state/json"
 )
 
 // Agent runs a set of plugins.
@@ -40,8 +41,14 @@ func (a *Agent) Run(ctx context.Context) error {
 		return ctx.Err()
 	}
 
+	log.Printf("D! [agent] Initializing state")
+	err := a.openStateStore()
+	if err != nil {
+		return err
+	}
+
 	log.Printf("D! [agent] Initializing plugins")
-	err := a.initPlugins()
+	err = a.initPlugins()
 	if err != nil {
 		return err
 	}
@@ -137,6 +144,12 @@ func (a *Agent) Run(ctx context.Context) error {
 
 	log.Printf("D! [agent] Closing outputs")
 	a.closeOutputs()
+
+	log.Printf("D! [agent] Closing state")
+	err = a.closeStateStore()
+	if err != nil {
+		return err
+	}
 
 	log.Printf("D! [agent] Stopped Successfully")
 	return nil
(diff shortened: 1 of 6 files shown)


rawkode:
Thoughts appreciated. If they're positive, I will clean up the code and add the appropriate tests.

rawkode:
@danielnelson @ssoroka

danielnelson:
I'd like to setup a data-directory, on Linux `/var/lib/telegraf`, that can be configured like --config-directory.  Plugins should be able to just write into this directory as needed, I don't think we need a hook in the interface.  This way we don't need locking for concurrency, state can be updated at a plugin dependent time, and non-json state can be written.

ssoroka:
Might be interesting to provide some kind of key-value store service to plugins, as long as nobody tries to use it to communicate/coordinate between Telegraf instances. Local filesystem is probably good for this, but won't automatically survive restarts/redeploys (thinking docker/k8s here; you'd have to explicitly declare a shared-folder/volume). Lol-local stores have availability issues, but that might not be the worst thing in the world as long as we communicate that plugins need fallback behavior when their state isn't available (They need this anyway).

Sounds like Daniel is suggesting a more hands-off approach, which can be nice.. though I'd really want to expose the config object to the plugin for that. They can't assume a data-directory path.

rawkode:
@ssoroka That's kind of what is offered at present, but I maybe went a little too crazy and offered an `interface{}` store for each plugin.

I see a lot of duplication hitting the repo if we ask plugins authors to do this themselves, instead of providing the hook. I've spent a bit of time on this today and there's not really a nice way to handle it

danielnelson:
> They can't assume a data-directory path.

We can have a function that returns the data-directory if one is set.

> I see a lot of duplication hitting the repo if we ask plugins authors to do this themselves, instead of providing the hook. 

It will be better to see some duplication so we know how to properly design the interface, rather than assume we know what is needed with a single use case.

rawkode:
I still think this is a pretty flexible and nice implementation to provide a long requested feature. It would be nice to get feedback from the current maintainers.

cc @timhallinflux @russorat

srebhan:
I agree. I have a similar draft in #9476 so a general decision would be nice.

sjwang90:
Let's prioritize #9476 to move forward with reviewing. We'll re-open this PR if needed. Thanks @rawkode!

PR #7540, opened 2020-05-18
Add processor to look up service name by port

### LINKED ISSUE #7244: Add processor that does service port to name translation
### Proposal:

Add a new processor that can convert a tag containing a well-known port number to the registered service name, according to the net.LookupPort function or similar.

### Current behavior:

N/A

### Desired behavior:

```toml
[[processors.lookup_port]]
  tag = "port"

  ## If set, write port to new tag key
  # dest = "service"
```

```diff
- foo,port=443 bar=42
+ foo,port=443,service=https bar=42
```

### Use case: 

Enrich the output of the sflow input.

reimda:
closes #7244


plugins/processors/port_name/port_name.go
```diff
@@ -0,0 +1,151 @@
...
+	DestTag         string `toml:"dest"`
+	DefaultProtocol string `toml:"default_protocol"`
+}
+
+func (d *PortName) SampleConfig() string {
+	return sampleConfig
+}
+
+func (d *PortName) Description() string {
+	return "Filter metrics with repeating field values"
```
danielnelson, line 41: Description needs updated

plugins/processors/port_name/port_name.go
```diff
@@ -0,0 +1,151 @@
...
+	file, err := os.Open(servicesPath())
+	if err != nil {
+		return
+	}
+	defer file.Close()
+
+	services = readServices(file)
+}
+
+func readServices(r io.Reader) sMap {
```
danielnelson, line 54: Maybe mention that this is based on the Go stdlib with reference to the file its in.

plugins/processors/port_name/port_name.go
```diff
@@ -0,0 +1,151 @@
...
+	for _, m := range metrics {
+		portProto, ok := m.GetTag(d.SourceTag)
+		if !ok {
+			// Nonexistent tag
+			continue
+		}
+		portProtoSlice := strings.SplitN(portProto, "/", 2)
+		l := len(portProtoSlice)
+
+		if l == 0 {
```
danielnelson, line 99: I'd probably try this as a switch statement and see how it looks:
```
switch len(portProtoSlice) {
default:
case 0:
case 1:
case 2:
}
```

plugins/processors/port_name/port_name.go
```diff
@@ -0,0 +1,151 @@
...
+			// Empty tag
+			continue
+		}
+
+		var port int
+		if l > 0 {
+			var err error
+			port, err = strconv.Atoi(portProtoSlice[0])
+			if err != nil {
+				// Can't convert port to string
```
danielnelson, line 109: Log this as an error.

plugins/processors/port_name/port_name.go
```diff
@@ -0,0 +1,151 @@
...
+		}
+
+		proto := d.DefaultProtocol
+		if l > 1 && len(portProtoSlice[1]) > 0 {
+			proto = portProtoSlice[1]
+		}
+
+		protoMap, ok := services[proto]
+		if !ok {
+			// Unknown protocol
```
danielnelson, line 121: I'm wondering if we should log at debug level?  I can see just wanting to ignore.

plugins/processors/port_name/port_name.go
```diff
@@ -0,0 +1,151 @@
...
+			continue
+		}
+
+		service, ok2 := protoMap[port]
```
danielnelson, line 125: I'd probably shadow the variable name for `ok`, similar to how `err` is generally shadowed.

plugins/processors/port_name/services_path_notwin.go
```diff
@@ -0,0 +1,7 @@
+// +build !windows
```
danielnelson, line 1: > A foolish consistency is the hobgoblin of little minds

With that in mind, let's rename this file to either `services_path_notwindows.go`, or `services_path_posix.go`

plugins/processors/port_name/services_path.go
```diff
@@ -0,0 +1,11 @@
+// +build windows
+
+package portname
+
+import (
+	"os"
+)
+
+func servicesPath() string {
+	return os.Getenv("WINDIR") + `\system32\drivers\etc\services`
```
danielnelson, line 10:
```
filepath.Join(os.Getenv("WINDIR"), `\system32\drivers\etc\services`)
```

plugins/processors/port_name/README.md
```diff
@@ -0,0 +1,25 @@
...
+  # dest = "service"
+
+  ## Default tcp or udp
+  # default_protocol = "tcp"
+```
+
+### Example
+
+```diff
++ measurement,port=80,service=http field=123 1560540094000000000
```
danielnelson, line 24: Add a `-` (before) line too.

plugins/processors/port_name/port_name.go
reimda: Yes, I missed that when stealing this from dedup.  fixed

plugins/processors/port_name/port_name.go
reimda: I added a comment pointing to the golang source.

plugins/processors/port_name/port_name.go
reimda: I had this as a switch initially but I couldn't make it work smoothly for both cases (port only and port and protocol separated by slash) because I need to read the port in both cases.  Since golang's switch stops evaluating after the first matched case and I didn't want to copy/paste the atoi call in both cases, I switched to ifs.  I thought it didn't look much like "idiomatic go" but it works.

plugins/processors/port_name/port_name.go
reimda: I added logging here and in most other cases in Apply.  It now logs an error if they provide the port tag but it's empty, or if it's not empty but won't parse as an int, or if they supply a protocol that isn't in the services file (meaning not tcp or udp).  It logs a debug message if the port/protocol isn't in the services file.

plugins/processors/port_name/port_name.go
reimda: see above

plugins/processors/port_name/port_name.go
reimda: makes sense

plugins/processors/port_name/services_path_notwin.go
reimda: No problem.  I renamed it _notwindows

PR #7545, opened 2020-05-19
add character selection to separating the metric name and field for graphite tag

### LINKED ISSUE #7539: Optional replacement of a dot in the metric name when using Graphite Tags
Graphite serializer (https://github.com/influxdata/telegraf/blob/master/plugins/serializers/graphite/graphite.go) when creating a metric for sending, combines the metric name with fields separated by a dot. 
This meets the specification, but is inconvenient if these metrics are later used for queries via PromQL, which does not support a dot in the metric names. 
For local use, I re-assembled telegraf by changing one line. 
```
        if field != "value" {
                out += "_" + field
        }
```
Is it possible to avoid bulkheads and make replacements using the processor? 
If not, will a pull request be accepted that adds a character selection to combine the metric name and field?

ihard:
Fix #7539

Add character selection to separating the metric name and field for graphite tag metrics.

* [x]  Signed [CLA](https://influxdata.com/community/cla/).
* [x]  Associated README.md updated.
* [x]  Has appropriate unit tests.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -1951,6 +1951,14 @@ func buildSerializer(name string, tbl *ast.Table) (serializers.Serializer, error
 		}
 	}
 
+	if node, ok := tbl.Fields["graphite_separator"]; ok {
+		if kv, ok := node.(*ast.KeyValue); ok {
+			if str, ok := kv.Value.(*ast.String); ok {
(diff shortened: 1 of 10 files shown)


ihard:
@danielnelson pls review

PR #7564, opened 2020-05-22
Update to Go 1.14.3 with testing using 1.13.11

danielnelson:
Linux/Official packages: 1.14.3 + 1.13.11
Mac: 1.13.? (unchanged)
Windows: 1.14.2

### Required for all PRs:


appveyor.yml
```diff
@@ -1,5 +1,7 @@
 version: "{build}"
 
+image: Visual Studio 2019
```
ssoroka, line 3: ?

danielnelson: This gives a different base image that has Go 1.14.2 preinstalled in C:\go, it appears there is always a version of Go installed so if we install it ourselves we need to put it in another location and adjust the PATH and perhaps some other Go environment variables.

https://www.appveyor.com/docs/windows-images-software/#golang

internal/tls/common.go
```diff
@@ -6,6 +6,7 @@ var tlsVersionMap = map[string]uint16{
 	"TLS10": tls.VersionTLS10,
 	"TLS11": tls.VersionTLS11,
 	"TLS12": tls.VersionTLS12,
+	"TLS13": tls.VersionTLS13,
```
reimda, line 9: Are these here so people can manually use TLS13 or the new ciphers?  Are we not defaulting to the newest yet because they're not available in all go versions we target?

plugins/inputs/clickhouse/clickhouse.go
```diff
@@ -205,6 +205,10 @@ func (ch *ClickHouse) Gather(acc telegraf.Accumulator) (err error) {
 	return nil
 }
 
+func (ch *ClickHouse) Stop() {
+	ch.client.CloseIdleConnections()
```
reimda, line 209: This looks unrelated to the rest of the PR.

internal/tls/common.go
ssoroka: fyi looks like go 1.12 had support for tls 1.3

plugins/inputs/clickhouse/clickhouse.go
danielnelson: I removed the Go 1.11 support code and this is moved over from `clickhouse_go1.12.go`.  I could split thee pull request?

reimda: Sorry, I missed the deleted clickhouse files.  It makes sense that this was moved from clichouse_go1.12.go.  No need to split the pr.

internal/tls/common.go
danielnelson: These are moved over from `common_go112.go`, it is possible to set some advanced TLS options in additiona to the options that are shown in the sample config: https://github.com/influxdata/telegraf/blob/master/docs/TLS.md.  Actually looks like I need to update these docs.

danielnelson:
Updated TLS documentation, removing references to Go 1.12 and older.  Also corrected documentation to indicate that the advanced settings are for the **server** config only.

PR #7569, opened 2020-05-23
Add a --once flag that will run the entire pipeline exactly once and exit

jacquesh:
I have need of #2546 so I figured I'd give implementing it a shot.
Instead of taking the approach of #6537 I decided instead to see what would happen if we didn't modify *any* existing code and just duplicated the relevant bits. As was suggested on that PR, there isn't a huge amount of code to be shared between a single run and the usual long-running loop.

Its possible that bits could be refactored out (e.g the filtering and overrides before writing to output) but even in those cases its note a huge amount of code that has been duplicated anyways.

### Required for all PRs:

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"math"
 	"os"
 	"runtime"
 	"sync"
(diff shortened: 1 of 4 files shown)


danielnelson:
I have an open PR that implements this, can you review/test #7474?

PR #7583, opened 2020-05-26
shorten config

ssoroka:
Ideas behind this change: 
- reduce duplication by providing standard methods for parsing field types.
- allow reusing fields by not deleting them
- delay unused field errors by instead adding them to a list and removing it if it's later processed
- reduce repetitious error handling by collecting errors on the config
- provide our own toml.Config settings that allows us to capture missingField errors and handle them ourselves.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -2,15 +2,14 @@ package config
 
 import (
 	"bytes"
-	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


ssoroka:
on hold until after 1.15

PR #7585, opened 2020-05-26
Add support for Solus distribution

YarmoM:
Since the pre/post install scripts check for distribution names, a lot of perfectly valid distributions are left out just because they are not "ubuntu", "debian" or "redhat". This PR simply adds a check for Solus to (un)install the systemd file.

### Required for all PRs:

### DIFF
--- a/scripts/post-install.sh
+++ b/scripts/post-install.sh
@@ -101,5 +101,8 @@ elif [[ -f /etc/os-release ]]; then
         else
             install_chkconfig
         fi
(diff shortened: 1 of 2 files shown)


danielnelson:
Thanks, I'm not familiar with Solus, which package do you install from? deb/rpm/tar?

YarmoM:
Solus uses eopkg but telegraf is not included. Therefore I build from source but that failed due to scripts not recognising the OS (which is fixed by this PR).

I've never built a package for Solus but that is something worth investigating.

PR #7596, opened 2020-05-28
7595 - Be able to add the body's content as field with http_response

### LINKED ISSUE #7595: Be able to add the body's content as field with http_response
### Proposal:

The idea is to have a way to get the content of the body as field, this feature could be disabled by default to remain backward compatible.

### Current behavior:

There is no way to get the content of the body with the plugin http_response.

### Desired behavior:

Have the ability to get the content of the body by enabling the feature by configuration.

### Use case:

If you access to an http endpoint and the server return an error, it could be interesting to have more details about the failure to eventually troubleshoot. Knowing that generally speaking, those details are available from the body of the response, it could be very helpful to provide a way to get also the body of the response.

essobedo:
closes #7595 

### Required for all PRs:

### DIFF
--- a/plugins/inputs/http_response/README.md
+++ b/plugins/inputs/http_response/README.md
@@ -31,6 +31,15 @@ This input plugin checks HTTP/HTTPS connections.
   # {'fake':'data'}
   # '''
 
+  ## Optional name of the field that will contain the body of the response. 
+  ## By default it is set to an empty String indicating that the body's content won't be added 
+  # response_body_field = ''
+
+  ## Maximum allowed HTTP response body size in bytes.
+  ## 0 means to use the default of 32MiB.
+  ## If the response body size exceeds this limit a "body_read_error" will be raised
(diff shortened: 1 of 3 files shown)


essobedo:
@danielnelson thx for your feedback. I did the corresponding changes however please note that regarding `response_body_max_size`:

1. I made it non optional with a default value of 32 Mo (it is big but we did not have any limit before this change)
2. I decided not to use `MaxBytesReader` because it seems to be for request body only such that you end up with an error message of type "http: request body too large" (as you can see [here]( https://github.com/golang/go/blob/master/src/net/http/request.go#L1158)) which would be misleading I believe so I used `io.LimitReader` instead and made it read one byte above the limit to know if the limit has been reached.

WDYT of these changes?

essobedo:
@danielnelson please note that the CI fails because of another test that doesn't seem to be related to those changes

essobedo:
@danielnelson thx

PR #7603, opened 2020-05-30
Add support for env variables to shim config

ykzts:
Enable `shim` to interpret environment variables by changing `config.parseConfig` to public.

### Required for all PRs:

### DIFF
--- a/plugins/inputs/execd/shim/goshim.go
+++ b/plugins/inputs/execd/shim/goshim.go
@@ -2,7 +2,6 @@ package shim
 
 import (
 	"bufio"
-	"bytes"
 	"context"
 	"errors"
 	"fmt"
@@ -10,7 +9,6 @@ import (
 	"io/ioutil"
 	"os"
 	"os/signal"
-	"regexp"
 	"strings"
 	"sync"
 	"syscall"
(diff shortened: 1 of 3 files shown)


danielnelson:
I'd like to add the env replacement into the shim configuration loader instead, but stick with burntsushi to parse the toml since it has fewer bugs vs naoina.

ykzts:
Rewritten to use `BurntSushi/toml`.

plugins/inputs/execd/shim/goshim.go
```diff
@@ -274,6 +282,30 @@ func LoadConfig(filePath *string) ([]telegraf.Input, error) {
 	return loadedInputs, err
 }
 
+func parseConfig(contents []byte, conf interface{}) (toml.MetaData, error) {
```
danielnelson, line 285: I know this is a little different from what we have in config.go, but let's have only the env var replacement in a new function  `expandEnvVars([]byte)`.  After calling it we can call toml.Decode as it is now.

ykzts: @danielnelson good catch!

https://github.com/influxdata/telegraf/compare/c38465e0c4512b7f3e12ef9e0b1532cf4f813bf4..d4991289c7504d3d85df7cfc718c9052038513c2

ykzts:
@ssoroka

In case of TOML, character string escape is required. So I didn't use `os.ExpandEnv`.
However, I was able to solve it by giving `os.Getenv` wrapped in the second argument of `os.Expand` (f75f4ace4e2d0c18150ea2d7c5a66b1259ace2a6).

ssoroka:
This is great. Do you mind adding a test or two so we don't accidentally break this going forward?

ykzts:
@ssoroka I added test 5bd286213d49df94634c473c51bd856439cdf3b7

ssoroka:
Thank you!

PR #7630, opened 2020-06-04
support packaging for aarch64 - same as goarch arm64 but requires arc…

sboeckelmann:
Missing aarch64 specific packaging makes it hard to install using existing packages on linux/aarch64 systems. Solution is simple because go arch build for linux/arm64 are the same. Package managers require aarch64 to run smoothly. Maybe you can pull and review this change to make aarch64 packages available to the whole community.
Thanks!

### Required for all PRs:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@
 
 #### Features
 
+- [#7634](https://github.com/influxdata/telegraf/pull/7634): Add support for streaming processors.
 - [#6905](https://github.com/influxdata/telegraf/pull/6905): Add commands stats to mongodb input plugin.
 - [#7193](https://github.com/influxdata/telegraf/pull/7193): Add additional concurrent transaction information.
 - [#7223](https://github.com/influxdata/telegraf/pull/7223): Add ability to specify HTTP Headers in http_listener_v2 which will added as tags.
@@ -59,6 +60,7 @@
 
 #### Bugfixes
 
+- [#7631](https://github.com/influxdata/telegraf/issues/7617): Fix issue with influx stream parser blocking when data is in buffer.
 - [#7371](https://github.com/influxdata/telegraf/issues/7371): Fix unable to write metrics to CloudWatch with IMDSv1 disabled.
 - [#7233](https://github.com/influxdata/telegraf/issues/7233): Fix vSphere 6.7 missing data issue.
 - [#7448](https://github.com/influxdata/telegraf/issues/7448): Remove debug fields from splunkmetric serializer.


danielnelson:
> Package managers require aarch64 to run smoothly.

Can you tell me more about the issue you are experiencing here?

sboeckelmann:
> > Package managers require aarch64 to run smoothly.
> 
> Can you tell me more about the issue you are experiencing here?

Yes, of course. Might be a getting bit lengthy, because I will give some details about the background.

We are running webservices as docker containers on AWS EC2 instances and started off with Intel (amd64) archicture using telegraf to ship performance and usage stats to influxdb.
A few weeks ago Amazon introduced new arm64 based EC2 instances that are cheaper and serving our requirements just as good as the intel based systems. So we decided to migrate our webservices to arm64 architecture to safe some money ;-)
see https://aws.amazon.com/en/ec2/instance-types/m6/

For this kind of machine we are using AWS optimized Amazon Linux AMI which is based on Fedora/CentOS.
see https://aws.amazon.com/marketplace/pp/B07KWFRJV3
Therefore the telegraf rpm package is a perfect match for us and we have been using it successfully for amd64 and are now also using it for aarch64 after building it ourselves by applying the changes from this pull request.

You might ask yourself, why does someone need an aarch64 package? Just use the arm64 it's a perfect binary match for this architecture! And, yes that's valid question because the arm64 go architecture binary runs perfectly on aarch64, because they are basically the same. But unfortunately rpm package manager does not like it, when running on aarch64:
```
# rpm -ivh ./telegraf-1.14.3-1.arm64.rpm 
Preparing...                          ################################# [100%]
	package telegraf-1.14.3-1.arm64 is intended for a different architecture
```

My change just allows to create packages with proper aarch64 architecture setup for rpm on Fedora/CentOS derived aarch64 systems, which will allow users to install it using rpm. I am not 100% sure, but I got the impression that in Fedora/CentOs the distrubutions architecture name for packaging is aarch64 while in Debian they call it arm64. So another solution might be to package rpms for aarch64 and dpkg deb for arm64. go arch is the same, it's arm64.
But like I said I am not sure if my assumption rpm=aarch64, deb=arm64 is correct. That's why I introduced it the way I did.

All I can say is that installing aarch64 package I introduced is working just fine for CentOS aarch64 ;-) And I'd rather be downloading the aarch64 rpm from your releases page in future than having to package it myself. Also I assume that more users might come across that arm64/aarch64 rpm discrepancy, therefore I felt that it was necessary to share it.

danielnelson:
Okay sounds good, I think we do want to keep the `arm64` name for Debian, perhaps we can just add another case around here that sets the package_arch value for the rpm only:

https://github.com/influxdata/telegraf/blob/62302a0f37ea0204de08a7827187b69882b0b686/scripts/build.py#L617-L618

sboeckelmann:
I like your proposal very much and will create another pull request containing your proposed solution. After a little research I could any rpm based bistro that's using arm64, they are all using aarch64 as package arch.

PR #7633, opened 2020-06-04
Add tags to snmp_trap input for context name and engine ID

reimda:
In V3 traps, the trap sender sends a context name and engine ID.  Before this change, telegraf's snmp_trap input passed a context name and engine ID to Gosnmp.  Gosnmp doesn't use these when receiving traps, only when sending, so this change removed them.  The input now also adds tags for the context name and engine ID that we receive in the trap.

Also clean up variable naming in tests.

### DIFF
--- a/plugins/inputs/snmp_trap/snmp_trap.go
+++ b/plugins/inputs/snmp_trap/snmp_trap.go
@@ -34,7 +34,6 @@ type SnmpTrap struct {
 	Version        string            `toml:"version"`
 
 	// Settings for version 3
-	ContextName string `toml:"context_name"`
 	// Values: "noAuthNoPriv", "authNoPriv", "authPriv"
 	SecLevel string `toml:"sec_level"`
 	SecName  string `toml:"sec_name"`
@@ -44,9 +43,6 @@ type SnmpTrap struct {
 	// Values: "DES", "AES", "". Default: ""
 	PrivProtocol string `toml:"priv_protocol"`
 	PrivPassword string `toml:"priv_password"`
-	EngineID     string `toml:"-"`
-	EngineBoots  uint32 `toml:"-"`
-	EngineTime   uint32 `toml:"-"`
 
 	acc      telegraf.Accumulator
 	listener *gosnmp.TrapListener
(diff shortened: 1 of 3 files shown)


plugins/inputs/snmp_trap/snmp_trap.go
```diff
@@ -34,7 +34,6 @@ type SnmpTrap struct {
 	Version        string            `toml:"version"`
 
 	// Settings for version 3
-	ContextName string `toml:"context_name"`
```
danielnelson, line 37: Usually we would leave the option, but deprecate it.  Since I expect the option isn't in wide use I suggest we remove the documentation for it too.

https://github.com/influxdata/telegraf/wiki/Deprecation#deprecate-options

plugins/inputs/snmp_trap/snmp_trap_test.go
```diff
@@ -408,11 +416,13 @@ func TestReceiveTrap(t *testing.T) {
...
-						"mib":     "SNMPv2-MIB",
-						"version": "3",
-						"source":  "127.0.0.1",
+						"oid":          ".1.3.6.1.6.3.1.1.5.1",
+						"name":         "coldStart",
+						"mib":          "SNMPv2-MIB",
+						"version":      "3",
+						"source":       "127.0.0.1",
+						"context_name": "foo_context_name",
+						"engine_id":    "6261725f656e67696e655f6964",
```
danielnelson, line 425: The `engine_id` looks high cardinality, is it different for every remote agent?  I don't know too much about how this is used, is it really important to be able to lookup on this value and if not perhaps we should skip adding it.

plugins/inputs/snmp_trap/snmp_trap.go
danielnelson: Nevermind, I see this is just added to master recently.

plugins/inputs/snmp_trap/snmp_trap_test.go
reimda: Engine ID is unique per remote agent and never changes, even if an attribute like the IP or MAC address changes.  I think context name is meant to be human readable but not change often.  For cardinality to become a problem with either you'd have to have as many unique remote agents sending traps as your cardinality ceiling.

I added them because they were available and seemed useful.  If you're querying v3 traps from a specific remote agent but its IP address isn't static, engine_id would be the best way to get what you want.

I'm open to removing either or both of them or adding settings so that they're disabled by default if you're concerned about cardinality.

danielnelson: From your description of engine_id I think it will be fine.

PR #7640, opened 2020-06-05
Execd processor

ssoroka:


plugins/processors/execd/execd.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+func (e *Execd) Add(m telegraf.Metric, acc telegraf.Accumulator) {
+	b, err := e.serializer.Serialize(m)
+	if err != nil {
+		log.Println(fmt.Errorf("Metric serializing error: %s", err))
+		return
+	}
+	// if osStdin, ok := e.process.Stdin.(*os.File); ok {
+	// 	osStdin.SetWriteDeadline(time.Now().Add(1 * time.Second))
```
ssoroka, line 97: I think I'm going to remove these comments as it's not the right place to drop writes, but I wanted feedback here as this could potentially prevent plugins from stopping properly. On the plus side a write deadline will keep us from stalling forever on broken/stuck process.

plugins/inputs/execd/execd_posix.go
```diff
@@ -6,30 +6,33 @@ import (
...
+		osProcess.Signal(syscall.SIGHUP)
 	case "SIGUSR1":
-		e.cmd.Process.Signal(syscall.SIGUSR1)
+		osProcess.Signal(syscall.SIGUSR1)
 	case "SIGUSR2":
-		e.cmd.Process.Signal(syscall.SIGUSR2)
+		osProcess.Signal(syscall.SIGUSR2)
 	case "STDIN":
-		if osStdin, ok := e.stdin.(*os.File); ok {
+		if osStdin, ok := e.process.Stdin.(*os.File); ok {
```
danielnelson, line 32: Is this right, it doesn't seem like it would ever be an os.File.

plugins/processors/execd/README.md
```diff
@@ -0,0 +1,73 @@
+# Execd Processor Plugin
+
+The `execd` processor plugin runs an external program as a separate process and pipes metrics in to the process's STDIN and reads processed metrics from its STDOUT.
```
danielnelson, line 3: Wrap lines

plugins/processors/execd/execd.go
```diff
@@ -0,0 +1,147 @@
...
+  restart_delay = "10s"
+`
+
+type Execd struct {
+	Command      []string
+	RestartDelay config.Duration
+
+	parserConfig     *parsers.Config
+	parser           parsers.Parser
+	serializerConfig *serializers.Config
```
danielnelson, line 32: How do you set the parser/serializer config?

plugins/processors/execd/execd.go
```diff
@@ -0,0 +1,147 @@
...
+		return fmt.Errorf("error creating parser: %w", err)
+	}
+	e.serializer, err = serializers.NewSerializer(e.serializerConfig)
+	if err != nil {
+		return fmt.Errorf("error creating serializer: %w", err)
+	}
+	e.acc = acc
+
+	if len(e.Command) == 0 {
+		return fmt.Errorf("FATAL no command specified")
```
danielnelson, line 71: Remove the `FATAL` bit from this error message.

plugins/processors/execd/execd.go
danielnelson: I agree it is best to block here, if the processor is not working any data sent to it will cannot be processed, and I don't think we just want to drop the data.  Later,  we may want to make it so that data not being routed through this processor can bypass without blocking.

plugins/processors/execd/execd.go
```diff
@@ -0,0 +1,147 @@
...
+		return fmt.Errorf("failed to start process %s: %w", e.Command, err)
+	}
+
+	return nil
+}
+
+func (e *Execd) Add(m telegraf.Metric, acc telegraf.Accumulator) {
+	b, err := e.serializer.Serialize(m)
+	if err != nil {
+		log.Println(fmt.Errorf("Metric serializing error: %s", err))
```
danielnelson, line 93: Here we need to use a telegraf.Logger on the Execd type

plugins/processors/execd/execd.go
```diff
@@ -0,0 +1,147 @@
...
+		return
+	}
+	// if osStdin, ok := e.process.Stdin.(*os.File); ok {
+	// 	osStdin.SetWriteDeadline(time.Now().Add(1 * time.Second))
+	// }
+
+	_, err = e.process.Stdin.Write(b)
```
danielnelson, line 100: In this processor we cannot maintain tracking metrics, because input/output is done asynchronously and we don't have any other way to tie the input to output.

Let's call metric.Drop() on every metric and document that tracking is removed in the README.

We can try to handle the tracking metrics in the future with the more tightly integrated process plugin using a different ipc method.

plugins/processors/execd/execd_test.go
```diff
@@ -0,0 +1,91 @@
...
+			now)
+		require.NoError(t, err)
+		metrics = append(metrics, m)
+		now = now.Add(1)
+
+		e.Add(m, acc)
+	}
+
+	acc.Wait(1)
+	m := acc.Metrics[0]
```
danielnelson, line 44: Use GetTelegrafMetrics(), I'm planning to remove the testutil.Metric in my Starlark branch.

I suggest setting up the expected metrics when you create the input metrics, then just calling `testutil.RequireMetricsEqual`

plugins/processors/execd/execd_test.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+	return nil
+}
+
+func fileShellScriptPath() string {
+	return "./examples/multiplier_line_protocol/multiplier_line_protocol.rb"
+}
+
+func shell() string {
+	return "ruby"
```
danielnelson, line 90: The only language we should require for tests to run is Go.

plugins/processors/execd/execd_test.go
```diff
@@ -0,0 +1,91 @@
...
+	// read the other 9 and make sure they're ordered properly
+	acc.Wait(9)
+	for i := 0; i < 9; i++ {
+		m = acc.Metrics[i+1]
+		require.EqualValues(t, metricTime+1, m.Time.UnixNano())
+		metricTime = m.Time.UnixNano()
+	}
+}
+
+func readChanWithTimeout(t *testing.T, metrics chan telegraf.Metric, timeout time.Duration) telegraf.Metric {
```
danielnelson, line 73: Not used?

internal/process/process.go
```diff
@@ -0,0 +1,162 @@
...
+
+	if err := p.Cmd.Start(); err != nil {
+		return fmt.Errorf("Error starting process: %s", err)
+	}
+
+	return nil
+}
+
+// cmdLoop watches an already running process, restarting it when appropriate.
+func (p *Process) cmdLoop(ctx context.Context) error {
```
danielnelson, line 95: Setup this function to return only once the process has exited.  This way you know when the process is finished and there won't be any calls to ReadStdoutFn later on.

internal/process/process_posix.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	"os/exec"
+	"syscall"
+	"time"
+)
+
+func gracefulStop(cmd *exec.Cmd, timeout time.Duration) {
+	go func() {
+		<-time.NewTimer(timeout).C
```
danielnelson, line 13: Look into using timer.AfterFunc for this.

plugins/processors/execd/README.md
```diff
@@ -0,0 +1,73 @@
+# Execd Processor Plugin
+
+The `execd` processor plugin runs an external program as a separate process and pipes metrics in to the process's STDIN and reads processed metrics from its STDOUT. 
+The programs must output metrics in any one of the accepted 
+[Processor Data Formats](https://github.com/influxdata/telegraf/blob/master/docs/DATA_FORMATS_INPUT.md) on its standard output.
+
+The `signal` can be configured to send a signal the running daemon on each collection interval.
```
danielnelson, line 7: We aren't supporting this right?  Also in the examples below and in the examples directory.

plugins/processors/execd/README.md
```diff
@@ -0,0 +1,73 @@
...
+```toml
+[[processor.execd]]
+  ## Program to run as daemon
+  command = ["telegraf-smartctl", "-d", "/dev/sda"]
+
+  ## Delay before the process is restarted after an unexpected termination
+  restart_delay = "10s"
+
+  ## Data format your plugin will consume AND output. 
+  ## Must be supported by both a serializer and a parser!
```
danielnelson, line 22: Make sure to update this, I think we need to split these.

plugins/processors/execd/execd.go
```diff
@@ -0,0 +1,147 @@
...
+const sampleConfig = `
+  ## Program to run as daemon
+  command = ["telegraf-smartctl", "-d", "/dev/sda"]
+
+  ## Delay before the process is restarted after an unexpected termination
+  restart_delay = "10s"
+`
+
+type Execd struct {
+	Command      []string
```
danielnelson, line 27: Add toml struct tags.

internal/process/process.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+	if len(command) > 1 {
+		p.Cmd = exec.Command(command[0], command[1:]...)
+	} else {
+		p.Cmd = exec.Command(command[0])
+	}
+	var err error
+	p.Stdin, err = p.Cmd.StdinPipe()
+	if err != nil {
+		return nil, fmt.Errorf("Error opening stdin pipe: %s", err)
```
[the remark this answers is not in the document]
ssoroka: 👍 I wonder if we could have a vet formatter that did this automatically.

internal/process/process.go
```diff
@@ -0,0 +1,162 @@
...
+			p.ReadStderrFn(p.Stderr)
+			wg.Done()
+		}()
+	}
+
+	wg.Wait()
+	return p.Cmd.Wait()
+}
+
+func isQuitting(ctx context.Context) bool {
```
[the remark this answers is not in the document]
ssoroka: Thanks, learned something new.

internal/process/process.go
```diff
@@ -0,0 +1,162 @@
...
+
+// cmdLoop watches an already running process, restarting it when appropriate.
+func (p *Process) cmdLoop(ctx context.Context) error {
+	for {
+		// Use a buffered channel to ensure goroutine below can exit
+		// if `ctx.Done` is selected and nothing reads on `done` anymore
+		done := make(chan error, 1)
+		go func() {
+			done <- p.cmdWait()
+		}()
```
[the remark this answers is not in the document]
ssoroka: Not sure what you mean. it writes the error from cmdWait to `done` when it finishes.

plugins/inputs/execd/execd_posix.go
ssoroka: yeah. I actually inspected the underlying type to figure out what it was. I was surprised, but it's definitely an *os.File.

plugins/processors/execd/execd.go
ssoroka: I have no idea. Open to suggestions here. Maybe *Config needs to be public with toml tags. At the moment I don't think anything other than influx is going to work for this processor, so it may be moot. json has the issue we previously discussed where the parse format doesn't match the serialize format. A new json parser/serializer pair would resolve that, if we're willing to go that direction.

plugins/processors/execd/execd.go
ssoroka: looks like acc.AddError() passes through to the correct log.. I'll use that for now.

plugins/processors/execd/execd.go
ssoroka: 👍 a case for metric metadata :D

plugins/processors/execd/README.md
ssoroka: needs more thought. At the moment I would change this to say it's not possible to set the parser/serializer... which isn't ideal.. but as far as i know no parser/serializer combo would work except influx

plugins/processors/execd/README.md
```diff
@@ -0,0 +1,73 @@
...
+  ## This should probably be either "json" or "influx"
+  data_format = "json"
+```
+
+### Example
+
+##### Go daemon using SIGHUP
+
+```go
+package main
```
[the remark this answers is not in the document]
ssoroka: yeah. I think this just got copied over from the input execd and ignored. good catch.

internal/process/process.go
ssoroka: This function looks more complicated than it is. rewriting it.

plugins/processors/execd/examples/count.go
```diff
@@ -0,0 +1,24 @@
+package main
```
reimda, line 1: The only reason you'd write a execd processor is to modify metrics.  This example doesn't do that so I'm concerned that it will mislead beginners.

plugins/inputs/execd/execd_windows.go
```diff
@@ -32,7 +31,3 @@ func (e *Execd) Gather(acc telegraf.Accumulator) error {
 
 	return nil
 }
-
```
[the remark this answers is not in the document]
ssoroka: I think so. The posix version supports signals that aren't supported in windows.

plugins/processors/execd/examples/count.go
ssoroka: removed

plugins/processors/execd/execd.go
```diff
@@ -0,0 +1,148 @@
...
+
+	_, err = e.process.Stdin.Write(b)
+	if err != nil {
+		return fmt.Errorf("error writing to process stdin: %w", err)
+	}
+
+	// We cannot maintain tracking metrics at the moment because input/output
+	// is done asynchronously and we don't have any metric metadata to tie the
+	// output metric back to the original input metric.
+	m.Drop()
```
danielnelson, line 104: It occurs to me that we need to work out the protocol when this function returns an error.  I suppose we should call Drop() in the agent when an error is returned, can you add this?  agent.go:L512 and also update the StreamingProcessor docs

plugins/processors/execd/execd.go
```diff
@@ -0,0 +1,148 @@
...
+	return nil
+}
+
+func (e *Execd) cmdReadOut(out io.Reader) {
+	scanner := bufio.NewScanner(out)
+
+	for scanner.Scan() {
+		metrics, err := e.parser.Parse(scanner.Bytes())
+		if err != nil {
+			log.Println(fmt.Errorf("Parse error: %s", err))
```
danielnelson, line 119: telegraf.Logger, here and below

plugins/processors/execd/execd.go
```diff
@@ -0,0 +1,148 @@
...
+	"github.com/influxdata/telegraf/config"
+	"github.com/influxdata/telegraf/internal/process"
+	"github.com/influxdata/telegraf/plugins/parsers"
+	"github.com/influxdata/telegraf/plugins/processors"
+	"github.com/influxdata/telegraf/plugins/serializers"
+)
+
+const sampleConfig = `
+  ## Program to run as daemon
+  command = ["telegraf-smartctl", "-d", "/dev/sda"]
```
danielnelson, line 20: Let's come up with a better example command, one that could plausibly work.  Maybe something like `tee -a /dev/null`?

testutil/accumulator.go
```diff
@@ -55,6 +55,8 @@ func (a *Accumulator) NMetrics() uint64 {
 }
 
 func (a *Accumulator) GetTelegrafMetrics() []telegraf.Metric {
+	a.Lock()
```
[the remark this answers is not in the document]
ssoroka: it doesn't really matter; if anything uses the accumulator concurrently, then all access to internals need to be locked, or -race will fail.

testutil/accumulator.go
```diff
@@ -251,6 +253,8 @@ func (a *Accumulator) SetDebug(debug bool) {
 
 // Get gets the specified measurement point from the accumulator
 func (a *Accumulator) Get(measurement string) (*Metric, bool) {
+	a.Lock()
+	defer a.Unlock()
```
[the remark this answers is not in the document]
ssoroka: same answer as previous. this was just to remove races.

internal/process/process_posix.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"os/exec"
+	"syscall"
+	"time"
+)
+
+func gracefulStop(cmd *exec.Cmd, timeout time.Duration) {
+	time.AfterFunc(timeout, func() {
+		if cmd == nil || cmd.ProcessState == nil {
```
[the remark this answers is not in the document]
ssoroka: I guess you're right about that, but I did have some panics that lead to me adding this. It could be that the panic was only around cmd.ProcessState, but I don't recall.

plugins/processors/execd/execd.go
ssoroka: how about just `cat`?

danielnelson: that works

PR #7644, opened 2020-06-06
Document that string fields do not produce prometheus metrics

### LINKED ISSUE #7623: Prometheus plugin log
## Feature Request

### Proposal:

Without "int" type on data flow, the prometheus_client return no data, without any log, I guess it could be very usefull to have some trace about this behavior.

### Current behavior:

An input conf like this : 
```
[[inputs.tail]]
  files = ["./data.sample"]
  data_format = "grok"
  grok_patterns = ['^%{LOGLEVEL:logLevel}\s*%{DATE:date} %{TIME:time}\s*%{NOTSPACE:fieldName}\s*Value = %{INT:value}$']
 
```
With output like this :
```
[[outputs.prometheus_client]]
  listen = ":9273"
```

Grok pattern match with lines in file, but no data are provided on page http://loclahost:9273/metrics and no log about what and why happens

### Desired behavior:

In the same configuration seen before, just one log line could be very usefull, something like :

`
2020-06-03T05:05:00Z D! [outputs.prometheus_client] No integer in the data
`
### Use case: 

Debuging this situation/configuration, burns couple of hours, but with this logs, I spend a couple of minutes I guess.

danielnelson:
closes #7623

### Required for all PRs:

### DIFF
--- a/plugins/serializers/prometheus/README.md
+++ b/plugins/serializers/prometheus/README.md
@@ -10,7 +10,7 @@ somewhat, but not fully, mitigated by using outputs that support writing in
 "batch format".  When using histogram and summary types, it is recommended to
 use only the `prometheus_client` output.
 
-## Configuration
+### Configuration
 
 ```toml
 [[outputs.file]]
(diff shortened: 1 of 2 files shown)


plugins/serializers/prometheus/README.md
```diff
@@ -35,6 +35,19 @@ use only the `prometheus_client` output.
   data_format = "prometheus"
 ```
 
+### Metrics
+
+A Prometheus metrics is created for each integer, float, boolean or unsigned
```
reimda, line 40:
```suggestion
A Prometheus metric is created for each integer, float, boolean or unsigned
```

PR #7652, opened 2020-06-08
Remove trailing backslash from tag keys/values

### LINKED ISSUE #7558: Tag value cannot end with \
This can be tried just by using Chronograf or any other means to write points (so is probably related to InfluxDB and not with telegraf)

### System info:
OS: Windows 10
InfluxDB: 1.8

### Steps to reproduce:
Just try to write the following points
```
#this does not work (ends with backslash)
Test, Tag1=C:\\,Tag2=Something Field1=10
#this works (ends with space)
Test,Tag1=C:\\\ ,Tag2=Something Field1=10
```

### Expected behavior:

being able to write tags that end with "\\"

### Actual behavior:

A point with a tag that ends with "\\" cannot be written (a path could end with the backslash in windows), but other escaped chars are valid (ie: space).
The error is: _invalid tag format_

### Additional info:
I'm not sure this is a telegraf related problem, it might be related to InfluxDB itself

danielnelson:
Currently a tag key or value with a trailing backslash will produce an error like:
```
 E! [outputs.influxdb] When writing to [http://localhost:8086]: received error unable to parse 'mem,host=loaner,path=C:\ free=3166146560i 1591645472000000000': invalid tag format; discarding points
```

With this change the trailing slashes are stripped.  Other methods were considered include using a space or a replacement char such as` _`.  I decided not to use a space since it might cause confusion when writing queries by hand and in the cli.  I decided against underscore because due to how ugly it is.

I know ideally we would have some form of https://github.com/influxdata/influxdb/issues/6037, but I believe this needs to be worked out separately from a fix for the current line protocol.

closes #7558

### Required for all PRs:

### DIFF
--- a/plugins/serializers/influx/influx_test.go
+++ b/plugins/serializers/influx/influx_test.go
@@ -388,6 +388,38 @@ var tests = []struct {
 		output: []byte("disk,path=/ value=42i 0\n"),
 	},
 	{
+		name: "tag key backslash is trimmed and removed",
+		input: MustMetric(
+			metric.New(
+				"disk",
+				map[string]string{
+					`\`: "example.org",
+				},
+				map[string]interface{}{
+					"value": 42,
+				},
+				time.Unix(0, 0),
+			),
+		),
+		output: []byte("disk value=42i 0\n"),
+	},
+	{
+		name: "tag value backslash is trimmed and removed",
+		input: MustMetric(
+			metric.New(
+				"disk",
+				map[string]string{
+					"host": `\`,
+				},
+				map[string]interface{}{
+					"value": 42,
+				},
+				time.Unix(0, 0),
+			),
+		),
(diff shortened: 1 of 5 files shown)


plugins/serializers/influx/influx_test.go
```diff
@@ -323,6 +323,70 @@ var tests = []struct {
...
+		),
+		output: []byte("cpu value=42i 0\n"),
+	},
+	{
+		name: "empty tag key is removed",
+		input: MustMetric(
+			metric.New(
+				"cpu",
+				map[string]string{
+					"": "example.org",
```
ssoroka, line 348: might also add a test for key `\` just to be thorough

danielnelson: Fixed

danielnelson:
I added documentation about these corner cases.

PR #7656, opened 2020-06-09
Flaky shim test

ssoroka:
pre-patch was seeing this fail about 1 in 50. 
now passes 10000 test runs with no failures.

### DIFF
--- a/plugins/inputs/execd/shim/goshim.go
+++ b/plugins/inputs/execd/shim/goshim.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"os/signal"
 	"strings"
+	"sync"
 	"syscall"
 	"time"
 
@@ -83,6 +84,7 @@ func (s *Shim) Run(pollInterval time.Duration) error {
 
 	s.metricCh = make(chan telegraf.Metric, 1)
 
+	wg := sync.WaitGroup{}
 	quit := make(chan os.Signal, 1)
 	signal.Notify(quit, syscall.SIGINT, syscall.SIGTERM)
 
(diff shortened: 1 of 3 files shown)


plugins/inputs/execd/shim/shim_test.go
```diff
@@ -172,3 +160,15 @@ func (i *serviceInput) Start(acc telegraf.Accumulator) error {
 
 func (i *serviceInput) Stop() {
 }
+
+// we can get stuck if stdout gets clogged up and nobody's reading from it.
+// make sure we keep it going
+func readUntilEmpty(r *bufio.Reader) {
```
danielnelson, line 166: Shouldn't the shim code we are testing be doing this?

plugins/inputs/execd/shim/goshim.go
```diff
@@ -106,19 +104,17 @@ func (s *Shim) Run(pollInterval time.Duration) error {
...
 		s.gatherPromptChans = append(s.gatherPromptChans, gatherPromptCh)
-		wg.Add(1)
 		go func(input telegraf.Input) {
 			startGathering(ctx, input, acc, gatherPromptCh, pollInterval)
 			if serviceInput, ok := input.(telegraf.ServiceInput); ok {
 				serviceInput.Stop()
 			}
 			close(gatherPromptCh)
-			wg.Done()
+			close(s.metricCh)
```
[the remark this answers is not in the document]
ssoroka: oh maybe that's why I had the waitgroup in there. Clearly not obvious enough. .. will take a look at that. maybe it needs a test.

plugins/inputs/execd/shim/shim_test.go
ssoroka: no, because this is the read end of the out pipe coming from the shim.

Issue #7679, opened 2020-06-14
x509_cert: add support for SMTP certificates over STARTTLS


tryfail:
## Feature Request

Some mail servers I support, cannot be monitored through telegraf's x509_cert input plugin. This happens, because these services only expose certificates after receiving a STARTTLS SMTP message.

I believe, this will also make the plugin more complete as per its current description.

### Proposal:

Add support for SMTP+STARTTLS through telegraf's x509_cert input plugin.

### Current behavior:

Using a source like tcp://smtp.example.com:587, will produce the following error:

[inputs.x509_cert] Error in plugin: cannot get SSL cert 'tcp://smtp.example.com:587': tls: first record does not look like a TLS handshake

### Desired behavior:

Using something like smtp+starttls://smtp.example.com:587 will produce the certificates as usual for HTTPS connections and other TLS-wrapped protocols.

### Use case: 

The use case is to support certificate monitoring for SMTP servers that do not have a TLS-wrapped endpoint and will only support TLS with STARTTLS in the long term.

danielnelson:
I not sure we want this plugin to become a plugin that speaks various protocols to extract the certificate, outside of the very common tcp handshake special case.  There are so many potential ways you could get a certificate.  Wouldn't be better to run Telegraf on the SMTP server and report the certificate from there as a local file?

danielnelson:
> Alternatively, do you believe this would make more sense as a separate plugin (e.g. x509_cert_smtp)?

No, definitely not a standalone plugin.

> For example, I would also be very interested in TLS validation of UDP OpenVPN servers, but that is complex on another level. This is the kind of thing that I imagine the plugin not supporting.

I guess this is what I'm struggling to decide on, how many protocols for obtaining the certificate to we want to support and what should be considered in scope and out of scope for the plugin.

Right now we have file system and tcp/https (also apparently udp?).  However there are many ways the certificate could be fetched over the network: SMTP+STARTTLS, LDAP+STARTTLS, OpenVPN, etc.

I don't think we want to say that we will allow any protocol.  I'd like to come up with a strategy for deciding what this plugin should and should not handle that we can apply consistently.

@reimda @ssoroka What are your thoughts on how much to expand this plugin?

reimda:
x509_cert seems like the right place to add new cert sources.  I don't see a problem adding lots of new protocols to it as long as we can keep the configuration clear.

[...]

PR #7680, opened 2020-06-14
add support for SMTP server certificates over STARTTLS

tryfail:
### Required for all PRs:

closes https://github.com/influxdata/telegraf/issues/7679 .

I am not too familiar with the unit test structure for golang and telegraf. Additionally, I did not find a way to create an analogous test to that of the local TLS server.

Currently known limitations:
- slower execution for mail tarpits with slow replies like OpenBSD's spamd. Still works fine with the default timeout
- cannot define ehlo/helo values (similar situation prior to adding SNI support in this plugin)
  - could be incompatible with some SMTP server non-default configurations which may demand a specific EHLO parameter that is different than the FQDN. So far, compatible with outlook, google, servers I care for

### DIFF
--- a/plugins/inputs/x509_cert/x509_cert.go
+++ b/plugins/inputs/x509_cert/x509_cert.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net"
+	"net/smtp"
 	"net/url"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


plugins/inputs/x509_cert/x509_cert.go
```diff
@@ -100,6 +101,58 @@ func (c *X509Cert) getCert(u *url.URL, timeout time.Duration) ([]*x509.Certifica
...
+		id, err := sclient.Text.Cmd("STARTTLS")
+		if err != nil {
+			return nil, err
+		}
+
+		sclient.Text.StartResponse(id)
+		defer sclient.Text.EndResponse(id)
+		_, _, err = sclient.Text.ReadResponse(220)
+		if err != nil {
+			return nil, fmt.Errorf("did not get 220 after STARTTLS: %s", err.Error())
```
danielnelson, line 143: Nit: switch over to the new %w format for wrapping errors
```
"did not get 220 after STARTTLS: %w", err)
```

srebhan:
@tryfail any chance you can work on the comments?

powersj:
Hi,

Thanks!

PR #7691, opened 2020-06-16
feat: add additional tags/fields to pod metrics

rawkode:
Add state reason, which can eventually deprecated `terminated_reason`, to pod_container metrics; as well as a tag to represent the readiness of the container.

### Required for all PRs:

### DIFF
--- a/plugins/inputs/kube_inventory/pod.go
+++ b/plugins/inputs/kube_inventory/pod.go
@@ -3,7 +3,7 @@ package kube_inventory
 import (
 	"context"
 
-	"github.com/ericchiang/k8s/apis/core/v1"
+	v1 "github.com/ericchiang/k8s/apis/core/v1"
 
 	"github.com/influxdata/telegraf"
 )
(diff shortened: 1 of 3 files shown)


rawkode:
I've removed `terminated_status` from the `README.md`, but it is still omitted from the plugin.

My thoughts were that we can hide / discourage it's use from the README and deprecate in favour of `state_reason` later.

Would like feedback

plugins/inputs/kube_inventory/pod.go
```diff
@@ -37,30 +37,42 @@ func (ki *KubernetesInventory) gatherPod(p v1.Pod, acc telegraf.Accumulator) err
...
+		stateReason = cs.State.Terminated.GetReason()
 	case cs.State.Waiting != nil:
 		stateCode = 2
 		state = "waiting"
+		stateReason = cs.State.Waiting.GetReason()
+	}
+
+	ready := "unready"
+	if cs.GetReady() {
+		ready = "ready"
```
danielnelson, line 59: Consider naming this `readiness`, might flow better than `read=unready`

plugins/inputs/kube_inventory/pod.go
```diff
@@ -37,30 +37,42 @@ func (ki *KubernetesInventory) gatherPod(p v1.Pod, acc telegraf.Accumulator) err
 
 func gatherPodContainer(nodeName string, p v1.Pod, cs v1.ContainerStatus, c v1.Container, acc telegraf.Accumulator) {
 	stateCode := 3
+	stateReason := ""
```
danielnelson, line 40: When this string is empty don't add it to the fields.

rawkode:
Done. Thank you, @danielnelson

rawkode:
Thanks, @danielnelson

PR #7697, opened 2020-06-17
Export internal/tls package for use in execd plugins

### LINKED ISSUE #7687: Make telegraf/internal/tls public interface for plugins written using execd/shim
## Feature Request
Make telegraf/internal/tls public interface for plugins written using execd/shim

Expose telegraf/internal/tls for plugins written using execd/shim, at the moment, execd/shim doesn't provide same capabilities s when plugin is in the core. For example vsphere input plugin, relies on tls parser from Telegraf core, thus it's impossible to move it to execd/shim architecture.

### Proposal:
Make telegraf/internal/tls public, thus execd/shim plugins can re-use it, rather then re-implement.

### Current behavior:
At the moment attempt to add TLS support into plugin configuration file lead to **use of internal package github.com/influxdata/telegraf/internal/tls not allowed** error during build.

### Desired behavior:
Make use of tls client/server configuration settings from Telegraf core possible

### Use case: 
Make use of tls client/server configuration settings from Telegraf core possible, like below:
```
import "github.com/influxdata/telegraf/internal/tls"

type pluginConfig struct {
  // Mix in the TLS/SSL goodness from core
  tls.ClientConfig

  Log telegraf.Logger
}
```

danielnelson:
closes #7687

### Required for all PRs:


reimda:
Looking in internal/ there is other code that is useful in plugins.  A quick grep shows plugins using tls most but also rcon, limiter, docker, globpath, syslog, rotate, etc.  internal.Duration is used everywhere.

If we fix just tls we're going to have other execd plugin writers request these as they need them.  Should we make a more comprehensive fix that covers more of the code in internal?

danielnelson:
Some of these should be exposed but I can do them in separate pull requests.

ssoroka:
fyi @reimda internal.Duration is deprecated in favor of config.Duration, it just hasn't been removed from the project yet.

PR #7701, opened 2020-06-17
tmp files should have their own folders

ssoroka:
Based on a conversation from community slack, it was requested that sysstat create its own folder rather than dumping temp files in the parent /tmp dir.

### DIFF
--- a/plugins/inputs/sysstat/sysstat.go
+++ b/plugins/inputs/sysstat/sysstat.go
@@ -7,6 +7,7 @@ import (
 	"encoding/csv"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
 	"os/exec"
 	"path"
@@ -178,7 +179,12 @@ func (s *Sysstat) collect() error {
 	for _, act := range s.Activities {
 		options = append(options, "-S", act)
 	}
-	s.tmpFile = path.Join("/tmp", fmt.Sprintf("sysstat-%d", time.Now().Unix()))
+
+	var err error
+	s.tmpFile, err = createTempFile()
+	if err != nil {
+		return err
+	}
 	// collectInterval has to be smaller than the telegraf data collection interval
 	collectInterval := s.interval - parseInterval
 
@@ -197,6 +203,17 @@ func (s *Sysstat) collect() error {
 		return fmt.Errorf("failed to run command %s: %s - %s", strings.Join(cmd.Args, " "), err, string(out))
 	}
 	return nil
+}
+
+func createTempFile() (string, error) {
+	tmpPath := path.Join(os.TempDir(), "tg-sysstat")
+	os.Mkdir(tmpPath, os.ModeDir)
+	f, err := ioutil.TempFile(tmpPath, "sysstat")
+	if err != nil {
+		return "", fmt.Errorf("unable to create temp file in path %s: %w", tmpPath, err)
+	}
+	defer f.Close()
+	return f.Name(), nil
 }
 
 func filterEnviron(env []string, prefix string) []string {


plugins/inputs/sysstat/sysstat.go
```diff
@@ -199,6 +205,17 @@ func (s *Sysstat) collect() error {
...
 }
 
+func createTempFile() (string, error) {
+	tmpPath := path.Join(os.TempDir(), "tg-sysstat")
+	os.Mkdir(tmpPath, os.ModeDir)
+	f, err := ioutil.TempFile(tmpPath, "sysstat")
+	if err != nil {
+		return "", fmt.Errorf("unable to create temp file in path %s: %w", tmpPath, err)
+	}
+	defer f.Close()
```
danielnelson, line 215: If we close the file here, we won't be able to use it.  Best to run this on a real system.

plugins/inputs/sysstat/sysstat.go
```diff
@@ -199,6 +205,17 @@ func (s *Sysstat) collect() error {
 	return nil
 }
 
+func createTempFile() (string, error) {
+	tmpPath := path.Join(os.TempDir(), "tg-sysstat")
+	os.Mkdir(tmpPath, os.ModeDir)
```
danielnelson, line 210: This directory needs to be removed, but since it isn't randomized we can't safely do this if there are multiple plugin instances running.  I suggest we use ioutil.TempDir and place the file inside.

plugins/inputs/sysstat/sysstat.go
```diff
@@ -178,7 +179,12 @@ func (s *Sysstat) collect() error {
 	for _, act := range s.Activities {
 		options = append(options, "-S", act)
 	}
-	s.tmpFile = path.Join("/tmp", fmt.Sprintf("sysstat-%d", time.Now().Unix()))
+
+	var err error
+	s.tmpFile, err = createTempFile()
```
danielnelson, line 184: Look into having this either not be a struct field, or perhaps leave it on the struct and reuse a directory each interval, removing only when the plugin is stopped.

plugins/inputs/sysstat/sysstat.go
ssoroka: the goal is to let another process write into it so Telegraf can read it later

takeahike: There appears to be a race condition when Sophos antivirus tries to read the file and it is removed. Since the files are created and removed very quickly this is causing issues.  We would like to choose an arbitrary location for these files which we can then exclude from the active scans.

plugins/inputs/sysstat/sysstat.go
ssoroka: if Telegraf always reuses the same dir, there's no reason to ever remove it.

plugins/inputs/sysstat/sysstat.go
takeahike: Using a glob exclusion in Sophos such as /tmp/sysstat* does not work unfortunately, and we really would like to not have to exclude all of /tmp

ssoroka: I'm bumping this up to the top as it's related to why this was opened.

danielnelson: @takeahike I see, the old code is indeed not creating the file correctly because it uses the timestamp to make the file "random", this could be predicted and I'm not sure exactly what could be done, sometimes this is a security issue.  At any rate the current code is not follow best tempfile practices, we need to fix this but I wouldn't necessarily exclude it from the scan.

plugins/inputs/sysstat/sysstat.go
danielnelson: We should always clean up our temporary files.

plugins/inputs/sysstat/sysstat.go
danielnelson: @ssoroka Do we need a temporary file here at all?  Maybe sadf can write to stdout?

plugins/inputs/sysstat/sysstat.go
ssoroka: You mean the directory, right? The files are getting removed as far as I can tell.

plugins/inputs/sysstat/sysstat.go
ssoroka: that'd be a lot cleaner.

plugins/inputs/sysstat/sysstat.go
danielnelson: Yes, all temporary files and directories should be cleaned up before Telegraf closes.

plugins/inputs/sysstat/sysstat.go
danielnelson: It seems like it would be possible, read the stdout from sadc and feed it back in to all sadf processes on stdin,  but a bit larger change so we could do it later.

I also read in `man sadc`:
> In this case, if the file already exists, sadc will overwrite it if it is from a previous month.

So we don't really need to recreate the tempfile, we could just use the same one each iteration and then clean up the directory and the file once when the plugin stops.

sjwang90:
@takeahike It's been awhile since this PR has been worked on - is this still an issue you're running into?

plugins/inputs/sysstat/sysstat.go
reimda: Sysstat's sadf program doesn't have the ability to read from stdin. It can read from an arbitrary file or daily files in a system location or in an arbitrary directory. On linux, if you try to force it to read from stdin by giving /proc/self/fd/0 as the filename, it fails with an "illegal seek" error.

It looks like this plugin is stuck with using temp files.

PR #7708, opened 2020-06-19
Create aggregate stats of some basic system stats

johnseekins:
### Required for all PRs:

When we had 600+ hosts with 72 disks a piece, getting the aggregate usage of disks across the hosts was causing performance problems. Adding these aggregate stats lets us query one value per host to get aggregate data on disk usage, IO stats, and networking information. This should significantly improve performance for larger queries without impacting existing collection.

We weren't able to find a way with existing aggregator/processor plugins to replicate this function because it requires reducing the dimensions (tags) that the resulting data has. Doing a general aggregation also wouldn't be valid for `disk` metrics, because the `used_percent` field would balloon to >100% quickly.


danielnelson:
Thanks for the pull request, but the way we would recommend collecting these types of metris is by using one of the aggregator plugins such as the [basicstats](https://github.com/influxdata/telegraf/tree/master/plugins/aggregators/basicstats) aggregator.  This allows us to decouple aggregation so it can be used optionally with any plugin.

johnseekins:
As I mentioned in my initial comment: 

```
We weren't able to find a way with existing aggregator/processor plugins to replicate this function because it requires reducing the dimensions (tags) that the resulting data has. Doing a general aggregation also wouldn't be valid for disk metrics, because the used_percent field would balloon to >100% quickly.
```

I added this code because `basicstats` and other aggregators didn't work correctly for this use case.

danielnelson:
> reducing the dimensions (tags) that the resulting data has

Here is an example of how to aggregate across tags:
```
disk,host=a,device=sda free=25i
disk,host=a,device=sdb free=50i
```

If you want to aggregate fields across all devices, remove the device tag using `tagexclude`:
```
disk,host=a free=25i
disk,host=a free=50i
```

Now if you do a `sum` you will get:
```
disk,host=a free=75i
```

If you need a mixture of aggregation types then you may need to define the aggregator multiple times for specific fields.

I hope that helps, if I'm not understanding correctly though can you show an example of what exactly you are trying?

Issue #5687, opened 2019-04-05
GPFS Statistics Input


russorat:
## Feature Request

Opening a feature request kicks off a discussion.

### Proposal:
Add a new input for gathering GPFS Statistics (available with GPFS’s mmpmon) including read/write throughput, read/write operations (IOps), read/write latency, and GPFS Waiters.

https://www.ibm.com/support/knowledgecenter/en/STXKQY_5.0.2/com.ibm.spectrum.scale.v5r02.doc/bl1adv_mmpmonch.htm

### Current behavior:
No plugin exists

### Desired behavior:
Plugin exists to pull the metrics mentioned above

### Use case: [Why is this important (helps with prioritizing requests)]
user would like to gather GPFS Statistics for the purpose of monitoring and alerting

powersj:
Hi,

Have not seen any recent requests for this going to close for now and can revisit if a strong desire to shown.

Thanks

PR #7709, opened 2020-06-20
Spectrum scale plugin

hanleyja:
### Required for all PRs:

We've been using a version of this for a while. I saw https://github.com/influxdata/telegraf/issues/5687 and figured it might help some others. A list of all possible sensors is probably out of scope for this PR, but I've included some that we've found useful.

### DIFF
--- a/plugins/inputs/spectrum_scale/README.md
+++ b/plugins/inputs/spectrum_scale/README.md
@@ -1,0 +1,29 @@
+# Spectrum Scale (GPFS) Input Plugin
+
+[Spectrum Scale](https://www.ibm.com/ca-en/marketplace/scale-out-file-and-object-storage) input plugin gathers metrics from GPFS.
+
+This plugin requires read-write access to Spectrum Scale monitoring socket (`/var/mmfs/mmpmon/mmpmonSocket`).
+This can be accomplished by granting telegraf root privileges, but an appropriate ACL should be sufficient.
+Tested against versions 4.2+ and 5.0+
+
+### Configuration:
+
+```toml
+# SampleConfig
+[[inputs.spectrum_scale]]
+  # An array of Spectrum Scale (GPFS) sensors
+  #
+  # These will be monitored through the mmpmon socket
+  sensors = ["fis", "nsd_ds", "vfss"]
+  socketLocation = "/var/mmfs/mmpmon/mmpmonSocket"
+
+```
+
+#### `sensors`
+
+List of spectrum scale subsystems to query. Examples include: `fis`, `nsd_ds`, and `vfss`
+
+#### `socketLocation`
+
+Location of the mmpmon monitoring socket. By default, `/var/mmfs/mmpmon/mmpmonSocket`
+
--- a/plugins/inputs/spectrum_scale/spectrum_scale.go
+++ b/plugins/inputs/spectrum_scale/spectrum_scale.go
@@ -1,0 +1,420 @@
+package spectrum_scale
+
+import (
+	"bufio"
+	"fmt"
+	"net"
+	"os"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/plugins/inputs"
+)
+
+type spectrum_scale struct {
+	Sensors        []string
+	SocketLocation string
+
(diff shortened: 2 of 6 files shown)


hanleyja:
Hi, I wanted to check in and see if I missed something or if this might land over the next couple of releases. Thanks!

plugins/inputs/spectrum_scale/README.md
```diff
@@ -0,0 +1,29 @@
...
+```
+
+#### `sensors`
+
+List of spectrum scale subsystems to query. Examples include: `fis`, `nsd_ds`, and `vfss`
+
+#### `socketLocation`
+
+Location of the mmpmon monitoring socket. By default, `/var/mmfs/mmpmon/mmpmonSocket`
+
```
reimda, line 29: Could you add an example metric and a list of tags and fields per sensor?  If the lists are very long, could you find a link to spectrum docs where we can find them?

Looking at the tests I see around 5 tags and 20 fields.  I think that could fit here.

plugins/inputs/spectrum_scale/spectrum_scale.go
```diff
@@ -0,0 +1,416 @@
...
+// Returns one of the errors encountered while gather stats (if any).
+func (s *Spectrum_Scale) Gather(acc telegraf.Accumulator) error {
+	var sensors []string
+	var socketLocation string
+
+	sensors = s.Sensors
+	socketLocation = s.SocketLocation
+
+	if len(sensors) == 0 {
+		sensors = []string{"nsd_ds", "gfis", "fis"}
```
reimda, line 49: Put defaults logic in Init() so you don't have to do it every gather.

plugins/inputs/spectrum_scale/spectrum_scale.go
```diff
@@ -0,0 +1,416 @@
...
+	}
+
+	if len(socketLocation) == 0 {
+		socketLocation = "/var/mmfs/mmpmon/mmpmonSocket"
+	}
+
+	// For every sensor specified, connect to the socket and parse the result
+	for _, sensor := range sensors {
+		err := s.getSensor(socketLocation, sensor, acc)
```
reimda, line 58: I'm not familiar with spectrums socket api so this might be a dumb question, but is there a way to get more than one sensor at once?  Iterating over all the sensors might mean the data returned for each sensor isn't from the exact same time which could be confusing.

plugins/inputs/spectrum_scale/spectrum_scale_test.go
```diff
@@ -0,0 +1,274 @@
...
+	socketPath := tempdir + socketName
+
+	t.Logf("socketPath: %s", socketPath)
+
+	go socketHelper(t, socketPath)
+
+	// Initialize "server"
+	time.Sleep(2 * time.Second)
+
+	verifySensorGfis(t, socketPath)
```
reimda, line 268: I'd prefer to see this structured as a table driven test instead of these three verify functions. https://github.com/golang/go/wiki/TableDrivenTests https://dave.cheney.net/2019/05/07/prefer-table-driven-tests

Table tests make the test easier to read because the input data and output data is in the same place and separate from the test code.  They're also easier to extend.

sjwang90:
@hanleyja Can you please take a look at the latest comments?

sjwang90:
Hi @hanleyja, checking in if you're still working on this plugin. If you are willing to submit this as an [external plugin](https://github.com/influxdata/telegraf/blob/master/docs/EXTERNAL_PLUGINS.md) that can be used with `execd` to run seamlessly with Telegraf, Telegraf users would be able to use the plugin sooner since it wouldn't have to go through the typical review process. 

Feel free to respond with any questions you may have.

sjwang90:
Due to inactivity I'm going to close the issue for now, but we'd love to see this as an external plugin.

BlueskyFR:
Hi! Is there any way this can be reopened?

MyaLongmire:
@BlueskyFR are you interested in taking over this pr? it looks like @hanleyja never got back to us.

hanleyja:
We're still using a version of this on our internal build. Haven't had the cycles to update the PR.

MyaLongmire:
@hanleyja would you like to continue this pr?

PR #7715, opened 2020-06-23
Fix incorrect Azure SQL DB server properties

denzilribeiro:
The sql_server_properties measure had a few incorrect values for Azure SQL DB
a. Number of DB's online/offline/suspect is irrelevant for a single DB and was counting master had incorrect values of 2
b. Space usage (Max/unallocated) was also set to NULL prior
c. Given all the 3 parts of this collector required adhoc queries, eliminated the unnecessarily table variable to simplify query

### Required for all PRs:

### DIFF
--- a/plugins/inputs/sqlserver/sqlserver.go
+++ b/plugins/inputs/sqlserver/sqlserver.go
@@ -502,111 +502,109 @@ END
 const sqlServerPropertiesV2 = `
 SET DEADLOCK_PRIORITY -10;
 DECLARE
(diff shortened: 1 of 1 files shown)


denzilribeiro:
@Trovalo can you take quick look.

Trovalo:
I've tested it on on-prem and the output is the same. 
The only thing I would change is some indentation, so if you print out the query is displays a bit more nicely. (if you want I can make a review for this)

Issue #7622, opened 2020-06-03
Telegraf 1.13.4-1  sarama lib "because consuming was taking too long" messages


avneradania:
After upgraded telegraf to 1.13.4-1, lots of messages in log file regarding the time it take to consume from kafka. 

 # Relevant telegraf.conf:
[global_tags]

[agent]
  hostname = "telegraf-host"
  omit_hostname = false
  interval = "60s"
  round_interval = true
  metric_batch_size = 10000
  metric_buffer_limit = 200000
  collection_jitter = "0s"
  flush_interval = "60s"
  flush_jitter = "1s"
  precision = "200ms"
  debug = true
  quiet = false
  logfile = "/var/log/telegraf/telegraf.log"

### System info:
telegraf version: Telegraf 1.13.4 (git: HEAD ffabd6b5)
Ubuntu 3.13.0-158-generic

### Additional info:

2020-03-30T10:11:02Z D! [sarama] consumer/broker/2 abandoned subscription to telegraf_metrics/13 because consuming was taking too long
2020-03-30T10:11:02Z D! [sarama] consumer/broker/2 abandoned subscription to telegraf_metrics/1 because consuming was taking too long
2020-03-30T10:11:02Z D! [sarama] consumer/broker/2 abandoned subscription to telegraf_metrics/31 because consuming was taking too long
2020-03-30T10:11:02Z D! [sarama] consumer/broker/2 abandoned subscription to telegraf_metrics/37 because consuming was taking too long
2020-03-30T10:11:02Z D! [sarama] consumer/broker/2 abandoned subscription to telegraf_metrics/19 because consuming was taking too long
2020-03-30T10:11:02Z D! [sarama] consumer/broker/2 abandoned subscription to telegraf_metrics/25 because consuming was taking too long
2020-03-30T10:11:02Z D! [sarama] consumer/broker/2 abandoned subscription to telegraf_metrics/7 because consuming was taking too long

avneradania:
consumer: 
[[inputs.kafka_consumer]]
    brokers = ["kafka-21:9092", "kafka-22:9092", "kafka-23:9092", "kafka-24:9092", "kafka-25:9092", "kafka-26:9092"]
  topics = ["telegraf_metrics"]
  client_id = "telegraf-22"
  version = "2.1.1"

  ## Optional TLS Config
  # tls_ca = "/etc/telegraf/ca.pem"
  # tls_cert = "/etc/telegraf/cert.pem"
[...]

PR #7730, opened 2020-06-24
Add the ability to control MaxProcessingTime on Sarama lib

EladLeev:
Hi :smile:
This PR adds the ability to control MaxProcessingTime on the [Sarama](https://github.com/Shopify/sarama/blob/master/config.go#L325) lib.

We reached to this solution after this https://github.com/influxdata/telegraf/issues/7622 issue, where we wanted to slightly increased it, but anyway I think that it could be nice to allow the users to control it if he wants to.

I've updated the README files and wrote unit tests, but I didn't cover a scenario where a user set this parameter as a `string` (same goes to other `int` parameters that already exist before - let me know if you want me add it). Anyway, the error that a user gets while setting it as `string` is decent and understandable.

### Required for all PRs:

Thanks :pray:

### DIFF
--- a/plugins/inputs/kafka_consumer/README.md
+++ b/plugins/inputs/kafka_consumer/README.md
@@ -65,6 +65,15 @@ and use the old zookeeper connection method.
   ## waiting until the next flush_interval.
   # max_undelivered_messages = 1000
 
+  ## The maximum amount of time the consumer expects a message takes to
+  ## process for the user. If writing to the Messages channel takes longer
+  ## than this, that partition will stop fetching more messages until it
+  ## can proceed again.
+  ##
+  ## Note that, since the Messages channel is buffered, the actual grace time is
+  ## (MaxProcessingTime * ChannelBufferSize). Defaults - MaxProcessingTime: 100ms, ChannelBufferSize: 256.
+  # max_processing_time = 100
+
   ## Data format to consume.
   ## Each data format has its own unique set of configuration options, read
   ## more about them here:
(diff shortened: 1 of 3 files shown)


danielnelson:
I'm not sure if we should expose this setting, how would we use it along with the `max_undelivered_messages` setting?  Let's think through a use-case and how the behavior and log messages would be different.

EladLeev:
Thanks for your quick response, Daniel.

I'm not quite sure that I understand why this parameter needs to be directly affected by `max_undelivered_messages`.

Correct me if I'm wrong, but the `max_undelivered_messages`  parameter represents the "buffer" of this consumer - "Maximum messages to read from the broker that have not been written by any output."
The plugin will take up to N amount of messages and won't take new messages  until they been written to the configured outputs.

The `MaxProcessingTime` parameter is more focusing on the "other side" of the plugin - for how long should I wait before calling a TimeOut?

Having said that, I do think that we need to be able to control this setting from the plugin itself. I'm sure that it will be useful for more advanced users who want to control it and will take into consideration the side effects and downfalls of it.

What do you think? did I miss something? do you want to create more verbose documentation about it?

danielnelson:
Let's document what issue we run into using only `max_undelivered_messages`, is there an error or undesired behavior and how is it manifest.  Then let's compare that with the addition of `max_processing_time` and document how the issue is resolved.

I think I have a general idea how it works, but we should be able to concretely show the difference between the two cases before merging.

powersj:
Based on the last comment I am going to close the PR itself. If you disagree or want to further pursue this please feel free to resubmit the PR, rebased on master.

Thanks!

PR #7731, opened 2020-06-24
Fix SNMP trap test race

reimda:
Clean up race detected by 'go test -race' in snmp_trap input plugin


reimda:
The race was that the oid lookup mock was inserted after the plugin was started.  Inserting the mock to before starting the plugin prevents the race.

reimda:
related #7729

Issue #7726, opened 2020-06-24
Add FreeBSD laundry memory usage


zyun-i:
## Feature Request

After FreeBSD 11.1, mem number of cached replaced by laundry.

https://reviews.freebsd.org/D8302

gopsutil already supported laundry.

https://github.com/shirou/gopsutil/pull/566

### Proposal:

Add FreeBSD laundry memory usage to memory input plugin.

### Current behavior:

Not collect memory laundry.

### Use case: 

FreeBSD memory usage check.

PR #7736, opened 2020-06-24
Add laundry to mem input plugin on FreeBSD

danielnelson:
closes: #7726

### Required for all PRs:


plugins/inputs/mem/memory.go
```diff
@@ -61,6 +62,11 @@ func (s *MemStats) Gather(acc telegraf.Accumulator) error {
 		"write_back":        vm.Writeback,
 		"write_back_tmp":    vm.WritebackTmp,
 	}
+
+	if runtime.GOOS == "freebsd" {
+		fields["laundry"] = vm.Laundry
```
reimda, line 67: Need to add this to the docs, maybe with a note that it's specific to freebsd

reimda: We already support active, inactive, and wired which are only valid on OS X and BSD.  They're not handled conditionally at runtime.  Maybe this doesn't need to be either.

danielnelson: Active/inactive are set on Linux, despite the comment in gopsutil.  We really shouldn't be adding fields that aren't set, so I went through the stats again in more detail.  One situation I'm not handling yet is the non-cgo version of Darwin.

danielnelson:
Now that we are returning different fields depending on the platform, we are running into the issue where we need the test to be different depending on the host running the tests.  To get around this I added a field to the struct that can be overridden in the tests to change the platform.

PR #7752, opened 2020-06-26
Set user agent when requesting http config

danielnelson:
closes #7374

### Required for all PRs:

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -863,6 +863,7 @@ func fetchConfig(u *url.URL) ([]byte, error) {
 		req.Header.Add("Authorization", "Token "+v)
 	}
 	req.Header.Add("Accept", "application/toml")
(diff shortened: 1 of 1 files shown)


ronopitz:
Hi Daniel,

thank for your work, but your changes are not the functionality I need.

To send the correct config to telegraf, I need the operating system. This information is not included in the ProductToken.
The ProductToken is not enough to identify the OS and I cannot write the config dynamically.

The solution is as I wrote in the thread [7374](https://github.com/influxdata/telegraf/pull/7374).

    req.Header.Set("User-Agent", internal.ProductToken()+" ("+runtime.GOOS+" "+runtime.GOARCH+")")

Thanks Ron

danielnelson:
@ronopitz I'm not planning to add that information to the product token, maybe you could have each platform collect from a unique URLs instead.

PR #7762, opened 2020-06-29
Allow overriding the collection_jitter and precision per input

### LINKED ISSUE #1589: Allow service inputs to set precision
Have tried "ms" and "s".  telegraf receives and writes to influxDB, but always turns up in ns.  ...  The bare-bones, almost stock telegraf config : 

```
[global_tags]
[agent]
  interval = "60s"
  metric_batch_size = 1000
  metric_buffer_limit = 10000
  collection_jitter = "0s"
  flush_interval = "60s"
  flush_jitter = "0s"
  precision = "ms"
  debug = false
  quiet = false
  hostname = ""
  omit_hostname = true
[[outputs.influxdb]]
  precision = "s"
  urls = ["http://localhost:8086"] # required
  database = "server_stats" # required
  retention_policy = ""
  write_consistency = "any"
  timeout = "20s"
[[inputs.udp_listener]]
  service_address = ":8092"
  data_format = "influx"
```

### LINKED ISSUE #3634: Feature: Support "collection_jitter" on a per-input basis
### Proposal:

Each input should be allowed to specify its own ``collection_jitter``. 

Some input collections happen less frequently and with greater expense than others. Having a single, global ``collection_jitter`` makes it hard for expensive, long-running collections to co-exist with cheap, short-running collections, and still allow appropriate randomization to avoid collection stampedes while maintaining collection SLAs.

### Current behavior:

The ``collection_jitter`` option is an agent configuration that applies equally to all inputs.

### Desired behavior:

The ``collection_jitter`` should additionally be configurable per input. If one input occurs once per hour, it should be allowed to jitter within a large period (e.g. 30min) that might not be appropriate for an input that occurs once per minute.

### Use case:

In my specific use-case, we have lots of agents distributed across many machines coordinating with a centralized service to create some statistics per machine. This collection happens once per hour. These agents need to be randomly distributed across the 1 hour period to lessen the load on that central service.

### LINKED ISSUE #5853: Precision timestamp for jti_openconfig_telemetry input plugin
## Feature Request

Allow to set the precision of timestamp set by the input jti_openconfig_telemetry. 
Currently timestamp precision is in ns. It could be interesting to have precision in second to avoid ms jitter between sensor paths. 

### Proposal:

Allow to set :

precision: "s | ms | us | ns" 

### Current behavior:

Send ns timestamp

### Desired behavior:

Still sending ns timestamp (epoch) but with less precision that a ns. Ex. 

Before you have timestamp like that : 

1557755590589715534

With precision="s" you should have something like :

1557755590000000000

### Use case: 

If I request to Junos from telegraf the sensor : /interfaces/interface 

I could receive from the network equipment : 

/interfaces/interface/state/counters/xxxx with timestamp 1557755590589715534
/interfaces/interface/state/cdescription with timestamp 1557755590685780881

To aggregate (merge) those 2 values in a third one on influxdb based on time it's more difficult. 
It could help to have at the output of telegraf : 

/interfaces/interface/state/counters/xxxx with timestamp 1557755590000000000
/interfaces/interface/state/cdescription with timestamp 1557755590000000000

danielnelson:
Allow overriding the agent precision and collection jitter on a per-plugin basis.  Unlike the agent precision, this setting is applied to service inputs as well.

It is possible to shoot yourself in the foot with this, care will need to be taken to avoid have event-style metrics merged in the database.  See linked issues for use-cases.

closes #3634
closes #1589
closes #5853

### Required for all PRs:

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -1373,6 +1373,30 @@ func buildInput(name string, tbl *ast.Table) (*models.InputConfig, error) {
 			}
 		}
 	}
+	if node, ok := tbl.Fields["precision"]; ok {
+		if kv, ok := node.(*ast.KeyValue); ok {
+			if str, ok := kv.Value.(*ast.String); ok {
+				dur, err := time.ParseDuration(str.Value)
+				if err != nil {
+					return nil, err
+				}
+
+				cp.Precision = dur
+			}
+		}
+	}
+	if node, ok := tbl.Fields["collection_jitter"]; ok {
+		if kv, ok := node.(*ast.KeyValue); ok {
+			if str, ok := kv.Value.(*ast.String); ok {
+				dur, err := time.ParseDuration(str.Value)
+				if err != nil {
+					return nil, err
+				}
+
+				cp.CollectionJitter = dur
+			}
+		}
+	}
 
 	if node, ok := tbl.Fields["name_prefix"]; ok {
 		if kv, ok := node.(*ast.KeyValue); ok {
@@ -1420,6 +1444,8 @@ func buildInput(name string, tbl *ast.Table) (*models.InputConfig, error) {
 	delete(tbl.Fields, "name_override")
 	delete(tbl.Fields, "alias")
 	delete(tbl.Fields, "interval")
+	delete(tbl.Fields, "collection_jitter")
+	delete(tbl.Fields, "precision")
 	delete(tbl.Fields, "tags")
 	var err error
 	cp.Filter, err = buildFilter(tbl)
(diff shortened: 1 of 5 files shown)


config/config.go
```diff
@@ -1373,6 +1373,30 @@ func buildInput(name string, tbl *ast.Table) (*models.InputConfig, error) {
 			}
 		}
 	}
+	if node, ok := tbl.Fields["precision"]; ok {
```
reimda, line 1376: Lets get #7583 in first.  It shouldn't take 12 lines to grab a single value.

danielnelson: I feel that #7583 has too many additional error handling changes and has a high chance of introducing new errors.  Since I'd like to replace our TOML loading with the burntsushi library I'd rather not go through two rounds bug squashing.

danielnelson: I've brought over a function from this PR to handle parsing time.Duration from the config with less duplication.

reimda: That works

Issue #7300, opened 2020-04-09
SNMP input slow warning with just 18 snmp clients


leonv024:
### Relevant telegraf.conf:

```toml
[[inputs.snmp]]
  agents = [ "8.8.8.8" ]
  version = 2
  community = "public"
  interval = "60s"
  timeout = "10s"
  retries = 3
  [inputs.snmp.tags]
    name = "Home Router"

  [[inputs.snmp.field]]
    name = "hostname"
    oid = "RFC1213-MIB::sysName.0"
    is_tag = true

  [[inputs.snmp.field]]
    name = "uptime"
    oid = "DISMAN-EXPRESSION-MIB::sysUpTimeInstance"

  # IF-MIB::ifTable contains counters on input and output traffic as well as errors and discards.
  [[inputs.snmp.table]]
    name = "interface"
    inherit_tags = [ "hostname" ]
    oid = "IF-MIB::ifTable"

    # Interface tag - used to identify interface in metrics database
    [[inputs.snmp.table.field]]
      name = "ifDescr"
      oid = "IF-MIB::ifDescr"
      is_tag = true

  # IF-MIB::ifXTable contains newer High Capacity (HC) counters that do not overflow as fast for a few of the ifTable counters
  [[inputs.snmp.table]]
    name = "interface"
    inherit_tags = [ "hostname" ]
    oid = "IF-MIB::ifXTable"

    # Interface tag - used to identify interface in metrics database
    [[inputs.snmp.table.field]]
      name = "ifDescr"
      oid = "IF-MIB::ifDescr"
      is_tag = true

  # EtherLike-MIB::dot3StatsTable contains detailed ethernet-level information about what kind of errors have been logged on an interface (such as FCS error, frame too long, etc)
  [[inputs.snmp.table]]
    name = "interface"
    inherit_tags = [ "hostname" ]
    oid = "EtherLike-MIB::dot3StatsTable"

    # Interface tag - used to identify interface in metrics database
    [[inputs.snmp.table.field]]
      name = "ifDescr"
      oid = "IF-MIB::ifDescr"
```

### System info:

Ubuntu Server 18.04
Fiber connection
Telegraf 1.14.0
### Steps to reproduce:

1. add multiple configs with snmp clients, remote routers in my case. With a interval of 60 seconds.
2. Eventually, in my case around 18 clients, it can not longer retrieve the data
3. Telegraf stops working entirely. 

### Expected behavior:
Telegraf collects snmp data.

[...]

PR #7763, opened 2020-06-29
Add ifname processor plugin

### LINKED ISSUE #7246: Add processor to lookup SNMP ifName from ifIndex
### Proposal:

Add a new processor that can convert a tag containing an SNMP ifIndex into an ifName, using SNMP.

### Current behavior:

N/A

### Desired behavior:

```toml
[[processors.lookup_interface]]
  tag = "ifIndex"

  ## If set, write port to new tag key
  # dest = "ifName"
```

```diff
- foo,ifIndex=1 bar=42
+ foo,ifIndex=1,ifName=eth0 bar=42
```

### Use case: 

Enrich the output of the sflow input.

reimda:
closes #7246

### DIFF
--- a/internal/snmp/config.go
+++ b/internal/snmp/config.go
@@ -1,0 +1,34 @@
+package snmp
+
+import (
+	"github.com/influxdata/telegraf/internal"
+)
+
+type ClientConfig struct {
+	// Timeout to wait for a response.
+	Timeout internal.Duration `toml:"timeout"`
+	Retries int               `toml:"retries"`
+	// Values: 1, 2, 3
+	Version uint8 `toml:"version"`
+
+	// Parameters for Version 1 & 2
+	Community string `toml:"community"`
+
+	// Parameters for Version 2 & 3
+	MaxRepetitions uint8 `toml:"max_repetitions"`
+
+	// Parameters for Version 3
+	ContextName string `toml:"context_name"`
+	// Values: "noAuthNoPriv", "authNoPriv", "authPriv"
+	SecLevel string `toml:"sec_level"`
+	SecName  string `toml:"sec_name"`
+	// Values: "MD5", "SHA", "". Default: ""
+	AuthProtocol string `toml:"auth_protocol"`
+	AuthPassword string `toml:"auth_password"`
+	// Values: "DES", "AES", "". Default: ""
+	PrivProtocol string `toml:"priv_protocol"`
+	PrivPassword string `toml:"priv_password"`
+	EngineID     string `toml:"-"`
+	EngineBoots  uint32 `toml:"-"`
+	EngineTime   uint32 `toml:"-"`
+}
--- a/internal/snmp/wrapper.go
+++ b/internal/snmp/wrapper.go
@@ -59,38 +59,8 @@ func (gsw GosnmpWrapper) Get(oids []string) (*gosnmp.SnmpPacket, error) {
 	return nil, err
 }
 
-func NewWrapper(s ClientConfig, agent string) (GosnmpWrapper, error) {
+func NewWrapper(s ClientConfig) (GosnmpWrapper, error) {
 	gs := GosnmpWrapper{&gosnmp.GoSNMP{}}
-
-	if !strings.Contains(agent, "://") {
-		agent = "udp://" + agent
-	}
-
-	u, err := url.Parse(agent)
-	if err != nil {
-		return GosnmpWrapper{}, err
-	}
-
-	switch u.Scheme {
-	case "tcp":
-		gs.Transport = "tcp"
-	case "", "udp":
-		gs.Transport = "udp"
-	default:
-		return GosnmpWrapper{}, fmt.Errorf("unsupported scheme: %v", u.Scheme)
-	}
-
-	gs.Target = u.Hostname()
-
-	portStr := u.Port()
-	if portStr == "" {
-		portStr = "161"
-	}
-	port, err := strconv.ParseUint(portStr, 10, 16)
-	if err != nil {
-		return GosnmpWrapper{}, fmt.Errorf("parsing port: %w", err)
-	}
-	gs.Port = uint16(port)
 
 	gs.Timeout = s.Timeout.Duration
 
@@ -170,4 +140,41 @@ func NewWrapper(s ClientConfig, agent string) (GosnmpWrapper, error) {
 		sp.AuthoritativeEngineTime = s.EngineTime
 	}
 	return gs, nil
+}
+
+// SetAgent takes a url (scheme://host:port) and sets the wrapped
+// GoSNMP struct's corresponding fields.  This shouldn't be called
+// after using the wrapped GoSNMP struct, for example after
+// connecting.
+func (gs *GosnmpWrapper) SetAgent(agent string) error {
+	if !strings.Contains(agent, "://") {
+		agent = "udp://" + agent
+	}
+
+	u, err := url.Parse(agent)
+	if err != nil {
+		return err
+	}
+
+	switch u.Scheme {
+	case "tcp":
+		gs.Transport = "tcp"
+	case "", "udp":
+		gs.Transport = "udp"
+	default:
+		return fmt.Errorf("unsupported scheme: %v", u.Scheme)
+	}
+
+	gs.Target = u.Hostname()
+
+	portStr := u.Port()
+	if portStr == "" {
+		portStr = "161"
+	}
+	port, err := strconv.ParseUint(portStr, 10, 16)
+	if err != nil {
+		return fmt.Errorf("parsing port: %w", err)
+	}
+	gs.Port = uint16(port)
+	return nil
 }
--- a/internal/snmp/wrapper.go
+++ b/internal/snmp/wrapper.go
@@ -1,0 +1,173 @@
+package snmp
+
+import (
+	"fmt"
+	"net/url"
+	"strconv"
+	"strings"
+
+	"github.com/soniah/gosnmp"
+)
+
+// GosnmpWrapper wraps a *gosnmp.GoSNMP object so we can use it as a snmpConnection.
+type GosnmpWrapper struct {
+	*gosnmp.GoSNMP
+}
+
+// Host returns the value of GoSNMP.Target.
+func (gsw GosnmpWrapper) Host() string {
+	return gsw.Target
+}
+
+// Walk wraps GoSNMP.Walk() or GoSNMP.BulkWalk(), depending on whether the
+// connection is using SNMPv1 or newer.
+// Also, if any error is encountered, it will just once reconnect and try again.
+func (gsw GosnmpWrapper) Walk(oid string, fn gosnmp.WalkFunc) error {
+	var err error
+	// On error, retry once.
+	// Unfortunately we can't distinguish between an error returned by gosnmp, and one returned by the walk function.
+	for i := 0; i < 2; i++ {
+		if gsw.Version == gosnmp.Version1 {
+			err = gsw.GoSNMP.Walk(oid, fn)
+		} else {
+			err = gsw.GoSNMP.BulkWalk(oid, fn)
+		}
+		if err == nil {
+			return nil
+		}
+		if err := gsw.GoSNMP.Connect(); err != nil {
+			return fmt.Errorf("reconnecting: %w", err)
+		}
+	}
+	return err
+}
+
+// Get wraps GoSNMP.GET().
+// If any error is encountered, it will just once reconnect and try again.
+func (gsw GosnmpWrapper) Get(oids []string) (*gosnmp.SnmpPacket, error) {
+	var err error
+	var pkt *gosnmp.SnmpPacket
+	for i := 0; i < 2; i++ {
+		pkt, err = gsw.GoSNMP.Get(oids)
+		if err == nil {
+			return pkt, nil
+		}
+		if err := gsw.GoSNMP.Connect(); err != nil {
+			return nil, fmt.Errorf("reconnecting: %w", err)
+		}
+	}
+	return nil, err
+}
+
+func NewWrapper(s ClientConfig, agent string) (GosnmpWrapper, error) {
+	gs := GosnmpWrapper{&gosnmp.GoSNMP{}}
+
+	if !strings.Contains(agent, "://") {
+		agent = "udp://" + agent
+	}
+
+	u, err := url.Parse(agent)
+	if err != nil {
+		return GosnmpWrapper{}, err
+	}
+
+	switch u.Scheme {
+	case "tcp":
+		gs.Transport = "tcp"
+	case "", "udp":
+		gs.Transport = "udp"
+	default:
+		return GosnmpWrapper{}, fmt.Errorf("unsupported scheme: %v", u.Scheme)
+	}
+
+	gs.Target = u.Hostname()
+
+	portStr := u.Port()
+	if portStr == "" {
+		portStr = "161"
+	}
+	port, err := strconv.ParseUint(portStr, 10, 16)
+	if err != nil {
+		return GosnmpWrapper{}, fmt.Errorf("parsing port: %w", err)
+	}
+	gs.Port = uint16(port)
+
+	gs.Timeout = s.Timeout.Duration
+
+	gs.Retries = s.Retries
+
+	switch s.Version {
+	case 3:
+		gs.Version = gosnmp.Version3
+	case 2, 0:
+		gs.Version = gosnmp.Version2c
+	case 1:
+		gs.Version = gosnmp.Version1
+	default:
+		return GosnmpWrapper{}, fmt.Errorf("invalid version")
+	}
+
+	if s.Version < 3 {
+		if s.Community == "" {
+			gs.Community = "public"
+		} else {
+			gs.Community = s.Community
+		}
+	}
+
+	gs.MaxRepetitions = s.MaxRepetitions
+
+	if s.Version == 3 {
+		gs.ContextName = s.ContextName
+
+		sp := &gosnmp.UsmSecurityParameters{}
+		gs.SecurityParameters = sp
+		gs.SecurityModel = gosnmp.UserSecurityModel
+
+		switch strings.ToLower(s.SecLevel) {
+		case "noauthnopriv", "":
+			gs.MsgFlags = gosnmp.NoAuthNoPriv
+		case "authnopriv":
+			gs.MsgFlags = gosnmp.AuthNoPriv
+		case "authpriv":
+			gs.MsgFlags = gosnmp.AuthPriv
+		default:
+			return GosnmpWrapper{}, fmt.Errorf("invalid secLevel")
+		}
+
+		sp.UserName = s.SecName
+
+		switch strings.ToLower(s.AuthProtocol) {
+		case "md5":
+			sp.AuthenticationProtocol = gosnmp.MD5
+		case "sha":
+			sp.AuthenticationProtocol = gosnmp.SHA
+		case "":
+			sp.AuthenticationProtocol = gosnmp.NoAuth
+		default:
+			return GosnmpWrapper{}, fmt.Errorf("invalid authProtocol")
+		}
+
+		sp.AuthenticationPassphrase = s.AuthPassword
+
+		switch strings.ToLower(s.PrivProtocol) {
+		case "des":
+			sp.PrivacyProtocol = gosnmp.DES
+		case "aes":
+			sp.PrivacyProtocol = gosnmp.AES
+		case "":
+			sp.PrivacyProtocol = gosnmp.NoPriv
+		default:
+			return GosnmpWrapper{}, fmt.Errorf("invalid privProtocol")
+		}
+
+		sp.PrivacyPassphrase = s.PrivPassword
+
+		sp.AuthoritativeEngineID = s.EngineID
+
+		sp.AuthoritativeEngineBoots = s.EngineBoots
+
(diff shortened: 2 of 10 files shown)


plugins/processors/ifname/ifname.go
```diff
@@ -0,0 +1,328 @@
...
+	d.cache.Put(agent, m)
+	return m, nil
+}
+
+func (d *IfName) getMapRemote(agent string) (*nameMap, error) {
+	gs, err := snmp.NewWrapper(d.ClientConfig, agent)
+	if err != nil {
+		//can't translate toml to gosnmp.GoSNMP
+		return nil, err
+	}
```
danielnelson, line 232: Could we do this in Start so that we can report the configuration error?  Seems like not knowing the agent might be a problem.

plugins/processors/ifname/ifname.go
```diff
@@ -0,0 +1,328 @@
...
+	var m *nameMap
+	m, err = buildMap(&gs, d.ifXTable, "ifName")
+	if err != nil {
+		var err2 error
+		m, err2 = buildMap(&gs, d.ifTable, "ifDescr")
+		if err2 != nil {
+			return nil, err2
+		}
+	}
+	return m, nil
```
danielnelson, line 251: Suggestion: invert err check so you can return on first success and avoid the ugly unseemly err2 variable.

plugins/processors/ifname/ifname.go
```diff
@@ -0,0 +1,328 @@
...
+		return nil, err
+	}
+
+	if len(rtab.Rows) == 0 {
+		return nil, fmt.Errorf("empty table")
+	}
+
+	t := make(nameMap)
+	for _, v := range rtab.Rows {
+		//fmt.Printf("tags: %v, fields: %v\n", v.Tags, v.Fields)
```
danielnelson, line 302: remove

plugins/processors/ifname/ifname.go
```diff
@@ -0,0 +1,328 @@
...
+
+	err = gs.Connect()
+	if err != nil {
+		//can't connect
+		return nil, err
+	}
+
+	//try ifXtable and ifName first.  if that fails, fall back to
+	//ifTable and ifDescr
+	var m *nameMap
```
danielnelson, line 242: By convention maps aren't normally pointers, it isn't a problem but not needed since the actual type is small with the data stored on the heap.

plugins/processors/ifname/ifname_test.go
```diff
@@ -0,0 +1,106 @@
+//+build localsnmp
```
danielnelson, line 1: This is a switch to avoid running unless you have a local snmp server?  The normal way we would do this is to add to each test a few lines like:
```
	if testing.Short() {
		t.Skip("Skipping integration test in short mode")
	}
```

plugins/processors/ifname/ifname_test.go
```diff
@@ -0,0 +1,106 @@
...
+	"time"
+
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/internal"
+	"github.com/influxdata/telegraf/internal/snmp"
+	"github.com/influxdata/telegraf/testutil"
+	"github.com/stretchr/testify/require"
+)
+
+func _TestTable(t *testing.T) {
```
danielnelson, line 17: Why the leading underscore?

plugins/processors/ifname/ifname.go
```diff
@@ -0,0 +1,328 @@
...
+	if ok {
+		return m, nil
+	}
+
+	m, err = d.getMapRemote(agent)
+	if err != nil {
+		return nil, err
+	}
+
+	d.cache.Put(agent, m)
```
danielnelson, line 223: Can these maps be updated if the interfaces change?  Maybe something for later.

internal/snmp/wrapper.go
```diff
@@ -0,0 +1,173 @@
...
+		default:
+			return GosnmpWrapper{}, fmt.Errorf("invalid privProtocol")
+		}
+
+		sp.PrivacyPassphrase = s.PrivPassword
+
+		sp.AuthoritativeEngineID = s.EngineID
+
+		sp.AuthoritativeEngineBoots = s.EngineBoots
+
```
ssoroka, line 169: probably don't need the spacing here.

internal/snmp/wrapper.go
```diff
@@ -0,0 +1,173 @@
...
+	"net/url"
+	"strconv"
+	"strings"
+
+	"github.com/soniah/gosnmp"
+)
+
+// GosnmpWrapper wraps a *gosnmp.GoSNMP object so we can use it as a snmpConnection.
+type GosnmpWrapper struct {
+	*gosnmp.GoSNMP
```
ssoroka, line 14: (optional) Not a huge issue, but if you make the wrapper wrap `gosnmp.GoSNMP` instead of `*gosnmp.GoSNMP` you won't have the struct creation for error returns. Pretty minor though.

plugins/processors/ifname/ifname.go
```diff
@@ -0,0 +1,328 @@
...
+  # timeout = "5s"
+
+  ## SNMP version; can be 1, 2, or 3.
+  # version = 2
+
+  ## SNMP community string.
+  # community = "public"
+
+  ## Number of retries to attempt.
+  # retries = 3
```
ssoroka, line 37: can we default to zero retries?

plugins/processors/ifname/ifname.go
```diff
@@ -0,0 +1,328 @@
...
+  ## max_parallel_lookups is the maximum number of SNMP requests to
+  ## make at the same time.
+  # max_parallel_lookups = 100
+
+  ## ordered controls whether or not the metrics need to stay in the
+  ## same order this plugin received them in. If false, this plugin
+  ## may change the order when data is cached.  If you need metrics to
+  ## stay in order set this to true.  keeping the metrics ordered may
+  ## be slightly slower
+  #ordered = false
```
ssoroka, line 68:
```suggestion
  # ordered = false
```

plugins/processors/ifname/README.md
```diff
@@ -0,0 +1,75 @@
...
+  ## max_parallel_lookups is the maximum number of SNMP requests to
+  ## make at the same time.
+  # max_parallel_lookups = 100
+
+  ## ordered controls whether or not the metrics need to stay in the
+  ## same order this plugin received them in. If false, this plugin
+  ## may change the order when data is cached.  If you need metrics to
+  ## stay in order set this to true.  keeping the metrics ordered may
+  ## be slightly slower
+  #ordered = false
```
ssoroka, line 59:
```suggestion
  # ordered = false
```

plugins/processors/ifname/ifname.go
```diff
@@ -0,0 +1,328 @@
...
+	m, ok := d.cache.Get(agent)
+	d.rwLock.RUnlock()
+	if ok {
+		return m, nil
+	}
+
+	// The cache missed so make an SNMP request.  Write the response
+	// to cache and return it
+	d.rwLock.Lock()
+	defer d.rwLock.Unlock()
```
ssoroka, line 208: just lock and defer unlock once for the whole function, you're not going to save anything here by splitting it up (depending on how you solve the getMapRemote comment)

plugins/processors/ifname/ifname.go
```diff
@@ -0,0 +1,328 @@
...
+
+	// Check cache again while holding write lock to prevent duplicate
+	// SNMP requests.
+	var err error
+	m, ok = d.cache.Get(agent)
+	if ok {
+		return m, nil
+	}
+
+	m, err = d.getMapRemote(agent)
```
ssoroka, line 218: are we okay with the lock staying in place over a network call? seems highly unusual and problematic.

plugins/processors/ifname/ifname.go
```diff
@@ -0,0 +1,328 @@
...
+func (d *IfName) getMapRemote(agent string) (*nameMap, error) {
+	gs, err := snmp.NewWrapper(d.ClientConfig, agent)
+	if err != nil {
+		//can't translate toml to gosnmp.GoSNMP
+		return nil, err
+	}
+
+	err = gs.Connect()
+	if err != nil {
+		//can't connect
```
ssoroka, line 236: can wrap the error here for context

plugins/processors/ifname/ifname.go
ssoroka: We never do see what `err` is, if that matters.

plugins/processors/ifname/ifname.go
```diff
@@ -0,0 +1,328 @@
...
+		//fmt.Printf("tags: %v, fields: %v\n", v.Tags, v.Fields)
+		i_str, ok := v.Tags["index"]
+		if !ok {
+			//should always have an index tag because the table should
+			//always have IndexAsTag true
+			continue
+		}
+		i, err := strconv.ParseUint(i_str, 10, 64)
+		if err != nil {
+			//index value isn't a uint?
```
ssoroka, line 311: these should probably be errors.

plugins/processors/ifname/ifname.go
```diff
@@ -0,0 +1,328 @@
...
+
+var sampleConfig = `
+  ## Name of tag holding the interface number
+  # tag = "ifIndex"
+
+  ## Name of output tag where service name will be added
+  # dest = "ifName"
+
+  ## Name of tag of the SNMP agent to request the interface name from
+  # agent = "agent"
```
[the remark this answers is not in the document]
reimda: I considered having an agent setting but didn't think it was worth the potential trouble.  The interface index is a required tag and it's only valid with the agent it came from.  If we add in a way for the index to potentially refer to some other agent, people are going to do that and silently get an interface name from the wrong agent.  I thought requiring the agent to come in as a tag would decrease the possibility that an index is ever used on the wrong agent.

plugins/processors/ifname/ifname.go
reimda: Interfaces change so rarely I didn't think it was important to add a way to invalidate and reload the maps in the first version.  I agree it would be a good future change.

plugins/processors/ifname/ifname_test.go
reimda: The underscore is there to disable the test temporarily.  I didn't mean to leave it in.

plugins/processors/ifname/ifname.go
```diff
@@ -0,0 +1,328 @@
...
+	if !ok {
+		//source tag missing
+		return
+	}
+
+	num, err := strconv.ParseUint(num_s, 10, 64)
+	if err != nil {
+		//source tag not a uint
+		d.Log.Infof("couldn't parse source tag as uint")
+		return
```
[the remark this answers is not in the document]
reimda: I don't know if there's currently a good way for streaming processors to report errors.  The add func in StreamingProcessor lets you return an error, but that's too early if you're processing in a different thread.  If processing is successful you call Accumulator.AddMetric, so it feels like you should also give errors to the accumulator.

I changed IfName.addTag to return an error and changed the caller to log the error in one place with a capitalized message.

plugins/processors/ifname/ifname.go
```diff
@@ -0,0 +1,328 @@
...
+	}
+	metric.AddTag(d.DestTag, name)
+}
+
+func (d *IfName) Apply(metrics ...telegraf.Metric) []telegraf.Metric {
+	for _, metric := range metrics {
+		d.addTag(metric)
+	}
+	return metrics
+}
```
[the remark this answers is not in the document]
reimda: I left it in because one of the tests was calling Apply.  I changed the test to call addTag directly and removed Apply.

plugins/processors/ifname/ifname.go
reimda: done

plugins/processors/ifname/ifname.go
reimda: I tried to have the LRU cache be reusable so I chose to use a pointer as the cache value type.  I didn't want to assume the value was small enough to copy.  Since nameMap is the cache value type it made sense to hold pointers elsewhere.

Since it looked strange to both of you, I unpointerized it everywhere and I won't worry as much about reusing the cache.

plugins/processors/ifname/ifname_test.go
reimda: It's a switch to turn off these tests in CI because they don't have a local snmp agent.  I've been running 'go test -tags localsnmp'.

I've changed the tests to avoid doing this.  TestIfName now mocks the snmp response.  TestTable and TestXTable are now skipped for short tests.

reimda: Automated unit testing just reminded me that this needs snmptranslate and mibs to be installed locally so mocking the snmp response isn't enough to get TestIfName as a short test.  It's now skipped too.

plugins/processors/ifname/ifname.go
reimda: I split the initialization so it only parses the host tag here and it parses the config in Start.  Now it will tell the user if any toml values are not the expected type or format.  Any snmp errors will still happen here.

internal/snmp/config.go
```diff
@@ -0,0 +1,34 @@
...
+	SecName  string `toml:"sec_name"`
+	// Values: "MD5", "SHA", "". Default: ""
+	AuthProtocol string `toml:"auth_protocol"`
+	AuthPassword string `toml:"auth_password"`
+	// Values: "DES", "AES", "". Default: ""
+	PrivProtocol string `toml:"priv_protocol"`
+	PrivPassword string `toml:"priv_password"`
+	EngineID     string `toml:"-"`
+	EngineBoots  uint32 `toml:"-"`
+	EngineTime   uint32 `toml:"-"`
```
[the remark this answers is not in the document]
reimda: This and everything in internal/snmp came from the snmp input plugin.  When I moved it into internal I noticed it has ugly spots like here and in your other review comments.  Since the integration tests are lacking, there is real potential to break stuff without knowing.  (The snmp input unit tests mock all network traffic.  It also mocks running the snmptranslate cli tool.)  In this PR I avoided modifying gosnmpwrapper because of the risk.

Cleaning up the snmp input is a good idea but it's not something we need to do for the ifname processor.

internal/snmp/wrapper.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+	return err
+}
+
+// Get wraps GoSNMP.GET().
+// If any error is encountered, it will just once reconnect and try again.
+func (gsw GosnmpWrapper) Get(oids []string) (*gosnmp.SnmpPacket, error) {
+	var err error
+	var pkt *gosnmp.SnmpPacket
+	for i := 0; i < 2; i++ {
```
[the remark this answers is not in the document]
reimda: I haven't seen it fail but this has been in the snmp input for a long time and I expect users are used to how it behaves so I don't want to change it with this PR.

internal/snmp/wrapper.go
reimda: It seems to be this way intentionally and isn't a trivial change.  The way snmp input's connection caching was written, the wrapper itself needs to be copyable but the GoSNMP library's struct itself isn't copyable.

plugins/processors/ifname/ifname.go
reimda: This is another case of copying the snmp input.  I thought it would be best to have the same behavior there and here but I'm open to changing it there's something wrong with it.

plugins/processors/ifname/ifname.go
reimda: done

plugins/processors/ifname/README.md
reimda: done

plugins/processors/ifname/ifname.go
reimda: I thought since the network call is so infrequent it wouldn't matter that it blocks reads.  Looking at it again I agree the network call should be outside.  I moved it and added some synchronization in case multiple requests for the same agent before the response comes back.

plugins/processors/ifname/ifname.go
reimda: I changed the rest of this function so that it makes sense now to use the RW lock for adding to the cache

plugins/processors/ifname/ifname.go
reimda: Agreed, I changed them to be errors

internal/snmp/wrapper.go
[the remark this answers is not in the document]
ssoroka: Sounds good. There's an open issue for one of the other plugins that retries like this, so I'd like to avoid creating more issues if we can help it, but I'm okay with solving this in a later PR.

plugins/processors/ifname/ifname.go
ssoroka: See the last comment on https://github.com/influxdata/telegraf/issues/7300

plugins/processors/ifname/ifname.go
Hipska: Virtual interfaces get added and removed all the time. Even frequently for physical ones…

reimda: Hi @Hipska, do you have a suggestion on how to handle caching of interface names? We can't get the names from the agent every time they're needed because it's a relatively slow network transaction. We also don't want to cache the names forever because they change. Would it be sufficient to expire cached names after a set amount of time, configurable in the plugin settings?  Maybe we should also invalidate the cache entry when a request comes in for an interface number that isn't in the cached list?

Hipska: Yeah, a timeout based on the polling interval is what I would expect, that also handles the invalidate problem when a new interface is found. Otherwise indeed invalidate the cache if number of interfaces change and also after some timeout.

reimda: Since ifname is a processor it doesn't have a polling interval.  Metrics will come into ifname from various inputs (snmp, sflow, possibly others) each of which could have different intervals.  The processor api also doesn't have a way to determine which input a given metric came from.  These limitations mean it's more practical to add a new setting for cache time to live.

I'm working on changes to do this.  I'll also have it invalidate the cache entry if the interface number isn't in the cached list.

plugins/processors/ifname/ifname.go
```diff
@@ -0,0 +1,370 @@
...
+
+var sampleConfig = `
+  ## Name of tag holding the interface number
+  # tag = "ifIndex"
+
+  ## Name of output tag where service name will be added
+  # dest = "ifName"
+
+  ## Name of tag of the SNMP agent to request the interface name from
+  # agent = "agent"
```
Hipska, line 25: Can't this be default set to `agent_host` as this is the tag that is being returned from `snmp` plugin? It would reduce the need to set this value every time.

plugins/processors/ifname/ifname.go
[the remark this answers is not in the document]
Hipska: Can't this be set to `agent_host` as default? Since this is the tag that is being returned from snmp plugin, It would reduce the need to set this value every time.

Hipska:
While trying out this plugin, I came to the conclusion of maybe allow the community string also coming from a tag, or some other way to retrieve all the snmp settings?

ssoroka:
cc @reimda

plugins/processors/ifname/ifname.go
reimda: I don't think we should change the default now that this has been released.  Changing the default would silently break anyone who is using the default.

reimda:
> While trying out this plugin, I came to the conclusion of maybe allow the community string also coming from a tag, or some other way to retrieve all the snmp settings?

I can see this being useful but I'm not sure how it would work.  Providing the settings in tags could work for v1 and v2c because they only need the snmp version and community string, so two tags.  Snmp v3 would need version and up to seven other settings.  Providing 8 tags just for snmp connection info doesn't sound pleasant.  Combining multiple settings into one tag also doesn't sound good.

We could have multiple sets of connection settings in the config file and then add a tag to select which settings to use, but that isn't as flexible as passing the settings in tags.

If you can think of a nice way to do this, I'd be happy to consider a PR.

Hipska:
I don't know of any way how to do this. Should I request it via an issue?

I also noticed that when using this plugin, after a while all inputs stop working (even like cpu and disk), will create an issue for that or should we look into it on Slack first?

plugins/processors/ifname/ifname.go
Hipska: I can't figure out why anyone would have "agent" as default? There is no plugin that provides this tag by default.

ssoroka:
a new issue (or issues) would probably be best.

plugins/processors/ifname/ifname.go
reimda: You have a valid point that it's unlikely anyone is using the default.  They would have to rename the tag from the snmp input or the sflow input (which uses "agent_address").

reimda:
I think we've identified three issues: 1) a feature request for providing snmp connection info outside of the conf file; 2) a bug that the default tag doesn't match the snmp or sflow inputs; 3) a bug that all inputs stop working after a while.

@Hipska would you open issues for these?  For the third one, let's start with opening an issue and then use slack if needed.  Please include your telegraf.conf, the telegraf output, and the conditions that cause it, like how long it takes to happen.

Hipska:
I'm still figuring out on the last one what and how. As I can't get this plugin run stable, I'm still figuring out how. Sometimes it doesn't do anything and without any error.

PR #7773, opened 2020-07-01
cloudfoundry input plugin

chrisfarms:
### What

Adds a cloudfoundry service input plugin for receiving logs and metrics from the cloudfoundry platform's loggregator reverse log proxy event streaming systems.

If the given credentials have the `logs.admin` scope, then It is capable of functioning like a "nozzle" (`firehose=true`) streaming all possible platform events.

If the given credentials are from a regular cloudfoundry SpaceDeveloper/SpaceAuditor user then event streams will be discovered for applications visible to the given user/client.

### Required for all PRs:


chrisfarms:
Note: dropped event stream "discovery" .... 

an [earlier experimental branch](https://github.com/chrisfarms/telegraf/tree/cloudfoundry-with-discovery/plugins/inputs/cloudfoundry) attempted to simulate the "firehose" (streaming of all metric envelopes from all authorised sources) for cloudfoundry users without the admin-level creds required to perform such action. That path has been abandoned as I don't think the additional "discovery" logic is inline with telegraf's model.

This change now focuses only on connecting to a single cloudfoundry RLP event stream for a single source ID (application ID etc) to reduce complexity.

trevorwhitney:
FWIW, there is unlikely to be any future development on #5471 and #5433, so +1 to picking them here and closing the originals.

sjwang90:
I'm worried this adds a lot of dependencies to Telegraf. A method we can now re-route plugins with a lot of dependencies is to have it as an [external plugin](https://github.com/influxdata/telegraf/blob/master/docs/EXTERNAL_PLUGINS.md) that can be used with `execd` to run seamlessly with Telegraf.

srebhan:
@chrisfarms are you still interested in getting this merged?

PR #7792, opened 2020-07-07
Support utf-16 in file and tail inputs

### LINKED ISSUE #6876: Support non utf-8 encodings in the file input
## Feature Request

### Proposal:

The file input should allow processing text files that are not utf-8 encoded.

### Current behavior:

Files with alternative encodings either cannot be parsed or is not encoded properly.

### Desired behavior:

Be able to select a text encoding in the input file, if selected the data will be converted internally to utf-8.
```
[[inputs.file]]
  encoding = "ISO-8859"
```

### Use case: 

https://community.influxdata.com/t/unable-to-rename-fieldname-containing-unicode-character/12293/5

### LINKED ISSUE #7299: Support non utf-8 encodings in the tail input
## Feature Request

Proposal:
The tail input should allow processing text files that are not utf-8 encoded.

Current behavior:
Files with alternative encodings either cannot be parsed or are not encoded properly.

Desired behavior:
Be able to select a text encoding in the input file, if selected the data will be converted internally to utf-8.

[[inputs.tail]]
  encoding = "UTF-16"

Same to this feature request for file plugin, but new one is for tail input plugin

https://github.com/influxdata/telegraf/issues/6876

danielnelson:
closes #7299
closes #6876

Also see changes on influxdata/tail: https://github.com/influxdata/tail/compare/master...influxdata:open-reader-func

### Required for all PRs:


plugins/parsers/csv/parser.go
```diff
@@ -208,8 +251,10 @@ outer:
 
 	// will default to plugin name
 	measurementName := p.MetricName
-	if recordFields[p.MeasurementColumn] != nil && recordFields[p.MeasurementColumn] != "" {
-		measurementName = fmt.Sprintf("%v", recordFields[p.MeasurementColumn])
+	if p.MeasurementColumn != "" {
+		if recordFields[p.MeasurementColumn] != nil && recordFields[p.MeasurementColumn] != "" {
```
ssoroka, line 255: how can you compare this without a type assertion?

danielnelson: I wasn't sure about this existing code, so I added some tests, seems to be working.  The conditional I added fixes a bug where the measurement name could be set to the value of an empty string named column.

Found this in the spec:

> Interface values are comparable. Two interface values are equal if they have identical dynamic types and equal dynamic values or if both have value nil.

PR #7814, opened 2020-07-10
Send metrics in FIFO order

### LINKED ISSUE #5633: Send metrics from oldest to newest, always
### Proposal:

Telegraf should always send metrics point from the older to the newer. During "normal" behavior and while catch-up backlog if an output become slow.

### Current behavior:

Since #5287, when buffer is not empty and contains more than one gather cycle, metrics are sent from newest to older.

### Desired behavior:

Batch() should send oldest metrics and metrics in the batch should be in increasing order of age.

We should still drop the oldest metrics if the buffer become full, so #5194 is still fixed

### Use case:

Currently metric points are not in consistent order. When buffer is empty, the order is from older to newer, but if buffer start to fill, batch with be in the opposite order (newer to older).

In practice, the buffer while often contains few metrics from previous gather cycle, so it may happen even if output is not down.

At the end, that means that the output could not rely on having metrics ordered, which was (mosly) true before #5287.
Having metrics ordered allow to easily do some transformation on the fly (e.g. difference between two point to compute the rate) and generally make working with time series easier.

If this change seems good, I could come with a PR.

PierreF:
As requested in https://github.com/influxdata/telegraf/issues/5633#issuecomment-655687905, here the PR :)

closes #5633

### Required for all PRs:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -105,6 +105,7 @@
 - [#7558](https://github.com/influxdata/telegraf/issues/7558): Remove trailing backslash from tag keys/values in influx serializer.
 - [#7715](https://github.com/influxdata/telegraf/issues/7715): Fix incorrect Azure SQL DB server properties.
 - [#7431](https://github.com/influxdata/telegraf/issues/7431): Fix json unmarshal error in the kibana input.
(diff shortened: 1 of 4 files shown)


danielnelson:
Thanks @PierreF, that was fast!

@reimda @ssoroka We should consider including this in 1.15.0 along with rc3.

PR #7822, opened 2020-07-11
add nsd input plugin

gearnode:
### Required for all PRs:

### DIFF
--- a/plugins/inputs/nsd/README.md
+++ b/plugins/inputs/nsd/README.md
@@ -1,0 +1,176 @@
+# NSD Input Plugin
+
+This plugin gathers stats from
+[NSD](https://www.nlnetlabs.nl/projects/nsd/about) - an authoritative DNS name
+server.
+
+### Configuration:
+
+```toml
+# A plugin to collect stats from the NSD DNS resolver
+[[inputs.nsd]]
+  ## Address of server to connect to, read from nsd conf default, optionally ':port'
+  ## Will lookup IP if given a hostname
+  server = "127.0.0.1:8953"
+
+  ## If running as a restricted user you can prepend sudo for additional access:
+  # use_sudo = false
+
+  ## The default location of the nsd-control binary can be overridden with:
+  # binary = "/usr/sbin/nsd-control"
+
+  ## The default location of the nsd config file can be overridden with:
+  # config_file = "/etc/nsd/nsd.conf"
+
+  ## The default timeout of 1s can be overridden with:
+  # timeout = "1s"
+```
+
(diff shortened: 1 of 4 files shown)


gearnode:
@danielnelson Any ETA to merge this patch ?

plugins/inputs/nsd/README.md
```diff
@@ -0,0 +1,176 @@
...
+This plugin gathers stats from
+[NSD](https://www.nlnetlabs.nl/projects/nsd/about) - an authoritative DNS name
+server.
+
+### Configuration:
+
+```toml
+# A plugin to collect stats from the NSD DNS resolver
+[[inputs.nsd]]
+  ## Address of server to connect to, read from nsd conf default, optionally ':port'
```
reimda, line 12:
```suggestion
  ## Address of server to connect to, optionally with ':port'.  Defaults to the address in the nsd config file.
```

plugins/inputs/nsd/nsd.go
```diff
@@ -0,0 +1,184 @@
...
+		ctx, lookUpCancel := context.WithTimeout(context.Background(), Timeout.Duration)
+		defer lookUpCancel()
+		serverIps, err := resolver.LookupIPAddr(ctx, host)
+		if err != nil {
+			return nil, fmt.Errorf("error looking up ip for server: %s: %s", Server, err)
+		}
+		if len(serverIps) == 0 {
+			return nil, fmt.Errorf("error no ip for server: %s: %s", Server, err)
+		}
+		server := serverIps[0].IP.String()
```
reimda, line 86: I understand that nsd-control's -s option requires an IP address, not a hostname, so we need to resolve the name here.  I'm a little concerned that this ignores secondary IPs.  Ordinary, connecting through Dial() will try each IP in order until it succeeds (https://golang.org/pkg/net/#Dial).  It may be unlikely that we can't connect on the first IP and we can on a secondary, but if this is the case and we never try the secondary IPs, telegraf will fail silently and it will be difficult for a user to debug.

I wonder if it would be worth it to remove the ability to resolve the server hostname and require the user to provide an IP, since this is what nsd-control's -s option requires.

reimda (changes requested):
Thanks for the PR! Daniel has moved on and isn't directly involved with telegraf now.

plugins/inputs/nsd/nsd.go
gearnode: Agree. In fact I add this because the `unbound` input plugin have this behavior.

reimda (approved):

Issue #6626, opened 2019-11-06
InfluxDB v2.0 support for inputs.influxdb_listener plugin


sjwang90:
### Proposal:
Requesting v2 support for `influxdb_listener` input plugin

### Current behavior:
`influxdb_listener` plugin currently only supports InfluxDB v1

### Desired behavior:
`influxdb_listener` capable of listening to both InfluxDB v1 AND v2. Ideally one plugin can support both versions.  

### Use case: 
Help with growth and utilization of Influx2.0

danielnelson:
> influxdb_listener capable of listening to both InfluxDB v1 AND v2. Ideally one plugin can support both versions.

I think we may want to do this as a separate plugin, in order to support the query parameters without causing confusion in the configuration (database_tag vs bucket_tag/organization_tag).  There are few key implementation differences between the two write endpoints that likely mean we need separate code anyway.

gustavooferreira:
Hi, is anyone aware of any input plugin with this support, I can't seem to find anything that supports v2 input for telegraf.

danielnelson:
Depending on your requirements, you may be able to use `http_listener_v2` with `data_format = "influx"`.  The bucket, org, token will just be ignored.  You will want to set up the plugin to use POST and `path = "/api/v2/write"`

magichair:
I've opened #7828 which ports the `influxdb_listener` to support the InfluxDB 1.8 / 2.x `/api/v2/write` endpoint and authentication scheme.

Issue #4935, opened 2018-10-29
HTTP 204 returned from /ping endpoint is unsuitable for Google Kubernetes Cloud healthcheck


noodle-rubrik:
## Feature Request

The "/ping" endpoint of influxdb_listener should have an option to return HTTP status code 200

### Proposal:

Google Cloud does not allow you to configure the behavior of its health checks, other than the endpoint, it requires a 200 response. This means that the "/ping" endpoint, which seems intended for this sort of use case, can't be used on this cloud provider. The return code should be changed to 200, either by default, or by adding a param to the endpoint, as suggested in a similar issue for influxdb itself.
[1] https://github.com/influxdata/influxdb/issues/9772

### Current behavior:

Calling the "/ping" endpoint returns 204. Using "/ping" as your Google HTTP(s) Load Balancer backend's health check fails in a (to me) surprising way. As a workaround, "/query" can be used to return the correct status code, but this is more obtuse, and the behavior of an empty query might change in the future, or consume more resources than a simple ping.

### Desired behavior:

If "/ping" can't be configured to return 200 by default, then configuring "/ping?statusCode=200" as the health check should return 200, as proposed in the linked influxdb issue. If that proposal is not accepted, then both implementations of "/ping" should have the same API to request a different status code, so that influxdb_listener is as similar to influxdb as possible.

### Use case: [Why is this important (helps with prioritizing requests)]

We are deploying telegraf to GCP as a proxy to forward traffic from a second telegraf instance to influxdb. The influxdb_listener plugin is definitely the right tool for this use case, and we can put this telegraf instance behind an endpoint of our load balancer, as long as we have a health check that returns 200 OK.

noodle-rubrik:
Maybe this should be a separate issue, but I also am running into problems when trying to use the basic authentication options, since there's no option to send those with Google's healthchecks. I don't think that /ping needs to be protected by this authentication, does it? Removing the call to AuthenticateIfSet for /ping would also help when running telegraf influxdb_listener on GKE.

glinton:
It looks like `/ping` isn't behind authentication in `influxdb v1.6.4`, but I agree that `influxdb_listener` should mirror influx's api. if it changes to allow a different response code, let us know so we can make that change here too

danielnelson:
Here is what will be, I assume, in 1.7:
https://github.com/influxdata/influxdb/blob/master/services/httpd/handler.go#L846

[...]

PR #7828, opened 2020-07-13
Influxdb v2 listener

magichair:
closes https://github.com/influxdata/telegraf/issues/6626

Most code is ported directly from the `influxdb_listener` input plugin. Naming convention for `influxdb_v2*` matches the paradigm with the existing output plugin.

I've manually tested functionality of this plugin with two of the official InfluxDB 2.x client libraries:
* https://github.com/influxdata/influxdb-client-python
* https://github.com/influxdata/influxdb-client-js

~Only open question, I ported the `handlePing` method from `influxdb_listener` to instead support the `/api/v2/ready` endpoint. But it could easily be removed if there is no requirement from the client libraries to test readiness before writing.~
Removed `handleReady`

Also to note, the `handleQuery` method was removed, the comments indicated that was to support some client libraries that would attempt a basic query to test connectivity. I didn't see any proof of that in the above two client libraries, so I removed the stubbed query response.

Oh, another thing updated to support this change, a new Auth handler is added to manage InfluxDB's idiosyncratic `Authorization: Token <token>` authentication strategy.

### Required for all PRs:

- [x] Awaiting Legal review from my company regarding the applicability of the [CCLA](https://www.influxdata.com/legal/ccla/) Update: Do not need any additional approvals


magichair:
I'll have a fix for the time related test failure.

plugins/inputs/influxdb_v2_listener/influxdb_v2_listener.go
```diff
@@ -0,0 +1,396 @@
...
+  ## You probably want to make sure you have TLS configured above for this.
+  # token = "foobar"
+`
+
+func (h *InfluxDBV2Listener) SampleConfig() string {
+	return sampleConfig
+}
+
+func (h *InfluxDBV2Listener) Description() string {
+	return "Accept metrics over InfluxDB 1.x HTTP API"
```
ssoroka, line 99: update description.

plugins/inputs/influxdb_v2_listener/influxdb_v2_listener.go
```diff
@@ -0,0 +1,396 @@
...
+
+type InfluxDBV2Listener struct {
+	ServiceAddress string `toml:"service_address"`
+	port           int
+	tlsint.ServerConfig
+
+	ReadTimeout  internal.Duration `toml:"read_timeout"`
+	WriteTimeout internal.Duration `toml:"write_timeout"`
+	MaxBodySize  internal.Size     `toml:"max_body_size"`
+	MaxLineSize  internal.Size     `toml:"max_line_size"` // deprecated in 1.14; ignored
```
ssoroka, line 35: I'd remove this. if it's already deprecated, there's no reason to add it to a new plugin.

plugins/inputs/influxdb_v2_listener/influxdb_v2_listener.go
```diff
@@ -0,0 +1,396 @@
...
+)
+
+type InfluxDBV2Listener struct {
+	ServiceAddress string `toml:"service_address"`
+	port           int
+	tlsint.ServerConfig
+
+	ReadTimeout  internal.Duration `toml:"read_timeout"`
+	WriteTimeout internal.Duration `toml:"write_timeout"`
+	MaxBodySize  internal.Size     `toml:"max_body_size"`
```
ssoroka, line 34: Not sure we really need a max body size. If we do, 32mb is possibly too small. We should be streaming the request through, so this won't affect memory used.

magichair: ah, yeah - good catch, InfluxDB 2.x now no longer has a `max-body-size` config parameter - https://v2.docs.influxdata.com/v2.0/reference/config-options/

plugins/inputs/influxdb_v2_listener/influxdb_v2_listener.go
magichair: Perfect, missed this, will remove.

magichair:
> ok so is it magi chair, magic chair, or magic hair? :D Either way, bravo, fellow mage. 🧙

magic hair, an ol' high school nickname that became my de facto online alias. Thanks for the review and approval!

internal/http.go
```diff
@@ -48,6 +49,50 @@ func (h *basicAuthHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request)
 	h.next.ServeHTTP(rw, req)
 }
 
+type TokenAuthErrorFunc func(rw http.ResponseWriter)
+
+// TokenAuthHandler returns a http handler that requires `Authorization: Token <token>`
+// Introduced to support InfluxDB 2.x style authentication
+// https://v2.docs.influxdata.com/v2.0/reference/api/#authentication
+func TokenAuthHandler(token string, onError TokenAuthErrorFunc) func(h http.Handler) http.Handler {
```
danielnelson, line 57: Set this up to handle a direct comparison with any Authorization header value, not treating the scheme and credentials separately, just as a single big string.  I think this will be more reusable and less tied to the InfluxDBv2 plugin.

internal/http.go
```diff
@@ -48,6 +49,50 @@ func (h *basicAuthHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request)
...
+func (h *schemeAuthHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
+	if h.credentials != "" {
+		// Scheme checking
+		authHeader := req.Header.Get("Authorization")
+		authParts := strings.SplitN(authHeader, " ", 2)
+		if len(authParts) != 2 ||
+			subtle.ConstantTimeCompare(
+				[]byte(strings.ToLower(strings.TrimSpace(authParts[0]))),
+				[]byte(strings.ToLower(h.scheme))) != 1 ||
+			subtle.ConstantTimeCompare([]byte(strings.TrimSpace(authParts[1])), []byte(h.credentials)) != 1 {
```
danielnelson, line 85: Let's remove the normalization steps here, just do a subtle.ConstantTimeCompare against the header value.

magichair:
Excellent points @danielnelson, I'll get right on these 👍 Thank you!

plugins/inputs/influxdb_v2_listener/influxdb_v2_listener.go
```diff
@@ -0,0 +1,359 @@
...
+}
+
+const sampleConfig = `
+  ## Address and port to host InfluxDB listener on
+  service_address = ":8186"
+
+  ## maximum duration before timing out read of the request
+  read_timeout = "10s"
+  ## maximum duration before timing out write of the response
+  write_timeout = "10s"
```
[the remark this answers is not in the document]
magichair: They do surface two different configurations on the `http.Server` and are carried over from the `influxdb_listener` implementation. I'm not sure how they might be used, but I'd like to keep them separate.

plugins/inputs/influxdb_v2_listener/influxdb_v2_listener.go
```diff
@@ -0,0 +1,359 @@
...
+
+	h.mux.Handle("/api/v2/write", authHandler(h.handleWrite()))
+	h.mux.Handle("/api/v2/ready", h.handleReady())
+	h.mux.Handle("/", authHandler(h.handleDefault()))
+}
+
+func (h *InfluxDBV2Listener) Init() error {
+	tags := map[string]string{
+		"address": h.ServiceAddress,
+	}
```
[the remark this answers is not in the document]
magichair: This `tags` is just used to add tagging to the internal `selfstat` below and does not munge the actual telegraf metric that is collected and passed on from the plugin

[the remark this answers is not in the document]
danielnelson: :+1:

plugins/inputs/influxdb_v2_listener/influxdb_v2_listener.go
```diff
@@ -0,0 +1,359 @@
...
+			body, err = gzip.NewReader(body)
+			if err != nil {
+				h.Log.Debugf("Error decompressing request body: %v", err.Error())
+				badRequest(res, err.Error())
+				return
+			}
+			defer body.Close()
+		}
+
+		parser := influx.NewStreamParser(body)
```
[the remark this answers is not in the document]
magichair: 🤔 so maybe I should still enforce some form of max_body_size if I have to read the entire http response body into memory and pass into `influx.Parse(bodyBytes)` https://github.com/influxdata/telegraf/pull/7828#discussion_r453742459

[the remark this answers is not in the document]
magichair: I definitely agree we want to match, not pushing back on that, just trying to brainstorm on the best way to approach this since I presume the accumulator isn't transactional, and I'd have to scrap the streaming code (or again store the telegraf metrics in memory and throw them away if we encounter an error):
https://github.com/influxdata/telegraf/blob/af42bc96bb18bc17360fcf262eb94cebf99cda40/plugins/inputs/influxdb_v2_listener/influxdb_v2_listener.go#L277

magichair:
Comments from Daniel still remaining:
* Handle all or nothing writes
* Normalized error handling to match API spec

plugins/inputs/influxdb_v2_listener/influxdb_v2_listener.go
danielnelson: I would keep this option, 32mb seems like plenty. The typical batch size is 1000-5000 metrics and is gzip compressed.

plugins/inputs/influxdb_v2_listener/influxdb_v2_listener.go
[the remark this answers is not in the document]
danielnelson: My thought was that the response is typically very small, so there isn't a lot of value in having separate options vs a simpler combined timeout.

plugins/inputs/influxdb_v2_listener/influxdb_v2_listener.go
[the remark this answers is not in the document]
danielnelson: I do think we should keep `max_body_size`, commented over there.  You are right that the accumulator isn't transactional, so we aren't safe in the case of SIGKILL, power loss or the likes.  Anyway, the output plugin is unsyncronized with the inputs, so the data could go out in separate writes.

Still we can do a close approximation that will work under normal operation (outputs up/clean shutdown): after switching to the Parse() function just check and return the error or add all the metrics if no error.

plugins/inputs/influxdb_v2_listener/influxdb_v2_listener.go
ssoroka: The limit seems pretty arbitrary if we're taking results of any size in a stream.

danielnelson: This API from InfluxDBv2 is built around accepting batches of data, it's not a streaming API.  As everything is loaded into memory, I think this is is a useful limit to avoid OOM.

plugins/inputs/influxdb_v2_listener/influxdb_v2_listener.go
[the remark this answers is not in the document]
magichair: Decided that given these should be small responses, to entirely remove the timeouts for a simpler config. Should someone need to add that functionality in the future, it should be straightforward to forward port from the `influxdb_listener` or other examples in the code base.

plugins/inputs/influxdb_v2_listener/influxdb_v2_listener.go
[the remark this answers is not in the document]
magichair: Added max_body_size back, implemented all or nothing writes.

magichair:
I think all of @danielnelson 's concerns have been addressed at this point. A fair bit of changes since last review - each split in their own commit (hopefully for easier review).

magichair:
Just poking in to see if there is a planned release timeline for 1.16.0. I've been running this as a custom built telegraf in production fairly successfully, not a lot of load on this for proving it out. However, it has kept up with what I'm using it for.

sjwang90:
@magichair We're looking to get a 1.16.0 release soon.  Ideally to get this plugin in to align with the InfluxDB OSS 2 GA.

ssoroka:
Merged. Thanks again for the great effort here!

Eraac:
Can this listener for v2 also support a `/ping` route, when you deploy this to kubernetes or/and behind a load balancer a health check route (without auth) is required to make this work

magichair:
> Can this listener for v2 also support a `/ping` route? When you deploy this to kubernetes or/and behind a load balancer a health check route (without auth) is required to make this work
> 
> **EDIT**: Which should return 200 instead of 204, because of this #4935

@Eraac I debating putting the /ping endpoint in, however it didn't seem necessary. The original Influx 1.x listener had it in place for compatibility with the existing InfluxDB client libraries where they tested /ping before allowing a write. In this case, the 2.x client libraries have no such check.

Is there a reason you can't add https://github.com/influxdata/telegraf/tree/master/plugins/outputs/health to setup a generic health endpoint for your telegraf node overall? This is what I use for my AWS ALB health checks.

Eraac:
> Is there a reason you can't add https://github.com/influxdata/telegraf/tree/master/plugins/outputs/health to setup a generic health endpoint for your telegraf node overall? This is what I use for my AWS ALB health checks.

@magichair I just try, but I can't start `output.health` and `input.influxdb_v2_listener` on the same port (which is perfectly logic) :/ 

Our telegraf agent is deployed on Kubernetes, and we can't set a different port for the health check on the load balancer. this feature is only available for 1.17 and is still "beta"

magichair:
That makes sense, if we wanted to resurrect it in a separate PR, it was roughly stubbed out and then removed in this commit: https://github.com/influxdata/telegraf/pull/7828/commits/e1571b914dfc60e0ac73cb2c2ad9b232909d7f02

PR #7834, opened 2020-07-14
sqlserver - Changed requests to include blocking sessions.

spaghettidba:
### Required for all PRs:

Changed query to gather requests to also include blocking sessions. The original query started from sys.dm_exec_requests, which excludes all idle queries (which have a session but not a request).
Also fixed query text truncation of the last character.

### DIFF
--- a/plugins/inputs/sqlserver/sqlserver.go
+++ b/plugins/inputs/sqlserver/sqlserver.go
@@ -1579,11 +1579,11 @@ SELECT
     , s.session_id
     , ISNULL(r.request_id,0) as request_id
     , DB_NAME(s.database_id) as session_db_name
-    , s.status
-    , s.cpu_time as cpu_time_ms
-    , s.total_elapsed_time as total_elapsed_time_ms
-    , s.logical_reads
-    , s.writes
+    , COALESCE(r.status,s.status) AS status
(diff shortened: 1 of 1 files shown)


danielnelson:
Thanks for the pull request, but could you update the pr to minimize any whitespace changes?  This will make the changes to the query a bit easier to understand and improve the history of the file.

spaghettidba:
@denzilribeiro No problem, I'll change it to a COALESCE. 

@danielnelson I certainly could reduce the number of whitespace changes, but the query is formatted in a way that makes it a bit hard to read. How do you suggest I can improve it? Should I open a second PR? Wouldn't that "clutter" the history of the file anyway? Should I avoid changing formatting at all?

spaghettidba:
Hi! Sorry to bother. Will this PR be merged now that it's approved? Do I need to do anything else?

Trovalo:
@ssoroka can you have a look at this?

denzilribeiro:
> It seems like coalescing the values from two different tables into a single column has the potential to provide really weird behavior. Would it be better to report these as separate columns instead?

No that is how it should be actually. Don't merge this, I will take care of it under PR #7934  given the major change if that is ok with you and @Trovalo

ssoroka:
Fine by me. I'll close this out, then.

PR #7838, opened 2020-07-14
ifname processor: expire old cached entries

reimda:
Also refetch interfaces when interface number is missing

### DIFF
--- a/plugins/processors/ifname/ifname.go
+++ b/plugins/processors/ifname/ifname.go
@@ -93,7 +93,7 @@ type IfName struct {
 	CacheSize          uint            `toml:"max_cache_entries"`
 	MaxParallelLookups int             `toml:"max_parallel_lookups"`
 	Ordered            bool            `toml:"ordered"`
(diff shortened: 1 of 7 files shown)


plugins/processors/ifname/ifname.go
```diff
@@ -82,17 +90,18 @@ type IfName struct {
 
 	snmp.ClientConfig
 
-	CacheSize          uint `toml:"max_cache_entries"`
-	MaxParallelLookups int  `toml:"max_parallel_lookups"`
-	Ordered            bool `toml:"ordered"`
+	CacheSize          uint            `toml:"max_cache_entries"`
+	MaxParallelLookups int             `toml:"max_parallel_lookups"`
+	Ordered            bool            `toml:"ordered"`
+	CacheTtl           config.Duration `toml:"cache_ttl"`
```
ssoroka, line 96: I think this should be CacheTTL, and some linters will complain about it.

PR #7848, opened 2020-07-16
ifname: avoid unpredictable conditions in getMap test

reimda:


plugins/processors/ifname/ifname_test.go
```diff
@@ -93,65 +92,47 @@ func TestIfName(t *testing.T) {
...
 	}
 
-	//signal mocked remote call to finish
-	wgRemote.Done()
-
-	//wait for requests to finish
-	wgReq.Wait()
-
-	//remote call should only happen once
+	// Remote call should not happen subsequent times getMap runs
```
danielnelson, line 136: I feel like you still need a waitgroup for these goroutines, otherwise it's possible they haven't event started running yet when the function ends.

PR #7861, opened 2020-07-20
Fix Ping Input plugin for FreeBSD's ping6

DasSkelett:
## Problem
FreeBSD has separate ping binaries for IPv4 and IPv6, `ping(8)` and `ping6(8)`.
Both have differently named options (or sometimes even with the same name but different function...),
and differently formatted outputs.

When using the `exec` method, the ping plugin, searches for a `ttl=` string to get the remaining TTL of the echo response, however `ping6` has a different output format and returns the TTL as `hlim=<integer>`. Thus the plugin throws this error:
```
E! [inputs.ping] Error in plugin: Fatal error processing ping output: 2a09::
```

Furthermore the ping plugin isn't aware of the different command options and always tries to use the arguments for the IPv4 `ping` command, which fails:
```
E! [inputs.ping] Error in plugin: host 2a09::: ping6: Non-recoverable failure in name resolution, exit status 1
```
`ping`'s `-W` is called `-x` for `ping6`
`ping`'s `-t` is called `-X` for `ping6`

## Changes
`processPingOutput()` searches for both `ttl=` and `hlim=` in the returned lines. The regex in `getTTL()` is updated accordingly, to match both cases.

`p.args()` now checks whether the binary string contains `ping6` and if the system / GOOS is `freebsd`, and if yes, adds the arguments needed for `ping6`.
This effectively will result in the following default command lines:
```sh
# IPv4 ping command, nothing changed
ping -c 1 -n -s 16 -i 1 -W 1000 -t 10 2a09::
# IPv6 ping6 command to reach the same effect
ping6 -c 1 -n -s 16 -i 1 -x 1000 -X 10 2a09::
```

## Unit tests
I've also added tests for this. I had to duplicate `TestArgs()`, since the expected result now also depends on other factors (`p.Binary`) besides the system.
With `iputils-ping` on Linux `ping6` is a symlink to `ping`, so the options are the same. Don't know for `inetuitls-ping`, but it isn't supported by this plugin anyways.
The test also assumes that "anything else" uses the same args for ping and ping6, since that is basically what's implemented now.

## PR testing

I have tested this myself on OPNsense 20.1.8 / FreeBSD 11.2-RELEASE-p21-HBSD
Here's an example config excerpt to reproduce the problem and to test the changes of this PR:

```ini
[[inputs.ping]]
  urls = ["8.8.8.8","216.58.207.36"]

[[inputs.ping]]
  urls = ["2001:4860:4860::8888","2a00:1450:4001:824::2004","2a09::"]
  binary = "ping6"
  ipv6 = true
```

### Required for all PRs:

- [x] Associated README.md updated. -> Not really, don't think there's anything to change in the README, is there?

### DIFF
--- a/plugins/inputs/ping/ping_notwindows.go
+++ b/plugins/inputs/ping/ping_notwindows.go
@@ -104,7 +104,13 @@ func (p *Ping) args(url string, system string) []string {
 		switch system {
 		case "darwin":
 			args = append(args, "-W", strconv.FormatFloat(p.Timeout*1000, 'f', -1, 64))
-		case "freebsd", "netbsd", "openbsd":
+		case "freebsd":
+			if strings.Contains(p.Binary, "ping6") {
+				args = append(args, "-x", strconv.FormatFloat(p.Timeout*1000, 'f', -1, 64))
+			} else {
+				args = append(args, "-W", strconv.FormatFloat(p.Timeout*1000, 'f', -1, 64))
+			}
+		case "netbsd", "openbsd":
 			args = append(args, "-W", strconv.FormatFloat(p.Timeout*1000, 'f', -1, 64))
 		case "linux":
 			args = append(args, "-W", strconv.FormatFloat(p.Timeout, 'f', -1, 64))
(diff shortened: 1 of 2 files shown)


DasSkelett:
If there's something else I can do or provide to simplify review and/or testing, please tell me.

DasSkelett:
Do I need to do something else to get someone to look at this? Did I forget something?

PR #7865, opened 2020-07-20
shim logger improvements

ssoroka:
- warn on loading default config
- move stdin reader to a goroutine so it doesn't block forever on local ctrl-c shutdown
- add shim logger interface that writes to stderr by default to work well with the shim environment.
- set `Log telegraf.Logger` if it's defined on plugins before calling plugin.Init() to avoid panics from plugins that expect to use it.
- add deprecation warning about using old shim.

### DIFF
--- a/plugins/common/shim/logger.go
+++ b/plugins/common/shim/logger.go
@@ -1,0 +1,87 @@
+package shim
+
+import (
+	"fmt"
+	"log"
+	"os"
+	"reflect"
+
+	"github.com/influxdata/telegraf"
+)
+
+func init() {
+	log.SetOutput(os.Stderr)
+}
+
+// Logger defines a logging structure for plugins.
+// external plugins can only ever write to stderr and writing to stdout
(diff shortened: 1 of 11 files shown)


plugins/common/shim/logger.go
```diff
@@ -0,0 +1,87 @@
...
+	log.Printf("I! "+format, args...)
+}
+
+// Info logs an information message, patterned after log.Print.
+func (l *Logger) Info(args ...interface{}) {
+	log.Print("I! ", fmt.Sprint(args...))
+}
+
+// setLoggerOnPlugin used to be called setLogIfExist
+func setLoggerOnPlugin(i interface{}, log telegraf.Logger) {
```
reimda, line 67: Why rename this when it's copied verbatim from models/log.go?  I don't particularly like the name setLogIfExist but it should match models/log.go.  If you're going to rename it, change both places.

ssoroka: I like this suggestion.

Issue #7859, opened 2020-07-20
Input execd - cannot use command parameters (on Windows)


Trovalo:
### Relevant telegraf.conf:

```toml
# Global tags can be specified here in key="value" format.
[global_tags]

# Configuration for telegraf agent
[agent]
  interval = "10s"
  round_interval = true
  metric_batch_size = 1000
  metric_buffer_limit = 10000
  collection_jitter = "0s"
  flush_interval = "10s"
  flush_jitter = "0s"

  precision = ""

  debug = true
  # quiet = false

  logtarget = 'stderr'
  # logfile = ''
  # logfile_rotation_interval = "14d"
  # logfile_rotation_max_size = "0MB"
  # logfile_rotation_max_archives = 5

  hostname = ""
  omit_hostname = false

###################################
#         OUTPUT PLUGINS           
###################################

[[outputs.file]]
  ## Files to write to, "stdout" is a specially handled file.
  files = ['C:\Users\gluisotto\Desktop\test_Telegraf\execd_output.out']

  data_format = "influx"

###################################
#         INPUT PLUGINS           
###################################
[[inputs.execd]]

  ## use bat file
  #command = ['C:\Users\gluisotto\Desktop\test_Telegraf\StartPrintrandom.bat']
  ## use command without params (default values)
  command = ['Powershell.exe -File "C:\Users\gluisotto\Desktop\test_Telegraf\PrintRandom.ps1"']
  ## use command with param
  #command = ["Powershell.exe -File 'C:\Users\gluisotto\Desktop\test_Telegraf\PrintRandom.ps1' -MaxWaitSec 2 -LoopNum 4"]

  signal = "none"

  restart_delay = "10s"

  data_format = "influx"

```

### System info:
Telegraf 1.15 (nightly)

### Steps to reproduce:
Here you can find all the files I've used:
[test_Telegraf.zip](https://github.com/influxdata/telegraf/files/4948259/test_Telegraf.zip)

1. Execute the PowerShell script on its own, it works (with or without params)
2. Execute the PowerShell script using _execd_ and with parameters - it fails
3. Execute the PowerShell script using _execd_ and without parameters - it fails
4. Execute the PowerShell script using _execd_ and by using a .bat file - it works

the same syntax in the _exec_ plugin works fine
> command = ['Powershell.exe -File "C:\Users\gluisotto\Desktop\test_Telegraf\PrintRandom.ps1"']

in the execd input I get the following error:
```
[...]

PR #7866, opened 2020-07-20
clarify docs and add warning if execd is misconfigured

ssoroka:
Addresses some of the concerns raised in https://github.com/influxdata/telegraf/issues/7859. Clarify docs and warn users if the plugin appears to be used in an unintended way.

### DIFF
--- a/plugins/inputs/execd/execd.go
+++ b/plugins/inputs/execd/execd.go
@@ -79,7 +79,9 @@ func (e *Execd) Start(acc telegraf.Accumulator) error {
 		// if there was only one argument, and it contained spaces, warn the user
 		// that they may have configured it wrong.
 		if len(e.Command) == 1 && strings.Contains(e.Command[0], " ") {
-			e.Log.Warn("Command contained spaces but no arguments. Check docs to make sure you've configured inputs.execd Command correctly.")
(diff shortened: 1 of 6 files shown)


plugins/inputs/execd/execd.go
```diff
@@ -75,6 +76,11 @@ func (e *Execd) Start(acc telegraf.Accumulator) error {
 	e.process.ReadStderrFn = e.cmdReadErr
 
 	if err = e.process.Start(); err != nil {
+		// if there was only one argument, and it contained spaces, warn the user
+		// that they may have configured it wrong.
+		if len(e.Command) == 1 && strings.Contains(e.Command[0], " ") {
+			e.Log.Warn("Command contained spaces but no arguments. Check docs to make sure you've configured inputs.execd Command correctly.")
```
reimda, line 82: This warning comes across a little like "RTFM".  I think we can tell the user how to fix the problem instead.

How about this: "The execd input's command setting contained spaces but no arguments.  This setting expects the program and arguments as an array of strings, not as a space delimited string."

ssoroka: I like your wording better, but it could possibly be correct, eg: `Command = ["c:\Program Files\Program.exe"] has no arguments, and a space, but is correct.

PR #7488, opened 2020-05-10
Fix float32 negative value/ add validation for range limit

garciaolais:
### Required for all PRs:

- [x] Signed [CLA](https://influxdata.com/community/cla/).
- [x] Associated README.md updated.
- [x] Has appropriate unit tests.

- Fix [7317](https://github.com/influxdata/telegraf/issues/7317)
- Add test for negative float32 values
- Updated Readme inside code

--------------------------------------

- Fix [7227](https://github.com/influxdata/telegraf/issues/7227)
- add validation for range limit

Issue #7317, opened 2020-04-11
Modbus input - negative values not supported with FLOAT32


kolmodin:
### Relevant telegraf.conf:

```toml
[[inputs.modbus]]
holding_registers = [
  { name = "outdoor_temperature",   byte_order = "AB",   data_type = "FLOAT32",   scale=0.1,     address = [903]},
]
```

Related issue: #7266.

The register holds a temperature scaled 10x. I want to store the decimals so I use FLOAT32.
When the temperature outdoors is 5.3° the register holds the int 53, and I want to use scaling to store 5.3.
When the temperature is below 0° the register holds a signed integer representing a negative value.
The problem is when the modbus plugin treats negative values.

### Expected behavior:

The modbus plugin should handle negative values as well as positive. When the register holds -13 it should be intepreted as -1.3 after scaling.

### Actual behavior:

Negative values are treated as unsigned. With a few negative degrees it stores the float 6554.0.
I see no way around this in telegraf except using the type INT16 instead of FLOAT32 and thus throw away the decimals.

### Additional info:

FLOAT32 is handled [here](https://github.com/influxdata/telegraf/blob/cc6c77f3012f2407780bd037a36dbaf9c5cac5a2/plugins/inputs/modbus/modbus.go#L504
). It assumes a UINT16 but INT16 values are also possible.

sensor-freak:
It seems to me that there is a need to differntiate between two scenarios regarding the handling of FLOAT32 fields:
1. the old behaviour: read as unsigned int, convert to float, scale
2. the new behaviour: read as signed int, convert to float, scale

Since there are for sure some people relying on the old behaviour, things will break for them when the new behaviour is introduced. But on the other hand, if we do not change anything, there will be people who cannot use this type at all.

Maybe we should resolve this conflict by introducing an additional type SFLOAT32, which implements the new behaviour. Old configurations will not be affected, and new configurations could make use of signed values.

Btw: FLOAT32 is misnamed, since the field value could be generated from 16, 32 or 64 byte integer values. Instead we should name it FLOAT (and SFLOAT) in the future. Or maybe even better add UFLOAT and FLOAT and leave FLOAT32 as it is.

sensor-freak:
I have updated PR #7869 which should now solve the issue from @kolmodin. That PR should now also supersede PR #7488.

srebhan:
[...]

PR #7869, opened 2020-07-21
Fix/extend support of fixed point values on input (modbus plugin)

sensor-freak:
### Required for all PRs:

This is a fix/extension for #7317, and it presumably superseedes PR #7488 (except for range validation)

The change introduces new input type FIXED and UNFIXED, and deprecates usage of the type FIXED32 (in docs only)


ssoroka:
Two main thoughts:
1. I'm a bit concerned that this would be a breaking change for anyone upgrading. 
2. Why is it that float32 needs to have a signed and unsigned version?

sensor-freak:
> Two main thoughts:
>     1. I'm a bit concerned that this would be a breaking change for anyone upgrading.

No, this change should not break anything for now. It just adds new type FIXED and UFIXED. The behaviour for all existing types, especially for FLOAT32 is (supposed to) be unchanged and therefore all existing configurations should work as currently.

>     2. Why is it that float32 needs to have a signed and unsigned version?

The FLOAT32 reads an integer value from modbus registers, converts this integer value to a float value, applies the given scale (by multiplication the float value with the configured scale value), and stores the result as a floating point type. When the sign bit of the input value from modbus registers is set, the result of the conversion from int to float will be different depending on the type of the input: for signed integer, the conversion must result in a negative value, while for unsigned integer the sign bit is the MSB of a positive value and thus the result must be a (large) positive value. Therefore we must be able to specify in the configuration which of these cases should be applied (it cannot be determined automatically, since there is no indication from modbus which one is the right one.)

Example:
- Modbus register value 0xFFFF (16 Bit), scale 0.01
- Interpretation with FIXED (signed): 
  -- (FLOAT64)(INT16)0xFFFF = -1.0; 
  -- -1.0 * scale = -0.01;
- Interpretation with UFIXED (unsigned):
  -- (FLOAT64)(UINT12)0xFFFF = 65535.0;
  -- 65535.0 * scale = 655.35;

(Please note that this is something else than FLOAT32-IEEE, which directly reads the bits from modbus registers into a float32 variable, without interpreting the bits as integer in the first place.)

And the need for both convertions is reality. I'm currently working with a modbus device, which uses both variants (signed and unsigned) with different scaling values in many combinations:

> 30795 Netzstrom AC (A) [Iac] 2 U32 FIX3 RO
> 30803 Netzfrequenz (Hz) [Fac] 2 U32 FIX2 RO
> 30805 Blindleistung (var) [Qac] 2 S32 FIX2 RO
> 30813 Scheinleistung3 (VA) [Sac] 2 S32 FIX0 RO

Here S32/U32 specifies INT32 and UINT32 resp., and FIX3, FIX2,... specifies the number of decimal digits contained in the integer value.

sensor-freak:
I've just re-checked again, it seems that I made a mistake in handling of FLOAT32 data type. Please assume this PR on hold until I've verified and probably corrected this. (The current state might be a breaking change for other users.)

sensor-freak:
The last commit fixes the break that I've introduced. With this commit the type FLOAT32 is handled identical to UFIXED, which is how it is handled in the current main branch. 

I have verified this by running the new unit tests with the code from the main branch. 

(The confusion came from PR #7488, which tried to introduced a breaking change. And after that I didn't realize that I had incorporated the idea of that change into my brain.)

sensor-freak:
Failing tests are not related to modbus plugin, and (seem to) do not fail in main branch. Can I force a re-run of the test somehow?

PR #7872, opened 2020-07-21
Recv next message after send returns EOF in gnmi input

danielnelson:
Should fix the intermittent test failure.

#Required for all PRs:


ssoroka:
Might be worth rerunning it a couple of times to see what happens?

danielnelson:
BTW, It seems the gRPC library is setup like this so that normally you would call Send and Recv from separate goroutines and each would know when to exit.  It is also documented that Send can block if Recv is not being called, so it may even be possible for the plugin as it is now to deadlock, although I think it is unlikely or even impossible since we Send once/Recv once in lockstep.

danielnelson:
Ran the workflow 3 times with no failures: https://circleci.com/gh/influxdata/workflows/telegraf/tree/fix-gnmi-test

PR #7874, opened 2020-07-22
Removing the master-slave terminology

bopopescu:
For diversity reasons, it would be nice to try to avoid 'master' and 'slave' terminology in this repository which can be associated to slavery. The master-slave terminology could be problematic for people in several countries which has the history of slavery like Romania, USA and many others. Thank you for considering the proposal. Let me know if any changes in the PR are needed, I would be happy to implement them.


ssoroka:
I support the spirit of this PR, but it's more complicated than a _find-and-replace_. There are a couple of issues we need to figure out how to navigate in the process of moving away from non-inclusive terminology:

1. Most references left in the project are matched to upstream projects. We can't make changes where it's going to break the build (the build is currently broken in this PR), and if the upstream project hasn't changed yet, we would need an adapter to switch to the upstream terminology; by definition, this means you can't remove the term from this project, and you're better off opening an issue with the upstream project. This _may_ mean following up their dependencies until you get to the root dependency or spec. For example, we've submitted a request to the MODBUS organization that manages the MODBUS spec to move from master/slave to client/server, which better describes what they're trying to communicate. I would encourage anyone interested in this to do the same with MODBUS and all dependencies.
2. We work to maintain backwards compatibility as much as possible. Flat-out removing a term is going to break existing deployments if it's mentioned in the config. The proper course of action here would be to deprecate it.
3. We need to be sensitive to documentation, guides, scripts, and other things that still might mention these terms and create a plan to migrate. 
4. The new term needs to make sense for the use. master/server is typically a lazy default that doesn't say much about the relationship. In the case of MODBUS, I believe the relationship their documentation describes is client/server. In the case of a master branch, "main" makes a lot of sense (though this migration will require a cut-over and a plan to update deploy scripts at the same time). For other plugins, the correct term may not be clear, but I would like to aim to improve code communication in the process, as much as we can.
5. We need to be sensitive to changing metric field and tag names, which would break existing user dashboards. It's not clear to me how to best deprecate the older uses yet.

Again, fully supportive, but it's going to be a long process.

ssoroka:
Since we need to look at this at a per-plugin approach (there are already some PRs to this effect), I'm going to close this out.

Snoopy9:
Each bus system defines the names and roles of its participants. For example "Modbus TCP" uses Client/Server and "Modbus RTU" use Master/Slave. Accordingly their function are also differently. To describe the telegram-plugin general for modbus and use some of this names to describe some plugin internal behaviours could confuse the reader a lot .

A specific telegram plug-in should therefore always use the names of the connected front end bus for the description of the bus behaviour . For the description of the interaction between the telegram plug-in  and InfluxDB etc.  the common names  used in the description of the connected bus should not be used again.

Issue #7871, opened 2020-07-21
Input consul - incompatible with output opentsdb


unflag:
### Relevant telegraf.conf:
```toml
[[inputs.consul]]
  address = "consul.service.consul:8500"
  scheme = "http"
  tag_delimiter = "="

[[outputs.opentsdb]]
  host = "http://opentsdb.service.consul"
  port = 8429
  httpBatchSize = 50000
```

### System info:
Telegraf 1.14.5 (git: HEAD e77ce3d1)

### Steps to reproduce:
Run telegraf using given configuration.

### Expected behavior:
All metrics successfully converted to opentsdb format.

### Actual behavior:
Some fields dropped due to invalid values:
```
2020-07-21T16:59:15Z D! OpenTSDB does not support metric value: [Service 'mesos-slave' check] of type [string].
```

### Additional info:
This behavior connected with this part of plugin:
```
func (c *Consul) GatherHealthCheck(acc telegraf.Accumulator, checks []*api.HealthCheck) {
	for _, check := range checks {
		record := make(map[string]interface{})
		tags := make(map[string]string)

		record["check_name"] = check.Name
		record["service_id"] = check.ServiceID

		record["status"] = check.Status
```
The fact is that OpenTSDB metric format doesn't allow string values, but output plugin uses fields as metric's value source and if field has no proper value - it just lost. So, I propose to place check_name, service_id and status fields into tags map - in this case all fields with string value would be safe with other tags. In my opinion it wouldn't disagree to anything. Prometheus has similar to opentsdb metric format and InfluxDB advises to not place commonly-queried metadata(which are check_name, service_id and status) in fields.

Maybe I missing something, if so - please point me. And if there is no objections I would like to submit PR.

danielnelson:
I think it would make sense to have these as tags, however since this is a breaking change it needs to be opt-in.  This could be done as a new `metric_version = 2` option, we could deprecate the current version and add a log message to recommend its use.  This would be similar to what was recently done in the `prometheus` input.

related: #2646

vikkyomkar:
@danielnelson I would like to work on this request. Pls let me know if its open to take ?

unflag:
@vikkyomkar sorry, i already took it up and submitted PR. Hope this allright.

Issue #2646, opened 2017-04-11
Make consul service_id a tag.


skygragon:
For example, the consul input will generate metric like:

> consul_health_checks,node=consul01,service_name=,check_id=serfHealth,host=consul01 check_name="Serf Health Status",service_id="",status="passing" 1491927292000000000 

Please note that those kvs like "check_name/service_id/status" will failed to send to openTSDB because it doesn't support string values....

skygragon:
Btw, just wondering why we use "service_name" as a tag, but "service_id" as a key name? Can we make those strings always as tags(or part of metric name) and only expose number as the value? e.g. metric="consul_service_<service_name/id>", node=xxx,xxx=xxx status=1|0

danielnelson:
I think it would make sense to make service_id a tag.  However, we shouldn't have empty tags or fields either, so we should also remove those tags.

danielnelson:
On the status part, I don't want to do that because it will make the point worse for other outputs, as well as be a breaking change.  Perhaps the opentsdb output could handle it or a processor that converts types.

danielnelson:
Also, making `service_id` a tag would be a breaking change, so we would need an option to opt-in.

skygragon:
@danielnelson Yep, I agree. As a input plugin, it's not proper to make special changes for openTSDB only. So who is in charge of openTSDB plugin,  better to hear his thoughts as well.

skygragon:
What about using separate metric for different service? put everything in the same metric "consul_health_checks" seems not necessary?

danielnelson:
I think having them in a single metric with the `service_name` as a tag is the proper way to structure them for most outputs.  How does the data look that is being sent to OpenTSDB?

PR #7875, opened 2020-07-22
input consul - added metric_version flag

unflag:
…  to control how to populate tags/fields of telegraf metrics. Addresses:
https://github.com/influxdata/telegraf/issues/7871
https://github.com/influxdata/telegraf/issues/2646

### Required for all PRs:

### DIFF
--- a/plugins/inputs/consul/README.md
+++ b/plugins/inputs/consul/README.md
@@ -17,6 +17,13 @@ report those stats already using StatsD protocol if needed.
   ## URI scheme for the Consul server, one of "http", "https"
   # scheme = "http"
 
+  ## Metric version controls the mapping from Consul metrics into
+  ## Telegraf metrics.
+  ##
+  ##   example: metric_version = 1; deprecated in 1.15
+  ##            metric_version = 2; recommended version
+  # metric_version = 1
+
   ## ACL token used in every request
   # token = ""
 
--- a/plugins/inputs/consul/consul.go
+++ b/plugins/inputs/consul/consul.go
@@ -20,6 +20,8 @@ type Consul struct {
 	Datacenter string
 	tls.ClientConfig
 	TagDelimiter string
+	MetricVersion int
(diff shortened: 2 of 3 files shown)


plugins/inputs/consul/consul.go
```diff
@@ -110,15 +127,21 @@ func (c *Consul) GatherHealthCheck(acc telegraf.Accumulator, checks []*api.Healt
...
-		record["status"] = check.Status
 		record["passing"] = 0
 		record["critical"] = 0
 		record["warning"] = 0
 		record[check.Status] = 1
 
+		if c.MetricVersion == 2 {
+			tags["check_name"] = check.Name
+			tags["service_id"] = check.ServiceID
+			tags["status"] = check.Status
```
ssoroka, line 138: status seems like it should stay as a field?

plugins/inputs/consul/README.md
```diff
@@ -17,6 +17,13 @@ report those stats already using StatsD protocol if needed.
   ## URI scheme for the Consul server, one of "http", "https"
   # scheme = "http"
 
+  ## Metric version controls the mapping from Consul metrics into
+  ## Telegraf metrics.
+  ##
+  ##   example: metric_version = 1; deprecated in 1.15
```
ssoroka, line 23: Maybe also explain what changed between versions here.

unflag: Sure, thanks. I''ll add details about v2 into Metrics section.

plugins/inputs/consul/consul.go
unflag: I think it should not. It will duplicate other fields in such case. Tag 'status' is just string representation of fields passing/critical/warning, it's value could be convenient to use in alert templates, for example. In my opinion it is more like metadata than value that one can add into some computations like get avg, max or min.

ssoroka (approved):

PR #7886, opened 2020-07-23
Fix tail stops following on first EOF

### LINKED ISSUE #7880: Tailing files stopped working since v1.15
I've upgraded from Telegraf v1.14 to v1.15.1 in our systems today and found that it broke one of our use cases.

We are tailing some log-like files from a service, the service outputs lines in InfluxDB format which is then read by Telegraf and passed on to our metrics system. I made a minimal example of this:

`telegraf.conf`
```toml
[[inputs.tail]]
  files = ["clock.log"]
  data_format = "influx"

[[processors.printer]]

[[outputs.discard]]
```

I've then run this loop in bash in order to write to `clock.log`:
```bash
while :; do
    echo "clock now=$(date +%s)"
    sleep 1
done >> clock.log
```

Using that example I was able to use Git bisect which pointed at this stopping working when #7792 was merged.

I've tried to set `character_encoding` in the config file to both an empty string and "utf-8" but it doesn't seem to make any difference, so I think the problem might more deeply rooted.

danielnelson:
The Reader from x/text stops reading after the first io.EOF error is returned, preventing further tailing of the file when new lines are appended.

All code below is brought over from x/text in support of adding two lines to the Read() function:
```
r.err = nil
r.transformComplete = false
```

closes #7880

### Required for all PRs:

### DIFF
--- a/plugins/common/encoding/decoder.go
+++ b/plugins/common/encoding/decoder.go
@@ -2,12 +2,12 @@ package encoding
 
 import (
 	"errors"
+	"io"
 
 	"golang.org/x/text/encoding"
 	"golang.org/x/text/encoding/unicode"
+	"golang.org/x/text/transform"
 )
-
-type Decoder = encoding.Decoder
 
 // NewDecoder returns a x/text Decoder for the specified text encoding.  The
 // Decoder converts a character encoding into utf-8 bytes.  If a BOM is found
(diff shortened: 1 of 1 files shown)


danielnelson:
Testing builds

- [telegraf_1.16.0~4cdc3501-0_amd64.deb](https://61028-33258973-gh.circle-artifacts.com/0/build/dist/telegraf_1.16.0~4cdc3501-0_amd64.deb)
- [telegraf-1.16.0~4cdc3501-0.x86_64.rpm](https://61028-33258973-gh.circle-artifacts.com/0/build/dist/telegraf-1.16.0~4cdc3501-0.x86_64.rpm)
- [telegraf-1.16.0~4cdc3501_windows_amd64.zip](https://61028-33258973-gh.circle-artifacts.com/0/build/dist/telegraf-1.16.0~4cdc3501_windows_amd64.zip)
- [telegraf-1.16.0~4cdc3501_freebsd_amd64.tar.gz](https://61028-33258973-gh.circle-artifacts.com/0/build/dist/telegraf-1.16.0~4cdc3501_freebsd_amd64.tar.gz)
- [telegraf-1.16.0~4cdc3501_darwin_amd64.tar.gz](https://61028-33258973-gh.circle-artifacts.com/0/build/dist/telegraf-1.16.0~4cdc3501_darwin_amd64.tar.gz)

PR #7890, opened 2020-07-24
Update procstat to support cgroup globs & include systemd unit children

DEvil0000:
* cgroups now support glob
* systemd_unit in combination with systemd_all will now report for all children and not just the main pid of the service
* in case cgroups or systemd_unit+all will return a group of pids they get tagged with their actual cgroup/service matched (cgroup_full)

### DIFF
--- a/plugins/inputs/procstat/procstat.go
+++ b/plugins/inputs/procstat/procstat.go
@@ -4,11 +4,11 @@ import (
 	"bytes"
 	"fmt"
 	"io/ioutil"
+	"os"
 	"os/exec"
 	"path/filepath"
 	"strconv"
 	"time"
-	"os"
 
 	"github.com/influxdata/telegraf"
 	"github.com/influxdata/telegraf/plugins/inputs"
@@ -32,7 +32,7 @@ type Procstat struct {
 	ProcessName string
 	User        string
 	SystemdUnit string
-	SystemdAll  bool `toml:"systemd_all"`
+	SystemdAll  bool   `toml:"systemd_all"`
 	CGroup      string `toml:"cgroup"`
 	PidTag      bool
 	WinService  string `toml:"win_service"`
@@ -388,7 +388,7 @@ func (p *Procstat) findPids() []PIDS_TAGS_ERR_GROUP {
 		return groups
 	} else {
 		pids, tags, err := p.SimpleFindPids(f)
-                pids_tags_err_groups = append(pids_tags_err_groups, PIDS_TAGS_ERR_GROUP{pids, tags, err})
+		pids_tags_err_groups = append(pids_tags_err_groups, PIDS_TAGS_ERR_GROUP{pids, tags, err})
 	}
 
 	return pids_tags_err_groups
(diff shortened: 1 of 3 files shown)


TelmoM:
Hi @sjwang90, is this update going to be merged soon?
Thank you

sjwang90:
Sorry for the delay - we'll get this reviewed.

plugins/inputs/procstat/procstat.go
```diff
@@ -92,7 +95,14 @@ func (_ *Procstat) Description() string {
 	return "Monitor process cpu and memory usage"
 }
 
+type PIDS_TAGS_ERR_GROUP struct {
```
sspaink, line 98: I think that putting PID and tags together in a struct makes sense, but it might be more idiomatic to keep error separate.

plugins/inputs/procstat/procstat.go
```diff
@@ -354,16 +372,34 @@ func (p *Procstat) getPIDFinder() (PIDFinder, error) {
...
+	} else {
+		pids, tags, err := p.SimpleFindPids(f)
+		pids_tags_err_groups = append(pids_tags_err_groups, PIDS_TAGS_ERR_GROUP{pids, tags, err})
 	}
 
+	return pids_tags_err_groups
+}
+
+// Get matching PIDs and their initial tags
+func (p *Procstat) SimpleFindPids(f PIDFinder) ([]PID, map[string]string, error) {
```
sspaink, line 398: Can you explain why the word "simple" is being used to describe this function ? It seems it could return a `PIDS_TAGS_ERR_GROUP` struct as well.

sspaink (changes requested):
Thank you for working on this pull request! It does look it will improve the procstat plugin, I've added some review comments for you to look at.

DEvil0000:
@sspaink thanks for the review!
However go is not one of the programming languages I use (at all) so I tried to do it in a kind of clean way and sticking close to the original code. About 1y after creating the PR I don't feel like I want to mess with those details (at least currently I don't have time to do so - sorry). Please fix or change what you feel meets the projects coding style or standards best e.g. in a additional commit.
I would absolutely appreciate having this features in the master so I may be able to go back to the official version instead of maintaining my own one.

sspaink:
@DEvil0000 thank you for letting us know, sorry for taking so long with the review. I'll look into taking over the pull request and work with the team to get it merged. I will be sure mention your contribution and let you know once it is in master/released.

DEvil0000:
Thank you! Thank you for taking care!

plugins/inputs/procstat/procstat.go
sspaink: Looking more into this, I see why it is structured this way.

plugins/inputs/procstat/procstat.go
sspaink: Looking more into this, I see why it is structured this way.

sspaink:
#9488 will continue this pull request.

PR #7897, opened 2020-07-25
common/tls: Allow specifying SNI hostnames

### LINKED ISSUE #7598: Override IP address in http_response
### Proposal:
Add a config setting to the http_response input plugin to override the IP address, to use instead of the resolved hostname. Alternatively, add a way to override the hostname when giving an IP address to `urls`.

### Current behavior:
If I set `urls=["http://somehostname/"]`, telegraf will resolve that hostname in DNS and connect to that IP address. If there are multiple IP addresses, I guess it connects to an arbitrary one. I can't work around it with `urls=["http://192.0.2.42/"]` and overriding `inputs.http_response.headers.Host`, because then it would fail to verify SSL. Even with `insecure_skip_verify` it would fail because it wouldn't send the correct hostname in SNI.

### Desired behavior:
A new config setting lets me change the IP address the plugin connects to, so I can use the hostname in the URL but connect to another IP.

Alternatively, a new config setting lets me change the hostname used in the Host: header, the SSL SNI, and the SSL certificate verification, so I can use the IP address in the URL but everything else still behaves as if the URL had the hostname.

In either case, both the hostname and the IP address should be somewhere in the resulting tags, so that requests to same hostname but different IP, or viceversa, can be distinguished later.

### Use case: 
A hostname may point to multiple IP addresses corresponding to different servers. Or it may have both IPv4 and IPv6. Or it may use geodns to return a single IP out of multiple depending on the source's location, and a telegraf instance would always get the same one.

I want to send a request to all IP addresses individually, and collect the response_time for each, so I can know if there is a failure or increased latency for only one server, or for only IPv6.

antifuchs:
This PR adds a new configration field `sni_server_name` that allows specifying the server name that'll be sent in the ClientHello when telegraf makes an HTTP request to TLS servers. This allows checking against load balancers responding to specific hostnames that otherwise wouldn't resolve to their addresses, or other more complex TLS configurations.

Fixes #7598.

### Required for all PRs:

### DIFF
--- a/docs/TLS.md
+++ b/docs/TLS.md
@@ -18,6 +18,8 @@ For client TLS support we have the following options:
 # tls_key = "/etc/telegraf/key.pem"
 ## Skip TLS verification.
 # insecure_skip_verify = false
+## Send the specified TLS server name via SNI.
+# tls_server_name = "foo.example.com"
 ```
 
 ### Server Configuration
--- a/plugins/common/tls/config.go
+++ b/plugins/common/tls/config.go
@@ -46,10 +46,13 @@ func (c *ClientConfig) TLSConfig() (*tls.Config, error) {
 		c.TLSKey = c.SSLKey
 	}
 
-	// TODO: return default tls.Config; plugins should not call if they don't
-	// want TLS, this will require using another option to determine.  In the
-	// case of an HTTP plugin, you could use `https`.  Other plugins may need
-	// the dedicated option `TLSEnable`.
+	// This check returns a nil (aka, "use the default")
+	// tls.Config if no field is set that would have an effect on
+	// a TLS connection. That is, any of:
+	//     * client certificate settings,
+	//     * peer certificate authorities,
+	//     * disabled security, or
+	//     * an SNI server name.
 	if c.TLSCA == "" && c.TLSKey == "" && c.TLSCert == "" && !c.InsecureSkipVerify && c.ServerName == "" {
 		return nil, nil
 	}
--- a/plugins/common/tls/config.go
+++ b/plugins/common/tls/config.go
@@ -14,6 +14,7 @@ type ClientConfig struct {
 	TLSCert            string `toml:"tls_cert"`
 	TLSKey             string `toml:"tls_key"`
 	InsecureSkipVerify bool   `toml:"insecure_skip_verify"`
+	ServerName         string `toml:"tls_server_name"`
 
 	// Deprecated in 1.7; use TLS variables above
 	SSLCA   string `toml:"ssl_ca"`
@@ -49,7 +50,7 @@ func (c *ClientConfig) TLSConfig() (*tls.Config, error) {
 	// want TLS, this will require using another option to determine.  In the
 	// case of an HTTP plugin, you could use `https`.  Other plugins may need
 	// the dedicated option `TLSEnable`.
-	if c.TLSCA == "" && c.TLSKey == "" && c.TLSCert == "" && !c.InsecureSkipVerify {
+	if c.TLSCA == "" && c.TLSKey == "" && c.TLSCert == "" && !c.InsecureSkipVerify && c.ServerName == "" {
 		return nil, nil
 	}
 
@@ -71,6 +72,10 @@ func (c *ClientConfig) TLSConfig() (*tls.Config, error) {
 		if err != nil {
 			return nil, err
 		}
+	}
+
+	if c.ServerName != "" {
+		tlsConfig.ServerName = c.ServerName
 	}
 
 	return tlsConfig, nil
--- a/plugins/common/tls/config_test.go
+++ b/plugins/common/tls/config_test.go
@@ -15,11 +15,10 @@ var pki = testutil.NewPKI("../../../testutil/pki")
 
 func TestClientConfig(t *testing.T) {
 	tests := []struct {
-		name       string
-		client     tls.ClientConfig
-		expNil     bool
-		expErr     bool
-		serverName string
+		name   string
+		client tls.ClientConfig
+		expNil bool
+		expErr bool
 	}{
 		{
 			name:   "unset",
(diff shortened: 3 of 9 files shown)


antifuchs:
The idea behind this PR is to test with something like the following setting:

```toml
[[inputs.http_response]]
insecure_skip_verify = false
response_string_match = "<title>Login</title>"
response_timeout = "10s"
urls = [
    "https://load-balancer-a.internal.example.com/login/",
    "https://load-balancer-b.internal.example.com/login/",
]
tls_ca = "/run/secrets/tls-cacert"
sni_server_name = "www.extremely-important-apex.com"
[inputs.http_response.headers]
Host = "www.extremely-important-apex.com"
```

In this example, `load-balancer-a` and `-b` are internal addresses and don't get resolved by `www.extremely-important-apex.com`; but they require the correct SNI server name to serve the content that they should serve (as well as the right Host header).

Hope that makes sense!

srebhan:
I think this is not the correct approach. We rather should add this option to the `tls.ClientConfig` so it is available for all plugins using TLS.
@antifuchs are you willing to look into this!?

antifuchs:
Huh, I didn't realize there was a common way to configure plugins' TLS client config. I'll look into it!

docs/TLS.md
```diff
@@ -44,23 +46,23 @@ for the interest of brevity.
...
-## 	         "TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA",
-## 	         "TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256",
-## 	         "TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA",
-## 	         "TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA",
-## 	         "TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA",
-## 	         "TLS_RSA_WITH_AES_128_GCM_SHA256",
-## 	         "TLS_RSA_WITH_AES_256_GCM_SHA384",
-## 	         "TLS_RSA_WITH_AES_128_CBC_SHA256",
-## 	         "TLS_RSA_WITH_AES_128_CBC_SHA",
-## 	         "TLS_RSA_WITH_AES_256_CBC_SHA"
```
antifuchs, line 63: Ugh, I'm sorry, that was my editor - I'll rebase that away.

plugins/common/tls/config.go
```diff
@@ -49,7 +50,7 @@ func (c *ClientConfig) TLSConfig() (*tls.Config, error) {
 	// want TLS, this will require using another option to determine.  In the
 	// case of an HTTP plugin, you could use `https`.  Other plugins may need
 	// the dedicated option `TLSEnable`.
-	if c.TLSCA == "" && c.TLSKey == "" && c.TLSCert == "" && !c.InsecureSkipVerify {
+	if c.TLSCA == "" && c.TLSKey == "" && c.TLSCert == "" && !c.InsecureSkipVerify && c.ServerName == "" {
```
srebhan, line 53: I don't think this is correct. If you neither specify a CA nor a KEY or CERT but a `ServerName` then this check will not trigger but I doubt it makes sense to check the servername without validating the certificate!?

srebhan (commented):
One question in the code and a request to add a test-case for only adding a server-name but no other TLS configuration... :-)

plugins/common/tls/config.go
antifuchs: The `ServerName` directive [does not affect only the host name in certificate checking](https://golang.org/src/crypto/tls/common.go#L588): It's also the hostname requested via SNI in the TLS [`ClientHello`](https://en.wikipedia.org/wiki/Server_Name_Indication#Technical_principles) message - that's the thing that I need to customize that prompted me to submit the PR in the first place (:

Even if that weren't the case, an empty TLS Config is [documented to use the host's root CA](https://golang.org/src/crypto/tls/common.go#L579) set if no root CAs are provided, so setting that value on an otherwise-empty TLS config would be meaningful.

srebhan: Could you please put a comment there linking to the docu you mentioned. Just for the next one asking himself why this should make sense!? Thanks for the explaination!

antifuchs: Hm, so I re-read the TODO comment above & started auditing our code's usage of `TLSConfig()`, where I found out that the only case where a nil tls.Config matters (because the stdlib function tls.Client requires a non-nil value) is the x509_cert.go code... where we override the ServerName based on a configuration setting (incompatibly with the "common" ServerName setting now)! 

So I'll have to fix the ServerName usage in x509_cert.go, but more importantly: I think we should remove that TODO, as it could be a pretty meaningful distinction in terms of "were there any settings that the operator customized" (but ultimately, it looks like there's not much of a difference). What do you think?

antifuchs:
@srebhan I think this should cover what we discussed above - lmk if you see any other things that need improving!

srebhan (approved):
Nice catch with the x509 plugin! Looks good to me.

plugins/common/tls/config_test.go
```diff
@@ -15,10 +15,11 @@ var pki = testutil.NewPKI("../../../testutil/pki")
...
 	tests := []struct {
-		name   string
-		client tls.ClientConfig
-		expNil bool
-		expErr bool
+		name       string
+		client     tls.ClientConfig
+		expNil     bool
+		expErr     bool
+		serverName string
```
sspaink, line 22: `serverName` doesn't seem to be used in this test. Can you remove it please?

sspaink (commented):
Great work @antifuchs ! I just added a minor comment.

plugins/common/tls/config_test.go
antifuchs: Yup, vestige of a previous idea for a test - Removing it now.

PR #7908, opened 2020-07-28
Fix for opentsdb output-plugin for NaN and Inf in float64 value

Aladex:
### Required for all PRs:

When in prometheus metrics we have a value with NaN or Inf, Telegraf cannot to send metrics in TSDB and write the error in stdout:

```
2020-07-28T08:42:32Z E! [agent] Error writing to outputs.opentsdb: Metric serialization error json: unsupported value: NaN
```

I fixed this error like in json serializer: https://github.com/influxdata/telegraf/blob/master/plugins/serializers/json/json.go#L61

### DIFF
--- a/plugins/outputs/opentsdb/opentsdb.go
+++ b/plugins/outputs/opentsdb/opentsdb.go
@@ -3,6 +3,7 @@ package opentsdb
 import (
 	"fmt"
 	"log"
+	"math"
 	"net"
 	"net/url"
 	"regexp"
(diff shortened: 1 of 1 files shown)


plugins/outputs/opentsdb/opentsdb.go
```diff
@@ -136,10 +137,14 @@ func (o *OpenTSDB) WriteHttp(metrics []telegraf.Metric, u *url.URL) error {
...
 
 		for fieldName, value := range m.Fields() {
-			switch value.(type) {
+			switch fv := value.(type) {
 			case int64:
 			case uint64:
 			case float64:
+				// JSON does not support these special values
+				if math.IsNaN(fv) || math.IsInf(fv, 0) {
+					continue
```
ssoroka, line 146: Looks like elsewhere these are serialized as `null`. should we be doing that instead?

Aladex: > 
> 
> Looks like elsewhere these are serialized as `null`. should we be doing that instead?

I do not think so. I dont found any of. So. Here we have a type checking, its logical to add checking of float64 for NaN or Inf. And we have errors for NaN from 1.14 version =)

PR #7922, opened 2020-07-30
Proxmox plugin

effitient:
### Required for all PRs:

### DIFF
--- a/plugins/inputs/proxmox/proxmox.go
+++ b/plugins/inputs/proxmox/proxmox.go
@@ -37,13 +37,7 @@ func (px *Proxmox) Description() string {
 }
 
 func (px *Proxmox) Gather(acc telegraf.Accumulator) error {
-	hostname, err := os.Hostname()
-	if err != nil {
-		return err
-	}
-	px.hostname = hostname
-
-	err = getNodeSearchDomain(px)
+	err := getNodeSearchDomain(px)
 	if err != nil {
 		return err
 	}
@@ -54,10 +48,32 @@ func (px *Proxmox) Gather(acc telegraf.Accumulator) error {
 	return nil
 }
 
+func (px *Proxmox) Init() error {
+	hostname, err := os.Hostname()
+	if err != nil {
+		return err
+	}
+	px.hostname = hostname
+
+	tlsCfg, err := px.ClientConfig.TLSConfig()
+	if err != nil {
+		return err
+	}
+	px.httpClient = &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: tlsCfg,
+		},
+		Timeout: px.ResponseTimeout.Duration,
+	}
+
+	return nil
+}
+
 func init() {
 	px := Proxmox{
 		requestFunction: performRequest,
 	}
+
 	inputs.Add("proxmox", func() telegraf.Input { return &px })
 }
 
@@ -76,24 +92,13 @@ func getNodeSearchDomain(px *Proxmox) error {
 }
 
 func performRequest(px *Proxmox, apiUrl string, method string, data url.Values) ([]byte, error) {
-	tlsCfg, err := px.ClientConfig.TLSConfig()
+	request, err := http.NewRequest(method, px.BaseURL+apiUrl, strings.NewReader(data.Encode()))
 	if err != nil {
 		return nil, err
 	}
-	client := &http.Client{
-		Transport: &http.Transport{
-			TLSClientConfig: tlsCfg,
-		},
-		Timeout: px.ResponseTimeout.Duration,
-	}
+	request.Header.Add("Authorization", "PVEAPIToken="+px.APIToken)
 
-	request, err := http.NewRequest(method, px.BaseUrl+apiUrl, strings.NewReader(data.Encode()))
-	if err != nil {
-		return nil, err
-	}
-	request.Header.Add("Authorization", "PVEAPIToken="+px.ApiToken)
-
-	resp, err := client.Do(request)
+	resp, err := px.httpClient.Do(request)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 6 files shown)


ssoroka:
This looks interesting. Thanks! I'll take a thorough look at it shortly.

plugins/inputs/proxmox/proxmox.go
```diff
@@ -0,0 +1,259 @@
...
+func (px *Proxmox) Description() string {
+	return "Provides metrics from Proxmox nodes (Proxmox Virtual Environment > 6.2)."
+}
+
+func (px *Proxmox) Gather(acc telegraf.Accumulator) error {
+	hostname, err := os.Hostname()
+	if err != nil {
+		return err
+	}
+	px.hostname = hostname
```
ssoroka, line 44: I'm thinking we should do this once on Init. if you define a function with the signature `Init() error` it will be called once when the plugin is initialized.

plugins/inputs/proxmox/proxmox.go
```diff
@@ -0,0 +1,259 @@
...
+
+func init() {
+	px := Proxmox{
+		requestFunction: performRequest,
+	}
+	inputs.Add("proxmox", func() telegraf.Input { return &px })
+}
+
+func getNodeSearchDomain(px *Proxmox) error {
+	apiUrl := "/nodes/" + px.hostname + "/dns"
```
ssoroka, line 65: does this need to be escaped?

plugins/inputs/proxmox/proxmox.go
```diff
@@ -0,0 +1,259 @@
...
+	tlsCfg, err := px.ClientConfig.TLSConfig()
+	if err != nil {
+		return nil, err
+	}
+	client := &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: tlsCfg,
+		},
+		Timeout: px.ResponseTimeout.Duration,
+	}
```
ssoroka, line 88: the client should be set once and reused. Try moving it to the `Init() error` method

plugins/inputs/proxmox/proxmox.go
```diff
@@ -0,0 +1,259 @@
...
+	var vmStats VmStats
+	err = json.Unmarshal(jsonData, &vmStats)
+	if err != nil {
+		return VmStats{}, err
+	}
+
+	return vmStats, nil
+}
+
+func getVmConfig(px *Proxmox, vmId string, lxc bool) (VmConfig, error) {
```
ssoroka, line 162: not a big fan of boolean args. a resource type might be better?
```
type resourceType string
var (
  rtQEMU resourceType = "qemu"
  rtLXC resourceType = "lxc"
)
```
This would let you make the `lxc bool` arg just `rt resourceType` and include it directly in the path.

plugins/inputs/proxmox/structs.go
```diff
@@ -0,0 +1,51 @@
...
+	"encoding/json"
+	"github.com/influxdata/telegraf/internal"
+	"github.com/influxdata/telegraf/plugins/common/tls"
+	"net/url"
+)
+
+type Proxmox struct {
+	BaseUrl         string            `toml:"base_url"`
+	ApiToken        string            `toml:"api_token"`
+	ResponseTimeout internal.Duration `toml:"response_timeout"`
```
ssoroka, line 13: Optional, we have config.Duration now which wraps the time.Duration instead of composing it. Fine either way.

plugins/inputs/proxmox/structs.go
```diff
@@ -0,0 +1,51 @@
...
+
+import (
+	"encoding/json"
+	"github.com/influxdata/telegraf/internal"
+	"github.com/influxdata/telegraf/plugins/common/tls"
+	"net/url"
+)
+
+type Proxmox struct {
+	BaseUrl         string            `toml:"base_url"`
```
ssoroka, line 11:
```suggestion
	BaseURL         string            `toml:"base_url"`
```

plugins/inputs/proxmox/structs.go
```diff
@@ -0,0 +1,51 @@
...
+	ApiToken        string            `toml:"api_token"`
```
ssoroka, line 12:
```suggestion
	APIToken        string            `toml:"api_token"`
```

plugins/inputs/proxmox/structs.go
```diff
@@ -0,0 +1,51 @@
...
+
+	requestFunction func(px *Proxmox, apiUrl string, method string, data url.Values) ([]byte, error)
+}
+
+type VmStats struct {
+	Data []VmStat `json:"data"`
+}
+
+type VmStat struct {
+	Id        string      `json:"vmid"`
```
ssoroka, line 27:
```suggestion
	ID        string      `json:"vmid"`
```

plugins/inputs/proxmox/structs.go
ssoroka: Running linters should highlight casing issues like this one.

plugins/inputs/proxmox/proxmox_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+	return bytedata, nil
+}
+
+func setUp() *Proxmox {
+	return &Proxmox{
+		requestFunction: performTestRequest,
+		hostname:        "testnode",
+	}
+}
```
ssoroka, line 41:
```suggestion
func setUp(t *testing.T) *Proxmox {
	p := &Proxmox{
		requestFunction: performTestRequest,
		hostname:        "testnode",
	}
	require.NoError(t, p.Init())
}
```

plugins/inputs/proxmox/proxmox.go
effitient: It's not escaped, it is a path.

plugins/inputs/proxmox/proxmox.go
```diff
@@ -0,0 +1,251 @@
...
+}
+
+func gatherQemuData(px *Proxmox, acc telegraf.Accumulator) {
+	gatherVmData(px, acc, QEMU)
+}
+
+func gatherVmData(px *Proxmox, acc telegraf.Accumulator, rt ResourceType) {
+	vmStats, err := getVmStats(px, rt)
+	if err != nil {
+		log.Fatal(err)
```
ssoroka, line 126: I missed this the first time around. Don't use Fatal as it won't handle errors gracefully. Instead, add the following to the struct:
```
```

At runtime it'll be given a logger instance, and you can use it to do `px.Log.Errorf("Error getting vm stats: %v", err)`

You can optionally update the tests to set to set the logger instance with a test logger: `px.Log = testutil.Logger{}` so they don't error when a log line is called.

ssoroka: After this I think you're good.

ChappIO:
Sweet! I really hope this gets released soon so I no longer have to run a binary I built from source.

ssoroka:
It'll be in 1.16, which should be soon.

PR #7808, opened 2020-07-09
SQL Server - added new counter - Lock Timeouts (timeout > 0)/sec

Trovalo:
### Required for all PRs:

- [x] Signed [CLA](https://influxdata.com/community/cla/).
- [ ] Associated README.md updated.
- [ ] Has appropriate unit tests.

closes #6978

Added a new counter 'Lock Timeouts (timeout > 0)/sec' to the sqlserver_performance measurement.

Issue #7724, opened 2020-06-24
Sql Server - remove tag "database_name" is some collectors


Trovalo:
## Feature Request

In most collectors the "database_name" tag is useless (at least when obtained using DB_NAME() ), since the queries are instance wide and do not use a database scope, so in general it will just return a fixed value, probably "master". 

Those are the queries in which the tag looks useless to me.
- PerformanceCounters
- WaitStatsCategorized
- ServerProperties
- MemoryClerk
- Schedulers
- SqlRequests (here the useful tag is called "session_db_name")

### Proposal:
Remove the tag "database_name" when not useful

denzilribeiro:
This was added primarily for the Azure SQL DB case as a logical server isn't a n instance in Azure and there is no way to correlate DB's otherwise. Once we split things up  per # 7713 will be much easier to remove for on-prem. Right now some of the collectors are shared.

Trovalo:
THis has been done in the query splitting and also in the later query refactoring

PR #7934, opened 2020-08-03
Database_type config to Split up sql queries by engine type

### LINKED ISSUE #7713: Split up sql script and config options for SQL Server plugin on database_type
## Feature Request
a. Split up collector queries base don whether they are for Cloud versions ( Azure SQL DB / Managed instance) or on-prem
b.  Enable a new configuration option called database_type or database_engine_type to simplify the queries and make maintainability of the plugin and test matrix much easier.

### Proposal:
Introduce a configuration parameter which could have one of 4 values
database_type =  "azure_sql_db" / "azure_sql_managed_instance" / "sqlserver"

Based on the choice, then a set of queries will be enabled for this collector allowing the normal inclusion/exclusion within that set. This will allow us to separate queries out and make them smaller rather than each collector having to support all 3 engine types. In future I could add "azure_sql_elastic_pools" support which is a different concept totally.

If the user does NOT specify this parameter, it will default to the older mechanisms based on V1/V2 choice of queries. If they do, then will follow this path

In code then could have a set of queries enabled by each type:
a. If  database_type="sql_server", then  Queries allowed =  Perfstats, WaitStats, DatabaseIO ,AvailabilityGroups ...
b, If database_type = "azure_sql_db" then Queries allowed = PerfStats, AzureDBWaitstats, WaitStats, 
, DatabaseIO, AzureDBResourceStats

This would also enable Config to be simpler for example could have a section each with servers for each type of SQL edition
 [[inputs.sqlserver]]
     servers = ["Server=192.168.1.10;Port=1433;UserId=<user>;Password=<pw>;app name=telegraf;log=1;"]
database_type = "azure_sql_db"

 [[inputs.sqlserver]]
     servers = ["Server=201.168.1.20;Port=1433;UserId=<user>;Password=<pw>;app name=telegraf;log=1;"]
database_type = "sql_server"

### Current behavior:
Currently from a configuration option you have 3 possible alternatives
a. You choose V1 queries - which are obsolete today and won't really cover this as not many people using this today.
b. You choose V2 queries - This covers all the queries relevant for on-premise version of SQL server. These queries are also elligible for all the cloud versions which means each of these queries when changed have to be tested for all versions SQL 2008 - SQL 2019, Azure SQL Managed instance Azure SQL Database
c. You choose azuredb = true/false - this enables few additional queries for cloud  versions ( Azure SQL Managed instance/ Azure SQL DB.

Irrespective if a query is "relevant" to cloud versions or not, they are run unless excluded, each query today now has to check Engine Edition and branch out  which means in all cases a connection is still made to SQL server, and the check is within the SQL query.

### Desired behavior:
It would be much easier for a user to have 3 config sections and split out their connection strings based on database type.
This would enable "less" queries to run based on engine type

### Use case: 
Today there are fundamentally 2 problems
a. Most collectors have to special case things across cloud versions and on-premise making maintenance more and more difficult.
b. Changing any core collectors has a big testing matrix today and results in bugs.
c. Configuration for cloud is only and on or off switch and all it does is now runs these cloud specific collectors on every server in the config file but by default still runs all the collectors for on-premise.

Some examples:
a. On azure SQL DB today I may want to collect sys.dm_os_wait_stats and sys.dm_db_wait_stats, today they are clubbed together.
b.  On-premise any of the resource governance collection is meaningless

The challenge is a bit of backward compabitibility. Would be much easier of we could deprecate V1 queries support in a shorter window to reduce that code which is expensive to run and not comprehensive.

denzilribeiro:
### Required for all PRs:

This is a big PR that addresses the issues in #7713

For example  I just realized that DatabaseIO was broken on Managed instances with the prior fix for on-prem stuff.
Fundamental principals
a. Split out queries by each database type - that simplifies queries, bug in a change will only affect a subset of input target platforms.
b. Maintain the same measures if the underlying core DMV is the same.
c. Reduces number of queries that are executed against each target or have ability to reduce them.
d. Ability to change the "default queries" per target.

Resolves #7713


denzilribeiro:
@Trovalo  , @danielnelson  can you both review this please given a big PR - still may make a small change or 2 but took a stab at this. 
Also not sure about the failure as has to do with : influxdata/telegraf/plugins/inputs/tcp_listener.TestRunParserInvalidMsg()

denzilribeiro:
There may not be a conflict with #7842 , as that is one of the few queries I haven't changed 
@ssoroka  / @reimda  the failed check doesn't seem to be related to my PR - could you help resolve
@danielnelson  - Good luck in your new job.

plugins/inputs/sqlserver/README.md
```diff
@@ -46,6 +46,11 @@ GO
   #  "Server=192.168.1.10;Port=1433;User Id=<user>;Password=<pw>;app name=telegraf;log=1;",
   # ]
 
+  ## This enables a specific set of queries depending on the database type. If specified, it replaces azuredb = true/false and query_version = 2
```
Trovalo, line 49: I would point out that "query_version" is now outdated/deprecated and "database_type" should be used instead.
I'll also edit its description in order to match version 2 current state.

```md
  ## The "query_version" parameter is deprecated, use the new "database_type" parameter.
  ## maintained for retro-compatibility reasons,version 1 is no longer mantained, therefore it's suggested to use version 2
  ## version 2 supports SQL Server from SQL 2008 SP3 and SQL 2008 R2, and the cloud version of SQL Server AzureSQLDB and Managed Instance
  query_version = 2
```

plugins/inputs/sqlserver/README.md
```diff
@@ -56,6 +61,27 @@ GO
   # azuredb = false
```
Trovalo, line 61: add parameter outdated/deprecated info
```md
  ## The "azuredb" parameter is deprecated, use the new "database_type" parameter.
  ## If you are using AzureDB, setting this to true will gather resource utilization metrics
  # azuredb = false
```

plugins/inputs/sqlserver/README.md
```diff
@@ -56,6 +61,27 @@ GO
...
+  ## - AzureDBWaitStats
+  ## - AzureDBResourceStats 
+  ## - AzureDBResourceGovernance
+  ## - AzureDBDatabaseIO
+  ## - AzureDBServerProperties
+  ## - AzureDBSQLOsWaitstats
+  ## - AzureDBMemoryClerks
+  ## - AzureDBPerformanceCounters
+
+   ## database_type =  AzureSQLManagedInstance
```
Trovalo, line 76: there is an extra space in the query list

plugins/inputs/sqlserver/README.md
```diff
@@ -6,7 +6,7 @@ lightweight and use Dynamic Management Views supplied by SQL Server.
```
Trovalo, line 6: We might add a section which lists the minimum requirements or supported editions for the plugin, something like:
```md
### THe SQL Server plugin supports the following editions/versions of SQL Server
- SQL Server
  - 2008 SP3 (with CU3)
  - SQL Server 2008 R2 SP2
  - any newer version
- Azure SQL DB
- Azure Managed Instance
```

plugins/inputs/sqlserver/README.md
```diff
@@ -46,6 +46,11 @@ GO
   #  "Server=192.168.1.10;Port=1433;User Id=<user>;Password=<pw>;app name=telegraf;log=1;",
   # ]
 
+  ## This enables a specific set of queries depending on the database type. If specified, it replaces azuredb = true/false and query_version = 2
+  ## In the config file, the sql server plugin section should be repeated multiple times each with a set of servers for a specific database_type
+  ## possible values for database_type are  "AzureSQLDB" or  "SQLServer" or "AzureSQLManagedInstance"
```
Trovalo, line 51: I just think a list is easier to spot and read
```
## possibles values are:
## - "SQLServer"
## - "AzureSQLDB"
## - "AzureSQLManagedInstance"
```

plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -2,6 +2,8 @@ package sqlserver
```
Trovalo, line 2: All the SQL Related queries have been moved in their own file, what about putting the "standard" edition SQL Queries in their own file too?
I think this will be even cleaner as we won't have queries in the main GO file

plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -46,6 +50,12 @@ const sampleConfig = `
 #  "Server=192.168.1.10;Port=1433;User Id=<user>;Password=<pw>;app name=telegraf;log=1;",
 # ]
 
+## This enables a specific set of queries depending on the database type. It replaces azuredb = true/false.
+## In the config file, the sql server plugin section can be repeated multiple times each with a set of servers for a specific database_type
+## possible values for database_type are  "AzureSQLDB" or  "SQLServer" or "AzureSQLManagedInstance"
```
Trovalo, line 55: as in the docs, a list is easier to read and identify
```
## possible values are
## - SQLServer
## - AzureSQLDB
## - AzureSQLManagedInstance
```

plugins/inputs/sqlserver/sqlserver.go
denzilribeiro: Yeah I was thinking about doing that ran out of time, will do that in next commit, will split V2 in own file and clean up queries to just have SQLServer queries in own file .

denzilribeiro:
@Trovalo  if you want me to just include the PR #7842  in this too let me know will be small change maybe easier to do.

Trovalo:
About open PRs which might need to be handled before or after this one, those are the one I've found
#7808 - additional field in perf-counters
#7842 - new query "query_stats"
#7834 - edit of existing query "sql_requests"

The first 2 are mine, so I've no problems with merging them later, personally I think there won't be any problem even with the last one since the query is the same for all the versions

plugins/inputs/sqlserver/sqlserver.go
denzilribeiro: Done

plugins/inputs/sqlserver/sqlserver.go
denzilribeiro: Split V2 in latest commit - Will do SQLServer as separate file later.

plugins/inputs/sqlserver/README.md
denzilribeiro: Done

plugins/inputs/sqlserver/README.md
denzilribeiro: Done

plugins/inputs/sqlserver/README.md
denzilribeiro: Changed

plugins/inputs/sqlserver/README.md
denzilribeiro: Done

plugins/inputs/sqlserver/README.md
denzilribeiro: Done, also added deprecation notice in log file.

Trovalo:
> @Trovalo if you want me to just include the PR #7842 in this too let me know will be small change maybe easier to do.

That's fine for me. If you do so just tell me so I can close my PR

plugins/inputs/sqlserver/README.md
```diff
@@ -82,7 +91,8 @@ GO
    ## - AzureMIMemoryClerks
    ## - AzureMIPerformanceCounters
 
-  ## Version 2:
+  ## Version 2 by default collects the following queries
+  ## Version 1 is being deprecated, please consider using database_type.
```
Trovalo, line 95: just a small error
## Version 1 → ## Version 2

plugins/inputs/sqlserver/README.md
Trovalo: I don't any comment about the deprecation of "azuredb" parameter (at least on the parameter itself)

plugins/inputs/sqlserver/README.md
```diff
@@ -47,23 +53,26 @@ GO
...
+  ## "AzureSQLDB" 
+  ## "SQLServer"
+  ## "AzureSQLManagedInstance"
+  # database_type = "AzureSQLDB"
 
   ## Optional parameter, setting this to 2 will use a new version
-  ## of the collection queries that break compatibility with the original
-  ## dashboards.
+  ## of the collection queries that break compatibility with the original dashboards.
   ## Version 2 - is compatible from SQL Server 2012 and later versions and also for SQL Azure DB
```
Trovalo, line 65: "from SQL Server 2012 and" → "from SQL Server 2008 SP3 and"

plugins/inputs/sqlserver/README.md
```diff
@@ -1,9 +1,15 @@
 # SQL Server Input Plugin
-
 The `sqlserver` plugin provides metrics for your SQL Server instance. It
 currently works with SQL Server 2008 SP3 and newer. Recorded metrics are
 lightweight and use Dynamic Management Views supplied by SQL Server.
 
+### THe SQL Server plugin supports the following editions/versions of SQL Server
```
Trovalo, line 6: there is an extra capital H "THe" → "The"

denzilribeiro:
Had to rebase, due to other commits, sorry about that.

denzilribeiro:
@ssoroka  - any idea why the test was failing? 

--- FAIL: TestAlignedTickerJitter (0.01s)
    tick_test.go:64: 
        	Error Trace:	tick_test.go:64
        	Error:      	Should be true
        	Test:       	TestAlignedTickerJitter

denzilribeiro:
@Trovalo  I did take care of ensuring these are included. the query_stats one can revisit after merge.
#7808 
#7834
#7724

ssoroka:
@denzilribeiro Sorry, was just a flakey test.

denzilribeiro:
@Trovalo  any other feedback?
@ssoroka -- can we merge this?

ssoroka:
@Trovalo, did you get a chance to run your tests on this branch? happy?

Trovalo:
I'm on vacation and I won't be able to run any test until this weekend.
I'll let you know the result as soon as I run them

plugins/inputs/sqlserver/sqlserverqueries.go
```diff
@@ -0,0 +1,1133 @@
...
+const sqlServerRequests string = `
+SET NOCOUNT ON; 
+SELECT  blocking_session_id into #blockingSessions FROM sys.dm_exec_requests WHERE blocking_session_id != 0
+create index ix_blockingSessions_1 on #blockingSessions (blocking_session_id)
+SELECT	
+'sqlserver_requests' AS [measurement]
+, REPLACE(@@SERVERNAME,'\',':') AS [sql_instance]
+, s.session_id
+, ISNULL(r.request_id,0) as request_id
+, DB_NAME(s.database_id) as session_db_name
```
Trovalo, line 1027: For 2008 and 2008R2 compatibility get col from dm_exec_requests
```SQL
, DB_NAME(r.database_id) as session_db_name
```

plugins/inputs/sqlserver/sqlserverqueries.go
```diff
@@ -0,0 +1,1133 @@
...
+, COALESCE(r.writes,s.writes) AS writes
+, r.command
+, r.wait_time as wait_time_ms
+, r.wait_type
+, r.wait_resource
+, r.blocking_session_id
+, s.program_name
+, s.host_name
+, s.nt_user_name
+, s.open_transaction_count  AS open_transaction
```
Trovalo, line 1041: For 2008 and 2008R2 compatibility the same can be obtained with
```SQL
,open_transaction =
        CASE
            WHEN r.session_id IS NOT NULL THEN r.open_transaction_count
            ELSE (SELECT COUNT(*) FROM sys.dm_tran_session_transactions AS t WHERE t.session_id = s.session_id)
        END
```
Thanks @spaghettidba

Trovalo (changes requested):
I run all the tests for SQL Server 2008 to 2019.
I've just found 2 errors in the _sqlServerRequests_ query, for compatibility reasons (with 2008 and 2008R2) two columns should be fetched in a different way.

All the other queries are working properly

plugins/inputs/sqlserver/sqlserverqueries.go
denzilribeiro: Will do, this was a change I picked up from https://github.com/influxdata/telegraf/pull/7834/commits/45a83b9ab487b3ac5bfafc24ce30257bce91fa75  and can revert it.

plugins/inputs/sqlserver/sqlserverqueries.go
denzilribeiro: This wasn't something I changed, haven't committed so far, can discuss under different PR. Having a count(*) from open transactions  for every request is a cheap operation ( think of 5000 active requests). This results in a nested loop join executed for every transaction. SQL 2008/R2 is past extended support from a SQL Server perspective, at some point trying to get parity for everything new isn't going to be possible and will complicate existing queries.

Trovalo: This has been changed in the 45a83b9, the priority has been given to the sessions table (s) before thew column was picked from the request table (r). (which to me still makes sense tbh, since the table is detailed "by request" and not by session. 
Probably the best solution is to use ```COALESCE(r.open_transaction_count, s.open_transaction_count) AS [open_transaction_count]```, by doing so we get the data per-request, but if there is no request then it makes sense to fetch it from the session table.

We can revert it back to ```r.___``` and get the data from the requests, or I can split the query in 2 versions using dynamic SQL in order to manage this better.

I'm aware of the fact that SQL 2008/R2 is past extended support, and I hope to don't see it around anymore.... but for now it's still around...
So I'd like to keep the full compatibility at least for the V2 of the queries, but I agree on the fact that a certain point the support for it should be stopped

denzilribeiro: These are totally 2 different discussions
a. I get the Requests v/s Sessions argument for open_transactions, that is a trivial change and shouldn't be different or if ti is the session count should override anyways as tran is still open
b. what I was commenting on is the fact that open_transactions isn't there before 2012 and hence the select to get the count from dm_tran_session_transactions is more expensive.. Either way this wasn't a change

I am not sure where you are seeing it changed in https://github.com/influxdata/telegraf/commit/45a83b9ab487b3ac5bfafc24ce30257bce91fa75 I just took a look a the line says   
, s.open_transaction_count  AS open_transaction

denzilribeiro:
@ssoroka are we good to merge this?
@Trovalo  any other concerns?

plugins/inputs/sqlserver/sqlserverqueries.go
Trovalo: In this case I would point to solution "a", so just get the column form requests instead of sessions (for compatibility reasons, otherwise the query explodes on SQL 2008). 

Just abort the idea of the subquery on "dm_tran_session_transactions" (it was not a great idea), as you said it's heavier and it's not worth running it just to keep compatibility with SQL 2008 and R2

Trovalo:
> @ssoroka are we good to merge this?
> @Trovalo any other concerns?

Except for the open point  about the column ```s."open_transaction_count"``` (sessions) vs ```r."open_transaction_count``` (requests) we are good to go. 
(if we get the column form requests the query will run on SQL 2008 and R2, otherwise, it will break)

plugins/inputs/sqlserver/sqlserverqueries.go
denzilribeiro: I ended up changing to dynamic SQL  to get appropriate column. Changing to r.open_transactions means that even for versions after 2008R2 you will not get the open tran count for an Idle session ( think of a session that has an open transaction and no request).

denzilribeiro:
> > @ssoroka are we good to merge this?
> > @Trovalo any other concerns?
> 
> Except for the open point about the column `s."open_transaction_count"` (sessions) vs `r."open_transaction_count` (requests) we are good to go.
> (if we get the column form requests the query will run on SQL 2008 and R2, otherwise, it will break)

Changes done with Dynamic SQL

plugins/inputs/sqlserver/sqlqueriesV2.go
```diff
@@ -1222,16 +1222,28 @@ ELSE
...
-'sqlserver_requests' AS [measurement]
-, REPLACE(@@SERVERNAME,'\',':') AS [sql_instance]
+SELECT 
+''sqlserver_requests'' AS [measurement]
+, REPLACE(@@SERVERNAME,''\'','':'') AS [sql_instance]
 , DB_NAME() as [database_name]
 , s.session_id
 , ISNULL(r.request_id,0) as request_id
-, DB_NAME(r.database_id) as session_db_name
+, DB_NAME(s.database_id) as session_db_name
```
Trovalo, line 1246: s.database_id is not available in SQL 2008 and R2, it should be r.database_id

But since this is dynamic SQL now you can just pass it in the column parameter
```TSQL
IF @MajorMinorVersion >= 1200  
	SET @Columns = ',s.open_transaction_count as open_transaction
	,DB_NAME(s.database_id) as session_db_name'
ELSE
	SET @Columns = ',r.open_transaction_count as open_transaction
	,DB_NAME(r.database_id) as session_db_name'
```

after this the PR is approved to me

Trovalo (changes requested):
One last fix is needed.

Then this is approved to me.

plugins/inputs/sqlserver/sqlqueriesV2.go
denzilribeiro: Dang it sorry  moving target should install a 2008/R2 instance at some point. should be done now.

Trovalo (approved):

PR #7944, opened 2020-08-05
Support kafka avro and fetching schema from registry server

johntdyer:
We use Avro schemas for all data transported in Kafka.  This PR supports fetching and validating against a schema prior to sending data into Kafka

 ### Required for all PRs:


ssoroka:
This is really interesting, but it doesn't quite look ready yet. I'm curious why you chose to copy the kafka plugins rather than extending them with new config options. 

It also looks like you're intended to drop all non-conforming metrics? seems like we might want other options for production readiness? drop/pass/quarantine ?

also curious about other processes surrounding this. How do you manage the schemas? other tooling, etc.

ssoroka:
Thinking about this a bit more. If I understand this correctly, you're both doing schema enforcement and using the schema for serializing/deserializing the payload. maybe it makes more sense for these to be implemented as parser/serializers instead of their own custom input/outputs.

johntdyer:
> This is really interesting, but it doesn't quite look ready yet. I'm curious why you chose to copy the kafka plugins rather than extending them with new config options.
> 
> It also looks like you're intended to drop all non-conforming metrics? seems like we might want other options for production readiness? drop/pass/quarantine ?
> 
> also curious about other processes surrounding this. How do you manage the schemas? other tooling, etc.

We use confluents schema registry 

https://docs.confluent.io/current/schema-registry/index.html

internal/avro/avro.go
```diff
@@ -0,0 +1,36 @@
...
+	// Schema is the Avro schema string.
+	Schema string `json:"schema"`
+	// Subject where the schema is registered for.
+	Subject string `json:"subject"`
+	// Version of the returned schema.
+	Version int `json:"version"`
+	ID      int `json:"id,omitempty"`
+}
+
+func GetSchema(avroRegistry string) (string, int) {
```
jsteenb2, line 21: we'll likely want a way to dependency inject the http client here. Things like TLS support and the like seem pertinent.

karrick:
I wrote https://github.com/linkedin/goavro. This effort sounds interesting and I had started working on an update to `goavro` to provide this sort of codec and schema support out of the box. I had done something similar to this for use at LinkedIn, but the new codec branch was meant to make it more general for everyone to use.

If there is interest in that effort, maybe I could prioritize getting my codec work merged into goavro master so we can simply this particular merge.

ssoroka:
Sounds interesting, @karrick. I would be interested in chatting more about it on our [community slack](https://www.influxdata.com/slack), if you're so inclined.

rodrigobrochado:
Hi guys,
Have you looked in this PR https://github.com/influxdata/telegraf/pull/7732 that already have an Avro parser plugin implementation in progress? We are testing it in a PoC with some modifications.

sjwang90:
Please let any comments for preferences of using this as an input plugin or as a parser in @rodrigobrochado's PR #7732

ssoroka:
closing in favor of something like https://github.com/influxdata/telegraf/pull/7732

PR #7960, opened 2020-08-10
Remedy malformed data in http post request

dwelch-r7:
### Required for all PRs:

This PR fixes an issue I noticed with the http input plugin, specifically when sending a POST request with a body it would be malformed as seen here:
image

These changes correctly send the request body shown here:
image

I borrowed the code from here https://github.com/influxdata/telegraf/blob/c73ed8ca6c4a81b70d1e1aeb80f6d277bdde5b77/plugins/inputs/http_response/http_response.go#L258 after I did some digging around and noticed this plugin seemed to send the data correctly

### DIFF
--- a/plugins/inputs/http/http.go
+++ b/plugins/inputs/http/http.go
@@ -156,11 +156,6 @@ func (h *HTTP) gatherURL(
 	acc telegraf.Accumulator,
 	url string,
 ) error {
-	//body, err := makeRequestBodyReader(h.ContentEncoding, h.Body)
-	//if err != nil {
-	//	return err
-	//}
-	//defer body.Close()
 
 	var body io.Reader
 	if h.Body != "" {
@@ -168,7 +163,6 @@ func (h *HTTP) gatherURL(
 	}
 	request, err := http.NewRequest(h.Method, url, body)
 
-	//request, err := http.NewRequest(h.Method, url, body)
 	if err != nil {
 		return err
 	}
--- a/plugins/inputs/http/http.go
+++ b/plugins/inputs/http/http.go
@@ -157,10 +157,11 @@ func (h *HTTP) gatherURL(
 	url string,
 ) error {
 
-	var body io.Reader
-	if h.Body != "" {
-		body = strings.NewReader(h.Body)
+	body, err := makeRequestBodyReader(h.ContentEncoding, h.Body)
+	if err != nil {
+		return err
 	}
+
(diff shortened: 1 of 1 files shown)


dwelch-r7:
@ssoroka yea that was my bad, didn't check things over properly, I'm fairly new to go and to be honest I don't _fully_ understand why these changes seem to fix it, happy to work with y'all on this if I need to do more

ssoroka:
No worries. So at this point your change only removes a body close, which seems like it's not related to the problem.

I'd be curious as to what is causing the issue, and I'd like to see a failing test to verify the failure and that it's resolved. We also need to resolve it without removing the _defer Close_. Also are we sure this isn't part of one of the many http specs? What started you on this quest? :)

dwelch-r7:
I can work on that test for sure
yea I found it super weird that doing that fixed the issue I was having...I'd love to know why/what the proper fix should be

I was graphing out some performance metrics, we had a JSON-RPC endpoint that we wanted to ping and it needed some data POST-ed to it, that's when I noticed the issues and I shopped around a few of the other plugins to see if any did a similar thing, spotted the http_response one I linked above which seemed to work so I just copy/pasted that to get the job done, I figured it might be worth pushing it back upstream as well

dwelch-r7:
oh hold on, on the one that's broken for me the `Transfer-Encoding` header is set to `chunked` but that goes away whenever I remove `body.Close()`

dwelch-r7:
Ok this looks like it was a problem on our end, apparently we weren't handling chunked encoding, my bad!
thanks for the help @ssoroka

dwelch-r7:
Incase anyone else comes across the same issue I did:
```
// If Body is present, Content-Length is <= 0 and TransferEncoding
// hasn't been set to "identity", Write adds "Transfer-Encoding:
// chunked" to the header. Body is closed after it is sent.
```
This is what causes the Transfer-Encoding: to be set to chunked, when you have a body and the Content-Length is set to <= 0

Now why the Content-Length isn't being calculated
```
// If body is of type *bytes.Buffer, *bytes.Reader, or
// *strings.Reader, the returned request's ContentLength is set to its
// exact value (instead of -1), GetBody is populated (so 307 and 308
// redirects can replay the body), and Body is set to NoBody if the
// ContentLength is 0.
```
This is why when I removed the code wrapping the Reader in a NopCloser it started working, because now it was able to calculate the length, and since the length was > 0 it didn't fall back to using a chunked Transfer-Encoding

PR #7971, opened 2020-08-11
Removed outdated information about Windows support

bencarlton:
Newer versions of Telegraf for Windows do not appear to support using WMI query patterns as values for the "pattern" option. Regular expressions appear to be the only patterns now supported.

### Required for all PRs:


ssoroka:
I wonder if we unintentionally broke support for this at some point...

ssoroka (approved):

PR #7987, opened 2020-08-14
update godirwalk to v1.16.1 (dragonfly support)

karrick:
1. Most importantly, this version runs on Dragonfly BSD, which might
   be an issue that potential users would care about.

2. Re-issues syscall request after EINTR.

   More resiliant on Go v1.14 and above, where syscall is more likely
   to return syscall.EINTR.

   Also, alongside the place where it would instead call
   golang.org/x/sys/unix, but it is commented out, I include the
   equivalent code for checking for unix.EINTR, to make it less likely
   that I forget to change the EINTR check if I ever convert it to use
   golang.org/x/sys/unix rather than syscall.

3. Performance improvements for unix and Windows.

### Required for all PRs:

NOTE: While updated godirwalk package now runs on Dragonfly BSD, I did not test whether the remainder of telegraf also runs on Dragonfly BSD, so I did not add Dragonfly BSD to the README.md file.

NOTE 2: Additional unit tests are in the updated godirwalk package.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -79,7 +79,7 @@ require (
 	github.com/jackc/pgx v3.6.0+incompatible
 	github.com/jcmturner/gofork v1.0.0 // indirect
 	github.com/kardianos/service v1.0.0
-	github.com/karrick/godirwalk v1.12.0
+	github.com/karrick/godirwalk v1.16.1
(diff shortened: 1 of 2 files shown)


ssoroka (approved):

ssoroka:
Thank you!

PR #7992, opened 2020-08-16
Make able statsd input plugin to push 0 to collector on the first time is the gather beign requested

angel9484:
Add option to send a 0 in the first gather for the metric. The intention of this new option is to solve things like: https://github.com/prometheus/prometheus/issues/1673
The next iterations will send the value collected.
If you use this with DeleteCounter enabled will create a series like [0,value,0,value,0,value...] 

The first idea was to use another layer to keep this behavior (like in a processor) and make able all input plugins able to do this but I saw that the effort in the processor was huge (creating a middle cache and retaining values in the first time...) so finally was done in the input plugin where we had this behavior with prometheus

### DIFF
--- a/plugins/inputs/statsd/README.md
+++ b/plugins/inputs/statsd/README.md
@@ -34,6 +34,15 @@
   ## Reset timings & histograms every interval (default=true)
   delete_timings = true
 
+  ## When resetting counter or the telegraf daemon, first gather will put a 0
+  ## This is useful for metric collectors systems like prometheus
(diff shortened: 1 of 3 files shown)


ssoroka:
how would Telegraf know enough to know whether or not this is the "first time" it has seen a metric? This sounds very problematic. Doesn't statsd already correctly handle increments from unknown metrics?

srebhan:
@angel9484 can you please comment on Steven's question!?

sspaink:

PR #7995, opened 2020-08-17
feat: Add Stomp (Active MQ) output plugin

amus-sal:
### Required for all PRs:


plugins/outputs/stomp/stomp.go
```diff
@@ -0,0 +1,95 @@
...
+		q.Conn, err = ts.Dial("tcp", q.Host, tlsConfig)
+
+		q.Stomp, err = stomp.Connect(q.Conn, stomp.ConnOpt.HeartBeat(0, 0), stomp.ConnOpt.Login(q.Username, q.Password))
+	} else {
+		q.NetConn, err = net.Dial("tcp", q.Host)
+		q.Stomp, err = stomp.Connect(q.NetConn, stomp.ConnOpt.HeartBeat(0, 0), stomp.ConnOpt.Login(q.Username, q.Password))
+
+	}
+	if err != nil {
+		println("cannot connect to server", err.Error())
```
ssoroka, line 43: add `	Log telegraf.Logger `toml:"-"`` to the STOMP struct instead, and Telegraf will give you a logger instance to use. in tests, set it to `testutil.Logger{}`

plugins/outputs/stomp/stomp.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/influxdata/telegraf/plugins/serializers"
+)
+
+//STOMP ...
+type STOMP struct {
+	Host      string `toml:"host"`
+	Username  string `toml:"username,omitempty"`
+	Password  string `toml:"password,omitempty"`
+	QueueName string `toml:"queueName"`
+	SSL       bool   `toml:"ssl"`
```
ssoroka, line 20: might not need this? could determine if it's set from the certs and insecure_skip_verify

plugins/outputs/stomp/stomp.go
```diff
@@ -0,0 +1,95 @@
...
+	NetConn   net.Conn
+	Stomp     *stomp.Conn
+	serialize serializers.Serializer
+}
+
+//Connect ...
+func (q *STOMP) Connect() error {
+	var err error
+	if q.SSL == true {
+		tlsConfig, _ := q.ClientConfig.TLSConfig()
```
ssoroka, line 32: hidden error?

plugins/outputs/stomp/stomp.go
```diff
@@ -0,0 +1,95 @@
...
+
+		q.Conn, err = ts.Dial("tcp", q.Host, tlsConfig)
+
+		q.Stomp, err = stomp.Connect(q.Conn, stomp.ConnOpt.HeartBeat(0, 0), stomp.ConnOpt.Login(q.Username, q.Password))
+	} else {
+		q.NetConn, err = net.Dial("tcp", q.Host)
```
ssoroka, line 38: Lots of ignored errors in this code. Gotta catch them all. :D

plugins/outputs/stomp/stomp.go
```diff
@@ -0,0 +1,95 @@
...
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+//Close ...
+func (q *STOMP) Close() error {
+	println("Closiong is starting .....")
```
ssoroka, line 78: Confusing message. Can probably just remove

plugins/outputs/stomp/stomp.go
```diff
@@ -0,0 +1,95 @@
...
+	q.Stomp.Disconnect()
+	q.Conn.Close()
+	return nil
+}
+
+//SampleConfig ...
+func (q *STOMP) SampleConfig() string {
+	return `ok = true`
```
ssoroka, line 86: needs to return the text config from the readme

plugins/outputs/stomp/stomp_test.go
```diff
@@ -0,0 +1,35 @@
...
+		QueueName: "test_queue",
+		SSL:       false,
+		serialize: s,
+	}
+	err = st.Connect()
+	require.NoError(t, err)
+
+	err = st.Write(testutil.MockMetrics())
+	require.NoError(t, err)
+}
```
ssoroka, line 35: Needs more tests. In short mode you need to mock a server and test the client does what you expect for message delivery.

plugins/outputs/stomp/stomp.go
amus-sal: If you ask about ssl field , it is for if i use tls connection or the tcp connection

plugins/outputs/stomp/stomp.go
```diff
@@ -23,32 +23,38 @@ type STOMP struct {
...
+	Log       telegraf.Logger `toml:"-"`
 }
 
 //Connect ...
 func (q *STOMP) Connect() error {
-	var err error
 	if q.SSL == true {
-		tlsConfig, _ := q.ClientConfig.TLSConfig()
-
+		tlsConfig, err := q.ClientConfig.TLSConfig()
```
ssoroka, line 32: missing ssl settings to configure certificates.
You copied the settings for this into the readme, but didn't implement these:
```
```

ssoroka (changes requested):
~Just needs work for certificate support~

ssoroka:
Readme format indentation doesn't match existing examples. Still needs more tests.

amus-sal:
> ust needs work for certificate support

I need more info, i think I follow  the implementation of the plugin "mqtt"  in correct way .

amus-sal:
> minimum one green-path test that runs as -short)

for the test case ,  I have already added the min test case with green path.

plugins/outputs/stomp/stomp.go
```diff
@@ -0,0 +1,136 @@
...
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+//SampleConfig ...
+func (q *STOMP) SampleConfig() string {
+	return `
+[[outputs.STOMP]]
```
ssoroka, line 99:
```suggestion
```
heading isn't used in SampleConfig as it's generated automatically.

plugins/outputs/stomp/stomp.go
```diff
@@ -0,0 +1,136 @@
...
+    	# tls_ca = "/etc/telegraf/ca.pem"
+    	# tls_cert = "/etc/telegraf/cert.pem"
+    	# tls_key = "/etc/telegraf/key.pem"
+    ## Use TLS but skip chain & host verification
+    	# insecure_skip_verify = false
+
+
+
+	## Data format to output.
+    	# data_format = "json"
```
ssoroka, line 126:
```suggestion
  ## Host of Active Mq broker
  host = "localhost:61613"

  ## Queue name for producer messages
  queueName = "telegraf"

  ## Optional username and password if Required to connect Active MQ server.
  username = ""
  password = ""

  ## Default No TLS Connecton 
  # SSL = false

  ## Optional TLS Config
  # SSL = true
  # tls_ca = "/etc/telegraf/ca.pem"
  # tls_cert = "/etc/telegraf/cert.pem"
  # tls_key = "/etc/telegraf/key.pem"
  ## Use TLS but skip chain & host verification
  # insecure_skip_verify = false


  ## Data format to output.
  # data_format = "json"
```

ssoroka: same for readme (except readme has header)

plugins/outputs/stomp/stomp_test.go
ssoroka: still looking for a short test.

ssoroka (changes requested):

ssoroka:
Your test case is skipped when `testing.Short()` is true. This means most of the time zero test cases will run as part of the automated test suite, and you can't be sure that future changes won't break things.

AShabana:
Hi all , Did anyone managed to solve the conflict  ?

docs/LICENSE_OF_DEPENDENCIES.md
```diff
@@ -91,7 +91,9 @@ following works:
 - github.com/go-ping/ping [MIT License](https://github.com/go-ping/ping/blob/master/LICENSE)
 - github.com/go-redis/redis [BSD 2-Clause "Simplified" License](https://github.com/go-redis/redis/blob/master/LICENSE)
 - github.com/go-sql-driver/mysql [Mozilla Public License 2.0](https://github.com/go-sql-driver/mysql/blob/master/LICENSE)
-- github.com/go-stack/stack [MIT License](https://github.com/go-stack/stack/blob/master/LICENSE.md)
+- github.com/go-stomp/stomp [Apache-2.0 License](https://github.com/go-stomp/stomp/blob/master/LICENSE.txt)
+- github.com/goburrow/modbus [BSD 3-Clause "New" or "Revised" License](https://github.com/goburrow/modbus/blob/master/LICENSE)
+- github.com/goburrow/serial [MIT License](https://github.com/goburrow/serial/LICENSE)
```
srebhan, line 96: Where does this come from?

go.mod
```diff
@@ -101,6 +101,13 @@ require (
 	github.com/fatih/color v1.10.0
 	github.com/form3tech-oss/jwt-go v3.2.5+incompatible // indirect
 	github.com/ghodss/yaml v1.0.1-0.20190212211648-25d852aebe32
+	github.com/glinton/ping v0.1.4-0.20200311211934-5ac87da8cd96
+	github.com/go-logfmt/logfmt v0.4.0
+	github.com/go-ole/go-ole v1.2.1 // indirect
+	github.com/go-redis/redis v6.12.0+incompatible
+	github.com/go-sql-driver/mysql v1.5.0
+	github.com/go-stomp/stomp v2.0.6+incompatible
+	github.com/goburrow/modbus v0.1.0
```
srebhan, line 110: There seem to be quite some unrelated libraries. Did you run a `make tidy`? Do you really need those?

plugins/outputs/stomp/README.md
```diff
@@ -0,0 +1,39 @@
+# STOMP Producer Output Plugin
+
+This plugin writes to a [Active MQ Broker](http://activemq.apache.org/) Applying STOMP Protocol.
+
+It also support Amazon MQ  (https://aws.amazon.com/amazon-mq/)
```
srebhan, line 5:
```suggestion
This plugin sends metrics to an [Active MQ Broker](http://activemq.apache.org/) applying the STOMP protocol.
It also support Amazon MQ  (https://aws.amazon.com/amazon-mq/)
```
Do you have a reference for the `STOMP` protocol? It would be nice to link to it here...

plugins/outputs/stomp/README.md
```diff
@@ -0,0 +1,39 @@
...
+    # tls_ca = "/etc/telegraf/ca.pem"
+    # tls_cert = "/etc/telegraf/cert.pem"
+    # tls_key = "/etc/telegraf/key.pem"
+    ## Use TLS but skip chain & host verification
+    # insecure_skip_verify = false
+
+
+
+  ## Data format to output.
+    # data_format = "json"
```
srebhan, line 37: Seems like you mix spaces and tabs here. Please use two leading spaces for the config.

plugins/outputs/stomp/stomp.go
```diff
@@ -0,0 +1,135 @@
...
+
+import (
+	ts "crypto/tls"
+	"net"
+
+	"github.com/go-stomp/stomp"
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/plugins/common/tls"
+	"github.com/influxdata/telegraf/plugins/outputs"
+	"github.com/influxdata/telegraf/plugins/serializers"
```
srebhan, line 11: Maybe using this aliases would be better
```suggestion
	"crypto/tls"
	"net"

	"github.com/go-stomp/stomp"
	
	"github.com/influxdata/telegraf"
	commontls "github.com/influxdata/telegraf/plugins/common/tls"
	"github.com/influxdata/telegraf/plugins/outputs"
	"github.com/influxdata/telegraf/plugins/serializers"
```

plugins/outputs/stomp/stomp.go
```diff
@@ -0,0 +1,135 @@
...
+)
+
+//STOMP ...
+type STOMP struct {
+	Host      string `toml:"host"`
+	Username  string `toml:"username,omitempty"`
+	Password  string `toml:"password,omitempty"`
+	QueueName string `toml:"queueName"`
+	SSL       bool   `toml:"ssl"`
```
srebhan, line 20: Please don't introduce this field. Instead check if a valid TLS/SSL configuration was given and determine if you should do SSL or not.

plugins/outputs/stomp/stomp.go
```diff
@@ -0,0 +1,135 @@
...
+	tls.ClientConfig
+	Conn      *ts.Conn
+	NetConn   net.Conn
+	Stomp     *stomp.Conn
```
srebhan, line 24: I think you do not need to export this fields.

plugins/outputs/stomp/stomp.go
```diff
@@ -0,0 +1,135 @@
...
+	Log       telegraf.Logger `toml:"-"`
+}
+
+//Connect ...
+func (q *STOMP) Connect() error {
+	if q.SSL == true {
+		tlsConfig, err := q.ClientConfig.TLSConfig()
+		if err != nil {
+			return err
+		}
```
srebhan, line 35:
```suggestion
	tlsConfig, err := q.ClientConfig.TLSConfig()
	if err != nil {
		return err
	}
	
	if tlsConfig != nil {
```

plugins/outputs/stomp/stomp.go
```diff
@@ -0,0 +1,135 @@
...
+	for _, metric := range metrics {
+		values, err := q.serialize.Serialize(metric)
+		if err != nil {
+			return err
+		}
+		err = q.Stomp.Send(q.QueueName, "text/plain",
+			[]byte(values), nil)
+		if err != nil {
+			return err
+		}
```
srebhan, line 77:
```suggestion
		if err := q.Stomp.Send(q.QueueName, "text/plain", []byte(values), nil);  err != nil {
			return err
		}
```

plugins/outputs/stomp/stomp.go
```diff
@@ -0,0 +1,135 @@
...
+	var err error
+	err = q.Stomp.Disconnect()
+	if err != nil {
+		return err
+	}
+	err = q.Conn.Close()
+	if err != nil {
+		return err
+	}
+	return nil
```
srebhan, line 93:
```suggestion
	if err := q.Stomp.Disconnect(); err != nil {
		return err
	}
	return q.Conn.Close()
```

plugins/outputs/stomp/stomp.go
```diff
@@ -0,0 +1,135 @@
...
+    	# tls_cert = "/etc/telegraf/cert.pem"
+    	# tls_key = "/etc/telegraf/key.pem"
+    ## Use TLS but skip chain & host verification
+    	# insecure_skip_verify = false
+
+
+
+	## Data format to output.
+    	# data_format = "json"
+`
```
srebhan, line 126: Usually people define a `const sampleConfig`. Anyway, your whitespaces seem to be mixed and the empty lines are nonstandard. Please prefix the config lines with two spaces to guarantee for a consistent style in the generated config
```suggestion
	return `
  ## Host of Active Mq broker
  host = "localhost:61613"

  ## Queue name for producer messages
  queueName = "telegraf"

  ## Optional credentials
  # username = ""
  # password = ""

  ## Optional TLS Config (defaults to unencrypted connection)
  # tls_ca = "/etc/telegraf/ca.pem"
  # tls_cert = "/etc/telegraf/cert.pem"
  # tls_key = "/etc/telegraf/key.pem"
 
  ## Use TLS but skip chain & host verification
  # insecure_skip_verify = false

  ## Data format to output.
  ## Each data format has it's own unique set of configuration options, read
  ## more about them here:
  ## https://github.com/influxdata/telegraf/blob/master/docs/DATA_FORMATS_OUTPUT.md
  # data_format = "json"
`
```

plugins/outputs/stomp/stomp.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+//Description ...
+func (q *STOMP) Description() string {
+	return "Telegraf Output Plugin For Stomp"
```
srebhan, line 131:
```suggestion
	return "Send metrics to ActiveMQ using the STOMP protocol"
```

plugins/outputs/stomp/stomp_test.go
```diff
@@ -0,0 +1,35 @@
+package stomp
+
+import (
+	"testing"
+
+	"github.com/influxdata/telegraf/plugins/serializers"
+	"github.com/influxdata/telegraf/testutil"
+
+	"github.com/stretchr/testify/require"
```
srebhan, line 9:
```suggestion
	"testing"

	"github.com/stretchr/testify/require"

	"github.com/influxdata/telegraf/plugins/serializers"
	"github.com/influxdata/telegraf/testutil"
```

plugins/outputs/stomp/stomp_test.go
```diff
@@ -0,0 +1,35 @@
...
+)
+
+// TestiConnectAndWrite ...
+func TestConnectAndWrite(t *testing.T) {
+	if testing.Short() {
+		t.Skip("Skipping integration test in short mode")
+	}
+
+	var url = testutil.GetLocalHost() + ":61613"
```
srebhan, line 18:
```suggestion
	url := testutil.GetLocalHost() + ":61613"
```

srebhan (commented):
Thanks @amus-sal for this plugin! I have some comments and want to ask you if there is a way to mockup the ActiveMQ side for unit-testing. It would be nice to test of metrics are sent in `short`testing mode...

srebhan:
@amus-sal any news here?

go.mod
amus-sal: done

plugins/outputs/stomp/stomp.go
amus-sal: done

plugins/outputs/stomp/stomp_test.go
amus-sal: done

plugins/outputs/stomp/stomp.go
amus-sal: done

plugins/outputs/stomp/stomp.go
amus-sal: done

plugins/outputs/stomp/stomp.go
amus-sal: done

amus-sal:
update with the required comments

amus-sal:
@srebhan  check now

plugins/outputs/stomp/README.md
```diff
@@ -0,0 +1,31 @@
...
+  ## Optional TLS Config
+    # tls_ca = "/etc/telegraf/ca.pem"
+    # tls_cert = "/etc/telegraf/cert.pem"
+    # tls_key = "/etc/telegraf/key.pem"
+    ## Use TLS but skip chain & host verification
+    # insecure_skip_verify = false
+
+  ## Data format to output.
+    # data_format = "json"
+```
```
srebhan, line 30: Please fix the formatting here and comment sensible defaults as documented [here](https://github.com/influxdata/telegraf/blob/master/docs/developers/SAMPLE_CONFIG.md#style). Please also adapt the `sampleConfig` accordingly.
```suggestion
```toml
  ## Host of Active MQ broker
  host = "localhost:61613"

  ## Queue name for producer messages
  # queueName = "telegraf"

  ## Username and password if required by the Active MQ server.
  # username = ""
  # password = ""

  ## Optional TLS Config

  ## Data format to output.
  # data_format = "json"
```
```

File: plugins/outputs/stomp/stomp.go [hunk 26]
```diff
@@ -0,0 +1,120 @@
...
+
+	"github.com/go-stomp/stomp"
+
+	"github.com/influxdata/telegraf"
+	commontls "github.com/influxdata/telegraf/plugins/common/tls"
+	"github.com/influxdata/telegraf/plugins/outputs"
+	"github.com/influxdata/telegraf/plugins/serializers"
+)
+
+//STOMP ...
```
srebhan, line 15: Please remove those stubs. Same for the other functions.

plugins/outputs/stomp/stomp.go
```diff
@@ -0,0 +1,120 @@
...
+	Host      string `toml:"host"`
+	Username  string `toml:"username,omitempty"`
+	Password  string `toml:"password,omitempty"`
+	QueueName string `toml:"queueName"`
+	commontls.ClientConfig
+	conn      *ts.Conn
+	netConn   net.Conn
+	stomp     *stomp.Conn
+	serialize serializers.Serializer
+	Log       telegraf.Logger `toml:"-"`
```
srebhan, line 26: We usually group the fields like
```suggestion
	Host      string `toml:"host"`
	Username  string `toml:"username,omitempty"`
	Password  string `toml:"password,omitempty"`
	QueueName string `toml:"queueName"`
	Log       telegraf.Logger `toml:"-"`

	commontls.ClientConfig
	
	conn      *ts.Conn
	netConn   net.Conn
	stomp     *stomp.Conn
	
	serialize serializers.Serializer
```

plugins/outputs/stomp/stomp.go
```diff
@@ -0,0 +1,120 @@
+package stomp
+
+import (
+	ts "crypto/tls"
```
srebhan, line 4: I think you should not use the alias here
```suggestion
	"crypto/tls"
```

plugins/outputs/stomp/stomp.go
```diff
@@ -0,0 +1,120 @@
...
+
+//STOMP ...
+type STOMP struct {
+	Host      string `toml:"host"`
+	Username  string `toml:"username,omitempty"`
+	Password  string `toml:"password,omitempty"`
+	QueueName string `toml:"queueName"`
+	commontls.ClientConfig
+	conn      *ts.Conn
+	netConn   net.Conn
```
srebhan, line 23: You can use a common interface (`net.Conn`) for this
```suggestion
	conn      net.Conn
```

plugins/outputs/stomp/stomp.go
```diff
@@ -0,0 +1,120 @@
...
+		q.netConn, err = net.Dial("tcp", q.Host)
+		if err != nil {
+			return err
+		}
+		q.stomp, err = stomp.Connect(q.netConn, stomp.ConnOpt.HeartBeat(0, 0), stomp.ConnOpt.Login(q.Username, q.Password))
+		if err != nil {
+			return err
+		}
+
+	}
```
srebhan, line 53: This can be simplified as `tls.Dial` returns a struct compatible with the `net.Conn` interface. So we can do
```suggestion
	var err error
	if tlsConfig != nil {
		q.conn, err = ts.Dial("tcp", q.Host, tlsConfig)
		if err != nil {
			return fmt.Errorf("connecting with TLS failed: %s", err)
		}
	} else {
		q.conn, err = net.Dial("tcp", q.Host)
		if err != nil {
			return fmt.Errorf("connecting without TLS failed: %s", err)
		}
	}
	q.stomp, err = stomp.Connect(q.conn, stomp.ConnOpt.HeartBeat(0, 0), stomp.ConnOpt.Login(q.Username, q.Password))
	if err != nil {
		return fmt.Errorf("connecting to STOMP failed: %s", err)
	}
```

plugins/outputs/stomp/stomp.go
```diff
@@ -0,0 +1,120 @@
...
+		values, err := q.serialize.Serialize(metric)
+		if err != nil {
+			return err
+		}
+		err = q.stomp.Send(q.QueueName, "text/plain",
+			[]byte(values), nil)
+		if err != nil {
+			return err
+		}
+	}
```
srebhan, line 76: In case there is a single metric that e.g. cannot be serialized, the whole output is stuck as the metrics are kept in the output buffer when `Write` returns an error. This means that (as the metric is kept) the `Write` function will fail until the output buffer will overflow and the malicious metric is overwritten...

I think you should log an error (`q.Log.Error()`) for the failing metric in case serialization fails and continue with the other metrics. This would drop the malicious metrics immediately.

Furthermore, you should use specific error message prefixes to ease finding the source of an error during debugging...
```suggestion
	for _, metric := range metrics {
		values, err := q.serialize.Serialize(metric)
		if err != nil {
			q.Log.Errorf("Serializing metric %v failed: %s", metric, err)
			continue
		}
		err = q.stomp.Send(q.QueueName, "text/plain", []byte(values), nil)
		if err != nil {
			return fmt.Errorf("sending metric failed: %s", err)
		}
	}
```

plugins/outputs/stomp/stomp.go
```diff
@@ -0,0 +1,120 @@
...
+	return nil
+}
+
+//Close ...
+func (q *STOMP) Close() error {
+	err := q.stomp.Disconnect()
+	if err != nil {
+		return err
+	}
+	return nil
```
srebhan, line 86:
```suggestion
	return q.stomp.Disconnect()
```

srebhan (commented):
Thanks for the update @amus-sal! I have some more comments. Please also check the linter warnings. You can check those offline using `make lint-branch` and use `markdownlint` on your `README.md` file.

amus-sal:
Thanks @srebhan,  check the updated pr  with your suggestions

plugins/outputs/stomp/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+  ## Optional TLS Config
+  # tls_ca = "/etc/telegraf/ca.pem"
+  # tls_cert = "/etc/telegraf/cert.pem"
+  # tls_key = "/etc/telegraf/key.pem"
+  ## Use TLS but skip chain & host verification
+  # insecure_skip_verify = false
+
+  ## Data format to output.
+  # data_format = "json"
```
srebhan, line 29: I think you are missing the `toml`-section closing tripple-quotes at the end of the file.

plugins/outputs/stomp/stomp.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/influxdata/telegraf"
+	commontls "github.com/influxdata/telegraf/plugins/common/tls"
+	"github.com/influxdata/telegraf/plugins/outputs"
+	"github.com/influxdata/telegraf/plugins/serializers"
+)
+
+type STOMP struct {
+	Host      string          `toml:"host"`
+	Username  string          `toml:"username,omitempty"`
+	Password  string          `toml:"password,omitempty"`
```
srebhan, line 19: Please remove the `omitempty` setting
```suggestion
	Username  string          `toml:"username"`
	Password  string          `toml:"password"`
```

plugins/outputs/stomp/stomp.go
```diff
@@ -0,0 +1,114 @@
...
+			return err
+		}
+	} else {
+		q.conn, err = net.Dial("tcp", q.Host)
+		if err != nil {
+			return err
+		}
+	}
+
+	q.stomp, err = stomp.Connect(q.conn, stomp.ConnOpt.HeartBeat(0, 0), stomp.ConnOpt.Login(q.Username, q.Password))
```
srebhan, line 48: Does it make sense to let the user configure the heart-beat parameters?

plugins/outputs/stomp/stomp.go
```diff
@@ -0,0 +1,114 @@
...
+	if err != nil {
+		return err
+	}
+	q.Log.Info("STOMP Connected...")
+	return nil
+}
+
+//SetSerializer ...
```
srebhan, line 56: Please remove the superfluous comments.

plugins/outputs/stomp/stomp.go
```diff
@@ -0,0 +1,114 @@
...
+//Write ...
+func (q *STOMP) Write(metrics []telegraf.Metric) error {
+	for _, metric := range metrics {
+		values, err := q.serialize.Serialize(metric)
+		if err != nil {
+			q.Log.Errorf("Serializing metric %v failed: %s", metric, err)
+			continue
+		}
+		err = q.stomp.Send(q.QueueName, "text/plain",
+			[]byte(values), nil)
```
srebhan, line 70:
```suggestion
		err = q.stomp.Send(q.QueueName, "text/plain", values, nil)
```

plugins/outputs/stomp/stomp.go
```diff
@@ -0,0 +1,114 @@
...
+		if err != nil {
+			return fmt.Errorf("sending metric failed: %s", err)
+		}
+	}
+	return nil
+}
+
+//Close ...
+func (q *STOMP) Close() error {
+	return q.stomp.Disconnect()
```
srebhan, line 80: Will this also close the TCP connection?

plugins/outputs/stomp/stomp.go
```diff
@@ -0,0 +1,114 @@
...
+   
+	## Use TLS but skip chain & host verification
+	# insecure_skip_verify = false
+	## Data format to output.
+	## Each data format has it's own unique set of configuration options, read
+	## more about them here:
+	## https://github.com/influxdata/telegraf/blob/master/docs/DATA_FORMATS_OUTPUT.md
+	# data_format = "json"
+`
+}
```
srebhan, line 106: The sample-config needs to start with two-spaces, therefore most plugins declare a `sampleConfig` constant at the root level to make this nice. However, you can also write
```suggestion
func (q *STOMP) SampleConfig() string {
	return `
  ## Host of Active Mq broker
  host = "localhost:61613"
  ## Queue name for producer messages
  queueName = "telegraf"
  ## Optional credentials
  # username = ""
  # password = ""
  ## Optional TLS Config (defaults to unencrypted connection)
  # tls_ca = "/etc/telegraf/ca.pem"
  # tls_cert = "/etc/telegraf/cert.pem"
  # tls_key = "/etc/telegraf/key.pem"
   
  ## Use TLS but skip chain & host verification
  # insecure_skip_verify = false
  ## Data format to output.
  ## Each data format has it's own unique set of configuration options, read
  ## more about them here:
  ## https://github.com/influxdata/telegraf/blob/master/docs/DATA_FORMATS_OUTPUT.md
  # data_format = "json"
`
}
```

Please also synchronize the settings with the README, e.g. spaces between "sections" and comment of the `queueName` option...

plugins/outputs/stomp/stomp_test.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/influxdata/telegraf/testutil"
+)
+
+// TestiConnectAndWrite ...
+func TestConnectAndWrite(t *testing.T) {
+	if testing.Short() {
+		t.Skip("Skipping integration test in short mode")
+	}
+
+	url := testutil.GetLocalHost() + ":61613"
```
srebhan, line 20: Can you maybe use a random port here to avoid conflicts with multiple tests running at the same time...

srebhan (commented):
Nice update @amus-sal! I have some more smaller comments... Please take a look.

plugins/outputs/stomp/stomp.go
amus-sal: I  need them as there are versions of  AMQ  supports integration without authentication.

plugins/outputs/stomp/stomp.go
amus-sal: yes go-stomp close the tcp connection  as well

plugins/outputs/stomp/stomp.go
amus-sal: nice idea

amus-sal:
Check now

amus-sal:
any updates ?

amus-sal:
any updates regarding my PR ?

plugins/outputs/stomp/stomp.go
```diff
@@ -0,0 +1,117 @@
...
+func (q *STOMP) Close() error {
+	return q.stomp.Disconnect()
+}
+
+func (q *STOMP) SampleConfig() string {
+	return `
+  ## Host of Active Mq broker
+  host = "localhost:61613"
+  ## Queue name for producer messages
+  queueName = "telegraf"
```
srebhan, line 88: Does it make sense to have this as a default? If so, please change to
```suggestion
  #queueName = "telegraf"
```
and set this as a default value in `Init()`.

plugins/outputs/stomp/stomp.go
srebhan: Well that's true, but in any case the value will be `nil` which means an empty string. `omitempty` only has an effect when serializing this struct which should never be the case actually.

plugins/outputs/stomp/stomp.go
```diff
@@ -0,0 +1,117 @@
...
+		if err != nil {
+			return err
+		}
+	}
+
+	q.stomp, err = stomp.Connect(q.conn, stomp.ConnOpt.HeartBeat(q.HeartBeatSend, q.HeartBeatRec), stomp.ConnOpt.Login(q.Username, q.Password))
+	if err != nil {
+		return err
+	}
+	q.Log.Info("STOMP Connected...")
```
srebhan, line 56: Can we please make this a debug message?

srebhan (commented):
Hey @amus-sal, sorry the late reply and thanks for the nice update. I have three minor comments but we are very close...

amus-sal:
any updates ?

plugins/outputs/stomp/stomp.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+func (q *STOMP) Close() error {
+	return q.stomp.Disconnect()
+}
+
+func (q *STOMP) SampleConfig() string {
+	return `
+  ## Host of Active Mq broker
+  host = "localhost:61613"
```
srebhan, line 86: As this is a default now you can change to 
```suggestion
  # host = "localhost:61613"
```

plugins/outputs/stomp/stomp.go
```diff
@@ -0,0 +1,124 @@
...
+  ## Queue name for producer messages
+  # queueName = "telegraf"
+  ## Optional credentials
+  # username = ""
+  # password = ""
+
+  ## Optional HeatBeats Configuration
+  #heartBeatSend = 0
+  #heartBeatRec = 0
```
srebhan, line 95: Maybe use
```suggestion
  # heartBeatSend = "0s"
  # heartBeatRec = "0s"
```
to make clear that you can specify a duration here.

plugins/outputs/stomp/stomp.go
```diff
@@ -0,0 +1,124 @@
...
+
+type STOMP struct {
+	Host      string          `toml:"host"`
+	Username  string          `toml:"username"`
+	Password  string          `toml:"password"`
+	QueueName string          `toml:"queueName"`
+	Log       telegraf.Logger `toml:"-"`
+
+	HeartBeatSend time.Duration `toml:"heartBeatSend"`
+	HeartBeatRec  time.Duration `toml:"heartBeatRec"`
```
srebhan, line 25: These should be `config.Duration` to allow time suffices like "s", "ms" or "m" for seconds, milliseconds, minutes etc. Furthermore, we usually use "snake-case" in TOML, so you maybe want to change to
```suggestion
	HeartBeatSend config.Duration `toml:"heartbeat_timeout_send`
	HeartBeatRec  config.Duration `toml:"heartbeat_timeout_receive"`
```

plugins/outputs/stomp/stomp.go
```diff
@@ -0,0 +1,124 @@
...
+	return "Send metrics to ActiveMQ using the STOMP protocol"
+}
+func init() {
+	outputs.Add("stomp", func() telegraf.Output {
+		return &STOMP{
+			Host:      "localhost:61613",
+			QueueName: "telegraf",
+			Username:  "",
+			Password:  "",
+		}
```
srebhan, line 122: You don't need to specify empty values as this is Golang's default
```suggestion
		return &STOMP{
			Host:      "localhost:61613",
			QueueName: "telegraf",
		}
```

srebhan (commented):
Almost @amus-sal... :-) Some minor things and I'd like you to make the `HeartBeat...` options a bit more expressive in the naming but also by allowing time suffices like milliseconds ("ms"), seconds ("s") etc.

amus-sal:
any updates ?

srebhan (commented):
Sorry @amus-sal, this somehow slipped through. Just one more sync of the option name and the sample-config and a question regarding `host`...

plugins/outputs/stomp/stomp.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func (q *STOMP) Close() error {
+	return q.stomp.Disconnect()
+}
+
+func (q *STOMP) SampleConfig() string {
+	return `
+  ## Host of Active Mq broker
+  # host = "localhost:61613"
```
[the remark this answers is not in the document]
amus-sal: yes

srebhan (approved):
Looks good to me. Thanks for your patience @amus-sal!

plugins/outputs/stomp/README.md
```diff
@@ -0,0 +1,28 @@
+# STOMP Producer Output Plugin
+
+This plugin writes to a [Active MQ Broker](http://activemq.apache.org/)/
+ for STOMP <http://stomp.github.io>.
```
powersj, line 4:
```suggestion
for STOMP <http://stomp.github.io>.
```

plugins/outputs/stomp/stomp.go
```diff
@@ -0,0 +1,123 @@
...
+  ## Each data format has it's own unique set of configuration options, read
+  ## more about them here:
+  ## https://github.com/influxdata/telegraf/blob/master/docs/DATA_FORMATS_OUTPUT.md
+  # data_format = "json"
+`
+}
+
+func (q *STOMP) Description() string {
+	return "Send metrics to ActiveMQ using the STOMP protocol"
+}
```
powersj, line 115: Since you started this PR and now, on master we have gotten rid of the Description variable across plugins. Let's remove this.
```suggestion
```

plugins/outputs/stomp/stomp.go
```diff
@@ -0,0 +1,123 @@
...
+   
+  ## Use TLS but skip chain & host verification
+  # insecure_skip_verify = false
+  ## Data format to output.
+  ## Each data format has it's own unique set of configuration options, read
+  ## more about them here:
+  ## https://github.com/influxdata/telegraf/blob/master/docs/DATA_FORMATS_OUTPUT.md
+  # data_format = "json"
+`
+}
```
powersj, line 111:
```suggestion
func (q *STOMP) SampleConfig() string {
	return sampleConfig
}
```

plugins/outputs/stomp/stomp.go
```diff
@@ -0,0 +1,123 @@
+package stomp
+
+import (
+	"crypto/tls"
```
powersj, line 4:
```suggestion
import (
    _ "embed"
	"crypto/tls"
```

plugins/outputs/stomp/stomp.go
```diff
@@ -0,0 +1,123 @@
+package stomp
```
powersj, line 1:
```suggestion
//go:generate ../../../tools/readme_config_includer/generator
package stomp
```

plugins/outputs/stomp/stomp.go
```diff
@@ -0,0 +1,123 @@
...
+	"github.com/go-stomp/stomp"
+
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/config"
+	commontls "github.com/influxdata/telegraf/plugins/common/tls"
+	"github.com/influxdata/telegraf/plugins/outputs"
+	"github.com/influxdata/telegraf/plugins/serializers"
+)
+
+type STOMP struct {
```
powersj, line 18:
```suggestion
)

// DO NOT REMOVE THE NEXT TWO LINES! This is required to embed the sampleConfig data.
//go:embed sample.conf
var sampleConfig string

type STOMP struct {
```

plugins/outputs/stomp/stomp_test.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/influxdata/telegraf/plugins/serializers"
+	"github.com/influxdata/telegraf/testutil"
+)
+
+func TestConnectAndWrite(t *testing.T) {
+	if testing.Short() {
+		t.Skip("Skipping integration test in short mode")
+	}
+
+	port := rand.Intn(7000-3000) + 3000
```
powersj, line 21: you are grabbing some random port?

Is it possible to use testcontainers, like we do with other integration tests to launch an ActiveMQ instance and write to it?

as-is are these tests even useful?

plugins/outputs/stomp/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+It also support Amazon MQ  <https://aws.amazon.com/amazon-mq/>
+
+## CONFIGURATON
+
+```toml
```
powersj, line 10:
```suggestion
```toml @sample.conf
```

powersj (commented):
Thanks for this! We have changed how we create our sample config, so I have tried my best to suggested changes to make it easy for you. However, we do still need you to create the `sample.conf` file with the sample configuration.

I also have a question about a) default port, it seems to always be documented as something different than what you have and b) ensuring tests that actually test something.
Thanks for this!

plugins/outputs/stomp/README.md
```diff
@@ -0,0 +1,28 @@
...
+  ## Host of Active MQ broker
+  host = "localhost:61613"
```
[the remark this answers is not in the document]
amus-sal: you can fine this one
https://activemq.apache.org/components/artemis/documentation/1.4.0/protocols-interoperability.html 
the part of Native Stomp support

plugins/outputs/stomp/stomp_test.go
amus-sal: I updated the test to use test containers

amus-sal:
sorry for delay, please check now

plugins/outputs/stomp/stomp_test.go
```diff
@@ -0,0 +1,55 @@
...
+		serialize:     s,
+	}
+	require.NoError(t, err)
+
+	err = st.Connect()
+	require.NoError(t, err)
+
+	err = st.Write(testutil.MockMetrics())
+	require.NoError(t, err)
+	require.NoError(t, err)
```
powersj, line 54: Are these double NoError accidental duplicate lines?

powersj (commented):
Thank you for the responses and for moving to test containers!  I think there are two items left and then this will be ready for final review.

1. I added one comment about a duplicate line
2. The various lint issues are resolved. I think if you rebase on master and clean up your imports you should be good to go

Thanks again!

amus-sal:
check now

powersj (commented):
Hi! Thanks for the quick turnaround. The diff looks a bit wrong, like a bunch of other changes were added and you still have some duplicate imports in stomp_test.go

amus-sal:
I removed the duplicates now but what do you mean by  "The diff looks a bit wrong" ?

powersj:
Take a look at https://github.com/influxdata/telegraf/pull/7995/files if shows 23 files are changed. When you are only adding two originally. It does look like those other files contain changes we added recently.

amus-sal:
> Take a look at https://github.com/influxdata/telegraf/pull/7995/files if shows 23 files are changed. When you are only adding two originally. It does look like those other files contain changes we added recently.

I did fetch upstream with the original repo and this is the output correct me if I am wrong is this right step  ?

sspaink:
@amus-sal looks like there are changes in this pull request that are unrelated to your Stomp output plugin, usually rebasing vs merging your changes avoids this I think. What I recommend doing is to save your changes, copy your Stomp output plugin into a separate directory and perhaps also copy this branch just in case. Then reset your branch to master, then copy your output plugin back into the branch and force push your changes. This should essentially be a manual rebase, let me know if that makes sense or if you need more help!

srebhan:
@amus-sal I think your rebase went out-of-control. I suggest you save your plugin code and reset your branch to master. Then add the plugin back in. Let me know if you need help!

amus-sal:
can you check it now ?

srebhan (approved):
Looks much better! Thanks for the update @amus-sal.

go.mod
```diff
@@ -406,3 +406,5 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.1 // indirect
 	sigs.k8s.io/yaml v1.2.0 // indirect
 )
+
+require github.com/go-stomp/stomp v2.1.4+incompatible // indirect
```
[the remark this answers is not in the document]
amus-sal: but indirect related to the state if using the lib inside the module but I did an update for it but still exist, I think the problems  that the plugin path "o export data for \"github.com/influxdata/telegraf/plugins/outputs/stomp\"\n\n" "is not yet  pushed to remote repo. btw I run Golangci-lint locally and it works well with both

srebhan:
@amus-sal did you run `make tidy` before pushing?

amus-sal:
yes I did and no changes found

amus-sal:
any help in that ?

srebhan:
@amus-sal I see you referenced a `sample.conf` file (the one that is embedded into your plugin code), but there is none in your PR. Did you maybe forget to check that in?

powersj:
Hi @amus-sal,

Thanks for all the work on this PR! We are nearly there.

@sspaink got tests running, and it looks like there is a missing license now in the docs/LICENSE_OF_DEPENDENCIES.md file because of pulling in go-stomp/stomp as a direct dependency:

```
@@ -108,0 +109 @@
+github.com/go-stomp/stomp [Apache License 2.0](https://github.com/go-stomp/stomp/blob/master/LICENSE.txt)
```

Can you add that package and specify it uses the Apache License 2.0 license?

Thanks!

powersj (approved):

powersj:
Thanks for all the work on this!\

Issue #7911, opened 2020-07-28
Telegraf Execd Go Shim on Windows Error "file does not exist"


ashishkaransingh:
https://github.com/influxdata/telegraf/tree/master/plugins/common/shim

I downloaded and build rand using "go build -o rand cmd/main.go"

![image](https://user-images.githubusercontent.com/29729110/88697390-1e8ef500-d122-11ea-9416-f0e25e337d75.png)

Then under telegraf.conf I used it as following:

```
[[inputs.execd]]
  command = ["C:/Users/test_user/Downloads/rand-master/rand-master/rand", "-config", "C:/Users/test_user/Downloads/rand-master/rand-master/plugin.conf"]
  signal = "none"
```

Then finally i ran it using: `telegraf.exe --config telegraf.conf -test`
But getting the following error

```
C:\Users\test_user\Downloads\telegraf-master\telegraf-master>telegraf.exe --config telegraf.conf -test
2020-07-28T16:58:38Z I! Starting Telegraf
2020-07-28T16:58:38Z I! [inputs.execd] Starting process: C:/Users/test_user/Downloads/rand-master/rand-master/rand [-config C:/Users/test_user/Downloads/rand-master/rand-master/plugin.conf]
2020-07-28T16:58:38Z E! [agent] Starting input inputs.execd: failed to start process [C:/Users/test_user/Downloads/rand-master/rand-master/rand -config C:/Users/test_user/Downloads/rand-master/rand-master/plugin.conf]: error starting process: exec: "C:/Users/test_user/Downloads/rand-master/rand-master/rand": file does not exist
```

What i am missing?

Please help!

ssoroka:
Hi, a couple thoughts:
1. In windows the file should probably be called `rand.exe`, as this is normal for windows executable files. You can just rename the rand one, or regenerate it with `go build -o rand.exe cmd/main.go`
2. I'm not sure if the slash matters or not. Do forward slashes work in place of backslahes? Or do you need to do something like `command = ["C:\\Users\\test_user\\Downloads\\rand-master\\rand-master\\rand", "-config", "C:\\Users\\test_user\\Downloads\\rand-master\\rand-master\\plugin.conf"]`?

Let me know where you're at after that.

ashishkaransingh:
> 1. rand.ex

Hi @ssoroka,

Thank you for your response, greatly appreciate it.
Sure i will try and let you know.

Thanks Again

ashishkaransingh:
[...]

PR #7996, opened 2020-08-17
fix issue with shim use of config.Duration

ssoroka:
The execd shim loads config in a slightly different way. As a result, the config.Duration wasn't being handled properly. 

This PR fixes it, and resolves issues mentioned in https://github.com/influxdata/telegraf/issues/7911

### DIFF
--- a/plugins/inputs/vsphere/endpoint.go
+++ b/plugins/inputs/vsphere/endpoint.go
@@ -887,7 +887,7 @@ func (e *Endpoint) chunkify(ctx context.Context, res *resourceKind, now time.Tim
 			// Determine time of last successful collection
 			metricName := e.getMetricNameForId(metric.CounterId)
 			if metricName == "" {
-				e.log.Info("Unable to find metric name for id %d. Skipping!", metric.CounterId)
+				e.log.Infof("Unable to find metric name for id %d. Skipping!", metric.CounterId)
 				continue
 			}
 			start, ok := e.hwMarks.Get(object.ref.Value, metricName)
--- a/plugins/processors/reverse_dns/reversedns_test.go
+++ b/plugins/processors/reverse_dns/reversedns_test.go
@@ -54,24 +54,3 @@ func TestLoadingConfig(t *testing.T) {
 
 	require.Len(t, c.Processors, 1)
 }
-
-func TestConfigDuration(t *testing.T) {
-	c := config.NewConfig()
-	err := c.LoadConfigData([]byte(`
-[[processors.reverse_dns]]
-  cache_ttl = "3h"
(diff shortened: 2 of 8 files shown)


plugins/inputs/vsphere/endpoint.go
```diff
@@ -887,7 +887,7 @@ func (e *Endpoint) chunkify(ctx context.Context, res *resourceKind, now time.Tim
 			// Determine time of last successful collection
 			metricName := e.getMetricNameForId(metric.CounterId)
 			if metricName == "" {
-				e.log.Info("Unable to find metric name for id %d. Skipping!", metric.CounterId)
+				e.log.Infof("Unable to find metric name for id %d. Skipping!", metric.CounterId)
```
reimda, line 890: This is a good change but looks unrelated.  Was it supposed to be in this pr?

plugins/processors/reverse_dns/reversedns_test.go
```diff
@@ -54,3 +54,24 @@ func TestLoadingConfig(t *testing.T) {
 
 	require.Len(t, c.Processors, 1)
 }
+
```
reimda, line 57: This test belongs in config/config_test.go instead of here.  TestConfig_LoadSpecialTypes is similar.

plugins/inputs/vsphere/endpoint.go
ssoroka: Unrelated; just don't want to make a new pr for one line. :D

plugins/processors/reverse_dns/reversedns_test.go
ssoroka: you can't have config require reverse_dns due to cyclical dependencies. Since it mentions both, it has to go here.

ssoroka: .. though, a config_test package might work.

ssoroka: Looking at this with fresh eyes, the test above it is almost identical in setup. I'm not so sure that it belongs in config. The functionality in config is covered by another test (in config_test above), and this was just an extra sanity test I added to make sure the plugin had what it needed, and wasn't falling back on default values. In the end this was not needed (because it's not a regression test per se, and didn't highlight the bug I was fixing), but I figured it didn't hurt to add the test to the test suite. I'd recommend leaving it as is.

reimda: I understand it's convenient to put it here because of dependencies, but it doesn't belong here because it's testing config parsing, not reverse dns functionality.  I'll leave it to you to move if you want.

PR #8001, opened 2020-08-18
Create external plugins doc file and update contributing guide

sjwang90:
update documentation for external plugins


sjwang90:
@mhall119 Could you take a look?

Issue #7409, opened 2020-04-24
[Wavefront Output Plugin] make internal sender buffer size configurable


lihaiswu:
## Feature Request

Make internal sender buffer size configurable in Wavefront Output Plugin

### Current behavior:
Current defaultBufferSize is 50000 in wavefront sdk https://github.com/wavefrontHQ/wavefront-sdk-go/blob/master/senders/configs.go#L5
senders/configs.go:5 . When user configure telegraf with metric_buffer_limit exceed 50000, then extra metrics will be failed to write wavefront because of buffer full returned.

### Desired behavior:
Wavefront Output Plugin should make buffer size configurable to align with telegraf metric_buffer_limit. Otherwise, it's a limitation of Wavefront Output Plugin which can only accept maximum 50000 metrics.

### Use case: 

- Set metric_buffer_limit to 250000 in telegraf.conf.

- Telegraf write 51228 metrics to Wavefront

- Only 50000 metrics are accepted, extra 1228 are failed due to "Wavefront sending error: buffer full, dropping line"

`2020-04-24T20:38:11Z D! [outputs.wavefront] Buffer fullness: 51228 / 250000 metrics
2020-04-24T20:38:38Z D! [outputs.wavefront] Wrote batch of 1000 metrics in 14.219747ms
2020-04-24T20:38:38Z D! [outputs.wavefront] Wrote batch of 1000 metrics in 11.761473ms
2020-04-24T20:38:38Z D! [outputs.wavefront] Wrote batch of 1000 metrics in 10.877892ms
2020-04-24T20:38:38Z D! [outputs.wavefront] Wrote batch of 1000 metrics in 16.069646ms
....
2020-04-24T20:38:39Z D! [outputs.wavefront] Wrote batch of 1000 metrics in 10.612196ms
2020-04-24T20:38:39Z D! [outputs.wavefront] Buffer fullness: 1228 / 250000 metrics
2020-04-24T20:38:39Z E! [agent] Error writing to outputs.wavefront: Wavefront sending error: buffer full, dropping line`

danielnelson:
@prydin Can you take a look?

The outputs don't know what the `metric_buffer_limit` is by design, but they should always fully send the batch during the write before returning.  This will prevent them from falling behind and it is also essential in order to work correctly with retries and "tracking metrics" from inputs such as the `kafka_consumer`.  Maybe we need to call Flush on the sender before returning from Write()?

prydin:
[...]

PR #8002, opened 2020-08-18
Make Wavefront internal sender's buffer size configurable

hanwavefront:
### Required for all PRs:

Updating Wavefront Output Plugin with buffer size configuration option (#7409) and related config options.

### DIFF
--- a/plugins/outputs/wavefront/README.md
+++ b/plugins/outputs/wavefront/README.md
@@ -1,6 +1,7 @@
 # Wavefront Output Plugin
 
-This plugin writes to a [Wavefront](https://www.wavefront.com) proxy, in Wavefront data format over TCP.
+This plugin writes in [Wavefront](https://www.wavefront.com) data format to a Wavefront proxy 
+(over HTTP or TCP) or directly to the Wavefront service (over HTTP).
 
 
 ### Configuration:
@@ -18,6 +19,18 @@ This plugin writes to a [Wavefront](https://www.wavefront.com) proxy, in Wavefro
 
   ## Port that the Wavefront proxy server listens on. Do not use if url is specified
   #port = 2878
+
+  ## Size of internal buffers beyond which received data is dropped. Applicable only if url is specified.
+  ## Buffers are not pre-allocated to max size and vary based on actual usage.
+  ## Defaults to 50,000. Higher values could use more memory.
+  #max_buffer_size = 50000
+
+  ## Max batch of data sent per flush interval. Applicable only if url is specified.
+  ## Defaults to 10,000. Recommended not to exceed 40,000. 
+  #batch_size = 10000
+
+  ## Interval (in seconds) at which to flush data. Defaults to 5 seconds.
+  #flush_interval_seconds = 5
 
   ## prefix for metrics keys
   #prefix = "my.specific.prefix."
--- a/plugins/outputs/wavefront/wavefront.go
+++ b/plugins/outputs/wavefront/wavefront.go
@@ -13,20 +13,23 @@ import (
 const maxTagLength = 254
 
 type Wavefront struct {
-	Url             string
-	Token           string
-	Host            string
-	Port            int
-	Prefix          string
-	SimpleFields    bool
-	MetricSeparator string
-	ConvertPaths    bool
-	ConvertBool     bool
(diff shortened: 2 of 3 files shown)


sjwang90:
Thanks for the PR! We'll take a look and let you know if we have any comments. 

@prydin If you have any feedback that'd be great.

prydin:
> @prydin If you have any feedback that'd be great.

Looks good to me.

plugins/outputs/wavefront/README.md
```diff
@@ -19,6 +20,18 @@ This plugin writes to a [Wavefront](https://www.wavefront.com) proxy, in Wavefro
...
+  ## Buffers are not pre-allocated to max size and vary based on actual usage.
+  ## Defaults to 50,000. Higher values could use more memory.
+  #max_buffer_size = 50000
+
+  ## Max batch of data sent per flush interval. Applicable only if url is specified.
+  ## Defaults to 10,000. Recommended not to exceed 40,000. 
+  #batch_size = 10000
+
+  ## Interval (in seconds) at which to flush data. Defaults to 5 seconds.
+  #flush_interval_seconds = 5
```
ssoroka, line 33: flush_interval, metric_buffer_limit and metric_batch_size are set automatically and override-able on outputs. Typically we don't add these to the output config readme, as they can be overridden on any output, and adding them to the config is going to conflict with those values.

ssoroka (changes requested):
I guess at this point I don't understand why we would want to define these on the plugin. These are settings that are globally available already, and it doesn't appear that the plugin needs to access these.

ssoroka:
Ok, I see the issue. To clarify, the problem is that `wavefront-sdk-go`'s BatchSize, MaxBufferSize, (though not necessarily FlushIntervalSeconds) conflict with our naming of BatchSize, MaxBufferSize, and FlushInterval, and setting these on the output is going to block the ability to override these settings for this output, so we probably don't want to do that.

Possible options: 
A) Name them WavefrontBatchSize, WavefrontMaxBufferSize, and WavefrontFlushInterval (the seconds part is kind of pointless if you have duration types).
B) Don't expose them and set these values to whatever the influx values are set to. This might be difficult because I don't think they're directly exposed to the plugin.

What's the problem we're trying to solve here? I don't see an open issue.

hanwavefront:
@ssorathia Thanks for the review.  The corresponding issue is #7409 -- `wavefront-sdk-go`'s config params (and specifically the buffer limit) need to be customizable to align with the influx values and prevent points from being dropped.  I'm hesitant about option B because I can envision a scenario where we would want the values to be different than the influx values (e.g., internal metrics in `wavefront-sdk-go`).  But I'm not clear on where the name conflict is and why option A is necessary?  The names are different from the influx names (MetricBatchSize, MetricBufferLimit, and FlushInterval), and they would only be configured on `[[outputs.wavefront]]`.  Let me know if I'm missing something.

hanwavefront:
@ssoroka Please take a look at my previous comment if you get a chance.  These new config params pertain only to the Wavefront output plugin's behavior and I don't believe they conflict with their influx counterparts, but let me know if I'm mistaken.

ssoroka:
After reading the issue, I'm concerned that this change won't resolve the problem. Wavefront should not have its own internal buffer. You end up with two buffers, and it doesn't matter what the size is, it's going to eventually drop metrics. Wavefront should instead be blocking and _actually writing the metrics out_ when passed batches of 1000 (agent.metric_batch_size) metrics, because Telegraf is managing the buffer, and keeping track of retries on failures. 

Said another way, because Wavefront is not blocking and applying backpressure, it's not communicating with Telegraf that the buffer is full and to slow down. Due to this lack of communication, Telegraf happily sends all the metrics in, resulting in overflowing the wavefront buffer.

I'd recommend removing the wavefront buffer and sending the metrics immediately when passed a batch, and blocking until it succeeds or fails. If for some reason you really must have double buffers (I still think this is a mistake), then wavefront needs to _block_ when the buffer is full, until there is room in the buffer to accept the new batch, not return an error and drop messages.

cc @prydin

ssoroka:
closing this in favor of https://github.com/influxdata/telegraf/pull/8165

PR #8005, opened 2020-08-19
Fix docker-image make target

pmalek-sumo:
`COMMIT` doesn't seem to be defined anywhere which causes:

```
docker build -f scripts/stretch.docker -t "telegraf:" .
invalid argument "telegraf:" for "-t, --tag" flag: invalid reference format
See 'docker build --help'.
make: *** [docker-image] Error 125
```

whereas `commit` is defined in https://github.com/influxdata/telegraf/blob/master/Makefile#L10

### Required for all PRs:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -160,7 +160,7 @@ clean:
 
 .PHONY: docker-image
 docker-image:
-	docker build -f scripts/stretch.docker -t "telegraf:$(COMMIT)" .
+	docker build -f scripts/stretch.docker -t "telegraf:$(commit)" .
 
 plugins/parsers/influx/machine.go: plugins/parsers/influx/machine.go.rl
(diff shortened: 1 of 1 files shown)


sjwang90:
Thanks for the contribution. We'll take a look at the PR and let you know if we have any comments.

pmalek-sumo:
Actually I'm not sure if we'd like to keep maintaining this as the make target that is referred in [scripts/stretch.docker](https://github.com/influxdata/telegraf/blob/9a5fd65/scripts/stretch.docker#L5): `go-install` doesn't exist in the repo anymore.

It was removed in https://github.com/influxdata/telegraf/pull/7759.

Maybe we should just remove the remnants of the previous build scripts/targets?

PR #8006, opened 2020-08-19
Fix string to int conversion in eventhub consumer

pmalek-sumo:
The existing conversion produces a vet check (which is enabled by default starting with 1.15) when running `go test` in 1.15 as per https://golang.org/doc/go1.15#vet.

```
$ make test
go test -short -race ./...
?       github.com/influxdata/telegraf  [no test files]
# github.com/influxdata/telegraf/plugins/inputs/eventhub_consumer
plugins/inputs/eventhub_consumer/eventhub_consumer.go:379:40: conversion from int16 to string yields a string of one rune, not a string of digits (did you mean fmt.Sprint(x)?)
...
```

### Required for all PRs:


reimda:
Thanks!

MSIT-BPSC-IOT:
Why Prefetch_count is not working?

PR #7213, opened 2020-03-22
New opc ua client plugin 

fdamador:
related #3041

### Required for all PRs:

- [x] Signed [CLA](https://influxdata.com/community/cla/).
- [x] Associated README.md updated.
- [x] Has appropriate unit tests.

PR #8009, opened 2020-08-20
Opcua input plugin + authentication options

haylesnortal:
### Required for all PRs:

Opcua plugin. Adapted from PR: 7213
https://github.com/influxdata/telegraf/pull/7213

Changes include authentication and encryption options.


plugins/inputs/opcua/opcua_client.go
```diff
@@ -0,0 +1,432 @@
...
+			TimestampsToReturn: ua.TimestampsToReturnBoth,
+		}
+
+		err = o.getData()
+		if err != nil {
+			return fmt.Errorf("Get Data Failed: %v", err)
+		}
+
+	default:
+		return fmt.Errorf("invalid opc.tcp endpoint")
```
ssoroka, line 318:
```suggestion
		return fmt.Errorf("unsupported scheme %q in endpoint. Expected opc.tcp", u.Scheme)
```

plugins/inputs/opcua/opcua_client.go
```diff
@@ -0,0 +1,432 @@
...
+	Endpoint       string   `toml:"endpoint"`
+	SecurityPolicy string   `toml:"security_policy"`
+	SecurityMode   string   `toml:"security_mode"`
+	Certificate    string   `toml:"certificate"`
+	PrivateKey     string   `toml:"private_key"`
+	Username       string   `toml:"username"`
+	Password       string   `toml:"password"`
+	AuthMethod     string   `toml:"auth_method"`
+	Interval       string   `toml:"time_interval"`
+	TimeOut        int      `toml:"timeout"`
```
ssoroka, line 28: should both be a `config.Duration`?

plugins/inputs/opcua/README.md
```diff
@@ -0,0 +1,76 @@
...
+## s=Temperature is indicting that the identifier type is a 'string' and the indentifier value is 'Temperature'
+## This temperature node may have a current value of 79.0, which would possibly make the value a 'float'.
+## To gather data from this node you would need to enter the following line into 'nodes' property above:
+##     {name="SomeLabel", namespace="3", identifier_type="s", identifier="Temperature", data_type="float", description="Some description."},
+
+```
+### Example Output:
+
+```
+opcua,host=3c70aee0901e,name=Random,type=double Random=0.018158170305814902 1597820490000000000
```
ssoroka, line 74: why report the data type every metric? This is a bit unusual.

metric/builder.go
```diff
@@ -0,0 +1,55 @@
...
+type TimeFunc func() time.Time
+
+type Builder struct {
+	TimeFunc
+	TimePrecision time.Duration
+
+	*metric
+}
+
+func NewBuilder() *Builder {
```
ssoroka, line 18: Really don't like metric builders. They are not so complicated that you need a builder to return a metric. This also isn't used anywhere and should be removed.

plugins/inputs/opcua/opcua_client.go
```diff
@@ -0,0 +1,432 @@
...
+			return fmt.Errorf("Get Data Failed: %v", err)
+		}
+
+	default:
+		return fmt.Errorf("invalid opc.tcp endpoint")
+	}
+	return nil
+}
+
+func (o *OpcUA) setupOptions() error {
```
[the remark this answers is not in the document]
haylesnortal: Yup. Makes sense. I'll get that moved.

plugins/inputs/opcua/README.md
haylesnortal: Agreed. Especially as it's known in advance and required as part of the plugin configurations. i.e. it's not adding any new information.

plugins/inputs/opcua/opcua_client.go
haylesnortal: There may have been a solid reason for interval using a string. I'll look into it and change as appropriate.

metric/builder.go
haylesnortal: Yeah, this isn't a part of the plugin. Likely a file from an older version that slipped through a merge/rebase/some git weirdness. I'll remove.

sjwang90:
@fdamador if you're interesting in testing this code with authentication options based off your original PR that would be really helpful.

plugins/inputs/opcua/README.md
```diff
@@ -0,0 +1,76 @@
+# Telegraf Input Plugin: opcua_client
+
+The opcua_client plugin retrieves data from OPCUA slave devices
+
+### Configuration:
+
+```toml
+
```
sjwang90, line 8: Include `[[inputs.opcua_client]]` into the same config for the the README

plugins/inputs/opcua/README.md
```diff
@@ -0,0 +1,76 @@
...
+		{name="ManufacturerName", namespace="0", identifier_type="i", identifier="2263", data_type="string", description="open62541"},
+]
+
+## Guide:
+## An OPC UA node ID may resemble: "n=3,s=Temperature"
+## In this example, n=3 is indicating the namespace is '3'.
+## s=Temperature is indicting that the identifier type is a 'string' and the indentifier value is 'Temperature'
+## This temperature node may have a current value of 79.0, which would possibly make the value a 'float'.
+## To gather data from this node you would need to enter the following line into 'nodes' property above:
+##     {name="SomeLabel", namespace="3", identifier_type="s", identifier="Temperature", data_type="float", description="Some description."},
```
sjwang90, line 68: Does this need to be in the sample config? Could bring it out as just plain text in the README

plugins/inputs/opcua/README.md
```diff
@@ -0,0 +1,76 @@
+# Telegraf Input Plugin: opcua_client
+
+The opcua_client plugin retrieves data from OPCUA slave devices
```
sjwang90, line 3: change "slave devices" to "client devices"

plugins/inputs/opcua/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+### Configuration:
+
+```toml
+
+# ## Connection Configuration
```
reimda, line 9: Clean up the comments in the example config here and in sampleConfig.  They need to match the commenting and indentation style because they are used when generating a config.  See https://github.com/influxdata/telegraf/wiki/SampleConfig

Accurate default values should be included here and commented out with a single hash.  Documentation is commented out with two hashes, not hash space hash hash

plugins/inputs/opcua/README.md
```diff
@@ -0,0 +1,76 @@
...
+#  ##
+#  ## The plugin supports connections to PLCs via OPCUA
+#  ##
+#  ## Device name
+name = "opcua_rocks"
+#
+#  # OPC UA Endpoint URL
+endpoint = "opc.tcp://opcua.rocks:4840"
```
reimda, line 17: Don't use opcua.rocks here.  If someone generates a config file with opcua enabled we don't want to point telegraf there.  It might be good to default to localhost if you expect people to run telegraf on the same machine as their opcua endpoint or leave it unset so people will have to provide their endpoint for it to work.

plugins/inputs/opcua/opcua_client.go
reimda: Did you look into this?  It seems error prone to allow people to set the interval as a string.  The docs say 5s or 10s which is the format config.Duration uses.

plugins/inputs/opcua/README.md
```diff
@@ -0,0 +1,76 @@
...
+#
+#  ## Read Timeout
+#  ## add an arbitrary timeout (seconds) to demonstrate how to stop a subscription
+#  ## with a context.
+timeout = 30
+#
+#  # Time Inteval, default = 10s
+time_interval = "5s"
```
reimda, line 25: The code in opcua_client.go says time_interval defaults to opcua.DefaultSubscriptionInterval.  The gopcua docs say that's 100ms.  https://godoc.org/github.com/gopcua/opcua#pkg-constants

Which default is right? 10s, 5s, 100ms?

plugins/inputs/opcua/README.md
reimda: That sounds clearer to me too.  Maybe make a subsection under configuration to describe how to use the nodes setting.  See https://github.com/influxdata/telegraf/blob/master/plugins/inputs/EXAMPLE_README.md

plugins/inputs/opcua/README.md
```diff
@@ -0,0 +1,76 @@
...
+#  ## namespace  			- integer value 0 thru 3
+#  ## identifier_type		- s=string, i=numeric, g=guid, b=opaque
+#  ## identifier			- tag as shown in opcua browser
+#  ## data_type  			- boolean, byte, short, int, uint, uint16, int16, uint32, int32, float, double, string, datetime, number
+#  ## Template 			- {name="", namespace="", identifier_type="", identifier="", data_type="", description=""},
+nodes = [
+		{name="ProductName", namespace="0", identifier_type="i", identifier="2261", data_type="string", description="open62541 OPC UA Server"},
+		{name="ProductUri", namespace="0", identifier_type="i", identifier="2262", data_type="string", description="http://open62541.org"},
+		{name="ManufacturerName", namespace="0", identifier_type="i", identifier="2263", data_type="string", description="open62541"},
+]
```
reimda, line 60: Move this out of the sample config.  It's a good config example and it makes sense to have it in the readme but we don't want to include it in everyone's config.

plugins/inputs/opcua/README.md
```diff
@@ -0,0 +1,76 @@
...
+#
+#  # OPC UA Endpoint URL
+endpoint = "opc.tcp://opcua.rocks:4840"
+#
+#  ## Read Timeout
+#  ## add an arbitrary timeout (seconds) to demonstrate how to stop a subscription
+#  ## with a context.
+timeout = 30
+#
+#  # Time Inteval, default = 10s
```
reimda, line 24: typo Inteval should be Interval

plugins/inputs/opcua/README.md
```diff
@@ -0,0 +1,76 @@
...
+time_interval = "5s"
+#
+#  # Security policy: None, Basic128Rsa15, Basic256, Basic256Sha256. Default: auto
+security_policy = "None"
+#
+#  # Security mode: None, Sign, SignAndEncrypt. Default: auto
+security_mode = "None"
+#
```
reimda, line 32: Include auto in the list of valid values.  Don't set them to "None" in the sample config if they default to auto.

Also, where in the code do these default to auto?  I didn't find it in a quick review.

plugins/inputs/opcua/README.md
```diff
@@ -0,0 +1,76 @@
...
+#  # auth_method = "Anonymous"
+#
+#  # Required for auth_method = "UserName"
+#  # username = "myusername"
+#
+#  # Required for auth_method = "UserName"
+#  # password = "mypassword"
+#
+#  ## Measurements
+#  ## node id to subscribe to
```
reimda, line 49: Use full sentences with punctuation and capitalization.

PR #8013, opened 2020-08-20
modbus_gateway input plugin

wz2b:
### Required for all PRs:

The motivation behind this plugin is to address the situation where there are multiple devices
behind a gateway, but the gateway can't accept many TCP connections at once. This is
a typical case (unfortunately) because many gateways on the market are made using
small microcontrollers with small RAM and limited TCP/IP stacks. Even some
expensive devices, like the PowerLogic CM4200, only allow 4 simultaneous connections.
The MBAP
protocol includes a unit identifier in each request specifically so that one TCP
connection can be shared talking to multiple devices. (Historical note: Modbus/UDP exists
as well, but is used less often).

The Modbus Gateway plugin collects Input Registers and Holding
Registers via Modbus TCP. It is similar to the "modbus" driver (and uses the same
underlying protocol implementation) but has a different configuration format suited to
communicating with Modbus/TCP devices acting as gateways. A gateway is still a modbus server,
but instead of reaching one device a gateway has many real or virtual devices attached to it.
The devices beyond the gateway are often connected via modems, radios, or multi-drop RS-485
or RS-422 buses.

Before writing this, I attempted to modify existing modbus plugin but the changes were too extensive to support multiple unit addresses in a single PDU, so I started over. Extensive explanation why this is needed in the README.md file. I plan additional enhancements, but this is good enough for many purposes as-is and since this is my first contribution I'm going to keep it simple and start here for now.


plugins/inputs/all/all.go
```diff
@@ -95,6 +95,7 @@ import (
 	_ "github.com/influxdata/telegraf/plugins/inputs/mesos"
 	_ "github.com/influxdata/telegraf/plugins/inputs/minecraft"
 	_ "github.com/influxdata/telegraf/plugins/inputs/modbus"
+	_ "github.com/influxdata/telegraf/plugins/inputs/modbusgw"
```
ssoroka, line 98: would really like to see the module, folder, and plugin name change to `modbus_gateway`

plugins/inputs/modbusgw/sample.go
```diff
@@ -0,0 +1,104 @@
+package modbusgw
+
+const sampleConfig = `
+[[inputs.modbusgw]]
```
ssoroka, line 4: don't include the plugin name in the sample config (it gets included automatically)

ssoroka:
cc @garciaolais @srebhan @sensor-freak in case you're interested and have any feedback

plugins/inputs/modbusgw/sample.go
wz2b: should i fix that, and also change the name, and re-submit?  Or add that as another PR?  Or you fx it when/after you merge?

ssoroka: just make the change in a new commit (same branch) and push it here in addition to the first commit. They'll all get squashed into one later when it gets merged.

plugins/inputs/modbus_gateway/README.md
```diff
@@ -0,0 +1,178 @@
...
+but you don't want to store the middle registers at all.  Why would you do that? Because it
+can be less "expensive" to have fewer requests.  If you were to request all four values,
+but only assign 1000 and 1993 to measurements, you sent 8 unwanted bytes across the
+wire.  That's far cheaper than requesting 1000 and 1003 in totally separate requests.
+To accomodate this type of request a field can be marked _omit=true_ meaning the
+response is received but not stored as a field of the measurement. 
+
+## Sample Configuration
+```toml
+[[inputs.modbusgw]]
```
ssoroka, line 82: missed one here
```suggestion
[[inputs.modbus_gateway]]
```

sensor-freak:
Good work, thank you!

But please let me comment on one point: the usage and implicit conversion of the data type.

You implicitly convert INT types to float values, even if no offset/scale is given. This might not always be desireable. As far as I can see, all INT measurements are converted to floating point values on the output (Influx or whatever). But sometimes one wants to keep the INT-property on the output. For this the original modbus plugin differentiates between INT types and FIXED types. Using INT types keps the values as integers, even when scaling/offset is applied, while the FIXED types behave similar to your implementation. (See also discussions on PR #7869, and maybe also issue #7317 and PR #7488.)

Maybe this can be kept in mind when the plugin gets extended in the future?

sensor-freak:
I'm sorry to add more requirements...

I'd like to request support for INT64 and UINT64. This is used for example by SMA for some energy counters. Probably there are also applications that provide IEE 754 double precision floats, but for this I actually do not have an example.

And regarding the other types:

size | format | order | old plugin terminology | old plugin order | output format
-- | -- | -- | -- | -- | --
32 | IEEE 754 single precision float | MSW, LSW | FLOAT32<b>-IEEE</b> | ABCD | usually float
32 | IEEE 754 single precision float | LSW, MSW | FLOAT32<b>-IEEE</b> | CDAB | usually float

Please note that FLOAT32 in the old modbus plugin is handles the same as the (newer) type (U)FIXED, e.g. it converts binary integer type to float (by applying scale/offset). The binary IEEE float input type has always been named FLOAT32-IEEE.

wz2b:
> I'd like to request support for INT64 and UINT64. This is used for example by SMA for some energy counters. Probably there are also applications that provide IEE 754 double precision floats, but for this I actually do not have an example.

Most PowerLogic meters do, too, for energy (kWh).  I'll add it to the short list.

> size	format	order	old plugin terminology	old plugin order	output format
> 32	IEEE 754 single precision float	MSW, LSW	FLOAT32-IEEE	ABCD	usually float
> 32	IEEE 754 single precision float	LSW, MSW	FLOAT32-IEEE	CDAB	usually float
> Please note that FLOAT32 in the old modbus plugin is handled the same as the (newer) type (U)FIXED, e.g. it converts binary integer type to float (by applying scale/offset). The binary IEEE float input type has always been named FLOAT32-IEEE.

ok.  I'm splitting those byte ordering things out to their own reusable unit ...

wz2b:
```
./scripts/check-deps.sh
--- /tmp/tmp.EzaAinpP8Q/LICENSE_OF_DEPENDENCIES.md	2020-08-24 23:05:06.262716479 +0000
+++ /tmp/tmp.EzaAinpP8Q/HEAD	2020-08-24 23:05:06.258716471 +0000
@@ -16,0 +17 @@
+github.com/alecthomas/template
@@ -151,0 +153 @@
+gopkg.in/alecthomas/kingpin.v2
```

I'm not sure what is causing this - I don't think it's something I did, because I haven't added/removed any dependencies or changed go.mod or go.sum.  I think maybe some transitive dependency changed somewhere, and I was unfortunate enough to commit at the wrong time.

plugins/inputs/modbus_gateway/convert.go
```diff
@@ -0,0 +1,117 @@
...
+	orderSpecUC := strings.ToUpper(orderSpec)
+	orderSpecLen := len(orderSpecUC)
+	orderSpecBytes := []byte(orderSpecUC)
+
+	converter := CustomByteOrder{
+		order: orderSpecUC,
+	}
+
+	i := 0
+	for ; i < orderSpecLen; i++ {
```
garciaolais, line 42: I suggest  `for i := 0; i < orderSpecLen; i++ `

wz2b: No, i need i to pick up where it left off in the next for loop down

garciaolais: I see, but at the final of the first loop always `i` be equal to `orderSpecLen`

you can use `orderSpecLen `in the next loop

`for ;  orderSpecLen < 8; i++`

wz2b: Right, thank you

srebhan:
Hmmm I still don't get it. We have all the features requested additionally in the existing plugin. We only miss:

1. Specifying multiple slave-IDs per plugin
2. Allowing to specify but omit registers in order to optimize the query chunks.

I'm pretty convinced that this can be added to the existing plugin... What am I missing here?

wz2b:
> I'm pretty convinced that this can be added to the existing plugin... What am I missing here?

It's a fair question.

First off, I personally prefer this mechanism of declaring the frame specifications and interpretations separately.  But I get that to some people that might be just 'more verbose.'  To me it's more explicit.  That, however, is a matter of personal preference - others may disagree.

But my main reason was what I felt would be creeping complexity.  The old driver works by building up ranges of registers to request

```
	/* These are only used for configuration */
	DiscreteInputs   []fieldContainer  `toml:"discrete_inputs"`
	Coils            []fieldContainer  `toml:"coils"`
	HoldingRegisters []fieldContainer  `toml:"holding_registers"`
	InputRegisters   []fieldContainer  `toml:"input_registers"`

	/* InitRegister() reorganizes them to this */
	registers        []register
```

To work with multiple slaves, `registers` has to become a map, and the loops in that function have to be wrapped with outer loops so that they do their work on a per-slave basis, so that it can repeat the adjacent-register groupings _per slave_.  I started down this path but I found the resulting code too hard to follow, and I was afraid to make that much of a change and risk breaking existing code.  That's why I started fresh.

I also have the idea that this ultimately should be a service plugin.  In my use case I'm polling a lot of different gateways, and there's no reason they shouldn't all be happening concurrently.  I'm just taking an agile approach so I haven't broached the details of that yet.

So where do we go from here?  How do we decide?  If we decide to reject modbus_gateway and try to roll these features into the modbus plugin, is anybody volunteering to do that?

ssoroka (approved):

mjf:
@wz2b How difficult would it be to add support for the *Modbus RTU* protocol (or perhaps all of the protocols supported by the underlying *Modbus* library)?

wz2b:
> @wz2b How difficult would it be to add support for the _Modbus RTU_ protocol (or perhaps all of the protocols supported by the underlying _Modbus_ library)?

It would not be that difficult if the underlying library supports it, but that would be a direct serial connection, not a "gateway."  It could be added, though.  I'm wondering why, though - do you prefer this configuration format over the older one?  (That was one of my reasons for doing this).

lampra1:
> It would not be that difficult if the underlying library supports it, but that would be a direct serial connection, not a "gateway." It could be added, though. I'm wondering why, though - do you prefer this configuration format over the older one? (That was one of my reasons for doing this).

Though, as an example based on your sample configuration, I expect that it would make sense for the gateway to have a direct serial (rs485) connection to one or more energy meters and batteries and poll at the same time multiple solar inverters using modbus tcp.

ssoroka:
Is there anything left for this or are we good to merge?

ssoroka:
@srebhan does #9141 and/or #7941 need to get merged before this?

wz2b:
> Is there anything left for this or are we good to merge? Looks like it needs some updates; `internal.Duration` should now be `config.Duration`

I have been detached from this for a while as I had to move on to other things, but if updates are needed I can jump back in.  I just need to know what needs to be done.  I've been using the current version (as-is) without having updated it, and it hasn't done anything funny.  I did see an earlier question about being able to control polling speed with the original modbus plugin (not this one) and adding a little more control over poll rates seems like a useful feature to add to this one.  I would be interested in doing that but I think it'd be better off as a separate branch and PR.  So my preference would be to close out this one before I start making any enhancements.  That said, if something needs fixing now, I will fix it.

srebhan:
@ssoroka This is a completely orthogonal approach to what I did in #9141 and will add *another* modbus plugin.
I have a PR (based on PR #9141) that covers this PR and also allows to fold in #7941 into the current plugin while keeping all the functionality...

wz2b:
> @ssoroka This is a completely orthogonal approach to what I did in #9141 and will add _another_ modbus plugin.
> I have a PR (based on PR #9141) that covers this PR and also allows to fold in #7941 into the current plugin while keeping all the functionality...

As I mentioned in #9141 I'm not sure I am okay with the idea of merging the configurations and not making another but I need the ability to specify the unit id as part of each request specification, otherwise it doesn't really work as a modbus gateway.  Most of the other changes I made here were embellishments on that original requirement.  The other part that I don't see in #9141 is the ability to 'omit' fields.

srebhan:
With #9279 being merged, I guess we can close this PR. @wz2b please reopen if you disagree.

wz2b:
> With #9279 being merged, I guess we can close this PR. @wz2b please reopen if you disagree.

Agree. If there are other changes/improvements to this plugin I think they would warrant a new issue.  All the changes thus far are working very well, I am using in production and keeping a careful eye on things and it's 👍 

Nice work (and collaboration) on this, Sven, thank you for your effort.

srebhan:
Thanks @wz2b for your kind words. I think we also discussed on per-request tags... If so, please take a look at (and test :-)) PR #10231.

Issue #8014, opened 2020-08-20
Dynatrace output plugin - setting insecure_skip_verify broken


dlopes7:
### Relevant telegraf.conf:
```toml
# # Send telegraf metrics to a Dynatrace environment
[[outputs.dynatrace]]
#   ## For usage with the Dynatrace OneAgent you can omit any configuration,
#   ## the only requirement is that the OneAgent is running on the same host.
#   ## Only setup environment url and token if you want to monitor a Host without the OneAgent present.
#   ##
#   ## Your Dynatrace environment URL.
#   ## For Dynatrace OneAgent you can leave this empty or set it to "http://127.0.0.1:14499/metrics/ingest" (default)
#   ## For Dynatrace SaaS environments the URL scheme is "https://{your-environment-id}.live.dynatrace.com/api/v2/metrics/ingest"
#   ## For Dynatrace Managed environments the URL scheme is "https://{your-domain}/e/{your-environment-id}/api/v2/metrics/ingest"
  url = "https://localhost:9999/e/eaa50379/api/v2/metrics/ingest"
#
#   ## Your Dynatrace API token.
#   ## Create an API token within your Dynatrace environment, by navigating to Settings > Integration > Dynatrace API
#   ## The API token needs data ingest scope permission. When using OneAgent, no API token is required.
  api_token = "<REDACTED>"
#
#   ## Optional prefix for metric names (e.g.: "telegraf.")
  prefix = "telegraf."
#
#   ## Optional TLS Config
#   # tls_ca = "/etc/telegraf/ca.pem"
#   # tls_cert = "/etc/telegraf/cert.pem"
#   # tls_key = "/etc/telegraf/key.pem"
#
#   ## Optional flag for ignoring tls certificate check
  insecure_skip_verify = true
#
#
#   ## Connection timeout, defaults to "5s" if not set.
#   timeout = "5s"
```

### Steps to reproduce:

1. Set insecure_skip_verify to any value
2. Run telegraf

### Expected behavior:

Expected the certificate to be ignored

### Actual behavior:

Telegraf won't start, with the error:

```
2020-08-20T13:17:00Z E! [telegraf] Error running agent: Error loading config file telegraf.conf: Error parsing dynatrace array, line 494: field corresponding to `insecure_skip_verify' in dynatrace.Dynatrace cannot be set through TOML
```
### Additional info:

[...]

PR #8015, opened 2020-08-20
Fix InsecureSkipVerify in Dynatrace Plugin

thschue:
### Required for all PRs:


sjwang90:
Issue https://github.com/influxdata/telegraf/issues/8014

ssoroka:
Rather than just removing the option, wouldn't it be better to add support for this and client certificates?

thschue:
> Rather than just removing the option, wouldn't it be better to add support for this and client certificates?

I think we implemented this by using the telegraf tls library (tls import and tls.ClientConfig in the struct, used in lines 253:264). This was also the cause for this Bug, as we switched to this without removing the "old" field.

PR #8019, opened 2020-08-21
added force_document_id option to ES output enable resend data avoiding duplicated ES documents, fix #7891

### LINKED ISSUE #7891: Ability in telegraf to use ID field from the metric, if available
## Feature Request

Currently, when we are saving metrics to output plugin (for e.g. Elasticsearch), there is no ability for telegraf to use the ID field from the payload (metrics) as ID of the metric being pushed. 
As a result, any metrics pushed to ES will **always** be taken as "New" document in the absence of ID field. 

### Proposal:

We provide a JSON parser configuration, similar to **json_time_key**,  named **json_id_key**. 
This will be an optional configuration just like _time_ key, if present, then the ID will be available to output plugins for use and assignment. 
If not present, then the current behaviour can continue to work in the absence of ID field.  

### Use case: 

One of the scenario where this is immensely useful is when saving metrics data to Elasticsearch. 
In a scenario where the metrics data from an input plugin is duplicated or same metrics is returned multiple times, then in the absence of this ID functionality, ES output plugin sends the metrics without an identifier.  This will cause ES to treat it as a new document and create a new ID and insert it into its datastore, hence resulting in duplicates of metrics. 
When the metrics is sent with an ID, then the ES will perform the "UPSERT" correctly, modifying any existing documents with an ID or create one if it does not ecxists.

Note - We would also need to make a change in desired output plugins to use the available ID property or not. That could depend on individual plugin choices. 

I have made changes to my local telegraf code, and have run the suggested steps to ensure its working fine. Let me know if this is something beneficial to all, then I can raise a PR accordingly. 
Thanks.

toni-moreno:
This PR solves #7891 
ID has been computed as hash computed with sha256(concat(timestamp,measurement,series-hash)),enables resend or update data avoiding ES duplicated documents.
Tested with several loads with the same data and no duplicated documents have been generated.

Hope this PR could help everybody.

### DIFF
--- a/etc/telegraf.conf
+++ b/etc/telegraf.conf
@@ -512,7 +512,9 @@
 #   template_name = "telegraf"
 #   ## Set to true if you want telegraf to overwrite an existing template
 #   overwrite_template = false
-
+#   ## If set to true a unique ID hash will be sent as sha256(concat(timestamp,measurement,series-hash)) string
+#   ## it will enable data resend and update metric points avoiding duplicated metrics with diferent id's
+#   force_document_id = false
 
 # # Send metrics to command as input over stdin
 # [[outputs.exec]]
--- a/plugins/outputs/elasticsearch/README.md
+++ b/plugins/outputs/elasticsearch/README.md
@@ -196,6 +196,9 @@ This plugin will format the events in the following way:
   template_name = "telegraf"
   ## Set to true if you want telegraf to overwrite an existing template
   overwrite_template = false
+  ## If set to true a unique ID hash will be sent as sha256(concat(timestamp,measurement,series-hash)) string
+  ## it will enable data resend and update metric points avoiding duplicated metrics with diferent id's
+  force_document_id = false
 ```
 
 #### Permissions
@@ -232,6 +235,7 @@ Additionally, you can specify dynamic index names by using tags with the notatio
 * `manage_template`: Set to true if you want telegraf to manage its index template. If enabled it will create a recommended index template for telegraf indexes.
 * `template_name`: The template name used for telegraf indexes.
 * `overwrite_template`: Set to true if you want telegraf to overwrite an existing template.
+* `force_document_id`: Set to true will compute a unique hash from as sha256(concat(timestamp,measurement,series-hash)),enables resend or update data withoud ES duplicated documents.
 
 ### Known issues
 
--- a/plugins/outputs/elasticsearch/elasticsearch.go
+++ b/plugins/outputs/elasticsearch/elasticsearch.go
@@ -11,6 +11,8 @@ import (
 	"text/template"
 	"time"
 
+	"crypto/sha256"
+
 	"github.com/influxdata/telegraf"
 	"github.com/influxdata/telegraf/internal"
 	"github.com/influxdata/telegraf/plugins/common/tls"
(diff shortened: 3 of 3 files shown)


toni-moreno:
Hello @ssoroka , thank you for the fast review.

About the chosen name "force_document_id".  I've chosen because "document_id" is something known for people used to play with ES and Logstash, and could help people to identify the property in both tools ( telegraf / logstash ).

* https://www.elastic.co/es/blog/logstash-lessons-handling-duplicates 
* https://www.elastic.co/guide/en/logstash/current/plugins-outputs-elasticsearch.html#plugins-outputs-elasticsearch-document_id) 

About turn on the property by default.   I put false by default , to make this change backwards compatible, I'm not in the head of the people working  with telegraf sending to elastic, and set the property to true by default would be a breaking change in some rare cases  , so I'd prefer maintain backwards compatibility right now and perhaps  change it in the future.

Anyway I'm open to consensus and change both things if other people could give us their opinion.

Thank you very much.

ssoroka:
@lpic10 do you want to weigh in here before I merge?

lpic10:
Concerning the default option, I don't know if there is a valid scenario in that the telegraf users would want the same data stored twice (maybe when there are really duplicated log lines or metric points?)

I understand that InfluxDB does that deduplication by timestamp and tags/fields automatically, so maybe that is not really something expected by most people. It could make sense to have this enabled by default on ES output even if there is a potential performance impact on telegraf side.

About the config name there seems to be no consistency on the other tools sending data to ES. (eg. fluentd calls it "hash_id" and beats/logstash calls it "fingerprint"). But in all cases this option is configurable.

For me both `force_document_id` and `avoid_duplicates` make sense but I'm bad at naming things.

ssoroka:
Ok. I think I like the idea of changing this to avoid duplicates by default, since this is more likely the expected behavior. In that case, I'd make a new config option called `allow_duplicates` or `disable_document_id` and default it to `false`. We should really strive to do the right thing most of the time, even if that's a breaking change. In this case, this change is more of a _fix_ than a _break_, anyway, as it probably should have been the default behavior in the first place, and likely anyone surprised by the change will be pleasantly surprised. Duplicate data is a pain.

melodous:
IMHO in the elasticsearch world it is not natural to perform this kind of deduplication, probably elasticsearch users does not expect this behavior by default. 

Also this behavior has a penalty in write throughput, from the official Elastic documentation:

'Use auto-generated ids
When indexing a document that has an explicit id, Elasticsearch needs to check whether a document with the same id already exists within the same shard, which is a costly operation and gets even more costly as the index grows. By using auto-generated ids, Elasticsearch can skip this check, which makes indexing faster'

ssoroka:
That's good feedback @melodous. I'll change my position to say we should choose the faster option by default and allow users interested in this to turn it on.

ssoroka:
@toni-moreno I'd say it's good to merge whenever you want. let me know if you plan to rename or want to merge as is.

toni-moreno:
Hello @ssoroka , INHO and regarding @melodous comment,  if set document_id from origin could be a costly operation in the ES backend  ,  the name  "force_document_id" ,gives people idea of something with possible penalty , and  looks good for me  if everybody agree.

Issue #7878, opened 2020-07-23
clustername tag value missing after upgrading to 1.15.1


ybinnenwegin2ip:
### Relevant telegraf.conf:
```toml
# Host & VM stats
[[inputs.vsphere]]
   interval = "60s"
   ## List of vCenter URLs to be monitored. These three lines must be uncommented
   ## and edited for the plugin to work.
   vcenters = [ "https://FQDN/sdk" ]
   username = "username"
   password = "hunter2"

   force_discover_on_init = true

   ## VMs
   ## Typical VM metrics (if omitted or empty, all metrics are collected)
   vm_metric_include = [
     "cpu.demand.average",
     "cpu.idle.summation",
     "cpu.latency.average",
     "cpu.readiness.average",
     "cpu.ready.summation",
     "cpu.run.summation",
     "cpu.usage.average",
     "cpu.usagemhz.average",
     "cpu.used.summation",
     "cpu.wait.summation",
     "mem.active.average",
     "mem.granted.average",
     "mem.latency.average",
     "mem.swapin.average",
     "mem.swapinRate.average",
     "mem.swapout.average",
     "mem.swapoutRate.average",
     "mem.usage.average",
     "mem.vmmemctl.average",
     "net.bytesRx.average",
     "net.bytesTx.average",
     "net.droppedRx.summation",
     "net.droppedTx.summation",
     "net.usage.average",
     "power.power.average",
     "virtualDisk.numberReadAveraged.average",
     "virtualDisk.numberWriteAveraged.average",
     "virtualDisk.read.average",
     "virtualDisk.readOIO.latest",
     "virtualDisk.throughput.usage.average",
     "virtualDisk.totalReadLatency.average",
     "virtualDisk.totalWriteLatency.average",
     "virtualDisk.write.average",
     "virtualDisk.writeOIO.latest",
     "sys.uptime.latest",
   ]
   # vm_metric_exclude = [] ## Nothing is excluded by default
   vm_instances = true ## true by default

   ## Hosts
   ## Typical host metrics (if omitted or empty, all metrics are collected)
   host_metric_include = [
     "cpu.coreUtilization.average",
     "cpu.costop.summation",
     "cpu.demand.average",
     "cpu.idle.summation",
     "cpu.latency.average",
     "cpu.readiness.average",
     "cpu.ready.summation",
     "cpu.swapwait.summation",
     "cpu.usage.average",
     "cpu.usagemhz.average",
     "cpu.used.summation",
[...]

PR #8026, opened 2020-08-24
Fixed issue 7878 (vSphere missing clustername)

prydin:
### Required for all PRs:

* Fixed issue #7878 (clustername missing)
* Fixed an issue causing clustername to be missing if data collection is turned off for clusters (unrelated to #7878)
* Added tests to make sure this doesn't happen again.

### DIFF
--- a/plugins/inputs/vsphere/endpoint.go
+++ b/plugins/inputs/vsphere/endpoint.go
@@ -844,7 +844,6 @@ func (e *Endpoint) Collect(ctx context.Context, acc telegraf.Accumulator) error 
 	if e.Parent.ObjectDiscoveryInterval.Duration == 0 {
 		err := e.discover(ctx)
 		if err != nil {
-			acc.AddError(err)
 			return err
 		}
 	}
--- a/plugins/inputs/vsphere/endpoint.go
+++ b/plugins/inputs/vsphere/endpoint.go
@@ -349,7 +349,7 @@ func (e *Endpoint) getMetadata(ctx context.Context, obj *objectRef, sampling int
 }
 
 func (e *Endpoint) getDatacenterName(ctx context.Context, client *Client, cache map[string]string, r types.ManagedObjectReference) (string, bool) {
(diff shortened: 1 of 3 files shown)


plugins/inputs/vsphere/endpoint.go
```diff
@@ -348,7 +348,11 @@ func (e *Endpoint) getMetadata(ctx context.Context, obj *objectRef, sampling int
 	return metrics, nil
 }
 
-func (e *Endpoint) getDatacenterName(ctx context.Context, client *Client, cache map[string]string, r types.ManagedObjectReference) string {
+func (e *Endpoint) getDatacenterName(ctx context.Context, client *Client, cache map[string]string, r types.ManagedObjectReference) (string, bool) {
+	return e.getAncestorName(ctx, client, "Datacenter", cache, r)
+}
+
+func (e *Endpoint) getAncestorName(ctx context.Context, client *Client, resourceType string, cache map[string]string, r types.ManagedObjectReference) (string, bool) {
```
ssoroka, line 355: It doesn't look like the returned bools are used anywhere. Maybe we can remove?

plugins/inputs/vsphere/endpoint.go
```diff
@@ -834,6 +844,7 @@ func (e *Endpoint) Collect(ctx context.Context, acc telegraf.Accumulator) error
 	if e.Parent.ObjectDiscoveryInterval.Duration == 0 {
 		err := e.discover(ctx)
 		if err != nil {
+			acc.AddError(err)
```
ssoroka, line 847: I think this will end up logging the error twice. no need to both log and return it.

plugins/inputs/vsphere/endpoint.go
prydin: It's used in some of the tests.

plugins/inputs/vsphere/endpoint.go
prydin: Will fix this.

PR #8032, opened 2020-08-25
http_response: status_code match

Eraac:
### Required for all PRs:

I've added the possibility to match on the status code instead of only the response body, can be useful if your service return a 204. 

I'm not fully satisfied because actually we can only check the body OR the status code (not both). Technically isn't difficult to be able to do both, but that require some decisions about the `result` value, if only one of these fail the value is pretty obvious but in case both check fail, do we need to create another value (`response_mismatch`?) or set the field to `response_status_code_mismatch` or `response_string_mismatch`

Thanks!

### DIFF
--- a/plugins/inputs/http_response/README.md
+++ b/plugins/inputs/http_response/README.md
@@ -45,8 +45,11 @@ This input plugin checks HTTP/HTTPS connections.
   # response_string_match = "ok"
   # response_string_match = "\".*_status\".?:.?\"up\""
 
-  ## Optional status code match of the response
-  # response_status_code_match = 204
+  ## Expected response status code.
+  ## The status code of the response is compared to this value. If they match, the field
+  ## "response_status_code_match" will be 1, otherwise it will be 0. If the
+  ## expected status code is 0, the check is disabled and the field won't be added.
+  # response_status_code = 0
 
   ## Optional TLS Config
   # tls_ca = "/etc/telegraf/ca.pem"
--- a/plugins/inputs/http_response/README.md
+++ b/plugins/inputs/http_response/README.md
@@ -45,6 +45,9 @@ This input plugin checks HTTP/HTTPS connections.
   # response_string_match = "ok"
   # response_string_match = "\".*_status\".?:.?\"up\""
 
+  ## Optional status code match of the response
+  # response_status_code_match = 204
+
   ## Optional TLS Config
   # tls_ca = "/etc/telegraf/ca.pem"
   # tls_cert = "/etc/telegraf/cert.pem"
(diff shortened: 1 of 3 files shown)


plugins/inputs/http_response/README.md
```diff
@@ -45,6 +45,9 @@ This input plugin checks HTTP/HTTPS connections.
   # response_string_match = "ok"
   # response_string_match = "\".*_status\".?:.?\"up\""
 
+  ## Optional status code match of the response
+  # response_status_code_match = 204
```
reimda, line 49: The name of the setting is a little confusing.  It is similar to the previous setting "response_string_match" but in that case you're doing a regex match.  I think calling this new setting response_status_code or response_expected_status_code would be clearer.

The convention with config examples is to show the default value in the line that's commented out.  It looks like the new setting defaults to 0.


Could you also add in the config example some text describing how the setting works?  Something like this:
```suggestion
  ## Expected response status code.
  ## The status code of the response is compared to this value.  If they match, the field
  ## "response_status_code_match" will be 1, otherwise it will be 0.  If the
  ## expected status code is 0, the check is disabled and the field won't be added.
  # response_status_code = 0
```

plugins/inputs/http_response/http_response.go
```diff
@@ -106,6 +107,9 @@ var sampleConfig = `
   # response_string_match = "ok"
   # response_string_match = "\".*_status\".?:.?\"up\""
 
+  ## Optional status code match of the response
+  # response_status_code_match = 204
```
reimda, line 111: Make this match the readme changes.

plugins/inputs/http_response/http_response.go
```diff
@@ -361,6 +366,14 @@ func (h *HTTPResponse) httpGather(u string) (map[string]interface{}, map[string]
 			setResult("response_string_mismatch", fields, tags)
 			fields["response_string_match"] = 0
 		}
+	} else if h.ResponseStatusCodeMatch > 0 {
```
reimda, line 369: I don't think you want an else here.  That would make the response code check only happen if "response_string_match" is empty.

If you remove the else here, give the block its own comment and undo the comment change above.
```suggestion
	}
	// Check the reponse status code
	if h.ResponseStatusCodeMatch > 0 {
```

Eraac: I've done the change, my little worry here is when the both condition fail, the `result` tag will have `response_status_code_mismatch` which it's can be confusing because isn't just this problem

Eraac:
The CI has failed but it's seems to be an error from another plugin, not sure I've an impact on this

Eraac:
@reimda They're a way to re-triggrer circleci without commit?

ssoroka:
would it make more sense to instead just record the status code, instead of a binary match/no-match?

Eraac:
> would it make more sense to instead just record the status code, instead of a binary match/no-match?

The status code is already record under `http_response_code`

reimda (approved):

PR #8039, opened 2020-08-26
Remove Event field as it is causing issues with pre-trained source types

samm-git:
### Required for all PRs:

- [N/A] Associated README.md updated.

### Description

I found that telegraf metrics export to splunk does not work on some of our environments. After all i found that reason was "event: metric"  object in the json. To understand if it is needed here and why it works on some other environments i opened ticket to the vendor.

Summary of replies is:

1. `event` field is not needed for the normal splunk metrics and it shoud work without it in all cases.
1. On source types with pre-trained event types (in our case it was `_json` or `collectd_http`) splunk would fail if this field is exists, probably due to more strict json checking.

After this i decided to raise this PR to remove this field from the upstream. I tested it on my environments and it works fine.


samm-git:
Few quotes from support replies, just in case

>  I also get the following error: 
> 07-23-2020 20:59:53.540 -0400 ERROR JsonLineBreaker - JSON StreamId:0 had parsing error:Unexpected character while looking for value: 'm' - data_source="disk", data_host="host_1.splunk.com", data_sourcetype="_json"
> I also confirmed with senior engineers that the "event: metric" field is not necessary, also, as per the error above, it is causing an issue during parsing, which could be causing Splunk to drop the event.
>  My advice would be to leave a message in the feedback section located in the lowest part of the page, in the link where you found the curl command, mentioning that the documentation should be updated.

samm-git:
@sjwang90 any chance to get this merged? We are using own fork due to that and it may affect other users.

ssoroka (approved):
seems reasonable. Thanks :)

PR #7988, opened 2020-08-15
gopsutil: bump to fix vendoring problem

Mic92:
This was fixed here https://github.com/shirou/gopsutil/pull/889

We use go mod vendor in NixOS to archive all dependencies and allow
reproducible offline builds.

Changelog is minimal otherwise:

- https://github.com/shirou/gopsutil/releases/tag/v2.20.6
- https://github.com/shirou/gopsutil/releases/tag/v2.20.7

### Required for all PRs:

- [x] Signed [CLA](https://influxdata.com/community/cla/).
- [ ] Associated README.md updated.
- [ ] Has appropriate unit tests.

PR #8047, opened 2020-08-28
Add XML parser using XPath queries

### LINKED ISSUE #1758: XML input

srebhan:
### Required for all PRs:

This PR adds a XML parser using XPath expressions to define fields, tags, etc. To achieve this, we use the underlying [antchfx/xpath library](https://github.com/antchfx/xpath). Please check there to see what expressions are supported.

The PR closes #1758 and #6968. The configuration syntax is borrowed from suggestions made by @danielnelson in https://github.com/influxdata/telegraf/issues/1758#issuecomment-304418082.
Furthermore, the code is heavily inspired by the json parser plugin.


ssoroka:
Looks like the build failure is related to https://github.com/shirou/gopsutil/issues/912

srebhan:
This is also related to PR #7988 trying to bump gopsutils...

M0rdecay:
Hello!

It looks like we are trying to solve one problem in parallel (https://github.com/M0rdecay/telegraf/tree/xml_parser/plugins/parsers/xml).

I like your solution, but I also have a couple of questions:
 - with your parser, is it possible to solve the problem when I need to get all the values ​​from a node without listing each of them in `fields`?
 - a similar question about tags
 - are all child nodes of the current node retrieved recursively?
 - is it possible to limit the array of analyzed nodes across the nodes?

srebhan:
Hey @M0rdecay,

indeed we are trying to solve the same problem... :-) Let me answer your questions:

>     * with your parser, is it possible to solve the problem when I need to get all the values ​​from a node without listing each of them in `fields`? In the case of very large documents, this would be useful
>     * a similar question about tags

I thought about this one and I think it can be done. I want to be able to specify something like 
```
field_query = some_node/*    # relative to selected node
field_query_name = @name    # relative to field_query or nothing for using the node name
field_query_value = @value    # relative to field_query or nothing for using the node value
```
and similar for tags. When the *field_query_name/_value* is given we take the values of the query relative to each *field_query* node. This allows to use attributes as names/values. If not given we simply take the node-names as field/tag names and values.

The problems I see:
- While this works nicely for tags, the type of the fields cannot be specified. This way everything will be a string... Is this acceptable?
- What if the name/values are given as attribute pairs? Ok we maybe could craft a XPath for that one...
- What do we do if both the queries above and field/tag-definitions are given?

>     * are all child nodes of the current node retrieved recursively?

I think you can do this as a XPath query (see `descendant`).

>     * is it possible to limit the array of analyzed nodes across the nodes?

This should also be possible with XPath syntax I think.

This being said, I first want to get this merged without adding more complexity. Let's extend this in a second round. I'm looking forward to your opinion on my thoughts above! Any help appreciated! :-)

M0rdecay:
Thank you for the quick response!

I agree, let`s continue in the second round)

I ask you to see the solution I proposed for your example - https://github.com/influxdata/telegraf/pull/7460#issuecomment-683879879 - I will wait for your opinion.

Answering your questions and suggestions:
 - `This way everything will be a string... Is this acceptable?` For myself, I found such a solution for converting a string - https://github.com/M0rdecay/telegraf/blob/xml_parser/plugins/parsers/xml/parser.go#L211, maybe it will work for you too
 - `What if the name/values are given as attribute pairs?` - Is it possible to take a key-value pair directly from a node/attribute? Does the library you use allow it? If so, it should make the task a lot easier. So far, the configuration without this feature seems to me very difficult, I would like a simpler configuration, like in `json` parser

It seems to me that trying to provide maximum flexibility, we may fall into the hell of configuration.  
The name and value of a field/tag can be formed directly from the name and value of a node or attribute. If you need to rename them, you can use processors - fortunately, the telegraf offers a rich set of tools for this.

srebhan:
@M0rdecay here is the config to reproduce your output on the given XML:
```toml
[[inputs.file]]
  files = ["M0rdecay.xml"]
  data_format = "xml"

  [[inputs.file.xml]]
    selected_nodes = "//Hosts/*[starts-with(name(), 'Host_')]"

    [inputs.file.xml.tags]
      name = "Name"
      xml_node_name="name()"

    [inputs.file.xml.fields_int]
      current = "Connections/Current"
      total = "Connections/Total"
      uptime = "Uptime"
```
It is a bit more verbose than your config without the field/tag_queries we discussed earlier, but as I said, I'm willing to add this later.

M0rdecay:
@srebhan hello!

Please, tell me, does your parser, when parsing an array, allow you to get tags / fields from the beginning of the document (like `//some_node`) or relative to the current parsed node, but at a higher level (`../../some_element`)?

In one of my cases, I found that I was missing such functionality.

srebhan:
Hey @M0rdecay,

sure. The readme states that if you start the path with a `/` you specify an absolute path, otherwise it is relative to the selected nodes (or the root node if non selected, which is the same as an absolute path :-)).
If you have more questions please join the slack-channel (I'm _srebhan_ :-)) as it is easier than filling the Pull-Request...

srebhan:
@M0rdecay:
I've found some time to implement the field selection similar to what you do. The only missing feature is the automagic type-conversion. I've we can settle on this parser you could easily implement that on top of the PR...
Could you give it a try and see if it works for you so we can get this rolling!?

M0rdecay:
> @M0rdecay:
> I've found some time to implement the field selection similar to what you do. The only missing feature is the automagic type-conversion. I've we can settle on this parser you could easily implement that on top of the PR...
> Could you give it a try and see if it works for you so we can get this rolling!?

Sure. I'll try to do it by tomorrow night.

M0rdecay:
I have checked if this works for me with one example. 
We discussed a case with Sven - Unfortunately, it didn't work for me :(

For a part of document like this, information will be lost:
```xml
                <COMMAND>
                    <CMD>41</CMD>
                    <ALL>
                        <PENDING_COUNT>0</PENDING_COUNT>
                        <MIN_DURATION>0</MIN_DURATION>
                        <MAX_DURATION>0</MAX_DURATION>
                        <DURATION_SUM>0</DURATION_SUM>
                        <DURATION_SQR_SUM>0</DURATION_SQR_SUM>
                        <DET_COUNT>0</DET_COUNT>
                    </ALL>
                    <SUCC>
                        <PENDING_COUNT>0</PENDING_COUNT>
                        <MIN_DURATION>0</MIN_DURATION>
                        <MAX_DURATION>0</MAX_DURATION>
                        <DURATION_SUM>0</DURATION_SUM>
                        <DURATION_SQR_SUM>0</DURATION_SQR_SUM>
                        <DET_COUNT>0</DET_COUNT>
                    </SUCC>
                    <FAIL>
                        <PENDING_COUNT>0</PENDING_COUNT>
                        <MIN_DURATION>0</MIN_DURATION>
                        <MAX_DURATION>0</MAX_DURATION>
                        <DURATION_SUM>0</DURATION_SUM>
                        <DURATION_SQR_SUM>0</DURATION_SQR_SUM>
                        <DET_COUNT>0</DET_COUNT>
                    </FAIL>
                </COMMAND>
```

M0rdecay:
Okay, we've discussed the case again - it looks like it's working very well now.

It will take me a little time for other tests.

srebhan:
One more question from my side: Currently the configuration options have inconsistent naming schemes e.g. "selected_nodes" (this should probably be "selected_metrics" anyway) and "field_selection". There are more inconsistencies...

Which naming scheme do you prefer?
@ssoroka, @M0rdecay  others?

M0rdecay:
> Which naming scheme do you prefer?

I like the current. After all, we first parse XML, and only then generate metrics.

M0rdecay:
I think the parser is flexible and functional. 

The only thing that scares me is the complexity of the configuration and the requests themselves. I didn't succeed the first time -_-

M0rdecay:
@srebhan you need to resolve merge conflicts.
Steven made a decision in favor of your implementation. We are waiting for the merger with the upstream.

plugins/parsers/xml/README.md
```diff
@@ -0,0 +1,344 @@
...
+      consumers = "Variable/@consumers"
+
+    ## Non-integer field definitions using XPath queries.
+    ## The field type is defined using XPath expressions such as number(), boolean() or string(). If no conversion is performed the field will be of type string.
+    [inputs.file.xml.fields]
+      temperature = "number(Variable/@temperature)"
+      power       = "number(Variable/@power)"
+      frequency   = "number(Variable/@frequency)"
+      ok          = "Mode != 'ok'"
+```
```
ssoroka, line 53: a lot of the other plugins make this explicit with `int_fields`, `string_fields`, `float_fields`, which seems like a decent pattern?
Note also tags can only ever be strings.

SimonMcN:
Thanks for all your work on this.  It will be useful for me.

M0rdecay:
@srebhan when are you planning to return to this plugin?
We would like to see it in 1.17

M0rdecay:
@srebhan @ssoroka release 1.17 is coming soon. Will this PR be in time?

ssoroka:
I'm testing this out locally and it seems that there are no warnings when your xpath doesn't find anything, so it seems really easy to end up with no metric and no clear reason why. I think we need to help the user with xpath more, since this is such an integral part of this plugin

M0rdecay:
Heeey, we have some progress!
Can we expect this parser in 1.18?

srebhan:
@M0rdecay this can only be answered by @ssoroka. :-) In the meantime you can check the [artifacts built by CircleCI](https://app.circleci.com/pipelines/github/influxdata/telegraf?branch=pull%2F8047) in the package category:
https://app.circleci.com/pipelines/github/influxdata/telegraf/3121/workflows/308e87c2-a2a3-49a0-9dd6-8ae73e832d63/jobs/75956/artifacts

M0rdecay:
I will it test this week
Thanks!

M0rdecay:
Well, so far so good
Honestly, I like it

@ssoroka when is this PR expected to merge?

Jagularr (approved):

reimda (approved):

reimda:
Merged in #8931.  Thanks @srebhan!

Issue #3520, opened 2017-11-28
Exec Input Plugin Parallel Processing? 


mm-manu:
Hello,
this is not an issue but rather a question.
In the simple example:
```
[[inputs.exec]]
  # Shell/commands array
  # Full command line to executable with parameters, or a glob pattern to run all matching files.
  commands = ["/tmp/test.sh", "/tmp/test2.sh", "/tmp/collect_*.sh"]

```
Do the scripts run in parallel or sequential?

danielnelson:
Parallel

mm-manu:
thanks

PR #8049, opened 2020-08-28
Update README.md

Alestrix:
Adding info from https://github.com/influxdata/telegraf/issues/3520. Thanks to this undocumented behavior I can throw away a bunch of data now.

### Required for all PRs:

- [ N/A ] Has appropriate unit tests.

### DIFF
--- a/plugins/inputs/exec/README.md
+++ b/plugins/inputs/exec/README.md
@@ -1,6 +1,6 @@
 # Exec Input Plugin
 
-The `exec` plugin executes the `commands` on every interval and parses metrics from
+The `exec` plugin executes all the `commands` in parralel on every interval and parses metrics from
 their output in any one of the accepted [Input Data Formats](https://github.com/influxdata/telegraf/blob/master/docs/DATA_FORMATS_INPUT.md).
 
 This plugin can be used to poll for custom metrics from any source.


ssoroka:
that seems like... unexpected behavior. I don't know how that could possibly do the right thing.

Alestrix:
The issue with a script file is that the resulting output would be multiple concatenated JSON strings, which is not a correct JSON string anymore. So I need to keep the commands separate and add delays in the hopes that the previous command is finished by then.

Running those in parallel is IMHO a horrible design decision.

Alestrix:
Shouldn't cause this line the commands to run after each other?
https://github.com/influxdata/telegraf/blob/6324b1fd630fb42232b82badaa05ed2acb4bbe09/plugins/inputs/exec/exec.go#L227

In my case the DSL speed tests to different servers definitely run in parallel (if I don't defer them with an added delay) as the results add up to the total available bandwidth.

ssoroka:
I'd be in favor of adding a configuration option to say whether you want them to run in parallel or not

srebhan:
> The issue with a script file is that the resulting output would be multiple concatenated JSON strings, which is not a correct JSON string anymore. [...]

What happens if you concatenate the JSON objects (`jobj1..N`) like this `[jobj1, jobj2, ...jobjN]`? This is valid JSON and the parser should be able to handle this IMO.

PR #8064, opened 2020-09-01
Add optional trim of field data for plugins/parsers/form_urlencoded

mjf:
### Required for all PRs:


reimda:
It looks like your input data is urlencoded but also has spaces at the beginning and/or end that you want to get rid?  You can already do this by adding a strings processor and configuring it as a trim.  You can have it act on all fields with the setting field = "*".  See https://github.com/influxdata/telegraf/blob/release-1.15/plugins/processors/strings/README.md

If that doesn't work for you we'll need to have you open an issue and give a better description of your problem before we'll accept a PR.  There are also merge conflicts in this PR that need to be fixed.

mjf:
@reimda We use `[[http_listener_2]]` and devices using `GET` method. These devices are buggy and are sending numeric values as space-padded text in field datum. Therefor we need to have the field datum sanitized before it reaches the `form_urlencoded` parses. That's exactly what my patch does. We also want if just for some of the listeners we use.

mjf:
Moved here #8148.

PR #8065, opened 2020-09-01
Exec with tags

pyriand3r:
### Required for all PRs:

We needed the feature to add additional tags to metrics right in telegraf before pushing the data to the influxdb. As it is a very general feature we wanted to share it with you in the hope of getting this into upstream.

With this feature you can add a `tags` array to a `[input.exec]` configuration. Within this tags array you can add an array for every command in which you define arrays for tags that are added automatically to the metrics of that command. Tag lines correspond to the lines of commands, so first tag array is applied to first commands metrics and so on. 

Example: 

```
commands = [
  "/tmp/test.sh",
  "/usr/bin/mycollector --foo=bar",
  "/tmp/collect_*.sh"
]

## additional parameters that will be added to the metrics as tags
## Lines correspond to the command lines, so first line of parameters are 
## added to metrics of first command and so on. If one command should not have additional tags,
## add an empty array.
tags = [
  [ ],
  [ [ "tag_1", "value_tag_1" ], [ "tag_2", "value_tag_2" ] ],
  [ [ "tag_3", "value_tag_3" ] ]
]
```
First command `/tmp/test.sh` will get no tags, second command `/usr/bin/mycollector --foo=bar" will get `tag_1` and `tag_2` as tags and third command will get `tag_3`.

### DIFF
--- a/plugins/inputs/exec/README.md
+++ b/plugins/inputs/exec/README.md
@@ -16,6 +16,15 @@ This plugin can be used to poll for custom metrics from any source.
     "/tmp/collect_*.sh"
   ]
 
+  ## additional parameters that will be added to the metrics as tags
+  ## Lines correspond to the command lines, so first line of parameters are 
+  ## added to metrics of first command and so on. 
+  ## If one command should not have additional tags, add an empty array.
+  tags = [
+    [ ],
+    [ [ "tag_1", "value_tag_1" ], [ "tag_2", "value_tag_2" ] ],
+  ]
+ 
   ## Timeout for each command to complete.
   timeout = "5s"
 
--- a/plugins/inputs/exec/README.md
+++ b/plugins/inputs/exec/README.md
@@ -23,6 +23,7 @@ This plugin can be used to poll for custom metrics from any source.
   tags = [
     [ ],
     [ [ "tag_1", "value_tag_1" ], [ "tag_2", "value_tag_2" ] ],
+    [ [ "tag_3", "value_tag_3" ] ]
   ]
  
   ## Timeout for each command to complete.
--- a/plugins/inputs/exec/dev/telegraf.conf
+++ b/plugins/inputs/exec/dev/telegraf.conf
@@ -10,6 +10,11 @@
     "echo 'deal,computer_name=hostb message=\"stuff\" 1530654676316265790'",
   ]
 
+  tags = [
+    [ ["tag_1", "value_hosta_1"] ],
+    [ ["tag_1", "value_hostb_1"], ["tag_2", "value_hostb_2"] ]
+  ]
+
 [[processors.regex]]
   [[processors.regex.tags]]
     key = "computer_name"
--- a/plugins/inputs/exec/exec.go
+++ b/plugins/inputs/exec/exec.go
@@ -32,7 +32,8 @@ const sampleConfig = `
   ## add an empty array.
   tags = [
     [ ],
-    [ [ "tag_1", "value_tag_1" ], [ "tag_2", "value_tag_2" ] ],
+	[ [ "tag_1", "value_tag_1" ], [ "tag_2", "value_tag_2" ] ],
+	[ [ "tag_3", "value_tag_3" ] ]
   ]
 
   ## Timeout for each command to complete.
--- a/plugins/inputs/exec/exec.go
+++ b/plugins/inputs/exec/exec.go
@@ -26,6 +26,14 @@ const sampleConfig = `
     "/tmp/collect_*.sh"
   ]
 
+  ## additional parameters that will be added to the metrics as tags
+  ## Lines correspond to the command lines, so first line of parameters are 
+  ## added to metrics of first command and so on
+  tags = [
+    [ [ "tag_1", "value_tag_1" ], [ "tag_2", "value_tag_2" ] ],
+    [ [ "tag_1", "value_tag_1" ], [ "tag_2", "value_tag_2" ] ],
+  ]
+
   ## Timeout for each command to complete.
   timeout = "5s"
 
@@ -45,6 +53,7 @@ type Exec struct {
 	Commands []string
 	Command  string
 	Timeout  internal.Duration
+	Tags [][][]string
 
 	parser parsers.Parser
 
@@ -140,7 +149,7 @@ func removeCarriageReturns(b bytes.Buffer) bytes.Buffer {
 
 }
 
-func (e *Exec) ProcessCommand(command string, acc telegraf.Accumulator, wg *sync.WaitGroup) {
+func (e *Exec) ProcessCommand(command string, tags [][]string, acc telegraf.Accumulator, wg *sync.WaitGroup) {
 	defer wg.Done()
 	_, isNagios := e.parser.(*nagios.NagiosParser)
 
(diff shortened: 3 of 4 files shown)


srebhan:
Please rebase on master to fix the AppVeyor problem.

pyriand3r:
Okay, sorry, first i did the rebase wrong so i had to revert and redo. now it should be a clean rebase... and everything is green...

wz2b:
I really don't love the configuration design where you have parallel arrays and you try to match one against the other by index.  I prefer, in that case, to convert the individual rows to their own structs, then have an array of those structs that more obviously binds the command and the tag together.

For instance, you could create an 'extended' command list

```
commands_ext = [
  { command="/tmp/test.sh" },
  { command="/usr/bin/mycollector --foo=bar", tags=[ "value_tag_1", "value_tag"2] },
  { command="/tmp/collect_*.sh"", tags=["value_tag_3"] }
]
```

What that also does is *sets you up for future enhancement* - if we want to add something else to the command definition (for example metric name override, or a uid/gid for chuid()/chgrp(), etc.) you can do it without adding a _third_ parallel array.

Unfortunately, I don't think TOML supports any kind of polymorphism, so you can't replace the 'command' parameter without breaking backward compatibility, which would be bad.  But you could merge the two lists on execution.

You can ignore this comment unless there's some consensus around it - this is a personal preference.  When parallel arrays have 2-3 items in them, it's fine.  The real problem is when they have 10 or 20 things in them, and you're counting with your thumb to line them together when you're making a change, and you're counting rows on your screen using your index finger.  Maybe that doesn't bother others like it does me.

reimda:
You can already add tags to any input with the tags setting.  If you want different tags for each exec command, you should try separate [[inputs.exec]] sections.

[[inputs.exec]]
commands = ["/tmp/test.sh"]

[[inputs.exec]]
commands = ["/usr/bin/mycollector --foo=bar"]
[inputs.exec.tags]
tag_1 = "value_tag_1"
tag_2 = "value_tag_2"

[[inputs.exec]
commands = ["/tmp/collect_*.sh"]
[inputs.exec.tags]
tag_3 = "value_tag_3"

Is there a reason why this config style won't work for you?

pyriand3r:
> Is there a reason why this config style won't work for you?

We are using the `exec` plugin extensivly and need for every command different tag values. We found Telegraf struggling with thousands of `[[inputs.exec]]` configurations, so we came up with this approach.

We know that the design is not the most pleasant one but it is backward compatible and won't break older configurations. Although it was the best we could imagine with the limitations of TOML

I find @wz2b 's idea very interesting, although i am not shure if this structure ist possible with TOML. If so, we could maybe add it as an additional field `command_extended` that can be used along `command`? Just an idea... That would keep backward compatibility.

ssoroka:
> We found Telegraf struggling with thousands of [[inputs.exec]] configurations

Can you tell me more about what you mean by "struggling"? 
I would think your operating system would have trouble launching thousands of processes on a tight interval in general. You might want to look at something like execd, which expects daemon processes that start once and continue to run. 

Is there specifically a performance difference between reusing exec vs a separate exec per process?

wz2b:
> I find @wz2b 's idea very interesting, although i am not shure if this structure ist possible with TOML. If so, we could maybe add it as an additional field `command_extended` that can be used along `command`? Just an idea... That would keep backward compatibility.

That's one of the reasons I did what I did to execd but not to exec.  The startup string I added to execd only works if there's one command, not a list.  So to do the same thing with exec you'd have to either define a second type of command list (your example called it command_extended) that takes objects ... or go back to having parallel arrays, which I don't particularly like.  However, this sort of thing (where you have arrays and you have to synchronize the indices to match them up) is elsewhere in telegraf already, so maybe I should get off this soapbox.

wz2b:
> > We found Telegraf struggling with thousands of [[inputs.exec]] configurations
> Can you tell me more about what you mean by "struggling"?
> I would think your operating system would have trouble launching thousands of processes on a tight interval in general. You might want to look at something like execd, which expects daemon processes that start once and continue to run.

I think what he is saying is that telegraf is trying to launch all of those thousand instances of [[input.exec]] concurrently, and what he's saying is that he doesn't want or need them to occur in parallel.  So instead, he made a single [[input.exec]] with a list of commands, which causes them to run sequentially.  However, that removes his ability to specify distinct tags for each command.

@pyriand3r please correct me if I got this wrong.

pyriand3r:
> I would think your operating system would have trouble launching thousands of processes on a tight interval in general.

That is true. The system wasn't able to handle all these exec commands aka processes. So we tried around and found this config suitable. I can't imagine that starting thousands of daemons would be better though? Okay there would be less overhead of creating and terminating threads. Allthough we exec nagios checks with our execs, so i cannot think of a way we would benefit from execd... but maybe i am missing something here ;)

> I think what he is saying is that telegraf is trying to launch all of those thousand instances of [[input.exec]] concurrently, and what he's saying is that he doesn't want or need them to occur in parallel. So instead, he made a single [[input.exec]] with a list of commands, which causes them to run sequentially. However, that removes his ability to specify distinct tags for each command.

Correct

ssoroka:
I see what you're getting at. Maybe we can provide an alternate syntax that allows for better configuration per-command. Let me get back to you on that.

pyriand3r:
Finaly i found time to update the pull request. I have added a `command_extended` config to the exec command. It looks like this:

```
commands_extended = [
    { command = "/tmp/test.sh", tags = [ ["tag_1", "value_1"], ["tag_2", "value_2"] ] },
    { command = "/usr/bin/mycollector --foo=bar", tags = [ ["tag_1", "value_2"], [ "tag_3", "value_3" ] ] }
  ]
```
every entry has a `command` and a `tags` field. Tags are still an array of arrays as it is not possible to write a map[string]string like this:

```
{ "key" = "value }
```
The burntSushi TOML parser is not capable of reading this, allthough it should be possible in TOML as far as I know. I would have prefered key-value syntax as it is more readable, but atm it seems not possible.

In the background the exec command merges all command fields ( `command`, `commands` and `commands_exended`) into a single slice of commandExtended structs to unify the further processing of the commands.

Hope this is a better approach. Should be fully downwards compatible as i have changed as little as possible. In addition it is possible to add more functionality/fields to the CommandExtended struct if needed in the future ;)

pyriand3r:
okay i don't know why the test fails. I have go 1.16.3 on arch linux and all tests are good... I have run them several time now, no errors... :(

pyriand3r:
So i found the failure. Although i fixed it by changing my tests it assumes to me, that the `AssertContainsField` method in the `testutil.Accumulator` has a bug. I had a slice with two different metrics in it, checked for one of it. If it was the first in the slice the check went well, if the other one was first `assert.Equal` would throw an error. The test will continue and find the correct metric but due to the error the test will fail. Seems to me a bug, but maybe it is wanted behaviour? As there are no function comments i only can guess...

pyriand3r:
Any comments for this? :)

Hipska:
> > I would think your operating system would have trouble launching thousands of processes on a tight interval in general.
> 
> That is true. The system wasn't able to handle all these exec commands aka processes. So we tried around and found this config suitable. I can't imagine that starting thousands of daemons would be better though? Okay there would be less overhead of creating and terminating threads. Allthough we exec nagios checks with our execs, so i cannot think of a way we would benefit from execd... but maybe i am missing something here ;)
> 
> > I think what he is saying is that telegraf is trying to launch all of those thousand instances of [[input.exec]] concurrently, and what he's saying is that he doesn't want or need them to occur in parallel. So instead, he made a single [[input.exec]] with a list of commands, which causes them to run sequentially. However, that removes his ability to specify distinct tags for each command.
> 
> Correct

What about playing with `collection_jitter` to make sure the execs are not all executed on the same time?

PR #8092, opened 2020-09-07
NATS Streaming output plugin

RashadAnsari:
### Required for all PRs:


plugins/outputs/stan/stan.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	if s.Secure {
+		tlsConfig, err := s.ClientConfig.TLSConfig()
+		if err != nil {
+			return err
+		}
+
+		natsOpts = append(natsOpts, nats.Secure(tlsConfig))
+	}
```
ssoroka, line 92: All the setup stuff belongs in an `Init() error` function, you don't want it repeated each connect.

plugins/outputs/stan/stan_test.go
```diff
@@ -0,0 +1,33 @@
...
+	}
+
+	// Verify that we can connect to the NATS Streaming daemon
+	err := n.Connect()
+	require.NoError(t, err)
+
+	// Verify that we can successfully write data to the NATS Streaming daemon
+	err = n.Write(testutil.MockMetrics())
+	require.NoError(t, err)
+}
```
ssoroka, line 33: I'd like to see some kind of -short test, otherwise we'll likely not catch it if this breaks.

ssoroka (changes requested):
looks pretty good. Would like to see more tests for it.

ssoroka:
Looks like there's some package versioning issues.

Kuzyashin:
Any updates here? How can I help speed up approving this PR ?

plugins/outputs/stan/stan_test.go
Kuzyashin: This tests looks like tests for other outputs.

sspaink:
It looks like NATS Streaming service is deprecated (https://github.com/nats-io/nats-streaming-server) is there still a reason for this plugin? It seems it was replaced by a service called Jetstream.

1995parham:
@sspaink Yes, I think it would be great if we can replace this plugin by jetstream which is based on nats.

sspaink:
Ok great I will go ahead and close this PR for now then, and if anyone would like to submit a Jetstream plugin pull request we can review it.

PR #8102, opened 2020-09-09
Fix execd processor test

pmalek-sumo:
On some occasions ingest part of the test (runCountMultiplierProgram)
does not manage to read all the metrics from stdin that are sent to it
before the process is stopped (and stdin closed) by the test. We can
then observe errors like "Error reading stderr" or "Error reading stdout"
(from the test process). This happens more often when the number of
metrics sent in the test increases.

Even less often, a deadlock might occur when the test does not consume
all the metrics and it waits on acc.Wait() for sync.Cond broadcast that
never arrives as in [1].

To fix this spin a separate goroutine which will check that accumulator
received all the expected metrics and only then stop the count
multiplier process.

[1]: https://app.circleci.com/pipelines/github/SumoLogic/telegraf/39/workflows/4a7fd631-25f9-4f99-8134-fe9ff267fc31/jobs/210

### Required for all PRs:

### DIFF
--- a/internal/process/process.go
+++ b/internal/process/process.go
@@ -78,6 +78,8 @@ func (p *Process) Stop() {
 		// signal our intent to shutdown and not restart the process
 		p.cancel()
 	}
+	// close stdin so the app can shut down gracefully.
+	p.Stdin.Close()
 	p.mainLoopWg.Wait()
 }
 
--- a/internal/process/process.go
+++ b/internal/process/process.go
@@ -78,8 +78,6 @@ func (p *Process) Stop() {
 		// signal our intent to shutdown and not restart the process
 		p.cancel()
 	}
-	// close stdin so the app can shut down gracefully.
-	p.Stdin.Close()
 	p.mainLoopWg.Wait()
 }
 
--- a/plugins/processors/execd/execd_test.go
+++ b/plugins/processors/execd/execd_test.go
@@ -50,9 +50,18 @@ func TestExternalProcessorWorks(t *testing.T) {
 		e.Add(m, acc)
 	}
 
-	acc.Wait(1)
-	require.NoError(t, e.Stop())
-	acc.Wait(9)
+	go func(t *testing.T) {
+		for {
+			n := acc.NMetrics()
+			if n == 10 {
+				if err := e.Stop(); err != nil {
+					fmt.Fprintf(os.Stderr, "failed to stop execd: %v", err)
+				}
+				return
+			}
+		}
+	}(t)
+	acc.Wait(10)
 
 	metrics = acc.GetTelegrafMetrics()
 	m := metrics[0]
--- a/testutil/accumulator.go
+++ b/testutil/accumulator.go
@@ -55,6 +55,8 @@ func (a *Accumulator) NMetrics() uint64 {
 }
 
 func (a *Accumulator) GetTelegrafMetrics() []telegraf.Metric {
+	a.Lock()
+	defer a.Unlock()
 	metrics := []telegraf.Metric{}
 	for _, m := range a.Metrics {
 		metrics = append(metrics, FromTestMetric(m))


pmalek-sumo:
Actually I've noticed there are other issues related to this which I'd like to point out:
* `GetTelegrafMetrics` (which is used in this test) requires a lock, example of a failed CI with a data race stack trace: https://github.com/influxdata/telegraf/blob/bb94c04bcda0f2b460d5519ac6933d3c4ff9bd74/testutil/accumulator.go#L57-L65
* `process.Process.Stop` doesn't need to explicitly call `p.Stdin.Close()` this is already handled when context gets canceled (via `p.canel()`). Otherwise we get 

```
2020/09/09 11:27:50 I! [] Starting process: /var/folders/78/xm1r7h7s7lqdp36w81v765500000gn/T/go-build471038421/b001/execd.test [-countmultiplier]
2020/09/09 11:27:51 E! [] Error reading stdout: read |0: file already closed
2020/09/09 11:27:51 E! [] Error reading stderr: read |0: file already closed
2020/09/09 11:27:51 I! [] Process /var/folders/78/xm1r7h7s7lqdp36w81v765500000gn/T/go-build471038421/b001/execd.test shut down
```

I've added those 2 changes in this PR. All tests seem to pass. Let me know if that seems OK.

internal/process/process.go
```diff
@@ -78,8 +78,6 @@ func (p *Process) Stop() {
 		// signal our intent to shutdown and not restart the process
 		p.cancel()
 	}
-	// close stdin so the app can shut down gracefully.
-	p.Stdin.Close()
```
ssoroka, line 82: It's important to notify the process that it needs to stop. You don't want the first notification to the process that it needs to stop to be trying to terminate the process, otherwise you drop metrics and don't allow it to finish up the work it's doing. Instead you need to close the pipe, give it time to process what's in process, and then close gracefully before being sent termination signals. We can't remove this.

plugins/processors/execd/execd_test.go
```diff
@@ -50,9 +50,18 @@ func TestExternalProcessorWorks(t *testing.T) {
...
 	}
 
-	acc.Wait(1)
-	require.NoError(t, e.Stop())
-	acc.Wait(9)
+	go func(t *testing.T) {
+		for {
+			n := acc.NMetrics()
+			if n == 10 {
+				if err := e.Stop(); err != nil {
```
ssoroka, line 57: stop can only be called once.

ssoroka (changes requested):
The lock is good, but the stdin close is required for graceful termination. The whole point of the test is to read one metric, call stop, then read the remaining.

plugins/processors/execd/execd_test.go
pmalek-sumo: It is, isn't it? It's an endless loop that upon reaching the desired metrics count - 10 - stops the process.
Note the return below.

internal/process/process.go
pmalek-sumo: Ok, fair point. Nonetheless, this way the error in the test: `Error reading stdout: read |0: file already closed` will not go away.

plugins/processors/execd/execd_test.go
ssoroka: ah, yes, sorry, I misread it.

internal/process/process.go
ssoroka: error reading **stdout** though, sounds like something else, not this stdin pipe.

pmalek-sumo: Yes, but the error does go away when the `p.Stdin.Close()` is removed.

I've reverted the change now but I'd say that this might require a closer look.

For future reference: the error is coming from: https://github.com/influxdata/telegraf/blob/de313fcde6c59376208a07ca9ace88aecaaf42a7/plugins/processors/execd/execd.go#L134

which is basically reading what `runCountMultiplierProgram()` prints out
https://github.com/influxdata/telegraf/blob/da1cb18df788a5b7a522c34b41e13c1264c8672b/plugins/processors/execd/execd_test.go#L104-L145

and when the stdin is being closed I get an error in https://github.com/influxdata/telegraf/blob/da1cb18df788a5b7a522c34b41e13c1264c8672b/plugins/processors/execd/execd_test.go#L110-L113

ssoroka: it goes away because you're no longer shutting down the process after reading the first metric, but this is what I'm intending and testing. This should work: 
- write 10 metrics
- read 1
- close() _(stop the plugin)_
- read the remaining 9 metrics

I believe the real issue is that in some cases the `cmdReadOut` goroutine+func isn't running until after the process is already shutting down, so the read runs into the error. Ideally I'd like to wait until that goroutine starts before triggering shutdown.

Ignoring this specific error is an option but then we're potentially missing stdout/stderr output.

What OS are you using? I'm assuming you're able to reliably reproduce the failure?

pmalek-sumo: I use the following one liner which reliably gives me an error (with a passing test) every 2-3 seconds or so.

I'm running this on mac ATM.

ssoroka: Ok I did some debugging and I was seeing this error come up at the end of the stream after the data was already read. If that's the case, it's not a real error at all, it's just indicating that the pipe is closed, and can be safely ignored

ssoroka: this is good. If the pipe was being closed before the data was read, we'd actually see the test fail, so we can be pretty confident it's doing the right thing.

ssoroka: Okay... so I'm still seeing cases where some of the metrics are missing afterwards. Maybe the pipe's being closed with metrics still in the pipe? I'm missing something.

pmalek-sumo: > I'm still seeing cases where some of the metrics are missing afterwards
I'll try to take a look at this tomorrow. Can you share your approach of testing this?

ssoroka: I'm using `go test github.com/influxdata/telegraf/plugins/processors/execd -count=100 -v -timeout 60s`, and adjusting count for about 15s of runtime.

Here's the diff of my workspace with local changes. It's not quite right as it's still failing periodically, though maybe not as much. https://gist.github.com/ssoroka/fe9c604f6b3cee6dc965948d43c20ee5

I'm not a fan of the sleep, I'm just testing with it. There might be something going on where the process ends before stdout has been completely read and it tears down the pipes or something. Not sure. I'm under the impression that this should work without changing the basic structure of the test (and really it does like 99.9% of the time).

ssoroka: (I also noticed that `acc.Wait(9)` should be `acc.Wait(10)` as it's the total count, not the count since last call.

pmalek-sumo: After look at it again I do believe that this not a problem (the error returned at https://github.com/influxdata/telegraf/blob/de313fcde6c59376208a07ca9ace88aecaaf42a7/plugins/processors/execd/execd.go#L134)

When `p.Stdin.Close()` is called the loop in the test https://github.com/influxdata/telegraf/blob/da1cb18df788a5b7a522c34b41e13c1264c8672b/plugins/processors/execd/execd_test.go#L109-L113 is terminated via a `return` and my suspicion is that sometimes this will be faster than the stderr/stdout read in `cmdReadErr`/`cmdReadOut` where it will return `file already closed`.

I'd say it is safe to ignore this specific error. As a safeguard one could add a check whether the process has already returned and if so then don't report `os.ErrClosed` like so:

```
func (e *Execd) cmdReadOut(out io.Reader) {
	scanner := bufio.NewScanner(out)

	for scanner.Scan() {
		metrics, err := e.parser.Parse(scanner.Bytes())
		if err != nil {
			e.Log.Errorf("Parse error: %s", err)
		}

			e.acc.AddMetric(metric)
		}
	}

	if err := scanner.Err(); err != nil {
		ps := e.process.Cmd.ProcessState
		// If the process already returned don't report ErrClosed errors
		if (ps != nil && ps.Exited()) && (errors.Unwrap(err) == os.ErrClosed) {
			return
		}

		e.Log.Errorf("Error reading stdout: %s", err)
	}
}
```

ssoroka:
so I did actually want to test receiving one metric, calling stop, and verifying that the remaining metrics are still read. This change avoids that test scenario.
Maybe there's a better way to do that.

Issue #8085, opened 2020-09-04
Add more missing NVMe attributes to smart plugin


p-zak:
## Feature Request

We would like to add support for reading additional NVMe attributes to smart plugin.
New attributes from [NVMe Specification 1.4](https://nvmexpress.org/wp-content/uploads/NVM-Express-1_4-2019.06.10-Ratified.pdf) we want to add:

- Temperature Sensor 1
- Temperature Sensor 2
- Temperature Sensor 3
- Temperature Sensor 4
- Temperature Sensor 5
- Temperature Sensor 6
- Temperature Sensor 7
- Temperature Sensor 8
- Thermal Management Temperature 1 Transition Count
- Thermal Management Temperature 2 Transition Count
- Total Time For Thermal Management Temperature 1
- Total Time For Thermal Management Temperature 2

Moreover, we would like to add support for NVMe vendor specific attributes on an example of INTEL<sup>&reg;</sup> device in accordance with [INTEL<sup>&reg;</sup> SSD SMART metrics for NVMe devices](https://www.intel.com/content/www/us/en/support/articles/000056596/memory-and-storage.html).

Plan is to add support for reading following attributes:
- Program Fail Count
- Erase Fail Count
- Wear Leveling Count
- End to End Error Detection Count
- CRC Error Count
- Timed Workload, Media Wear
- Timed Workload, Host Reads
- Timed Workload, Timer
- Thermal Throttle Status
- Retry Buffer Overflow Counter
- PLL Lock Loss Count
- NAND Bytes Written
- Host Bytes Written

### Proposal:
In the near future we will propose PR which will add support for mentioned NVMe attributes.

### Current behavior:
Not all NVMe attributes (in accordance to NVMe Specification 1.4) are still supported. There is no support for vendor specific attributes at all.

### Desired behavior:
Smart plugin should support most of NVMe attributes (in accordance to NVMe Specification 1.4) and vendor specific attributes.

### Use case: 
To have support for most NVMe attributes (in accordance to NVMe Specification 1.4) and vendor specific attributes.

What do you think?

p-zak:
Merged in https://github.com/influxdata/telegraf/issues/8085

PR #8113, opened 2020-09-11
Add more missing NVMe attributes to smart plugin

p-zak:
### Required for all PRs:

- [x] Signed [CCLA](https://www.influxdata.com/legal/ccla/).

CCLA completed and signed by Intel Corporation (I hope ICLA is not required since this github account is business account).

This PR solves https://github.com/influxdata/telegraf/issues/8085

## Feature Request
We would like to add support for reading additional NVMe attributes to smart plugin.
New attributes from [NVMe Specification 1.4](https://nvmexpress.org/wp-content/uploads/NVM-Express-1_4-2019.06.10-Ratified.pdf) we want to add:

- Temperature Sensor 1
- Temperature Sensor 2
- Temperature Sensor 3
- Temperature Sensor 4
- Temperature Sensor 5
- Temperature Sensor 6
- Temperature Sensor 7
- Temperature Sensor 8
- Thermal Management Temperature 1 Transition Count
- Thermal Management Temperature 2 Transition Count
- Total Time For Thermal Management Temperature 1
- Total Time For Thermal Management Temperature 2

Moreover, we would like to add support for NVMe vendor specific attributes on an example of INTEL<sup>&reg;</sup> device in accordance with [INTEL<sup>&reg;</sup> SSD SMART metrics for NVMe devices](https://www.intel.com/content/www/us/en/support/articles/000056596/memory-and-storage.html).

Plan is to add support for reading following attributes:
- Program Fail Count
- Erase Fail Count
- Wear Leveling Count
- End to End Error Detection Count
- CRC Error Count
- Timed Workload, Media Wear
- Timed Workload, Host Reads
- Timed Workload, Timer
- Thermal Throttle Status
- Retry Buffer Overflow Counter
- PLL Lock Loss Count
- NAND Bytes Written
- Host Bytes Written

Generic attributes from [NVMe Specification 1.4](https://nvmexpress.org/wp-content/uploads/NVM-Express-1_4-2019.06.10-Ratified.pdf) are still gathered by [smartctl](https://github.com/smartmontools/smartmontools) tool.
NVMe vendor specific attributes are optionally gathered by [nvme-cli](https://github.com/linux-nvme/nvme-cli) if it exists in OS.


plugins/inputs/smart/smart.go
```diff
@@ -146,31 +160,152 @@ var (
...
+	Excludes         []string
+	Devices          []string
+	UseSudo          bool
+	Timeout          internal.Duration
 }
 
 var sampleConfig = `
   ## Optionally specify the path to the smartctl executable
-  # path = "/usr/bin/smartctl"
+  # path_smartctl = "/usr/bin/smartctl"
```
ssoroka, line 291: need to provide backwards-support for "path" if we're going to change the name

plugins/inputs/smart/smart.go
```diff
@@ -491,12 +935,30 @@ func parseTemperature(fields, deviceFields map[string]interface{}, str string) e
...
+	}
+
+	fields["raw_value"] = temp
+
+	return nil
+}
+
 func init() {
+	// Set LC_NUMERIC to uniform numeric output from cli tools
+	_ = os.Setenv("LC_NUMERIC", "en_US.UTF-8")
```
ssoroka, line 951: a bit odd, but I imagine it's fine.

plugins/inputs/smart/smart.go
```diff
@@ -491,12 +935,30 @@ func parseTemperature(fields, deviceFields map[string]interface{}, str string) e
...
+
 	inputs.Add("smart", func() telegraf.Input {
 		m := NewSmart()
-		path, _ := exec.LookPath("smartctl")
-		if len(path) > 0 {
-			m.Path = path
+		pathSmartctl, _ := exec.LookPath("smartctl")
```
ssoroka, line 955: plugins can have an `Init() error` function. looking up the path names (and erroring if they're missing) might be better there.

ssoroka (changes requested):
Really awesome. Couple small change requests and I'm good to merge it.

plugins/inputs/smart/smart.go
zak-pawel: @ssoroka 
It depends. If I return any error in `Init() error`, Telegraf won't start. In approach we proposed (which is aligned with previous version), plugin won't gather anything but Telegraf (with other plugins) will work properly.

It is up to you :)

plugins/inputs/smart/smart.go
zak-pawel: Format of `smartctl` and `nvme` outputs could be based on `LC_NUMERIC` value. Setting it to `en_US.UTF-8` ensures us that it can be easily parsable independently from local settings in system.

p-zak: Format of `smartctl` and `nvme` outputs could be based on `LC_NUMERIC` value. Setting it to `en_US.UTF-8` ensures us that it can be easily parsable independently from local settings in system.

plugins/inputs/smart/smart.go
p-zak: @ssoroka 
It depends. If I return any error in `Init() error`, Telegraf won't start. In approach we proposed (which is aligned with previous version), plugin won't gather anything but Telegraf (with other plugins) will work properly.

It is up to you :)

plugins/inputs/smart/smart.go
p-zak: It would be easier to understand readme and code if there is just `path` for `smartctl` executable. But it would be inconsistent with `path_nvme`.

@ssoroka what do you think?

ssoroka: keep `path_smartctl`, but add an undocumented 
```go
Path string `toml:"path"`
```
to the app and if it's set, set PathSmartctl to that value.

Ideally all the fields settable from toml should have the toml tags defined.

plugins/inputs/smart/smart.go
p-zak: Did as you said! Thanks for guidance!

plugins/inputs/smart/smart.go
p-zak: Kept `path` undocumented for backwards-support as proposed. Added toml tags where as you suggested.

p-zak:
> Really awesome. Couple small change requests and I'm good to merge it.

@ssoroka Small changes delivered! If you are ok with them, you can merge whenever is fine for you 👍

ssoroka (approved):

Issue #25, opened 2015-06-23
Working demo


skynet:
Amazing solution but the documentation doesn't really go all the way through with a working example, which makes it relatively hard for a non-expert to start using it. Can someone fix that?

pauldix:
We'll be adding docs on that soon.

skynet:
:+1:

sparrc:
@skynet I believe this is solved, please re-open if that's not the case!

PR #8115, opened 2020-09-11
Fix Sumo Logic output plugin not splitting requests properly (#25)

pmalek-sumo:
This should fix 2 problems:

* sending the last request with empty body (the condition when calculating boundary was incorrect)
* sending requests bigger than `MaxRequstBodySize` (now it will only happen when it's configured so small so that it's not able to accomodate even one metric - a warning will be printed out in such case but the request will go through).

### Required for all PRs:


ssoroka (changes requested):
Once the requests are chunked, if any one succeeds, you need to **log** the error (not return it) and continue to attempt to deliver all the rest of the chunks. This means dropping the ones that don't deliver correctly, as there's no way to partially deliver a batch.

A better approach might be to ensure your batch size is small enough that you don't ever need to chunk, as once you start chunking your only options when a chunk fails to deliver is to: 1. Drop chunks that don't make it and accept the batch as delivered (returning no error), or 2. accept delivering duplicates and return an error, retrying the entire batch.

pmalek-sumo:
> A better approach might be to ensure your batch size is small enough that you don't ever need to chunk

Can you explain a bit more how should that work?

> 1. Drop chunks that don't make it and accept the batch as delivered (returning no error), or 2. accept delivering duplicates and return an error, retrying the entire batch.

I'd rather just do 1.

pmalek-sumo:
I've changed the logic to just log the error when a chunk fails to get sent.

On top of that: I have a patch that checks each request's body additionally to checking the request count, I can add it here or push a separate PR if that would be better.

plugins/outputs/sumologic/sumologic_test.go
```diff
@@ -592,26 +603,35 @@ func TestMaxRequestBodySize(t *testing.T) {
...
 	}
 
 	for _, tt := range testcases {
 		t.Run(tt.name, func(t *testing.T) {
+			var (
+				requestCount int
+				linesCount   int
+			)
+			ts.Config.Handler = http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				requestCount++
```
ssoroka, line 620: might need `atomic.AddInt32` / `atomic.LoadInt32` for these counters. It's a different goroutine, and isn't race safe.

ssoroka (changes requested):
Nearly there. two small change requests.

ssoroka (approved):

PR #8126, opened 2020-09-15
add influxdata slack icon to telegraf readme

sjwang90:


ssoroka (approved):
If you intended it to be next to the other buttons, it has to be on the same line. (but it's fine either way)

sjwang90:
Yeah looks a little too clutter all on the same line.  Cool gonna merge.

PR #8132, opened 2020-09-16
Sumo Logic output plugin: carbon2 default to include field in metric

pmalek-sumo:
This PR will make Sumo Logic output plugin fallback to include field in metric name when using Carbon2 metrics format without a specified `data_format` field

### Required for all PRs:


pmalek-sumo:
Would be super cool if we could also make this into [`v1.16.0`](https://github.com/influxdata/telegraf/milestone/77) together with https://github.com/influxdata/telegraf/pull/8115 as that would allow to ship the 'desired' default to users already with the very first version of this plugin.

plugins/serializers/carbon2/carbon2.go
```diff
@@ -93,6 +94,24 @@ func (s *Serializer) createObject(metric telegraf.Metric) []byte {
 	return m.Bytes()
 }
 
+func (s *Serializer) SetMetricsFormat(f format) {
+	s.metricsFormatLock.Lock()
```
ssoroka, line 98: are locks needed here?

ssoroka (approved):

plugins/serializers/carbon2/carbon2.go
pmalek-sumo: In general yes because we don't know how it will be called.
Perhaps calling `SetMetricsFormat` and `IsMetricsFormatUnset` from different goroutines will never happen but this is just to be sure.
It's not a hot path so my assumption is that it's ok.
Unless there's a specific concern here I'd leave it this way.

ssoroka: Unless you've specifically seen race conditions, this should be unnecessary. Plugins have their own dedicated goroutines and the interface functions are never called in parallel by Telegraf. We try to avoid this type of defensive programming, as it tends to be confusing and sets a bad example for other plugins.

pmalek-sumo: Got it. Removed.

ssoroka (approved):

PR #8133, opened 2020-09-16
Adding ready route to inputs.influxdb_v2_listener

Eraac:
### Required for all PRs:

Following this discussion https://github.com/influxdata/telegraf/pull/7828#issuecomment-692747173, that can be great if this come with 1.16 

All credits for @magichair, I only revert the commit who remove the ready route

Thanks!

### DIFF
--- a/plugins/inputs/influxdb_v2_listener/influxdb_v2_listener.go
+++ b/plugins/inputs/influxdb_v2_listener/influxdb_v2_listener.go
@@ -53,6 +53,7 @@ type InfluxDBV2Listener struct {
 	bytesRecv       selfstat.Stat
 	requestsServed  selfstat.Stat
 	writesServed    selfstat.Stat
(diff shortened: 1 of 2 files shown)


ssoroka:
this endpoint is deprecated and was specifically removed. Any comments on why it should be added back in?

magichair:
@ssoroka is there a deprecation note I'm missing? https://docs.influxdata.com/influxdb/v2.0/api/#tag/Ready still looks fully formed and planned for InfluxDB v2. The original "removal" was more a "don't put it in the first version of this plugin if we don't need it"-type choice.

ssoroka:
Perhaps I misunderstood. I was under the impression it's not used anymore, so I'm curious where this use case is coming from. I'm not against adding this in.

ssoroka (approved):

PR #8144, opened 2020-09-17
fix: fix sticky field `running` of `procstat_lookup`for inputs/procstat

agebhar1:
On update the processes by the given PIDs, remove PIDs if they are not alive anymore.

### Required for all PRs:

### DIFF
--- a/plugins/inputs/procstat/procstat.go
+++ b/plugins/inputs/procstat/procstat.go
@@ -323,6 +323,10 @@ func (p *Procstat) updateProcesses(pids []PID, tags map[string]string, prevInfo 
 			if name, _ := info.Name(); name == "" {
 				continue
 			}
+			if exists, err := info.Exists(); err != nil || !exists {
+				// process has gone
+				continue
+			}
 			procs[pid] = info
 		} else {
 			proc, err := p.createProcess(pid)
--- a/plugins/inputs/procstat/process.go
+++ b/plugins/inputs/procstat/process.go
@@ -10,6 +10,7 @@ import (
 
 type Process interface {
 	PID() PID
+	Exists() (bool, error)
 	Tags() map[string]string
 
 	PageFaults() (*process.PageFaultsStat, error)
@@ -54,6 +55,14 @@ func NewProc(pid PID) (Process, error) {
 		tags:        make(map[string]string),
 	}
 	return proc, nil
+}
+
+func (p *Proc) Exists() (bool, error) {
+	exists, err := process.PidExists(int32(p.PID()))
+	if err != nil {
+		return false, err
+	}
+	return exists, nil
 }
 
 func (p *Proc) Tags() map[string]string {
--- a/plugins/inputs/procstat/procstat_test.go
+++ b/plugins/inputs/procstat/procstat_test.go
@@ -95,15 +95,21 @@ func (pg *testPgrep) FullPattern(_ string) ([]PID, error) {
 }
 
 type testProc struct {
-	pid  PID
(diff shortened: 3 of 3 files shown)


sspaink:
Thank you @agebhar1 for the pr, sorry for the slow response. I was wondering if you could provide more information why this change is necessary? Was the issue that you were receiving process info for ones that weren't running anymore?

agebhar1:
Hi @sspaink. Exactly, I received process info for already stopped processes.

plugins/inputs/procstat/procstat.go
```diff
@@ -323,6 +323,10 @@ func (p *Procstat) updateProcesses(pids []PID, tags map[string]string, prevInfo
 			if name, _ := info.Name(); name == "" {
 				continue
 			}
+			if _, err := p.createProcess(pid); err != nil {
```
sspaink, line 326: Perhaps it would be better to call this [PidExists](https://pkg.go.dev/github.com/shirou/gopsutil@v3.21.6+incompatible/process#PidExists) function instead? It seems createProcess calls [NewProcess](https://pkg.go.dev/github.com/shirou/gopsutil@v3.21.6+incompatible/process#NewProcess) which seems to do the same thing but we don't need the resulting process structure at all.

agebhar1:
At the end [NewProcess](https://github.com/shirou/gopsutil/blob/v3.21.6/process/process.go#L167) calls the same function as [PidExists](https://github.com/shirou/gopsutil/blob/v3.21.6/process/process.go#L187) which is one of the OS dependent `PidExistsWithContext` functions. But I'm totally with you to use `PidExists` instead.

agebhar1:
@sspaink I've update the PR. WDYT?

sspaink (approved):
lgtm!

plugins/inputs/procstat/procstat.go
```diff
@@ -323,6 +323,10 @@ func (p *Procstat) updateProcesses(pids []PID, tags map[string]string, prevInfo
 			if name, _ := info.Name(); name == "" {
 				continue
 			}
+			if exists, err := info.Exists(); err != nil || !exists {
+				// process has gone
+				continue
+			}
```
reimda, line 329: This does filter out pids that are no longer running. I'm wondering if returning non-running pids is really a problem though. I assume the pids we filter out here were running when the original query was run. In other words if we query pids, then one or more of them finishes before we get here, why should we drop them?

Are you using the data in a way that needs the pids to be running or is this intended to drop pids that have been finished for a long time?

reimda:
We recently added a CI check for commit message format. Since this PR only has one commit, the merged message comes from the commit message, not the PR title. Could you amend your commit to match the required format ("fix: ") or add another commit so the PR title will be used? Thanks

agebhar1:
Updated commit message.

plugins/inputs/procstat/procstat.go
agebhar1: In the exported data they are shown as 'alive'. Yes, we need the information that the process is gone.

reimda: The list of processes produced by the plugin is always going to be a little out of date because it's a snapshot in time. When the sample interval happens, the plugin checks which processes are running. Adding the .Exists() check here only modifies the snapshot.

I don't think it's valuable to double check with .Exists(). The snapshot was accurate when it was taken.

There are other use cases that don't want non-running pids to be removed like this. If we accept this change I suspect it will cause trouble for other users.

Issue #8142, opened 2020-09-17
 [processors.execd] Make the buffer size configurable


M0rdecay:
## Feature Request

Opening a feature request kicks off a discussion.

### Proposal:
Add the size of the buffer used in the [cmdReadOut](https://github.com/influxdata/telegraf/blob/master/plugins/processors/execd/execd.go#L119) function to the processor parameters

### Current behavior:
The buffer size is currently not configurable. For this reason, an attempt to get a large metric from an external processor fails with an error:
```
E! [processors.execd] Error reading stdout: bufio.Scanner: token too long
```

### Desired behavior:
The buffer size can be adjusted so that the processor can handle large metrics.

### Use case:
To work with XML, I have made a wrapper for the parser and work with it as with an external processor. The processor is written following the example from the [README](https://github.com/influxdata/telegraf/tree/master/plugins/processors/execd#example).
When processing large documents (~7 thousand lines), receiving data back from the processor fails. The reason is as follows - as a result of parsing, the processor can return a metric of 70-80 KiB.

Also, if this error occurs, the processor stops working and does not process the following metrics, which could fit in the default buffer.

ssoroka:
Wow. That's a large metric. It looks like line protocol limits a single field to 64 KB anyway. You might run into other problems even if we expand it, but I'm not against expanding the buffer size available for the metric. the change should be pretty straightforward, and doesn't necessarily have to be configurable.

something like:
```
	scanner := bufio.NewScanner(out)
	scanBuf := make([]byte, 4_000) // default size
	scanner.Buffer(scanBuf, 300_000) // max size
```

M0rdecay:
I will try to do it, test it, and, if it works, prepare a PR.
What do you think?

M0rdecay:
Yeah, this works as expected.

This means that the error was not related to a protocol limitation - otherwise, I think, I would have received an error in the external processor even at the serialization stage.

PR #8121, opened 2020-09-13
Alternative implementation of XML parser

M0rdecay:
### Required for all PRs:

- [x] Signed [CLA](https://influxdata.com/community/cla/).
- [x] Associated README.md updated.
- [x] Has appropriate unit tests.

@ssoroka, @srebhan, please take a look.

This is a slightly different implementation of the parser from PR https://github.com/influxdata/telegraf/pull/7460.
This version is closer to the json parser.

This parser, on the one hand, is less flexible than other proposed solutions, on the other hand, due to naming similar to that used in the json parser, it does not lose data if the names of nodes or attributes are repeated.

I would like to know your opinion.

PR #8145, opened 2020-09-17
Configurable execd processor buffer size

M0rdecay:
### Required for all PRs:

Closes https://github.com/influxdata/telegraf/issues/8142
Also related with https://github.com/influxdata/telegraf/pull/8121

### DIFF
--- a/plugins/processors/execd/README.md
+++ b/plugins/processors/execd/README.md
@@ -31,6 +31,10 @@ Telegraf minimum version: Telegraf 1.15.0
 
   ## Delay before the process is restarted after an unexpected termination
   # restart_delay = "10s"
+
+  ## Buffer size for storing metrics received from the process in bytes
+  ## The minimum allowed value is 4096
+  # buffer_size = 65536
 ```
 
 ### Example
--- a/plugins/processors/execd/README.md
+++ b/plugins/processors/execd/README.md
@@ -32,7 +32,7 @@ Telegraf minimum version: Telegraf 1.15.0
   ## Delay before the process is restarted after an unexpected termination
   # restart_delay = "10s"
 
-  ## Buffer size for storing metrics received from the process in bytes
+  ## Buffer size in bytes for storing metrics received from the process
   ## The minimum allowed value is 4096
   # buffer_size = 65536
 ```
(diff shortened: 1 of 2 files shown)


plugins/processors/execd/execd.go
```diff
@@ -23,13 +23,20 @@ const sampleConfig = `
 
   ## Delay before the process is restarted after an unexpected termination
   restart_delay = "10s"
+
+  ## Buffer size in bytes for storing metrics received from the process
+  ## The minimum allowed value is 4096
+  buffer_size = 65536
```
ssoroka, line 29: this is actually the buffer size used for storing a single metric, not multiple metrics.

ssoroka (changes requested):
I'm still not sure I like that it's configurable. we should just do the right thing. Too many knobs and dials to configure is just confusing.

This could literally be a two line change that works for 100% of people. No configuration options.

M0rdecay:
> I'm still not sure I like that it's configurable. we should just do the right thing. Too many knobs and dials to configure is just confusing.
> 
> This could literally be a two line change that works for 100% of people. No configuration options.

What about these values?
Hope this will be enough...
```
	scanner := bufio.NewScanner(out)
	scanBuf := make([]byte, 4096) // default size
	scanner.Buffer(scanBuf, 262144) // max size
```

ssoroka (approved):
Looks good.

ssoroka:
You may run into size issues in other places, but at least this won't be one of them.

M0rdecay:
> You may run into size issues in other places, but at least this won't be one of them.

Yes, I understand, but for now we prefer to use an external processor for working with XML - until a built-in parser appears in the telegraf.

I tested processing larger documents (10-12 thousand lines), everything is fine with the new buffer.
Also, there are no errors when processing such metrics when using an assembly with the parser inside the agent.

Many thanks for the help!

plugins/processors/execd/README.md
```diff
@@ -31,6 +31,10 @@ Telegraf minimum version: Telegraf 1.15.0
 
   ## Delay before the process is restarted after an unexpected termination
   # restart_delay = "10s"
+
+  ## Buffer size in bytes for storing metrics received from the process
+  ## The minimum allowed value is 4096
+  # buffer_size = 65536
```
meseeif: its the same when execd  used as inputs plugin？

PR #8157, opened 2020-09-20
Add support for fields and protocol lookups in port_name.

a-bali:
Improve functionality of the `port_name` plugin by:
- allowing source to be provided as a field (not only as a tag)
- output will be provided as a field if input was a field as well
- additional tag/field can be provided to specify the protocol (tcp/udp)

Documentation and tests are updated accordingly.

### Required for all PRs:

### DIFF
--- a/plugins/processors/port_name/port_name.go
+++ b/plugins/processors/port_name/port_name.go
@@ -124,9 +124,9 @@ func (d *PortName) Apply(metrics ...telegraf.Metric) []telegraf.Metric {
 					d.Log.Errorf("Unexpected type %t in source field; must be string or int", v)
 					continue
 				case int64:
-					portProto = strconv.FormatInt(field.(int64), 10)
+					portProto = strconv.FormatInt(v, 10)
 				case string:
-					portProto = field.(string)
+					portProto = v
 				}
 				fromField = true
 			}
@@ -173,7 +173,7 @@ func (d *PortName) Apply(metrics ...telegraf.Metric) []telegraf.Metric {
 					d.Log.Errorf("Unexpected type %t in protocol field; must be string", v)
 					continue
 				case string:
-					proto = field.(string)
+					proto = v
 				}
 			}
 		}
--- a/plugins/processors/port_name/port_name.go
+++ b/plugins/processors/port_name/port_name.go
@@ -125,6 +125,8 @@ func (d *PortName) Apply(metrics ...telegraf.Metric) []telegraf.Metric {
 					continue
 				case int64:
 					portProto = strconv.FormatInt(v, 10)
+				case uint64:
+					portProto = strconv.FormatUint(v, 10)
 				case string:
 					portProto = v
 				}
--- a/plugins/processors/port_name/port_name.go
+++ b/plugins/processors/port_name/port_name.go
@@ -15,12 +15,20 @@ var sampleConfig = `
 [[processors.port_name]]
   ## Name of tag holding the port number
   # tag = "port"
+  ## Or name of the field holding the port number
+  # field = "port"
 
-  ## Name of output tag where service name will be added
+  ## Name of output tag or field (depending on the source) where service name will be added
   # dest = "service"
 
   ## Default tcp or udp
   # default_protocol = "tcp"
+
+  ## Tag containing the protocol (tcp or udp, case-insensitive)
+  # protocol_tag = "proto"
+
+  ## Field containing the protocol (tcp or udp, case-insensitive)
+  # protocol_field = "proto"
 `
 
 type sMap map[string]map[int]string // "https" == services["tcp"][443]
@@ -29,8 +37,11 @@ var services sMap
 
 type PortName struct {
 	SourceTag       string `toml:"tag"`
+	SourceField     string `toml:"field"`
 	DestTag         string `toml:"dest"`
 	DefaultProtocol string `toml:"default_protocol"`
+	ProtocolTag     string `toml:"protocol_tag"`
+	ProtocolField   string `toml:"protocol_field"`
 
 	Log telegraf.Logger `toml:"-"`
 }
@@ -40,7 +51,7 @@ func (d *PortName) SampleConfig() string {
 }
 
 func (d *PortName) Description() string {
-	return "Given a tag of a TCP or UDP port number, add a tag of the service name looked up in the system services file"
+	return "Given a tag/field of a TCP or UDP port number, add a tag/field of the service name looked up in the system services file"
 }
 
 func readServicesFile() {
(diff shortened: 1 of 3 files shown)


plugins/processors/port_name/port_name.go
```diff
@@ -97,11 +108,26 @@ func readServices(r io.Reader) sMap {
...
+		var fromField bool
+
+		if len(d.SourceTag) > 0 {
+			if tag, ok := m.GetTag(d.SourceTag); ok {
+				portProto = string([]byte(tag))
+			}
+		}
+		if len(d.SourceField) > 0 {
+			if field, ok := m.GetField(d.SourceField); ok {
+				portProto = field.(string)
```
ssoroka, line 122: This will panic when the field is not a string. You should probably detect the type, support integers, and warn on fields that don't match the supported types.

plugins/processors/port_name/port_name.go
```diff
@@ -97,11 +108,26 @@ func readServices(r io.Reader) sMap {
 
 func (d *PortName) Apply(metrics ...telegraf.Metric) []telegraf.Metric {
 	for _, m := range metrics {
-		portProto, ok := m.GetTag(d.SourceTag)
-		if !ok {
-			// Nonexistent tag
+
+		var portProto string
+		var fromField bool
```
ssoroka, line 113: fromField isn't used.

plugins/processors/port_name/port_name.go
```diff
@@ -127,6 +153,17 @@ func (d *PortName) Apply(metrics ...telegraf.Metric) []telegraf.Metric {
...
 			proto = portProtoSlice[1]
 		}
+		if len(d.ProtocolTag) > 0 {
+			if tag, ok := m.GetTag(d.ProtocolTag); ok {
+				proto = tag
+			}
+		}
+		if len(d.ProtocolField) > 0 {
+			if field, ok := m.GetField(d.ProtocolField); ok {
+				proto = field.(string)
```
ssoroka, line 163: panic here too. Should instead check the type and warn if not string.

plugins/processors/port_name/port_name.go
```diff
@@ -151,7 +188,11 @@ func (d *PortName) Apply(metrics ...telegraf.Metric) []telegraf.Metric {
 			continue
 		}
 
-		m.AddTag(d.DestTag, service)
+		if fromField {
+			m.AddField(d.DestTag, service)
+		} else {
+			m.AddTag(d.DestTag, service)
```
ssoroka, line 194: rename DestTag to Dest because it's now a tag or a field.

ssoroka (changes requested):

plugins/processors/port_name/port_name.go
a-bali: It is used in the end to determine if the looked up service name is to be added as a tag or as a field.

reimda:
Hi @a-bali, thanks for the PR.

I wrote port_name to fill a specific need of the sflow input plugin so I'm curious how you are reusing it.  I'm not opposed to changing it but I want to have a better idea of the problem you're trying to solve.  Could you describe how you use port_name?  Where did your data with ports as fields come from?

Have you considered using the converter processor to change the field to tag or tag to field?

a-bali:
> I wrote port_name to fill a specific need of the sflow input plugin so I'm curious how you are reusing it. I'm not opposed to changing it but I want to have a better idea of the problem you're trying to solve. Could you describe how you use port_name? Where did your data with ports as fields come from?

Thank you for writing this plugin. I'm processing firewall logs with the tail input plugin and a grok pattern that parses the various fields in the log. I am using the port_name plugin (and others) to enrich/transform this data for reporting and visualization.
 
> Have you considered using the converter processor to change the field to tag or tag to field?

I guess that is an option, but it would require additional steps and I saw that other processor plugins (e.g. reverse_dns) also support using fields/tags in the input/output.  Also the PR contains functionality to use a dedicated field to specify the protocol (tcp/udp) which I have in the firewall log.

plugins/processors/port_name/port_name.go
```diff
@@ -119,7 +119,15 @@ func (d *PortName) Apply(metrics ...telegraf.Metric) []telegraf.Metric {
 		}
 		if len(d.SourceField) > 0 {
 			if field, ok := m.GetField(d.SourceField); ok {
-				portProto = field.(string)
+				switch v := field.(type) {
+				default:
+					d.Log.Errorf("Unexpected type %t in source field; must be string or int", v)
```
ssoroka, line 124: I think unsigned ints might be possible as well

plugins/processors/port_name/port_name.go
```diff
@@ -119,7 +119,15 @@ func (d *PortName) Apply(metrics ...telegraf.Metric) []telegraf.Metric {
...
+					continue
+				case int64:
+					portProto = strconv.FormatInt(field.(int64), 10)
+				case string:
+					portProto = field.(string)
```
ssoroka, line 129:
```suggestion
					portProto = v
```

plugins/processors/port_name/port_name.go
```diff
@@ -119,7 +119,15 @@ func (d *PortName) Apply(metrics ...telegraf.Metric) []telegraf.Metric {
 		}
 		if len(d.SourceField) > 0 {
 			if field, ok := m.GetField(d.SourceField); ok {
-				portProto = field.(string)
+				switch v := field.(type) {
+				default:
+					d.Log.Errorf("Unexpected type %t in source field; must be string or int", v)
+					continue
+				case int64:
+					portProto = strconv.FormatInt(field.(int64), 10)
```
ssoroka, line 127:
```suggestion
					portProto = strconv.FormatInt(v, 10)
```

plugins/processors/port_name/port_name.go
```diff
@@ -160,7 +168,13 @@ func (d *PortName) Apply(metrics ...telegraf.Metric) []telegraf.Metric {
 		}
 		if len(d.ProtocolField) > 0 {
 			if field, ok := m.GetField(d.ProtocolField); ok {
-				proto = field.(string)
+				switch v := field.(type) {
+				default:
+					d.Log.Errorf("Unexpected type %t in protocol field; must be string", v)
+					continue
+				case string:
+					proto = field.(string)
```
ssoroka, line 176:
```suggestion
					proto = v
```

ssoroka (commented):
This looks good, but there's a couple small changes, as well we should probably support uint64

plugins/processors/port_name/port_name.go
a-bali: OK, added.

ssoroka (approved):

PR #8165, opened 2020-09-22
[output.wavefront] Introduced "immediate_flush" flag

prydin:
### Required for all PRs:

Addresses an issue mentioned in #7409 concerning the "double buffering" done by the Wavefront SDK. The SDK is designed to send metrics asynchronously in order to minimize the delay experienced by a caller. This behavior is not ideal when used with Telegraf, since Telegraf will assume the connection to Wavefront can accept data very quickly. When the internal buffer in the Wavefront SDK overflows, it will drop metrics. Meanwhile, Telegraf will still think the metrics are flowing at a very high rate and will keep flooding the SDK. 

To avoid extensive redesign of the SDK, this PR adds a flag, ```immediate_flush``` that flushes the internal SDK buffers after each batch. This will ensure that the buffer never fills with more records than the batch size and also has the side effect of providing "back pressure" to the Telegraf core. Thanks to that back pressure, Telegraf will no longer be "tricked" into thinking the connection has virtually unlimited capacity and can take whatever the action it needs to manage the throughput gracefully.

### DIFF
--- a/plugins/outputs/wavefront/README.md
+++ b/plugins/outputs/wavefront/README.md
@@ -54,7 +54,7 @@ This plugin writes to a [Wavefront](https://www.wavefront.com) proxy, in Wavefro
   ## normally done by the Wavefront SDK. This can be used if you are experiencing buffer overruns. The sending 
   ## of metrics will block for a longer time, but this will be handled gracefully by the internal buffering in
   ## Telegraf.
-  #immediate_flush = false
+  #immediate_flush = true
 ```
 
 
--- a/plugins/outputs/wavefront/README.md
+++ b/plugins/outputs/wavefront/README.md
@@ -49,6 +49,12 @@ This plugin writes to a [Wavefront](https://www.wavefront.com) proxy, in Wavefro
   ## Truncate metric tags to a total of 254 characters for the tag name value. Wavefront will reject any 
   ## data point exceeding this limit if not truncated. Defaults to 'false' to provide backwards compatibility.
   #truncate_tags = false
+
+  ## Flush the internal buffers after each batch. This effectively bypasses the background sending of metrics
+  ## normally done by the Wavefront SDK. This can be used if you are experiencing buffer overruns. The sending 
+  ## of metrics will block for a longer time, but this will be handled gracefully by the internal buffering in
+  ## Telegraf.
+  #immediate_flush = false
 ```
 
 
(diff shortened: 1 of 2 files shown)


plugins/outputs/wavefront/README.md
```diff
@@ -49,6 +49,12 @@ This plugin writes to a [Wavefront](https://www.wavefront.com) proxy, in Wavefro
   ## Truncate metric tags to a total of 254 characters for the tag name value. Wavefront will reject any 
   ## data point exceeding this limit if not truncated. Defaults to 'false' to provide backwards compatibility.
   #truncate_tags = false
+
+  ## Flush the internal buffers after each batch. This effectively bypasses the background sending of metrics
+  ## normally done by the Wavefront SDK. This can be used if you are experiencing buffer overruns. The sending 
+  ## of metrics will block for a longer time, but this will be handled gracefully by the internal buffering in
+  ## Telegraf.
+  #immediate_flush = false
```
ssoroka, line 57: I don't think we should provide this as an option, and instead just always assume it's true.

ssoroka (commented):
This is great, but I think there's no reason to still support the buffer or allow configuration of the flush. Keep in mind Telegraf is already doing periodic flush (which is essentially the write method) based on batch size.

plugins/outputs/wavefront/README.md
prydin: I would agree, but I don't want to break backwards compatibility.

ssoroka: Seems like it was already broken. Maybe I'm missing something. Is the plugin in use elsewhere?

prydin: It's in use with hundreds of customers. I don't have a way of testing the behavior in an extremely busy environment and with heavily overloaded connections. I'm not comfortable changing the behavior without a way to roll back to the current behavior. Can we let it be a flag for now and deprecate it in the next release if we don't see any issues?

ssoroka: I see what you're getting at; I'm willing to leave it in temporarily. I'd like to default it to `true`, because I don't believe the `false` behavior works correctly for anyone.

prydin: Done. Can we merge this now?

ssoroka (approved):

PR #8179, opened 2020-09-25
Add support for datadog distributions metric

### LINKED ISSUE #8134: [inputs/statsd] Add support for Distribution metrics for compatibility with DataDog (DogStatsD)
## Feature Request

telegraf's statsd input plugin supports datadog statsd format with `parse_data_dog_tags=true` flag.
This is a great feature which allows using telegraf as a drop-in replacement for DataDog's agent.
Unfortunately, it does not support [Distributions](https://docs.datadoghq.com/metrics/distributions/#counting-distribution-metrics) metric type.

### Proposal:

Add support for distributions type of metrics to provide full compatibility with DataDog's statsd (DogStatsD).

### Current behavior:

When sending distributions metrics to telegraf, the metrics are dropped generating the following error in the logs:

```
Aug 10 08:55:44 mytest-i-02d65c1fc7f587007.sonic-us telegraf[3719]: 2020-08-10T08:55:44Z E! [inputs.statsd] Metric type "d" unsupported
```

### Desired behavior:
Same behaviour as in DogStatsD, ideally implemented the same way to provide identical behaviour. That is, send all raw data collected during the flush interval to the output like DogStatsD does.

### Use case: 

Seamless migration from DataDog.

ppsreejith:
### Required for all PRs:

Thanks for the awesome work you folks do :)

This PR is my attempt to add support for the [distributions](https://docs.datadoghq.com/developers/metrics/types/?tab=distribution) metric (alternate [link](https://docs.datadoghq.com/metrics/distributions/)) to maintain compatibility with Datadog metrics. This would allow us to aggregate metrics on the server rather than on the agent (i.e. aggregate on Influx rather than on Telegraf).

I've added unit tests and tested this on my local Influx and Telegraf setup. Let me know of all the changes required 👍 

Closes #8134

### DIFF
--- a/plugins/inputs/statsd/statsd.go
+++ b/plugins/inputs/statsd/statsd.go
@@ -770,9 +770,9 @@ func (s *Statsd) aggregate(m metric) {
 	switch m.mtype {
 	case "d":
 		cached := cacheddistributions{
-			name:   m.name,
+			name:  m.name,
 			value: m.floatvalue,
-			tags:   m.tags,
+			tags:  m.tags,
 		}
 		s.distributions = append(s.distributions, cached)
 	case "ms", "h":
--- a/plugins/inputs/statsd/statsd.go
+++ b/plugins/inputs/statsd/statsd.go
@@ -70,6 +70,11 @@ type Statsd struct {
 	// http://docs.datadoghq.com/guides/dogstatsd/
 	DataDogExtensions bool `toml:"datadog_extensions"`
 
+	// Parses distribution metrics in the datadog statsd format.
+	// Requires the DataDogExtension flag to be enabled.
+	// https://docs.datadoghq.com/developers/metrics/types/?tab=distribution#definition
+	DataDogDistributions bool `toml:"datadog_distributions"`
+
 	// UDPPacketSize is deprecated, it's only here for legacy support
 	// we now always create 1 max size buffer and then copy only what we need
 	// into the in channel
@@ -244,6 +249,10 @@ const sampleConfig = `
 
   ## Parses datadog extensions to the statsd format
   datadog_extensions = false
+
+  ## Parses distributions metric as specified in the datadog statsd format
+  ## https://docs.datadoghq.com/developers/metrics/types/?tab=distribution#definition
+  datadog_distributions = false
 
   ## Statsd data translation templates, more info can be read here:
   ## https://github.com/influxdata/telegraf/blob/master/docs/TEMPLATE_PATTERN.md
@@ -769,12 +778,14 @@ func (s *Statsd) aggregate(m metric) {
 
 	switch m.mtype {
 	case "d":
-		cached := cacheddistributions{
-			name:  m.name,
-			value: m.floatvalue,
-			tags:  m.tags,
+		if s.DataDogExtensions && s.DataDogDistributions {
+			cached := cacheddistributions{
+				name:  m.name,
+				value: m.floatvalue,
+				tags:  m.tags,
+			}
+			s.distributions = append(s.distributions, cached)
 		}
-		s.distributions = append(s.distributions, cached)
 	case "ms", "h":
 		// Check if the measurement exists
 		cached, ok := s.timings[m.hash]
--- a/plugins/inputs/statsd/statsd.go
+++ b/plugins/inputs/statsd/statsd.go
@@ -615,7 +615,7 @@ func (s *Statsd) parseStatsdLine(line string) error {
 
 		// Validate metric type
 		switch pipesplit[1] {
-		case "g", "c", "s", "ms", "h":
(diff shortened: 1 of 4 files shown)


sjwang90:
Awesome! Thanks @ppsreejith. We currently have a backlog of PR reviews for our next 1.16 release which we want out soon. We'll get this reviewed shortly after.

gerilya:
> Awesome! Thanks @ppsreejith. We currently have a backlog of PR reviews for our next 1.16 release which we want out soon. We'll get this reviewed shortly after.

@sjwang90 as v1.16 went out approx one month ago, I am wondering whether you have an ETA for merging this nice feature into the master branch. Thank you

plugins/inputs/statsd/statsd.go
```diff
@@ -254,6 +262,11 @@ func (s *Statsd) Gather(acc telegraf.Accumulator) error {
 	defer s.Unlock()
 	now := time.Now()
 
+	for _, m := range s.distributions {
+		acc.AddFields(m.name, m.fields, m.tags, now)
+	}
+	s.distributions = []cacheddistributions{}
```
srebhan, line 268: How about
```suggestion
	s.distributions = make([]cacheddistributions, 0)
```

plugins/inputs/statsd/statsd.go
```diff
@@ -322,6 +335,7 @@ func (s *Statsd) Start(ac telegraf.Accumulator) error {
 	s.counters = make(map[string]cachedcounter)
 	s.sets = make(map[string]cachedset)
 	s.timings = make(map[string]cachedtimings)
+	s.distributions = []cacheddistributions{}
```
srebhan, line 338: Same here to keep the same style as above
```suggestion
	s.distributions = make([]cacheddistributions, 0)
```

plugins/inputs/statsd/statsd.go
```diff
@@ -735,6 +751,15 @@ func parseKeyValue(keyvalue string) (string, string) {
...
 func (s *Statsd) aggregate(m metric) {
 	switch m.mtype {
+	case "d":
+		fields := make(map[string]interface{})
+		fields[m.field] = m.floatvalue
+		cached := cacheddistributions{
+			name:   m.name,
+			fields: fields,
+			tags:   m.tags,
+		}
```
srebhan, line 761: You are always setting just one field, why do you use a map for this? You could also store the name and value instead of allocating over and over.

srebhan (changes requested):
Have some comments in the code, nothing big. Please also add the new option to the `sampleConfig` variable!

srebhan:
@ppsreejith any news?

ppsreejith:
@srebhan Sorry for the (really) long pause (Life's been hectic 😅). I've some time now so I'll pick this up this week 👍

plugins/inputs/statsd/statsd.go
ppsreejith: So I updated the `cacheddistributions` interface and replaced field with value. The repeated allocation now happens in the `Gather` method where the accumulator's `addFields` requires a map argument.

ppsreejith:
@srebhan I've addressed the above comments :+1: Let me know if there are more changes.

> Please also add the new option to the sampleConfig variable!

Do you mean the `sampleConfig` defined in the `plugins/inputs/statsd/statsd.go` file? I'm not sure what changes would be required here.

ppsreejith:
Also, small side note: I set up a [playground](https://github.com/ppsreejith/telegraf-sandbox) to make it easier to test out Telegraf changes end-to-end (In case this makes it easier for others).

srebhan:
@ppsreejith to be honest I can't remember what I meant with 

> Please also add the new option to the sampleConfig variable!

:-) However, as you add new fields it might be good to give the user a chance to enable or disable the new field(s). We want to keep the output series stable between versions by default. Is there a way to enable/disable the _distribution_ fields? If not, could you please add a flag to the config where those fields are disabled by default!?

ppsreejith:
@srebhan Got it. I'll add this 👍

ppsreejith:
@srebhan I've added a new config flag `datadog_distributions` to enable parsing the distributions metric.

The distributions metric will only be parsed if both the `datadog_extensions` config _and_ the `datadog_distributions` config are enabled. My reasoning is that since the distributions metric type _is_ a datadog extension, we should only enable parsing if the `datadog_extensions` config is enabled. Furthermore, to keep the option of enabling/disabling this feature, I've added the new `datadog_distributions` config.

I've documented this config in the `sampleConfig` variable as well as the Readme file (with links to Datadog's documentation). I've also extended the tests to cover this config enabled parsing. Let me know if this looks good or if there are other changes required :+1:

srebhan (approved):
Thank you very much for your effort! Looks good to me!

PR #8183, opened 2020-09-27
Fix metric SetAggregate didn't work

stitchcula:
Fixes the bug that metric SetAggregate didn't work, when using sync.Pool to reserve metrics from outputs.

### Required for all PRs:
- [x] Signed CLA.

### DIFF
--- a/metric/metric.go
+++ b/metric/metric.go
@@ -252,7 +252,7 @@ func (m *metric) Copy() telegraf.Metric {
 }
 
 func (m *metric) SetAggregate(b bool) {
-	m.aggregate = true
+	m.aggregate = b
 }
 
 func (m *metric) IsAggregate() bool {
--- a/metric/metric_test.go
+++ b/metric/metric_test.go
@@ -334,9 +334,21 @@ func TestValueType(t *testing.T) {
 	assert.Equal(t, telegraf.Gauge, m.Type())
 }
 
+func TestSetAggregate(t *testing.T) {
+	m1 := baseMetric()
+	m1.SetAggregate(true)
+	assert.True(t, m1.IsAggregate())
+	m1.SetAggregate(false)
+	assert.False(t, m1.IsAggregate())
+}
+
 func TestCopyAggregate(t *testing.T) {
 	m1 := baseMetric()
 	m1.SetAggregate(true)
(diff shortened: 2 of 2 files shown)


reimda:
SetAggregate is a deprecated interface.  See https://github.com/influxdata/telegraf/blob/57cd20a26c01c47854bd0dd9975e4f962929748d/metric.go#L128

It's not used in telegraf any more.  Can you share more detail why you need it to work?

stitchcula:
> SetAggregate is a deprecated interface. See
> 
> https://github.com/influxdata/telegraf/blob/57cd20a26c01c47854bd0dd9975e4f962929748d/metric.go#L128
> 
> It's not used in telegraf any more. Can you share more detail why you need it to work?

In my fork, I use sync.Pool to reuse metrics, just like:

```
type reservable struct {
	telegraf.Metric
}

var pool = &sync.Pool{New: func() interface{} {
	m, _ := metric.New("", nil, nil, time.Time{})
	return &reservable{Metric: m}
}

// call it to get blank Metric in parser plugin
func NewMetric() telegraf.Metric {
	return pool.Get().(*reservable)
}

func reserve(m *reservable) {
	m.SetName("")
	m.SetTime(zeroTime)
	m.SetAggregate(false) // it always set aggregate true

	//...

	pool.Put(m)
}

//...

func (m *metric) Accept() {
	reserve(m)
}

func (m *metric) Drop() {
	reserve(m)
}
```
In function `reserve`, `m.SetAggregate(false)` **always set aggregate true**, because of an obvious mistake in metric.go:

> 
> https://github.com/influxdata/telegraf/blob/57cd20a26c01c47854bd0dd9975e4f962929748d/metric/metric.go#L255
> 

As well, reusing metric can significantly improve the pressure of GC. Maybe I can start a new pull request?

reimda:
Since SetAggregate() is deprecated and not used anywhere I would rather remove it along with IsAggregate() and metric.aggregate rather than fix it and add tests for it.

srebhan:
@stitchcula any progress on this?

sspaink:
Closing due to no response, feel free to re-open if you are still interested in completing this pr. Thanks!

PR #8186, opened 2020-09-28
New sql server queries (Azure) - refactoring and formatting

Trovalo:
### Required for all PRs:

I've done some formatting also for the Azure queries part.
There are no breaking changes, but a pair of edits to align some queries with the on-prem ones.
- sqlAzureMIRequests --> added [session_db_name]
- sqlAzureMIRequests and sqlAzureDBRequests --> added COLASCE for [open_transaction]

Some testing, especially for the Managed instance section is needed (I have an Azure SQL DB but no Managed Instance)
@denzilribeiro


reimda:
It would be better to put formatting style and behavior changes in separate PRs.  The code behavior changes are important for a reviewer to check but putting them in with formatting/style changes make reviewing like finding a needle in a haystack.

@denzilribeiro Can you review this?

Trovalo:
> It would be better to put formatting style and behavior changes in separate PRs. The code behavior changes are important for a reviewer to check but putting them in with formatting/style changes make reviewing like finding a needle in a haystack.
> 
> @denzilribeiro Can you review this?

I'm aware of that, but those are just small things that got lost in the big PR of splitting the different sections of the plugin.
For actual structural changes I'll make different PRs

plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -4,93 +4,112 @@ import (
...
-                cast(avg_login_rate_percent as float) as avg_login_rate_percent  ,
-                end_time,
-                cast(avg_instance_memory_percent as float) as avg_instance_memory_percent ,
-                cast(avg_instance_cpu_percent as float) as avg_instance_cpu_percent
-        FROM
-                sys.dm_db_resource_stats WITH (NOLOCK)
-        ORDER BY
-                end_time DESC
+const sqlAzureDBResourceStats string = `
+SET DEADLOCK_PRIORITY -10;
```
denzilribeiro, line 12: What is the nead for Deadlock_priority - given DMVs they are memory locations not updatable. I had removed many of them :)... do you see any explicit need for them?

plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -4,93 +4,112 @@ import (
...
+IF SERVERPROPERTY('EngineEdition') <> 5 BEGIN /*not Azure SQL DB*/
+	DECLARE @ErrorMessage AS nvarchar(500) = 'Telegraf - the instance "'+ @@SERVERNAME +'" is not an Azure SQL DB. Check the database_type parameter in the telegraf configuration.';
```
denzilribeiro, line 14: Can we change the message thrown to include DB name? on Azure DB there are many SQL Databases.. in particular... this is an example of the MI message -- giving just a logical server doesn't help customer identify which DB.. This is accross the board I think in Azure and on-prem
	DECLARE @ErrorMessage AS nvarchar(500) = 'Telegraf - Connection string Server:"'+ @@SERVERNAME + ',Database:' + DB_NAME() +'" is not an Azure Managed Instance. Check the database_type parameter in the telegraf configuration.';

denzilribeiro (changes requested):
Made 2 comments that are accross the board
a. Do we need deadlock priority? These are DMVs not updatable really
b. Error message change to include DB_Name as well as just a logical server on Azure DB in connection string won't help identify what connection string is in the wrong section

Ex:
	DECLARE @ErrorMessage AS nvarchar(500) = 'Telegraf - Connection string Server:"'+ @@SERVERNAME + ',Database:' + DB_NAME() +'" is not an Azure Managed Instance. Check the database_type parameter in the telegraf configuration.';

plugins/inputs/sqlserver/azuresqlqueries.go
Trovalo: my bad, I placed it everywhere when I added the check on the engine edition (I mad the code block once and then pasted it everywhere)

plugins/inputs/sqlserver/azuresqlqueries.go
Trovalo: Makes sense for Azure DB since the DB is part of the identifier, I'm not sure about Managed instance, but for the on-prem that's not really useful since DB_NAME() will probably just return "master" all the time as database name.

I'll add it in the Azure section as you asked

denzilribeiro: You need it everywhere.. think of when you put an Azure DB connection string in say an on-prem section or a managed instance section

plugins/inputs/sqlserver/azuresqlqueries.go
Trovalo: I had a look at this, DEADLOCK_PRIORITY was present in 11 queries out of 17 for the Azure part and in 9 out of 9 for the on-prem.
I'll remove it in case queries use only dmv

plugins/inputs/sqlserver/azuresqlqueries.go
Trovalo: I've edited the error message to include the Database name.

plugins/inputs/sqlserver/azuresqlqueries.go
Trovalo: I've removed some DEADLOCK_PRIORITY (where only dmv were referenced), it is now used only in DatabaseIO and PerfCounters queries (both azure DB and MI)

denzilribeiro (approved):
Looks good!

Trovalo:
@ssoroka can you merge this?

ssoroka (approved):

Issue #1027, opened 2016-04-14
Extend redis plugin capabilities


mathieuboniface:
We are massively using redis lists (and telegraf too). We would like export using telegraf the sizes of thoses lists stored into redis instances. 

Currently, the redis input plugin gather metrics from the `INFO` command. And there is no way to get metrics stored into redis database.

For us, a good solution could be to permit the redis plugin to gather metrics from command given in the configuration file, something like this : 

``` toml
[[inputs.redis]]
  servers = ["tcp://localhost:6379"]
  commands = ["LLEN myList", "GET myKey", "CLUSTER COUNT-FAILURE-REPORTS myNode"]
```

WDYT ?

radcheb:
+1

sparrc:
is this the same as a redis subscriber plugin? https://github.com/influxdata/telegraf/issues/759

mathieuboniface:
Hi sparcc, and thanks for your quick reply.

#759 is focused on pub/sub mechanism of redis. This new feature is more like the [postgresql_extended](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/postgresql_extensible) plugin principle. The idea is to permit the execution of some commands (queries) to redis.

UserNotFound:
@mathieuboniface Did you find a solution? I am about to just starting writing something using the telegraf exec input, and some scripting using `redis-cli llen {listname}`

athoune:
you need something more dynamic like http_json.

A list of requests is not enough. You need a key per request (with mandatory number value), and handling hash response. With hash, you can handle dynamic key/value.

Here is the hash example for _sidekiq_, for watching queues length :

```
EVAL "local a={}; local queues=redis.call('smembers', 'resque:gitlab:queues'); for i,v in ipairs(queues) do a[i*2-1] = 'queue:' .. v; a[i*2] = redis.call('llen', 'resqueue:gitlab:queue:' .. v) end; return a" 0
```
Do you need some PR ?

danielnelson:
We do need someone to put together a pull request still, but let's finalize an example plugin configuration first.  Can you mock this up and show how the output would look given some input?

athoune:
Something like that :
```toml
# Read Redis's basic status information
[[inputs.redis]]
  ## specify servers via a url matching:
  ##  [protocol://][:password]@address[:port]
  ##  e.g.
  ##    tcp://localhost:6379
  ##    tcp://:password@192.168.99.100
  ##
[...]

PR #8196, opened 2020-09-29
feat: add functionality to get values from redis commands

yquansah:
### Required for all PRs:

This PR is an attempt to pull additional metrics on redis from arbitrary commands. The logic in this PR accounts for types (`string`, `integer`, and `float`), and we could add more types.

Issue addressed: https://github.com/influxdata/idpe/issues/8711
Community issue Addressed: https://github.com/influxdata/telegraf/issues/1027

### DIFF
--- a/plugins/inputs/redis/redis.go
+++ b/plugins/inputs/redis/redis.go
@@ -19,7 +19,7 @@ import (
 
 type RedisCommand struct {
 	Command []interface{}
-	Key     string
+	Field   string
 	Type    string
 }
 
@@ -51,29 +51,13 @@ func (r *RedisClient) Do(returnType string, args ...interface{}) (interface{}, e
 
 	switch returnType {
 	case "integer":
-		intVal, err := rawVal.Int64()
-		if err != nil {
-			return nil, err
-		}
-		return intVal, nil
+		return rawVal.Int64()
 	case "string":
-		strVal, err := rawVal.String()
-		if err != nil {
-			return nil, err
-		}
-		return strVal, nil
+		return rawVal.String()
 	case "float":
-		floatVal, err := rawVal.Float64()
-		if err != nil {
-			return nil, err
-		}
-		return floatVal, nil
+		return rawVal.Float64()
 	default:
-		strVal, err := rawVal.String()
-		if err != nil {
-			return nil, err
-		}
-		return strVal, nil
+		return rawVal.String()
 	}
 }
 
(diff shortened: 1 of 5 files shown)


yquansah:
Was a little unsure of how exactly to go about editing README

plugins/inputs/redis/redis.go
```diff
@@ -38,6 +46,37 @@ type RedisClient struct {
...
+func (r *RedisClient) Do(returnType string, args ...interface{}) (interface{}, error) {
+	rawVal := r.client.Do(args...)
+
+	switch returnType {
+	case "integer":
+		intVal, err := rawVal.Int64()
+		if err != nil {
+			return nil, err
+		}
+		return intVal, nil
```
ssoroka, line 58:
```suggestion
		return rawVal.Int64()
```

plugins/inputs/redis/redis.go
```diff
@@ -64,6 +103,11 @@ var sampleConfig = `
   ## If no port is specified, 6379 is used
   servers = ["tcp://localhost:6379"]
 
+  ## Optional. Specify redis commands to retrieve values, under key "commands"
+  # command = ["get", "sample-key"]
+  # key = "sample-key-value"
+  # type = "string"
+
```
ssoroka, line 110: this looks wrong. Don't you need to specify `[[inputs.redis.commands]]` ahead of this?

ssoroka: consider naming "key" `field` instead.

ssoroka: changes should be copied into readme.

ssoroka (changes requested):
Looks good. The only thing I'd say is that this might be better off as a new plugin. Already you're going to have a hard time changing the metric name if you don't like `redis_commands`. Also not clear how to control which commands go into which metric, which I guess is not possible currently. That said, I'm not necessarily against merging it into the existing plugin.

reimda:
> Was a little unsure of how exactly to go about editing README

Describe how to use the feature you added.  Give examples so people unfamiliar with the feature will be able to understand how to configure it and know what data it will produce.

plugins/inputs/redis/redis.go
yquansah: This is addressed! Thank you for the catch, let me know if there is anything else needed.

plugins/inputs/redis/redis_test.go
```diff
@@ -7,11 +7,27 @@ import (
...
 	"github.com/influxdata/telegraf/testutil"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
+type testClient struct {
+}
+
+func (t *testClient) BaseTags() map[string]string {
+	return map[string]string{"host": "redis.net"}
```
ssoroka, line 20: why is this hardcoded?

yquansah: Oh just for testing purposes, i am testing the logic around external calls. The struct needs this method to adhere to the interface

ssoroka: ah, it's just a test. never mind.

ssoroka (approved):

PR #8200, opened 2020-09-30
SqlServer - Fixed Query mapping

Trovalo:
### Required for all PRs:

For reasons I got to test the Managed instance section of the queries, and I've found a mismatch between the available queries listed in the README.md and what was actually mapped in the code. 
This means that those queries will never be called if the name used is the one specified in the readme

A mismatch was also discovered for Azure SQL DB.
Those are the queries that have been fixed
- AzureSQLDBOsWaitstats
- AzureSQLMIRequests

@denzilribeiro

### DIFF
--- a/plugins/inputs/sqlserver/README.md
+++ b/plugins/inputs/sqlserver/README.md
@@ -78,7 +78,7 @@ GO
   ## - AzureSQLDBResourceGovernance
   ## - AzureSQLDBDatabaseIO
   ## - AzureSQLDBServerProperties
-  ## - AzureSQLDBSQLOsWaitstats
+  ## - AzureSQLDBOsWaitstats
   ## - AzureSQLDBMemoryClerks
   ## - AzureSQLDBPerformanceCounters
   ## - AzureSQLDBRequests
@@ -92,7 +92,7 @@ GO
    ## - AzureSQLMIOsWaitstats 
    ## - AzureSQLMIMemoryClerks
    ## - AzureSQLMIPerformanceCounters
-   ## - AzureSQLMIDBRequests
+   ## - AzureSQLMIRequests
    ## - AzureSQLMISchedulers
 
    ## database_type =  SQLServer by default collects the following queries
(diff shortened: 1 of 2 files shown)


denzilribeiro (approved):

denzilribeiro (approved):
Just readme could have been changed correct yep was a Readme typo

Trovalo:
> Just readme could have been changed correct yep was a Readme typo

Yes, but I was not sure about it, so I've applied the "standard" naming in a strict way (Prefix + QueryName)

reimda:
Changing both doc and code to AzureSQLMIRequests may be more standard, but if anyone was using AzureSQLMISqlRequests in their config before in include_query or exclude_query, this change will break their config and require them to figure out what the new value is.  After a quick read of the filter code, I think if it's in include_query it won't be included as before and if it's in exclude_query it won't be excluded as before, and there won't be any warning to the user that the behavior is opposite.

Why not just change the doc to AzureSQLMISqlRequests and preserve backward compatibility?

Trovalo:
> include_query

That's because this new set of features (query splitted by version) has not reached any released version of Telegraf yet, if someone is using it, then they are using custom builds or nightly versions.
As of now, I see no problems in fixing and editing this kind of behavior.

About backward compatibility, the old queries are still there and will keep working unedited. That "old" version is still the default  configuration proposed by the telegraf executable, see:
https://github.com/influxdata/telegraf/blob/bfdb0d5c83e46d7e65faae5112f2ba148e4e4fb2/plugins/inputs/sqlserver/sqlserver.go#L41

Trovalo:
What do we want to do in the end? @denzilribeiro @reimda

To me the PR is ok as it is but let me know

denzilribeiro:
> What do we want to do in the end? @denzilribeiro @reimda
> 
> To me the PR is ok as it is but let me know

This is a fairly new change so am ok with it, if it was later on that would be different

Trovalo:
@ssoroka can you merge this?

PR #8201, opened 2020-09-30
Fix carbon2 serializer not falling through to field separate when carbon2_format field is unset

pmalek-sumo:
When `carbon2_format` is unspecified or is specified as empty string there's a missing `fallthrough` which makes carbon2 serializer not produce metric names at all (noop case in `switch`).

This PR addresses that and adds a handful of tests for that case.

### Required for all PRs:

### DIFF
--- a/plugins/serializers/carbon2/carbon2_test.go
+++ b/plugins/serializers/carbon2/carbon2_test.go
@@ -39,11 +39,6 @@ func TestSerializeMetricFloat(t *testing.T) {
 			expected: fmt.Sprintf("metric=cpu field=usage_idle cpu=cpu0  91.5 %d\n", now.Unix()),
 		},
 		{
-			// Fall back to separate field name
-			format:   Carbon2FormatFieldEmpty,
-			expected: fmt.Sprintf("metric=cpu field=usage_idle cpu=cpu0  91.5 %d\n", now.Unix()),
-		},
-		{
 			format:   Carbon2FormatMetricIncludesField,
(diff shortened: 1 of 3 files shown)


plugins/serializers/carbon2/carbon2_test.go
```diff
@@ -38,6 +38,11 @@ func TestSerializeMetricFloat(t *testing.T) {
 			format:   Carbon2FormatFieldSeparate,
 			expected: fmt.Sprintf("metric=cpu field=usage_idle cpu=cpu0  91.5 %d\n", now.Unix()),
 		},
+		{
```
reimda, line 41: It doesn't make sense to me to duplicate the tests here.  Empty doesn't provide different functionality than Separate.  It would be simpler to change NewSerializer so that if metricsFormat argument is empty string then it returns a Serializer with metricsformat Carbon2FormatFieldSeparate.  Then you could remove Carbon2FormatFieldEmpty altogether.  Then you'd only need a single test that NewSerializer("") returns a Serializer with metricsFormat of Carbon2FormatFieldSeparate.

reimda (changes requested):

plugins/serializers/carbon2/carbon2_test.go
pmalek-sumo: Ok, sure.
So I've made `Carbon2FormatFieldEmpty` unexported (i.e. `carbon2FormatFieldEmpty`) and with this the only place where we need to check for empty string is the constructor as it's impossible to pass `format("")` even in `SetMetricsFormat()` because of `format` being unexported as well.

PR #8202, opened 2020-09-30
A new Logz.io output plugin

idohalevi:
### Required for all PRs:


idohalevi:
@p-zak @ssoroka ready for the review :)

plugins/outputs/logzio/README.md
```diff
@@ -0,0 +1,43 @@
+# Logz.io Output Plugin
+
+This plugin sends metrics to Logz.io over HTTPs.
+
+### Configuration:
+
+```toml
+# A plugin that can send metrics over HTTPs to Logz.io
+[[outputs.http]]
```
ssoroka, line 9: this looks wrong.

plugins/outputs/logzio/logzio.go
```diff
@@ -0,0 +1,171 @@
...
+		lg.SetDrainDuration(drainDuration),
+		lg.SetTempDirectory(l.QueueDir),
+		lg.SetUrl(l.URL),
+	)
+
+	if err != nil {
+		return fmt.Errorf("[logzio] failed to create new logzio sender: %v", err)
+	}
+
+	log.Printf("I! [logzio] Successfuly created Logz.io sender: %s %s %s %d\n", l.URL, l.QueueDir,
```
ssoroka, line 98: add 
```go
Log Telegraf.Logger `toml:"-"`
```
to the struct and use that instead. Telegraf will initialize it for you, and you can use testutil.Logger{} in tests.
This uses the Telegraf logger interface which will set warning level and plugin name for you.

plugins/outputs/logzio/logzio.go
```diff
@@ -0,0 +1,171 @@
...
+		l.Token,
+		lg.SetCheckDiskSpace(l.CheckDiskSpace),
+		lg.SetDrainDiskThreshold(l.DiskThreshold),
+		lg.SetDrainDuration(drainDuration),
+		lg.SetTempDirectory(l.QueueDir),
+		lg.SetUrl(l.URL),
+	)
+
+	if err != nil {
+		return fmt.Errorf("[logzio] failed to create new logzio sender: %v", err)
```
ssoroka, line 95: don't include the plugin name in errors.

plugins/outputs/logzio/logzio_test.go
```diff
@@ -0,0 +1,35 @@
...
+	require.Equal(t, l.URL, defaultLogzioURL)
+
+	require.Equal(t, l.Token, "")
+}
+
+func TestConnectAndWrite(t *testing.T) {
+	l := CreateDefultLogizoOutput()
+	l.Token = "123456789"
+
+	err := l.Connect()
```
ssoroka, line 23: would be best if it didn't actually connect to a server here and instead mocked out an http server and tested the format.

ssoroka (changes requested):

plugins/outputs/logzio/README.md
idohalevi: @ssoroka nice catch thank you!

plugins/outputs/logzio/logzio.go
idohalevi: @ssoroka I think it wasn't part of the telegraf project when I first sent the PR. I keep getting errors when I try to use Telegraf logger locally. Any ideas why?

image

plugins/outputs/logzio/logzio_test.go
idohalevi: @ssoroka I think this can be problematic since you need to have an account at logz.io to test it with real server

ssoroka: just this one remaining item to get it merged. :)

idohalevi: @ssoroka thank you! I changed the logic a bit. The sending portion is tested in the sender so instead of testing it here I've extracted the parsing part and tested that it's valid

idohalevi:
@ssoroka what do you think?

ssoroka:
It's an improvement, but you still have other tests connecting to the logz.io service. I was kind of expecting a `net/http/httptest` `httptest.NewServer` ([example](https://golang.org/pkg/net/http/httptest/#example_Server)) that would check the request and mock the response.

idohalevi:
@ssoroka I think I'm missing something, which test is connecting to the Logz.io server? The connect function just go to initializeSender but not actually connecting to Logz.io's server

ssoroka (changes requested):
Ok, so I took a closer look at everything and I have two main concerns. 
1) There's no test of the Write() function. If anything was broken in there, you probably wouldn't know, but at least the code is simple right now.
2) I looked closer at the Send() function from https://github.com/logzio/logzio-go/blob/master/logziosender.go, and I noticed that it has its own internal buffer, which competes with Telegraf's buffer, so essentially you've double buffered the output at twice the cost of memory for it. Not only this, but when the second logzio buffer is full, you drop metrics, meaning users can't control the buffer size from the config like other plugins anymore. The fix for this is to remove the use of logzio-go and just use a simple http connection to write the messages out. 90% of that package is buffering anyway, and you can just copy out the one http write function that you need. 

Let me know what you think and if you have any questions.

plugins/outputs/logzio/logzio.go
```diff
@@ -0,0 +1,176 @@
...
+	return sampleConfig
+}
+
+// Write takes in group of points to be written to the Output
+func (l *Logzio) Write(metrics []telegraf.Metric) error {
+	if len(metrics) == 0 {
+		return nil
+	}
+
+	l.Log.Debugf("Recived %d metrics", len(metrics))
```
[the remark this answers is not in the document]
idohalevi: I didn't include a test for the Write() because most of it is tested in the logziosender.go file. I'll remove logziosender and add a simple HTTP client + tests :)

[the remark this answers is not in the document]
idohalevi: @ssoroka done

idohalevi:
@ssoroka I've changed the implementation to use a default HTTP client

Doron-Bargo:
Hi @ssoroka 
Any update on this PR?

ssoroka (approved):
I think you're good now. I kind of don't like that the connect function is used as an initializer (there's an `Init() error` function signature you can use for that), but you'd still need an empty `Connect()` function if you did that, so, I'll leave it up to you. Let me know when you're ready to merge.

idohalevi:
@ssoroka we're ready to merge :)

Issue #7710, opened 2020-06-21
[Logging] Sort the loaded inputs, aggregators, processors, and outputs 


FixTestRepeat:
## Feature Request

telegraf (v1.14.3) currently doesn't sort the various loaded inputs, aggregators, processors, and outputs.  This makes troubleshooting more onerous.

### Proposal:
When Telegraf starts, the modules loaded should be logged in a sorted fashion. 

This should be applied to all loads:

- inputs
- aggregators
- outputs
- inputs

### Current behavior:
5 successive reloads of the telegraf service yield different order of the loaded modules:

> for i in `seq 1 5` ; do sudo systemctl restart telegraf.service ; sleep 2 ; systemctl status telegraf.service | grep Loaded ; done
   Loaded: loaded (/lib/systemd/system/telegraf.service; enabled; vendor preset: enabled)
Jun 21 12:36:34 raspiM telegraf[25946]: 2020-06-21T02:36:34Z I! Loaded inputs: cpu disk diskio mem net socket_listener processes swap system
Jun 21 12:36:34 raspiM telegraf[25946]: 2020-06-21T02:36:34Z I! Loaded aggregators:
Jun 21 12:36:34 raspiM telegraf[25946]: 2020-06-21T02:36:34Z I! Loaded processors:
Jun 21 12:36:34 raspiM telegraf[25946]: 2020-06-21T02:36:34Z I! Loaded outputs: influxdb_v2
   Loaded: loaded (/lib/systemd/system/telegraf.service; enabled; vendor preset: enabled)
Jun 21 12:36:36 raspiM telegraf[25973]: 2020-06-21T02:36:36Z I! Loaded inputs: **diskio mem net system socket_listener cpu disk processes swap**
Jun 21 12:36:36 raspiM telegraf[25973]: 2020-06-21T02:36:36Z I! Loaded aggregators:
Jun 21 12:36:36 raspiM telegraf[25973]: 2020-06-21T02:36:36Z I! Loaded processors:
Jun 21 12:36:36 raspiM telegraf[25973]: 2020-06-21T02:36:36Z I! Loaded outputs: influxdb_v2
   Loaded: loaded (/lib/systemd/system/telegraf.service; enabled; vendor preset: enabled)
Jun 21 12:36:39 raspiM telegraf[25999]: 2020-06-21T02:36:39Z I! Loaded inputs: **disk diskio mem net processes cpu swap system socket_listener**
[...]

PR #8218, opened 2020-10-02
Sort plugin names in logs on load

jonathanlloyd:
Previously, plugins were printed non-deterministically in the log.
This made it difficult to compare active plugins between runs.

This commit formats the plugin list in a deterministic order for easy comparison.

### Required for all PRs:

Closes: #7710

### DIFF
--- a/cmd/telegraf/telegraf.go
+++ b/cmd/telegraf/telegraf.go
@@ -180,10 +180,19 @@ func runAgent(ctx context.Context,
 		return ag.Test(ctx, wait)
 	}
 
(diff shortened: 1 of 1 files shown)


ssoroka (changes requested):
I think this was already done. See PluginNameCounts in master. I think it was resolved in https://github.com/influxdata/telegraf/pull/8127 and we missed closing the issue. Does this PR cover anything else?

jonathanlloyd:
Ah, missed that - never mind!

PR #8232, opened 2020-10-06
add minimum versions to plugins

sjwang90:


ssoroka:
I find this a bit weird. In my mind, plugins don't really have a minimum version, just a version where they were added. If you took any plugin and ran it on an older version of Telegraf (you'd have to build it yourself), there's usually a good chance it will work, since the plugin interface almost never changes (we do sometimes add new plugin types).

So it seems what you're really signaling is what Telegraf version the plugin first became available in, and not that the plugin wouldn't work if you built it in an older Telegraf version.

sjwang90:
lets not do this. this is already highlighted on the plugins with [our docs](https://docs.influxdata.com/telegraf/v1.16/plugins/).

PR #8243, opened 2020-10-08
Sumo Logic output plugin: fix unparsable config.Size from sample config

pmalek-sumo:
After deciding to use `config.Size` in https://github.com/influxdata/telegraf/commit/e9dcade0a85701a645f11ecd11a1463b4ecc4d7f I've noticed that it uses `ParseStrictBytes` from https://github.com/alecthomas/units/blob/b9c82cb/bytes.go#L77-L85 which doesn't understand underscores in numbers.

This PR addresses that by changing what's in the sample config not to confuse users.

### Required for all PRs:

### DIFF
--- a/etc/telegraf.conf
+++ b/etc/telegraf.conf
@@ -1337,7 +1337,7 @@
 #   ## Bear in mind that in some serializer a metric even though serialized to multiple
 #   ## lines cannot be split any further so setting this very low might not work
 #   ## as expected.
-#   # max_request_body_size = 1_000_000
+#   # max_request_body_size = 1000000
 #
 #   ## Additional, Sumo specific options.
 #   ## Full list can be found here:
--- a/plugins/outputs/sumologic/sumologic.go
+++ b/plugins/outputs/sumologic/sumologic.go
@@ -55,7 +55,7 @@ const (
   ## Bear in mind that in some serializer a metric even though serialized to multiple
   ## lines cannot be split any further so setting this very low might not work
   ## as expected.
-  # max_request_body_size = 1_000_000
+  # max_request_body_size = 1000000
 
   ## Additional, Sumo specific options.
   ## Full list can be found here:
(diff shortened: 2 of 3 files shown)


pmalek-sumo:
Any chance to get this before `v1.16` release? cc: @reimda

etc/telegraf.conf
```diff
@@ -1337,7 +1337,7 @@
 #   ## Bear in mind that in some serializer a metric even though serialized to multiple
 #   ## lines cannot be split any further so setting this very low might not work
 #   ## as expected.
-#   # max_request_body_size = 1_000_000
+#   # max_request_body_size = 1000000
```
reimda, line 1340: You don't need to edit etc/telegraf.conf.  It gets updated once before the release to incorporate changes from all plugins.

plugins/outputs/sumologic/sumologic.go
```diff
@@ -55,7 +55,7 @@ const (
   ## Bear in mind that in some serializer a metric even though serialized to multiple
   ## lines cannot be split any further so setting this very low might not work
   ## as expected.
-  # max_request_body_size = 1_000_000
+  # max_request_body_size = 1000000
```
reimda, line 58: Could you also make this change in the readme?

reimda (changes requested):

etc/telegraf.conf
pmalek-sumo: Ok, noted. Thanks!

plugins/outputs/sumologic/sumologic.go
pmalek-sumo: Done

reimda (approved):

reimda:
Thanks @pmalek-sumo!

pmalek-sumo:
No problem, thanks for merging this

Issue #8234, opened 2020-10-07
[processors.regex] Add fully overwrite ability


HongChenTW:
## Feature Request

https://github.com/influxdata/telegraf/tree/master/plugins/processors/regex

### Proposal:

If `pattern` is matched, we can use `replacement` to fully overwrite existed key or fill into `result_key`.

### Current behavior:

Currently regex plugin can only replace matched patterns.

Ex. 

```
  [[processors.regex.fields]]
    key = "path"
    pattern = "^/blog/"
    replacement = "blog"
    result_key = "url-group"
```

input: `path="/blog/telegraf"`
output: `path="/blog/telegraf",url-group="blogtelegraf"`

### Desired behavior:

```
  [[processors.regex.fields]]
    key = "path"
    pattern = "^/blog/"
    replacement = "blog"
    result_key = "url-group"
    overwrite = true
```

input: `path="/blog/telegraf"`
output: `path="/blog/telegraf",url-group="blog"`

### Use case: 

Assume that the pattern is defined by other users, patterns can not be controlled, so if user's pattern regex only match part of value, the result_key value will be mixed with original value.

If plugin has ability to fill value exactly we defined, not just replacement, it would be more widely used with more use cases.

I'm happy to contribute if this feature is needed.

PR #8247, opened 2020-10-08
Add fully overwrite ability to processors.regex plugin #8234

HongChenTW:
### Required for all PRs:

### DIFF
--- a/plugins/processors/regex/README.md
+++ b/plugins/processors/regex/README.md
@@ -4,6 +4,8 @@ The `regex` plugin transforms tag and field values with regex pattern. If `resul
 
 For tags transforms, if `append` is set to `true`, it will append the transformation to the existing tag value, instead of overwriting it.
 
+If `overwrite` is set to `true`, it will be fully overwritten with replacement field, instead of only patterns match part.
+
 ### Configuration:
 
 ```toml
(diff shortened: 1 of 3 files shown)


ssoroka:
I'm not sure I understand the need for this change. In your test example, you could add `.*` to the end of the match to include the rest of the string in the replacement text. Ultimately this is a lot more flexible than a boolean all-or-nothing replacement flag.

HongChenTW:
Hi @ssoroka , thanks for reviewing.

In our scenario, we develop a mechanism for automatically generating telegraf config with user's setting (both `pattern` & `replacement`), and we use this plugin to help users filter their logs.

Since we cannot control what regex user will give and also want to leverage this plugin `replacement` feature for classification, add a all-or-nothing replacement flag will be flexible for using telegraf as SaaS.

ssoroka:
Even so, I'm hesitant to add a second way to do something when one already exists. I would think if you want to support this behavior as a layer on top of Telegraf, you could do so by modifying the supplied regex. 
eg (pseudocode)
```
if overwrite 
  if !startsWith "^" or "\A" 
    regex = ".*" + regex
  if !endsWith "$" or "\z"
    regex = regex + ".*"
```

ssoroka:
going to close this as it's already supported via regex and we don't want to add a second way to do this.

HongChenTW:
Sure, thanks for the suggestion.

Issue #7397, opened 2020-04-23
influxdb output (v1) should gzip by default


ssoroka:
should default to `content_encoding = "gzip"` for influxdb

### Relevant telegraf.conf:

```toml
[[outputs.influxdb]]
  ## HTTP Content-Encoding for write request body, can be set to "gzip" to
  ## compress body or "identity" to apply no encoding.
  # content_encoding = "identity"
```

### System info:

Telegraf 1.14

### Expected behavior:

gzip compression by default

### Actual behavior:

no data compression

### Additional info:

Need to make sure this won't break any old InfluxDB servers; but as far as I can tell it's not going to be an issue for any version of InfluxDB.

timhallinflux:
any reason this wouldn't be the default for v2 as well?

danielnelson:
On v2 gzip is the default.  If we make this change, with the default settings network bandwidth will be reduced substantially but it will most likely increase CPU requirements for the database.

ssoroka:
I suspect the cpu cost of uncompressing a gzip body is so minimal that it won't have a measurable impact, yet it has a huge impact on the throughput rate of metrics; a highly visible negative performance when it is disabled.

e-dard:
I don't have a good feeling on whether this would be a good idea from a Cloud 1 perspective, but that is something to think about.

My initial feeling on this is that regardless of whether this is enabled or not InfluxDB is processing the same amount of metrics. 

With it enabled InfluxDB has to do more work to decompress the payload, but the client can send metrics in a smaller payload and more quickly. If metrics are arriving more quickly after a sampling interval, and they're now more expensive to process I wonder if that could have some impact in some cases..

If you put to one side the reduction in network, what real benefit is there to a user in pushing their (let's say) 10 second interval metrics slightly quicker to the DB? 

> I suspect the cpu cost of uncompressing a gzip body is so minimal that it won't have a measurable impact

I have seen InfluxDB profiles where gzip is showing up quite heavily in the profiles.  I can't recall whether that was due to the Go implementation of the gzip decoder, or the volume of writes on those clusters in those cases.

ssoroka:
> If you put to one side the reduction in network, what real benefit is there to a user in pushing their (let's say) 10 second interval metrics slightly quicker to the DB?

[...]

PR #8269, opened 2020-10-14
turn gzip on by default for influxdb v1

ssoroka:
resolves https://github.com/influxdata/telegraf/issues/7397

### DIFF
--- a/plugins/outputs/influxdb/README.md
+++ b/plugins/outputs/influxdb/README.md
@@ -75,7 +75,7 @@ The InfluxDB output plugin writes metrics to the [InfluxDB v1.x] HTTP or UDP ser
 
   ## HTTP Content-Encoding for write request body, can be set to "gzip" to
   ## compress body or "identity" to apply no encoding.
-  # content_encoding = "identity"
+  # content_encoding = "gzip"
 
   ## When true, Telegraf will output unsigned integers as unsigned values,
   ## i.e.: "42u".  You will need a version of InfluxDB supporting unsigned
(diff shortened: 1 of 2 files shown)


reimda:
I wish gzip would have been the default from the beginning but I have reservations about changing the default at this point.  It was uncompressed from the beginning of telegraf (almost 5 years ago) and default uncompressed since the content encoding option was added 3 years ago.

It should be safe in most cases to switch to gzip, but there are cases where it might not be, such as when the device running telegraf was sized with the expectation of low cpu usage (think embedded device or small virtual instance on a cloud provider).  The same goes for influxdb which has to gunzip if telegraf is sending gzipped.

I like the idea of gradually defaulting to gzip.  The v2 output is already default gzip, so people upgrading that way will get it.  I'd also be be fine changing the v1 output default to gzip in a major release (2.0) where backwards compatibility isn't as expected as a minor release (1.x).

I haven't seen a lot of user reports of uncompressed v1 being a problem.  You mentioned one case in https://github.com/influxdata/telegraf/issues/7397#issuecomment-628742703 where enabling gzip helped.  I don't think it's worth the risk when people aren't asking for it.

ssoroka:
I understand where you're coming from, but the potential gains here outweigh any potential risk, especially since you can opt to turn it off. Anyone who has set this setting explicitly will not be affected, and we can document the change.

Some other things to consider:
- I've talked to almost a dozen customers now where the answer was to enable gzip. Nobody has ever complained about gzip support in v2 being the default, and the server has no problems handling this.
- the cost of gzip is order of magnitude less than the network costs of transmitting the data. 
- gzip compression is the standard everywhere. This is an extremely reasonable and safe default.
- No server has ever died or become slow, to my knowledge, due to the cost of doing gzip decompression. TLS will be an issue before gzip is.
- this isn't breaking backwards compatibility, because all past users will continue to work.

lastly, we can't be afraid that we might make performance worse for a few people with underpowered devices at the cost of performance at high-volume for the majority.

PR #8273, opened 2020-10-15
Fix suricata crash on startup when unix socket missing

odedlaz:
Suricata plugin tries to dial into the configured stats socket
if one is missing, the plugin doesn't start up instead of
accumulating the error.

### Required for all PRs:

### DIFF
--- a/plugins/inputs/suricata/suricata.go
+++ b/plugins/inputs/suricata/suricata.go
@@ -56,16 +56,16 @@ func (s *Suricata) SampleConfig() string {
 	return sampleConfig
 }
 
-// Start initiates background collection of JSON data from the socket
-// provided to Suricata.
-func (s *Suricata) Start(acc telegraf.Accumulator) error {
+// Gather reads stats from the socket provided to Suricata.
+func (s *Suricata) Gather(acc telegraf.Accumulator) error {
 	var err error
 	s.inputListener, err = net.ListenUnix("unix", &net.UnixAddr{
 		Name: s.Source,
 		Net:  "unix",
 	})
 	if err != nil {
-		return err
+		acc.AddError(err)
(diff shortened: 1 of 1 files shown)


srebhan:
Hey @odedlaz!

The build failed with errors and I think your change is not quite right. With your change, a new "connection" is made everytime `Gather()` is called without closing this connection. I'd rather split the functions into a `connect()/disconnect()` and "get and parse metric" part.
Then you try to connect in init. If this does not work, you silently ignore the error and set a flag `connected` to false (or true if connection was successful). Then in `Gather()` you should check if connected is true and otherwise try to connect (maybe also incrementing the error count). If reading the socket fails, set connected to false again and either retry connection or drop out of `Gather()`...

Do you think you can rework the PR?

srebhan (changes requested):
Hey @odedlaz!

The build failed with errors and I think your change is not quite right. With your change, a new "connection" is made everytime Gather() is called without closing this connection. I'd rather split the functions into a connect()/disconnect() and "get and parse metric" part.
Then you try to connect in init. If this does not work, you silently ignore the error and set a flag connected to false (or true if connection was successful). Then in Gather() you should check if connected is true and otherwise try to connect (maybe also incrementing the error count). If reading the socket fails, set connected to false again and either retry connection or drop out of Gather()...

Do you think you can rework the PR?

odedlaz:
@srebhan, thank you so much for answering! after further debugging I came to the conclusion that this PR is redundant since the problem was on my end. I'm closing the PR.

srebhan:
@odedlaz can you elaborate on what the problem was? I mean telegraf should _never_ crash I think, so maybe we can catch the "problem on your side" and at least avoid crashing...

PR #8278, opened 2020-10-16
Add glibc version check for linux builds

reimda:
This version check makes sure that telegraf on linux platforms doesn't dynamically link to a glibc that is too new.  Requiring a new glibc would cause telegraf to fail to run on older supported linux distributions.  The version it compares against is configurable in the makefile and set to 2.17 which is what RHEL 7 provides.

The check runs as part of the makefile's packages target which already runs during CI, so this will run on every PR.

CI for this PR is failing intentionally to show that it works.


ssoroka:
lgtm

ssoroka (approved):

PR #8282, opened 2020-10-17
Fix using empty string as the namespace prefix in azure_monitor output plugin

### LINKED ISSUE #8256: [azure_monitor] Cannot use empty string as the nampspace prefix
### Relevant telegraf.conf:

```toml
[[outputs.azure_monitor]]
  namespace_prefix = ""

# For debugging
[[outputs.file]]
  data_format = "influx"
  files = ["stdout"]
```

### System info:

Telegraf 1.15.3

### Steps to reproduce:

1. Run telegraf with the above configuration

### Expected behavior:

Metrics are emitted to azure monitor in the AZM namespace which matches the field name.

### Actual behavior:

Metrics are emitted to azure monitor in the AZM namespace `telegraf/{fieldName}`.

### Additional info:

[This](https://github.com/influxdata/telegraf/blob/a8810b2b7ae93000105b20842cb7ffb5cf94d6f8/plugins/outputs/azure_monitor/azure_monitor.go#L158-L160) seems to be the suspect part of the code. Can it be compared to `nil` instead of the empty string?

As a side note, the [regex](https://github.com/influxdata/telegraf/tree/master/plugins/processors/regex) processor may be a better solution to add namespace prefixes.

~ATTN: @ danielnelson~

simnv:
Fixes #8256

### Required for all PRs:

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -12,6 +12,7 @@ import (
 	"github.com/influxdata/telegraf/plugins/inputs/http_listener_v2"
 	"github.com/influxdata/telegraf/plugins/inputs/memcached"
 	"github.com/influxdata/telegraf/plugins/inputs/procstat"
+	"github.com/influxdata/telegraf/plugins/outputs/azure_monitor"
 	httpOut "github.com/influxdata/telegraf/plugins/outputs/http"
 	"github.com/influxdata/telegraf/plugins/parsers"
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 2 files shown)


ssoroka:
this is awesome. Can you add a test or confirm you've tested the original scenario and that it resolves it?

simnv:
@ssoroka It may be a bit overkill, but I hope I've added this test in a correct place.

PR #8287, opened 2020-10-19
Add support for an inclusive job list in Jenkins plugin

JS1010111:
### Required for all PRs:

### DIFF
--- a/plugins/inputs/jenkins/README.md
+++ b/plugins/inputs/jenkins/README.md
@@ -41,7 +41,7 @@ This plugin does not require a plugin on jenkins and it makes use of Jenkins API
 
   ## Jobs to include or exclude from gathering
   ## When using both lists, job_exclude has priority.
-  # job_include = [ "jobA", "jobB/subjobA/subjobB", "jobC/*"]
+  # job_include = [ "jobA", "jobB/subjobA/subjobB", "jobC*"]
   # job_exclude = [ "job1", "job2/subjob1/subjob2", "job3/*"]
 
   ## Nodes to exclude from gathering
--- a/plugins/inputs/jenkins/README.md
+++ b/plugins/inputs/jenkins/README.md
@@ -39,7 +39,9 @@ This plugin does not require a plugin on jenkins and it makes use of Jenkins API
   ## empty will use default value 10
   # max_subjob_per_layer = 10
 
-  ## Jobs to exclude from gathering
+  ## Jobs to include or exclude from gathering
+  ## When using both lists, job_exclude has priority.
+  # job_include = [ "jobA", "jobB/subjobA/subjobB", "jobC/*"]
   # job_exclude = [ "job1", "job2/subjob1/subjob2", "job3/*"]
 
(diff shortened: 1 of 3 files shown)


plugins/inputs/jenkins/README.md
```diff
@@ -39,7 +39,9 @@ This plugin does not require a plugin on jenkins and it makes use of Jenkins API
   ## empty will use default value 10
   # max_subjob_per_layer = 10
 
-  ## Jobs to exclude from gathering
+  ## Jobs to include or exclude from gathering
+  ## When using both lists, job_exclude has priority.
+  # job_include = [ "jobA", "jobB/subjobA/subjobB", "jobC*"]
```
srebhan, line 44: The default should be an empty list with the meaning of "include all" as I guess that was the behavior before.

plugins/inputs/jenkins/jenkins.go
```diff
@@ -77,7 +79,9 @@ const sampleConfig = `
   ## empty will use default value 10
   # max_subjob_per_layer = 10
 
-  ## Jobs to exclude from gathering
+  ## Jobs to include or exclude from gathering
+  ## When using both lists, job_exclude has priority.
+  # job_include = [ "jobA", "jobB/subjobA/subjobB", "jobC*"]
```
srebhan, line 84: Same as in README.

srebhan (commented):
The only comment is that the README (and `sampleConfig`) should show the actual default for the option. In this case the default is an empty list... Code looks good otherwise.

JS1010111:
Thanks for the review  @srebhan !

Actually the default behavior is not changed here. This PR should not break or change in any way the current configs in use.

The default was no excluded list defined (it's commented in config), which let all jobs to be collected.

With this PR If you pass an empty `job_include` list no jobs will be collected because there's nothing to match.

To get all jobs I see 2 options:
- No lists defined (default behavior, not changed)
- A `job_include` list like: `[*]`

Please let me know if you think we should improve the README or if it's ok. Thanks!

srebhan:
Hey @JS1010111, I think there is a misunderstanding. This is not about **code** change but only about change in the config documentation (README.md and `sampleConfig`). As far as I understand your code, if the user does not specify an `job_include` or `job_exclude`, both `JobInclude` and `JobExclude` variables are empty string lists. In turn the filters `j.jobFilterInclude` and `j.jobFilterExclude` will be `nil` leading to no filtering. Perfect this **code** keeps the previous behavior, there is absolutely no dissent!
My comment targeted the entry in README.md and `sampleConfig`. Here the convention says that the commented variable shall contain the default value. In your code, the default value is an empty list! So please change
`# job_include = [ "jobA", "jobB/subjobA/subjobB", "jobC*"]` to `# job_include = [ ]` to match the code! You can do the same for the exclude option. Just add a comment that an empty include list means to include all. The same is done for multiple other plugins which all define an empty include list as "use all".

srebhan (approved):
Looks good to me.

Issue #7263, opened 2020-04-01
input kafka_consumer uses old partition leader during Kafka cluster node failure


dpajin:
### Relevant telegraf.conf:

```toml

[agent]
  interval = "30s"
  round_interval = true
  flush_buffer_when_full = true
  metric_buffer_limit = 10000000
  metric_batch_size = 100000
  flush_interval = "2s"
  flush_jitter = "0s"
  debug = true
  omit_hostname = true

[[outputs.influxdb]]
  urls = ["http://db:8086"]
  database = "kpi"
  username = 
  password = 
  precision = "s"
  retention_policy = "default"
  timeout = "5s"
  skip_database_creation = true

[[inputs.kafka_consumer]]
  brokers = [ "node1:9094", "node2:9094", "node3:9094" ]
  topics = ["kpi"] 
  version = "2.0.0"
  max_message_len = 0
  max_undelivered_messages = 1000
  data_format = "influx"
```

### System info:
Telegraf 1.13.4

Linux 5.3.0-42-generic #34~18.04.

Kafka version 2.4.0 (Commit:77a89fcf8d7fa018)

### Docker
Docker Swarm 19.03.7

### Steps to reproduce:

I have a Kafka cluster of 3 instances, running in the Docker Swarm environment with 3 nodes. 
Each Kafka instance is on one node.

I have 3 Telegraf instances which uses kafka_consumer plugin to read the data from the Kafka cluster. It reads 1 topic which has **1 partition and 3 replicas** - for each Kafka node

When Kafka instance which is the leader for the partition goes down, it happens that the Telegraf kafka_consumer does not switch to use another Kafka node (new partition leader) for consuming metric, but it tries to access the Kafka node (old partition leader) which is down.

#### Normal situation, everything up:
Kafka metadata - 3 brokers: node1, node2 and node3, partition leader is node2.
```
Metadata for all topics (from broker 1049: node1:9094/1049):
 3 brokers:
  broker 1049 at node1:9094
  broker 1003 at node3:9094
  broker 1001 at node2:9094
 2 topics:
  topic "kpi" with 1 partitions:
    partition 0, leader 1001, replicas: 1001,1003,1049, isrs: 1049,1003,1001
  topic "__consumer_offsets" with 1 partitions:
    partition 0, leader 1001, replicas: 1001,1003,1049, isrs: 1049,1003,1001
``` 
Telegraf log:
```

2020-04-01T09:29:14Z I! Starting Telegraf 1.13.4
[...]

PR #8289, opened 2020-10-19
Update sarama library to 1.27.1

reimda:
Sarama 1.24.1 has known regressions and the project recommends it should not be used.  https://github.com/Shopify/sarama/releases/tag/v1.24.1


reimda:
Sarama 1.27.1 requires go 1.15 but telegraf CI runs on 1.15 and 1.14.  We might need to use an older sarama version that builds on 1.14 but also fixes the issues in 1.24.1.

sjwang90:
https://github.com/influxdata/telegraf/issues/7263

PR #8306, opened 2020-10-23
RAS plugin - fix for too many open files handlers

p-zak:
### Required for all PRs:

### Details
Database driver doesn't close file handler properly when `db.Close()` is called. 
This may lead to `too many open files` for Telegraf process (returned by OS).

In this fix `sql.Open()` is called once in `Start()` and `db.Close()` is also called only once in `Stop()`.

This is also optimization (open DB just once, not in every `Gather()`) and following along guide from comment for `func (db *DB) Close() error` in `sql.go`:

> // It is rare to Close a DB, as the DB handle is meant to be
> // long-lived and shared between many goroutines.

The same approach was introduced in `inputs/postgresql` plugin.

### DIFF
--- a/plugins/inputs/ras/ras.go
+++ b/plugins/inputs/ras/ras.go
@@ -107,11 +107,6 @@ func (r *Ras) Stop() {
 
 // Gather reads the stats provided by RASDaemon and writes it to the Accumulator.
 func (r *Ras) Gather(acc telegraf.Accumulator) error {
-	err := r.db.Ping()
-	if err != nil {
-		return fmt.Errorf("connection to DB (%s) couldn't be established: %v", r.DBPath, err)
-	}
-
 	rows, err := r.db.Query(mceQuery, r.latestTimestamp)
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


plugins/inputs/ras/ras.go
```diff
@@ -77,20 +80,39 @@ func (r *Ras) Description() string {
...
+		}
+	}
+}
+
+// Gather reads the stats provided by RASDaemon and writes it to the Accumulator.
+func (r *Ras) Gather(acc telegraf.Accumulator) error {
+	err := r.db.Ping()
+	if err != nil {
+		return fmt.Errorf("connection to DB (%s) couldn't be established: %v", r.DBPath, err)
+	}
```
ssoroka, line 113: I'd consider skipping this and just doing the db.Query. If it's possible to detect the failed connection error and try reconnecting that would be fine.

p-zak: You're right! `db.Query -> db.QueryContext -> db.query` do the same potential reconnect as `db.Ping` so even direct reconnection is not needed.

ssoroka (approved):

PR #7176, opened 2020-03-14
Add Syncthing plugin for gathering metrics

kradalby:
Monitor Syncthing instance, gather useful metrics :)

### Required for all PRs:

- [x] Signed [CLA](https://influxdata.com/community/cla/).
- [x] Associated README.md updated.
- [x] Has appropriate unit tests.

PR #8312, opened 2020-10-25
inputs: add syncthing

blockloop:
This adds an input for [Syncthing REST API](https://docs.syncthing.net/dev/rest.html).
It creates two measurements described in detail in the README. I would like to improve
on this over time as there is much more information that would be useful, but for now
I think this covers the major parts. We can identify if a device/folder is "Out of sync".

### syncthing_folder

Tags:
 * `id` - the folder ID
 * `label` - the label (name) you have assigned to the folder
 * `path` - the sync path on the local filesystem

Fields: 
 * `paused` - if the folder is paused
 * `needed` - how many files are needed to be in sync

### syncthing_device

Tags:
 * `device_id` - the device id that is connected
 * `name` - the name of the connected device 

Fields:
 * `address` - the IP address of the connected device
 * `client_version` - the version of Syncthing that the connected device is using
 * `connected` - if the device is connected
 * `crypto` - which version of TLS the device is using for encryption
 * `in_bytes_total` - how many bytes have been received by this device
 * `out_bytes_total` - how many bytes have been sent to this device
 * `paused` - if the device connection has been paused

### Required for all PRs:

### DIFF
--- a/plugins/inputs/all/all.go
+++ b/plugins/inputs/all/all.go
@@ -170,6 +170,7 @@ import (
 	_ "github.com/influxdata/telegraf/plugins/inputs/statsd"
 	_ "github.com/influxdata/telegraf/plugins/inputs/suricata"
 	_ "github.com/influxdata/telegraf/plugins/inputs/swap"
+	_ "github.com/influxdata/telegraf/plugins/inputs/syncthing"
 	_ "github.com/influxdata/telegraf/plugins/inputs/synproxy"
 	_ "github.com/influxdata/telegraf/plugins/inputs/syslog"
 	_ "github.com/influxdata/telegraf/plugins/inputs/sysstat"
--- a/plugins/inputs/syncthing/README.md
+++ b/plugins/inputs/syncthing/README.md
@@ -1,0 +1,60 @@
+# Syncthing Input Plugin
+
+The Syncthing input plugin collects information from one Syncthing API endpoint.
+
+## Configuration:
+
+```toml
+## Syncthing host
+url = "http://localhost:8384"
+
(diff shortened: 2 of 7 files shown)


mhall119:
Will the `syncthing_folder` measurement have a different `path` for every sync'd folder, including subfolders? Or just the top-level ones?  If it's every folder this may result in high cardinality in InfluxDB

blockloop:
> Will the `syncthing_folder` measurement have a different `path` for every sync'd folder, including subfolders? Or just the top-level ones? If it's every folder this may result in high cardinality in InfluxDB

Only the top-level folder that is being synced [see more here](https://docs.syncthing.net/rest/system-config-get.html)

blockloop:
@sjwang90 any chance of getting this merged?

sjwang90:
@blockloop Thanks for the contribution and sorry for the delay! How does this compare to https://github.com/influxdata/telegraf/pull/7176? We had that PR open for awhile and were hoping to have it added to the [list of external plugins](https://github.com/influxdata/telegraf/blob/master/EXTERNAL_PLUGINS.md).

Would this be a reasonable route for your Syncthing input plugin?

blockloop:
> @blockloop Thanks for the contribution and sorry for the delay! How does this compare to #7176? We had that PR open for awhile and were hoping to have it added to the [list of external plugins](https://github.com/influxdata/telegraf/blob/master/EXTERNAL_PLUGINS.md).
> 
> Would this be a reasonable route for your Syncthing input plugin?

Ah yes. I didn't know that was already opened :grimacing:

blockloop:
Closing in favor of #7176

PR #8314, opened 2020-10-26
Feat/add stepped processor plugin

tomhollingworth:
### Required for all PRs:

## What?

This pull request is for a community contributed processor plugin. It emits a metric of the previous unique field value and tag set with the timestamp just before the current one to display field as stepped. See [Step Function](https://en.wikipedia.org/wiki/Step_function).

## Why?

Some metrics are discrete by nature, for example string or boolean fields. When querying and visualizing data downstream it is advantageous to know the previous value on stepped metrics for calculations and visualization. Whilst this could be considered part of the visualization effort and query effort, those responsible parties for capturing the data in the first place are better situated to indicate that a metric is stepped or not.

## How?

The stepped processor plugin caches the last field value for a set of tags. When processing, it compares the processing metrics against the cache, if a field in `unique_fields` has changed between the current value and cache, then this processor emits an addition metric that is the last value at the timestamp of the current record minus the `step_offset`. Add a metric to the cache if it has a `unique_field` and isn't already in the cache. Remove a metric from the cache if it hasn't been updated in `cache_interval`. Identical `unique_field` values will update the timestamp in the cache.

### DIFF
--- a/plugins/processors/all/all.go
+++ b/plugins/processors/all/all.go
@@ -20,6 +20,7 @@ import (
 	_ "github.com/influxdata/telegraf/plugins/processors/reverse_dns"
 	_ "github.com/influxdata/telegraf/plugins/processors/s2geo"
 	_ "github.com/influxdata/telegraf/plugins/processors/starlark"
+	_ "github.com/influxdata/telegraf/plugins/processors/stepped"
 	_ "github.com/influxdata/telegraf/plugins/processors/strings"
 	_ "github.com/influxdata/telegraf/plugins/processors/tag_limit"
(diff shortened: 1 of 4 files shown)


sjwang90:
Hi @tomhollingworth - Sorry for the delay into looking at this but this functionality actually seems like something the [Starlark processor](https://github.com/influxdata/telegraf/tree/master/plugins/processors/starlark) could do well. We try not to add new processors who's functionality can be done in Starlark.

If you find it successful, we can add it to the list of [examples](https://github.com/influxdata/telegraf/tree/master/plugins/processors/starlark#examples) for others to use as well.

tomhollingworth:
Closing this as I no longer require this functionality

PR #8172, opened 2020-09-23
New sql server queries (on-prem) - refactoring and formatting

Trovalo:
### Required for all PRs:

- [x] Signed [CLA](https://influxdata.com/community/cla/).
- [ ] Associated README.md updated.
- [ ] Has appropriate unit tests.

This PR affects only the SQL Server on-prem queries (database_type = "SQLServer"), a different PR can be opened for the Azure queries (if this is approved).
**The PR does not contain any breaking change**, the output is the same as before (with minor fixes)

This PR contains:
- SQL code indentation and formatting
- SQL code refactoring (nothing breaking)
  - Removed @EngineEdition variables (not used anymore)
  - minor edits/fixes to the "sqlServerRequests" query
    - Removed ```DB_NAME() as [database_name]```
    - added COALESCE for [session_db_name] and [open_transaction] (priority to dm_exec_requests, then dm_exec_sessions)

An additional not breaking fix can be made in the PerfCounters query. on line 436.
```SQL
	INNER JOIN sys.dm_resource_governor_resource_pools AS rgrp /*No fields from this table. remove?*/
		ON rgwg.[pool_id] = rgrp.[pool_id]
```
this table is joined, but no data is fetched from it. @denzilribeiro, to me this join can be removed, can you confirm? do you have ideas about why it was there?

PR #8325, opened 2020-10-27
SQL Server - PerformanceCounters - removed synthetic counters

Trovalo:
### Required for all PRs:

This PR removes some synthetic performance counters from the sqlserver_performance_counters measurement.

We believe that those counters were added to create unavailable counters on older editions of SQL Server, as of now, some of those counters already exist even in SQL 2008. In any case, we don't really want to fetch performance counters that do not exist.

- affects only the latest version of the queries
- affect only the on-prem query, for Azure those synthetic counter never existed
- The query has been simplified

### DIFF
--- a/plugins/inputs/sqlserver/sqlserverqueries.go
+++ b/plugins/inputs/sqlserver/sqlserverqueries.go
@@ -292,8 +292,6 @@ END
 DECLARE
 	 @SqlStatement AS nvarchar(max)
 	,@MajorMinorVersion AS int = CAST(PARSENAME(CAST(SERVERPROPERTY('ProductVersion') AS nvarchar),4) AS int)*100 + CAST(PARSENAME(CAST(SERVERPROPERTY('ProductVersion') AS nvarchar),3) AS int)
-	,@Columns AS nvarchar(MAX) = '
-	,@PivotColumns AS nvarchar(MAX) = '
 
 DECLARE @PCounters TABLE
 (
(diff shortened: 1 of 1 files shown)


plugins/inputs/sqlserver/sqlserverqueries.go
```diff
@@ -292,8 +292,8 @@ END
 DECLARE
 	 @SqlStatement AS nvarchar(max)
 	,@MajorMinorVersion AS int = CAST(PARSENAME(CAST(SERVERPROPERTY('ProductVersion') AS nvarchar),4) AS int)*100 + CAST(PARSENAME(CAST(SERVERPROPERTY('ProductVersion') AS nvarchar),3) AS int)
-	,@Columns AS nvarchar(MAX) = ''
-	,@PivotColumns AS nvarchar(MAX) = ''
+	,@Columns AS nvarchar(MAX) = '
+	,@PivotColumns AS nvarchar(MAX) = '
```
ssoroka, line 296: looks unintentional

ssoroka (changes requested):

plugins/inputs/sqlserver/sqlserverqueries.go
Trovalo: You are right, I replaced some single quotes and that introduced that error, as it should be 2 single-quotes.
Anyway turns out those variables are no longer needed and can be removed, as there were only used in the shythetic counters section.

Thanks for spotting it.

Trovalo: I already made a commit to remove the variables

ssoroka (approved):
Can I ask why the synthetic counters were removed? I don't really have any context on it, but someone might find the answer useful in the future.

Trovalo:
When we had a look at the code (me and @denzilribeiro), we had no idea about what those counters were doing there, so we concluded they were there in the first version of the plugin, to compensate for something (maybe specific) that was missing.

But since they just don't belong to SQL Server performance counters, It's probably better to remove, they might cause confusion as you can't find them in the SQL server dmv (sys.dm_os_performance_counters), also the "Workload Group" related counters are already there, in the default counters

here is part of the conversation 
https://github.com/influxdata/telegraf/pull/8172#issuecomment-697576901

PR #8326, opened 2020-10-27
proxmox: ignore QEMU templates and iron out a few bugs

krzysdabro:
### Changes
- Ignore QEMU templates
- Fix logger calls (`Error` -> `Errorf`)
- Changed confusing error `node_name not found` to `search domain is not set`
- Fix creator to return a pointer to new struct on every call (now it uses the same struct every time, which results in gathering only the last input of proxmox in the configuration)
- Updated README.md to acknowledge the default value of `node_name`

### Required for all PRs:

### DIFF
--- a/plugins/inputs/proxmox/proxmox.go
+++ b/plugins/inputs/proxmox/proxmox.go
@@ -3,19 +3,22 @@ package proxmox
 import (
 	"encoding/json"
 	"errors"
-	"github.com/influxdata/telegraf"
-	"github.com/influxdata/telegraf/plugins/inputs"
 	"io/ioutil"
 	"net/http"
 	"net/url"
 	"os"
 	"strings"
+
+	"github.com/influxdata/telegraf"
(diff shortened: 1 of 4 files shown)


krzysdabro:
@ssoroka @danielnelson Can you review this PR?

plugins/inputs/proxmox/proxmox.go
```diff
@@ -49,9 +52,10 @@ func (px *Proxmox) Gather(acc telegraf.Accumulator) error {
 }
 
 func (px *Proxmox) Init() error {
-
+	// Set hostname as default node name for backwards compatibility
 	if px.NodeName == "" {
-		return errors.New("node_name must be configured")
+		hostname, _ := os.Hostname()
+		px.NodeName = hostname
 	}
```
ssoroka, line 59: Just a thought, if you move this to the init() function below, you don't need the `""` check, and it'll allow users to unset it if they don't want it

ssoroka (approved):
Looks great, thanks!

PR #8333, opened 2020-10-28
add the ability to configure inputs.exec error truncation

ssoroka:


plugins/inputs/exec/README.md
```diff
@@ -22,6 +22,9 @@ This plugin can be used to poll for custom metrics from any source.
   ## measurement name suffix (for separating different commands)
   name_suffix = "_mycollector"
 
+  ## truncate errors after this many characters. Use 0 to disable error truncation.
+  # error_truncate_length = 512
```
srebhan, line 26: Shouldn't this be zero to keep the previous behavior?

plugins/inputs/exec/exec.go
```diff
@@ -30,7 +30,10 @@ const sampleConfig = `
   timeout = "5s"
 
   ## measurement name suffix (for separating different commands)
-  name_suffix = "_mycollector"
+	name_suffix = "_mycollector"
```
srebhan, line 33: There seems to be a formatting problem...

plugins/inputs/exec/exec.go
```diff
@@ -39,12 +42,13 @@ const sampleConfig = `
   data_format = "influx"
 `
 
-const MaxStderrBytes = 512
+const MaxStderrBytes int = 512
```
srebhan, line 45: What happens if the user specifies more than this max? You maybe should check that in `Init()`, clip to max and issue a warning...

srebhan (commented):
I only have a minor comment (see code). Could you please also add a test-case where you actually truncate with a non-default argument!? Just to make sure... :-)

srebhan:
@ssoroka ping... ;-P

srebhan:
@ssoroka any news on this?

helenosheaa:
When Dave and I paired on reviews we were wondering what problem this is trying to solve? When would the user know that they need this?

plugins/inputs/exec/exec.go
ssoroka:
```suggestion
  name_suffix = "_mycollector"
```

sspaink (approved):
lgtm

ssoroka:
exec truncates the output by default, so you might not be able to see the whole message about what is causing the script to fail. There's an open issue or user request for this somewhere; not sure why it's not linked.

An alternative (or corollary) to this is to show the whole output when running in debug, and then adjusting the length is less of an issue.

PR #8335, opened 2020-10-29
Add setting to enable caching in ipmitool

DEvil0000:
I had the issue that ipmi data was reported with gaps on a 30s interval on HP G10 servers (latest firmware).
The gaps where due to the ipmitool command was not always finishing within the 20s timeout. Even changeing the timeout to 30s was not always giving it enough time to respond.
ipmi-sensors from freeipmi tools which is used in some prometheus ipmi exporters is running way faster but does not provide all information I needed.
At some point I just realized there is a cache option for ipmitool. Using this option speeds the tool up by something between 3x and 6x for most of my runs so it holds the deadline given by timeout and I can record the data not even in 30s intervals but also now in 15s intervals.

In case you merge this or agree with it, I would appreciate a hacktoberfest-accepted label.

### DIFF
--- a/plugins/inputs/ipmi_sensor/ipmi.go
+++ b/plugins/inputs/ipmi_sensor/ipmi.go
@@ -12,6 +12,7 @@ import (
 	"sync"
 	"time"
 	"os"
+	"path/filepath"
 
 	"github.com/influxdata/telegraf"
 	"github.com/influxdata/telegraf/internal"
@@ -76,8 +77,8 @@ var sampleConfig = `
   ## If ipmitool should use a cache
   use_cache = true
 
-  ## Path to the ipmitools cache file
-  cache_path = "/tmp/"
+  ## Path to the ipmitools cache file (defaults to OS temp dir)
+  cache_path = "/tmp"
 `
 
 // SampleConfig returns the documentation about the sample configuration
@@ -129,7 +130,7 @@ func (m *Ipmi) parse(acc telegraf.Accumulator, server string) error {
 	}
 	opts = append(opts, "sdr")
 	if m.UseCache {
-		cacheFile := m.CachePath + "/" + server + "_ipmi_cache"
+		cacheFile := filepath.Join(m.CachePath, server + "_ipmi_cache")
 		_, err := os.Stat(cacheFile)
 		if os.IsNotExist(err) {
 			dumpOpts := opts
@@ -328,7 +329,7 @@ func init() {
 	}
 	m.Timeout = internal.Duration{Duration: time.Second * 20}
 	m.UseCache = true
-	m.CachePath = "/tmp"
+	m.CachePath = os.TempDir()
 	inputs.Add("ipmi_sensor", func() telegraf.Input {
 		m := m
 		return &m
--- a/plugins/inputs/ipmi_sensor/ipmi.go
+++ b/plugins/inputs/ipmi_sensor/ipmi.go
@@ -5,14 +5,14 @@ import (
 	"bytes"
 	"fmt"
 	"log"
+	"os"
 	"os/exec"
+	"path/filepath"
 	"regexp"
 	"strconv"
 	"strings"
 	"sync"
 	"time"
-	"os"
-	"path/filepath"
 
 	"github.com/influxdata/telegraf"
 	"github.com/influxdata/telegraf/internal"
@@ -132,7 +132,7 @@ func (m *Ipmi) parse(acc telegraf.Accumulator, server string) error {
 	}
 	opts = append(opts, "sdr")
 	if m.UseCache {
-		cacheFile := filepath.Join(m.CachePath, server + "_ipmi_cache")
+		cacheFile := filepath.Join(m.CachePath, server+"_ipmi_cache")
 		_, err := os.Stat(cacheFile)
 		if os.IsNotExist(err) {
 			dumpOpts := opts
--- a/plugins/inputs/ipmi_sensor/ipmi.go
+++ b/plugins/inputs/ipmi_sensor/ipmi.go
@@ -77,7 +77,7 @@ var sampleConfig = `
   ## If ipmitool should use a cache
   ## for me ipmitool runs about 2 to 10 times faster with cache enabled on HP G10 servers (when using ubuntu20.04)
   ## the cache file may not work well for you if some sensors come up late
-  # use_cache = true
+  # use_cache = false
 
   ## Path to the ipmitools cache file (defaults to OS temp dir)
   # cache_path = ""
@@ -329,7 +329,7 @@ func init() {
 		m.Path = path
 	}
 	m.Timeout = internal.Duration{Duration: time.Second * 20}
-	m.UseCache = true
+	m.UseCache = false
 	m.CachePath = os.TempDir()
 	inputs.Add("ipmi_sensor", func() telegraf.Input {
 		m := m
--- a/plugins/inputs/ipmi_sensor/ipmi.go
+++ b/plugins/inputs/ipmi_sensor/ipmi.go
@@ -75,10 +75,12 @@ var sampleConfig = `
   # hex_key = ""
 
   ## If ipmitool should use a cache
-  use_cache = true
+  ## for me ipmitool runs about 2 to 10 times faster enabled on HP G10 servers (when using ubuntu20.04)
+  ## the cache file may not work well for you if some sensensors come up late
+  # use_cache = true
 
-  ## Path to the ipmitools cache file (defaults to OS temp dir)
-  cache_path = "/tmp"
+  ## Path to the ipmitools cache file (defaults to OS temp dir)	  ## Path to the ipmitools cache file (defaults to OS temp dir e.g. /tmp)
+  # cache_path = ""
 `
 
 // SampleConfig returns the documentation about the sample configuration
@@ -149,7 +151,6 @@ func (m *Ipmi) parse(acc telegraf.Accumulator, server string) error {
 				return fmt.Errorf("failed to run command %s: %s - %s", strings.Join(cmd.Args, " "), err, string(out))
 			}
 		}
-		// -S /tmp/cache_file speeds up ipmitool runs (by 2 to 10 times on HP servers)
 		opts = append(opts, "-S")
 		opts = append(opts, cacheFile)
 	}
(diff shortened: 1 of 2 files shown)


plugins/inputs/ipmi_sensor/ipmi.go
```diff
@@ -65,6 +68,12 @@ var sampleConfig = `
 
   ## Schema Version: (Optional, defaults to version 1)
   metric_version = 2
+
+  ## If ipmitool should use a cache
+  use_cache = true
+
+  ## Path to the ipmitools cache file
+  cache_path = "/tmp/"
```
p-zak, line 76: Add them also to `README.md`

plugins/inputs/ipmi_sensor/ipmi.go
```diff
@@ -115,6 +124,30 @@ func (m *Ipmi) parse(acc telegraf.Accumulator, server string) error {
 		opts = conn.options()
 	}
 	opts = append(opts, "sdr")
+	if m.UseCache {
+		cacheFile := m.CachePath + "/" + server + "_ipmi_cache"
```
p-zak, line 128: Is this plugin meant for Linux only? I don't see any build tags or information in `README.md`.

But here I read that `ipmitool` can be run on Windows and MacOS: https://github.com/ipmitool/ipmitool/blob/master/INSTALL 
Does this plugin support this?

If yes, you should correct creating path - it should be OS independent.
If no, maybe this plugin should be built only for Linux?

@ssoroka @reimda what do you think?

reimda: This plugin should work on any platform that provides the ipmitool program.  I'm not sure which platforms actually have the BMC hardware in practice but the ipmitool program is available on many platforms.  For example, stable debian ships it for x86, arm, mips, ppc64, s390.  (https://packages.debian.org/buster/ipmitool)  It's also available on windows and mac os.

With this in mind I think it's appropriate to leave out the build tags and let all telegraf platforms get the plug, and also require platform independent path manipulation in the code.

DEvil0000: sorry, was not aware that this tool is also working on windows and mac.
this should be fixed with 6c652b7

plugins/inputs/ipmi_sensor/ipmi.go
DEvil0000: this should be fixed with 6c652b7

reimda (changes requested):

plugins/inputs/ipmi_sensor/ipmi.go
```diff
@@ -115,6 +125,30 @@ func (m *Ipmi) parse(acc telegraf.Accumulator, server string) error {
...
+				dumpOpts = append([]string{"-n", name}, dumpOpts...)
+				name = "sudo"
+			}
+			cmd := execCommand(name, dumpOpts...)
+			out, err := internal.CombinedOutputTimeout(cmd, m.Timeout.Duration)
+			if err != nil {
+				return fmt.Errorf("failed to run command %s: %s - %s", strings.Join(cmd.Args, " "), err, string(out))
+			}
+		}
+		// -S /tmp/cache_file speeds up ipmitool runs (by 2 to 10 times on HP servers)
```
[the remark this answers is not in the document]
DEvil0000: done with 7dab887
since I don't know the inner workings of ipmitool I can only do a informed guess about when you should not cache.
In most cases caching will help performance but nothing else however it may not hurt most people not using it.

plugins/inputs/ipmi_sensor/ipmi.go
```diff
@@ -290,6 +324,8 @@ func init() {
 		m.Path = path
 	}
 	m.Timeout = internal.Duration{Duration: time.Second * 20}
+	m.UseCache = true
```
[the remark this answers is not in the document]
DEvil0000: * caching uses less then 10kB per host for me.
* just in case cache is a issue for whatever unseen reason for someone i like it to be optional
* I don't know the inner workings of ipmitool so idk

plugins/inputs/ipmi_sensor/ipmi.go
```diff
@@ -65,6 +69,12 @@ var sampleConfig = `
 
   ## Schema Version: (Optional, defaults to version 1)
   metric_version = 2
+
+  ## If ipmitool should use a cache
+  use_cache = true
```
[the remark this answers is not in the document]
DEvil0000: done with 7dab887
this however is not standard in the readme or conifig examples.
other values like 20s timeout are also default and not commented out

plugins/inputs/ipmi_sensor/ipmi.go
```diff
@@ -65,6 +69,12 @@ var sampleConfig = `
...
+
+  ## Path to the ipmitools cache file (defaults to OS temp dir)
+  cache_path = "/tmp"
```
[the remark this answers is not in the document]
DEvil0000: done with 7dab887
I was under the impression everyone knows what /tmp means while not everyone may know what strange path windows uses. Especially since it was made clear in the comment what to expect.

DEvil0000:
@reimda I am under the impression your requested changes marked the checks failed. Can you tell me whats wrong about it now?

p-zak:
> @reimda I am under the impression your requested changes marked the checks failed. Can you tell me whats wrong about it now?

```
cd plugins/inputs/ipmi_sensor
go fmt ipmi.go
```

Moreover, please rebase on master.

DEvil0000:
rebased and formatted as requested.

Please add hacktoberfest-accepted label on merge

plugins/inputs/ipmi_sensor/ipmi.go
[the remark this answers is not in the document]
p-zak: Not these lines look like this:
`cache_path = "/tmp"	  # cache_path = ""`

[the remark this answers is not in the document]
DEvil0000: whoopsie, fixed

DEvil0000:
@reimda do you still see something?

zak-pawel:
@DEvil0000 PR still needs to be rebased.

DEvil0000:
@zak-pawel rebase done.

plugins/inputs/ipmi_sensor/ipmi.go
```diff
@@ -69,6 +73,14 @@ var sampleConfig = `
 
   ## Optionally provide the hex key for the IMPI connection.
   # hex_key = ""
+
+  ## If ipmitool should use a cache
+  ## for me ipmitool runs about 2 to 10 times faster enabled on HP G10 servers (when using ubuntu20.04)
+  ## the cache file may not work well for you if some sensensors come up late
```
zak-pawel, line 79: `sensensors` ?

plugins/inputs/ipmi_sensor/ipmi.go
```diff
@@ -69,6 +73,14 @@ var sampleConfig = `
...
+  # use_cache = true
+
+  ## Path to the ipmitools cache file (defaults to OS temp dir)	  ## Path to the ipmitools cache file (defaults to OS temp dir e.g. /tmp)
```
zak-pawel, line 82: This `## Path to the ipmitools cache file (defaults to OS temp dir e.g. /tmp)` seems to be redundant.

plugins/inputs/ipmi_sensor/ipmi.go
```diff
@@ -69,6 +73,14 @@ var sampleConfig = `
...
+  # cache_path = ""
```
zak-pawel, line 83: Content of `sampleConfig` should be exactly the same as sample config in `README.md`. Please sync these differences.

plugins/inputs/ipmi_sensor/ipmi.go
```diff
@@ -119,6 +131,29 @@ func (m *Ipmi) parse(acc telegraf.Accumulator, server string) error {
 		opts = conn.options()
 	}
 	opts = append(opts, "sdr")
+	if m.UseCache {
+		cacheFile := filepath.Join(m.CachePath, server+"_ipmi_cache")
```
zak-pawel, line 135: You should check if `CachePath` provided by user exists and is writeable directory.

zak-pawel (changes requested):

plugins/inputs/ipmi_sensor/ipmi.go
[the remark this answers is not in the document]
zak-pawel: @reimda Are you ok with this explanation?

plugins/inputs/ipmi_sensor/ipmi.go
DEvil0000: CachePath defaults to systems tmp dir which exists and is writable by default. Don't you think checking the directory is a bit out of scope of the application and should be left to the user when changing config?

plugins/inputs/ipmi_sensor/ipmi.go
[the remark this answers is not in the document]
reimda: The ipmitool program itself defaults to no caching and requires their users to turn it on with a command line switch.  I'm not comfortable with having telegraf change this and cache by default.  I understand that caching has worked well for @DEvil0000 but I don't feel like we have an explanation of whether switching caching on has the potential to cause trouble for existing telegraf users who upgrade.

I'm fine merging this PR if we make caching opt-in by defaulting to "use_cache = false" (and fix the typos you brought up in your review)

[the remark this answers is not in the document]
DEvil0000: okay, will change it to be turned off by default

plugins/inputs/ipmi_sensor/ipmi.go
zak-pawel: Honestly, I think either it should be checked here, or in README (and sample config) it should be stated that user needs to provide path to writeable directory where cache file can be stored. 
For security reasons, I prefer the former.

plugins/inputs/ipmi_sensor/ipmi.go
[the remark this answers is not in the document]
reimda: Thanks!

reimda (approved):

plugins/inputs/ipmi_sensor/ipmi.go
DEvil0000: In my eyes thats just a logical consequence for a given tmp dir (cache dir) that is must exist and must be writable. Anyway I added that to the description and readme now.
Since use cache is now off by default I don't see the need for adding more code for bizarre cases.

zak-pawel (approved):

PR #8346, opened 2020-10-30
Support prometheus remote write with http output

aarnaud:
### Required for all PRs:

```
  ## Exemple prometheus remote_write
  [[outputs.http]]
    url = "http://127.0.0.1:1234/receive"
    data_format = "prometheus"
    prometheus_write_format = "remote_write"
    [outputs.http.headers]
      Content-Type = "application/x-protobuf"
      Content-Encoding = "snappy"
      X-Prometheus-Remote-Write-Version = "0.1.0"
```

Tested with Cortex : https://cortexmetrics.io/


rickdonato:
Any ideas if/when this may be merged in?

aarnaud:
Maybe it's better to rename `prometheus_remote_write_format` by `prometheus_remote_write_protocol` ?

aarnaud:
> Maybe it's better to rename `prometheus_remote_write_format` by `prometheus_remote_write_protocol` ?

@ssoroka do you have preference ?

plugins/serializers/prometheus/prometheus.go
```diff
@@ -32,10 +38,20 @@ const (
 	StringAsLabel
 )
 
+// ProtobufFormating defines how to format body request.
+type Formating string
+
+const (
+	FormatText        Formating = "text"
+	FormatRemoteWrite Formating = "remote_write"
+	Format
```
ssoroka, line 47: blank intentional?

plugins/serializers/registry.go
```diff
@@ -103,6 +103,9 @@ type Config struct {
 	// Output string fields as metric labels; when false string fields are
 	// discarded.
 	PrometheusStringAsLabel bool `toml:"prometheus_string_as_label"`
+
+	// Output as text or protobuf
+	PrometheusRemoteWriteFormat bool `toml:"prometheus_remote_write_format"`
```
ssoroka, line 108: Seems like this would be better off as a string of either "text" or "remote_write", and maybe a generic name like WriteFormat

ssoroka (changes requested):
Maybe this should be its own serializer instead of attaching it to the existing prometheus serializer. I'm also pretty concerned it's loading way too many libraries, like all of the github.com/prometheus/prometheus project.

aarnaud:
thanks for the review, i will change that.

plugins/serializers/registry.go
aarnaud: No, I forgot to remove it, because before use boolean, I wrote code with string config.

plugins/serializers/prometheus/prometheus.go
aarnaud: No, I forgot to remove it, because before use boolean, I wrote code with string config.

plugins/serializers/registry.go
aarnaud: PR updated with a string configuration for prometheus write format.

aarnaud:
> Maybe this should be its own serializer instead of attaching it to the existing prometheus serializer. I'm also pretty concerned it's loading way too many libraries, like all of the github.com/prometheus/prometheus project.

I can try to use a dedicated serializer but, I think I have to duplicate code, like `collection.go` or `converter.go` that is present in prometheus serializer

aarnaud:
> I'm also pretty concerned it's loading way too many libraries, like all of the github.com/prometheus/prometheus project.

I need github.com/prometheus/prometheus/prompb but I can't import without prometheus project.

sjwang90:
Closing for separate Prometheus remote write serializer in PR #8360

PR #8383, opened 2020-11-10
Add dataset metrics to zfs input

zozoh94:
### Required for all PRs:

### DIFF
--- a/plugins/inputs/zfs/zfs.go
+++ b/plugins/inputs/zfs/zfs.go
@@ -2,13 +2,16 @@ package zfs
 
 type Sysctl func(metric string) ([]string, error)
 type Zpool func() ([]string, error)
+type Zdataset func(properties []string) ([]string, error)
 
 type Zfs struct {
-	KstatPath    string
-	KstatMetrics []string
-	PoolMetrics  bool
-	sysctl       Sysctl
-	zpool        Zpool
+	KstatPath      string
+	KstatMetrics   []string
+	PoolMetrics    bool
+	DatasetMetrics bool
+	sysctl         Sysctl
+	zpool          Zpool
+	zdataset       Zdataset
 }
 
 var sampleConfig = `
@@ -24,6 +27,8 @@ var sampleConfig = `
   #   "dmu_tx", "fm", "vdev_mirror_stats", "zfetchstats", "zil"]
   ## By default, don't gather zpool stats
   # poolMetrics = false
+  ## By default, don't gather zdataset stats
+  # datasetMetrics = false
 `
 
 func (z *Zfs) SampleConfig() string {
@@ -31,5 +36,5 @@ func (z *Zfs) SampleConfig() string {
 }
 
 func (z *Zfs) Description() string {
-	return "Read metrics of ZFS from arcstats, zfetchstats, vdev_cache_stats, and pools"
+	return "Read metrics of ZFS from arcstats, zfetchstats, vdev_cache_stats, pools and datasets"
 }
--- a/plugins/inputs/zfs/zfs.go
+++ b/plugins/inputs/zfs/zfs.go
@@ -1,5 +1,9 @@
 package zfs
 
+import (
+	"github.com/influxdata/telegraf"
+)
+
 type Sysctl func(metric string) ([]string, error)
 type Zpool func() ([]string, error)
 type Zdataset func(properties []string) ([]string, error)
@@ -12,6 +16,7 @@ type Zfs struct {
 	sysctl         Sysctl
 	zpool          Zpool
 	zdataset       Zdataset
+	Log            telegraf.Logger `toml:"-"`
 }
 
 var sampleConfig = `
--- a/plugins/inputs/zfs/zfs_freebsd.go
+++ b/plugins/inputs/zfs/zfs_freebsd.go
@@ -87,6 +87,46 @@ func (z *Zfs) gatherPoolStats(acc telegraf.Accumulator) (string, error) {
 	return strings.Join(pools, "::"), nil
 }
 
+func (z *Zfs) gatherDatasetStats(acc telegraf.Accumulator) (string, error) {
+	properties := []string{"name", "avail", "used", "usedsnap", "usedds"}
+
+	lines, err := z.zdataset(properties)
+	if err != nil {
+		return "", err
+	}
+
+	datasets := []string{}
+	for _, line := range lines {
+		col := strings.Split(line, "\t")
+
+		datasets = append(datasets, col[0])
+	}
+
+	if z.DatasetMetrics {
+		for _, line := range lines {
+			col := strings.Split(line, "\t")
+			if len(col) != len(properties) {
+				continue
+			}
+
+			tags := map[string]string{"dataset": col[0]}
+			fields := map[string]interface{}{}
+
+			for i, key := range properties[1:] {
+				value, err := strconv.ParseInt(col[i+1], 10, 64)
+				if err != nil {
+					return "", fmt.Errorf("Error parsing %s: %s", key, err)
+				}
+				fields[key] = value
+			}
+
+			acc.AddFields("zfs_dataset", fields, tags)
+		}
+	}
+
+	return strings.Join(datasets, "::"), nil
+}
+
 func (z *Zfs) Gather(acc telegraf.Accumulator) error {
 	kstatMetrics := z.KstatMetrics
 	if len(kstatMetrics) == 0 {
@@ -99,6 +139,11 @@ func (z *Zfs) Gather(acc telegraf.Accumulator) error {
 		return err
 	}
 	tags["pools"] = poolNames
+	datasetNames, err := z.gatherDatasetStats(acc)
+	if err != nil {
+		return err
+	}
+	tags["datasets"] = datasetNames
 
 	fields := make(map[string]interface{})
 	for _, metric := range kstatMetrics {
@@ -137,6 +182,10 @@ func zpool() ([]string, error) {
 	return run("zpool", []string{"list", "-Hp", "-o", "name,health,size,alloc,free,fragmentation,capacity,dedupratio"}...)
 }
 
+func zdataset(properties []string) ([]string, error) {
+	return run("zfs", []string{"list", "-Hp", "-o", strings.join(properties, ',')}...)
+}
+
 func sysctl(metric string) ([]string, error) {
 	return run("sysctl", []string{"-q", fmt.Sprintf("kstat.zfs.misc.%s", metric)}...)
 }
@@ -144,8 +193,9 @@ func sysctl(metric string) ([]string, error) {
 func init() {
 	inputs.Add("zfs", func() telegraf.Input {
 		return &Zfs{
-			sysctl: sysctl,
-			zpool:  zpool,
+			sysctl:   sysctl,
+			zpool:    zpool,
+			zdataset: zdataset,
 		}
 	})
 }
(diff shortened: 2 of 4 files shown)


GilgOuille:
Wow, excellent !

This is exactly what I need ! Very thanks @zozoh94 !

plugins/inputs/zfs/zfs_freebsd.go
```diff
@@ -137,15 +197,20 @@ func zpool() ([]string, error) {
 	return run("zpool", []string{"list", "-Hp", "-o", "name,health,size,alloc,free,fragmentation,capacity,dedupratio"}...)
 }
 
+func zdataset() ([]string, error) {
```
srebhan, line 200: How about getting the information to gather as a parameter? You could then synchronize with the `gatherDatasetStats()` function and do not keep the names in two different locations...

plugins/inputs/zfs/zfs_freebsd.go
```diff
@@ -87,6 +87,61 @@ func (z *Zfs) gatherPoolStats(acc telegraf.Accumulator) (string, error) {
 	return strings.Join(pools, "::"), nil
 }
 
+func (z *Zfs) gatherDatasetStats(acc telegraf.Accumulator) (string, error) {
+
+	lines, err := z.zdataset()
```
srebhan, line 92: You could define the information to gather here and pass it on to the `zdataset()` call like:
```suggestion
	properties := []string {"name", "avail", "used", "usedsnap", "usedds"}
	
	lines, err := z.zdataset(properties)
```

plugins/inputs/zfs/zfs_freebsd.go
```diff
@@ -87,6 +87,61 @@ func (z *Zfs) gatherPoolStats(acc telegraf.Accumulator) (string, error) {
...
+	if err != nil {
+		return "", err
+	}
+
+	datasets := []string{}
+	for _, line := range lines {
+		col := strings.Split(line, "\t")
+
+		datasets = append(datasets, col[0])
+	}
```
srebhan, line 102: The same is done when actually constructing the metric. Can we remove the duplication and keep the split information for metric gathering?

plugins/inputs/zfs/zfs_freebsd.go
```diff
@@ -87,6 +87,61 @@ func (z *Zfs) gatherPoolStats(acc telegraf.Accumulator) (string, error) {
...
+
+	if z.DatasetMetrics {
+		for _, line := range lines {
+			col := strings.Split(line, "\t")
+			if len(col) != 5 {
+				continue
+			}
```
srebhan, line 109: With the `properties` definition above we can make this easily extensible via
```suggestion
			if len(col) != len(properties) {
				continue
			}
```

plugins/inputs/zfs/zfs_freebsd.go
```diff
@@ -87,6 +87,61 @@ func (z *Zfs) gatherPoolStats(acc telegraf.Accumulator) (string, error) {
...
+			if err != nil {
+				return "", fmt.Errorf("Error parsing usedsnap: %s", err)
+			}
+			fields["usedsnap"] = usedsnap
+
+			usedds, err := strconv.ParseInt(col[4], 10, 64)
+			if err != nil {
+				return "", fmt.Errorf("Error parsing usedds: %s", err)
+			}
+			fields["usedds"] = usedds
```
srebhan, line 136: This is a re-occurring pattern. Assuming that all returned properties (except for the name) are integers we can do
```suggestion
		for i, key := range properties[1:] {
			value, err := strconv.ParseInt(col[i+1], 10, 64)
			if err != nil {
				return "", fmt.Errorf("Error parsing %s: %s", key, err)
			}
			fields[key] = value
		}
```

srebhan (changes requested):
Nice work. However, I think we can simplify this a bit more, see comments.

plugins/inputs/zfs/zfs_freebsd.go
```diff
@@ -87,6 +87,46 @@ func (z *Zfs) gatherPoolStats(acc telegraf.Accumulator) (string, error) {
...
+		col := strings.Split(line, "\t")
+
+		datasets = append(datasets, col[0])
+	}
+
+	if z.DatasetMetrics {
+		for _, line := range lines {
+			col := strings.Split(line, "\t")
+			if len(col) != len(properties) {
+				continue
```
srebhan, line 109: I still would like to see an INFO or WARN message here using the logging framework. Otherwise if e.g. the zfs tool changes the output we silently drop metrics. Is there a reason _NOT_ to add it?

srebhan (commented):
Almost there. :-) Could you please comment or fix the one comment on logging dropped lines!

I have one more general question tough: Your PR adds dataset-collection to FreeBSD but it sould also work on e.g. Linux. What about those poor souls? Do you think you can make it also work on Linux? I'll be able to test it here...

zozoh94:
@srebhan I added a log but I'm sorry but I have no experience with ZFS on linux.

plugins/inputs/zfs/zfs_freebsd.go
```diff
@@ -106,6 +107,7 @@ func (z *Zfs) gatherDatasetStats(acc telegraf.Accumulator) (string, error) {
 		for _, line := range lines {
 			col := strings.Split(line, "\t")
 			if len(col) != len(properties) {
+				log.Println("The number of columns in the command result doesn't match the number of columns expected.")
```
srebhan, line 110: Not quite. You have to add the logger
```
 Log      telegraf.Logger `toml:"-"`
```
to your data structure and here use something like
```
z.Log.Warnf("Invalid number of columns for line: %s", line)
```

srebhan:
Hey @zozoh94, I think linux behaves exactly the same as *BSD there. When calling `zfs list -Hp -o name,avail,used,usedsnap,usedds` on linux I get 
```
mytestpool      7318564864      482809856       0       482642432
mytestpool/mydataset1   7318564864      24576   0       24576
mytestpool/mydataset2   7318564864      24576   0       24576
mytestpool/mydataset3   7318564864      24576   0       24576
```
as output. So I guess it's pretty much the same you're seeing on FreeBSD!? However, I realized the Linux code relies on directly reading `/proc`... So never mind.

plugins/inputs/zfs/zfs_freebsd.go
```diff
@@ -8,6 +8,7 @@ import (
 	"os/exec"
 	"strconv"
 	"strings"
+	"log"
```
srebhan, line 11: You need to remove this otherwise golang will generate "unused import" warnings/errors.

plugins/inputs/zfs/zfs.go
```diff
@@ -12,6 +12,7 @@ type Zfs struct {
 	sysctl         Sysctl
 	zpool          Zpool
 	zdataset       Zdataset
+	Log            telegraf.Logger `toml:"-"`
```
srebhan, line 15: You are missing the import here.

plugins/inputs/zfs/zfs_freebsd.go
```diff
@@ -137,15 +183,20 @@ func zpool() ([]string, error) {
 	return run("zpool", []string{"list", "-Hp", "-o", "name,health,size,alloc,free,fragmentation,capacity,dedupratio"}...)
 }
 
+func zdataset(properties []string) ([]string, error) {
+	return run("zfs", []string{"list", "-Hp", "-o", strings.join(properties, ',')}...)
```
srebhan, line 187: This has to be
```suggestion
	return run("zfs", []string{"list", "-Hp", "-o", strings.Join(properties, ',')}...)
```
Did you compile and run `make check` on a FreeBSD machine?

plugins/inputs/zfs/zfs_freebsd.go
srebhan: You can mark this as resolved.

plugins/inputs/zfs/zfs_freebsd.go
srebhan: You can mark this as resolved.

plugins/inputs/zfs/zfs.go
srebhan: You can mark this as resolved.

plugins/inputs/zfs/zfs_freebsd.go
srebhan: You can mark this as resolved.

srebhan (approved):
Looks good. Thank you!

plugins/inputs/zfs/zfs_freebsd.go
```diff
@@ -87,6 +87,47 @@ func (z *Zfs) gatherPoolStats(acc telegraf.Accumulator) (string, error) {
...
+	properties := []string{"name", "avail", "used", "usedsnap", "usedds"}
+
+	lines, err := z.zdataset(properties)
+	if err != nil {
+		return "", err
+	}
+
+	datasets := []string{}
+	for _, line := range lines {
+		col := strings.Split(line, "\t")
```
ssoroka, line 100: is it always exactly one tab?

plugins/inputs/zfs/zfs_freebsd.go
```diff
@@ -87,6 +87,47 @@ func (z *Zfs) gatherPoolStats(acc telegraf.Accumulator) (string, error) {
...
+
+		datasets = append(datasets, col[0])
+	}
+
+	if z.DatasetMetrics {
+		for _, line := range lines {
+			col := strings.Split(line, "\t")
```
ssoroka, line 107: is it always exactly one tab?

plugins/inputs/zfs/zfs_freebsd.go
```diff
@@ -87,6 +87,47 @@ func (z *Zfs) gatherPoolStats(acc telegraf.Accumulator) (string, error) {
...
+				continue
+			}
+
+			tags := map[string]string{"dataset": col[0]}
+			fields := map[string]interface{}{}
+
+			for i, key := range properties[1:] {
+				value, err := strconv.ParseInt(col[i+1], 10, 64)
+				if err != nil {
+					return "", fmt.Errorf("Error parsing %s: %s", key, err)
```
ssoroka, line 119: parseInt uses col[i+1] but the error uses key. maybe change them both to key? can also use %q to quote strings

ssoroka (commented):
Looks pretty good. just some minor comments

plugins/inputs/zfs/zfs_freebsd.go
zozoh94: Yes

plugins/inputs/zfs/zfs_freebsd.go
zozoh94: @ssoroka key and col are not the same. Key is for column name and col is an array containing the values for 1 line of result.

zozoh94: @ssoroka I don't understand want you want to do?

ssoroka: Ok in that case I'd like to change the error message to include what was difficult to parse. eg:
```suggestion
				value, err := strconv.ParseInt(col[i+1], 10, 64)
				if err != nil {
					return "", fmt.Errorf("Error parsing %s %q: %s", key, col[i+1], err)
```

ssoroka (approved):

ssoroka:
Thanks!

toelke:
FWIW, the freebsd-code also works with linux, see a crappy port at https://github.com/toelke/telegraf/tree/port-dataset-to-linux

PR #8389, opened 2020-11-11
Add node groups to opcua input plugin

reimda:
Remove unneeded node settings (data_type and description).

Add the concept of a node group.  Groups have default settings for all nodes in the group.  They also are able to override the plugin's metric name so multiple metric names can be used in a single plugin instance.

Remove the name tag which wasn't useful because it duplicated the OPC UA value's field name.

Expose stats to telegraf internal plugin.

### DIFF
--- a/plugins/inputs/opcua/README.md
+++ b/plugins/inputs/opcua/README.md
@@ -98,6 +98,6 @@ To gather data from this node enter the following line into the 'nodes' property
 ### Example Output
 
 ```
-opcua,name=temp,id=n\=3;s\=Temperature temp=79.0,quality="OK (0x0)" 1597820490000000000
+opcua,id=n\=3;s\=Temperature temp=79.0,quality="OK (0x0)" 1597820490000000000
 
 ```
--- a/plugins/inputs/opcua/README.md
+++ b/plugins/inputs/opcua/README.md
@@ -10,7 +10,7 @@ Plugin minimum tested version: 1.16
 ```toml
 [[inputs.opcua]]
   ## Metric name
-  # name = "localhost"
+  # metric_name = "opcua"
   #
   ## OPC UA Endpoint URL
   # endpoint = "opc.tcp://localhost:4840"
@@ -47,15 +47,15 @@ Plugin minimum tested version: 1.16
   # password = ""
   #
   ## Node ID configuration
-  ## name       			- the variable name
-  ## namespace  			- integer value 0 thru 3
+  ## field_name			- the field name
+  ## namespace				- integer value 0 thru 3
   ## identifier_type		- s=string, i=numeric, g=guid, b=opaque
   ## identifier			- tag as shown in opcua browser
   ## Example:
   ## {name="ProductUri", namespace="0", identifier_type="i", identifier="2262"}
   nodes = [
-    {name="", namespace="", identifier_type="", identifier=""},
-    {name="", namespace="", identifier_type="", identifier=""},
+    {field_name="", namespace="", identifier_type="", identifier=""},
+    {field_name="", namespace="", identifier_type="", identifier=""},
   ]
 ```
 
@@ -66,13 +66,13 @@ An OPC UA node ID may resemble: "n=3;s=Temperature". In this example:
 - This example temperature node has a value of 79.0
 To gather data from this node enter the following line into the 'nodes' property above:
 ```
-{name="temp", namespace="3", identifier_type="s", identifier="Temperature"},
+{field_name="temp", namespace="3", identifier_type="s", identifier="Temperature"},
 ```
 
 
 ### Example Output
 
 ```
-localhost,name=temp,id=n\=3;s\=Temperature temp=79.0,quality="OK (0x0)" 1597820490000000000
+opcua,name=temp,id=n\=3;s\=Temperature temp=79.0,quality="OK (0x0)" 1597820490000000000
 
 ```
--- a/plugins/inputs/opcua/README.md
+++ b/plugins/inputs/opcua/README.md
@@ -10,7 +10,7 @@ Plugin minimum tested version: 1.16
 ```toml
 [[inputs.opcua]]
   ## Metric name
-  # metric_name = "opcua"
+  # name = "opcua"
   #
   ## OPC UA Endpoint URL
   # endpoint = "opc.tcp://localhost:4840"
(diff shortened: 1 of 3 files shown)


plugins/inputs/opcua/README.md
```diff
@@ -47,34 +47,57 @@ Plugin minimum tested version: 1.16
   # password = ""
   #
   ## Node ID configuration
-  ## name       			- the variable name
-  ## namespace  			- integer value 0 thru 3
-  ## identifier_type		- s=string, i=numeric, g=guid, b=opaque
-  ## identifier			- tag as shown in opcua browser
-  ## data_type  			- boolean, byte, short, int, uint, uint16, int16,
-  ##                        uint32, int32, float, double, string, datetime, number
+  ## field_name        - field name to use in the output
```
ssoroka, line 50: this documentation is a bit confusing; I thought these were top-level settings until I got to the example below. Not sure what we can do to improve it.

plugins/inputs/opcua/opcua_client.go
```diff
@@ -48,13 +49,24 @@ type OpcUA struct {
 }
 
 // OPCTag type
-type OPCTag struct {
-	Name           string `toml:"name"`
+type NodeSettings struct {
+	FieldName      string `toml:"field_name"`
```
ssoroka, line 53: backwards support for name might be good here.

plugins/inputs/opcua/opcua_client.go
```diff
@@ -13,11 +13,12 @@ import (
 	"github.com/influxdata/telegraf"
 	"github.com/influxdata/telegraf/config"
 	"github.com/influxdata/telegraf/plugins/inputs"
+	"github.com/influxdata/telegraf/selfstat"
 )
 
 // OpcUA type
 type OpcUA struct {
-	Name           string          `toml:"name"`
+	MetricName     string          `toml:"metric_name"`
```
ssoroka, line 21: backwards support for name might be good here.

plugins/inputs/opcua/opcua_client.go
```diff
@@ -48,13 +49,24 @@ type OpcUA struct {
...
 // OPCTag type
-type OPCTag struct {
-	Name           string `toml:"name"`
+type NodeSettings struct {
+	FieldName      string `toml:"field_name"`
 	Namespace      string `toml:"namespace"`
 	IdentifierType string `toml:"identifier_type"`
 	Identifier     string `toml:"identifier"`
-	DataType       string `toml:"data_type"`
-	Description    string `toml:"description"`
```
ssoroka, line 57: removing these will make old configs fail. A concern?

ssoroka (commented):
👍🏼

ssoroka (approved):

plugins/inputs/opcua/opcua_client.go
```diff
@@ -392,16 +441,15 @@ func (o *OpcUA) Gather(acc telegraf.Accumulator) error {
...
-			"name": n.Name,
-			"id":   BuildNodeID(n),
+			"id": n.idStr,
 		}
 
-		fields[o.NodeData[i].TagName] = o.NodeData[i].Value
-		fields["Quality"] = strings.TrimSpace(fmt.Sprint(o.NodeData[i].Quality))
-		acc.AddFields(o.Name, fields, tags)
+		fields[o.nodeData[i].TagName] = o.nodeData[i].Value
+		fields["quality"] = strings.TrimSpace(fmt.Sprint(o.nodeData[i].Quality))
```
srebhan, line 451: For case-sensitive downstream applications (such as InfluxDB IIRC) this might be a problem. First it will generate new series AFAIK as field names are used for hashing **and** it might break queries further down...
Should we really risk this breakage?

plugins/inputs/opcua/README.md
reimda: Nodes was a top level setting before this PR.  I added groups in this PR and each group also has a set of nodes.  I left in the top level nodes for backward compatibility.  I can see that this is confusing so I'll have to work on the docs.

plugins/inputs/opcua/opcua_client.go
reimda: I wasn't worried about backward compatibility when I started this PR because I thought this plugin was new enough and broken enough that there wouldn't be enough existing users to worry about.  Your and Sven's concerns made me change my mind so I've reverted this.

plugins/inputs/opcua/opcua_client.go
reimda: done

plugins/inputs/opcua/opcua_client.go
reimda: I added them back for compatibility but removed them from the readme because they were unnecessary and were never actually used.

plugins/inputs/opcua/opcua_client.go
reimda: Telegraf fields are usually lower case.  I wasn't worried about backward compatibility when I changed this, but you have valid concerns.  reverted.

reimda:
I reverted the changes that affect backward compatibility.  I think you both are right that this plugin has been released long enough that we should keep compatibility.

I also added a feature OPC UA users have requested, which is to be able to add tags to specific node IDs or to groups.  Sorry for sneaking this feature into this PR after the initial review but it seemed closely related and not big enough to start a separate PR.

plugins/inputs/opcua/opcua_client.go
```diff
@@ -210,50 +312,75 @@ func (o *OpcUA) InitNodes() error {
...
+	}
+	return x
+}
+
 func (o *OpcUA) validateOPCTags() error {
-	nameEncountered := map[string]bool{}
-	for i, item := range o.NodeList {
+	nameEncountered := map[metricParts]struct{}{}
+	for i := range o.nodes {
+		node := &o.nodes[i]
```
srebhan, line 347: Couldn't you get this directly in the for loop above?

plugins/inputs/opcua/opcua_client.go
```diff
@@ -210,50 +312,75 @@ func (o *OpcUA) InitNodes() error {
...
+		sb.WriteString("=")
+		sb.WriteString(n.metricTags[key])
+	}
+	x := metricParts{
+		metricName: n.metricName,
+		fieldName:  n.tag.FieldName,
+		tags:       sb.String(),
+	}
+	return x
+}
```
srebhan, line 342: This function might benefit speedwise from the optimization done here: https://github.com/influxdata/telegraf/pull/8391.

srebhan (approved):
Two very minor comments. Looks good to me.

plugins/inputs/opcua/opcua_client.go
reimda: Since this validation is only needed once during plugin initialization I didn't think optimization was worthwhile

plugins/inputs/opcua/opcua_client.go
reimda: fixed, thanks

PR #8391, opened 2020-11-11
Optimize SeriesGrouper & aggregators.merge

phemmer:
This introduces some optimizations for use with the aggregators.merge plugin. In my current usage of telegraf, it's heavily bottle-necking on this plugin.

----

The previous implementation of `SeriesGrouper` required breaking a metric object apart into its constituents, converting tags and keys into unoptimized maps, only to have it put them back together into another metric object. This resulted in a significant performance overhead. This overhead was further compounded when the number of fields was large.

This change adds a new `AddMetric` method to `SeriesGrouper` which preserves the metric object and removes the back-and-forth conversion.

Additionlly the method used for calculating the metric's hash was switched to use `maphash`, which is optimized for this case.

----

Benchmarks

Before:

    BenchmarkMergeOne-16          106012	     11790 ns/op
    BenchmarkMergeTwo-16           48529	     24819 ns/op
    BenchmarkGroupID-16           780018	      1608 ns/op

After:

    BenchmarkMergeOne-16          907093	      1173 ns/op
    BenchmarkMergeTwo-16          508321	      2168 ns/op
    BenchmarkGroupID-16         11217788	      99.4 ns/op

Note the `BenchmarkGroupID` isn't a direct apples-to-apples comparison, since the old method had the sort code present, and the new one does not. However earlier in my development, switching just the hash, while keeping all the other code resulted in about a 2x performance increase (831 ns/op).

----

Also since the vast majority of metrics are likely not to result in any mergers, a further optimization might be to utilize copy-on-write. Meaning use the input metric without copying it, and if a second metric is merged into it, then copy it. However the current change is a good first step, and that can be investigated later.

### Required for all PRs:

### DIFF
--- a/metric/series_grouper.go
+++ b/metric/series_grouper.go
@@ -1,10 +1,9 @@
 package metric
 
 import (
-	"hash/fnv"
-	"io"
+	"encoding/binary"
+	"hash/maphash"
 	"sort"
-	"strconv"
 	"time"
 
 	"github.com/influxdata/telegraf"
@@ -23,14 +22,17 @@ import (
 // + cpu,host=localhost idle_time=42,usage_time=42
 func NewSeriesGrouper() *SeriesGrouper {
 	return &SeriesGrouper{
-		metrics: make(map[uint64]telegraf.Metric),
-		ordered: []telegraf.Metric{},
+		metrics:  make(map[uint64]telegraf.Metric),
+		ordered:  []telegraf.Metric{},
+		hashSeed: maphash.MakeSeed(),
 	}
 }
 
 type SeriesGrouper struct {
 	metrics map[uint64]telegraf.Metric
 	ordered []telegraf.Metric
+
+	hashSeed maphash.Seed
 }
 
 // Add adds a field key and value to the series.
@@ -41,8 +43,15 @@ func (g *SeriesGrouper) Add(
 	field string,
 	fieldValue interface{},
 ) error {
+	taglist := make([]*telegraf.Tag, 0, len(tags))
+	for k, v := range tags {
+		taglist = append(taglist,
+			&telegraf.Tag{Key: k, Value: v})
+	}
+	sort.Slice(taglist, func(i, j int) bool { return taglist[i].Key < taglist[j].Key })
+
 	var err error
-	id := groupID(measurement, tags, tm)
+	id := groupID(g.hashSeed, measurement, taglist, tm)
 	metric := g.metrics[id]
 	if metric == nil {
 		metric, err = New(measurement, tags, map[string]interface{}{field: fieldValue}, tm)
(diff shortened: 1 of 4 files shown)


phemmer:
I haven't a clue whats up with the test failure.

macdeps:

```
github.com/influxdata/telegraf/metric imports
	hash/maphash: malformed module path "hash/maphash": missing dot in first path element
```

... No it's not

reimda:
I took a look at the macdeps failure but I haven't figured out yet why it's failing.  First I reran it and it failed again.  I noticed when rerunning it that macdeps installs go version 1.13.6 when it intends to install the latest go package from homebrew.  Maybe 'go mod tidy' fails only on go 1.13.6?  (The other CI steps use 1.14 or 1.15.)

phemmer:
Ah, that would explain it. maphash was introduced in 1.14.

Given that `go.mod` has the version set to 1.15, should we not even be testing against lower versions?

reimda:
I just noticed that hash/maphash is new in 1.14 too.
https://golang.org/doc/go1.14#hash/maphash

Telegraf is currently supporting go 1.14 and 1.15 even though go.mod has the version set to 1.15.  Fortunately the 1.14 CI steps are working even with the go.mod version set to 1.15.

To fix this failure we'll have to fix the macdeps job to use a newer golang version.  I think it's just a matter of updating .circleci/config.yml to run 'brew update' before running 'brew install go'.  Could you make that change and try it out in this PR?

reimda:
It looks like that worked.  Unfortunately 'brew update' took over four minutes to run.  We may need to look for a faster way to get an up to date golang install, but you don't need to worry about that for this PR.  Thanks for your help debugging CI!

phemmer:
Yay.

> We may need to look for a faster way to get an up to date golang install

Maybe it should use the official package/archive. I know I've seen issues on the golang issue tracker where they won't support people using the homebrew package, and tell these people to go reproduce issues using the official package. I'm unfamiliar with the background, but some googling suggests that homebrew does janky things with the package.

metric/series_grouper.go
```diff
@@ -57,30 +63,48 @@ func (g *SeriesGrouper) Add(
...
 // Metrics returns the metrics grouped by series and time.
 func (g *SeriesGrouper) Metrics() []telegraf.Metric {
 	return g.ordered
 }
 
-func groupID(measurement string, tags map[string]string, tm time.Time) uint64 {
-	h := fnv.New64a()
-	h.Write([]byte(measurement))
-	h.Write([]byte("\n"))
+var seed = maphash.MakeSeed()
```
srebhan, line 88: Can't this be done in `NewSeriesGrouper()` maybe together with the `SetSeed()` call?

srebhan, line 88: How about initializing the seed in `NewSeriesGrouper()` and pass it as an argument to `groupID()`?

srebhan (commented):
Code looks good. Maybe we could move the seed init to the `NewSeriesGrouper()` call? This would get rid of the global variable and generates an individual seed per grouper if I'm not mistaken...

Could we also move the CircleCI fix out here? It's just a minor thing, but nobody will afterwards search this PR for that... :-)

reimda:
I updated the golang version for mac ci in #8516.  Could you remove the 'brew update' from .circleci/config.yml and rebase this PR?  Thanks!

srebhan:
@phemmer any news?

phemmer:
Updated to address comments.

srebhan (approved):
Looks good to me.

ssoroka (approved):

ssoroka:
Great work, thanks!

PR #8392, opened 2020-11-11
Add output plugin for Honeycomb

puckpuck:
Adds an output plugin for [Honeycomb](https://honeycomb.io). 

### Required for all PRs:


puckpuck:
@sjwang90 do I need to do anything to get this plugin reviewed?

sjwang90:
Hey @puckpuck, thanks for the PR!

Would you be willing to have this plugin as an [external plugin](https://github.com/influxdata/telegraf/blob/master/docs/EXTERNAL_PLUGINS.md) that can be used with `execd` to run seamlessly with Telegraf? 

External plugins don't need to go through the typical review process and people can start using the plugin sooner. 

Otherwise we'll add this to the backlog of items to review.

puckpuck:
If we can get this added to the backlog that would be great.  We have a way to use `execd` as well, but making it part of telegraf itself enables a better process and discoverability.

Thanks.

plugins/outputs/honeycomb/README.md
```diff
@@ -0,0 +1,34 @@
...
+  dataset = "my-dataset"  
+
+  ## Special tags that will not get prefixed by the measurement name
+  ## This should be set if you specified global tags, and it should include the list of all global tags + host
+  ## Default value for this list is: host
+  #specialTags = ["host"]
+  
+  ## Optional: the hostname for the Honeycomb API server
+  #apiHost = "https://api.honeycomb.io/""
+```
```
srebhan, line 22: Please use snake_case in TOML.
```suggestion
```toml
  ## Honeycomb authentication token
  api_key = "API_KEY"

  ## Dataset name in Honeycomb to send data to
  dataset = "my-dataset"  

  ## Special tags that will not get prefixed by the measurement name
  ## This should be set if you specified global tags, and it should include the list of all global tags + host
  ## Default value for this list is: host
  #non_prefixed_tags = ["host"]
  
  ## Optional: the hostname for the Honeycomb API server
  #url = "https://api.honeycomb.io/""
```
```

File: plugins/outputs/honeycomb/honeycomb.go [hunk 2]
```diff
@@ -0,0 +1,133 @@
...
+)
+
+type Honeycomb struct {
+	APIKey      string   `toml:"apiKey"`
+	Dataset     string   `toml:"dataset"`
+	APIHost     string   `toml:"apiHost"`
+	SpecialTags []string `toml:"specialTags"`
+}
+
+var sampleConfig = `
```
srebhan, line 18:
```suggestion
const sampleConfig = `
```

We now follow a different scheme to handle `sampleConfig`, which adds a file `sample.conf` containing the config and the embedding this config into the code and the README. You can checkout the [example plugin](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/example) for details.

plugins/outputs/honeycomb/honeycomb.go
```diff
@@ -0,0 +1,133 @@
...
+  ## This should be set if you specified global tags, and it should include the list of all global tags + host
+  ## Default value for this list is: host
+  #specialTags = ["host"]
+  
+  ## Optional: the hostname for the Honeycomb API server
+  #apiHost = "https://api.honeycomb.io/""
+`
+
+func (h *Honeycomb) Connect() error {
+
```
srebhan, line 35: Please remove the newline.

plugins/outputs/honeycomb/honeycomb.go
```diff
@@ -0,0 +1,133 @@
...
+	if h.APIKey == "" {
+		return errors.New("Honeycomb apiKey can not be empty")
+	}
+
+	if h.Dataset == "" {
+		return errors.New("Honeycomb dataset can not be empty")
+	}
+
```
srebhan, line 43: These checks should be done in an `func (h *Honeycomb) Init() error` function that is called by Telegraf _after_ setting the user-configured parameters.

plugins/outputs/honeycomb/honeycomb.go
```diff
@@ -0,0 +1,133 @@
...
+		APIKey:       h.APIKey,
+		Dataset:      h.Dataset,
+		APIHost:      h.APIHost,
+		MaxBatchSize: 500,
+	})
+	if err != nil {
+		return fmt.Errorf("Honeycomb Init error: %s", err.Error())
+	}
+
+	return nil
```
srebhan, line 54: As this is the only error here, you could just return it.
```suggestion
	return libhoney.Init(libhoney.Config{
		APIKey:       h.APIKey,
		Dataset:      h.Dataset,
		APIHost:      h.APIHost,
		MaxBatchSize: 500,
	})
```

plugins/outputs/honeycomb/honeycomb.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+func (h *Honeycomb) Write(metrics []telegraf.Metric) error {
+
+	for _, m := range metrics {
+
```
srebhan, line 60: Please remove the extra newlines.

plugins/outputs/honeycomb/honeycomb.go
```diff
@@ -0,0 +1,133 @@
...
+		// BuildEvent from metric
+		ev, err := h.BuildEvent(m)
+		if err != nil {
+			return fmt.Errorf("Honeycomb event creation error: %s", err.Error())
+		}
+
+		// send event
+		if err = ev.Send(); err != nil {
+			return fmt.Errorf("Honeycomb Send error: %s", err.Error())
```
srebhan, line 69: Error messages should start lower-case. Furthermore, the plugin is always prefixed during logging, so please use this format
```suggestion
			return fmt.Errorf("event creation failed: %v", err)
		}

		// send event
		if err = ev.Send(); err != nil {
			return fmt.Errorf("send failed: %v", err)
```

plugins/outputs/honeycomb/honeycomb.go
```diff
@@ -0,0 +1,133 @@
...
+		}
+	}
+
+	libhoney.Flush()
+
+	return nil
+}
+
+func (h *Honeycomb) BuildEvent(m telegraf.Metric) (*libhoney.Event, error) {
```
srebhan, line 78: I think you don't need to export this function, do you?
```suggestion
func (h *Honeycomb) buildEvent(m telegraf.Metric) (*libhoney.Event, error) {
```

plugins/outputs/honeycomb/honeycomb.go
```diff
@@ -0,0 +1,133 @@
...
+			if t.Key == st {
+				prefixTag = false
+				break
+			}
+		}
+		if prefixTag {
+			data[m.Name()+"."+t.Key] = t.Value
+		} else {
+			data[t.Key] = t.Value
+		}
```
srebhan, line 101: Either use [`choice.Contains`](https://github.com/influxdata/telegraf/blob/master/internal/choice/choice.go) or a [filter](https://github.com/influxdata/telegraf/blob/master/filter/filter.go) here
```suggestion
		if choice.Contains(t,Key, h.SpecialTags) {
			data[t.Key] = t.Value
		} else {
			data[m.Name()+"."+t.Key] = t.Value
		}
```

plugins/outputs/honeycomb/honeycomb_test.go
```diff
@@ -0,0 +1,124 @@
...
+	"github.com/honeycombio/libhoney-go"
+	"github.com/honeycombio/libhoney-go/transmission"
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/metric"
+	"reflect"
+	"testing"
+	"time"
+)
+
+func MockInit() {
```
srebhan, line 13: Please do not export this function.

plugins/outputs/honeycomb/honeycomb_test.go
```diff
@@ -0,0 +1,124 @@
...
+		APIHost:      "http://localhost:1234",
+		Transmission: &transmission.MockSender{},
+	})
+}
+
+func TestWrite(t *testing.T) {
+	h := Honeycomb{}
+	MockInit()
+
+	testMetric, _ := metric.New("testName", map[string]string{"tag1": "value1", "tag2": "value2"}, map[string]interface{}{"foo": 1, "bar": 100.123}, time.Now())
```
srebhan, line 26:
```suggestion
	testMetric := metric.New(
		"testName", 
		map[string]string{"tag1": "value1", "tag2": "value2"},
		map[string]interface{}{"foo": 1, "bar": 100.123}
		time.Now(),
	)
```

plugins/outputs/honeycomb/honeycomb_test.go
```diff
@@ -0,0 +1,124 @@
...
+
+	if err := h.Write([]telegraf.Metric{testMetric}); err != nil {
+		t.Error(err)
+	}
```
srebhan, line 30:
```suggestion
	require.NoError(t, h.Write([]telegraf.Metric{testMetric}))
```

plugins/outputs/honeycomb/honeycomb_test.go
```diff
@@ -0,0 +1,124 @@
...
+		"testName.bar":  100.123,
+	})
+	testEvent.Timestamp = testTime
+	testMetric, _ := metric.New("testName", map[string]string{"tag1": "value1", "tag2": "value2"}, map[string]interface{}{"foo": 1, "bar": 100.123}, testTime)
+
+	retEvent, err := h.BuildEvent(testMetric)
+
+	if err != nil {
+		t.Error(err)
+	}
```
srebhan, line 53:
```suggestion
	retEvent, err := h.BuildEvent(testMetric)
	require.NoError(t, err)
```

plugins/outputs/honeycomb/honeycomb_test.go
```diff
@@ -0,0 +1,124 @@
...
+
+	if !reflect.DeepEqual(testEvent.Fields(), retEvent.Fields()) {
+		t.Errorf("\nexpected\t%+v \nreceived\t%+v \n", testEvent.Fields(), retEvent.Fields())
+	}
+	if testEvent.Timestamp != retEvent.Timestamp {
+		t.Errorf("\nexpected\t%+v \nreceived\t%+v \n", testEvent.Timestamp, retEvent.Timestamp)
+	}
```
srebhan, line 60: Maybe
```suggestion
	require.Equal(t, testEvent.Fields(), retEvent.Fields())
	require.Equal(t, testEvent.Timestamp, retEvent.Timestamp)
```

srebhan (commented):
Thanks for your contribution @puckpuck! The code looks quite good. I have some minor suggestions/comments in the code. Please have a look.

PR #8397, opened 2020-11-12
Dynatrace Plugin: Make conversion to counters possible / Changed large bulk handling

thschue:
### Required for all PRs:

Some input plugins do not set a ValueType for counters, which causes them to be handled as gauges in dynatrace. This PR enables the definition of such metrics to be handled as counters


zak-pawel:
@thschue Is this PR ready for review? Still see some `Println`s, lack of `README.md` update and so on...

thschue:
> @thschue Is this PR ready for review? Still see some `Println`s, lack of `README.md` update and so on..

The PR is ready now

plugins/outputs/dynatrace/README.md
```diff
@@ -35,6 +35,8 @@ You will also need to configure an API token for secure access. Find out how to
   prefix = "telegraf."
   ## Flag for skipping the tls certificate check, just for testing purposes, should be false by default
   insecure_skip_verify = false
+  ## If you want to convert values represented as gauges to counters, add the metric names here
```
zak-pawel, line 38: According to this: https://github.com/influxdata/telegraf/wiki/Review#reviewing-plugin-code: `SampleConfig must match the readme, but not include the plugin name.`.
Right now they are not in sync (both comments and variables)

plugins/outputs/dynatrace/dynatrace.go
```diff
@@ -219,6 +232,15 @@ func (d *Dynatrace) Write(metrics []telegraf.Metric) error {
 				default:
 					fmt.Fprintf(&buf, "%s%s %v\n", metricID, tagb.String(), value)
 				}
+
+				if metricCounter%1000 == 0 {
```
zak-pawel, line 236: Shouldn't `1000` be configurable or at least not used as a magic number?

zak-pawel (changes requested):

plugins/outputs/dynatrace/README.md
thschue: Fixed this in the SampleConfig

plugins/outputs/dynatrace/dynatrace.go
thschue: This is to ensure that the bulks sent to the api are not too large (that's also the reason, why it isn't configurable)

ssoroka: keep in mind if this splits it into chunks and the last chunk fails, all chunks will be retried(re-sent) when returning an error. When returning a nil, no chunks will be retried(re-sent).
I wonder if instead we should allow plugins to request that the batch size not exceed a certain limit.

ssoroka (approved):

ssoroka:
see last comment

plugins/outputs/dynatrace/dynatrace.go
dyladan: Since that isn't possible now, I think the solution here is sufficient. 1000 is a hard limit set by the dynatrace api rate limiting so a configuration wouldn't do any good.

Retrying the whole batch if the last chunk fails is unfortunate and something i'm taking a note of that I might fix in a follow up. Thanks for the heads-up.

arminru: Hey @zak-pawel! You still have a _request changes_ in place for this comment. Would you be fine with merging this now? Thanks.

zak-pawel: @arminru I understand why it should not be configurable. 
But I don't understand why this magic number cannot be converted to const. You can name it `dynatraceAPIRateLimit` (or similar) and everyone will be happy :)

arminru: Sure, making it a const for better readability makes sense - will do!

zak-pawel (approved):

plugins/outputs/dynatrace/dynatrace.go
zak-pawel: @arminru Now it's great.

arminru: Thanks!
@ssoroka can this be merged now? 🙂

arminru: Thanks!
@ssoroka can this be merged now? 🙂

arminru:
Hey, any chance this could be merged? Thanks!

arminru:
@helenosheaa @ssoroka I updated the PR to pick up the latest Revive fixes from the master branch.

PR #8398, opened 2020-11-12
Sensu Go Output Plugin for Telegraf

calebhailey:
Hello all 👋 we (Sensu) are looking forward to promoting Telegraf as a compatible agent for Sensu Go by adding upstream support for sending Telegraf metrics to Sensu Go (via this output plugin PR). 

I have signed the CLA, but it looks like our branch can't be automatically merged, so we'll look into that and see what changes need to be made and update the PR accordingly. 

Thanks in advance for your help to get this merged! #monitoringlove ❤️ 

/cc @pauldix :wave: 

### Required for all PRs:


plugins/outputs/sensu/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+This plugin is based off of the HTTP output plugin to send metrics
+to the Sensu Events API.
+
+### Configuration:
+
+```toml
+  ## Configure check configurations
+  [outputs.sensu-go.check]
+    name = "telegraf"
```
srebhan, line 11: Can you please detail a bit more on what it does? What is the `name` parameter about?

plugins/outputs/sensu/README.md
```diff
@@ -0,0 +1,69 @@
...
+  # [outputs.sensu-go.entity]
+  #   name = "server-01"
+  #   namespace = "default"
+
+  # [outputs.sensu-go.tags]
+  #   source = "telegraf"
+
+  # [outputs.sensu-go.metrics]
+  #   handlers = ["elasticsearch","timescaledb"]
+```
```
srebhan, line 69: Please provide some output example for the people to see what to expect. You might also want to explain some output parameters (e.g. `name` or `namespace`) below. Just imagine someone new to Sensu will read this and wants to set this up. Maybe provide references to the Sensu docu for explaining some workings.

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,468 @@
...
+}
+
+// SampleConfig provides a sample configuration for the plugin
+func (s *Sensu) SampleConfig() string {
+	return sampleConfig
+}
+
+func (s *Sensu) SetSerializer(serializer serializers.Serializer) {
+	s.serializer = serializer
+}
```
srebhan, line 126: You don't need this function as you don't support different serialization formats.

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,468 @@
...
+func (s *Sensu) createClient(ctx context.Context) (*http.Client, error) {
+	tlsCfg, err := s.ClientConfig.TLSConfig()
+	if err != nil {
+		return nil, err
+	}
+
+	client := &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: tlsCfg,
+			Proxy:           http.ProxyFromEnvironment,
```
srebhan, line 137: Maybe it would be nice to add a user-configurable option for the proxy. IIRC getting it from environment is the default behavior of `http.Transport` so you would not require setting it explicitly. However, I believe that allowing the user to set a proxy for the agent is a good thing especially in enterprise environments.

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,468 @@
...
+
+	return client, nil
+}
+
+func (s *Sensu) Connect() error {
+	if s.Timeout.Duration == 0 {
+		s.Timeout.Duration = defaultClientTimeout
+	}
+
+	ctx := context.Background()
```
srebhan, line 150: Would it make sense to also apply the `timeout` option here?

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,468 @@
...
+		return err
+	}
+
+	s.client = client
+
+	return nil
+}
+
+func (s *Sensu) Close() error {
+	return nil
```
srebhan, line 162: Please call `s.Client.CloseIdleConnections()` here.

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,468 @@
...
+	Name      string       `json:"name"`
+	Tags      []*OutputTag `json:"tags"`
+	Value     interface{}  `json:"value"`
+	Timestamp int64        `json:"timestamp"`
+}
+
+type OutputTag struct {
+	Name  string `json:"name"`
+	Value string `json:"value"`
+}
```
srebhan, line 468: Can you please group this with the other struct definitions on top so it's easier to find them as you only need to look at one place.

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,468 @@
...
+		}
+	})
+}
+
+type OutputEvent struct {
+	Entity    *OutputEntity  `json:"entity,omitempty"`
+	Check     *OutputCheck   `json:"check"`
+	Metrics   *OutputMetrics `json:"metrics"`
+	Timestamp int64          `json:"timestamp"`
+}
```
srebhan, line 320: Please move this up to the place where the other structs are defined so we only have to check one place when looking for structs.

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,468 @@
...
+
+func (s *Sensu) Write(metrics []telegraf.Metric) error {
+	var points []*OutputMetric
+	for _, metric := range metrics {
+		// Add tags from config to each metric point
+		var tagList []*OutputTag
+		for name, value := range s.Tags {
+			metric.AddTag(name, value)
+		}
```
srebhan, line 172: I don't like modifying the input metrics. Please add those tags to `tagList` directly! Maybe it would also be nice to pre-allocate the `tagList` as you can expect `len(s.Tags) + len(metric.TagList())` entries.

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,468 @@
...
+
+		// Get all valid numeric values, convert to float64
+		for _, fieldSet := range metric.FieldList() {
+			key := fieldSet.Key
+			// don't need err since math.NaN is returned with err
+			value, _ := getFloat(fieldSet.Value)
+			// JSON does not support these special values
+			if math.IsNaN(value) || math.IsInf(value, 0) {
+				continue
```
srebhan, line 188: How about replace at least the `Inf`s with `+-math.MaxFloat64`? Anyway, you should output at least some debug message notifying the user that you dropped a metric. If we silently drop metrics he/she has no way to find out what happened to their data...

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,468 @@
...
+			}
+
+			point := &OutputMetric{
+				Name:      metric.Name() + "." + key,
+				Tags:      tagList,
+				Timestamp: metric.Time().Unix(),
```
srebhan, line 194: So you only support second resolution?

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,468 @@
...
+	Entity    *OutputEntity  `json:"entity,omitempty"`
+	Check     *OutputCheck   `json:"check"`
+	Metrics   *OutputMetrics `json:"metrics"`
+	Timestamp int64          `json:"timestamp"`
+}
+
+func (s *Sensu) Wrapper(metricPoints []*OutputMetric) ([]byte, error) {
+	timestamp := time.Now().Unix()
+
+	var err error
```
srebhan, line 325: I think this is superfluous.

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,468 @@
...
+}
+
+type OutputEvent struct {
+	Entity    *OutputEntity  `json:"entity,omitempty"`
+	Check     *OutputCheck   `json:"check"`
+	Metrics   *OutputMetrics `json:"metrics"`
+	Timestamp int64          `json:"timestamp"`
+}
+
+func (s *Sensu) Wrapper(metricPoints []*OutputMetric) ([]byte, error) {
```
srebhan, line 322: Can we make this a private function?

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,468 @@
...
+				Tags:      tagList,
+				Timestamp: metric.Time().Unix(),
+				Value:     value,
+			}
+			points = append(points, point)
+		}
+	}
+
+	reqBody, err := s.Wrapper(points)
+
```
srebhan, line 202: Remove newline.

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,468 @@
...
+	if err != nil {
+		return err
+	}
+
+	if err := s.write(reqBody); err != nil {
+		return err
+	}
+
+	return nil
```
srebhan, line 211: You could do
```suggestion
	return s.write(reqBody)
```

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,468 @@
...
+}
+
+func (s *Sensu) write(reqBody []byte) error {
+	var reqBodyBuffer io.Reader = bytes.NewBuffer(reqBody)
```
srebhan, line 215:
```suggestion
	reqBodyBuffer := bytes.NewBuffer(reqBody)
```?

File: plugins/outputs/sensu/sensu.go [hunk 17]
```diff
@@ -0,0 +1,468 @@
...
+	}
+
+	return client, nil
+}
+
+func (s *Sensu) Connect() error {
+	if s.Timeout.Duration == 0 {
+		s.Timeout.Duration = defaultClientTimeout
+	}
+
```
srebhan, line 149: You should also check if the other options are valid to avoid user configuration errors. E.g. currently only `gzip` and `identity` is supported for `ContentEncoding`. You maybe should error out if the user specifies something else. Check the `internal.choice` package providing some convenient functions for doing so.

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,468 @@
...
+	return nil
+}
+
+func (s *Sensu) write(reqBody []byte) error {
+	var reqBodyBuffer io.Reader = bytes.NewBuffer(reqBody)
+	method := http.MethodPost
+
+	var err error
```
srebhan, line 218: This is not needed I think.

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,468 @@
...
+
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
+		return fmt.Errorf("when writing to [%s] received status code: %d", endpointUrl, resp.StatusCode)
+	}
+
+	return nil
+}
+
+// Resolves the event write endpoint
+func (s *Sensu) GetEndpointUrl() (string, error) {
```
srebhan, line 275: Can you please make this private.

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,468 @@
...
+			return err
+		}
+		defer rc.Close()
+		reqBodyBuffer = rc
+	}
+
+	endpointUrl, err := s.GetEndpointUrl()
+	if err != nil {
+		return err
+	}
```
srebhan, line 231: Can this be done in `Connect()` instead of doing it for each `Write()`? It seems the function does not require any metric-specific parameters.

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,468 @@
...
+
+	req.Header.Set("User-Agent", internal.ProductToken())
+
+	req.Header.Set("Content-Type", defaultContentType)
+	if s.ContentEncoding == "gzip" {
+		req.Header.Set("Content-Encoding", "gzip")
+	}
```
srebhan, line 243: Maybe also do this decision in `Connect()` and just set the resulting encoding. This removes some lines from here.

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,468 @@
...
+
+	if s.ApiKey != nil {
+		bearerType := "Bearer"
+		if s.BackendApiUrl != nil {
+			bearerType = "Key"
+		}
+		req.Header.Set("Authorization", bearerType+" "+*s.ApiKey)
```
srebhan, line 250: Please determine the `Authorization` parameter in `Connect()`. I like to keep `Write()` as slim as possible.

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,468 @@
...
+	}
+
+	for k, v := range s.Headers {
+		if strings.ToLower(k) == "host" {
+			req.Host = v
+		}
```
srebhan, line 256: `continue` here or do you also need to set it in the headers?

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,468 @@
...
+		req.Header.Set(k, v)
+	}
+
+	resp, err := s.client.Do(req)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+	_, err = ioutil.ReadAll(resp.Body)
```
srebhan, line 265: Why do you read the body if you are neither interested in the `err` nor in the content of the body?

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,468 @@
...
+	return endpointUrl, nil
+}
+
+func init() {
+	outputs.Add("sensu-go", func() telegraf.Output {
+		// Default configuration values
+		defaultUrl := defaultUrl
+
+		return &Sensu{
+			AgentApiUrl: &defaultUrl,
```
srebhan, line 309: Please check the URLs in `Connect()` as otherwise someone might specify `agent_api_url = ""` in the config and make you end up with both url-options being empty.

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,468 @@
...
+func (s *Sensu) GetEndpointUrl() (string, error) {
+	var endpointUrl string
+
+	if s.BackendApiUrl != nil {
+		endpointUrl = *s.BackendApiUrl
+	} else if s.AgentApiUrl != nil {
+		endpointUrl = *s.AgentApiUrl
+	}
+
+	u, err := url.Parse(endpointUrl)
```
srebhan, line 284: Currently `endpointUrl` might be `nil` in case the user specifies `""` for the `agent_api_url`. Please see my comment in the `init()` function.

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,468 @@
...
+
+	if s.BackendApiUrl != nil {
+		namespace := "default"
+		if s.Entity != nil && s.Entity.Namespace != nil {
+			namespace = *s.Entity.Namespace
+		}
+		u.Path = path.Join(u.Path, "/api/core/v2/namespaces", namespace, "events")
+	} else {
+		u.Path = path.Join(u.Path, "/events")
+	}
```
srebhan, line 297: Determine the "path_suffix" in the `if` above and then just do
```suggestion
	u.Path = path.Join(u.Path, path_suffix)
```
here. This way you keep the special handling of the URL options in one place.

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,468 @@
...
+	endpointUrl = u.String()
+
+	return endpointUrl, nil
+}
+
+func init() {
+	outputs.Add("sensu-go", func() telegraf.Output {
+		// Default configuration values
+		defaultUrl := defaultUrl
```
srebhan, line 306: Why?

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,468 @@
...
+		},
+		Timestamp: timestamp,
+	})
+
+	return output, err
+}
+
+// Constructs the entity payload
+// Throws when no entity name is provided and fails resolve to hostname
+func (s *Sensu) GetEntity() (*OutputEntity, error) {
```
srebhan, line 351: Can we make that private please.

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,468 @@
...
+	Check     *OutputCheck   `json:"check"`
+	Metrics   *OutputMetrics `json:"metrics"`
+	Timestamp int64          `json:"timestamp"`
+}
+
+func (s *Sensu) Wrapper(metricPoints []*OutputMetric) ([]byte, error) {
+	timestamp := time.Now().Unix()
+
+	var err error
+	entity, err := s.GetEntity()
```
srebhan, line 326: Can we determine the `entity` a single time in `Connect()` please to avoid e.g. asking for the hostname of the current machine in each `Write()`. :-)

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,468 @@
...
+		return []byte{}, err
+	}
+
+	check, err := s.GetCheck(metricPoints)
+	if err != nil {
+		return []byte{}, err
+	}
+
+	output, err := json.Marshal(&OutputEvent{
+		Entity: entity,
```
srebhan, line 337: `entity` might be `nil` here, is this ok?

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,468 @@
...
+// Constructs the entity payload
+// Throws when no entity name is provided and fails resolve to hostname
+func (s *Sensu) GetEntity() (*OutputEntity, error) {
+	if s.BackendApiUrl != nil {
+		var entityName string
+		if s.Entity != nil && s.Entity.Name != nil {
+			entityName = *s.Entity.Name
+		} else {
+			defaultHostname, err := os.Hostname()
+
```
srebhan, line 358: Remove newline.

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,468 @@
...
+			if err != nil {
+				return &OutputEntity{}, fmt.Errorf("when resolving hostname")
```
srebhan, line 360: Carrying the error upwards by
```suggestion
				return &OutputEntity{}, fmt.Errorf("resolving hostname failed: %v", err)
```
would be nice.

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,468 @@
...
+				Name: entityName,
+			},
+		}, nil
+	}
+	return nil, nil
+}
+
+// Constructs the check payload
+// Throws if check name is not provided
+func (s *Sensu) GetCheck(metricPoints []*OutputMetric) (*OutputCheck, error) {
```
srebhan, line 376: Please make this function private.

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,468 @@
...
+				Value: tagSet.Value,
+			}
+			tagList = append(tagList, tag)
+		}
+
+		// Get all valid numeric values, convert to float64
+		for _, fieldSet := range metric.FieldList() {
+			key := fieldSet.Key
+			// don't need err since math.NaN is returned with err
+			value, _ := getFloat(fieldSet.Value)
```
srebhan, line 185: Please check the error potentially returned here!

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,468 @@
...
+			Name: *s.Check.Name,
+		},
+		Status:               0, // Always OK
+		Issued:               time.Now().Unix(),
+		Output:               "Telegraf agent processed " + strconv.Itoa(count) + " metrics",
+		OutputMetricHandlers: s.GetHandlers(),
+	}, nil
+}
+
+func (s *Sensu) GetHandlers() []string {
```
srebhan, line 394: Please make this function private.

plugins/outputs/sensu/sensu_test.go
```diff
@@ -0,0 +1,69 @@
...
+	defer ts.Close()
+
+	_, err := url.Parse(fmt.Sprintf("http://%s", ts.Listener.Addr().String()))
+	require.NoError(t, err)
+
+	t.Run("default-user-agent", func(t *testing.T) {
+		var err = error(nil)
+		require.NoError(t, err)
+	})
+}
```
srebhan, line 24: What is this test for? Are you testing if the `httptest` server can be set-up? If this fails we will be in big trouble anyway I think.

srebhan (changes requested):
@calebhailey thanks for submitting this! The code looks quite nice already but I do have some comments. The most important one is to add more tests especially to also test the data(format) submitted.
I also have some minor comments (see code) where making internal functions private and moving more code to `Connect()` are the most outstanding ones.

I'm looking forward to the next round!

plugins/outputs/sensu/sensu.go
nixwiz: Your suggestion leads to the following error:

`plugins/outputs/sensu/sensu.go:284:17: cannot use rc (type io.ReadCloser) as type *bytes.Buffer in assignment`

Also this same line of code can be found in the [http outputs plugin](https://github.com/influxdata/telegraf/blob/master/plugins/outputs/http/http.go#L182).

plugins/outputs/sensu/sensu.go
nixwiz: Yes.  [Docs link](https://docs.sensu.io/sensu-go/latest/observability-pipeline/observe-events/events/#points-attributes) to type specification.

plugins/outputs/sensu/sensu.go
nixwiz: Agreed, removed.

plugins/outputs/sensu/sensu.go
nixwiz: Fixed.

plugins/outputs/sensu/sensu.go
nixwiz: To get the pointer to the string, defaultUrl is a `const` for the default URL.  Sensu.AgentApiUrl is a `*string` so we assign the constant to a string to get the pointer.  I will change the string variable name to avoid this confusion.

plugins/outputs/sensu/sensu.go
nixwiz: Since it is the default, I have removed it.  The agent we run on entities also makes use of environment variables for proxies, if necessary ([docs link](https://docs.sensu.io/sensu-go/latest/observability-pipeline/observe-schedule/agent/#use-environment-variables-to-specify-an-http-proxy-for-agent-use)) so I don't believe an option for it is necessary.

plugins/outputs/sensu/sensu.go
nixwiz: This was a holdover from using the http output plugin as a base, we are not setting headers in our config, so this was removed.

plugins/outputs/sensu/sensu.go
nixwiz: Removed.

plugins/outputs/sensu/sensu.go
srebhan: Ok, I just wanted to make sure... ;-) Close this as resolved then. (I cannot do this unfortunately...)

plugins/outputs/sensu/sensu.go
srebhan: Ok did't stumble upon this one. Then please keep as is and sorry for the noise.

srebhan:
@calebhailey I see you merged master and added some comments but the code didn't change. Just want to make sure this is intentional!? Let me know when feel ready for the next round. :-)

nixwiz:
> @calebhailey I see you merged master and added some comments but the code didn't change. Just want to make sure this is intentional!? Let me know when feel ready for the next round. :-)

@srebhan that is on me.  Yes, it is intentional.  Given the number of comments/requested changes I commented on many of them as I worked on them, but I've not yet checked in any of that code.

plugins/outputs/sensu/sensu.go
nixwiz: Yes, because if sending to the localhost agent API no entity is necessary.  It is only needed for sending to the backend API.

srebhan: Ok then please close as resolved.

calebhailey:
@srebhan thanks for your help with this BTW!

plugins/outputs/sensu/sensu.go
calebhailey: Added a timeout.

plugins/outputs/sensu/sensu.go
calebhailey: Added 👍

plugins/outputs/sensu/sensu.go
calebhailey: This should be fixed now.

plugins/outputs/sensu/sensu.go
calebhailey: This is fixed.

plugins/outputs/sensu/sensu.go
calebhailey: 👍

plugins/outputs/sensu/sensu.go
calebhailey: We're doing config validation now.

plugins/outputs/sensu/sensu.go
calebhailey: Removed.

plugins/outputs/sensu/sensu.go
nixwiz: Leaving this here with all of the other headers.

plugins/outputs/sensu/sensu.go
calebhailey: Added validation for this.

plugins/outputs/sensu/sensu.go
calebhailey: 👍

plugins/outputs/sensu/sensu_test.go
nixwiz: Removed this essentially useless test and added sufficient testing for connect and write, increased coverage from 12% to 70%.

calebhailey:
@srebhan - @nixwiz submitted his changes which should have resolved almost all of your initial feedback. We're looking into the failing tests now, but let us know if there are any other changes required for conformity. 👍

plugins/outputs/sensu/README.md
srebhan: Still open.

plugins/outputs/sensu/README.md
```diff
@@ -0,0 +1,85 @@
...
+  ## append the corresponding backend or agent API path (e.g. /events or 
+  ## /api/core/v2/namespaces/:entity_namespace/events/:entity_name/:check_name).
+  ## 
+  ## NOTE: if backend_api_url and agent_api_url and api_key are set, the output 
+  ## plugin will use backend_api_url. If backend_api_url and agent_api_url are 
+  ## not provided, the output plugin will default to use an agent_api_url of 
+  ## http://127.0.0.1:3031
+  ## 
+  # backend_api_url = "http://127.0.0.1:8080"
+  # agent_api_url = "http://127.0.0.1:3031"
```
srebhan, line 20: Could you also add a sentence regarding `agent_api_url`. Sorry I don't know the details of Sensu (yet) but I don't understand how `backend_api_url` and `agent_api_url` differ and what to use when...

plugins/outputs/sensu/README.md
```diff
@@ -0,0 +1,85 @@
...
+  ## - Metrics - https://docs.sensu.io/sensu-go/latest/reference/events#metrics
+  ## 
+  ## Check specification
+  ## The check name is the name to give the Sensu check associated with the event
+  ## created.
+  [outputs.sensu-go.check]
+    name = "telegraf"
+
+  ## Entity specification
+  ## Configure the entity name and namepsace, if necessary.
```
srebhan, line 67: Typo
```suggestion
  ## Configure the entity name and namespace, if necessary.
```

plugins/outputs/sensu/README.md
```diff
@@ -0,0 +1,85 @@
...
+  ## - Events - https://docs.sensu.io/sensu-go/latest/reference/events
+  ## - Checks -  https://docs.sensu.io/sensu-go/latest/reference/checks
+  ## - Entities - https://docs.sensu.io/sensu-go/latest/reference/entities
+  ## - Metrics - https://docs.sensu.io/sensu-go/latest/reference/events#metrics
+  ## 
+  ## Check specification
+  ## The check name is the name to give the Sensu check associated with the event
+  ## created.
+  [outputs.sensu-go.check]
+    name = "telegraf"
```
srebhan, line 64: So does this relate to the `metadata/name` property in the documentation you linked?

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,500 @@
...
+	"path"
+	"strconv"
+	"time"
+
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/internal"
+	"github.com/influxdata/telegraf/internal/choice"
+	"github.com/influxdata/telegraf/plugins/common/tls"
+	"github.com/influxdata/telegraf/plugins/outputs"
+	"github.com/influxdata/telegraf/plugins/serializers"
```
srebhan, line 23: Please remove, it's not required as you do not support serializers.

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,500 @@
...
+	Timeout         internal.Duration `toml:"timeout"`
+	ContentEncoding string            `toml:"content_encoding"`
+
+	EndpointUrl string
+	OutEntity   *OutputEntity
+
+	tls.ClientConfig
+	client *http.Client
+
+	serializer serializers.Serializer
```
srebhan, line 103: Please remove, it's not required as you do not support serializers.

plugins/outputs/sensu/README.md
nixwiz: Yes.

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,500 @@
...
+			TLSClientConfig: tlsCfg,
+		},
+		Timeout: s.Timeout.Duration,
+	}
+
+	return client, nil
+}
+
+func (s *Sensu) Connect() error {
+	if s.Timeout.Duration == 0 {
```
srebhan, line 212: I guess zero is not a valid value here (e.g. wait forever)? If however, zero is a valid value, then please set the default value in `init()` when constructing the struct. Asking, just in case... :-)

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,500 @@
...
+		return err
+	}
+
+	err = s.setEntity()
+	if err != nil {
+		return err
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), s.Timeout.Duration)
+	defer cancel()
```
srebhan, line 240: I don't think you need this as `createClient()` as no operation using the context and it will probably also send no packets over the line or am I mistaken?

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,500 @@
...
+func (s *Sensu) Close() error {
+	s.client.CloseIdleConnections()
+	return nil
+}
+
+func (s *Sensu) Write(metrics []telegraf.Metric) error {
+	var points []*OutputMetric
+	for _, metric := range metrics {
+		// Add tags from config to each metric point
+		var tagList []*OutputTag
```
srebhan, line 261: You might want to preallocate `tagList` by
```suggestion
		tagList := make([]*OutputTag, 0, len(s.Tags)+len(metric.TagList()))
```
to avoid reallocation.

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,500 @@
...
+	Tags          map[string]string `toml:"tags"`
+	Metrics       *SensuMetrics     `toml:"metrics"`
+	Check         *SensuCheck       `toml:"check"`
+
+	Timeout         internal.Duration `toml:"timeout"`
+	ContentEncoding string            `toml:"content_encoding"`
+
+	EndpointUrl string
+	OutEntity   *OutputEntity
+
```
srebhan, line 99: Please add 
```
Log        telegraf.Logger `toml:"-"`
```
here for later use (see comment below).

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,500 @@
...
+			tagList = append(tagList, tag)
+		}
+
+		// Get all valid numeric values, convert to float64
+		for _, fieldSet := range metric.FieldList() {
+			key := fieldSet.Key
+			value := getFloat(fieldSet.Value)
+			// JSON does not support these special values
+			if math.IsInf(value, 1) {
+				log.Printf("D! [outputs.sensu-go] metric %s returned positive infity, setting value to %f", key, math.MaxFloat64)
```
srebhan, line 283: Please use the telegraf logger (see the structure comment about `Log`)! You should do
```suggestion
				s.Log.Debugf("metric %s returned positive infity, setting value to %f", key, math.MaxFloat64)
```
instead to ensure consistent formatting of the log messages. Same for the other instances of `fmt.Printf`.

plugins/outputs/sensu/sensu.go
srebhan: Please mark as resolved.

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,500 @@
...
+	if s.ContentEncoding == "gzip" {
+		rc, err := internal.CompressWithGzip(reqBodyBuffer)
+		if err != nil {
+			return err
+		}
+		defer rc.Close()
+		reqBodyBuffer = rc
+	}
+
+	req, err := http.NewRequest(method, s.EndpointUrl, reqBodyBuffer)
```
srebhan, line 326: Hmm how about using a context with timeout here using `http.NewRequestWithContext()` to prevent some blocking requests to stall the output?

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,500 @@
...
+	var endpointUrl string
+
+	if s.BackendApiUrl != nil {
+		endpointUrl = *s.BackendApiUrl
+	} else if s.AgentApiUrl != nil {
+		endpointUrl = *s.AgentApiUrl
+	}
+
+	if len(endpointUrl) == 0 {
+		log.Printf("D! [outputs.sensu-go] no backend or agent API URL provided, falling back to default agent API URL %s", defaultUrl)
```
srebhan, line 366: Also convert to use the telegraf logger.

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,500 @@
...
+	var path_suffix string
+	if s.BackendApiUrl != nil {
+		namespace := "default"
+		if s.Entity != nil && s.Entity.Namespace != nil {
+			namespace = *s.Entity.Namespace
+		}
+		path_suffix = "/api/core/v2/namespaces/" + namespace + "/events"
+	} else {
+		path_suffix = "/events"
+	}
```
srebhan, line 384: Can you please fuse this with the if some lines above? It's easier to realize what's going on then.

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,500 @@
...
+func init() {
+	outputs.Add("sensu-go", func() telegraf.Output {
+		// Default configuration values
+
+		// make a string from the defaultUrl const
+		agentApiUrl := defaultUrl
+
+		return &Sensu{
+			AgentApiUrl: &agentApiUrl,
+			Timeout:     internal.Duration{Duration: defaultClientTimeout},
```
srebhan, line 400: If a zero timeout is not permitted, you should remove this line as you check for an empty condition anyway in `Connect()`.

srebhan (changes requested):
Very close! :-) I have some points require fixing:

1. Please remove the serializer leftovers.
2. Use `telegraf.Logger` instead of `fmt.Printf` to ensure a consistent output format.
3. Please check the context stuff again. Not sure we need them at all.
4. Please add `github.com/sensu/sensu-go/api/core/v2` to the list of dependencies in `docs/LICENSE_OF_DEPENDENCIES.md` (and please keep the alphabetic order ;-))

You might also want to add your plugin to the list of output plugins in the toplevel `README.md` file. I have some more comments in the code, please also take a look there.

plugins/outputs/sensu/sensu_test.go
srebhan: Good job! Please consider this resolved.

plugins/outputs/sensu/README.md
srebhan: Ok, so probably everybody using this plugins knows. If you like you might want to add a short notice that the properties are identical to the metadata ones in the documentation!?

plugins/outputs/sensu/sensu.go
srebhan: I'm not sure here and the documentation is vague in my view. So probably setting the timeout in `http.Client` is sufficient. What do you think?

plugins/outputs/sensu/README.md
nixwiz: This one is confusing me. I am not generating any new metrics that would be output via `telegraf --test`.  This output plugin takes telegraf metrics and creates Sensu events that contain those metrics.  On the backend, Sensu uses handlers as its destination for the metrics.  That is what is being configured here, is how the Sensu backend routes the metrics it ingests from the events generated by this Telegraf output plugin.

plugins/outputs/sensu/sensu.go
nixwiz: I think so.  I will remove the context bits.

plugins/outputs/sensu/README.md
srebhan: You are absolutely right! Sorry too much input plugins lately... Thank god it's Friday already. :-) Please consider this as resolved and accept my apologies.

plugins/outputs/sensu/sensu.go
srebhan: This one is also resolved I think.

plugins/outputs/sensu/sensu.go
nixwiz: Yeah, this seems redundant and zero should be a valid value.  I think I'll remove this portion and leave the setting of Timeout in `init()`.

plugins/outputs/sensu/README.md
nixwiz: No worries, we've all been there.

plugins/outputs/sensu/sensu.go
srebhan: Sounds reasonable.

nixwiz:
I submitted my latest batch of changes, but now have a failing test (should have ran a test locally before pushing).  If I don't get that sorted today, I will work on it early next week.

plugins/outputs/sensu/README.md
srebhan: Mark as resolved.

plugins/outputs/sensu/sensu.go
srebhan: Mark as resolved.

srebhan (commented):
The code looks good now and if you initialize the logger in your tests via `tt.plugin.Log = testutil.Logger{}` you also get those errors fixed! ;-)

nixwiz:
@srebhan I've got the code fixed, thanks for the pointer on that.

Now the CI jobs are failing due to the line you requested I add to docs/LICENSE_OF_DEPENDENCIES.md

The `make check-deps` will fail with the line I added for the Sensu license.

srebhan (approved):
Looks good to me. **The failing test is `tail` and not related to this plugin!**

I try to find out if or if not an addition to the list of dependencies is necessary, will come back to you once this is sorted.

Thanks for all your effort!

nixwiz:
And, thank you for your patience as I worked through this process.

srebhan (approved):

Jagularr (approved):
Looks good overall.

plugins/outputs/sensu/README.md
```diff
@@ -0,0 +1,96 @@
+# HTTP Output Plugin
+
+This plugin writes metrics events to [Sensu Go](https://sensu.io)  via its 
+HTTP events API.
+
+### Configuration:
+
+```toml
+  ## BACKEND API URL is the Sensu Backend API root URL to send metrics to
```
ssoroka, line 9:
```suggestion
```toml
[[outputs.sensu-go]]
  ## BACKEND API URL is the Sensu Backend API root URL to send metrics to 
```

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,501 @@
...
+	OutEntity   *OutputEntity
+
+	Log telegraf.Logger `toml:"-"`
+
+	tls.ClientConfig
+	client *http.Client
+}
+
+var sampleConfig = `
+## BACKEND API URL is the Sensu Backend API root URL to send metrics to
```
ssoroka, line 104: config should be indented with two spaces here.

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,501 @@
...
+}
+
+func (s *Sensu) Connect() error {
+	if len(s.ContentEncoding) != 0 {
+		validEncoding := []string{"identity", "gzip"}
+		if !choice.Contains(s.ContentEncoding, validEncoding) {
+			return fmt.Errorf("Unsupported content_encoding [%q] specified", s.ContentEncoding)
+		}
+	} else {
+		s.ContentEncoding = "identity"
```
ssoroka, line 226: set defaults in the init() function

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,501 @@
...
+
+	return client, nil
+}
+
+func (s *Sensu) Connect() error {
+	if len(s.ContentEncoding) != 0 {
+		validEncoding := []string{"identity", "gzip"}
+		if !choice.Contains(s.ContentEncoding, validEncoding) {
+			return fmt.Errorf("Unsupported content_encoding [%q] specified", s.ContentEncoding)
+		}
```
ssoroka, line 224: validate settings in a new function like this: 
```go
func (s *Sensu) Init() error {
  // ... return error on configuration issue.
}
```

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,501 @@
...
+			tagList = append(tagList, tag)
+		}
+
+		// Get all valid numeric values, convert to float64
+		for _, fieldSet := range metric.FieldList() {
+			key := fieldSet.Key
+			value := getFloat(fieldSet.Value)
+			// JSON does not support these special values
+			if math.IsInf(value, 1) {
+				s.Log.Debugf("metric %s returned positive infity, setting value to %f", key, math.MaxFloat64)
```
ssoroka, line 284: infity: check this spelling in multiple spots.

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,501 @@
...
+		req.Header.Set("Authorization", "Key "+*s.ApiKey)
+	}
+
+	resp, err := s.client.Do(req)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
```
ssoroka, line 349: may want to check for 4xx level codes and log but not return the error.

plugins/outputs/sensu/sensu.go
```diff
@@ -0,0 +1,501 @@
...
+func init() {
+	outputs.Add("sensu-go", func() telegraf.Output {
+		// Default configuration values
+
+		// make a string from the defaultUrl const
+		agentApiUrl := defaultUrl
+
+		return &Sensu{
+			AgentApiUrl: &agentApiUrl,
+			Timeout:     internal.Duration{Duration: defaultClientTimeout},
```
ssoroka, line 401: can you switch to config.Duration? the usage is nicer and we're planning to deprecate internal.Duration.

ssoroka (changes requested):
Have a few more change requests for you.

plugins/outputs/sensu/sensu.go
```diff
@@ -346,7 +335,12 @@ func (s *Sensu) write(reqBody []byte) error {
...
 	defer resp.Body.Close()
 
-	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
+	if resp.StatusCode != http.StatusCreated {
+		bodyData, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			s.Log.Debugf("Couldn't read response body: %v", err)
+		}
+		s.Log.Debugf("Failed to write, response: %v", string(bodyData))
 		return fmt.Errorf("when writing to [%s] received status code: %d", s.EndpointUrl, resp.StatusCode)
```
ssoroka: only return the error if it's non-4xx

ssoroka (approved):

ssoroka:
thanks for the fast turn-around! :)

Issue #8354, opened 2020-11-02
Starlark processor: Allow to manage errors that occur in the apply function


essobedo:
Thanks to the Starlark processor, it is now possible to execute code written in Starlark language which is very cool. The problem is how to deal with errors that occur within the apply function. Indeed, for example if we want to parse a JSON content, and the JSON content is invalid. Up to now, we get an error in the logger of Telegraf while we would like to somehow catch the exception (which is not supported by the Starlark language) to add a field in my metric with an error message.

ssoroka:
This is a good idea. I think we could add some extra functions to help with this. Let me test out some ideas and get back to you

srebhan:
How about we add an extra `error()` API that you can fill just like the `apply()` function? It could have the exact same API returning some metrics and additionally get the error message?
@essobedo: What do you think about this? For example it could look something like:
```
def on_error(metric, error):
    metric.fields["ok"] = False
    metric.fields["error"] = error.message

    return metric
```
Would that work for you?

essobedo:
@ssoroka 
> Would that work for you?

Yes, clearly, It would be good enough

srebhan:
@ssoroka: What do you think about my proposal? How should the function be called? Do you want to or should I? :-)

essobedo:
@ssoroka Hi, were you able to test something? do you want me to provide a PR based on what @srebhan proposed or something else?

srebhan:
@essobedo: If you are up to putting together a PR that would be cool. I'm willing to review it if you don't mind.

essobedo:
@srebhan @ssoroka here is my PR proposal https://github.com/influxdata/telegraf/pull/8401

ssoroka:
merged! I think we could do a little bit more in this area, I've added a couple more issues under [the starlark label](https://github.com/influxdata/telegraf/labels/area%2Fstarlark).

I'm going to close this out as the original issue is resolved.

srebhan:
@essobedo: I think `catch` should belong to the Starlark language. Maybe you can also try to get it into the upstream repository at https://github.com/google/starlark-go!?

essobedo:
@srebhan AFAIK It is not supported intentionally to keep it simple so I don't think they will accept it anyway https://github.com/google/skylark/issues/125

PR #8401, opened 2020-11-13
Allow to manage errors that occur in the apply function

essobedo:
### Required for all PRs:

Fix for https://github.com/influxdata/telegraf/issues/8354

## Motivation

There is no way to manage error returned in the apply function.

## Modifications:

* Adds a new built-in function called `catch` based on the function of the same name of [`starlarktest`](https://github.com/google/starlark-go/blob/master/starlarktest/starlarktest.go#L79-L88)
* Adds a new test to ensure that it works as expected
* Adds a new sub section of **Common Questions** into the `README.md` to describe how to use it

### DIFF
--- a/plugins/processors/starlark/README.md
+++ b/plugins/processors/starlark/README.md
@@ -154,6 +154,27 @@ def apply(metric):
 Telegraf freezes the global scope, which prevents it from being modified.
 Attempting to modify the global scope will fail with an error.
 
+**How to manage errors that occur in the apply function?**
+
+In case you need to call some code that may return an error, you can delegate the call
+to the built-in function `catch` which takes as argument a `Callable` and returns the error
+that occured if any, `None` otherwise.
+
+So for example:
+
+```python
+load("json.star", "json")
+
+def apply(metric):
+    error = catch(lambda: failing(metric))
+    if error != None:
+        # Some code to execute in case of an error
+        metric.fields["error"] = error
+    return metric
+
+def failing(metric):
+    json.decode("non-json-content")
+```
 
 ### Examples
 
--- a/plugins/processors/starlark/builtins.go
+++ b/plugins/processors/starlark/builtins.go
@@ -34,6 +34,19 @@ func deepcopy(thread *starlark.Thread, _ *starlark.Builtin, args starlark.Tuple,
 	return &Metric{metric: dup}, nil
 }
 
+// catch(f) evaluates f() and returns its evaluation error message
+// if it failed or None if it succeeded.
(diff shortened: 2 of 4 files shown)


ssoroka:
I like where you're going with this. I'm guessing in this case you specifically don't want the error to prevent the metric from being parsed? or are you just trying to make the error visible?

essobedo:
@ssoroka 
> I like where you're going with this. I'm guessing in this case you specifically don't want the error to prevent the metric from being parsed? or are you just trying to make the error visible?

I would answer both, I want to have the full control if an error occurs. But initially my use case was indeed to prevent the metric being processed in case of an error. With the current code, if you have an error, the metric is lost while I would like to keep it and add the error message to it.

ssoroka:
I agree the user should have full control over what happens. I was thinking there should be some logging primitives, too, to tie into the Telegraf logger. something like `log.info`, `log.warn`, `log.error`, and `log.debug`, so you can customize what to log around error conditions.

I'd also like the ability to explicitly return an error.  I believe starlark has a built-in `fail()` command that can do this, but I haven't tested it.

ssoroka (approved):

essobedo:
@ssoroka 
> I'd also like the ability to explicitly return an error. I believe starlark has a built-in `fail()` command that can do this, but I haven't tested it.
Yes the `fail` function is part starlarktest and is for testing purpose. It actually leverages the `catch` function  as you can see [here](https://github.com/google/starlark-go/blob/30ae18b8564f6cd89b550b19dbdae92e0ec5b0a3/starlarktest/assert.star#L32-L38)

ssoroka:
Have any interest in adding the other functions I mentioned in another PR? I've got a list of outstanding starlark issues here, and I'd be more than happy for the assistance. https://github.com/influxdata/telegraf/labels/area%2Fstarlark

ssoroka:
also feel free to join the #Telegraf channel on [InfluxData slack](http://influxdata.com/slack), and consider writing [starlark-based tests](https://github.com/influxdata/telegraf/tree/master/plugins/processors/starlark/testdata) in the future as they get executed and double as documentation

essobedo:
> Have any interest in adding the other functions I mentioned in another PR? I've got a list of outstanding starlark issues here, and I'd be more than happy for the assistance. https://github.com/influxdata/telegraf/labels/area%2Fstarlark

Thanks for proposing, yes I'm interested in helping you if I can.

I will try to propose a PR asap for the logging, I guess it is https://github.com/influxdata/telegraf/issues/8402

ssoroka:
feel free to add to that issue. it's mostly just a placeholder to attach work and descriptions to later.

PR #8407, opened 2020-11-14
CloudWatch output plugin enhancement to exclude dimensions

dougyoung:
Adds feature to remove specified dimensions before sending to CloudWatch. For example CloudWatch has a [maximum limit on the length of dimension names and values](https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_Dimension.html) of 255 characters. In our particular use case, Flink metrics gathered by Prometheus and sent to CloudWatch, Flink can often generate `task_name` dimensions that far exceed this limit. When this occurs the following error is logged by Telegraf:

```
2020-11-14T08:50:12Z E! [agent] Error writing to outputs.cloudwatch: InvalidParameterValue: The parameter MetricData.member.2.Dimensions.member.8.Value must be shorter than 257 characters.
```

Another option to solving this is to automatically remove dimensions with length greater than 255. Upon consideration I thought this would be too unexpected for the user. I'm open to reconsidering this.

This feature is used by declaring the dimensions to be excluded like this:

```
[[outputs.cloudwatch]]
  region = "us-east-1"
  namespace = "namespace"
  [outputs.cloudwatch.dimensions_excluded]
    dimension_to_remove = true
```

### Required for all PRs:

### DIFF
--- a/plugins/outputs/cloudwatch/cloudwatch.go
+++ b/plugins/outputs/cloudwatch/cloudwatch.go
@@ -25,6 +25,7 @@ type CloudWatch struct {
 	Token       string `toml:"token"`
 	EndpointURL string `toml:"endpoint_url"`
 
+	DimensionsExcluded    map[string]bool `toml:"dimensions_excluded"`
 	Namespace             string `toml:"namespace"` // CloudWatch Metrics Namespace
 	HighResolutionMetrics bool   `toml:"high_resolution_metrics"`
 	svc                   *cloudwatch.CloudWatch
@@ -44,7 +45,7 @@ const (
 
 type cloudwatchField interface {
 	addValue(sType statisticType, value float64)
-	buildDatum() []*cloudwatch.MetricDatum
+	buildDatum(dimensionsExcluded map[string]bool) []*cloudwatch.MetricDatum
 }
 
 type statisticField struct {
@@ -62,7 +63,7 @@ func (f *statisticField) addValue(sType statisticType, value float64) {
 	}
 }
 
-func (f *statisticField) buildDatum() []*cloudwatch.MetricDatum {
+func (f *statisticField) buildDatum(dimensionsExcluded map[string]bool) []*cloudwatch.MetricDatum {
 
 	var datums []*cloudwatch.MetricDatum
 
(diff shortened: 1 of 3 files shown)


plugins/outputs/cloudwatch/cloudwatch.go
```diff
@@ -25,6 +25,7 @@ type CloudWatch struct {
 	Token       string `toml:"token"`
 	EndpointURL string `toml:"endpoint_url"`
 
+	DimensionsExcluded    map[string]bool `toml:"dimensions_excluded"`
```
matiasgarciamarset, line 28: very small improvement but using struct{} you don't need to store the bool
```suggestion
	DimensionsExcluded    map[string]struct{} `toml:"dimensions_excluded"`
```

plugins/outputs/cloudwatch/cloudwatch.go
```diff
@@ -376,7 +377,7 @@ func BuildDimensions(mTags map[string]string) []*cloudwatch.Dimension {
 		}
 
 		value := mTags[k]
-		if value == "" {
+		if dimensionExcluded[k] || value == "" {
```
matiasgarciamarset, line 380: using struct{} here would be something like
```suggestion
		if _, ok := dimensionExcluded[k]; !ok || value == "" {
```

matiasgarciamarset (approved):
Thanks for this improvement, this would be very helpful for us!

esteban-mallen (approved):

gtulipani (approved):

debu99:
when this commit can be merged?

sspaink:
@dougyoung sorry you haven't received a response in a long time, are you still interested in working on this pull request?

srebhan:
One more question regarding this PR... Couldn't you achieve the same using the already existing [`tagexclude` modifier](https://github.com/influxdata/telegraf/blob/master/docs/CONFIGURATION.md#modifiers)? You could drop tags (dimensions in your terms) by doing
```toml
    tagexclude = ["dimension_to_exclude", "a_whole_bunch_of_tags_*",...]
```

zan-xhipe:
@srebhan I am working on #16499 and something like this would be very useful.  If I use `tagexclude` then the tag is not available for use in the namespace template, but if I don't then that tag gets included as a dimension, which uses up one of the cloudwatch dimensions, and is redundant as the namespace already included the information from that tag.

Another option would be changing how `tagexclude` works, but that seems like it will be much more of an issue.

Issue #8402, opened 2020-11-13
support logging in starlark


ssoroka:
I was thinking there should be some logging primitives, too, to tie into the Telegraf logger. something like log.info, log.warn, log.error, and log.debug, so you can customize what to log around error conditions.

essobedo:
@ssoroka I could not figure out a better way to do it, so here a potential fix for it https://github.com/influxdata/telegraf/pull/8408 hoping that it was what you had in mind

ssoroka:
resolved in https://github.com/influxdata/telegraf/pull/8408

Issue #8371, opened 2020-11-06
Starlark Processor: Return multiple metrics from plugin


sjwang90:
## Feature Request

Opening a feature request kicks off a discussion.

### Proposal: 
Return multiple metrics from Starlark processor plugin

### Current behavior:
Currently can only return (create) one metric from the Starlark

### Desired behavior:
- Return multiple metrics.
- Approaches:
  1. Add a new function `add()` to add new metrics in the starlark processor. _Use case: parse an array of json items and add one metric per item in the array_
  2. Return an array or a tuple in the apply function. Currently we return the metric if we want to keep it.  If we want to make multiple metrics from one, more natural to return multiple metrics rather than call an `add()` function

### Use case: 
- Parse an array of json items and add one metric per item in the array

srebhan:
Hey @sjwang90, back when I discussed the starlark processor, @danielnelson was against handling multiple metrics as once (basically enabling aggregation etc.). I don't remember the reasoning.
Do you want multiple in, multiple out or single in multiple out?

ssoroka:
always single in, 0 or more out.

essobedo:
The corresponding PR https://github.com/influxdata/telegraf/pull/8423 that shows how to do it

ssoroka:
fixed in https://github.com/influxdata/telegraf/pull/8423

PR #8408, opened 2020-11-14
Support logging in starlark

essobedo:
### Required for all PRs:

Fix for #8402

## Motivation

Logging primitives to tie into the Telegraf logger are missing

## Modifications:

* Adds a new module to manage all the logging primitives
* Uses a static logger to log the message as there is no easy way to use the current one

### DIFF
--- a/plugins/processors/starlark/logging.go
+++ b/plugins/processors/starlark/logging.go
@@ -1,0 +1,47 @@
+package starlark
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/influxdata/telegraf"
+	"go.starlark.net/starlark"
+	"go.starlark.net/starlarkstruct"
+)
+
+// Builds a module that defines all the supported logging functions which will log using the provided logger
+func LogModule(logger telegraf.Logger) *starlarkstruct.Module {
+	var logFunc = func(t *starlark.Thread, b *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+		return log(t, b, args, kwargs, logger)
+	}
+	return &starlarkstruct.Module{
+		Name: "log",
+		Members: starlark.StringDict{
+			"debug": starlark.NewBuiltin("log.debug", logFunc),
+			"info":  starlark.NewBuiltin("log.info", logFunc),
+			"warn":  starlark.NewBuiltin("log.warn", logFunc),
+			"error": starlark.NewBuiltin("log.error", logFunc),
+		},
+	}
+}
+
+// Logs the provided message according to the level chosen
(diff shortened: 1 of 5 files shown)


srebhan:
While your changes look good, I'm a bit hesitant to add this to to telegraf as IMO this should really go to starlark itself. Otherwise we are basically forking starlark... I've e.g. some time/duration module in the pipe and could also add this, but this means that we bloat telegraf by stuff that actually belongs to starlark.
@ssoroka what is your opinion on this?

ssoroka:
@srebhan This is what I was expecting. I don't see anything specific to logging in Starlark, and even so it'd need some tie in to our logger. This does sort of bring us to a situation where we have a sort of sub-dialect of Starlark, but I don't necessarily see that as a problem.

plugins/processors/starlark/logging.go
```diff
@@ -0,0 +1,45 @@
...
+	Members: starlark.StringDict{
+		"debug": starlark.NewBuiltin("log.debug", log),
+		"info":  starlark.NewBuiltin("log.info", log),
+		"warn":  starlark.NewBuiltin("log.warn", log),
+		"error": starlark.NewBuiltin("log.error", log),
+	},
+}
+
+// The logger used by the module
+var logger = models.NewLogger("processors", "starlak", "logging")
```
ssoroka, line 24: this is the only line that doesn't sit well with me. 
1. Ideally I'd like the alias to default to the script file name, if it's given. If not maybe we should consider suggesting setting the alias in the plugin config.
2. Loggers are typically initialized with the plugin name and allow overriding the alias. I think with some work we could inject the logger into this function.

essobedo: @ssoroka it should be better now, please check again

ssoroka (approved):
looks great.

srebhan:
@ssoroka: Well you could easily define an interface for the logger and allow to specify this when initializing the logging module... IF we want to go the route of adding all that stuff from _InfluxData_ side, we should actually fork starklark and add all that modules in the fork. This way we will not clutter this plugin. In the end, the real problem is where to stop. I've a date/time/duration module, someone else will come up with a math module and the next guy submits a timeseries prediction... Where do you cut this? Again, **using** those modules in this plugin is not my concern, just plugging more and more stuff in here...

@essobedo this is no objection to your code (quite the contrary), IMO we need to get this straight now as otherwise it is hard to stop somewhere.

ssoroka:
Going to merge and then pull out the module to https://github.com/influxdata/starlark-go-modules afterwards. 

Thanks for the PR @essobedo ❤️

ssoroka:
@essobedo find me on the [community slack](https://influxdata.com/slack) for detailed conversation... but maybe you could post your ideas on https://github.com/influxdata/telegraf/issues/8371 and start thinking about that. Possible options are multiple return values, or maybe if there are problems with that an exposed method to add new metrics.

essobedo:
@ssoroka ok, I can look at #8371, you can assign it to, I will do it asap

Issue #3716, opened 2018-01-23
Add support for converting integers encoded as Hex-STRING


Soldrym:
### Relevant telegraf.conf:
[[inputs.snmp]]
agents = ["10.41.16.16"]
community = "public"
interval = "60s"
[[inputs.snmp.table]]
name = "testing"
[[inputs.snmp.table.field]]
name = "connUnitPortStatIndex"
oid = "FCMGMT-MIB::connUnitPortStatIndex"
is_tag = true
[[inputs.snmp.table.field]]
oid = "FCMGMT-MIB::connUnitPortStatCountTxElements"
conversion = "int"

### System info:
Server:
Telegraf v1.5.1 (git: release-1.5 0605af7c)
Red Hat Enterprise Linux Server release 7.4 (Maipo)
Switches tested: 
Brocade 7800, DCX8510 and DCX-4S
Fabric OS v7.4.1e and v8.0.2a

### Steps to reproduce:
1. Run snmpwalk on the above OID
2. Compare results using Telegraf --test

### Expected behavior:
snmpwalk shows Hex-STRING results. Telegraf would convert the hex-string to a decimal integer. Example:

Snmpwalk:
> snmpwalk -v2c -cpublic 10.41.16.16 FCMGMT-MIB::connUnitPortStatCountTxElements
FCMGMT-MIB::connUnitPortStatCountTxElements.'....31.$........'.1 = Hex-STRING: 00 09 3E E3 F6 D5 3B 60
FCMGMT-MIB::connUnitPortStatCountTxElements.'....31.$........'.2 = Hex-STRING: 00 16 F7 B2 C1 07 C5 30

Telegraf (truncated to 2 ports):
> telegraf --test --config telegraf.conf
* Plugin: inputs.snmp, Collection 1
* Internal: 1m0s
> testing,connUnitPortStatIndex=1,agent_host=10.41.16.16,host=testhost connUnitPortStatCountTxElements=2602423610063712 1516743861000000000
> testing,connUnitPortStatIndex=2,agent_host=10.41.16.16,host=testhost connUnitPortStatCountTxElements=6464796602385712 1516743861000000000

### Actual behavior:
Telegraf converts the data to a '0i' 

Snmpwalk:
>snmpwalk -v2c -cpublic 10.41.16.16 FCMGMT-MIB::connUnitPortStatCountTxElements
[...]

PR #8426, opened 2020-11-17
Add support to convert snmp hex strings to integers

wiardvanrij:
### Required for all PRs:

Relates to #3716

### DIFF
--- a/plugins/inputs/snmp/README.md
+++ b/plugins/inputs/snmp/README.md
@@ -113,15 +113,16 @@ option operate similar to the `snmpget` utility.
     # is_tag = false
 
     ## Apply one of the following conversions to the variable value:
-    ##   float(X) Convert the input value into a float and divides by the
-    ##            Xth power of 10. Effectively just moves the decimal left
-    ##            X places. For example a value of `123` with `float(2)`
-    ##            will result in `1.23`.
-    ##   float:   Convert the value into a float with no adjustment. Same
-    ##            as `float(0)`.
-    ##   int:     Convert the value into an integer.
-    ##   hwaddr:  Convert the value to a MAC address.
-    ##   ipaddr:  Convert the value to an IP address.
+    ##   float(X):  Convert the input value into a float and divides by the
+    ##              Xth power of 10. Effectively just moves the decimal left
+    ##              X places. For example a value of `123` with `float(2)`
+    ##              will result in `1.23`.
+    ##   float:     Convert the value into a float with no adjustment. Same
+    ##              as `float(0)`.
+    ##   int:       Convert the value into an integer.
+    ##   hwaddr:    Convert the value to a MAC address.
+    ##   ipaddr:    Convert the value to an IP address.
+    ##   hextoint:  Convert a hex string value to integer.
     # conversion = ""
 ```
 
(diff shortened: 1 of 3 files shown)


srebhan (commented):
Looks good so far. One minor thing:
In the comment above the declaration of function `fieldConvert()` the possible formats are described. It would be nice to either add the new option there or remove the documentation altogether...

wiardvanrij:
Thanks, I've removed the comments as I think the code speaks for itself and the readme should be sufficient.

wiardvanrij:
@srebhan - Just letting you know I've build this myself for a team at our company. They have tested this in production today and it's working as intended. I hope we can have this released soon. Thanks!

srebhan (approved):

reimda (approved):

phemmer:
Copying my comment from over on https://github.com/influxdata/telegraf/issues/3716#issuecomment-750885510 for visibility:

"hextoint" is not the correct name for the conversion. The data isn't in hex, it's in binary. "Hex" is an encoding format representing for numbers as strings using ASCII characters 0-9 A-F. That's not what this data being converted there is. The data is in binary, hence why there's no code to actually convert from a string in that PR.

If we were to ever actually add support for converting hex, we'd have to give it a different name, and it'd be extremely confusing to have hextoint which sounds like it does hex conversion, but doesn't, and something else that does.

srebhan:
Well reading the code this is exactly what is done here. The *data-to-convert* is a string containing a HEX representation... What is your critique concretely? In my interpretation `hextoint` is a conversion of a string e.g. `0xFF2A` to the integer datatype in this case `65322`. What is your interpretation?

phemmer:
> Well reading the code this is exactly what is done here. The data-to-convert is a string containing a HEX representation

I'm sorry, but you must be misreading the code. `binary.LittleEndian.Uint64` and all the neighboring functions convert raw binary, not hex strings.

srebhan:
Very true I completely missed that. Sorry! So basically we should rename this to `bytetoint` right?
@ssoroka and @reimda should we add a PR to rename the option before a release!?!???!?

phemmer:
> Very true I completely missed that. Sorry! So basically we should rename this to `bytetoint` right?

`bytetoint` seems reasonable. The name still nags at me a little bit, but I can't think of anything better. I think it would also have made sense as a parameter to the `int` converter, as that converter can already accept multiple input types and outputs to int. But a separate converter works fine too.

wiardvanrij:
> Very true I completely missed that. Sorry! So basically we should rename this to `bytetoint` right?
> @ssoroka and @reimda should we add a PR to rename the option before a release!?!???!?

This is already released: https://docs.influxdata.com/telegraf/v1.17/about_the_project/release-notes-changelog/

> `bytetoint` seems reasonable. The name still nags at me a little bit, but I can't think of anything better. I think it would also have made sense as a parameter to the `int` converter, as that converter can already accept multiple input types and outputs to int. But a separate converter works fine too.

That's the great thing about OSS, pull-requests can be made. 

Currently, it would be breaking to change it again. My suggestion is to just close the 3-year-old issue that had no priority and call it a day. With perhaps some lessons learned.

srebhan:
@wiardvanrij yeah the really bad thing is that we now "never-again" can reuse the "hextoint" name as it would change the semantics of an existing option. :-( Sorry for having missed it in the first place.

phemmer:
My bigger concern is that when someone comes looking for functionality to convert from binary (byte encoding) to native integer, they're not going to see this "hextoint" and realize it does what they want.

PR #8438, opened 2020-11-19
Starlark example dropbytype

srebhan:
### Required for all PRs:

Starlark example for dropping fields of metrics by the value type. This is a suitable replacement for PR #3763.

### DIFF
--- a/plugins/processors/starlark/testdata/drop_fields_with_unexpected_type.star
+++ b/plugins/processors/starlark/testdata/drop_fields_with_unexpected_type.star
@@ -10,15 +10,15 @@
 # measurement,host=hostname a=1i,b=4.2,c=42.0,d="v3.14",e=true,f=23.0 1597255410000000000
 # measurement,host=hostname a=1i,c=42.0,d="v3.14",e=true,f=23.0 1597255410000000000
 
+expected_type = {
+    "a": "int",
+    "b": "float",
+    "c": "float",
+    "d": "string",
+    "e": "bool"
+}
+    
 def apply(metric):
-    expected_type = {
-        "a": "int",
-        "b": "float",
-        "c": "float",
-        "d": "string",
-        "e": "bool"
-    }
-
     for k, v in metric.fields.items():
         if type(v) != expected_type.get(k, type(v)):
             metric.fields.pop(k)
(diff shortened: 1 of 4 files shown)


plugins/processors/starlark/testdata/dropfields_by_type.star
```diff
@@ -0,0 +1,19 @@
...
+# dropped. If no such tag exists we assume "string".
+#
+# Example Input:
+# measurement,host=hostname a=1,b="somestring" 1597255410000000000
+#
+# Example Output:
+# measurement,host=hostname a=1 1597255410000000000
+
+def apply(metric):
+    drop = metric.tags.get("type", "string")
```
ssoroka, line 13: This seems like an unusual example. I don't think I'd expect metrics to contain instructions about what field type to drop. It might be a clearer example if it was just assumed that it's dropping string fields.

ssoroka:
Left a comment. A further example of a real life use case: sometimes data comes in with dirty types. The field is _usually_ an int, but once in a while some system sticks a string in the type. It would make sense in that case to either drop the field with the unexpected type.  You could expand this example to enforcing a schema for a metric.

srebhan:
Sure will try to handle your comments during lunch break... ;-)

srebhan:
Ok removed the "tag option for removal" to make it more real-life and also added a "drop field by unexpected type example".

plugins/processors/starlark/testdata/drop_fields_with_unexpected_type.star
```diff
@@ -0,0 +1,26 @@
...
+
+def apply(metric):
+    expected_type = {
+        "a": "int",
+        "b": "float",
+        "c": "float",
+        "d": "string",
+        "e": "bool"
+    }
+
```
ssoroka, line 21: It probably makes little difference to performance, but this is possible:

```suggestion
expected_type = {
    "a": "int",
    "b": "float",
    "c": "float",
    "d": "string",
    "e": "bool"
}
    
def apply(metric):
```

ssoroka (approved):

ssoroka:
LGTM.

plugins/processors/starlark/testdata/drop_fields_with_unexpected_type.star
```diff
@@ -0,0 +1,26 @@
...
+    "b": "float",
+    "c": "float",
+    "d": "string",
+    "e": "bool"
+}
+    
+def apply(metric):
+    for k, v in metric.fields.items():
+        if type(v) != expected_type.get(k, type(v)):
+            metric.fields.pop(k)
```
ssoroka, line 24: we have access to a logger now, let's add a log entry to this example that says that we're dropping it because it's the wrong type, and what the actual and expected types were.

Issue #8403, opened 2020-11-13
support returning a custom error in starlark


ssoroka:
the custom error should get logged. It'd also be great if the starlark processor had a configurable option to say whether the original metric should be dropped if the processor errors.

essobedo:
@ssoroka I can check this one if you want. If so, could you clarify a little bit about what you expect exactly? I'm mean we can already create a custom error with the built-in function `fail(message)` unless I misunderstand what you want to do?

essobedo:
The corresponding PR https://github.com/influxdata/telegraf/pull/8439

essobedo:
@ssoroka I guess you can close this one since the PR has been merged

sjwang90:
Closed #8439

PR #8439, opened 2020-11-19
Show how to return a custom error with the Starlark processor

essobedo:
### Required for all PRs:

Fix for #8403 

## Motivation

It seems to be unclear for end-users to return a custom error with the Starlark processor

## Modifications

* Adds a new Starlark script to show how to return a custom error thanks to the function [fail](https://docs.bazel.build/versions/master/skylark/lib/globals.html#fail)
* Adds the corresponding unit test that proves that it works as expected
* Refers the new Starlark script from the `README.md`

### DIFF
--- a/plugins/processors/starlark/testdata/fail.star
+++ b/plugins/processors/starlark/testdata/fail.star
@@ -1,4 +1,5 @@
 # Example of the way to return a custom error thanks to the built-in function fail
+# Returning an error will drop the current metric. Consider using logging instead if you want to keep the metric.
 #
 # Example Input:
 # fail value=1 1465839830100400201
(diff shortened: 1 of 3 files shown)


plugins/processors/starlark/testdata/fail.star
```diff
@@ -0,0 +1,12 @@
+# Example of the way to return a custom error thanks to the built-in function fail
+#
```
ssoroka, line 2: let's add a statement along the lines of "returning an error will drop the current metric."

ssoroka (approved):

srebhan (approved):
Looks good.

PR #8440, opened 2020-11-19
Keep field name for csv timestamp column (don't implicitly convert)

### LINKED ISSUE #7288: timestamp format parsing is confusing yyyyMMddHHmm for a unix timestamp
### Relevant telegraf.conf:
```toml

[[inputs.http]]
   urls = [
     "https://data.geo.admin.ch/ch.meteoschweiz.messwerte-aktuell/VQHA80.csv"
   ]
   interval = "10m"
   method = "GET"
   timeout = "15s"
   data_format = "csv"
   csv_delimiter = ";"
   csv_skip_rows = 2
   csv_header_row_count = 1 
   csv_tag_columns = ["stn"]
   csv_timestamp_column = "time"
   csv_timestamp_format = "200601021504"
```

### System info:

Telegraf 1.11.2 (git: HEAD a80d173f)

### Steps to reproduce:

1. Add the input config above to your Telegraf deamon 
2. Execute the command `telegraf --test --config=/etc/telegraf/<your_conf>.conf`

### Expected behavior:

The `time` timestamp is interpreted as yyyyMMddHHmm.
e.g. `200601021504` for `Mon Jan 2 15:04:05 -0700 MST 2006` as in the Go documentation (https://golang.org/pkg/time/#Parse)

### Actual behavior:

The `time` is interpreted as a UNIX format and is therefore, invalid as seen in the related log line below.

`2020-04-05T06:55:51Z E! [inputs.http]: Error in plugin: [url=https://data.geo.admin.ch/ch.meteoschweiz.messwerte-aktuell/VQHA80.csv]: Invalid unix format`

stephanie-engel:
### Required for all PRs:

**Root Cause Analysis:**
Within the csv parser, the field name for the `csv_timestamp_column` is, initially, converted from a `string` into an `int`. This causes the field name (for example: `202011131605`) to be interpreted as a Unix format (instead of the Go reference time `200601021504`).

**The Fix:**
- Prevent implicit type conversion on the timestamp column, so that `parseTimestamp` can correctly parse the time as unix or UTC
- Added a unit test to verify `TimestampFormat:   "200601021504"`

closes #7288

### DIFF
--- a/plugins/parsers/csv/parser.go
+++ b/plugins/parsers/csv/parser.go
@@ -204,6 +204,12 @@ outer:
 				}
 			}
 
+			// If the field name is the timestamp column, then keep field name as is.
+			if fieldName == p.TimestampColumn {
+				recordFields[fieldName] = value
+				continue
+			}
+
 			// Try explicit conversion only when column types is defined.
 			if len(p.ColumnTypes) > 0 {
 				// Throw error if current column count exceeds defined types.
--- a/plugins/parsers/csv/parser_test.go
+++ b/plugins/parsers/csv/parser_test.go
@@ -85,6 +85,26 @@ func TestTimestamp(t *testing.T) {
 	require.Equal(t, metrics[1].Time().UnixNano(), int64(1257609906000000000))
 }
 
+func TestTimestampYYYYMMDDHHmm(t *testing.T) {
+	p, err := NewParser(
+		&Config{
+			HeaderRowCount:    1,
+			ColumnNames:       []string{"first", "second", "third"},
+			MeasurementColumn: "third",
+			TimestampColumn:   "first",
+			TimestampFormat:   "200601021504",
+			TimeFunc:          DefaultTime,
+		},
+	)
+	testCSV := `line1,line2,line3
+200905231605,70,test_name
+200907111605,80,test_name2`
+	metrics, err := p.Parse([]byte(testCSV))
+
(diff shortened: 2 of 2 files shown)


srebhan (changes requested):
Hey @stephanie-engel, nice work! However, I think we also have to handle the cases where the column-type is given explicitly. This is the section above you change. How about handling both cases by adding
```
		if fieldName == p.TimestampColumn {
￼				recordFields[fieldName] = value
                                continue
                }
```
directly before line 208
```
			if len(p.ColumnTypes) > 0 {
```

This way you can leave the lines 241--250 unchanged.

stephanie-engel:
> Hey @stephanie-engel, nice work! However, I think we also have to handle the cases where the column-type is given explicitly. This is the section above your change. How about handling both cases by adding
> 
> ```
> 		if fieldName == p.TimestampColumn {
> ￼				recordFields[fieldName] = value
>                                 continue
>                 }
> ```
> 
> directly before line 208
> 
> ```
> 			if len(p.ColumnTypes) > 0 {
> ```
> 
> This way you can leave the lines 241--250 unchanged.

Thanks for the great feedback, @srebhan ! I just went ahead and made the changes you requested. I confirmed that the timestamp is still parsed correctly and the unit tests still pass 😄

ssoroka (approved):
Looks great, congratulations. 🥇

Issue #5939, opened 2019-06-01
Add Grafana Loki logging output for telegraf.


devopstales:
## Feature Request
`Submission of telegraf output plugin for Grafana Loki`

### Proposal:
`Create telegraf output plugin to push data to  Grafana Loki`

### Current behavior:
 `No plugin exists`

### Desired behavior:
` Ability to configure output plugin for Grafana Loki`
### Use case: 
`I can use telegraf to get megrics for Prometheus. Loki is a Prometheus-inspired logging service so it would be nice, if I could send the logs of the application to Loki with telegraf.`

rdxmb:
telegraf sends metrics, not logs. Have you tried out promtail for loki?

danielnelson:
We do logs too, metrics/logs are basically treated the same for Telegraf.  For example we are adding a syslog output in the next release that can send metrics as logs, as well as log style data from the syslog, tail, and docker_log (coming soon) plugins.  So as long as we can come up with a way to encode metrics as logs, maybe logfmt or the serializers, it could work.

cricket007:
@rdxmb 

> telegraf sends metrics, not logs

Refer - https://github.com/influxdata/telegraf/tree/master/plugins/outputs/syslog

rdxmb:
Oh, I've just learned something new. Thanks!

kOstovic:
Any status on this one?

PR #8450, opened 2020-11-22
plugin: output loki

Eraac:
fixes: #5939

### Required for all PRs:

README and unit tests will be added soon (next week)

I open this PR to get review, is my first plugin I hope doing well

Documentation about Loki endpoint: https://github.com/grafana/loki/blob/master/docs/sources/api/_index.md#user-content-post-lokiapiv1push

### DIFF
--- a/plugins/outputs/loki/loki.go
+++ b/plugins/outputs/loki/loki.go
@@ -27,14 +27,21 @@ var sampleConfig = `
   ## Connection timeout, defaults to "5s" if not set.
   timeout = "5s"
 
-  ## The URLs of the Loki instance
+  ## The URL of Loki
   # url = "https://loki.domain.tld"
+
+  ## Basic auth credential
+  # username = "loki"
+  # password = "pass"
 
   ## Additional HTTP headers
   # http_headers = {"X-Scope-OrgID" = "1"}
 
   ## The field containing the log
   # field_line = "log"
+
+  ## If the request must be gzip encoded
+  # gzip_request = false
 
   ## Optional TLS Config
   # tls_ca = "/etc/telegraf/ca.pem"
@@ -65,7 +72,7 @@ func (l *Loki) SampleConfig() string {
 }
 
 func (l *Loki) Description() string {
-	return "Send logs to Loki instance"
+	return "Send logs to Loki"
 }
 
 func (l *Loki) createClient(ctx context.Context) (*http.Client, error) {
--- a/plugins/outputs/loki/loki.go
+++ b/plugins/outputs/loki/loki.go
@@ -2,11 +2,13 @@ package loki
 
 import (
 	"bytes"
-	"compress/gzip"
 	"context"
 	"encoding/json"
 	"fmt"
+	"io"
+	"io/ioutil"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/influxdata/telegraf"
@@ -146,60 +148,64 @@ func (l *Loki) Write(metrics []telegraf.Metric) error {
 			continue
 		}
 
-		s.insertLog(tags, Log{
-			Timestamp: m.Time().UnixNano(),
-			Line:      line.(string),
-		})
+		s.insertLog(tags, Log{fmt.Sprintf("%d", m.Time().UnixNano()), line.(string)})
 	}
 
 	return l.write(s)
 }
 
-func (l *Loki) write(s Streams) (err error) {
-	var (
-		buf     bytes.Buffer
-		encoder *json.Encoder
-	)
+func (l *Loki) write(s Streams) error {
+	bs, err := json.Marshal(s)
+	if err != nil {
+		return fmt.Errorf("json.Marshal: %w", err)
+	}
+
+	var reqBodyBuffer io.Reader = bytes.NewBuffer(bs)
 
 	if l.GZipRequest {
-		gz := gzip.NewWriter(&buf)
-		defer gz.Close()
-
-		encoder = json.NewEncoder(gz)
-	} else {
-		encoder = json.NewEncoder(&buf)
+		rc, err := internal.CompressWithGzip(reqBodyBuffer)
+		if err != nil {
+			return err
+		}
+		defer rc.Close()
+		reqBodyBuffer = rc
 	}
 
-	if err := encoder.Encode(s); err != nil {
-		return fmt.Errorf("new json encoder fail: %w", err)
-	}
-
-	req, err := http.NewRequest(http.MethodPost, fmt.Sprintf("%s%s", l.URL, l.Endpoint), &buf)
+	url := fmt.Sprintf("%s%s", l.URL, l.Endpoint)
+	req, err := http.NewRequest(http.MethodPost, url, reqBodyBuffer)
 	if err != nil {
-		return fmt.Errorf("new request fail: %w", err)
+		return err
+	}
+
+	if l.Username != "" || l.Password != "" {
+		req.SetBasicAuth(l.Username, l.Password)
 	}
 
 	for k, v := range l.Headers {
-		req.Header.Add(k, v)
+		if strings.ToLower(k) == "host" {
+			req.Host = v
+		}
+		req.Header.Set(k, v)
 	}
 
-	req.SetBasicAuth(l.Username, l.Password)
-	req.Header.Add("Content-Type", "application/json")
+	req.Header.Set("User-Agent", internal.ProductToken())
+	req.Header.Set("Content-Type", "application/json")
 	if l.GZipRequest {
-		req.Header.Add("Content-Encoding", "gzip")
+		req.Header.Set("Content-Encoding", "gzip")
 	}
 
 	resp, err := l.client.Do(req)
 	if err != nil {
-		return fmt.Errorf("client do fail: %w", err)
+		return err
 	}
 	defer resp.Body.Close()
+	_, err = ioutil.ReadAll(resp.Body)
 
-	if resp.StatusCode < 200 || resp.StatusCode > 299 {
-		return fmt.Errorf("received bad status code %d", resp.StatusCode)
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
+		return fmt.Errorf("when writing to [%s] received status code: %d", url, resp.StatusCode)
 	}
 
-	return
+	return nil
 }
 
 func init() {
--- a/plugins/outputs/loki/loki.go
+++ b/plugins/outputs/loki/loki.go
@@ -6,7 +6,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"net/http"
 	"strings"
 	"time"
@@ -194,11 +193,7 @@ func (l *Loki) write(s Streams) error {
 	if err != nil {
 		return err
 	}
-	defer resp.Body.Close()
-	_, err = ioutil.ReadAll(resp.Body)
-	if err != nil {
-		return err
-	}
+	_ = resp.Body.Close()
 
 	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
 		return fmt.Errorf("when writing to [%s] received status code: %d", l.url, resp.StatusCode)
--- a/plugins/outputs/loki/loki.go
+++ b/plugins/outputs/loki/loki.go
@@ -1,0 +1,204 @@
+package loki
+
+import (
+	"bytes"
+	"compress/gzip"
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"time"
+
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/internal"
+	"github.com/influxdata/telegraf/plugins/common/tls"
+	"github.com/influxdata/telegraf/plugins/outputs"
+	"golang.org/x/oauth2"
+	"golang.org/x/oauth2/clientcredentials"
+)
+
+const (
+	defaultEndpoint      = "/loki/api/v1/push"
+	defaultClientTimeout = 5 * time.Second
+	defaultFieldLine     = "line"
+)
+
+var sampleConfig = `
+  ## Connection timeout, defaults to "5s" if not set.
+  timeout = "5s"
+
+  ## The URLs of the Loki instance
+  # url = "https://loki.domain.tld"
+
+  ## Additional HTTP headers
+  # http_headers = {"X-Scope-OrgID" = "1"}
+
+  ## The field containing the log
+  # field_line = "log"
+
+  ## Optional TLS Config
+  # tls_ca = "/etc/telegraf/ca.pem"
+  # tls_cert = "/etc/telegraf/cert.pem"
+  # tls_key = "/etc/telegraf/key.pem"
+`
+
+type Loki struct {
+	URL          string            `toml:"url"`
+	Endpoint     string            `toml:"endpoint"`
+	Timeout      internal.Duration `toml:"timeout"`
+	Username     string            `toml:"username"`
+	Password     string            `toml:"password"`
+	Headers      map[string]string `toml:"headers"`
+	ClientID     string            `toml:"client_id"`
+	ClientSecret string            `toml:"client_secret"`
+	TokenURL     string            `toml:"token_url"`
+	Scopes       []string          `toml:"scopes"`
+	FieldLine    string            `toml:"field_line"`
+	GZipRequest  bool              `toml:"gzip_request"`
+
+	client *http.Client
+	tls.ClientConfig
+}
+
+func (l *Loki) SampleConfig() string {
+	return sampleConfig
+}
+
+func (l *Loki) Description() string {
+	return "Send logs to Loki instance"
+}
+
+func (l *Loki) createClient(ctx context.Context) (*http.Client, error) {
+	tlsCfg, err := l.ClientConfig.TLSConfig()
+	if err != nil {
+		return nil, fmt.Errorf("tls config fail: %w", err)
+	}
+
+	client := &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: tlsCfg,
+			Proxy:           http.ProxyFromEnvironment,
+		},
+		Timeout: l.Timeout.Duration,
+	}
+
+	if l.ClientID != "" && l.ClientSecret != "" && l.TokenURL != "" {
+		oauthConfig := clientcredentials.Config{
+			ClientID:     l.ClientID,
+			ClientSecret: l.ClientSecret,
+			TokenURL:     l.TokenURL,
+			Scopes:       l.Scopes,
+		}
+		ctx = context.WithValue(ctx, oauth2.HTTPClient, client)
+		client = oauthConfig.Client(ctx)
+	}
+
+	return client, nil
+}
+
+func (l *Loki) Connect() (err error) {
+	if l.URL == "" {
+		return fmt.Errorf("url is required")
+	}
+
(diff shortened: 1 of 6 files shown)


sjwang90:
Going to set this as draft. Feel free to select Ready for Review when you're done with all your commits and want a review.

Eraac:
Hi @sjwang90 any update for a review? This plugin can save us a lot of work

loganmc10:
```field_line = "log"```

What happens if the input has multiple fields? Would it not be better to just send all the comma separated fields as the log line and let users seperate/parse them inside Loki as desired?

**Comment by loganmc10** [comment]
I'm hoping to use this to send SNMP traps to Loki (https://github.com/influxdata/telegraf/tree/master/plugins/inputs/snmp_trap) but that input plugin (and I'm sure others) generates multiple fields based on the trap. I'd rather send all the fields (comma separated) to Loki and use LogQL to separate it there.

File: plugins/outputs/loki/loki.go [hunk 1]
```diff
@@ -0,0 +1,217 @@
...
+	defaultClientTimeout = 5 * time.Second
+	defaultFieldLine     = "line"
+)
+
+var sampleConfig = `
+  ## Connection timeout, defaults to "5s" if not set.
+  timeout = "5s"
+
+  ## The URL of Loki
+  # url = "https://loki.domain.tld"
```
srebhan, line 33: Please uncomment the `url` as it is required to be set by the user!

plugins/outputs/loki/loki.go
```diff
@@ -0,0 +1,217 @@
...
+	l.client, err = l.createClient(ctx)
+	if err != nil {
+		return fmt.Errorf("http client fail: %w", err)
+	}
+
+	return
+}
+
+func (l *Loki) Close() error {
+	return nil
```
srebhan, line 135: You should close the client shouldn't you?

plugins/outputs/loki/loki.go
```diff
@@ -0,0 +1,217 @@
...
+	s := Streams{
+		Streams: make([]Stream, 0, len(metrics)),
+	}
+
+	for _, m := range metrics {
+		tags := m.TagList()
+		line, ok := m.GetField(l.FieldLine)
+
+		if !ok {
+			continue
```
srebhan, line 148: Probably you should output a log-entry here as I guess hitting metrics where the field is missing is not the default behavior. Maybe make it a _debug_ message if this case is likely to happen often.
The reason for my comment is that we otherwise silently drop metrics which is bad for a metric _collection_ agent. ;-)

plugins/outputs/loki/loki.go
```diff
@@ -0,0 +1,217 @@
...
+		}
+
+		s.insertLog(tags, Log{fmt.Sprintf("%d", m.Time().UnixNano()), line.(string)})
```
srebhan, line 151: This will panic if the line does not contain a string (e.g. an integer or similar). Please catch this corner-case!

plugins/outputs/loki/loki.go
```diff
@@ -0,0 +1,217 @@
...
+	if l.GZipRequest {
+		rc, err := internal.CompressWithGzip(reqBodyBuffer)
+		if err != nil {
+			return err
+		}
+		defer rc.Close()
+		reqBodyBuffer = rc
+	}
+
+	url := fmt.Sprintf("%s%s", l.URL, l.Endpoint)
```
srebhan, line 174: Please do this once in the `Init()` function.

plugins/outputs/loki/loki.go
```diff
@@ -0,0 +1,217 @@
...
+	req, err := http.NewRequest(http.MethodPost, url, reqBodyBuffer)
+	if err != nil {
+		return err
+	}
+
+	if l.Username != "" || l.Password != "" {
```
srebhan, line 180: Not sure if it makes sense to do this if `Username` is empty but `Password` is not!?

plugins/outputs/loki/loki.go
```diff
@@ -0,0 +1,217 @@
...
+		return fmt.Errorf("when writing to [%s] received status code: %d", url, resp.StatusCode)
+	}
+
+	return nil
+}
+
+func init() {
+	outputs.Add("loki", func() telegraf.Output {
+		return &Loki{
+			Timeout: internal.Duration{Duration: defaultClientTimeout},
```
srebhan, line 214: This is not necessary as you are setting the default in `Connect()` already.

plugins/outputs/loki/loki_test.go
```diff
@@ -0,0 +1,372 @@
...
+	"time"
+
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/internal"
+	"github.com/influxdata/telegraf/metric"
+	"github.com/stretchr/testify/require"
+)
+
+func getMetric() telegraf.Metric {
+	m, err := metric.New(
```
srebhan, line 21: Check `testutil.MustMetric()`.

plugins/outputs/loki/stream.go
```diff
@@ -0,0 +1,66 @@
...
+	s.Streams[index].Logs = append(s.Streams[index].Logs, l)
+}
+
+func uniqKeyFromTagList(ts []*telegraf.Tag) (k string) {
+	for _, t := range ts {
+		k += fmt.Sprintf("%s%s", t.Key, t.Value)
+	}
+
+	return
+}
```
srebhan, line 52: Are you really sure this is unique and correct? 

1. This will fail if tags are reordered but are otherwise the same. In this case you will generate a new "key" even though the tags are the same.
2. This will probably fail in case of overlapping key/value pairs in the tags. Assume a first metric that has `target=local, host=host, service=dns` and a second metric with `target=localhost, hostservice=dns`. With your "key-construction" you will get `targetlocalhosthostservicedns` in both cases and the metrics will be merged...

Is there a reason you don't use the metric ID that already exists (`metric.HashID()`)?

plugins/outputs/loki/stream.go
```diff
@@ -0,0 +1,66 @@
...
+		index int
+		found bool
+	)
+
+	for i, s := range s.Streams {
+		if s.key == key {
+			index, found = i, true
+			break
+		}
+	}
```
srebhan, line 36: This looks a bit like the poor-man's (slow) version of `map[string][]Stream`. The only reason to do it is the marshaling but I'm not sure it is worth the burden instead of writing a custom marshaller as described in the [json docu](https://golang.org/pkg/encoding/json/#Marshaler)

It might look like this:
```
type Stream struct {
	Labels map[string]string `json:"stream"`
	Logs   []Log             `json:"values"`
}

type Streams map[string][]Stream

func (s Streams) MarshalJSON() ([]byte, error) {
    var out struct {
        S []Stream `json:"streams"`
    }
    for _, streams := range s {
	for _, v := range streams {
	        out.S = append(out.S, v)
	}
    }
    return json.Marshal(out)
}
```

plugins/outputs/loki/loki.go
Eraac: indeed

plugins/outputs/loki/loki.go
Eraac: The only close function the http.Client has is `CloseIdleConnections()`, I've added it

plugins/outputs/loki/loki.go
Eraac: I've change this, now I write all fields like @loganmc10 mention it

plugins/outputs/loki/loki.go
Eraac: done

plugins/outputs/loki/loki.go
Eraac: TBH I've copy the code from `outputs/http` for http.Client relative stuff and there are special test unit to handle this case, I guess it's relevant?

plugins/outputs/loki/stream.go
Eraac: `metric.HashID` seems to return an unique identifier for each metric, regarding is the same tags or not. I need the same identifier if the tags are the same

1. `[]*telegraf.Tag` are got from `TagList()` method, which mention tags are "by the tag key in lexical bytewise ascending order", seems to be good for me?
2. Indeed, I should improve this and the unit test

plugins/outputs/loki/stream.go
Eraac: I probably can improve this part too, I will do it

plugins/outputs/loki/stream.go
Eraac: point 2 should be good now

plugins/outputs/loki/loki.go
srebhan: TBH I guess it is an oversight. Is there any service out there that accepts an empty username and a password? Does loki support this? If not we might want to catch that as it is _very likely_ a configuration error by the user, so he or she should know.

plugins/outputs/loki/loki_test.go
```diff
@@ -0,0 +1,370 @@
...
+		name     string
+		plugin   *Loki
+		expected string
+	}{
+		{
+			name: "default is text plain",
+			plugin: &Loki{
+				Domain: u.String(),
+			},
+			expected: "application/json",
```
srebhan, line 121: This test suggests that the default is `json` and not plain-text as the test-name suggests. Am I missing something?

plugins/outputs/loki/stream.go
```diff
@@ -0,0 +1,71 @@
...
+)
+
+func (s Streams) insertLog(ts []*telegraf.Tag, l Log) {
+	key := uniqKeyFromTagList(ts)
+	_, ok := s[key]
+
+	if !ok {
+		s[key] = newStream(ts)
+	}
```
srebhan, line 32: You could shorten this to a very common pattern:
```suggestion
	if _, ok := s[key]; !ok {
		s[key] = newStream(ts)
	}
```

plugins/outputs/loki/stream.go
```diff
@@ -0,0 +1,71 @@
...
+
+func uniqKeyFromTagList(ts []*telegraf.Tag) (k string) {
+	for _, t := range ts {
+		k += fmt.Sprintf("%s-%s-",
+			strings.ReplaceAll(t.Key, "-", "--"),
+			strings.ReplaceAll(t.Value, "-", "--"),
+		)
+	}
+
+	return
```
srebhan, line 57: I know golang supports this but in telegraf explicit returns are used. Please change this to `return k`.

plugins/outputs/loki/stream.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+func (s Streams) MarshalJSON() ([]byte, error) {
+	r := Request{
+		Streams: make([]Stream, 0, len(s)),
+	}
+
+	for _, stream := range s {
+		r.Streams = append(r.Streams, *stream)
+	}
```
srebhan, line 44: Just a heads-up: In golang access to maps is randomized, so if you rely on a certain order of streams you have to enforce this order after this loop. We had quite some problems with e.g. tests ignoring that fact and fail randomly.

As I said, no request for change bit just a heads-up.

srebhan (commented):
Almost there! I have two minor things left, one being the explicit `return k` in `uniqKeyFromTagList()` and the other is the handling of username/password. Please change the former and check the latter. :-)

Regarding the ordering of the tags you are absolutely right!

Furthermore there are some more cosmetic comments and one heads-up. Those you may or may not change depending on your taste and the necessity. :-)

Eraac:
Thanks for the review @srebhan

plugins/outputs/loki/loki.go
```diff
@@ -0,0 +1,214 @@
...
+		defer rc.Close()
+		reqBodyBuffer = rc
+	}
+
+	req, err := http.NewRequest(http.MethodPost, l.url, reqBodyBuffer)
+	if err != nil {
+		return err
+	}
+
+	if l.Username != "" && l.Password != "" {
```
srebhan, line 176: This is also not quite correct, sorry! :-/ It is possible (at least for most services not sure about loki) to have a user name with empty password. So either we do
```suggestion
	if l.Username != "" {
```
or you check in `Connect()` that a username is given if the password is not empty and throw an error!

plugins/outputs/loki/loki.go
srebhan: Please mark this as resolved.

srebhan (commented):
Still not statisfied with the username/password handling, sorry! :-/

Eraac:
@srebhan No problem, password is now optional

srebhan (approved):
This is beautiful! Thank you very much @Eraac! Looks good to me.

loganmc10:
It looks like the example config still has this:
```
## The field containing the log
  # field_line = "log"
```
even though it's no longer an option

loganmc10:
Sorry I should have noticed this put the first sentence in the readme also needs to be updated:

> This plugin sends logs to Loki, using tags as labels and **one field as log line**

srebhan (approved):
Good catch @loganmc10!

plugins/outputs/loki/loki.go
```diff
@@ -0,0 +1,214 @@
...
+
+	resp, err := l.client.Do(req)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+	_, err = ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return err
+	}
```
ssoroka, line 201: maybe check the body after the response status. There are plenty of response statuses that don't return bodies, and the Go library no longer leaks if you don't read the body.

ssoroka (approved):
approved with two minor suggestions

plugins/outputs/loki/loki.go
```diff
@@ -0,0 +1,214 @@
...
+
+  ## Basic auth credential
+  # username = "loki"
+  # password = "pass"
+
+  ## Additional HTTP headers
+  # http_headers = {"X-Scope-OrgID" = "1"}
+
+  ## If the request must be gzip encoded
+  # gzip_request = false
```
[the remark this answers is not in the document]
Eraac: Without a custom proxy Loki didn't accept gzip request bu default

plugins/outputs/loki/loki.go
Eraac: I think the ReadAll is useless here, the body isn't used. I've remove it and only check the status code

srebhan (approved):
Looks good. @ssoroka are you satisfied with not enabling `gzip` by default if the server doesn't support it without additional infrastructure?

loganmc10:
Gzip support was only merged into Loki master about 2 weeks ago, it's still not part of any released version. I wouldn't enable it by default.

nazar3n0:
Promtail allows to group labels by job while sending logs to Loki.
Is there a way to group labels by jobs with the Telegraf - Gragana Loki plugin ?

loganmc10:
@nazar3n0 what do you mean exactly by "group labels by job"? 'job' is usually one of the labels

nazar3n0:
> @nazar3n0 what do you mean exactly by "group labels by job"? 'job' is usually one of the labels

Promtail Config Example:
```

scrape_configs:
- job_name: TEST
  static_configs:
  - targets:
      - localhost
    labels:
      job: TEST
      __path__: /data/log/SERVERS/TEST.log
  pipeline_stages:
    - match:
        selector: '{job="TEST"}'
        stages:
        - regex:
            expression: ^\w+\s{1,}\d+ \S+ (?P<server>[\S+]+).*?\'(?P<asset>\b\S+\b).*
        - labels:
            server:
            asset:
```

This let me search at `job=TEST` strings at labels
`server or assets
`
Like this:
`topk(10,sum(rate({job="TEST",asset!=""} [15m])) by (asset))
`

And here I have Telegraf sending labels `server and assets`, but do not find the way to associate them to a `label "job"` as I can at promtail.

/etc/telegraf/telegraf.d/test.conf:

```
[[inputs.tail]]
  files = ["/data/log/SERVERS/TEST.log"]

  from_beginning = false
  watch_method = "poll"
  grok_patterns = [
'%{TIMESTAMP_ISO8601:time:ts-"2006-01-02T15:04:05.000-0700"} %{GREEDYDATA:message:string}'
]
  name_override = "edge_access_logs"
....
....
[[processors.converter]]
   order = 2
   namepass = ["edge_access_logs"]

  [processors.converter.fields]
    float = ["response_time"]
    tag = ["server", "assets"]

```

/etc/telegraf/telegraf.conf


```
[[outputs.loki]]
  namepass = ["edge_access_logs"]
  ## The domain of Loki
  domain = "http://10.200.181.240:3100"
..
....

```

I hope my question is understood. Thanks

nazar3n0:
> > @nazar3n0 what do you mean exactly by "group labels by job"? 'job' is usually one of the labels
> 
> Promtail Config Example:
> 
> ```
> 
> scrape_configs:
> - job_name: TEST
>   static_configs:
>   - targets:
>       - localhost
>     labels:
>       job: TEST
>       __path__: /data/log/SERVERS/TEST.log
>   pipeline_stages:
>     - match:
>         selector: '{job="TEST"}'
>         stages:
>         - regex:
>             expression: ^\w+\s{1,}\d+ \S+ (?P<server>[\S+]+).*?\'(?P<asset>\b\S+\b).*
>         - labels:
>             server:
>             asset:
> ```
> 
> This let me search at `job=TEST` strings at labels
> `server or assets `
> Like this:
> `topk(10,sum(rate({job="TEST",asset!=""} [15m])) by (asset)) `
> 
> And here I have Telegraf sending labels `server and assets`, but do not find the way to associate them to a `label "job"` as I can at promtail.
> 
> /etc/telegraf/telegraf.d/test.conf:
> 
> ```
> [[inputs.tail]]
>   files = ["/data/log/SERVERS/TEST.log"]
> 
>   from_beginning = false
>   watch_method = "poll"
>   grok_patterns = [
> '%{TIMESTAMP_ISO8601:time:ts-"2006-01-02T15:04:05.000-0700"} %{GREEDYDATA:message:string}'
> ]
>   name_override = "edge_access_logs"
> ....
> ....
> [[processors.converter]]
>    order = 2
>    namepass = ["edge_access_logs"]
> 
>   [processors.converter.fields]
>     float = ["response_time"]
>     tag = ["server", "assets"]
> ```
> 
> /etc/telegraf/telegraf.conf
> 
> ```
> [[outputs.loki]]
>   namepass = ["edge_access_logs"]
>   ## The domain of Loki
>   domain = "http://10.200.181.240:3100"
> ..
> ....
> ```
> 
> I hope my question is understood. Thanks

Solution:  [inputs.tail.tags]
    job = "mnv_edge_access_log"

gudujohn:
How should I set the 'inputs.tail' config?
Why is my 'inputs.tail' config is invalid.
Is there any tail configuration description for loki?
No error reported, just can't collect logs!
Here is some of my configuration information:
`[[inputs.tail]]
  files = ["/root/app/logs/xxx.log"]
  [inputs.tail.tags]
    target = "loki"`

gudujohn:
> How should I set the 'inputs.tail' config?
> Why is my 'inputs.tail' config is invalid.
> Is there any tail configuration description for loki?
> No error reported, just can't collect logs!
> Here is some of my configuration information:
> 
> > [[inputs.tail]]
> > files = ["/root/app/logs/xxx.log"]
> > data_format = "influx"
> > [inputs.tail.tags]
> > target = "loki"

I found the problem:(I hope my questions can help my friends who have the same problems)
I use the systemctl command. Its telegraf.service script uses the telegraf user and does not have access to the corresponding folder. But there is no error information in the log. It took a long time to find out.

PR #8463, opened 2020-11-23
sqlAzureMIResourceStats - add extra columns

IgorKuchmienko:
### Required for all PRs:

Add extra columns to query sqlAzureMIResourceStats.


denzilribeiro:
What will you get from capturing IO's at the resource stat layer? io bytes read/writte/IOs don't make a whole lot of sense to collect from sys.resource_stats, they are already in virtual file stats

Am fine with the rest of the changes.

IgorKuchmienko:
Per @denzilribeiro's advice, I could replace these metrics with virtual file stats so they are not necessary and we do not need this PR. We can close it.

PR #8464, opened 2020-11-23
Add column measurement_db_type to output of all queries

IgorKuchmienko:
### Required for all PRs:

This column is used to distinguish between outputs of queries that have the same value in column [measurement], such as sqlAzureDBResourceStats and sqlAzureMIResourceStats.

### DIFF
--- a/plugins/inputs/sqlserver/sqlserver.go
+++ b/plugins/inputs/sqlserver/sqlserver.go
@@ -307,7 +307,9 @@ func (s *SQLServer) accRow(query Query, acc telegraf.Accumulator, row scanner) e
 		}
 	}
 
-	tags["measurement_db_type"] = s.DatabaseType
+	if s.DatabaseType != "" {
+		tags["measurement_db_type"] = s.DatabaseType
+	}
 
 	if query.ResultByRow {
 		// add measurement to Accumulator
--- a/plugins/inputs/sqlserver/sqlserver.go
+++ b/plugins/inputs/sqlserver/sqlserver.go
@@ -307,6 +307,8 @@ func (s *SQLServer) accRow(query Query, acc telegraf.Accumulator, row scanner) e
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


denzilribeiro (approved):
Looks fine to me.  @Trovalo  - any objections to this

denzilribeiro (commented):
Realized that DatabaseType won't always exist if someone isn't using it in the config file ( aka are using the older V2/V1 queries). May have to check if it is configured in conf file.

Trovalo:
@denzilribeiro I see no problem in adding this.
Personally, I'll leave the V2 queries as they are and add the column only on the new ones. If we want to provide that info also for old query we can easily use SERVERPORPERTY('EngineEdition') to identify the correct value

denzilribeiro:
> Personally, I'll leave the V2 queries as they are and add the column only on the new ones. If we want to provide that info also for old query we can easily use SERVERPORPERTY('EngineEdition') to identify the correct value

But where it was being added was when results were being sent , appending a tag to all result sets which won't be conditional at this point.

Trovalo:
Sorry I didn't notice that, in any case it's fine to me.
The other option is to add the info in the query itself as a static tag (like the measurement), I have no preference, both methods are fine to me.

If we choose this way, adding a check to ensure that there is a valid value seems reasonable to me, otherwise we could end up with an empty tag, which is not that handy to query

IgorKuchmienko:
I added a check to ensure that the new column is only added when DatabaseType is not empty.

Trovalo (approved):
Looks fine to me, @denzilribeiro anything else to add?

denzilribeiro (approved):
Looks good to me

IgorKuchmienko:
Looks like there is nothing else to be done here. @ssoroka can you please merge it?

ssoroka (approved):

PR #8465, opened 2020-11-23
Add queries for collecting query store runtime and wait statistics

IgorKuchmienko:
### Required for all PRs:

Add queries for collecting query store runtime and wait statistics, for Azure SQL DB and MI.

Add ability to store in a hashtable a custom value associated with the query result, and use it when the query is run next time.
For query store queries, we store the time when the query was run last, and in the next run we only return the data between the last run time and the current time, rounded down to a multiple of the reporting interval.

The implementation takes care of two cases:
1) query store aggregation interval (configured in SQL) is larger than or equal to the telegraf interval
In this case, the query will return data only once per aggregation interval, and will return an empty set for each subsequent telegraf call within the same aggregation interval. 
2) query store aggregation interval is smaller than the telegraf interval
In this case, the query will return records from several aggregation intervals, i.e. all the new data since the previous telegraf call.


plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -382,6 +382,116 @@ HAVING
...
+END;
+`
+
+const sqlAzureDBQueryStoreWaitStatistics = sqlAzureDBPartQueryPeriod + `
+SET DEADLOCK_PRIORITY -10;
+IF SERVERPROPERTY('EngineEdition') = 5  -- Is this Azure SQL DB?
+BEGIN
+SELECT
+	'sqlserver_azuredb_querystore_wait_stats' AS [measurement],
+	CONVERT(varchar(36), NEWID()) AS [measurement_id],
```
denzilribeiro, line 468: I don't understand need for measurement_id

plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -382,6 +382,116 @@ HAVING
...
+
+IF @QueryData != '' AND @QueryData IS NOT NULL
+	SET @lastIntervalEndTimestamp = CAST(@QueryData AS BIGINT);
+
+DECLARE @magicNumberDays BIGINT = 4294967296, @magicNumberSeconds BIGINT = 300, @secondsInDay INT = 86400;
+
+DECLARE @timeGrain BIGINT = (SELECT interval_length_minutes * 60 FROM sys.database_query_store_options WITH (NOLOCK)); --use query store time grain
+
+DECLARE @currDate DATETIME = GETUTCDATE();
+DECLARE @currIntervalEndTimestamp BIGINT = CONVERT(BIGINT, DATEDIFF(day, 0, @currDate)) * @secondsInDay + DATEDIFF(second, DATEADD(day, DATEDIFF(day, 0, @currDate), 0), @currDate) / @timeGrain * @timeGrain;
```
denzilribeiro, line 398: Can you add comments of what you are trying to achieve here?  I know has to do with interval config but right now reviewing the code isn't easy

plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -382,6 +382,116 @@ HAVING
...
+	REPLACE(@@SERVERNAME, '\', ':') AS [sql_instance],
+	DB_NAME() AS [database_name],
+	CONVERT(datetime, rsi.start_time, 1) AS interval_start_time,
+	CONVERT(datetime, rsi.end_time, 1) AS interval_end_time,
+	CONVERT(BIGINT, DATEDIFF(day, 0, CONVERT(datetime, rsi.start_time, 1))) * @magicNumberDays + DATEDIFF(second, 0, DATEADD(day,-DATEDIFF(day, 0, CONVERT(datetime, rsi.start_time, 1)), CONVERT(datetime, rsi.start_time, 1))) * @magicNumberSeconds AS interval_start_time_d,
+	CONVERT(BIGINT, DATEDIFF(day, 0, CONVERT(datetime, rsi.end_time, 1))) * @magicNumberDays + DATEDIFF(second, 0, DATEADD(day,-DATEDIFF(day, 0, CONVERT(datetime, rsi.end_time, 1)), CONVERT(datetime, rsi.end_time, 1))) * @magicNumberSeconds AS interval_end_time_d,
+	q.query_id,
+	CONVERT(nvarchar(max), q.query_hash, 1) as query_hash,
+	p.plan_id,
+	CONVERT(nvarchar(max), q.last_compile_batch_sql_handle, 1) as last_compile_batch_sql_handle,
```
denzilribeiro, line 478: Do we really need last compile batch handle?

plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -382,6 +382,116 @@ HAVING
...
+	CONVERT(nvarchar(max), p.query_plan_hash, 1) as query_plan_hash,
```
denzilribeiro, line 479: Why nvarchar(max) it is a binary 8.. so can cast to string but don't need a max.

plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -382,6 +382,116 @@ HAVING
...
+	REPLACE(@@SERVERNAME, '\', ':') AS [sql_instance],
+	DB_NAME() AS [database_name],
+	CONVERT(datetime, rsi.start_time, 1) AS interval_start_time,
+	CONVERT(datetime, rsi.end_time, 1) AS interval_end_time,
+	CONVERT(BIGINT, DATEDIFF(day, 0, CONVERT(datetime, rsi.start_time, 1))) * @magicNumberDays + DATEDIFF(second, 0, DATEADD(day,-DATEDIFF(day, 0, CONVERT(datetime, rsi.start_time, 1)), CONVERT(datetime, rsi.start_time, 1))) * @magicNumberSeconds AS interval_start_time_d,
+	CONVERT(BIGINT, DATEDIFF(day, 0, CONVERT(datetime, rsi.end_time, 1))) * @magicNumberDays + DATEDIFF(second, 0, DATEADD(day,-DATEDIFF(day, 0, CONVERT(datetime, rsi.end_time, 1)), CONVERT(datetime, rsi.end_time, 1))) * @magicNumberSeconds AS interval_end_time_d,
+	q.query_id,
+	CONVERT(nvarchar(max), q.query_hash, 1) as query_hash,
+	p.plan_id,
+	CONVERT(nvarchar(max), q.last_compile_batch_sql_handle, 1) as last_compile_batch_sql_handle,
```
denzilribeiro, line 427: Why do we need last_compile_batch_sql_handle?

plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -382,6 +382,116 @@ HAVING
...
+	'sqlserver_azuredb_querystore_runtime_stats' AS [measurement],
+	CONVERT(varchar(36), NEWID()) AS [measurement_id],
+	REPLACE(@@SERVERNAME, '\', ':') AS [sql_instance],
+	DB_NAME() AS [database_name],
+	CONVERT(datetime, rsi.start_time, 1) AS interval_start_time,
+	CONVERT(datetime, rsi.end_time, 1) AS interval_end_time,
+	CONVERT(BIGINT, DATEDIFF(day, 0, CONVERT(datetime, rsi.start_time, 1))) * @magicNumberDays + DATEDIFF(second, 0, DATEADD(day,-DATEDIFF(day, 0, CONVERT(datetime, rsi.start_time, 1)), CONVERT(datetime, rsi.start_time, 1))) * @magicNumberSeconds AS interval_start_time_d,
+	CONVERT(BIGINT, DATEDIFF(day, 0, CONVERT(datetime, rsi.end_time, 1))) * @magicNumberDays + DATEDIFF(second, 0, DATEADD(day,-DATEDIFF(day, 0, CONVERT(datetime, rsi.end_time, 1)), CONVERT(datetime, rsi.end_time, 1))) * @magicNumberSeconds AS interval_end_time_d,
+	q.query_id,
+	CONVERT(nvarchar(max), q.query_hash, 1) as query_hash,
```
denzilribeiro, line 425: No need to cast to nvarchar(max) smaller will do as hash is binary 8 if I remember

plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -382,6 +382,116 @@ HAVING
...
+	'sqlserver_azuredb_querystore_wait_stats' AS [measurement],
+	CONVERT(varchar(36), NEWID()) AS [measurement_id],
+	REPLACE(@@SERVERNAME, '\', ':') AS [sql_instance],
+	DB_NAME() AS [database_name],
+	CONVERT(datetime, rsi.start_time, 1) AS interval_start_time,
+	CONVERT(datetime, rsi.end_time, 1) AS interval_end_time,
+	CONVERT(BIGINT, DATEDIFF(day, 0, CONVERT(datetime, rsi.start_time, 1))) * @magicNumberDays + DATEDIFF(second, 0, DATEADD(day,-DATEDIFF(day, 0, CONVERT(datetime, rsi.start_time, 1)), CONVERT(datetime, rsi.start_time, 1))) * @magicNumberSeconds AS interval_start_time_d,
+	CONVERT(BIGINT, DATEDIFF(day, 0, CONVERT(datetime, rsi.end_time, 1))) * @magicNumberDays + DATEDIFF(second, 0, DATEADD(day,-DATEDIFF(day, 0, CONVERT(datetime, rsi.end_time, 1)), CONVERT(datetime, rsi.end_time, 1))) * @magicNumberSeconds AS interval_end_time_d,
+	q.query_id,
+	CONVERT(nvarchar(max), q.query_hash, 1) as query_hash,
```
denzilribeiro, line 476: Cast to smaller string.

plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -382,6 +382,116 @@ HAVING
...
 
+const sqlAzureDBPartQueryPeriod = `
+DECLARE @QueryData NVARCHAR(MAX) = ?1; --input parameter
+
+DECLARE @lastIntervalEndTimestamp BIGINT = 0;
+
+IF @QueryData != '' AND @QueryData IS NOT NULL
+	SET @lastIntervalEndTimestamp = CAST(@QueryData AS BIGINT);
+
+DECLARE @magicNumberDays BIGINT = 4294967296, @magicNumberSeconds BIGINT = 300, @secondsInDay INT = 86400;
```
denzilribeiro, line 393: Why is Magic Number seconds 5 mins and not 15 mins? Any logic to it?

plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -1178,3 +1288,235 @@ SELECT
...
+	REPLACE(@@SERVERNAME, ''\'', '':'') AS [sql_instance],
+	DB_NAME() AS [database_name],
+	CONVERT(datetime, rsi.start_time, 1) AS interval_start_time,
+	CONVERT(datetime, rsi.end_time, 1) AS interval_end_time,
+	CONVERT(BIGINT, DATEDIFF(day, 0, CONVERT(datetime, rsi.start_time, 1))) * @magicNumberDays + DATEDIFF(second, 0, DATEADD(day,-DATEDIFF(day, 0, CONVERT(datetime, rsi.start_time, 1)), CONVERT(datetime, rsi.start_time, 1))) * @magicNumberSeconds AS interval_start_time_d,
+	CONVERT(BIGINT, DATEDIFF(day, 0, CONVERT(datetime, rsi.end_time, 1))) * @magicNumberDays + DATEDIFF(second, 0, DATEADD(day,-DATEDIFF(day, 0, CONVERT(datetime, rsi.end_time, 1)), CONVERT(datetime, rsi.end_time, 1))) * @magicNumberSeconds AS interval_end_time_d,
+	q.query_id,
+	CONVERT(nvarchar(max), q.query_hash, 1) as query_hash,
+	p.plan_id,
+	CONVERT(nvarchar(max), q.last_compile_batch_sql_handle, 1) as last_compile_batch_sql_handle,
```
denzilribeiro, line 1503: Do we need this column ( last_compile_atch_sql_handle)

plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -382,6 +382,116 @@ HAVING
...
+	rs.avg_duration as avg_duration,
+	rs.max_num_physical_io_reads as max_num_physical_io_reads,
+	rs.avg_num_physical_io_reads as avg_num_physical_io_reads,
+	rs.max_log_bytes_used as [max_log_bytes_used],
+	rs.avg_log_bytes_used as avg_log_bytes_used
+FROM sys.query_store_query q WITH (NOLOCK)
+	JOIN sys.query_store_plan p WITH (NOLOCK) ON q.query_id = p.query_id
+	JOIN sys.query_store_runtime_stats rs WITH (NOLOCK) ON p.plan_id = rs.plan_id
+	JOIN sys.query_store_runtime_stats_interval rsi WITH (NOLOCK) ON rs.runtime_stats_interval_id = rsi.runtime_stats_interval_id
+WHERE rsi.start_time >= @currIntervalStartDate AND rsi.end_time <= @currIntervalEndDate;
```
denzilribeiro, line 456: Wouldn't it reduce amount of data (specially for MI) to collect aggregated data by query_id / plan_id even if you bucket say in 1 hour intervals - right now if someone changes collection interval to 5 mins or 10 mins will be a lot of data being collected.

plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -382,6 +382,116 @@ HAVING
...
+	p.plan_id,
+	CONVERT(nvarchar(max), q.last_compile_batch_sql_handle, 1) as last_compile_batch_sql_handle,
+	CONVERT(nvarchar(max), p.query_plan_hash, 1) as query_plan_hash,
+	ws.wait_category_desc AS wait_category_s,
+	ws.execution_type AS exec_type_d,
+	ws.total_query_wait_time_ms AS total_query_wait_time_ms_d,
+	ws.max_query_wait_time_ms AS max_query_wait_time_ms_d,
+	q.query_parameterization_type AS query_param_type_d
+FROM sys.query_store_query q WITH (NOLOCK)
+	JOIN sys.query_store_plan p WITH (NOLOCK) ON q.query_id = p.query_id
```
denzilribeiro, line 486: Do we need this dmv? only thing being returned is plan_id which I beleive is in runtimestats

plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -1178,3 +1288,235 @@ SELECT
...
```
denzilribeiro, line 1511: Do we need sys.query-store_plan? only column is plan_id which is in the waitstats DMV.

plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -1178,3 +1288,235 @@ SELECT
...
+DECLARE @lastIntervalEndTimestamp BIGINT = 0;
+SELECT TOP 1 @lastIntervalEndTimestamp = [timestamp] FROM #CachedDbTimestamps WHERE [id] = DB_ID();
+
+INSERT INTO #DbTimestamps([id], [timestamp]) VALUES(DB_ID(), @lastIntervalEndTimestamp)
+
+DECLARE @magicNumberDays BIGINT = 4294967296, @magicNumberSeconds BIGINT = 300, @secondsInDay INT = 86400;
+
+DECLARE @timeGrain BIGINT = (SELECT interval_length_minutes * 60 FROM sys.database_query_store_options WITH (NOLOCK)); --use query store time grain
+
+DECLARE @currDate DATETIME = GETUTCDATE();
```
denzilribeiro, line 1356: Please add comments of what these time calculations are doing for maintainability.

plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -382,6 +382,116 @@ HAVING
...
+DECLARE @currIntervalStartTimestamp BIGINT = @currIntervalEndTimestamp - @timeGrain;
+IF @lastIntervalEndTimestamp != 0
+	SET @currIntervalStartTimestamp = @lastIntervalEndTimestamp;
+
+DECLARE @currIntervalEndDate datetime = DATEADD(second, @currIntervalEndTimestamp % @secondsInDay, DATEADD(day, @currIntervalEndTimestamp / @secondsInDay, 0));
+DECLARE @currIntervalStartDate datetime = DATEADD(second, @currIntervalStartTimestamp % @secondsInDay, DATEADD(day, @currIntervalStartTimestamp / @secondsInDay, 0));
+
+`
+
+const sqlAzureDBQueryStoreRuntimeStatistics = sqlAzureDBPartQueryPeriod + `
```
denzilribeiro, line 411: You  may want to check Updateability (until we have query store on read replicas). If not you will be collecting query store data (duplicate or same) from 2 replicas if  read replica in BC/Hyperscale is configured for monitoring. Collect only if it is READ_WRITE.
https://docs.microsoft.com/en-us/sql/t-sql/functions/databasepropertyex-transact-sql?view=sql-server-ver15

denzilribeiro (changes requested):
All the comments put for SQLDB apply to MI as well. I have looked closer at SQLDB for now but haven't tested on running workload. For all the timestamp calculations please add comments for maintainability. I also feel we should have some "check" given we now have a cache to NOT run this query if it has been run say < 15 mins ago or something. Think of a person who had an old config file, who hasn't added this new addition to their "exclude" list so this would for them run by default at a 10 second interval. Min collection should be 15 mins. Also have you done any perf testing on collecting this on a SQL with workload? And for MI on an MI with multiple databases running a workload?
The other thing I noticed, is if someone changes the interval below the collection time as is we are not "merging' or deduping right so will get queries counted twice.

plugins/inputs/sqlserver/azuresqlqueries.go
IgorKuchmienko: Removed

plugins/inputs/sqlserver/azuresqlqueries.go
IgorKuchmienko: Removed magic number. Now we return start_time and end_time as strings.

plugins/inputs/sqlserver/azuresqlqueries.go
IgorKuchmienko: Removed.

plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -382,6 +382,116 @@ HAVING
...
+SET DEADLOCK_PRIORITY -10;
+IF SERVERPROPERTY('EngineEdition') = 5  -- Is this Azure SQL DB?
+BEGIN
+SELECT
+	'sqlserver_azuredb_querystore_runtime_stats' AS [measurement],
+	CONVERT(varchar(36), NEWID()) AS [measurement_id],
```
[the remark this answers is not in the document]
IgorKuchmienko: Removed. Measurement_ID was needed because in the output recordset we had multiple records for the same query in the same time interval, but it is no longer needed now that aggregation is added https://github.com/influxdata/telegraf/pull/8465/commits/3cf25f6ccab56ae377db346adef5d42c60ed7d70

plugins/inputs/sqlserver/azuresqlqueries.go
IgorKuchmienko: Removed.

plugins/inputs/sqlserver/azuresqlqueries.go
IgorKuchmienko: Added comments

plugins/inputs/sqlserver/azuresqlqueries.go
IgorKuchmienko: Removed

plugins/inputs/sqlserver/azuresqlqueries.go
IgorKuchmienko: Added comments

plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -382,6 +382,116 @@ HAVING
...
+	rs.max_dop as max_dop,
+	rs.avg_dop as avg_dop,
+	rs.max_rowcount as max_rowcount,
+	rs.avg_rowcount as avg_rowcount,
+	rs.max_query_max_used_memory AS max_query_max_used_memory,
+	rs.avg_query_max_used_memory AS avg_query_max_used_memory,
+	rs.max_duration as [max_duration],
+	rs.avg_duration as avg_duration,
+	rs.max_num_physical_io_reads as max_num_physical_io_reads,
+	rs.avg_num_physical_io_reads as avg_num_physical_io_reads,
```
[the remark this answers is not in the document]
IgorKuchmienko: Added.

plugins/inputs/sqlserver/azuresqlqueries.go
IgorKuchmienko: Changed to nvarchar(20)

plugins/inputs/sqlserver/azuresqlqueries.go
IgorKuchmienko: Changed to nvarchar(20)

plugins/inputs/sqlserver/azuresqlqueries.go
IgorKuchmienko: Changed to nvarchar(20)

IgorKuchmienko:
> I also feel we should have some "check" given we now have a cache to NOT run this query if it has been run say < 15 mins ago or something. 

> The other thing I noticed, is if someone changes the interval below the collection time as is we are not "merging' or deduping right so will get queries counted twice.

Yes, this is exactly what the cache is used for. Whatever the interval in telegraf is set to, by using the cached time we guarantee that nothing is returned from telegraf while there is no new data in query store, and no duplicates are returned. I edited the comment in the description of this PR to elaborate on this, and I added comments in the implementation.

plugins/inputs/sqlserver/azuresqlqueries.go
IgorKuchmienko: If the aggregation interval in SQL is set to 1 hour, the telegraf will return data only once per hour, whatever the collection interval is set to.

I did add aggregation by plan_id, execution_type and runtime_stats_interval_id, as suggested here:
https://docs.microsoft.com/en-us/sql/relational-databases/system-catalog-views/sys-query-store-runtime-stats-transact-sql?view=sql-server-ver15 
-- to guarantee that there are no multiple records for the same time interval and plan_id.

IgorKuchmienko:
These tests were run for the queries.

1. On Azure SQL DB
Telegraf VM tier: Basic A1 (1 vcpu, 1.75 GiB memory)
SQL DB tier: Basic

- case 1
query store aggregation interval = 60 min
telegraf interval = 15 min

- case 2
query store aggregation interval = 5 min
telegraf interval = 15 min

In both cases, a CPU load was generated with a T-SQL script on the DB.

2. On Azure MI
Telegraf VM tier: Basic A1 (1 vcpu, 1.75 GiB memory)
MI tier: Business Critical Gen5 (32 GB, 4 vCores)
Number of DBs: 7
query store aggregation interval = 60 min 
telegraf interval = 15 min

The plug-in was configured to output the data into a socket, and it was then streamed into an Azure Log Analytics workspace by the Log Analytics agent installed on the VM. The tests showed that the queries produce the data as expected: once per aggregation interval and with no duplicate records. There are no gaps in data and telegraf log file tgf.log contains no errors.

Test results are here: [Tests.zip](https://github.com/influxdata/telegraf/files/5749095/Tests.zip).
They contain the data produced by the queries (exported from Azure LA), CPU load generation script, telegraf config and log files.

plugins/inputs/sqlserver/azuresqlqueries.go
denzilribeiro: We still need this check here aka collect only if a read/write database and not a read only replica

plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -1196,3 +1320,261 @@ SELECT
...
+IF OBJECT_ID ('tempdb.dbo.#ExecForeachDb') IS NOT NULL DROP PROCEDURE #ExecForeachDb;
+
+/* Create a temporary procedure that iterates through all DBs that the user has access to in this managed instance, except system DBs, and execute the query specified with @queryText */
+EXEC('CREATE PROCEDURE #ExecForeachDb @queryText NVARCHAR(MAX)
+AS
+BEGIN
+	SET NOCOUNT ON;
+
+	DECLARE dbCursor CURSOR LOCAL FOR 
+	SELECT
```
denzilribeiro, line 1342: Need other conditions, aka database is online, and read/write ( not read replica) - aka sys.database state column
This also needs error handling, check if query store is enabled as there is no Try Catch block here and it is iterating through multiple databases

plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -1196,3 +1320,261 @@ SELECT
 	,DATABASEPROPERTYEX(DB_NAME(), 'Updateability') as replica_updateability
 FROM sys.dm_os_schedulers AS s
 `
+
+const sqlAzureMIPartQueryPeriod = `
```
denzilribeiro, line 1324: This is a sizable query, have we thought of limiting what we retrieve? Aka take for example MI with 20 databases, creating a temp table with everything in query store could be expensive? Should you get like the top 50 from each?

plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -1196,3 +1320,261 @@ SELECT
...
+const sqlAzureMIQueryStoreRuntimeStatistics = sqlAzureMIPartQueryPeriod + `
+/* Create a temp table for accumulating the result for all DBs */
+IF OBJECT_ID ('tempdb.dbo.#QueryStoreMetrics') IS NOT NULL DROP TABLE #QueryStoreMetrics;
+CREATE TABLE #QueryStoreMetrics(
+	measurement varchar(42),
+	sql_instance nvarchar(256),
+	[database_name] nvarchar(128),
+	interval_start_time nvarchar(30),
+	interval_end_time nvarchar(30),
+	query_id bigint,
```
denzilribeiro, line 1431: query_id should be a TAG ( string) as not a measurement

denzilribeiro (changes requested):
Need some changes, error handling in the MI part, checks for databases that they are online, that it isn't a read replica and some minor chances of some columns to tags.

plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -1196,3 +1320,261 @@ SELECT
...
+	WHERE HAS_DBACCESS([name]) = 1 
+		AND [name] NOT IN (''master'', ''tempdb'', ''model'', ''msdb'');
+	DECLARE @CurrSqlText nvarchar(max);
+	DECLARE @dbName sysname;
+
+	OPEN dbCursor;
+	FETCH NEXT FROM dbCursor INTO @dbName;
+
+	WHILE @@Fetch_Status=0 BEGIN
+		SET @CurrSqlText = N''USE ''+ QUOTENAME(@dbName ,''"'') + @queryText
```
dimitri-furman, line 1354: A database may be dropped between the time the cursor is open and the time this USE executes.

mhall119:
Hi @IgorKuchmienko, are you still able to working on this PR? Please let me know if there's anything we can do to help

masree:
> Hi @IgorKuchmienko, are you still able to working on this PR? Please let me know if there's anything we can do to help

Yes, we'll be working on this PR. Unfortunately, the PR got delayed and ETA is around July.

masree:
> > Hi @IgorKuchmienko, are you still able to working on this PR? Please let me know if there's anything we can do to help
> 
> Yes, we'll be working on this PR. Unfortunately, the PR got delayed and ETA is around June/July.

plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -389,6 +389,130 @@ HAVING
...
+DECLARE @currIntervalStartTimestamp BIGINT = @currIntervalEndTimestamp - @timeGrain;
+
+/* If the query was called before, start time is when the query was previously called, so that we return all the records since the previous call */
+IF @lastIntervalEndTimestamp != 0
+	SET @currIntervalStartTimestamp = @lastIntervalEndTimestamp;
+
+/* Convert start and end times to datetime */
+DECLARE @currIntervalEndDate datetime = DATEADD(second, @currIntervalEndTimestamp % @secondsInDay, DATEADD(day, @currIntervalEndTimestamp / @secondsInDay, 0));
+DECLARE @currIntervalStartDate datetime = DATEADD(second, @currIntervalStartTimestamp % @secondsInDay, DATEADD(day, @currIntervalStartTimestamp / @secondsInDay, 0));
+
```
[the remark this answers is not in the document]
masree: @dimitri-furman Is the recommendation to use datetime2 at all places? datetime2 seems to have higher precision than datetime, however it is not aware of timezone offset.

masree:
Splitting up this PR into 2 for easier review. New PR - https://github.com/influxdata/telegraf/pull/9150

plugins/inputs/sqlserver/azuresqlqueries.go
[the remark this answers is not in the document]
dimitri-furman: Query Store uses datetimeoffset in sys.query_store_runtime_stats_interval. We should stay consistent and use the same data type to avoid type conversion problems. But you should check if the current time math that uses datetime still works correctly with datetimeoffset.

masree:
> All the comments put for SQLDB apply to MI as well. I have looked closer at SQLDB for now but haven't tested on running workload. For all the timestamp calculations please add comments for maintainability. I also feel we should have some "check" given we now have a cache to NOT run this query if it has been run say < 15 mins ago or something. Think of a person who had an old config file, who hasn't added this new addition to their "exclude" list so this would for them run by default at a 10 second interval. Min collection should be 15 mins. Also have you done any perf testing on collecting this on a SQL with workload? And for MI on an MI with multiple databases running a workload?
> The other thing I noticed, is if someone changes the interval below the collection time as is we are not "merging' or deduping right so will get queries counted twice.

@denzilribeiro : Re " Think of a person who had an old config file, who hasn't added this new addition to their "exclude" list so this would for them run by default at a 10 second interval. Min collection should be 15 mins. " - for an user with old config and latest telegraf bits, yes it would start to collect QDS data at the interval specified.  One possible way to avoid it is to introduce another config parameter "query_store_collection" set to false by default. If user wants to enables QDS queries, they will need to set query_store_collection = true and also remove the qds query list from exclude. if they just change query_store_collection to true without  modifying exclude list, data will still not be collected

masree:
#9150

sjwang90:
Closing in favor of #9355

Issue #6330, opened 2019-08-29
SNMP integer fields are not translated


Hipska:
It seems that integer fields that have translations available, are actually not translated. I thought it should have been set in the field `conversion` parameter, but it seems there is no `translate` or similar option.

It seems like I could work around by using something as the reverse of the enum processor plugin. (not sure if this is possible)

### Relevant telegraf.conf:
```toml
  [[inputs.snmp.table]]
    name = "interface"

    [[inputs.snmp.table.field]]
      oid = "IF-MIB::ifName"
      is_tag = true

    [[inputs.snmp.table.field]]
      oid = "IF-MIB::ifAdminStatus"
      is_tag = true

    [[inputs.snmp.table.field]]
      name = "ifInOctets"
      oid = "IF-MIB::ifHCInOctets"

    [[inputs.snmp.table.field]]
      name = "ifOutOctets"
      oid = "IF-MIB::ifHCOutOctets"
```

### System info:

Telegraf 1.11.4 (git: HEAD d9ca76e4)
RedHat 7.4

### Expected behavior:
```
> interface,ifAdminStatus=up(1),ifName=InLoopBack0 ifInOctets=0i,ifOutOctets=0i
```
### Actual behavior:
```
> interface,ifAdminStatus=1,ifName=InLoopBack0 ifInOctets=0i,ifOutOctets=0i
```
### Additional info:
Mib:
```mib
ifAdminStatus OBJECT-TYPE
    SYNTAX INTEGER {
        up(1),
        down(2),
        testing(3)
        }
    ACCESS read-write
    STATUS mandatory
    DESCRIPTION
        "The desired state of the interface.  The testing(3) state
        indicates that no operational packets can be passed.  When a
        managed system initializes, all interfaces start with
        ifAdminStatus in the down(2) state.  As a result of either
        explicit management action or per configuration information
        retained by the managed system, ifAdminStatus is then
        changed to either the up(1) or testing(3) states (or remains
        in the down(2) state)."
    ::= { ifEntry 7 }
```
Output from `snmpget`:
```
IF-MIB::ifAdminStatus.1 = INTEGER: up(1)
```

goller:
@Hipska I like your idea of using the entire enum value as the tag.

comes from:

```
SYNTAX INTEGER {
       up(1),
       down(2),
       testing(3)
       }
```

Previously, I had been thinking of just `up` rather than `up(1)`... I like your way better for sure.

Hipska:
[...]

PR #8466, opened 2020-11-24
Translate snmp field values

### LINKED ISSUE #8458: Enable snmptranslate to field values
## Feature Request

The current implementation for the snmp input plugin does not enable snmptranslate for the field values, so there are some cases when the user is left with OID values for fields. I ran into this when ENTITY-MIB::entPhysicalVendorType returned an OID value. 

### Proposal:
Enable translating value for individual fields

### Current behavior:
E.g.
```
[[inputs.snmp.table]]
    oid = "ENTITY-MIB::entPhysicalTable"
 
    [[inputs.snmp.table.field]]
      oid = "ENTITY-MIB::entPhysicalVendorType"
```

### Desired behavior:
```
[[inputs.snmp.table]]
    oid = "ENTITY-MIB::entPhysicalTable"
 
    [[inputs.snmp.table.field]]
      oid = "ENTITY-MIB::entPhysicalVendorType"
      translate = true
```

### Use case: 
I have the need this kind of behavior enabled and all the prerequisites for snmptranslate already exist. I'd be happy to implement this if other people feel that this feature is good enough to be included in the snmp plugin.

LeadingMoominExpert:
### Required for all PRs:

Fixes #8458


plugins/inputs/snmp/snmp.go
```diff
@@ -460,6 +462,21 @@ func (t Table) Build(gs snmpConnection, walk bool) (*RTable, error) {
...
+					case string:
+						{
+							entOid := ent.Value.(string)
+							_, _, oidText, _, err := SnmpTranslate(entOid)
+							if err == nil {
+								// If no error translating, the original value for ent.Value should be replaced
+								ent.Value = oidText
+							}
+						}
+					}
```
srebhan, line 477: I think this is overly complex. How about
```suggestion
					if entOid, ok := ent.Value.(string); ok {
						_, _, oidText, _, err := SnmpTranslate(entOid)
						if err == nil {
							// If no error translating, the original value for ent.Value should be replaced
							ent.Value = oidText
						}
					}
```

srebhan (commented):
@LeadingMoominExpert you need to add the description of the new option also to the `sampleConfig` variable.
Furthermore, I'm not sure if your testcase actually is suited to test the translate functionality as you specify the name in the `tbl`. How can you see if it is *actually* overwritten using the translation?

plugins/inputs/snmp/snmp.go
LeadingMoominExpert: Your'e right, the switch is redundant here.

LeadingMoominExpert:
@srebhan I modified the tests so that a field returns an OID value with cases of this value being translated and not translated with mocked snmptranslate command.

LeadingMoominExpert:
I also did not include the translate functionality in `sampleConfig` as it seems to only have descriptions of the plugin definitions and not table or field definitions. 
https://github.com/influxdata/telegraf/blob/a51a9a86a7b9f1be510b6c1a9e8ef9ef17aad9a2/plugins/inputs/snmp/snmp.go#L25-L69

srebhan (approved):
Fair enough. :-) Looks good.

ssoroka (approved):

Hipska:
Does this also solve #6330? And could you give an example output with the function enabled? Thanks.

ssoroka:
Looks like it doesn't solve it for integer fields.

PR #8478, opened 2020-11-27
GTFS-realtime vehicle locations input plugin

gavincabbage:
### Required for all PRs:


go.mod
```diff
@@ -158,3 +159,7 @@ require (
 
 // replaced due to https://github.com/satori/go.uuid/issues/73
 replace github.com/satori/go.uuid => github.com/gofrs/uuid v3.2.0+incompatible
+
+replace github.com/golang/protobuf => github.com/golang/protobuf v1.3.3
+
+replace google.golang.org/genproto => google.golang.org/genproto v0.0.0-20200218151345-dad8c97a84f5
```
srebhan, line 165: Is this really required?

plugins/inputs/gtfs/README.md
```diff
@@ -0,0 +1,81 @@
...
+### Configuration
+
+This plugin can be configured to collect any combination of vehicle positions, trip updates
+and service alerts. Enable any one of these three input types by populating its URL in the
+plugin configuration. At least one of the three must be configured.
+
+```toml
+[[inputs.gtfs]]
+  ## API Key
+  # key = "${GTFS_API_KEY}"
```
srebhan, line 16: Does it mean "from environment"? If so, please add a comment above.

plugins/inputs/gtfs/README.md
```diff
@@ -0,0 +1,81 @@
...
+  ## URL for fetching vehicle positions
+  # vehicle_positions_url = "https://host.test/VehiclePositions.pb"
+  ## URL for fetching vehicle positions
+  # trip_updates_url = "https://host.test/TripUpdates.pb"
+  ## URL for fetching vehicle positions
+  # service_alerts_url = "https://host.test/ServiceAlerts.pb"
```
srebhan, line 22: You maybe should add a base-url here or is it a common use-case to use different servers to serve those?
Furthermore, can the suffices (`VehiclePositions.pb`, `TripUpdates.pb` and `ServiceAlerts.pb`) actually vary? If not it would be better to move them to the code...

plugins/inputs/gtfs/README.md
```diff
@@ -0,0 +1,81 @@
...
+  ## Use TLS but skip chain & host verification
+  # insecure_skip_verify = false
+
+  ## Amount of time allowed to complete the HTTP request
+  # timeout = "5s"
+```
+
+### Metrics
+
+Metrics can be collected for each of three GTFS-realtime feed types: vehicle positions, trip updates and service alerts.
```
srebhan, line 41: Please name the output metrics here as a list in form
```
- gtfs_position
- gtfs_updates
- gtfs_alerts
```
so a user gets an idea what to expect.

plugins/inputs/gtfs/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+#### Vehicle Positions
+
+The vehicle `position` measurement captures information about the vehicles including location and congestion level.
```
srebhan, line 45: Similar to the metrics, please be verbose here and list the fieldnames that are produced.

plugins/inputs/gtfs/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+##### Fields
```
srebhan, line 47: Put this before the "Vehicle positions header". You might also want to define this in the form
```
- gtfs_position
    - lon
    - lat
    - ...
- gtfs_updates
    - ...
- gtfs_alerts
    - ...
```

plugins/inputs/gtfs/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+Metrics can be collected for each of three GTFS-realtime feed types: vehicle positions, trip updates and service alerts.
+
+#### Vehicle Positions
+
+The vehicle `position` measurement captures information about the vehicles including location and congestion level.
+
+##### Fields 
+
+##### Tags
```
srebhan, line 49: Same as above.

plugins/inputs/gtfs/README.md
```diff
@@ -0,0 +1,81 @@
...
+The service `alert` measurement captures information about stop moves and unforeseen events 
+affecting a station, route or the entire network.
+
+##### Fields 
+
+##### Tags
+
+### Examples
+
+An example of vehicle positions from the Boston's MBTA GTFS-realtime feed:
```
srebhan, line 70: Is there a public URL for trying this out? If so, please add here so users can check what to expect from the plugin.

plugins/inputs/gtfs/gfts_test.go
```diff
@@ -0,0 +1,191 @@
+package gtfs_test
```
srebhan, line 1: No this should be
```suggestion
package gtfs
```

plugins/inputs/gtfs/gfts_test.go
```diff
@@ -0,0 +1,191 @@
...
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	gtfsbindings "github.com/MobilityData/gtfs-realtime-bindings/golang/gtfs"
+	"github.com/golang/protobuf/proto"
+	"github.com/stretchr/testify/require"
+
+	"github.com/influxdata/telegraf/plugins/inputs/gtfs"
```
srebhan, line 12: Don't import this and rather put the test into the same package.

plugins/inputs/gtfs/gfts_test.go
```diff
@@ -0,0 +1,191 @@
...
+		testLat          = float32(42.42)
+		testLng          = float32(20.20)
+		testBearing      = float32(270)
+		testOdometer     = float64(1000)
+		testSpeed        = float32(60)
+		testVersion      = "v0.0.0"
+		testHeader       = &gtfsbindings.FeedHeader{GtfsRealtimeVersion: &testVersion}
+	)
+
+	for _, test := range []struct {
```
srebhan, line 37: Please define the structs and assign to a variable. The iterate over the variable to separate the test-data from the testing logic. This eases reading.

plugins/inputs/gtfs/gtfs.go
```diff
@@ -0,0 +1,305 @@
+// Package gtfs is an input plugin for collecting GTFS-realtime data.
```
srebhan, line 1: Remove.

plugins/inputs/gtfs/gtfs.go
```diff
@@ -0,0 +1,305 @@
...
+	}
+
+	return req, nil
+}
+
+func init() {
+	inputs.Add("gtfs", func() telegraf.Input { return &GTFS{} })
+}
+
+var errNoSourceConfigured = errors.New("No source url configured; at least one of vehicle_positions_url, trip_updates_url or service_alerts_url must be populated")
```
srebhan, line 305: Please remove this global variable and move the error to the only place where you use it.

plugins/inputs/gtfs/gtfs.go
```diff
@@ -0,0 +1,305 @@
...
+			"odometer":      v.Position.Odometer,
+			"speed":         v.Position.Speed,
+			"vehicle_id":    v.Vehicle.Id,
+			"vehicle_label": v.Vehicle.Label,
+			"trip_id":       v.Trip.TripId,
+			"direction_id":  v.Trip.DirectionId,
+			"stop_id":       v.StopId,
+			"status":        v.CurrentStatus,
+			"congestion":    v.CongestionLevel,
+			"json":          b,
```
srebhan, line 168: Why do you add the whole JSON here? Can we please remove this field.

plugins/inputs/gtfs/gtfs.go
```diff
@@ -0,0 +1,305 @@
...
+		}
+
+		tags := map[string]string{
+			"route_id":     *v.Trip.RouteId,
+			"direction_id": fmt.Sprintf("%d", v.Trip.DirectionId),
```
srebhan, line 173: Use `strconv.FormatInt()` here.

plugins/inputs/gtfs/gtfs.go
```diff
@@ -0,0 +1,305 @@
...
+		}
+
+		acc.AddFields("position", fields, tags, time.Unix(int64(*v.Timestamp), 0))
```
srebhan, line 176: The metric name is too generic. Use `gtfs_position` at least.

plugins/inputs/gtfs/gtfs.go
```diff
@@ -0,0 +1,305 @@
...
+	}
+
+	if g.serviceAlertsReq != nil {
+		work.Go(func() error { return g.gatherServiceAlerts(acc, now) })
+	}
+
+	return work.Wait()
+}
+
+func (g *GTFS) gatherVehiclePositions(acc telegraf.Accumulator, t time.Time) error {
```
srebhan, line 138: I don't see where you use `t`. Please remove the parameter.

go.mod
gavincabbage: This is required, unfortunately, due to incompatibilities between different versions of the `protobuf` and `grpc` packages. I realize Telegraf currently uses `1.3.5` and not `1.3.3`, however, so I will try to use that instead to maintain the status quo. We're facing similar ecosystem challenges in other InfluxData projects and replacing `1.4.x` with `1.3.x` has side-stepped this issue in the past.

plugins/inputs/gtfs/gtfs.go
```diff
@@ -0,0 +1,305 @@
...
+	entities, err := g.gather(g.vehiclePositionsReq)
+	if err != nil {
+		return err
+	}
+
+	for _, entity := range entities {
+		if entity.Vehicle == nil {
+			continue
```
srebhan, line 146: Is this a common case or an error case. In case of the latter, please log a warning or an error here to make the user aware that metrics were dropped.

plugins/inputs/gtfs/README.md
gavincabbage: It does, but that syntax shouldn't be in the example config. I will change it and comment accordingly.

plugins/inputs/gtfs/gtfs.go
```diff
@@ -0,0 +1,305 @@
...
+			"direction_id": fmt.Sprintf("%d", v.Trip.DirectionId),
+		}
+
+		acc.AddFields("position", fields, tags, time.Unix(int64(*v.Timestamp), 0))
+	}
+
+	return nil
+}
+
+func (g *GTFS) gatherTripUpdates(acc telegraf.Accumulator, t time.Time) error {
```
srebhan, line 182: See comment about unused time above.

plugins/inputs/gtfs/gtfs.go
```diff
@@ -0,0 +1,305 @@
...
+			fields := map[string]interface{}{
+				"vehicle_id":    u.Vehicle.Id,
+				"vehicle_label": u.Vehicle.Label,
+				"trip_id":       u.Trip.TripId,
+				"stop_id":       update.StopId,
+				"stop_sequence": update.StopSequence,
+				"arrival":       update.Arrival.Time,
+				"departure":     update.Departure.Time,
+				"delay":         u.Delay,
+				"json":          b,
```
srebhan, line 209: Please remove the `json` field to avoid clutter similar to above.

plugins/inputs/gtfs/gtfs.go
```diff
@@ -0,0 +1,305 @@
...
+		}
+
+		fields := map[string]interface{}{
+			"latitude":      v.Position.Latitude,
+			"longitude":     v.Position.Longitude,
+			"bearing":       v.Position.Bearing,
+			"odometer":      v.Position.Odometer,
+			"speed":         v.Position.Speed,
+			"vehicle_id":    v.Vehicle.Id,
+			"vehicle_label": v.Vehicle.Label,
```
srebhan, line 162: Should these be fields or rather tags? If you might want to search for or group by this a tag is preferable.

plugins/inputs/gtfs/README.md
gavincabbage: I am mostly developing this plugin against Boston transit system data. I am not sure how other systems expose these endpoints. I will look into this, but unless I find is a clear standard, I would like to provide ultimate flexibility to accommodate any transit system using GTFS-realtime, even if they use different suffixes.

plugins/inputs/gtfs/gtfs.go
```diff
@@ -0,0 +1,305 @@
...
+
+		b, err := json.Marshal(u)
+		if err != nil {
+			return err
+		}
+
+		for _, update := range u.StopTimeUpdate {
+			fields := map[string]interface{}{
+				"vehicle_id":    u.Vehicle.Id,
+				"vehicle_label": u.Vehicle.Label,
```
srebhan, line 202: Similar to above.

plugins/inputs/gtfs/README.md
gavincabbage: Will do.

plugins/inputs/gtfs/README.md
gavincabbage: Will do.

plugins/inputs/gtfs/gtfs.go
```diff
@@ -0,0 +1,305 @@
...
+				"stop_sequence": update.StopSequence,
+				"arrival":       update.Arrival.Time,
+				"departure":     update.Departure.Time,
+				"delay":         u.Delay,
+				"json":          b,
+			}
+
+			tags := map[string]string{
+				"route_id":     *u.Trip.RouteId,
+				"direction_id": fmt.Sprintf("%d", u.Trip.DirectionId),
```
srebhan, line 214: `strconv.FormatInt()` see above.

plugins/inputs/gtfs/README.md
gavincabbage: Will do.

plugins/inputs/gtfs/README.md
gavincabbage: Will do.

plugins/inputs/gtfs/gtfs.go
```diff
@@ -0,0 +1,305 @@
...
+			}
+
+			acc.AddFields("update", fields, tags, time.Unix(int64(*u.Timestamp), 0))
```
srebhan, line 217: Metric name too generic, see above.

plugins/inputs/gtfs/gtfs.go
```diff
@@ -0,0 +1,305 @@
...
+		}
+	}
+
+	return nil
+}
+
+func (g *GTFS) gatherServiceAlerts(acc telegraf.Accumulator, t time.Time) error {
```
srebhan, line 224: `t` unused see above.

plugins/inputs/gtfs/README.md
gavincabbage: There is no public URL without an API key, but I can provide more details on how to get an example setup working against the Boston feed, including links to acquiring an API key (its a quick form request).

plugins/inputs/gtfs/gtfs.go
```diff
@@ -0,0 +1,305 @@
...
+
+func (g *GTFS) gatherTripUpdates(acc telegraf.Accumulator, t time.Time) error {
+	entities, err := g.gather(g.tripUpdatesReq)
+	if err != nil {
+		return err
+	}
+
+	for _, entity := range entities {
+		if entity.TripUpdate == nil {
+			continue
```
srebhan, line 190: Log in case this is an error. See above.

plugins/inputs/gtfs/gtfs.go
```diff
@@ -0,0 +1,305 @@
...
+
+func (g *GTFS) gatherServiceAlerts(acc telegraf.Accumulator, t time.Time) error {
+	entities, err := g.gather(g.serviceAlertsReq)
+	if err != nil {
+		return err
+	}
+
+	for _, entity := range entities {
+		if entity.Alert == nil {
+			continue
```
srebhan, line 232: Log in case of error. See above.

plugins/inputs/gtfs/gtfs.go
```diff
@@ -0,0 +1,305 @@
...
+		a := entity.Alert
+
+		b, err := json.Marshal(a)
+		if err != nil {
+			return err
+		}
+
+		fields := map[string]interface{}{
+			"cause": a.Cause.String(),
+			"json":  b,
```
srebhan, line 243: Remove to avoid clutter. See above.

plugins/inputs/gtfs/gtfs.go
```diff
@@ -0,0 +1,305 @@
...
+		}
+		a := entity.Alert
+
+		b, err := json.Marshal(a)
+		if err != nil {
+			return err
+		}
+
+		fields := map[string]interface{}{
+			"cause": a.Cause.String(),
```
srebhan, line 242: Is this the only relevant field?

plugins/inputs/gtfs/gtfs.go
```diff
@@ -0,0 +1,305 @@
...
+			"json":  b,
+		}
+
+		tags := map[string]string{
+			"severity": a.SeverityLevel.String(),
+		}
+
+		acc.AddFields("alert", fields, tags, time.Now())
```
srebhan, line 250: Metric name is too generic. See above.

plugins/inputs/gtfs/gtfs.go
gavincabbage: A package-level comment summarizing the package's purpose is [idiomatic Go](https://blog.golang.org/godoc). I disagree this should be removed.

plugins/inputs/gtfs/gfts_test.go
gavincabbage: Testing a package as an external consumer is an idiomatic pattern in Go. This ensures the subject package is exercised as a consumer would exercise it, and avoids common pitfalls in unit testing where the test becomes less meaningful because it has too much knowledge of and interaction with package internals. These tests do not need internal access to the package and are therefore not given it.

Further evidence that this is idiomatic and intended is the fact that a directly may only contain one package in Go, except for this specific case, where `package foo` is accompanied by a black-box `package foo_test`.

plugins/inputs/gtfs/gfts_test.go
gavincabbage: See above.

srebhan (commented):
The code looks quite clean and nice. Had some minor comments in the code. There are three major comments:

1. The changes to CHANGELOG.md are unrelated and should not be there. Please get rid of those changes.
2. The `replace` declarations in go.mod are strange. Do you really need them and if yes, why?
3. The definition of the three URLs are a bit strange to me. I think you need one base-url and maybe configurable paths to the entities. Otherwise, a user could specify multiple servers...

Please rebase your code and resolve the merge-conflict.

plugins/inputs/gtfs/gtfs.go
gavincabbage: The raw JSON contains many attributes not explicitly captured as InfluxDB fields or tags. Adding the raw JSON entity here provides flexibility in case it is necessary to inspect attributes of the entity that are not captured as first-class tags or fields.

plugins/inputs/gtfs/gtfs.go
gavincabbage: Will do.

plugins/inputs/gtfs/gtfs.go
gavincabbage: Will do.

plugins/inputs/gtfs/gtfs.go
gavincabbage: I believe this is an oversight and it should be used as the timestamp. Will double check and use or remove it.

plugins/inputs/gtfs/gtfs.go
gavincabbage: I'm not aware of how common this is in the wild, but it is also not an error. Given the GTFS-realtime specification, it is entirely possible for a transit service to expose all three feed types (position, alerts and updates) in a single feed. This check allows the plugin to work gracefully in that case.

plugins/inputs/gtfs/gtfs.go
gavincabbage: Will double check this one as well.

plugins/inputs/gtfs/gtfs.go
gavincabbage: These were added as fields to avoid cardinality issues given the number of Vehicle IDs and Labels is unbounded (and when considering bus services specifically, vehicles may server a number of routes and trips, exploding the cardinality of this measurement).

plugins/inputs/gtfs/gtfs.go
gavincabbage: I believe so. I added this measurement for completion. In reality, I expect vehicle positions to be the most useful and commonly used measurement provided by this plugin.

gavincabbage:
@srebhan Thank you for the thorough review! I expect I'll address your comments over the next few weeks. 

Much appreciated and happy holidays!

srebhan:
@gavincabbage cool plugin and nice code! Thanks and also have happy and safe holidays!

plugins/inputs/gtfs/README.md
srebhan: Well the issue is that you need to (potentially) provide three TLS configs in this case. I guess if you have different servers it would be easier to duplicate the plugin instance in the telegraf config...

plugins/inputs/gtfs/README.md
srebhan: It would be helpful for people playing with this to get a jump-start. Maybe a link to a tutorial would suffice.

plugins/inputs/gtfs/gfts_test.go
srebhan: Well almost all plugins in telegraf put the test into the same package with the main code, probably because there is no external customer...

plugins/inputs/gtfs/gtfs.go
srebhan: While I agree that this is standard in Go programs, this is a plugin to telegraf and I think most (if not all) other plugins do not have such a comment. I like to have things identical if there is no actual need to deviate from "the-usual" as this eases reviews over a large number of plugins. If this comment is really important to you, I will take to the core-devs on how to handle these. Just let me know.

plugins/inputs/gtfs/gtfs.go
srebhan: I disagree here. You force almost everybody to either swallow this (and live with a cluttered DB) or to drop the field requiring additional config effort. If you require this information, I prefer to print it in a debug output or provide a flag to enable this additional field.

plugins/inputs/gtfs/gtfs.go
srebhan: The background of the question is that if you _expect_ that field to be there, we don't want to silently drop the received data but rather make the user aware that we threw away data. In this case I do expect a info or even warn message.
If an the other hand receiving multiple types of messages is a common thing, then we should not bombard the user with messages in the log...
I rely on your estimate how likely it is we drop data erroneously here... But maybe it is worth a comment for the next guy looking at the code.

plugins/inputs/gtfs/gtfs.go
srebhan: Understood. Please mark this one as resolved. Maybe (optionally) you could also put a small comment in the code explaining this design decision...

plugins/inputs/gtfs/gtfs.go
srebhan: Also mark as resolved see above.

plugins/inputs/gtfs/gtfs.go
srebhan: Thanks, please mark as resolved.

srebhan:
@gavincabbage any news on this?

srebhan:
Any news here? Anybody else willing to work on the PR?

Issue #8485, opened 2020-11-30
New Intel PowerStat input plugin


p-zak:
## Feature Request
We have created a Telegraf PowerStat input plugin which enables monitoring of platform metrics (power, TDP) and per-CPU metrics like temperature, power and utilization.

Intel PowerStat plugin supports _Intel_ based platforms and assumes presence of _Linux_ based OS.

#### In the near future we will propose PR which adds new plugin we have created. It will be able to provide following metrics:
**Platform metrics:**
- thermal_design_power (Maximum Thermal Design Power (TDP) available for processor package [Watts])
- current_power_consumption (Current power consumption of processor package [Watts])
- current_dram_power_consumption (Current power consumption of processor package DRAM subsystem [Watts])

**Per-CPU metrics:**
- cpu_frequency (Current operational frequency of CPU Core [MHz])
- cpu_temperature (Current temperature of CPU Core [Celsius degrees])
- cpu_busy_cycles (CPU Core Busy cycles as a ratio of Cycles spent in C0 state residency to all cycles executed by CPU Core [%])
- cpu_c6_state_residency (Percentage of time that CPU Core spent in C6 Core residency state [%])
- cpu_c1_state_residency (Percentage of time that CPU Core spent in C1 Core residency state [%])
- cpu_busy_frequency (CPU Core Busy Frequency measured as frequency adjusted to CPU Core busy cycles [MHz])

### Use case:
Key source of platform telemetry is power domain that is beneficial for MANO/Monitoring&Analytics systems to take preventive/corrective actions based on platform busyness, CPU temperature, actual CPU utilization and power statistics. 
Main use cases are power saving and workload migration.

Leynakindle:
@p-zak Hello Zak, I encountered an issue while using the powerstat plugin. 
Specifically, when I run Telegraf with Docker Compose, I receive an error about loading MSR. However, I do not encounter this issue when running it directly on the host machine. Do you have any suggestions on how to resolve this? 
Attached is my compose.yaml file for reference. Host OS is CentOS Stream 9.
`services:
    telegraf:
        image: telegraf:1.26.2
        container_name: telegraf
        hostname: intel-101
        restart: always
        privileged: true
        volumes:
            - ./telegraf.conf:/etc/telegraf/telegraf.conf:ro
            - /sys/devices/virtual/powercap/intel-rapl:/sys/devices/virtual/powercap/intel-rapl:ro
[...]

PR #8488, opened 2020-11-30
New Intel PowerStat input plugin

MaciejMis:
### Required for all PRs:

- [x] Signed [CCLA](https://www.influxdata.com/legal/ccla/).

CCLA completed and signed by Intel Corporation.

## Feature Request
We have created a Telegraf PowerStat input plugin which enables monitoring of platform metrics (power, TDP) and per-CPU metrics like temperature, power and utilization.

Intel PowerStat plugin supports _Intel_ based platforms and assumes presence of _Linux_ based OS.

#### The plugin we have created is able to provide following metrics:
**Platform metrics:**
- thermal_design_power (Maximum Thermal Design Power (TDP) available for processor package [Watts])
- current_power_consumption (Current power consumption of processor package [Watts])
- current_dram_power_consumption (Current power consumption of processor package DRAM subsystem [Watts])

**Per-CPU metrics:**
- cpu_frequency (Current operational frequency of CPU Core [MHz])
- cpu_temperature (Current temperature of CPU Core [Celsius degrees])
- cpu_busy_cycles (CPU Core Busy cycles as a ratio of Cycles spent in C0 state residency to all cycles executed by CPU Core [%])
- cpu_c6_state_residency (Percentage of time that CPU Core spent in C6 Core residency state [%])
- cpu_c1_state_residency (Percentage of time that CPU Core spent in C1 Core residency state [%])
- cpu_busy_frequency (CPU Core Busy Frequency measured as frequency adjusted to CPU Core busy cycles [MHz])

### Use case:
Key source of platform telemetry is power domain that is beneficial for MANO/Monitoring&Analytics systems to take preventive/corrective actions based on platform busyness, CPU temperature, actual CPU utilization and power statistics.
Main use cases are power saving and workload migration.

Closes: https://github.com/influxdata/telegraf/issues/8485


plugins/inputs/intel_powerstat/README.md
```diff
@@ -0,0 +1,199 @@
...
+  ## telemetry will be exposed by Intel PowerStat plugin.
+  ## Supported options:
+  ## "cpu_frequency", "cpu_c1_state_residency", "cpu_c6_state_residency", "cpu_busy_cycles", "cpu_temperature", "cpu_busy_frequency"
+  # cpu_metrics = []
+```
+### Example: Configuration with no per-CPU telemetry
+This configuration allows getting global metrics (processor package specific), no per-CPU metrics are collected:
+```toml
+[[inputs.intel_powerstat]]
+  cpu_metrics = []
```
srebhan, line 26: This can also be removed, can't it?

plugins/inputs/intel_powerstat/README.md
```diff
@@ -0,0 +1,199 @@
...
+[[inputs.intel_powerstat]]
+  cpu_metrics = ["cpu_frequency", "cpu_c1_state_residency", "cpu_c6_state_residency", "cpu_busy_cycles", "cpu_temperature", "cpu_busy_frequency"]
+```
+
+### SW Dependencies:
+Plugin is based on Linux Kernel modules that expose specific metrics over `sysfs` or `devfs`.
+The following dependencies are expected by plugin:
+- _intel-rapl_ module which exposes Intel Runtime Power Limiting metrics over `sysfs` (`/sys/devices/virtual/powercap/intel-rapl`)
+Minimum kernel version required is 3.13 (to make sure RAPL driver is included in current distro),
+- _msr_ kernel module that provides access to processor model specific registers from user space,
```
srebhan, line 48: Can you also provide a vanilla kernel version here?

plugins/inputs/intel_powerstat/README.md
```diff
@@ -0,0 +1,199 @@
...
+Plugin is based on Linux Kernel modules that expose specific metrics over `sysfs` or `devfs`.
+The following dependencies are expected by plugin:
+- _intel-rapl_ module which exposes Intel Runtime Power Limiting metrics over `sysfs` (`/sys/devices/virtual/powercap/intel-rapl`)
+Minimum kernel version required is 3.13 (to make sure RAPL driver is included in current distro),
+- _msr_ kernel module that provides access to processor model specific registers from user space, 
+- _cpufreq_ kernel module - which exposes per-CPU Frequency over `sysfs`. Since kernel 3.4 the necessary modules are 
+loaded automatically and the recommended on demand governor is enabled by default.
+
+Please make sure that kernel modules are loaded and running. You might have to manually enable them by using `modprobe`.
+Plugin requires root access to read MSR files.
```
srebhan, line 53: This mean the whole telegraf needs to run as root right? Ouch! Maybe make it a bold hint!

plugins/inputs/intel_powerstat/intel_powerstat.go
```diff
@@ -0,0 +1,493 @@
...
+	cpuTemperature       = "cpu_temperature"
+	percentageMultiplier = 100
+)
+
+var (
+	allowedProcessorModelsForC1C6 = []string{"0x37", "0x4D", "0x5C", "0x5F", "0x7A", "0x4C", "0x86", "0x96", "0x9C",
+		"0x1A", "0x1E", "0x1F", "0x2E", "0x25", "0x2C", "0x2F", "0x2A", "0x2D", "0x3A", "0x3E", "0x4E", "0x5E", "0x55", "0x8E",
+		"0x9E", "0x6A", "0x6C", "0x7D", "0x7E", "0x9D", "0x3C", "0x3F", "0x45", "0x46", "0x3D", "0x47", "0x4F", "0x56",
+		"0x66", "0x57", "0x85", "0xA5", "0xA6", "0x8F", "0x8C", "0x8D"}
+)
```
srebhan, line 32: Maybe declare this in `verifyProcessor()` as this is the only user...

plugins/inputs/intel_powerstat/intel_powerstat.go
```diff
@@ -0,0 +1,493 @@
...
+
+	if !strings.Contains(firstCPU.Flags, "dts") {
+		p.coreCPUTemperature = false
+	}
+
+	return nil
+}
+
+// ParseCPUMetricsConfig enables specific metrics based on config values.
+func (p *PowerStat) ParseCPUMetricsConfig() {
```
srebhan, line 431: This should be a private function `parseCPUMetricsConfig()`

plugins/inputs/intel_powerstat/README.md
p-zak: Yes, that's right. But it shows explicitly what is the result of putting empty array in config. Is that bad?

plugins/inputs/intel_powerstat/intel_powerstat.go
```diff
@@ -0,0 +1,493 @@
...
+// Description returns a one-sentence description on the plugin.
+func (p *PowerStat) Description() string {
+	return `Intel PowerStat plugin enables monitoring of platform metrics (power, TDP) and Core metrics like temperature, power and utilization.`
+}
+
+// Gather takes in an accumulator and adds the metrics that the Input gathers.
+func (p *PowerStat) Gather(acc telegraf.Accumulator) error {
+	p.addGlobalMetrics(acc)
+
+	if p.perCoreMetrics && len(p.msr.GetCpuIDs()) > 0 {
```
srebhan, line 94: If no core metric is specified, `p.msr` can still be unititialized leading to a nil-pointer exception in `p.msr.GetCpuIDs()` or am I missing something?

plugins/inputs/intel_powerstat/intel_powerstat.go
```diff
@@ -0,0 +1,493 @@
...
+func (p *PowerStat) Init() error {
+	p.ParseCPUMetricsConfig()
+	err := p.verifyProcessor()
+	if err != nil {
+		return err
+	}
+	// Set perCoreMetrics flag only when there is at least one core metric enabled.
+	if p.c1CoreStateResidency || p.c6CoreStateResidency || p.processorBusyCycles || p.coreCPUTemperature ||
+		p.processorBusyFrequency || p.coreCPUFrequency {
+		p.perCoreMetrics = true
```
srebhan, line 64: This is in the later code just a guard against `p.msr` being `nil`. How about directly checking for `p.msr` being nil and getting rid of `p.perCoreMetrics` altogether. It was hard to me to read the later checks because I had to remember that `p.msr` cannot be nil if  `p.perCoreMetrics` is true. Checking directly for nil would have been easier to read.

plugins/inputs/intel_powerstat/intel_powerstat.go
srebhan, line 94:
```suggestion
	if p.msr !=nil && len(p.msr.GetCpuIDs()) > 0 {
```

plugins/inputs/intel_powerstat/intel_powerstat.go
```diff
@@ -0,0 +1,493 @@
...
+func (p *PowerStat) addPerCoreMetrics(acc telegraf.Accumulator) {
+	var wg sync.WaitGroup
+	cpuIDs := p.msr.GetCpuIDs()
+	wg.Add(len(cpuIDs))
+
+	for _, cpuID := range cpuIDs {
+		go p.addMetricsForSingleCore(cpuID, acc, &wg)
+	}
+
+	wg.Wait()
```
srebhan, line 113: How about moving this code to `Gather()`? There seems to be no further user...

plugins/inputs/intel_powerstat/README.md
p-zak: It's really hard to find exact version :D For sure _msr_ kernel module worked for 2.6 (~15 years ago), maybe even for earlier versions.
But you need at least 3.1 to run Go application for Linux (and this plugin aims only for Linux): https://github.com/golang/go/wiki/MinimumRequirements
So I would say: "If you can run Telegraf on Linux, you should have access to _msr_ kernel module" and I would not write anything here in README.

plugins/inputs/intel_powerstat/intel_powerstat.go
```diff
@@ -0,0 +1,493 @@
...
+	tags := map[string]string{
+		"package_id": cpu.PhysicalID,
+		"core_id":    cpu.CoreID,
+		"cpu_id":     cpu.CPUID,
+	}
+
+	fields := map[string]interface{}{
+		"name":  "cpu_frequency",
+		"unit":  "MHz",
+		"value": services.RoundFloatToNearestTwoDecimalPlaces(services.ConvertKiloHertzToMegaHertz(frequency)),
```
srebhan, line 175: I think this is not so nice as querying the data will be hard. Looking up `cpu_frequencies` might be costly as you can only constrain by field value which means you have to walk all rows in the table. How about either making the "name" and "unit" a tag **or**
```suggestion
		"cpu_frequency_mhz": services.RoundFloatToNearestTwoDecimalPlaces(services.ConvertKiloHertzToMegaHertz(frequency)),
```
and maybe unit a tag?

srebhan: Same for all other instances.

plugins/inputs/intel_powerstat/intel_powerstat.go
```diff
@@ -0,0 +1,493 @@
...
+	// Since counter collection is not atomic it may happen that sum of C0, C1, C3, C6 and C7
+	// is bigger value than TSC, in such case C1 residency shall be set to 0.
+	if timestampDelta > (p.msr.GetMperfDelta(cpuID) + p.msr.GetC3Delta(cpuID) +
+		p.msr.GetC6Delta(cpuID) + p.msr.GetC7Delta(cpuID)) {
+		c1 := timestampDelta - p.msr.GetMperfDelta(cpuID) - p.msr.GetC3Delta(cpuID) -
+			p.msr.GetC6Delta(cpuID) - p.msr.GetC7Delta(cpuID)
+
+		c1Value = services.RoundFloatToNearestTwoDecimalPlaces(percentageMultiplier * float64(c1) /
+			float64(timestampDelta))
+	}
```
srebhan, line 264: Wouldn't the clipping
```suggestion
	c1 := timestampDelta - p.msr.GetMperfDelta(cpuID) - p.msr.GetC3Delta(cpuID) -
		p.msr.GetC6Delta(cpuID) - p.msr.GetC7Delta(cpuID)
	if c1 < 0.0 {
		c1 = 0.0
	}
	c1Value = services.RoundFloatToNearestTwoDecimalPlaces(percentageMultiplier * float64(c1) /
		float64(timestampDelta))
```
be clearer. As a bonus we avoid calling all the functions twice.

plugins/inputs/intel_powerstat/intel_powerstat.go
```diff
@@ -0,0 +1,493 @@
...
+	tsc := services.ConvertProcessorCyclesToHertz(p.msr.GetTimestampDelta(cpuID))
+	timeNow := time.Now().UnixNano()
+	interval := services.ConvertNanoSecondsToSeconds(timeNow - p.msr.GetReadDate(cpuID))
+	p.msr.SetReadDate(cpuID, timeNow)
+
+	if p.skipFirstIteration {
+		return
+	}
+
+	busyMhzValue := services.RoundFloatToNearestTwoDecimalPlaces(tsc * aperfMperf / interval)
```
srebhan, line 297: Maybe check `interval` for being zero.

plugins/inputs/intel_powerstat/intel_powerstat.go
```diff
@@ -0,0 +1,493 @@
...
+
+	for _, socketID := range p.rapl.GetSocketIDs() {
+		err := p.rapl.RetrieveAndCalculateData(socketID)
+		if err != nil {
+			// In case of an error skip calculating metrics for this socket
+			p.Log.Errorf("error fetching rapl data for socket %s, err: %v", socketID, err)
+			continue
+		}
+		p.addThermalDesignPowerMetric(socketID, acc)
+		if !p.skipFirstIteration {
```
srebhan, line 326: How about
```suggestion
		if p.skipFirstIteration {
			continue
		}
```

plugins/inputs/intel_powerstat/intel_powerstat.go
```diff
@@ -0,0 +1,493 @@
...
+		skipFirstIteration:     true,
+		fs:                     fs,
+	}
+
+	return p
+}
+
+func init() {
+	inputs.Add("intel_powerstat", func() telegraf.Input {
+		return NewPowerStat(services.NewFileService())
```
srebhan, line 491: Can't you just do
```suggestion
		return &PowerStat{}
```
and set `fs` in `Init()`?

plugins/inputs/intel_powerstat/intel_powerstat.go
p-zak: It can be done of course, but it was extracted from `Gather()` by design - to have it as small as possible.

srebhan (commented):
Hey @MaciejMis, cool stuff. I've some comment nevertheless...

1. I think the metric output format is costly on the query side. If you e.g. want the "cpu_frequency" you need to constrain the **field** value if "name" to "cpu_frequency". That means you need to touch all rows of the DB. Making it a field i.e. field: "cpu_frequency_mhz"=value would be much more efficient.
1. The whole "services" thing feels like a standalone golang package. Can maybe Intel host that as a package (just the "services", "data", "mocks" thing) as a golang package that we can use here? I'm asking because that's a lot of code not solely bound with telegraf (e.g. someone could use it in a different golang project)...
1. The `skipfirst` feels strange. While I couldn't come up with an ad-hoc replacement it feels like circumventing an `Init()` case...

plugins/inputs/intel_powerstat/intel_powerstat.go
p-zak: If `fs` is set in `Init()` it would be very difficult to test it. So it was done on purpose.

plugins/inputs/intel_powerstat/README.md
p-zak: Yep, it should be bolded.

plugins/inputs/intel_powerstat/README.md
srebhan: No it's ok, but it suggests you need to specify it to be empty. Fine if you want to keep it if you want to make it explicit.

plugins/inputs/intel_powerstat/README.md
srebhan: Would be fine if you put "3.1 maybe earlier". People < 3.1 then know it might or might not work. It's about the "no worries it will work anyway" portion of ppl. It's just you put a version on any other module... I'd like to see the "first-to-be-sure" version here if it exists.

plugins/inputs/intel_powerstat/intel_powerstat.go
srebhan: It's a matter of taste. It was just a suggestion. Proceed as you like.

plugins/inputs/intel_powerstat/README.md
p-zak: It will be exactly the same behavior if we do not specify this empty array - no per-CPU metrics gathered. 
As guided here: https://github.com/influxdata/telegraf/wiki/Review#configuration: `have one # for defaults, which should always match the default value of the plugin` :)

p-zak:
> Hey @MaciejMis, cool stuff. I've some comment nevertheless...
> 
> 1. I think the metric output format is costly on the query side. If you e.g. want the "cpu_frequency" you need to constrain the **field** value if "name" to "cpu_frequency". That means you need to touch all rows of the DB. Making it a field i.e. field: "cpu_frequency_mhz"=value would be much more efficient.
> 2. The whole "services" thing feels like a standalone golang package. Can maybe Intel host that as a package (just the "services", "data", "mocks" thing) as a golang package that we can use here? I'm asking because that's a lot of code not solely bound with telegraf (e.g. someone could use it in a different golang project)...
> 3. The `skipfirst` feels strange. While I couldn't come up with an ad-hoc replacement it feels like circumventing an `Init()` case...

1. Good point!
2. Hosting "services" thing somewhere else is not an option right now. There are just helper functions extracted from main file to read and calculate values and crafted to be used by this plugin. 
3. `skipfirst` is just to omit few metrics which are calculated deltas. In first iteration they doesn't make any sense (need at least two iterations of `Gather` to calculate something).

plugins/inputs/intel_powerstat/services/msr.go
```diff
@@ -0,0 +1,297 @@
...
+	GetThrottleTemperature(CPU string) uint64
+	GetTemperature(CPU string) uint64
+	GetTimestampDelta(CPU string) uint64
+	GetMperfDelta(CPU string) uint64
+	GetAperfDelta(CPU string) uint64
+	GetC6Delta(CPU string) uint64
+	GetC3Delta(CPU string) uint64
+	GetC7Delta(CPU string) uint64
+	GetReadDate(CPU string) int64
+	SetReadDate(CPU string, date int64)
```
srebhan, line 48: Get rid of the getter/setter and use the exported fields directly.

plugins/inputs/intel_powerstat/services/msr.go
```diff
@@ -0,0 +1,297 @@
...
+}
+
+// MsrServiceImpl is implementation of MsrService.
+type MsrServiceImpl struct {
+	cpuCoresData map[string]*data.MsrData
+	msrOffsets   map[string]int64
+	fs           FileService
+	log          telegraf.Logger
```
srebhan, line 56: You should not rely on telegraf in this abstraction layer IMO. Just return the error back to telegraf and let it handle the stuff.

plugins/inputs/intel_powerstat/services/msr.go
```diff
@@ -0,0 +1,297 @@
...
+
+// GetReadDate returns last read date.
+func (m *MsrServiceImpl) GetReadDate(cpuID string) int64 {
+	return m.cpuCoresData[cpuID].ReadDate
+}
+
+// SetReadDate sets last read date.
+func (m *MsrServiceImpl) SetReadDate(cpuID string, date int64) {
+	m.cpuCoresData[cpuID].ReadDate = date
+}
```
srebhan, line 118: I don't think we need this. The fields are public anyway and we are the only internal user.

plugins/inputs/intel_powerstat/services/msr.go
```diff
@@ -0,0 +1,297 @@
...
+
+				err := m.readValueFromFileAtOffset(ctx, ch, reader, m.msrOffsets[off])
+				if err != nil {
+					return fmt.Errorf("error reading MSR file, err: %v", err)
+				}
+
+				return nil
+			})
+		}(offset, channel)
+	}
```
srebhan, line 171: What is the rational behind parallelizing this? All data is read from the same file right?

plugins/inputs/intel_powerstat/services/msr.go
```diff
@@ -0,0 +1,297 @@
...
+			C7Delta:      0,
+		}
+	}
+
+	return nil
+}
+
+func (m *MsrServiceImpl) getMsrOffsets() map[string]int64 {
+	return m.msrOffsets
+}
```
srebhan, line 258: This is superfluous.

plugins/inputs/intel_powerstat/services/msr.go
```diff
@@ -0,0 +1,297 @@
...
+	m.msrOffsets = map[string]int64{
+		c3StateResidencyLocation:           m.parseStringHexToInt(c3StateResidencyLocation),
+		c6StateResidencyLocation:           m.parseStringHexToInt(c6StateResidencyLocation),
+		c7StateResidencyLocation:           m.parseStringHexToInt(c7StateResidencyLocation),
+		maximumFrequencyClockCountLocation: m.parseStringHexToInt(maximumFrequencyClockCountLocation),
+		actualFrequencyClockCountLocation:  m.parseStringHexToInt(actualFrequencyClockCountLocation),
+		TimestampCounterLocation:           m.parseStringHexToInt(TimestampCounterLocation),
+		throttleTemperatureLocation:        m.parseStringHexToInt(throttleTemperatureLocation),
+		temperatureLocation:                m.parseStringHexToInt(temperatureLocation),
+	}
```
srebhan, line 270: Don't. Just specify the offsets here as int instead of doing the indirections! This
```suggestion
	m.msrOffsets = map[string]int64{
		c3StateResidencyLocation:           0x3FC,
		c6StateResidencyLocation:           0x3FD,
		c7StateResidencyLocation:           0x3FE,
		maximumFrequencyClockCountLocation: 0xE7,
		actualFrequencyClockCountLocation:  0xE8,
		TimestampCounterLocation:           0x10,
		throttleTemperatureLocation:        0x1A2,
		temperatureLocation:                0x19C,
	}
```
is much cleaner and easier to read.

plugins/inputs/intel_powerstat/services/msr.go
```diff
@@ -0,0 +1,297 @@
...
+
+func (m *MsrServiceImpl) parseStringHexToInt(s string) int64 {
+	parsedInt, err := strconv.ParseInt(s, 0, 64)
+	if err != nil {
+		m.log.Errorf("error on parsing offset %s, err: %v", s, err)
+		return 0
+	}
+
+	return parsedInt
+}
```
srebhan, line 281: With the change above this can be scraped.

plugins/inputs/intel_powerstat/services/msr.go
```diff
@@ -0,0 +1,297 @@
...
+
+// NewMsrServiceWithFs returns new RaplServiceImpl struct with given FileService.
+func NewMsrServiceWithFs(logger telegraf.Logger, fs FileService) *MsrServiceImpl {
+	msrService := &MsrServiceImpl{
+		fs:  fs,
+		log: logger,
+	}
+	err := msrService.setCPUCores()
```
srebhan, line 289: The only location where this call can fail is if the well-formatted patterns passed to `filepath.Glob()` are broken. We then have a much deeper problem. So I'd rather return an error here.

plugins/inputs/intel_powerstat/services/msr.go
zak-pawel: Yes, we read 8 bytes at different offsets from the same "character device file" - MSR. And it is faster to do that in parallel.

plugins/inputs/intel_powerstat/services/rapl.go
```diff
@@ -0,0 +1,274 @@
...
+// GetCurrentPackagePowerConsumption returns package power consumption for socket.
+func (r *RaplServiceImpl) GetCurrentPackagePowerConsumption(socketID string) float64 {
+	return r.data[socketID].SocketCurrentEnergy
+}
+
+// GetCurrentDramPowerConsumption returns dram package power consumption for socket.
+func (r *RaplServiceImpl) GetCurrentDramPowerConsumption(socketID string) float64 {
+	return r.data[socketID].DramCurrentEnergy
+}
+
```
srebhan, line 74: I'd rather get rid of this.

plugins/inputs/intel_powerstat/services/rapl.go
```diff
@@ -0,0 +1,274 @@
...
+	defer maxEnergyFile.Close()
+	maxEnergy, _, err := r.fs.ReadFileToFloat64(maxEnergyFile)
+
+	return maxEnergy, err
+}
+
+// getDramFolders returns Dram folder names from RAPL.
+func (r *RaplServiceImpl) getDramFolders() map[string]string {
+	return r.dramFolders
+}
```
srebhan, line 141: Use field directly. It's an internal usage anyway.

plugins/inputs/intel_powerstat/services/rapl.go
```diff
@@ -0,0 +1,274 @@
...
+	GetSocketIDs() []string
+	GetCurrentPackagePowerConsumption(socketID string) float64
+	GetCurrentDramPowerConsumption(socketID string) float64
+	GetDramMaxEnergyRangeUj(socketID string) (float64, error)
+	GetMaxEnergyRangeUj(socketID string) (float64, error)
+}
+
+// RaplServiceImpl is implementation of RaplService.
+type RaplServiceImpl struct {
+	log          telegraf.Logger
```
srebhan, line 41: Again I think you should not rely on telegraf here. Remove the logger and return errors where needed.

plugins/inputs/intel_powerstat/services/rapl.go
```diff
@@ -0,0 +1,274 @@
...
+
+// getDramFolders returns Dram folder names from RAPL.
+func (r *RaplServiceImpl) getDramFolders() map[string]string {
+	return r.dramFolders
+}
+
+func (r *RaplServiceImpl) prepareData() {
+	intelRaplPrefix := "intel-rapl:"
+	intelRapl := fmt.Sprintf("%s%s", intelRaplPrefix, "[0-9]*")
+	raplPaths, err := r.fs.GetStringsMatchingPatternOnPath(fmt.Sprintf("%s/%s", IntelRaplPath, intelRapl))
```
srebhan, line 146: There should never be an error as the underlying `filepath.Glob()` only returns an error on malformed patterns. As you specify the pattern this should never happen. So return an error if there is a problem as probably more things are broken then.

plugins/inputs/intel_powerstat/services/rapl.go
```diff
@@ -0,0 +1,274 @@
...
+	if err != nil {
+		if r.isFirstError {
+			r.isFirstError = false
+			r.log.Errorf("error while preparing RAPL data: %v", err)
+		} else {
+			r.log.Debugf("error while preparing RAPL data: %v", err)
+		}
```
srebhan, line 153: Simply return the error.

plugins/inputs/intel_powerstat/services/rapl.go
```diff
@@ -0,0 +1,274 @@
...
+
+		r.data = make(map[string]*data.RaplData)
+		return
+	}
+
+	r.isFirstError = true
```
srebhan, line 159: I don't see a use of `isFirstError`, you only _write_ this variable but don't read it as far as I can see except for deciding if you want to log an error or a debug message above.

plugins/inputs/intel_powerstat/services/rapl.go
```diff
@@ -0,0 +1,274 @@
...
+
+func (r *RaplServiceImpl) findDramFolders() {
+	intelRaplPrefix := "intel-rapl:"
+	intelRaplDram := fmt.Sprintf("%s%s", intelRaplPrefix, "[0-9]*[0-9]*")
+	// Clean existing map
+	r.dramFolders = make(map[string]string)
+
+	for socketID := range r.data {
+		path := fmt.Sprintf(IntelRaplSocketPartialPath, IntelRaplPath, socketID)
+		pathsToRaplFolders, err := r.fs.GetStringsMatchingPatternOnPath(fmt.Sprintf("%s/%s", path, intelRaplDram))
```
srebhan, line 183: Again this should not fail, if it does, there is a big problem anyway.

plugins/inputs/intel_powerstat/services/rapl.go
```diff
@@ -0,0 +1,274 @@
...
+
+func (r *RaplServiceImpl) findDramFolder(raplFolders []string, socketID string) {
+	for _, raplFolder := range raplFolders {
+		potentialDramPath := fmt.Sprintf(IntelRaplDramPartialPath, IntelRaplPath, socketID, raplFolder)
+		nameFilePath := fmt.Sprintf(IntelRaplDramNamePartialPath, potentialDramPath)
+		read, err := r.fs.ReadFile(nameFilePath)
+		if err != nil {
+			r.log.Errorf("error reading file on path: %s, err: %v", nameFilePath, err)
+			continue
+		}
```
srebhan, line 248: Should we really continue here of fail? Is there a real use-case (hot-plugging CPU?) where we want to go on and accept the error?

plugins/inputs/intel_powerstat/services/rapl.go
zak-pawel: Yep, exactly. If you don't have proper directories for reading RAPL data (when it is disabled but plugin will gather other metrics anyway), it will be logged as an error for the first time and on debug level afterwards to make default log nice and clean. If you turn on RAPL when Telegraf is on, it will eventually start gathering metrics from RAPL subsystem.

plugins/inputs/intel_powerstat/intel_powerstat.go
srebhan: Oh and make them integers right from the start as you don't use them as string anywhere. This allows to get rid of ` services.ConvertHexArrayToIntegerArray()`.

srebhan (changes requested):
Nice code, but I have the feeling it has too many indirections. Please see my comments.

plugins/inputs/intel_powerstat/services/rapl.go
p-zak: Sure: hot-plugging RAPL system (when RAPL's paths appear and disappear in filesystem).
```
sudo modprobe intel_rapl
sudo modprobe -r intel_rapl
```

plugins/inputs/intel_powerstat/README.md
reimda: On my machine everything in /sys/devices/virtual/powercap/intel-rapl is world readable.  I think this means the plugin would work without root access.  I'd be surprised if other linux distributions had different permissions.

A quick search of similar tools also seems to imply root access isn't needed. e.g. http://web.eece.maine.edu/~vweaver/projects/rapl/

I'd strongly prefer not to encourage users to run telegraf as root, especially if it's not necessary.  What exactly needs root access?  Does the plugin write to files in sysfs to configure RAPL?

zak-pawel: @reimda 
You're right - to read `/sys/devices/virtual/powercap/intel-rapl` you don't need root access (probably on every distro).

But here is written: `Plugin requires root access to read MSR files.` which are located here: `/dev/cpu/%s/msr` (`%s` for logical cpu core). And this is the character device file to which you need the root access (no workarounds for that :()

We will try change README here to show what can be read without root and for which metrics root is required.

(and no, plugin does not write to any file)

plugins/inputs/intel_powerstat/README.md
p-zak: Example without array also added to `README.md`

plugins/inputs/intel_powerstat/README.md
p-zak: Section restructured, "first-to-be-sure" version added (globally, for types of metrics)

plugins/inputs/intel_powerstat/README.md
p-zak: @srebhan corrected.
@reimda I found out that on some environments (OS/conf) meaningful files under `/sys/devices/virtual/powercap/intel-rapl` (text files, not directories) need root access and on some they don't. Section restructured.

plugins/inputs/intel_powerstat/intel_powerstat.go
p-zak: Great point, fixed.

plugins/inputs/intel_powerstat/intel_powerstat.go
p-zak: Fixed.

plugins/inputs/intel_powerstat/intel_powerstat.go
p-zak: You're right. Corrected.

plugins/inputs/intel_powerstat/intel_powerstat.go
p-zak: Done.

plugins/inputs/intel_powerstat/intel_powerstat.go
p-zak: Thanks.

plugins/inputs/intel_powerstat/intel_powerstat.go
p-zak: Excellent point. Fixed.

plugins/inputs/intel_powerstat/intel_powerstat.go
p-zak: Done. And also corrected to avoid overflows.

plugins/inputs/intel_powerstat/intel_powerstat.go
p-zak: Done.

plugins/inputs/intel_powerstat/intel_powerstat.go
p-zak: Sure, changed.

plugins/inputs/intel_powerstat/intel_powerstat.go
p-zak: Would like to leave it as it is right now.

plugins/inputs/intel_powerstat/services/msr.go
p-zak: Done.

plugins/inputs/intel_powerstat/services/msr.go
p-zak: This is integral part of Telegraf plugin. Abstraction layer simplified.

plugins/inputs/intel_powerstat/services/msr.go
p-zak: You're right. Removed.

plugins/inputs/intel_powerstat/services/msr.go
p-zak: Sure, removed.

plugins/inputs/intel_powerstat/services/msr.go
p-zak: 100% agree.

plugins/inputs/intel_powerstat/services/msr.go
p-zak: Scraped.

plugins/inputs/intel_powerstat/services/rapl.go
p-zak: Removed.

plugins/inputs/intel_powerstat/services/rapl.go
p-zak: This is integral part of Telegraf plugin. Abstraction layer simplified.

plugins/inputs/intel_powerstat/services/rapl.go
p-zak: Simplified.

plugins/inputs/intel_powerstat/services/msr.go
p-zak: I understand, this is almost impossible to happen that this pattern is not well-formatted but for security reasons and for the sake of handling every error we would like to leave it as it is.

plugins/inputs/intel_powerstat/services/rapl.go
p-zak: Ups, you're right.

plugins/inputs/intel_powerstat/services/rapl.go
p-zak: I understand, this is almost impossible to happen that this pattern is not well-formatted but for security reasons and for the sake of handling every error we would like to leave it as it is.

plugins/inputs/intel_powerstat/services/rapl.go
p-zak: I understand, this is almost impossible to happen that this pattern is not well-formatted but for security reasons and for the sake of handling every error we would like to leave it as it is.
Simplified, `isFirstError` removed.

plugins/inputs/intel_powerstat/services/rapl.go
p-zak: I understand, this is almost impossible to happen that this pattern is not well-formatted but for security reasons and for the sake of handling every error we would like to leave it as it is.

plugins/inputs/intel_powerstat/file.go
```diff
@@ -0,0 +1,154 @@
...
+	currentInfo := &cpuInfo{}
+
+	for scanner.Scan() {
+		line := scanner.Text()
+
+		processorRes := processorRegexp.FindStringSubmatch(line)
+		if len(processorRes) > 1 {
+			currentInfo = &cpuInfo{
+				cpuID: processorRes[1],
+			}
```
srebhan, line 60: Shouldn't this be 
```suggestion
			currentInfo.cpuID = processorRes[1]
```

plugins/inputs/intel_powerstat/file.go
```diff
@@ -0,0 +1,154 @@
...
+
+	scanner := bufio.NewScanner(cpuInfoFile)
+
+	processorRegexp := regexp.MustCompile(`^processor\t+:\s([0-9]+)\n*$`)
+	physicalIDRegexp := regexp.MustCompile(`^physical id\t+:\s([0-9]+)\n*$`)
+	coreIDRegexp := regexp.MustCompile(`^core id\t+:\s([0-9]+)\n*$`)
+	vendorIDRegexp := regexp.MustCompile(`^vendor_id\t+:\s([a-zA-Z]+)\n*$`)
+	cpuFamilyRegexp := regexp.MustCompile(`^cpu\sfamily\t+:\s([0-9]+)\n*$`)
+	modelRegexp := regexp.MustCompile(`^model\t+:\s([0-9]+)\n*$`)
+	flagsRegexp := regexp.MustCompile(`^flags\t+:\s(.+)\n*$`)
```
srebhan, line 48: This could be moved to `newFileService()` to save some cycles but it's cosmetics only.

srebhan (approved):
LGTM. I really enjoy reading the code!

One thing worth mentioning is the atomic organization of the metrics. There are multiple metrics emitted with the same name but different fields. One could argue to collect those fields before emitting, but on the other hand the current structure keeps the code clean.

plugins/inputs/intel_powerstat/file.go
p-zak: Agreed that this needs to created for every new logical cpu found in `/proc/cpuinfo`

plugins/inputs/intel_powerstat/file.go
p-zak: It it is only cosmetic, maybe it can stay as it is? ;)

plugins/inputs/intel_powerstat/README.md
```diff
@@ -0,0 +1,206 @@
...
+
+Please make sure that kernel modules are loaded and running. You might have to manually enable them by using `modprobe`.
+Exact commands to be executed are:
+```
+sudo modprobe cpufreq-stats
+sudo modprobe msr
+sudo modprobe intel_rapl
+```
+
+**Telegraf with Intel PowerStat plugin enabled requires root access to read model specific registers (MSRs)**
```
reimda, line 66: I don't think root is required even for MSRs.
```suggestion
**Telegraf with Intel PowerStat plugin enabled may require root access to read model specific registers (MSRs)** 
```

Issue #7548, opened 2020-05-19
Add plugin that implements a simple file based queue system


danielnelson:
### Proposal:

Many teams are using files and directories to implement ETL jobs.

It would be useful to have a plugin that parses a file once and when it reaches the EOF moves the file to a target success or failure directory.

Like the `tail` plugin, this plugin would process the file once and would support durable delivery with tracking metrics.  Unlike `tail` it would stop processing when reaching EOF.

The plugin should support transparent handling of gzipped files, which would great to support in both `file` and `tail` as well.

The plugin should contain `internal` metrics for the number of files processed.

### Current behavior:

This cannot be accomplished reliable, but some aspects can be faked by using a outside process to stage and move files.

### Use case: 

See https://github.com/influxdata/telegraf/issues/5781

sjwang90:
This is a common use case for major Industrial IoT setups that have decides that collect data in remote areas with poor connectivity and end up backlogging a lot of data and transmitting in compressed formats (gzip)

SimonMcN:
proposal:
add flag rename_after_process = 'true'
add flag rename_prefix = 'processed'
add flag rename_suffix = ''
then after the file is read, if  rename_after_process == 'true' then rename according to rename_prefix and rename_suffix

Would that work ?

sjwang90:
Requirements for large scale IIoT environments: @boris377 @suseelapattamatta
- Translation of the file lines into line protocol (our origin file doesn’t match line protocol).
- Potential large batches of incoming files of 1000+ (not uncommon)
- Continuous processing (directory monitoring portion) – We get 1-3 files per minute per asset except when the above happens
- Data integrity (no lost metrics)

PR #8495, opened 2020-12-01
Adding dirmon

suseelapattamatta:
### Required for all PRs:


sjwang90:
Related to #7548

srebhan:
Hey @suseelapattamatta! What are you trying to do with the plugin? Could you detail a bit on what is missing `tail` or `file` to achieve your goal!?

sjwang90:
Hey @srebhan, take a look at #7548 if you wanna read more. Essentially file directory monitoring plugin.

plugins/inputs/dirmon/README.md
```diff
@@ -0,0 +1,51 @@
+# tail Input Plugin
```
sjwang90, line 1: Looks like the README wasn't really edited. I gave feedback to @reimda that we may want a clearer name for this plugin besides dirmon. I suggested simply `inputs.directory`.

suseelapattamatta: We had a scenario where backlog files are processed when dirmon is restarted after  a network outage or similar incident that crashes the application. Upon successfully processing the backlog, dirmon stops scanning the directory for real time events. Yesterday, I resolved this issue. I am adding the code in dirmon.go.  You can find it at 
func (ddo DirDefObject) FileProcessor(id int).

The plugin is named "dirmon" because it monitors the directory where we expect files to be available for processing. These files are further processed to extract the metrics etc.

sjwang90:
From @suseelapattamatta: "Dirmon (directory monitor) scans folders and returns total number of files to be processed (bunch of config settings to determine what files). These are all csv.gz files of varying length and size, all pretty consistent. It then processes this file in using our specific CSV format to make line protocols to pass in. This is all good and works really well, **however the problem becomes that now dirmon runs so efficiently on the new application, that it is processing more files than telegraf can possibly keep up with writing to influx.**

Dirmon now, will process so quickly, that it puts the metrics to telegraf who cannot keep up. The buffer is full or it won’t finish writing within the flush interval, we actually cannot even be sure if dirmon is handling all the metrics properly because its doing so many. Right now it will do as many files as the processor will let it handle which thankfully on these servers is pretty limited. Since the plugin cannot determine how many metrics exist PER file, we have to be smart in how we configure this choosing to use a lower limit as to not hit the telegraf buffer limit."


Outstanding questions for the PR:
- How much does the flush time write failures matter (if at all)?
- What is the suggested max/optimal batch size? 5k right?
- What other way can we “slow” the total point ingestion coming? It will always be our worst case scenario that 1000+ files just hit at any one point in time.
- What is the correlation between flush and collect? Obviously Dirmon is getting way way more than it can handle, but we are unclear what we should be adjusting based on the wiki descriptions.
- Should we ever hit the max buffer size? If no, what are the downsides of hitting it?

sjwang90:
@suseelapattamatta Our team is continuing this work on #8751 so I'm going to close this PR. I'll reach out to your team via e-mail when we we're ready for you to test.

PR #8496, opened 2020-12-02
Increase build verbosity with go mod download -x

lesam:
Initial build took several minutes on my machine, and it looked like it
was hanging until I added the -x flag.

### Required for all PRs:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,7 +83,7 @@ help:
 
 .PHONY: deps
 deps:
-	go mod download
+	go mod download -x
 
 .PHONY: telegraf
 telegraf:


p-zak:
`The -x flag causes download to print the commands download executes.` 
I wonder how this helped you.

But what is more important: this flag was added in Go `1.14`. And seems that CI checks for Mac/Darwin uses `1.13.6`.

Similar problem (with Go in `1.13.6`) was encounter here: https://github.com/influxdata/telegraf/pull/8391

LeSam:
> `The -x flag causes download to print the commands download executes.`
> I wonder how this helped you.
> 
> But what is more important: this flag was added in Go `1.14`. And seems that CI checks for Mac/Darwin uses `1.13.6`.
> 
> Similar problem (with Go in `1.13.6`) was encounter here: #8391

Darn.

The `-x` flag helps because I can know that `go mod download` is actually doing something instead of just wondering why it is hanging for minutes - if there are no modules to download it's pretty quiet, so I think it would be an improvement for people new to the project.

How long will 1.13 be supported? I can withdraw the PR and maybe re-raise when 1.13 support is dropped.

LeSam:
Actually, judging by that comment looks like the circleci change for macdeps to 1.14 is coming soon: https://github.com/influxdata/telegraf/pull/8391#issuecomment-725688523 . So I'll leave this up.

LeSam:
Cherry-picked the commit from the other PR to bump the go version in macdeps

ssoroka:
That's probably the last holdout from Go 1.13, and I suspect we can merge this without further problems

ssoroka (approved):

reimda:
Hi Sam, thanks for the PR.

I don't want to merge the 'brew update' change because it adds 4-5 minutes to the build.  I think circleci provides a newer mac image than we're using.  If we can switch to a new image instead of updating the old one, we should get golang 1.14+ without increased build time.

I'll look into switching to a new image today so we can merge this and #8391.

.circleci/config.yml
```diff
@@ -12,7 +12,7 @@ defaults:
       - image: 'quay.io/influxdb/telegraf-ci:1.15.5'
   mac: &mac
     macos:
-      xcode: 11.3.1
+      xcode: 12.2.0
```
lesam, line 15: Playing around with circleci to make sure I understand it - this bumps the macos version (see commit message).

reimda:
I updated the golang version for mac ci in #8516.  Could you remove changes to .circleci/config.yml and rebase this PR?  Thanks!

lesam:
> I updated the golang version for mac ci in #8516. Could you remove changes to .circleci/config.yml and rebase this PR? Thanks!

Done

lesam:
Rebased

lesam:
Rebased again

ssoroka (approved):

PR #8498, opened 2020-12-02
fixed network test 

helenosheaa:
Fixed network test by adding a mock client for failing test coming back with connection failure instead of dns failure

### DIFF
--- a/plugins/inputs/http_response/http_response.go
+++ b/plugins/inputs/http_response/http_response.go
@@ -52,7 +52,11 @@ type HTTPResponse struct {
 	Log telegraf.Logger
 
 	compiledStringMatch *regexp.Regexp
-	client              *http.Client
+	client              httpClient
+}
+
+type httpClient interface {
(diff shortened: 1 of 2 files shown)


srebhan:
The problem **only** occurs if you set a proxy in your environment. Unsetting it will make the test pass. The reason behind is that you get a "Bad gateway" error message from your proxy (Squid at least) instead of the DNS error expected. 
So I'd rather like to ask you to implement a way for the plugin to ignore the environment proxy, e.g. by setting the `http_proxy` option to `"-"`. This is helpful in production environments and allows us to force the tests into not using any proxy a.k.a. reproducible environment. :-)

Are you willing to do this?

ssoroka:
@srebhan I think it's a bigger issue that this test relies on DNS failure from a live request. There are lots of ways for internet-destined requests to fail, and I generally don't like tests requiring internet access. stubbing out the response accomplishes what the test intended.

ssoroka (approved):

PR #8499, opened 2020-12-02
Adding support for new lines in influx line protocol fields.

### LINKED ISSUE #8095: execd processor does not handle newline characters in strings
### Actual behavior:
If the external processor returns data in the `influxdb` format in several lines, a parsing error occurs:
```
2020-09-08T12:00:39Z E! [processors.execd] Parse error: metric parse error: expected tag at 1:65: "some data here"
```

### Additional info:
The situation is well described here - https://github.com/influxdata/telegraf/issues/7461
But now it occurs in the execd processor

Jagularr:
closes #8095

### DIFF
--- a/plugins/processors/execd/execd.go
+++ b/plugins/processors/execd/execd.go
@@ -12,6 +12,7 @@ import (
 	"github.com/influxdata/telegraf/config"
 	"github.com/influxdata/telegraf/internal/process"
 	"github.com/influxdata/telegraf/plugins/parsers"
+	"github.com/influxdata/telegraf/plugins/parsers/influx"
 	"github.com/influxdata/telegraf/plugins/processors"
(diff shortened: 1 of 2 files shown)


plugins/processors/execd/execd.go
```diff
@@ -137,6 +144,33 @@ func (e *Execd) cmdReadOut(out io.Reader) {
 	}
 }
 
+func (e *Execd) cmdReadOutStream(out io.Reader) {
+	parser := influx.NewStreamParser(out)
```
srebhan, line 148: I assume you need this to avoid side effects!?

Jagularr: Yeah, the stream parser reads directly from that io.Reader stream and is able to correctly parse new line characters.

ssoroka (approved):

PR #8500, opened 2020-12-02
[inputs.github] Add query of pull-request statistics

srebhan:
### Required for all PRs:

This adds the possibility to query the number of "open" and "closed" pull-requests from a GitHub repository. We hide this behind a new `additional_fields` option to limit the effect to the rate-limit as two new queries are necessary for the new fields.

### DIFF
--- a/plugins/inputs/github/README.md
+++ b/plugins/inputs/github/README.md
@@ -23,6 +23,14 @@ alternative method for collecting repository information.
 
   ## Timeout for HTTP requests.
   # http_timeout = "5s"
+
+  ## List of additional fields to query.
+	## NOTE: Getting those fields might involve issuing additional API-calls, so please
+	##       make sure you do not exceed the rate-limit of GitHub.
+	##
+	## Available fields are:
+	## 	- pull-requests			-- number of open and closed pull requests (2 API-calls per repository)
+  # additional_fields = []
 ```
 
 ### Metrics
@@ -52,11 +60,21 @@ When the [internal][] input is enabled:
     - remaining - How many requests you have remaining (per hour)
     - blocks - How many requests have been blocked due to rate limit
 
+When specifying `additional_fields` the plugin will collect the specified properties.
+**NOTE:** Querying this additional fields might require to perform additional API-calls.
+Please make sure you don't exceed the query rate-limit by specifying too many additional fields.
+In the following we list the available options with the required API-calls and the resulting fields
+
+- "pull-requests" (2 API-calls per repository)
+  - fields:
+    - open_pull_requests (int)
+    - closed_pull_requests (int)
+
 ### Example Output
 
 ```
 github_repository,language=Go,license=MIT\ License,name=telegraf,owner=influxdata forks=2679i,networks=2679i,open_issues=794i,size=23263i,stars=7091i,subscribers=316i,watchers=7091i 1563901372000000000
-internal_github,access_token=Unauthenticated rate_limit_remaining=59i,rate_limit_limit=60i,rate_limit_blocks=0i 1552653551000000000
+internal_github,access_token=Unauthenticated closed_pull_requests=3522i,rate_limit_remaining=59i,rate_limit_limit=60i,rate_limit_blocks=0i,open_pull_requests=260i 1552653551000000000
 ```
 
 [GitHub]: https://www.github.com
--- a/plugins/inputs/github/github.go
+++ b/plugins/inputs/github/github.go
@@ -135,18 +135,11 @@ func (g *GitHub) Gather(acc telegraf.Accumulator) error {
 			}
 
 			repositoryInfo, response, err := g.githubClient.Repositories.Get(ctx, owner, repository)
-
-			if _, ok := err.(*github.RateLimitError); ok {
-				g.RateLimitErrors.Incr(1)
-			}
-
+			g.handleRateLimit(response, err)
 			if err != nil {
 				acc.AddError(err)
 				return
 			}
-
-			g.RateLimit.Set(int64(response.Rate.Limit))
-			g.RateRemaining.Set(int64(response.Rate.Remaining))
 
 			now := time.Now()
 			tags := getTags(repositoryInfo)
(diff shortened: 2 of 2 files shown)


plugins/inputs/github/README.md
```diff
@@ -23,6 +23,9 @@ alternative method for collecting repository information.
 
   ## Timeout for HTTP requests.
   # http_timeout = "5s"
+
+  ## Enable collection of additional fields (e.g. open/closed pull requests)
+	# additional_fields = true
```
p-zak, line 28: Spaces, not tabs.

plugins/inputs/github/github.go
```diff
@@ -46,6 +47,9 @@ const sampleConfig = `
 
   ## Timeout for HTTP requests.
   # http_timeout = "5s"
+
+	## Enable collection of additional fields (e.g. open/closed pull requests)
+	# additional_fields = true
```
p-zak, line 52: Spaces, not tabs. 

Moreover it means that by default it should be `true` and it is `false` (default zero value for boolean)

plugins/inputs/github/github.go
```diff
@@ -144,6 +147,17 @@ func (g *GitHub) Gather(acc telegraf.Accumulator) error {
...
 			fields := getFields(repositoryInfo)
 
+			if g.AdditionalFields {
+				// Pull request properties
+				prFields, err := g.getPullRequestFields(ctx, owner, repository)
+				if err != nil {
+					acc.AddError(err)
+				}
+				for k, v := range prFields {
+					fields[k] = v
```
p-zak, line 157: It may happen that `prFields` is not empty when `err != nil`. Do you still want to get these fields in case of error?

plugins/inputs/github/github.go
```diff
@@ -191,6 +205,33 @@ func getFields(repositoryInfo *github.Repository) map[string]interface{} {
...
+		},
+	}
+
+	classes := []string{"open", "closed"}
+	fields := make(map[string]interface{})
+	for _, class := range classes {
+		q := fmt.Sprintf("repo:%s/%s is:pr is:%s", owner, repo, class)
+		searchResult, response, err := g.githubClient.Search.Issues(ctx, q, &options)
+		if err != nil {
+			return fields, err
```
p-zak, line 223: It may happen that `fields` is not empty when err != nil. Do you still want to return it in case of error?

p-zak:
LGTM!

plugins/inputs/github/github.go
```diff
@@ -152,6 +163,35 @@ func (g *GitHub) Gather(acc telegraf.Accumulator) error {
...
+
+func (g *GitHub) gatherAdditionalFields(ctx context.Context, owner, repo string) (map[string]interface{}, error) {
+	fields := make(map[string]interface{})
+
+	// Pull request properties
+	options := github.SearchOptions{TextMatch: false, ListOptions: github.ListOptions{PerPage: 1, Page: 1}}
+	classes := []string{"open", "closed"}
+	for _, class := range classes {
+		q := fmt.Sprintf("repo:%s/%s is:pr is:%s", owner, repo, class)
+		searchResult, response, err := g.githubClient.Search.Issues(ctx, q, &options)
```
reimda, line 183: It looks like this makes an additional http round trip to get the PR counts?  This is a cost we should mention in the config file and readme so when users choose to enable this new functionality they understand why they might not want to.

reimda (changes requested):

plugins/inputs/github/README.md
```diff
@@ -23,6 +23,9 @@ alternative method for collecting repository information.
 
   ## Timeout for HTTP requests.
   # http_timeout = "5s"
+
+  ## Enable collection of additional fields (e.g. open/closed pull requests)
+  # additional_fields = false
```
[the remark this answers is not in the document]
srebhan: Will make it a list like
```
additional_fields=["pull-requests"]
```
so we can extend this later without hassle.

plugins/inputs/github/github.go
```diff
@@ -152,6 +163,35 @@ func (g *GitHub) Gather(acc telegraf.Accumulator) error {
 	return nil
 }
 
+func (g *GitHub) handleRateLimit(response *github.Response, err error) {
+	if err == nil {
+		g.RateLimit.Set(int64(response.Rate.Limit))
```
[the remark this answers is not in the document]
srebhan: I don't think so. The code before called `g.RateLimitErrors.Incr(1)` if a `RateLimitError` occured which also means the error is not `nil`. Afterwards, `acc.AddError()` is called and the function **returns** if `err != nil`. That means the `g.RateLimit.Set()` is not reached in the error case. That hard-to-understand logic made me change it in the first place. :-)

plugins/inputs/github/github.go
srebhan: Agreed will do.

plugins/inputs/github/github.go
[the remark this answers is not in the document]
reimda: My mistake.  I didn't read the code closely enough yesterday.

plugins/inputs/github/github.go
```diff
@@ -191,6 +218,33 @@ func getFields(repositoryInfo *github.Repository) map[string]interface{} {
...
+			Page:    1,
+		},
+	}
+
+	classes := []string{"open", "closed"}
+	fields := make(map[string]interface{})
+	for _, class := range classes {
+		q := fmt.Sprintf("repo:%s/%s is:pr is:%s", owner, repo, class)
+		searchResult, response, err := g.githubClient.Search.Issues(ctx, q, &options)
+		if err != nil {
```
reimda, line 235: Looks like we need to check that we hit the rate limit here too?

```suggestion
		if _, ok := err.(*github.RateLimitError); ok {
			g.RateLimitErrors.Incr(1)
		}
		if err != nil {
```

srebhan: Seem like I messed-up the commit and lost `handleRateLimit()` on the way. Brought it back to fix this instance.

reimda (approved):

PR #8505, opened 2020-12-03
Fixed minor issue in datacenter name lookup code

prydin:
### Required for all PRs:

Due to a spelling error in an if-statement, the code to resolve names of datacenters was called even when the name was known. This also caused a somewhat misleading debug message to be shown.

### DIFF
--- a/plugins/inputs/vsphere/endpoint.go
+++ b/plugins/inputs/vsphere/endpoint.go
@@ -437,7 +437,7 @@ func (e *Endpoint) discover(ctx context.Context) error {
 			}
 
 			// Fill in datacenter names where available (no need to do it for Datacenters)
-			if res.name != "Datacenter" {
+			if res.name != "datacenter" {
 				for k, obj := range objects {
 					if obj.parentRef != nil {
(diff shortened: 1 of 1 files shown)


srebhan (approved):
Nice. As I cannot test, I take your word on it.

plugins/inputs/vsphere/endpoint.go
```diff
@@ -437,7 +437,7 @@ func (e *Endpoint) discover(ctx context.Context) error {
 			}
 
 			// Fill in datacenter names where available (no need to do it for Datacenters)
-			if res.name != "Datacenter" {
+			if res.name != "datacenter" {
```
ssoroka, line 440: Is there any value in making this comparison case insensitive, or is it always capitalized?

ssoroka (approved):

reimda:
@prydin We'd like to merge this for the 1.17 release tomorrow.  Could you respond to Steven's question about case sensitivity?

plugins/inputs/vsphere/endpoint.go
prydin: It’s always capitalized. It’s just a key used internally in the code.

PR #8507, opened 2020-12-03
Improve SQL connection lifecycle

IgorKuchmienko:
Instead of opening a new SQL connection for every query, open it for a server, and use it for all queries within this server. We observed that opening a new connection for each query lead to sporadic connection errors in the Telegraf log file when monitoring 100 Azure SQL DBs with all queries enabled.

### Required for all PRs:

### DIFF
--- a/plugins/inputs/sqlserver/sqlserver.go
+++ b/plugins/inputs/sqlserver/sqlserver.go
@@ -250,7 +250,7 @@ func (s *SQLServer) Gather(acc telegraf.Accumulator) error {
 	return nil
 }
 
-func (s *SQLServer) gatherServer(conn *sql.DB, server string, query Query, acc telegraf.Accumulator) error {	
+func (s *SQLServer) gatherServer(conn *sql.DB, server string, query Query, acc telegraf.Accumulator) error {
 	// execute query
 	rows, err := conn.Query(query.Script)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


ssoroka (changes requested):
It looks like this changes the functionality to run queries serially instead of in parallel. A better approach might be to use a connection pool that can reuse the connection for multiple queries in parallel up to some max parallelism.

denzilribeiro:
Running serially is not an option as would take the 10-15 seconds interval once # queries increase. Are you sure you have multiple connections to the same DB? I was under impression driver had connection pooling.

IgorKuchmienko:
This change was made to address this issue:
https://github.com/influxdata/telegraf/issues/8585

@ssoroka, yes, this change makes the queries run serially. I agree that with a large number of queries this can cause problems; on the other hand the parallel approach leads to the issue above. We need to understand where the bottleneck is. It may be a problem with the sql server driver https://github.com/denisenkom/go-mssqldb on a large number of parallel connections (roughly estimated as 200 DBs * 10 queries = 2000 connections).

srebhan:
Anyone willing to work on this PR?

srebhan:
I'm closing this as it is unclear if the fix is still needed. If so, please reopen the PR and/or drop me a note.

PR #8509, opened 2020-12-03
Add configurable Max TTL duration for statsd input plugin entries

### LINKED ISSUE #8348: Max ttl for gauges in statsd input plugin
### Proposal:

Allow for setting "gaugesMaxTTL" option in the `statsd` input plugin. Essentially, this from the core statsd repo:

https://github.com/statsd/statsd/pull/599/files

### Current behavior:

If `deleteGauges` is set to "False" then once a gauge metric is reported it is reported for the life of the `telegraf` server. There is no way to set a TTL value where if the gauge metric is not received at the ingress endpoint after a certain time it is reset to 0.

### Desired behavior:

An optional `gaugesMaxTTL` option is available for the `statsd` input -- this has exactly the same effect as the `deleteGauges` option, except that it only executes that logic after no input has been received for the value specified in `gaugesMaxTTL`.

### Use case: 

Telegraf server `A` on host `X` is taking input from service `B` and service `C`. Both of these services are reporting metrics via a gauge with "deleteGauges" set to `False` -- let's assume this is required w/o getting distracted by details. :-)

Now, service `C` is turned down by some load-managing automation. However, telegraf `A` will still report the last gauge metrics to Influx until restarted, with no way of ever reflecting the fact that service `C` is permanently gone.

Jagularr:
The new config value will establish how long any metric can sit in the gauges/counters/timings/sets without being updated / received. When the Max TTL elapses, the metric will stop being reported until it was received again.

closes #8348

### DIFF
--- a/plugins/inputs/statsd/statsd_test.go
+++ b/plugins/inputs/statsd/statsd_test.go
@@ -2,14 +2,16 @@ package statsd
 
 import (
 	"fmt"
-	"github.com/influxdata/telegraf"
-	"github.com/influxdata/telegraf/internal"
-	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
 	"net"
 	"sync"
 	"testing"
 	"time"
+
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/config"
+	"github.com/influxdata/telegraf/internal"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	"github.com/influxdata/telegraf/testutil"
 )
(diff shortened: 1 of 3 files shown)


plugins/inputs/statsd/statsd_test.go
```diff
@@ -1077,6 +1079,65 @@ func TestParse_MeasurementsWithSameName(t *testing.T) {
...
+				"valid",
+				map[string]string{
+					"metric_type": "counter",
+				},
+				map[string]interface{}{
+					"value": 90,
+				},
+				time.Now(),
+				telegraf.Counter,
+			),
```
ssoroka, line 1123: I'm surprised that we get two metrics back with a value of `90`, any idea why this is?

Jagularr: The reason it's coming out like that is that the expire function comes after it accumulates the metrics. So it's going to end up removed on the next run. If you think it would be a better idea to have that function come first, then we could make that change.

ssoroka (approved):
Nevermind, I got confused about the number of Gather calls. :D it's all good.

LINKIWI:
After upgrading from Telegraf 1.16.3 to 1.17.0, I've noticed that tags from statsd emissions are being mangled with unrelated metrics. I'm not pointing to this PR as the root cause for the regression, but this is also the only thing that's changed with the statsd plugin since the last release. If I have some more time I'll compile a more thorough bug report with a reproducible example.

Jagularr:
Interesting. Thank you for reporting that, let me know when/if you get more info!

gnjack:
We've also seen the same mangled tags from statsd metrics after upgrading to 1.17.0

gnjack:
My golang is very rusty, but I think this is the cause of the issue:

The `s.Lock()` [was moved](https://github.com/influxdata/telegraf/commit/2187baceea458e571560b5a0b9f16ba382137ac4#r45966095) from the start of `parseStatsdLine` to the start of `aggregate`. This means the lock has been moved from before tag calculation for a metric to after the tag calculation.

As the parser [was recently multithreaded](https://github.com/influxdata/telegraf/pull/6922), this means there may be multiple goroutines calling the following line in `parseStatsdLine` which is now no longer inside a lock:
https://github.com/influxdata/telegraf/blob/76c2201bbe9f098a09e01fa1e39790a973d7da8e/plugins/inputs/statsd/statsd.go#L648-L649

`func (s *Statsd) parseName(bucket string)` does not appear to be thread safe. First it pulls out tags for the current metric it was invoked with
https://github.com/influxdata/telegraf/blob/76c2201bbe9f098a09e01fa1e39790a973d7da8e/plugins/inputs/statsd/statsd.go#L691-L696

then creates a `GraphiteParser` on the shared `Statsd` struct `s`
https://github.com/influxdata/telegraf/blob/76c2201bbe9f098a09e01fa1e39790a973d7da8e/plugins/inputs/statsd/statsd.go#L707-L710

and sets the default tags on that shared `GraphiteParser` to the current metrics tags
https://github.com/influxdata/telegraf/blob/76c2201bbe9f098a09e01fa1e39790a973d7da8e/plugins/inputs/statsd/statsd.go#L713

this means multiple goroutines will be concurrently modifying `s.graphiteParser.DefaultTags` to different values and calling `name, tags, field, _ = s.graphiteParser.ApplyTemplate(name)` which will return incorrect tags if another goroutine has modified  `s.graphiteParser.DefaultTags`

PR #8511, opened 2020-12-03
Add cloud controlled configuration management

TheGoldenKyle:
This PR outlines the details of Telegraf Assistant, which adds the functionality of stopping, starting, updating, and retrieving information about plugins without having to stop the Telegraf service.

Telegraf Assistant aims to take the hassle out of setting up config files, allowing users to directly edit their configurations through an easy-to-use web dashboard (not yet created) rather than editing the raw TOML files. To do so, the assistant handles CRUD operations for a single plugin while Telegraf is running. The plugins are referenced by id, rather than name, so that a user may have multiple instances of the same plugin. Any changes made to the plugin configurations will be saved back to the user's config file (currently updated_config.conf) to retain the proper settings when Telegraf is started again.

See assistant/README.md for a detailed explanation of how the Assistant is integrated with the Agent, and for examples of request payloads and responses.

Comprehensive list of Plugin API:
- Start plugin
- Stop plugin
- Update plugin
- Get plugin
- Get plugin schema
- Get all plugins
- Get all running plugins

Unit and integration tests can be found in assistant/assistant_test.go and config/config_test.go. Changes were also made to existing tests to account for the new plugin-id implementation and to fix data race conditions.

### Required for all PRs:


ssoroka:
CI says `please run go mod tidy and check in changes`

sjwang90:
We loved working on this project with you all! However the API implementation in Telegraf has been de-prioritized for the near future. We'll definitely take over this PR if it does get implemented in the future but thank you for all your hard work back in 2020!

PR #8512, opened 2020-12-03
Update string parsing of allocatable cpu cores in kube_inventory

j03wang:
The Allocatable cpu value returned from Kubernetes can have a milli suffix, which gets parsed as 0 when stored in allocatable_cpu_cores.

This change updates the metric to be millicpus to prevent truncation of decimals.

### Required for all PRs:

### DIFF
--- a/plugins/inputs/kube_inventory/README.md
+++ b/plugins/inputs/kube_inventory/README.md
@@ -190,10 +190,10 @@ subjects:
   - tags:
     - node_name
   - fields:
-    - capacity_cpu_cores
+    - capacity_millicpu_cores
     - capacity_memory_bytes
     - capacity_pods
-    - allocatable_cpu_cores
+    - allocatable_millicpu_cores
     - allocatable_memory_bytes
     - allocatable_pods
 
--- a/plugins/inputs/kube_inventory/README.md
+++ b/plugins/inputs/kube_inventory/README.md
@@ -190,9 +190,11 @@ subjects:
   - tags:
     - node_name
   - fields:
+    - capacity_cpu_cores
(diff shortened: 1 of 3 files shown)


srebhan (approved):
Looks good to me. I cannot test, so I take your word on it. :-)

j03wang:
Now that I think about it, would it make sense to update the metric itself to be "allocatable_millicpu_cores" and scale the value by 1000? Some tsdbs can't handle decimal values and would likely round/truncate the current value.

srebhan (approved):
Also fine with me. :-)

plugins/inputs/kube_inventory/README.md
```diff
@@ -190,10 +190,10 @@ subjects:
   - tags:
     - node_name
   - fields:
-    - capacity_cpu_cores
```
reimda, line 193: Removing fields breaks backward compatibility so we won't be able to merge this as is.

For example, if a user of this plugin saves metrics to a database and uses the capacity_cpu_cores field, then they upgrade to a telegraf that doesn't send the capacity_cpu_cores field, their queries will stop returning what they expected.

Adding new fields is ok.  It would also be ok to add a setting to let the user choose whether they want whole cpu or millicpu units and defaulted the setting to the previous behavior of using whole cpu units.

reimda (changes requested):

plugins/inputs/kube_inventory/README.md
j03wang: Sounds good - I've added back the original fields.

reimda (approved):

PR #8529, opened 2020-12-08
feat: Added possibility to detect local config file changes and reload

### LINKED ISSUE #7187: Telegraf watch and restart with new configuration file
## Feature Request

Telegraf should watch and restart with new configuration file

### Proposal:

Telegraf watch and restart with new configuration file

### Current behavior:
If a new config file is placed in the same place where telegraf is running, the only way of picking up the changes is either to restart the process or sending SIGHUP signal

### Desired behavior:

Telegraf making a md5 checksum of the whole config file once it's fully templated out and then trying to template it out every 10/30/60 seconds to see if something changed. If the md5 checksum has changed, then trigger a restart 

### Use case: 

This would be very useful in a container environment because config files can be updated very often and it would be. and it's very easy to run hundreds/thousands of telegraf so the more automation we have the better

vlastahajek:
Closes #7187

There are situations (e.g. Kubernetes environment, Windows service) when restarting Telegraf due to config changes is cumbersome, and sending _SIGHUP_ signal to achieve that is difficult.

Telegraf watching its config for changes will make this easy.

**This PR adds watching for a single local config file only, not config dir or URL .**

Watching is made using the `infuxdata/tail` lib and it is possible to be enabled via cmd option `watch-config` with two possible values:
 - inotify - uses fsnotify from tail
 - poll - uses polling from tail 

However, polling from tail needs a fix (https://github.com/influxdata/tail/pull/9) and until the tail is fixed, it is redirected to the fork with a simple fix.

### Required for all PRs:

### DIFF
--- a/cmd/telegraf/telegraf.go
+++ b/cmd/telegraf/telegraf.go
@@ -15,6 +15,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/influxdata/tail/watch"
 	"github.com/influxdata/telegraf/agent"
 	"github.com/influxdata/telegraf/config"
 	"github.com/influxdata/telegraf/internal"
@@ -26,6 +27,7 @@ import (
 	"github.com/influxdata/telegraf/plugins/outputs"
 	_ "github.com/influxdata/telegraf/plugins/outputs/all"
 	_ "github.com/influxdata/telegraf/plugins/processors/all"
+	"gopkg.in/tomb.v1"
 )
 
 // If you update these, update usage.go and usage_windows.go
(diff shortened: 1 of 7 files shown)


ssoroka:
Note that you can already do this with --config-directory, as watching for file changes and reloading configuration is the default behavior for that.

sjwang90:
@vlastahajek do you know how this could be implemented for URL-based configs? This would be really helpful for our cloud customers and address use cases like this: https://github.com/influxdata/telegraf/issues/8730

vlastahajek:
@sjwang90, this shouldn't be hard. 
As there is not guarantee that server holding the remote config supports HEAD (e.g. InfluxDB server doesn't) and also there is not guarantee that server will return info about the resource state, e.g. `ETag` header or `Date-Modified` header, easiest solution could be to store hash (MD5) of the current config and periodically (there should be parametrized time interval) download copy of the remote config, compute hash and compare. If changed, restart. 

The question is, what should be default time interval for check  to not bother server much? 1 minute?

PR #8543, opened 2020-12-10
Add configurable number of 'most recent' date-stamped indices to gather in Elasticsearch input

### LINKED ISSUE #8517: Add configuration option to elasticsearch input to collect only the most recent fields
## Feature Request

### Current behavior:
The plugin currently gathers all historical data, which creates too much data and increases cardinality. Customer would like to filter on a specific time frame, like last 24 hours.

### Desired behavior:
New configuration option: If my indices are datestamped - I'll get X of the newest datestamps depending on what I configure

### Use case: 

Customer using the elasticsearch input plugin and only wants the following fields:
`indexing_index_failed `(float)

He only wanted it for last 24 hours, not all historical data. Add this feature to this specific plugin so it just matches a daily pattern, and ignores everything that’s old.

Jagularr:
Add configurable number of 'most recent' date-stamped indices to gather in the Elasticsearch input plugin, and allow wildcards to account for date-suffixed index names. Configuring '3' for num_most_recent_indices will only gather the 3 latest indices, based on the date or number they end with. Finding the date or number is dependent on the targeted indices being configured with wildcards at the end of their 'base' names.

closes #8517

### DIFF
--- a/plugins/inputs/elasticsearch/elasticsearch.go
+++ b/plugins/inputs/elasticsearch/elasticsearch.go
@@ -166,6 +166,7 @@ type Elasticsearch struct {
 	client          *http.Client
 	serverInfo      map[string]serverInfo
 	serverInfoMutex sync.Mutex
+	indexMatchers   map[string]filter.Filter
 }
 type serverInfo struct {
 	nodeID   string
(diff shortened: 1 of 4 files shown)


plugins/inputs/elasticsearch/elasticsearch.go
```diff
@@ -527,66 +535,130 @@ func (e *Elasticsearch) gatherIndicesStats(url string, acc telegraf.Accumulator)
...
-						"type":       shardType,
-					}
+	// Bucket each returned index with its associated configured index wildcard (if any match).
+	for indexName := range indices {
+		matchingPrefix := indexName
+		for _, configuredIndex := range e.IndicesInclude {
+			// If an index name was configured with a wildcard, check if the index found matches the configured one.
+			if strings.HasSuffix(configuredIndex, "*") && strings.HasPrefix(indexName, strings.TrimSuffix(configuredIndex, "*")) {
+				matchingPrefix = configuredIndex
+			}
```
ssoroka, line 595: can we switch to filter/filter.go's matching here? We'll have much more robust matching support.

ssoroka (changes requested):
Looks good. I'd like to change to our built-in glob library

plugins/inputs/elasticsearch/elasticsearch.go
Jagularr: Updated.

PR #8544, opened 2020-12-11
Add an optional health metric for the sqlserver input plugin

coquagli:
### Required for all PRs:

Add the ability to collect a per-connection health metric for the sqlserver input plugin. This health metric is optional and is disabled by default. Users can leverage this metric to detect if their SQL Server monitoring is not working as intended. The details below are fairly comprehensive and are copied from the updated README.

**Health Metric**
All collection versions (version 1, version 2, and database_type) support an optional plugin health metric called `sqlserver_telegraf_health`. This metric tracks if connections to SQL Server are succeeding or failing. Users can leverage this metric to detect if their SQL Server monitoring is not working as intended.

In the configuration file, toggling `health_metric` to true will enable collection of this metric. By default, this value is set to false and the metric is not collected. The health metric emits one record for each connection specified by servers in the configuration file.

The health metric emits the following tags:

- `sql_instance` - Name of the server specified in the connection string. This value is emitted as-is in the connection string. If the server could not be parsed from the connection string, a constant placeholder value is emitted
- `database_name` - Name of the database or (initial catalog) specified in the connection string. This value is emitted as-is in the connection string. If the database could not be parsed from the connection string, a constant placeholder value is emitted

The health metric emits the following fields:

- `attempted_queries` - Number of queries that were attempted for this connection
- `successful_queries` - Number of queries that completed successfully for this connection
- `database_type` - Type of database as specified by `database_type` in the configuration file. If `database_type` from the configuration file is empty, the `QueryVersion` and `AzureDB` fields are concatenated instead

If `attempted_queries` and `successful_queries` are not equal for a given connection, some metrics were not successfully gathered for that connection. If `successful_queries` is 0, no metrics were successfully gathered.


ssoroka (changes requested):
I think this is largely covered already by the regular output stats collected by `inputs.internal`

plugins/inputs/sqlserver/connectionstring.go
```diff
@@ -0,0 +1,121 @@
+package sqlserver
+
+import (
+	"net"
+	"net/url"
+	"strings"
+)
+
+const emptyServerName = "<empty-server-name>"
```
masree, line 9: nit: Maybe it can go into a single const ( emptyServername = ""
                                                                 emptyDBName = ""
                                                                 otherVariables = "" 
                                                                 )

plugins/inputs/sqlserver/connectionstring.go
```diff
@@ -0,0 +1,121 @@
...
+
+		var value string = ""
+		if len(keyAndValue) > 1 {
+			value = strings.TrimSpace(keyAndValue[1])
+		}
+
+		if isInstanceIdentifier(key) {
+			sqlInstance = value
+			if databaseName != emptyDatabaseName {
+				break
```
masree, line 61: nit: Can we 'return' from here instead of break as there are no further statements to execute outside of 'for' loop?

plugins/inputs/sqlserver/connectionstring.go
```diff
@@ -0,0 +1,121 @@
...
+	return
+}
+
+// parseConnectionStringURL parses a URL-formatted connection string and returns the SQL instance and database name
+func parseConnectionStringURL(connectionString string) (sqlInstance string, databaseName string) {
+	sqlInstance = emptyServerName
+	databaseName = emptyDatabaseName
+
+	u, err := url.Parse(connectionString)
+	if err != nil {
```
masree, line 80: Do we need to log the error if unable to parse?

plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -36,8 +38,25 @@ type Query struct {
...
 
+// HealthMetric struct tracking the number of attempted vs successful connections for each connection string
+type HealthMetric struct {
+	AttemptedQueries  int
+	SuccessfulQueries int
+}
+
 const defaultServer = "Server=.;app name=telegraf;log=1;"
 
+const typeAzureSQLDB = "AzureSQLDB"
```
masree, line 49: nit: same as earlier -it can be grouped under single const variable for readability

plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -327,6 +351,62 @@ func (s *SQLServer) accRow(query Query, acc telegraf.Accumulator, row scanner) e
 	return nil
 }
 
+// gatherHealth stores info about any query errors in the healthMetrics map
+func (s *SQLServer) gatherHealth(healthMetrics map[string]*HealthMetric, serv string, queryError error, mutex *sync.Mutex) {
+	if !s.HealthMetric {
+		return
+	}
+
+	mutex.Lock()
```
masree, line 360: I understand why we're trying to do this. Does this mean that now we're serializing instead of parallel? Even though gatherServer is parallel, all processes will need to execute gatherHealth in a serial fashion right coz of the lock? If yes, do you think having pub-sub mechanism will help to retain the parallelism? Like gatherServer publises the queryError and notifies gatherHealth whenever there are new records?

plugins/inputs/sqlserver/connectionstring.go
```diff
@@ -0,0 +1,121 @@
...
+				break
+			}
+		}
+	}
+
+	return
+}
+
+// parseConnectionStringURL parses a URL-formatted connection string and returns the SQL instance and database name
+func parseConnectionStringURL(connectionString string) (sqlInstance string, databaseName string) {
```
masree, line 75: nit: Since we are setting value at beginning of method to emptyServerName and emptyDatabaseName, having named return value might be redundant. Named return will initialize the string to "" if declared. Is it for readability we added named returns?

masree:
> I think this is largely covered already by the regular output stats collected by `inputs.internal`

Thanks @ssoroka ! I took a look at inputs.internal and it looks like these metrics are reported either at the agent level or at the plugin level. However, we're looking to gather metrics per connection (i.e server/database combination). Having this data point will help us check which connection failed vs which connection succeeded in a plugin and also number of queries that were successfully executed per connection. I think it is not possible to gather such metrics per connection from inputs.internal as of today. It would be great if you can help clarify and confirm if this approach seems ok.

masree:
> I think this is largely covered already by the regular output stats collected by `inputs.internal`



> > I think this is largely covered already by the regular output stats collected by `inputs.internal`
> 
> Thanks @ssoroka ! I took a look at inputs.internal and it looks like these metrics are reported either at the agent level or at the plugin level. However, we're looking to gather metrics per connection (i.e server/database combination). Having this data point will help us check which connection failed vs which connection succeeded in a plugin and also number of queries that were successfully executed per connection. From my understanding, I think it is not possible to gather such metrics per connection from inputs.internal as of today. It would be great if you can help clarify and let us know your thoughts. Thanks!

@ssoroka Please let us know if per connection level metric is possible from inputs.internal. From the docs, we could not its usage for such granularity. Thanks!

plugins/inputs/sqlserver/sqlserver.go
coquagli: Queries are where a bulk of the time are spent. Queries are still executed in parallel. Once queries are complete, their results are added to the dictionary in serial. That critical section is pretty small. I don't see it causing any more than ones of milliseconds of delay. 

I'd be in favor of keeping this as-is (with a single lock and dictionary). Other strategies would add more complexity than would be needed for this kind of simple parallelization.

plugins/inputs/sqlserver/connectionstring.go
coquagli: I actually thought about this for a while when I originally wrote it. I decided not to log to avoid exposing any secrets possibly contained in the connection string. If we fail to parse here, then the SQL connection will certainly fail, so it won't be failing silently.

plugins/inputs/sqlserver/connectionstring.go
coquagli: Done.

plugins/inputs/sqlserver/connectionstring.go
coquagli: Done.

plugins/inputs/sqlserver/sqlserver.go
coquagli: Done. Created multiple groups to clarify which constants are related.

plugins/inputs/sqlserver/connectionstring.go
coquagli: Correct, I added the named returns for readability. Feel free to reopen this conversation if you want to discuss more.

plugins/inputs/sqlserver/sqlserver.go
masree: ok, as discussed maybe we can add some testing results to know the time spent in this block. (timer to keep track of time during its execution in this mutex). Negligible amount of delay as you called out ( almost 0 or ones of milliseconds) might be ok.

plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -327,6 +355,62 @@ func (s *SQLServer) accRow(query Query, acc telegraf.Accumulator, row scanner) e
...
```
masree, line 364: publishing the time spent on this block might be helpful to indicate it is light weight and fast

masree (approved):

plugins/inputs/sqlserver/connectionstring.go
```diff
@@ -0,0 +1,124 @@
...
+	emptyServerName   = "<empty-server-name>"
+	emptyDatabaseName = "<empty-database-name>"
+)
+
+// getConnectionIdentifiers returns the sqlInstance and databaseName from the given connection string.
+// The name of the server is returned as-is in the connection string
+// If the connection string could not be parsed or sqlInstance/databaseName were not present, a placeholder value is returned
+func getConnectionIdentifiers(connectionString string) (sqlInstance string, databaseName string) {
+	if len(connectionString) == 0 {
+		return
```
srebhan, line 21: Please use explicit returns.

plugins/inputs/sqlserver/connectionstring.go
```diff
@@ -0,0 +1,124 @@
...
+		i := strings.Index(trimmedConnectionString, odbcPrefix)
+		connectionStringWithoutOdbc := trimmedConnectionString[i+len(odbcPrefix):]
+		sqlInstance, databaseName = parseConnectionStringKeyValue(connectionStringWithoutOdbc)
+	} else if strings.HasPrefix(trimmedConnectionString, "sqlserver://") {
+		sqlInstance, databaseName = parseConnectionStringURL(trimmedConnectionString)
+	} else {
+		sqlInstance, databaseName = parseConnectionStringKeyValue(trimmedConnectionString)
+	}
+
+	return
```
srebhan, line 36: Please use
```suggestion
	if strings.HasPrefix(trimmedConnectionString, odbcPrefix) {
		i := strings.Index(trimmedConnectionString, odbcPrefix)
		connectionStringWithoutOdbc := trimmedConnectionString[i+len(odbcPrefix):]
		return parseConnectionStringKeyValue(connectionStringWithoutOdbc)
	}
	if strings.HasPrefix(trimmedConnectionString, "sqlserver://") {
		return parseConnectionStringURL(trimmedConnectionString)
	}
	return parseConnectionStringKeyValue(trimmedConnectionString)
```
to save some indention and make the code easier to follow.

plugins/inputs/sqlserver/connectionstring.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+//isInstanceIdentifier returns true if the input is equal to "server" (case insensitive)
+func isInstanceIdentifier(input string) bool {
+	// The GoLang SQL driver only supports "server" in the connection string, not any of the following others:
+	// "data source", "address", "addr", "network address"
+	return strings.EqualFold("server", input)
```
srebhan, line 117: Can we please move this up to the actual use of `isInstanceIdentifier()` to get rid of this function and save some lines and indirection.

plugins/inputs/sqlserver/connectionstring.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+//isDatabaseIdentifier returns true if the input is equal to "database" (case insensitive)
+func isDatabaseIdentifier(input string) bool {
+	// The GoLang SQL driver only supports "database" in the connection string, not "initial catalog"
+	return strings.EqualFold("database", input)
```
srebhan, line 123: Can we please move this up to the actual use of `isDatabaseIdentifier()` to get rid of this function and save some lines and indirection.

plugins/inputs/sqlserver/sqlserver.go
coquagli: Sure, good call. For others: I timed the code block from right before `Lock()` until right after `Unlock()` it was consistently taking 1s of nanoseconds to `Lock`, execute the codeblock, then `Unlock`.

plugins/inputs/sqlserver/connectionstring.go
```diff
@@ -0,0 +1,124 @@
...
+			continue
+		}
+
+		var value string = ""
+		if len(keyAndValue) > 1 {
+			value = strings.TrimSpace(keyAndValue[1])
+		}
+
+		if isInstanceIdentifier(key) {
+			sqlInstance = value
```
srebhan, line 62: What happens if you hit a `keyValuePair` with `"server"` only. In this case `value` will be empty (as `len(keyAndValue)==1` and in turn `sqlInstance` is empty. Shouldn't we return `emptyServerName` in this case?
The same holds true for `databaseName`.

plugins/inputs/sqlserver/connectionstring.go
```diff
@@ -0,0 +1,124 @@
...
+		} else if isDatabaseIdentifier(key) {
+			databaseName = value
+			if sqlInstance != emptyServerName {
+				return
+			}
+		}
+	}
+
+	return
+}
```
srebhan, line 75: Maybe check at the end and spend the extra cycles in the loop for readability...
```suggestion
func parseConnectionStringKeyValue(connectionString string) (sqlInstance string, databaseName string) {
	sqlInstance := ""
	databaseName := ""

	keyValuePairs := strings.Split(connectionString, ";")
	for _, keyValuePair := range keyValuePairs {
		if len(keyValuePair) == 0 {
			continue
		}

		keyAndValue := strings.SplitN(keyValuePair, "=", 2)
		key := strings.TrimSpace(strings.ToLower(keyAndValue[0]))
		if len(key) == 0 {
			continue
		}

		value := ""
		if len(keyAndValue) > 1 {
			value = strings.TrimSpace(keyAndValue[1])
		}

		if isInstanceIdentifier(key) {
			sqlInstance = value
			continue
		} 
		if isDatabaseIdentifier(key) {
			databaseName = value
		}
	}

        if sqlInstance == "" {
                sqlInstance = emptyServerName
        }

        if databaseName == "" {
                databaseName = emptyDatabaseName
        }
        return sqlInstance, databaseName
}
```

plugins/inputs/sqlserver/connectionstring.go
```diff
@@ -0,0 +1,124 @@
...
+
+// parseConnectionStringURL parses a URL-formatted connection string and returns the SQL instance and database name
+func parseConnectionStringURL(connectionString string) (sqlInstance string, databaseName string) {
+	sqlInstance = emptyServerName
+	databaseName = emptyDatabaseName
+
+	u, err := url.Parse(connectionString)
+	if err != nil {
+		return
```
srebhan, line 84: Please use explicit returns in telegraf!

plugins/inputs/sqlserver/connectionstring.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	host, _, err := net.SplitHostPort(u.Host)
+	if err != nil {
+		// The host did not have a port
+		sqlInstance = u.Host
```
srebhan, line 90: Well there are more errors returned by `net.SplitHostPort()` e.g. indicating a malformed address etc. I think it is not save to assume that the host part didn't have a port in case of any error. So either check for the appropriate error or check for the port separator beforehand:
```suggestion
	if host, _, err := net.SplitHostPort(u.Host); err != nil {
		if ! string.Contains(u.Host, ":") {
			// The host did not have a port
			sqlInstance = u.Host
		}
```

plugins/inputs/sqlserver/connectionstring.go
```diff
@@ -0,0 +1,124 @@
...
+	} else {
+		// The host did have a port (e.g. "the.host.com:1234")
+		sqlInstance = host
+	}
+
+	if len(u.Path) > 1 {
+		// There was a SQL instance name specified in addition to the host
+		// E.g. "the.host.com:1234/InstanceName" or "the.host.com/InstanceName"
+		sqlInstance = sqlInstance + "\\" + u.Path[1:]
```
srebhan, line 99: Do you really need backslashes here or can we do with slash?

plugins/inputs/sqlserver/connectionstring.go
```diff
@@ -0,0 +1,124 @@
...
+
+	query := u.Query()
+	for key, value := range query {
+		if isDatabaseIdentifier(key) {
+			databaseName = value[0]
+			break
+		}
+	}
+
+	return
```
srebhan, line 110: Again, explicit return.

plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -327,6 +355,62 @@ func (s *SQLServer) accRow(query Query, acc telegraf.Accumulator, row scanner) e
 	return nil
 }
 
+// gatherHealth stores info about any query errors in the healthMetrics map
+func (s *SQLServer) gatherHealth(healthMetrics map[string]*HealthMetric, serv string, queryError error, mutex *sync.Mutex) {
+	if !s.HealthMetric {
+		return
+	}
```
srebhan, line 362: How can this happen? And shouldn't this be a big fat error in case it _really_ happens?

plugins/inputs/sqlserver/sqlserver.go
srebhan: I feel more safe if you put a `defer mutex.Unlock()` right below that line.

plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -327,6 +355,62 @@ func (s *SQLServer) accRow(query Query, acc telegraf.Accumulator, row scanner) e
...
+
+	mutex.Unlock()
+}
+
+// accHealth accumulates the query health data contained within the healthMetrics map
+func (s *SQLServer) accHealth(healthMetrics map[string]*HealthMetric, acc telegraf.Accumulator) {
+	if !s.HealthMetric {
+		return
+	}
```
srebhan, line 382: Same as above.

plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -327,6 +355,62 @@ func (s *SQLServer) accRow(query Query, acc telegraf.Accumulator, row scanner) e
...
+
+	var b strings.Builder
+	b.WriteString("QueryVersion-")
+	fmt.Fprintf(&b, "%d", s.QueryVersion)
+
+	if s.AzureDB {
+		b.WriteString("-AzureDB")
+	}
+
+	return b.String()
```
srebhan, line 411: Why not using
```suggestion
	logname := fmt.Sprintf("QueryVersion-%d", s.QueryVersion)
	if s.AzureDB {
		logname += "-AzureDB"
	}

	return b.String()
```

plugins/inputs/sqlserver/sqlserver.go
srebhan: Btw: If this is really a discussion point about holding the log for that few operations, just change the pattern to return if the query succeeded or not (i.e 1 if successful 0 otherwise). Then in the caller hold the lock and sum up the successful ones in `SuccessfulQueries` and add one to `AttemptedQueries` in the `healthMetrics` map.

In general I get the feeling that this is a very complicated way to count. :-)

plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -36,8 +38,29 @@ type Query struct {
...
+)
+
+const (
+	healthMetricName              = "sqlserver_telegraf_health"
+	healthMetricInstanceTag       = "connection_sql_instance"
+	healthMetricDatabaseTag       = "connection_database_name"
+	healthMetricAttemptedQueries  = "AttemptedQueries"
+	healthMetricSuccessfulQueries = "SuccessfulQueries"
+	healthMetricDatabaseType      = "DatabaseType"
+)
```
srebhan, line 62: I'm still not sure I like this definitions. You use it in one place in the code, so it seems to me you are adding too much lines here, on the other hand you have everything in one place... As it is a matter of taste, please proceed as you prefer...

However, please unify the label styles here. I don't think having snake-case (tags) and CamelCase (fields) in one metric is very intuitive.

plugins/inputs/sqlserver/sqlserver_test.go
```diff
@@ -138,13 +139,59 @@ func TestSqlServer_MultipleInstanceIntegration(t *testing.T) {
...
-	// acc2 includes memory metrics, and excludes size metrics
+	// acc2 includes memory metrics and the health metric, and excludes size metrics
+	assert.True(t, acc2.HasMeasurement(healthMetricName))
 	assert.True(t, acc2.HasMeasurement("Memory breakdown (%)"))
 	assert.False(t, acc2.HasMeasurement("Log size (bytes)"))
+
+	sqlInstance, database := getConnectionIdentifiers(testServer)
+	tags := map[string]string{healthMetricInstanceTag: sqlInstance, healthMetricDatabaseTag: database}
+	assert.True(t, acc2.HasPoint(healthMetricName, tags, healthMetricAttemptedQueries, 9))
+	assert.True(t, acc2.HasPoint(healthMetricName, tags, healthMetricSuccessfulQueries, 0))
```
srebhan, line 156: Please don't change this test-case! Here we have a nice indicator if the old behavior still exists with the previous settings. You can easily duplicate this test with the changes you propose but leave the old one unchanged!

srebhan (commented):
Hey @coquagli thanks for submitting this PR. I think the code looks quite good already. I have one major remark about using explicit returns. In telegraf we use explicit returns instead of implicit  ones, so please change those.

plugins/inputs/sqlserver/sqlserver_test.go
coquagli: Good point.

plugins/inputs/sqlserver/sqlserver.go
coquagli: Good gatch on the snake-case vs. CamelCase. I changed that. 

I added these as constants here because I wanted to reference them in the sqlserver_test.go. It made testing a bit cleaner.

plugins/inputs/sqlserver/sqlserver.go
coquagli: I like it. Done.

plugins/inputs/sqlserver/sqlserver.go
coquagli: This is actually the only check that enables/disables gathering the health metric values. You can see that on line 261, `gatherHealth` will always be called even when the health metric is disabled. 

I chose to do it this way to improve readability and reduce indentation in the nested for loops on lines 254 - 264 in the `Gather` method. I originally had more logic within those nested for loops. But I felt it made the code too convoluted.

Let me know what you think.

plugins/inputs/sqlserver/sqlserver.go
coquagli: See my response above.

plugins/inputs/sqlserver/connectionstring.go
coquagli: While rather annoying, the `\` characters are the official format for SQL Server named instances. See here:
- https://docs.microsoft.com/en-us/previous-versions/sql/sql-server-2005/ms165614(v=sql.90)#sql-server-express-default-and-named-instances
- https://docs.microsoft.com/en-us/dotnet/framework/data/adonet/connection-string-syntax#connect-to-a-named-instance-of-sql-server


It's not a deal breaker to deviate from the format. But it is the format that users would expect to see.

plugins/inputs/sqlserver/connectionstring.go
coquagli: Looked into this more. Turns out the `Hostname()` function handles all this for me.

plugins/inputs/sqlserver/connectionstring.go
coquagli: Ooh, good catch. I used your suggestion below to fix that.

plugins/inputs/sqlserver/connectionstring.go
coquagli: Good suggestion.

plugins/inputs/sqlserver/connectionstring.go
coquagli: Done.

plugins/inputs/sqlserver/connectionstring.go
coquagli: Done.

plugins/inputs/sqlserver/connectionstring.go
coquagli: Good suggestion. Much cleaner.

plugins/inputs/sqlserver/connectionstring.go
coquagli: I've gone through all the `return`s and have attempted to clean them up to fix this Let me know if the current iteration still doesn't follow the guidelines.

coquagli:
Thanks for the review @srebhan! I've pushed another iteration to address your comments.

plugins/inputs/sqlserver/connectionstring.go
srebhan: Ok then please keep it and close this as resolved.

plugins/inputs/sqlserver/connectionstring.go
srebhan: Looks good to me. Please close all comments of this type as resolved. Sorry I don't have the permission to do so...

plugins/inputs/sqlserver/connectionstring.go
srebhan: Please close as resolved.

plugins/inputs/sqlserver/connectionstring.go
srebhan: Please close as resolved.

plugins/inputs/sqlserver/connectionstring.go
srebhan: Please close as resolved.

plugins/inputs/sqlserver/connectionstring.go
```diff
@@ -0,0 +1,100 @@
...
+	emptyServerName   = "<empty-server-name>"
+	emptyDatabaseName = "<empty-database-name>"
+)
+
+// getConnectionIdentifiers returns the sqlInstance and databaseName from the given connection string.
+// The name of the server is returned as-is in the connection string
+// If the connection string could not be parsed or sqlInstance/databaseName were not present, a placeholder value is returned
+func getConnectionIdentifiers(connectionString string) (sqlInstance string, databaseName string) {
+	if len(connectionString) == 0 {
+		return "", ""
```
srebhan, line 18: IIRC you want `emptyServerName` and `emptyDatabaseName` here!?!? :-)

plugins/inputs/sqlserver/connectionstring.go
srebhan: Please close as resolved.

plugins/inputs/sqlserver/connectionstring.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+// parseConnectionStringURL parses a URL-formatted connection string and returns the SQL instance and database name
+func parseConnectionStringURL(connectionString string) (sqlInstance string, databaseName string) {
+	sqlInstance = emptyServerName
+	databaseName = emptyDatabaseName
+
+	u, err := url.Parse(connectionString)
+	if err != nil {
+		return sqlInstance, databaseName
```
srebhan, line 80: You _could_ use `return emptyServerName, emptyDatabaseName)` here to make clear you are returning the defaults, but as it is a matter of taste I leave this decision to you.

plugins/inputs/sqlserver/connectionstring.go
srebhan: Oh one more question, if the path has more than one "segment" they are probably separated by `/` aren't they? Do you want to replace those with backslash also?

plugins/inputs/sqlserver/connectionstring.go
srebhan: Please close as resolved.

plugins/inputs/sqlserver/connectionstring.go
srebhan: Please close as resolved.

plugins/inputs/sqlserver/connectionstring.go
srebhan: Please close as resolved.

plugins/inputs/sqlserver/connectionstring.go
srebhan: Perfect solution! Please close as resolved.

plugins/inputs/sqlserver/connectionstring.go
srebhan: Please close as resolved.

plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -226,18 +248,23 @@ func (s *SQLServer) Gather(acc telegraf.Accumulator) error {
...
+	var healthMetrics = make(map[string]*HealthMetric)
 
 	for _, serv := range s.Servers {
 		for _, query := range s.queries {
 			wg.Add(1)
 			go func(serv string, query Query) {
 				defer wg.Done()
-				acc.AddError(s.gatherServer(serv, query, acc))
+				queryError := s.gatherServer(serv, query, acc)
+				s.gatherHealth(healthMetrics, serv, queryError, &mutex)
```
srebhan, line 260: How about
```suggestion
			if s.HealthMetric {
				mutex.Lock()
				s.gatherHealth(healthMetrics, serv, queryError)
				mutex.Unlock()
			}
```
so it's easier for people to see that this operation depends on the switch. You might want to do the same for `s.accHealth` but this keep the functions more simple. It's only a suggestion...

plugins/inputs/sqlserver/sqlserver.go
srebhan: Right I confused this with `healthMetrics`. As I outlined above, we might also want to do this check in the level above so everybody directly sees that the gathering here is optional but I can live with both solutions...

plugins/inputs/sqlserver/sqlserver.go
srebhan: As I said above, we might want to do this one level up. Furthermore, you don't need this here at all (besides for documentation) as the gather method above only fills the `healthMetrics` map in case the condition is true. So the loop below will not be run.

plugins/inputs/sqlserver/sqlserver.go
srebhan: Please close as resolved.

plugins/inputs/sqlserver/sqlserver.go
srebhan: Fair enough. Please close as resolved.

plugins/inputs/sqlserver/sqlserver_test.go
srebhan: Please close as resolved.

srebhan (commented):
The only remaining thing you _have to_ look into is the first `return` in `getConnectionIdentifiers()`, not sure but you might want to return the defaults!?

All other comments are suggestions or a matter of taste, so please change or keep as you like. Just leave me a comment on those what your decision is and resolve the resolved comments. :-)

plugins/inputs/sqlserver/connectionstring.go
coquagli: Good point. It's unclear how to handle that situation. The SQL Server driver for Go just ignores any extra segments, so that's also what I did here. (see https://github.com/denisenkom/go-mssqldb/blob/master/conn_str.go#L313).

plugins/inputs/sqlserver/sqlserver.go
coquagli: I like it. Done.

plugins/inputs/sqlserver/connectionstring.go
coquagli: Done.

plugins/inputs/sqlserver/connectionstring.go
coquagli: Good catch.

plugins/inputs/sqlserver/sqlserver.go
coquagli: The way you suggested below works. I'll go with that.

plugins/inputs/sqlserver/sqlserver.go
coquagli: The way you suggested below works. I'll go with that. 

I think keeping the `if s.HealthMetric` on line 274 helps with readability. It looks visually similar to the same check on line 261, so it's clear they are related bits of code.

coquagli:
Hey, @srebhan thanks for the great suggestions. I pushed some more changes.

srebhan (approved):
Looks perfect to me. Thank you very much for the nice collaboration!

Jagularr (approved):

coquagli:
@sjwang90 @srebhan @Jagularr @ssoroka what else is needed to merge this?

ssoroka (approved):
ok, looks good. Just needs a merge conflict fixed.

coquagli:
Should be good now @ssoroka.

srebhan (approved):

Issue #7581, opened 2020-05-26
fatal error: concurrent map writes with MQTT plugin


phoenix147:
Hi!

I'm using testing Telegraf with the mqtt plugin and the latest InfluxDB 2.0 beta.
It works fine for several ours, but after that it crashes with "fatal error: concurrent map writes"
It seems related to the MQTT Plugin and also occurs often after it has some timeouts writing to influxdb, but not always

### Relevant telegraf.conf:
```
[agent]
  omit_hostname = true

[[outputs.influxdb_v2]]
  urls = ["http://localhost:9999"]

  ## Token for authentication.
  token = "token omitted"

  ## Organization is the name of the organization you wish to write to; must exist.
  organization = "development"

  ## Destination bucket to write into.
  bucket = "stations"

[[processors.regex]]
  namepass = ["mqtt_consumer"]

  [[processors.regex.tags]]
    key = "topic"
    pattern = '^(\d+)\/(\d+)\/(\w+)$'
    replacement = "${1}"
    result_key = "station"
  [[processors.regex.tags]]
    key = "topic"
    pattern = '^(\d+)\/(\d+)\/(\w+)$'
    replacement = "${2}"
    result_key = "sensor"
  [[processors.regex.tags]]
    key = "topic"
    pattern = '^(\d+)\/(\d+)\/(\w+)$'
    replacement = "${3}"
    result_key = "measurement"
# station topic
  [[processors.regex.tags]]
    key = "topic"
    pattern = '^(\d+)\/(\w+)$'
    replacement = "${1}"
    result_key = "station"
  [[processors.regex.tags]]
    key = "topic"
    pattern = '^(\d+)\/(\w+)$'
    replacement = "${2}"
    result_key = "measurement"

[[inputs.mqtt_consumer]]
  servers = ["tcp://localhost:1883"]

  ## Topics that will be subscribed to.
  topics = [
    "#",
  ]
  topic_tag = "topic"
  data_format = "json"
  json_time_key = "ts"
  json_time_format = "2006-01-02T15:04:05Z07:00"

```

### System info:

Telegraf Version: 1.14.3
InfluxDB Version: 2.0beta 10
OS: Ubuntu 18.04.3

### Steps to reproduce:

1. Start InfluxDB and Telegraf
2. Wait for several hours

### Expected behavior:

Telegraf runs until stopped

### Actual behavior:

Telegraf crashes (see log)

### Additional info:
[...]

PR #8562, opened 2020-12-14
fix issue with mqtt concurrent map write

ssoroka:
closes https://github.com/influxdata/telegraf/issues/7581

### DIFF
--- a/plugins/inputs/mqtt_consumer/mqtt_consumer.go
+++ b/plugins/inputs/mqtt_consumer/mqtt_consumer.go
@@ -5,9 +5,10 @@ import (
 	"errors"
 	"fmt"
 	"strings"
+	"sync"
 	"time"
 
-	"github.com/eclipse/paho.mqtt.golang"
+	mqtt "github.com/eclipse/paho.mqtt.golang"
 	"github.com/influxdata/telegraf"
 	"github.com/influxdata/telegraf/internal"
 	"github.com/influxdata/telegraf/plugins/common/tls"
@@ -69,6 +70,7 @@ type MQTTConsumer struct {
 	state         ConnectionState
 	sem           semaphore
 	messages      map[telegraf.TrackingID]bool
+	messagesMutex sync.RWMutex
 	topicTag      string
 
 	ctx    context.Context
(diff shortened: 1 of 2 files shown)


plugins/inputs/mqtt_consumer/mqtt_consumer.go
```diff
@@ -258,13 +262,17 @@ func (m *MQTTConsumer) recvMessage(c mqtt.Client, msg mqtt.Message) {
...
 			_, ok := m.messages[track.ID()]
+			m.messagesMutex.RUnlock()
 			if !ok {
 				// Added by a previous connection
 				continue
 			}
 			// No ack, MQTT does not support durable handling
+			m.messagesMutex.Lock()
 			delete(m.messages, track.ID())
+			m.messagesMutex.Unlock()
```
srebhan, line 275: Doing
```suggestion
			m.messagesMutex.RLock()
			_, ok := m.messages[track.ID()]
			if ok {
				// No ack, MQTT does not support durable handling
				delete(m.messages, track.ID())
			}
			m.messagesMutex.Unlock()
```
might also work but is holding the mutex a bit longer... On the other hand it prevents that exactly this ID is deleted in the meantime...

srebhan (approved):
One suggestion but the code looks good!

plugins/inputs/mqtt_consumer/mqtt_consumer_test.go
```diff
@@ -4,7 +4,7 @@ import (
 	"testing"
 	"time"
 
-	"github.com/eclipse/paho.mqtt.golang"
+	mqtt "github.com/eclipse/paho.mqtt.golang"
```
srebhan, line 7: Could you please elaborate on why this is necessary!?

srebhan (approved):
Looks good now, but I'm not sure why you alias the mqtt library in the test...

plugins/inputs/mqtt_consumer/mqtt_consumer_test.go
ssoroka: this was automatically added by the formatting library, (I think goreturns, in this case). It doesn't like the implicit import of the mqtt package via a path that doesn't match the package name.

srebhan: Ok so this was my "TIL". :-) Thanks for the explanation!

Issue #8566, opened 2020-12-15
Add cmdline as a field in procstat plugin


prashanthjbabu:
## Feature Request

Opening a feature request kicks off a discussion.

### Proposal:
    Adding `cmdline` as a field in procstat plugin
### Current behavior:
 `cmdline` can only be added as a tag
### Desired behavior:
 option to add `cmdline` as a field or tag
### Use case: 
Adding `cmdline` as a tag can introduce a very high cardinality in databases like influxdb because if you're running 5-10 shell scripts or java programs which show as `shell` or `java` as`process_name` you are bound to need `cmdline` as mentioned here https://github.com/influxdata/telegraf/issues/1873 but it comes with a cost . Therefore just like `pid` there should be an option to allow `cmdline` as a field as well . So if we use `pattern` as a way to get all java programs in a system we can identify them in `cmdline` as a field without having to bear the cost of high cardinality by having `cmdline` as a tag . I'm talking about a case where there are tens of thousands of hosts .

reimda:
Closing in favor of using the converter plugin.  See https://github.com/influxdata/telegraf/pull/8567#issuecomment-763227148

PR #8567, opened 2020-12-15
procstat : support to add cmdline as a field

prashanthjbabu:
Addresses https://github.com/influxdata/telegraf/issues/8566
### Required for all PRs:

### DIFF
--- a/plugins/inputs/procstat/README.md
+++ b/plugins/inputs/procstat/README.md
@@ -41,7 +41,7 @@ Processes can be selected for monitoring using one of several methods:
   ## Field name prefix
   # prefix = ""
 
-  ## When true add the full cmdline as a tag.
+  ## When true add the full cmdline as a tag else add it as a field.
   # cmdline_tag = false
 
   ## Add the PID as a tag instead of as a field.  When collecting multiple
@@ -79,6 +79,7 @@ the `win_perf_counters` input plugin as a more mature alternative.
     - cgroup (when defined)
     - win_service (when defined)
   - fields:
+    - cmdline (when 'cmdline_tag' is false)
     - child_major_faults (int)
     - child_minor_faults (int)
     - created_at (int) [epoch in nanoseconds]
--- a/plugins/inputs/procstat/procstat.go
+++ b/plugins/inputs/procstat/procstat.go
@@ -178,12 +178,14 @@ func (p *Procstat) addMetric(proc Process, acc telegraf.Accumulator) {
 	}
 
 	//If cmd_line tag is true and it is not already set add cmdline as a tag
-	if p.CmdLineTag {
-		if _, ok := proc.Tags()["cmdline"]; !ok {
-			Cmdline, err := proc.Cmdline()
-			if err == nil {
+	Cmdline, err := proc.Cmdline()
+	if err == nil {
+		if p.CmdLineTag {
+			if _, ok := proc.Tags()["cmdline"]; !ok {
 				proc.Tags()["cmdline"] = Cmdline
 			}
+		} else {
+			fields["cmdline"] = Cmdline
 		}
 	}
 


plugins/inputs/procstat/README.md
```diff
@@ -41,7 +41,7 @@ Processes can be selected for monitoring using one of several methods:
   ## Field name prefix
   # prefix = ""
 
-  ## When true add the full cmdline as a tag.
+  ## When true add the full cmdline as a tag else add it as a field.
```
srebhan, line 44: I think we cannot do it this way as you will alter the series for everyone of we merge as is. That is, you automagically add a field for everyone having this set to false. We need to stay backward compatible as much as we can...

That being said, you might want to add a new boolean option`cmdline_field`...

srebhan (changes requested):
To add this feature you should use a new option instead of hijacking `cmdline_tags`. The code need to follow this change but looks good otherwise.

plugins/inputs/procstat/README.md
prashanthjbabu: @srebhan  Thanks for your review ! I do feel it is backward compatible for those who have this set as false as it would not affect the existing set of tags and only create a new field . This is similar to how we've added extra fields to procstat plugins in the past . I dont see it breaking anything . Also wrt hijacking of `cmdline_tags` , this works very similar to an existing boolean in procstat called `pid_tag` ( Reference : https://github.com/influxdata/telegraf/blob/master/plugins/inputs/procstat/procstat.go#L35 ) . If `pid_tag` is true , pid would come up as a tag , else pid would come up as a field . I'm trying to follow the same convention using the same boolean!

srebhan: Well an influxdb series is defined by the measurement, the tag-set and **the set of field keys** (aka field-names)! This way you **always** create new series with your implementation if `cmdline_tags` is false. 

@reimda  or @ssoroka could you please check how to proceed in this case!?

reimda: I would also prefer not forcing users to take a new field if they don't want cmdline_tag.  I suspect plenty of people won't want either.

prashanthjbabu: @srebhan  @reimda  a lot of fields in the proctstat plugin were added much later after the plugin was introduced without no new additional flag to allow users to enable/disable the field . What would you suggest we do to proceed further?

srebhan: The current option `cmdline_tag` and the other option `pid_tag` do only state if this should be added as a tag. My point here is that, if  `cmdline_tag` is false, you automatically add a field. To me this is unintuitive and unexpected. I think the best way would be to allow specifying a list of tags and a list of additional fields in the form
```
additional_tags = ["pid", "cmdline"]
additional_fields = ["cmdline"]
```
with a default that matches the current state i.e.
```
# additional_tags = []
# additional_fields = []
```
This is extensible and allows to keep the output constant at the same time. Remember you have to keep the current options to be backward compatible. This is my take on the matter.

prashanthjbabu: @srebhan  That is incorrect . If `pid_tag` is false it will not be added as a tag and it will automatically be added as a field . Here's the reference https://github.com/influxdata/telegraf/blob/master/plugins/inputs/procstat/procstat.go#L175 . I'm following the same behavior as `pid_tag` for `cmdline_tag` .

srebhan: @prashanthjbabu I see your point, but it doesn't address the fact that you unexpectedly add a new field for existing users with existing configs. There are IMO two ways to proceed:

1. Add an extra boolean flag `cmdline_field` and default this to `false` avoiding to overload the `cmdline_tag` and changing outputs for existing users.
2. Add a `additional_fields = []` option where you can (future proof) add a list of new additional fields as they come up. It is really not that hard and is my preferred solution. The first available entry would then be `cmdline`.

As stated 2 is my preferred option. Let me know if you need help with implementing it.

prashanthjbabu: @srebhan  I was under the impression that it is normal practice to include new fields to plugins as and when required ( without the addition of a flag to support it )  . I've seen this in quite a lot of plugins where fields are added later on as they deemed fit . Here are a few examples : 
1.  https://github.com/influxdata/telegraf/commit/219bf93566515e46b98a91fee309e63c5f818e4b [docker plugin ]
2. https://github.com/influxdata/telegraf/commit/9f4752ba123d9c50180b54248cc1448f1948bb87 [ docker plugin ]
3. https://github.com/influxdata/telegraf/commit/9c3af1e6ac1c6e619cef0b56fa53798c4f9369d5 [ procstat plugin ]
4. https://github.com/influxdata/telegraf/commit/d727a6f85cdf7ac33e3bd66fed34da25b2355f84 [ mem plugin ]
5. https://github.com/influxdata/telegraf/commit/c6b8947c154da75d82984f01ae8323d2f315c188 [ diskio plugin ]

These are just a few to quote as an example  , where new fields were added without addition of any new configs . Is this a new practice that we're following going forward for all plugins? If so, and if this is something that we're strictly following going forward globally for all plugins, then the suggestions you mentioned make complete sense.

srebhan: Let me discuss this with the core-devs. I'll come back to you, but I guess it will be earliest in mid of January... :-(

prashanthjbabu: @srebhan  Sure no worries . Thank you!

reimda:
Hi @prashanthjbabu.  @srebhan asked me to take a closer look at this PR.

One of the converter processor's purposes is to change tags to fields.  It is the recommended solution for cases like this.

I configured telegraf locally to use procstat and the following config to change the procstat cmdline tag to a field:
[[processors.converter]]
  namepass = ["procstat"]
  [processors.converter.tags]
    string = ["cmdline"]

This is a simpler solution that avoids the code change and also the concerns about backward compatibility that @srebhan and I shared earlier.

Please try this and let us know how it goes.

prashanthjbabu:
@reimda thanks for sharing this ! Yes this is the best way to do it rather than changing code . Will work with this and close this PR .

Issue #831, opened 2016-03-11
Oracle input plugin for database performance


abarban:
I'm looking for the ability to have statistical data / performace by Oracle.

I am analyzing two solutions:

A. Go live using driver as:
- https://github.com/rana/ora
- https://github.com/mattn/go-oci8

B. by a demon read in Java that expose the data in JSON

Both have positive and negative sides, the A may have a long latency due to the connection to be restored each time, the B would be an external object to Telegraf.

As always willing to test and development.

sparrc:
From the Telegraf point of view, A would be the way to go

menardorama:
By an Oracle Point of view, having frequent connection/disconnection is a bad practice for the listener.

abarban:
The reason that move me to try a little demon with java, single connection and light SQL!
there is a project like https://sourceforge.net/projects/ashv/ that can be a start point.
But I think that can be so useful try both solutions.

how can I be most useful?

menardorama:
If you want to try the A solution, you can have a look at what I did on Postgres #792 

Moving it to Oracle wouldn't be a big deal

I guess that you are trying to build a dashboard based on ASH ?

abarban:
First milestone is to have a dashboard with Top Activity graphs on multiple Oracle instance, without detail on sessions and query.

I'm thinking about live monitoring with a way to build dashboard for applications stack (accelerator, application, db).

sparrc:
going with A, why would it have to do frequent connect/disconnect? couldn't the plugin just cache and hold onto a connection?

abarban:
Sure! I must try how it could work on Go! ;) I'm a junior on it.

menardorama:
Didn't know it was possible.

Is it just by not closing the connection and add a defer close ?

sparrc:
just making the connection a field on the plugin struct

menardorama:
ok it's a bit out of the subject of the issue, but  do you have a link of an example ?

bbczeuz:
I started writing a universal DB Query plugin. Just to avoid duplication: Has anybody else started working on such a plugin?

It's based on golang database/sql, so Oracle, Postgres, Mysql should be supported (tested: Oracle XE 12.1 via go-oci8). It supports dynamic tagging based on column names, multiple queries.
But it's my first golang project, so the design might not be very advanced (I'm a c++ guy). It's very alpha now, but it works for my case. I'll test larger datasets in the next weeks.

sparrc:
[...]

PR #8568, opened 2020-12-15
Oracle db input plugin

unflag:
This plugin runs queries on Oracle DB and creates metrics from received data. It has abilities to control which columns would be tags/fields, to switch schema and to set query execution timeout. Logging is handy and can be really helpful when multiple plugin instances run on different databases and executes a lot of queries. It uses [godror](https://github.com/godror/godror) driver, so free [Oracle Client libraries](https://www.oracle.com/database/technologies/instant-client/downloads.html) are required at run time. Tests could be run without it, because they use [go-sqlmock](github.com/DATA-DOG/go-sqlmock) package. I already use it several month in production and now hope it ready to be helpful to someone else.

Also closes https://github.com/influxdata/telegraf/issues/831

### Required for all PRs:


sjwang90:
Besides being just OracleDB focused, can you state the discrepancies between your PR and the GenericSQL Plugin PR #2785?

unflag:
> Besides being just OracleDB focused, can you state the discrepancies between your PR and the GenericSQL Plugin PR #2785?

Sure. 
First of all, this plugin doesn't have to be specifically compiled to support Oracle, it just uses external oracle libs at run time which makes it easier to use. 

It gathers database metadata and place it in tags automatically, so it makes configuration less verbose and error prone, also  metadata contains more data(because it OracleDB focused) compared to GenericSQL.

It supports per-query execution timeout, which can be helpful if some query starts being slow and takes all plugin interval time. From monitoring and alerting side it much more easy to investigate, when only one or two metrics lost instead of loosing all data and trying to understand why so many alerts rang.

And, as I can see in GenericSQL's code, it doesn't consider that Oracle can provide NUMBER data type, which is internally varchar and can store numbers which not fit in 2^64. In Go such values would have string type. There is not much we can do if we get too large numbers, of course, but I tried to save as much precision as possible by parsing string and using float64, int64 and uint64 types. GenericSQL, if I looked properly, would store such values as strings and it may be impossible to use with some outputs which doesn't accept fields with string values like opentsdb. Or to use it in math expressions. I hadn't enough time to test it or to look into the code more attentively, so I may be incorrect - please, point me to it.

unflag:
As far as I can see there is some cross compilation issue with 386 arch because of CGO. Is there any way to fix it?
And windows builder complains that there is no gcc in path, is it true? I haven't any clue where to start...

Hipska:
If this plugin is only returning metrics coming from SQL queries, I would prefer to improve GenericSQL instead. 

The Oracle plugin can be kept if this one also produces specific info of the Oracle DB, similar as the [MySQL](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/mysql) and [PostgreSQL](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/postgresql) plugins.

unflag:
Plugin generates metrics using sql queries only, but it also gathers some Oracle-specific runtime metadata to populate tags. It also has some Oracle-specific type conversion and use specific driver types. Unfortunately, I don't see a way how all of it can be added to existing GenericSQL logic.

Hipska:
The team will have a look on it and we will come back to you. Our preference is still towards the GenericSQL input plugin.

megastef:
A very interesting plugin. Many Oracle users would love it, including me. I was happy to see some news here after years of discussions. +1 to add it to telegraf.

srebhan:
Hey guys, sorry for being late to the party...

@unflag I do understand that you use certain Oracle-specific data. The whole problem with Oracle go-clients is that they all require CGO. You are using `godror` and in their README they are stating:

> At least Go 1.13 is required! Cgo is required, so cross-compilation is hard, and you cannot set CGO_ENABLED=0!

While Go 1.13 is not a problem, the CGO requirement is a problem! Telegraf will not have CGO enabled AFAIK! This also hinders the GenericSQL plugin to support OracleDBs. :-(

One way to circumvent that issue might be to add this as external plugin. @ssoroka would this be possible, an external plugin with CGO enabled?

ssoroka:
building as an external plugin with CGO is possible using inputs.execd. Unless we can find a native Go version of the driver, I think the external execd plugin is probably the best route. 
check out the [inputs.execd](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/execd) docs and the [Telegraf shim](https://github.com/influxdata/telegraf/tree/master/plugins/common/shim#telegraf-execd-go-shim) docs which together allows a plugin to run as a separate binary.

Let me know if you have any questions or the status of non-cgo drivers changes

Issue #3673, opened 2018-01-15
Feature Request - Add wildcard support to "Windows Services" plugin


glcx:
### Proposal:

Add wildcard support to "Windows Services" plugin. e.g. [ "Win*", "net*"]

### Current behavior:

To add service monitoring we need to either add services one per one or check the status of all the services.

### Desired behavior:

As per proposal

### Use case: [Why is this important (helps with prioritizing requests)]

Wildcard could permit more "automation-ready" plugin, and so permit the autodiscovery of new services with a specific pattern e.g. "Win*".

That way it could remove Manual & Automated modification of the config file.

dsvware:
@danielnelson do we have an ETA for this? Thanks in advance!

danielnelson:
No current ETA, any chance you could tackle this one?

dsvware:
Hi Daniel,

Sorry, I'm afraid I'm not a developer :(

sspaink:
Merged a pull request that resolves this issue. Let us know if you have any problems with this change! thanks.

PR #8575, opened 2020-12-16
Added Glob pattern matching for "Windows Services" plugin

sspaink:
### Required for all PRs:

This pull request addresses the issue  #3673
Updated the "Windows Services" plugin to allow glob patterns in the input config for including services.
I based the changes on the Docker plugin which also allows glob pattern matching.

Tested the changes locally and uploaded some information:
```
[[inputs.win_services]]
  ## Names of the services to monitor. Leave empty to monitor all the available services on the host
  service_names = [
    "Win*",
  ]
```
image

### DIFF
--- a/plugins/inputs/win_services/README.md
+++ b/plugins/inputs/win_services/README.md
@@ -8,10 +8,11 @@ Monitoring some services may require running Telegraf with administrator privile
 
 ```toml
 [[inputs.win_services]]
-  ## Names of the services to monitor. Leave empty to monitor all the available services on the host
+  ## Names of the services to monitor. Leave empty to monitor all the available services on the host. Globs accepted.
   service_names = [
     "LanmanServer",
     "TermService",
+    "Win*",
   ]
 ```
 
--- a/plugins/inputs/win_services/win_services.go
+++ b/plugins/inputs/win_services/win_services.go
@@ -7,6 +7,7 @@ import (
 	"os"
 
 	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/filter"
 	"github.com/influxdata/telegraf/plugins/inputs"
 	"golang.org/x/sys/windows/svc"
 	"golang.org/x/sys/windows/svc/mgr"
@@ -78,7 +79,7 @@ func (rmr *MgProvider) Connect() (WinServiceManager, error) {
 }
 
 var sampleConfig = `
-  ## Names of the services to monitor. Leave empty to monitor all the available services on the host
+  ## Names of the services to monitor. Leave empty to monitor all the available services on the host. Globs accepted.
   service_names = [
     "LanmanServer",
     "TermService",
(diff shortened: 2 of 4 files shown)


plugins/inputs/win_services/README.md
```diff
@@ -8,10 +8,11 @@ Monitoring some services may require running Telegraf with administrator privile
 
 ```toml
 [[inputs.win_services]]
-  ## Names of the services to monitor. Leave empty to monitor all the available services on the host
+  ## Names of the services to monitor. Leave empty to monitor all the available services on the host. Globs accepted.
```
ssoroka, line 11: I think globs are the technical name that isn't really used outside tech circles. Let's maybe refer to it as wildcards in the docs?

plugins/inputs/win_services/win_services.go
```diff
@@ -117,7 +122,16 @@ func (m *WinServices) Gather(acc telegraf.Accumulator) error {
...
 
-	serviceNames, err := listServices(scmgr, m.ServiceNames)
+	// Create label filters if not already created
+	if !m.filtersCreated {
+		err := m.createServicesFilter()
+		if err != nil {
+			return err
+		}
+		m.filtersCreated = true
+	}
```
ssoroka, line 132: let's move this to an Init() function as it only has to happen once, and then you don't need the boolean guard around it.

plugins/inputs/win_services/win_services.go
```diff
@@ -151,17 +165,30 @@ func (m *WinServices) Gather(acc telegraf.Accumulator) error {
...
-	if len(userServices) != 0 {
-		return userServices, nil
+func (m *WinServices) createServicesFilter() error {
+	filter, err := filter.NewIncludeExcludeFilter(m.ServiceNames, nil)
+	if err != nil {
+		return err
 	}
+	m.servicesFilter = filter
+	return nil
+}
```
ssoroka, line 175: can probably collapse this all into one line, might not need a function for it.

plugins/inputs/win_services/README.md
sspaink: I agree, the word "wildcard" would probably be easier to understand. Although the example config that Telegraf produces does use the word "glob" multiple times, do you think we should make an issue to update it as well? 
Example: https://github.com/influxdata/telegraf/blob/7d3b7fc2f9d379aa21d09eb3e905e644e876be6f/etc/telegraf.conf#L2949

ssoroka: ok perhaps it's a larger discussion. feel free to ignore

plugins/inputs/win_services/win_services.go
sspaink: Updated like suggested, Thanks!

plugins/inputs/win_services/README.md
sspaink: ignoring for now then.

plugins/inputs/win_services/win_services.go
sspaink: Updated like suggested, thanks!

ssoroka (approved):

PR #8576, opened 2020-12-16
docs: Add list of 3rd party builds to the README

mhall119:
### Required for all PRs:

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,17 @@ These builds are generated from the master branch:
 - [telegraf-nightly.x86_64.rpm](https://dl.influxdata.com/telegraf/nightlies/telegraf-nightly.x86_64.rpm)
 - [telegraf-static-nightly_linux_amd64.tar.gz](https://dl.influxdata.com/telegraf/nightlies/telegraf-static-nightly_linux_amd64.tar.gz)
 
+### 3rd Party Builds
+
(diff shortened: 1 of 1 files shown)


ssoroka:
My only thoughts on this is that there are a lot of others. eg, homebrew for OSX. I do feel a bit like we are recommending these, and I have no idea what quality they are.

sjwang90:
I like this. I don't necessarily think it seems like we're recommending it - if we want to add another note in the header `Not supported by InfluxData`?. @ssoroka How exhaustive is this list to just add of it?

sspaink:
My concern is that these links will be hard to maintain because no one on the team uses them and could become unsupported without us knowing. Maybe there is a better place to share this information, @popey do you maybe have a suggestion on this?

popey:
I think this is great, and a common pattern on other open source projects. It's reasonable for people to repackage and distribute our software, and it's equally reasonable for us to describe them as unsupported. If a support request comes in, we can ask to confirm that the issue is reproducible with our official builds before proceeding.

If the wording is clear that these are third party, unsupported builds, then I am happy for them to be there. I also thing we can add an item in our calendar to check them once in a while, in terms of their existence, not functionality or fit. If - for example - one disappears, or is no longer up to date, we can remove it from the readme.

sspaink (approved):
@popey makes a good case, if this is common amongst other projects then I am good with merging.

PR #8587, opened 2020-12-17
Worktable workfile stats

m82labs:
### Required for all PRs:

This is a small PR to add some new stats to help in troubleshooting tempdb contention from sort/hash spills. I have added two new metrics to the `sqlServerPerformanceCounters` query:
- `Workfiles Created/sec`
- `Worktables Created/sec`

I also noticed some mertrics from the V2 query that seemed to be omitted when the queries were split out and added them back:
- `Distributed Query`
- `DTC calls`
- `Query Store CPU usage`


m82labs:
I might need a review from @denzilribeiro on this one. I don't think the metrics above were removed on purpose, but I don't honestly know.

denzilribeiro (approved):
Mark, can you add these to the sqlAzureDBPerformanceCounters  and sqlAzureMIPerformanceCounters  as well?   Did these get dropped along the way?

m82labs:
I'll add them. No the worktables/files stats were never there, but the Query Store CPU and the DTC related metrics were at one point.

m82labs:
All set @denzilribeiro.

Jagularr (approved):

denzilribeiro (approved):

m82labs:
I haven't contributed in a while, will this automatically get merged in at some point in the future or do I have to do something else? Thanks!

m82labs:
@denzilribeiro I also noticed all the resource governor metrics were removed when the queries were split out. Do you know if the normal RG DMVs exist on all versions (Azure SQL DB, Managed instances, etc.)?

Specifically in the perf counter query I was inserting some "fake" metrics via a query: 
```SET  @SqlStatement = N'
SELECT
	 ''SQLServer:Workload Group Stats'' AS [object]
	,[counter]
	,[instance]
	,CAST(vs.[value] AS BIGINT) AS [value]
	,1
FROM
(
    SELECT
		 rgwg.name AS instance
		,rgwg.total_request_count AS [Request Count]
		,rgwg.total_queued_request_count AS [Queued Request Count]
		,rgwg.total_cpu_limit_violation_count AS [CPU Limit Violation Count]
		,rgwg.total_cpu_usage_ms AS [CPU Usage (time)]
		,rgwg.total_lock_wait_count AS [Lock Wait Count]
		,rgwg.total_lock_wait_time_ms AS [Lock Wait Time]
		,rgwg.total_reduced_memgrant_count AS [Reduced Memory Grant Count]
		' + @Columns + N'
	FROM sys.[dm_resource_governor_workload_groups] AS rgwg
	INNER JOIN sys.[dm_resource_governor_resource_pools] AS rgrp
		ON rgwg.[pool_id] = rgrp.[pool_id]
) AS rg
UNPIVOT (
    value FOR counter IN ( [Request Count], [Queued Request Count], [CPU Limit Violation Count], [CPU Usage (time)], [Lock Wait Count], [Lock Wait Time], [Reduced Memory Grant Count] ' + @PivotColumns + N')
) AS vs'

INSERT INTO @PCounters
EXEC( @SqlStatement )```

I can add it to this same PR unless there were compatibility issues.

denzilribeiro:
@m82labs  semantics of RG are different on SQL DB (MI is the same as on-prem). 
Why don't you get what you can get from the actual counters rather than simulated? Aka perfmon Workload Group Stats Object ? It can be confusing otherwise right?
https://docs.microsoft.com/en-us/sql/relational-databases/performance-monitor/sql-server-workload-group-stats-object?view=sql-server-ver15 
If it is just for backward compat, then no need to add for MI.

denzilribeiro (changes requested):
Can you remove these 2 from the SQL DB portion ( given no DTC support)- sorry I missed that

,'Distributed Query'
,'DTC calls'

m82labs:
> @m82labs semantics of RG are different on SQL DB (MI is the same as on-prem).
> Why don't you get what you can get from the actual counters rather than simulated? Aka perfmon Workload Group Stats Object ? It can be confusing otherwise right?
> https://docs.microsoft.com/en-us/sql/relational-databases/performance-monitor/sql-server-workload-group-stats-object?view=sql-server-ver15
> If it is just for backward compat, then no need to add for MI.

You have made me re-evaluate how much I actually need the few extra metrics the DMVs provide, I don't think I need them. :)

m82labs:
Is there anything I need to do to get these changes merged @Jagularr ?

denzilribeiro (approved):

sspaink (approved):
lgtm. @m82labs sorry for the slow response, thanks for waiting!

PR #8590, opened 2020-12-17
[ethtool] interface_up field added

zak-pawel:
### Required for all PRs:

`interface_up` field added, cosmetic changes


reimda (approved):

plugins/inputs/ethtool/ethtool_linux.go
```diff
@@ -80,19 +77,27 @@ func (e *Ethtool) gatherEthtoolStats(iface net.Interface, acc telegraf.Accumulat
...
 }
 
+func (e *Ethtool) interfaceUp(iface net.Interface) int {
+	interfaceUp := 0
+	if iface.Flags&net.FlagUp != 0 {
+		interfaceUp = 1
+	}
+	return interfaceUp
+}
+
```
srebhan, line 95: Isn't a boolean more appropriate given that you have only two states?
```suggestion
func (e *Ethtool) interfaceUp(iface net.Interface) bool {
	return (iface.Flags&net.FlagUp) != 0
}

```

srebhan (approved):
I suggest to switch the `interface_up` field to a boolean (suggested change is in the code) which seems more natural to me. But I can also live with the `0`/`1` setup.
So approved in any case. :-)

plugins/inputs/ethtool/ethtool_linux.go
Hipska: I also think this is more logical. What do you think @zak-pawel?

Hipska (approved):

ssoroka (approved):

PR #6713, opened 2019-11-24
Add service and save connection pools

nwneisen:
### Required for all PRs:

- [x] Signed [CLA](https://influxdata.com/community/cla/).
- [ ] Associated README.md updated.
- [x] Has appropriate unit tests.

Closes #5898

PR #8596, opened 2020-12-18
input/sqlserver: Add service and save connection pools

### LINKED ISSUE #5898: Reuse connections from connection pool in sqlserver input
### Proposal:

The sqlserver input should reuse connections from the connection pool.

### Current behavior:

New connection pool is created every interval.

### Desired behavior:

Hang on to the connection pool and reuse.

### Use case: 

Should be a lighter resource load on Telegraf and the server.

See also https://github.com/influxdata/telegraf/issues/3485#issuecomment-494887920

mjiderhamn:
### Required for all PRs:

Since the issue persists while #6713 was closed by author @nwneisen , I'm opening up a new PR to fix #5898.


helenosheaa:
Thanks for reopening @mjiderhamn is this something you or @nwneisen are currently working on?

mjiderhamn:
I'd be willing to try to take this to a mergeable state.

I have rebased onto `master` and made CircleCI happy. Anything else you're thinking about @helenosheaa ?

helenosheaa:
@mjiderhamn sounds good, that's great! Nothing currently on my side.

mjiderhamn:
Any news on getting this merged, @helenosheaa

helenosheaa:
@mjiderhamn it looks like it has conflicts with master, if you can address those.

After that it needs to be reviewed, I'll take a look and request another reviewer :)

mjiderhamn:
@helenosheaa `master` conflict resolved

helenosheaa:
@denzilribeiro or @Trovalo would either of you mind taking a look at this?

sjwang90:
@avinash-nigam would you or anyone else on your team be able to take a review on this PR?

plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -214,43 +212,56 @@ func initQueries(s *SQLServer) error {
...
+			}(pool, query)
 		}
 	}
 
+	wg.Wait()
+	return nil
+}
+
+// Start initialize a list of connection pools
+func (s *SQLServer) Start(acc telegraf.Accumulator) error {
```
masree, line 232: I see this func only used by unit test and not called from plugin. Is that intenional?

plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -15,14 +15,14 @@ import (
...
-	ExcludeQuery  []string `toml:"exclude_query"`
-	queries       MapQuery
-	isInitialized bool
+	Servers      []string `toml:"servers"`
+	QueryVersion int      `toml:"query_version"`
+	AzureDB      bool     `toml:"azuredb"`
+	DatabaseType string   `toml:"database_type"`
+	IncludeQuery []string `toml:"include_query"`
+	ExcludeQuery []string `toml:"exclude_query"`
+	pools        []*sql.DB
```
masree, line 24: AFAIK, for connection pooling, we usually have a default pool size and add connection to pool if the number of connections is less than pool size. was it there a specific reason of not having that mechanism? Is it safe to have any number?

masree:
Looks good to me!

Jagularr (commented):
Looks good! Would be curious to better understand how masree's comment might effect this but other than that this seems to make sense.

mjiderhamn:
I did leave a reply to @masree , which is marked as "Pending" :-/ 

I wrote
> From [the docs](https://golang.org/pkg/database/sql/#DB) it seems to me you can only set max open vs idle, so I'm not sure what you mean with default pool size?
> 
> The idea is that each pool will only ever have 1 connection in them, since there is no concurrency involved. Yes, there should be 1 pool per connection string/server. The goal is to avoid opening and closing connections over and over for each interval, the way it currently does.

helenosheaa:
@mjiderhamn could you address the merge conflicts again, then we can merge. Thanks for your work on this.

masree:
> I did leave a reply to @masree , which is marked as "Pending" :-/
> 
> I wrote
> 
> > From [the docs](https://golang.org/pkg/database/sql/#DB) it seems to me you can only set max open vs idle, so I'm not sure what you mean with default pool size?
> > The idea is that each pool will only ever have 1 connection in them, since there is no concurrency involved. Yes, there should be 1 pool per connection string/server. The goal is to avoid opening and closing connections over and over for each interval, the way it currently does.

Sorry, I meant to say if we need to set max open and idle connections. Thanks for clarifying.

mjiderhamn:
@helenosheaa I fixed the merge issues, but I don't understand what's up with CircleCI. If I run `make tidy` in an `quay.io/influxdb/telegraf-ci:1.15.5` container it is successful.

ssoroka (approved):

helenosheaa (approved):

PR #8609, opened 2020-12-21
Use the 'measurement' json field from the particle webhook event as the measurement name, or if it's blank, use the 'name' field of the event's json.

### LINKED ISSUE #6879: Particle.io webhook plugin wrong measurement name
IMHO webhook plugin writes data to wrong measurement in InfluxDb. Adding 
`{
    "measurement": "your_measurement_name"
}`
have no effect since in plugin code measurement string from json is written to event.database (but what for?):

> type event struct {
> 	Name        string `json:"event"`
> 	Data        data   `json:"data"`
> 	TTL         int    `json:"ttl"`
> 	PublishedAt string `json:"published_at"`
> 	Database    string `json:"measurement"`
> }

After that when submitting data to telegraf with eventHandler () as a measurement name there's event.name which is name of the event which triggered webhook.

As a solution I propose:

> type event struct {
> 	Name        string `json:"event"`
> 	Data        data   `json:"data"`
> 	TTL         int    `json:"ttl"`
> 	PublishedAt string `json:"published_at"`
> 	Measurement    string `json:"measurement"`
> }

and in eventHandler () (line 62): 

> 	rb.acc.AddFields(e.Measurement, e.Data.Fields, e.Data.Tags, pTime)

Jagularr:
resolves #6879

### DIFF
--- a/plugins/inputs/webhooks/particle/particle_webhooks.go
+++ b/plugins/inputs/webhooks/particle/particle_webhooks.go
@@ -14,7 +14,7 @@ type event struct {
 	Data        data   `json:"data"`
 	TTL         int    `json:"ttl"`
 	PublishedAt string `json:"published_at"`
-	Database    string `json:"measurement"`
+	Measurement string `json:"measurement"`
 }
 
 type data struct {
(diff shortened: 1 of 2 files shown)


helenosheaa (commented):
This looks good to me

PR #8613, opened 2020-12-22
Support exclamation mark to create non-matching list in tail plugin

### LINKED ISSUE #7419: Glob Pattern at inputs.tail not working properly
### Relevant telegraf.conf:

```
[[inputs.tail]]
  ## File names or a pattern to tail.
  ## These accept standard unix glob matching rules, but with the addition of
  ##   "/var/log/**.log"  -> recursively find all .log files in /var/log
  ##   "/var/log/*/*.log" -> find all .log files with a parent dir in /var/log
  ##   "/var/log/apache.log" -> just tail the apache log file
  ##
  ## See https://github.com/gobwas/glob for more examples
  ##
  files = ["/home/test/2020041[!1-8]*"]

  ## Read file from beginning.
  # from_beginning = false

  ## Whether file is a named pipe
  # pipe = false
   grok_patterns = ["'''%{GREEDYDATA:message}'''"]
   data_format = "grok"

```

### System info:
Telegraf 1.14.1

### Steps to reproduce:

1. create files:
touch /home/test/202004182350
touch /home/test/202004190000
touch /home/test/202004190010

2. start telegraf

### Expected behavior:
telegraf should open the files 202004190000 and 202004190010 

### Actual behavior:
telegraf opens the file 202004182350
2020-04-28T06:43:25Z D! [inputs.tail] Tail added for "/home/test/202004182350"

### Additional info:
I think there is an problem with the glob parser, all glob testing-tools say that the glob-pattern is correct:

https://www.digitalocean.com/community/tools/glob?comments=true&glob=%2Fhome%2Ftest%2F2020041%5B%211-8%5D%2A&matches=false&tests=%2Fhome%2Ftest%2F202004180010&tests=%2Fhome%2Ftest%2F202004190010

sspaink:
### Required for all PRs:

Fixes #7419 by converting `[!` to `[^` which allows both characters to be used.

The Tail plugin depends on the standard Go function [filepath.Glob](https://golang.org/pkg/path/filepath/#Glob) which uses [filepath.Match](https://golang.org/pkg/path/filepath/#Match) and this function does not support `!` only `^`, evidently this was an oversight during implementation: https://github.com/golang/go/issues/41394. Hopefully in the second version of the filepath package both characters will be supported natively and we can remove this change.

Also updated the relevant unit test in globpath_test.go to use table driven tests.

Thanks!


Jagularr (approved):
Makes sense to me.

internal/globpath/globpath.go
```diff
@@ -45,6 +45,7 @@ func Compile(path string) (*GlobPath, error) {
 // If it's a static path, returns path.
 // All returned path will have the host platform separator.
 func (g *GlobPath) Match() []string {
+	g.path = strings.Replace(g.path, "[!", "[^", -1)
```
srebhan, line 48: Can we please use
```suggestion
	g.path = strings.ReplaceAll(g.path, "[!", "[^")
```

srebhan (commented):
Would love to use `ReplaceAll()` instead of the `Replace()` function with `-1` for readability. Furthermore, `\\[!` is a valid pattern matching literal `[!` but is now replaced to `\\[^` with your change. Even though I think it is unlikely to be used in the wild, the probability is not zero. Could you please check this corner-case!?

sspaink:
@srebhan thanks for the review! So because Go doesn't support [negative lookahead](https://stackoverflow.com/questions/26771592/negative-look-ahead-in-go-regular-expressions) in regular expressions I am having a hard time thinking of a clean way to check for this edge case, do you have any suggestions?

srebhan:
Well 
```
	re := regexp.MustCompile(`(?:[^\\])(\[!)`)
	re.ReplaceAllString(s, "[^")
```
should work according to [my test](https://play.golang.org/p/ei1qPz6fqD9)...

sspaink:
@srebhan So I couldn't get the regex to work exactly like we would want to. I did find another open source project [doublestar](https://github.com/bmatcuk/doublestar) that supports both double star and !/^ for negation, it does remove the custom logic that globpath has for handling doublestar which is nice. What do you think?

@Jagularr the code has changed quite a bit from when you reviewed it after my latest change in case you wanted to look at it again as well.

plugins/inputs/phpfpm/phpfpm_test.go
```diff
@@ -327,7 +327,7 @@ func TestPhpFpmGeneratesMetrics_Throw_Error_When_Socket_Path_Is_Invalid(t *testi
 
 	err = acc.GatherError(r.Gather)
 	require.Error(t, err)
-	assert.Equal(t, `dial unix /tmp/invalid.sock: connect: no such file or directory`, err.Error())
+	assert.Equal(t, `socket doesn't exist "/tmp/invalid.sock": <nil>`, err.Error())
```
ssoroka, line 330: should not be nil. remove error from internal error format.

plugins/inputs/tail/README.md
```diff
@@ -29,8 +29,9 @@ The plugin expects messages in one of the
   ##   "/var/log/**.log"  -> recursively find all .log files in /var/log
   ##   "/var/log/*/*.log" -> find all .log files with a parent dir in /var/log
   ##   "/var/log/apache.log" -> just tail the apache log file
-  ##
-  ## See https://github.com/gobwas/glob for more examples
+  ##   "/var/log/log[!1-2]* -> tail files without 1-2
+  ##   "/var/log/log[^1-2]* -> identical behavior as above
+  ## See https://github.com/gobwas/glob and https://golang.org/pkg/path/filepath/#Match for more examples
```
ssoroka, line 34: add note about `**/**`

ssoroka (changes requested):

ssoroka:
replace `**` with `**/**` for backwards compatibility

srebhan:
@sspaink what's wrong with the regexp I posted? Could you give me an example where it fails!? Requiring an external dependency to replace a regexp seems odd...

sspaink:
@srebhan The two issues I noticed when you run the playground example you provided: `lala[!w]ee --> lal[^w]ee` the result overrides the 'a' in the second 'la' and then if there isn't any characters trailing the negation class it doesn't update: `[!p]laid --> [!p]laid`.

@ssoroka provided a regex that solved the second issue, [playground](https://play.golang.org/p/lhgw5uskGKP) but the first problem where the previous character gets replaced still happens.

plugins/inputs/tail/README.md
```diff
@@ -26,11 +26,14 @@ The plugin expects messages in one of the
   ## File names or a pattern to tail.
   ## These accept standard unix glob matching rules, but with the addition of
   ## ** as a "super asterisk". ie:
-  ##   "/var/log/**.log"  -> recursively find all .log files in /var/log
-  ##   "/var/log/*/*.log" -> find all .log files with a parent dir in /var/log
+  ##   /var/log/**           -> recursively find all directories in /var/log and count files in each directories
+  ##   "/var/log/**.log"     -> recursively find all .log files in /var/log
+  ##   "/var/log/**/*.log"   -> recursively find all .log files in /var/log and subdirectories
```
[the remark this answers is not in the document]
ssoroka: what if we do the replace `**` -> `**/**` ?

ssoroka:
re using a library, glob is used in a lot of different places and I'd really love consistent behavior that–ideally–meet some standard that we are not managing ourselves. It seems too easy to accidentally create our own fork of glob behavior, which it seems is exactly what we've done and now we have to support this old broken behavior...

plugins/inputs/tail/README.md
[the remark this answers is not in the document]
ssoroka: I guess the only potential concern there is not to replace existing `**/**` with `**/**/**/**`, which I believe implies a minimum number of directories descended. for matches

srebhan:
@sspaink fixed both with
```golang
package main

import (
	"fmt"
	"regexp"
)

func main() {
	test := []string {
		"lala[!w]ee",
		"[!p]laid",
		"foo\\[!wja",
		"\\[!wbar",
		"foo[!abc] and \\[!wbar are wunderbar",
	}

	re := regexp.MustCompile(`([^\\]?)(\[!)`)
	for _, s := range test {
		fmt.Printf("%s --> %s\n", s, re.ReplaceAllString(s, "$1[^"))
	}
}
```
But if I understand @ssoroka correctly he prefers to use the doublestar-package. Also fine with me. :-)

sspaink:
@srebhan thank you for the updated regular expression, I've updated the pull request to use the suggested change and it looks like it works without having to change any of the existing tests.
@ssoroka I did like the idea of moving to the library but I couldn't figure out how to cleanly replace **->**/** without causing the issue you describe (**/**->**/**/**/**). Maybe we can move this to a library safely for a 2.0 release.

srebhan (approved):
Looks good to me.

ssoroka (approved):

PR #8616, opened 2020-12-23
Add Event Log support for Windows

### LINKED ISSUE #7487: Windows: Unsupported logtarget: eventlog, using stderr
When running from console, this erroneous message is output during startup if using eventlog as logtarget.

### Relevant telegraf.conf:
```toml
[agent]
logtarget = "eventlog"
```

### System info:
Telegraf 1.13.3
OS: All Windows Server 2012 R2, 2016, 2019
Bittiness: 64

### Steps to reproduce:

1.  Set logtarget = "eventlog" in config file
2. Start telegraf from command line (either using config from service, or for example with --test)

### Expected behavior:
Not to see this message.

### Actual behavior:
See this message logged to stdout / stderr (not sure which) 

### Additional info:
This is generated in logger.go lines 106 - 122 due to missing case to handle eventlog.

I've never written Go before, so while I think I could add the case - I can't work out the impact given Windows logging seems to override what's set in logger.go anyway.   Hence not sending PR, but opening issue so someone with better knowledge can assess. 

This message confused me a lot when I first started using Telegraf, until I learned to ignore it.

Jagularr:
This should add support to be able to configure the following in telegraf.conf:

`logtarget = "eventlog"`

Currently it only works when running telegraf as an already-installed windows service. This should enable it to work always, as long as it's configured and running on a windows machine.

resolves #7487

### DIFF
--- a/cmd/telegraf/telegraf_windows.go
+++ b/cmd/telegraf/telegraf_windows.go
@@ -96,10 +96,9 @@ func runAsWindowsService(inputFilters, outputFilters, aggregatorFilters, process
 		}
 		os.Exit(0)
 	} else {
-		winlogger, err := s.Logger(nil)
 		if err == nil {
 			//When in service mode, register eventlog target andd setup default logging to eventlog
-			logger.RegisterEventLogger(winlogger)
+			logger.RegisterEventLogger(*fServiceName)
 			logger.SetupLogging(logger.LogConfig{LogTarget: logger.LogTargetEventlog})
 		}
 		err = s.Run()
--- a/cmd/telegraf/telegraf_windows.go
+++ b/cmd/telegraf/telegraf_windows.go
@@ -12,6 +12,9 @@ import (
 )
 
 func run(inputFilters, outputFilters, aggregatorFilters, processorFilters []string) {
+	// Register the eventlog logging target for windows.
+	logger.RegisterEventLogger(*fServiceName)
+
 	if runtime.GOOS == "windows" && windowsRunAsService() {
 		runAsWindowsService(
 			inputFilters,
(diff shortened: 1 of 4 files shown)


cmd/telegraf/telegraf_windows.go
```diff
@@ -96,10 +99,7 @@ func runAsWindowsService(inputFilters, outputFilters, aggregatorFilters, process
 		}
 		os.Exit(0)
 	} else {
-		winlogger, err := s.Logger(nil)
 		if err == nil {
```
ssoroka, line 102: looks like you left the error handling behind, but there's no call.

logger/event_logger.go
```diff
@@ -1,49 +1,60 @@
...
-	logger service.Logger
+	logger *eventlog.Log
 }
 
 func (t *eventLogger) Write(b []byte) (n int, err error) {
 	loc := prefixRegex.FindIndex(b)
 	n = len(b)
 	if loc == nil {
-		err = t.logger.Info(b)
+		err = t.logger.Info(1, fmt.Sprint(b))
```
ssoroka, line 27:
```suggestion
		err = t.logger.Info(1, string(b))
```

logger/event_logger.go
```diff
@@ -1,49 +1,60 @@
...
 	} else if n > 2 { //skip empty log messages
 		line := strings.Trim(string(b[loc[1]:]), " \t\r\n")
 		switch rune(b[loc[0]]) {
 		case 'I':
-			err = t.logger.Info(line)
+			err = t.logger.Info(1, line)
```
ssoroka, line 32: what are the numbers here?

ssoroka (changes requested):
does this cause any problems with telegraf's logging output?

logger/event_logger.go
Jagularr: Something called eid, the package we were using before this was calling these methods that require the 'eid' itself, and putting 1 for info, 2 for warning, and 3 for error. I'm calling it directly here because I had to ditch that package to make this work

cmd/telegraf/telegraf_windows.go
Jagularr: updated.

logger/event_logger.go
ssoroka: could make consts for it.. eidInfo, eidWarn, eidError

Jagularr: updated.

ssoroka (approved):

PR #8627, opened 2020-12-30
Added ability to define skip values in csv parser

### LINKED ISSUE #6436: Add the ability to replace or ignore values in the CSV parser
## Feature Request

### Proposal:
When trying to parse data from https://www.ndbc.noaa.gov/data/latest_obs/latest_obs.txt , they substitute `MM` in for fields that are missing a value. This data cannot be parsed by the CSV parser because if it hits that `MM`, then it throws a field conversion error.

It would be nice to be able to define a set of values to "ignore" when parsing CSV.

### Current behavior:
There is no way to configure ignoring missing values.

### Desired behavior:
A way to ignore specific values for parsing csv data where a value is substituted in for a missing value, such as in the link above.

### Use case: 
This would allow me to parse additional CSV data without needed to pre-process it or write a custom script.

helenosheaa:
Closes #6436 

Adds the ability to define an array of string skip values to the csv parser, such as an empty string, where the value matches the provided skip value it will skip the field entirely. Addresses the issue of ignoring missing values

### DIFF
--- a/plugins/parsers/csv/parser.go
+++ b/plugins/parsers/csv/parser.go
@@ -31,6 +31,7 @@ type Config struct {
 	TimestampFormat   string   `toml:"csv_timestamp_format"`
 	Timezone          string   `toml:"csv_timezone"`
 	TrimSpace         bool     `toml:"csv_trim_space"`
+	SkipValues        []string `toml:"csv_skip_values"`
 
 	gotColumnNames bool
 
@@ -195,6 +196,15 @@ outer:
 			value := record[i]
 			if p.TrimSpace {
 				value = strings.Trim(value, " ")
+			}
+
+			// don't record fields where the value matches a skip value
+			if len(p.SkipValues) > 0 {
+				for _, s := range p.SkipValues {
+					if value == s {
+						continue outer
+					}
+				}
 			}
 
 			for _, tagName := range p.TagColumns {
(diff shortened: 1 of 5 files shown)


plugins/parsers/csv/parser.go
```diff
@@ -197,6 +198,15 @@ outer:
 				value = strings.Trim(value, " ")
 			}
 
+			// don't record fields where the value matches a skip value
+			if len(p.SkipValues) > 0 {
```
Jagularr, line 202: Is checking the length necessary?

Jagularr (commented):
Looks good to me

plugins/parsers/csv/parser.go
helenosheaa: You're right it's redundant, I'll take it out

plugins/parsers/csv/parser.go
```diff
@@ -197,6 +198,14 @@ outer:
 				value = strings.Trim(value, " ")
 			}
 
+			// don't record fields where the value matches a skip value
+			for _, s := range p.SkipValues {
+				fmt.Println("in here")
```
reimda, line 203: Remove debugging println

reimda (changes requested):

plugins/parsers/csv/parser.go
helenosheaa: Done, I'll make sure I check more carefully for println in the future before I commit!

reimda: No problem, they sneak in from time to time for everybody.  That's what reviews are for.

reimda (approved):

Issue #7227, opened 2020-03-25
Error while reading multiple holding registers on modbus input plugin


gianniforesi:
### Relevant telegraf.conf:

```toml
holding_registers = [
	{ name = "MW0", byte_order = "AB", data_type = "UINT16", scale=1.0,  address = [0]},
	{ name = "MW1", byte_order = "AB", data_type = "UINT16", scale=1.0,  address = [1]},
	{ name = "MW2", byte_order = "AB", data_type = "UINT16", scale=1.0,  address = [2]},
	{ name = "MW3", byte_order = "AB", data_type = "UINT16", scale=1.0,  address = [3]},
... all registers up to the last one ...
	{ name = "MW345", byte_order = "AB", data_type = "UINT16", scale=1.0,  address = [345]},
	{ name = "MW346", byte_order = "AB", data_type = "UINT16", scale=1.0,  address = [346]},
 ]

```

### Expected behavior:
I expect to read all register listed in the configuration file.

### Actual behavior:
Error on telegraf log appears:
"quantity 346 must be between 1 and 125".

### Additional info:
If I delete register 100, 200 and 300 from register list, plugin work correct and i can read the olding registers (except fo the ones deleted).

garciaolais:
Hello @gianniforesi 

Today I tested read 1000 registers and I don't see any error 😕 

![image](https://user-images.githubusercontent.com/14133711/81487675-1dd27b00-9225-11ea-9674-5865032cbdea.png)

I attached my config file
[telegraf.txt](https://github.com/influxdata/telegraf/files/4604901/telegraf.txt)

garciaolais:
@gianniforesi 

I sended a PR to fix this issue 

[7488](https://github.com/influxdata/telegraf/pull/7488)

helenosheaa:
@garciaolais @gianniforesi is this issue now closed with the closed above PR?

garciaolais:
@helenosheaa 

https://github.com/influxdata/telegraf/pull/7488 tried solved two issues this one https://github.com/influxdata/telegraf/issues/7227 and https://github.com/influxdata/telegraf/issues/7317.

But https://github.com/influxdata/telegraf/pull/7869 solved https://github.com/influxdata/telegraf/issues/7317 first, for this reason I closed https://github.com/influxdata/telegraf/pull/7488.

Today I downloaded the last version of telegraf and seems that this issue persist, we need to create a new PR.

helenosheaa:
@garciaolais 

[...]

PR #8628, opened 2020-12-30
fix: reading multiple holding registers in modbus input plugin

garciaolais:
### Required for all PRs:

- Fix https://github.com/influxdata/telegraf/issues/7227

if you try to read multiple consecutive addresses , modbus return error "quantity 346 must be between 1 and 125".
this happens because we have a limitation with the number of registers that we can retrieve, 125 for holding registers and Input registers and 2000 for discrete inputs and coils

### DIFF
--- a/plugins/inputs/modbus/modbus.go
+++ b/plugins/inputs/modbus/modbus.go
@@ -229,19 +229,20 @@ func (m *Modbus) InitRegister(fields []fieldContainer, name string) error {
 	// [1, 2, 3, 5, 6, 10, 11, 12, 14]
 	// (1, 3) , (5, 2) , (10, 3), (14 , 1)
 	for range addrs {
-		quantity := 1
-		if ii < len(addrs) {
-			start := addrs[ii]
-			end := start
-
-			for ii < len(addrs)-1 && addrs[ii+1]-addrs[ii] == 1 && quantity < maxQuantity {
-				end = addrs[ii+1]
-				ii++
-				quantity++
-			}
-			ii++
-			registersRange = append(registersRange, registerRange{start, end - start + 1})
+		if ii >= len(addrs) {
+			break
 		}
+		quantity := 1
+		start := addrs[ii]
+		end := start
+
+		for ii < len(addrs)-1 && addrs[ii+1]-addrs[ii] == 1 && quantity < maxQuantity {
+			end = addrs[ii+1]
+			ii++
+			quantity++
+		}
+		ii++
+		registersRange = append(registersRange, registerRange{start, end - start + 1})
 	}
 
 	m.registers = append(m.registers, register{name, registersRange, fields})
--- a/plugins/inputs/modbus/modbus.go
+++ b/plugins/inputs/modbus/modbus.go
@@ -241,7 +241,7 @@ func (m *Modbus) InitRegister(fields []fieldContainer, name string) error {
 			ii++
 			quantity++
 		}
-		ii++
+
 		registersRange = append(registersRange, registerRange{start, end - start + 1})
 	}
 
--- a/plugins/inputs/modbus/modbus.go
+++ b/plugins/inputs/modbus/modbus.go
@@ -221,9 +221,7 @@ func (m *Modbus) InitRegister(fields []fieldContainer, name string) error {
 	var registersRange []registerRange
 	if name == cDiscreteInputs || name == cCoils {
 		maxQuantity = 2000
-	}
-
-	if name == cInputRegisters || name == cHoldingRegisters {
+	} else if name == cInputRegisters || name == cHoldingRegisters {
 		maxQuantity = 125
 	}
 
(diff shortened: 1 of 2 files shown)


Jagularr (commented):
Logic looks good to me

helenosheaa:
Looks good @garciaolais thanks for opening up another PR!

plugins/inputs/modbus/modbus.go
```diff
@@ -217,19 +217,29 @@ func (m *Modbus) InitRegister(fields []fieldContainer, name string) error {
 	sort.Slice(addrs, func(i, j int) bool { return addrs[i] < addrs[j] })
 
 	ii := 0
+	maxQuantity := 1
 	var registersRange []registerRange
+	if name == cDiscreteInputs || name == cCoils {
+		maxQuantity = 2000
+	}
+
+	if name == cInputRegisters || name == cHoldingRegisters {
```
srebhan, line 226: Can we make this an `else if` please.

plugins/inputs/modbus/modbus.go
```diff
@@ -217,19 +217,29 @@ func (m *Modbus) InitRegister(fields []fieldContainer, name string) error {
...
+		maxQuantity = 125
+	}
 
 	// Get range of consecutive integers
 	// [1, 2, 3, 5, 6, 10, 11, 12, 14]
 	// (1, 3) , (5, 2) , (10, 3), (14 , 1)
 	for range addrs {
+		quantity := 1
 		if ii < len(addrs) {
```
srebhan, line 235: Not directly related to the patch, but wouldn't it make sense to rather do
```suggestion
		if ii >= len(addrs) {
			break
		}
```

srebhan (commented):
Looks good basically, but I'd prefer to have the checking for max grouped registers in an else-if to make it obvious that it is exclusive.

plugins/inputs/modbus/modbus.go
garciaolais: I've tested but that change broke some tests

srebhan (approved):
Perfect thanks!

helenosheaa (approved):

plugins/inputs/modbus/modbus.go
srebhan: Very strange as from a logic point-of-view this should be equivalent. This is not related to the PR anyway, so consider this as resolved.

garciaolais: Let me double check that

garciaolais: @srebhan  you are right, your suggestion works 👍

plugins/inputs/modbus/modbus.go
```diff
@@ -217,23 +217,32 @@ func (m *Modbus) InitRegister(fields []fieldContainer, name string) error {
...
-				ii++
-			}
+		for ii < len(addrs)-1 && addrs[ii+1]-addrs[ii] == 1 && quantity < maxQuantity {
+			end = addrs[ii+1]
 			ii++
-			registersRange = append(registersRange, registerRange{start, end - start + 1})
+			quantity++
 		}
+		ii++
+		registersRange = append(registersRange, registerRange{start, end - start + 1})
```
srebhan, line 245: Hmmm I'm pretty sure that the additional `i++` here means we should change `end-start+1` to `end-start`!? However, I would feel much more safe if we remove the additional `i++` and keep the range calculation as is... :-/

srebhan (commented):
I think the register-range changed. Could you please check my comment in the code!?

srebhan:
@garciaolais can you please comment on my question above regarding the new increment of `end`!?

garciaolais:
@srebhan sorry for the delay but I really doesn't have free time to check this. Let me chance to retake it and I hope to test it this week.

plugins/inputs/modbus/modbus.go
garciaolais: we can remove `i++` but the other change _`end-start+1 `to `end-start`_  brake some test

srebhan:
@garciaolais sure, just wanted to make sure it is not forgotten. Please let me know about the results of your tests!

plugins/inputs/modbus/modbus.go
garciaolais: Hi @srebhan 

Do you have any other comments about this issue?

srebhan (approved):
Looks good to me assuming that you tested the code @garciaolais ... :-)

ssoroka (changes requested):
This all looks good. One ask is that the tests don't validate the generated metric. It's possible that it runs without errors but doesn't give you the response you'd expect. 
Does it make sense to add some lines to at least one of the tests to make sure the metric was generated as expected?

ssoroka (approved):

ssoroka (changes requested):
Actually, would still be great to get a test showing the expected metrics are generated properly.

srebhan (approved):
Looks good to me.

ssoroka (approved):

Issue #8630, opened 2020-12-31
Google BigQuery Output Plugin


gkatzioura:
## Feature Request

BigQuery is a serverless, highly scalable, and cost-effective data warehouse designed to help you turn big data into informed business decisions

### Proposal:

Implementation of a BigQuery plugin for output.

### Desired behavior:

The telegraph metrics will get mapped to a corresponding table in a BigQuery Dataset. The fields and tags would be fields in a BigQuery table with the corresponding data types. Timestamp should be a mandatory field.

### Use case: 

BigQuery can be used as a data warehouse. Can be handy for data that get in silo status or for processing vast amounts of data. Also is very well integrated with the GCP ecosystem.

helenosheaa:
Closing from pr #8634

PR #8634, opened 2020-12-31
Bigquery output Plugin

gkatzioura:
### Required for all PRs:

The pull request for the Issue #8630


plugins/outputs/bigquery/bigquery.go
```diff
@@ -0,0 +1,215 @@
...
+
+func (b *BigQuery) setUpClient() error {
+	if endpoint, present := os.LookupEnv(testingHostEnv); present {
+		return b.setUpTestClient(endpoint)
+	}
+
+	return b.setUpDefaultClient()
+}
+
+func (b *BigQuery) setUpTestClient(endpoint string) error {
```
Jagularr, line 61: Feels like this method could be moved to the test file

Jagularr (commented):
Looks solid to me

plugins/outputs/bigquery/bigquery.go
```diff
@@ -0,0 +1,215 @@
+package bigquery
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"reflect"
+
+	"cloud.google.com/go/bigquery"
+	"github.com/influxdata/telegraf"
```
srebhan, line 10: You might want to put this one in a separate "group" below as it is an "internal" package.

plugins/outputs/bigquery/bigquery.go
```diff
@@ -0,0 +1,215 @@
...
+
+	ctx := context.Background()
+
+	if c, err := bigquery.NewClient(ctx, b.Project, noAuth, endpoints); err != nil {
+		return err
+	} else {
+		b.client = c
+		return nil
+	}
+}
```
srebhan, line 73: Can you please move this to the testing! You can guard the actual function by a check for `b.client` not being `nil` if you have to.

plugins/outputs/bigquery/bigquery.go
```diff
@@ -0,0 +1,215 @@
...
+func (b *BigQuery) Connect() error {
+	if b.Project == "" {
+		return fmt.Errorf("Project is a required field for BigQuery output")
+	}
+
+	if b.Dataset == "" {
+		return fmt.Errorf("Dataset is a required field for BigQuery output")
+	}
+
+	b.setUpClient()
```
srebhan, line 48: For testing you could do
```suggestion
	// To allow testing we do not override the client. The client will never be initialized in
	// normal environment. 
	if b.client == nil {
		b.setUpClient()
	}
```
and init `b.Client` with what you have in `setUpTestClient()`. This way you can move all

plugins/outputs/bigquery/bigquery.go
```diff
@@ -0,0 +1,215 @@
...
+	endpoints := option.WithEndpoint("http://" + endpoint)
+
+	ctx := context.Background()
+
+	if c, err := bigquery.NewClient(ctx, b.Project, noAuth, endpoints); err != nil {
+		return err
+	} else {
+		b.client = c
+		return nil
+	}
```
srebhan, line 72: As you return in the if you can save some indention:
```suggestion
	}
	b.client = c

	return nil
```

plugins/outputs/bigquery/bigquery.go
```diff
@@ -0,0 +1,215 @@
...
+}
+
+func (b *BigQuery) setUpDefaultClient() error {
+	var credentialsOption option.ClientOption
+
+	ctx := context.Background()
```
srebhan, line 78: Maybe you want a user-definable timeout and use it here!? Do not forget to `defer cancel()` in this case.

plugins/outputs/bigquery/bigquery.go
```diff
@@ -0,0 +1,215 @@
...
+	return err
+}
+
+// Write the metrics to Google Cloud BigQuery.
+func (b *BigQuery) Write(metrics []telegraf.Metric) error {
+	groupedMetrics := b.groupByMetricName(metrics)
+
+	for k, v := range groupedMetrics {
+		if err := b.insertToTable(k, v); err != nil {
+			return err
```
srebhan, line 103: This means you drop all remaining metrics. Skipping over this erroneous one but process the others might be better
```suggestion
			b.Log.Errorf("inserting metric %q failed: %v", <metric id/name>, err)
			continue
```

plugins/outputs/bigquery/bigquery.go
```diff
@@ -0,0 +1,215 @@
...
+	b.client = client
+	return err
+}
+
+// Write the metrics to Google Cloud BigQuery.
+func (b *BigQuery) Write(metrics []telegraf.Metric) error {
+	groupedMetrics := b.groupByMetricName(metrics)
+
+	for k, v := range groupedMetrics {
+		if err := b.insertToTable(k, v); err != nil {
```
srebhan, line 102: Does it make sense to do this in parallel using a go-function?

plugins/outputs/bigquery/bigquery.go
```diff
@@ -0,0 +1,215 @@
...
+		return bigquery.FloatFieldType
+	case reflect.Bool:
+		return bigquery.BooleanFieldType
+	default:
+		return bigquery.StringFieldType
+	}
+}
+
+func (b *BigQuery) insertToTable(metricName string, metrics []bigquery.ValueSaver) error {
+	ctx := context.Background()
```
srebhan, line 190: Again, you might want a user-definable timeout...

plugins/outputs/bigquery/bigquery.go
```diff
@@ -0,0 +1,215 @@
...
+
+// SampleConfig returns the formatted sample configuration for the plugin.
+func (s *BigQuery) SampleConfig() string {
+	return sampleConfig
+}
+
+// Description returns the human-readable function definition of the plugin.
+func (s *BigQuery) Description() string {
+	return "Configuration for Google Cloud BigQuery to send entries"
+}
```
srebhan, line 215: Usually those are at the top.

plugins/outputs/bigquery/bigquery.go
```diff
@@ -0,0 +1,215 @@
...
+
+func (b *BigQuery) tableForMetric(metricName string) *bigquery.Table {
+	return b.client.DatasetInProject(b.Project, b.Dataset).Table(metricName)
+}
+
+// Close will terminate the session to the backend, returning error if an issue arises.
+func (b *BigQuery) Close() error {
+	return b.client.Close()
+}
+
```
srebhan, line 206: You are missing a `init()` function to register this output plugin.

srebhan (changes requested):
Hey this looks quite good already. My main concerns are

1. You are missing an `init()` function to register the plugin.
2. You might want a user-definable `timeout` option to avoid blocking forever in operations.
3. You should move the testing machinery out of the main plugin file.
4. Please add `cloud.google.com/go/bigquery` to `docs/LICENSE_OF_DEPENDENCIES.md`.

See also my comments in the code.

plugins/outputs/bigquery/bigquery.go
gkatzioura: I updated it, to achieve it through a Wait Group.

plugins/outputs/bigquery/bigquery.go
```diff
@@ -0,0 +1,233 @@
...
+	defaultOffSetKey   = "offset-key.json"
+	timeStampFieldName = "timestamp"
+)
+
+var defaultTimeout = 5 * time.Second
+
+const sampleConfig = `	
+  ## Credentials File
+  credentials_file = "/path/to/service/account/key.json"
+  ## GCP Project
```
srebhan, line 29: Could you please avoid abbreviations here as some first-time-users might have difficulties to understand what it is about... :-)

plugins/outputs/bigquery/bigquery.go
```diff
@@ -0,0 +1,233 @@
...
+	}
+
+	return nil
+}
+
+func (b *BigQuery) setUpTimeout() {
+	if b.Timeout.Duration < 1*time.Second {
+		b.Timeout.Duration = defaultTimeout
+	}
+}
```
srebhan, line 83: Could you please also move this to `Connect()` to have a nice consistent structure. It will add to the beauty of this code.

plugins/outputs/bigquery/bigquery.go
```diff
@@ -0,0 +1,233 @@
...
+const sampleConfig = `	
+  ## Credentials File
+  credentials_file = "/path/to/service/account/key.json"
+  ## GCP Project
+  project = "my-gcp-project"
+
+  ## The namespace for the metric descriptor
+  dataset = "telegraf"
+
+  ## Timeout for BigQuery operations.  Minimum timeout is "1s".
```
srebhan, line 35: Is there a reason you enforce a timeout of one second? If it is about default-value handling, the easy solution is to set the default in `init()` when initializing the structure. This way the value is only overwritten when set by the user.

plugins/outputs/bigquery/bigquery.go
```diff
@@ -0,0 +1,233 @@
...
+
+// Write the metrics to Google Cloud BigQuery.
+func (b *BigQuery) Write(metrics []telegraf.Metric) error {
+	groupedMetrics := b.groupByMetricName(metrics)
+
+	var wg sync.WaitGroup
+
+	for k, v := range groupedMetrics {
+		wg.Add(1)
+		go b.insertToTable(k, v, &wg)
```
srebhan, line 117: You might use an anonymous function here to put the `defer wg.Done()` here rather than the function body for everybody to see it's done correctly. But that is a matter of taste and I leave it to you.

plugins/outputs/bigquery/bigquery.go
srebhan: Please mark as resolved.

plugins/outputs/bigquery/bigquery_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+func TestMain(t *testing.M) {
+	srv := localBigQueryServer(t)
+
+	testingHost = strings.ReplaceAll(srv.URL, "http://", "")
+
+	defer srv.Close()
+
+	os.Exit(t.Run())
+}
```
srebhan, line 33: I don't get the intention behind this test... If you only want to test if you can setup a local test-server, I think you don't need to do it.

plugins/outputs/bigquery/bigquery_test.go
```diff
@@ -0,0 +1,96 @@
...
+		Dataset: "test-dataset",
+	}
+
+	mockMetrics := testutil.MockMetrics()
+	b.setUpTestClient()
+	b.Connect()
+	err := b.Write(mockMetrics)
+	require.NoError(t, err)
+}
+
```
srebhan, line 59: Could you add a test where you actually compare what is sent to the server with what you expect? Do you have an example of the on-wire format?

srebhan (commented):
Very nice code! Thank you! Only few comments in the code. One nice-to-have thing is a test where you actually compare what is sent to the server with your expectations. But I'm not sure if this is doable!?

plugins/outputs/bigquery/bigquery.go
gkatzioura: Moved to anonymous function

srebhan:
Hey @gkatzioura you are writing the code faster than I can review! ;-) Good job! :-)

plugins/outputs/bigquery/bigquery.go
```diff
@@ -110,7 +110,10 @@ func (b *BigQuery) Write(metrics []telegraf.Metric) error {
 
 	for k, v := range groupedMetrics {
 		wg.Add(1)
-		go b.insertToTable(k, v, &wg)
+		go func(k string, v []bigquery.ValueSaver) {
+			b.insertToTable(k, v)
+			wg.Done()
```
srebhan, line 115: Make it `defer wg.Done()` and move it to the top of the anonymous function so we can be sure it will be done in any case.

plugins/outputs/bigquery/bigquery.go
srebhan: Please mark as resolved.

plugins/outputs/bigquery/bigquery.go
```diff
@@ -0,0 +1,229 @@
...
+	groupedMetrics := b.groupByMetricName(metrics)
+
+	var wg sync.WaitGroup
+
+	for k, v := range groupedMetrics {
+		wg.Add(1)
+		go func(k string, v []bigquery.ValueSaver) {
+			defer wg.Done()
+			b.insertToTable(k, v)
+		}(k, v)
```
srebhan, line 116: This should also work
```suggestion
		go func() {
			defer wg.Done()
			b.insertToTable(k, v)
		}()
```

srebhan:
@gkatzioura can you please comment on why you enforce a minimum timeout of 1s!?

gkatzioura:
> @gkatzioura can you please comment on why you enforce a minimum timeout of 1s!?

Hi @srebhan will do, bit busy now due to work.

srebhan:
@gkatzioura ok don't worry, just liked the live coding/review in the morning... ;-)

plugins/outputs/bigquery/bigquery.go
gkatzioura: Hi @srebhan no specific reason. Was just checking the zookeeper connector that did the same. Did the default as you proposed.

plugins/outputs/bigquery/bigquery_test.go
gkatzioura: I run the tests through TestMain. I use TestMain in order to spin up the server that produces a fake BigQuery Server. Then through T.Run() the tests `TestWrite` and  `TestConnect` will run using the fake BigQuery Server.

plugins/outputs/bigquery/bigquery.go
srebhan: Please mark as resolved.

plugins/outputs/bigquery/bigquery_test.go
srebhan: This is an unusual pattern for telegraf. :-) Is there any reason not to spin up a server in every test (that's what everybody else in telegraf does)?

plugins/outputs/bigquery/bigquery.go
```diff
@@ -0,0 +1,227 @@
...
+	"github.com/influxdata/telegraf/internal"
+	"github.com/influxdata/telegraf/plugins/outputs"
+)
+
+const (
+	defaultOffSetKey   = "offset-key.json"
+	timeStampFieldName = "timestamp"
+)
+
+var defaultTimeout = internal.Duration{Duration: time.Duration(time.Second * 5)}
```
srebhan, line 24: Can you please do:
```suggestion
var defaultTimeout = internal.Duration{Duration: 5 * time.Second}
```
as there is a big PR fixing those superfluous expressions (#8835) and Pawel will not be happy if we add more. ;-)

plugins/outputs/bigquery/bigquery_test.go
```diff
@@ -21,6 +23,7 @@ const (
 )
 
 var testingHost string
+var testDuration = internal.Duration{Duration: time.Duration(time.Second * 5)}
```
srebhan, line 26: Please remove `time.Duration`.

srebhan:
Almost. Please remove the `time.Duration` as the timeout is already a duration. Furthermore, you can remove`cloud.google.com/go/bigquery` from the list-of-dependencies document as `cloud.google.com/go/*` is covered already.

Last thing is the test-structure. Can you please start the server in each test individually to match "the usual pattern"!?

plugins/outputs/bigquery/bigquery_test.go
gkatzioura: Changed to spinning up per test.

plugins/outputs/bigquery/README.md
```diff
@@ -0,0 +1,46 @@
...
+  ## GCP Project
+  project = "erudite-bloom-151019"
+
+  ## The BigQuery dataset
+  dataset = "telegraf"
+```
+
+### Restrictions
+
+Avoid hyphens on BigQuery tables, underlying sdk cannot handle straming inserts to Table with hyphens.
```
srebhan, line 32:
```suggestion
Avoid hyphens on BigQuery tables, underlying SDK cannot handle streaming inserts to Table with hyphens.
```?

File: plugins/outputs/bigquery/README.md [hunk 23]
```diff
@@ -0,0 +1,46 @@
...
+
+Available data type options are:
+* integer
+* float or long
+* string
+* boolean
+
+All field naming restrictions that apply to BigQuery sould apply to the measurements to be imported.
```
srebhan, line 40:
```suggestion
All field naming restrictions that apply to BigQuery should apply to the measurements to be imported.
```

plugins/outputs/bigquery/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+Tables on BigQuery should be created before hand and they are not created during persistence
+
+Pay attention to the columnd `timestamp` since it is reserved upfront and cannot change.
```
srebhan, line 44:
```suggestion
Pay attention to the column `timestamp` since it is reserved upfront and cannot change. 
```?

File: plugins/outputs/bigquery/README.md [hunk 25]
```diff
@@ -0,0 +1,46 @@
...
+
+Available data type options are:
+* integer
+* float or long
+* string
+* boolean
+
+All field naming restrictions that apply to BigQuery sould apply to the measurements to be imported.
+
+Tables on BigQuery should be created before hand and they are not created during persistence
```
srebhan, line 42:
```suggestion
Tables on BigQuery should be created beforehand and they are not created during persistence
```

plugins/outputs/bigquery/bigquery.go
```diff
@@ -0,0 +1,227 @@
...
+
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/internal"
+	"github.com/influxdata/telegraf/plugins/outputs"
+)
+
+const (
+	defaultOffSetKey   = "offset-key.json"
+	timeStampFieldName = "timestamp"
+)
```
srebhan, line 22: I think this isn't used.
```suggestion
const timeStampFieldName = "timestamp"
```

plugins/outputs/bigquery/bigquery.go
```diff
@@ -0,0 +1,227 @@
...
+
+var defaultTimeout = internal.Duration{Duration: 5 * time.Second}
+
+const sampleConfig = `	
+  ## Credentials File
+  credentials_file = "/path/to/service/account/key.json"
+  ## Google Cloud Platform Project
```
srebhan, line 29:
```suggestion
  credentials_file = "/path/to/service/account/key.json"

  ## Google Cloud Platform Project
```

plugins/outputs/bigquery/bigquery.go
```diff
@@ -0,0 +1,227 @@
...
+
+  ## Timeout for BigQuery operations.
+  # timeout = "5s"
+`
+
+type BigQuery struct {
+	CredentialsFile string `toml:"credentials_file"`
+	Project         string `toml:"project"`
+	Dataset         string `toml:"dataset"`
+
```
srebhan, line 43:
```suggestion
```

plugins/outputs/bigquery/bigquery.go
```diff
@@ -0,0 +1,227 @@
...
+	Timeout internal.Duration `toml:"timeout"`
+
+	Log telegraf.Logger
```
srebhan, line 46:
```suggestion
	Log telegraf.Logger `toml: "-"`
```

srebhan (commented):
I went through everything a (almost) last time and found some more typos and formatting things. Only cosmetics. After fixing those I think we are good to go.

gkatzioura:
Hi @srebhan I added a map in order to do a lookup for cases of metrics with hyphens. Due to the SDK and the api having issues inserting to tables with hyphens this connector would have been limited to only non-hyphen metrics. 
Instead by having this map the user can define an association between a metric with hyphens and the corresponding table.

srebhan:
I'm a bit ambivalent here. On the one hand-side I think letting the user specify a replacement (e.g. underscore, space or nothing, i.e. empty) would be good enough in my view, on the other hand this is not much code. What makes me tend to the "user specifies a replacement" (with a sensible default) solution is the fact that this way you can _ensure_ that you can process the metric.
In the table case, the user might simply forget, mistype or do some other error that leaves you with a hyphen-metric without a mapping in the table. What do you do in such cases?

Any thoughts?

gkatzioura:
Hi @srebhan. 
My concern was users dropping the plugin due to the naming limitation, since many measurement will have hyphens.If the users specify a map (for the hyphens cases, there is no need to do it for everything, the connector should also work out of the box) they are covered on this issue but as you mentioned they are not covered in case they forget that they have a hyphen metric.
It seems ok in case of a hyphen metric where the user has not specified a table mapping, to fallback on removing the hyphens (and print a warning). 
Thus co-existence of table mapping exists and also a fallback.
However the user still needs to create the BigQuery tables upfront. So somehow he still needs to be aware on how to name the tables and there also might be a rename collision.

srebhan:
@gkatzioura I can follow your argumentation, but a mixture of concepts (replacement of hyphen vs. replacement of metric name) is always the worst thing to do... ;-)

> If the users specify a map (for the hyphens cases, there is no need to do it for everything, the connector should also work out of the box) they are covered on this issue but as you mentioned they are not covered in case they forget that they have a hyphen metric.

Right. That's my concern. In large installations with many plugins and maybe multiple users modifying the config you will hit the  "forgotten mapping" case with certainty! That's why I would prefer to specify a replacement character (e.g. underscore) for the hyphens. If the user wants to really rename the metrics he/she can still use the [rename processor](https://github.com/influxdata/telegraf/tree/master/plugins/processors/rename) for doing so. This would keep the config options for your plugins short (it's just so beautifully brief!) and makes sure we cover _all_ cases by default.

gkatzioura:
Hi @srebhan I think the [rename processor](https://github.com/influxdata/telegraf/tree/master/plugins/processors/rename) is amazing and there no need for what I did with the map (was not fully aware :/ on it). Thus I shall remove the map mechanism.

On your comment you proposed to swap hyphens with underscores. Are you ok for me to proceed?

srebhan:
@gkatzioura yeah that's my proposal. However, I would like to see a config option with the replacement, having underscore as default and removing hyphens if empty. Do you think this makes sense?

For logging on replace, please track which metric you already warned on to avoid spamming the log-file. You can do this easily by using a `map[string]bool` that you set if you warned and before warning check `if ! alreadyWarned[metric]`...

gkatzioura:
Hi @srebhan will proceed as suggested.

gkatzioura:
Hi @srebhan did proceed on the changes and updated the Readme.
* Added a default hyphen replacement
* Added configuration on the character replacing
* Print warning only once

plugins/outputs/bigquery/bigquery.go
```diff
@@ -0,0 +1,248 @@
...
+
+  ## Google Cloud Platform Project
+  project = "my-gcp-project"
+
+  ## The namespace for the metric descriptor
+  dataset = "telegraf"
+
+  ## Timeout for BigQuery operations.
+  # timeout = "5s"
+`
```
srebhan, line 36: This misses the `replace_hyphen_to` option. Please also sync with `README.md`.

plugins/outputs/bigquery/bigquery.go
```diff
@@ -0,0 +1,248 @@
...
+
+		if warned, _ := b.warnedOnHyphens[metricName]; !warned {
+			b.Log.Warn("Metric %q contains hyphens please consider using the rename processor plugin, falling back to %q", dhm)
+			b.warnedOnHyphens[metricName] = true
+		}
+
+		return dhm
+	}
+
+	return metricName
```
srebhan, line 228: Please invert the logic to save indention levels:
```suggestion
	if ! strings.Contains(metricName, "-") {
		return metricName
	}

	dhm := strings.ReplaceAll(metricName, "-", b.ReplaceHyphenTo)
	if warned, _ := b.warnedOnHyphens[metricName]; !warned {
		b.Log.Warn("Metric %q contains hyphens please consider using the rename processor plugin, falling back to %q", dhm)
		b.warnedOnHyphens[metricName] = true
	}

	return dhm
```

plugins/outputs/bigquery/bigquery.go
```diff
@@ -0,0 +1,248 @@
...
+// Close will terminate the session to the backend, returning error if an issue arises.
+func (b *BigQuery) Close() error {
+	return b.client.Close()
+}
+
+func init() {
+	outputs.Add("bigquery", func() telegraf.Output {
+		return &BigQuery{
+			Timeout:         defaultTimeout,
+			warnedOnHyphens: make(map[string]bool),
```
srebhan, line 244: Please do this in `Init()`.

srebhan (commented):
Very good, please check the minor comments and update `sampleConfig` and `README.md` to the new option(s).

plugins/outputs/bigquery/bigquery.go
gkatzioura: I moved this above. Did you mean doing the map initialization on the Connect()?

srebhan: I mean moving `b.warnOnHyphens = make(map[string]bool)` to `Connect()` instead of doing it here. Sorry for the confusion, `Init()` is the `Connect()` in input plugins... :-)

srebhan (approved):
This code looks beautiful IMO. Thank you very much for your effort!

srebhan:
Oh just one more request @gkatzioura... Please add this plugin to the [top-level README file's output list](https://github.com/influxdata/telegraf/blob/master/README.md#output-plugins).

srebhan (approved):
Perfect.

srebhan:
@gkatzioura please resolve the merge conflicts!

srebhan (approved):
It's off again, please take a look from time to time so this one stays merge-able. I'll try to find out when this might be merged.

ssoroka:
I'm good to merge this after you've had a look at my last two comments.

plugins/outputs/bigquery/bigquery.go
```diff
@@ -0,0 +1,251 @@
...
+	ctx, cancel := context.WithTimeout(ctx, b.Timeout.Duration)
+	defer cancel()
+
+	tableName := b.metricToTable(metricName)
+	table := b.client.DatasetInProject(b.Project, b.Dataset).Table(tableName)
+	inserter := table.Inserter()
+
+	if err := inserter.Put(ctx, metrics); err != nil {
+		b.Log.Errorf("inserting metric %q failed: %v", metricName, err)
+	}
```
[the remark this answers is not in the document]
gkatzioura: Yes this is a no retry case. In case of a failure due to the availability of BigQuery the assumption is that the connector is misconfigured (BigQuery Table does not exist, wrong fields). Let me know what would work better.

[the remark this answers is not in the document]
ssoroka: I'm fine with it if you are. I do wonder if maybe the Goroutines aren't really necessary. The retry logic would be easier without them, as on 5xx status code you'd return an error, and on 4xx you'd log it and return nil. If they're all going over the same client connection, I think they're being serialized anyway.

ssoroka:
I'm going to merge and you can adjust the goroutines if you want in a new PR. [See my last reply](https://github.com/influxdata/telegraf/pull/8634#discussion_r593324308)

plugins/outputs/bigquery/bigquery.go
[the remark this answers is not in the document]
gkatzioura: Yes, initially the pr was not using Goroutines, but I was encouraged to parallelize the inserts. I will raise a PR removing them, thus more control and will return on error.

PR #7749, opened 2020-06-26
Docker log track message offset

i-prudnikov:
### Required for all PRs:

- [X] Signed [CLA](https://influxdata.com/community/cla/).
- [X] Associated README.md updated.
- [X] Has appropriate unit tests.

This PR comes from #7068, as discussed with @danielnelson, we agreed to split #7068 to smaller parts. This the second part - track message timestamps. This allows to resume from particular message, and prevent delivering messages that already delivered. This feature inspired by fluentd.

PR #8635, opened 2020-12-31
Ali cloud CMS input plugin

i-prudnikov:
### Required for all PRs:

Here the input to grab metrics from aliyun (aka alibaba) cloud monitor service (CMS)  - this is analogue of AWS cloudwatch. The input allows to scrape any metrics that is provided by CMS service - https://www.alibabacloud.com/help/product/28572.htm?spm=a3c0i.63551.6516238130.1.1b3f112enCNPQF.
Also the input allows to enrich metric information with some attributes of monitored objects. For that an additional API calls to ali cloud objects is used, this feature called `discovery`. Discovery is supported for a limited list of object types and can be extended to support other types later. Discovery data can be extracted using jmespath - https://jmespath.org/  and added in a form of tags to input's metrics.


Hipska:
Hi @i-prudnikov, would it not be better to have the project name as a tag as well instead of merged into the measurement name?

i-prudnikov:
Hi Thomas, sure we can do that. The rationale of current design is this:
 - project describes type of the alicloud object (it  is exactly the same as AWS namespace). Since we can configure multiple inputs to scrape different types of objects, then we have different measurement names, that look very robust to me.
 - the aliyun CMS service look & feel is very similar to AWS cloudwatch. So I follow the structure of cloudwatch input plugin in terms of  the configurable  attributes  etc.. And In  AWS cloudwatch input we have merged namespace  into measurement name.
 
Anyway if this is a blocker for merging/reviewing this MR then we can of course move project name to tag, I see no technical issues here.

pyogesh2:
We are very much interested in the feature and alicloud metrics. Are there any possibilities of bringing this as soon as possible? It would be of great help for our alicloud usecases. Thanks!

Hipska:
Hi, I'm wondering why this plugin is called Aliyun instead of Alibaba, since it seems the product is still called Alibaba Cloud? What am I missing here?

Hipska (approved):

i-prudnikov:
Hi, @Hipska , 'yun' 云  is Chinese word for 'cloud'. The provider itself used it across documentation and web console. So it looks very native.

Hipska:
Okay, thanks. Could you just reference that it is Alibaba Cloud in the readme file? I will contact someone who can do the merge then 😉

i-prudnikov:
Sure, done.

Hipska (approved):

i-prudnikov:
Hello Thomas, could you please also review these 2:
https://github.com/influxdata/telegraf/pull/8639
https://github.com/influxdata/telegraf/pull/7749

Hipska:
Hi @i-prudnikov, the team wanted to know if this plugin could not be merged with the AWS plugin? Are there any Aliyun specific stuff, or is the API really a clone of AWS?

i-prudnikov:
HI @Hipska , I think the structure of the plugin under the hood is quite different since we have a completely different structure of the API and we have discovery option that is not provided by AWS cloudwatch plugin. Similarities are: user experience and general settings, since Alicloud borrows a lot of concepts from AWS.

Hipska (approved):

pyogesh2:
Hello, We are waiting for this plugin for a quite a long time for our business cases in alicloud. Happy to see it is with ready to merge label. Can we get this merged at the earliesrt?

vipinvkmenon:
Yes seems like a much sought after need....would be great to merge this.

plugins/inputs/aliyuncms/README.md
```diff
@@ -0,0 +1,142 @@
...
+  ## - acs_rds_dashboard
+  ## - acs_slb_dashboard
+  ## - acs_vpc_eip
+  ##
+  ## If not set, all regions would be covered, it can provide a significant load on API, so the recommendation here
+  ## is to limit the list as much as possible. Allowed values: https://www.alibabacloud.com/help/zh/doc-detail/40654.htm
+  discovery_regions = ["cn-hongkong"]
+  
+  ## how often the discovery API call executed (default 1m)
+  #discovery_interval = "1m"
```
ssoroka, line 79: Why have discovery_interval if there's already an interval?

plugins/inputs/aliyuncms/aliyuncms.go
```diff
@@ -0,0 +1,597 @@
...
+func (l logger) logW(format string, v ...interface{}) {
+	log.Printf(l.headerFormat, l.warnHeader, fmt.Sprintf(format, v...))
+}
+func (l logger) logI(format string, v ...interface{}) {
+
+	log.Printf(l.headerFormat, l.infoHeader, fmt.Sprintf(format, v...))
+}
+func (l logger) logD(format string, v ...interface{}) {
+	log.Printf(l.headerFormat, l.debugHeader, fmt.Sprintf(format, v...))
+}
```
ssoroka, line 154: instead of a custom logger, add 
```
Logger Telegraf.Logger `toml:"-"`
```
to the main struct and it will be populated with a logger instance you can use. Set it to testutils.Logger{} in tests

plugins/inputs/aliyuncms/aliyuncms.go
```diff
@@ -0,0 +1,597 @@
...
+		RoleArn           string `toml:"role_arn"`
+		RoleSessionName   string `toml:"role_session_name"`
+		PrivateKey        string `toml:"private_key"`
+		PublicKeyID       string `toml:"public_key_id"`
+		RoleName          string `toml:"role_name"`
+
+		DiscoveryRegions  []string          `toml:"discovery_regions"`
+		DiscoveryInterval internal.Duration `toml:"discovery_interval"`
+		Period            internal.Duration `toml:"period"`
+		Delay             internal.Duration `toml:"delay"`
```
ssoroka, line 171: is it possible to take out the aggregation parts? usually we use separate plugins for that. Is there something special about how this does aggregation?

plugins/inputs/aliyuncms/aliyuncms.go
```diff
@@ -0,0 +1,597 @@
...
+	s.discoveryData, err = s.dt.getDiscoveryDataAllRegions(nil)
+	if err != nil {
+		lg.logE("Discovery tool is not activated: %v", err)
+		s.dt = nil
+		return nil
+	}
+
+	lg.logI("%d object(s) discovered...", len(s.discoveryData))
+	//Start periodic discovery process
+	s.dt.Start()
```
ssoroka, line 292: this shouldn't happen here, it should probably be in the Start function.

plugins/inputs/aliyuncms/aliyuncms.go
```diff
@@ -0,0 +1,597 @@
...
+	req.EndTime = strconv.FormatInt(s.windowEnd.Unix()*1000, 10)
+	req.StartTime = strconv.FormatInt(s.windowStart.Unix()*1000, 10)
+	req.Dimensions = metric.requestDimensionsStr
+
+	for more := true; more; {
+		resp, err := s.client.DescribeMetricList(req)
+		if err != nil {
+			return errors.Errorf("failed to query metricName list: %v", err)
+		} else if resp.Code != "200" {
+			return errors.Errorf("failed to query metricName list: %v", resp.Message)
```
ssoroka, line 379: you may not want to return an error here. probably log and break the loop instead.

plugins/inputs/aliyuncms/aliyuncms.go
```diff
@@ -0,0 +1,597 @@
...
+		}
+
+		var datapoints []map[string]interface{}
+		if err = json.Unmarshal([]byte(resp.Datapoints), &datapoints); err != nil {
+			return errors.Errorf("failed to decode response datapoints: %v", err)
+		}
+
+		if len(datapoints) == 0 {
+			lg.logW("No metrics returned from CMS, response msg: %s", resp.Message)
```
ssoroka, line 388: Probably a good debug level. it should be pretty quiet in normal operation.

plugins/inputs/aliyuncms/aliyuncms.go
```diff
@@ -0,0 +1,597 @@
...
+L:
+	for {
+		select {
+		case s.discoveryData = <-s.dt.dataChan:
+			newData = true
+			continue
+		default:
+			break L
+		}
+	}
```
ssoroka, line 484: doesn't look like the label is required here?

plugins/inputs/aliyuncms/discovery.go
```diff
@@ -0,0 +1,508 @@
...
+	)
+
+	//Initializing channel
+	dt.done = make(chan bool)
+
+	dt.wg.Add(1)
+	go func() {
+		defer dt.wg.Done()
+
+		ticker := time.NewTicker(dt.interval)
```
ssoroka, line 457: Why start another internal service to do the discovery when you could potentially do it in the Gather function?

plugins/inputs/aliyuncms/aliyuncms_test.go
```diff
@@ -0,0 +1,485 @@
...
+			require.Empty(t, acc.GatherError(plugin.Gather))
+			require.Equal(t, acc.HasMeasurement("aliyuncms_acs_slb_dashboard"), tt.hasMeasurment)
+			if tt.hasMeasurment {
+				acc.AssertContainsTaggedFields(t, "aliyuncms_acs_slb_dashboard", tt.expected[0].Fields(), tt.expected[0].Tags())
+			}
+		})
+	}
+}
+
+//func TestParseDiscoveryResponse(t *testing.T) {
```
ssoroka, line 331: should probably fix or remove the commented-out test

ssoroka (changes requested):
Thanks for the PR! I have some questions below

plugins/inputs/aliyuncms/README.md
i-prudnikov: The reason is that the nature of the API calls is very different.
 `interval` defines how often we ask API about ali-cloud object metrics. `discovery_interval` defines how often we ask API about what ali-cloud objects exist and what attributes they have. There is no need to ask too often about new objects as once they are created we eventually can get the info about them and enrich metrics with objects attributes.

To sum up: different intervals is to decreasing API calls frequency (we are rate-limitied) and the overhead it produce, especially in the case when we ask across all regions (`discovery_regions` attribute set to empty list).

i-prudnikov:
Hello Steven, thank you for reviewing. I will provide comments shortly.

plugins/inputs/aliyuncms/aliyuncms.go
i-prudnikov: Good point, thank you. Fixed.

plugins/inputs/aliyuncms/aliyuncms.go
i-prudnikov: Yes, there is some special logic here. The API supports different aggregation per object, we need to clearly specify which aggregation period we are querying (that's the reason for `period` parameter). The logic is similar to cloudwatch input plugin, since Alicloud borrows a lot of concepts from AWS.

plugins/inputs/aliyuncms/aliyuncms.go
i-prudnikov: Moved to debug level

plugins/inputs/aliyuncms/aliyuncms.go
i-prudnikov: Agree, make more sense to move it into Start function, done.

plugins/inputs/aliyuncms/aliyuncms.go
i-prudnikov: Agree, changed to returning error with braking loop

plugins/inputs/aliyuncms/aliyuncms.go
i-prudnikov: The outer `for` cycle is running forever and we need to break it once all buffered data read, otherwise we never go out from the `for` cycle, as `break` without label breaks from inner `select` statement not outer  `for` loop. So I believe the label is needed here.

plugins/inputs/aliyuncms/discovery.go
i-prudnikov: Well, discovery process is independent from plugin `gather` as they gathers quite a different info (metrics vs discovery data). Discovery runs on it's own pace, as I commented earlier about the reasons why have `interval` and `discovery_interval`

plugins/inputs/aliyuncms/aliyuncms_test.go
i-prudnikov: Agree, removed

plugins/inputs/aliyuncms/aliyuncms.go
Hipska: I think @ssoroka was hinting on using a different method instead of endless `for` and a label.

i-prudnikov: @Hipska, thanks for the hint. I oversaw the point.
Maybe smth. like this would be ok:
```go
var moreData = true
for ;moreData;{
select {
  case s.discoveryData = <-s.dt.dataChan:
	newData = true
	continue
  default:
	moreData = false
    }
}
```
What do you think?

Hipska: I have no opinion on this, we'll wait on @ssoroka's comments.

ssoroka (approved):

ssoroka:
able to resolve the merge conflict?

plugins/inputs/aliyuncms/aliyuncms.go
Hipska: @ssoroka You resolved this conversation, but didn't give feedback? Do you mean it is okay to leave like this or what?

helenosheaa:
Resolved conflicts and merged to master manually, thank you

Hipska:
Could you reference the commit?

helenosheaa:
Yes, [here](https://github.com/influxdata/telegraf/commit/ba8452d61dcd8c8ce750c4db5f1fcf527d9a4864)

pyogesh2:
Thanks ! Glad to see this PR is merged to Master. Would be great if we can know this feature would be part of which telegraf release ?

sjwang90:
This was merged right after the cut off of our last feature release so this will be part of Telegraf 1.19 which will be released sometime towards the end of June 2021.

In the mean time you can use it if you build off of master or if you download one of the [nightly builds](https://github.com/influxdata/telegraf#nightly-builds).

pyogesh2:
Thank you @sjwang90

PR #8636, opened 2021-01-01
Disable fmt/fmtcheck targets on Windows

slai:
I noticed that in Windows CI builds, the `make check` target, specifically the `fmtcheck` target, silently fails, e.g. https://ci.appveyor.com/project/sspaink/telegraf/builds/37051098#L71

```
make check
process_begin: CreateProcess(C:\go\bin\gofmt.exe, gofmt -l -s accumulator.go agent/accumulator.go agent/accumulator_test.go
...
testutil/metric_test.go testutil/testutil.go testutil/testutil_test.go testutil/tls.go, ...) failed.
Makefile:107: pipe: Bad file descriptor
```

This is because the Makefile attempts to run `gofmt` by passing it a list of every `*.go` file except a certain generated file. This list is almost 40000 characters long, and on Windows, the [maximum is 32768 characters](https://stackoverflow.com/a/28452546), so it fails to execute. This is a Windows CreateProcess API call limitation, so it will fail regardless of the shell used.

CI however, continued working, because make does not fail when variables with shell function values fail. The end result is a lot of error logging bloating out the CI logs for no useful purpose.

This PR changes the following -
* moves the `gofmt` execution into the `fmtcheck` target script itself and configures the script to fail if it fails. 
* replaces `make fmt` and `make fmtcheck` on Windows with a message instead (this was broken anyway, so there's nothing is being lost, plus the Linux builds will fail if there are any formatting errors)
* some minor refactoring of the variables

The longer term solution will probably be to move to something like golangci-lint, which has its own config for [excluding files](https://golangci-lint.run/usage/configuration/#command-line-options), removing the need for lengthy `git ls-files` listing. I didn't make that change because it's an additional tool that all devs will need to install on their systems, so it probably needs further discussion.

Alternatively, the generated code could be moved into a separate Go package, and that entire package could be excluded from `gofmt` instead. The list of packages is smaller (only 19074 characters), under Windows' 32k limit. 

### Required for all PRs:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,7 +53,7 @@ endif
 
 
 GOFILES ?= $(shell git ls-files '*.go')
-GOFMT ?= $(shell gofmt -l -s $(filter-out plugins/parsers/influx/machine.go, $(GOFILES)))
+GOFILES_FMT ?= $(filter-out plugins/parsers/influx/machine.go, $(GOFILES))
 
 prefix ?= /usr/local
 bindir ?= $(prefix)/bin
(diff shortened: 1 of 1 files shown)


Jagularr (approved):
This makes sense to me

ssoroka:
Could probably do a list of packages for all other packages, then a list of files for the one package with issues? 
I'd also be okay with adding golangci-lint as a dependency of the project to make sure it's installed. we're thinking about using it anyway.

sspaink:
Closing as we are using golangci-lint now, thank you for preparing this solution and explaining it so well.

PR #8639, opened 2021-01-03
AWS Cloudwatch log output

i-prudnikov:
### Required for all PRs:

The primary motivation is to fix messy log storing when using native AWS tools to deliver logs from docker containers (for example) running on EC2 (with or without kuberentes). This plugin allows to sort out logs by log streams that can be parsed from metric tag. Log streams is created automatically.

### DIFF
--- a/plugins/outputs/cloudwatch_logs/cloudwatch_logs.go
+++ b/plugins/outputs/cloudwatch_logs/cloudwatch_logs.go
@@ -1,0 +1,434 @@
+package cloudwatch_logs
+
+import (
+	"fmt"
+	"log"
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/aws/aws-sdk-go/service/cloudwatchlogs"
+	"github.com/influxdata/telegraf"
+	internalaws "github.com/influxdata/telegraf/config/aws"
+	"github.com/influxdata/telegraf/plugins/outputs"
+)
+
+type messageBatch struct {
+	logEvents    []*cloudwatchlogs.InputLogEvent
+	messageCount int
+}
+type logStreamContainer struct {
+	currentBatchSizeBytes int
+	currentBatchIndex     int
+	messageBatches        []messageBatch
+	sequenceToken         string
+}
+
+//Cloudwatch Logs service interface
+type cloudWatchLogs interface {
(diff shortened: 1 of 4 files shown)


plugins/outputs/cloudwatch_logs/cloudwatch_logs.go
```diff
@@ -0,0 +1,434 @@
+package cloudwatch_logs
+
+import (
+	"fmt"
+	"log"
```
sspaink, line 5: Instead of using the log package can you update this to use the Telegraf logger:
```
Logger Telegraf.Logger `toml:"-"`
```
(in tests, you can do myPlugin.Log = testutil.Logger{})

plugins/outputs/cloudwatch_logs/cloudwatch_logs.go
```diff
@@ -0,0 +1,434 @@
...
+		} else {
+			return fmt.Errorf("Log data source is not properly formatted, ':' is missed.\n" +
+				"Should be 'tag:<tag_mame>' or 'field:<field_name>'")
+		}
+
+		if c.lsSource == "" {
+			c.lsSource = c.LogStream
+			log.Printf("D! %s Log stream '%s'...", logId, c.lsSource)
+		}
+	}
```
sspaink, line 184: Initialization and config checking should be done on the Init() error function, not in the Connect function. Can you move these lines to an Init() function please? I think the rest of the function looks like it belongs in the Connect function.

sspaink (changes requested):
Thank you for your work on this plugin, overall I think it looks great. I added some comments for you to look at.

plugins/outputs/cloudwatch_logs/cloudwatch_logs.go
i-prudnikov: Hello, sure, will fix that.

plugins/outputs/cloudwatch_logs/cloudwatch_logs.go
i-prudnikov: Agree, will fix that

i-prudnikov:
Hello @sspaink , just provided the fixes. Could you please review?

sspaink (approved):

PR #8641, opened 2021-01-03
Support Landing page on Prometheus landing page

### LINKED ISSUE #8640: Support Landing Page on Prometheus Client Output Plugin
## Feature Request

Request to add Landing page on Prometheus Client Output Plugin.

The Landing page is one of the recommended developments for the Prometheus exporter.

ref: https://prometheus.io/docs/instrumenting/writing_exporters/#landing-page

We are going to use the landing page for health check about service discovery.

We can use  "/metrics" path for health check, but if the metric size is large, unnecessary network costs occur.

Actually, we currently have a system that generates metrics over 20MB, so we want to do health check using the landing page.

### Proposal:

Add Landing page on Prometheus Client Output Plugin.

### Current behavior:

Use the "/metric" path for health check.

### Desired behavior:

Use the "/" path (landing page) for health check.

### Use case: 

Prometheus is commonly integrated with service discovery to identify target that is being metric collection.such as consul.

A health check is required to maintain list of healthy service, and a landing page is used to reduce unnecessary network costs.

hwanjin-jeong:
### Required for all PRs: 

Closes #8640

### DIFF
--- a/plugins/outputs/prometheus_client/prometheus_client.go
+++ b/plugins/outputs/prometheus_client/prometheus_client.go
@@ -159,12 +159,16 @@ func (p *PrometheusClient) Init() error {
 	authHandler := internal.AuthHandler(p.BasicUsername, p.BasicPassword, "prometheus", onAuthError)
 	rangeHandler := internal.IPRangeHandler(ipRange, onError)
 	promHandler := promhttp.HandlerFor(registry, promhttp.HandlerOpts{ErrorHandling: promhttp.ContinueOnError})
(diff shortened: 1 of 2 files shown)


sspaink (approved):
lgtm, at first I was concerned this would break backwards compatibility but looks like the readme says it should default to "/metric".

@hwanjin-jeong think you could rebase this pr with the latest master branch? Then the new linter will run and the tests will use the latest version of Go. Thank you!

helenosheaa (approved):

PR #8646, opened 2021-01-04
Open Hardware Monitor

marianob85:
### Required for all PRs:

- [x ] Signed [CLA](https://influxdata.com/community/cla/).
- [ x] Associated README.md updated.
- [ x] Has appropriate unit tests.

PR #8647, opened 2021-01-05
add Open Hardware Monitor as external plugin

sjwang90:
- add Open Hardware Monitor as external plugin
- alphabetize plugins

Closes #4596


srebhan:
See https://github.com/influxdata/telegraf/pull/8646 for adding the external label by the contributor of the plugin. Can we please split this PR and only fix the alphabetical order here?

PR #8657, opened 2021-01-06
outputs/warp10: url encode comma in tags value

BlackYoup:
Hello,

The warp10 specification requires the commas in tag values to be URL encoded as `%2C`. This PR fixes this behavior. 

I'm not very familiar with Golang, let me know if there is anything I should do better.

EDIT: Without this, Warp10 will reject the input and reply a parse error. This is the case with the current Elasticsearch input plugin for example.

### Required for all PRs:

### DIFF
--- a/plugins/outputs/warp10/warp10.go
+++ b/plugins/outputs/warp10/warp10.go
@@ -7,6 +7,7 @@ import (
 	"log"
 	"math"
 	"net/http"
+	"net/url"
 	"sort"
 	"strconv"
 	"strings"
@@ -174,7 +175,9 @@ func buildTags(tags []*telegraf.Tag) []string {
 	tagsString := make([]string, len(tags)+1)
 	indexSource := 0
 	for index, tag := range tags {
-		tagsString[index] = fmt.Sprintf("%s=%s", tag.Key, tag.Value)
+		key := url.QueryEscape(tag.Key)
+		value := url.QueryEscape(tag.Value)
+		tagsString[index] = fmt.Sprintf("%s=%s", key, value)
 		indexSource = index
 	}
 	indexSource++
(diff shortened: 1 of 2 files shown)


Jagularr (approved):
Makes sense to me

srebhan (approved):
Looks good to me.

plugins/outputs/warp10/warp10.go
```diff
@@ -174,7 +174,8 @@ func buildTags(tags []*telegraf.Tag) []string {
 	tagsString := make([]string, len(tags)+1)
 	indexSource := 0
 	for index, tag := range tags {
-		tagsString[index] = fmt.Sprintf("%s=%s", tag.Key, tag.Value)
+		value := strings.Replace(tag.Value, ",", "%2C", -1)
+		tagsString[index] = fmt.Sprintf("%s=%s", tag.Key, value)
```
ssoroka, line 178: What do you think about query escaping this with the same function used for URL encoding, so all special characters are handled? you could do this for the tag key as well.
```suggestion
		value := url.QueryEscape(tag.Value)
		tagsString[index] = fmt.Sprintf("%s=%s", tag.Key, value)
```

BlackYoup: Mh, it might be better indeed. After looking again at the Warp10 specification, it seems that both tag key and value should be url encoded, I just didn't encounter an issue with it. But at the same time, most of the characters would be fine, only `,`, `{` and `}` would be a real issue.

I think I'll adapt your commit to url encode the key too, would that be okay? Or I can commit your changes if you prefer to do it.

srebhan: I guess both is fine for @ssoroka. Will remove the "ready-to-merge" label and wait for your changes... :-) Thanks for taking care of this!

BlackYoup: Alright, I'll make the changes withing a few days, thanks :)

BlackYoup:
I rebased this branch onto master and I amended the commit to also url encode the tag key. Let me know if there is anything else I should do.

srebhan (approved):
LGTM.

PR #8661, opened 2021-01-07
cisco_telemetry_mdt enhancement

dsx1123:
### Required for all PRs:

### Summary of Changes
1.  Fixed NXAPI to handle more complicated output
2.  DME add support for Events and addd support for Class based query
3.  Property transformation logicSummary of Changes


plugins/inputs/cisco_telemetry_mdt/cisco_telemetry_mdt.go
```diff
@@ -50,13 +52,976 @@ type CiscoTelemetryMDT struct {
...
+    key := "show ip pim statistics vrf all"
+    c.nxpathMap[key] = make(map[string]string, 20)
+    c.nxpathMap[key]["vrf-name"] = "string"
+
+    return nil
+}
+
+func (c *CiscoTelemetryMDT) InitIntfBrief() error {
+    key := "show interface brief"
+    c.nxpathMap[key] = make(map[string]string, 20)
```
Jagularr, line 767: is there any reason to give all of these maps a length of 20?

plugins/inputs/cisco_telemetry_mdt/cisco_telemetry_mdt.go
```diff
@@ -37,6 +38,7 @@ type CiscoTelemetryMDT struct {
 	ServiceAddress string            `toml:"service_address"`
 	MaxMsgSize     int               `toml:"max_msg_size"`
 	Aliases        map[string]string `toml:"aliases"`
+        Dmes           map[string]string `toml:"dmes"`
```
Jagularr, line 41: this file needs go format ran on it

dsx1123: fixed in latest commit

plugins/inputs/cisco_telemetry_mdt/cisco_telemetry_mdt.go
dsx1123: we just give it a little bit extra space instead of calculate the exact number of keys of each command, in latest commit we fixed everything with number of entries, in few scenario we just use 20 as approximation.

dsx1123:
@Jagularr  do you mind take a look at the change?

plugins/inputs/cisco_telemetry_mdt/cisco_telemetry_mdt.go
```diff
@@ -50,13 +52,974 @@ type CiscoTelemetryMDT struct {
...
+		return str
+	case *telemetry.TelemetryField_Sint64Value:
+		str = strconv.FormatInt(val.Sint64Value, 10)
+		return str
+	}
+	return ""
+}
+
+//xform Uint64 to int64
+func (c *CiscoTelemetryMDT) nxosValueXformUint64Toint64(field *telemetry.TelemetryField, value interface{}) interface{} {
```
srebhan, line 98: I think you don't use `c` so this doesn't have to be a member function of `CiscoTelemetryMDT`. You might still want to keep as is but then please also change `xformValueString` to follow the same pattern.

plugins/inputs/cisco_telemetry_mdt/cisco_telemetry_mdt.go
```diff
@@ -50,13 +52,974 @@ type CiscoTelemetryMDT struct {
...
+	if field.GetUint64Value() != 0 {
+		return int64(value.(uint64))
+	}
+	return nil
+}
+
+//xform string to float
+func (c *CiscoTelemetryMDT) nxosValueXformStringTofloat(field *telemetry.TelemetryField, value interface{}) interface{} {
```
srebhan, line 106: You might want to remove `(c *CiscoTelemetryMDT)` as it is not needed. See `nxosValueXformUint64Toint64`.

plugins/inputs/cisco_telemetry_mdt/cisco_telemetry_mdt.go
```diff
@@ -50,13 +52,974 @@ type CiscoTelemetryMDT struct {
...
+		}
+		if val64, err := strconv.ParseInt(vals, 10, 64); err == nil {
+			return val64
+		}
+	} // switch
+	return nil
+}
+
+//auto-xform float properties
+func (c *CiscoTelemetryMDT) nxosValueAutoXformFloatProp(field *telemetry.TelemetryField, value interface{}) interface{} {
```
srebhan, line 160: You might want to remove `(c *CiscoTelemetryMDT)` as it is not needed. See `nxosValueXformUint64Toint64`.

plugins/inputs/cisco_telemetry_mdt/cisco_telemetry_mdt.go
```diff
@@ -50,13 +52,974 @@ type CiscoTelemetryMDT struct {
...
+	if vals != "" {
+		if valf, err := strconv.ParseFloat(vals, 64); err == nil {
+			return valf
+		}
+	} // switch
+	return nil
+}
+
+//xform uint64 to string
+func (c *CiscoTelemetryMDT) nxosValueXformUint64ToString(field *telemetry.TelemetryField, value interface{}) interface{} {
```
srebhan, line 172: You might want to remove `(c *CiscoTelemetryMDT)` as it is not needed. See `nxosValueXformUint64Toint64`.

plugins/inputs/cisco_telemetry_mdt/cisco_telemetry_mdt.go
```diff
@@ -50,13 +52,974 @@ type CiscoTelemetryMDT struct {
...
+	case *telemetry.TelemetryField_Uint64Value:
+		return (strconv.FormatUint(val.Uint64Value, 10))
+	}
+	return nil
+}
+
+//Xform value field.
+func (c *CiscoTelemetryMDT) nxosValueXform(field *telemetry.TelemetryField, value interface{}, path string) interface{} {
+	//if isNXOS
+	if !strings.ContainsRune(path, ':') {
```
srebhan, line 187: Please invert the logic to safe some indention levels
```suggestion
	if strings.ContainsRune(path, ':') {
		// not NXOS
		return nil
	}
```

plugins/inputs/cisco_telemetry_mdt/cisco_telemetry_mdt.go
```diff
@@ -50,13 +52,974 @@ type CiscoTelemetryMDT struct {
...
+		if _, ok := c.propMap[field.Name]; ok {
+			return (c.propMap[field.Name](field, value))
+		} else {
```
srebhan, line 190: No need for `else` as you return in the if already. This again saves an indention level and makes the code more readable.

plugins/inputs/cisco_telemetry_mdt/cisco_telemetry_mdt.go
```diff
@@ -50,13 +52,974 @@ type CiscoTelemetryMDT struct {
...
+			//check if we want auto xformation
+			if _, ok := c.propMap["auto-prop-xfromi"]; ok {
+				v1 := c.propMap["auto-prop-xfrom"](field, value)
+				if v1 != nil {
+					return v1
+				}
```
srebhan, line 196: What do you return if `v1` is `nil`? If I interpret the logic correctly you will return `nil` don't you? If so, you can simplify to
```suggestion
				return c.propMap["auto-prop-xfrom"](field, value)
```
and in turn remove the `else` below.

plugins/inputs/cisco_telemetry_mdt/cisco_telemetry_mdt.go
```diff
@@ -50,13 +52,974 @@ type CiscoTelemetryMDT struct {
...
+			} else {
+				//Now check path based conversion.
+				//If mapping is found then do the required transformation.
+				if c.nxpathMap[path] != nil {
```
srebhan, line 200: Again invert the logic
```suggestion
				if c.nxpathMap[path] == nil {
					return nil
				}
```
and save some indention level.

plugins/inputs/cisco_telemetry_mdt/cisco_telemetry_mdt.go
```diff
@@ -50,13 +52,974 @@ type CiscoTelemetryMDT struct {
...
+					case "integer":
+						switch val := field.ValueByType.(type) {
+						case *telemetry.TelemetryField_StringValue:
+							if vali, err := strconv.ParseInt(val.StringValue, 10, 32); err == nil {
+								return vali
+							}
+						case *telemetry.TelemetryField_Uint32Value:
+							return int(value.(uint32))
+						case *telemetry.TelemetryField_Uint64Value:
+							return int64(value.(uint64))
```
srebhan, line 212: Please check the type assertions before using them as otherwise telegraf will panic in case the assertion doesn't hold.

plugins/inputs/cisco_telemetry_mdt/cisco_telemetry_mdt.go
```diff
@@ -50,13 +52,974 @@ type CiscoTelemetryMDT struct {
...
+					case "float":
+						switch val := field.ValueByType.(type) {
+						case *telemetry.TelemetryField_StringValue:
+							if valf, err := strconv.ParseFloat(val.StringValue, 64); err == nil {
+								return valf
+							}
+						} //switch
+						return nil
+					case "string":
+						return (xformValueString(field))
```
srebhan, line 225: Please remove extra brackets.

plugins/inputs/cisco_telemetry_mdt/cisco_telemetry_mdt.go
```diff
@@ -50,13 +52,974 @@ type CiscoTelemetryMDT struct {
...
+	c.nxpathMap[key]["det_pinb"] = "string"
+	c.nxpathMap[key]["det_vinb"] = "string"
+	c.nxpathMap[key]["ps_state"] = "string"
+	c.nxpathMap[key]["det_sw_alarm"] = "string"
+	c.nxpathMap[key]["regnum"] = "string"
+	c.nxpathMap[key]["amps_requested"] = "string"
+	c.nxpathMap[key]["fanrpm"] = "string"
+	c.nxpathMap[key]["actual_input"] = "string"
+	c.nxpathMap[key]["outlet_temp"] = "string"
+	c.nxpathMap[key]["tot_capa"] = "string"
```
srebhan, line 330: To make more visible that all of those are `"string"` you might want to do:
```suggestion
	stringFields := []string {
		"reserve_sup",
		"det_volt",
		"heatsink_temp",
		...,
		"tot_capa",
	}
	for _, n := range stringFields {
		c.nxpathMap[key][n] = "string"
	}
```

plugins/inputs/cisco_telemetry_mdt/cisco_telemetry_mdt.go
```diff
@@ -50,13 +52,974 @@ type CiscoTelemetryMDT struct {
...
+
+	return nil
+}
+
+func (c *CiscoTelemetryMDT) InitMemPhys() error {
+	key := "show processes memory physical"
+	c.nxpathMap[key] = make(map[string]string, 1)
+	c.nxpathMap[key]["processname"] = "string"
+
+	return nil
```
srebhan, line 340: How about
```suggestion
	c.nxpathMap["show processes memory physical"] = map[string]string{"processname": "string"}

	return nil
```

srebhan: Btw. If there cannot be an error, please remove the return values!

plugins/inputs/cisco_telemetry_mdt/cisco_telemetry_mdt.go
```diff
@@ -50,13 +52,974 @@ type CiscoTelemetryMDT struct {
...
+	c.nxpathMap[key]["amps_requested"] = "string"
+	c.nxpathMap[key]["fanrpm"] = "string"
+	c.nxpathMap[key]["actual_input"] = "string"
+	c.nxpathMap[key]["outlet_temp"] = "string"
+	c.nxpathMap[key]["tot_capa"] = "string"
+
+	return nil
+}
+
+func (c *CiscoTelemetryMDT) InitMemPhys() error {
```
srebhan, line 335: Please make this function private.

plugins/inputs/cisco_telemetry_mdt/cisco_telemetry_mdt.go
```diff
@@ -50,13 +52,974 @@ type CiscoTelemetryMDT struct {
...
+	return nil
+}
+
+func (c *CiscoTelemetryMDT) InitBgpV4() error {
+	key := "show bgp ipv4 unicast"
+	c.nxpathMap[key] = make(map[string]string, 1)
+	c.nxpathMap[key]["aspath"] = "string"
+
+	return nil
+}
```
srebhan, line 349: See above.

plugins/inputs/cisco_telemetry_mdt/cisco_telemetry_mdt.go
```diff
@@ -50,13 +52,974 @@ type CiscoTelemetryMDT struct {
...
+	c.InitIpMroute()
+	c.InitIpv6Mroute()
+	c.InitVpc()
+	c.InitBgp()
+	c.InitCh()
+	c.InitIntf()
+	c.InitProcsys()
+	c.InitProc()
+	c.InitBfd()
+	c.InitLldp()
```
srebhan, line 1018: As seemingly those functions cannot error remove the `error` from the function signature of all those functions to indicate this fact.

plugins/inputs/cisco_telemetry_mdt/cisco_telemetry_mdt.go
```diff
@@ -50,13 +52,974 @@ type CiscoTelemetryMDT struct {
...
+
+	return nil
```
srebhan, line 1020: Same here. No error so remove the error return value.

plugins/inputs/cisco_telemetry_mdt/cisco_telemetry_mdt.go
```diff
@@ -391,9 +1394,72 @@ func (c *CiscoTelemetryMDT) parseKeyField(tags map[string]string, field *telemet
...
+	for _, subfield := range field.Fields {
+		//For Every table fill the keys which are vrfName, address and masklen
+		switch subfield.Name {
+		case "vrfName", "address", "maskLen":
+			tags[subfield.Name] = decodeTag(subfield)
+		}
+		if value := decodeValue(subfield); value != nil {
+			grouper.Add(measurement, tags, timestamp, subfield.Name, value)
+		}
+		if subfield.Name == "nextHop" {
```
srebhan, line 1409: Invert the logic here to save some indention level(s).
```suggestion
		if subfield.Name != "nextHop" {
			continue
		}
```

plugins/inputs/cisco_telemetry_mdt/cisco_telemetry_mdt.go
```diff
@@ -391,9 +1394,72 @@ func (c *CiscoTelemetryMDT) parseKeyField(tags map[string]string, field *telemet
...
+				for _, ff := range subf.Fields {
+					switch ff.Name {
+					case "address", "vrfName":
+						key := "nextHop/" + ff.Name
+						tags[key] = decodeTag(ff)
+					}
+					if value := decodeValue(ff); value != nil {
+						name := "nextHop/" + ff.Name
+						grouper.Add(measurement, tags, timestamp, name, value)
+					}
```
srebhan, line 1421: Did you mean
```suggestion
					case "address", "vrfName":
						key := "nextHop/" + ff.Name
						tags[key] = decodeTag(ff)
					default:
						if value := decodeValue(ff); value != nil {
							name := "nextHop/" + ff.Name
							grouper.Add(measurement, tags, timestamp, name, value)
						}
					}
```

plugins/inputs/cisco_telemetry_mdt/cisco_telemetry_mdt.go
```diff
@@ -391,9 +1394,72 @@ func (c *CiscoTelemetryMDT) parseKeyField(tags map[string]string, field *telemet
...
+	if path == "rib" {
+		//handle native data path rib
+		c.parseRib(grouper, field, prefix, path, tags, timestamp)
+		return
+	}
+	if field == nil || !isDme {
+		return
+	}
+
+	if field.Fields[0] != nil && field.Fields[0].Fields != nil && field.Fields[0].Fields[0] != nil && field.Fields[0].Fields[0].Fields[0].Name != "attributes" {
```
srebhan, line 1442: Please check beforehand that `field.Fields` has at least one entry as otherwise accessing the first element will segfault!

plugins/inputs/cisco_telemetry_mdt/cisco_telemetry_mdt.go
```diff
@@ -430,11 +1500,26 @@ func (c *CiscoTelemetryMDT) parseContentField(grouper *metric.SeriesGrouper, fie
...
+			if !isEVENT {
+				nxChildren = subfield
+			} else {
+				sub := subfield.Fields
+				if sub[0] != nil && sub[0].Fields[0].Name == "subscriptionId" && len(sub[0].Fields) >= 2 {
+					nxAttributes = sub[0].Fields[1].Fields[0].Fields[0].Fields[0].Fields[0].Fields[0]
+				}
+			}
+			//if nxAttributes == NULL then class based query.
+			if nxAttributes == nil {
```
srebhan, line 1517: Please initialize `nxChildren` as `nil` as in case `isEVENT == true` and `if sub[0] != nil && sub[0].Fields[0].Name == "subscriptionId" && len(sub[0].Fields) >= 2` being `false`  this variable is not set and might contain values from the previous loop-cycle.

plugins/inputs/cisco_telemetry_mdt/cisco_telemetry_mdt.go
```diff
@@ -516,21 +1608,20 @@ const sampleConfig = `
...
  # tls_cert = "/etc/telegraf/cert.pem"
  # tls_key = "/etc/telegraf/key.pem"
-
  ## Enable TLS client authentication and define allowed CA certificates; grpc
  ##  transport only.
  # tls_allowed_cacerts = ["/etc/telegraf/clientca.pem"]
-
  ## Define (for certain nested telemetry measurements with embedded tags) which fields are tags
  # embedded_tags = ["Cisco-IOS-XR-qos-ma-oper:qos/interface-table/interface/input/service-policy-names/service-policy-instance/statistics/class-stats/class-name"]
-
```
srebhan, line 530: Please keep the newlines.

plugins/inputs/cisco_telemetry_mdt/cisco_telemetry_mdt.go
```diff
@@ -516,21 +1608,20 @@ const sampleConfig = `
...
  ## Define aliases to map telemetry encoding paths to simple measurement names
  [inputs.cisco_telemetry_mdt.aliases]
    ifstats = "ietf-interfaces:interfaces-state/interface/statistics"
+ [inputs.cisco_telemetry_mdt.dmes]
+   ModTs = "ignore"
+   CreateTs = "ignore"
```
srebhan, line 1624: Can you please give an explanation what those parameters mean (maybe including a reference to a manual or similar) including which options and values are possible. Otherwise a non-Cisco-employee will not know what to specify here.

plugins/inputs/cisco_telemetry_mdt/cisco_telemetry_mdt_test.go
```diff
@@ -21,7 +21,6 @@ func TestHandleTelemetryTwoSimple(t *testing.T) {
 	err := c.Start(acc)
 	// error is expected since we are passing in dummy transport
 	require.Error(t, err)
-
```
srebhan, line 24: Keep.

plugins/inputs/cisco_telemetry_mdt/cisco_telemetry_mdt_test.go
```diff
@@ -238,6 +237,7 @@ func TestHandleNXAPI(t *testing.T) {
 	// error is expected since we are passing in dummy transport
 	require.Error(t, err)
 
+
```
srebhan, line 240: Remove. :-)

plugins/inputs/cisco_telemetry_mdt/cisco_telemetry_mdt_test.go
```diff
@@ -311,14 +311,172 @@ func TestHandleNXAPI(t *testing.T) {
...
 }
 
+func TestHandleNXAPIXformNXAPI(t *testing.T) {
+	c := &CiscoTelemetryMDT{Log: testutil.Logger{}, Transport: "dummy", Aliases: map[string]string{"nxapi": "show nxapi"}}
+	acc := &testutil.Accumulator{}
+	err := c.Start(acc)
+	// error is expected since we are passing in dummy transport
+	require.Error(t, err)
+
+
```
srebhan, line 329: Only one newline. Please remove one.

plugins/inputs/cisco_telemetry_mdt/parser.json
```diff
@@ -0,0 +1 @@
...": "unit","Value": "string"}, {"Key": "hwVer","Value": "string"}]},{"Name": "sys/intf","prop": [{"Key": "descr","Value": "string"}, {"Key": "name","Value": "string"}, {"Key": "lastStCause","Value": "string"}, {"Key": "description","Value": "string"}, {"Key": "unit","Value": "string"}, {"Key": "operFECMode","Value": "string"}, {"Key": "operBitset","Value": "string"}, {"Key": "mdix","Value": "string"}]},{"Name": "sys/procsys","prop": [{"Key": "name","Value": "string"}, {"Key": "id","Value": "string"}, {"Key": "upTs","Value": "string"}, {"Key": "interval","Value": "string"}, {"Key": "memstatus","Value": "string"}]},{"Name": "sys/proc","prop": [{"Key": "processName","Value": "string"}, {"Key": "procArg","Value": "string"}]},{"Name": "sys/bfd/inst","prop": [{"Key": "descr","Value": "string"}, {"Key": "vrfName","Value": "string"}, {"Key": "name","Value": "string"}, {"Key": "name","Value": "string"}]},{"Name": "sys/lldp","prop": [{"Key": "sysDesc","Value": "string"}, {"Key": "portDesc","Value": "string"}, {"Key": "portIdV","Value": "string"}, {"Key": "chassisIdV","Value": "string"}, {"Key": "sysName","Value": "string"}, {"Key": "name","Value": "string"}, {"Key": "id","Value": "string"}]}]
```
srebhan, line 1: Do we still need this file?

srebhan (changes requested):
This code looks much better! Some general comments (more in the code):

1. Please make as many functions private as possible.
2. Please remove `error` return values if the function cannot fail.
3. Please save some indention levels by returning early from functions, this makes the code more readable.
4. Maybe you can move the `Init...()` and data-conversion functions to a separate file withing the same package to not clutter the "main" file too much. It's only a suggestion.
5. Please check if we still need `parser.json`.

plugins/inputs/cisco_telemetry_mdt/cisco_telemetry_mdt.go
dekumar23: All these functions are updating nxpathMap that's why they are defined with
func (c *CiscoTelemetryMDT) InitMemPhys()
c.nxpathMap[key]["processname"] = "string"

dekumar23: Did you mean following changes to make function private 
-func (c *CiscoTelemetryMDT) InitProc() error {
+func InitProc(c *CiscoTelemetryMDT) {

plugins/inputs/cisco_telemetry_mdt/cisco_telemetry_mdt.go
srebhan: Well the maps are dynamic and can be initialized with zero. Giving a capacity only helps to reduce the reallocations in case you have an idea where you end-up.

plugins/inputs/cisco_telemetry_mdt/cisco_telemetry_mdt.go
srebhan: No I want to make it `func (c *CiscoTelemetryMDT) initProc() error` do it will definitively stay inside of this package. In case a function starts with lower-case it will not be visible from outside the package.

plugins/inputs/cisco_telemetry_mdt/cisco_telemetry_mdt.go
```diff
@@ -391,9 +1328,72 @@ func (c *CiscoTelemetryMDT) parseKeyField(tags map[string]string, field *telemet
...
+				for _, ff := range subf.Fields {
+					switch ff.Name {
+					case "address", "vrfName":
+						key := "nextHop/" + ff.Name
+						tags[key] = decodeTag(ff)
+					}
+					if value := decodeValue(ff); value != nil {
+						name := "nextHop/" + ff.Name
+						grouper.Add(measurement, tags, timestamp, name, value)
+					}
```
srebhan, line 1355: Do you mean
```suggestion
					case "address", "vrfName":
						key := "nextHop/" + ff.Name
						tags[key] = decodeTag(ff)
						continue
					}
					if value := decodeValue(ff); value != nil {
						name := "nextHop/" + ff.Name
						grouper.Add(measurement, tags, timestamp, name, value)
					}
```

as otherwise you process `ff` twice in case of `address` and `vrfName`...

srebhan (changes requested):
Can you please move all `nxosValueXform*` functions to a separate file (e.g. `cisco_telemetry_nxos.go`) that also belongs to `package cisco_telemetry_mdt`! This way the main file is not so cluttered with the conversion and init functions for the new feature.
Please also check the issue brought up by LGTM!

dsx1123:
> Can you please move all `nxosValueXform*` functions to a separate file (e.g. `cisco_telemetry_nxos.go`) that also belongs to `package cisco_telemetry_mdt`! This way the main file is not so cluttered with the conversion and init functions for the new feature.
> Please also check the issue brought up by LGTM!

we should have fixed all the comment issue, just one more thing, the lgtm is complained about 
> This definition of nxAttributes is never used.

but it is used blow:
```
nxAttributes = field.Fields[0].Fields[0].Fields[0].Fields[0]

for _, subfield := range nxAttributes.Fields {
```

plugins/inputs/cisco_telemetry_mdt/cisco_telemetry_mdt.go
```diff
@@ -50,13 +52,910 @@ type CiscoTelemetryMDT struct {
...
+}
+
+//Xform value field.
+func (c *CiscoTelemetryMDT) nxosValueXform(field *telemetry.TelemetryField, value interface{}, path string) interface{} {
+	if strings.ContainsRune(path, ':') {
+		// not NXOS
+		return nil
+	}
+	if _, ok := c.propMap[field.Name]; ok {
+		return (c.propMap[field.Name](field, value))
```
srebhan, line 191: Please remove the extra parentheses
```suggestion
		return c.propMap[field.Name](field, value)
```

plugins/inputs/cisco_telemetry_mdt/cisco_telemetry_mdt.go
```diff
@@ -50,13 +52,910 @@ type CiscoTelemetryMDT struct {
...
+	} else {
```
srebhan, line 192: You don't require an `else` here as you `return` in the `if` path. Please remove the else and save another indention level.

srebhan:
@dsx1123 very good, almost there. Can you please move all new functions starting with `nx` or `init` to an own file!? This way the main file is not so cluttered and it's easier to find the functions you are looking for.

srebhan:
@dsx1123 please fix the formatting errors using `make fmt`. Please also run `make check` locally to detect such errors before pushing...

plugins/inputs/cisco_telemetry_mdt/cisco_telemetry_mdt.go
```diff
@@ -50,13 +52,48 @@ type CiscoTelemetryMDT struct {
 
 	// Internal state
 	aliases   map[string]string
+	dmes      map[string]string
```
srebhan, line 55: Can you please rename this (e.g. to `dmesFuncs` or similar) or `Dmes` above to avoid confusion between the two.

plugins/inputs/cisco_telemetry_mdt/cisco_telemetry_mdt.go
```diff
@@ -50,13 +52,48 @@ type CiscoTelemetryMDT struct {
...
+		return str
+	case *telemetry.TelemetryField_Sint32Value:
+		str = strconv.FormatInt(int64(val.Sint32Value), 10)
+		return str
+	case *telemetry.TelemetryField_Sint64Value:
+		str = strconv.FormatInt(val.Sint64Value, 10)
+		return str
+	}
+	return ""
+}
```
srebhan, line 95: Should this also move to the `_util.go` file?

plugins/inputs/cisco_telemetry_mdt/cisco_telemetry_mdt.go
```diff
@@ -66,12 +103,50 @@ func (c *CiscoTelemetryMDT) Start(acc telegraf.Accumulator) error {
...
+			err := json.Unmarshal(js, &jsStruct)
+			if err == nil {
+				//Build 2 level Hash nxpathMap Key = jsStruct.Name, Value = map of jsStruct.Prop
+				//It will override the default of code if same path is provided in configuration.
+				c.nxpathMap[jsStruct.Name] = make(map[string]string, len(jsStruct.Prop))
+				for _, prop := range jsStruct.Prop {
+					c.nxpathMap[jsStruct.Name][prop.Key] = prop.Value
+				}
+			}
+		}
```
srebhan, line 148: We should make this a switch statement
```suggestion
		switch path {
		case "uint64 to int":
			c.propMap[dme] = nxosValueXformUint64Toint64
		case "uint64 to string":
			c.propMap[dme] = nxosValueXformUint64ToString
		case "string to float64":
			c.propMap[dme] = nxosValueXformStringTofloat
		case "string to uint64":
			c.propMap[dme] = nxosValueXformStringToUint64
		case "string to int64":
			c.propMap[dme] = nxosValueXformStringToInt64
		case "auto-float-xfrom"
			c.propMap[dme] = nxosValueAutoXformFloatProp
		default:
			if ! strings.HasPrefix(dme, "dnpath") { // not path based property map
				continue
			}
			
			var jsStruct NxPayloadXfromStructure
			if err := json.Unmarshal([]byte(path), &jsStruct); if err != nil {
				continue
			}
				
			// Build 2 level Hash nxpathMap Key = jsStruct.Name, Value = map of jsStruct.Prop
			// It will override the default of code if same path is provided in configuration.
			c.nxpathMap[jsStruct.Name] = make(map[string]string, len(jsStruct.Prop))
			for _, prop := range jsStruct.Prop {
				c.nxpathMap[jsStruct.Name][prop.Key] = prop.Value
			}
		}
```

plugins/inputs/cisco_telemetry_mdt/cisco_telemetry_mdt.go
```diff
@@ -296,7 +371,9 @@ func (c *CiscoTelemetryMDT) handleTelemetry(data []byte) {
 		// Parse keys
 		tags = make(map[string]string, len(keys.Fields)+3)
 		tags["source"] = msg.GetNodeIdStr()
-		tags["subscription"] = msg.GetSubscriptionIdStr()
+		if msg.GetSubscriptionIdStr() != "" {
+			tags["subscription"] = msg.GetSubscriptionIdStr()
+		}
```
srebhan, line 376: We can save a function call here
```suggestion
		if msgID := msg.GetSubscriptionIdStr(); msgID != "" {
			tags["subscription"] = msgID
		}
```

plugins/inputs/cisco_telemetry_mdt/cisco_telemetry_mdt.go
srebhan: Please mark this as resolved.

srebhan (commented):
The (almost) final pass brought up two or three more minor requests. Can you please fix them and I think then we are good to go.

srebhan (approved):
Looks good to me.

plugins/inputs/cisco_telemetry_mdt/cisco_telemetry_mdt.go
```diff
@@ -430,11 +555,28 @@ func (c *CiscoTelemetryMDT) parseContentField(grouper *metric.SeriesGrouper, fie
...
 	for _, subfield := range field.Fields {
 		if isNXOS && subfield.Name == "attributes" && len(subfield.Fields) > 0 {
 			nxAttributes = subfield.Fields[0]
 		} else if isNXOS && subfield.Name == "children" && len(subfield.Fields) > 0 {
-			nxChildren = subfield
+			if !isEVENT {
+				nxChildren = subfield
+			} else {
+				sub := subfield.Fields
+				if sub[0] != nil && sub[0].Fields[0].Name == "subscriptionId" && len(sub[0].Fields) >= 2 {
```
ssoroka, line 569: does sub[0] need a length check here?

plugins/inputs/cisco_telemetry_mdt/cisco_telemetry_mdt.go
ssoroka: this is still relevant

plugins/inputs/cisco_telemetry_mdt/cisco_telemetry_mdt.go
```diff
@@ -391,9 +449,72 @@ func (c *CiscoTelemetryMDT) parseKeyField(tags map[string]string, field *telemet
...
+		return
+	}
+	if field == nil || !isDme {
+		return
+	}
+
+	if len(field.Fields) >= 1 && field.Fields[0] != nil && field.Fields[0].Fields != nil && field.Fields[0].Fields[0] != nil && field.Fields[0].Fields[0].Fields[0].Name != "attributes" {
+		return
+	}
+	nxAttributes = field.Fields[0].Fields[0].Fields[0].Fields[0]
```
ssoroka, line 500: This logic seems to have a bunch of bad paths through it. eg:
- if `len(field.Fields)` is 0, line 500 will panic
- if len(field.Fields[0].Fields) is 0, 500 will panic.
etc.
try exiting on all the cases where len() is 0. nil checks aren't necessary on slices because len() of a nil slice returns 0.

ssoroka (changes requested):

plugins/inputs/cisco_telemetry_mdt/cisco_telemetry_mdt.go
srebhan: You are right and I think the check on line 497 is the opposite of what it should be, isn't it?

plugins/inputs/cisco_telemetry_mdt/cisco_telemetry_mdt.go
sspaink: Looks like the code is now checking the length of `fields.Fields` is greater than zero, marking this as resolved.

plugins/inputs/cisco_telemetry_mdt/cisco_telemetry_mdt.go
sspaink: Latest change adds a len check

plugins/inputs/cisco_telemetry_mdt/cisco_telemetry_mdt.go
sspaink: Looks like the updated checks address the concern of Fields being empty or nil, marking this conversation as resolved.

sspaink:
@ssoroka I looked over your requested changes and looks like @dsx1123 has addressed them and I marked the conversations as resolved. I don't see anything else, but do you have anything else you would like to add before merging this? Thanks!

ssoroka (approved):

helenosheaa (approved):

Issue #8571, opened 2020-12-16
outputs.influxdb_v2 keeps looping on 500 response.


eremmel:
### Relevant telegraf.conf:
[[outputs.influxdb_v2]]
  Any working config will do.

[[inputs.influxdb_listener]]
  Any local input will do

### System info:

telegraf 1.16.3
influxdb 2.x

### Steps to reproduce:

I've configured a bucket with a retention policy of 2 weeks. When I write some measurements that are older than those two weeks, influxdb response with an 500 error. 

1. setup influxdb v2 
2. create bucket with a retention policy of N weeks
3. setup outputs.influxdb_v2 to connect to this bucket
4. start telegraf with debug
5. send some data to this bucket that is older than N weeks. E.g. via curl.
6. the outputs.influxdb_v2 will keep trying to resend the data to influxdb 2.
7. New measurements pile up after this request. 

### Expected behavior:

Report the error and discard the batch of measurements.

### Actual behavior:

Error is reported but the batch of measurements is offered over and over again to influxdb

### Error trace:
2020-12-16T11:02:09Z D! [outputs.influxdb_v2] Buffer fullness: 278 / 10000 metrics
2020-12-16T11:02:09Z E! [agent] Error writing to outputs.influxdb_v2: 500 Internal Server Error: internal error: unexpected error writing points to database: partial write: points beyond retention policy dropped=139
2020-12-16T11:02:19Z D! [outputs.http] Buffer fullness: 0 / 10000 metrics
2020-12-16T11:02:19Z D! [outputs.influxdb] Wrote batch of 139 metrics in 24.9356ms
2020-12-16T11:02:19Z D! [outputs.influxdb] Buffer fullness: 0 / 10000 metrics
2020-12-16T11:02:19Z D! [outputs.influxdb] Wrote batch of 139 metrics in 25.9331ms
2020-12-16T11:02:19Z D! [outputs.influxdb] Buffer fullness: 0 / 10000 metrics
2020-12-16T11:02:19Z E! [outputs.influxdb_v2] when writing to [http://54.226.129.163]: 500 Internal Server Error: internal error: unexpected error writing points to database: partial write: points beyond retention policy dropped=139
2020-12-16T11:02:19Z D! [outputs.influxdb_v2] Buffer fullness: 417 / 10000 metrics
[...]

PR #8662, opened 2021-01-08
[outputs.influxdb_v2] add exponential backoff, and respect client error responses

ssoroka:
- change max wait on error to 30s
- handle more http response codes, both for future-proofing and for playing nice with gateways and proxies
- 4xx response codes now result in metrics being dropped, because server will never accept them. This prevents infinite delivery of failing metrics, blocking transmission at head of line.
- add exponential backoff as well as respecting retry-after header. Pick whichever is more up to 30s max
- include the server response code in the logs, as well as clarify if the metric was dropped.

resolves https://github.com/influxdata/telegraf/issues/8571

### DIFF
--- a/plugins/outputs/influxdb_v2/http.go
+++ b/plugins/outputs/influxdb_v2/http.go
@@ -37,7 +37,7 @@ func (e APIError) Error() string {
 
 const (
 	defaultRequestTimeout = time.Second * 5
-	defaultMaxWait        = 30 // seconds
+	defaultMaxWait        = 60 // seconds
 	defaultDatabase       = "telegraf"
 )
 
--- a/plugins/outputs/influxdb_v2/http.go
+++ b/plugins/outputs/influxdb_v2/http.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"io/ioutil"
 	"log"
+	"math"
 	"net"
 	"net/http"
 	"net/url"
@@ -36,7 +37,7 @@ func (e APIError) Error() string {
 
 const (
 	defaultRequestTimeout = time.Second * 5
-	defaultMaxWait        = 10 // seconds
+	defaultMaxWait        = 30 // seconds
 	defaultDatabase       = "telegraf"
 )
 
@@ -70,6 +71,7 @@ type httpClient struct {
 	serializer *influx.Serializer
 	url        *url.URL
 	retryTime  time.Time
+	retryCount int
 }
 
 func NewHTTPClient(config *HTTPConfig) (*httpClient, error) {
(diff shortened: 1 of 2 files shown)


Jagularr (approved):

plugins/outputs/influxdb_v2/http.go
```diff
@@ -287,6 +304,29 @@ func (c *httpClient) writeBatch(ctx context.Context, bucket string, metrics []te
...
+		var err error
+		retryAfterHeader, err = strconv.ParseFloat(retryAfterHeaderString, 64)
+		if err != nil {
+			// there was a value but we couldn't parse it? guess minimum 10 sec
+			retryAfterHeader = 10
+		}
+	}
+	// take the highest value from both, but not over the max wait.
+	retry := math.Max(backoff, retryAfterHeader)
+	retry = math.Min(retry, defaultMaxWait)
```
reimda, line 326: I would expect telegraf to obey the Retry-After header even if it's longer than defaultMaxWait.  If the backend is in trouble and ops needs to quiet down retries, they will increase Retry-After and telegraf should obey it.

sspaink:
!signed-cla

ssoroka:
!signed-cla

plugins/outputs/influxdb_v2/http.go
ssoroka: yeah, agreed. I had checked with the db team, and all of their use cases either don't use it (OSS 2.x doesn't use it at all), or it's only used as rate limiting. I suspect it would be prudent to still have a maximum as I wouldn't really want it to sit idle for 2 hours due to an expiry bug. I think we should set defaultMaxWait to whatever we're comfortable with as a max wait time and leave it at that. I can't really see anything above 60s being hugely valuable, but it seems a lot more likely to cause problems.

I'm proposing upping the defaultMaxWait to 60s. let me know what you think.

reimda: I prefer trusting the header unconditionally but if you don't want to, it's not a big deal for me because the situation we're talking about is unlikely anyway.  Increasing absolute max is the next best thing, whether it's 1 or 5 or 10 minutes.

reimda (approved):

plugins/outputs/influxdb_v2/http.go
```diff
@@ -36,7 +37,7 @@ func (e APIError) Error() string {
 
 const (
 	defaultRequestTimeout = time.Second * 5
-	defaultMaxWait        = 10 // seconds
+	defaultMaxWait        = 30 // seconds
```
ssoroka, line 40:
```suggestion
	defaultMaxWait        = 60 // seconds
```

srebhan (approved):
LGTM!

PR #8665, opened 2021-01-09
fix bug: inputs.chrony :Init function is not executed resulting in empty path

fcddk:
### Required for all PRs:

fix bug: inputs.chrony :Init function is not executed resulting in empty path

### DIFF
--- a/plugins/inputs/chrony/chrony.go
+++ b/plugins/inputs/chrony/chrony.go
@@ -43,12 +43,11 @@ func (c *Chrony) init() error {
 }
 
 func (c *Chrony) Gather(acc telegraf.Accumulator) error {
-	// manually execute chronyc lookup
-	err := c.init() 
-        if err != nil {
+	var err error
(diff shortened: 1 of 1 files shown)


srebhan (changes requested):
No! This is definitively not correct! The `Init()` function (capital I) is executed_once_ by telegraf _after_ parsing (and setting) the config variables! Now you execute the function in _EVERY_ gather cycle and I'm pretty sure that looking-up the executable path every interval is not necessary...
If there is a problem with `Init()` please let us know, but as I said this is definitively not the correct fix!

ssoroka (changes requested):
As sven pointed out, Init is called automatically by Telegraf. if you're seeing problems with empty paths, it's not because Init isn't being called.

srebhan:
@fcddk I'll close this as it is not the right fix and I'm not even sure I got your problem. If you still see the issue, please reopen this PR or create an issue!

PR #8671, opened 2021-01-11
add input odyssey plugin

Xrias:
### Required for all PRs:

### DIFF
--- a/plugins/inputs/odyssey/README.md
+++ b/plugins/inputs/odyssey/README.md
@@ -1,0 +1,73 @@
+### Configuration example
+
+```toml
+[[inputs.odyssey]]
+  ## specify address via a url matching:
+  ##   postgres://[console[:password]]@host:port[/dbname]\
+  ##       ?sslmode=[disable|verify-ca|verify-full]
+  ## or a simple string:
+  ##   host=localhost port=6432 user=postgres password=... sslmode=... dbname=app_production
+  ##
+  ## All connection parameters are optional.
+  ##
+  address = "host=localhost user=postgres sslmode=disable"
+```
+
+#### `address`
+
+Specify address via a postgresql connection string:
+
+  `host=/run/postgresql port=6432 user=postgres database=console`
+
+Or via an url matching:
+
+  `postgres://[console[:password]]@host:port[/dbname]?sslmode=[disable|verify-ca|verify-full]`
+
+All connection parameters are optional.
+
+Without the dbname parameter, the driver will default to a database with the same name as the user.
+This dbname is just for instantiating a connection with the server and doesn't restrict the databases we are trying to grab metrics for.
+
+### Metrics
+
+- odyssey
+  - tags:
+    - db
+    - server
+  - fields:
+    - avg_query_count
+    - avg_query_time
+    - avg_wait_time
+    - avg_xact_count
+    - avg_xact_time
+    - total_query_count
+    - total_query_time
+    - total_received
+    - total_sent
+    - total_wait_time
+    - total_xact_count
+    - total_xact_time
+
++ odyssey_pools
+  - tags:
+    - db
+    - pool_mode
+    - server
+    - user
+  - fields:
+    - cl_active
+    - cl_waiting
+    - maxwait
+    - maxwait_us
+    - sv_active
+    - sv_idle
+    - sv_login
+    - sv_tested
+    - sv_used
+
+### Example Output
+
+```
+console,db=console,server=host\=debian-buster-postgres\ user\=dbn\ port\=6432\ dbname\=console\  avg_query_count=0i,avg_query_time=0i,avg_wait_time=0i,avg_xact_count=0i,avg_xact_time=0i,total_query_count=26i,total_query_time=0i,total_received=0i,total_sent=0i,total_wait_time=0i,total_xact_count=26i,total_xact_time=0i 1581569936000000000
+odyssey_pools,db=console,pool_mode=statement,server=host\=debian-buster-postgres\ user\=dbn\ port\=6432\ dbname\=console\ ,user=postgres cl_active=1i,cl_waiting=0i,maxwait=0i,maxwait_us=0i,sv_active=0i,sv_idle=0i,sv_login=0i,sv_tested=0i,sv_used=0i 1581569936000000000
+```
--- a/plugins/inputs/odyssey/README.md
+++ b/plugins/inputs/odyssey/README.md
@@ -1,3 +1,10 @@
+# Odyssey Input Plugin
+
+The `odyssey` plugin provides metrics for your Odyssey connection pooler.
+
+More information about the meaning of these metrics can be found in the
+[Odyssey Documentation](https://github.com/yandex/odyssey).
+
 ### Configuration example
 
 ```toml
@@ -12,6 +19,8 @@
   ##
   address = "host=localhost user=postgres sslmode=disable"
 
+  ## Default queries - ['SHOW_POOLS', 'SHOW_ERRORS', 'SHOW_CLIENTS', 'SHOW_STATS']
+  ## A list of queries to include. If not specified, all the above listed queries are used.
   include_query = []
 ```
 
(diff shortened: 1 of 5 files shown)


reshke:
🆗 👍

reshke (approved):
as an odyssey developer i think this plugin is cool and would love to use it

plugins/inputs/odyssey/README.md
```diff
@@ -0,0 +1,75 @@
+### Configuration example
+
```
srebhan, line 2: Could you please add a short description what this input is about and what it is supposed to do. Furthermore, for people not knowing what _odyssey_ is about, it would help to add a reference or link to the project.

plugins/inputs/odyssey/odyssey.go
```diff
@@ -0,0 +1,158 @@
...
+  ## or a simple string:
+  ##   host=localhost user=console password=... sslmode=... dbname=app_production
+  ##
+  ## include_query = []string - commands for psql
+  ##
+  ## All connection parameters are optional.
+  ##
+  address = "host=localhost user=postgresql sslmode=disable dbname=console port=6432"
+
+  include_query = ['SHOW STATS', 'SHOW POOLS']
```
srebhan, line 35: Hmmm... You require the user to specify an `include_query` as otherwise you will not collect any metric. How about making the specified queries the default ones (i.e. use those if no include_query is given) or at least check (e.g. in an `Init()` function) if the parameter is empty?
It would also help to move the parameter description directly above the parameter instead of having a common description section. You might want to specify more possible values here...

srebhan (commented):
I have some first comments in the code. But what puzzles me is how this plugin differs from the existing `postgresql_extensible` plugin. Could you please outline what the difference is and why this cannot be included in the mentioned plugin!?

Xrias:
@srebhan Thanks for the review Sven. Odyssey does not support Extended Query https://www.postgresql.org/docs/8.3/protocol-flow.html (45.2.3). We would also like to set our default settings for the Odyssey.

srebhan:
@Xrias if I do understand your comment correctly `SHOW STATS` and `SHOW POOLS` are the only (???) options for queries and you would like to have them as default. If this is correct I fully agree with you that this makes sense.

The only difference we have is on how to implement this default. Almost all plugins sent such a sensible default if the option in question is not specified by the user. So what I want you to do is, to comment the option in the config and the implement the following logic: If `include_query` is empty, set the default set of queries. Otherwise accept what the user specifies after validating the input.
That's it. The advantage is that the user _only_ has to know the address and probably gets what he/she wants... :-)

Xrias:
@srebhan My last comment was about the need for a separate plugin. Your remark about «include query» will definetely be corrected =)

srebhan:
Ah I see, sorry for missing it in my first answer. Could you also take a look at https://github.com/influxdata/telegraf/pull/8735. It's my attempt to get a generic SQL input-plugin. It could use some testing and I wonder if it also covers your use-case!?

Anyway looking forward to your code and comments. :-)

plugins/inputs/odyssey/odyssey.go
Jagularr: agreed

Jagularr (commented):
(outdated)

srebhan (commented):
Looks good to me.

@Xrias if you find some time testing #8735 I would be very interested in your opinion, comment or potential issues.

srebhan (approved):
Looks good to me.

@Xrias if you find some time testing #8735 I would be very interested in your opinion, comment or potential issues.

plugins/inputs/odyssey/odyssey.go
```diff
@@ -0,0 +1,165 @@
...
+  ## A list of queries to include. If not specified, all the above listed queries are used.
+  include_query = []
+`
+
+func (p *Odyssey) SampleConfig() string {
+	return sampleConfig
+}
+
+func (p *Odyssey) Description() string {
+	return "Read metrics from Odyssey"
```
Jagularr, line 43: nit: should be "Reads"

plugins/inputs/odyssey/odyssey.go
```diff
@@ -0,0 +1,165 @@
...
+
+func (p *Odyssey) Gather(acc telegraf.Accumulator) error {
+	var (
+		err     error
+		query   string
+		queries []string
+		columns []string
+	)
+
+	queries = []string{`SHOW POOLS`, `SHOW CLIENTS`, `SHOW ERRORS`, `SHOW STATS`}
```
Jagularr, line 54: this should probably be a constant variable set at the top of the file

plugins/inputs/odyssey/odyssey.go
```diff
@@ -0,0 +1,165 @@
...
+	return "Read metrics from Odyssey"
+}
+
+func (p *Odyssey) Gather(acc telegraf.Accumulator) error {
+	var (
+		err     error
+		query   string
+		queries []string
+		columns []string
+	)
```
Jagularr, line 52: I don't think any of this is necessary, all can be initiated with := when they are first made.

plugins/inputs/odyssey/odyssey.go
```diff
@@ -0,0 +1,165 @@
...
+
+	queries = []string{`SHOW POOLS`, `SHOW CLIENTS`, `SHOW ERRORS`, `SHOW STATS`}
+
+	if len(p.IncludeQuery) > 0 {
+		queries = p.IncludeQuery
+	}
+
+	for _, value := range queries {
+		query = value
```
Jagularr, line 61: unnecessary, use value variable directly

plugins/inputs/odyssey/odyssey.go
```diff
@@ -0,0 +1,165 @@
...
+)
+
+type Odyssey struct {
+	postgresql.Service
+	IncludeQuery []string `toml:"include_query"`
+}
+
+var ignoredColumns = map[string]bool{"user": true, "database": true, "pool_mode": true,
+	"avg_req": true, "avg_recv": true, "avg_sent": true, "avg_query": true,
+}
```
Jagularr, line 20: should this be configurable?

plugins/inputs/odyssey/odyssey.go
```diff
@@ -0,0 +1,165 @@
...
+
+type scanner interface {
+	Scan(dest ...interface{}) error
+}
+
+func (p *Odyssey) accRow(row scanner, acc telegraf.Accumulator, columns []string) (map[string]string,
+	map[string]*interface{}, error) {
+	var columnVars []interface{}
+	var dbname bytes.Buffer
```
Jagularr, line 115: you're using a bytes.Buffer to assign convert strings into and then converting it back to a string at the end. I'd suggest just using a string lol

plugins/inputs/odyssey/odyssey.go
```diff
@@ -0,0 +1,165 @@
...
+			if poolMode, ok := columnMap["pool_mode"]; ok {
+				if s, ok := (*poolMode).(string); ok && s != "" {
+					tags["pool_mode"] = s
+				}
+			}
+
+			fields := make(map[string]interface{})
+			for col, val := range columnMap {
+				_, ignore := ignoredColumns[col]
+				if !ignore {
```
Jagularr, line 97: can be combined

Jagularr (changes requested):
Took another look with more comments, overall this looks to be heavily copied from the existing postgresql input plugin and thus some coding issues came along with it.

plugins/inputs/odyssey/odyssey.go
srebhan: I'd prefer to keep it here if only used in one location (which is the case here I think). The advantage is to have it close to the location where it is used. Otherwise this results in a lot of scrolling forth and back during review.

Jagularr: Yeah that's fair.

srebhan:
@Xrias can you please take a look at the comments!

srebhan:
@Xrias any news here?

sjwang90:
@Xrias We're going to open up this to community to see if we can get anyone to pick up this PR. If you are still interested in working on this please comment.

sjwang90:

PR #8678, opened 2021-01-12
Fix redis output type inconsistencies

### LINKED ISSUE #7375: redis input - type confusion float vs string
```toml
[[inputs.redis]]
servers = ["tcp://1.1.2.3:5"]
```

### System info:

Influxdb: 1.7.10-1
telegraf: Telegraf 1.14.0 (git: HEAD fefd7ff1)
redis: redis_version="5.0.8"
os: fedora 31 x86_64

### Steps to reproduce:

1. enable redis
2. utilize redis a bit
3. restart both telegraf and redis (order doesn't matter)

### Expected behavior:

Telegraf reports proper redis metrics

### Actual behavior:

telegraf log chunk:
```
field type conflict: input field "mem_fragmentation_ratio" on measurement "redis" is type float, already exists as type string dropped=1; discarding poin..
```

### Additional information:

Ref #5957

helenosheaa:
Closes #7375

Currently there can be unreliable typing for outgoing data types. 

- Defined a struct for redis field types. Used reflection to coerce existing fields into defined types
- Also updated redis test with newer redis version output string to account for typing of fields


plugins/inputs/redis/redis.go
```diff
@@ -352,6 +463,126 @@ func gatherInfoOutput(
...
+	fields["used_memory"] = o.UsedMemory
+	fields["used_memory_dataset"] = o.UsedMemoryDataset
+	fields["used_memory_dataset_perc"] = o.UsedMemoryDatasetPerc
+	fields["used_memory_lua"] = o.UsedMemoryLua
+	fields["used_memory_overhead"] = o.UsedMemoryOverhead
+	fields["used_memory_peak"] = o.UsedMemoryPeak
+	fields["used_memory_peak_perc"] = o.UsedMemoryPeakPerc
+	fields["used_memory_rss"] = o.UsedMemoryRss
+	fields["used_memory_scripts"] = o.UsedMemoryScripts
+	fields["used_memory_startup"] = o.UsedMemoryStartup
```
ssoroka, line 584: so one downside of this is that you're going to get all these fields reported in InfluxDB even if Redis didn't report them. One way you could solve this is to only add the field if it also exists in the `fields` map.

ssoroka (changes requested):

ssoroka (approved):

Issue #7892, opened 2020-07-24
[Ping Input Plugin] Strange behavior with method = "native"


zszia:
### Actual behavior:

Config file
```
[[inputs.ping]]
  urls = [
    "10.153.4.7",
    "10.153.0.5"
  ]
  method = "native"
  #method = "exec"
  count = 2
  timeout = 1.5
```

systemd Config
```
[Service]
LimitNOFILE=8192
CapabilityBoundingSet=CAP_NET_RAW
AmbientCapabilities=CAP_NET_RAW
```

Additionally I did:
```
sudo setcap cap_net_raw=eip /opt/telegraf/usr/bin/telegraf
```

Ping is working:
```
ping -c 2 10.153.4.7
PING 10.153.4.7 (10.153.4.7) 56(84) bytes of data.
64 bytes from 10.153.4.7: icmp_seq=1 ttl=246 time=12.4 ms
64 bytes from 10.153.4.7: icmp_seq=2 ttl=246 time=11.8 ms

--- 10.153.4.7 ping statistics ---
2 packets transmitted, 2 received, 0% packet loss, time 1000ms
rtt min/avg/max/mdev = 11.869/12.155/12.442/0.306 ms
```

Exec method is working too:
method = "exec"
```
2020-07-24T11:05:23Z I! Starting Telegraf 1.15.1
> ping,host=xxxxx,url=10.153.4.7 average_response_ms=12.215,maximum_response_ms=12.683,minimum_response_ms=11.747,packets_received=2i,packets_transmitted=2i,percent_packet_loss=0,result_code=0i,standard_deviation_ms=0.468,ttl=246i 1595588724000000000
> ping,host=xxxxx,url=10.153.0.5 average_response_ms=177.759,maximum_response_ms=177.961,minimum_response_ms=177.558,packets_received=2i,packets_transmitted=2i,percent_packet_loss=0,result_code=0i,standard_deviation_ms=0.467,ttl=244i 1595588724000000000
```

native method is working only for one of the hosts. For 10.153.4.7 I get every time packets_received=0, percent_packet_loss=100 and result_code=1
method = "native"
```
2020-07-24T11:04:24Z I! Starting Telegraf 1.15.1
[...]

Issue #7386, opened 2020-04-22
[inputs.ping] Error when trying to lookup hosts


oijkn:
### Relevant telegraf.conf:

```toml
[global_tags]


# Configuration for telegraf agent
[agent]
    interval = "60s"
    debug = false
    hostname = "CentOS8"
    round_interval = true
    flush_interval = "10s"
    flush_jitter = "0s"
    collection_jitter = "0s"
    metric_batch_size = 1000
    metric_buffer_limit = 10000
    quiet = false
    logfile = "/var/log/telegraf/telegraf.log"
    omit_hostname = false

###############################################################################
#                                  OUTPUTS                                    #
###############################################################################

[[outputs.influxdb]]
    urls = ["http://192.168.1.19:8086"]
    skip_database_creation = true
    database = "Ping"
    timeout = "10s"
    retention_policy = ""

###############################################################################
##                                  INPUTS                                     #
################################################################################

# Ping given url(s) and return statistics
[[inputs.ping]]
  interval = '300s'
  urls = ['host01']
  count = 5
  ping_interval = 1.0
  timeout = 3.0
  deadline = 3
  method = 'native'
  interface = 'eth0'
  [inputs.ping.tags]
     plaque = '01'
     equipement = 'bng'

[[inputs.ping]]
  interval = '300s'
  urls = ['host02']
  count = 5
  ping_interval = 1.0
  timeout = 3.0
  deadline = 3
  method = 'native'
  interface = 'eth0'
  [inputs.ping.tags]
     plaque = '02'
     equipement = 'bng'

[[inputs.ping]]
  interval = '300s'
  urls = ['host03']
  count = 5
  ping_interval = 1.0
  timeout = 3.0
  deadline = 3
  method = 'native'
  interface = 'eth0'
  [inputs.ping.tags]
     plaque = '03'
     equipement = 'bng'

[[inputs.ping]]
  interval = '300s'
  urls = ['host04']
  count = 5
  ping_interval = 1.0
  timeout = 3.0
  deadline = 3
  method = 'native'
  interface = 'eth0'
  [inputs.ping.tags]
     plaque = '04'
     equipement = 'bng'

[[inputs.ping]]
  interval = '300s'
  urls = ['host05']
  count = 5
  ping_interval = 1.0
  timeout = 3.0
  deadline = 3
  method = 'native'
  interface = 'eth0'
  [inputs.ping.tags]
     plaque = '05'
     equipement = 'bng'
```

### System info:

```
[...]

PR #8679, opened 2021-01-12
Use go-ping for "native" execution in Ping plugin

sspaink:
### Required for all PRs:

Updated the `input plugin Ping` to use the package [go-ping](https://github.com/go-ping/ping) when using "native" in the configuration. I believe this improves code readability without breaking any backwards compatible functionality. I was having 100% packet loss with the original implementation that I couldn't resolve even after providing the correct permissions as specified in the README, using the new implementation solved this problem for me. Moving all the implementation details to go-ping should help with future maintenance of this plugin.

This might help address the following issues:

- https://community.influxdata.com/t/input-ping-not-working-with-native-method/17582
- https://github.com/influxdata/telegraf/issues/7892
- https://github.com/influxdata/telegraf/issues/7386


Jagularr (approved):

plugins/inputs/ping/README.md
```diff
@@ -19,7 +19,7 @@ apt-get install iputils-ping
 
 When using `method = "native"` a ping is sent and the results are reported in
 native Go by the Telegraf process, eliminating the need to execute the system
-`ping` command.
+`ping` command. This is dependent on the [go-ping](https://github.com/go-ping/ping) package.
```
ssoroka, line 22: probably not all that relevant to users.

plugins/inputs/ping/ping.go
```diff
@@ -157,204 +141,103 @@ func (p *Ping) Gather(acc telegraf.Accumulator) error {
...
+	go func() {
+		if p.Deadline <= 0 {
+			return
+		}
+		// If deadline is set ping exits regardless of how many packets have been sent or received
+		timer := time.AfterFunc(time.Duration(p.Deadline)*time.Second, func() {
+			pinger.Stop()
+		})
+		defer timer.Stop()
+	}()
```
ssoroka, line 180: AfterFunc returns immediately, this is going to stop the timer immediately after it starts.
Looks like you don't need a goroutine for this.
```suggestion
	if p.Deadline > 0 {
		// If deadline is set ping exits regardless of how many packets have been sent or received
		timer := time.AfterFunc(time.Duration(p.Deadline)*time.Second, func() {
			pinger.Stop()
		})
	}
```

plugins/inputs/ping/README.md
sspaink: Good point, I was thinking might help them debug but they can just look at the source code if they are going to do that.

plugins/inputs/ping/ping.go
sspaink: That looks a lot better, thanks!

sspaink:
!signed-cla

plugins/inputs/ping/ping.go
```diff
@@ -157,204 +141,104 @@ func (p *Ping) Gather(acc telegraf.Accumulator) error {
...
+	if p.PingInterval < 0.2 {
+		pinger.Interval = time.Duration(.2 * float64(time.Second))
 	} else {
-		ips = filterIPs(ips, isV4)
-	}
-
-	if len(ips) == 0 {
-		return nil, errors.New("Cannot resolve ip address")
+		pinger.Interval = time.Duration(p.PingInterval * float64(time.Second))
 	}
```
srebhan, line 166: Can you please do this in the `Init()` function.

plugins/inputs/ping/ping.go
```diff
@@ -157,204 +141,104 @@ func (p *Ping) Gather(acc telegraf.Accumulator) error {
...
-
-	timeout := p.Timeout
-	if timeout == 0 {
-		timeout = 5
+	// If no timeout is given default to 5 seconds, matching original implementation
+	if p.Timeout == 0 {
+		pinger.Timeout = time.Duration(5) * time.Second
+	} else {
+		pinger.Timeout = time.Duration(p.Timeout) * time.Second
 	}
```
srebhan, line 173: Can you please do it in the `Init()` function.

plugins/inputs/ping/ping_test.go
```diff
@@ -436,10 +457,9 @@ func TestDNSLookupError(t *testing.T) {
 
 	var acc testutil.Accumulator
 	p := Ping{
-		Urls:        []string{"localhost"},
-		Method:      "native",
-		IPv6:        false,
-		resolveHost: mockHostResolverError,
+		Urls:   []string{"localhost"},
+		Method: "native",
+		IPv6:   false,
```
srebhan, line 462: I believe this now does the opposite of what was originally intended! Now the resolve will very likely not fail! Am I missing something?

srebhan (changes requested):
I have two minor comments (see code) and one regarding the "name-resolve test". It seems like before the test was checking for name-resolution errors but now it seems you are testing exactly the opposite... Could you please comment on that!

plugins/inputs/ping/ping.go
sspaink: So to get pinger with `pinger, err := ping.NewPinger(destination)` it resolves the destination address. I assumed we want to re-do this at each gather in case something changes? What do you think?

I would have to move ping.NewPinger to Init() as well to move setting the Interval and Timeout to Init().

plugins/inputs/ping/ping_test.go
sspaink: Good catch! Without the `mockHostResolverError` it doesn't make sure an error is thrown. I made a new commit changing "localhost" to "fakehost" that should ensure that the test fails. There shouldn't be a case were someone has a "fakehost"?

plugins/inputs/ping/ping.go
srebhan: No I was think about clipping `p.PingInterval` in `Init()` and get rid of the `if p.PingInterval < 0.2 {` here.

plugins/inputs/ping/ping_test.go
srebhan: Yeah but I remember some discussions in the past where someone from InfluxData stated that it's not an option to rely on this. Could you please check back with @ssoroka or @reimda!?

ssoroka: yeah I'd rather not depend on a real network address failing, as it fails differently in a lot of various scenarios.

sspaink: I removed the test with the assumption that it is now the 3d party libraries responsibility to detect DNS failures.

srebhan (commented):
You now removed the resolution test completely!? @ssoroka has to decide if this is ok or not. :-)

ssoroka (changes requested):
needs some changes. ping me

ssoroka (approved):

sspaink:
Thanks @srebhan and @ssoroka  for the help!

redrumdk:
I just tried this nightly build, now when using native and i ping a IP which permanently down . 
i get various results, between 0 = no packetsloss, 100 = 100% packetloss, -300 = ??? i have no idea. 

i get no errors in the debug logs for telegraf, and i have uploaded a screenshot of the influxdb ping content. 

https://ibb.co/kQk8RCm

Running on windows 2016
w. around 200+ ip's i would like to ping.

srebhan:
@redrumdk please open an issue with the information above, so we can look at the problem. I've no access to Windows 2016, but I will try in my setup and find someone to reproduce the issue otherwise. Would be nice if you could mention me in the issue so myself or @sspaink can take a look.

Hipska:
He did in #8704

PR #8720, opened 2021-01-19
fix: remove ambiguity on '\v' from line-protocol parser

adrian-thurston:
The line protocol parser allows vertical tab ('\v') in whitespace, but also
allows it in measurements, tags keys, tag values, and field keys. This
ambiguity causes a blowup in the parsing state machine and triggers undefined
behaviour when the vertical tab character is seen. The parser will attempt to
simultaneously extend a compoent of the line, and move on to the next one.

This patch removes vertical tab from measurements, tags, and fields. The
resulting state machine goes from 745 states to 90, with a similar reduction in
the number of lines of generated code.


ssoroka (approved):
❤️ 🎉

srebhan:
While this change makes definitively sense I wonder if this is a breaking change. Could you guy please outline if the vertical tab _ever_ worked in _any_ case!? If not we should merge this right away but otherwise ppl might rely on it...

ssoroka:
This change would only affect key names, key values, and field names, not field values.
I suspect it's more likely that if you're using odd characters in these you're writing data that you won't be able to query or do anything with. I'm in support of merging this.

adrian-thurston:
@srebhan In general, when it comes to parser ambiguities in ragel parsers, the parser would be entering into undefined behaviour and bad things would be happening. It would be simultaneously considering both possibilities as it moves forward.

In this particular case though, the ambiguity is between recognizing whitespace, which simply terminates the previous string, and extending the string. So all that happens is some strings get terminated early or have the wrong value in them. But it only happens in places where whitespace is permitted.. So to sum up what happened before this patch, a `\v` in:

1. measurement name: measurement parses wrong
2. key name: \v accepted 
3. key value: parses wrong
4. value name: \v accepted

So to answer your question, `\v` worked only in key name and value name. In the other places it parsed wrong. So we would be breaking for anywone who is currently using `\v` successfully in those places.

PR #8721, opened 2021-01-20
kubernetes_inventory: Set $HOSTIP in default URL

paulchoi:
Telegraf daemonset pod cannot contact kube-apiserver via localhost.
Should be `$HOSTIP`, and the default port is 6443.


paulchoi:
!signed-cla

sspaink (approved):

PR #8724, opened 2021-01-20
Add snmp_trap integration test with docker-compose

reimda:
This is an integration test for the snmp_trap plugin.  It uses netsnmp to send a trap to telegraf.  The two processes are run in containers managed by docker-compose.

Snmp_trap is a plugin that listens on UDP.  Unlike TCP, it's not possible for a remote host to test whether a program is listening on a UDP port.  That means for this test, synchronization of the containers is critical to prevent a race condition between the containers.  This PR uses docker-compose v2's ability to enforce container health dependency to make sure telegraf is listening before letting netsnmp send the trap.

Currently only version 2 trap is tested.  I would like eventually to expand to version 3 traps including encryption, but I want to get this running regularly by an automated process before expanding coverage.

### DIFF
--- a/plugins/inputs/snmp_trap/integration_test/docker-compose.test.yml
+++ b/plugins/inputs/snmp_trap/integration_test/docker-compose.test.yml
@@ -1,0 +1,40 @@
+#version 3 doesn't have depends_on condition service_healthy
+version: '2.4'
+
+networks:
+  integration:
+    driver: bridge
+    ipam:
+      config:
+        - subnet: 10.0.0.0/8 # test expects netsnmp containter to be 10.0.0.3
+
+services:
+  telegraf_test:
+    build:
+      context: .
+      dockerfile: ./Dockerfile.test
+    networks:
+      - integration
+    volumes:
+      - ../../../..:/go/src/github.com/influxdata/telegraf
+    healthcheck:
+#I haven't been able to get any shell command to work when provided as a string
+#      test: "ss -lun|grep :12399"
+#A list starting with CMD-SHELL is supposed to be equivalent but does work
+      test: ["CMD-SHELL", "ss -lun|grep :12399"]
+      interval: 1s
+      retries: 40
+      #enough to download and build module dependencies,
+      #build and run the test, then open the port
+  netsnmp:
+    build:
+      context: .
+      dockerfile: ./Dockerfile.netsnmp
+    restart: on-failure
+    networks:
+      - integration
+    depends_on:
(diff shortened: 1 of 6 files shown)


reimda:
!signed-cla

plugins/inputs/snmp_trap/integration_test/docker-compose.test.yml
```diff
@@ -0,0 +1,40 @@
...
+  netsnmp:
+    build:
+      context: .
+      dockerfile: ./Dockerfile.netsnmp
+    restart: on-failure
+    networks:
+      - integration
+    depends_on:
+      telegraf_test:
+        condition: service_healthy
```
helenosheaa, line 38: You may have already tried this but I saw a [stack overflow](https://stackoverflow.com/questions/47710767/what-is-the-alternative-to-condition-form-of-depends-on-in-docker-compose-versio#:~:text=Version%203%20no%20longer%20supports,that%20behaviour%20using%20V3%20upwards.) which says you can replicate depends on conditional behaviour in v3 using dockerize -wait

helenosheaa (approved):
Looks good, just posted in something to look at if you hadn't already about v3. Also would be helpful for me to have clarification around this having it's own docker/docker-compose files whereas in other integration tests we are using the one docker-compose?

plugins/inputs/snmp_trap/integration_test/docker-compose.test.yml
reimda: @helenosheaa dockerize -wait works well for tcp services like http.  It waits for the service to accept connections.  Snmp traps use udp which is connectionless so there's no way to know if telegraf is accepting connections yet.

Jagularr (approved):

powersj:
@reimda next steps on this one? or happy to close?

reimda:
This PR hasn't been a priority because the existing snmp_trap unit tests already provide coverage. They send test traps using gosnmp which is what telegraf also uses to receive traps.

This test's purpose was to use a different snmp implementation and one that's more widely used. I was trying to catch failures caused if traps that gosnmp sends aren't the same as net-snmp's.

The risk of such a bug in gosnmp's traps seems low now that this plugin has been in use for a while.

If we were to move this forward, we would want to remove the dependency on docker-compose and add this to automated tests. To add it to CI as a unit test we could replace docker-compose with testcontainers-go to run the net-snmp image.

I'm ok closing because it doesn't seem worth the trouble.

Issue #8643, opened 2021-01-04
Azure IoT/Event hub Telegraf Plugin not sending data to influxdb cloud 2.0


goodvirus136:
As I have stated in the [influxdb forum](https://community.influxdata.com/t/telegraf-not-sending-data-to-influxdb-azure-event-hub-influxdb-cloud-2/17240/13) I have a problem getting data from and azure iot hub (or event hub, I tried both) to the influxdb cloud 2.0
The setup is as following (As described in the [documentation](https://github.com/influxdata/telegraf/blob/release-1.14/plugins/inputs/eventhub_consumer/README.md)):

I can see outgoing traffic on the azure side, but nothing shows up in my bucket. Also I tried a different output (stdout), but nothing there either. The forum advised me to make a bug report.

### Relevant telegraf.conf:
```
# Configuration for telegraf agent
[agent]
  ## Default data collection interval for all inputs
  interval = "10s"
  ## Rounds collection interval to 'interval'
  ## ie, if interval="10s" then always collect on :00, :10, :20, etc.
  round_interval = true

  ## Telegraf will send metrics to outputs in batches of at most
  ## metric_batch_size metrics.
  ## This controls the size of writes that Telegraf sends to output plugins.
  metric_batch_size = 1000

  ## For failed writes, telegraf will cache metric_buffer_limit metrics for each
  ## output, and will flush this buffer on a successful write. Oldest metrics
  ## are dropped first when this buffer fills.
  ## This buffer only fills when writes fail to output plugin(s).
  metric_buffer_limit = 10000

  ## Collection jitter is used to jitter the collection by a random amount.
  ## Each plugin will sleep for a random time within jitter before collecting.
  ## This can be used to avoid many plugins querying things like sysfs at the
  ## same time, which can have a measurable effect on the system.
  collection_jitter = "0s"

  ## Default flushing interval for all outputs. Maximum flush_interval will be
  ## flush_interval + flush_jitter
  flush_interval = "10s"
  ## Jitter the flush interval by a random amount. This is primarily to avoid
  ## large write spikes for users running a large number of telegraf instances.
  ## ie, a jitter of 5s and interval 10s means flushes will happen every 10-15s
  flush_jitter = "0s"

  ## By default or when set to "0s", precision will be set to the same
  ## timestamp order as the collection interval, with the maximum being 1s.
  ##   ie, when interval = "10s", precision will be "1s"
  ##       when interval = "250ms", precision will be "1ms"
[...]

PR #8731, opened 2021-01-21
Improve eventhub_consumer input documentation

marinsky:
This PR improves documentation (README and sample config) for eventhub_consumer input plugin:

* clarifies the usage of env vars and adds example.
* addresses issues like: #8643

### DIFF
--- a/plugins/inputs/eventhub_consumer/README.md
+++ b/plugins/inputs/eventhub_consumer/README.md
@@ -18,8 +18,6 @@ The main focus for development of this plugin is Azure IoT hub:
   ## This requires one of the following sets of environment variables to be set:
   ##
   ## 1) Expected Environment Variables:
-  ##    - "EVENTHUB_NAMESPACE"
-  ##    - "EVENTHUB_NAME"
   ##    - "EVENTHUB_CONNECTION_STRING"
   ##
   ## 2) Expected Environment Variables:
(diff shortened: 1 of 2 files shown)


marinsky:
!signed-cla

plugins/inputs/eventhub_consumer/README.md
```diff
@@ -88,11 +95,70 @@ The main focus for development of this plugin is Azure IoT hub:
   ## Each data format has its own unique set of configuration options, read
   ## more about them here:
   ## https://github.com/influxdata/telegraf/blob/master/docs/DATA_FORMATS_INPUT.md
-  data_format = "influx"
```
ssoroka, line 91: default should probably be left at influx.

plugins/inputs/eventhub_consumer/README.md
```diff
@@ -8,7 +8,7 @@ The main focus for development of this plugin is Azure IoT hub:
 
 1. Create an Azure IoT Hub by following any of the guides provided here: https://docs.microsoft.com/en-us/azure/iot-hub/
 2. Create a device, for example a [simulated Raspberry Pi](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-raspberry-pi-web-simulator-get-started)
-3. The connection string needed for the plugin is located under *Shared access policies*, both the *iothubowner* and *service* policies should work
```
ssoroka, line 11: this change is missing from sampleConfig, they should match.

ssoroka (changes requested):
JSON example probably doesn't belong here and is better off with the json configuration documentation. We could link there instead.

plugins/inputs/eventhub_consumer/README.md
marinsky: Thank you for a quick and thorough review!

No problem, I can change it back, but are you absolutely sure about it?  influx format is not what you see every day in kafka topics, json from the other side is one of the most popular options. The most popular should probably be default.

That line was the main reason of spending 2h investigating what's going on when I started using the plugin. The same line is the main reason #8643 was opened, because if the format is not correct, the message silently disappears.

Please, consider and let me know.

marinsky: Nvm, I moved back to influx, json docs have enough examples and there is a clear link to other data formats above this line.
To address issues like #8643 I will open another PR to improve logging when received data doesn't correspond to the specified format.

marinsky:
@ssoroka Can we please find time to re-review this. Should be quick. Just a documentation improvement. I have another eventhub improvement coming. Thanks.

ssoroka (approved):

ssoroka:
Thanks for the contribution :)

PR #8734, opened 2021-01-22
Add line to template docs about measurement

helenosheaa:
If you supply a template without measurement included it returns an error asking you to specify measurement. This was not currently mentioned in the templating docs.


srebhan (approved):
Makes sense to me. :-)

PR #8736, opened 2021-01-22
Update README to list only supported versions of SQL server

masree:
Unsupported versions of Sql Server: 2011 and older versions of SQL server 
Azure SQL Database: VIEW DATABASE STATE permission does not apply for certain DMV like sys.dm_os_performance_counters where databases fall under tier Basic, S0 and S1 service level objectives. It is supported for higher version of service level objective(S2 and above) - https://docs.microsoft.com/en-us/sql/relational-databases/system-dynamic-management-views/sys-dm-os-performance-counters-transact-sql?view=sql-server-ver15

### Required for all PRs:

### DIFF
--- a/plugins/inputs/sqlserver/README.md
+++ b/plugins/inputs/sqlserver/README.md
@@ -5,8 +5,7 @@ lightweight and use Dynamic Management Views supplied by SQL Server.
 
 ### The SQL Server plugin supports the following editions/versions of SQL Server
 - SQL Server
-  - 2008 SP3 (with CU3)
-  - SQL Server 2008 R2 SP3 and newer versions
+  - SQL Server 2012 and newer versions
 - Azure SQL Database (Single)
 - Azure SQL Managed Instance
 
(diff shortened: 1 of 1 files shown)


denzilribeiro (changes requested):
For SQL Server I believe this still supports 2008R2 Sp3 so  can't say 2012 and newer as of now. At some point would have to deprecate it. @Trovalo  thoughts? AT what point would we go to "supported versions" of SQL?

Trovalo:
Recently I discovered that even older versions are supported, like 2008 "vanilla", you just have to play around with the connection string and add a magic ```{...};encrypt=disable``` and the connection will be possible.

About 2008 vanilla, I should have some edited queries to make everything work even without some sys objects (as of now something should break), with a better way to check if a column exists or not for the given version, I just need to find the time to contribute again (which won't be easy as of now).

Going back to the issue itself, I think we can say that the plugin provides support for `SQL Server 2008 and newer versions`, with a note about the connection parameter.
About when to switch to supported editions only, tbh I don't know. As of now, I'm not bothered by supporting SQL 2008. (as most of the used sys objects and dmv are available)

helenosheaa:
This [pr](https://github.com/influxdata/telegraf/pull/9040/files) was merged recently updating the readme causing a conflict here, not sure if you want to alter the wording / still make the change to the Azure SQL Database part.

masree:
> This [pr](https://github.com/influxdata/telegraf/pull/9040/files) was merged recently updating the readme causing a conflict here, not sure if you want to alter the wording / still make the change to the Azure SQL Database part.

Thanks @helenosheaa! I'll link the above PR and mark it closed

masree:
[PR](https://github.com/influxdata/telegraf/pull/9040/files) have already addressed the changes that I had done in the PR.

Issue #7393, opened 2020-04-23
Telegraf log gets spammed/flooded if Elasticsearch receiver is in read-only state


cmakaloclifford:
### Relevant telegraf.conf:
```toml
# Configuration for telegraf agent
[agent]
  interval = "10s"
  round_interval = true
  metric_batch_size = 1000
  metric_buffer_limit = 10000
  collection_jitter = "0s"
  flush_interval = "10s"
  flush_jitter = "0s"
  precision = ""
  logtarget = "file"
  logfile = "/dev/null" # (Temporary fix for infinity loop of logs when Elasticsearch is in read only mode.)
  #logfile_rotation_interval = "1d"
  #logfile_rotation_max_archives = 7
  hostname = ""
  omit_hostname = false

[[outputs.elasticsearch]]
  urls = [ "https://***:9200" ] # required.
  timeout = "5s"
  enable_sniffer = false
  health_check_interval = "10s"
  username = "***"
  password = "***"
  index_name = "***-%Y.%m.%d" # required.
  tls_ca = "/etc/telegraf/***-x509.pem"
  tls_cert = "/etc/telegraf/***-x509.pem"
  tls_key = "/etc/telegraf/***.key.pem"
  manage_template = true
  template_name = "edge"
  overwrite_template = false

# Telegraf logs
[[inputs.tail]]
  files = ["/var/log/telegraf/telegraf.log"]
  from_beginning = false
  pipe = false
  data_format = "value"
  data_type = "string"
  name_override = "log"

  [inputs.tail.tags]
     metric_type = "logs"
     log_source = "telegraf"
```

### System info:

Telegraf version: 1.14.1 (installed via Apt/Deb package)
OS version: Ubuntu 18.04

### Docker

No Docker instance is being used

### Steps to reproduce:

1. Wait for Elasticsearch server to reach 95% used storage and automatically go into read-only mode.
2. Tail telegraf.log and see that every 10 second, the buffer (1000 lines) gets written with an error that Elasticsearch is in read-only mode.
3. I we also have the Tail input activated on Telegrafs log file, then this happens about 4 times every second, resulting in about 4000 lines get written every second.

Please note that right now we have set the logging to /dev/null just to save us from getting our disks full each and every day. ;)

### Expected behavior:

It seams reasonable that either these lines where deduplicated in the log or it did not try all writes at all if the first write against the server fails.
[...]

PR #8748, opened 2021-01-25
Fix issue with elasticsearch output being really noisy about some errors

ssoroka:
resolves https://github.com/influxdata/telegraf/issues/7393


helenosheaa (approved):

srebhan:
Hmm now only the first error is shown. If this is what you want, please at least add a comment to the break to prevent the next one to remove it because he/she cannot guess the intention behind it.

However I would prefer to group the errors by `err.Error.Reason` and output the unique set of reasons or whatever describes the error best. This way you would only get one error per _error type_ but you do see all errors! Additionally we could collect the IDs per error type so we can output the same information as before in less log-lines...

What do you think? I could even do it if you want me to...

Jagularr (approved):

ssoroka:
@srebhan going to merge this, but open to improving it in another PR.

PR #8749, opened 2021-01-26
Add HTTP proxy setting to New Relic output plugin

msummers-nr:
### Required for all PRs:

Update New Relic output plugin:
- Add http_proxy as a configuration value override


plugins/outputs/newrelic/newrelic.go
```diff
@@ -4,12 +4,13 @@ package newrelic
 import (
 	"context"
 	"fmt"
+	"github.com/influxdata/telegraf/plugins/outputs"
```
srebhan, line 7: Moving this line here is not correct. We like to group the influx imports below.

plugins/outputs/newrelic/newrelic.go
```diff
@@ -51,14 +57,16 @@ func (nr *NewRelic) Connect() error {
 	if nr.InsightsKey == "" {
 		return fmt.Errorf("InsightKey is a required for newrelic")
 	}
+	//TODO init nr.client here
```
srebhan, line 60: Well, please do it in this PR then. What should this line tell me? :-)

plugins/outputs/newrelic/newrelic.go
```diff
@@ -115,9 +123,7 @@ func (nr *NewRelic) Write(metrics []telegraf.Metric) error {
 					mvalue = float64(1)
 				}
 			case string:
-				// Do not log everytime we encounter string
-				// we just skip
-				continue
+				tags[mname] = n
```
srebhan, line 126: This change is strange. The pattern for string now does not match the one of the other types. Are you sure this is correct or did you want to do `mname = n` here?

plugins/outputs/newrelic/newrelic.go
```diff
@@ -152,7 +158,29 @@ func init() {
 	outputs.Add("newrelic", func() telegraf.Output {
 		return &NewRelic{
 			Timeout: internal.Duration{Duration: time.Second * 15},
-			Client:  http.Client{},
+			// TODO Don't set the client here, wait
+			// client: http.Client{},
```
srebhan, line 162: Please get rid of the TODOs! I don't want to get half-finished stuff in and let the TODOs pile-up in the repo. It is also not clear what needs to be done for this TODO to vanish. Waiting probably won't solve the problem... ;-)

plugins/outputs/newrelic/newrelic.go
```diff
@@ -51,14 +57,16 @@ func (nr *NewRelic) Connect() error {
 	if nr.InsightsKey == "" {
 		return fmt.Errorf("InsightKey is a required for newrelic")
 	}
+	//TODO init nr.client here
+	nr.getHTTPClient()
```
srebhan, line 61: The function returns `error` and it really might do so. Ignoring those errors is extremely bad! Either move the code of the function up here _or_ check the return code!

plugins/outputs/newrelic/newrelic.go
```diff
@@ -152,7 +158,29 @@ func init() {
...
 		return &NewRelic{
 			Timeout: internal.Duration{Duration: time.Second * 15},
-			Client:  http.Client{},
+			// TODO Don't set the client here, wait
+			// client: http.Client{},
 		}
 	})
 }
+
+func (nr *NewRelic) getHTTPClient() error {
```
srebhan, line 167: Maybe call this function `setupClient()` or `initClient()` as the current name suggests you return the client which is not the case.

srebhan (changes requested):
Thanks for this PR! To get this merged you have to

1. Get rid of all TODOs in the code!
2. Check the errors returned by `getHTTPClient()` or change the code structure! See my comments in the code.

There are also some minor comments in the code. Furthermore it would be nice to change the PR title to something reflecting your change e.g. "Add HTTP proxy setting to New Relic output plugin.".

plugins/outputs/newrelic/newrelic.go
msummers-nr: Done

plugins/outputs/newrelic/newrelic.go
msummers-nr: Removed

plugins/outputs/newrelic/newrelic.go
msummers-nr: This is correct, comment added to code to explain.

plugins/outputs/newrelic/newrelic.go
msummers-nr: Done

plugins/outputs/newrelic/newrelic.go
msummers-nr: Thanks for catching this, my mistake.

Corrected.

plugins/outputs/newrelic/newrelic.go
msummers-nr: Changed

plugins/outputs/newrelic/newrelic.go
```diff
@@ -108,18 +118,21 @@ func (nr *NewRelic) Write(metrics []telegraf.Metric) error {
...
 			case string:
-				// Do not log everytime we encounter string
-				// we just skip
-				continue
+				// If we see a string send it as a tag so it can be parsed as an attribute, max 255 chars
+				tagLength := len(n)
+				if tagLength > 255 {
+					tagLength = 255
+				}
+				tags[mname] = n[:tagLength]
```
srebhan, line 133: This is a change in behavior compared to what is done currently. As we try to keep the output stable over releases you need to guard this with an option that keeps the old behavior by default. I.e. add an option `submit_string_values_as_tag` or similar and keep it set to false by default. Then wrap your code here with `if nr.StringsAsTags`.
If this is controversal, you maybe should split out this change to another PR so we do not hold-up the proxy part.

srebhan (changes requested):
Nice. The only thing is the change in behavior for string values. To keep the output of the plugin stable, please add an option to allow to keep the current behavior by default and opt-in for the new behavior (string values as tags). Comment in the code.

plugins/outputs/newrelic/newrelic.go
msummers-nr: Reverted to original

srebhan (approved):
Looks good to me.

msummers-nr:
> Looks good to me.

@srebhan Thanks for your help

Jagularr (approved):

ssoroka (approved):

PR #8751, opened 2021-01-26
Adding a new directory monitor input plugin. 

### LINKED ISSUE #7548: Add plugin that implements a simple file based queue system

Jagularr:
This will monitor and parse all files coming into a single configured directory.

Closes #7548


Jagularr:
!signed-cla

sjwang90:
Closes #7548

Jagularr:
!signed-cla

Jagularr:
!signed-cla

Jagularr:
!signed-cla

reimda (changes requested):
It's coming along!

plugins/inputs/directory_monitor/directory_monitor.go
```diff
@@ -0,0 +1,404 @@
...
+		reader, err = gzip.NewReader(file)
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		reader, _ = utfbom.Skip(monitor.decoder.Reader(file))
+	}
+
+	// Read the file and parse with the configured parse method.
+	fileContents, err := ioutil.ReadAll(reader)
```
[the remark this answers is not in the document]
Jagularr: Changed, and made exception for JSON files, for now, as discussed.

plugins/inputs/directory_monitor/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+This plugin monitors a single directory (without looking at sub-directories), and takes in each file placed in the directory.
+The plugin will gather all files in the directory at a configurable interval (`monitor_interval`), and parse the ones that haven't been picked up yet.
+
+Please be advised that this plugin pulls files directly after they've been in the directory for the length of the configurable `directory_duration_threshold`, and thus files should not be written 'live' to the monitored directory unless they are guaranteed to finish writing before the `directory_duration_threshold`. This plugin is intended to read files that are moved or copied to the monitored directory, and thus files should also not be used by another process or else they may fail to be gathered.
+
+### Configuration:
+
+```toml
+## The directory to monitor and read files from.
```
ssoroka, line 11: should include plugin name here with indentation

plugins/inputs/directory_monitor/README.md
```diff
@@ -0,0 +1,59 @@
...
+directory = ""
+#
+## The directory to move finished files to.
+finished_directory = "5s"
```
ssoroka, line 15: looks like this is set to a duration?

plugins/inputs/directory_monitor/README.md
```diff
@@ -0,0 +1,59 @@
+# Directory Monitor Input Plugin
+
+This plugin monitors a single directory (without looking at sub-directories), and takes in each file placed in the directory.
+The plugin will gather all files in the directory at a configurable interval (`monitor_interval`), and parse the ones that haven't been picked up yet.
+
+Please be advised that this plugin pulls files directly after they've been in the directory for the length of the configurable `directory_duration_threshold`, and thus files should not be written 'live' to the monitored directory unless they are guaranteed to finish writing before the `directory_duration_threshold`. This plugin is intended to read files that are moved or copied to the monitored directory, and thus files should also not be used by another process or else they may fail to be gathered.
```
ssoroka, line 6: Seems like this approach is asking for trouble. It's friendlier to the users to allow them to write directly into the directory, but if they're wrong about how long the write takes they'll lose data. At the least I think we should encourage them to write the files elsewhere then move them into the folder when complete.

plugins/inputs/directory_monitor/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+```toml
+## The directory to monitor and read files from.
+directory = ""
+#
+## The directory to move finished files to.
+finished_directory = "5s"
+#
+## Whether or not to move files that error out to an error directory.
+use_error_directory = "true"
```
ssoroka, line 18: use the path to a directory instead, and if it's blank don't use an error directory.

plugins/inputs/directory_monitor/README.md
```diff
@@ -0,0 +1,59 @@
...
+#
+## The directory to move files to upon file error, given that 'use_error_directory' is enabled.
+## If not is given, the error directory will be auto-generated.
+# error_directory = ""
+#
+## The interval at which to check the directory for new files.
+# monitor_interval = "50ms"
```
ssoroka, line 25: We should use the built-in `interval` and Gather() function for these rather than recreating something new.

plugins/inputs/directory_monitor/README.md
```diff
@@ -0,0 +1,59 @@
...
+#
+## The amount of time a file is allowed to sit in the directory before it is picked up.
+## This time can generally be low but if you choose to have a very large file written to the directory and it's potentially slow,
+## set this higher so that the plugin will wait until the file is fully copied to the directory.
+# directory_duration_threshold = "50ms"
```
ssoroka, line 30: uneasy about this. would prefer it not to exist :D

plugins/inputs/directory_monitor/README.md
```diff
@@ -0,0 +1,59 @@
...
+#
+## Character encoding to use when interpreting the file contents. Invalid
+## characters are replaced using the unicode replacement character. Defaults to utf-8.
+##   ex: character_encoding = "utf-8"
+##       character_encoding = "utf-16le"
+##       character_encoding = "utf-16be"
+# character_encoding = "utf-8"
```
ssoroka, line 37: Line protocol can't process anything other than UTF-8, so if we want to support others we will need to convert them to UTF-8 anyway. Not sure we want to make that the responsibility of this plugin.

plugins/inputs/directory_monitor/README.md
```diff
@@ -0,0 +1,59 @@
...
+## A list of files to ignore, if necessary. Supports regex.
+# files_to_ignore = [".DS_Store"]
+#
+## Maximum lines of the file to process that have not yet be written by the
+## output. For best throughput set based on the number of metrics and the size of the output's metric_batch_size.
+## Warning: setting this number too high can lead to a high number of dropped metrics.
+# max_buffered_metrics = 1000
+#
+## The maximum amount of files to process at once. A very high number can lead to bigger memory use and potential file system errors.
+# max_concurrent_files = 3000
```
ssoroka, line 51: we may want to only process a couple files at once anyway. There's likely no performance gain to processing many files in parallel as you're going to hit disk IO limits you can't fix with parallelism. Consider hard-coding this to something small, like 1-5

plugins/inputs/directory_monitor/README.md
```diff
@@ -0,0 +1,59 @@
...
+#
+## The dataformat to be read from the files.
+## Each data format has its own unique set of configuration options, read
+## more about them here:
+## https://github.com/influxdata/telegraf/blob/master/docs/DATA_FORMATS_INPUT.md
+## NOTE: We do not currently support JSON file streaming and thus JSON files will be fully loaded into memory when they are processed.
```
ssoroka, line 57: jsonp is line-delimited JSON and it should be easy enough to support it. Most of the time we support this transparently as "json"

plugins/inputs/directory_monitor/directory_monitor_test.go
```diff
@@ -0,0 +1,187 @@
...
+			Length: 25.97,
+		},
+	}
+
+	for count, data := range fileData {
+		writeJSONFile(data, filepath.Join(processDirectory, "test"+fmt.Sprint(count)+".json"))
+	}
+
+	require.NoError(t, err)
+	time.Sleep(50 * time.Millisecond)
```
ssoroka, line 154: remove the sleeps from the tests, set the wait to 0.

plugins/inputs/directory_monitor/directory_monitor.go
```diff
@@ -0,0 +1,468 @@
...
+	"compress/gzip"
+	"io"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"sync"
+)
+
+const sampleConfig = `
+## The directory to monitor and read files from.
```
ssoroka, line 31: should be indented with two spaces

plugins/inputs/directory_monitor/directory_monitor.go
```diff
@@ -0,0 +1,468 @@
...
+## NOTE: We do not currently support JSON file streaming and thus JSON files will be fully loaded into memory when they are processed.
+data_format = "influx"
+`
+
+var (
+	defaultFilesToMonitor             = []string{}
+	defaultFilesToIgnore              = []string{}
+	defaultMaxBufferedMetrics         = 1000
+	defaultMaxConcurrentFiles         = 3000
+	defaultMonitorInterval            = internal.Duration{Duration: 50 * time.Millisecond}
```
ssoroka, line 86: internal.Duration is deprecated. use config.Duration

plugins/inputs/directory_monitor/directory_monitor.go
```diff
@@ -0,0 +1,468 @@
...
+	defaultDirectoryDurationThreshold = internal.Duration{Duration: 50 * time.Millisecond}
+)
+
+type empty struct{}
+type semaphore chan empty
```
ssoroka, line 91: see also golang.org/x/sync/semaphore

plugins/inputs/directory_monitor/directory_monitor.go
```diff
@@ -0,0 +1,468 @@
...
+
+type DirectoryMonitor struct {
+	Directory         string `toml:"directory"`
+	FinishedDirectory string `toml:"finished_directory"`
+	UseErrorDirectory bool   `toml:"use_error_directory"`
```
ssoroka, line 96: avoid bool fields by having a meaningful zero value (`""`) for ErrorDirectory

plugins/inputs/directory_monitor/directory_monitor.go
```diff
@@ -0,0 +1,468 @@
...
+	ErrorDirectory    string `toml:"error_directory"`
+
+	CharacterEncoding          string            `toml:"character_encoding"`
+	FilesToMonitor             []string          `toml:"files_to_monitor"`
+	FilesToIgnore              []string          `toml:"files_to_ignore"`
+	MaxBufferedMetrics         int               `toml:"max_buffered_metrics"`
+	MonitorInterval            internal.Duration `toml:"monitor_interval"`
+	DirectoryDurationThreshold internal.Duration `toml:"directory_duration_threshold"`
```
ssoroka, line 104: config.Duration

plugins/inputs/directory_monitor/directory_monitor.go
```diff
@@ -0,0 +1,468 @@
...
+	MaxConcurrentFiles         int               `toml:"max_concurrent_files"`
+
+	filesInUse          cmap.ConcurrentMap
```
ssoroka, line 107: Would prefer using sync.Map instead of importing a new library. Any reason we can't do that? We can also potentially get rid of the shared map by using a channel to communicate instead, and getting rid of the parallelism.

plugins/inputs/directory_monitor/directory_monitor.go
```diff
@@ -0,0 +1,468 @@
...
+	Log                 telegraf.Logger
```
ssoroka, line 108:
```suggestion
	Log                 Telegraf.Logger `toml:"-"`
```

plugins/inputs/directory_monitor/directory_monitor.go
```diff
@@ -0,0 +1,468 @@
...
+	parserFunc          parsers.ParserFunc
+	decoder             *encoding.Decoder
+	filesProcessed      selfstat.Stat
+	filesDropped        selfstat.Stat
+	waitGroup           *sync.WaitGroup
+	acc                 telegraf.TrackingAccumulator
+	sem                 semaphore
+	quit                chan bool
```
ssoroka, line 116: use a context.Context instead.

plugins/inputs/directory_monitor/directory_monitor.go
```diff
@@ -0,0 +1,468 @@
...
+func (monitor *DirectoryMonitor) Start(acc telegraf.Accumulator) error {
+	// Use tracking to determine when more metrics can be added without overflowing the outputs.
+	monitor.acc = acc.WithTracking(monitor.MaxBufferedMetrics)
+	go func() {
+		for {
+			select {
+			case <-monitor.acc.Delivered():
+				<-monitor.sem
+			}
+		}
```
ssoroka, line 138: range works really well with channels and closes cleanly
```suggestion
		for range monitor.acc.Delivered() {
				<-monitor.sem
		}
```

plugins/inputs/directory_monitor/directory_monitor.go
```diff
@@ -0,0 +1,468 @@
...
+				monitor.filesInUse.Set(filePath, struct{}{})
+				filesToProcess = append(filesToProcess, fileInfo)
+			}
+		}
+
+		monitor.processFileBatch(filesToProcess, acc)
+
+		select {
+		// Monitor in intervals.
+		case <-time.After(monitor.MonitorInterval.Duration):
```
ssoroka, line 198: this sort of ignores how long it's been running for already.

plugins/inputs/directory_monitor/directory_monitor.go
```diff
@@ -0,0 +1,468 @@
...
+
+	// Handle gzipped files.
+	var reader io.Reader
+	if filepath.Ext(filePath) == ".gz" {
+		reader, err = gzip.NewReader(file)
+		if err != nil {
+			return err
+		}
+	} else {
+		reader, _ = utfbom.Skip(monitor.decoder.Reader(file))
```
ssoroka, line 273: we're discarding the actual detected encoding here, but I doubt we can handle anything other than UTF-8, and if it's not UTF-8 we would need to error or convert it (but none of the other plugins do this)

plugins/inputs/directory_monitor/directory_monitor.go
```diff
@@ -0,0 +1,468 @@
...
+	}
+
+	return monitor.parseFile(parser, reader)
+}
+
+func (monitor *DirectoryMonitor) parseFile(parser parsers.Parser, reader io.Reader) error {
+	// JSON files can't be read line-by-line with the current parser.
```
ssoroka, line 280: check other plugins. this is supported elsewhere. Reading the whole file into memory for a data dump is not an option. needs further discussion.

plugins/inputs/directory_monitor/directory_monitor.go
```diff
@@ -0,0 +1,468 @@
...
+		select {
+		case monitor.sem <- empty{}:
+			monitor.acc.AddTrackingMetricGroup([]telegraf.Metric{m})
+			continue
+		default:
+		}
+
+		// Block until room is available to add metrics.
+		select {
+		case monitor.sem <- empty{}:
```
ssoroka, line 351: Ideally we need to watch for quit here, unless you intend to block Telegraf shutdown, which might be the case. may require discussion.

plugins/inputs/directory_monitor/directory_monitor.go
```diff
@@ -0,0 +1,468 @@
...
+}
+
+func (monitor *DirectoryMonitor) Init() error {
+	if monitor.Directory == "" || monitor.FinishedDirectory == "" {
+		return errors.New("Missing one of the following required config options: directory, finished_directory.")
+	}
+
+	// Finished directory can be created if not exists for convenience.
+	if _, err := os.Stat(monitor.FinishedDirectory); os.IsNotExist(err) {
+		os.Mkdir(monitor.FinishedDirectory, 0777)
```
ssoroka, line 406: os.Mkdir can fail.

plugins/inputs/directory_monitor/directory_monitor.go
```diff
@@ -0,0 +1,468 @@
...
+		}
+
+		monitor.ErrorDirectory = errorDirectoryName
+	}
+
+	if monitor.CharacterEncoding == "" {
+		monitor.CharacterEncoding = "utf-8"
+	}
+
+	monitor.waitGroup = new(sync.WaitGroup)
```
ssoroka, line 432:
```suggestion
	monitor.waitGroup = &sync.WaitGroup{}
```

ssoroka (changes requested):
Looks well thought through. I think we can clean it up further. Mostly minor feedback, but I think I'd like the parallelism to go away. It complicates the code and we're not going to get any better throughput for it. Keep in mind the limiting factor is going to be the TrackingMetrics by a long shot.

plugins/inputs/directory_monitor/directory_monitor.go
```diff
@@ -0,0 +1,422 @@
...
+  ## If not provided, erroring files will stay in the monitored directory.
+  # error_directory = ""
+  #
+  ## The amount of time a file is allowed to sit in the directory before it is picked up.
+  ## This time can generally be low but if you choose to have a very large file written to the directory and it's potentially slow,
+  ## set this higher so that the plugin will wait until the file is fully copied to the directory.
+  # directory_duration_threshold = "50ms" 
+  #
+  ## A list of the only file names to monitor, if necessary. Supports regex. If left blank, all files are ingested.
+  # files_to_monitor = [".*.csv"]
```
ssoroka, line 45: this is a misleading regex example. it will match the file "foo.notcsv"

plugins/inputs/directory_monitor/directory_monitor.go
```diff
@@ -0,0 +1,422 @@
...
+  #
+  ## A list of files to ignore, if necessary. Supports regex.
+  # files_to_ignore = [".DS_Store"]
+  #
+  ## Maximum lines of the file to process that have not yet be written by the
+  ## output. For best throughput set based on the number of metrics and the size of the output's metric_batch_size.
+  ## Warning: setting this number too high can lead to a high number of dropped metrics.
```
ssoroka, line 52:
```suggestion
  ## output. For best throughput set to the size of the output's metric_buffer_limit.
  ## Warning: setting this number higher than the output's metric_buffer_limit can cause dropped metrics.
```

Jagularr:
!retry-checks

plugins/inputs/directory_monitor/directory_monitor.go
```diff
@@ -0,0 +1,422 @@
...
+  # max_buffered_metrics = 1000
```
ssoroka, line 53:
```suggestion
  # max_buffered_metrics = 10000
```

plugins/inputs/directory_monitor/directory_monitor.go
```diff
@@ -0,0 +1,422 @@
...
+  ## more about them here:
+  ## https://github.com/influxdata/telegraf/blob/master/docs/DATA_FORMATS_INPUT.md
+  ## NOTE: We currently only support parsing newline-delimited JSON. See the format here: https://github.com/ndjson/ndjson-spec
+  data_format = "influx"
+`
+
+var (
+	defaultFilesToMonitor             = []string{}
+	defaultFilesToIgnore              = []string{}
+	defaultMaxBufferedMetrics         = 1000
```
ssoroka, line 73:
```suggestion
	defaultMaxBufferedMetrics         = 10000
```

ssoroka (commented):
consider removing the parallelism as I don't think it's adding anything here.

plugins/inputs/directory_monitor/directory_monitor.go
Jagularr: ha, yup

ssoroka (approved):

PR #8754, opened 2021-01-26
chore: remove references to outdated InfluxDB settings.

danxmoran:
The `max-series-per-database` and `max-values-per-tag` settings are/were only used by the `inmem` index option in InfluxDB. `inmem` is going away for good in v2.0.4 (releasing late this week / early next), so the recommendations to tweak those settings aren't relevant any more.

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -50,8 +50,6 @@ You can use the following techniques to avoid cardinality issues:
 
 - Use [metric filtering][] options to exclude unneeded measurements and tags.
 - Write to a database with an appropriate [retention policy][].
-- Limit series cardinality in your database using the
-  [max-series-per-database][] and [max-values-per-tag][] settings.
 - Consider using the [Time Series Index][tsi].
 - Monitor your databases using the [show cardinality][] commands.
 - Consult the [InfluxDB documentation][influx docs] for the most up-to-date techniques.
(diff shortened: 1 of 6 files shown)


srebhan (approved):
Looks good code-wise. Please remember to sign a CLA as otherwise the PR cannot be merged!

srebhan:
Oh I see you are a member of InfluxData, please ask Sebastian on how to handle this in the tiger-bot... :-)

Jagularr (approved):

ssoroka:
isn't this still relevant to anyone running InfluxDB pre v2.0.4?

danxmoran:
> isn't this still relevant to anyone running InfluxDB pre v2.0.4?

Yes, if they're running an `inmem` index. Some mitigations:
* The messaging from 2.0 GA has been that `inmem` is dead (the fact that it was still there until 2.0.4 was an oversight)
* The Telegraf plugin documentation embedded in the InfluxDB UI will still show these options for versions < 2.0.4

@russorat asked me to make the changes here to prevent the defunct settings from reappearing next time we update the plugin pages in the OSS UI code.

ssoroka:
I still don't understand. Telegraf is often used with 1.x, and those settings are still relevant there.

ssoroka:
Closing as this seems to do the wrong thing. Please comment if this should be reopened.

russorat:
Reopening this PR. This may not be the right approach, but the goal of all of these readmes is to help users avoid high cardinality when working with influxdb. 

Maybe a better solution rather than duplicating the information in the various readmes would be linking to a dedicated docs page to resolve these cardinality issues? maybe https://docs.influxdata.com/influxdb/v2.0/write-data/best-practices/schema-design/

If there is content that is in those links that isn't accessible from those docs pages, can we open up issues to add it to the docs?

ssoroka (approved):
alright I understand this a bit better now. these links are dead, so there's no point keeping them. Hopefully we can improve the docs further in another PR.

PR #8658, opened 2021-01-06
Procstat input plugin should use the same timestamp in all metrics in the same Gather() cycle.

Jagularr:
resolves #8648

PR #8755, opened 2021-01-27
Procstat: collect connections and listener for each proc

adrianlzt:
Procstat modified to add a new metric "procstat_tcp" with to values
"conn" and "listen".
It also adds number of connections for each TCP state.

Those metrics will be added only if it they are activated specifically.

Those values will contain a comma separated value of the endpoints where
the proc is connecting to or listening (IPv4 and IPv6).
Only for linux.

Local, virtual and docker interfaces are ignored.

If some proc is listening on 0.0.0.0, an endpoint for each of the
"public" (those not ignored as internal) IPv4 IPs is created.
If it is listening on :: (IPv6) an endpoint is created for each IPv4 and
IPv6.

For programs with one parent and several children, all listening in some
endpoint, only the parent process is taken into account. Child endpoints
are ignored.

Connections made to this cost (local port is one of the listening ports)
are ignored. To avoid having servers with thousands of connections.
We prefer to collect that info in the clients.

It is also added connection info (number of connections in each of the
TCP states) for each proc. Improving PR #5402

To gather tcp connections netlink is used, to avoid the cost of parsing
/proc/net/tcp(6), but /proc should be readed to get the mapping between
inodes and pids.

### Required for all PRs:


adrianlzt:
!signed-cla

adrianlzt:
Linux tests are failing because a new test added in #8658 

Darwin errors are mine.

adrianlzt:
I want to add someway to only gather the procs with connections, to avoid the extra cost of parsing all procs.

sspaink:
!signed-cla

sspaink:
@adrianlzt Sorry for the bot being a nuisance, there was a bug. Thanks for working on this PR!

adrianlzt:
Added an option to skip some metrics [as suggested](https://github.com/influxdata/telegraf/pull/5402#issuecomment-463022446)

plugins/inputs/procstat/procstat_linux.go
```diff
@@ -0,0 +1,109 @@
...
+	fields[prefix+"tcp_last_ack"] = counts[linux.TCP_LAST_ACK]
+	fields[prefix+"tcp_listen"] = counts[linux.TCP_LISTEN]
+	fields[prefix+"tcp_closing"] = counts[linux.TCP_CLOSING]
+}
+
+// addConnectionEnpoints add listen and connection endpoints to the procstat_tcp metric.
+// If listen is 0.0.0.0 or ::, it will be added one value for each of the IP addresses of the host.
+// Listeners in private IPs are ignored (maybe a flag could be added, but now the reasoning is matching connections between hosts).
+// Connections made to this server are ignored (the local port is one of the listening ports).
+func addConnectionEnpoints(acc telegraf.Accumulator, proc Process, netInfo NetworkInfo) error {
```
helenosheaa, line 39: missing d change to `addConnectionEndpoints`

plugins/inputs/procstat/README.md
```diff
@@ -157,11 +186,49 @@ the `win_perf_counters` input plugin as a more mature alternative.
     - running (int)
     - result_code (int, success = 0, lookup_error = 1)
 
+If ``connections_stats`` enabled, added fields:
```
helenosheaa, line 189: should this be tcp_stats?

ssoroka (changes requested):
looks pretty thorough. I think it needs a bit of cleanup on error handling and logging, and some flow control around that. if you add `Log telegraf.Logger` to the struct the plugin system will populate it, and you can use `testutils.Logger{}` in tests. 

Q: does github.com/elastic/gosigar/ use cgo?

sspaink:
Closing as there hasn't been any response to the latest reviews, but if you are still seeing this pull request merged please comment or re-open and we can help work through any outstanding issues.

Issue #8705, opened 2021-01-15
Prometheus [Input] plugin - Optimizing for bigger kubernetes clusters (500+ pods) when scraping thru 'monitor_kubernetes_pods'


vishiy:
## Feature Request

Prometheus [Input] plugin - Optimizing for bigger kubernetes clusters (500+ pods) when scraping thru 'monitor_kubernetes_pods'

### Current behavior:
Currently  when 'monitor_kubernetes_pods=true' , telegraf watches for pods with specific annotations to scrape metrics, as pods come & go (in all namespaces or in specified namespaces). This approach works for smaller clusters, its almost 100% not scaling in bigger clusters (more than 500+ pods), especially when running telegraf in A pod which does this scraping for all pods in the cluster. This is also a single point of scale failure/unreliability when scraping thru pod annotations using telegraf promethus input plugin.

### Proposal:
To introduce an additional option (may be like 'local_mode' or something more intuitive), which when TRUE, will get ONLY pods that are running in that node. It will fetch podlist locally for that node from the node's kubelet (instead of watching them thru API server as it does today) and scrape the ones with the same annotations as it is today. This will require running Telegraf as daemonset (in every node) in the cluster, which will do pod scraping in each node locally, when enabled. By default, this will be backward compatible (meaning this new option will be turned OFF/false by default and users can turn ON as they see the need)

### Desired behavior:
Pod annotation based scraping thru Telegraf, scale as k8s cluster scales.

### Use case: 
As Kubernetes starts to become defacto for running workloads, most production clusters are growing, and prometheus metric sources & metrics are widely available. To monitor them thru telegraf, we need Telegraf to have reliable way to scale & collect metrics as the cluster grows.

ssoroka:
This issue looks good. Do you think instead of a boolean flag it should ask for a list of nodes to query?

Either way I think we'd be in support of this, please do feel free to write up the PR if that's what you are intending.

gracewehner:
PR is out: #8762. Thanks.

sjwang90:
Closed in #8762

PR #8762, opened 2021-01-27
Prometheus [Input] plugin - Optimizing for bigger kubernetes clusters (500+ pods) when scraping thru 'monitor_kubernetes_pods'

gracewehner:
### Required for all PRs:

PR for the issue https://github.com/influxdata/telegraf/issues/8705.

If monitor_kubernetes_pods = true and the new monitor_kubernetes_pods_version = 2 is specified, the pod list is queried locally on the node every 60s instead of using the watch api. This allows for scalability for large clusters so the pods to scrape are distributed among the nodes, which requires telegraf to be run as a daemonset. The same annotations to choose which pods to scrape are used and the optional filtering by namespace and selectors is also the same, with this filtering done afterwards. The feature is backwards compatible and is used if and only if monitor_kubernetes_pods_version = 2 is specified in the config.

### DIFF
--- a/plugins/inputs/prometheus/README.md
+++ b/plugins/inputs/prometheus/README.md
@@ -35,7 +35,7 @@ in Prometheus format.
   # monitor_kubernetes_pods = true
   ## Get the list of pods to scrape from either:
   ##    - version 1 (default): the kubernetes watch api (cluster-wide)
-  ##    - version 2: the local cadvisor api (node-wide); for scalability
+  ##    - version 2: the local cadvisor api (node-wide); for scalability. Note that the environment variable NODE_IP must be set to the host IP.
   # monitor_kubernetes_pods_version = 1
   ## Restricts Kubernetes monitoring to a single namespace
   ##   ex: monitor_kubernetes_pods_namespace = "default"
@@ -92,7 +92,14 @@ Currently the following annotation are supported:
 
 Using the `monitor_kubernetes_pods_namespace` option allows you to limit which pods you are scraping.
 
-Using `monitor_kubernetes_pods_version = 2` allows more scalable scraping for pods which will scrape pods only in the node that telegraf is running. It will fetch the pod list locally from the node's kubelet. This will require running Telegraf as a daemonset in the cluster.
+Using `monitor_kubernetes_pods_version = 2` allows more scalable scraping for pods which will scrape pods only in the node that telegraf is running. It will fetch the pod list locally from the node's kubelet. This will require running Telegraf as a daemonset in the cluster. Note that the environment variable NODE_IP must be set to the host IP. This can be done in the yaml of the pod running telegraf: 
+```
+env:
+  - name: NODE_IP
+    valueFrom:
+      fieldRef:
+        fieldPath: status.hostIP
+ ```
 
 #### Bearer Token
 
--- a/plugins/inputs/prometheus/README.md
+++ b/plugins/inputs/prometheus/README.md
@@ -33,9 +33,12 @@ in Prometheus format.
   ## - prometheus.io/path: If the metrics path is not /metrics, define it with this annotation.
   ## - prometheus.io/port: If port is not 9102 use this annotation
   # monitor_kubernetes_pods = true
-  ## Get the list of pods to scrape from either:
-  ##    - version 1 (default): the kubernetes watch api (cluster-wide)
-  ##    - version 2: the local cadvisor api (node-wide); for scalability. Note that the environment variable NODE_IP must be set to the host IP.
+  ## Get the list of pods to scrape with either the scope of
+  ## - cluster: the kubernetes watch api (default), no need to specify
+  ## - node: the local cadvisor api; for scalability. Note that the environment variable NODE_IP must be set to the host IP.
+  # pod_scrape_scope = "cluster"
+  ## Only for node scrape scope: interval in seconds for how often to get updated pod list for scraping
+  # pod_scrape_interval = 60
   # monitor_kubernetes_pods_version = 1
   ## Restricts Kubernetes monitoring to a single namespace
   ##   ex: monitor_kubernetes_pods_namespace = "default"
@@ -92,7 +95,7 @@ Currently the following annotation are supported:
 
 Using the `monitor_kubernetes_pods_namespace` option allows you to limit which pods you are scraping.
 
-Using `monitor_kubernetes_pods_version = 2` allows more scalable scraping for pods which will scrape pods only in the node that telegraf is running. It will fetch the pod list locally from the node's kubelet. This will require running Telegraf as a daemonset in the cluster. Note that the environment variable NODE_IP must be set to the host IP. This can be done in the yaml of the pod running telegraf: 
+Using `pod_scrape_scope = "node"` allows more scalable scraping for pods which will scrape pods only in the node that telegraf is running. It will fetch the pod list locally from the node's kubelet. This will require running Telegraf in every node of the cluster. Note that the environment variable NODE_IP must be set to the host IP. This can be done in the yaml of the pod running telegraf:
 ```
 env:
   - name: NODE_IP
@@ -100,6 +103,8 @@ env:
       fieldRef:
         fieldPath: status.hostIP
  ```
+
+If using node level scrape scope, `pod_scrape_interval` specifies how often (in seconds) the pod list for scraping should updated.
 
 #### Bearer Token
 
--- a/plugins/inputs/prometheus/README.md
+++ b/plugins/inputs/prometheus/README.md
@@ -39,7 +39,6 @@ in Prometheus format.
   # pod_scrape_scope = "cluster"
   ## Only for node scrape scope: interval in seconds for how often to get updated pod list for scraping
   # pod_scrape_interval = 60
-  # monitor_kubernetes_pods_version = 1
   ## Restricts Kubernetes monitoring to a single namespace
   ##   ex: monitor_kubernetes_pods_namespace = "default"
   # monitor_kubernetes_pods_namespace = ""
--- a/plugins/inputs/prometheus/README.md
+++ b/plugins/inputs/prometheus/README.md
@@ -35,9 +35,13 @@ in Prometheus format.
   # monitor_kubernetes_pods = true
   ## Get the list of pods to scrape with either the scope of
   ## - cluster: the kubernetes watch api (default), no need to specify
-  ## - node: the local cadvisor api; for scalability. Note that the environment variable NODE_IP must be set to the host IP.
+  ## - node: the local cadvisor api; for scalability. Note that the config node_ip or the environment variable NODE_IP must be set to the host IP.
   # pod_scrape_scope = "cluster"
+  ## Only for node scrape scope: node IP of the node that telegraf is running on.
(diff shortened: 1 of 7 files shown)


gracewehner:
#8705

vishiy:
@ssoroka - can you please look for this to be merged ?

plugins/inputs/prometheus/README.md
```diff
@@ -33,6 +33,10 @@ in Prometheus format.
   ## - prometheus.io/path: If the metrics path is not /metrics, define it with this annotation.
   ## - prometheus.io/port: If port is not 9102 use this annotation
   # monitor_kubernetes_pods = true
+  ## Get the list of pods to scrape from either:
+  ##    - version 1 (default): the kubernetes watch api (cluster-wide)
+  ##    - version 2: the local cadvisor api (node-wide); for scalability. Note that the environment variable NODE_IP must be set to the host IP.
```
ssoroka, line 38: How do you feel about accepting node_ip as a string parameter? I find environment variables aren't great for discoverability. I'm not against it also setting the value from env NODE_IP by default in init()

plugins/inputs/prometheus/README.md
```diff
@@ -33,6 +33,10 @@ in Prometheus format.
...
+  # monitor_kubernetes_pods_version = 1
```
ssoroka, line 39: What do you think about giving this a more meaningful name like:
`pod_scrape_scope = "cluster"` and `pod_scrape_scope = "node"`

plugins/inputs/prometheus/kubernetes.go
```diff
@@ -15,13 +19,29 @@ import (
...
+}
+
+type podResponse struct {
+	Kind       string        `json:"kind"`
+	ApiVersion string        `json:"apiVersion"`
+	Metadata   podMetadata   `json:"metadata"`
+	Items      []*corev1.Pod `json:"items,string,omitempty"`
+}
+
+const cAdvisorCallInterval = 60
```
ssoroka, line 43: should this be configurable?

plugins/inputs/prometheus/kubernetes.go
```diff
@@ -126,6 +150,169 @@ func (p *Prometheus) watch(ctx context.Context, client *k8s.Client) error {
 	}
 }
 
+func (p *Prometheus) cAdvisor(ctx context.Context, client *k8s.Client) error {
+	p.Log.Infof("Using monitor pods version 2 to get pod list using cAdvisor.")
+
+	nodeIP := os.Getenv("NODE_IP")
+	if nodeIP == "" {
+		return errors.New("The environment variable NODE_IP is not set. Cannot get pod list for monitor_kubernetes_pods using version 2.")
+	}
```
ssoroka, line 159: based on my other suggestions, you'd want to do this in the init() function, and not error as it would just be a default. You'd move the error to a `func (p *Prometheus) Init() error{}` function, as that's called after config is loaded, and if it's not set at that point, neither the ENV variable nor the config set it.

plugins/inputs/prometheus/kubernetes.go
```diff
@@ -126,6 +150,169 @@ func (p *Prometheus) watch(ctx context.Context, client *k8s.Client) error {
...
+		p.Log.Errorf("Error parsing the specified label selector(s): %s", err.Error())
+	}
+	fieldSelector, err := fields.ParseSelector(p.KubernetesFieldSelector)
+	if err != nil {
+		p.Log.Errorf("Error parsing the specified field selector(s): %s", err.Error())
+	}
+	if fieldSelectorIsSupported(fieldSelector) {
+		p.Log.Errorf("A specified field selector is not supported as a field selector for pods")
+	}
+	p.Log.Infof("Using the label selector: %v and field selector: %v", labelSelector, fieldSelector)
```
ssoroka, line 173: let's do any config-related validation in the `func (p *Prometheus) Init() error{}` function, and return errors instead of logging them

plugins/inputs/prometheus/kubernetes.go
```diff
@@ -126,6 +150,169 @@ func (p *Prometheus) watch(ctx context.Context, client *k8s.Client) error {
...
+
+	// Set InsecureSkipVerify for cAdvisor client since Node IP will not be a SAN for the CA cert
+	tlsConfig := client.Client.Transport.(*http.Transport).TLSClientConfig
+	tlsConfig.InsecureSkipVerify = true
```
ssoroka, line 177: Hmm. no way to make this work otherwise?

plugins/inputs/prometheus/kubernetes.go
```diff
@@ -126,6 +150,169 @@ func (p *Prometheus) watch(ctx context.Context, client *k8s.Client) error {
...
+	req, err := http.NewRequest("GET", podsUrl, nil)
+	if err != nil {
+		return err
+	}
+	client.SetHeaders(req.Header)
+
+	// Update right away so code is not waiting the length of the cAdvisorCallInterval initially
+	err = updateCadvisorPodList(ctx, p, client, req, labelSelector, fieldSelector)
+	if err != nil {
+		return err
```
ssoroka, line 190: Consider wrapping more context in certain places where you think it might be helpful to aid in debugging problems. for example:
```
		return fmt.Errorf("updating pod list: %w", err)
```

plugins/inputs/prometheus/kubernetes.go
```diff
@@ -126,6 +150,169 @@ func (p *Prometheus) watch(ctx context.Context, client *k8s.Client) error {
...
+}
+
+func updateCadvisorPodList(ctx context.Context, p *Prometheus, client *k8s.Client, req *http.Request,
+	labelSelector labels.Selector, fieldSelector fields.Selector) error {
+
+	resp, err := client.Client.Do(req)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
```
ssoroka, line 213: check the response code is a success before trying to process the body. Also consider using a json.Decoder instead of ioutil.ReadAll if the bodies can be large.

plugins/inputs/prometheus/kubernetes.go
```diff
@@ -180,14 +367,21 @@ func registerPod(pod *corev1.Pod, p *Prometheus) {
 		return
 	}
 	podURL := p.AddressToURL(URL, URL.Hostname())
-	p.lock.Lock()
+
+	// Locks earlier if using cAdvisor calls - makes a new list each time
+	// rather than updating and removing from the same list
+	if p.MonitorPodsVersion != 2 {
+		p.lock.Lock()
```
ssoroka, line 374: if you use defer you can drop the second conditional.
```
		p.lock.Lock()
    defer p.lock.Unlock()
```
But you may want to move the locks out of here as the caller should probably lock.

plugins/inputs/prometheus/kubernetes.go
```diff
@@ -126,6 +150,169 @@ func (p *Prometheus) watch(ctx context.Context, client *k8s.Client) error {
...
+	responseBody, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return err
+	}
+	cadvisorPodsResponse := podResponse{}
+	json.Unmarshal([]byte(responseBody), &cadvisorPodsResponse)
+	pods := cadvisorPodsResponse.Items
+
+	// Updating pod list to be latest cadvisor response
+	p.lock.Lock()
```
ssoroka, line 224: registerPod already locks, and this will cause double locks and panic. You'll need to unlock before the loop, which means kubernetesPods will need a default value of an empty map or something. May be better off removing the locks from registerPod.

plugins/inputs/prometheus/kubernetes.go
```diff
@@ -126,6 +150,169 @@ func (p *Prometheus) watch(ctx context.Context, client *k8s.Client) error {
...
+	p.kubernetesPods = nil
```
ssoroka, line 225: should be an empty map, and not nil?

plugins/inputs/prometheus/prometheus.go
```diff
@@ -57,12 +57,13 @@ type Prometheus struct {
...
-	MonitorPods    bool   `toml:"monitor_kubernetes_pods"`
-	PodNamespace   string `toml:"monitor_kubernetes_pods_namespace"`
-	lock           sync.Mutex
-	kubernetesPods map[string]URLAndAddress
-	cancel         context.CancelFunc
-	wg             sync.WaitGroup
+	MonitorPods        bool   `toml:"monitor_kubernetes_pods"`
+	MonitorPodsVersion int    `toml:"monitor_kubernetes_pods_version"`
+	PodNamespace       string `toml:"monitor_kubernetes_pods_namespace"`
+	lock               sync.Mutex
```
ssoroka, line 63: optional suggestion:
```suggestion
	sync.Mutex
```
then you can call `p.Lock()` instead of `p.lock.Lock()`

ssoroka (changes requested):
Looks pretty good. I'd like to see more tests around the cAdvisor code if possible, as it looks as if it will panic when run.

plugins/inputs/prometheus/prometheus.go
gracewehner: I would prefer to keep as is for now as to not change too much of the default code path

plugins/inputs/prometheus/kubernetes.go
gracewehner: I have added the defer to drop the second conditional

gracewehner:
Hi @ssoroka, I apologize for the delay. I have addressed and/or responded to your comments. Could please take another look? Thank you!

plugins/inputs/prometheus/kubernetes.go
ssoroka: It might simplify the logic, but I see you've resolved the main issue so I'll leave that up to you.

plugins/inputs/prometheus/prometheus.go
```diff
@@ -137,6 +155,34 @@ func (p *Prometheus) Init() error {
 		p.Log.Warnf("Use of deprecated configuration: 'metric_version = 1'; please update to 'metric_version = 2'")
 	}
 
+	// Config proccessing for node scrape scope for monitor_kubernetes_pods
+	p.isNodeScrapeScope = strings.EqualFold(p.PodScrapeScope, "node")
+	if p.isNodeScrapeScope {
+		p.nodeIP = os.Getenv("NODE_IP")
+		if p.nodeIP == "" {
+			return errors.New("The environment variable NODE_IP is not set. Cannot get pod list for monitor_kubernetes_pods using node scrape scope")
+		}
```
ssoroka, line 164: consider defaulting this to os.Getenv("NODE_IP") and allowing it to be set from config. ... actually, I think it's a requirement that all settings need to be configurable from config settings, though env variable defaults are fine

gracewehner: I have added this as a config setting. If the config for this is empty or the specified ip is invalid, it then checks if the env var exists and is valid

ssoroka (approved):

ssoroka:
Thanks for all the work here, much appreciated!

rashmichandrashekar:
> Thanks for all the work here, much appreciated!

Thanks @ssoroka, @vishiy and @gracewehner!

Issue #4507, opened 2018-08-02
Telegraf RPM install script can't detect Suse 15/SLE 15


chrmeyer:
### Relevant telegraf.conf:
n/a
### System info:

Tested with telegraf-1.7.2-1.x86_64 on SLE15

### Steps to reproduce:

Install Telegraf RPM under SLE 15

### Expected behavior:
Telegraf service is available in systemd
### Actual behavior:
telegraf.service is not available in systemd
### Additional info:
It seems that the telegraf RPM install script only tests for the path "/etc/SuSE-release" to detect if it's on Suse OS. Under SLE15 the path do not longer exist. You should better aditional test for content in /etc/os-release.

powersj:
Hi,

I have a PR up at https://github.com/influxdata/telegraf/pull/10570 with a change. I know there can still be an issue with the dependency but wanted to let you know this part is getting looked at.

Thanks!

PR #8765, opened 2021-01-28
fix: telegraf.service startup script on sles15 during instalation

sanekkurt:
https://github.com/influxdata/telegraf/issues/4507 solved
Due to the absence of the /etc/SuSE-release file, telegraph.service was not started in the new versions of sles 15. Therefore, the logic for running the service in linux sles was added to the logic for checking the $NAME variable. Now the script can work both in old versions of linux sles and in new versions.


sanekkurt:
!signed-cla

scripts/rpm/post-remove.sh
```diff
@@ -44,5 +44,16 @@ elif [[ -f /etc/os-release ]]; then
...
+        if [[ "$1" = "0" ]]; then
+            rm -f /etc/default/telegraf
+
+            if [[ "$(readlink /proc/1/exe)" == */systemd ]]; then
+                disable_systemd /usr/lib/systemd/system/telegraf.service
+            else
+                # Assuming sysv
+                disable_chkconfig
+            fi
+        fi
```
reimda, line 57: This is copied from line 20.  Would you put it in a function and call the function in both places?  Maybe name the function disable_systemd_or_chkconfig

scripts/rpm/post-install.sh
```diff
@@ -84,5 +84,18 @@ elif [[ -f /etc/os-release ]]; then
...
+        else
+            # Assuming SysVinit
+            install_init
+            # Run update-rc.d or fallback to chkconfig if not available
+            if which update-rc.d &>/dev/null; then
+                install_update_rcd
+            else
+                install_chkconfig
+            fi
+        fi
```
reimda, line 99: This is copied from line 58.  Would you put it in a function and call it from both places?  Maybe name the function install_systemd_or_chkconfig

reimda (changes requested):

scripts/rpm/post-remove.sh
sanekkurt: Yes, of course. Added a function in a new commit

scripts/rpm/post-install.sh
sanekkurt: Yes, of course. Added a function in a new commit

scripts/rpm/post-install.sh
```diff
@@ -24,6 +24,22 @@ function install_chkconfig {
     chkconfig --add telegraf
 }
 
+function install_systemd_or_chkonfig {
```
reimda, line 27: Thanks!  But it looks like there's a typo here and also in the disable function 
```suggestion
function install_systemd_or_chkconfig {
```

scripts/rpm/post-remove.sh
```diff
@@ -15,7 +15,7 @@ function disable_chkconfig {
     rm -f /etc/init.d/telegraf
 }
 
-if [[ -f /etc/redhat-release ]] || [[ -f /etc/SuSE-release ]]; then
+function disable_systemd_or_chkonfig {
```
reimda, line 18: :)
```suggestion
function disable_systemd_or_chkconfig {
```

reimda (changes requested):

scripts/rpm/post-install.sh
sanekkurt: Sorry I didn't notice the mistake )

sspaink:
@sanekkurt sorry for not getting back to you for a while, are you still interested in working on this PR? Could you re-base it and we can give it a final review.

reimda:
The bot didn't notice new commits from two days ago so I'm reopening

powersj:
Hi,

I'm going to close this PR as a fix for #4507 has landed. That fix makes the install scripts OS-agnostic to avoid needing to add specific distro names.

Thanks!

Issue #8776, opened 2021-01-30
proxmox plugin - [inputs.proxmox] Error in plugin: search domain is not set


er-pratyush:
### Relevant telegraf.conf:

```toml
# Configuration for telegraf agent
[agent]
  interval = "10s"
  round_interval = true
  metric_batch_size = 1000
  metric_buffer_limit = 10000
  collection_jitter = "0s"
  flush_interval = "10s"
  flush_jitter = "0s"
  precision = ""
  debug = true
  quiet = false
  logfile = ""
  hostname = ""
  omit_hostname = false
[[outputs.influxdb_v2]]
  urls = ["http://192.x.x.x:8086"]
  token = "xxxx=="
  organization = "xxx"
  bucket = "proxmox"
[[inputs.proxmox]]
  ## API connection configuration. The API token was introduced in Proxmox v6.2. Required permissions for user and token: PVEAuditor role on /.
  base_url = "https://192.x.x.x:8006/api2/json"
  api_token = "infTele1@pve!monitor=xxx-xxx-xxx-xxx-xxx"
  ## Node name, defaults to OS hostname
  node_name = "pve"

  ## Optional TLS Config
  # tls_ca = "/etc/telegraf/ca.pem"
  # tls_cert = "/etc/telegraf/cert.pem"
  # tls_key = "/etc/telegraf/key.pem"
  ## Use TLS but skip chain & host verification
  insecure_skip_verify = true

  # HTTP response timeout (default: 5s)
  response_timeout = "5s"
```

### System info:

telegraf -version
Telegraf 1.17.2 (git: HEAD 74011e22)

pve version
proxmox-ve: 6.3-1 (running kernel: 5.4.78-2-pve)
pve-manager: 6.3-3 (running version: 6.3-3/eee5f901)
pve-kernel-5.4: 6.3-3
pve-kernel-helper: 6.3-3
pve-kernel-5.4.78-2-pve: 5.4.78-2
pve-kernel-5.4.73-1-pve: 5.4.73-1

telegraf host : ubuntu 20.04-1 container on proxmox
### Docker
no Docker

### Steps to reproduce:

running telegraf throws error "search domain is not set"

### Expected behavior:

should not throw error

### Actual behavior:

2021-01-30T01:18:45Z I! Starting Telegraf 1.17.2
2021-01-30T01:18:45Z I! Using config file: /etc/telegraf/telegraf.conf
2021-01-30T01:18:45Z I! Loaded inputs: proxmox
[...]

PR #8777, opened 2021-01-30
Update proxmox.go

er-pratyush:
### Required for all PRs:

### DIFF
--- a/plugins/inputs/proxmox/proxmox.go
+++ b/plugins/inputs/proxmox/proxmox.go
@@ -93,10 +93,10 @@ func getNodeSearchDomain(px *Proxmox) error {
 		return err
 	}
 
-	if nodeDns.Data.Searchdomain == "" {
+	if nodeDns.Data.Search == "" {
 		return errors.New("search domain is not set")
 	}
-	px.nodeSearchDomain = nodeDns.Data.Searchdomain
+	px.nodeSearchDomain = nodeDns.Data.Search
 
 	return nil
 }


Jagularr (changes requested):
The field 'Search' does not exist on the 'Data' struct. The field would need to be renamed from 'Searchdomain' for the build to succeed. however, can you add some context as to what this change is needed for?

er-pratyush:
refer https://github.com/influxdata/telegraf/issues/8776
JSON data from Proxmox API returns key as data.search and the if condition in plugin was checking for Data.Searchdomain.
it seemed a quick fix and i made the change without looking at struct (mistake on my end). though later i realized that it was in line with data struct.

easterism:
The call https://localhost:8006/api2/json/nodes/nodename/dns returns
 `{"data":{"search":"node.local","dns2":"8.8.8.8","dns1":"192.168.0.22"}}`
 Confirmed on PVE 6.2-4

aantuness:
The call https://localhost:8006/api2/json/nodes/nodename/dns returns
{"data":{"search":"<name>.local","dns1":"8.8.8.8"}}
Confirmed on PVE 6.4-4

connormckelvey:
I don't think there is anything wrong with the use of `Searchdomain` as it stands. if you look here: https://github.com/influxdata/telegraf/blob/master/plugins/inputs/proxmox/structs.go#L65, Searchdomain has a JSON tag of `search`, which matches the JSON payload referenced above.

PR #6882, opened 2020-01-09
Non-systemd udev path + adjusted unit test

red15:
### Required for all PRs:

- [X] Signed [CLA](https://influxdata.com/community/cla/).
- [X] Associated README.md updated.
- [X] Has appropriate unit tests.

Now only probing 2 paths at most.

This is a rewrite with lessons learned by writing #5180 .
Closes #3900

PR #8785, opened 2021-02-02
Non systemd support with unittest

red15:
### Required for all PRs:

Fixes https://github.com/influxdata/telegraf/issues/3900


red15:
Could replace #6882 as the CircleCI integration there has gone bonkers.
@srebhan @ssoroka , can we replace the old one ? Just close it and then I will pop the draft off from this MR.

red15:
> 
> 
> Could replace #6882 as the CircleCI integration there has gone bonkers.
> @srebhan @ssoroka , can we replace the old one ? Just close it and then I will pop the draft off from this MR.

Actually looks like CircleCI has suddenly realized it needs to test this also on the other MR.
So the old MR might contain more relevant information.

Jagularr (approved):

plugins/inputs/diskio/diskio_linux_test.go
```diff
@@ -19,19 +19,31 @@ S:foo/bar/devlink1
...
 // setupNullDisk sets up fake udev info as if /dev/null were a disk.
-func setupNullDisk(t *testing.T) func() error {
-	td, err := ioutil.TempDir("", ".telegraf.TestDiskInfo")
+func setupNullDisk(t *testing.T, s *DiskIO, devName string) func() error {
+	td, err := ioutil.TempFile("", ".telegraf.DiskInfoTest")
 	require.NoError(t, err)
 
-	origUdevPath := udevPath
+	if s.infoCache == nil {
+		s.infoCache = map[string]diskInfoCache{}
```
srebhan, line 27: I'd prefer
```suggestion
		s.infoCache = make(map[string]diskInfoCache, 0)
```
but it seems cosmetics...

plugins/inputs/diskio/diskio_linux.go
```diff
@@ -37,9 +35,22 @@ func (s *DiskIO) diskInfo(devName string) (map[string]string, error) {
...
+		major := unix.Major(uint64(stat.Rdev))
+		minor := unix.Minor(uint64(stat.Rdev))
+		udevDataPath = fmt.Sprintf("/run/udev/data/b%d:%d", major, minor)
+
+		f, err := os.Open(udevDataPath)
+		defer f.Close()
+		if err != nil {
+			udevDataPath = fmt.Sprintf("/dev/.udev/db/block:%s", devName)
+		}
+	}
```
srebhan, line 53: Could you please move the check for last-standing `udevDataPath` up here so we have all the checking for that property in one place.

srebhan (commented):
Hey @red15! Nice to see you could circumvent CircleCI problems. Could you please address moving the check of the udev path  up to the construction place so we have all in one place. Otherwise we have to jump in code to see it's checked...

And please close #6882 in favor of this PR. Thanks.

plugins/inputs/diskio/diskio_linux_test.go
red15: Doing this introduced an error:
```plugins/inputs/diskio/diskio_linux.go:30:46: map[string]diskInfoCache literal is not a type```

Also this line wasn't touched by me at all so it feels a bit out of place to be changing this.

File: plugins/inputs/diskio/diskio_linux.go [hunk 3]
```diff
@@ -37,9 +35,31 @@ func (s *DiskIO) diskInfo(devName string) (map[string]string, error) {
...
+		major := unix.Major(uint64(stat.Rdev))
+		minor := unix.Minor(uint64(stat.Rdev))
+		udevDataPath = fmt.Sprintf("/run/udev/data/b%d:%d", major, minor)
+
+		f, err := os.Open(udevDataPath)
+		defer f.Close() // Closing this when it goes out of scope.
+		if err != nil {
+			// This path failed, try the fallback .udev style (non-systemd)
+			udevDataPath = fmt.Sprintf("/dev/.udev/db/block:%s", devName)
+			f, err2 := os.Open(udevDataPath)
```
srebhan, line 53: How about 
```suggestion
			f, err = os.Open(udevDataPath)
```
to avoid the ugly `err2`?

plugins/inputs/diskio/diskio_linux.go
srebhan: Nice! Please close this as resolved.

plugins/inputs/diskio/diskio_linux.go
```diff
@@ -37,9 +35,31 @@ func (s *DiskIO) diskInfo(devName string) (map[string]string, error) {
...
+			defer f.Close() // Closing this when it goes out of scope.
+			if err2 != nil {
+				return nil, err2
+			}
+		}
+	}
+	// Final open of the confirmed (or the previously detected/used) udev file
+	f, err := os.Open(udevDataPath)
```
srebhan, line 61: Maybe you should check `err` just in case...

red15: Added the error checking after replacing the `Open` call(s) above with stat calls to avoid having to close a lingering file handle.

plugins/inputs/diskio/diskio_linux.go
```diff
@@ -45,15 +45,14 @@ func (s *DiskIO) diskInfo(devName string) (map[string]string, error) {
...
 			// This path failed, try the fallback .udev style (non-systemd)
 			udevDataPath = fmt.Sprintf("/dev/.udev/db/block:%s", devName)
-			f, err2 := os.Open(udevDataPath)
-			defer f.Close() // Closing this when it goes out of scope.
-			if err2 != nil {
-				return nil, err2
+			f, err := os.Stat(udevDataPath)
+			if err != nil {
+				// Giving up, cannot retrieve disk info
+				return nil, err
```
srebhan: Please avoid the stats call as this gets race in the worst case. What happens if the file is deleted/moved right between the stats and the open call? You could do the open-fallthrough and only place the `defer f.Close()` call directly below the error handling!?

srebhan:
Please fix the code and LGTM-bot comment! Finally if you think you are done and want me to review, please change the "Draft" state of your PR. :-)

srebhan (approved):
Looks good to me. Thank you for your work @red15!

srebhan (approved):

PR #8797, opened 2021-02-03
Revive fixes - part 1

zak-pawel:
**[Revive](https://github.com/mgechev/revive)** is fast, configurable, extensible, flexible, and beautiful linter for Go. Drop-in replacement of golint.

This is part1 of fixes for problems found for following set of rules:
- [rule.blank-imports]
  - Disallows blank imports -> Blank import should be only in a main or test package, or have a comment justifying it.
- [rule.context-as-argument]
  - `context.Context` should be the first argument of a function. -> By [convention](https://github.com/golang/go/wiki/CodeReviewComments#contexts), `context.Context` should be the first parameter of a function. This rule spots function declarations that do not follow the convention.
- [rule.dot-imports]
  - Forbids `.` imports. -> Importing with `.` makes the programs much harder to understand because it is unclear whether names belong to the current package or to an imported package. More information [here](https://github.com/golang/go/wiki/CodeReviewComments#import-dot)
- [rule.error-strings]
  - Conventions around error strings. -> Description: By convention, for better readability, error messages should not be capitalized or end with punctuation or a newline. More information [here](https://github.com/golang/go/wiki/CodeReviewComments#error-strings)
- [rule.error-naming]
  - Naming of error variables. -> By convention, for the sake of readability, variables of type `error` must be named with the prefix `err`.
- [rule.indent-error-flow]
  - Prevents redundant else statements. -> To improve the readability of code, it is recommended to reduce the indentation as much as possible. This rule highlights redundant else-blocks that can be eliminated from the code. More information [here](https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow)
- [rule.errorf]
  - Should replace `errors.New(fmt.Sprintf())` with `fmt.Errorf()` -> It is possible to get a simpler program by replacing `errors.New(fmt.Sprintf())` with `fmt.Errorf()`. This rule spots that kind of simplification opportunities.

PR #8798, opened 2021-02-04
Run revive linter in CI

sspaink:
Run the [revive linter](https://github.com/mgechev/revive) for each pull request, but don't fail on it just output the results in a friendly format. Added a config toml with the rule "exported" removed that had 2K issues reported for things exported without a comment.

Once the errors are more cleaned up, this should be updated to warn/error if a pull request introduces issues. Related to the pr #8797.


zak-pawel:
LGTM!

zak-pawel (approved):

revivelint_config.toml
```diff
@@ -0,0 +1,24 @@
...
+[rule.increment-decrement]
+[rule.var-naming]
+[rule.var-declaration]
+[rule.package-comments]
+[rule.range]
+[rule.receiver-naming]
+[rule.time-naming]
+[rule.unexported-return]
+[rule.indent-error-flow]
+[rule.errorf]
```
srebhan, line 24: Missing newline. :-)

Makefile
```diff
@@ -130,6 +130,15 @@ vet:
 		exit 1; \
 	fi
 
+.PHONY: lint
+lint:
+ifeq (, $(shell which revive))
+	$(shell go install github.com/mgechev/revive)
+	$(info revive installed)
+endif
```
srebhan, line 138: Is this really a good idea to install it here without asking the user? Maybe we better invert the if and tell the user to install `revive`!?

srebhan:
Not sure if this is the right place but I'll try anyway. ;-)
Working through #8797 I found the following things that we might want to have as a rule:

1. Do not use `log` in a plugin of type input, output, aggregator, processor (parser and serializer are special) but rather use `telegraf.Logger`.
2. Finding `\"%s\"` in fmt functions and use `%q` instead.
3. Patterns of the form
```golang
if condition {
    return true
}
return false
```
or the inverse can be simplified to `return condition`.

zak-pawel:
> Not sure if this is the right place but I'll try anyway. ;-)
> Working through #8797 I found the following things that we might want to have as a rule:
> 
> 1. Do not use `log` in a plugin of type input, output, aggregator, processor (parser and serializer are special) but rather use `telegraf.Logger`.
> 2. Finding `\"%s\"` in fmt functions and use `%q` instead.
> 3. Patterns of the form
> 
> ```go
> if condition {
>     return true
> }
> return false
> ```
> 
> or the inverse can be simplified to `return condition`.

@srebhan (and @ssoroka, @reimda): Unfortunately it seems that to add own rule, accepted PR to https://github.com/mgechev/revive or fork is needed: https://github.com/mgechev/revive#extensibility

sspaink:
@zak-pawel that is an unfortunate limitation, maybe we should use golangci-lint for now which would let us create a Telegraf specific linter without having to push it upstream ([more info](https://golangci-lint.run/contributing/new-linters/)). I can just update this PR to change from revive->golangci-lint.
We could potentially run both linters if we want or wait until golangci-lint supports running revive.

Makefile
sspaink: @srebhan good point, from a CI perspective this was easier but I can move the installation logic out and into the circleci config and just have it print a error message if it isn't installed.

p-zak:
> @zak-pawel that is an unfortunate limitation, maybe we should use golangci-lint for now which would let us create a Telegraf specific linter without having to push it upstream ([more info](https://golangci-lint.run/contributing/new-linters/)). I can just update this PR to change from revive->golangci-lint.
> We could potentially run both linters if we want or wait until golangci-lint supports running revive.

I think that we can start with revive (as wew did :)) - because it is pretty good tool. We will try to fix as many problems as possible and after that (or in parallel) we can start to work with golangci-lint also.
In the meantime we need to think about custom rules.

srebhan (approved):
Perfect thanks!

srebhan:
@sspaink I discussed with @p-zak and I think at least my requests can be covered. 1. can be checked by not allowing to include `log` in the plugins. 2. might be an upstream acceptable change anyway, same as 3. The probably also exist at least partially.

p-zak:
> @sspaink I discussed with @p-zak and I think at least my requests can be covered. 1. can be checked by not allowing to include `log` in the plugins. 2. might be an upstream acceptable change anyway, same as 3. The probably also exist at least partially.

@srebhan we can cover these in the future, they should not be part of this PR :)

srebhan:
@p-zak I fully agree.

Makefile
```diff
@@ -130,6 +131,15 @@ vet:
 		exit 1; \
 	fi
 
+.PHONY: lint
+lint:
+ifeq (, $(shell which golangci-lint))
+	$(info golangci-lint can't be found, please install it: https://golangci-lint.run/usage/install/)
```
ssoroka, line 137: should this just try to install it?

sspaink: @srebhan made a good point that we might not want to install anything without prompting the user. I also want to have this download the release binary instead of installing from source like it is doing now (should be faster).

zak-pawel:
@sspaink Don't we need to wait for next release (> v1.36.0) of [golangci-lint](https://github.com/golangci/golangci-lint) to contain `revive` linter?

Makefile
srebhan: @ssoroka I vote for not installing it automatically as it's the machine of the user. There is a link to the upstream install instructions that should suffice for anyone building/developing for telegraf I guess. :-)

zak-pawel:
> @sspaink Don't we need to wait for next release (> v1.36.0) of [golangci-lint](https://github.com/golangci/golangci-lint) to contain `revive` linter? Right now we have unknown linter: https://app.circleci.com/pipelines/github/influxdata/telegraf/3456/workflows/54ea1dc6-8eda-4b53-8dc5-d79ec57cba0b/jobs/78136

And [v1.37.0](https://github.com/golangci/golangci-lint/releases/tag/v1.37.0) has been just released :)

.golangci.yml
```diff
@@ -0,0 +1,9 @@
+linters:
+  enable:
+    - revive
+
+run:
+  skip-dirs:
+    - scripts
+    - docs
+    - etc
```
zak-pawel, line 9: Maybe we can try with:
```
linters:
  enable:
    - revive

linters-settings:
  revive:
    rules:
      - name: blank-imports
      - name: context-as-argument
      - name: context-keys-type
      - name: dot-imports
      - name: error-return
      - name: error-strings
      - name: error-naming
      - name: exported
      - name: if-return
      - name: increment-decrement
      - name: var-naming
      - name: var-declaration
      - name: package-comments
      - name: range
      - name: receiver-naming
      - name: time-naming
      - name: unexported-return
      - name: indent-error-flow
      - name: errorf
      - name: empty-block
      - name: superfluous-else
      - name: unused-parameter
      - name: unreachable-code
      - name: redefines-builtin-id

run:
  skip-dirs:
    - scripts
    - docs
    - etc
  skip-files:
    - plugins/parsers/influx/machine.go*

issues:
  exclude:
    - don't use an underscore in package name
    - exported.*should have comment.*or be unexported
```

zak-pawel: It will:
- enable all suggested `revive` rules
- skip two "problematic" files: `plugins/parsers/influx/machine.go*`
- filter out underscore in package name and lack of comments on exported types/functions issues

PR #7662, opened 2020-06-10
Added retry to remote config get on startup

schmorgs:
### Required for all PRs:

- [X] Signed [CLA](https://influxdata.com/community/cla/).
- [X] Associated README.md updated.
- [X] Has appropriate unit tests.

PR #8803, opened 2021-02-04
Add default retry for load config via url

### LINKED ISSUE #7338: Agent won't start if remote HTTP config endpoint is down

helenosheaa:
Closes #7338

Related #7349, #7662 

This is a temporary fix which we intend to improve on:

- Retries three times at 10s intervals when receiving an error on loading config from a url incase of the endpoint being down.

- Log attempts and the specific failure response codes.

- Does not use env variables or add new flags 

- Tests added to demonstrate that on the final failure it will return the error.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -924,14 +924,9 @@ func fetchConfig(u *url.URL) ([]byte, error) {
 	req.Header.Set("User-Agent", internal.ProductToken())
 
 	retries := 3
-	for i := 1; i <= retries; i++ {
+	for i := 0; i <= retries; i++ {
 		resp, err := http.DefaultClient.Do(req)
 		if err != nil {
-			if i < retries {
-				log.Printf("Unable to connect to HTTP config server.  Retry %d of %d in %s. %s", i, retries, httpLoadConfigRetryInterval, err)
-				time.Sleep(httpLoadConfigRetryInterval)
-				continue
-			}
 			return nil, fmt.Errorf("Retry %d of %d failed connecting to HTTP config server %s", i, retries, err)
 		}
 
--- a/config/config.go
+++ b/config/config.go
@@ -49,6 +49,7 @@ var (
 		`"`, `\"`,
 		`\`, `\\`,
 	)
+	httpLoadConfigRetryInterval = 10 * time.Second
 )
 
 // Config specifies the URL/user/password for the database that telegraf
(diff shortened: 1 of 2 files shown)


config/config.go
```diff
@@ -921,17 +922,32 @@ func fetchConfig(u *url.URL) ([]byte, error) {
...
 	req.Header.Add("Accept", "application/toml")
 	req.Header.Set("User-Agent", internal.ProductToken())
-	resp, err := http.DefaultClient.Do(req)
-	if err != nil {
-		return nil, err
-	}
 
-	if resp.StatusCode != http.StatusOK {
-		return nil, fmt.Errorf("failed to retrieve remote config: %s", resp.Status)
+	retries := 3
```
Jagularr, line 926: nitpick but I guess this (retries) could be added to the constants at the top too, for consistency

Jagularr (approved):

config/config.go
helenosheaa: Yeah I get you. I only added the interval var up top just so I could access and change it for the test so it wouldn't take 30s to run

config/config.go
```diff
@@ -921,17 +922,32 @@ func fetchConfig(u *url.URL) ([]byte, error) {
...
+	for i := 1; i <= retries; i++ {
```
ssoroka, line 927: if retries is 0, this won't attempt at all. If we later move to user-configurable retries, that won't quite make sense

config/config.go
```diff
@@ -921,17 +922,32 @@ func fetchConfig(u *url.URL) ([]byte, error) {
...
+		resp, err := http.DefaultClient.Do(req)
+		if err != nil {
```
ssoroka, line 929: non-200 responses don't return errors, so we might want to not retry on err. err could indicate something more fundamental is wrong. Check the docs for what conditions Do returns errors.

config/config.go
helenosheaa: Right changed i to 0

config/config.go
helenosheaa: So it only returns on error either for client policy or failure to speak HTTP. So the use case was for when someones network is temporarily down I'm thinking the retries on http status code covers that so will just break if there is an err

ssoroka (approved):

Issue #7879, opened 2020-07-23
SNMP input - force IPv4/IPv6


jacobreid:
## Feature Request

Opening a feature request kicks off a discussion.

### Proposal:

Add an option for the SNMP input to be configured to use/prefer IPv4 or v6

### Current behavior:

(telegraf 1.14.5) SNMP input seems to prefer IPv6 over IPv4 where there are both A and AAAA records for a hostname, and will not fall back to IPv4 if attempting to gather data via IPv6 fails.

### Desired behavior:

Something in the config like `ipv6 = [true, false, auto]` where true uses ipv6 only if available (ignoring ipv4 if existing), false uses ipv4 only, and auto will try one and then the other if it fails (and ideally, remember which one worked for next run?) - in particular, trying IPv4 if an AAAA record is found but an attempt to gather data from the IPv6 address fails. True would be backwards compatible with current behaviour (fall back to IPv4 if no AAAA record, use only IPv6 and don't try v4 if there is an AAAA record) so probably the default setting.

### Use case: 
I have some network equipment with both A and AAAA records for its hostname that has a bug where SNMP only works over IPv4 but telegraf only attempts to reach it over IPv6.

It can be worked around by either pointing the SNMP input to the IPv4 address, or creating another hostname with only an A record, but it would be nice to use the canonical hostname instead.

synfinatic:
sadly, poor support for IPv6 is really common for a lot of vendors.  While ideally, they would all fix it it's unlikely that will happen in the near future.

jaydizz:
I can only second this. 
What comes to mind is specifying the transport protocol. 
GoSNMP aallows the following transport-protocols when setting up a connection:
```// connect to address addr on the given network
//
// https://golang.org/pkg/net/#Dial gives acceptable network values as:
//   "tcp", "tcp4" (IPv4-only), "tcp6" (IPv6-only), "udp", "udp4" (IPv4-only),"udp6" (IPv6-only), "ip",
//   "ip4" (IPv4-only), "ip6" (IPv6-only), "unix", "unixgram" and "unixpacket"```

Why is this not exposed in the configuration files?
This way we could easily specify `udp4://` or `udp6://` for the different agend_hosts.

jaydizz:
I created a PR (https://github.com/influxdata/telegraf/pull/8804)  that enables one to pass the connection-scheme to Gosnmp through the GosnmpWrapper.
It works perfectly in my test-setup.

Hipska:
[...]

PR #8804, opened 2021-02-04
Expose v4/v6-only connection-schemes through GosnmpWrapper

jaydizz:
Hi, 
This PR fixes https://github.com/influxdata/telegraf/issues/7879.

I modified the Gosnmpwrapper to expose tcp{4,6} and udp{4,6} as connection-scheme and pass them to Gosnmp. 
This way, one can indirectly specify the address-family to be used for snmp-requests. 
I updated the according README-files. 
In my small test-setup everything is working as intended. 

However, I did not find a unit-test for the snmp input-plugin or the wrapper to alter, so I assume this small 4 line change can (hopefully) be reviewed by hand. 
This is my first PR on a go-project ever, so please be gentle! :)

Have a nice day. 

### Required for all PRs:

### DIFF
--- a/internal/snmp/wrapper.go
+++ b/internal/snmp/wrapper.go
@@ -165,18 +165,8 @@ func (gs *GosnmpWrapper) SetAgent(agent string) error {
 	}
 
 	switch u.Scheme {
-	case "tcp":
-		gs.Transport = "tcp"
-	case "tcp4":
-		gs.Transport = "tcp4"
-	case "tcp6":
-		gs.Transport = "tcp6"
-	case "udp":
-		gs.Transport = "udp"
-	case "udp4":
-		gs.Transport = "udp4"
-	case "udp6":
-		gs.Transport = "udp6"
+	case "tcp", "tcp4", "tcp6", "udp", "udp4", "udp6":
+		gs.Transport = u.Scheme
 	default:
 		return fmt.Errorf("unsupported scheme: %v", u.Scheme)
 	}
--- a/internal/snmp/wrapper.go
+++ b/internal/snmp/wrapper.go
@@ -164,6 +164,11 @@ func (gs *GosnmpWrapper) SetAgent(agent string) error {
 		return err
 	}
 
+	// Only allow udp{4,6} and tcp{4,6}.
+	// Allowing ip{4,6} does not make sense as specifying a port
+	// requires the specification of a protocol.
+	// gosnmp does not handle these errors well, which is why
+	// they can result in cryptic errors by net.Dial.
 	switch u.Scheme {
 	case "tcp", "tcp4", "tcp6", "udp", "udp4", "udp6":
 		gs.Transport = u.Scheme
--- a/internal/snmp/wrapper.go
+++ b/internal/snmp/wrapper.go
@@ -171,12 +171,12 @@ func (gs *GosnmpWrapper) SetAgent(agent string) error {
 		gs.Transport = "tcp4"
 	case "tcp6":
 		gs.Transport = "tcp6"
+	case "udp":
+		gs.Transport = "udp"
 	case "udp4":
 		gs.Transport = "udp4"
 	case "udp6":
 		gs.Transport = "udp6"
-	case "udp":
-		gs.Transport = "udp"
 	default:
 		return fmt.Errorf("unsupported scheme: %v", u.Scheme)
 	}
--- a/internal/snmp/wrapper.go
+++ b/internal/snmp/wrapper.go
@@ -167,6 +167,14 @@ func (gs *GosnmpWrapper) SetAgent(agent string) error {
 	switch u.Scheme {
 	case "tcp":
 		gs.Transport = "tcp"
+	case "tcp4":
+		gs.Transport = "tcp4"
+	case "tcp6":
+		gs.Transport = "tcp6"
+	case "udp4":
+		gs.Transport = "udp4"
+	case "udp6":
+		gs.Transport = "udp6"
 	case "", "udp":
 		gs.Transport = "udp"
 	default:
--- a/internal/snmp/wrapper.go
+++ b/internal/snmp/wrapper.go
@@ -175,7 +175,7 @@ func (gs *GosnmpWrapper) SetAgent(agent string) error {
 		gs.Transport = "udp4"
 	case "udp6":
 		gs.Transport = "udp6"
-	case "", "udp":
+	case "udp":
 		gs.Transport = "udp"
 	default:
 		return fmt.Errorf("unsupported scheme: %v", u.Scheme)
--- a/plugins/inputs/snmp/README.md
+++ b/plugins/inputs/snmp/README.md
@@ -22,11 +22,13 @@ information.
 ```toml
 [[inputs.snmp]]
   ## Agent addresses to retrieve values from.
-  ##   format:  agents = ["<scheme>://<hostname>:<port>"]
-  ##   scheme:  either udp, udp4, udp6, tcp, tcp4, tcp6.  
+  ##   format:  agents = ["<scheme://><hostname>:<port>"]
+  ##   scheme:  optional, either udp, udp4, udp6, tcp, tcp4, tcp6.  
   ##            default is udp
+  ##   port:    optional
   ##   example: agents = ["udp://127.0.0.1:161"]
-  ##            agents = ["udp4://v4only-snmp-agent"]
+  ##            agents = ["tcp://127.0.0.1:161"]
+  ##            agents = ["udpv4://v4only-snmp-agent"]
   agents = ["udp://127.0.0.1:161"]
 
   ## Timeout for each request.
--- a/plugins/inputs/snmp/README.md
+++ b/plugins/inputs/snmp/README.md
@@ -28,7 +28,7 @@ information.
   ##   port:    optional
   ##   example: agents = ["udp://127.0.0.1:161"]
   ##            agents = ["tcp://127.0.0.1:161"]
-  ##            agents = ["udpv4://v4only-snmp-agent"]
+  ##            agents = ["udp4://v4only-snmp-agent"]
   agents = ["udp://127.0.0.1:161"]
 
   ## Timeout for each request.
--- a/plugins/inputs/snmp/README.md
+++ b/plugins/inputs/snmp/README.md
@@ -22,8 +22,11 @@ information.
 ```toml
 [[inputs.snmp]]
   ## Agent addresses to retrieve values from.
+  ##   format:  agents = ["<scheme>://<hostname>:<port>"]
+  ##   scheme:  either udp, udp4, udp6, tcp, tcp4, tcp6.  
+  ##            default is udp
   ##   example: agents = ["udp://127.0.0.1:161"]
-  ##            agents = ["tcp://127.0.0.1:161"]
+  ##            agents = ["udp4://v4only-snmp-agent"]
   agents = ["udp://127.0.0.1:161"]
 
   ## Timeout for each request.
--- a/plugins/inputs/snmp/snmp.go
+++ b/plugins/inputs/snmp/snmp.go
@@ -565,8 +565,8 @@ func (s *Snmp) getConnection(idx int) (snmpConnection, error) {
 	if err != nil {
 		return nil, err
 	}
-	
-  err = gs.SetAgent(agent)
+
+	err = gs.SetAgent(agent)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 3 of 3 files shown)


jaydizz:
!singed-cla

jaydizz:
!signed-cla

jaydizz:
!signed-cla

plugins/inputs/snmp/README.md
```diff
@@ -22,8 +22,11 @@ information.
 ```toml
 [[inputs.snmp]]
   ## Agent addresses to retrieve values from.
+  ##   format:  agents = ["<scheme>://<hostname>:<port>"]
+  ##   scheme:  either udp, udp4, udp6, tcp, tcp4, tcp6.  
+  ##            default is udp
   ##   example: agents = ["udp://127.0.0.1:161"]
-  ##            agents = ["tcp://127.0.0.1:161"]
```
Hipska, line 26: I would keep the `tcp` example.

plugins/inputs/snmp/README.md
```diff
@@ -22,8 +22,11 @@ information.
 ```toml
 [[inputs.snmp]]
   ## Agent addresses to retrieve values from.
+  ##   format:  agents = ["<scheme>://<hostname>:<port>"]
```
Hipska, line 25: I would mention that `scheme` and `port` are optional.

Hipska (changes requested):

Hipska:
Could you check your markup for the checkbox? Should be `[x]`.
Could you also add "Fixes #7879"?

plugins/inputs/snmp/README.md
jaydizz: I have re-added the example.

plugins/inputs/snmp/README.md
jaydizz: Done, thank you!

Hipska (approved):

internal/snmp/wrapper.go
```diff
@@ -167,6 +167,14 @@ func (gs *GosnmpWrapper) SetAgent(agent string) error {
...
 		gs.Transport = "tcp"
+	case "tcp4":
+		gs.Transport = "tcp4"
+	case "tcp6":
+		gs.Transport = "tcp6"
+	case "udp4":
+		gs.Transport = "udp4"
+	case "udp6":
+		gs.Transport = "udp6"
 	case "", "udp":
```
Hipska, line 178: I assume the empty string can be removed?
```suggestion
	case "udp":
```

jaydizz: Gosnmp will default to "udp" if no scheme is specified. 
However, in userconfig we can omit the "udp://" which will then cause the switch-case to fall through to the default-case. 

Regarding #7879:
The default case is reached when an invalid scheme is specified, but telegraf does not output the error nor stop execution. I will look into it further.

plugins/inputs/snmp/snmp.go
```diff
@@ -24,9 +24,15 @@ import (
 
 const description = `Retrieves SNMP values from remote agents`
 const sampleConfig = `
+  [[inputs.snmp]]
```
Hipska, line 27: I think this line addition is causing the test suite to fail..

Hipska (changes requested):

internal/snmp/wrapper.go
Hipska: Correct, but isn't that handled by this part? https://github.com/influxdata/telegraf/blob/3b8df55b9c8c15353b8778ab3a19f0b1a5021d29/internal/snmp/wrapper.go#L158-L160

jaydizz: yes, you right. 
I removed the line.

plugins/inputs/snmp/snmp.go
jaydizz: I removed the line.
Thank you for your patience!

internal/snmp/wrapper.go
```diff
@@ -167,7 +167,15 @@ func (gs *GosnmpWrapper) SetAgent(agent string) error {
...
+	case "tcp4":
+		gs.Transport = "tcp4"
+	case "tcp6":
+		gs.Transport = "tcp6"
+	case "udp4":
+		gs.Transport = "udp4"
+	case "udp6":
+		gs.Transport = "udp6"
+	case "udp":
 		gs.Transport = "udp"
```
Hipska, line 179: Maybe keep a more logical ordering?
```suggestion
	case "udp":
		gs.Transport = "udp"
	case "udp4":
		gs.Transport = "udp4"
	case "udp6":
		gs.Transport = "udp6"
```

Hipska (approved):

plugins/inputs/snmp/snmp.go
```diff
@@ -560,9 +565,8 @@ func (s *Snmp) getConnection(idx int) (snmpConnection, error) {
 	if err != nil {
 		return nil, err
 	}
-	gs.SetAgent(agent)
-	if err != nil {
-		return nil, err
+	if err := gs.SetAgent(agent); err != nil {
+		return nil, fmt.Errorf("setting gosnmpwrapper agent: %w", err)
 	}
```
Hipska, line 570: I think the idea was to behave it similar like the lines above, so it should have been `err = gs.SetAgent(agent)`

internal/snmp/wrapper.go
```diff
@@ -165,18 +165,8 @@ func (gs *GosnmpWrapper) SetAgent(agent string) error {
...
-		gs.Transport = "udp"
-	case "udp4":
-		gs.Transport = "udp4"
-	case "udp6":
-		gs.Transport = "udp6"
+	case "tcp", "tcp4", "tcp6", "udp", "udp4", "udp6":
+		gs.Transport = u.Scheme
 	default:
 		return fmt.Errorf("unsupported scheme: %v", u.Scheme)
 	}
```
Hipska: This got me thinking, why even bother checking it? This block could be replaced with `gs.Transport = u.Scheme` and gosnmp will handle it further when connecting.

jaydizz: Hi and thank you for you reviews! This is a pleasant experience! 

I thought the same thing but I am not sure of the side-effects that the remaining schemes like unix-sockets have. I am not sure if this plays well with windows or if it would play well with non-privileged users on Unix.
After all, the only schemes we do not support with the above code are ` "unix", "unixgram" and "unixpacket" `. 
Usually I prefer to just expose everything the library has to offer and let the user handle the aftermath, but in this case I am a little intimidated by the size of the project here. 
I think it would be better to limit the connection possibilities to the "well-known" ones. 

What do you think of that?

plugins/inputs/snmp/snmp.go
```diff
@@ -560,9 +565,8 @@ func (s *Snmp) getConnection(idx int) (snmpConnection, error) {
 	if err != nil {
 		return nil, err
 	}
-	gs.SetAgent(agent)
-	if err != nil {
-		return nil, err
+	if err := gs.SetAgent(agent); err != nil {
+		return nil, fmt.Errorf("setting gosnmpwrapper agent: %w", err)
```
Hipska, line 569:
```suggestion
	err = gs.SetAgent(agent)
 	if err != nil {
 		return nil, err
```

internal/snmp/wrapper.go
Hipska: My opinion is still to let gosnmp handle it. I'm calling @reimda in to hear his opinion.

Hipska:
Do you run `go test` before you commit? Anyway, last commit is not well formatted.

Hipska (changes requested):
Do you run `go test` before you commit? Anyway, last commit is not well formatted.

internal/snmp/wrapper.go
jaydizz: I did a little digging and I think I know what the intention behind the check in gosnmpwrapper was: 
gosnmp will happily take arbitrary values as connetion scheme and not complain about them. The first time the code can fail due to a wrong connection scheme is when `gs.Connect` in `plugins/input/snmp.go` is called.

However, the error raised by gosnmp is quite descriptive of what is going on:
```agent upd4://test: setting up connection: error establishing connection to host: dial upd4:test: unknown network upd```
Additionally, this is error is handeled here (line 576 ongoing):
```
if err := gs.Connect(); err != nil {
   return nil, fmt.Errorf("setting up connection: %w", err)
}
```
So indeed: There should be no problem in letting gosnmp handle the error and just feeding it with anything in`u.Scheme`.
Additionally this is more future-proof... maybe gosnmp will add some more schemes later on.

**Review by Hipska — APPROVED** [review]

File: internal/snmp/wrapper.go (hunk 7 above)
Comment (Hipska) [reply internal/snmp/wrapper.go]: Correct indeed, that's what I meant.

Comment (jaydizz) [reply internal/snmp/wrapper.go]: Currently I am looking into why the schemes `ip`, `ip4` and `ip6` always seem to fail and return UnknownNetworkError. 
I traced it down to `dial.go` in the go-sources line 175. 
```
167 func parseNetwork(ctx context.Context, network string, needsProto bool) (afnet string, proto int, err error) {
168     i := last(network, ':')
169     if i < 0 { // no colon
170         switch network {
171         case "tcp", "tcp4", "tcp6":
172         case "udp", "udp4", "udp6":
173         case "ip", "ip4", "ip6":
174             if needsProto {
175                 return "", 0, UnknownNetworkError(network)
176             }
177         case "unix", "unixgram", "unixpacket":
178         default:
179             return "", 0, UnknownNetworkError(network)
180         }
181         return network, 0, nil
182     }
```
Of course: As soon as you specify a port, you also need to specify the protocol! However, the error is not really clear on whats happening. 

Since we are using port 161 by default for snmp, it is clear that the aformentiones schemes will not work. And in fact, they don't really make any sense to me... I really don't know why go-snmp offers them in the first place. 
For the sockets as connection-schemes: 
A socket does not need a port. Unfortunately, specifying something like `unix://test:123` will also result in an `UnknownNetworkError`. I think this can be quite confusing for the user. 

Maybe, after all, it does make sense to limit the connection-schemes to udp{4,6} and tcp{4,6}. Or we have to implement checks in `snmp.go` to ensure correct formatting of the schemes. 
Maybe this is the reason why the connection-schemes were limited in the first place. 
Otherwise, mis-configuration may result in confusing error-messages for the user. 

Penny for your thoughts :)

Comment (Hipska) [reply internal/snmp/wrapper.go]: Okay, you convinced me :laughing: Fine for me to keep it as is.

Comment (jaydizz) [reply internal/snmp/wrapper.go]: I added some documentation, so the next guy is spared of this discovery-process :)
Thanks for your input. Greatly appreciated!

**Review by Hipska — APPROVED** [review]

File: plugins/inputs/snmp/README.md [hunk 9]
```diff
@@ -22,8 +22,13 @@ information.
 ```toml
 [[inputs.snmp]]
   ## Agent addresses to retrieve values from.
+  ##   format:  agents = ["<scheme://><hostname>:<port>"]
+  ##   scheme:  optional, either udp, udp4, udp6, tcp, tcp4, tcp6.  
+  ##            default is udp
+  ##   port:    optional
   ##   example: agents = ["udp://127.0.0.1:161"]
   ##            agents = ["tcp://127.0.0.1:161"]
+  ##            agents = ["udpv4://v4only-snmp-agent"]
```
reimda, line 31: Is there an extra "v" in the scheme in this example?
```suggestion
  ##            agents = ["udp4://v4only-snmp-agent"]
```

plugins/inputs/snmp/snmp.go
```diff
@@ -25,8 +25,13 @@ import (
 const description = `Retrieves SNMP values from remote agents`
 const sampleConfig = `
   ## Agent addresses to retrieve values from.
+  ##   format:  agents = ["<scheme://><hostname>:<port>"]
+  ##   scheme:  optional, either udp, udp4, udp6, tcp, tcp4, tcp6.            
+  ##            default is udp
+  ##   port:    optional
   ##   example: agents = ["udp://127.0.0.1:161"]
   ##            agents = ["tcp://127.0.0.1:161"]
+  ##            agents = ["udpv4://v4only-snmp-agent"]
```
reimda, line 34: extra v here too

reimda (changes requested):
Thanks for your work!

plugins/inputs/snmp/README.md
Hipska: Wow, good catch!

Hipska (approved):

reimda (approved):

plugins/inputs/snmp/snmp.go
jaydizz: Uhhh. 
How many Typos can there be in a few-line-change! 
Thanks for your input! :)

reimda: That's what reviews are for!

Hey I noticed you said you're just getting started with go.  If you're looking for more opportunities to write go code, feel free to grab issues with the "enhancement" or "good first issue" label.  The influx and community devs and I are happy to give you feedback on PRs.

Thanks again for this PR!

PR #8813, opened 2021-02-05
added member_id as tag(as it is a unique value) for redfish plugin and added address of the server when the status is other than 200 for better debugging

sarvanikonda:
### Required for all PRs:


sarvanikonda:
!signed-cla

ssoroka (approved):

PR #8817, opened 2021-02-05
outputs.kinesis - log record error count

JeffAshton:
Step 1 for [Issue #8818 - outputs.kinesis silently drops metrics](https://github.com/influxdata/telegraf/issues/8818)

### Required for all PRs:


plugins/outputs/kinesis/kinesis.go
```diff
@@ -154,26 +155,33 @@ func (k *KinesisOutput) SetSerializer(serializer serializers.Serializer) {
...
 	start := time.Now()
 	payload := &kinesis.PutRecordsInput{
 		Records:    r,
 		StreamName: aws.String(k.StreamName),
 	}
 
+	resp, err := k.svc.PutRecords(payload)
+	if err != nil {
+		log.Printf("E! kinesis: Unable to write to Kinesis : %s", err.Error())
+		return time.Since(start)
```
JeffAshton, line 173: If the service errored, then `resp` will not be set

plugins/outputs/kinesis/kinesis.go
```diff
@@ -154,26 +155,33 @@ func (k *KinesisOutput) SetSerializer(serializer serializers.Serializer) {
...
+	}
 
-	} else {
-		_, err := k.svc.PutRecords(payload)
-		if err != nil {
-			log.Printf("E! kinesis: Unable to write to Kinesis : %s", err.Error())
-		}
+	var failed = *resp.FailedRecordCount
+	if failed > 0 {
+		log.Printf("E! kinesis: Unable to write %+v of %+v record(s) to Kinesis", failed, len(r))
```
JeffAshton, line 182: Previously we have been silently dropping metrics

plugins/outputs/kinesis/kinesis_test.go
```diff
@@ -83,3 +89,219 @@ func TestPartitionKey(t *testing.T) {
...
+	m.requests = append(m.requests, input)
+
+	var resp = m.responses[reqNum]
+	return resp.Output, resp.Err
+}
+
+func (m *mockKinesisPutRecords) AssertRequests(
+	assert *assert.Assertions,
+	expected []*kinesis.PutRecordsInput,
+) {
```
JeffAshton, line 262: Not overly exciting assertions for this PR, but this will be helpful for testing batching behavior in future PRs.  The current test coverage is lacking, so if I can set a good precedent.  :)

plugins/outputs/kinesis/kinesis.go
```diff
@@ -29,7 +30,7 @@ type (
 		RandomPartitionKey bool       `toml:"use_random_partitionkey"`
 		Partition          *Partition `toml:"partition"`
 		Debug              bool       `toml:"debug"`
-		svc                *kinesis.Kinesis
+		svc                kinesisiface.KinesisAPI
```
helenosheaa, line 33: I saw within the docs for kinesisAPI there's a warning about this interface having breaking changes
when the service model is updated. Do we need this over the previous kinesis client struct?

helenosheaa (changes requested):
Change around logs and errors looks good. I've left a few comments with minor things and questions around the kinesis api mock.

plugins/outputs/kinesis/kinesis.go
JeffAshton: Yes, in order to mock the Kinesis responses.  If the there were breaking changes to the interface however, you would be required to update this code as well.  I'm only mock the one `PutRecords` method.

helenosheaa (approved):
Changes look good, thanks for doing that so quickly!

plugins/outputs/kinesis/kinesis.go
ssoroka: are you saying resp needs a nil check in addition to an err check?

plugins/outputs/kinesis/kinesis.go
ssoroka: would be great if someone updated this plugin to use 
```go
```
then the code could use `k.Log.Debug()` without conditionals, etc.
tests can set the logger with `testutil.Logger{}`

plugins/outputs/kinesis/kinesis.go
JeffAshton: No, I'm just giving insight into how the AWS APIs tend to work.

plugins/outputs/kinesis/kinesis.go
JeffAshton: I plan to do a few more PRs after this one.  Be glad to tackle that for you.

JeffAshton:
@helenosheaa, Are you waiting on anything from me for this PR?

helenosheaa:
@JeffAshton no looks good, happy to merge. Would be great if you could change this in your next pr to use telegraf.Logger instead :)

Issue #8808, opened 2021-02-04
Azure Active Directory (AAD) authentication support in SQL Server input plugin


avinash-nigam:
## Feature Request

Enable support for Azure Active Directory (AAD) authentication in the SQL Server input plugin

### Proposal:

- Azure SQL Database supports 2 main methods of authentication: [SQL authentication and AAD authentication](https://docs.microsoft.com/en-us/azure/azure-sql/database/security-overview#authentication). The recommended practice is to [use AAD authentication when possible](https://docs.microsoft.com/en-us/azure/azure-sql/database/authentication-aad-overview). AAD is a more modern authentication protocol, allows for easier credential/role management, and can eliminate the need to include passwords in a connection string
- To enable support for AAD authentication, we can leverage the existing AAD authentication support in the [SQL Server driver for Go](https://github.com/denisenkom/go-mssqldb#azure-active-directory-authentication---preview)

### Current behavior:

- The only authentication method currently supported by the SQL Server input plugin is SQL authentication

### Desired behavior:

- AAD authentication would also be supported by the SQL Server input

### Use case: 

- The recommended practice for Azure SQL Database is to [use AAD authentication when possible](https://docs.microsoft.com/en-us/azure/azure-sql/database/authentication-aad-overview). AAD is a more modern authentication protocol, allows for easier credential/role management, and can eliminate the need to include passwords in a connection string
- Many organizations using Azure SQL Database have security policies that require AAD authentication. Such organizations would be unable to use telegraf to monitor Azure SQL Database without support for AAD authentication

avinash-nigam:
@ssoroka @srebhan @Trovalo @denzilribeiro - Thoughts?
@masree @coquagli @bhsu-ms - FYI

srebhan:
Reviewed the PR. The feature makes totally sense to me.

PR #8822, opened 2021-02-05
SQL Server input plugin - Enable Azure Active Directory (AAD) authentication support 

avinash-nigam:
### Required for all PRs:

Associated to feature request - [Azure Active Directory (AAD) authentication support in SQL Server input plugin](https://github.com/influxdata/telegraf/issues/8808#issue-801695311)


plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -244,11 +247,36 @@ func (s *SQLServer) Gather(acc telegraf.Accumulator) error {
...
 
 func (s *SQLServer) gatherServer(server string, query Query, acc telegraf.Accumulator) error {
-	// deferred opening
-	conn, err := sql.Open("mssql", server)
-	if err != nil {
-		return err
+	// deferred connection opening
+	var conn *sql.DB
+
+	if strings.Contains(strings.ToLower(server), strings.ToLower("password=")) {
```
srebhan, line 253: I rather like to see a regular expression that makes sure that `password` is the beginning of the parameter and not e.g. part of `o_i_lost_my_password=damn` or `ignorepassword=true`. :-)

go.mod
```diff
@@ -12,6 +12,7 @@ require (
 	github.com/Azure/azure-storage-queue-go v0.0.0-20181215014128-6ed74e755687
 	github.com/Azure/go-ansiterm v0.0.0-20170929234023-d6e3b3328b78 // indirect
 	github.com/Azure/go-autorest/autorest v0.9.3
+	github.com/Azure/go-autorest/autorest/adal v0.8.1
```
srebhan, line 15: Please also add this to `docs/LICENSE_OF_DEPENDENCIES.md`. Currently it seems like our check-script is on strike, so watch the output with care. :-)

plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -244,11 +247,36 @@ func (s *SQLServer) Gather(acc telegraf.Accumulator) error {
...
+		var err error
+		conn, err = sql.Open("mssql", server)
+
+		if err != nil {
+			return err
+		}
+	} else {
+		// when password is not provided in connection string, use AAD auth with an MSI token
+
+		tokenProvider, err := getMSITokenProvider()
```
masree, line 265: Do we need to get the token for all the servers or can we get the token once and leverage it for all servers since t is from the same VM?

masree:
I think we should update SQL Server plugin doc to indicate AAD support using MSI

plugins/inputs/sqlserver/sqlserver.go
masree, line 265: I believe for all servers under a single VM, token will be same and we can re-use. Will fetching of token happen once per VM or is this once per server?  If it is latter, is it possible to fetch the token once and re-use for all servers in one VM?

plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -242,11 +245,36 @@ func (s *SQLServer) Gather(acc telegraf.Accumulator) error {
 }
 
 func (s *SQLServer) gatherServer(server string, query Query, acc telegraf.Accumulator) error {
-	// deferred opening
-	conn, err := sql.Open("mssql", server)
-	if err != nil {
-		return err
+	// deferred connection opening
```
masree, line 248: nit: this comment should go below when opening the connection

plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -242,11 +245,36 @@ func (s *SQLServer) Gather(acc telegraf.Accumulator) error {
...
+	var conn *sql.DB
+
+	if strings.Contains(strings.ToLower(server), strings.ToLower("password=")) {
+		// when password is provided in connection string, use SQL auth
+
```
masree, line 253: Can we log the database and server saying SQL Auth vs AAD auth? That would help in troubleshooting when something fails

plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -242,11 +245,36 @@ func (s *SQLServer) Gather(acc telegraf.Accumulator) error {
...
+		var err error
+		conn, err = sql.Open("mssql", server)
+
+		if err != nil {
+			return err
+		}
+	} else {
+		// when password is not provided in connection string, use AAD auth with an MSI token
+
```
masree, line 262: Add logging for AAD auth maybe?

plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -242,11 +245,36 @@ func (s *SQLServer) Gather(acc telegraf.Accumulator) error {
...
+		tokenProvider, err := getMSITokenProvider()
```
masree, line 263: For SQL VM, should we throw an exception and fail sooner here? SQL VM does not support AAD auth

plugins/inputs/sqlserver/sqlserver.go
avinash-nigam: Addressed the same using regular expression

avinash-nigam:
> I think we should update SQL Server plugin doc (readme) to indicate AAD support using MSI. Also indicate how an AAD login is created on SQL side using TSQL

Added content to readme file about its usage and instructions

plugins/inputs/sqlserver/sqlserver.go
avinash-nigam: fixed

plugins/inputs/sqlserver/sqlserver.go
avinash-nigam: Added a conditional check for DatabaseType = SQLServer i.e. SQL VM to error out if AAD auth is used

plugins/inputs/sqlserver/sqlserver.go
avinash-nigam: Yes, I think it'd be useful to have the Server/Database name and Auth mode details while troubleshooting. Once the above mentioned PR is merged, we can extend the metrics to capture this.

plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -275,6 +282,7 @@ func (s *SQLServer) gatherServer(server string, query Query, acc telegraf.Accumu
 		conn = sql.OpenDB(connector)
 	}
 
+	// deferred connection closure
```
srebhan, line 285: Please remove as it does not add any information.

plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -260,6 +262,11 @@ func (s *SQLServer) gatherServer(server string, query Query, acc telegraf.Accumu
 	} else {
 		// when password is not provided in connection string, use AAD auth with an MSI token
 
+		// return error if AAD Auth is used for DatabaseType SQLServer since AAD Auth is only supported for Azure SQL Database or Azure SQL Managed Instance
```
srebhan, line 265: Please shorten this to `AAD Auth is only supported for Azure SQL Database or Azure SQL Managed Instance`.

plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -245,10 +245,12 @@ func (s *SQLServer) Gather(acc telegraf.Accumulator) error {
 }
 
 func (s *SQLServer) gatherServer(server string, query Query, acc telegraf.Accumulator) error {
-	// deferred connection opening
+	// connection object
 	var conn *sql.DB
 
-	if strings.Contains(strings.ToLower(server), strings.ToLower("password=")) {
+	// setup connection based on authentication
+	rx := regexp.MustCompile("(?i)\\b(?:;(Password=\\w+))\\b")
```
srebhan, line 252: Not sure but is `dsn="Password=verysecret"` also a valid DSN (without the semicolon)?

go.mod
srebhan: Ok seems like it is covered by `github.com/Azure/go-autorest/autorest` already. Please mark as resolved.

srebhan (commented):
Please check my comments in the code. @masree any comments?

plugins/inputs/sqlserver/README.md
```diff
@@ -166,9 +166,34 @@ GO
...
+        DROP USER [<Monitoring_VM_Name>]
+    END')
+EXECUTE ('CREATE USER [<Monitoring_VM_Name>] FROM EXTERNAL PROVIDER')
+EXECUTE ('GRANT VIEW DATABASE STATE TO [<Monitoring_VM_Name>]')
+```
+- On the SQL Server resource of the database(s) being monitored, go to "Firewalls and Virtual Networks" tab and allowlist the monitoring VM IP address.
+- On the Monitoring VM, update the telegraf config file with the database connection string in the following format.
+```toml
+  servers = [
+    "Server=<Azure_SQL_Server_Name>.database.windows.net;Port=1433;Database=<Azure_SQL_Database_Name>;Authentication=Active Directory Integrated;app name=telegraf;log=1;",
```
masree, line 195: Can we explicitly mention is supported for SQL MI and SQL DB? Also call out it is not supported for SQL server on a VM

masree:
> Please check my comments in the code. @masree any comments?

Thank you for the review!

plugins/inputs/sqlserver/README.md
```diff
@@ -172,9 +172,34 @@ GO
...
+
+Azure SQL Database supports 2 main methods of authentication: [SQL authentication and AAD authentication](https://docs.microsoft.com/en-us/azure/azure-sql/database/security-overview#authentication). The recommended practice is to [use AAD authentication when possible](https://docs.microsoft.com/en-us/azure/azure-sql/database/authentication-aad-overview).
 
+AAD is a more modern authentication protocol, allows for easier credential/role management, and can eliminate the need to include passwords in a connection string.
 
+To enable support for AAD authentication, we leverage the existing AAD authentication support in the [SQL Server driver for Go](https://github.com/denisenkom/go-mssqldb#azure-active-directory-authentication---preview)
 
+#### How to use AAD Auth with MSI
+
+- Configure managed identity for Azure resources on the Monitoring VM (the VM that'd connect to the SQL server/database) [using the Azure portal](https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/qs-configure-portal-windows-vm).
```
masree, line 187: would it work with both user and system identity?

avinash-nigam: User identity won't work here, since we retrieve token based on managed identity.

plugins/inputs/sqlserver/sqlserver.go
avinash-nigam: done

plugins/inputs/sqlserver/sqlserver.go
avinash-nigam: done

plugins/inputs/sqlserver/sqlserver.go
avinash-nigam: Added in-memory token caching mechanism, it gets a token for the managed identity and use it until it finds it expired.

plugins/inputs/sqlserver/README.md
avinash-nigam: Updated

plugins/inputs/sqlserver/README.md
```diff
@@ -172,9 +172,34 @@ GO
...
+- On the database being monitored, create/update a USER with the name of the Monitoring VM as the principal using the below script. This might require allow-listing the client machine's IP address (from where the below SQL script is being run) on the SQL Server resource.
+```sql
+EXECUTE ('IF EXISTS(SELECT * FROM sys.database_principals WHERE name = ''<Monitoring_VM_Name>'')
```
masree, line 190: I would exclude the drop, and add only create user

plugins/inputs/sqlserver/sqlserver.go
avinash-nigam: I've updated regex condition used for AAD auth applicability where I look for Authentication parameter instead of Password. However with regards to your concern about well-formed dsn or connection string, the intention of this check if the connection string provides a way to identify auth mechanism and not if its well-formed. We could check that but it should be part of validating in the input config file instead of this check alone.

Just for clarity, the new regex I'm using now "(?i)(^|[;])(?:(\s*Authentication=Active Directory Integrated\s*))([;]|$)" would take care of scenarios where the connection string contains the "Authentication=Active Directory Integrated" string anywhere (at the start or at the end or in the middle somewhere).

plugins/inputs/sqlserver/README.md
avinash-nigam: The intention is to provide a "defensive" script, running it would not cause any harm. And the user always has the option to remove part of it if they deem it unnecessary.

plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -275,11 +286,41 @@ func (s *SQLServer) Gather(acc telegraf.Accumulator) error {
 }
 
 func (s *SQLServer) gatherServer(server string, query Query, acc telegraf.Accumulator) error {
-	// deferred opening
-	conn, err := sql.Open("mssql", server)
-	if err != nil {
-		return err
+	// connection object
+	var conn *sql.DB
```
masree, line 290: Heads up: Depending on which [PR ](https://github.com/influxdata/telegraf/pull/8596/files)gets merged to master, there  may be a possibility of merge conflict

avinash-nigam: thank you, yes I did notice that PR

plugins/inputs/sqlserver/README.md
masree: User Identity is also a managed identity- https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/how-to-manage-ua-identity-portal#:~:text=1%20Sign%20in%20to%20the%20Azure%20portal%20using,of%20a%20user-assigned%20managed%20identity%20click%20its%20name. Can you help check? If it does not work, documentation might have to be explicit saying system-assigned identity.

avinash-nigam: Yes, this is not meant for a user-assigned managed identity, I'll update the verbiage to call out that only a system-assigned managed identity should be created for this case.

plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -275,11 +286,41 @@ func (s *SQLServer) Gather(acc telegraf.Accumulator) error {
...
+
+	// setup connection based on authentication
+	rx := regexp.MustCompile(`(?i)(^|[;])(?:(\s*Authentication=Active Directory Integrated\s*))([;]|$)`)
+
+	// when auth mechanism is mentioned as AAD in the connection string, use managed-identity token
+	if rx.MatchString(server) {
```
masree, line 296: Is "Authentication=Active Directory Integration" property a mandatory field in connection string? If that property is missing, and password is not specified, connection would still work right?

avinash-nigam: Yes, for AAD auth the connection string needs to include "Authentication=Active Directory Integration" property (I'm referring to [this ](https://docs.microsoft.com/en-us/azure/azure-sql/database/authentication-aad-configure?tabs=azure-powershell#active-directory-integrated-authentication-1) doc). With the current code, if the property is not included it'll be processed as SQL Auth, and for your example, it won't connect since there are no credentials.

masree: I think that is AAD and different from MSI auth. Can you help remove the property Authentication from your connection string and see if you're able to connect in your test systems. My understanding is that this is not a required field and connection should succeed. Please help verify.

avinash-nigam: Thank you @masree for highlighting and clarifying this. Yes, MSI auth does not require the Authentication property in the connection string. I have reverted the code to use existence of "Password" as the definition of whether to use SQL Auth or AAD/MSI Auth. Readme file has been correspondingly modified as well.

masree:
How have you testing it? is there a testing plan you would like to link for us to take a look?
Example - 
Scenario1: Telegraf installed on VM and with connection strings using sqlauth only 
Scenario2: Telegraf installed on VM with connection strings using sqlauth and aadMSI (multiple connection strings) 
and so on

plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -408,6 +450,96 @@ func (s *SQLServer) Init() error {
 	return nil
 }
 
+// Get Token Provider by loading cached token or refreshed token
+func getTokenProvider() (func() (string, error), error) {
```
masree, line 454: Do you think it is better to move token related methods to a new class? That would help to keep this main class minimal.

plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -34,6 +36,12 @@ type Query struct {
 	OrderedColumns []string
 }
 
+// var to cache ADAL MSI token
+var adalToken *adal.Token
+
+// var for mutual exclusion lock
+var muCacheLock sync.RWMutex
+
```
srebhan, line 44: Is there a reason why you don't move this inside the struct? I assume this is the same for all `servers`!?

plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -272,15 +280,48 @@ func (s *SQLServer) Start(acc telegraf.Accumulator) error {
...
+			if err != nil {
+				acc.AddError(err)
+				return err
+			}
+		} else {
+			// otherwise assume AAD Auth with system-assigned managed identity (MSI)
+
+			// AAD Auth is only supported for Azure SQL Database or Azure SQL Managed Instance
+			if s.DatabaseType == "SQLServer" {
+				return fmt.Errorf("Database connection failed : AAD auth is not supported for SQL VM i.e. DatabaseType=SQLServer")
```
srebhan, line 306: Error texts have to start lowercase.

plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -272,15 +280,48 @@ func (s *SQLServer) Start(acc telegraf.Accumulator) error {
...
+			}
+
+			// get token from im-memory cache variable or from Azure Active Directory
+			tokenProvider, err := getTokenProvider()
+			if err != nil {
+				fmt.Printf("Error creating AAD token provider for system assigned Azure managed identity : %s", err.Error())
```
srebhan, line 312: Never `fmt.Printf`! Add a logger facility 
```
```
to the `SQLServer` struct and then use `s.Log.Warn()` or whatever to log the message. Alternatively you can just return the message as error message!?

plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -272,15 +280,48 @@ func (s *SQLServer) Start(acc telegraf.Accumulator) error {
...
+				acc.AddError(err)
+				return err
+			}
+
+			connector, err := mssql.NewAccessTokenConnector(serv, tokenProvider)
+			if err != nil {
+				fmt.Printf("Error creating the SQL connector : %s", err.Error())
```
srebhan, line 319: See above.

plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -423,6 +464,96 @@ func (s *SQLServer) Init() error {
...
+			}
+
+			// use the refreshed token
+			tokenString = spt.OAuthToken()
+		} else {
+			// use locally cached token
+			tokenString = token.OAuthToken()
+		}
+
+		muCacheLock.Unlock()
```
srebhan, line 499: Oh boy, if you return in between you will not release the lock!!!!!!!!!!!!!
Please use
```golang
muCacheLock.Lock()
defer muCacheLock.Unlock()
```
as a robust pattern wherever possible!

plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -423,6 +464,96 @@ func (s *SQLServer) Init() error {
...
+}
+
+// Load token from in-mem cache
+func loadToken() (*adal.Token, error) {
+	if adalToken == nil {
+		return nil, fmt.Errorf("Token is nil or Failed to load existing token")
+	}
+
+	return adalToken, nil
+}
```
srebhan, line 518: I don't get the intention behind this function... You could check for the token being `nil` on the caller side or am I missing something...

plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -423,6 +464,96 @@ func (s *SQLServer) Init() error {
...
+
+// Refresh token for the resource, and save to in-mem cache
+func refreshToken() (*adal.Token, error) {
+	// resource id for Azure SQL Database
+	const sqlAzureResourceID = "https://database.windows.net/"
```
srebhan, line 523: Should this be really fixed? And if so, is there a reason to put it here and not side-by-side with the other const declarations?

plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -423,6 +464,96 @@ func (s *SQLServer) Init() error {
...
+	spt, err := adal.NewServicePrincipalTokenFromMSI(msiEndpoint, sqlAzureResourceID)
+	if err != nil {
+		return nil, err
+	}
+
+	// ensure token is fresh
+	err = spt.EnsureFresh()
+	if err != nil {
+		return nil, err
+	}
```
srebhan, line 541: You might want to use
```suggestion
	if err := spt.EnsureFresh(); err != nil {
		return nil, err
	}
```
here.

srebhan (changes requested):
I have some comments in the code. The most sever is the potential deadlock when an error occurs in `getTokenProvider()`. Please take a look and fix the leaking lock.

plugins/inputs/sqlserver/sqlserver.go
avinash-nigam: When I initially wrote this I did have a reason, however now I don't think it makes sense to keep it out of the struct. Moved them to the struct.

plugins/inputs/sqlserver/sqlserver.go
avinash-nigam: done

plugins/inputs/sqlserver/sqlserver.go
avinash-nigam: Yes, I noticed it after merging the latest changes (for connection pooling). I was already adding the error to accumulator and this printf is not required, refactored it.

plugins/inputs/sqlserver/sqlserver.go
avinash-nigam: refactored it

plugins/inputs/sqlserver/sqlserver.go
avinash-nigam: Thank you for catching this, fixed it

plugins/inputs/sqlserver/sqlserver.go
avinash-nigam: Yes, it a fixed resource url, moved it with other const declarations

avinash-nigam:
> I have some comments in the code. The most sever is the potential deadlock when an error occurs in `getTokenProvider()`. Please take a look and fix the leaking lock.

Thank you @srebhan for the review, I've addressed your comments.

plugins/inputs/sqlserver/sqlserver.go
avinash-nigam: This was added while looking for different options of caching the token (for example, disk based), so would be useful if we use a different caching mechanism. I'd prefer to keep it structured this way, unless you have a strong opinion on it.

srebhan: Fine with me, maybe you could add a comment about this reasoning to the code?

srebhan: Btw: Error messages should start lowercase... ;-P

srebhan (commented):
Only some small (cosmetic) things left. Looks good so far.

plugins/inputs/sqlserver/sqlserver.go
avinash-nigam: Added a comment to the method and the error message with lowercase was already fixed in the previous commit (above screen shot is "outdated" :-) )

plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -296,29 +296,29 @@ func (s *SQLServer) Start(acc telegraf.Accumulator) error {
...
+				continue
 			}
 		} else {
 			// otherwise assume AAD Auth with system-assigned managed identity (MSI)
 
 			// AAD Auth is only supported for Azure SQL Database or Azure SQL Managed Instance
 			if s.DatabaseType == "SQLServer" {
-				return fmt.Errorf("Database connection failed : AAD auth is not supported for SQL VM i.e. DatabaseType=SQLServer")
+				er := errors.New("database connection failed : AAD auth is not supported for SQL VM i.e. DatabaseType=SQLServer")
+				acc.AddError(er)
```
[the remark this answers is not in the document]
avinash-nigam: changed to err

avinash-nigam:
> Only some small (cosmetic) things left. Looks good so far.

Thank you @srebhan for quick review, I've addressed your comments.

masree (approved):

plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -423,6 +463,94 @@ func (s *SQLServer) Init() error {
 	return nil
 }
 
+// Get Token Provider by loading cached token or refreshed token
+func (s *SQLServer) getTokenProvider() (func() (string, error), error) {
```
masree, line 467: Can we move token related methods to a different class? It would keep the main class short and simple

avinash-nigam: I feel the methods are not reusable as such to warrant a refactor to another class. And after moving the cache variable and it's lock into the SQLServer struct, those methods are more bound and separating the methods to another class won't add value. I'd prefer to keep them here for now unless we change the caching mechanism to a more reusable structure.

srebhan (approved):
Looks good to me.

sspaink:
> How have you been testing it? is there a testing plan you would like to link for us to take a look?
> Example -
> Scenario1: Telegraf installed on VM and with connection strings using sqlauth only
> Scenario2: Telegraf installed on VM with connection strings using sqlauth and aadMSI (multiple connection strings)
> and so on

@avinash-nigam I was also wondering about how you tested this? There aren't any unit tests covering this new auth method, what are your thoughts on including tests?

avinash-nigam:
> > How have you been testing it? is there a testing plan you would like to link for us to take a look?
> > Example -
> > Scenario1: Telegraf installed on VM and with connection strings using sqlauth only
> > Scenario2: Telegraf installed on VM with connection strings using sqlauth and aadMSI (multiple connection strings)
> > and so on
> 
> @avinash-nigam I was also wondering about how you tested this? There aren't any unit tests covering this new auth method, what are your thoughts on including tests?

@sspaink - This has been manually tested for databases of following variants individually and a combination of multiple variants by deploying the bits on an Ubuntu VM and settings connection strings to use SqlAuth and AADAuth (both as supported cases and not supported cases) - 

Azure SQL Database - S1, S2 (AAD supported)
Azure SQL Elastic Pool Database - S0, S2 (AAD supported)
SQL VM - 2012, 2014, 2016, 2019 (AAD not supported)
SQL AG (HADR) - 2019 (AAD not supported)

Currently this plugin does not support integration testing, so there are no tests for authentication yet. Once we build a mechanism to write integration tests, then we would be able to add tests for the same.

cerilewis:
This change appears to break Windows Integrated authentication where no User ID or Password is given in the connection string. If I run telegraf version 1.19.0-rc0 with configuration that works with 1.18.3 I get the following error:
`2021-06-14T09:35:28Z E! [inputs.sqlserver] Error in plugin: database connection failed : AAD auth is not supported for SQL VM i.e. DatabaseType=SQLServer`

This appears to come from the check for Password attribute being used to determine whether to user SQL Server authentication or AAD.

srebhan:
@cerilewis thanks for reporting this problem! Can you please open an issue and link to this PR. I fear that we need to add an option to specify which auth-scheme to use as the magic here will likely not be able to detect it... :-(

cerilewis:
@srebhan I have created the issue  #9362 but either can't or don't know how to link this PR to it correctly.

srebhan:
You already did by mentioning this PR in the issue. :-) Thank you for reporting.

srebhan:
@arkapravasinha please do not post support questions to old PRs! Nobody will notice them besides me. ;-) Better use the forum or Slack!

This being said, how about reading the [docs ](https://github.com/influxdata/telegraf/blob/master/plugins/inputs/sqlserver/README.md) and use `auth_method = "aad"`?

PR #8828, opened 2021-02-07
Add MessagePack output data format

dialogbox:
### Required for all PRs:

In addition to the current output data formats, I added a new binary output data format `msgpack`.

### Purporse:

To define msgpack messages and add output data format to support it.

### Current behavior:

Only text based output is supported.

### Desired behavior:

New format `msgpack` generates MessagePack encoded metrics as output.

### Use case:

Even the text based formats are easy to use, these are not the most efficient to parse and process. MessagePack can be used to define language-neutral, platform-neutral binary data encoding and the result can be easily encoded, decoded with many built-in safety checks. MessagePack is an efficient binary serialization format. It lets you exchange data among multiple languages like JSON. But it's faster and smaller.

https://msgpack.org/index.html


dialogbox:
This is the new PR to keep working on the previous PR (https://github.com/influxdata/telegraf/pull/3427) since I deleted the original fork.

docs/DATA_FORMATS_OUTPUT.md
```diff
@@ -13,6 +13,7 @@ plugins.
 1. [ServiceNow Metrics](/plugins/serializers/nowmetric)
 1. [SplunkMetric](/plugins/serializers/splunkmetric)
 1. [Wavefront](/plugins/serializers/wavefront)
+1. [MessagePack](/plugins/serializers/msgpack)
```
srebhan, line 16: Would you please order alphabetically!? I think the file is organized like this.

docs/LICENSE_OF_DEPENDENCIES.md
```diff
@@ -126,6 +126,7 @@ following works:
 - github.com/opencontainers/go-digest [Apache License 2.0](https://github.com/opencontainers/go-digest/blob/master/LICENSE)
 - github.com/opencontainers/image-spec [Apache License 2.0](https://github.com/opencontainers/image-spec/blob/master/LICENSE)
 - github.com/openzipkin/zipkin-go-opentracing [MIT License](https://github.com/openzipkin/zipkin-go-opentracing/blob/master/LICENSE)
+- github.com/philhofer/fwd [MIT License](https://github.com/philhofer/fwd/blob/master/LICENSE.md)
```
srebhan, line 129: You only need to add direct dependencies (only the stuff you _added_ to go.mod).

plugins/serializers/msgpack/metric.go
```diff
@@ -0,0 +1,66 @@
+package msgpack
+
+import (
+	"encoding/binary"
+	"time"
+
+	"github.com/tinylib/msgp/msgp"
+)
+
+//go:generate msgp
```
srebhan, line 10: Should we leave this in here?

plugins/serializers/msgpack/metric.go
```diff
@@ -0,0 +1,66 @@
...
+	Name   string                 `msg:"name"`
+	Time   MessagePackTime        `msg:"time,extension"`
+	Tags   map[string]string      `msg:"tags"`
+	Fields map[string]interface{} `msg:"fields"`
+}
+
+// MessagePackTime implements the official timestamp extension type
+// https://github.com/msgpack/msgpack/blob/master/spec.md#timestamp-extension-type
+//
+// tinylib/msgp has been used their own custom extenstion type and the offcial extension
```
srebhan, line 24:
```suggestion
// tinylib/msgp has been using their own custom extension type and the official extension
```

plugins/serializers/msgpack/metric.go
```diff
@@ -0,0 +1,66 @@
...
+// is not available. (https://github.com/tinylib/msgp/issues/214)
+type MessagePackTime struct {
+	time time.Time
+}
+
+func init() {
+	msgp.RegisterExtension(-1, func() msgp.Extension { return new(MessagePackTime) })
+}
```
srebhan, line 32: Usually this function is the last function in the file, but I can also live with this location.

plugins/serializers/msgpack/metric.go
```diff
@@ -0,0 +1,66 @@
...
+	Time   MessagePackTime        `msg:"time,extension"`
+	Tags   map[string]string      `msg:"tags"`
+	Fields map[string]interface{} `msg:"fields"`
+}
+
+// MessagePackTime implements the official timestamp extension type
+// https://github.com/msgpack/msgpack/blob/master/spec.md#timestamp-extension-type
+//
+// tinylib/msgp has been used their own custom extenstion type and the offcial extension
+// is not available. (https://github.com/tinylib/msgp/issues/214)
```
srebhan, line 25: So wouldn't it be beneficial for portability if you use a usual type (e.g. nanosecond-timestamp as uint64) instead of a custom-built extension? I suppose in case of the extension the other end also needs to use the same extension!?

plugins/serializers/msgpack/metric.go
```diff
@@ -0,0 +1,66 @@
...
+func (*MessagePackTime) Len() int {
+	return 8
+}
+
+// MarshalBinaryTo implements the Extension interface
+func (t *MessagePackTime) MarshalBinaryTo(buf []byte) error {
+	sec := uint64(t.time.Unix())
+	nsec := uint64(t.time.Nanosecond())
+
+	data := nsec<<34 | sec
```
srebhan, line 49: How about [UnixNano](https://golang.org/pkg/time/#Time.UnixNano)?

plugins/serializers/msgpack/metric_test.go
```diff
@@ -0,0 +1,22 @@
...
+	t1 := MessagePackTime{time: time.Unix(sec, nsec)}
+
+	buf := make([]byte, t1.Len())
+	assert.NoError(t, t1.MarshalBinaryTo(buf))
+
+	t2 := new(MessagePackTime)
+	t2.UnmarshalBinary(buf)
+
+	assert.Equal(t, t1.time, t2.time)
+}
```
srebhan, line 22: Please add some more tests!

plugins/serializers/msgpack/msgpack.go
```diff
@@ -0,0 +1,51 @@
...
+// Serialize implements serializers.Serializer.Serialize
+// github.com/influxdata/telegraf/plugins/serializers/Serializer
+func (s *Serializer) Serialize(metric telegraf.Metric) ([]byte, error) {
+	buf, err := marshalMetric(nil, metric)
+
+	if err != nil {
+		return nil, err
+	}
+
+	return buf, nil
```
srebhan, line 34: You might even want to do
```suggestion
	return marshalMetric(nil, metric)
```
or even directly move the `marshalMetric()` function in there.

plugins/serializers/msgpack/msgpack.go
```diff
@@ -0,0 +1,51 @@
...
+// github.com/influxdata/telegraf/plugins/serializers/Serializer
+func (s *Serializer) SerializeBatch(metrics []telegraf.Metric) ([]byte, error) {
+	buf := make([]byte, 0)
+	for _, m := range metrics {
+		var err error
+		buf, err = marshalMetric(buf, m)
+
+		if err != nil {
+			return nil, err
+		}
```
srebhan, line 47: In case you have multiple metrics you would just die and take all the metrics with you. Instead I propose to print an error message and continue. So we only use the one metric that cannot be serialized instead of losing all...

plugins/serializers/msgpack/msgpack_test.go
```diff
@@ -0,0 +1,237 @@
...
+	buf, err = s.Serialize(m)
+	assert.NoError(t, err)
+
+	m2 := &Metric{}
+	left, err := m2.UnmarshalMsg(buf)
+	assert.NoError(t, err)
+
+	assert.Equal(t, len(left), 0)
+
+	assert.Equal(t, true, isIdentical(t, m, m2))
```
srebhan, line 74: You can use `testutils.RequireMetricEqual()` or `RequireMetricsEqual()` in case you want to compare two or more metrics.

plugins/serializers/msgpack/msgpack_test.go
```diff
@@ -0,0 +1,237 @@
...
+
+	left := buf
+	for len(left) > 0 {
+		decodeM := &Metric{}
+		left, err = decodeM.UnmarshalMsg(left)
+
+		assert.NoError(t, err)
+		assert.Equal(t, true, isIdentical(t, m, decodeM))
+	}
+}
```
srebhan, line 237: Could you please also add one test with a time set to something not equal zero!?

plugins/serializers/registry.go
```diff
@@ -2,14 +2,16 @@ package serializers
 
 import (
 	"fmt"
-	"github.com/influxdata/telegraf/plugins/serializers/prometheusremotewrite"
 	"time"
 
+	"github.com/influxdata/telegraf/plugins/serializers/prometheusremotewrite"
```
srebhan, line 7: If you move this, then please insert in the list below in alphabetical order.

srebhan (commented):
Very nice code! I have some comments/questions in the code. For testing please take a look at telegraf's `testuils` for comparing metrics and add one test for a metric with the time field set.

docs/DATA_FORMATS_OUTPUT.md
dialogbox: Ack. I didn't realize it because of the first line.

docs/LICENSE_OF_DEPENDENCIES.md
dialogbox: Ack. I way trying to fix the CI error.

plugins/serializers/msgpack/metric.go
dialogbox: I'm not sure. What is the general location for generator directives?

If you are asking why we need this line, it is required for the code generator. https://github.com/tinylib/msgp/wiki/Using-the-Code-Generator

plugins/serializers/msgpack/metric.go
dialogbox: Ack.

plugins/serializers/msgpack/metric.go
dialogbox: It is an official extension which is defined in the MessagePack spec. Because this was added later, the tinylib/msgp hasn't implemented it yet. But the other libraries can understand the extension.

Just FYI, negative extension ids are reserved for the predefined extensions.

https://github.com/msgpack/msgpack/blob/master/spec.md#extension-types

plugins/serializers/msgpack/msgpack.go
dialogbox: Ack.

plugins/serializers/msgpack/msgpack.go
dialogbox: Do you know which way is the protocol? I can not find the protocol from the interface code. 

https://github.com/influxdata/telegraf/blob/master/plugins/serializers/registry.go#L40-L43

plugins/serializers/msgpack/msgpack_test.go
dialogbox: Ack.

m2 here is not a telegraf.Metric but writing a simple converter is cleaner. I will change. And I also found the testutil.TestMetric function. I will update the code with it.

plugins/serializers/registry.go
dialogbox: Ack.

plugins/serializers/msgpack/msgpack_test.go
dialogbox: Could you please explain the request a little bit?

plugins/serializers/msgpack/msgpack.go
dialogbox: I think this is undefined. I created an issue regarding this. https://github.com/influxdata/telegraf/issues/8830

I'm fine with either way or we can add a config flag like `batch_serialize_ignore_errors` later. How does it sound to you?

docs/LICENSE_OF_DEPENDENCIES.md
dialogbox: Do you know why the `make check-deps` is complaining about this dependency?

```
./scripts/check-deps.sh
--- /tmp/tmp.f2BWDwwhyY/LICENSE_OF_DEPENDENCIES.md	2021-02-07 14:23:02.275902607 +0000
+++ /tmp/tmp.f2BWDwwhyY/HEAD	2021-02-07 14:23:02.255902512 +0000
@@ -123,0 +124 @@
+github.com/philhofer/fwd
@@ -144,0 +146 @@
+github.com/tinylib/msgp


The docs/LICENSE_OF_DEPENDENCIES.md file does not contain the expected entries.
```

plugins/serializers/msgpack/metric.go
dialogbox: The timestamp extension spec uses 34 bits for seconds and 30 bits for nanoseconds so we need the both separately.

And the `UnixNano` is made by adding the both. `(t.unixSec())*1e9 + int64(t.nsec())`

https://golang.org/src/time/time.go#L1137

So, IMO just using `Unix()` and `Nanosecond()` is better for our purpose.

docs/LICENSE_OF_DEPENDENCIES.md
srebhan: I already triggered InfluxData people. Don't expect this is a problem of your code. :-) Will drop you a note as soon as it is fixed.

plugins/serializers/msgpack/metric.go
srebhan: It's not about the location, but as you check in the generated code it's not required I think. However, maybe it is a good idea to keep it for being able to reproduce the generated artifacts. Please keep it and mark as resolved.

plugins/serializers/msgpack/metric.go
srebhan: Understood. So please keep it as is and mark this conversation as resolved.

plugins/serializers/msgpack/metric.go
srebhan: Understood. However, I feel better if you mask the second's additional bits before bitwise or... :-)
```suggestion
	sec := uint64(t.time.Unix())
	nsec := uint64(t.time.Nanosecond())

	data := nsec<<34 | (sec & 0x3FFFFFFF)
```

plugins/serializers/msgpack/msgpack.go
srebhan: No please don't add a config flag. I stumbled upon this problem earlier and multiple times... The question is a bit, do you expect metrics to fail while others won't _or_ is it unlikely and in case of failure something is fundamentally broken. If the latter, please keep the code as is. Otherwise change to logging + `continue`.

plugins/serializers/msgpack/msgpack_test.go
srebhan: Oh man I missed the `now` in the end... :-S So please forget about the comment above. :-/ Anyway, sometimes we have problems with different time resolutions across OSes, so it might be easier for testing when using a fixed time. You decide.

plugins/serializers/msgpack/metric.go
dialogbox: I think this location is good since this is directly related to the `MessagePackTime` struct. We can move it to the end of the file later if anything else is added to this `init` function.

docs/LICENSE_OF_DEPENDENCIES.md
srebhan: Now I see. As I stated with my comment above you should not add `github.com/philhofer/fwd` but only `github.com/tinylib/msgp`. Can you please change this!?

plugins/serializers/msgpack/metric.go
dialogbox: How about this? The supported max date is 2514-05-30.

```
	if sec & 0xFFFF_FFFC_0000_0000 != 0 {
		return fmt.Errorf("Time is out of supported range: %v", t)
	}

	data := nsec<<34 | (sec & 0x03_FFFF_FFFF)
```

srebhan: Perfect.

plugins/serializers/msgpack/msgpack.go
dialogbox: If we add the guard for the supported date range, it could fail partially. Otherwise, I don't have any example yet.

plugins/serializers/msgpack/msgpack_test.go
dialogbox: I already changed it to use `testutil.TestMetric` so I assume I can resolve this.

plugins/serializers/msgpack/msgpack.go
srebhan: Ok then leave it as you have it now because if the date really passes this limit more metrics will likely fail. In this case there is no point in doing the extra work.
Please mark as resolved.

srebhan:
So this leaves us with masking/checking time and the removal of the `github.com/philhofer/fwd` dependency...

docs/LICENSE_OF_DEPENDENCIES.md
dialogbox: Ack

srebhan:
To be honest I don't get the license stuff. Now it complains about missing `github.com/philhofer/fwd`. :-/ Can you please try to add it again!? Otherwise the code is perfect.

dialogbox:
@srebhan Done.

dialogbox:
Uhm.. I only tested my code but it looks like the generated code enforces me to accept the default value of each type. Let me try to fix this.

dialogbox:
The new timestamp implementation is pretty complicated now. I feel we need more test cases to be sure it can work with all the possible cases. I will try on it.

plugins/serializers/msgpack/metric.go
```diff
@@ -0,0 +1,101 @@
...
+	sec := t.time.Unix()
+	nsec := t.time.Nanosecond()
+
+	if sec < 0 || sec > 0x400000000 { // 96 bits encoding
+		return 12
+	} else if sec > 0xFFFFFFFF || nsec != 0 {
+		return 8
+	} else {
+		return 4
+	}
```
srebhan, line 50: Could you please use
```suggestion
	if sec < 0 || sec > (1 << 34) { // 96 bits encoding
		return 12
	} else if sec >= (1 << 32) || nsec != 0 {
		return 8
	} else {
		return 4
	}
```
so it's easier to understand what the bit-masks mean.

plugins/serializers/msgpack/metric.go
```diff
@@ -0,0 +1,101 @@
...
+	len := len(buf)
+
+	if len == 4 {
+		sec := binary.BigEndian.Uint32(buf)
+		t.time = time.Unix(int64(sec), 0)
+	} else if len == 8 {
+		data := binary.BigEndian.Uint64(buf)
+
+		nsec := (data & 0xfffffffc_00000000) >> 34
+		sec := (data & 0x00000003_ffffffff)
```
srebhan, line 90: Can you please use uppercase _or_ lowercase for the hex-part. I like the style used in the `len==8` case (i.e. `0x03_FFFF_FFFF`).

srebhan (commented):
The code looks good,. I have some readability comments, but nothing big. Regarding the test-cases, maybe you can add one test-case where you serialize metrics and increase the timestamp between the supported range with a fixed step (e.g. one year) and test the result? Do you think it's possible?

dialogbox:
@srebhan Oh. I found your comment just after I pushed the changes. I added the test case and I found one bug actually. Although it was not critical since only one time point has issue (2514-05-30T01:53:03).

plugins/serializers/msgpack/metric.go
dialogbox: I actually already changed it to all lower cases since all the others uses lower cases.

plugins/serializers/msgpack/metric.go
dialogbox: Ack.

README.md
```diff
@@ -370,6 +370,7 @@ For documentation on the latest development code see the [documentation index][d
 - [SplunkMetric](/plugins/serializers/splunkmetric)
 - [Carbon2](/plugins/serializers/carbon2)
 - [Wavefront](/plugins/serializers/wavefront)
+- [MessagePack](/plugins/serializers/msgpack)
```
srebhan, line 373: Can you please sort alphabetically here!?

go.mod
```diff
@@ -142,12 +143,13 @@ require (
 	go.starlark.net v0.0.0-20200901195727-6e684ef5eeee
 	golang.org/x/exp v0.0.0-20200224162631-6cc2880d07d6 // indirect
 	golang.org/x/lint v0.0.0-20200302205851-738671d3881b // indirect
-	golang.org/x/net v0.0.0-20200904194848-62affa334b73
+	golang.org/x/mod v0.4.1 // indirect
```
srebhan, line 146: Where does this come from? If it's needed please keep, but just want to make sure we don't add random stuff by accident. :-)

plugins/serializers/msgpack/metric.go
```diff
@@ -0,0 +1,106 @@
...
+	sec := t.time.Unix()
+	nsec := t.time.Nanosecond()
+
+	if sec < 0 || sec >= (1<<34) { // 96 bits encoding
+		return 12
+	} else if sec >= (1<<32) || nsec != 0 {
+		return 8
+	} else {
+		return 4
+	}
```
srebhan, line 55: Please make our future linter happy :-)
```suggestion
	}
	return 4
```

srebhan, line 55: Please make our future linter happy
```suggestion
	} 
	if sec >= (1<<32) || nsec != 0 {
		return 8
	}
	return 4
```
it doesn't like `else`... ;-)

plugins/serializers/msgpack/metric.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+// MarshalBinaryTo implements the Extension interface
+func (t *MessagePackTime) MarshalBinaryTo(buf []byte) error {
+	len := t.Len()
+
+	if len == 4 {
+		sec := t.time.Unix()
+		binary.BigEndian.PutUint32(buf, uint32(sec))
+
```
srebhan, line 65: Remove newline.

plugins/serializers/msgpack/metric.go
```diff
@@ -0,0 +1,106 @@
...
+	} else if len == 8 {
+		sec := t.time.Unix()
+		nsec := t.time.Nanosecond()
+
+		data := uint64(nsec)<<34 | (uint64(sec) & 0x03_ffff_ffff)
+		binary.BigEndian.PutUint64(buf, data)
+
```
srebhan, line 72: Remove newline.

plugins/serializers/msgpack/msgpack.go
```diff
@@ -0,0 +1,45 @@
+package msgpack
+
+import (
+	"github.com/influxdata/telegraf"
+)
+
+// Serializer encodes metrics in MessagePack format
+type Serializer struct {
+}
```
srebhan, line 9:
```suggestion
type Serializer struct {}
```

srebhan (commented):
(Almost) final review brought up some remaining cosmetic things but very minor. Please 
keep the alphabetical order in the top-level readme and cross-check the `golang.org/x/mod` addition to `go.mod`. It smells a bit like unintentionally so I thought to better ask. If this is required just leave it as is.

And I have to tell you it's a joy to read your code! It's just beautiful.

README.md
dialogbox: Ack.

go.mod
dialogbox: I'm not sure why but it is added by `go get -u github.com/tinylib/msgp`.

I tested it again in a blank project and it confirms it is added by it.

```
$ go get -u github.com/tinylib/msgp
go: github.com/tinylib/msgp upgrade => v1.1.5
go: golang.org/x/mod upgrade => v0.4.1
go: golang.org/x/xerrors upgrade => v0.0.0-20200804184101-5ec99f83aff1
go: golang.org/x/tools upgrade => v0.1.0
go: golang.org/x/sys upgrade => v0.0.0-20210124154548-22da62e12c0c
```

And this is the result.

```
require (
        github.com/tinylib/msgp v1.1.5 // indirect
        golang.org/x/mod v0.4.1 // indirect
        golang.org/x/sys v0.0.0-20210124154548-22da62e12c0c // indirect
        golang.org/x/tools v0.1.0 // indirect
)
```

plugins/serializers/msgpack/metric.go
dialogbox: Ack

plugins/serializers/msgpack/metric.go
dialogbox: Ack

plugins/serializers/msgpack/metric.go
dialogbox: Ack

plugins/serializers/msgpack/msgpack.go
dialogbox: Ack

go.mod
dialogbox: Oh. I unintentionally removed this line in the last commit. But it works without any issue. Can we just safely remove this?

dialogbox:
@srebhan Wow. That's great. Thank you for the kind words. It means a lot to me because it was my first contribution to  a project in this scale.

README.md
srebhan: Thanks for also cleaning up the other plugins as well! :-)

go.mod
srebhan: If you remove it and it still works I guess so. :-) Just keep the setup without that and we will see. You can view this as resolved. Thanks for checking again!

plugins/serializers/msgpack/metric.go
```diff
@@ -0,0 +1,103 @@
...
+// 96bits: [-584554047284-02-23 16:59:44 UTC, 584554051223-11-09 07:00:16.000000000 UTC) range.
+func (t *MessagePackTime) Len() int {
+	sec := t.time.Unix()
+	nsec := t.time.Nanosecond()
+
+	if sec < 0 || sec >= (1<<34) { // 96 bits encoding
+		return 12
+	} else if sec >= (1<<32) || nsec != 0 {
+		return 8
+	}
```
srebhan, line 53: I think you missed this part :-)
```suggestion
	} 
	
	if sec >= (1<<32) || nsec != 0 {
		return 8
	}
```

srebhan:
@dialogbox it's well deserved! One last small thing and you unfortunately need to rebase due to conflicting go.sum...

plugins/serializers/msgpack/metric.go
dialogbox: Ah.. Did you mean removing all else? I think you did. I will update the others as well.

plugins/serializers/msgpack/metric.go
```diff
@@ -0,0 +1,108 @@
...
+
+		t.time = time.Unix(int64(sec), int64(nsec))
+	}
+	if len == 12 {
+		nsec := binary.BigEndian.Uint32(buf)
+		sec := binary.BigEndian.Uint64(buf[4:])
+
+		t.time = time.Unix(int64(sec), int64(nsec))
+	}
+
```
srebhan, line 106: Sorry, those should stay with the `else`. The instance in `Len` just `return`s in the `if` so the `else` has no effect... However in this code the `else` is beneficial and should stay.

Sorry for the confusion caused! :-(

dialogbox: No problem. So whether or not retuning inside the block was the point. Got it.

srebhan (approved):
This code is awesome! Sorry again for the confusion caused with the `else` block!

dialogbox:
@srebhan Great! Thank you for all of your effort.

go.mod
```diff
@@ -121,7 +121,7 @@ require (
 	github.com/safchain/ethtool v0.0.0-20200218184317-f459e2d13664
 	github.com/samuel/go-zookeeper v0.0.0-20180130194729-c4fab1ac1bec // indirect
 	github.com/satori/go.uuid v1.2.1-0.20181028125025-b2ce2384e17b // indirect
-	github.com/shirou/gopsutil v2.20.9+incompatible
+	github.com/shirou/gopsutil v3.21.1+incompatible
```
ssoroka, line 124: this might have impacts on other plugins, as I think it's pretty widely used. is this a requirement for this plugin or did it get updated accidentally?

ssoroka (changes requested):
Looks really good!

go.mod
ssoroka: looks like the versions within the package have separate paths, so might not be as big a deal to upgrade it, thought we might want to address it in a separate PR.

dialogbox: I had to do this because of this issue. https://github.com/shirou/gopsutil/issues/986 It looks like some recent change on golang.org/x/sys broke it.
I couldn't figure out any other solution.

dialogbox: I also tried to pin the version of the `golang.org/x/sys` but it is updated to the most recent version whenever I run the `make`.

Do you have any idea?

dialogbox: I think I fixed it. Probably I mistakenly modified something in `go.sum`. I checked out the `go.sum` from the upstream and tried it again and it manages to build with the `github.com/shirou/gopsutil v2.20.9+incompatible` and `golang.org/x/sys v0.0.0-20200930185726-fdedc70b468f`.

dialogbox: It is interesting because I didn't know the content of the `go.sum` can cause the change of `go.mod`. I though it is unidirectional not bidirectional.

srebhan (approved):

ssoroka (approved):

PR #8831, opened 2021-02-08
Fixing Code Quality Issues 

HarshCasper:
### Description
Hi 👋 

I ran the DeepSource static analyzer on the forked copy of this repo and found some interesting code quality [issues](https://deepsource.io/gh/HarshCasper/telegraf/). This PR fixes a few of them.

### Summary of changes

- Remove unnecessary wrapping of function call
- Simplify `switch` cases with type assertion
- Replace `strings.Index` with `strings.Contains` 
- Added `.deepsource.toml` file

### DIFF
--- a/plugins/inputs/postgresql_extensible/postgresql_extensible_test.go
+++ b/plugins/inputs/postgresql_extensible/postgresql_extensible_test.go
@@ -293,9 +293,9 @@ func (f fakeRow) Scan(dest ...interface{}) error {
 	}
 
 	for i, d := range dest {
-		switch d.(type) {
+		switch d := d.(type) {
 		case (*interface{}):
-			*d.(*interface{}) = f.fields[i]
+			*d = f.fields[i]
 		default:
 			return fmt.Errorf("Bad type %T", d)
 		}
--- a/plugins/outputs/riemann/riemann.go
+++ b/plugins/outputs/riemann/riemann.go
@@ -145,14 +145,14 @@ func (r *Riemann) buildRiemannEvents(m telegraf.Metric) []*raidman.Event {
 			Tags:       r.tags(m.Tags()),
 		}
 
-		switch value.(type) {
+		switch value := value.(type) {
 		case string:
 			// only send string metrics if explicitly enabled, skip otherwise
 			if !r.StringAsState {
 				log.Printf("D! Riemann event states disabled, skipping metric value [%s]\n", value)
 				continue
 			}
-			event.State = value.(string)
+			event.State = value
 		case int, int64, uint64, float32, float64:
 			event.Metric = value
 		default:
--- a/plugins/outputs/riemann_legacy/riemann.go
+++ b/plugins/outputs/riemann_legacy/riemann.go
@@ -110,9 +110,9 @@ func buildEvents(p telegraf.Metric, s string) []*raidman.Event {
 			Service: serviceName(s, p.Name(), p.Tags(), fieldName),
 		}
 
-		switch value.(type) {
+		switch value := value.(type) {
 		case string:
-			event.State = value.(string)
+			event.State = value
 		default:
 			event.Metric = value
 		}
--- a/.deepsource.toml
+++ b/.deepsource.toml
@@ -1,0 +1,10 @@
+version = 1
+
+test_patterns = ["telegraf/testutil/**"]
+
+[[analyzers]]
+name = "go"
+enabled = true
+
+  [analyzers.meta]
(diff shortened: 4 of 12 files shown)


HarshCasper:
!signed-cla

HarshCasper:
!signed-cla

HarshCasper:
!signed-cla

HarshCasper:
!signed-cla

plugins/inputs/postgresql_extensible/postgresql_extensible_test.go
```diff
@@ -293,9 +293,9 @@ func (f fakeRow) Scan(dest ...interface{}) error {
 	}
 
 	for i, d := range dest {
-		switch d.(type) {
+		switch d := d.(type) {
 		case (*interface{}):
-			*d.(*interface{}) = f.fields[i]
+			*d = f.fields[i]
```
srebhan, line 298: Shouldn't we use a simple if/else statement here and avoid the whole switch clutter?

plugins/outputs/riemann/riemann.go
```diff
@@ -145,14 +145,14 @@ func (r *Riemann) buildRiemannEvents(m telegraf.Metric) []*raidman.Event {
 			Tags:       r.tags(m.Tags()),
 		}
 
-		switch value.(type) {
+		switch value := value.(type) {
```
srebhan, line 148: Please avoid overriding the variable. Use 
```suggestion
		switch v := value.(type) {
```
or similar instead.

plugins/outputs/riemann_legacy/riemann.go
```diff
@@ -110,9 +110,9 @@ func buildEvents(p telegraf.Metric, s string) []*raidman.Event {
 			Service: serviceName(s, p.Name(), p.Tags(), fieldName),
 		}
 
-		switch value.(type) {
+		switch value := value.(type) {
 		case string:
-			event.State = value.(string)
+			event.State = value
```
srebhan, line 115: Please use an if statement instead
```golang
if v, ok := value.(string); ok {
    event.State = v
} else {
    event.Metric := value
}

**Review by srebhan — CHANGES_REQUESTED** [review]
Thanks for your patches! Those are much appreciated.

However, especially overriding/shadowing variable names in type switches are not so nice. Please use distinctive names for type-switch assignments and use if/else statements in case only a single type-assertion is required.

**Comment by srebhan** [comment]
@HarshCasper any news on this? Those fixes are really appreciated, so can you please spend some time on addressing my comments!?

**Comment by HarshCasper** [comment]
Hi @srebhan 

Apologies for missing this out in a flood of Github Notifications. I will readdress your comments by tomorrow and ping you back for a re-review. Thanks for the reminding up on this.

File: plugins/inputs/postgresql_extensible/postgresql_extensible_test.go (hunk 1 above)
Comment (HarshCasper) [reply plugins/inputs/postgresql_extensible/postgresql_extensible_test.go:298]: Hi @srebhan 

We definitely can! The DeepSource checks have flagged all the issues related to this [here](https://deepsource.io/gh/HarshCasper/telegraf/issue/CRT-A0014/occurrences)

For example, a Switch statement like this: 

```
switch x := x.(type) {
case int:
    body()
}
```

can be written as: 

```
if x, ok := x.(int); ok {
    body()
}
``` 

The only issue here would be their occurrence a lot of times in a codebase. So shall I clean all of them up in this PR or shall I raise a separate PR for all of this cleanup?

plugins/outputs/riemann_legacy/riemann.go
HarshCasper: Hi @srebhan 

Can you please check my comment above? If you want, I can carry out a project-wide refactor for this instead of fixing a few instances here and there. Please let me know what you think about this.

plugins/inputs/postgresql_extensible/postgresql_extensible_test.go
srebhan: I think you should clean them in one go. I'm on some other linter errors and it's a mess as the code base changes often and you will touch a lot of places. So it might make sense to split the PR, but it's not a must. Be prepared to rebase often! ;-) You have been warned. :-)

srebhan:
@HarshCasper sorry for the late response but I have been on vacation for a few days... Please see my comment above.

srebhan:
@HarshCasper any update?

srebhan:
@HarshCasper did you have a chance to work on these?

PR #8834, opened 2021-02-08
Input plugin for RavenDB

ml054:
### Required for all PRs:

### Metrics

- ravendb_server
  - tags:
    - url
    - node_tag
    - cluster_id
    - public_server_url (optional)  
  - fields:
    - backup_current_number_of_running_backups
    - backup_max_number_of_concurrent_backups
    - certificate_server_certificate_expiration_left_in_sec (optional)
    - certificate_well_known_admin_certificates (optional, separated by ';')
    - cluster_current_term
    - cluster_index      
    - cluster_node_state
      - 0 -> Passive
      - 1 -> Candidate
      - 2 -> Follower
      - 3 -> LeaderElect
      - 4 -> Leader
    - config_public_tcp_server_urls (optional, separated by ';')
    - config_server_urls
    - config_tcp_server_urls (optional, separated by ';')
    - cpu_assigned_processor_count
    - cpu_machine_usage
    - cpu_machine_io_wait (optional)
    - cpu_process_usage
    - cpu_processor_count
    - cpu_thread_pool_available_worker_threads
    - cpu_thread_pool_available_completion_port_threads
    - databases_loaded_count
    - databases_total_count
    - disk_remaining_storage_space_percentage
    - disk_system_store_used_data_file_size_in_mb
    - disk_system_store_total_data_file_size_in_mb
    - disk_total_free_space_in_mb
    - license_expiration_left_in_sec (optional)
    - license_max_cores
    - license_type
    - license_utilized_cpu_cores
    - memory_allocated_in_mb  
    - memory_installed_in_mb
    - memory_low_memory_severity
      - 0 -> None
      - 1 -> Low
      - 2 -> Extremely Low
    - memory_physical_in_mb
    - memory_total_dirty_in_mb
    - memory_total_swap_size_in_mb
    - memory_total_swap_usage_in_mb
    - memory_working_set_swap_usage_in_mb
    - network_concurrent_requests_count
    - network_last_authorized_non_cluster_admin_request_time_in_sec (optional)
    - network_last_request_time_in_sec (optional)
    - network_requests_per_sec
    - network_tcp_active_connections
    - network_total_requests
    - server_full_version
    - server_process_id
    - server_version
    - uptime_in_sec
  
- ravendb_databases
  - tags:
    - url
    - database_name
    - database_id
    - node_tag
    - public_server_url (optional)
  - fields:
    - counts_alerts
    - counts_attachments
    - counts_documents
    - counts_performance_hints
    - counts_rehabs
    - counts_replication_factor
    - counts_revisions
    - counts_unique_attachments
    - statistics_doc_puts_per_sec
    - statistics_map_index_indexes_per_sec
    - statistics_map_reduce_index_mapped_per_sec
    - statistics_map_reduce_index_reduced_per_sec
    - statistics_request_average_duration
    - statistics_requests_count
    - statistics_requests_per_sec
    - indexes_auto_count
    - indexes_count
    - indexes_disabled_count
    - indexes_errors_count
    - indexes_errored_count
    - indexes_idle_count
    - indexes_stale_count
    - indexes_static_count
    - storage_documents_allocated_data_file_in_mb
    - storage_documents_used_data_file_in_mb
    - storage_indexes_allocated_data_file_in_mb
    - storage_indexes_used_data_file_in_mb
    - storage_total_allocated_storage_file_in_mb
    - storage_total_free_space_in_mb
    - time_since_last_backup_in_sec (optional)
    - uptime_in_sec

- ravendb_indexes
  - tags: 
    - database_name
    - index_name
    - node_tag
    - public_server_url (optional)
    - url
  - fields
    - errors
    - is_invalid
    - lock_mode
      - Unlock
      - LockedIgnore
      - LockedError
    - mapped_per_sec
    - priority
      - Low
      - Normal
      - High
    - reduced_per_sec
    - state
      - Normal
      - Disabled
      - Idle
      - Error
    - status
      - Running
      - Paused
      - Disabled
    - time_since_last_indexing_in_sec (optional)
    - time_since_last_query_in_sec (optional)
    - type
      - None
      - AutoMap
      - AutoMapReduce
      - Map
      - MapReduce
      - Faulty
      - JavaScriptMap
      - JavaScriptMapReduce

- ravendb_collections
  - tags:
    - collection_name
    - database_name
    - node_tag
    - public_server_url (optional)
    - url
  - fields
    - documents_count
    - documents_size_in_bytes
    - revisions_size_in_bytes
    - tombstones_size_in_bytes
    - total_size_in_bytes


ml054:
CLA signed!

plugins/inputs/ravendb/README.md
```diff
@@ -0,0 +1,220 @@
...
+
+### Configuration
+
+The following is an example config for RavenDB. **Note:** The client certificate used should have `Operator` permissions on the cluster.
+
+```toml
+## Global tags useful to group other metrics with the RavenDB Node metrics
+[global_tags]
+ cluster = "ECommerce Stage" 
+
```
srebhan, line 15: Remove this part and keep only the RavenDB part.

plugins/inputs/ravendb/README.md
```diff
@@ -0,0 +1,220 @@
...
+    - node_tag
+    - public_server_url (optional)
+    - url
+  - fields
+    - documents_count
+    - documents_size_in_bytes
+    - revisions_size_in_bytes
+    - tombstones_size_in_bytes
+    - total_size_in_bytes
+
```
srebhan, line 216: It would be nice to show an example output in line-protocol format for this plugin. Only two lines or so for the user to get an idea.

plugins/inputs/ravendb/ravendb.go
```diff
@@ -0,0 +1,680 @@
+package ravendb
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/influxdata/telegraf/plugins/common/tls"
```
srebhan, line 6: Please move down to the influxdata group.

plugins/inputs/ravendb/ravendb.go
```diff
@@ -0,0 +1,680 @@
...
+	"github.com/influxdata/telegraf/plugins/inputs"
+)
+
+// defaultURL will set a default value that corresponds to the default value
+// used by RavenDB
+const DefaultURL = "http://localhost:8080"
+
+// Default http timeouts
+const DefaultResponseHeaderTimeout = 3
+const DefaultClientTimeout = 4
```
srebhan, line 24: Make these private variables by lower-case first letter.

plugins/inputs/ravendb/ravendb.go
```diff
@@ -0,0 +1,680 @@
...
+	ResponseHeaderTimeout internal.Duration `toml:"header_timeout"`
+	ClientTimeout         internal.Duration `toml:"client_timeout"`
+
+	GatherServerStats     bool     `toml:"gather_server_stats"`
+	GatherDbStats         bool     `toml:"gather_db_stats"`
+	GatherIndexStats      bool     `toml:"gather_index_stats"`
+	GatherCollectionStats bool     `toml:"gather_collection_stats"`
+	DbStatsDbs            []string `toml:"db_stats_dbs"`
+	IndexStatsDbs         []string `toml:"index_stats_dbs"`
+	CollectionStatsDbs    []string `toml:"collection_stats_dbs"`
```
srebhan, line 48: Please group the user-definable options. Move `Log telegraf.Logger` at the end of the user-definable options and add `toml:"-"` to prevent toml to try to set this.

plugins/inputs/ravendb/ravendb.go
```diff
@@ -0,0 +1,680 @@
...
+
+	Client *http.Client `toml:"-"`
```
srebhan, line 50: Make this a private variable `client`. No need to export it.

plugins/inputs/ravendb/ravendb.go
```diff
@@ -0,0 +1,680 @@
...
+	RevisionsSizeInBytes  int64  `json:"RevisionsSizeInBytes"`
+}
+
+var sampleConfig = `
+  ## Node URL and port that RavenDB is listening on.
+  url = "https://localhost:8080"
+
+  ## RavenDB X509 client certificate setup
+  tls_cert = "/etc/telegraf/raven.crt"
+  tls_key = "/etc/telegraf/raven.key"
```
srebhan, line 249: Are those mandatory or do you also support non-TLS connections?

plugins/inputs/ravendb/ravendb.go
```diff
@@ -0,0 +1,680 @@
...
+
+  ## When true, collect per database stats
+  # gather_db_stats = true
+
+  ## When true, collect per index stats
+  # gather_index_stats = true
+
+  ## When true, collect per collection stats
+  # gather_collection_stats = true
+
```
srebhan, line 272: Does it make sense to have a selection list instead like `stats_include=["server", "db", "index",...]`? This is much easier to extend... You might event add an exclude list and gather everything by default if `stats_include` is empty. Maybe take a look at `plugins/inputs/cloudwatch` as a good example.

plugins/inputs/ravendb/ravendb.go
```diff
@@ -0,0 +1,680 @@
...
+			gf(r, acc)
+		}(f)
+	}
+	wg.Wait()
+
+	return nil
+}
+
+func (r *RavenDB) ensureClient() error {
+	if r.Client == nil {
```
srebhan, line 320: Please invert the logic here and save some indention level below:
```suggestion
	if r.Client != nil {
		return nil
	}
```
_or_ move the if up one level.

plugins/inputs/ravendb/ravendb.go
```diff
@@ -0,0 +1,680 @@
...
+// Description ...
+func (r *RavenDB) Description() string {
+	return "Reads metrics from RavenDB servers via the Monitoring Endpoints"
+}
+
+// Gather ...
+func (r *RavenDB) Gather(acc telegraf.Accumulator) error {
+	err := r.ensureClient()
+	if nil != err {
+		if nil != r.Log {
```
srebhan, line 300: `r.Log` should never be `nil`. For tests set `<your instance>.Log = testutils.Logger{}` to avoid errors.

plugins/inputs/ravendb/ravendb.go
```diff
@@ -0,0 +1,680 @@
...
+			r.Log.Errorf("Error with Client %s", err)
+		}
+		return err
+	}
+
+	var wg sync.WaitGroup
+	wg.Add(len(gatherFunctions))
+	for _, f := range gatherFunctions {
+		go func(gf gatherFunc) {
```
srebhan, line 309: Change this to 
```suggestion
	for _, f := range gatherFunctions {
		wg.Add(1)
		go func(gf gatherFunc) {
```

plugins/inputs/ravendb/ravendb.go
```diff
@@ -0,0 +1,680 @@
...
+		}
+	}
+	return nil
+}
+
+func (r *RavenDB) requestJSON(u string, target interface{}) error {
+	if r.URL == "" {
+		r.URL = DefaultURL
+	}
+	u = fmt.Sprintf("%s%s", r.URL, u)
```
srebhan, line 342: Move this to an `Init()` function that is only called once. Check the input-plugin-example or interface for the signature. The `Init()` function is called once after parsing and setting the config parameters in the structure. This means you do this evaluation only once instead of doing it in every iteration cycle.

plugins/inputs/ravendb/ravendb.go
```diff
@@ -0,0 +1,680 @@
...
+	}
+
+	resp, err := r.Client.Do(req)
+	if err != nil {
+		return err
+	}
+
+	defer resp.Body.Close()
+
+	if nil != r.Log {
```
srebhan, line 356: Again is never `nil`.

plugins/inputs/ravendb/ravendb.go
```diff
@@ -0,0 +1,680 @@
...
+		r.Log.Debugf("%s: %s", u, resp.Status)
+	}
+	if resp.StatusCode >= 400 {
+		return fmt.Errorf("invalid response code to request '%s': %d - %s", r.URL, resp.StatusCode, resp.Status)
+	}
+
+	json.NewDecoder(resp.Body).Decode(target)
```
srebhan, line 363: `Decode()` might produce an error. Please check this code by
```suggestion
	return json.NewDecoder(resp.Body).Decode(target)
```

plugins/inputs/ravendb/ravendb.go
```diff
@@ -0,0 +1,680 @@
...
+
+	err := r.requestJSON("/admin/monitoring/v1/server", &serverResponse)
+	if err != nil {
+		acc.AddError(err)
+		return
+	}
+
+	tags := map[string]string{"url": r.URL}
+	tags["node_tag"] = serverResponse.Cluster.NodeTag
+	tags["cluster_id"] = serverResponse.Cluster.Id
```
srebhan, line 383: How about
```suggestion
	tags := map[string]string{
		"url": r.URL,
		"node_tag": serverResponse.Cluster.NodeTag,
		"cluster_id": serverResponse.Cluster.Id,
	}
```

plugins/inputs/ravendb/ravendb.go
```diff
@@ -0,0 +1,680 @@
...
+		"cpu_process_usage":                                 serverResponse.Cpu.ProcessUsage,
+		"cpu_machine_usage":                                 serverResponse.Cpu.MachineUsage,
+		"cpu_processor_count":                               serverResponse.Cpu.ProcessorCount,
+		"cpu_assigned_processor_count":                      serverResponse.Cpu.AssignedProcessorCount,
+		"cpu_thread_pool_available_worker_threads":          serverResponse.Cpu.ThreadPoolAvailableWorkerThreads,
+		"cpu_thread_pool_available_completion_port_threads": serverResponse.Cpu.ThreadPoolAvailableCompletionPortThreads,
+		"memory_allocated_in_mb":                            serverResponse.Memory.AllocatedMemoryInMb,
+		"memory_installed_in_mb":                            serverResponse.Memory.InstalledMemoryInMb,
+		"memory_physical_in_mb":                             serverResponse.Memory.PhysicalMemoryInMb,
+		"memory_low_memory_severity":                        mapMemorySeverity(serverResponse.Memory.LowMemorySeverity),
```
srebhan, line 406: Why not using the `string` as is? The string is much more readable than an arbitrary number.

plugins/inputs/ravendb/ravendb.go
```diff
@@ -0,0 +1,680 @@
...
+		"disk_total_free_space_in_mb":                       serverResponse.Disk.TotalFreeSpaceInMb,
+		"disk_remaining_storage_space_percentage":           serverResponse.Disk.RemainingStorageSpacePercentage,
+		"license_type":                                      serverResponse.License.Type,
+		"license_utilized_cpu_cores":                        serverResponse.License.UtilizedCpuCores,
+		"license_max_cores":                                 serverResponse.License.MaxCores,
+		"network_tcp_active_connections":                    serverResponse.Network.TcpActiveConnections,
+		"network_concurrent_requests_count":                 serverResponse.Network.ConcurrentRequestsCount,
+		"network_total_requests":                            serverResponse.Network.TotalRequests,
+		"network_requests_per_sec":                          serverResponse.Network.RequestsPerSec,
+		"cluster_node_state":                                mapNodeState(serverResponse.Cluster.NodeState),
```
srebhan, line 422: Same, why not set the field to the string value?

plugins/inputs/ravendb/ravendb.go
```diff
@@ -0,0 +1,680 @@
...
+		fields["network_last_authorized_non_cluster_admin_request_time_in_sec"] = serverResponse.Network.LastAuthorizedNonClusterAdminRequestTimeInSec
+	}
+
+	if serverResponse.Certificate.WellKnownAdminCertificates != nil {
+		fields["certificate_well_known_admin_certificates"] = strings.Join(serverResponse.Certificate.WellKnownAdminCertificates, ";")
+	}
+
+	if serverResponse.Certificate.ServerCertificateExpirationLeftInSec != nil {
+		fields["certificate_server_certificate_expiration_left_in_sec"] = serverResponse.Certificate.ServerCertificateExpirationLeftInSec
+	}
```
srebhan, line 459: You might just add the fields as `acc` will ignore fields with a `nil` value.

plugins/inputs/ravendb/ravendb.go
```diff
@@ -0,0 +1,680 @@
...
+	if err != nil {
+		acc.AddError(err)
+		return
+	}
+
+	for _, dbResponse := range databasesResponse.Results {
+		tags := map[string]string{"url": r.URL}
+		tags["database_name"] = dbResponse.DatabaseName
+		tags["database_id"] = dbResponse.DatabaseId
+		tags["node_tag"] = databasesResponse.NodeTag
```
srebhan, line 511: Fuse the initialization.

plugins/inputs/ravendb/ravendb.go
```diff
@@ -0,0 +1,680 @@
...
+			"storage_documents_used_data_file_in_mb":      dbResponse.Storage.DocumentsUsedDataFileInMb,
+			"storage_indexes_allocated_data_file_in_mb":   dbResponse.Storage.IndexesAllocatedDataFileInMb,
+			"storage_indexes_used_data_file_in_mb":        dbResponse.Storage.IndexesUsedDataFileInMb,
+			"storage_total_allocated_storage_file_in_mb":  dbResponse.Storage.TotalAllocatedStorageFileInMb,
+			"storage_total_free_space_in_mb":              dbResponse.Storage.TotalFreeSpaceInMb,
+		}
+
+		if dbResponse.TimeSinceLastBackupInSec != nil {
+			fields["time_since_last_backup_in_sec"] = dbResponse.TimeSinceLastBackupInSec
+		}
```
srebhan, line 551: Set field directly as `nil` values for fields will be ignored.

plugins/inputs/ravendb/ravendb.go
```diff
@@ -0,0 +1,680 @@
...
+	if err != nil {
+		acc.AddError(err)
+		return
+	}
+
+	for _, indexResponse := range indexesResponse.Results {
+		tags := map[string]string{"url": r.URL}
+		tags["database_name"] = indexResponse.DatabaseName
+		tags["index_name"] = indexResponse.IndexName
+		tags["node_tag"] = indexesResponse.NodeTag
```
srebhan, line 574: Fuse.

plugins/inputs/ravendb/ravendb.go
```diff
@@ -0,0 +1,680 @@
...
+			"type":            indexResponse.Type,
+		}
+
+		if indexResponse.TimeSinceLastQueryInSec != nil {
+			fields["time_since_last_query_in_sec"] = indexResponse.TimeSinceLastQueryInSec
+		}
+
+		if indexResponse.TimeSinceLastIndexingInSec != nil {
+			fields["time_since_last_indexing_in_sec"] = indexResponse.TimeSinceLastIndexingInSec
+		}
```
srebhan, line 598: Add directly.

plugins/inputs/ravendb/ravendb.go
```diff
@@ -0,0 +1,680 @@
...
+	if err != nil {
+		acc.AddError(err)
+		return
+	}
+
+	for _, collectionMetrics := range collectionsResponse.Results {
+		tags := map[string]string{"url": r.URL}
+		tags["node_tag"] = collectionsResponse.NodeTag
+		tags["database_name"] = collectionMetrics.DatabaseName
+		tags["collection_name"] = collectionMetrics.CollectionName
```
srebhan, line 621: Fuse.

plugins/inputs/ravendb/ravendb.go
```diff
@@ -0,0 +1,680 @@
...
+		if indexesResponse.PublicServerUrl != "" {
+			tags["public_server_url"] = indexesResponse.PublicServerUrl
+		}
+
+		fields := map[string]interface{}{
+			"priority":        indexResponse.Priority,
+			"state":           indexResponse.State,
+			"errors":          indexResponse.Errors,
+			"lock_mode":       indexResponse.LockMode,
+			"is_invalid":      boolToInt(indexResponse.IsInvalid),
```
srebhan, line 585: We do understand bool... :-) Please keep it instead of converting to int.

plugins/inputs/ravendb/ravendb.go
```diff
@@ -0,0 +1,680 @@
...
+
+		acc.AddFields("ravendb_collections", fields, tags)
+	}
+}
+
+func prepareDbNamesUrlPart(dbNames []string) string {
+	if len(dbNames) == 0 {
+		return ""
+	}
+
```
srebhan, line 643: This is the result anyway, so you can skip this test.

plugins/inputs/ravendb/ravendb.go
```diff
@@ -0,0 +1,680 @@
...
+	for _, db := range dbNames {
+		if first {
+			result += "?"
+			first = false
+		} else {
+			result += "&"
+		}
+		result += "name=" + url.QueryEscape(db)
+	}
```
srebhan, line 655: How about
```suggestion
	if len(dbNames) == 0 {
		return ""
	}

	result := "&"+dbNames[0]
	for _, db := range dbNames[1:] {
		result += "&name=" + url.QueryEscape(db)
	}
```

plugins/inputs/ravendb/ravendb.go
```diff
@@ -0,0 +1,680 @@
...
+		return &RavenDB{
+			ResponseHeaderTimeout: internal.Duration{Duration: DefaultResponseHeaderTimeout * time.Second},
+			ClientTimeout:         internal.Duration{Duration: DefaultClientTimeout * time.Second},
+			GatherServerStats:     true,
+			GatherDbStats:         true,
+			GatherIndexStats:      true,
+			GatherCollectionStats: true,
+			DbStatsDbs:            []string{},
+			IndexStatsDbs:         []string{},
+			CollectionStatsDbs:    []string{},
```
srebhan, line 677: This is the default anyway. Please remove.

plugins/inputs/ravendb/ravendb_test.go
```diff
@@ -0,0 +1,435 @@
...
+	}))
+	defer ts.Close()
+
+	r := &RavenDB{
+		URL:                   ts.URL,
+		GatherDbStats:         true,
+		GatherIndexStats:      true,
+		GatherServerStats:     true,
+		GatherCollectionStats: true,
+	}
```
srebhan, line 51:
```suggestion
		Log: testutil.Logger{},
	}
```

plugins/inputs/ravendb/ravendb_test.go
```diff
@@ -0,0 +1,435 @@
...
+
+	serverTags := map[string]string{
+		"url":               ts.URL,
+		"node_tag":          "A",
+		"cluster_id":        "6b535a18-558f-4e53-a479-a514efc16aab",
+		"public_server_url": "http://raven1:8080",
+	}
+
+	compareFields(t, serverFields, acc, "ravendb_server")
+	compareTags(t, serverTags, acc, "ravendb_server")
```
srebhan, line 114: No need to do this. Check `testutil.RequireMetricEqual()` and `testutil.RequireMetricsEqual()`! You can also ignore the time by setting the `IgnoreTime` option.

plugins/inputs/ravendb/testdata/collections_full.json
```diff
@@ -0,0 +1,15 @@
...
+      "DatabaseName": "db1",
+      "CollectionName": "Orders",
+      "DocumentsCount": 830,
+      "TotalSizeInBytes": 2744320,
+      "DocumentsSizeInBytes": 868352,
+      "TombstonesSizeInBytes": 122880,
+      "RevisionsSizeInBytes": 1753088
+    }
+  ]
+}
```
srebhan, line 15: Newline at end of file? Same for all other jsons..

srebhan (commented):
Nice code! I have some minor comments in the code. Please make use of `testutil` functions and maybe rethink converting the gathering options to a list rather than a set of booleans.

srebhan:
@ml054 you have to use `!signed-cla` as message for tiger-bot instead of "CLA signed!". It's a bot so no natural language... :-)

ml054:
!signed-cla

plugins/inputs/ravendb/README.md
ml054: done

plugins/inputs/ravendb/ravendb.go
ml054: done

plugins/inputs/ravendb/ravendb.go
ml054: done

plugins/inputs/ravendb/ravendb.go
ml054: I'm not sure I fully understand. I end up with this:
```
type RavenDB struct {
	URL  string `toml:"url"`
	Name string `toml:"name"`
	tls.ClientConfig

	ResponseHeaderTimeout internal.Duration `toml:"header_timeout"`
	ClientTimeout         internal.Duration `toml:"client_timeout"`

	GatherServerStats     bool     `toml:"gather_server_stats"`
	GatherDbStats         bool     `toml:"gather_db_stats"`
	GatherIndexStats      bool     `toml:"gather_index_stats"`
	GatherCollectionStats bool     `toml:"gather_collection_stats"`
	DbStatsDbs            []string `toml:"db_stats_dbs"`
	IndexStatsDbs         []string `toml:"index_stats_dbs"`
	CollectionStatsDbs    []string `toml:"collection_stats_dbs"`

	Client *http.Client `toml:"-"`
}
```

plugins/inputs/ravendb/ravendb.go
ml054: done

plugins/inputs/ravendb/ravendb.go
ml054: we support both secured and non-secured servers, so I prepend this with `#`.

plugins/inputs/ravendb/ravendb.go
ml054: done

plugins/inputs/ravendb/ravendb.go
ml054: done

plugins/inputs/ravendb/ravendb.go
ml054: done

plugins/inputs/ravendb/ravendb.go
ml054: done

plugins/inputs/ravendb/ravendb.go
ml054: done

plugins/inputs/ravendb/ravendb.go
ml054: done

plugins/inputs/ravendb/ravendb.go
ml054: I leave items which does strings.join inside if.

plugins/inputs/ravendb/ravendb.go
ml054: done

plugins/inputs/ravendb/ravendb.go
ml054: done

plugins/inputs/ravendb/ravendb.go
ml054: done

plugins/inputs/ravendb/ravendb.go
ml054: done

plugins/inputs/ravendb/ravendb.go
ml054: done

plugins/inputs/ravendb/ravendb_test.go
ml054: done

plugins/inputs/ravendb/ravendb.go
srebhan: Well almost. Usually the user-configurable options (everything with a `toml` annotation) are listed first, `Log` is the last in this group. Afterwards,  `tls.ClientConfig` and internal fields (e.g. `client`) follow. That's the usual way so it's easier for developers they feel home in this plugin. :-)

srebhan:
@ml054 if you feel ready, just push your changes and I will do a next round. You might even want to push some intermediate state, but please let me know about the stuff that you didn't address yet so I can skip those on review. :-)

ml054:
@srebhan sure I will let you know once I finish and push changes. Look like I even have an access to click on re-request review. :)

plugins/inputs/ravendb/ravendb.go
ml054: that's the reason: https://community.grafana.com/t/is-there-a-text-to-value-mapping/13420 or https://github.com/grafana/grafana/issues/7631

so we can plot this value over time.

plugins/inputs/ravendb/ravendb.go
ml054: ok updated

plugins/inputs/ravendb/ravendb.go
ml054: since we have 4 switches I think let's leave includes only - I removed boolean - I agree this is redundant.

ml054: ok I just recall why we have bool + list. If we don't want to collect given stats then we set boolean to false. Having empty list might be ambiguous (does it mean collect all items are none?). Please have a look at mongodb input plugin. I follow the same pattern.

plugins/inputs/ravendb/ravendb.go
ml054: see note above

plugins/inputs/ravendb/ravendb.go
ml054: done

plugins/inputs/ravendb/ravendb.go
ml054: removed

plugins/inputs/ravendb/ravendb.go
ml054: applied - thanks

plugins/inputs/ravendb/ravendb.go
ml054: ok

plugins/inputs/ravendb/testdata/collections_full.json
ml054: I don't have newlines at the end of json files and I assume it is correct, no ?

plugins/inputs/ravendb/ravendb_test.go
ml054: done - thanks for this method

plugins/inputs/ravendb/README.md
ml054: sure - added

plugins/inputs/ravendb/ravendb.go
ml054: I think I don't follow this url maybe different over time if you deal with multiple nodes.

plugins/inputs/ravendb/ravendb.go
ml054: anyway Grafana has issues with reading bools

plugins/inputs/ravendb/README.md
srebhan: Thx, please mark as resolved.

plugins/inputs/ravendb/ravendb_test.go
srebhan: Thx, please mark as resolved.

plugins/inputs/ravendb/ravendb_test.go
```diff
@@ -199,8 +192,23 @@ func TestRavenDBGeneratesMetricsFull(t *testing.T) {
...
+		case "ravendb_server":
+			testutil.RequireMetricEqual(t, serverExpected, metric, testutil.IgnoreTime())
+		case "ravendb_databases":
+			testutil.RequireMetricEqual(t, dbExpected, metric, testutil.IgnoreTime())
+		case "ravendb_indexes":
+			testutil.RequireMetricEqual(t, indexExpected, metric, testutil.IgnoreTime())
+		case "ravendb_collections":
+			testutil.RequireMetricEqual(t, collectionsExpected, metric, testutil.IgnoreTime())
+		}
+	}
```
srebhan, line 211: If I'm not mistaken, you can add the expected metrics to an `[]telegraf.Metric` array and then test all metrics at once with `testutil.RequireMetricsEqual()` (note the **s** in Metric**s**)!

plugins/inputs/ravendb/ravendb_test.go
```diff
@@ -373,63 +373,23 @@ func TestRavenDBGeneratesMetricsMin(t *testing.T) {
...
+	for _, metric := range acc.GetTelegrafMetrics() {
+		switch metric.Name() {
+		case "ravendb_server":
+			testutil.RequireMetricEqual(t, serverExpected, metric, testutil.IgnoreTime())
+		case "ravendb_databases":
+			testutil.RequireMetricEqual(t, dbExpected, metric, testutil.IgnoreTime())
+		case "ravendb_indexes":
+			testutil.RequireMetricEqual(t, indexExpected, metric, testutil.IgnoreTime())
+		case "ravendb_collections":
+			testutil.RequireMetricEqual(t, collectionsExpected, metric, testutil.IgnoreTime())
```
srebhan, line 392: See above.

plugins/inputs/ravendb/ravendb.go
```diff
@@ -378,52 +375,59 @@ func gatherServer(r *RavenDB, acc telegraf.Accumulator) {
...
+		"disk_total_free_space_in_mb":                                   serverResponse.Disk.TotalFreeSpaceInMb,
+		"disk_remaining_storage_space_percentage":                       serverResponse.Disk.RemainingStorageSpacePercentage,
+		"license_type":                                                  serverResponse.License.Type,
+		"license_utilized_cpu_cores":                                    serverResponse.License.UtilizedCpuCores,
+		"license_max_cores":                                             serverResponse.License.MaxCores,
+		"network_tcp_active_connections":                                serverResponse.Network.TcpActiveConnections,
+		"network_concurrent_requests_count":                             serverResponse.Network.ConcurrentRequestsCount,
+		"network_total_requests":                                        serverResponse.Network.TotalRequests,
+		"network_requests_per_sec":                                      serverResponse.Network.RequestsPerSec,
+		"cluster_node_state":                                            mapNodeState(serverResponse.Cluster.NodeState),
```
srebhan, line 421: Please keep the string here instead of mapping it to a cryptic number. A sting is much easier to discover during data-analysis especially by people that don't know this code.

plugins/inputs/ravendb/ravendb.go
```diff
@@ -378,52 +375,59 @@ func gatherServer(r *RavenDB, acc telegraf.Accumulator) {
...
+		"cpu_process_usage":                                             serverResponse.Cpu.ProcessUsage,
+		"cpu_machine_usage":                                             serverResponse.Cpu.MachineUsage,
+		"cpu_processor_count":                                           serverResponse.Cpu.ProcessorCount,
+		"cpu_assigned_processor_count":                                  serverResponse.Cpu.AssignedProcessorCount,
+		"cpu_thread_pool_available_worker_threads":                      serverResponse.Cpu.ThreadPoolAvailableWorkerThreads,
+		"cpu_thread_pool_available_completion_port_threads":             serverResponse.Cpu.ThreadPoolAvailableCompletionPortThreads,
+		"memory_allocated_in_mb":                                        serverResponse.Memory.AllocatedMemoryInMb,
+		"memory_installed_in_mb":                                        serverResponse.Memory.InstalledMemoryInMb,
+		"memory_physical_in_mb":                                         serverResponse.Memory.PhysicalMemoryInMb,
+		"memory_low_memory_severity":                                    mapMemorySeverity(serverResponse.Memory.LowMemorySeverity),
```
srebhan, line 405: Please keep the string here instead of mapping it to a cryptic number. A sting is much easier to discover during data-analysis especially by people that don't know this code.

plugins/inputs/ravendb/ravendb.go
```diff
@@ -378,52 +375,59 @@ func gatherServer(r *RavenDB, acc telegraf.Accumulator) {
...
+		"cluster_node_state":                                            mapNodeState(serverResponse.Cluster.NodeState),
+		"cluster_current_term":                                          serverResponse.Cluster.CurrentTerm,
+		"cluster_index":                                                 serverResponse.Cluster.Index,
+		"databases_total_count":                                         serverResponse.Databases.TotalCount,
+		"databases_loaded_count":                                        serverResponse.Databases.LoadedCount,
+		"cpu_machine_io_wait":                                           serverResponse.Cpu.MachineIoWait,
+		"license_expiration_left_in_sec":                                serverResponse.License.ExpirationLeftInSec,
+		"network_last_request_time_in_sec":                              serverResponse.Network.LastRequestTimeInSec,
+		"network_last_authorized_non_cluster_admin_request_time_in_sec": serverResponse.Network.LastAuthorizedNonClusterAdminRequestTimeInSec,
+		"certificate_server_certificate_expiration_left_in_sec":         serverResponse.Certificate.ServerCertificateExpirationLeftInSec,
```
srebhan, line 430: Would it make sense to sort this alphabetically in case someone is looking for a certain field?

plugins/inputs/ravendb/ravendb.go
```diff
@@ -0,0 +1,644 @@
...
+var gatherFunctions = []gatherFunc{gatherServer, gatherDatabases, gatherIndexes, gatherCollections}
+
+// RavenDB defines the configuration necessary for gathering metrics,
+// see the sample config for further details
+type RavenDB struct {
+	URL  string `toml:"url"`
+	Name string `toml:"name"`
+
+	ResponseHeaderTimeout internal.Duration `toml:"header_timeout"`
+	ClientTimeout         internal.Duration `toml:"client_timeout"`
```
srebhan, line 37: Do we really need this fine-grained timeout definition or can we just work with the `client_timeout`? If so, please rename the option to `timeout`.

plugins/inputs/ravendb/ravendb.go
```diff
@@ -0,0 +1,644 @@
...
+
+type collectionMetrics struct {
+	DatabaseName          string `json:"DatabaseName"`
+	CollectionName        string `json:"CollectionName"`
+	DocumentsCount        int64  `json:"DocumentsCount"`
+	TotalSizeInBytes      int64  `json:"TotalSizeInBytes"`
+	DocumentsSizeInBytes  int64  `json:"DocumentsSizeInBytes"`
+	TombstonesSizeInBytes int64  `json:"TombstonesSizeInBytes"`
+	RevisionsSizeInBytes  int64  `json:"RevisionsSizeInBytes"`
+}
```
srebhan, line 242: Would it make sense to move these to an own file within the same package in order to make this file easier to read?

plugins/inputs/ravendb/ravendb.go
srebhan: Well, lately we had a lot of troubles with other plugins that started to extend the options or make the selection more fine-grained. You can easily end-up with dozens of options in the worst-case. A list is easy to extend and you can define and document the empty behavior. Most plugins (like `cloudwatch`) define an `_include` and an `_exclude` list with empty include-list means _all_ included.

So if there is a chance (even a small one) that at some point you want to add/split/remove items here, please go with the list. There is even a built-in `filter` implementation that makes using such lists dead easy (including glob patterns).

plugins/inputs/ravendb/ravendb.go
```diff
@@ -0,0 +1,644 @@
...
+	case "LeaderElect":
+		return 3
+	case "Leader":
+		return 4
+	default:
+		return -1
+	}
+}
+
+func gatherDatabases(r *RavenDB, acc telegraf.Accumulator) {
```
srebhan, line 478: Please change to
```suggestion
func (r *RavenDB) gatherDatabases(acc telegraf.Accumulator) {
```
same for all other members of the `gather` family.

plugins/inputs/ravendb/ravendb.go
```diff
@@ -0,0 +1,644 @@
...
+}
+
+func gatherIndexes(r *RavenDB, acc telegraf.Accumulator) {
+	if !r.GatherIndexStats {
+		return
+	}
+
+	indexesResponse := &indexesMetricResponse{}
+
+	err := r.requestJSON("/admin/monitoring/v1/indexes"+prepareDbNamesUrlPart(r.IndexStatsDbs), &indexesResponse)
```
srebhan, line 548: Please prepare the URL in a `Init()` function. It really doesn't hurt, saves some cycles and eases reading.

srebhan (commented):
Wow nice progress! Please think _again_ about an `_include` list to define what is gathered. Most recent plugins use this pattern. There is also potential to further simplify the tests.

One thing I really want to urge you to change is the mapping of the string-fields. As someone working with data I often stumble upon those _magic_ numbers. I always curse the author of such mappings as usually you have to dive down to the data-collection code to find out what the mapping means... not nice! ;-)

plugins/inputs/ravendb/testdata/collections_full.json
srebhan: Well it's a bit cumbersome if you for example just `cat` them and your prompt doesn't move on. Nothing big, but it would be nice to add that extra newline. :-)

plugins/inputs/ravendb/ravendb_test.go
srebhan: Thx, please mark as resolved.

plugins/inputs/ravendb/ravendb.go
srebhan: Thx, please mark as resolved.

plugins/inputs/ravendb/ravendb.go
srebhan: Thx, please mark as resolved.

plugins/inputs/ravendb/ravendb.go
srebhan: Thx, please mark as resolved.

plugins/inputs/ravendb/ravendb.go
srebhan: Well you can use a processor for your use-case to do the mapping you need, so we can have the best of two worlds.

plugins/inputs/ravendb/ravendb.go
srebhan: Thx, please mark as resolved.

plugins/inputs/ravendb/ravendb.go
srebhan: Thx, please mark as resolved.

plugins/inputs/ravendb/ravendb.go
srebhan: Thx, please mark as resolved.

plugins/inputs/ravendb/ravendb.go
srebhan: Thx, please mark as resolved.

plugins/inputs/ravendb/ravendb.go
srebhan: Thx, please mark as resolved.

plugins/inputs/ravendb/ravendb.go
srebhan: Thx, please mark as resolved.

plugins/inputs/ravendb/ravendb.go
srebhan: As I wrote for the bool, you can still use a processor in telegraf for doing the mapping if you need it. This way we can keep the readability and you can still use grafana. Anyway I think it's a bit ugly to do things in the data-collection just to work around a deficit of the visualization. :-)

plugins/inputs/ravendb/ravendb.go
srebhan: Please mark as resolved.

plugins/inputs/ravendb/ravendb.go
srebhan: Please mark as resolved.

plugins/inputs/ravendb/ravendb.go
srebhan: Please mark as resolved.

plugins/inputs/ravendb/ravendb.go
srebhan: Please mark as resolved.

plugins/inputs/ravendb/ravendb.go
srebhan: Sorry I don't understand your comment.

plugins/inputs/ravendb/ravendb.go
ml054: and what about disabling given feature? Should user do: collections_include = ["THIS_COLLECTION_DOES_NOT_EXIST"] ?

Maybe (it this is common solution), we could do:
nil -> get nothing
empty -> get all
some items -> get only those items? 

But such mapping is always problematic. That the reason I like that in GO string cannot be null (unless we use pointers)

plugins/inputs/ravendb/ravendb.go
ml054: ok I got this: https://github.com/influxdata/telegraf/blob/release-1.17/plugins/processors/enum/README.md#enum-processor-plugin will use enums than

plugins/inputs/ravendb/ravendb.go
ml054: ok removed and I'm using boolean

plugins/inputs/ravendb/testdata/collections_full.json
ml054: ok I get the point - thanks

plugins/inputs/ravendb/ravendb_test.go
ml054: actually I can't use that as all operations take place in parallel and order is not deterministic. That's the reason I have this switch.

plugins/inputs/ravendb/ravendb_test.go
ml054: see above ;-)

plugins/inputs/ravendb/ravendb.go
ml054: done

plugins/inputs/ravendb/ravendb.go
ml054: done

plugins/inputs/ravendb/ravendb.go
ml054: sorted

plugins/inputs/ravendb/ravendb.go
ml054: done - initially I took that from rabbitmq and thought it is standard way of doing that.

plugins/inputs/ravendb/ravendb.go
ml054: done

plugins/inputs/ravendb/ravendb.go
ml054: ok refactored that a bit and remove list of functions.

plugins/inputs/ravendb/ravendb.go
ml054: I think I finally undestand what needs to be done - please review.

ml054:
@srebhan I'm still not sure about those include/exclude patterns. Please have a look at my remarks above.

plugins/inputs/ravendb/README.md
srebhan: Please mark as resolved.

plugins/inputs/ravendb/ravendb.go
srebhan: Please mark as resolved.

plugins/inputs/ravendb/ravendb.go
srebhan: Please mark as resolved.

plugins/inputs/ravendb/ravendb.go
srebhan: Please mark as resolved.

plugins/inputs/ravendb/ravendb.go
srebhan: Please mark as resolved.

plugins/inputs/ravendb/ravendb.go
srebhan: No please don't invent anything new. Look at your `*StatsDbs` lists, there you also define a behavior for an empty list. To make an example, currently you have the possible values `server`, `index`, `collection` for the `stats_include` list. To match your current default this corresponds to a
```
stats_include = ["server", "index", "collection"]
```
setting in the config. To disable a certain feature, just remove it from the list. To e.g. disable `collection` just specify 
```
stats_include = ["server", "index"]
```

Now you only have to care for the case of the user specifying an empty list `stats_include = []`. There are two options.
1. An empty list means **nothing** is collected. You might output a warning or error out in this case. If you decide for this behavior, you can set the default `stats_include = ["server", "index", "collection"]` when initializing the structure in `init()`.
2. You can define an empty list as **everything** is collected as a number of other plugins do. This is beneficial if you have many possible options and the default case is to collect everything. In this case you might check the value of `StatsInclude` in `Init()` and fill in all options there to not clutter the `Gather()` function.

As you see the most difficult part is to decide which behavior you want for an empty list. Using `internal.choice` makes it easy to check if the options are valid or if a certain option is contained. In case you define _empty means all_, you can even accompany the `stats_include` with an `stats_exclude`, but for me this is overdoing for three options.

plugins/inputs/ravendb/ravendb.go
srebhan: Please mark as resolved.

plugins/inputs/ravendb/ravendb.go
srebhan: Please mark as resolved.

plugins/inputs/ravendb/ravendb.go
srebhan: I see you fixed this already. :+1: Please mark as resolved.

plugins/inputs/ravendb/ravendb.go
srebhan: Please mark as resolved.

plugins/inputs/ravendb/ravendb.go
srebhan: Please mark as resolved.

plugins/inputs/ravendb/ravendb.go
srebhan: Please mark as resolved.

plugins/inputs/ravendb/testdata/collections_full.json
srebhan: Thx! Please mark as resolved.

plugins/inputs/ravendb/ravendb_test.go
srebhan: Well I think `testutil.RequireMetricsEqual()` cares for the ordering. However to avoid race conditions you should call `acc.Wait(N)` before comparison, where `N` is the number of metrics you expect.

plugins/inputs/ravendb/ravendb.go
srebhan: Please mark as resolved.

plugins/inputs/ravendb/ravendb.go
srebhan: Please mark as resolved.

plugins/inputs/ravendb/ravendb.go
srebhan: Perfect, thx! Please mark as resolved.

plugins/inputs/ravendb/ravendb.go
srebhan: Yeah no problem. Please mark as resolved.

plugins/inputs/ravendb/ravendb.go
srebhan: Nice. Please mark as resolved.

plugins/inputs/ravendb/ravendb.go
srebhan: Please mark as resolved.

plugins/inputs/ravendb/ravendb.go
```diff
@@ -0,0 +1,443 @@
...
+
+func (r *RavenDB) Init() error {
+	if r.URL == "" {
+		r.URL = defaultURL
+	}
+
+	r.requestUrlServer = fmt.Sprintf("%s%s", r.URL, "/admin/monitoring/v1/server")
+	r.requestUrlDatabases = fmt.Sprintf("%s%s", r.URL, "/admin/monitoring/v1/databases"+prepareDbNamesUrlPart(r.DbStatsDbs))
+	r.requestUrlIndexes = fmt.Sprintf("%s%s", r.URL, "/admin/monitoring/v1/indexes"+prepareDbNamesUrlPart(r.IndexStatsDbs))
+	r.requestUrlCollection = fmt.Sprintf("%s%s", r.URL, "/admin/monitoring/v1/collections"+prepareDbNamesUrlPart(r.IndexStatsDbs))
```
srebhan, line 428: I think you can even get rid of the `fmt.Sprintf()`
```suggestion
	r.requestUrlServer = r.URL + "/admin/monitoring/v1/server"
	r.requestUrlDatabases = r.URL + "/admin/monitoring/v1/databases" + prepareDbNamesUrlPart(r.DbStatsDbs)
	r.requestUrlIndexes = r.URL + "/admin/monitoring/v1/indexes" + prepareDbNamesUrlPart(r.IndexStatsDbs)
	r.requestUrlCollection = r.URL + "/admin/monitoring/v1/collections"+prepareDbNamesUrlPart(r.IndexStatsDbs)
```

srebhan (commented):
Please check my comment regarding the include/exclude. Hope it gets more clear!? Anyway great progress. We are close to the finish line I think.

plugins/inputs/ravendb/ravendb_test.go
ml054: it won't help as we gather stats and push to accumulator in parallel, see: https://github.com/ml054/telegraf/blob/master/plugins/inputs/ravendb/ravendb.go#L111

plugins/inputs/ravendb/ravendb.go
ml054: done

ml054:
@srebhan 

Does it work for you:
```
  ## List of statistics which are collected
  # At least one is required
  # stats_include = ["server", "databases", "indexes", "collections"]

  ## List of db where database stats are collected
  ## If empty, all db are concerned
  # db_stats_dbs = []

  ## List of db where index status are collected
  ## If empty, all indexes from all db are concerned
  # index_stats_dbs = []

  ## List of db where collection status are collected
  ## If empty, all collections from all db are concerned
  # collection_stats_dbs = []
```
?

plugins/inputs/ravendb/README.md
```diff
@@ -24,17 +24,9 @@ The following is an example config for RavenDB. **Note:** The client certificate
...
-  ## When true, collect per database stats
-  # gather_db_stats = true
-
-  ## When true, collect per index stats
-  # gather_index_stats = true
-  
-  ## When true, collect per collection stats
-  # gather_collection_stats = true
+  ## List of statistics which are collected
+  # At least one is required
```
srebhan, line 28: Please add a  comment on possible options and what happens in case of this list being empty.

plugins/inputs/ravendb/ravendb.go
```diff
@@ -108,40 +97,39 @@ func (r *RavenDB) Gather(acc telegraf.Accumulator) error {
...
+			}()
+		case "collections":
+			wg.Add(1)
+
+			go func() {
+				defer wg.Done()
+				r.gatherCollections(acc)
+			}()
+		default:
+			panic(fmt.Sprintf("unhandled statistics type %s", statToCollect))
```
srebhan, line 131: Please don't panic here! You can still `acc.AddError()`.

srebhan (commented):
Very good, thank you! Two small things. Please add a comment on possible options to include and on the behavior in case of an empty list. Furthermore, an more importantly. Do _not_ `panic` as this will bring down all telegraf! Please error out instead or return with an error. Please make sure you cancel (or wait for) all ongoing operation before returning!

plugins/inputs/ravendb/ravendb.go
ml054: done

plugins/inputs/ravendb/README.md
ml054: done

plugins/inputs/ravendb/ravendb.go
```diff
@@ -0,0 +1,430 @@
...
+			}()
+		case "collections":
+			wg.Add(1)
+
+			go func() {
+				defer wg.Done()
+				r.gatherCollections(acc)
+			}()
+		default:
+			acc.AddError(fmt.Errorf("unhandled statistics type %s", statToCollect))
```
srebhan, line 133: Please check this in `Init()`. You can use internal/choice with `choice.CheckSlice(r.StatsInclude, []string{"server", "databases", "indexes", "collections"})` and error out early in `Init()`. This why we avoid spamming the logfile in case of an invalid choice.

ml054: done - moved check to init

srebhan (approved):
Looks good to me.

plugins/inputs/ravendb/README.md
```diff
@@ -0,0 +1,216 @@
...
+  # db_stats_dbs = []
+
+  ## List of db where index status are collected
+  ## If empty, all indexes from all db are concerned
+  # index_stats_dbs = []
+  
+  ## List of db where collection status are collected
+  ## If empty, all collections from all db are concerned
+  # collection_stats_dbs = []
+```
```
Jagularr, line 44: small nitpick: some of the comments on these settings end with periods and some don't, likely should be consistent

plugins/inputs/ravendb/ravendb.go
```diff
@@ -0,0 +1,434 @@
...
+  ## If empty, all collections from all db are concerned
+  # collection_stats_dbs = []
+`
+
+// SampleConfig ...
+func (r *RavenDB) SampleConfig() string {
+	return sampleConfig
+}
+
+// Description ...
```
Jagularr, line 88: `// Description ...` and such comments aren't useful

plugins/inputs/ravendb/ravendb.go
```diff
@@ -0,0 +1,434 @@
...
+		result += "&name=" + url.QueryEscape(db)
+	}
+
+	return result
+}
+
+func (r *RavenDB) Init() error {
+	if r.URL == "" {
+		r.URL = defaultURL
+	}
```
Jagularr, line 412: I believe instead of this block, you can just set the URL to defaultURL in the init() method, thus default will be used unless the user configures it.

plugins/inputs/ravendb/ravendb.go
```diff
@@ -0,0 +1,434 @@
...
+}
+
+// Description ...
+func (r *RavenDB) Description() string {
+	return "Reads metrics from RavenDB servers via the Monitoring Endpoints"
+}
+
+// Gather ...
+func (r *RavenDB) Gather(acc telegraf.Accumulator) error {
+	err := r.ensureClient()
```
Jagularr, line 95: I believe this method can be called in Init(). It appears that the client is only initialized once?

plugins/inputs/ravendb/ravendb.go
```diff
@@ -0,0 +1,434 @@
...
+	}
+
+	tags := map[string]string{
+		"cluster_id": serverResponse.Cluster.Id,
+		"node_tag":   serverResponse.Cluster.NodeTag,
+		"url":        r.URL,
+	}
+
+	if serverResponse.Config.PublicServerUrl != nil {
+		tags["public_server_url"] = *serverResponse.Config.PublicServerUrl
```
Jagularr, line 199: Should you check for the string being blank here as well? if that's at all possible to occur

plugins/inputs/ravendb/ravendb.go
```diff
@@ -0,0 +1,434 @@
...
+			go func() {
+				defer wg.Done()
+				r.gatherDatabases(acc)
+			}()
+		case "indexes":
+			wg.Add(1)
+
+			go func() {
+				defer wg.Done()
+				r.gatherDatabases(acc)
```
Jagularr, line 124: did you mean to say gatherIndexes(acc) here?

plugins/inputs/ravendb/ravendb.go
```diff
@@ -0,0 +1,434 @@
...
+		r.Log.Errorf("Error with Client %s", err)
+		return err
+	}
+
+	var wg sync.WaitGroup
+
+	for _, statToCollect := range r.StatsInclude {
+		switch statToCollect {
+		case "server":
+			wg.Add(1)
```
Jagularr, line 106: nit: I think you can remove the four `wg.Add(1)` lines and just add a single one at the top of the loop. In Init() you already assured that each entry is going to be one of these cases.

Jagularr (changes requested):
Looks like a very solid plugin. Mostly minor feedback attached. Thanks!

plugins/inputs/ravendb/ravendb.go
srebhan: No please leave this in here to prevent user explicitly specifying `url = ""` from being surprised.

srebhan:
Oh boy what did I review here... *blush*  Thanks @Jagularr for catching those oversights... :-/

plugins/inputs/ravendb/README.md
ml054: removed dots

plugins/inputs/ravendb/ravendb.go
ml054: removed

plugins/inputs/ravendb/ravendb.go
ml054: leaving as it is.

plugins/inputs/ravendb/ravendb.go
ml054: moved to init method.

plugins/inputs/ravendb/ravendb.go
ml054: no need to check - if option is defined in RavenDB it can't be empty.

plugins/inputs/ravendb/ravendb.go
ml054: THANKS!!! Wonderful catch. :)

plugins/inputs/ravendb/ravendb.go
ml054: done

ml054:
@Jagularr thanks a lot for review.

Jagularr (approved):

PR #8835, opened 2021-02-08
Revive fixes - part 2

zak-pawel:
**[Revive](https://github.com/mgechev/revive)** is fast, configurable, extensible, flexible, and beautiful linter for Go. Drop-in replacement of golint.

This is part2 of fixes for problems found for following set of rules:
- [rule.if-return]
  - Redundant if when returning an error. -> Checking if an error is nil to just after return the error or nil is redundant.
- [rule.increment-decrement]
  - Use `i++` and `i--` instead of `i += 1` and `i -= 1`. -> By convention, for better readability, incrementing an integer variable by 1 is recommended to be done using the `++` operator. This rule spots expressions like `i += 1` and `i -= 1` and proposes to change them into `i++` and `i--`.
- [rule.var-declaration]
  - Reduces redundancies around variable declaration. -> This rule proposes simplifications of variable declarations.
- [rule.package-comments]
  - Package commenting conventions. -> Packages should have comments. This rule warns on undocumented packages and when packages comments are detached to the `package` keyword. More information [here](https://github.com/golang/go/wiki/CodeReviewComments#package-comments)
- [rule.receiver-naming]
  - Conventions around the naming of receivers. -> By convention, receiver names in a method should reflect their identity. For example, if the receiver is of type `Parts`, `p` is an adequate name for it. Contrary to other languages, it is not idiomatic to name receivers as `this` or `self`.
- [rule.unexported-return]
  - Warns when a public return is from unexported type. -> This rule warns when an exported function or method returns a value of an un-exported type.
  - Few findings left untouched -> for some deeper review in the future.

For changed files, minor fixes were also made (got rid off "log" package where it was easy, CamelCase instead of snake_case, etc...)


models/running_output.go
```diff
@@ -110,17 +110,17 @@ func NewRunningOutput(
 	return ro
 }
 
-func (r *RunningOutput) LogName() string {
-	return logName("outputs", r.Config.Name, r.Config.Alias)
+func (ro *RunningOutput) LogName() string {
+	return logName("outputs", ro.Config.Name, ro.Config.Alias)
```
srebhan, line 114: Does this renaming really make sense? I mean almost all plugins do have one-letter var-names for this...

plugins/inputs/cpu/cpu.go
```diff
@@ -97,7 +97,7 @@ func (s *CPUStats) Gather(acc telegraf.Accumulator) error {
 		totalDelta := total - lastTotal
 
 		if totalDelta < 0 {
-			err = fmt.Errorf("Error: current total CPU time is less than previous total CPU time")
+			err = fmt.Errorf("error: current total CPU time is less than previous total CPU time")
```
srebhan, line 100: `error:` is redundant and should be removed IMO.

plugins/inputs/ipset/ipset.go
```diff
@@ -46,8 +46,8 @@ func (ipset *Ipset) SampleConfig() string {
 `
 }
 
-func (ips *Ipset) Gather(acc telegraf.Accumulator) error {
-	out, e := ips.lister(ips.Timeout, ips.UseSudo)
+func (ipset *Ipset) Gather(acc telegraf.Accumulator) error {
```
srebhan, line 49: Here we diverge from the usual one-letter variable name... Only because it's i? :-)

plugins/outputs/dynatrace/dynatrace.go
```diff
@@ -236,7 +236,7 @@ func (d *Dynatrace) send(msg []byte) error {
 	req, err := http.NewRequest("POST", d.URL, bytes.NewBuffer(msg))
 	if err != nil {
 		d.Log.Errorf("Dynatrace error: %s", err.Error())
-		return fmt.Errorf("Dynatrace error while creating HTTP request:, %s", err.Error())
+		return fmt.Errorf("dynatrace error while creating HTTP request:, %s", err.Error())
```
srebhan, line 239: Maybe remove `dynatrace` completely or otherwise leave capitalized as it is a name.

plugins/processors/topk/topk.go
```diff
@@ -276,7 +276,7 @@ func (t *TopK) push() []telegraf.Metric {
 	}
 
 	// The return value that will hold the returned metrics
-	var ret []telegraf.Metric = make([]telegraf.Metric, 0, 0)
+	var ret = make([]telegraf.Metric, 0, 0)
```
srebhan, line 279: Hmm isn't this equivalent to `var ret []telegraf.Metric` which I would prefer then...

srebhan (commented):
Looks good, found only very minor things... ;-)
Regarding the renaming of the "class" variable I'm not so sure about the exceptions, but if everybody can live with it...

models/running_output.go
zak-pawel: `rule.receiver-naming` requires that receiver names should be consistent (have the same var-name) and they weren't in this case. I changed all of them to `ro` (I like letter-per-word approach) but I can change them to `r` if you insist - still some changes to these receiver will have to be made...

plugins/inputs/ipset/ipset.go
zak-pawel: `rule.receiver-naming` requires that receiver names should be consistent (have the same var-name) and they weren't in this case. Previous receivers had used `ipset` so I adjusted,  but I can change all of them to `i` if you insist ;)

plugins/inputs/cpu/cpu.go
zak-pawel: Removed

plugins/outputs/dynatrace/dynatrace.go
zak-pawel: Removed

plugins/processors/topk/topk.go
zak-pawel: Then `ret` will be `nil` initially but it is ok for this case. Changed.

models/running_output.go
srebhan: I'm not insisting, I'm just asking if enforcing these names makes sense? If you think it does, please change those.

plugins/inputs/ipset/ipset.go
srebhan: Again I'm not insisting, but if you don't need to add an exception, I wouldn't add one. Just my opinion. :-)

models/running_output.go
zak-pawel: @srebhan changed to `r`

plugins/inputs/ipset/ipset.go
zak-pawel: @srebhan changed to `i`

srebhan (approved):
Looks good to me.

zak-pawel:
For retry

zak-pawel:
For retry of windows tests

plugins/inputs/passenger/passenger.go
```diff
@@ -17,13 +17,13 @@ type passenger struct {
 	Command string
 }
 
-func (p *passenger) parseCommand() (string, []string) {
+func (pssngr *passenger) parseCommand() (string, []string) {
```
ssoroka, line 20: seems like the short one was better?

plugins/inputs/webhooks/webhooks.go
```diff
@@ -28,14 +27,16 @@ func init() {
 }
 
 type Webhooks struct {
-	ServiceAddress string
+	ServiceAddress string `toml:"service_address"`
```
ssoroka, line 30: is this a breaking change?

agent/tick_test.go
```diff
@@ -32,7 +30,7 @@ func TestAlignedTicker(t *testing.T) {
 		time.Unix(60, 0).UTC(),
 	}
 
-	actual := []time.Time{}
+	var actual []time.Time
```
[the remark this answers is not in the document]
zak-pawel: Right, reverted.

models/running_output.go
```diff
@@ -114,8 +114,8 @@ func (r *RunningOutput) LogName() string {
 	return logName("outputs", r.Config.Name, r.Config.Alias)
 }
 
-func (ro *RunningOutput) metricFiltered(metric telegraf.Metric) {
-	ro.MetricsFiltered.Incr(1)
+func (r *RunningOutput) metricFiltered(metric telegraf.Metric) {
+	r.MetricsFiltered.Incr(1)
```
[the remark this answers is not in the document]
zak-pawel: @srebhan asked me to use one letter "vars" in receivers. The rule we would like to follow here is "receiver name should be consistent with previous receiver name'. Of course, I can change all of them to `ro`

plugins/common/shim/config.go
```diff
@@ -28,28 +28,28 @@ type loadedConfig struct {
...
+			return fmt.Errorf("failed to add Output: %w", err)
 		}
 	}
 	return nil
 }
 
-// LoadConfig loads the config and returns inputs that later need to be loaded.
-func LoadConfig(filePath *string) (loaded loadedConfig, err error) {
+// loadConfig loads the config and returns inputs that later need to be loaded.
+func loadConfig(filePath *string) (loaded loadedConfig, err error) {
```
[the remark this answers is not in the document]
zak-pawel: Right, reverted.

plugins/common/shim/config_test.go
```diff
@@ -81,7 +81,7 @@ func (i *testDurationInput) SampleConfig() string {
 func (i *testDurationInput) Description() string {
 	return ""
 }
-func (i *testDurationInput) Gather(acc telegraf.Accumulator) error {
+func (i *testDurationInput) Gather(_ telegraf.Accumulator) error {
```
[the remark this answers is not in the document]
zak-pawel: Reverted.

plugins/inputs/passenger/passenger.go
zak-pawel: Changed to `p`

plugins/inputs/webhooks/webhooks.go
zak-pawel: I don't think so. When I added logger, I tried to add also tags as suggested here: **https://github.com/influxdata/telegraf/wiki/Review#reviewing-plugin-code**

models/running_output.go
ssoroka: yeah. maybe we should use one letter receivers.

plugins/inputs/ipset/ipset.go
ssoroka: docs seem to agree. resolving.

ssoroka:
I think there's still a lot of var struct/map/slice cases left over. I'm guessing we don't want to use `rule.var-declaration`.

zak-pawel:
> I think there's still a lot of var struct/map/slice cases left over. I'm guessing we don't want to use `rule.var-declaration`.

@ssoroka All var struct/map/slice cases reverted. `rule.var-declaration` is just fine, I've just changed too much.

ssoroka (approved):

Issue #8805, opened 2021-02-04
ping plugin broken by commit c237989


0x0a11c0de:
[This commit](https://github.com/influxdata/telegraf/commit/c2379896312ba0582a1e7a9be79f9dc826a955a2#diff-1e6bedd7550baa6b9ff9469e1c2d12cfde952f9acce4b2abd928b4c46e6962d2) introduced two problems:

1. [It sets the go-ping timeout to the per-ping timeout value](https://github.com/influxdata/telegraf/blob/c2379896312ba0582a1e7a9be79f9dc826a955a2/plugins/inputs/ping/ping.go#L173). The timeout in `go-ping` is not the per-ping timeout. [It is the timeout for exiting the pinger](https://github.com/go-ping/ping/blob/b6486c6f1f2d7aa4e017c9571730d467c4d4a2b8/ping.go#L372). The `go-ping` timeout should be set to the `deadline` value. `go-ping` does not appear to have a way to set the per-ping timeout value.
2. It removed the sorting of the rtts before calculating the percentiles, which means the percentile calculation cannot work.

psarossy:
Was this change actually released?

The documentation is updated, stating that as of 1.7.0 the percentiles are available, but when running the 1.7.2 docker image I get:
```
│ 2021/02/17 18:52:12 E! Error parsing /etc/telegraf/telegraf.conf, line 25: field corresponding to `percentiles' is not defined in `*ping.Ping'                                                                                               │
```

PR #8594, opened 2020-12-18
Add quantile aggregator plugin

srebhan:
### Required for all PRs:

- [X] Signed [CLA](https://influxdata.com/community/cla/).
- [X] Associated README.md updated.
- [X] Has appropriate unit tests.

This PR implements a quantile aggregator allowing to aggregate multiple series and output the specified quantiles per field and metric. It provides three types of approximations, T-Digest as well as exact approximations using Hyndman & Fan's R7 (equivalent to Excel and Numpy) and R8 (recommended in the paper) approach.

The plugin adds *no* further dependencies, but it was necessary to update `github.com/wavefronthq/wavefront-sdk-go` from 0.9.2 to 0.9.7 to get an up-to-date `github.com/caio/go-tdigest` library used here.

The approach is similar to PR #6887 but has much less code. Closes #6440.

PR #8839, opened 2021-02-08
Fix percentile calculations and assign Deadline correctly

sspaink:
### Required for all PRs:

#8805

- Sort the RTTS slice before calculating the percentile, updated a unit test to verify this
- Changed inputs.ping "Deadline" config to be assigned to go-pings "Timeout"

go-ping doesn't support "timeout to wait for ping response (-W)" yet, but there is an open issue https://github.com/go-ping/ping/issues/63

### DIFF
--- a/plugins/inputs/ping/ping.go
+++ b/plugins/inputs/ping/ping.go
@@ -7,6 +7,7 @@ import (
 	"net"
 	"os/exec"
 	"runtime"
+	"sort"
 	"strings"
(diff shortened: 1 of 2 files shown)


srebhan (approved):
Looks good to me.
Just a hint: You could also use the quantile aggregator (#8594) but now we have this in here... :-(

Jagularr (approved):

PR #8844, opened 2021-02-09
Move windows tests to circle-ci, remove appveyor

sspaink:
Only the windows tests are run in appveyor, moved the execution of the tests to circle-ci config so we can have only one CI system.


.circleci/config.yml
```diff
@@ -186,6 +198,10 @@ workflows:
           filters:
             tags: # only runs on tags if you specify this filter
               only: /.*/
+      - 'test-go-windows':
+          filters:
+            tags:
+              only: /.*/
       - 'package':
```
Jagularr, line 205: shouldn't 'package' and 'release' also require test-go-windows?

sspaink: Good catch, added `test-go-windows` as requirements for package and release

.circleci/config.yml
```diff
@@ -1,4 +1,6 @@
 version: 2.1
+orbs:
+  win: circleci/windows@2.2.0
```
reimda, line 3: This uses 2.2.0 but 2.4.0 is the latest. https://circleci.com/developer/orbs/orb/circleci/windows

I'm new to circleci orbs, maybe the version isn't important.  I wasn't able to find the 2.2.0 source to compare them.  Did you come across docs or anything that explains the differences?

reimda (changes requested):
Looks good.  We probably ought to run it nightly too since the other platforms do.

Jagularr (approved):

.circleci/config.yml
sspaink: I couldn't find a history for the version either to see what changed, but using the latest version sounds like a good idea to me.

reimda (approved):
Looks good to me.  @sspaink Would you let the rest of the team know before we switch over, maybe on community slack #telegraf-dev?  Then would you deactivate the appveyor webhooks and merge this?  Deactivating the webhooks instead of deleting will give us the chance to revert if we have to for some reason.

To deactivate them, go into the github repo settings and webhooks section then edit the webhook and uncheck the active checkbox (at the very bottom) and click "update webhook".

sspaink:
@reimda thanks for the review! I've disabled appveyor webhooks in the settings and posted to #telegraf-dev.

PR #8849, opened 2021-02-12
Filter data out from system databases for Azure SQL DB only

masree:
Azure SQL DB has few system databases like msdb, model .... which emits data and are not useful for customers for monitoring. However, we're collecting data from telegraf queries and seems unnecessary as this data does not make sense for single DB customers. Excluding these system databases from performance DMVs to lower the amount of data emitted.

Please enter the commit message for your changes. Lines starting

### Required for all PRs:


plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -437,8 +437,17 @@ WITH PerfCounters AS (
...
-		counter_name IN (
+	WHERE 
+		spi.instance_name NOT IN (
+			'master'
+			,'model'
+			,'model_masterdb'
+			,'model_userdb'
+			,'msdb'
+			,'mssqlsystemresource'
+			,'tempdb'
```
denzilribeiro, line 448: We want Tempdb and master for sure, others are good to exclude as this is Single DB offering so doesn't make sense. However Tempdb is critical and there are background tasks counted in master.

denzilribeiro (changes requested):
Makes sense to remove model/replicated databases as this is SOLELY for the single Azure DB collector, leave master and tempdb though

plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -437,8 +437,17 @@ WITH PerfCounters AS (
 				THEN d.[name]
 			ELSE d.[physical_database_name]
 		END
-	WHERE
-		counter_name IN (
+	WHERE 
+		spi.instance_name NOT IN (
```
dimitri-furman, line 441: There is a couple of problems here:
1. The instance_name column is nullable. If there is a NULL in a row, the results from NOT IN () may be unexpected.
2. We should also check object_name, because the same instance_name for a different object could carry a different meaning.

Both of these can be addressed with the following rewrite:

```sql
        -- filter out unnecessary SQL DB system database counters, other than master and tempdb
        NOT (spi.object_name LIKE 'MSSQL%:Databases%' AND spi.instance_name IN ('model','model_masterdb','model_userdb','msdb','mssqlsystemresource'))
        AND
        (
		counter_name IN (
...
```

dimitri-furman (changes requested):

plugins/inputs/sqlserver/azuresqlqueries.go
masree: Ah got it, I've updated the PR with the changes you suggested. Thanks!

plugins/inputs/sqlserver/azuresqlqueries.go
masree: Sure, added exclusions for other DBs except master and tempdb. Thanks!

dimitri-furman (approved):

denzilribeiro (approved):

masree:
@ssoroka / @sjwang90 / @srebhan - Can you help review and merge if no concerns?

srebhan:
Hmmm as the discussion shows finding a consent on what ppl want is difficult. How about make this exclude list a config option that is, by default, set to the list you want to exclude? People really wanting to monitor that DBs for whatever reason are then free to do so.

masree:
> Hmmm as the discussion shows finding a consent on what ppl want is difficult. How about make this exclude list a config option that is, by default, set to the list you want to exclude? People really wanting to monitor that DBs for whatever reason are then free to do so.

Thank you for the suggestion. In Azure SQL database offering, Azure SQL manages system databases(modelDB, msdb ..) and customer will likely not find any useful information for monitoring in these system databases that are currently filtered out. We think the filtered list of system databases are standard and not useful for monitoring. Even if we make it as an option in config for customers to exclude, most/all customers might always keep the default config as including system databases just dumps more data without actionable results. We wanted to keep it simple by excluding them in the query, but if you think it is a concern, we can follow your suggestion. Let us know what you think. Thanks!

srebhan (approved):
Ok I now understand the use-case and I'm fine with the change. However, please keep in mind that if you ever want to touch this again to make it a config option! ;-)

ssoroka (approved):

ssoroka:
looks like there's conflicts to resolve. Also is it at all possible that this data is useful to some customers, and we should instead parameterize it?

denzilribeiro:
> looks like there's conflicts to resolve. Also is it at all possible that this data is useful to some customers, and we should instead parameterize it?

I don't think there is conflict - the change is only for "single database" offering of Azure , not for SQL Managed instance ( multiple DBs') or on-prem SQL Server ( multiple-DB's). I think you risk over parameterizing things too, there is no one that will get any benefit looking at these couners for master/model/MSDB databases. If you make it a config and then make the default as what backward compat was, then you remove usefulness of it, if we parameterize then would want default for param to be to exclude system Dbs not the other way around. I personally think no need of parameterization

srebhan:
@denzilribeiro I think @ssoroka is referring to the merge conflict in `plugins/inputs/sqlserver/azuresqlqueries.go
` that needs to be resolved...

masree:
> @denzilribeiro I think @ssoroka is referring to the merge conflict in `plugins/inputs/sqlserver/azuresqlqueries.go` that needs to be resolved...

@ssoroka I'll work on resolving the merge conflict and push a new iteration. Thanks everyone for the review!

masree:
Fetched the latest from upstream and resolved the conflict.

masree:
> > @denzilribeiro I think @ssoroka is referring to the merge conflict in `plugins/inputs/sqlserver/azuresqlqueries.go` that needs to be resolved...
> 
> @ssoroka I'll work on resolving the merge conflict and push a new iteration. Thanks everyone for the review!

PR #4615, opened 2018-08-30
Nfsclient input: (Reopen of PR 1491 and 1684)

pmoranga:
### Required for all PRs:

- [x] Signed [CLA](https://influxdata.com/community/cla/).
- [x] Associated README.md updated.
- [x] Has appropriate unit tests.

Resolves #820, I just fixed the README from PR#1648, as it was idle for a long time. I didn't squash to preserve the mention from original contributors.

PR #8865, opened 2021-02-16
feat: implement nfsclient stats input plugin on linux to count kernel…

write0nly:
… nfs client side statistics per host

### Required for all PRs:

- [README.md](plugins/inputs/nfsstat/README.md) Associated README.md updated.
- [/] Has appropriate unit tests.

### DIFF
--- a/plugins/inputs/nfsstat/nfsstat.go
+++ b/plugins/inputs/nfsstat/nfsstat.go
@@ -1,0 +1,289 @@
+package nfsstat
+
+import (
+	"strconv"
+	"strings"
+
+	"errors"
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/plugins/inputs"
+	"io/ioutil"
+)
+
+var (
+	ErrorInvalidNFSVersion   error = errors.New("invalid nfs version")
+	ErrorInvalidNFSOpIndex   error = errors.New("invalid nfs operation index")
+	ErrorInvalidCounterValue error = errors.New("invalid counter value")
+)
+
+const (
+	prefix_nfs3 string = "proc3 22 "
+	prefix_nfs4 string = "proc4 60 "
+)
+
+type NFSStat struct {
+	Log telegraf.Logger
+}
+
+type NFSStatCounter struct {
+	nfsvers string
+	op      string
+	val     int64
+}
+
+func (n *NFSStat) SampleConfig() string {
+	return "[[inputs.nfsstat]]"
+}
+
+func (n *NFSStat) Description() string {
+	return "Read global nfsclient stats"
+}
+
+func getNFSVers(s string) (int, bool) {
+	if strings.HasPrefix(s, prefix_nfs3) {
+		return 3, true
+	}
+
+	if strings.HasPrefix(s, prefix_nfs4) {
+		return 4, true
+	}
+
+	return 0, false
+}
+
+func getNFSv3Ops() []string {
+	return []string{
+		"null",
+		"getattr",
+		"setattr",
+		"lookup",
+		"access",
+		"readlink",
+		"read",
+		"write",
+		"create",
+		"mkdir",
+		"symlink",
+		"mknod",
+		"remove",
+		"rmdir",
+		"rename",
+		"link",
+		"readdir",
+		"readdirplus",
+		"fsstat",
+		"fsinfo",
+		"pathconf",
+		"commit",
+	}
+}
+
+func getNFSv4Ops() []string {
+	return []string{
+		"null",
+		"read",
+		"write",
+		"commit",
+		"open",
+		"open_confirm",
+		"open_noattr",
+		"open_downgrade",
+		"close",
+		"setattr",
+		"fsinfo",
+		"renew",
+		"setclientid",
+		"setclientid_confirm",
+		"lock",
+		"lockt",
+		"locku",
+		"access",
+		"getattr",
+		"lookup",
+		"lookup_root",
+		"remove",
+		"rename",
+		"link",
+		"symlink",
+		"create",
+		"pathconf",
+		"statfs",
+		"readlink",
+		"readdir",
+		"server_caps",
+		"delegreturn",
+		"getacl",
+		"setacl",
+		"fs_locations",
+		"release_lockowner",
+		"secinfo",
+		"fsid_present",
+		"exchange_id",
+		"create_session",
+		"destroy_session",
+		"sequence",
+		"get_lease_time",
+		"reclaim_complete",
+		"layoutget",
+		"getdeviceinfo",
+		"layoutcommit",
+		"layoutreturn",
+		"secinfo_no_name",
+		"test_stateid",
+		"free_stateid",
+		"getdevicelist",
+		"bind_conn_to_session",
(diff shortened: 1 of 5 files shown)


srebhan:
Thanks for submitting this. Can you please elaborate on how it is different than https://github.com/influxdata/telegraf/tree/master/plugins/inputs/nfsclient? 

In case it is different, can you please check if you can merge your code into this plugin to avoid having two plugins doing similar things...

write0nly:
> Thanks for submitting this. Can you please elaborate on how it is different than https://github.com/influxdata/telegraf/tree/master/plugins/inputs/nfsclient?
> 
> In case it is different, can you please check if you can merge your code into this plugin to avoid having two plugins doing similar things...

The reason I wrote this plugin instead of tagging along the mountstats one is because I believe that using mountstats is fraught with a large number of issues, and likely to give you results that can be easily mis-interpreted. eg:

1. mountstats counters are namespaced, so if you have docker containers (or just unshare -m) that mount nfs without doing a -v from the OS mount there will be innacurate counters, values that do not match, or mounts that are not accounted for.
2. if you run telegraf inside docker you may have a different and unexpected set of results.
3. even in the same namespace if you have 2 NFS mounts /mnt/1 and /mnt/2 that bind to the same vfsmount the counters will be the same for both so the metrics will show unexpected results. The note on the plugin page implies that this is a limitation of the metrics exposed by the kernel, however this is not really accurate as the counters should be per server:fsid not mountpoint:

>NOTE Many of the metrics, even if tagged with a mount point, are really per-server. Thus, if you mount these two shares: nfs01:/vol/foo/bar and nfs01:/vol/foo/baz, there will be two near identical entries in /proc/self/mountstats. This is a limitation of the metrics exposed by the kernel, not the telegraf plugin.

Therefore I would argue that you have 2 very different use cases that measure totally different things, and therefore should not be conflated together:

1. nfsstats, global counters from the kernel nfs client that will give you 100% accurate counters on all nfs operations done on the host via the kernel nfs code.
2. the input on #4615 based on mountstats , which will give you accurate measurements in a single namespace for a tuple (server, fsid), allows you to have per nfsmount stats, allows for latency and many other goodies that are not the scope of nfsstats.

I would argue that the name nfsclient for the plugin #4615 is innacurate, as it is actually nfsmount based, but I would not pursue this discussion. I would also argue that the points above should be added as notes to the other plugin, because it is likely to lead users into confusion and future issues unless they understand the peculiarities of the measurements in great detail.

write0nly:
> Thanks for submitting this. Can you please elaborate on how it is different than https://github.com/influxdata/telegraf/tree/master/plugins/inputs/nfsclient?
> 
> In case it is different, can you please check if you can merge your code into this plugin to avoid having two plugins doing similar things...

The reason I wrote this plugin instead of tagging along the mountstats one is because I believe that using mountstats is fraught with a large number of issues, and likely to give you results that can be easily mis-interpreted. eg:

1. mountstats counters are namespaced, so if you have docker containers (or just unshare -m) that mount nfs without doing a -v from the OS mount there will be innacurate counters, values that do not match, or mounts that are not accounted for.
2. if you run telegraf inside docker you may have a different and unexpected set of results.
3. even in the same namespace if you have 2 NFS mounts /mnt/1 and /mnt/2 that bind to the same vfsmount the counters will be the same for both so the metrics will show unexpected results. The note on the plugin page implies that this is a limitation of the metrics exposed by the kernel, however this is not really accurate as the counters should be per server:fsid not mountpoint:

>NOTE Many of the metrics, even if tagged with a mount point, are really per-server. Thus, if you mount these two shares: nfs01:/vol/foo/bar and nfs01:/vol/foo/baz, there will be two near identical entries in /proc/self/mountstats. This is a limitation of the metrics exposed by the kernel, not the telegraf plugin.

Therefore I would argue that you have 2 very different use cases that measure totally different things, and therefore should not be conflated together:

1. nfsstats, global counters from the kernel nfs client that will give you 100% accurate counters on all nfs operations done on the host via the kernel nfs code.
2. the input on #4615 based on mountstats , which will give you accurate measurements in a single namespace for a tuple (server, fsid), allows you to have per nfsmount stats, allows for latency and many other goodies that are not the scope of nfsstats.

I would argue that the name nfsclient for the plugin #4615 is innacurate, as it is actually nfsmount based, but I would not pursue this discussion. I would also argue that the points above should be added as notes to the other plugin, because it is likely to lead users into confusion and future issues unless they understand the peculiarities of the measurements in great detail.

plugins/inputs/nfsstat/nfsstat.go
```diff
@@ -0,0 +1,289 @@
...
+}
+
+type NFSStatCounter struct {
+	nfsvers string
+	op      string
+	val     int64
+}
+
+func (n *NFSStat) SampleConfig() string {
+	return "[[inputs.nfsstat]]"
```
ssoroka, line 35: SampleConfig() shouldn't include the name of the plugin
```suggestion
	return ""
```

plugins/inputs/nfsstat/nfsstat.go
```diff
@@ -0,0 +1,289 @@
...
+	return nfsops[index], nil
+}
+
+func parseCounters(s string) ([]int64, error) {
+	ss := strings.Split(s, " ")
+	sn := make([]int64, len(ss))
+	for i, val := range ss {
+		val_i, err := strconv.ParseInt(val, 0, 64)
+		if err != nil {
+			return []int64{}, err
```
ssoroka, line 181: do you want to return only the counters that parsed correctly?

plugins/inputs/nfsstat/nfsstat.go
```diff
@@ -0,0 +1,289 @@
...
+		s = line[len(prefix_nfs3):]
+	case 4:
+		s = line[len(prefix_nfs4):]
+	default:
+		return []int64{}, ErrorInvalidNFSVersion
+	}
+
+	stats, err := parseCounters(s)
+	if err != nil {
+		return []int64{}, err
```
ssoroka, line 208: Can return nils instead of empty slices. `len(nil) == 0` anyway.
```suggestion
		return nil, err
```

plugins/inputs/nfsstat/nfsstat.go
```diff
@@ -0,0 +1,289 @@
...
+
+	return nil
+}
+
+func (n *NFSStat) Gather(acc telegraf.Accumulator) error {
+
+	rawstats, err := ioutil.ReadFile("/proc/net/rpc/nfs")
+	if err != nil {
+		n.Log.Errorf("Failed to read /proc/net/rpc/nfs:  ", err)
+		return err
```
ssoroka, line 275: don't need to log as well as return. if you want the extra context use 
```suggestion
		return fmt.Errorf("reading /proc/net/rpc/nfs: %w", err)
```

ssoroka (changes requested):
It'd be good to also update the docs of both plugins with some of the context from your comments on the issue to help users decide which plugin to use.

helenosheaa:
@write0nly thanks for the PR. Are you still working on this? There's some review comments we'd like to get clarified before we proceed.

sjwang90:

PR #8872, opened 2021-02-17
Revive fixes - part 3

zak-pawel:
**[Revive](https://github.com/mgechev/revive)** is fast, configurable, extensible, flexible, and beautiful linter for Go. Drop-in replacement of golint.

This is part3 of fixes for problems found for following set of rules:
- [rule.var-naming]
  - Naming rules. -> This rule warns when [variable](https://github.com/golang/go/wiki/CodeReviewComments#variable-names) or [package](https://github.com/golang/go/wiki/CodeReviewComments#package-names) naming conventions are not followed.
  - This PR contains fixes for more than 400 issues found by this rule. 
  - All `don't use an underscore in package name` issues were not fixed -> Telegraf allows to have underscores in package names. This approach should be rethought or having underscores in packages should be ignored.


plugins/inputs/mailchimp/mailchimp.go
```diff
@@ -11,9 +11,9 @@ import (
 type MailChimp struct {
 	api *ChimpAPI
 
-	ApiKey     string
+	APIKey     string
 	DaysOld    int
-	CampaignId string
+	CampaignID string
```
srebhan, line 16: These might be exposed to `toml`. Are you sure that the `toml` config parser still does the same as before?

zak-pawel: Right, missed adding `toml` tags in this file. Fixed.

plugins/inputs/raindrops/raindrops.go
```diff
@@ -16,8 +16,8 @@ import (
 )
 
 type Raindrops struct {
-	Urls        []string
-	http_client *http.Client
+	Urls       []string
```
srebhan, line 19: Hmmm `Url` is changed to `URL` but `Urls` is not changed to `URLs`?

plugins/inputs/rethinkdb/rethinkdb_server.go
```diff
@@ -15,7 +15,7 @@ import (
 )
 
 type Server struct {
-	Url          *url.URL
+	URL          *url.URL
```
srebhan, line 18: This one might also be exposed to `toml` through the `RethinkDB` struct.

plugins/outputs/instrumental/instrumental.go
```diff
@@ -22,7 +22,7 @@ var (
 
 type Instrumental struct {
 	Host       string
-	ApiToken   string
+	APIToken   string
```
srebhan, line 25: Exposed to `toml`.

plugins/outputs/wavefront/wavefront.go
```diff
@@ -13,7 +13,7 @@ import (
 const maxTagLength = 254
 
 type Wavefront struct {
-	Url             string
+	URL             string
```
srebhan, line 16: Exposed to `toml`.

srebhan (commented):
Nice. Please double check the struct elements exposed to `toml` but not carrying an annotation as these might change the config option...

plugins/inputs/raindrops/raindrops.go
zak-pawel: Revive didn't find plurals. This rule can be configured to find all `Urls` but it will suggest to rename them to `URLS`...
I don't know if we want to follow that path...

plugins/inputs/rethinkdb/rethinkdb_server.go
zak-pawel: `Server` is only created manually. `RethinkDB` struct uses array of strings.

```
type RethinkDB struct {
	Servers []string
}
```

plugins/outputs/instrumental/instrumental.go
zak-pawel: Fixed

plugins/outputs/wavefront/wavefront.go
zak-pawel: Fixed

plugins/inputs/raindrops/raindrops.go
srebhan: Not sure but it caught my attention. :-)

srebhan (approved):
Looks good to me, but might have missed some `toml` issues... :-)

zak-pawel:
!retry-check

zak-pawel:
!retry-checks

zak-pawel:
!retry-failed

ssoroka (approved):

PR #8878, opened 2021-02-18
Mac Packaging / Notarization

### LINKED ISSUE #8580: Telegraf for macOS should be signed
### Proposal:

Apple has a hard requirement in current versions of macOS for binaries to be signed and notarized. This FR is to start signing the telegraf binary with a valid Apple certificate and notarize as needed. Here is some documentation from Apple on this topic: [https://developer.apple.com/support/code-signing/](url). 

This would ideally mean someone at a company level obtaining an company developer account from developer.apple.com, which is $300/year. Personal accounts could be used at $99/year but the signing would appear in that persons name. 

### Current behavior:
When running telegraf on macOS Big Sur, the following error messages appear: 

Screen Shot 2020-12-16 at 3 54 12 PM

Screen Shot 2020-12-16 at 3 54 34 PM

### Desired behavior:

telegraf should run without error. 

### Use case: 

For continued use of telegraf on the macOS platform, it will need to be signed and notarized.

Jagularr:
This update will run a script on 'release' deployments that will sign and notarize our Mac OS packages. Notarization requires the .app format and the binary will be distributed within the Resources of the .app, and the .app will be distributed within a DMG image. Note that those who need direct access to the Telegraf binary can still go into the .app 'folder' and retrieve it as normal.

What this PR accomplishes in a practical manner is that users can now double-click/open the Telegraf app directly after downloading it, and mac OS will not block them from opening it. Users can also run the Telegraf.app through the command line, as follows:

`open -a Telegraf.app`

you can also pass any Telegraf arguments as follows:

`open -a Telegraf.app --args (arguments)`

resolves #8580


Jagularr:
Right now this replaces the .tar.gz package that we've been distributing for mac. Let me know if I should leave that in, and we'll distribute both the signed .app as well as the original package. Note that either way, the telegraf binary itself can be accessed easily, as normal.

Hipska:
Not sure, but I think it might still be needed for homebrew: https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/telegraf.rb

I guess brew can also unpack from the dmg..

ssoroka (approved):

Issue #8759, opened 2021-01-27
New DPDK input plugin


p-zak:
## Feature Request
We have created a Telegraf DPDK input plugin which collects metrics exposed by applications built with [Data Plane Development Kit](https://www.dpdk.org/) which is an extensive set of open source libraries designed for accelerating packet processing workloads.

#### In the near future we will propose PR which adds new plugin we have created. It will be able to collect telemetry from:
- DPDK NICs (ethdev): 
  - stats - basic device statistics
  - xstats - extended device statistics
  - link_status - up/down link status
- DPDK Raw Devices (rawdev):
  -  xstats - extended device statistics
- DPDK applications (custom, application-specific metrics)

### Use case:
DPDK provides APIs that enable exposing various statistics from the devices used by DPDK applications and enable exposing
KPI metrics directly from applications. Device statistics include e.g. common statistics available across NICs, like:
received and sent packets, received and sent bytes etc. In addition to this generic statistics, an extended statistics API
is available that allows providing more detailed, driver-specific metrics that are not available as generic statistics.

PR #8883, opened 2021-02-19
New DPDK input plugin

p-zak:
### Required for all PRs:

CCLA completed and signed by Intel Corporation.

## Feature Request
We have created a Telegraf DPDK input plugin which collects metrics exposed by applications built with [Data Plane Development Kit](https://www.dpdk.org/) which is an extensive set of open source libraries designed for accelerating packet processing workloads.

#### The plugin we have created is able to provide telemetry from:
- DPDK NICs (ethdev): 
  - stats - basic device statistics
  - xstats - extended device statistics
  - link_status - up/down link status
- DPDK Raw Devices (rawdev):
  -  xstats - extended device statistics
- DPDK applications (custom, application-specific metrics)

(for more information, including real-life examples of gathered metrics, please refer to `README.md`)

### Use case:
DPDK provides APIs that enable exposing various statistics from the devices used by DPDK applications and enable exposing
KPI metrics directly from applications. Device statistics include e.g. common statistics available across NICs, like:
received and sent packets, received and sent bytes etc. In addition to this generic statistics, an extended statistics API
is available that allows providing more detailed, driver-specific metrics that are not available as generic statistics.

Closes: #8759


plugins/inputs/dpdk/dpdk.go
```diff
@@ -0,0 +1,240 @@
...
+
+  ## Duration that defines how long the connected socket client will wait for a response before terminating connection.
+  ## This includes both writing to and reading from socket.
+  ## Setting the value to 0 disables the timeout (not recommended)
+  # socket_access_timeout = "200ms"
+
+  ## Enables telemetry data collection for selected device types.
+  ## Adding "ethdev" enables collection of telemetry from DPDK NICs (stats, xstats, link_status).
+  ## Adding "rawdev" enables collection of telemetry from DPDK Raw Devices (xstats).
+  device_types = ["ethdev", "rawdev"]
```
srebhan, line 31: If this is the default, please comment the line.

plugins/inputs/dpdk/dpdk.go
```diff
@@ -0,0 +1,240 @@
...
+const (
+	dpdkDescription  = "Reads metrics from DPDK applications using v2 telemetry interface."
+	dpdkSampleConfig = `
+  ## Path to DPDK telemetry socket. This shall point to v2 version of DPDK telemetry interface.
+  # socket_path = "/var/run/dpdk/rte/dpdk_telemetry.v2"
+
+  ## Duration that defines how long the connected socket client will wait for a response before terminating connection.
+  ## This includes both writing to and reading from socket.
+  ## Setting the value to 0 disables the timeout (not recommended)
+  # socket_access_timeout = "200ms"
```
srebhan, line 26: Is this a sensible default that works for "most users"?

plugins/inputs/dpdk/dpdk.go
```diff
@@ -0,0 +1,240 @@
...
+
+  ## Enables telemetry data collection for selected device types.
+  ## Adding "ethdev" enables collection of telemetry from DPDK NICs (stats, xstats, link_status).
+  ## Adding "rawdev" enables collection of telemetry from DPDK Raw Devices (xstats).
+  device_types = ["ethdev", "rawdev"]
+
+  ## List of custom, application-specific telemetry commands to query 
+  ## For e.g. L3 Forwarding with Power Management Sample Application this could be: 
+  ##   additional_commands = ["/l3fwd-power/stats"]
+  # additional_commands = []
```
srebhan, line 36: Can you please add a reference where to find a list of available commands?

plugins/inputs/dpdk/dpdk.go
```diff
@@ -0,0 +1,240 @@
...
+type dpdk struct {
+	SocketPath         string          `toml:"socket_path"`
+	AccessTimeout      config.Duration `toml:"socket_access_timeout"`
+	DeviceTypes        []string        `toml:"device_types"`
+	EthdevConfig       ethdevConfig    `toml:"ethdev"`
+	AdditionalCommands []string        `toml:"additional_commands"`
+	Log                telegraf.Logger `toml:"-"`
+	connector          *dpdkConnector  `toml:"-"`
+	rawdevCommands     []string        `toml:"-"`
+	ethdevCommands     []string        `toml:"-"`
```
srebhan, line 66: Please remove the `toml` annotations as private variables are not processed by default.

plugins/inputs/dpdk/dpdk.go
```diff
@@ -0,0 +1,240 @@
...
+}
+
+type ethdevConfig struct {
+	EthdevExcludeCommands []string `toml:"exclude_commands"`
+}
+
+func init() {
+	inputs.Add(dpdkPluginName, func() telegraf.Input {
+		dpdk := &dpdk{
+			SocketPath:     defaultPathToSocket,
```
srebhan, line 76: No need to specify this here, you set it in `Init()` anyway.

plugins/inputs/dpdk/dpdk.go
```diff
@@ -0,0 +1,240 @@
...
+			AccessTimeout:  defaultAccessTimeout,
+			rawdevCommands: []string{"/rawdev/xstats"},
+			ethdevCommands: []string{"/ethdev/stats", "/ethdev/xstats", "/ethdev/link_status"},
```
srebhan, line 79: Can we please move this to `Init()`!

plugins/inputs/dpdk/dpdk.go
```diff
@@ -0,0 +1,240 @@
...
+	"github.com/influxdata/telegraf/config"
+
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/plugins/inputs"
+	jsonparser "github.com/influxdata/telegraf/plugins/parsers/json"
+)
+
+const (
+	dpdkDescription  = "Reads metrics from DPDK applications using v2 telemetry interface."
+	dpdkSampleConfig = `
```
srebhan, line 19: Please name those `description` and `sampleConfig` to match most other plugins.

plugins/inputs/dpdk/dpdk.go
```diff
@@ -0,0 +1,240 @@
...
+
+func (dpdk *dpdk) Description() string {
+	return dpdkDescription
+}
+
+// Performs validation of all parameters from configuration
+func (dpdk *dpdk) Init() error {
+	if dpdk.SocketPath == "" {
+		dpdk.SocketPath = defaultPathToSocket
+		dpdk.Log.Infof("using default '%v' path for socket_path", defaultPathToSocket)
```
srebhan, line 97: Please make this at most a `Debugf` or remove altogether. Usually there is more than one plugin configured with telegraf and cluttering the log with messages like this is not helpful.

plugins/inputs/dpdk/dpdk.go
```diff
@@ -0,0 +1,240 @@
...
+
+	if err := dpdk.validateCommands(); err != nil {
+		return err
+	}
+
+	if dpdk.AccessTimeout < 0 {
+		return fmt.Errorf("socket_access_timeout should be positive number or equal to 0 (to disable timeouts)")
+	}
+	if len(dpdk.AdditionalCommands) == 0 && len(dpdk.DeviceTypes) == 0 {
+		dpdk.Log.Warn("DPDK plugin is enabled and configured with nothing to read")
```
srebhan, line 111: It's a philosophical question, but does it really make sense to run a configuration like this?

plugins/inputs/dpdk/dpdk.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+
+	dpdk.connector = newDpdkConnector(dpdk.SocketPath, dpdk.AccessTimeout)
+	result, err := dpdk.connector.connect()
+	if result != "" {
+		dpdk.Log.Info(result)
+	}
```
srebhan, line 118: This looks a bit like a debug message left-over. Is the average user really interested in this?

plugins/inputs/dpdk/dpdk_utils.go
```diff
@@ -0,0 +1,123 @@
...
+
+	return result
+}
+
+func uniqueValues(values []string) []string {
+	in := make(map[string]bool)
+	result := make([]string, 0, len(values))
+
+	for _, value := range values {
+		if isAdded := in[value]; !isAdded {
```
srebhan, line 113:
```suggestion
		if ! in[value] {
```

plugins/inputs/dpdk/dpdk.go
```diff
@@ -0,0 +1,240 @@
...
+
+// Gathers all unique commands
+func (dpdk *dpdk) gatherCommands(accumulator telegraf.Accumulator) []string {
+	var commands []string
+	if contains(dpdk.DeviceTypes, "ethdev") {
+		commands = append(commands, dpdk.getCommandsAndParamsCombinations(dpdk.ethdevCommands, dpdk.EthdevConfig.EthdevExcludeCommands, ethdevListCommand, accumulator)...)
+	}
+
+	if contains(dpdk.DeviceTypes, "rawdev") {
+		commands = append(commands, dpdk.getCommandsAndParamsCombinations(dpdk.rawdevCommands, []string{}, rawdevListCommand, accumulator)...)
```
srebhan, line 167: I think you should only pass `acc` to this function as everything else is also reachable from within the function.

plugins/inputs/dpdk/dpdk.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+
+	return nil
+}
+
+// Gathers all unique commands
+func (dpdk *dpdk) gatherCommands(accumulator telegraf.Accumulator) []string {
+	var commands []string
+	if contains(dpdk.DeviceTypes, "ethdev") {
+		commands = append(commands, dpdk.getCommandsAndParamsCombinations(dpdk.ethdevCommands, dpdk.EthdevConfig.EthdevExcludeCommands, ethdevListCommand, accumulator)...)
```
srebhan, line 163: I think you should only pass acc to this function as everything else is also reachable from within the function.

plugins/inputs/dpdk/dpdk.go
```diff
@@ -0,0 +1,240 @@
...
+		}
+	}
+
+	return nil
+}
+
+// Gathers all unique commands and processes each command sequentially
+// Parallel processing could be achieved by running several instances of this plugin with different settings
+func (dpdk *dpdk) Gather(accumulator telegraf.Accumulator) error {
+	commands := dpdk.gatherCommands(accumulator)
```
srebhan, line 150: This should be done in `Init()` as it does not depend on anything related to the gather-cycle.

srebhan, line 150: This should be done in the`Init()` function as it does not depend on anything related to the gather-cycle.

plugins/inputs/dpdk/dpdk.go
```diff
@@ -0,0 +1,240 @@
...
+
+	return uniqueValues(commands)
+}
+
+func (dpdk *dpdk) getCommandsAndParamsCombinations(appendedCommands []string, excludedCommands []string, listCommand string, accumulator telegraf.Accumulator) []string {
+	validCommands := removeSubset(appendedCommands, excludedCommands)
+
+	commands, err := dpdk.appendCommandsWithParamsFromList(listCommand, validCommands)
+	if err != nil {
+		accumulator.AddError(fmt.Errorf("error occurred during fetching of %v params - %v", listCommand, err))
```
srebhan, line 180: Just return the error here and handle it in the function one level up. This way you don't need the `acc` passed down here.

plugins/inputs/dpdk/dpdk.go
```diff
@@ -0,0 +1,240 @@
...
+}
+
+// Gathers all unique commands and processes each command sequentially
+// Parallel processing could be achieved by running several instances of this plugin with different settings
+func (dpdk *dpdk) Gather(accumulator telegraf.Accumulator) error {
+	commands := dpdk.gatherCommands(accumulator)
+
+	for _, command := range commands {
+		dpdk.processCommand(command, accumulator)
+	}
```
srebhan, line 154: Is this required to be executed sequentially?

plugins/inputs/dpdk/dpdk.go
```diff
@@ -0,0 +1,240 @@
...
+
+	commands, err := dpdk.appendCommandsWithParamsFromList(listCommand, validCommands)
+	if err != nil {
+		accumulator.AddError(fmt.Errorf("error occurred during fetching of %v params - %v", listCommand, err))
+	}
+	return commands
+}
+
+// Fetches all identifiers of devices and then creates all possible combinations of commands for each device
+func (dpdk *dpdk) appendCommandsWithParamsFromList(listCommand string, commands []string) ([]string, error) {
```
srebhan, line 186: Please try to follow a consistent parameter ordering. In the functions above the commands come first, but here you apparently switch the order...

plugins/inputs/dpdk/dpdk_utils.go
```diff
@@ -0,0 +1,123 @@
...
+
+	var intArray []int
+	var stringArray []string
+	err = json.Unmarshal(rawMessage[command], &intArray)
+	if err != nil {
+		return nil, fmt.Errorf("failed to unmarshall json response - %v", err)
+	}
+
+	for _, value := range intArray {
+		stringArray = append(stringArray, fmt.Sprintf("%v", value))
```
srebhan, line 77: To be honest I would prefer an explicit conversion like `strconv.FormatInt` or similar here instead of relying on automatic type conversion through `%v`...

plugins/inputs/dpdk/dpdk.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+
+	result := make([]string, 0, len(commands)*len(params))
+	for _, command := range commands {
+		for _, param := range params {
+			result = append(result, commandWithParams(command, param))
+		}
+	}
+
+	return uniqueValues(result), nil
```
srebhan, line 204: I think the commands are already unique, so you could simply make sure the `params` are also unique to ensure unique results.

plugins/inputs/dpdk/dpdk.go
```diff
@@ -0,0 +1,240 @@
...
+}
+
+// Executes command, parses response and creates/writes metric from response
+func (dpdk *dpdk) processCommand(commandWithParams string, accumulator telegraf.Accumulator) {
```
srebhan, line 208: Please name the `telegraf.Accumulator` `acc` and make it always the first parameter.

plugins/inputs/dpdk/dpdk.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+
+	var parsedResponse map[string]interface{}
+	err = json.Unmarshal(buf, &parsedResponse)
+	if err != nil {
+		accumulator.AddError(fmt.Errorf("failed to unmarshall json response from %v command - %v", commandWithParams, err))
+		return
+	}
+
+	command := stripParams(commandWithParams)
```
srebhan, line 222: Hmm wouldn't it make more sense to concatenate `command` and `params` when calling `getCommandResponse` instead of first fuse the two and then split them here again? You could even prepare the params to be either `""` or `",param1..."` and simply do `buf, err := dpdk.connector.getCommandResponse(command+preparedParams)`...

plugins/inputs/dpdk/dpdk_connector.go
```diff
@@ -0,0 +1,145 @@
...
+
+	if initMessage.MaxOutputLen == 0 {
+		return "", fmt.Errorf("failed to read maxOutputLen information")
+	}
+
+	if !conn.messageShowed {
+		conn.maxOutputLen = initMessage.MaxOutputLen
+		conn.messageShowed = true
+		return fmt.Sprintf("Successfully connected to %v running as process with PID %v with len %v", initMessage.Version, initMessage.Pid, initMessage.MaxOutputLen), nil
+	}
```
srebhan, line 142: I think it would be simpler to return the `initMessage` itself and let the caller decide if and how to print the message.

plugins/inputs/dpdk/dpdk_connector.go
```diff
@@ -0,0 +1,145 @@
...
+		messageShowed: false,
+		accessTimeout: time.Duration(accessTimeout),
+	}
+}
+
+// Connects to the socket
+// Since DPDK is a local unix socket, it is instantly returns error or connection, so there's no need to set timeout for it
+func (conn *dpdkConnector) connect() (string, error) {
+	connection, err := net.Dial("unixpacket", conn.pathToSocket)
+
```
srebhan, line 42: Please remove the newline.

plugins/inputs/dpdk/dpdk_connector.go
```diff
@@ -0,0 +1,145 @@
...
+	if err != nil {
+		return "", fmt.Errorf("failed to connect to the socket - %v", err)
+	}
+
+	result, err := conn.readMaxOutputLen(connection)
+	if err != nil {
+		return result, err
+	}
+
+	conn.connection = connection
```
srebhan, line 52: In case `readMaxOutputLen()` for some reason fails, it might become impossible to close the connection as the variable is dropped. Can we please _first_ set `conn.connection` and afterwards read the init message!? Maybe even close the connection in case of an error?

plugins/inputs/dpdk/dpdk_connector.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+// Returns connections, if connection is not created then function tries to recreate it
+func (conn *dpdkConnector) getConnection() (net.Conn, error) {
+	if conn.connection == nil {
+		_, err := conn.connect()
+		if err != nil {
+			return nil, err
+		}
+	}
```
srebhan, line 111: I'm not so sure if I like the reconnect at this level. How about doing this in the `Gather()` function? You could even provide a helper to determine the connection status.

plugins/inputs/dpdk/dpdk_connector.go
```diff
@@ -0,0 +1,145 @@
...
+	closeErr := conn.connection.Close()
+	conn.connection = nil
+	newErr := fmt.Sprintf("failed to %v command - %v", operation, err)
+	if closeErr != nil {
+		return fmt.Errorf("%v and failed to close connection - %v", newErr, closeErr)
+	}
+	return fmt.Errorf(newErr)
+}
+
+func (conn *dpdkConnector) setTimeout(connection net.Conn) error {
```
srebhan, line 97: Why not use `conn.connection`?

plugins/inputs/dpdk/dpdk_connector.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	if messageLength == 0 {
+		return nil, fmt.Errorf("got empty response during execution of '%v' command", fullCommand)
+	}
+	return buf[:messageLength], nil
+}
+
+func (conn *dpdkConnector) tryClose(operation string, err error) error {
+	closeErr := conn.connection.Close()
```
srebhan, line 88: `conn.connection` might be `nil` in case `connect()` was never called (successfully).

plugins/inputs/dpdk/dpdk_connector.go
```diff
@@ -0,0 +1,145 @@
...
+	conn.connection = nil
+	newErr := fmt.Sprintf("failed to %v command - %v", operation, err)
```
srebhan, line 90: Wow this is so wrong. Passing down an `err` here just to reformat it and return it home!? Can we please handle this at the level where this function is called!

plugins/inputs/dpdk/dpdk_test.go
```diff
@@ -0,0 +1,330 @@
...
+		simulateResponse(mockConn, `{"/endpoint1":{"myvalue":[0,1,123]}}`, nil)
+
+		err := dpdk.Gather(mockAcc)
+
+		require.NoError(t, err)
+		require.Equal(t, 0, len(mockAcc.Errors))
+		require.Equal(t, 1, len(mockAcc.Metrics))
+		require.Equal(t, float64(0), mockAcc.Metrics[0].Fields["myvalue_0"])
+		require.Equal(t, float64(1), mockAcc.Metrics[0].Fields["myvalue_1"])
+		require.Equal(t, float64(123), mockAcc.Metrics[0].Fields["myvalue_2"])
```
srebhan, line 304: Please test the whole metric including name and tags. You can do this by defining a reference metric (e.g. call it `expected`) and the compare against your result using `testutil.RequireMetricEqual()` or `testutil.RequireMetricsEqual()`. You probably need to specify the option `IgnoreTime()` as you don't have a deterministic gathering time I think.

plugins/inputs/dpdk/dpdk_utils.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func contains(slice []string, value string) bool {
+	for _, elem := range slice {
+		if value == elem {
+			return true
+		}
+	}
+	return false
+}
```
srebhan, line 90: Please use `internal.choice` which already implements `Contains`.

plugins/inputs/dpdk/dpdk_utils.go
```diff
@@ -0,0 +1,123 @@
...
+
+	var result []string
+	for _, element := range elements {
+		if isExcluded := excludedMap[element]; !isExcluded {
+			result = append(result, element)
+		}
+	}
+
+	return result
+}
```
srebhan, line 106: Please use a `filter.Filter` to determine if an entry matches the subset.

srebhan (changes requested):
Thanks for submitting this plugin. Please check my comments in the code. My major points are

1. Please rethink on which level you handle errors! I think your code can benefit from doing it one level up in most cases.
2. Please use the structure you get passed instead of mixing passing parameter through the receiver struct and the function parameters.
3. Check inverting the `command`/`parameter` approach to only fuse the two when calling the command.
4. Prepare the commands in `Init()` instead of `Gather()`.

plugins/inputs/dpdk/dpdk.go
Hipska: No, that's maybe why there is the warning? Users then can figure out why they don't receive any metrics.

p-zak: Exactly, this is why there is the warning. 
Alternatively, we can return error which will cause Telegraf to fail and exit. But I wouldn't do that for somehow valid configuration.
@srebhan ?

plugins/inputs/dpdk/dpdk.go
p-zak: @srebhan 
I agree that everything except `acc` is also reachable from within `dpdk.getCommandsAndParamsCombinations` function but this was done intentionally -> to not double operations which are done inside `dpdk.getCommandsAndParamsCombinations` function. I would leave it as it is.

However, if you see how these two functions can be improved, I would be more than happy to get meaningful example.

```
func (dpdk *dpdk) gatherCommands(accumulator telegraf.Accumulator) []string {
	var commands []string
	if contains(dpdk.DeviceTypes, "ethdev") {
		commands = append(commands, dpdk.getCommandsAndParamsCombinations(dpdk.ethdevCommands, dpdk.EthdevConfig.EthdevExcludeCommands, ethdevListCommand, accumulator)...)
	}

	if contains(dpdk.DeviceTypes, "rawdev") {
		commands = append(commands, dpdk.getCommandsAndParamsCombinations(dpdk.rawdevCommands, []string{}, rawdevListCommand, accumulator)...)
	}

	commands = append(commands, dpdk.AdditionalCommands...)

	return uniqueValues(commands)
}

func (dpdk *dpdk) getCommandsAndParamsCombinations(appendedCommands []string, excludedCommands []string, listCommand string, accumulator telegraf.Accumulator) []string {
	validCommands := removeSubset(appendedCommands, excludedCommands)

	commands, err := dpdk.appendCommandsWithParamsFromList(listCommand, validCommands)
	if err != nil {
		accumulator.AddError(fmt.Errorf("error occurred during fetching of %v params - %v", listCommand, err))
	}
	return commands
}
```

plugins/inputs/dpdk/dpdk.go
p-zak: @srebhan 
I agree that everything except `acc` is also reachable from within `dpdk.getCommandsAndParamsCombinations` function but this was done intentionally -> to not double operations which are done inside `dpdk.getCommandsAndParamsCombinations` function. I would leave it as it is.

However, if you see how these two functions can be improved, I would be more than happy to get meaningful example.

```
func (dpdk *dpdk) gatherCommands(accumulator telegraf.Accumulator) []string {
	var commands []string
	if contains(dpdk.DeviceTypes, "ethdev") {
		commands = append(commands, dpdk.getCommandsAndParamsCombinations(dpdk.ethdevCommands, dpdk.EthdevConfig.EthdevExcludeCommands, ethdevListCommand, accumulator)...)
	}

	if contains(dpdk.DeviceTypes, "rawdev") {
		commands = append(commands, dpdk.getCommandsAndParamsCombinations(dpdk.rawdevCommands, []string{}, rawdevListCommand, accumulator)...)
	}

	commands = append(commands, dpdk.AdditionalCommands...)

	return uniqueValues(commands)
}

func (dpdk *dpdk) getCommandsAndParamsCombinations(appendedCommands []string, excludedCommands []string, listCommand string, accumulator telegraf.Accumulator) []string {
	validCommands := removeSubset(appendedCommands, excludedCommands)

	commands, err := dpdk.appendCommandsWithParamsFromList(listCommand, validCommands)
	if err != nil {
		accumulator.AddError(fmt.Errorf("error occurred during fetching of %v params - %v", listCommand, err))
	}
	return commands
}
```

plugins/inputs/dpdk/dpdk.go
p-zak: This was done intentionally to simplify code, because before execution of commands we have to leave only unique values. This could be done with approach you mentioned (separated command and params), but it will be much more complicated code than it is now, especially since splitting operation is pretty simple (both in code and in algorithmic complexity).

I wouldn't change it.

plugins/inputs/dpdk/dpdk_connector.go
p-zak: In case `connect()` was never successfully called, `tryClose` function will never be called.

plugins/inputs/dpdk/dpdk_connector.go
p-zak: Connectivity toys were extracted to `dpdk_connector.go` intentionally. This is some kind of "implementation detail" and `Gather()` should not be involved in fixing potential connection issues.

plugins/inputs/dpdk/dpdk.go
p-zak: Unfortunately. it cannot be done in the way you proposed. 
DPDK can be restarted between consecutive `Gather()` cycles which can cause that it will be exposing different set of metrics. This is perfectly valid scenario which is handled in provided approach.

plugins/inputs/dpdk/dpdk.go
p-zak: Connections and timeouts management for parallel approach (with goroutines) would introduce additional complexity we would like to avoid. We tested this approach initially, but this is much easier to understand and work with.

plugins/inputs/dpdk/dpdk.go
srebhan: To be honest, this is now the second or third plugin that allows this and I completely miss the use case. Why should I configure a plugin that will not report any metric in this (valid) config? This is with 99% certainty a configuration error. [Unix rule 12](https://homepage.cs.uri.edu/~thenry/resources/unix_art/ch01s06.html) says: "When you must fail, fail noisily and as soon as possible.". Now we can argue if we need to fail but having a plugin that produces nothing is not very helpful either....
Please enlighten me: What is your use-case to make this a valid config?

plugins/inputs/dpdk/dpdk.go
srebhan: Then why not fuse the two an duplicate the two lines?

plugins/inputs/dpdk/dpdk.go
srebhan: Ok. Please add the above statement as a comment and mark this as resolved.

plugins/inputs/dpdk/dpdk.go
srebhan: Perfectly fine with me. Keep as is and mark as resolved.

plugins/inputs/dpdk/dpdk.go
srebhan: Hmmm ok but I don't see the problem in making the stuff unique. If you really insist, keep it as is... :-)

plugins/inputs/dpdk/dpdk_connector.go
srebhan: Well a lot of plugins make sure they are still connected in the beginning of a `Gather()` cycle.

plugins/inputs/dpdk/dpdk_connector.go
srebhan: That's maybe the current state of afairs, but please also protect our children and grad-children from shooting themselves in the foot! ;-)

zak-pawel: I would definitely want to avoid checking in every unexported function if argument is `nil`. 
It would be different case for exported functions.

p-zak: I would definitely want to avoid checking in every unexported function if argument is `nil`.
It would be different case for exported functions.

srebhan: Really? The one `if` I'm requesting? Handling the connection state is often a difficult thing so I think an extra guard cannot hurt...

plugins/inputs/dpdk/README.md
```diff
@@ -0,0 +1,184 @@
...
+
+  [inputs.dpdk.tags]
+    dpdk_instance = "l2fwd-cat"
+```
+This utilizes Telegraf's standard capability of [adding custom tags](https://github.com/influxdata/telegraf/blob/master/docs/CONFIGURATION.md#input-plugins)
+to input plugin's measurements.
+
+## Metrics
+The DPDK socket accepts `command,params` requests and returns metric data in JSON format. All metrics from DPDK socket
+become flattened using [Telegraf's JSON Flattener](../../parsers/json/README.md) and exposed as fields.
```
reimda, line 145: I have concerns about the metric format in this plugin.  I don't think I like the field names that the json flattener produces.  It's not user friendly to have 80-200 fields with very similar names.  I'm not sure exactly what to do here because I'm not an expert on DPDK, but having so many fields with various prefixes and suffixes doesn't feel right.

Maybe that prefix/suffix information should be exposed as tags?  Something like direction='tx' or 'rx' would cut down on the number of fields in each metric.  (maybe direction isn't the right tag name, but you get the idea)  I see lots of names with i or o, presumably input and output?  It seems like that should be a tag too.

I also see histograms by packet size in the fields, like "rx_size_65_to_127_packets".  Would it make sense to output these more like the histogram aggregator? (https://github.com/influxdata/telegraf/tree/master/plugins/aggregators/histogram)  I could also see people using the prometheus output wanting the size histograms to work with prometheus's histogram metric type.

reimda (changes requested):

plugins/inputs/dpdk/README.md
```diff
@@ -0,0 +1,184 @@
...
+```
+dpdk,host=dpdk-host,dpdk_instance=l3fwd-power,command=/ethdev/stats,params=0 [fields] [timestamp]
+```
+
+| Tag | Description |
+|-----|-------------|
+| `host` | hostname of the machine (consult [Telegraf Agent configuration](https://github.com/influxdata/telegraf/blob/master/docs/CONFIGURATION.md#agent) for additional details) |
+| `dpdk_instance` | custom tag from `[inputs.dpdk.tags]` (optional) | 
+| `command` | executed command (without params) |
+| `params` | command parameter, e.g. for `/ethdev/stats` it is the id of NIC as exposed by `/ethdev/list`<br>For DPDK app that uses 2 NICs the metrics will output e.g. `params=0`, `params=1`. |
```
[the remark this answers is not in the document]
pperycz: Since DPDK allows libraries and applications to [register their own telemetry commands](https://doc.dpdk.org/guides-20.05/prog_guide/telemetry_lib.html#registering-commands) and gives freedom in how to use those params (or not to use them at all), it's not always possible to tell what is the actual nature of the param and translate it into a meaningful name like "nic", "device" or anything else. That's why we went with the approach of keeping it generic.

 

To have any hint on params, the only one place we've encountered is the help message for particular command. When registering a command, one can provide such string that can be later on read using `help` invoked for a command, the example would be:
```
--> /help,/ethdev/xstats
{"/help": {"/ethdev/xstats": "Returns the extended stats for a port. Parameters: int port_id"}}
```
Unfortunately those messags are not in an enforced, structured format that would be common across libraries and applications. If this format was standardized we could parse it and extract some detailed information about the nature of the parameter. But currently using such approach would not be reliable.

plugins/inputs/dpdk/README.md
```diff
@@ -0,0 +1,184 @@
...
+When running plugin configuration below...
+```toml
+[[inputs.dpdk]]
+  device_types = ["ethdev"]
+  additional_commands = ["/l3fwd-power/stats"]
+  [inputs.dpdk.tags]
+    dpdk_instance = "l3fwd-power"
+```
+
+...expected output for `dpdk` plugin instance running on host named `host=dpdk-host`:
```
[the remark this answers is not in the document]
pperycz: The observation is correct, list of fields changes depending on the command being run and also the amount and the naming of metrics vary across different drivers and apps. This is yet another effect of the freedom given by DPDK to the libraries and applications to expose their specific telemetry. Such custom metrics can be exposed from libraries e.g. through `xstats` ("extended statistics" that are by nature customizable) or through [registering its own telemetry commands](https://doc.dpdk.org/guides-20.05/prog_guide/telemetry_lib.html#registering-commands) which makes the set of exposed metrics highly dependent on the driver that is used and on the DPDK-based application. So right now this leaves it to the drivers and DPDK applications to maintain metrics compatibility across versions.

E.g. looking at [feature availability in networking drivers for DPDK](https://doc.dpdk.org/guides/nics/overview.html) we can tell how much differences there are between drivers and such differences are also observed in set of exposed metrics. There are no guarantees that metrics will be exposed in the same way across all the drivers or applications, even if they effectively intend to expose the same piece of information. 

There are ongoing efforts to standardize at least some of the common metrics (e.g. `/ethdev/stats` introduced in DPDK 20.11) but even the DPDK telemetry interface itself is still considered "experimental" with a note that "the output format may change in the future", as stated in [DPDK Telemetry User Guide](https://doc.dpdk.org/guides-20.05/howto/telemetry.html).

So since there's not many constants to currently rely on, we went with generic approach of reading metrics in this plugin and just expose what DPDK-based applications and underlying drivers are exposing.

plugins/inputs/dpdk/README.md
pperycz: Especially for applications that may expose their own metrics through DPDK telemetry interface we have no insights into the structure of those responses since they may contain completely custom metrics. If we want to expose all metrics, even the ones that are available in some nested structures (that we don't know the format of), then those not-so-friendly names are a cost of just having those metrics even collected.

Considering the set of metrics exposed in the example output provided in plugin readme: those come from an Intel driver that exposes quite a bit of detailed NIC metrics, including the packet size buckets that you mentioned. Unfortunately here we also don't have a common naming standard that we can rely on, that is available across all drivers. It means we would not be able to clearly interpret them and e.g. break down into groups of metrics sharing a common set of tags. 

Due to such per-driver, per-vendor and per-application differences in metrics we went with the approach of reducing transformation of metrics to a minimum that is necessary to expose them, and to just provide a set of tags to enhance querying capabilities (host, dpdk_instance, command, param).

plugins/inputs/dpdk/dpdk_connector.go
p-zak: @srebhan 
Ok, I will do that but since you want to check if connection was established (is not nil) in `dpdk_connector.go`, I will not make any changes for your comment regarding checking status of connection in `Gather(...)`. I would like to have connection-related activities in this file.

plugins/inputs/dpdk/dpdk_connector.go
p-zak: I still insist on having connection-related activities in this file (as stated in different comment). Although, I will make small adjustments in neighborhood.

plugins/inputs/dpdk/dpdk.go
srebhan: Ok leave this one as it is.

plugins/inputs/dpdk/dpdk_connector.go
srebhan: Ok.

plugins/inputs/dpdk/dpdk_connector.go
srebhan: Ok I can live with this.

ssoroka (approved):

plugins/inputs/dpdk/README.md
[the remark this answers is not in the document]
reimda: Since DPDK libraries and applications don't provide a standardized format, it's understandable that DPDK has to provide a generic metric format to telegraf.  I'm ok accepting the generic metric format, but I think this should be an area we to try to improve in the future.  Maybe with improvements to json support in telegraf we will be able to come up with a way to use tags and/or metric names instead of field prefixes.

plugins/inputs/dpdk/README.md
reimda: Since DPDK libraries and applications don't provide a standardized format, it's understandable that DPDK has to provide a generic metric format to telegraf. I'm ok accepting the generic metric format, but I think this should be an area we to try to improve in the future. Maybe with improvements to json support in telegraf we will be able to come up with a way to use tags and/or metric names instead of field prefixes.

reimda (approved):

plugins/inputs/dpdk/dpdk.go
p-zak: Done.

plugins/inputs/dpdk/dpdk.go
p-zak: Comment added.

plugins/inputs/dpdk/dpdk.go
p-zak: Added.

plugins/inputs/dpdk/dpdk.go
p-zak: Removed.

plugins/inputs/dpdk/dpdk.go
p-zak: Right! Removed.

plugins/inputs/dpdk/dpdk.go
p-zak: Moved, thank you.

plugins/inputs/dpdk/dpdk.go
p-zak: Renamed.

plugins/inputs/dpdk/dpdk.go
p-zak: Good point, changed to `Debugf`.

plugins/inputs/dpdk/dpdk.go
p-zak: @srebhan You're right. Will fail fast now.

plugins/inputs/dpdk/dpdk.go
p-zak: Changed to debug level.

plugins/inputs/dpdk/dpdk_utils.go
p-zak: Changed.

plugins/inputs/dpdk/dpdk.go
p-zak: Fused as suggested.

plugins/inputs/dpdk/dpdk.go
p-zak: Fused as suggested in previous comment.

plugins/inputs/dpdk/dpdk.go
p-zak: After merging functions (as suggested in previous comments) this situation no longer exists.

plugins/inputs/dpdk/dpdk.go
p-zak: Right now consistency in parameter ordering is preserved.

plugins/inputs/dpdk/dpdk.go
p-zak: Comment added.

plugins/inputs/dpdk/dpdk_utils.go
p-zak: Good point, changed.

plugins/inputs/dpdk/dpdk.go
p-zak: Done.

plugins/inputs/dpdk/dpdk.go
p-zak: Names changed, `acc` always first.

plugins/inputs/dpdk/dpdk.go
p-zak: Thank you, keeping it as is :)

plugins/inputs/dpdk/dpdk_connector.go
p-zak: Right, changed.

plugins/inputs/dpdk/dpdk_connector.go
p-zak: Removed!

plugins/inputs/dpdk/dpdk_connector.go
p-zak: Good catch, fixed.

plugins/inputs/dpdk/dpdk_connector.go
p-zak: Thank you for flexibility!

plugins/inputs/dpdk/dpdk_connector.go
p-zak: Sure!

plugins/inputs/dpdk/dpdk_connector.go
p-zak: Good to hear :)

plugins/inputs/dpdk/dpdk_test.go
p-zak: Nice, rewritten.

plugins/inputs/dpdk/dpdk_utils.go
p-zak: Changed to use `filter.Filter`.

plugins/inputs/dpdk/README.md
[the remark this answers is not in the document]
p-zak: No changes here, as we agreed.

plugins/inputs/dpdk/README.md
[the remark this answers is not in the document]
p-zak: No changes here, as we agreed.

plugins/inputs/dpdk/README.md
p-zak: No changes here right now, as we agreed.

plugins/inputs/dpdk/dpdk_connector.go
p-zak: Handled.

plugins/inputs/dpdk/dpdk.go
```diff
@@ -0,0 +1,263 @@
...
+	if err = isSocket(dpdk.SocketPath); err != nil {
+		return err
+	}
+
+	dpdk.rawdevCommands = []string{"/rawdev/xstats"}
+	dpdk.ethdevCommands = []string{"/ethdev/stats", "/ethdev/xstats", "/ethdev/link_status"}
+
+	if err = dpdk.validateCommands(); err != nil {
+		return err
+	}
```
srebhan, line 119: You can use local errors here
```suggestion
	if err := isSocket(dpdk.SocketPath); err != nil {
		return err
	}

	dpdk.rawdevCommands = []string{"/rawdev/xstats"}
	dpdk.ethdevCommands = []string{"/ethdev/stats", "/ethdev/xstats", "/ethdev/link_status"}

	if err := dpdk.validateCommands(); err != nil {
		return err
	}
```
but as it is only cosmetic, take this as FYI only. :-)

srebhan (approved):
Looks good to me. Please feel free to ignore my comment as it is purely cosmetic.

p-zak:
> Looks good to me. Please feel free to ignore my comment as it is purely cosmetic.

@srebhan I will ignore it this time, but I will remember about it in the future :)

PR #8885, opened 2021-02-19
Fix statsd concurrency bug

### LINKED ISSUE #8695: [inputs.statsd] Tags mangled with tags from other StatsD metrics
After upgrading from Telegraf 1.16.3 to 1.17.0, we've noticed that tags from StatsD metrics are being mangled with unrelated metrics.

### Relevant telegraf.conf:
This may require Graphite templates to be set on the statsd input as the thread safe issue appears to be partially in the `GraphiteParser`
```toml
[[inputs.statsd]]
  templates = ["example.* measurement.host"]
```

### Steps to reproduce:

1. Configure a StatsD input
2. Send a large number of metrics with Influx StatsD style tags (issue is worse at high concurrency when reading StatsD input). For example, send these two metrics a large number of times:
```
test,a=b:1|c
foobar,foo=bar:1|c
```

### Expected behavior:

Something similar to these two metrics are written to the output:
```
test,a=b count=100
foobar,foo=bar count=100
```

### Actual behavior:

Something similar to these metrics are written to the output, where sometimes the tags are swapped:
```
test,a=b count=93
test,foo=bar count=7
foobar,foo=bar count=89
foobar,a=b count=11
```

### Additional info:

Appears to be caused by a locking change in https://github.com/influxdata/telegraf/pull/8509, as I attempted to explain here https://github.com/influxdata/telegraf/pull/8509#issuecomment-760368301

Jagularr:
This bug was inadvertently caused by a small optimization I attempted to make in a previous PR, simply resetting that change here.

Closes #8695


ssoroka (commented):
we're sure this is the cause, and not the cached changes?

ssoroka (approved):

Jagularr:
Updated to keep somewhat of the 'optimization' by adding the lock only to parseName. From my various tests, this works as well.

Jagularr:
!retry-checks

Jagularr:
!retry-checks

Jagularr:
!retry-checks

Jagularr:
!retry-checks

Jagularr:
!retry-checks

PR #8891, opened 2021-02-20
Decompress option on kinesis-consumer

tootedom:
# Description

We are wishing to use the kinesis_consumer input to consume a kinesis stream that is being populated via a cloudwatch logs filter subscription.  When AWS pushes cloudwatch logs to the kinesis destination it is zipping the logs and base64 encoding them.

The AWS sdk for go, is under the hood is automatically base64 decoding the data (a blob) that the kinesis consumer obtains from the kinesis record (r.Data).  The result is a zip []byte that we cannot subsequently consume/manipulate/use in telegraf.

# PR update

Adding of a `decompress` option (default is false for backwards compatibility), to allow the r.Data to be decompressed if the user chooses to.  

This will then allow for the consuming of cloudwatch log data that is streamed to kinesis.

I used the standard go implementation for the uncompress, unsure if you wanted to use something like (I didn't want to add an dependency, without an ok so have left to using the golang standard lib):
- https://github.com/klauspost/compress


srebhan:
!retry-failed

plugins/inputs/kinesis_consumer/README.md
```diff
@@ -54,6 +54,15 @@ and creates metrics using one of the supported [input data formats][].
...
   data_format = "influx"
 
+  ##
+  ## Whether or not to uncompress the data from kinesis
+  ## If you are processing a cloudwatch logs kinesis stream then set this to true
+  ## as AWS compresses cloudwatch log data before it is sent to kinesis (aws
+  ## also base64 encodes the zip byte data before pushing to the stream.  The base64 decoding
+  ## is done automatically by the golang sdk, as data is read from kinesis)
+  ##
+  # decompress = false
```
srebhan, line 64: Does it make sense to allow to specify the compression algorithm in order to be more flexible? What if they decide to use `snappy` or similar in the future?

srebhan (commented):
Code looks good, but it might be beneficial to allow the user to specify the decompression/compression algorithm (i.e. `"gzip"` for now) to be future proof. What do you think?

plugins/inputs/kinesis_consumer/README.md
tootelldels: > Code looks good, but it might be beneficial to allow the user to specify the decompression/compression algorithm (i.e. `"gzip"` for now) to be future proof. What do you think?

Sounds sensible.  How's something like the following sound:

`decompress: "gzip"`

snappy could requested like the following, i.e:

`decompress: "snappy"`

I can look at adding in the support for snappy at the same time (would be an extra dependency I think).

srebhan: Sounds exactly like what I had in mind. :-) You don't need to add snappy now but I like to be prepared. :-) So only having `"gzip"` or `""` is fine for me. If we hit other compression types we can also add later. But in case you are already aware of a use-case, you can of course add it in one go.

srebhan: Oh and please use `internal.choice` for testing the valid values... ;-)

tootedom: Let me know how you feel about this change to `decompress_with`.  I've added `zlib` as an option; but left out the snappy for now (extra dependency required to do the snappy; whereas zlib is standard in golang).   I could always look at sending an additional PR for snappy.  I added a test for the gzip/zlib/none options.

Happy to make changes.  Thanks for the help.

tootedom:
!retry-failed

plugins/inputs/kinesis_consumer/README.md
tootedom: I've tried doing a !retry-failed; but it did not seem to make the CircleCI happier.   Unfortunately I can't see what the output of the circle ci tests are; would you be able to provide me the details?   Locally I'm unable to see failures related to the PR.  (The permissions CircleCI is wanting are very permissive; and is wanting access to all of my repos; private and public).

plugins/inputs/kinesis_consumer/README.md
```diff
@@ -54,6 +54,15 @@ and creates metrics using one of the supported [input data formats][].
...
   data_format = "influx"
 
+  ##
+  ## Whether or not to uncompress the data from kinesis
+  ## If you are processing a cloudwatch logs kinesis stream then set this to "gzip"
+  ## as AWS compresses cloudwatch log data before it is sent to kinesis (aws
+  ## also base64 encodes the zip byte data before pushing to the stream.  The base64 decoding
+  ## is done automatically by the golang sdk, as data is read from kinesis)
+  ##
+  # decompress_with = "none"
```
srebhan, line 64: TBH the name is a bit bulky. I think we should either stick with `decompress` or `compression`.

plugins/inputs/kinesis_consumer/kinesis_consumer.go
```diff
@@ -66,8 +72,15 @@ type (
 
 const (
 	defaultMaxUndeliveredMessages = 1000
+	gzipDecompression             = "gzip"
+	noDecompression               = "none"
+	zlibDecompression             = "zlib"
 )
 
+var supportedDecompressionAlgorithms = []string{gzipDecompression, noDecompression, zlibDecompression}
```
srebhan, line 80: Please move this to `Init()` instead of declaring a global variable and directly define the list instead of using the constants.

plugins/inputs/kinesis_consumer/kinesis_consumer.go
```diff
@@ -238,8 +260,47 @@ func (k *KinesisConsumer) Start(ac telegraf.Accumulator) error {
 	return nil
 }
 
+func gzipDecompress(data []byte) ([]byte, error) {
```
srebhan, line 263: Only cosmetics, but can you please change the name to `decompressGzip()` to match the other names starting with an action...

plugins/inputs/kinesis_consumer/kinesis_consumer.go
```diff
@@ -334,6 +395,14 @@ func (k *KinesisConsumer) Set(streamName, shardID, sequenceNumber string) error
 	return nil
 }
 
+func (k *KinesisConsumer) Init() error {
+	err := choice.Check(k.DecompressWith, supportedDecompressionAlgorithms)
+	if err != nil {
+		return fmt.Errorf(`cannot verify "decompress_with" setting: %v`, err)
```
srebhan, line 401: You also might want to log the valid options for the user.

plugins/inputs/kinesis_consumer/kinesis_consumer.go
```diff
@@ -118,6 +131,15 @@ var sampleConfig = `
...
   data_format = "influx"
 
+  ##
+  ## Whether or not to uncompress the data from kinesis
+  ## If you are processing a cloudwatch logs kinesis stream then set this to "gzip"
+  ## as AWS compresses cloudwatch log data before it is sent to kinesis (aws
+  ## also base64 encodes the zip byte data before pushing to the stream.  The base64 decoding
+  ## is done automatically by the golang sdk, as data is read from kinesis)
+  ##
+  # decompress_with = "none"
```
srebhan, line 141: How about making `""` a synonym for `none`? This way you do get the default for free. :-)

plugins/inputs/kinesis_consumer/kinesis_consumer.go
```diff
@@ -238,8 +260,47 @@ func (k *KinesisConsumer) Start(ac telegraf.Accumulator) error {
...
+	default:
+		return func(data []byte) ([]byte, error) {
+			return data, nil
+		}
+	}
+}
+
+func (k *KinesisConsumer) getRecordData(r *consumer.Record) ([]byte, error) {
+	return k.getDecompressionFunc(k.DecompressWith)(r.Data)
+}
```
srebhan, line 296: How about determining the decompression function once in `Init()` and store it in `KinesisConsumer`, e.g. as `decompress`? Then call `k.decompress(r.Data)` in `onMessage()`...

srebhan (commented):
Nice. I have some minor comments to simplify the code a bit more (hopefullly). :-) Oh and please fix the formatting issues found by CI...

plugins/inputs/kinesis_consumer/kinesis_consumer.go
```diff
@@ -334,6 +357,50 @@ func (k *KinesisConsumer) Set(streamName, shardID, sequenceNumber string) error
...
+
+	switch k.DecompressionType {
+	case gzipDecompression:
+		k.decompressionFunc = decompressGzip
+	case zlibDecompression:
+		k.decompressionFunc = decompressZlib
+	default:
+		k.decompressionFunc = decompressNoOp
+	}
+	return nil
```
srebhan, line 397: We can do:
```suggestion
	switch k.DecompressionType {
	case "gzip":
		k.decompressionFunc = decompressGzip
	case "zlib":
		k.decompressionFunc = decompressZlib
	case "none", "":
		k.decompressionFunc = decompressNoOp
	default:
		return fmt.Errorf("unknown decompression %q", k.DecompressionType)
	}
	return nil
```

srebhan (commented):
Still some small potential for beauty (at least in my view)... ;-) See my comment.

plugins/inputs/kinesis_consumer/kinesis_consumer_test.go
```diff
@@ -0,0 +1,140 @@
...
+
+type TestTrackingAccumulator struct {
+	telegraf.TrackingAccumulator
+	Metrics *[]telegraf.Metric
+}
+
+func (t TestTrackingAccumulator) AddTrackingMetricGroup(group []telegraf.Metric) telegraf.TrackingID {
+	*t.Metrics = append(*t.Metrics, group...)
+	return 1
+}
```
srebhan, line 22: Please use the tracking accumulator provided by [`testutil`](https://github.com/influxdata/telegraf/blob/8a6907a18625a760e5f0f3d95016c6198c83b76b/testutil/accumulator.go#L200).

srebhan (changes requested):
Use the tracking accumulator provided by [`testutil](https://github.com/influxdata/telegraf/blob/8a6907a18625a760e5f0f3d95016c6198c83b76b/testutil/accumulator.go#L200).

srebhan (approved):
Looks good to me.

srebhan:
!retry-failed

reimda:
Is decompression something the SDKs we use should do?  It seems like if decoding base64 is handled for us decompression should be too.  I noticed that aws-sdk-go has a new v2 release.  Is it included there?

I like how the new setting lets the user choose a compression algorithm.  Many of the existing plugins have a setting just like this but they call it content_encoding.  For example, the influxdb_v2 output, amqp_consumer input.  Even though this plugin doesn't use http directly, I think it would make sense to call the setting content_encoding here too.

tootedom:
Hi there,

> Many of the existing plugins have a setting just like this but they call it content_encoding. For example, the influxdb_v2 output, amqp_consumer input. Even though this plugin doesn't use http directly, I think it would make sense to call the setting content_encoding here too.

Yup, makes sense to keep a consistent language across the plugins.  I'm happy to update to call it `content_encoding`.  Let me know if you all agree, and I'll make that change.

----

> Is decompression something the SDKs we use should do? It seems like if decoding base64 is handled for us decompression should be too. I noticed that aws-sdk-go has a new v2 release. Is it included there?

I do not know if the v2 sdk performs the auto base64 decoding like that of the v1 (https://github.com/aws/aws-sdk-go/blob/master/service/kinesis/api.go#L7022).   

Also unknown I'm unsure if the v2 sdk would perform auto decompression of compressed data (my gut "guess" would be that it doesn't)

However, I'm thinking that that a SDK update for this plugin, isn't for this PR to implement; and would be best placed for a separate isolated upgrade PR (It may take a complete re-implementation).

reimda:
>  I'm happy to update to call it `content_encoding`. Let me know if you all agree, and I'll make that change.

Let's go ahead and rename it to content_encoding.  I'll merge it when this is complete

plugins/inputs/kinesis_consumer/kinesis_consumer_test.go
```diff
@@ -0,0 +1,178 @@
...
+			for _, metric := range acc.Metrics {
+				if logEventMessage, ok := metric.Fields["message"]; ok {
+					assert.Contains(t, logEventMessage.(string), tt.expected.messageContains)
+				} else {
+					t.Errorf("Expect logEvents to be present")
+				}
+			}
+		})
+	}
+
```
reimda, line 177: @tootedom I think the linter is complaining about this empty line.  We need CI to pass before we can merge, including the linter.
```suggestion
```

reimda (changes requested):

tootedom:
!retry-failed

plugins/inputs/kinesis_consumer/kinesis_consumer_test.go
tootedom: Many thanks for the pointer.  I was failing miserably at installing revive locally (still not fathomed it).  
Have removed the line, and pushed.  Linter looks happy.

reimda (approved):

PR #8895, opened 2021-02-22
Add draft version of timer based ticker used when round_interval = true

goslly:
### Required for all PRs:


sspaink:
Closing, looks like this might have been a mistake seeing it is only a commit from 2020.

PR #8900, opened 2021-02-22
add support for datastreams and elastic common schema doc format

michalg91:
### Required for all PRs:

### Changes:
- Added Elastic Common Schema format for metrics
- Added a way to use datastreams
- Added "metric_name" replacer in index name to allow writes to indices based on telegraf input name.

### DIFF
--- a/plugins/outputs/elasticsearch/elasticsearch.go
+++ b/plugins/outputs/elasticsearch/elasticsearch.go
@@ -1,6 +1,7 @@
 package elasticsearch
 
 import (
+	"os"
 	"bytes"
 	"context"
 	"fmt"
@@ -10,6 +11,7 @@ import (
 	"strings"
 	"text/template"
 	"time"
+	"net"
 
 	"crypto/sha256"
 
@@ -17,12 +19,14 @@ import (
 	"github.com/influxdata/telegraf/internal"
 	"github.com/influxdata/telegraf/plugins/common/tls"
 	"github.com/influxdata/telegraf/plugins/outputs"
-	"gopkg.in/olivere/elastic.v5"
+	"github.com/elastic/go-sysinfo"
+	"github.com/olivere/elastic/v7"
 )
 
 type Elasticsearch struct {
 	URLs                []string `toml:"urls"`
 	IndexName           string
+	IsDataStream        bool `toml:"datastream"`
 	DefaultTagValue     string
 	TagKeys             []string
 	Username            string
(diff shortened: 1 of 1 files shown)


gpolek (approved):

plugins/outputs/elasticsearch/elasticsearch.go
```diff
@@ -92,6 +97,14 @@ var sampleConfig = `
...
   ## it will enable data resend and update metric points avoiding duplicated metrics with diferent id's
   force_document_id = false
+
+  ## Enable Elastic Common Schema format
+  ## Set to true if you want to format output document with ECS standard
+  use_ecs = false
+  
+  ## Use datastream
+  ## Enable to save documents as datastream
+  datastream = false
```
reimda, line 107: Changes to sampleConfig need to be made in README.md too.

plugins/outputs/elasticsearch/elasticsearch.go
```diff
@@ -411,6 +450,13 @@ func (a *Elasticsearch) GetIndexName(indexName string, eventTime time.Time, tagK
 		indexName = dateReplacer.Replace(indexName)
 	}
 
+	if strings.Contains(indexName, "metric_name") {
+		var metricNameReplacer = strings.NewReplacer(
+			"metric_name", metricName,
+		)
+		indexName = metricNameReplacer.Replace(indexName)
+	}
+
```
reimda, line 459: This needs to be documented in the README.md. I see you mentioned it in the PR description but that won't be available to the end user.

plugins/outputs/elasticsearch/elasticsearch.go
```diff
@@ -262,28 +275,54 @@ func GetPointID(m telegraf.Metric) string {
...
+		} else {
+			t := make(map[string]interface{})
+			t["tags"] = metric.Tags()
+			t["metrics"] = map[string]interface{}{name: metric.Fields()}
+			m["telegraf"] = t
+			m["metricset"] = map[string]interface{}{"name": name}
+			m["event"] = ecs_event
+			m["host"] = ecs_host
+			m["agent"] = ecs_agent
+			m["ecs"] = map[string]interface{}{"version": "1.7.0"}
```
reimda, line 317: Is there a way to add a unit test for this new common schema format? How much would we need to mock of the elasticsearch client api?

reimda (changes requested):
Thanks for the PR, @michalg91!

smalenfant:
I was just about to use data stream and looks like something is in the works for Telegraf. I think all we needed at first is to set the indexing to create. That would actually let us index data without managing the whole datastream vs ecs. 

https://www.elastic.co/guide/en/elasticsearch/reference/current/set-up-a-data-stream.html#create-data-stream

sspaink:

PR #8911, opened 2021-02-26
Support float64 in enum processor

tuannh99:
This adds float64 support to enum processor.

### Required for all PRs:

### DIFF
--- a/plugins/processors/enum/README.md
+++ b/plugins/processors/enum/README.md
@@ -2,7 +2,7 @@
 
 The Enum Processor allows the configuration of value mappings for metric tags or fields.
 The main use-case for this is to rewrite status codes such as _red_, _amber_ and
-_green_ by numeric values such as 0, 1, 2. The plugin supports string, int and bool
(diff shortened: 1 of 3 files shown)


srebhan:
@tuannh99 while the code looks good I wonder what your use case is. As far as I get it you want to replace a float value with something else!?
However, float values are notoriously difficult to compare and your are trying to convert a dense quantity into a categorical value which is ill-defined anyway without quantization.
To cut a long story short, please outline your use-case! If you really need this I would expect at least an option `float_precision`  and maybe additionally `float_rounding` which can be `floor`, `ceil` and `round` at least...

srebhan:
@tuannh99 any news on this?

tuannh99:
@srebhan I should have a commit to update this in a few.

srebhan:
@tuannh99 can you please comment on my concern above that trying to map a raw floating point value is very fragile and explain what your use-case is!

ssoroka (approved):

ssoroka:
This is fine as is. In a future PR we may want to expose the FormatFloat() precision argument, to give users the ability to control how precisely the floats are matched.

PR #8912, opened 2021-02-26
Add Predicates on valuecounter to emit only fields whose aggregated value match the predicate(s)

tootedom:
# Description

The valuecounter plugin counts the occurrence of values in fields and emits the counter once every 'period' seconds.
We wish to add the ability to only emit the counter if it is above a certain value.  

An example use: aggregating the number of "errors" generated by applications, and outputting the list of those applications who have been erring above a threshold.

# PR

Have added the ability to specify one of more predicates on the valuecounter.  These are evaluated in the Push.
Have included predicates for greater_than, less_than, equal_to and not_equal_to.

### DIFF
--- a/plugins/aggregators/valuecounter/valuecounter.go
+++ b/plugins/aggregators/valuecounter/valuecounter.go
@@ -14,19 +14,19 @@ type aggregate struct {
 	fieldCount map[string]int
 }
 
-type predicate func(int) bool
+type predicate func(int, int) bool
 
 type Predicate struct {
-	Type string
-	value int
+	Type          string `toml:"type"`
+	Value         int    `toml:"value"`
 	predicateFunc predicate
 }
 
 // ValueCounter an aggregation plugin
 type ValueCounter struct {
-	cache  map[uint64]aggregate
-	Fields []string
-	Predicates []*Predicate `toml:predicate`
+	cache      map[uint64]aggregate
+	Fields     []string
+	Predicates []*Predicate `toml:"predicate"`
 }
 
 // NewValueCounter create a new aggregation plugin which counts the occurrences
@@ -47,9 +47,9 @@ var sampleConfig = `
   ## The fields for which the values will be counted
   fields = []
   ## Only include fields for which the predicate is true
-  [aggregators.valuecounter.predicate]
+  [[aggregators.valuecounter.predicate]]
 	type = "greater_than"
-	value = 10
+	value = 0
  
 `
 
@@ -97,7 +97,7 @@ func (vc *ValueCounter) Push(acc telegraf.Accumulator) {
 		for field, count := range agg.fieldCount {
 			var matched = 0
 			for _, predicate := range vc.Predicates {
-				if predicate.predicateFunc(count) {
+				if predicate.predicateFunc(count, predicate.Value) {
 					matched++
 				}
 			}
@@ -130,9 +130,9 @@ func (vc *ValueCounter) configurePredicates() error {
 	for _, predicate := range vc.Predicates {
 		switch predicate.Type {
 		case greaterThan:
-			predicate.predicateFunc = func(count int) bool { return count > predicate.value }
+			predicate.predicateFunc = func(count int, compareAgainst int) bool { return count > compareAgainst }
 		case lessThan:
-			predicate.predicateFunc = func(count int) bool { return count < predicate.value }
+			predicate.predicateFunc = func(count int, compareAgainst int) bool { return count < compareAgainst }
 		}
 	}
 	return nil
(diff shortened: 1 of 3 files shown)


plugins/aggregators/valuecounter/valuecounter.go
```diff
@@ -92,6 +116,46 @@ func (vc *ValueCounter) Reset() {
...
+	keys := make([]string, 0, len(predicates))
+	for k := range predicates {
+		keys = append(keys, k)
+	}
+
+	return predicates, keys
+}
+
+func (vc *ValueCounter) configurePredicates() error {
+
```
Jagularr, line 136: extra white space here

plugins/aggregators/valuecounter/valuecounter.go
```diff
@@ -80,7 +95,16 @@ func (vc *ValueCounter) Push(acc telegraf.Accumulator) {
...
+			var matched = 0
+			for _, predicate := range vc.Predicates {
+				if predicate.predicateFunc(count, predicate.Value) {
+					matched++
+				}
+			}
+
+			if matched == len(vc.Predicates) {
+				fields[field] = count
+			}
```
Jagularr, line 107: the above block can be simplified by removing `matched`, you could put a `label` for the field loop such as "FieldCount:" and then you could leave the assignment at the end without an if statement wrapping it and just put a `
if !predicate.predicateFunc(count, predicate.Value) {
	continue FieldCount
}
` on line 100. That way time isn't spent doing checks even after one failed.

Jagularr (commented):
Cool idea maybe minor refractorings

plugins/aggregators/valuecounter/valuecounter.go
tootedom: Great suggestion.   Hadn't even crossed my mind.
cheers
/dom

tootedom: !retry-failed

tootedom:
!retry-failed

plugins/aggregators/valuecounter/valuecounter.go
tootedom: Apologies I forgot to add a note here.  I made the suggested changes.  Let me know if this wasn't matching what you had in mind; happy to update if needs.

cheers
/dom

reimda:
This looks useful, but I'm not sure valuecounter is the right place for the predicate to be. Usually when we get requests to add conditionals like greater_than, less_than, etc to a plugin we recommend using the starlark processor. A small script expresses logic in a much more natural way than a new toml config section does.

Have you considered using starlark to do this?

tootedom:
> This looks useful, but I'm not sure valuecounter is the right place for the predicate to be. Usually when we get requests to add conditionals like greater_than, less_than, etc to a plugin we recommend using the starlark processor. A small script expresses logic in a much more natural way than a new toml config section does.
> 
> Have you considered using starlark to do this?

Hi there,

I'd not actually noticed the existence of starlark.  Reading the docs and blog on the this processor and agreed it does sound like it would work.  I'll shall have a take a look at it within a configuration and will post back on how I get on.

Many Thanks
/dom

tootedom:
> > This looks useful, but I'm not sure valuecounter is the right place for the predicate to be. Usually when we get requests to add conditionals like greater_than, less_than, etc to a plugin we recommend using the starlark processor. A small script expresses logic in a much more natural way than a new toml config section does.
> > Have you considered using starlark to do this?
> 
> Hi there,
> 
> I'd not actually noticed the existence of starlark. Reading the docs and blog on the this processor and agreed it does sound like it would work. I'll shall have a take a look at it within a configuration and will post back on how I get on.
> 
> Many Thanks
> /dom

Hi there,

Apologies for the delay in responding.  I managed to verify that the `starlark` processor will indeed meet the requirements. 
 This means this PR to add predicate functionality to the `valuecounter` can be achieved by using the current starlark processor.   

Feel free to close the PR as won't do; can be done using starlark.

Here is an example below (the metric deepcopy may or may not be appropriate based on the context in which the example is used) that removes any aggregated fields whose values are 600 or less (for anyone who is searching in the future).

```
[[processors.starlark]]
    namepass = ["incoming_web_log_aggregates"]
    source = '''
def apply(metric):
    m2 = deepcopy(metric)
    m2.name = "incoming_web_log_aggregates_filtered"
    for agg, val in metric.fields.items():
        if val <= 600:
            m2.fields.pop(agg,0)
    return [metric, m2]
'''
```

reimda:
Thanks @tootedom, I'm glad the starlark solution is working for you. Thanks for posting it for others!

PR #3537, opened 2017-12-03
New input jboss plugin

toni-moreno:
Resolves #824 

### Required for all PRs:

- [X] Signed [CLA](https://influxdata.com/community/cla/).
- [X] Associated README.md updated.
- [X] Has appropriate unit tests.

### New JBoss Plugin 

Based on the great work from @stefa975 in (https://github.com/stefa975/telegraf/tree/master/plugins/inputs/jboss )  fix  (https://github.com/influxdata/telegraf/issues/824)  with some fixes and improvements.

Tested on standalone and Domain Based JBoss 6.X and 7.X on working ok on producction servers.

PR #8926, opened 2021-03-02
New input jboss plugin

### LINKED ISSUE #824: add Jboss input plugin
I'm so interesting on TICK platform!
Working on enterprise and open platform, will be so interesting monitoring application stack objects.
One of this is Jboss/Wildfly.
Just try with httpjson plugin, but think that a dedicated input plugin like on Diamond will be more powerful.

I am also available for development and testing.

stefa975:
Resolves #824

- [x]  Signed CLA.

New JBoss Plugin

Based on @toni-moreno PR (https://github.com/influxdata/telegraf/pull/3537) 
and rebased to the latest master version.

Tested on standalone and Domain Based JBoss 6.X and 7.X on working ok on producction servers.


stefa975:
@dependabot rebase

stefa975:
!signed-cla

docs/LICENSE_OF_DEPENDENCIES.md
```diff
@@ -132,6 +132,7 @@ following works:
 - github.com/shirou/gopsutil [BSD 3-Clause Clear License](https://github.com/shirou/gopsutil/blob/master/LICENSE)
 - github.com/sirupsen/logrus [MIT License](https://github.com/sirupsen/logrus/blob/master/LICENSE)
 - github.com/soniah/gosnmp [BSD 2-Clause "Simplified" License](https://github.com/soniah/gosnmp/blob/master/LICENSE)
+- github.com/stefa975/go-http-digest-auth-client [BSD 2-Clause "Simplified" License](https://github.com/stefa975/go-http-digest-auth-client/blob/master/LICENSE.md)
```
srebhan, line 135: Any reason you don't use the "upstream" version at https://github.com/xinsnake/go-http-digest-auth-client?

plugins/inputs/jboss/README.md
```diff
@@ -0,0 +1,168 @@
+# JBoss plugin
+
+The JBoss plugin can collect data from JBoss management API.
+
+Plugin currently support JBoss Application server in domain modes:-
```
srebhan, line 5: You might want to be a bit more verbose in the README... At least:
```suggestion
Currently, the plugin supports JBoss Application servers in the following domain modes:
```

plugins/inputs/jboss/jboss.go
```diff
@@ -0,0 +1,1312 @@
...
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/internal"
+	"github.com/influxdata/telegraf/plugins/common/tls"
+	"github.com/influxdata/telegraf/plugins/inputs"
+	dac "github.com/stefa975/go-http-digest-auth-client"
```
srebhan, line 20: Any reason you don't use the "upstream" version at https://github.com/xinsnake/go-http-digest-auth-client?

plugins/inputs/jboss/jboss.go
```diff
@@ -0,0 +1,1312 @@
...
+)
+
+// GetHosts constan applied to jboss management query types
```
srebhan, line 23:
```suggestion
// GetHosts constants applied to jboss management query types
```

plugins/inputs/jboss/jboss.go
```diff
@@ -0,0 +1,1312 @@
...
+	ExpiredSessions   string                 `json:"expired-sessions"`
+	MaxActiveSessions string                 `json:"max-active-sessions"`
+	SessionsCreated   string                 `json:"sessions-created"`
+	Servlet           map[string]interface{} `json:"servlet"`
+}
+
+// JBoss the main collectod struct
+type JBoss struct {
+	Servers []string
+	Metrics []string
```
srebhan, line 176: Please add `toml` annotations
```suggestion
	Servers []string `toml:"servers"`
	Metrics []string `toml:"metrics"`
```

Same for the other config options...

plugins/inputs/jboss/jboss.go
```diff
@@ -0,0 +1,1312 @@
...
+
+// SetHTTPClient set http client
+func (c *RealHTTPClient) SetHTTPClient(client *http.Client) {
+	c.client = client
+}
+
+// HTTPClient return the HTTP client
+func (c *RealHTTPClient) HTTPClient() *http.Client {
+	return c.client
+}
```
srebhan, line 224: Why do you need this indirection and not use the code directly?

plugins/inputs/jboss/jboss.go
```diff
@@ -0,0 +1,1312 @@
...
+	return "Telegraf plugin for gathering metrics from JBoss AS"
+}
+
+// Gather Gathers data for all servers.
+func (h *JBoss) Gather(acc telegraf.Accumulator) error {
+
+	if h.ResponseTimeout.Duration < time.Second {
+		h.ResponseTimeout.Duration = time.Second * 5
+	}
+
```
srebhan, line 272: This will not allow to specify a timeout of zero aka. no timeout. Please set the default timeout in `init()` when initializing the structure.

plugins/inputs/jboss/jboss.go
```diff
@@ -0,0 +1,1312 @@
...
+}
+
+// Description just returns a short description of the JBoss plugin
+func (*JBoss) Description() string {
+	return "Telegraf plugin for gathering metrics from JBoss AS"
+}
+
+// Gather Gathers data for all servers.
+func (h *JBoss) Gather(acc telegraf.Accumulator) error {
+
```
srebhan, line 268: Avoid newlines if they don't help to visually structure the code.

plugins/inputs/jboss/jboss.go
```diff
@@ -0,0 +1,1312 @@
...
+		}
+		tr := &http.Transport{
+			TLSClientConfig: tlsCfg,
+		}
+		client := &http.Client{
+			Transport: tr,
+			Timeout:   h.ResponseTimeout.Duration,
+		}
+		h.client.SetHTTPClient(client)
+	}
```
srebhan, line 287: Please move this to an `Init() error` function that is automatically called by telegraf once all config parameters are parsed and set but before the first gather. This avoids doing the same init code over and over and also helps the reader to understand that it is intended to be done once.

plugins/inputs/jboss/jboss.go
```diff
@@ -0,0 +1,1312 @@
...
+
+	var wg sync.WaitGroup
+	for _, server := range h.Servers {
+
+		//Check Exec Mode for each servers
+
+		bodyContent, err := h.createRequestBody(GetExecStat, nil)
+		if err != nil {
+			acc.AddError(err)
+		}
```
srebhan, line 297: How about preparing all the request bodies in the `Init()` function mentioned above. You could either use a `map` for storing them or use explicit variables. This also allows you to get rid of the `OrderedMap` implementation and reduces the overall number of lines.

plugins/inputs/jboss/jboss.go
```diff
@@ -0,0 +1,1312 @@
+package jboss
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"log"
```
srebhan, line 8: Please don't use `log` but the internal `telegraf.Logger` by adding 
```golang
```
to your plugin struct and the use `h.Log.Debugf()` etc.

plugins/inputs/jboss/jboss.go
```diff
@@ -0,0 +1,1312 @@
...
+		bodyContent, err := h.createRequestBody(GetExecStat, nil)
+		if err != nil {
+			acc.AddError(err)
+		}
+
+		out, err := h.doRequest(server, bodyContent)
+		if err != nil {
+			log.Printf("E! JBoss Error handling ExecMode Test: %s\n", err)
+			acc.AddError(err)
+		}
```
srebhan, line 303: If you go for parallelism (which makes sense), please also move the request to the goroutine.

plugins/inputs/jboss/jboss.go
```diff
@@ -0,0 +1,1312 @@
...
+		// Unmarshal json
+		exec := ExecTypeResponse{}
+		if err = json.Unmarshal(out, &exec); err != nil {
+			acc.AddError(fmt.Errorf("Error decoding JSON response (ExecTypeResponse) %s,%s", out, err))
```
srebhan, line 307: Errors should start lowercase. Maybe
```suggestion
			acc.AddError(fmt.Errorf("decoding JSON response (ExecTypeResponse) %q failed: %v", out, err))
```

plugins/inputs/jboss/jboss.go
```diff
@@ -0,0 +1,1312 @@
...
+			return nil
+		}
+		var execAsDomain bool
+		var isEAP7 bool
+		if exec.Result["launch-type"] == "DOMAIN" {
+			execAsDomain = true
+		} else {
+			execAsDomain = false
+		}
+		isEAP7 = isEAP7Version(exec.Result["product-name"].(string), exec.Result["product-version"].(string))
```
srebhan, line 317: Please use the power of go
```suggestion
		execAsDomain := exec.Result["launch-type"] == "DOMAIN"
		isEAP7 := isEAP7Version(exec.Result["product-name"].(string), exec.Result["product-version"].(string))
```

plugins/inputs/jboss/jboss.go
```diff
@@ -0,0 +1,1312 @@
...
+		log.Printf("D! JBoss Plugin Working as Domain: %t EAP7: %t  for server %s \n", execAsDomain, isEAP7, server)
```
srebhan, line 318: Please use the telegraf Logging facility instead.

plugins/inputs/jboss/jboss.go
```diff
@@ -0,0 +1,1312 @@
...
+		out, err := h.doRequest(server, bodyContent)
+		if err != nil {
+			log.Printf("E! JBoss Error handling ExecMode Test: %s\n", err)
+			acc.AddError(err)
+		}
+		// Unmarshal json
+		exec := ExecTypeResponse{}
+		if err = json.Unmarshal(out, &exec); err != nil {
+			acc.AddError(fmt.Errorf("Error decoding JSON response (ExecTypeResponse) %s,%s", out, err))
+			return nil
```
srebhan, line 308: Well there was an error, why returning `nil`? In case of multiple servers you will skip all servers after the first one failing!?!? Maybe just `continue` the loop over the servers?

plugins/inputs/jboss/jboss.go
```diff
@@ -0,0 +1,1312 @@
...
+// Returns:
+//     error: Any error that may have occurred
+
+func isEAP7Version(productName string, productVersion string) bool {
+	log.Printf("D! JBoss product: %s version %s", productName, productVersion)
+	if strings.Contains(productName, "EAP") {
+		return strings.HasPrefix(productVersion, "7.")
+	} else {
+		return strings.HasPrefix(productVersion, "10.")
+	}
```
srebhan, line 372:
```suggestion
	if strings.Contains(productName, "EAP") {
		return strings.HasPrefix(productVersion, "7.")
	}
	return strings.HasPrefix(productVersion, "10.")
```

srebhan: Furthermore, this function is a four-liner only used once. Maybe just move the function body to the respective location and get rid of the function...

plugins/inputs/jboss/jboss.go
```diff
@@ -0,0 +1,1312 @@
...
+	serverURL string,
+	execAsDomain bool,
+	isEAP7 bool,
+	hosts []string,
+) error {
+	var wg sync.WaitGroup
+
+	for _, host := range hosts {
+		wg.Add(1)
+		go func(host string) {
```
srebhan, line 395: Is it really worth to get the servers in parallel?

plugins/inputs/jboss/jboss.go
```diff
@@ -0,0 +1,1312 @@
...
+// Returns:
+//     error: Any error that may have occurred
+
+func (h *JBoss) getServersOnHost(
+	acc telegraf.Accumulator,
+	serverURL string,
+	execAsDomain bool,
+	isEAP7 bool,
+	hosts []string,
+) error {
```
srebhan, line 390: AFAICS this function never returns an error, so please remove the return value indicating exactly this _OR_ change the error handling.

plugins/inputs/jboss/jboss.go
```diff
@@ -0,0 +1,1312 @@
...
+	acc telegraf.Accumulator,
+	serverURL string,
+	execAsDomain bool,
+	host string,
+	serverName string,
+) error {
+	adr := OrderedMap{}
+
+	if execAsDomain {
+		adr = OrderedMap{
```
srebhan, line 487: Does it really have to be ordered?

plugins/inputs/jboss/jboss.go
```diff
@@ -0,0 +1,1312 @@
...
+	bodyContent, err := h.createRequestBody(GetTransactionStat, adr)
+	if err != nil {
+		return fmt.Errorf("error on request to %s : %s\n", serverURL, err)
+	}
+
+	out, err := h.doRequest(serverURL, bodyContent)
+
+	if err != nil {
+		return fmt.Errorf("error on request to %s : %s\n", serverURL, err)
+	} else {
```
srebhan, line 507: Remove the `else` and save one indention level.

plugins/inputs/jboss/jboss.go
```diff
@@ -0,0 +1,1312 @@
...
+		server := TransactionResponse{}
+		if err = json.Unmarshal(out, &server); err != nil {
+			return fmt.Errorf("Error decoding JSON response: %s : %s", out, err)
+		}
+
+		fields := make(map[string]interface{})
+		for key, value := range server.Result {
+			if strings.Contains(key, "number-of") {
+				fields[key], _ = strconv.ParseInt(value.(string), 10, 64)
```
srebhan, line 516: Please check the type assertion here! If, for whatever reason, `value` is not a `string` you will tear-down telegraf completely with a panic!

plugins/inputs/jboss/jboss.go
```diff
@@ -0,0 +1,1312 @@
...
+				case string:
+					f, err := strconv.ParseFloat(value.(string), 64)
+					if err != nil {
+						log.Printf("E! JBoss Error decoding Float  from string : %s = %s\n", key, value.(string))
+					} else {
+						fields[key] = f
+					}
+				}
+			}
+		}
```
srebhan, line 598: This is hard to read. How about
```suggestion
		if value == nil {
			continue
		}

		switch key {
		case "bytesReceived", "bytesSent", "requestCount", "errorCount", "maxTime", "processingTime":
			switch v := value.(type) {
			case int:
				fields[key] = float64(v)
			case float64:
				fields[key] = v
			case string:
				if f, err := strconv.ParseFloat(v, 64); err == nil {
					fields[key] = f
				}
			}
		}
```

plugins/inputs/jboss/jboss.go
```diff
@@ -0,0 +1,1312 @@
...
+func (h *JBoss) getUndertowStatistics(
+	acc telegraf.Accumulator,
+	serverURL string,
+	execAsDomain bool,
+	host string,
+	serverName string,
+	listener string,
+) error {
+	adr := OrderedMap{}
+	var listenerName = "default"
```
srebhan, line 619:
```suggestion
	listenerName := "default"
```

plugins/inputs/jboss/jboss.go
```diff
@@ -0,0 +1,1312 @@
...
+	server := WebResponse{}
+	if err = json.Unmarshal(out, &server); err != nil {
+		return fmt.Errorf("Error decoding JSON response: OUT: %s  ERR: %s", out, err)
+	}
+
+	fields := make(map[string]interface{})
+	for key, value := range server.Result {
+		log.Printf("D! LISTERNER %s : %s \n", key, value)
+		switch key {
+		case "bytes-received", "bytes-sent", "request-count", "error-count", "max-processing-time", "processing-time":
```
srebhan, line 659: Same as above.

plugins/inputs/jboss/jboss.go
```diff
@@ -0,0 +1,1312 @@
...
+		"jboss_host":   host,
+		"jboss_server": serverName,
+		"type":         listener,
+	}
+	acc.AddFields("jboss_web", fields, tags)
+
+	return nil
+}
+
+func GetPoolFields(pool DBPoolStatistics) map[string]interface{} {
```
srebhan, line 687: Please make this a private function.

plugins/inputs/jboss/jboss.go
```diff
@@ -0,0 +1,1312 @@
...
+	retmap := make(map[string]interface{})
+	//Jboss EAP 6/AS 7.X returns "strings", wildfly 12 returns integers
+	switch pool.ActiveCount.(type) {
```
srebhan, line 690: Please use type-switches more efficiently.
```suggestion
	switch v := pool.ActiveCount.(type) {
```

plugins/inputs/jboss/jboss.go
```diff
@@ -0,0 +1,1312 @@
...
+
+	deployments := HostResponse{}
+	if err = json.Unmarshal(out, &deployments); err != nil {
+		return fmt.Errorf("Error decoding JSON response (HostResponse): %s : %s", out, err)
+	}
+
+	for _, deployment := range deployments.Result {
+		wg.Add(1)
+		go func(deployment string) {
+			defer wg.Done()
```
srebhan, line 1027: This is now the third nested wait-group. Do you _really_ need this parallelism?

plugins/inputs/jboss/jboss.go
```diff
@@ -0,0 +1,1312 @@
...
+			"address":         address,
+			"json.pretty":     1,
+		}
+	case GetHosts:
+		bodyContent = map[string]interface{}{
+			"operation":   "read-children-names",
+			"child-type":  "host",
+			"address":     address,
+			"json.pretty": 1,
+		}
```
srebhan, line 1179: Please move those to an extra file and declare them as private (plugin-global) variables like
```suggestion
var execStatBody = map[string]interface{}{
	"operation":       "read-resource",
	"attributes-only": "true",
	"include-runtime": "true",
	"json.pretty":     1,
}

var hostsBody = map[string]interface{}{
	"operation":   "read-children-names",
	"child-type":  "host",
	"json.pretty": 1,
}
...
```

Then in the functions use those variables and add/set the `address`.

plugins/inputs/jboss/jboss.go
```diff
@@ -0,0 +1,1312 @@
...
+			"json.pretty":     1,
+		}
+	}
+
+	return bodyContent, nil
+}
+
+func (h *JBoss) doRequest(domainURL string, bodyContent map[string]interface{}) ([]byte, error) {
+
+	serverURL, err := url.Parse(domainURL)
```
srebhan, line 1260: Do you have to do it in each iteration or will this stay the same per server?

plugins/inputs/jboss/jboss_test.go
```diff
@@ -0,0 +1,1561 @@
...
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/influxdata/telegraf/testutil"
+	"github.com/stretchr/testify/require"
+)
+
+var jboss_exec_mode_out = `
```
srebhan, line 14: Please move those to some files in a `testdata` directory and read them from there.

plugins/inputs/jboss/jboss_test.go
```diff
@@ -0,0 +1,1561 @@
...
+		if b.Operation == "read-resource" {
+			if b.AttributesOnly == "true" {
+				if eap7 {
+					fmt.Fprintln(w, jboss_exec_mode_out)
+				} else {
+					fmt.Fprintln(w, jboss_exec_mode_out_eap6)
+				}
+				return
+			}
+			if v, ok := b.Address[0]["core-service"]; ok {
```
srebhan, line 1344: You should check the length of `b.Address` before accessing the first entry.

srebhan (changes requested):
Thanks for caring about this plugin @stefa975! I have some major concerns but nothing we cannot overcome. The main ones are:

1. Why do you use your own version of `go-http-digest-auth-client` instead of the more maintained upstream version?
2. Please restructure the parallelism in your plugin! You have (at least) three nested waiting groups spawning goroutines. That's very hard to follow... Either rethink if you really need this parallelism and if so, try to come up with a way to construct the requests first and then run them in parallel in one wait-group.
3. Roughly 1,300 lines are a lot for a plugin. Try to unify functions with the same structure. Reduce your own implementations (OrderedMap) to an absolute minimum. Make use of existing telegraf-infrastructure as much as possible.
4. They to use golang language constructs (e.g. in type-switches) where applicable.
5. Maybe split the plugin-code into multiple files (e.g. separate-out the answer structs) where there are logical groups.

Looking forward to the next round!

stefa975:
Thanks for the review @srebhan, I'll look into this.

1. Yes, I have a fix on my version, but I have done a PR to upstream version.
2. I'll look in to that.
3.  Yes it's a large plugin. The OrderedMap in used for the special json that is posted to the jboss managent api.
4. and 5. Good points, I don't use Go dayly, so I'm still learning

srebhan:
@stefa975 thanks for pushing this forward. Regarding 3. so you need the OrderedMap to guarantee the order of the keys in the marshalled JSON? If not, I suggested to use dedicated variables for the different request templates removing the need for an ordered map... Anyway, please take a look and let me know where you see problems.

Regarding 4&5 I'll try to assist you as much as I can and give some hints and examples. Let me know if you have questions or even join the telegraf-dev channel on Slack.

plugins/inputs/jboss/jboss.go
```diff
@@ -0,0 +1,1262 @@
...
+	GetHosts
+	GetServers
+	GetDBStat
+	GetJVMStat
+	GetDeployments
+	GetDeploymentStat
+	GetWebStat
+	GetJMSQueueStat
+	GetJMSTopicStat
+	GetTransactionStat
```
srebhan, line 34: Those should be private variables starting with a lowercase letter. But I guess you don't need those if you switch to the variables scheme I suggested earlier.

plugins/inputs/jboss/jboss.go
```diff
@@ -0,0 +1,1262 @@
...
+)
+
+// KeyVal key / value struct
+type KeyVal struct {
```
srebhan, line 38: Should be a private type.

plugins/inputs/jboss/jboss.go
```diff
@@ -0,0 +1,1262 @@
...
+	Key string
+	Val interface{}
+}
+
+// OrderedMap Define an ordered map
+type OrderedMap []KeyVal
```
srebhan, line 44: Please also make this a non-exported/private type.

plugins/inputs/jboss/jboss.go
```diff
@@ -0,0 +1,1262 @@
...
+
+// WebMetrics  Web Modules related metrics
+type WebMetrics struct {
+	ActiveSessions    string                 `json:"active-sessions"`
+	ContextRoot       string                 `json:"context-root"`
+	ExpiredSessions   string                 `json:"expired-sessions"`
+	MaxActiveSessions string                 `json:"max-active-sessions"`
+	SessionsCreated   string                 `json:"sessions-created"`
+	Servlet           map[string]interface{} `json:"servlet"`
+}
```
srebhan, line 170: Can you please move those response structs to an own file (e.g. `jboss_api.go`) with the same `package` declaration (i.e. `package jboss`). This will reducing the clutter in this file.

plugins/inputs/jboss/jboss.go
```diff
@@ -0,0 +1,1262 @@
...
+
+// JBoss the main collectod struct
+type JBoss struct {
+	Servers []string `toml:"servers"`
+	Metrics []string `toml:"metrics"`
+
+	Username string
+	Password string
```
srebhan, line 178: Those also require a `toml` annotation.

plugins/inputs/jboss/jboss.go
```diff
@@ -0,0 +1,1262 @@
...
+
+	Authorization string
```
srebhan, line 180: Same here. Please add a `toml` declaration to _all_ config parameters!

plugins/inputs/jboss/jboss.go
```diff
@@ -0,0 +1,1262 @@
...
+  servers = [
+    "http://localhost:9090/management",
+  ]
+  
+  ## Username and password
+  username = ""
+  password = ""
+  
+  ## authorization mode could be "basic" or "digest"
+  authorization = "digest"
```
srebhan, line 202: Please comment those options as they contain the default value
```suggestion
  # username = ""
  # password = ""
  
  ## authorization mode could be "basic" or "digest"
  # authorization = "digest"
```

plugins/inputs/jboss/jboss.go
```diff
@@ -0,0 +1,1262 @@
...
+  # insecure_skip_verify = false
+  ## Metric selection
+  metrics =[
+	"jvm",
+	"web", 		# Handles both EAP <=6.X/AS <=7.X and EAP >=7.X/Widlfly > 8
+	"deployment",
+	"database",
+	"transaction",
+	"jms",
+  ]
```
srebhan, line 218: I this a sensible default? If so, please comment the option. Furthermore, I want you to think about slightly changing the mechanics here. How about an include and an exclude list? You can use the telegraf `filter` package to do this easily...

plugins/inputs/jboss/jboss.go
```diff
@@ -0,0 +1,1262 @@
...
+`
+
+func (*JBoss) SampleConfig() string {
+	return sampleConfig
+}
+
+func (*JBoss) Description() string {
+	return "Telegraf plugin for gathering metrics from JBoss AS"
```
srebhan, line 226: Please leave out `Telegraf` here
```suggestion
	return "Plugin for gathering metrics from JBoss AS"
```

plugins/inputs/jboss/jboss.go
```diff
@@ -0,0 +1,1262 @@
...
+}
+
+func (h *JBoss) Gather(acc telegraf.Accumulator) error {
+	var wg sync.WaitGroup
+
+	// Create an HTTP client that is re-used for each
+	// collection interval
+
```
srebhan, line 234: Can you please remove those newlines everywhere!? They make the code harder to read as they add kind of a new paragraph visually which is not what you do in the code.

plugins/inputs/jboss/jboss.go
```diff
@@ -0,0 +1,1262 @@
...
+	if h.client == nil {
+		client, err := h.createHttpClient()
+		if err != nil {
+			return err
+		}
+		h.client = client
+	}
```
srebhan, line 241: Please move this to an `Init()` function. See the example input plugin for an example.

plugins/inputs/jboss/jboss.go
```diff
@@ -0,0 +1,1262 @@
...
+
+	for _, server := range h.Servers {
+
+		//Check Exec Mode for each servers
+
```
srebhan, line 246: Newlines!

plugins/inputs/jboss/jboss.go
srebhan: Still an open question...

srebhan (commented):
Thanks @stefa975! There are still quite some comments open, so I guess you are in the middle of the rework. :-) Please let me know once you addressed the rest of the comments.

docs/LICENSE_OF_DEPENDENCIES.md
srebhan: Please try to merge your fork of that library with the upstream version!

plugins/inputs/jboss/jboss.go
srebhan: Can you please comment on this!?

plugins/inputs/jboss/jboss.go
```diff
@@ -0,0 +1,1153 @@
...
+}
+
+func (*JBoss) Description() string {
+	return "Gathering metrics from JBoss AS"
+}
+
+func (j *JBoss) Init() error {
+	if j.ResponseTimeout.Duration < time.Second {
+		j.ResponseTimeout.Duration = time.Second * 5
+	}
```
srebhan, line 136: Is zero a valid value for the timeout (e.g. wait forever)?

plugins/inputs/jboss/jboss.go
srebhan: Still valid.

plugins/inputs/jboss/jboss.go
```diff
@@ -0,0 +1,1153 @@
...
+				out, err := h.doRequest(server, bodyContent)
+
+				if err != nil {
+					h.Log.Errorf("JBoss Error handling response 1: %s", err)
+					h.Log.Errorf("JBoss server:%s bodyContent %s", server, bodyContent)
+					acc.AddError(err)
+					return
+				}
+				// Unmarshal json
+
```
srebhan, line 202: Remove newline or add before the comment.

plugins/inputs/jboss/jboss.go
srebhan: Please comment.

srebhan (commented):
There are still open comments in the code. Please take a look at those!

srebhan (commented):
Thanks for keep going @stefa975!
There are some remaining things in the code but it's getting less and less. :+1:

docs/LICENSE_OF_DEPENDENCIES.md
stefa975: I did a pull Request to the upstream project, but he replied it's not maintained any more.
I'll try to find som other library.

srebhan: The alternative is you merge your repo with the upstream one and maintain it. It's ok for me, but we need a good reason here. :-)

Hipska:
Any reason why @telegraf-tiger is not posting any builds?

OrganizedMayhem:
@stefa975 is this almost ready or still being worked on?

srebhan:
Any volunteer for working on this PR?

PR #8927, opened 2021-03-02
Get Tail tests to pass on windows

Jagularr:
'poll' mode works and is consistent, as opposed to the default fsnotify mode.


helenosheaa (approved):
Not sure whats failing on circle ci, when I click on it, it says its a legacy job view and I cannot rerun the workflow

PR #8930, opened 2021-03-03
Writing unit tests for Kinesis output plugin Write method

JeffAshton:
### Required for all PRs:

### DIFF
--- a/plugins/outputs/kinesis/kinesis.go
+++ b/plugins/outputs/kinesis/kinesis.go
@@ -30,9 +30,10 @@ type (
 		Partition          *Partition `toml:"partition"`
 		Debug              bool       `toml:"debug"`
 
-		Log        telegraf.Logger `toml:"-"`
-		serializer serializers.Serializer
-		svc        kinesisiface.KinesisAPI
+		Log                  telegraf.Logger `toml:"-"`
+		maxRecordsPerRequest uint32
+		serializer           serializers.Serializer
+		svc                  kinesisiface.KinesisAPI
 	}
 
 	Partition struct {
@@ -243,8 +244,7 @@ func (k *KinesisOutput) Write(metrics []telegraf.Metric) error {
 
 		r = append(r, &d)
 
-		if sz == 500 {
-			// Max Messages Per PutRecordRequest is 500
+		if sz >= k.maxRecordsPerRequest {
 			elapsed := k.writeKinesis(r)
 			k.Log.Debugf("Wrote a %d point batch to Kinesis in %+v.", sz, elapsed)
 			sz = 0
@@ -262,6 +262,10 @@ func (k *KinesisOutput) Write(metrics []telegraf.Metric) error {
 
 func init() {
 	outputs.Add("kinesis", func() telegraf.Output {
-		return &KinesisOutput{}
+		return &KinesisOutput{
+
+			// Limit set by AWS (https://docs.aws.amazon.com/kinesis/latest/APIReference/API_PutRecords.html)
+			maxRecordsPerRequest: 500,
+		}
 	})
 }
--- a/plugins/outputs/kinesis/kinesis.go
+++ b/plugins/outputs/kinesis/kinesis.go
@@ -13,6 +13,9 @@ import (
 	"github.com/influxdata/telegraf/plugins/serializers"
 )
 
+// Limit set by AWS (https://docs.aws.amazon.com/kinesis/latest/APIReference/API_PutRecords.html)
+const maxRecordsPerRequest uint32 = 500
+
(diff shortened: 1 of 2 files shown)


plugins/outputs/kinesis/kinesis.go
```diff
@@ -30,9 +30,10 @@ type (
 		Partition          *Partition `toml:"partition"`
 		Debug              bool       `toml:"debug"`
 
-		Log        telegraf.Logger `toml:"-"`
-		serializer serializers.Serializer
-		svc        kinesisiface.KinesisAPI
+		Log                  telegraf.Logger `toml:"-"`
+		maxRecordsPerRequest uint32
```
JeffAshton, line 34: Adding `maxRecordsPerRequest` to make unit testing easier.

plugins/outputs/kinesis/kinesis.go
```diff
@@ -243,8 +244,7 @@ func (k *KinesisOutput) Write(metrics []telegraf.Metric) error {
 
 		r = append(r, &d)
 
-		if sz == 500 {
-			// Max Messages Per PutRecordRequest is 500
+		if sz >= k.maxRecordsPerRequest {
```
JeffAshton, line 247: Switched to `>=` to be more defensive.  Should only ever evaluate to `true` on `==`, but if there was a bug, the alternative is worse.

JeffAshton:
Failure in `test-go-windows` is unrelated:

> FAIL	github.com/influxdata/telegraf/plugins/inputs/tail

Likely addressed by [Pull Request 8927](https://github.com/influxdata/telegraf/pull/8927)

Jagularr (approved):

plugins/outputs/kinesis/kinesis.go
srebhan: Hmmm I think this should be a `const` shouldn't it? It will probably never change at runtime...

plugins/outputs/kinesis/kinesis_test.go
```diff
@@ -227,6 +230,318 @@ func TestWriteKinesis_WhenServiceError(t *testing.T) {
 	})
 }
 
+func TestWrite_NoMetrics(t *testing.T) {
+
```
srebhan, line 234: Please remove newline.

plugins/outputs/kinesis/kinesis_test.go
```diff
@@ -323,3 +670,16 @@ func (m *mockKinesisPutRecords) AssertRequests(
...
+func serializeMetric(
+	t *testing.T,
+	serializer serializers.Serializer,
+	metric telegraf.Metric,
+) []byte {
+
+	data, err := serializer.Serialize(metric)
+	if err != nil {
+		t.Errorf("Should have serialized test metric: %s", err.Error())
+	}
```
srebhan, line 683:
```suggestion
	require.NoError(t, err)
```
I'm not so sure if those two-lines are worth an own function...

plugins/outputs/kinesis/kinesis_test.go
```diff
@@ -253,6 +568,38 @@ func (m *mockKinesisPutRecords) SetupResponse(
...
+) {
+
+	errorCode := "InternalFailure"
+	errorMessage := "Internal Service Failure"
+	shard := "shardId-000000000003"
+
+	records := []*kinesis.PutRecordsResultEntry{}
+
+	for i := int64(0); i < successfulRecordCount; i++ {
+
```
srebhan, line 583: Can you please reduce the usage of newlines to occasions where it helps structuring the code visually (e.g. two lines down).

srebhan (commented):
Looks mostly good, only very few comments. I'd like to see `maxRecordsPerRequest` being a constant and maybe dissolving of the `serializeMetric` function as this might help to identify the test that fails more easily.

plugins/outputs/kinesis/kinesis.go
JeffAshton: If it's a `const`, this would require creating 500+ records for testing batching.  While this is doable, it's not really practical.

plugins/outputs/kinesis/kinesis_test.go
JeffAshton: Lol.  I find the newlines helpful, but I'm visually impaired, so I I know I'm bias.  :)  Let me know if that's better for you.

plugins/outputs/kinesis/kinesis.go
JeffAshton: Done

srebhan (approved):
Perfect! Thank you very much.

PR #8931, opened 2021-03-03
Add XML parser using XPath queries

reimda:
Fixes go.mod from #8047


plugins/parsers/xml/parser.go
```diff
@@ -0,0 +1,422 @@
...
+				return nil, fmt.Errorf("cannot parse line with empty selection")
+			} else if len(selectedNodes) != 1 {
+				return nil, fmt.Errorf("cannot parse line with multiple selected nodes (%d)", len(selectedNodes))
+			}
+			selected = selectedNodes[0]
+		}
+
+		return p.parseQuery(t, doc, selected, config)
+	}
+	return nil, fmt.Errorf("cannot parse line with multiple (%d) configurations", len(p.Configs))
```
Hipska, line 106: @reimda What is the reason ParseLine cannot be used in combination with multiple configs?

reimda: I'm not sure. It looks like p.document holds some state, so that p.document.Parse() can't be called for multiple configs. @srebhan wrote this code, maybe he can give a better answer.

reimda: Oops, looks like you two are already discussing it in PR 11102

PR #8950, opened 2021-03-06
Teamspeak input plugin query clients

JoshKeegan:
### Required for all PRs:

Add additional field `query_clients_online` to the teamspeak input plugin.
This field is useful because the `clients_online` field already included in the plugin is a total of both end-users and server query users connected to a virtual server, so having this allows you to break down the total clients by type.


Jagularr (approved):

srebhan (approved):
Looks good to me.

PR #8951, opened 2021-03-07
Proposal of enabled linters and their settings.

zak-pawel:
- Most valuable (TBD?) linters for `golangci-lint` enabled, for `revive` linter almost all rules enabled. Both linters and rules should reviewed to get list of what should stay enabled.
- `golangci-lint` bumped from `v1.37.0` to `v1.38.0`
- settings from `Makefile` moved to config file

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -155,6 +155,7 @@ run:
     - scripts
     - docs
     - etc
+    - plugins/parsers/influx/plugins #workaround to skip weird "plugins/parsers/influx/plugins/parsers/influx" directory
 
   # which files to skip: they will be analyzed, but issues from them
   # won't be reported. Default value is empty list, but there is
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,36 +1,150 @@
 linters:
   enable:
+    - bodyclose
+    - deadcode
+    - dogsled
+    - errcheck
+    - funlen
+    - goprintffuncname
+    - gosec
+    - gosimple
+    - govet
+    - ifshort
+    - ineffassign
+    - nakedret
+    - nilerr
+    - predeclared
     - revive
+    - sqlclosecheck
+    - staticcheck
+    - structcheck
+    - typecheck
+    - unconvert
+    - unparam
+    - unused
+    - varcheck
+  disable:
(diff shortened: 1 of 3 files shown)


Jagularr (approved):

.golangci.yml
```diff
@@ -1,36 +1,150 @@
 linters:
   enable:
+    - bodyclose
+    - deadcode
+    - dogsled
```
sspaink, line 5: The only issues `dogsled` caught are in test files. Maybe not necessary?

zak-pawel: Right now only findings for `dogsled` are on test files, I agree. But it could be a good check for "production" code in the future!

srebhan: After looking it up, I'm sure we want dogsleds enabled. :-)

Makefile
```diff
@@ -138,7 +138,7 @@ ifeq (, $(shell which golangci-lint))
 	exit 1
 endif
 
-	golangci-lint run --timeout 5m0s --issues-exit-code 0
```
reimda, line 141: I'm concerned that enabling many linters will increase time too much.  Does removing this mean we expect it to take longer than 5 minutes?  Instead of removing the timeout maybe we should increase it.

zak-pawel: @reimda this setting was just moved to `.golangci.yml`

reimda: Aha, thanks

sspaink:
@zak-pawel I re-based to get the github action changes and updated the config. Let me know if all looks good!

.golangci.yml
```diff
@@ -1,46 +1,152 @@
...
+      - name: var-declaration
+      - name: var-naming
+      - name: waitgroup-by-value
+  nakedret:
+    # make an issue if func has more lines of code than this setting and it has naked returns; default is 30
+    max-func-lines: 1
 
 run:
+  # exit code when at least one issue was found, default is 1
+  issues-exit-code: 0
```
sspaink, line 134: This can be removed, it is handled in the github action config.

zak-pawel: Removed

PR #8953, opened 2021-03-08
Reset the flush interval timer when flush is requested or batch is ready.

### LINKED ISSUE #8941: did not complete within its flush interval
It's possible for the "did not complete within its flush interval" message to appear when there are no slow writes to outputs due to how the timers are set up. 

### Steps to reproduce:

1. write enough metrics out to trigger the batch writes before the flush timeout
2. if it lines up just right you'll see the "did not complete within its flush interval" on writes that did not take very long, even far below the flush interval.

### Expected behavior:

You don't see the message unless the write is > flush interval.

### Actual behavior:

you see it whenever a flush happens in the middle of a batch write

Jagularr:
Reset the flush interval timer when flush is requested or batch is ready, so that timer doesn't expire (and then falsely report that flush interval has been exceeded without writing) while one of those flushes is occurring.

resolves #8941

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -793,7 +793,7 @@ func (a *Agent) runOutputs(
 func (a *Agent) flushLoop(
 	ctx context.Context,
 	output *models.RunningOutput,
-	ticker Ticker,
+	ticker *RollingTicker,
 ) {
 	logError := func(err error) {
 		if err != nil {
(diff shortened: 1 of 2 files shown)


ssoroka (approved):

gnjack:
@Jagularr @ssoroka We've just upgraded to telegraf 1.19.3 and I believe this change has introduced a subtle bug where the buffer is not flushed fully when the metric throughput is higher than that of the `batch_size` /  `flush_interval` and there are large spikes in metrics to write.

In this scenario `ticker.Elapsed()` is never true due to `output.BatchReady` calling `ticker.Reset()` more frequently than the  `flush_interval`.

 `output.WriteBatch` **only flushes a single batch of metrics**, unlike `ticker.Elapsed()` which calls `output.Write` flushing as many batches as in the buffer in a single call.

If you then have a massive spike of metrics (several multiples of `batch_size`) then `BatchReady` will be set to `time.Now` multiple times in very quick succession.https://github.com/influxdata/telegraf/blob/74a1acd8144b6929171efffdf44a1135442b1588/models/running_output.go#L172

This may only actually trigger `output.WriteBatch` once on the next iteration of the `flushLoop` flushing only a single batch, leaving many possible batches unflushed. Due to `ticker.Reset()` being constantly called, `output.Write` is never called and this backlog of many possible batches in the backlog is never flushed as we are only ever flushing a single batch at a time using `output.WriteBatch`.

You can see this on the following graphs - metrics build up in the buffer unflushed after every spike in metrics gathered, until we get lucky and `flush_interval` passes without an early `BatchReady` induced flush, causing `output.Write` to be called instead of `output.WriteBatch` and actually flushing all buffered batches in one method call.
image

Issue #8171, opened 2020-09-23
Common plugin TLS module allows for insecure ciphers 


darinfisher:
The TLS module included for plugins allows insecure versions 1.0 and 1.1.

See the plugins common file: `plugins/common/tls/common.go`

ssoroka:
So I guess the question is should we allow users to use insecure versions if they really want to? I'd love to say "you must use secure settings", but I'm thinking that's going to cause a problem with some old client somewhere. Maybe we should have a list of secure ciphers and allow an agent setting to switch back to the old list if anyone has problems. We can name it something scary like "allow_insecure_ciphers" and "allow_insecure_tls_versions"

jdstrand:
To add some flavor to this, TLS1.0 and TLS1.1 are effectively EOL as of March 2020 due to their inherent problems:
* https://security.googleblog.com/2018/10/modernizing-transport-security.html
* https://blogs.windows.com/msedgedev/2018/10/15/modernizing-tls-edge-ie11/
* https://webkit.org/blog/8462/deprecation-of-legacy-tls-1-0-and-1-1-versions/
* https://blog.mozilla.org/security/2018/10/15/removing-old-versions-of-tls/

I agree we shouldn't be using them and should enforce TLS1.2 at a minimum by adjusting plugins/common/tls/config.go to set tlsConfig.MinVersion accordingly (this would prevent downgrade attacks. You can use tls.MaxVersion in your unittests to mock things that don't speak TLS1.2 to verify it is working as intended).

I'm new to telegraf (but looking at open security issues across influxdata); who is expected to be talking to the listener or put another way, what client software is expected to connect to telegraf? Adding an allow_insecure_tls_versions (or similar) that defaults to false but when true sets tlsConfig.MinVersion to TLS1.0 would be acceptable IME.

Also, this isn't assigned to anyone; should it be?

jdstrand:
@ssoroka - fyi, I submitted https://github.com/influxdata/telegraf/pull/8959 for this issue.

jdstrand:
I think it is clear that I don't have time to work on this. I would be happy for anyone to pick up my work and run with it. If priorities change, I'll circle back to it, but I don't see that happening in the foreseeable short to mid term future.

jdstrand:
[...]

PR #8959, opened 2021-03-08
require TLS1.2 in ClientConfig (with TLSMinVersion override) (#8171)

jdstrand:
### Required for all PRs:

This PR aims to address https://github.com/influxdata/telegraf/issues/8171 and makes clients require a server that supports TLS1.2, unless the TLSMinVersion configuration option is supplied in ClientConfig. Unit tests have been added, but I wasn't sure what README.md file to update (please advise).

### DIFF
--- a/plugins/common/tls/config.go
+++ b/plugins/common/tls/config.go
@@ -15,6 +15,7 @@ type ClientConfig struct {
 	TLSKey             string `toml:"tls_key"`
 	InsecureSkipVerify bool   `toml:"insecure_skip_verify"`
 	ServerName         string `toml:"tls_server_name"`
+	TLSMinVersion      string `toml:"tls_min_version"`
 
 	// Deprecated in 1.7; use TLS variables above
 	SSLCA   string `toml:"ssl_ca"`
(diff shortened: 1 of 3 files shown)


jdstrand:
Looking at the failing tests, I see I need to change the logic back to returning nil if no options are set and only set MinVersion if other options are actually set.

plugins/common/tls/config.go
```diff
@@ -57,9 +58,21 @@ func (c *ClientConfig) TLSConfig() (*tls.Config, error) {
 		return nil, nil
 	}
 
+	// TLS1.0 and TLS1.1 are effectively EOL, so make TLS1.2 the minimum. See
+	// https://github.com/influxdata/telegraf/issues/8171#issuecomment-777120753
 	tlsConfig := &tls.Config{
 		InsecureSkipVerify: c.InsecureSkipVerify,
 		Renegotiation:      tls.RenegotiateNever,
+		MinVersion:         tls.VersionTLS12,
```
ssoroka, line 66: I like this idea, but we support a lot of really old software and this will likely be a breaking change for some of them. I'd like to manage this by putting this one line in its own PR, and when they don't specify the version spit out a message saying something like: `TLS 1.0 is deprecated and will no longer be the default in a future version of Telegraf. To suppress this message specify the minimum TLS version explicitly, eg tls_min_version = "TLS13"`. ... alternatively maybe we could find a way to warn when we establish a connection with a really old TLS version.. though that seems more complicated.

There's also a MaxVersion we might want to support, though I don't currently have a use case for it.

ssoroka (changes requested):
Looks great. as far as docs go, search the project for "insecure_skip_verify", and you'll find all the places the TLS config is mentioned, and you can add the tls_min_version to those. I suggest adding it with 1.2 as the default in the config `tls_min_version = "TLS12"` (or whatever the format is). be sure to list all the valid options, and it can be commented out like the other tls options.

jdstrand:
From https://github.com/influxdata/telegraf/issues/8171#issuecomment-940426331 from me: "_I think it is clear that I don't have time to work on this. I would be happy for anyone to pick up my work and run with it. If priorities change, I'll circle back to it, but I don't see that happening in the foreseeable short to mid term future._"

Issue #8682, opened 2021-01-13
Add Prometheus Remote Write Parser


sjwang90:
### Proposal:
Add prometheus remote write parser to use a `data_format` for input plugins
### Current behavior:
None - only [Prometheus remote write serializer](https://github.com/influxdata/telegraf/tree/master/plugins/serializers/prometheusremotewrite)

### Desired behavior:
Add prometheus remote write parser where users can read in data with setting `data_format = "prometheusremotewrite"` in their input plugin configuration.

### Use case: 
Allows users to parse prometheus remote_write format with Telegraf and output to InfluxDB. 

There are currently no capabilities in InfluxDB 2.x for prometheus remote_write. With an Telegraf parser, users could transform their prometheus remote_write data into line protocol with Telegraf and send it into InfluxDB 2.x.  This would be a key enhancement for users being able to migrate from InfluxDB 1.x to 2.x. See https://github.com/influxdata/influxdb/issues/17958

aleksey-novikov:
Basic implementation: [https://github.com/influxdata/telegraf/pull/8702]

PR #8966, opened 2021-03-10
Adds snappy support for http_listener_v2

helenosheaa:
Adds snappy support plus test

Relates to #8682, #8967 - one use is for with prometheus remote write parsing


plugins/inputs/http_listener_v2/http_listener_v2.go
```diff
@@ -268,6 +269,16 @@ func (h *HTTPListenerV2) collectBody(res http.ResponseWriter, req *http.Request)
...
+	// Handle snappy request bodies
+	if req.Header.Get("Content-Encoding") == "snappy" {
+		bytes, err = snappy.Decode(nil, bytes)
+		if err != nil {
+			h.Log.Debug(err.Error())
+			badRequest(res)
+			return nil, false
+		}
+	}
+
```
srebhan, line 281: Is it possible to use the `snappy.NewReader` function and match the processing of `gzip`!?

srebhan (commented):
I would prefer to match the pattern of gzip and construct a reader. Is there any reason to _not_ do it this way?

helenosheaa:
Thanks for the review, I'll take a look at matching the gzip pattern

helenosheaa:
So after just having another look at the docs, we want to use Decode rather than a newReader as we need the block format rather than streaming format (specifically for the prometheus remote write use case). Bit more context [here](https://pkg.go.dev/github.com/golang/snappy#:~:text=The%20block%20format%20is%20the,at%20the%20time%20compression%20starts.)

plugins/inputs/http_listener_v2/http_listener_v2.go
helenosheaa: See comment below

srebhan: Could you maybe add a small comment here that we need the block-decode!?

helenosheaa: Yes, added

srebhan: Thanks!

srebhan (approved):
Looks good to me.

plugins/inputs/http_listener_v2/http_listener_v2.go
```diff
@@ -268,6 +269,17 @@ func (h *HTTPListenerV2) collectBody(res http.ResponseWriter, req *http.Request)
 		return nil, false
 	}
 
+	// Handle snappy request bodies
+	if req.Header.Get("Content-Encoding") == "snappy" {
```
reimda, line 273: About 20 lines above there's a similar block for gzip.  It looks like we should change the "if content-encoding == gzip" to a switch statement that also handles snappy, then move this new block into the switch statement.  The snappy decoding needs to happen at the same place gzip does, namely before the MaxBytesReader.

reimda (changes requested):

plugins/inputs/http_listener_v2/http_listener_v2.go
helenosheaa: The NewReader function of snappy would return a reader such as the gzip implementation so the switch would work but the block implementation of snappy that we need (for prometheus remote write) is a Decode which takes in bytes and returns bytes. So I think it needs to be after the reader on 266 unless I'm missing something.

plugins/inputs/http_listener_v2/http_listener_v2.go
```diff
@@ -268,6 +269,17 @@ func (h *HTTPListenerV2) collectBody(res http.ResponseWriter, req *http.Request)
...
+		// Requires snappy decode for block format
+		bytes, err = snappy.Decode(nil, bytes)
```
ssoroka, line 275: Let's use snappy.NewReader() instead which works on streams and not byte slices, then  ioutil.ReadAll is the last step of both cases, so definitely go with the switch statement

plugins/inputs/http_listener_v2/http_listener_v2_test.go
```diff
@@ -327,6 +328,37 @@ func TestWriteHTTPGzippedData(t *testing.T) {
...
+	require.NoError(t, err)
+	req.Header.Set("Content-Encoding", "snappy")
+
+	client := &http.Client{}
+	resp, err := client.Do(req)
+	require.NoError(t, err)
+	require.EqualValues(t, 204, resp.StatusCode)
+
+	hostTags := []string{"server01"}
+	acc.Wait(len(hostTags))
```
ssoroka, line 352: should this just be 1, since it's 1 metric?

ssoroka (changes requested):

plugins/inputs/http_listener_v2/http_listener_v2.go
helenosheaa: As discussed refactored to use a switch statement, but sticking with Decode as the NewReader doesn't support the block format

ssoroka (approved):

PR #8969, opened 2021-03-11
Get rid of deprecated internal.{Duration,Size,Number}

srebhan:
### Required for all PRs:

The datatypes `internal.Size` and `internal.Duration` are replaced by their `config` counterparts. This PR moves the code to the new types and fix the fallout. Furthermore, I removed `internal.Number` which seems to be used by `statsd` input plugin only to represent a `float64`.


internal/internal_test.go
```diff
@@ -171,52 +171,6 @@ func TestRandomSleep(t *testing.T) {
 	assert.True(t, elapsed < time.Millisecond*150)
 }
 
-func TestDuration(t *testing.T) {
```
ssoroka, line 174: do these need to be moved over to the config.Duration tests or are they duplicated? I can't remember.

plugins/parsers/influx/machine.go
```diff
@@ -1,4 +1,3 @@
-
```
ssoroka, line 1: this file is generated and should not be modified. update machine.go.rl instead.

ssoroka (changes requested):
epic work. Thank you! couple notes

config/testdata/azure_monitor.toml
```diff
@@ -0,0 +1,4 @@
+[[outputs.azure_monitor]]
+
+[[outputs.azure_monitor]]
+  namespace_prefix = ""
```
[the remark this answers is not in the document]
srebhan: This is factored out of the `config_test` which had this inlined in the code. Should I do another PR for factoring this out? Please not that additional PR will have a dependency on this PR... :-)

internal/internal_test.go
srebhan: Oh I was assuming it's there but it isn't will fix.

plugins/parsers/influx/machine.go
srebhan: Will fix. Seems like my editor was of the opinion running `go fmt` on it would be a good idea... :-)

config/testdata/azure_monitor.toml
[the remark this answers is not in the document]
srebhan: Will keep as is.

plugins/inputs/opensmtpd/opensmtpd_test.go
```diff
@@ -2,14 +2,18 @@ package opensmtpd
...
+
+	"github.com/influxdata/telegraf/config"
 	"github.com/influxdata/telegraf/testutil"
 	"github.com/stretchr/testify/assert"
-	"testing"
 )
 
-func SMTPCTL(output string) func(string, internal.Duration, bool) (*bytes.Buffer, error) {
-	return func(string, internal.Duration, bool) (*bytes.Buffer, error) {
+var TestTimeout = config.Duration(time.Second)
```
ssoroka, line 13: looks like we added this line back in after another PR removed it?

ssoroka (approved):

ssoroka (approved):

PR #8970, opened 2021-03-11
Simplifying the kinesis output tests

JeffAshton:
### Required for all PRs:


plugins/outputs/kinesis/kinesis_test.go
```diff
@@ -104,17 +109,11 @@ func TestPartitionKey(t *testing.T) {
...
 
-	partitionKey := "partitionKey"
-	shard := "shard"
-	sequenceNumber := "sequenceNumber"
-	streamName := "stream"
-
 	records := []*kinesis.PutRecordsRequestEntry{
 		{
-			PartitionKey: &partitionKey,
+			PartitionKey: aws.String(testPartitionKey),
```
JeffAshton, line 116: Learned of these helper functions after the fact:

```
// String returns a pointer to the string value passed in.
func String(v string) *string {
	return &v
}
```

plugins/outputs/kinesis/kinesis_test.go
```diff
@@ -578,11 +546,13 @@ func (m *mockKinesisPutRecords) AssertRequests(
...
 ) {
 
-	assert.Equal(
+	if !assert.Equal(
 		len(expected),
 		len(m.requests),
 		fmt.Sprintf("Expected %v requests", len(expected)),
-	)
+	) {
+		return
```
JeffAshton, line 554: Exit early to avoid index out of range in subsequent check

plugins/outputs/kinesis/kinesis_test.go
```diff
@@ -593,24 +563,26 @@ func (m *mockKinesisPutRecords) AssertRequests(
...
 		)
 
-		assert.Equal(
+		if !assert.Equal(
 			len(expectedInput.Records),
 			len(actualInput.Records),
 			fmt.Sprintf("Expected request %v to have %v Records", i, len(expectedInput.Records)),
-		)
+		) {
+			return
```
JeffAshton, line 571: Exit early to avoid index out of range in subsequent check

plugins/outputs/kinesis/kinesis_test.go
srebhan: You can achieve the same using `require.Equal`. Citing the [package documentation](https://pkg.go.dev/github.com/stretchr/testify/require):

> Package require implements the same assertions as the `assert` package but stops test execution when a test fails.

srebhan (commented):
Can you please check if using `require.Equals` can prevent the manual killing of the tests!?

plugins/outputs/kinesis/kinesis_test.go
JeffAshton: So pretty much avoid `testify/assert` and use `testify/require`.  :)

JeffAshton:
> Can you please check if using `require.Equals` can prevent the manual killing of the tests!?

It does.  Thanks for the tip.  That's too bad `assert` works differently.

srebhan:
I think it's good to have both, but tbh I usually prefer `require` as it often doesn't make sense to continue the test if something fails...

plugins/outputs/kinesis/kinesis_test.go
```diff
@@ -602,19 +570,19 @@ func (m *mockKinesisPutRecords) AssertRequests(
...
+				actualRecord.PartitionKey,
 				fmt.Sprintf("Expected (request %v, record %v) to have correct PartitionKey", i, r),
 			)
 
-			assert.Equal(
-				&expectedRecord.ExplicitHashKey,
-				&actualRecord.ExplicitHashKey,
+			require.Equal(t,
+				expectedRecord.ExplicitHashKey,
+				actualRecord.ExplicitHashKey,
```
srebhan, line 581: Oh one more improvement here... :-) You can use `require.Equalf` (and `assert.Equalf` likewise) to construct the error message without the need for `fmt.Springf`...

JeffAshton: No worries.  I appreciate it.

plugins/outputs/kinesis/kinesis_test.go
```diff
@@ -546,46 +546,46 @@ func (m *mockKinesisPutRecords) AssertRequests(
 	expected []*kinesis.PutRecordsInput,
 ) {
 
-	require.Equal(t,
+	require.Equalf(t,
```
srebhan, line 549: If you want to +1 for your `require`-skill you might want to use `require.Lenf` for length checks. :-) But just a remark not a real request.

srebhan (approved):
Looks good to me. If you want to use `require.Lenf()` for length checks, please just submit and I will review. :-)

srebhan (approved):
Looks good to me.

helenosheaa (approved):
Looks good

PR #8975, opened 2021-03-11
Move golangci-lint from circle-ci to github actions

sspaink:
Golangci-lint project provides a github action: https://github.com/golangci/golangci-lint-action

This pr adds a config to enable the github action (replacing circle-ci), this should be faster with caching enabled and not having to download the golangci-lint binary each time.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -71,14 +71,6 @@ commands:
           paths:
             - 'dist'
 jobs:
-  linter:
-    executor: go-1_16
-    steps:
-      - checkout
-      - restore_cache:
-          key: go-mod-v1-{{ checksum "go.sum" }}
-      - run: wget -O- -nv https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.37.0
-      - run: make lint
   deps:
     executor: go-1_16
     steps:
@@ -209,7 +201,6 @@ workflows:
   version: 2
   check:
     jobs:
-      - 'linter'
       - 'macdeps':
           filters:
             tags:
@@ -287,7 +278,6 @@ workflows:
                 only: /.*/
   nightly:
     jobs:
-      - 'linter'
       - 'deps'
       - 'macdeps'
       - 'test-go-1_15':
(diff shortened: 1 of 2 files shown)


.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,19 @@
...
+jobs:
+  golangci-pr:
+    name: lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v2
+        with:
+          version: v1.29
```
zak-pawel, line 18: `v1.38` (or `latest`, but particular version seems to be safer) - based on https://github.com/golangci/golangci-lint/blob/master/assets/github-action-config.json

sspaink: Good catch, I updated the version

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,19 @@
...
+  golangci-pr:
+    name: lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v2
+        with:
+          version: v1.38
+          only-new-issues: true
```
zak-pawel, line 19: + `args: --timeout 5m0s` ?

.circleci/config.yml
```diff
@@ -287,7 +278,6 @@ workflows:
                 only: /.*/
   nightly:
     jobs:
-      - 'linter'
```
zak-pawel, line 290: Don't we want to have non-failing linter run for nightly targets (for whole codebase)?

ssoroka (approved):
seems fine

.circleci/config.yml
sspaink: good suggestion, I added a daily schedule

.github/workflows/golangci-lint.yml
sspaink: I forgot your patch with the updated config hasn't been merged yet, I added the timeout arg

PR #8987, opened 2021-03-14
Add telegraf url env var

### LINKED ISSUE #8909: Add Environmental Variable To Fetch Config from URL
## Feature Request
Allow for fetching a configuration file from a URL set by an environmental variable.

### Proposal:
Add a new variable, such as `TELEGRAF_CONFIG_URL`, that will parse a configuration file from a URL.

New order for finding config file:
1. -config
2. $TELEGRAF_CONFIG_URL
3. $TELEGRAF_CONFIG_PATH
4. $HOME/.telegraf/telegraf.conf
5. /etc/telegraf/telegraf.conf

### Current behavior:
The `-config` option allows for either a URL or a file path and the `TELEGRAF_CONFIG_PATH` only allows a file path.

### Desired behavior:
Fetch a configuration from a URL when set by `TELEGRAF_CONFIG_URL`.

### Use case:
When setting up an InfluxDB and Telegraf stack using docker-compose the `-config` switch cannot be used. Therefore, an environmental variable that can resolve the URL is required.

Example:
``` shell
  telegraf:
    image: telegraf:1.17.2
    environment:
      - TELEGRAF_CONFIG_URL="http://influxdb.local:8096/api/v2/telegraf/0000000000000000"
      - INFLUX_TOKEN=/run/secrets/INFLUX_TOKEN
```

Kuzyashin:
Closes #8909.
Add $TELEGRAF_CONFIG_URL to config discovery


srebhan (approved):
Looks good to me.

Jagularr (approved):

config/config.go
```diff
@@ -686,11 +686,13 @@ func (c *Config) LoadDirectory(path string) error {
...
-//   2. $HOME/.telegraf/telegraf.conf
-//   3. /etc/telegraf/telegraf.conf
+//   1. $TELEGRAF_CONFIG_URL
+//   2. $TELEGRAF_CONFIG_PATH
+//   3. $HOME/.telegraf/telegraf.conf
+//   4. /etc/telegraf/telegraf.conf
 //
 func getDefaultConfigPath() (string, error) {
+	urlfile := os.Getenv("TELEGRAF_CONFIG_URL")
 	envfile := os.Getenv("TELEGRAF_CONFIG_PATH")
```
reimda, line 696: The -config command line option takes either a local path or a URL.  I think we should also have one environment variable that takes either a local path or a URL.  I would prefer to keep TELEGRAF_CONFIG_PATH for backward compatibility and allow it to accept URLs.

It is unfortunate that the "_PATH" might make users think it only accepts local paths, but I think this is better than having two environment variables.  We should make sure to add to the documentation to let users know URLs work too.

In the long term we may want to consider renaming the single env var to something more generic like "TELEGRAF_CONFIG" but that would be a breaking change so it would have to wait for a major release (2.0.0) where compatibility can be broken.

For this PR to be merged, could you remove the new TELEGRAF_CONFIG_URL env var?  It looks like we may also need to combine the getDefaultConfigPath and LoadConfig and loadConfig functions because getDefaultConfigPath checks if the string is a file path, but loadConfig checks if it's a URL.  I think we will need to do that in the same function.

reimda (changes requested):

config/config.go
reimda: @Kuzyashin What do you think of this approach?  Are you able to work on this PR more?

Kuzyashin: Yes. I can push new commits in 2-3 days.

reimda: Let me know if you can't find the time for this and we'll pass it on to someone else.

srebhan:
@Kuzyashin did you have any chance to take a look?

config/config.go
```diff
@@ -686,9 +686,10 @@ func (c *Config) LoadDirectory(path string) error {
 }
 
 // Try to find a default config file at these locations (in order):
-//   1. $TELEGRAF_CONFIG_PATH
-//   2. $HOME/.telegraf/telegraf.conf
-//   3. /etc/telegraf/telegraf.conf
+//   1. $TELEGRAF_CONFIG_URL
+//   2. $TELEGRAF_CONFIG_PATH
+//   3. $HOME/.telegraf/telegraf.conf
+//   4. /etc/telegraf/telegraf.conf
```
srebhan, line 692: I think you also need to revert this part...

config/config.go
```diff
@@ -702,6 +703,10 @@ func getDefaultConfigPath() (string, error) {
 		etcfile = programFiles + `\Telegraf\telegraf.conf`
 	}
 	for _, path := range []string{envfile, homefile, etcfile} {
+		if valid := isUrl(path); valid {
```
srebhan, line 706: Please simplify to
```suggestion
		if isUrl(path) {
```

srebhan (commented):
Hey @Kuzyashin! Thanks for your quick response. There are two minor comment in the code. Can you please take a look?

Kuzyashin:
> Hey @Kuzyashin! Thanks for your quick response. There are two minor comment in the code. Can you please take a look?

Yes. Sorry for the long answer. Lot of work lately.

srebhan (approved):
Looks good to me. Thanks again for the quick cycle!

srebhan:
@Kuzyashin looks like you need to resolve the conflict... :-)

config/config.go
srebhan: @reimda can you please check again, I think your comment was resolved with the latest commits.

reimda (approved):
Thanks!

srebhan:
@Kuzyashin the majority of the CI tests failed. Can you please take a look and fix the problem if there is any?

srebhan (approved):

PR #4215, opened 2018-05-31
Cpufreq plugin

R4scal:
Hi.

This plugin work on Linux systems and collect information about cpu frequency and throttles

### Required for all PRs:

- [x] Signed [CLA](https://influxdata.com/community/cla/).
- [x] Associated README.md updated.
- [x] Has appropriate unit tests.

PR #8988, opened 2021-03-15
feat(inputs.linux_cpu): Add CPUFreq input plugin for Linux (v3)

### LINKED ISSUE #2601: Fetch CPU frequency states
### Proposal:
Fetch CPU frequency states, governor

### Current behavior:
```[[inputs.system]]```
does not show CPU frequency, sleep states etc.

### Desired behavior:
`inputs.system` should also fetch cpu frequency, sleep states etc.

lower CPU frequencys might impact services; lower frequencies might be forced when the temperature is too high or there are problems with the PSU

### LINKED ISSUE #4256: Add cpu_speed (in mhz) to cpu or system measurement

### PR DESCRIPTION
**Pull request by fabianishere**
This pull request adds an input plugin for monitoring Linux' CPUFreq subsystem. This pull request builds upon the work #4215 and #8342, but given that they haven't seen any activity in a while, I would like to complete the integration of this plugin into the Telegraf repository.

Currently, the plugin only exposes `cpufreq/scaling_cur_freq`, `cpufreq/scaling_min_freq`, `cpufreq/scaling_cur_freq` and `thermal_throttle/core_throttle_count`. If desirable, we could also add other fields exposed by the CPUFreq subsystem. 
**Question to the maintainers:** the `thermal_throttle/core_throttle_count` field is a bit out of place in this plugin, since it is not part of the CPUFreq subsystem in Linux. We could limit this plugin to only expose properties from the CPUFreq subsystem and have a separate plugin for `thermal_throttle` or we could emit all properties that Linux exposes per CPU (and rename the plugin to `cpulinux` for instance)?

I have optimized the implementation to reduce the number of syscalls per gather cycle. However, a limitation of this implementation is that it does not support hotplugging CPUs, though I don't expect this to affect many people.

### Required for all PRs:
- [x] Signed CLA.

Resolves #2601
Resolves #4256

### DIFF
--- a/plugins/inputs/linux_cpu/README.md
+++ b/plugins/inputs/linux_cpu/README.md
@@ -1,0 +1,64 @@
+# Linux CPU Input Plugin
+
+The `linux_cpu` plugin gathers CPU metrics exposed on Linux-based systems.
+
+## Configuration
+
+```toml @sample.conf
+# Collects CPU metrics exposed on Linux
+[[inputs.linux_cpu]]
+  ## Path for sysfs filesystem.
+  ## See https://www.kernel.org/doc/Documentation/filesystems/sysfs.txt
+  ## Defaults:
+  # host_sys = "/sys"
+
+  ## CPU metrics collected by the plugin.
+  ## Supported options:
+  ## "cpufreq", "thermal"
+  ## Defaults:
+  # metrics = ["cpufreq"]
+```
+
+## Metrics
+
+The following tags are emitted by the plugin under the name `linux_cpu`:
+
+| Tag   | Description           |
+|-------|-----------------------|
+| `cpu` | Identifier of the CPU |
+
+The following fields are emitted by the plugin when selecting `cpufreq`:
+
+| Metric name (field) | Description                                                | Units |
+|---------------------|------------------------------------------------------------|-------|
+| `scaling_cur_freq`  | Current frequency of the CPU as determined by CPUFreq      | KHz   |
+| `scaling_min_freq`  | Minimum frequency the governor can scale to                | KHz   |
+| `scaling_max_freq`  | Maximum frequency the governor can scale to                | KHz   |
+| `cpuinfo_cur_freq`  | Current frequency of the CPU as determined by the hardware | KHz   |
+| `cpuinfo_min_freq`  | Minimum operating frequency of the CPU                     | KHz   |
+| `cpuinfo_max_freq`  | Maximum operating frequency of the CPU                     | KHz   |
+
+The following fields are emitted by the plugin when selecting `thermal`:
+
+| Metric name (field)   | Description                                                 | Units |
+|-----------------------|-------------------------------------------------------------|-------|
+| `throttle_count`      | Number of thermal throttle events reported by the CPU       |       |
+| `throttle_max_time`   | Maximum amount of time CPU was in throttled state           | ms    |
+| `throtlle_total_time` | Cumulative time during which the CPU was in throttled state | ms    |
+
+## Example Output
+
+```shell
+> linux_cpu,cpu=0,host=go scaling_max_freq=4700000i,cpuinfo_min_freq=400000i,cpuinfo_max_freq=4700000i,throttle_count=0i,throttle_max_time=0i,throttle_total_time=0i,scaling_cur_freq=803157i,scaling_min_freq=400000i 1617621150000000000
+> linux_cpu,cpu=1,host=go throttle_total_time=0i,scaling_cur_freq=802939i,scaling_min_freq=400000i,scaling_max_freq=4700000i,cpuinfo_min_freq=400000i,cpuinfo_max_freq=4700000i,throttle_count=0i,throttle_max_time=0i 1617621150000000000
+> linux_cpu,cpu=10,host=go throttle_max_time=0i,throttle_total_time=0i,scaling_cur_freq=838343i,scaling_min_freq=400000i,scaling_max_freq=4700000i,cpuinfo_min_freq=400000i,cpuinfo_max_freq=4700000i,throttle_count=0i 1617621150000000000
+> linux_cpu,cpu=11,host=go cpuinfo_max_freq=4700000i,throttle_count=0i,throttle_max_time=0i,throttle_total_time=0i,scaling_cur_freq=800054i,scaling_min_freq=400000i,scaling_max_freq=4700000i,cpuinfo_min_freq=400000i 1617621150000000000
+> linux_cpu,cpu=2,host=go throttle_total_time=0i,scaling_cur_freq=800404i,scaling_min_freq=400000i,scaling_max_freq=4700000i,cpuinfo_min_freq=400000i,cpuinfo_max_freq=4700000i,throttle_count=0i,throttle_max_time=0i 1617621150000000000
+> linux_cpu,cpu=3,host=go throttle_total_time=0i,scaling_cur_freq=800126i,scaling_min_freq=400000i,scaling_max_freq=4700000i,cpuinfo_min_freq=400000i,cpuinfo_max_freq=4700000i,throttle_count=0i,throttle_max_time=0i 1617621150000000000
+> linux_cpu,cpu=4,host=go cpuinfo_max_freq=4700000i,throttle_count=0i,throttle_max_time=0i,throttle_total_time=0i,scaling_cur_freq=800359i,scaling_min_freq=400000i,scaling_max_freq=4700000i,cpuinfo_min_freq=400000i 1617621150000000000
+> linux_cpu,cpu=5,host=go throttle_max_time=0i,throttle_total_time=0i,scaling_cur_freq=800093i,scaling_min_freq=400000i,scaling_max_freq=4700000i,cpuinfo_min_freq=400000i,cpuinfo_max_freq=4700000i,throttle_count=0i 1617621150000000000
+> linux_cpu,cpu=6,host=go cpuinfo_max_freq=4700000i,throttle_count=0i,throttle_max_time=0i,throttle_total_time=0i,scaling_cur_freq=741646i,scaling_min_freq=400000i,scaling_max_freq=4700000i,cpuinfo_min_freq=400000i 1617621150000000000
+> linux_cpu,cpu=7,host=go scaling_cur_freq=700006i,scaling_min_freq=400000i,scaling_max_freq=4700000i,cpuinfo_min_freq=400000i,cpuinfo_max_freq=4700000i,throttle_count=0i,throttle_max_time=0i,throttle_total_time=0i 1617621150000000000
+> linux_cpu,cpu=8,host=go throttle_max_time=0i,throttle_total_time=0i,scaling_cur_freq=700046i,scaling_min_freq=400000i,scaling_max_freq=4700000i,cpuinfo_min_freq=400000i,cpuinfo_max_freq=4700000i,throttle_count=0i 1617621150000000000
+> linux_cpu,cpu=9,host=go throttle_count=0i,throttle_max_time=0i,throttle_total_time=0i,scaling_cur_freq=700075i,scaling_min_freq=400000i,scaling_max_freq=4700000i,cpuinfo_min_freq=400000i,cpuinfo_max_freq=4700000i 1617621150000000000
+```
--- a/plugins/inputs/linux_cpu/linux_cpu.go
+++ b/plugins/inputs/linux_cpu/linux_cpu.go
@@ -1,0 +1,214 @@
+//go:build linux
+
+package linux_cpu
+
+import (
+	_ "embed"
+
+	"fmt"
+	"io"
+	"os"
+	"path"
+	"path/filepath"
+	"strconv"
+	"strings"
+
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/internal/choice"
+	"github.com/influxdata/telegraf/plugins/inputs"
+)
+
+const (
+	defaultHostSys = "/sys"
+	cpufreq        = "cpufreq"
+	thermal        = "thermal"
+)
+
+type LinuxCPU struct {
+	Log       telegraf.Logger `toml:"-"`
+	PathSysfs string          `toml:"host_sys"`
+	Metrics   []string        `toml:"metrics"`
+	cpus      []cpu
+}
+
+type cpu struct {
+	id    string
+	path  string
+	props map[string]string
+}
+
+type prop struct {
+	name     string
+	path     string
+	optional bool
+}
+
+//go:embed sample.conf
+var sampleConfig string
+
+func (g *LinuxCPU) SampleConfig() string {
+	return sampleConfig
+}
+
+func (g *LinuxCPU) Init() error {
+	if g.PathSysfs == "" {
+		g.PathSysfs = defaultHostSys
+	}
+
+	if len(g.Metrics) == 0 {
+		// The user has not enabled any of the metrics
+		return fmt.Errorf("no metrics selected")
+	}
+
+	cpus, err := g.discoverCpus()
+	if err != nil {
+		return err
+	} else if len(cpus) == 0 {
+		// Although the user has specified metrics to collect, `discoverCpus` failed to find the required metrics
+		return fmt.Errorf("no CPUs detected to track")
+	}
+	g.cpus = cpus
+
+	return nil
+}
+
+func (g *LinuxCPU) Gather(acc telegraf.Accumulator) error {
+	for _, cpu := range g.cpus {
+		fields := make(map[string]interface{})
+		tags := map[string]string{"cpu": cpu.id}
+
+		failed := false
+		for name, propPath := range cpu.props {
+			v, err := readUintFromFile(propPath)
+			if err != nil {
+				acc.AddError(err)
+				failed = true
+				break
+			}
+
+			fields[name] = v
+		}
+
+		if !failed {
+			acc.AddFields("linux_cpu", fields, tags)
+		}
+	}
+
+	return nil
+}
+
+func (g *LinuxCPU) discoverCpus() ([]cpu, error) {
+	var cpus []cpu
+
+	glob := path.Join(g.PathSysfs, "devices/system/cpu/cpu[0-9]*")
+	cpuDirs, err := filepath.Glob(glob)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(cpuDirs) == 0 {
+		return nil, fmt.Errorf("no CPUs detected at: %s", glob)
+	}
+
+	for _, dir := range cpuDirs {
+		_, cpuName := filepath.Split(dir)
+		cpuNum := strings.TrimPrefix(cpuName, "cpu")
+
+		cpu := cpu{
+			id:    cpuNum,
+			path:  dir,
+			props: make(map[string]string),
+		}
+
+		var props []prop
+
+		if choice.Contains(cpufreq, g.Metrics) {
+			props = append(props,
+				prop{name: "scaling_cur_freq", path: "cpufreq/scaling_cur_freq", optional: false},
+				prop{name: "scaling_min_freq", path: "cpufreq/scaling_min_freq", optional: false},
+				prop{name: "scaling_max_freq", path: "cpufreq/scaling_max_freq", optional: false},
+				prop{name: "cpuinfo_cur_freq", path: "cpufreq/cpuinfo_cur_freq", optional: true},
+				prop{name: "cpuinfo_min_freq", path: "cpufreq/cpuinfo_min_freq", optional: true},
+				prop{name: "cpuinfo_max_freq", path: "cpufreq/cpuinfo_max_freq", optional: true},
+			)
+		}
+
+		if choice.Contains(thermal, g.Metrics) {
+			props = append(
+				props,
+				prop{name: "throttle_count", path: "thermal_throttle/core_throttle_count", optional: false},
+				prop{name: "throttle_max_time", path: "thermal_throttle/core_throttle_max_time_ms", optional: false},
+				prop{name: "throttle_total_time", path: "thermal_throttle/core_throttle_total_time_ms", optional: false},
+			)
+		}
+
+		var failed = false
+		for _, prop := range props {
+			propPath := filepath.Join(dir, prop.path)
+			err := validatePath(propPath)
+			if err != nil {
+				if prop.optional {
+					continue
+				}
+
+				g.Log.Warnf("Failed to load property %s: %v", propPath, err)
+				failed = true
+				break
+			}
+
+			cpu.props[prop.name] = propPath
+		}
+
+		if len(cpu.props) == 0 {
+			g.Log.Warnf("No properties enabled/loaded for CPU %s", cpuNum)
+			failed = true
+		}
+
+		if !failed {
+			cpus = append(cpus, cpu)
+		}
+	}
+	return cpus, nil
+}
+
+func init() {
+	inputs.Add("linux_cpu", func() telegraf.Input {
+		return &LinuxCPU{
+			Metrics: []string{"cpufreq"},
+		}
+	})
+}
+
+func validatePath(propPath string) error {
+	f, err := os.Open(propPath)
+
+	if os.IsNotExist(err) {
+		return fmt.Errorf("CPU property does not exist: [%s]", propPath)
+	}
+
+	if err != nil {
+		return fmt.Errorf("cannot get system information for CPU property: [%s] - %v", propPath, err)
+	}
+
+	_ = f.Close() // File is not written to, closing should be safe
+	return nil
+}
+
+func readUintFromFile(propPath string) (uint64, error) {
+	f, err := os.Open(propPath)
+	if err != nil {
+		return 0, err
+	}
+	defer f.Close()
+
+	buffer := make([]byte, 22)
+
+	n, err := f.Read(buffer)
+	if err != nil && err != io.EOF {
+		return 0, fmt.Errorf("error on reading file, err: %v", err)
+	} else if n == 0 {
+		return 0, fmt.Errorf("error on reading file, file is empty")
+	}
+
+	return strconv.ParseUint(string(buffer[:n-1]), 10, 64)
+}
--- a/plugins/inputs/linux_cpu/linux_cpu_test.go
+++ b/plugins/inputs/linux_cpu/linux_cpu_test.go
@@ -1,0 +1,183 @@
+//go:build linux
+
+package linux_cpu
+
+import (
+	"github.com/influxdata/telegraf/testutil"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestNoMetrics(t *testing.T) {
+	plugin := &LinuxCPU{}
+	require.Error(t, plugin.Init())
+}
+
+func TestNoCPUs(t *testing.T) {
+	td := t.TempDir()
+
+	plugin := &LinuxCPU{
+		Log:       testutil.Logger{Name: "LinuxCPUPluginTest"},
+		Metrics:   []string{"cpufreq"},
+		PathSysfs: td,
+	}
+	require.Error(t, plugin.Init())
+}
+
+func TestNoCPUMetrics(t *testing.T) {
+	td := t.TempDir()
+
+	require.NoError(t, os.MkdirAll(td+"/devices/system/cpu/cpu0/cpufreq", os.ModePerm))
+
+	plugin := &LinuxCPU{
+		Log:       testutil.Logger{Name: "LinuxCPUPluginTest"},
+		Metrics:   []string{"cpufreq"},
+		PathSysfs: td,
+	}
+	require.Error(t, plugin.Init())
+}
+
+func TestGatherCPUFreq(t *testing.T) {
+	td := t.TempDir()
+
+	require.NoError(t, os.MkdirAll(td+"/devices/system/cpu/cpu0/cpufreq", os.ModePerm))
+	require.NoError(t, os.WriteFile(td+"/devices/system/cpu/cpu0/cpufreq/scaling_cur_freq", []byte("250\n"), 0644))
+	require.NoError(t, os.WriteFile(td+"/devices/system/cpu/cpu0/cpufreq/scaling_min_freq", []byte("100\n"), 0644))
+	require.NoError(t, os.WriteFile(td+"/devices/system/cpu/cpu0/cpufreq/scaling_max_freq", []byte("255\n"), 0644))
+
+	require.NoError(t, os.MkdirAll(td+"/devices/system/cpu/cpu1/cpufreq", os.ModePerm))
+	require.NoError(t, os.WriteFile(td+"/devices/system/cpu/cpu1/cpufreq/scaling_cur_freq", []byte("123\n"), 0644))
+	require.NoError(t, os.WriteFile(td+"/devices/system/cpu/cpu1/cpufreq/scaling_min_freq", []byte("80\n"), 0644))
+	require.NoError(t, os.WriteFile(td+"/devices/system/cpu/cpu1/cpufreq/scaling_max_freq", []byte("230\n"), 0644))
+
+	plugin := &LinuxCPU{
+		Log:       testutil.Logger{Name: "LinuxCPUPluginTest"},
+		Metrics:   []string{"cpufreq"},
+		PathSysfs: td,
+	}
+
+	require.NoError(t, plugin.Init())
+
+	var acc testutil.Accumulator
+	require.NoError(t, plugin.Gather(&acc))
+
+	tags1 := map[string]string{
+		"cpu": "0",
+	}
+
+	tags2 := map[string]string{
+		"cpu": "1",
+	}
+
+	fields1 := map[string]interface{}{
+		"scaling_cur_freq": uint64(250),
+		"scaling_min_freq": uint64(100),
+		"scaling_max_freq": uint64(255),
+	}
+
+	fields2 := map[string]interface{}{
+		"scaling_cur_freq": uint64(123),
+		"scaling_min_freq": uint64(80),
+		"scaling_max_freq": uint64(230),
+	}
+
+	acc.AssertContainsTaggedFields(t, "linux_cpu", fields1, tags1)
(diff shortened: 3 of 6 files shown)

### CODE REVIEW

File: plugins/inputs/cpufreq/cpufreq.go [hunk 1]
```diff
@@ -0,0 +1,157 @@
...
+
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/plugins/inputs"
+)
+
+const defaultHostSys = "/sys"
+
+type CPUFreq struct {
+	Log             telegraf.Logger `toml:"-"`
+	CPUs            []CPU           `toml:"-"`
```
zak-pawel, line 22: `CPUs` can be unexported => `cpus`

plugins/inputs/cpufreq/cpufreq.go
```diff
@@ -0,0 +1,157 @@
...
+	PathSysfs       string          `toml:"host_sys"`
+	GatherThrottles bool            `toml:"gather_throttles"`
+}
+
+type CPU struct {
+	ID    string
+	Path  string
+	Props map[string]*os.File
+}
```
zak-pawel, line 31: Struct and all these properties can be unexported

plugins/inputs/cpufreq/README.md
```diff
@@ -0,0 +1,46 @@
+# CPU Frequency Input Plugin
+
+The `cpufreq` plugin gathers metrics of the system CPU's frequency using Linux' sysfs.
+
+#### Configuration
+```toml
+[[inputs.cpufreq]]
```
zak-pawel, line 7:
```
```toml
# Collects current CPU's frequencies
[[inputs.cpufreq]]
```

plugins/inputs/cpufreq/cpufreq.go
```diff
@@ -0,0 +1,157 @@
...
+  ## Defaults:
+  # gather_throttles = false
+`
+
+func (g *CPUFreq) SampleConfig() string {
+	return sampleConfig
+}
+
+func (g *CPUFreq) Description() string {
+	return "Collect current CPU's frequencies"
```
zak-pawel, line 49: `Collects`

plugins/inputs/cpufreq/cpufreq.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+func (g *CPUFreq) Init() error {
+	if os.Getenv("HOST_SYS") != "" {
+		g.PathSysfs = os.Getenv("HOST_SYS")
+	} else {
+		g.PathSysfs = defaultHostSys
+	}
```
zak-pawel, line 57: Because of `if-else`, `host_sys` paramater from config will be never taken into consideration - it will be always overwritten by content of `os.Getenv("HOST_SYS")` or by `defaultHostSys`.

I would wrap this `if-else` with:
```
if g.PathSysfs == "" {
...
}
```

plugins/inputs/cpufreq/cpufreq.go
```diff
@@ -0,0 +1,157 @@
...
+
+		if !failed {
+			acc.AddFields("cpufreq", fields, tags)
+		}
+	}
+
+	return nil
+}
+
+func (g *CPUFreq) DiscoverCpus() ([]CPU, error) {
```
zak-pawel, line 93: `discoverCpus()`

plugins/inputs/cpufreq/cpufreq.go
```diff
@@ -0,0 +1,157 @@
...
+	var cpus []CPU
+
+	cpuDirs, err := filepath.Glob(path.Join(g.PathSysfs, "devices/system/cpu/cpu[0-9]*"))
+	if err != nil {
+		return nil, err
```
zak-pawel, line 98: If `g.PathSysfs` contains path to nonexistent directory (because of rubbish in `HOST_SYS` env or in `host_sys` parameter in config file), `cpuDirs` will be empty forever. 

User will not get any metrics and will not get any information why these metrics are not here...

I would fail right here and inform user what is wrong. You can also check: `if len(cpuDirs) == 0` and return an error containing content of this command: `path.Join(g.PathSysfs, "devices/system/cpu/cpu[0-9]*")`

plugins/inputs/cpufreq/cpufreq.go
```diff
@@ -0,0 +1,157 @@
...
+	return sampleConfig
+}
+
+func (g *CPUFreq) Description() string {
+	return "Collect current CPU's frequencies"
+}
+
+func (g *CPUFreq) Init() error {
+	if os.Getenv("HOST_SYS") != "" {
+		g.PathSysfs = os.Getenv("HOST_SYS")
```
zak-pawel, line 54: Setting `g.PathSysfs` based on `HOST_SYS` env is undocumented and may lead to problems with not getting any metrics.

Personally, I would remove this `if` and use `defaultHostSys` (only when `host_sys` was not set in config).

plugins/inputs/cpufreq/cpufreq.go
```diff
@@ -0,0 +1,157 @@
...
+		return nil, err
+	}
+
+	for _, dir := range cpuDirs {
+		_, cpuName := filepath.Split(dir)
+		cpuNum := strings.TrimPrefix(cpuName, "cpu")
+
+		if _, err := os.Stat(filepath.Join(dir, "cpufreq")); os.IsNotExist(err) {
+			g.Log.Warn("CPUFreq subsystem not available")
+			continue
```
zak-pawel, line 107: Is it possible that CPUFreq subsystem is not available only for some CPUs and is available for rest of them?

If not, I would return error instead of logging a warning (fail fast ;)).

plugins/inputs/cpufreq/cpufreq.go
```diff
@@ -0,0 +1,157 @@
...
+
+		if g.GatherThrottles {
+			props["thermal_throttle/core_throttle_count"] = "throttle_count"
+		}
+
+		var failed = false
+		for prop, name := range props {
+			fd, err := os.Open(filepath.Join(dir, prop))
+			if err != nil {
+				g.Log.Warn("Failed to load property of cpu ", cpuNum, ": ", filepath.Join(dir, prop))
```
zak-pawel, line 129: Can you add `err` to warning message?

Moreover, logging `cpuNum` seems to be redundant - it will be contained in `filepath.Join(dir, prop)`

zak-pawel (changes requested):
Please, check few minor comments.
It would be great to see a few unit tests for this plugin.

plugins/inputs/cpufreq/cpufreq.go
fabianishere: I don't think this is possible in Linux, but I'll check to make sure.

plugins/inputs/cpufreq/cpufreq.go
fabianishere: Agreed, will remove the code.

fabianishere:
@zak-pawel Thanks for the comments. I have tried to address them.

Do you have any advice regarding the following question:
> the thermal_throttle/core_throttle_count field is a bit out of place in this plugin, since it is not part of the CPUFreq subsystem in Linux. We could limit this plugin to only expose properties from the CPUFreq subsystem and have a separate plugin for thermal_throttle or we could emit all properties that Linux exposes per CPU (and rename the plugin to cpulinux for instance)?

I might be interested in tracking other CPU-related metrics exposed by Linux but that are not part of the CPUFreq subsystem (like `thermal_throttle/core_throttle_count`)? Do you prefer to have separate input plugins for each of the CPU subsystems or  would a single input for all per-CPU metrics exposed by Linux be a better approach for Telegraf?

plugins/inputs/cpufreq/cpufreq.go
```diff
@@ -0,0 +1,158 @@
...
+		cpuNum := strings.TrimPrefix(cpuName, "cpu")
+
+		if _, err := os.Stat(filepath.Join(dir, "cpufreq")); os.IsNotExist(err) {
+			return nil, fmt.Errorf("CPUFreq subsystem not available for cpu %s", cpuNum)
+		}
+
+		cpu := cpu{
+			id:    cpuNum,
+			path:  dir,
+			props: make(map[string]*os.File, 0),
```
zak-pawel, line 114: should use `make(map[string]*os.File)` instead

plugins/inputs/cpufreq/cpufreq.go
```diff
@@ -0,0 +1,158 @@
...
+		}
+
+		props := make(map[string]string, 0)
```
zak-pawel, line 117: should use `make(map[string]string)` instead

plugins/inputs/cpufreq/cpufreq.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+type cpu struct {
+	id    string
+	path  string
+	props map[string]*os.File
+}
+
+var sampleConfig = `
+  ## path for sysfs filesystem.
```
zak-pawel, line 34: `Path`

plugins/inputs/cpufreq/cpufreq.go
```diff
@@ -0,0 +1,158 @@
...
+  ## See https://www.kernel.org/doc/Documentation/filesystems/sysfs.txt
+  ## Defaults:
+  # host_sys = "/sys"
+
+  ## Gather cpu throttles per core
```
zak-pawel, line 39: `CPU`

plugins/inputs/cpufreq/cpufreq.go
```diff
@@ -0,0 +1,158 @@
...
+func (g *CPUFreq) discoverCpus() ([]cpu, error) {
+	var cpus []cpu
+
+	cpuDirs, err := filepath.Glob(path.Join(g.PathSysfs, "devices/system/cpu/cpu[0-9]*"))
+	if err != nil {
+		return nil, err
+	}
+
+	if len(cpuDirs) == 0 {
+		return nil, fmt.Errorf("no CPUs detected at: %s", path.Join(g.PathSysfs, "devices/system/cpu/cpu[0-9]*"))
```
zak-pawel, line 100: `path.Join(g.PathSysfs, "devices/system/cpu/cpu[0-9]*")` can be saved to `var`

plugins/inputs/cpufreq/cpufreq.go
```diff
@@ -0,0 +1,158 @@
...
+
+		if g.GatherThrottles {
+			props["thermal_throttle/core_throttle_count"] = "throttle_count"
+		}
+
+		var failed = false
+		for prop, name := range props {
+			fd, err := os.Open(filepath.Join(dir, prop))
+			if err != nil {
+				g.Log.Warn("Failed to load property: ", filepath.Join(dir, prop), err)
```
zak-pawel, line 130: `Warnf`

plugins/inputs/cpufreq/cpufreq.go
```diff
@@ -0,0 +1,158 @@
...
+	if len(cpuDirs) == 0 {
+		return nil, fmt.Errorf("no CPUs detected at: %s", path.Join(g.PathSysfs, "devices/system/cpu/cpu[0-9]*"))
+	}
+
+	for _, dir := range cpuDirs {
+		_, cpuName := filepath.Split(dir)
+		cpuNum := strings.TrimPrefix(cpuName, "cpu")
+
+		if _, err := os.Stat(filepath.Join(dir, "cpufreq")); os.IsNotExist(err) {
+			return nil, fmt.Errorf("CPUFreq subsystem not available for cpu %s", cpuNum)
```
zak-pawel, line 108: Maybe you can write something about this subsystem (and how to enable it) in `README.md`?

Please see `README.md` for `intel_powerstat` plugin.

plugins/inputs/cpufreq/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+```
+> cpufreq,cpu=0,host=Z370M-DS3H current_khz=1382522000i,max_khz=4900000000i,min_khz=800000000i 1604049750000000000
+> cpufreq,cpu=1,host=Z370M-DS3H current_khz=1094884000i,max_khz=4900000000i,min_khz=800000000i 1604049750000000000
+> cpufreq,cpu=2,host=Z370M-DS3H current_khz=1010482000i,max_khz=4900000000i,min_khz=800000000i 1604049750000000000
+> cpufreq,cpu=3,host=Z370M-DS3H current_khz=2089249000i,max_khz=4900000000i,min_khz=800000000i 1604049750000000000
+> cpufreq,cpu=4,host=Z370M-DS3H current_khz=1272475000i,max_khz=4900000000i,min_khz=800000000i 1604049750000000000
+> cpufreq,cpu=5,host=Z370M-DS3H current_khz=1374903000i,max_khz=4900000000i,min_khz=800000000i 1604049750000000000
+> cpufreq,cpu=6,host=Z370M-DS3H current_khz=1355753000i,max_khz=4900000000i,min_khz=800000000i 1604049750000000000
+> cpufreq,cpu=7,host=Z370M-DS3H current_khz=1153656000i,max_khz=4900000000i,min_khz=800000000i 1604049750000000000
```
zak-pawel, line 46: Cut some `0`s to have kHz ;)

zak-pawel (changes requested):

zak-pawel:
> @zak-pawel Thanks for the comments. I have tried to address them.
> 
> Do you have any advice regarding the following question:
> 
> > the thermal_throttle/core_throttle_count field is a bit out of place in this plugin, since it is not part of the CPUFreq subsystem in Linux. We could limit this plugin to only expose properties from the CPUFreq subsystem and have a separate plugin for thermal_throttle or we could emit all properties that Linux exposes per CPU (and rename the plugin to cpulinux for instance)?
> 
> I might be interested in tracking other CPU-related metrics exposed by Linux but that are not part of the CPUFreq subsystem (like `thermal_throttle/core_throttle_count`)? Do you prefer to have separate input plugins for each of the CPU subsystems or would a single input for all per-CPU metrics exposed by Linux be a better approach for Telegraf?

Hard to say... What metrics can be gathered from `thermal_throttle`? Is `thermal_throttle` available widely in Linux? Can it be enabled/disabled?
@ssoroka @reimda What do you think?

plugins/inputs/cpufreq/cpufreq.go
fabianishere: CPUFreq must be enabled during build time of the kernel, though it is enabled in all mainstream distributions AFAIK. Is it it sufficient to say that it must be supported by the user's kernel (along with a description of the subsystem)?

fabianishere:
@zak-pawel `thermal_throttle` is exposes on Intel systems a couple metrics that might indicate that the CPU is being thermally throttled. One of the previous authors added the `core_throttle_count` field, which counts  the amount of thermal throttle events reported by the CPU.

Perhaps to give a more concrete example of my question: Linux also exposes information about the idle states via `/sys/devices/system/cpu/cpuX/cpuidle` (e.g., C-state residency, similar to what's reported by `intel_powerstat`), which I would like to track. It would not be much work to support these metrics in this input plugin, but if we do so the name `cpufreq` for the plugin might not be appropriate.

fabianishere:
@zak-pawel Is there still anything I can do from my side?

zak-pawel:
> @zak-pawel Is there still anything I can do from my side?

I don't think so :) Just waiting for some comments from @reimda / @ssoroka about `thermal_throttle` and other subsytems you had mentioned before.

reimda:
Hi Fabian, thanks for picking up #4215.  It makes sense to me for this plugin to include things like /sys/devices/system/cpu/cpuX/cpuidle that are outside of the cpufreq linux subsystem.  In that case I agree a more general name would be useful to quickly communicate to the user what the plugin does.

Would the name linux_cpu be more appropriate?  There's already a linux_sysctl_fs plugin that matches this naming scheme.

cpu_linux is another option.  It has the advantage of showing up next to the cpu plugin in sorted list, so it might be more visible.  I think I prefer linux_cpu.

Both these options position this plugin as the place for linux specific metrics.  That leaves the cpu plugin as the place for cross platform metrics (through gopsutil).  It would be great to have only one cpu plugin but I don't think it's practical to expect gopsutil to provide the metrics this plugin does.

If this is going to be the place for all linux cpu metrics, I would expect the field names to match linux terms.  For example, scaling_cur_freq instead of current_khz.

fabianishere:
@reimda I agree with your points! My plan is to rename the plugin to `linux_cpu` and start adding other metrics that Linux exposes. I will work on this the coming weekend, so you can expect an update soon after.

fabianishere:
@zak-pawel I have re-requested your review. Although not all possible metrics are included yet, I want to get your opinion on the current design. I have two questions in particular:
1. When should the plugin fail? It could be the case that some metrics do not exist on the system, does that mean that we fail all together?
2. To what granularity should the user be able to control the metrics that are being collected by the plugin? I can imagine that some users are not interested in tracking all possible metrics and might want to reduce overhead and the amount of data collected.

plugins/inputs/all/all.go
```diff
@@ -30,6 +30,7 @@ import (
 	_ "github.com/influxdata/telegraf/plugins/inputs/couchbase"
 	_ "github.com/influxdata/telegraf/plugins/inputs/couchdb"
 	_ "github.com/influxdata/telegraf/plugins/inputs/cpu"
+	_ "github.com/influxdata/telegraf/plugins/inputs/linux_cpu"
```
zak-pawel, line 33: Please, sort this.

plugins/inputs/linux_cpu/README.md
```diff
@@ -0,0 +1,62 @@
+# Linux CPU Input Plugin
+The `linux_cpu` plugin gathers CPU metrics exposed on Linux-based systems.
+
+#### Configuration
+```toml
+# Collects current CPU's frequencies
+[[inputs.cpufreq]]
```
zak-pawel, line 7: `[[inputs.linux_cpu]]`

plugins/inputs/linux_cpu/README.md
```diff
@@ -0,0 +1,62 @@
+# Linux CPU Input Plugin
+The `linux_cpu` plugin gathers CPU metrics exposed on Linux-based systems.
+
+#### Configuration
+```toml
+# Collects current CPU's frequencies
```
zak-pawel, line 6: `# Collects CPU metrics exposed on Linux`

plugins/inputs/linux_cpu/README.md
```diff
@@ -0,0 +1,62 @@
...
+  ## Defaults:
+  # host_sys = "/sys"
+ 
+  ## Gather CPU frequency information
+  ## Defaults:
+  # gather_cpufreq = true
+
+  ## Gather CPU throttles per core
+  ## Defaults:
+  # gather_throttles = false
```
zak-pawel, line 19: Please sync it with content of `sampleConfig` var.

plugins/inputs/linux_cpu/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+```
+> linux_cpu,cpu=0,host=Z370M-DS3H scaling_cur_freq=1382522i,scaling_max_freq=4900000i,scaling_min_freq=800000i 1604049750000000000
+> linux_cpu,cpu=1,host=Z370M-DS3H scaling_cur_freq=1094884i,scaling_max_freq=4900000i,scaling_min_freq=800000i 1604049750000000000
+> linux_cpu,cpu=2,host=Z370M-DS3H scaling_cur_freq=1010482i,scaling_max_freq=4900000i,scaling_min_freq=800000i 1604049750000000000
+> linux_cpu,cpu=3,host=Z370M-DS3H scaling_cur_freq=2089249i,scaling_max_freq=4900000i,scaling_min_freq=800000i 1604049750000000000
+> linux_cpu,cpu=4,host=Z370M-DS3H scaling_cur_freq=1272475i,scaling_max_freq=4900000i,scaling_min_freq=800000i 1604049750000000000
+> linux_cpu,cpu=5,host=Z370M-DS3H scaling_cur_freq=1374903i,scaling_max_freq=4900000i,scaling_min_freq=800000i 1604049750000000000
+> linux_cpu,cpu=6,host=Z370M-DS3H scaling_cur_freq=1355753i,scaling_max_freq=4900000i,scaling_min_freq=800000i 1604049750000000000
+> linux_cpu,cpu=7,host=Z370M-DS3H scaling_cur_freq=1153656i,scaling_max_freq=4900000i,scaling_min_freq=800000i 1604049750000000000
```
zak-pawel, line 61: Would be great to have an example with all possible fields.

plugins/inputs/linux_cpu/linux_cpu.go
```diff
@@ -0,0 +1,188 @@
...
+	"github.com/influxdata/telegraf/plugins/inputs"
+)
+
+const defaultHostSys = "/sys"
+
+type LinuxCPU struct {
+	Log             telegraf.Logger `toml:"-"`
+	PathSysfs       string          `toml:"host_sys"`
+	GatherCPUFreq   bool            `toml:"gather_cpufreq"`
+	GatherThrottles bool            `toml:"gather_throttles"`
```
zak-pawel, line 24: Please, avoid boolean fields when a string or enumerated type would be better for future extension. Lots of boolean fields also make the code difficult to maintain.

(taken from: https://github.com/influxdata/telegraf/wiki/Review#reviewing-plugin-code )

plugins/inputs/linux_cpu/linux_cpu.go
```diff
@@ -0,0 +1,188 @@
...
+		var failed = false
+		for _, prop := range props {
+			fd, err := os.Open(filepath.Join(dir, prop.path))
+			if err != nil && !prop.optional {
+				g.Log.Warnf("Failed to load property %s: %s", filepath.Join(dir, prop.path), err)
+				failed = true
+				break
+			}
+
+			cpu.props[prop.name] = fd
```
zak-pawel, line 156: So we may get here when `err != nil && prop.optional`. 

Do we want to get here in that case? :)

plugins/inputs/linux_cpu/linux_cpu.go
```diff
@@ -0,0 +1,188 @@
...
+func (g *LinuxCPU) Init() error {
+	if g.PathSysfs == "" {
+		g.PathSysfs = defaultHostSys
+	}
+
+	cpus, err := g.discoverCpus()
+	if err != nil {
+		return err
+	}
+	g.cpus = cpus
```
zak-pawel, line 72: Theoretically, we may have `len(g.cpus) == 0`.

Shouldn't we fail fast in that case right here?

plugins/inputs/linux_cpu/linux_cpu.go
```diff
@@ -0,0 +1,188 @@
...
+			)
+		}
+
+		var failed = false
+		for _, prop := range props {
+			fd, err := os.Open(filepath.Join(dir, prop.path))
+			if err != nil && !prop.optional {
+				g.Log.Warnf("Failed to load property %s: %s", filepath.Join(dir, prop.path), err)
+				failed = true
+				break
```
zak-pawel, line 153: Don't we want to `continue` here? Is there a possibility to gather these non-optional properties which don't fail?

plugins/inputs/linux_cpu/linux_cpu.go
```diff
@@ -0,0 +1,188 @@
+// +build linux
+
+package linux_cpu
```
zak-pawel, line 3: I would love to see unit tests for this plugin.

zak-pawel (changes requested):
1. I think that plugin should fail (during `Init()`) when there are no metrics to gather at all. For other cases, it would be good to gather what is possible.
2. I think that these two flags (which should be converted to array of strings in config ;)) will control the metrics that are being collected by the plugin just fine. But I would fail as soon as possible if both of them are not enabled (== no metrics to gather).

plugins/inputs/linux_cpu/linux_cpu.go
```diff
@@ -0,0 +1,211 @@
...
+
+	n, err := fd.ReadAt(buffer, 0)
+	if err != nil && err != io.EOF {
+		return 0, fmt.Errorf("error on reading file, err: %v", err)
+	}
+
+	return strconv.ParseUint(string(buffer[:n-1]), 10, 64)
+}
+
+func contains(slice []string, str string) bool {
```
zak-pawel, line 203: There is such function in Telegraf internals. `choice.Contains` and this can be removed :)

plugins/inputs/linux_cpu/linux_cpu.go
```diff
@@ -0,0 +1,211 @@
...
+  ## Path for sysfs filesystem.
+  ## See https://www.kernel.org/doc/Documentation/filesystems/sysfs.txt
+  ## Defaults:
+  # host_sys = "/sys"
+
+  ## CPU metrics collected by the plugin.
+  ## Supported options:
+  ## "cpufreq", "thermal"
+  ## Defaults:
+  # metrics = ["cpufreq""]
```
zak-pawel, line 53: Too many `"`

plugins/inputs/linux_cpu/linux_cpu.go
```diff
@@ -0,0 +1,211 @@
...
+				props,
+				prop{ name: "throttle_count", path: "thermal_throttle/core_throttle_count", optional: false },
+				prop{ name: "throttle_max_time", path: "thermal_throttle/core_throttle_max_time_ms", optional: false },
+				prop{ name: "throttle_total_time", path: "thermal_throttle/core_throttle_total_time_ms", optional: false },
+			)
+		}
+
+		var failed = false
+		for _, prop := range props {
+			fd, err := os.Open(filepath.Join(dir, prop.path))
```
zak-pawel, line 158: You open file here and you never close it...
Better idea would be to check here if file exists and is a regular file (check `validateDbPath` function in `ras.go` for example) and to save fullPath to file in `cpu.props`. Then in `readUintFromFile` you would open, read, close every file.

zak-pawel (commented):
Still without unit tests :(

plugins/inputs/linux_cpu/linux_cpu.go
fabianishere: Correct. I am a bit hesitant on having the collector perform multiple syscalls (for each of these metrics) every collection. With the current approach, a file descriptor is openend for each of the metrics and performs just one read call every collection cycle. The downside of this is that the process has more open file descriptors (possibly too many).

fabianishere:
Thanks @zak-pawel for the review. I have been a bit busy these weeks with other things. 

Given that I don't have a lot of experience with Go, I am still researching what's the best way to test the plugin. I am probably going to approach it the same way as the `intel_powerstat` plugin where they mocked the FS interface. Let me know if you have any suggestions yourself!

sjwang90:
@fabianishere Do you have a chance to continue working on this? Or else we'll see if someone else wants to take it over.

fabianishere:
I was planning on finishing this at the start of next year. Currently rather busy with wrapping up some other things.

antst:
When run as unprivileged user, this would have no permissions to read /sys/devices/system/cpu/cpuXX/cpufreq/cpuinfo_cur_freq , but because file exists, it will keep trying to read and pollute logs.
validatePath() should also check for permissions.
Either really trying to open file, or check "o+r" flag.

fabianishere:
@antst Good catch! I’ll probably go with just trying to open the file.

powersj:
@zak-pawel I think this is ready for another review since you last looked at it? Is that true?

zak-pawel:
> @zak-pawel I think this is ready for another review since you last looked at it? Is that true?

I would love to see few unit tests :)

fabianishere:
@zak-pawel Sorry for the delay, I was quite busy with some other work last months. 

I have rebased the changes onto the master branch and added a few unit tests. Let me know if you want me to cover any additional cases.

plugins/inputs/linux_cpu/linux_cpu.go
```diff
@@ -0,0 +1,226 @@
...
+
+	if os.IsNotExist(err) {
+		return fmt.Errorf("CPU property does not exist: [%s]", propPath)
+	}
+
+	if err != nil {
+		return fmt.Errorf("cannot get system information for CPU property: [%s] - %v", propPath, err)
+	}
+
+	_ = f.Close() // File is not written to, closing should be safe
```
zak-pawel, line 205: Maybe this would be nicer?

```
func validatePath(propPath string) error {
	_, err := os.Lstat(propPath)
	if err != nil {
		if os.IsNotExist(err) {
			return fmt.Errorf("CPU property does not exist: [%s]", propPath)
		}
		return fmt.Errorf("cannot get system information for CPU property: [%s] - %v", propPath, err)
	}

	return nil
}
```

plugins/inputs/linux_cpu/linux_cpu.go
```diff
@@ -0,0 +1,226 @@
...
+	buffer := make([]byte, 22)
+
+	n, err := f.Read(buffer)
+	if err != nil && err != io.EOF {
+		return 0, fmt.Errorf("error on reading file, err: %v", err)
+	} else if n == 0 {
+		return 0, nil
+	}
+
+	return strconv.ParseUint(string(buffer[:n-1]), 10, 64)
```
zak-pawel, line 225: Maybe this would be a little bit cleaner?
```
	read, err := io.ReadAll(f)
	if err != nil {
		return 0, err
	}
	
	trimmedString := strings.TrimRight(string(read), "\n")
	return strconv.ParseUint(trimmedString, 10, 64)
```

plugins/inputs/linux_cpu/linux_cpu_test.go
```diff
@@ -0,0 +1,98 @@
...
+func TestNoCPUs(t *testing.T) {
+	td := t.TempDir()
+
+	plugin := &LinuxCPU{
+		Metrics:   []string{"cpufreq"},
+		PathSysfs: td,
+	}
+	require.Error(t, plugin.Init())
+}
+
```
zak-pawel, line 27: Would be also nice to test scenario when `discoverCpus()` returns no error and empty lists (it would be for case when directories for cpus exist but all properties cannot be loaded).

plugins/inputs/linux_cpu/linux_cpu_test.go
```diff
@@ -0,0 +1,98 @@
...
+	require.NoError(t, os.MkdirAll(td+"/devices/system/cpu/cpu0/cpufreq", os.ModePerm))
+	require.NoError(t, os.WriteFile(td+"/devices/system/cpu/cpu0/cpufreq/scaling_cur_freq", []byte("250\n"), 0644))
+	require.NoError(t, os.WriteFile(td+"/devices/system/cpu/cpu0/cpufreq/scaling_min_freq", []byte("100\n"), 0644))
+	require.NoError(t, os.WriteFile(td+"/devices/system/cpu/cpu0/cpufreq/scaling_max_freq", []byte("255\n"), 0644))
+
+	require.NoError(t, os.MkdirAll(td+"/devices/system/cpu/cpu1/cpufreq", os.ModePerm))
+	require.NoError(t, os.WriteFile(td+"/devices/system/cpu/cpu1/cpufreq/scaling_cur_freq", []byte("123\n"), 0644))
+	require.NoError(t, os.WriteFile(td+"/devices/system/cpu/cpu1/cpufreq/scaling_min_freq", []byte("80\n"), 0644))
+	require.NoError(t, os.WriteFile(td+"/devices/system/cpu/cpu1/cpufreq/scaling_max_freq", []byte("230\n"), 0644))
```
zak-pawel, line 39: Would be nice to test scenario when property cannot be loaded (property path doesn't exist and/or value of property cannot be parsed to uint)

zak-pawel (changes requested):

plugins/inputs/linux_cpu/linux_cpu.go
fabianishere: This relates to the comment from @antst. Checking whether the file exists is not enough, we need to verify that it is readable as well.

plugins/inputs/linux_cpu/linux_cpu.go
fabianishere: My preference would be the first approach, since we know that the input cannot exceed 22 bytes if the value is a `uint`, so we don't need to allocate a dynamic buffer.

plugins/inputs/linux_cpu/linux_cpu.go
zak-pawel: Ok.

plugins/inputs/linux_cpu/linux_cpu.go
zak-pawel: Ok, will not insist. Do you want to return `0` as a value when file was read by was empty?

fabianishere: Hmm, good question, perhaps that should also give an error. What do you think?

zak-pawel: I think that missing value probably will not mean that 0 should be used as a default value instead. So error could be better option.

plugins/inputs/linux_cpu/linux_cpu.go
```diff
@@ -0,0 +1,218 @@
...
+			cpus = append(cpus, cpu)
+		}
+	}
+	return cpus, nil
+}
+
+func init() {
+	inputs.Add("linux_cpu", func() telegraf.Input {
+		return &LinuxCPU{
+			Metrics: []string{"cpufreq", "thermal"},
```
zak-pawel, line 181: According to READMEmd, this would be default:
`Metrics: []string{"cpufreq"},`

zak-pawel (changes requested):

sspaink (approved):
Great job! Thank you so much for keeping this pull request alive, I think it looks great.

@zak-pawel is out, but it looks like all the comments have been resolved. If there are any other changes I think we should handle them in a separate PR, merging.

reimda (approved):

PR #8989, opened 2021-03-15
Update to Go 1.16.2

sspaink:
Update Go to 1.16.2, also update how Mac installs Go.
Migrated away from brew because it was so slow to get latest version and we couldn't pull exact minor version, also added check to see if Go is already installed.


sspaink:
!retry-checks

Jagularr (approved):

scripts/mac_installgo.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/sh
+
+# Install Go directly from tar, the reason we aren't using brew: it is slow to update and we can't pull specific minor versions
+install_go () {
+    echo "installing go"
+    curl -OL https://golang.org/dl/go1.16.2.darwin-amd64.tar.gz --output go1.16.2.darwin-amd64.tar.gz
+    sudo rm -rf /usr/local/go && sudo tar -C /usr/local -xzf go1.16.2.darwin-amd64.tar.gz
+    ln -s /usr/local/go/bin/go /usr/local/bin
```
ssoroka, line 8: might need -f for existing symlinks

scripts/mac_installgo.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/sh
+
+# Install Go directly from tar, the reason we aren't using brew: it is slow to update and we can't pull specific minor versions
+install_go () {
+    echo "installing go"
+    curl -OL https://golang.org/dl/go1.16.2.darwin-amd64.tar.gz --output go1.16.2.darwin-amd64.tar.gz
```
ssoroka, line 6: maybe parameterize the go version

sspaink:
I think this is the final version, I was having issues with the cache in circle-ci throwing a weird error that was evidently caused by failing to restore from the path `/usr/local/go` but using `/usr/local/Cellar/go` is working. I've also merged the jobs `mac-deps` and `test-go-mac`, this should hopefully be faster because there wasn't any benefit keeping these steps separate.

ssoroka (approved):

Issue #7937, opened 2020-08-04
sflow: invalid memory address or nil pointer dereference on Ubiquiti Security Gateway


midzelis:
Getting this stack trace when connecting Ubility Security Gateway flow accounting to this sflow input plugin. 

```
telegraf      | panic: runtime error: invalid memory address or nil pointer dereference
telegraf      | [signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x19517d0]
telegraf      |
telegraf      | goroutine 21 [running]:
telegraf      | github.com/influxdata/telegraf/plugins/inputs/sflow.RawPacketHeaderFlowData.GetTags(...)
telegraf      | 	/go/src/github.com/influxdata/telegraf/plugins/inputs/sflow/types.go:121
telegraf      | github.com/influxdata/telegraf/plugins/inputs/sflow.makeMetrics(0xc0007233e0, 0x194ab2f, 0xc000245520, 0x2ed53e0, 0xc000741a40, 0xc0001abb80)
telegraf      | 	/go/src/github.com/influxdata/telegraf/plugins/inputs/sflow/metricencoder.go:29 +0x6fc
telegraf      | github.com/influxdata/telegraf/plugins/inputs/sflow.(*SFlow).Start.func1(0xc0007233e0)
telegraf      | 	/go/src/github.com/influxdata/telegraf/plugins/inputs/sflow/sflow.go:66 +0x4d
telegraf      | github.com/influxdata/telegraf/plugins/inputs/sflow.(*PacketDecoder).Decode(0xc000245520, 0x2ed53e0, 0xc000741a40, 0x0, 0x0)
telegraf      | 	/go/src/github.com/influxdata/telegraf/plugins/inputs/sflow/packetdecoder.go:40 +0x42
telegraf      | github.com/influxdata/telegraf/plugins/inputs/sflow.(*SFlow).process(0xc000522000, 0x2f495c0, 0xc0002455a0, 0xc00065e000, 0x4c4, 0x10000)
telegraf      | 	/go/src/github.com/influxdata/telegraf/plugins/inputs/sflow/sflow.go:135 +0x83
telegraf      | github.com/influxdata/telegraf/plugins/inputs/sflow.(*SFlow).read(0xc000522000, 0x2f495c0, 0xc0002455a0, 0x2f426c0, 0xc000010528)
telegraf      | 	/go/src/github.com/influxdata/telegraf/plugins/inputs/sflow/sflow.go:129 +0x8c
[...]

PR #8995, opened 2021-03-16
fix segfaults in sflow plugin by checking if protocol headers are set

fmoessbauer:
prior to calling getFields() or getTags() on them.

If a protocol header is not available, an empty map is returned.

updates: #7937

Signed-off-by: Felix Moessbauer <felix.moessbauer@siemens.com>

### Required for all PRs:

### DIFF
--- a/plugins/inputs/sflow/types.go
+++ b/plugins/inputs/sflow/types.go
@@ -118,12 +118,22 @@ type RawPacketHeaderFlowData struct {
 }
 
 func (h RawPacketHeaderFlowData) GetTags() map[string]string {
-	t := h.Header.GetTags()
+	var t map[string]string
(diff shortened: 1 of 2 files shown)


jessingrass:
Hello, please add unit tests for the two types (RawPacketHeaderFlowData and EthHeader) to make sure they're not returning nil even when there are no tags or fields. Thanks very much!

fmoessbauer:
@jessingrass : I added the unit test. Actually, there are tags for all Ethernet packets, but not necessarily encapsulated protocol headers. This is checked in the unit tests.

ssoroka (approved):

ssoroka:
thank you!

Issue #7501, opened 2020-05-12
Support merging configs


ssoroka:
## Feature Request

Explicitly around the case of supporting multiple disparate Telegraf configurations, which comes in from 1. cloud, 2. community template, and 3. Telegraf configuration service, we need the ability to take multiple desired Telegraf configs merge them into a single Telegraf configuration that Telegraf can run.

An acceptable solution to this could be an external project or library, but that is not necessarily a requirement.

This is important because community template users currently need to manually merge configs. If we don't do this project, it will encourage template authors to work around the issue by merging separate templates into "template stacks" to make template installation easier, at a high cost to flexibility and an explosion in variants (vs what would be small composable templates if we had this).

### Use cases:

1. Installing multiple community templates installs multiple Telegraf configurations. It's not unusual for a single server to be collecting data for multiple templates, eg, redis, system, internal, kafka, or any stack you can imagine.
2. This avoids you having to predefine Telegraf "roles" where a role is a set of plugins to use. You could generate the "role" config on the fly from known configurations.

### Requirements/Nice-to-Haves: 

- [R] Take two configs and produce a single config from it
- [R] Combine multiple outputs pointing to the same destination, respecting buckets
- [N] Generate a default output (eg, for cloud), when none is defined (when desired)
- [N] auto pipelining/namespacing, (or tagpassing?) to ensure configs don't have unintended behavior when merged

### Unresolved issues:

- There will be issues around merging multiple configurations with complex workflows, eg, a lot of processors/aggregators, especially with ones using generic field names. Stand-alone configs can get away with not checking metric names, but can cause issues when merged. Config pipelining/namespacing could be a solution here.

### Proposal:

- Create a new library that knows how to understand/merge multiple configs, smart enough to remove duplicate configs, with idempotent merge behavior.
- [option 1] extend Telegraf to accept multiple --config arguments
- [option 2] create a stand-alone binary that can merge config files.

danielnelson:
My initial reaction is that InfluxDB Templates should make use of the `--config-directory` to add configuration, which in effect merges it.  Can you expand on why this functionality is needed?

M0rdecay:
Is it possible to extend the functionality of the `--config-directory` key so that several directories can be specified?

Usage example - we store agent configurations in directories with the names of the applications to which they relate. The scheme is approximately as follows:
```
.
├── bin
│   └── telegraf
├── conf
│   ├── telegraf.conf
│   ├── base.d
│   ├── app_1.d
[...]

PR #9007, opened 2021-03-18
Feature: merge multiple "--config" and "--config-directory" flags

gdunstone:
This adds a feature mentioned in this issue https://github.com/influxdata/telegraf/issues/7501

The only change in functionality is that telegraf will now merge multiple "--config" and "--config-directory" flags into the config, rather than only using value of the last flag.

### Required for all PRs:

### DIFF
--- a/cmd/telegraf/telegraf_windows.go
+++ b/cmd/telegraf/telegraf_windows.go
@@ -84,12 +84,18 @@ func runAsWindowsService(inputFilters, outputFilters, aggregatorFilters, process
 	// Handle the --service flag here to prevent any issues with tooling that
 	// may not have an interactive session, e.g. installing from Ansible.
 	if *fService != "" {
-		if *fConfig != "" {
-			svcConfig.Arguments = []string{"--config", *fConfig}
+		var err error
+		if len(fConfigs) > 0 {
+			svcConfig.Arguments = []string{}	
 		}
-		if *fConfigDirectory != "" {
-			svcConfig.Arguments = append(svcConfig.Arguments, "--config-directory", *fConfigDirectory)
+		for _, fConfig := range fConfigs {
+			svcConfig.Arguments = append(svcConfig.Arguments, "--config", fConfig)
 		}
+
+		for _, fConfigDirectory := range fConfigDirs {
+			svcConfig.Arguments = append(svcConfig.Arguments, "--config-directory", fConfigDirectory)
+		}
+
 		//set servicename to service cmd line, to have a custom name after relaunch as a service
 		svcConfig.Arguments = append(svcConfig.Arguments, "--service-name", *fServiceName)
 
--- a/cmd/telegraf/telegraf_windows.go
+++ b/cmd/telegraf/telegraf_windows.go
@@ -84,7 +84,6 @@ func runAsWindowsService(inputFilters, outputFilters, aggregatorFilters, process
 	// Handle the --service flag here to prevent any issues with tooling that
 	// may not have an interactive session, e.g. installing from Ansible.
 	if *fService != "" {
-		var err error
 		if len(fConfigs) > 0 {
 			svcConfig.Arguments = []string{}
 		}
--- a/cmd/telegraf/telegraf_windows.go
+++ b/cmd/telegraf/telegraf_windows.go
@@ -86,7 +86,7 @@ func runAsWindowsService(inputFilters, outputFilters, aggregatorFilters, process
 	if *fService != "" {
 		var err error
 		if len(fConfigs) > 0 {
-			svcConfig.Arguments = []string{}	
+			svcConfig.Arguments = []string{}
 		}
 		for _, fConfig := range fConfigs {
 			svcConfig.Arguments = append(svcConfig.Arguments, "--config", fConfig)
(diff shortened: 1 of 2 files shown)


srebhan (commented):
While the code looks good to me, you have to fix the CI errors for the windows build before this has a chance to be merged. The error is
```
# github.com/influxdata/telegraf/cmd/telegraf
cmd\telegraf\telegraf_windows.go:87:7: undefined: fConfig
cmd\telegraf\telegraf_windows.go:88:48: undefined: fConfig
cmd\telegraf\telegraf_windows.go:90:7: undefined: fConfigDirectory
cmd\telegraf\telegraf_windows.go:91:77: undefined: fConfigDirectory
```

srebhan:
And `make fmt` plz! :-)

gdunstone:
Sorry about that, I figured it was quicker to just fix it rather than reply. 

Thanks for reviewing this.

srebhan:
!retry-failed

srebhan:
!retry-failed

srebhan (approved):
Looks good to me.

Jagularr (approved):

Hipska:
Since, `config-directory` already walks recursively, I don't think it is needed to have the possibility to define multiple on the command-line?

srebhan:
@Hipska well imagine a case where you have one central (directory) repository that defines the "used-everywhere" plugins and then a machine-local directory for holding the machine-individual configs. Those will likely not be nested. We currently approximate such a use case with symbolic links in one "machine-local" directory but it's... a workaround. :-)

Hipska:
Both of these directories can be put in a parent directory and that one will be parsed recursively, so I still see no actual use case. It seems better to KISS as much as possible.

srebhan:
@Hipska let me put an example. The central repository is located on a server and mounted e.g. to `/myserver/configs/telegraf`, while on some of your machines you additionally want to put a config e.g. for special hardware on this machine only. Let's assume those config files reside on `/usr/local/configs/telegraf`. Of course you could now merge those two directories on a filesystem level (e.g. merge by copying the files, or (as I do) symlink the two, etc) but how is that better than being able to say
`telegraf ... --config-directory /myserver/configs/telegraf --config-directory /usr/local/configs/telegraf`?

gdunstone:
@Hipska Going through a directory recursively doesn't allow you to fetch and merge multiple config files over http.

Hipska:
> Of course you could now merge those two directories on a filesystem level (e.g. merge by copying the files, or (as I do) symlink the two, etc) but how is that better than being able to say
> `telegraf ... --config-directory /myserver/configs/telegraf --config-directory /usr/local/configs/telegraf`?

@srebhan symlink would indeed be better. I would avoid adding 'complex' logic on the command line arguments.

> Going through a directory recursively doesn't allow you to fetch and merge multiple config files over http.

@gdunstone Correct, that's why I have absolutely no objection towards multiple `--config` arguments. I'm only against multiple `--config-directory` flags.

gdunstone:
@Hipska I am ambivalent about the `--config-directory` flag as I don't use it. My main objective with this PR was to allow merging multiple http configs with the `--config`  flag, the `--config-directory` changes were just because it was convenient and I didn't see why not at the time.

However I can see the use case @srebhan provides.

srebhan:
@Hipska I still don't see the complex logic point, but to cut a long story short: I can live with only one config directory option even though I'd prefer multiple ones. ;-)

Hipska:
And I prefer not 😂 Best would be someone from Influx to give their thoughts.

If you want to include multiple different directories, I think it would be better to make that an option in the toml config rather than on the command line.

srebhan:
The only difference is: Now only you is happy, with the change we can both be happy as nobody forces you to _use_ the option multiple times.

Hipska:
You are right 😛 But maybe put it on the table for the weekly meeting to know their view on it?

cmd/telegraf/telegraf_windows.go
```diff
@@ -84,12 +84,17 @@ func runAsWindowsService(inputFilters, outputFilters, aggregatorFilters, process
 	// Handle the --service flag here to prevent any issues with tooling that
 	// may not have an interactive session, e.g. installing from Ansible.
 	if *fService != "" {
-		if *fConfig != "" {
-			svcConfig.Arguments = []string{"--config", *fConfig}
+		if len(fConfigs) > 0 {
+			svcConfig.Arguments = []string{}
```
ssoroka, line 88: this looks like it's not necessary. append doesn't need initialization and the 0 len case does not hit this.

ssoroka:
I'm okay with multiple --config and multiple --config-directory flags. It adds flexibility but doesn't harm the users who are not using it.  This is good to go once my comment is resolved.

cmd/telegraf/telegraf_windows.go
gdunstone: I think this is not initialization, its clearing the default config from here: https://github.com/influxdata/telegraf/blob/f26bfe6755a1c5c8750bdece387ddece56838e9c/cmd/telegraf/telegraf_windows.go#L71

If a user specifies config files via command line options we should use those config files only, not the default + their configs.

Hipska (approved):

ssoroka (approved):

PR #9016, opened 2021-03-18
Update supported regions in azure monitor

koudaiii:
### Required for all PRs:

- [x] ~~Has appropriate unit tests.~~ does not modify code

## Why

Azure Monitor Custom Metrics is supported in All Public Cloud Regions. ref https://docs.microsoft.com/en-us/azure/azure-monitor/essentials/metrics-custom-overview#supported-regions
I think that should update README.

## What

Update Azure Monitor README.
I deleted "The following regions are currently known to be supported", because of already supported All Public Regions.


koudaiii:
!signed-cla

sspaink (approved):

PR #9039, opened 2021-03-23
Replace github.com/kubernetes import with k8s.io and update protobuf to v1.5.1

### LINKED ISSUE #9021: fix "duplicate proto type registered"
Full error message: `2021/03/18 16:19:25 proto: duplicate proto type registered: k8s.io.apimachinery.pkg.api.resource.Quantity`

There are two different protobuf libaries being used, `github.com/gogo/protobuf v1.3.1` and `github.com/golang/protobuf v1.4.3`.
This could be causing the issue, but there might be another cause.

sspaink:
Closes #9021 

Fixing this import path to match the one used everywhere else removes the "duplicate proto" message.

Also updated protobuf library to the latest change, this didn't fix the issue but seemed like a good change to keep.

### DIFF
--- a/plugins/inputs/kube_inventory/kube_state.go
+++ b/plugins/inputs/kube_inventory/kube_state.go
@@ -10,7 +10,7 @@ import (
 	"sync"
 	"time"
 
-	"github.com/kubernetes/apimachinery/pkg/api/resource"
+	"k8s.io/apimachinery/pkg/api/resource"
 
 	"github.com/influxdata/telegraf"
 	"github.com/influxdata/telegraf/filter"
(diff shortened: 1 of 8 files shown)


Jagularr (approved):

plugins/inputs/kube_inventory/kube_state.go
```diff
@@ -10,7 +10,7 @@ import (
 	"sync"
 	"time"
 
-	"github.com/kubernetes/apimachinery/pkg/api/resource"
+	"k8s.io/apimachinery/pkg/api/resource"
```
reimda, line 13: github.com/kubernetes and k8s.io are mirrors of each other.  In order to not end up with two versions in go.mod and golang type conflicts in the future we need to use one mirror in telegraf.  The prometheus plugin uses github.com/kubernetes/apimachinery, so it makes sense to use github.com/kubernetes everywhere in telegraf and remove all references to k8s.io.

PR #9004, opened 2021-03-17
Add support of the time module in Starlark Processor

essobedo:
### Required for all PRs:

- [x] Associated README.md updated.
- [x] Has appropriate unit tests.

fixes #9000

## Motivation

The upstream project starlark-go has added [a new module called time](https://github.com/google/starlark-go/pull/327) to support dates and durations, thus the goal of this feature request is to add the support of this new module in the Starlark Processor.  

## Modifications

* Upgrade the version of `go.starlark.net`
* Import the module `time` of `go.starlark.net` in the Starlark Processor
* Add a test showing how to parse date, duration and timestamp
* Refer the tests into the `README.md`

## Result

In Starlark script, we can now import the module with `load('time.star', 'time')`, and then have access to all the corresponding building functions and constants.

**NB:** Beware the built-in function `from_timestamp` only supports timestamps in seconds.

PR #9042, opened 2021-03-24
Add the math module to the Starlark Processor

### LINKED ISSUE #8985: processors.starlark support math module
## Feature Request

Starlark supports basic math functions out of the box (+,-,/,*,...) but for some conversions more extended functions are desired (cos, sin, exp, ...). Starlark supports them by the [math.star](https://pkg.go.dev/github.com/qri-io/starlib@v0.2.0/math) module but it is not supported by Telegraf.

### Proposal:

Add support to load the math module in starlark scripts, or even load it by default (if there is no overhead, i have no knowledge of starlark's internals/design).

### Current behavior:

```
[[processors.starlark]]
  source = '''
load('math.star', 'math') 
def apply(metric):
	return metric
'''
```

Error: `[telegraf] Error running agent: could not initialize processor starlark: cannot load math.star: module math.star is not available`

### Desired behavior:

Use more extended math functions in starlark through `[[processors.starlark]]`; loading the math module.

### Use case:

In value conversions use functions like exp(...), cos(...), ect as they cannot be implemented by the default available math functions (+,-,/,*) and are required in some conversions.

essobedo:
### Required for all PRs:

resolves #8985 

## Motivation

The math module has been added recently to the upstream project `starlark-go`, we would like to make the Starlark processor support it too

## Modifications:

* Adds the math module in the list of external modules to load
* Adds a simple example showing how to use it
* Update the `README.md` to refer the new example

## Result

We can use math functions in the Starlark scripts executed by the Starlark Processor.

### DIFF
--- a/plugins/processors/starlark/README.md
+++ b/plugins/processors/starlark/README.md
@@ -222,6 +222,7 @@ def apply(metric):
 - [drop fields with unexpected type](/plugins/processors/starlark/testdata/drop_fields_with_unexpected_type.star) - Drop fields containing unexpected value types.
 - [iops](/plugins/processors/starlark/testdata/iops.star) - obtain IOPS (to aggregate, to produce max_iops)
 - [json](/plugins/processors/starlark/testdata/json.star) - an example of processing JSON from a field in a metric
+- [math](/plugins/processors/starlark/testdata/math.star) - Use a math function to compute the value of a field. [The list of the supported math functions and constants](https://pkg.go.dev/go.starlark.net/lib/math).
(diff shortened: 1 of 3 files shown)


Jagularr (approved):

ssoroka (approved):
do we need to update starlark versions for this?

essobedo:
> do we need to update Starlark versions for this? I guess not if it passes. :D

Yes indeed we don't need as it has already been updated within the context of https://github.com/influxdata/telegraf/pull/9004

PR #9046, opened 2021-03-24
Linter fixes - gosimple

zak-pawel:
Here are fixes for problems found by `gosimple`:
- S1003: should use `strings.ContainsAny(s, "*?[")` instead
  - instead of `strings.IndexAny(s, "*?[") >= 0`
- S1039: unnecessary use of `fmt.Sprintf`
  - example: `msg := fmt.Sprintf("catmaster is incorrect")`
- S1019: should use `make(map[string][]point)` instead
  - instead of `make(map[string][]point, 0)`
- S1023: redundant `return` statement
- S1023: redundant `break` statement 
- S1025: the argument is already a string, there's no need to use `fmt.Sprintf`
   - example: `fmt.Sprintf("%s", firstParseErrorStr)`
- S1005: unnecessary assignment to the blank identifier
   - exameple `min, _ := f.values[statisticTypeMin]`
- S1004: should use `bytes.Equal(value, []byte("No"))` instead
   - instead of `bytes.Compare(value, []byte("No")) == 0`
- S1021: should merge variable declaration with assignment on next line
- S1009: should omit nil check; `len()` for nil slices is defined as zero
   - example: `if rest == nil || len(rest) == 0 {`
- S1002: should omit comparison to bool constant, can be simplified to `!ok`
   - example: `if ok == false {`
- S1034: assigning the result of this type assertion to a variable `(switch i := i.(type))` could eliminate type assertions in switch cases
- S1033: unnecessary guard around call to `delete`
- S1008: should use `return f.namePass.Match(key)` instead of `if f.namePass.Match(key) { return true }; return false`
- S1011: should replace loop with `m.metrics = append(m.metrics, metrics...)`
- S1012: should use `time.Since` instead of `time.Now().Sub`

These need to be reviewed carefully and fixed in the near future:
- S1007: should use raw string (`...`) with `regexp.Compile` to avoid having to escape twice
- S1000: should use for range instead of `for { select {} }`
- S1000: should use a simple channel send/receive instead of `select` with a single case


plugins/inputs/graylog/graylog.go
```diff
@@ -203,13 +203,13 @@ func (h *GrayLog) flatten(item map[string]interface{}, fields map[string]interfa
 		id = id + "_"
 	}
 	for k, i := range item {
-		switch i.(type) {
+		switch i := i.(type) {
 		case int:
-			fields[id+k] = i.(float64)
+			fields[id+k] = i
```
ssoroka, line 208: wait, how did this ever work? the type assertion would have panicked. If the intent is to make sure we always write it as a float, it should be `float64(i)`

ssoroka (approved):

plugins/inputs/graylog/graylog.go
zak-pawel: I didn't :) 
Corrected to `float64(i)` - as you suggested.

Jagularr (approved):

helenosheaa (approved):

PR #9050, opened 2021-03-25
Fix inputs.snmp init when no mibs installed

Hipska:
### Required for all PRs:

The snmp plugin would log translating errors if the required mibs are not installed, even when a lookup is not actually needed, because the user has the numerical `oid` and `name` already configured.

Now the plugin will not attempt a lookup if these are configured like that.

### DIFF
--- a/plugins/inputs/snmp/snmp.go
+++ b/plugins/inputs/snmp/snmp.go
@@ -255,19 +255,22 @@ func (f *Field) init() error {
 		return nil
 	}
 
-	_, oidNum, oidText, conversion, err := SnmpTranslate(f.Oid)
-	if err != nil {
-		return fmt.Errorf("translating: %w", err)
-	}
-	f.Oid = oidNum
-	if f.Name == "" {
-		f.Name = oidText
-	}
-	if f.Conversion == "" {
-		f.Conversion = conversion
-	}
+	// check if oid needs translation or name is not set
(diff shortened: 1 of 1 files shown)


Jagularr (approved):

plugins/inputs/snmp/snmp.go
```diff
@@ -255,19 +255,22 @@ func (f *Field) init() error {
...
+		if err != nil {
+			return fmt.Errorf("translating: %w", err)
+		}
+		f.Oid = oidNum
+		if f.Name == "" {
+			f.Name = oidText
+		}
+		if f.Conversion == "" {
+			f.Conversion = conversion
+		}
```
reimda, line 271: I think this is the line the linter is upset about.
```suggestion
```

reimda (approved):
Other than the linter error, this looks good to me.

plugins/inputs/snmp/snmp.go
Hipska: I think we can ignore the linter warning? as it was your intention to leave an empty line to make that ToDo more clear?

reimda: Could you remove the line so the linter doesn't continue to flag it in the future?

Hipska: Done

PR #9051, opened 2021-03-25
Moved samara config out of init into connect

helenosheaa:
This is a fix for if Kafka doesn't connect it wont stop telegraf from starting up

### DIFF
--- a/plugins/outputs/kafka/kafka.go
+++ b/plugins/outputs/kafka/kafka.go
@@ -43,6 +43,7 @@ type Kafka struct {
 
 	Log telegraf.Logger `toml:"-"`
 
+	saramaConfig *sarama.Config
 	producerFunc func(addrs []string, config *sarama.Config) (sarama.SyncProducer, error)
 	producer     sarama.SyncProducer
 
@@ -278,6 +279,8 @@ func (k *Kafka) Init() error {
 		return err
 	}
 
+	k.saramaConfig = config
+
 	// Legacy support ssl config
 	if k.Certificate != "" {
 		k.TLSCert = k.Certificate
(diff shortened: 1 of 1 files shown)


Hipska:
This only fixes the case for Kafka, but what about the other plugins that have this behaviour? #3723

helenosheaa:
Thanks @Hipska, the team will revisit that open issue next week and add thoughts/plans to the issue

ssoroka:
If anyone reading this later is confused, the point of this change is to make it so that Telegraf will not fail to start this single output has a problem connecting. 

To @Hipska 's question, this needs to be addressed on a per-plugin basis. Plugins should use the `Init()` (capital letter `I`) function to check config and error on a configuration problem that should prevent Telegraf from starting, and the `Connect()`/`Start()` functions for actually connecting to external services; if those return errors they will not prevent Telegraf from starting. Most plugins follow this pattern but a few got it wrong and went unnoticed.

oplehto:
> To @Hipska 's question, this needs to be addressed on a per-plugin basis. Plugins should use the `Init()` (capital letter `I`) function to check config and error on a configuration problem that should prevent Telegraf from starting, and the `Connect()`/`Start()` functions for actually connecting to external services; if those return errors they will not prevent Telegraf from starting. Most plugins follow this pattern but a few got it wrong and went unnoticed.

100% this ^ Maybe the PR template should even have some checklist item of verifying what are the fatal/non-fatal failure modes and do they conform to the recommended pattern.

Hipska:
@ssoroka That would be very nice if it is actually true, but this code proves otherwise:
https://github.com/influxdata/telegraf/blob/17e86ab4ca97af454e19d34d6d0e815c248e3a08/agent/agent.go#L253-L257
When the `Start()` returns an error, it will stop all other plugins and then exit..

PR #9067, opened 2021-03-29
inputs.nfsclient: use uint64, also update error handling

### LINKED ISSUE #9028: inputs.nfsclient ErrRange NFS v4.1
### Relevant telegraf.conf:

```toml
[[inputs.nfsclient]]

```

### System info:

NFS v4 (v4.1 to be precise)

### Steps to reproduce:

1. Run telegraf 1.18.0 client
2. Get an error ErrRange: line: [[EXCHANGE_ID: .... ]]

sspaink:
Close #9028

The values received by [mountstats](https://man7.org/linux/man-pages/man8/mountstats.8.html) were out of the range of int64, this pr updates the type to uint64. I couldn't find any definitive documentation that says this is the right type, but this [Prometheus project](https://pkg.go.dev/github.com/prometheus/procfs#MountStats) that seems to do something similar also uses uint64. I added a unit test to test for the values received by @micoq (thanks!) which were out of int64 range. 

Also removed the call to `log.Fatalf`, having it return the error instead.

### DIFF
--- a/plugins/inputs/nfsclient/nfsclient.go
+++ b/plugins/inputs/nfsclient/nfsclient.go
@@ -300,7 +300,7 @@ func (n *NFSClient) processText(scanner *bufio.Scanner, acc telegraf.Accumulator
 		if !skip {
 			err := n.parseStat(mount, export, version, line, acc)
 			if err != nil {
-				return err
+				return fmt.Errorf("could not parseStat: %w", err)
 			}
 		}
 	}
--- a/plugins/inputs/nfsclient/nfsclient.go
+++ b/plugins/inputs/nfsclient/nfsclient.go
@@ -2,7 +2,7 @@ package nfsclient
 
 import (
 	"bufio"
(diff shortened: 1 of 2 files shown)


Jagularr (approved):

plugins/inputs/nfsclient/nfsclient.go
```diff
@@ -293,9 +298,14 @@ func (n *NFSClient) processText(scanner *bufio.Scanner, acc telegraf.Accumulator
 		}
 
 		if !skip {
-			n.parseStat(mount, export, version, line, acc)
+			err := n.parseStat(mount, export, version, line, acc)
+			if err != nil {
+				return err
```
ssoroka, line 303: The error returns in various places here could use some more context. example:
```suggestion
				return fmt.Errorf("could not parseStat: %w", err)
```

ssoroka (approved):

PR #9069, opened 2021-03-29
Add configurable option for the 'path' tag override in the Tail plugin.

### LINKED ISSUE #7649: tail plugin overwrites adds "path" tag
### Expected behavior:

I can configure Telegraf to pass metrics from named pipe into InfluxDB, without loss of data. This worked fine with Telegraf 1.2 (I know, that's ancient).

Writing of this tag should be configurable.

### Actual behavior:

tail input overwrites `path` tag: https://github.com/influxdata/telegraf/blob/release-1.14/plugins/inputs/tail/tail.go#L231 , breaking any `path` tag that can come from tailed file.

Jagularr:
resolves #7649


ssoroka (approved):
perfect

PR #9076, opened 2021-03-31
Only run CI for certain file changes

sspaink:
Updated circle-ci config to check for certain patterns in the latest commit to see if the jobs should run. Credit to this blog post for the idea: https://dev.to/acro5piano/exit-circleci-jobs-if-changed-files-do-not-match-specific-pattern-mel

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,11 +26,8 @@ executors:
 
 commands:
   check-changed-files-or-halt:
-      parameters:
-        pattern:
-          type: string
       steps:
-        - run: git show -m HEAD --name-only --pretty="" | egrep -q '<< parameters.pattern >>' || circleci step halt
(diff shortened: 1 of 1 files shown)


sspaink:
With this pull request, each job will check if the recent changes match the regular expression (needs to be circleci config, makefile, or go/mod/sum file) if it does then it will continue as normal. otherwise it will call "halt". This unfortunately doesn't cause the workflow to stop all it does is stop the job and mark it as green, so it needs to be done in each job. There is a way to cancel the entire workflow but this will make it look grayed out and might seem like it failed (and also requires a circleci token which causes issues with forked pull requests). 

Hope that makes sense, @Jagularr mind looking it over again?

Jagularr (approved):
I like it

PR #9081, opened 2021-04-01
Remove all log.Fatal() calls from within telegraf plugin code, replace with error handling.

Jagularr:
Plugins are meant to return errors that can be handled by Telegraf rather than terminating the process on their own. I've left log.Fatal usage in other places in the project, but only for tests.

### DIFF
--- a/plugins/inputs/opcua/opcua_util.go
+++ b/plugins/inputs/opcua/opcua_util.go
@@ -145,7 +145,7 @@ func pemBlockForKey(priv interface{}) *pem.Block {
 }
 
 // OPT FUNCTIONS
-
+//revive:disable-next-line:argument-limit
 func generateClientOpts(endpoints []*ua.EndpointDescription, certFile, keyFile, policy, mode, auth, username, password string, requestTimeout time.Duration) ([]opcua.Option, error) {
 	opts := []opcua.Option{}
 	appuri := "urn:telegraf:gopcua:client"
--- a/plugins/inputs/opcua/opcua_util.go
+++ b/plugins/inputs/opcua/opcua_util.go
@@ -145,8 +145,9 @@ func pemBlockForKey(priv interface{}) *pem.Block {
 }
 
 // OPT FUNCTIONS
-//revive:disable-next-line
+//revive:disable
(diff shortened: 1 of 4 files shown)


plugins/inputs/opcua/opcua_util.go
```diff
@@ -146,7 +146,7 @@ func pemBlockForKey(priv interface{}) *pem.Block {
 
 // OPT FUNCTIONS
 
-func generateClientOpts(endpoints []*ua.EndpointDescription, certFile, keyFile, policy, mode, auth, username, password string, requestTimeout time.Duration) []opcua.Option {
+func generateClientOpts(endpoints []*ua.EndpointDescription, certFile, keyFile, policy, mode, auth, username, password string, requestTimeout time.Duration) ([]opcua.Option, error) {
```
helenosheaa, line 149: You can switch off the linter for that one line something like:

//revive:disable-next-line:argument-limit

plugins/inputs/webhooks/github/github_webhooks_models.go
```diff
@@ -107,10 +106,7 @@ func (s CommitCommentEvent) NewMetric() telegraf.Metric {
 		"commit":  s.Comment.Commit,
 		"comment": s.Comment.Body,
 	}
-	m, err := metric.New(meas, t, f, time.Now())
-	if err != nil {
-		log.Fatalf("Failed to create %v event", event)
-	}
+	m, _ := metric.New(meas, t, f, time.Now())
```
helenosheaa, line 109: why don't we want to error check these?

helenosheaa (commented):
Looks good overall, two questions/comments in the code. Can you rebase to get the circleci checks?

plugins/inputs/webhooks/github/github_webhooks_models.go
Jagularr: so actually those metric.new methods don't actually have any possibility of returning an error. in the future we should refractor the hundreds of uses of it to handle removing that error possibility, but for now it could be ignored.

plugins/inputs/opcua/opcua_util.go
Jagularr: sweet thanks

PR #9086, opened 2021-04-02
Delete log.Fatal calls and replace with error returns.

Jagularr:
Removing all calls other than in tests. Calls to metric.new have no chance of returning an error, so we can ignore it (it's a mistake that it returns an 'error', and could be refactored soon)

### DIFF
--- a/plugins/inputs/opcua/opcua_util.go
+++ b/plugins/inputs/opcua/opcua_util.go
@@ -290,13 +290,13 @@ func generateAuth(a string, cert []byte, un, pw string) (ua.UserTokenType, opcua
 
 		if un == "" {
 			if err != nil {
-				return 0, nil, fmt.Errorf("error reading username input: %s", err)
+				return 0, nil, fmt.Errorf("error reading the username input: %s", err)
 			}
 		}
 
 		if pw == "" {
 			if err != nil {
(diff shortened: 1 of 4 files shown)


plugins/inputs/opcua/opcua_util.go
```diff
@@ -31,11 +31,11 @@ func newTempDir() (string, error) {
 	return dir, err
 }
 
-func generateCert(host string, rsaBits int, certFile, keyFile string, dur time.Duration) (string, string) {
+func generateCert(host string, rsaBits int, certFile, keyFile string, dur time.Duration) (cert string, key string, err error) {
 	dir, _ := newTempDir()
 
 	if len(host) == 0 {
-		log.Fatalf("Missing required host parameter")
+		return "", "", fmt.Errorf("Missing required host parameter")
```
sspaink, line 38: Error strings should not be capitalized, I think I saw a couple of other places with caps as well.

Reference: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

Jagularr: Yup thanks, fixed

sspaink (approved):

ssoroka (approved):

PR #9096, opened 2021-04-06
Speed up package step by running in parallel.

### LINKED ISSUE #9098: Enhance speed of 'package' job while building PRs with Parallelization
We can likely run build jobs in parallel based on OS type or something, and speed it up as it is currently heavily weighing down the time for PR checks to complete (takes about 20-22 minutes currently).

Jagularr:
resolves #9098

Before this change, a full check suite on a PR took about 30-33 minutes. After this change, I've seen consistent results around 19 minutes. Note that the `make package` command will no longer work for generating all packages, each OS type has to be specified in an argument (such as `windows=1 make package`). This will also require an update to telegraf tiger which will be a separate PR.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -233,7 +233,7 @@ jobs:
     steps:
       - run:
           command: |
-            echo "Tests complete."
+            echo "Go tests complete."
 
   share-artifacts:
     executor: aws-cli/default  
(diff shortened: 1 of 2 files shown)


sspaink (approved):

ssoroka (approved):
seems to make  sense

Hipska:
Also seems to introduce problems: #9233

PR #9097, opened 2021-04-06
add XML examples

sjwang90:
added XML examples:
- cdc COVID-19 vaccine HTTP/XML config 
- London bikeshare XML file
- London bikeshare file/XML config

### DIFF
--- a/plugins/parsers/xml/testcases/cdc-vaccines.conf
+++ b/plugins/parsers/xml/testcases/cdc-vaccines.conf
@@ -1,0 +1,50 @@
+# Example for parsing COVID-19 Vaccine Distribution Allocations by Jurisdiction from CDC
+# 
+#
+# Expected Output:
+# cdc-vaccines,state=Connecticut,url=https://data.cdc.gov/api/views/saz5-9hgg/rows.xml,vaccine_type=Pfizer 1st_dose_allocations=60840i,2nd_dose_allocations=60840i 1617580800000000000
+# cdc-vaccines,state=Maine,url=https://data.cdc.gov/api/views/saz5-9hgg/rows.xml,vaccine_type=Pfizer 1st_dose_allocations=23400i,2nd_dose_allocations=23400i 1617580800000000000
+# cdc-vaccines,state=Massachusetts,url=https://data.cdc.gov/api/views/saz5-9hgg/rows.xml,vaccine_type=Pfizer 1st_dose_allocations=117000i,2nd_dose_allocations=117000i 1617580800000000000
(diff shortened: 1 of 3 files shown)


Jagularr (approved):

plugins/parsers/xml/testcases/cdc-vaccines.conf
```diff
@@ -0,0 +1,50 @@
...
+        metric_name = "'cdc-vaccines'"
+        timestamp = "week_of_allocations"
+        timestamp_format = "2006-01-02T15:04:05"
+
+        [inputs.http.xml.tags]
+            state   = "jurisdiction"
+
+        [inputs.http.xml.fields_int]
+            1st_dose_allocations = "_1st_dose_allocations"
+            2nd_dose_allocations = "_2nd_dose_allocations"
```
srebhan, line 34: Not completely sure, but shouldn't this have the same indention level as `data_format` and `tagexclude`?

plugins/parsers/xml/testcases/cyclehire-london.conf
```diff
@@ -0,0 +1,34 @@
...
+    field_selection = "child::info"
+    field_name = "name(@*[1])"
+    field_value = "number(@*[1])"
+
+    [inputs.tail.xml.tags]
+      address = "@name"
+      id = "@id"
+
+    [inputs.tail.xml.fields]
+      placement = "string(temporary)"
```
srebhan, line 34: The test-cases (unfortunately) have another format. You should do:
```suggestion

metric_selection = "response/child::location"
metric_name = "string('bikes')"

timestamp = "/stations/@lastUpdate"
timestamp_format = "unix_ms"

field_selection = "child::info"
field_name = "name(@*[1])"
field_value = "number(@*[1])"

[tags]
  address = "@name"
  id = "@id"

[fields]
  placement = "string(temporary)"
```

srebhan (changes requested):
Hey Samantha, the test-case format is only a simplified version of the telegraf config and _only_ contains the parser's part without prefixes and can only read local files. Please take a look at my comment regarding the second test-case.

Furthermore, I fear that, due to these limitations, you cannot represent your first test case as is. You might download one of those XML files and put it locally and then parse it without the `enum` processor to generate a testable test-case.

Additionally or alternatively, we might want to generate an _examples_ subdir in the `xml` parser where you can put those full configs and then add a link to the parser's readme. What do you think?

sjwang90:
As discussed on slack, instead of creating a separate `examples` folder we can just add them to our customer facing docs - docs.influxdata.com

PR #9100, opened 2021-04-07
Add doc on supported platforms

reimda:


Jagularr (approved):

docs/SUPPORTED_PLATFORMS.md
```diff
@@ -0,0 +1,92 @@
...
+===================
+Telegraf is released on Linux, FreeBSD, Windows, and macOS. This doc helps define which distributions and releases are targeted.
+
+Linux
+-----
+Telegraf intent: *Support latest versions of the most popular distributions*
+
+Telegraf supports RHEL, Fedora, Debian, and Ubuntu. InfluxData provides package repositories for these distributions. Instructions for using the package repositories can be found on [docs.influxdata.com](https://docs.influxdata.com/telegraf/v1.16/introduction/installation/)
+
+https://distrowatch.com/dwres.php?resource=major
```
srebhan, line 11: Should we add a paragraph mentioning that telegraf _might_ also work for older versions and other distributions? I mean I almost never use the distribution version of telegraf but download the static binary from InfluxData directly instead.

srebhan (commented):
Very nicely written. I added a comment regarding mentioning that other distributions etc. might still work but are not directly supported by InfluxData. It's not a must, but as a user of an alternative OS or distribution you might get the impression that telegraf will not work there, while it does.

reimda:
> Very nicely written. I added a comment regarding mentioning that other distributions etc. might still work but are not directly supported by InfluxData. It's not a must, but as a user of an alternative OS or distribution you might get the impression that telegraf will not work there, while it does.

I updated the doc to clarify that other operating systems and linux distributions may work but are not supported.

docs/SUPPORTED_PLATFORMS.md
```diff
@@ -0,0 +1,102 @@
...
+
+https://github.com/golang/go/wiki/MinimumRequirements#operating-systems
+
+Linux
+-----
+Telegraf intent: *Support latest versions of the most popular distributions*
+
+https://distrowatch.com/dwres.php?resource=major
+
+Telegraf supports RHEL, Fedora, Debian, and Ubuntu. InfluxData provides package repositories for these distributions. Instructions for using the package repositories can be found on [docs.influxdata.com](https://docs.influxdata.com/telegraf/v1.16/introduction/installation/) Bug reports shuld be submitted only for supported distributions and releases.
```
sjwang90, line 17: use `latest` in docs link will link to the latest version: https://docs.influxdata.com/telegraf/latest/introduction/installation/

sjwang90: Add a `.` after the link before `Bug reports`

sjwang90: spelling: shuld >> should

sjwang90 (commented):
Super useful!

General comment is describing what the links are. Could even just describe and hyperlink. Just wasn't really sure what the purpose of the link I was clicking on was as I was going through.

srebhan (approved):
Fine with me but @sjwang90 has some more comments.

docs/SUPPORTED_PLATFORMS.md
```diff
@@ -0,0 +1,102 @@
+Supported Platforms
+===================
+Telegraf is a cross-platform application. This doc helps define which operating systems, distributions, and releases Telegraf supports.
+
+Telegraf is supported on Linux, FreeBSD, Windows, and macOS. It is written in Go which supports these operating systems and more. Telegraf may work on Go's other operating systems and users are welcome to build their own binaries for them. Bug reports should be submitted only for supported platforms.
```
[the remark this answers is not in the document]
reimda: Let's keep this it in mind if we find people ignore the sentence.  For now I'm hoping that putting it in the first paragraph and repeating the idea later on in the Linux section will be enough.

sjwang90 (approved):

SomKen:
Awesome for writing this! Does this mean that Influxdb will be building FreeBSD arm packages now?

PR #9109, opened 2021-04-08
OpenTelemetry Output Plugin

### LINKED ISSUE #9108: Add OpenTelemetry Output plugin
## Feature Request

This feature request is to provide an output plugin from telegraf to the OpenTelemetry protocol (OTLP).

### Proposal:
Add an output plugin for OTLP.

### Current behavior:
No such output exists.

### Use case:
Using telegraf I want to be able to send metrics data to OTLP backends.

codeboten:
### Required for all PRs:

resolves #9108 

This change takes the Stackdriver output plugin as a template and outputs the data in the OpenTelemetry protocol instead. It's currently state still needs:
- [x] refactoring to not bring in the prometheus sidecar OTLP client (this was done as part of the prototyping, I plan on removing this before marking the PR as ready for review)


codeboten:
!signed-cla

README.md
```diff
@@ -444,6 +444,7 @@ For documentation on the latest development code see the [documentation index][d
 * [newrelic](./plugins/outputs/newrelic)
 * [nsq](./plugins/outputs/nsq)
 * [opentsdb](./plugins/outputs/opentsdb)
+* [otlp](./plugins/outputs/otlp) (OpenTelemetry Protocol)
```
srebhan, line 447: Can we please call it `opentelemetry`!? It's much more obvious than the abbreviation...

plugins/outputs/otlp/README.md
```diff
@@ -0,0 +1,34 @@
...
+  # Additional grpc metadata
+  [outputs.otlp.headers]
+    key1 = "value1"
+
+```
+
+### Restrictions
+
+OTLP does not support string values in custom metrics, any string
+fields will not be written.
```
srebhan, line 34:
```suggestion
fields will be omitted not be written to the endpoint.
```

plugins/outputs/otlp/client.go
```diff
@@ -0,0 +1,314 @@
...
+				"DEADLINE_EXCEEDED",
+				"RESOURCE_EXHAUSTED",
+				"ABORTED",
+				"OUT_OF_RANGE",
+				"UNAVAILABLE",
+				"DATA_LOSS"
+			]
+		}
+	}]
+}`
```
srebhan, line 52: As far as I can see this is a one-to-one copy, so why don't you use the variable exported by the cited source:
```suggestion
	serviceConfig = optl.DefaultGRPCServiceConfig
```
Please note: You need to import `import "go.opentelemetry.io/otel/exporters/otlp"` to have access to this variable.

plugins/outputs/otlp/client.go
```diff
@@ -0,0 +1,314 @@
...
+	}
+	return &Client{
+		logger:           logger,
+		url:              conf.URL,
+		timeout:          conf.Timeout,
+		rootCertificates: conf.RootCertificates,
+		headers:          conf.Headers,
+		compressor:       conf.Compressor,
+	}
+}
```
srebhan, line 98: I think this is much too complicated. You _always_ directly specify a new config to `NewClient()`, so why don't you just leave
 ```suggestion
type client struct {
	logger           telegraf.Logger
	url              *url.URL
	timeout          time.Duration
	rootCertificates []string
	headers          metadata.MD
	compressor       string

	conn *grpc.ClientConn
}
```
and directly initialize the client without calling a funciton. Please note that I also un-exported the type as you should only use it internally.

File: plugins/outputs/otlp/client.go [hunk 5]
```diff
@@ -0,0 +1,314 @@
...
+
+// getConnection will dial a new connection if one is not set.  When
+// dialing, this function uses its a new context and the same timeout
+// used for Store().
+func (c *Client) getConnection(ctx context.Context) (_ *grpc.ClientConn, retErr error) {
```
srebhan, line 103: How about naming this `connect()`? Furthermore, you should leave out the first parameter completely if you don't need it!

plugins/outputs/otlp/client.go
```diff
@@ -0,0 +1,314 @@
...
+	if c.conn != nil {
+		return c.conn, nil
+	}
+
+	ctx, cancel := context.WithTimeout(ctx, c.timeout)
+	defer cancel()
+
+	useAuth := c.url.Scheme != "http"
+	_ = level.Debug(c.logger).Log(
```
srebhan, line 112: Please use the telegraf logging factility!

plugins/outputs/otlp/client.go
```diff
@@ -0,0 +1,314 @@
...
+		"msg", "new OTLP connection",
+		"auth", useAuth,
+		"url", c.url.String(),
+		"timeout", c.timeout)
+
+	dopts := []grpc.DialOption{
+		grpc.WithBlock(), // Wait for the connection to be established before using it.
+		grpc.WithDefaultServiceConfig(serviceConfig),
+	}
+	if useAuth {
```
srebhan, line 122: I'd prefer to directly write
```suggestion
	if  c.url.Scheme != "http" {
```
to see what is going on here, but it's a matter of taste...

plugins/outputs/otlp/client.go
```diff
@@ -0,0 +1,314 @@
...
+				ServerName: c.url.Hostname(),
+				RootCAs:    certPool,
+			}
+		}
+		_ = level.Debug(c.logger).Log(
+			"msg", "TLS configured",
+			"server", c.url.Hostname(),
+			"root_certs", fmt.Sprint(c.rootCertificates),
+		)
+		dopts = append(dopts, grpc.WithTransportCredentials(credentials.NewTLS(&tcfg)))
```
srebhan, line 149: I wonder you you do it here... You could specify `tls.ClientConfig` in the plugin struct and let telegraf parse it. Then construct the `tls.Config` in the `Init()` function of the main plugin code and just pass it in here...

plugins/outputs/otlp/client.go
```diff
@@ -0,0 +1,314 @@
...
+	}
+	if c.compressor != "" && c.compressor != "none" {
+		dopts = append(dopts, grpc.WithDefaultCallOptions(grpc.UseCompressor(c.compressor)))
+	}
+	address := c.url.Hostname()
+	if len(c.url.Port()) > 0 {
+		address = net.JoinHostPort(address, c.url.Port())
+	}
+	conn, err := grpc.DialContext(ctx, address, dopts...)
+	c.conn = conn
```
srebhan, line 161: Please check the error first before assigning the connection.

plugins/outputs/otlp/client.go
```diff
@@ -0,0 +1,314 @@
...
+	if err != nil {
+		_ = level.Debug(c.logger).Log(
+			"msg", "connection status",
+			"address", address,
+			"err", err,
+		)
```
srebhan, line 167: I don't think this is necessary as the error will pop-up anyway later in the logs.

plugins/outputs/otlp/client.go
```diff
@@ -0,0 +1,314 @@
...
+		return nil, err
+	}
+
+	return conn, err
```
srebhan, line 171: Why do you need to return the connection _and_ set it in the internal structure. Please decide for one of the two options!

plugins/outputs/otlp/client.go
```diff
@@ -0,0 +1,314 @@
...
+		default:
+			if isRecoverable(err) {
+				_ = level.Info(c.logger).Log("msg", "selftest recoverable error, still trying", "err", err)
+				continue
+			}
+		}
+		return fmt.Errorf(
+			"non-recoverable failure in selftest: %s",
+			truncateErrorString(err),
+		)
```
srebhan, line 202: Please don't truncate errors on this level as you might strip the important details!

plugins/outputs/otlp/client.go
```diff
@@ -0,0 +1,314 @@
...
+			"err", err,
+		)
+		return nil, err
+	}
+
+	return conn, err
+}
+
+// Selftest sends an empty request the endpoint.
+func (c *Client) Selftest(ctx context.Context) error {
```
srebhan, line 175: To me it reads like a `ping` command. If this is the case please name it `ping` and do not export the function, i.e. start the function name with a lowercase letter.

plugins/outputs/otlp/client.go
```diff
@@ -0,0 +1,314 @@
...
+		}
+		return fmt.Errorf(
+			"non-recoverable failure in selftest: %s",
+			truncateErrorString(err),
+		)
+	}
+}
+
+// Store sends a batch of samples to the endpoint.
+func (c *Client) Store(req *metricsService.ExportMetricsServiceRequest) error {
```
srebhan, line 207: Please do not export this function as it should only be used within the boundaries of this package.

plugins/outputs/otlp/client.go
```diff
@@ -0,0 +1,314 @@
...
+	tss := req.ResourceMetrics
+	if len(tss) == 0 {
```
srebhan, line 209:
```suggestion
	if len(req.ResourceMetrics) == 0 {
```

plugins/outputs/otlp/client.go
```diff
@@ -0,0 +1,314 @@
...
+			var err error
+
+			if _, err = service.Export(c.grpcMetadata(ctx), reqCopy, grpc.Trailer(&md)); err != nil {
+				_ = level.Error(c.logger).Log(
+					"msg", "export failure",
+					"err", truncateErrorString(err),
+					"size", proto.Size(reqCopy),
+					"trailers", fmt.Sprint(md),
+					"recoverable", isRecoverable(err),
+				)
```
srebhan, line 250: Please remove the extensive logging on this level. You might want some brief debug messages but errors will be logged anyway.

plugins/outputs/otlp/client.go
```diff
@@ -0,0 +1,314 @@
...
+				"size", proto.Size(reqCopy),
+				"trailers", fmt.Sprint(md),
+			)
+		}(i, end)
+	}
+	wg.Wait()
+	close(errs)
+	if err, ok := <-errs; ok {
+		return err
+	}
```
srebhan, line 267: Is there a gain in doing this in parallel? This makes the code overly complicated, so there should be a gain on the other side of the scale. :-)

plugins/outputs/otlp/client.go
```diff
@@ -0,0 +1,314 @@
...
+	return nil
+}
+
+func (c *Client) Close() error {
```
srebhan, line 271: Please do not export.

plugins/outputs/otlp/client.go
```diff
@@ -0,0 +1,314 @@
...
+	if c.conn == nil {
+		return nil
+	}
+	return c.conn.Close()
+}
+
+func (c *Client) grpcMetadata(ctx context.Context) context.Context {
+	return metadata.NewOutgoingContext(ctx, c.headers)
+}
```
srebhan, line 280: Can we please use the body of the function directly to ease reading the code. 
If you need to understand the code, every indirection means interrupting the reading (and understanding) flow, scroll/search/grep and returning to the initial code. Humans are not very good with this kinds of stacks and context-switches... :-)
So one-liner indirections are not helpful IMO.

plugins/outputs/otlp/client.go
```diff
@@ -0,0 +1,314 @@
...
+// truncateErrorString avoids printing error messages that are very
+// large.
+func truncateErrorString(err error) string {
+	tmp := fmt.Sprint(err)
+	if len(tmp) > maxErrorDetailStringLen {
+		tmp = fmt.Sprint(tmp[:maxErrorDetailStringLen], " ...")
+	}
+	return tmp
+}
+
```
srebhan, line 291: Please do not truncate error-messages on this level. If (at all) this should be done at the plugin level.

plugins/outputs/otlp/client_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+func TestClientWithRecoverableError(t *testing.T) {
+	listener := newLocalListener()
+	grpcServer := grpc.NewServer()
+	mockMetricsServer := metricServiceServer{
+		status: nil,
+	}
+	metricsService.RegisterMetricsServiceServer(grpcServer, &mockMetricsServer)
+	go func() {
+		_ = grpcServer.Serve(listener)
```
srebhan, line 24: Why do you ignore the returned error? I you really need to ignore it, please add a comment. Just silencing the linter is not a good reason. :-)

plugins/outputs/otlp/otlp.go
```diff
@@ -0,0 +1,379 @@
...
+
+import (
+	"context"
+	"fmt"
+	"net/url"
+	"sort"
+	"time"
+
+	// TODO: this should be simplified
+	// Imports the OTLP client from the prometheus sidecar
```
srebhan, line 11: If it should be done, please do it! :-)

plugins/outputs/otlp/otlp.go
```diff
@@ -0,0 +1,379 @@
...
+
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/internal"
+	"github.com/influxdata/telegraf/plugins/outputs"
+	metricsService "go.opentelemetry.io/proto/otlp/collector/metrics/v1"
+	otlpcommonpb "go.opentelemetry.io/proto/otlp/common/v1"
+	otlpmetricpb "go.opentelemetry.io/proto/otlp/metrics/v1"
+	otlpresourcepb "go.opentelemetry.io/proto/otlp/resource/v1"
```
srebhan, line 19: You might want to leave out the `otlp` prefix...

plugins/outputs/otlp/otlp.go
```diff
@@ -0,0 +1,379 @@
...
+	"google.golang.org/grpc/metadata"
+)
+
+// OTLP is the OpenTelemetry Protocol config info.
+type OTLP struct {
```
srebhan, line 24: As I wrote earlier, I prefer calling this `OpenTelemetry`.

plugins/outputs/otlp/otlp.go
```diff
@@ -0,0 +1,379 @@
...
+	Endpoint   string            `toml:"endpoint"`
+	Timeout    string            `toml:"timeout"`
+	Headers    map[string]string `toml:"headers"`
+	Attributes map[string]string `toml:"attributes"`
+
+	Namespace string
```
srebhan, line 30: According to the README this is not a user-configurable option, so please do not export the field and make it private.

plugins/outputs/otlp/otlp.go
```diff
@@ -0,0 +1,379 @@
...
+	Log       telegraf.Logger
```
srebhan, line 31: Please add
```suggestion
	Log       telegraf.Logger `toml:"-"`
```

plugins/outputs/otlp/otlp.go
```diff
@@ -0,0 +1,379 @@
...
+	otlpcommonpb "go.opentelemetry.io/proto/otlp/common/v1"
+	otlpmetricpb "go.opentelemetry.io/proto/otlp/metrics/v1"
+	otlpresourcepb "go.opentelemetry.io/proto/otlp/resource/v1"
+	"google.golang.org/grpc/metadata"
+)
+
+// OTLP is the OpenTelemetry Protocol config info.
+type OTLP struct {
+	Endpoint   string            `toml:"endpoint"`
+	Timeout    string            `toml:"timeout"`
```
srebhan, line 26: Please let telegraf do the heavy-lifting:
```suggestion
	Timeout    config.Duration            `toml:"timeout"`
```

plugins/outputs/otlp/otlp.go
```diff
@@ -0,0 +1,379 @@
...
+	Headers    map[string]string `toml:"headers"`
+	Attributes map[string]string `toml:"attributes"`
+
+	Namespace string
+	Log       telegraf.Logger
+
+	client       *Client
+	resourceTags []*telegraf.Tag
+	grpcTimeout  time.Duration
```
srebhan, line 35: With the change above you can directly use `Timeout`.

plugins/outputs/otlp/otlp.go
```diff
@@ -0,0 +1,379 @@
...
+	// to string length for label key.
+	QuotaStringLengthForLabelKey = 100
+	// QuotaStringLengthForLabelValue is the limit
+	// to string length for label value.
+	QuotaStringLengthForLabelValue = 1024
+
+	// StartTime for cumulative metrics.
+	StartTime = int64(1)
+	// MaxInt is the max int64 value.
+	MaxInt = int(^uint(0) >> 1)
```
srebhan, line 52: Do we need those exported?

plugins/outputs/otlp/otlp.go
```diff
@@ -0,0 +1,379 @@
...
+	defaultTimeout             = time.Second * 60
+	instrumentationLibraryName = "Telegraf"
+)
+
+var sampleConfig = `
+  ## OpenTelemetry endpoint
+  endpoint = "http://localhost:4317"
+
+  ## Timeout used when sending data over grpc
+  timeout = "10s"
```
srebhan, line 64: Please comment all options that have a sensible default
```suggestion
  # endpoint = "http://localhost:4317"

  ## Timeout used when sending data over grpc
  # timeout = "10s"
```

plugins/outputs/otlp/otlp.go
```diff
@@ -0,0 +1,379 @@
...
+
+  # Additional resource attributes
+  [outputs.otlp.attributes]
+  	"service.name" = "demo"
+
+  # Additional grpc metadata
+  [outputs.otlp.headers]
+    key1 = "value1"
+
```
srebhan, line 73: See above.

plugins/outputs/otlp/otlp.go
```diff
@@ -0,0 +1,379 @@
...
+	if o.Headers == nil {
+		o.Headers = make(map[string]string, 1)
+	}
+
+	o.Headers["telemetry-reporting-agent"] = fmt.Sprint(
+		"telegraf/",
+		internal.Version(),
+	)
+
```
srebhan, line 107: I think most of this should move to the `Init()` function. Even if it makes no difference on the program flow it makes clear what is initialization and what is connection handling.

plugins/outputs/otlp/otlp.go
```diff
@@ -0,0 +1,379 @@
...
+	if o.client == nil {
+		ctx := context.Background()
+		o.client = NewClient(ClientConfig{
+			URL:     endpoint,
+			Headers: metadata.New(o.Headers),
+			Timeout: o.grpcTimeout,
+		})
+		if err := o.client.Selftest(ctx); err != nil {
+			_ = o.client.Close()
```
srebhan, line 116: Please add a comment on why you ignore the returned error. You might even rather suppress the linter warnings using a `//nolint:...` statement to allow finding those again.

plugins/outputs/otlp/otlp.go
```diff
@@ -0,0 +1,379 @@
...
+func sorted(metrics []telegraf.Metric) []telegraf.Metric {
+	batch := make([]telegraf.Metric, 0, len(metrics))
+	for i := len(metrics) - 1; i >= 0; i-- {
+		batch = append(batch, metrics[i])
+	}
+	sort.Slice(batch, func(i, j int) bool {
+		return batch[i].Time().Before(batch[j].Time())
+	})
+	return batch
+}
```
srebhan, line 136: This might be an interesting function to have in the accumulator! Anyway I think you should rather return a sorted set of pointers instead of the full metric copies.

plugins/outputs/otlp/otlp.go
```diff
@@ -0,0 +1,379 @@
...
+
+// getNanos converts a millisecond timestamp into a OTLP nanosecond timestamp.
+func getNanos(t int64) uint64 {
+	return uint64(time.Duration(t) * time.Nanosecond)
+}
```
srebhan, line 141: Please don't and use the one.line directly.

plugins/outputs/otlp/otlp.go
```diff
@@ -0,0 +1,379 @@
...
+func doubleGauge(labels []*otlpcommonpb.StringKeyValue, ts int64, value float64) *otlpmetricpb.DoubleGauge {
+	double := &otlpmetricpb.DoubleDataPoint{
+		Labels:       labels,
+		TimeUnixNano: uint64(ts),
+		Value:        value,
+	}
+	return &otlpmetricpb.DoubleGauge{
+		DataPoints: []*otlpmetricpb.DoubleDataPoint{double},
+	}
+}
```
srebhan, line 247: It might be worth to move those to a separate file to reduce the clutter.

plugins/outputs/otlp/otlp.go
```diff
@@ -0,0 +1,379 @@
...
+						point.Data = &otlpmetricpb.Metric_IntSum{
+							IntSum: monotonicIntegerPoint(labels, ts, currentTs, 0),
+						}
+					}
+				case string:
+					o.Log.Error("get type failed: unsupported telegraf value type string")
+					continue
+				default:
+					o.Log.Errorf("get type failed: unsupported telegraf value type %v\n", f.Value)
+					continue
```
srebhan, line 297: I think you can skip the `string` case as it is handled by `default`.

plugins/outputs/otlp/otlp.go
```diff
@@ -0,0 +1,379 @@
...
+				case uint64:
+					if v <= uint64(MaxInt) {
+						point.Data = &otlpmetricpb.Metric_IntGauge{
+							IntGauge: intGauge(labels, ts, int64(v)),
+						}
+					} else {
+						point.Data = &otlpmetricpb.Metric_IntGauge{
+							IntGauge: intGauge(labels, ts, int64(MaxInt)),
+						}
+					}
```
srebhan, line 310: Why do you convert the `uint64` value to int? Anyway this might make more clear that you clip the value:
```suggestion
					x := int64(v)
					if v > uint64(MaxInt) {
						x := MaxInt
					}
					point.Data = &otlpmetricpb.Metric_IntGauge{
						IntGauge: intGauge(labels, ts, x),
					}
```

plugins/outputs/otlp/otlp.go
```diff
@@ -0,0 +1,379 @@
...
+							IntGauge: intGauge(labels, ts, 1),
+						}
+					} else {
+						point.Data = &otlpmetricpb.Metric_IntGauge{
+							IntGauge: intGauge(labels, ts, 0),
+						}
+					}
+				case string:
+					o.Log.Error("get type failed: unsupported telegraf value type string")
+					continue
```
srebhan, line 331: Contained in the `default` case.

plugins/outputs/otlp/otlp.go
```diff
@@ -0,0 +1,379 @@
...
+				default:
+					o.Log.Errorf("get type failed: unsupported telegraf value type %v\n", f.Value)
+					continue
+				}
+			// TODO: add support for histogram & summary
+			case telegraf.Histogram, telegraf.Summary:
+				fallthrough
```
srebhan, line 338: Please either implement it or leave it out!

plugins/outputs/otlp/otlp.go
```diff
@@ -0,0 +1,379 @@
...
+			}
+			samples = append(samples, sample)
+		}
+	}
+
+	if err := o.client.Store(&metricsService.ExportMetricsServiceRequest{
+		ResourceMetrics: samples,
+	}); err != nil {
+		o.Log.Errorf("unable to write to endpoint: %s", err)
+		return err
```
srebhan, line 351:
```suggestion
		return fmt.Errorf("unable to write to endpoint: %s", err)
```

plugins/outputs/otlp/otlp.go
```diff
@@ -0,0 +1,379 @@
...
+
+// Description returns the human-readable function definition of the plugin.
+func (o *OTLP) Description() string {
+	return "Configuration for OTLP to send metrics to"
+}
+
+func newOTLP() *OTLP {
+	return &OTLP{}
+}
+
```
srebhan, line 374: Please remove this function and initialize the structure directly.

plugins/outputs/otlp/otlp_test.go
```diff
@@ -0,0 +1,259 @@
...
+	mockMetricsServer = metricServiceServer{
+		status: nil,
+	}
+	metricsService.RegisterMetricsServiceServer(grpcServer, &mockMetricsServer)
+	go func() {
+		_ = grpcServer.Serve(listener)
+	}()
+	defer grpcServer.Stop()
+	os.Exit(m.Run())
+}
```
srebhan, line 68: Can you please create a listener in each test separately to prevent side-effects between the tests. You might also want to take a look at https://pkg.go.dev/golang.org/x/net/nettest for creating the listener...

plugins/outputs/otlp/otlp_test.go
```diff
@@ -0,0 +1,259 @@
...
+func TestConfigOptions(t *testing.T) {
+	o := OTLP{
+		Endpoint: ":::::",
+		Log:      testutil.Logger{},
+	}
+	err := o.Connect()
+	require.EqualError(t, err, "invalid endpoint configured")
```
srebhan, line 75:
```suggestion
	err := o.Connect()
	require.Error(t, err)
	require.EqualError(t, err, "invalid endpoint configured")
```

plugins/outputs/otlp/otlp_test.go
```diff
@@ -0,0 +1,259 @@
...
+	require.EqualError(t, err, "invalid timeout configured")
+
+	o = OTLP{
+		Endpoint: "http://" + listener.Addr().String(),
+		Log:      testutil.Logger{},
+	}
+	err = o.Connect()
+	if err != nil {
+		t.Fatal(err)
+	}
```
srebhan, line 91:
```suggestion
	err = o.Connect()
	require.NoError(t, err)
```

srebhan (changes requested):
Thanks for submitting this plugin! The overall code quality is quite nice.

Nevertheless, I have several comments in the code, the most pressing one is the naming of the plugin. Can we please name it `opentelemetry` to make it discoverable!? The other major issue is the logging in the client code.

plugins/outputs/otlp/otlp.go
codeboten: Whoops! I left the comment in there after cleaning things up 😄, fixed.

plugins/outputs/otlp/otlp.go
codeboten: They do not, I've removed the unused variables and those left are no longer exported

README.md
codeboten: Sure, that makes sense to me.

plugins/outputs/otlp/otlp.go
codeboten: Done.

plugins/outputs/otlp/otlp.go
codeboten: Renamed.

plugins/outputs/otlp/otlp.go
codeboten: Removed.

plugins/outputs/otlp/client.go
codeboten: Fixed.

plugins/outputs/otlp/client.go
codeboten: Done.

plugins/outputs/otlp/otlp.go
codeboten: Removed

plugins/outputs/otlp/otlp.go
codeboten: Done

plugins/outputs/otlp/client.go
codeboten: I've updated the link, the configuration was moved to an internal package: https://github.com/open-telemetry/opentelemetry-go/blob/a2cecb6e80f6a0712187b080a97f8efb5a61082a/exporters/otlp/internal/otlpconfig/options.go#L47

plugins/outputs/otlp/client.go
codeboten: cleaned this up, i agree this is much simpler.

plugins/outputs/otlp/client.go
codeboten: Fixed

plugins/outputs/otlp/client.go
codeboten: After the refactor this was only used in one place, makes sense to remove the variable.

plugins/outputs/otlp/client.go
codeboten: Done.

plugins/outputs/otlp/otlp.go
codeboten: I moved most of this to `Init`, thanks for the tip, i had no idea `Init` even existed :)

plugins/outputs/otlp/client.go
srebhan: Forgive me if this is a dumb comment, but I'm not deep into open-telemetry... Reading the new referenced version I think you can use [`NewDriver()`](https://pkg.go.dev/go.opentelemetry.io/otel/exporters/otlp@v0.19.0/otlpgrpc#NewDriver) and then call `ExportMetrics()` and `ExportTraces()` on [the resulting driver](https://pkg.go.dev/go.opentelemetry.io/otel/exporters/otlp@v0.19.0#ProtocolDriver)... Am I missing something?

plugins/outputs/otlp/README.md
codeboten: Fixed.

plugins/outputs/otlp/client.go
codeboten: Nice, this makes the code much cleaner!

plugins/outputs/otlp/otlp.go
codeboten: Removed

plugins/outputs/otlp/client.go
codeboten: Fixed

plugins/outputs/otlp/client.go
codeboten: Done.

plugins/outputs/otlp/client.go
codeboten: Agreed, removed.

plugins/outputs/otlp/client.go
codeboten: I've updated the code to use `c.conn` everywhere and only return the error

plugins/outputs/otlp/client_test.go
codeboten: No longer ignoring it.

plugins/outputs/otlp/client.go
codeboten: Removed.

plugins/outputs/otlp/client.go
codeboten: Removed.

plugins/outputs/otlp/otlp.go
codeboten: Thanks for the suggestion, this is much cleaner now!

plugins/outputs/otlp/otlp.go
codeboten: Done

plugins/outputs/otlp/otlp.go
codeboten: This was code leftover from the initial copy of the stackdriver output plugin. Ordering doesn't matter for sending data over OTLP.

Is it worth opening an issue to track the change to a slice of pointers? https://github.com/influxdata/telegraf/blob/ad942052f06bb13ce8dad899cba9066e9ad307ed/plugins/outputs/stackdriver/stackdriver.go#L106-L115

plugins/outputs/otlp/client.go
codeboten: The metrics API and SDK are still being defined so i'd rather not rely on them just yet for this code. Once it's solidified, i'll likely refactor this plugin to use those instead.

plugins/outputs/otlp/otlp_test.go
codeboten: Done. Thanks for pointing me to `nettest`!

plugins/outputs/otlp/client.go
codeboten: Done

plugins/outputs/otlp/otlp.go
codeboten: The conversion happens because OTLP only supports `int64`, I've updated the code to something similar to your suggestion.

plugins/outputs/otlp/otlp.go
codeboten: Done, moved this to `transform.go`

plugins/outputs/otlp/otlp.go
codeboten: Done

plugins/outputs/otlp/otlp.go
codeboten: Fixed.

plugins/outputs/otlp/client.go
codeboten: In most circumstances parallelizing RPC requests will be more performant, using grpc to handle the load balancing. Is there a reason why it wouldn’t be in this case here?

codeboten:
@srebhan I've addressed your suggestions, please take a look! thanks for a great review!

go.mod
```diff
@@ -3,9 +3,9 @@ module github.com/influxdata/telegraf
 go 1.16
 
 require (
-	cloud.google.com/go v0.56.0
-	cloud.google.com/go/bigquery v1.4.0
-	cloud.google.com/go/pubsub v1.2.0
+	cloud.google.com/go v0.72.0
+	cloud.google.com/go/bigquery v1.8.0
+	cloud.google.com/go/pubsub v1.3.1
```
srebhan, line 8: Are all of those package upgrades necessary or did they make it in by accident? Just wondering as it seems a massive bunch... Same for all updates below.

plugins/outputs/opentelemetry/README.md
```diff
@@ -0,0 +1,44 @@
+# OpenTelemetry Output Plugin
+
+This plugin writes to any backend that support the OpenTelemetry Protocol (OTLP)
+
+Metrics are named by combining the metric name and field key - eg: `cpu.usage_user`
+
+Additional resource attributes can be configured by `attributes`.
+
+Additional gRPC metadata can be configured by `headers`.
```
srebhan, line 9: You might want to remove the newlines. It's just cosmetics, but I might be nicer to read. Maybe ask @sjwang90, she's good at writing documentation! :-)

plugins/outputs/opentelemetry/README.md
```diff
@@ -0,0 +1,44 @@
...
+  ## Use TLS but skip chain & host verification
+  insecure_skip_verify = false
+
+  # Additional resource attributes
+  [outputs.opentelemetry.attributes]
+      "service.name" = "demo"
+
+  # Additional grpc metadata
+  [outputs.opentelemetry.headers]
+      key1 = "value1"
```
srebhan, line 37: Can you maybe sync this with the content of the `sampleConfig` in the code!?

plugins/outputs/otlp/client.go
srebhan: Hmm, it's versioned as 0.19, so I think you can rely on this. I think the interface of 0.19 referenced to above
a) simplifies the code massively
b) does not need copying internal code from the client that might be subject to change
c) should be guaranteed to work by the maintainers of the package

So to be honest I would feel better to rely on their code and maybe later change if 0.20, 1.0 or whatever comes out...

plugins/outputs/opentelemetry/client.go
```diff
@@ -0,0 +1,225 @@
...
+				"OUT_OF_RANGE",
+				"UNAVAILABLE",
+				"DATA_LOSS"
+			]
+		}
+	}]
+}`
+)
+
+var maxTimeseriesPerRequest = 500
```
srebhan, line 52: This can be a `const`

plugins/outputs/opentelemetry/client.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+	switch s.Code() {
+	case codes.DeadlineExceeded, codes.Canceled, codes.ResourceExhausted,
+		codes.Aborted, codes.OutOfRange, codes.Unavailable, codes.DataLoss:
+		// See https://github.com/open-telemetry/opentelemetry-specification/
+		// blob/master/specification/protocol/otlp.md#response
+		return true
+	default:
+		return false
+	}
```
srebhan, line 224: You might change this to 
```suggestion
	}
	return false
```
but it's a matter of taste...

plugins/outputs/opentelemetry/client_test.go
```diff
@@ -0,0 +1,87 @@
...
+	"google.golang.org/grpc/status"
+)
+
+func TestClientWithRecoverableError(t *testing.T) {
+	listener, err := nettest.NewLocalListener("tcp")
+	require.NoError(t, err)
+	grpcServer := grpc.NewServer()
+	mockMetricsServer := metricServiceServer{
+		status: nil,
+	}
```
srebhan, line 25: The member variables are `nil` by default, so you can simply do
```suggestion
	mockMetricsServer := metricServiceServer{}
```

plugins/outputs/opentelemetry/client_test.go
```diff
@@ -0,0 +1,87 @@
...
+		logger:  testutil.Logger{},
+		url:     u,
+		timeout: time.Second,
+	}
+
+	err = client.ping(context.Background())
+	require.False(t, isRecoverable(err), "expected unrecoverableError in error %v", err)
+
+	err = client.connect(context.Background())
+	require.False(t, isRecoverable(err), "expected unrecoverableError in error %v", err)
```
srebhan, line 72: How are you pinging before connect? Is this really the intended order? If so it might be worth a small comment.

plugins/outputs/opentelemetry/opentelemetry.go
```diff
@@ -0,0 +1,238 @@
...
+		instrumentationLibraryName,
+		internal.Version(),
+	)
+
+	tlsConfig, err := o.TLSConfig()
+	if err != nil {
+		return errors.Wrap(err, "invalid tls configuration")
+	}
+
+	if o.client == nil {
```
srebhan, line 107: This is always the case as `Init()` is called first and only once, so you can skip the `if`.

srebhan (commented):
This looks quite good. I have some small comments in the code, mostly cosmetic. The biggest issue I still see is the copying of the internal config. I would prefer to rely on the 0.19 version of the client rather than re-coding it.
But if the other reviewers are ok with it I don't want to be in the way as the client code is still nice. :-)

plugins/outputs/opentelemetry/client.go
codeboten: Fixed.

plugins/outputs/opentelemetry/README.md
codeboten: Done.

plugins/outputs/opentelemetry/opentelemetry.go
codeboten: Done.

plugins/outputs/opentelemetry/README.md
codeboten: Updated the documentation.

go.mod
codeboten: I reverted go.mod and started fresh, looks much better now!

plugins/outputs/opentelemetry/client_test.go
codeboten: ping calls connect initially, the test is to ensure it reports the connect error. I've updated the second test to call store instead, which also call connect.

codeboten:
Addressed the new comments, it looks like the failing test is caused by a race condition happening in a different plugin.

plugins/outputs/opentelemetry/client.go
```diff
@@ -0,0 +1,224 @@
...
+	c.conn = conn
+	return err
+}
+
+// ping sends an empty request the endpoint.
+func (c *client) ping(ctx context.Context) error {
+	// Loop until the context is canceled, allowing for retryable failures.
+	for {
+		err := c.connect(ctx)
```
srebhan, line 109: To be honest, this is somewhat unexpected from my perspective. The task of this function is to test connectivity, so I'm not sure if it's a good idea to connect. However, if everybody else accepts this as is, I can live with it.

plugins/outputs/opentelemetry/client_test.go
srebhan: Well tbh it's somewhat unexpected that `ping` also connects. But maybe it's only my view. :-)

srebhan (commented):
Very nice. Can you please say a few words regarding the possibility to use the v0.19 interface of the upstream client and thus get rid of most of your client implementation!?
Another comment is about connecting in `ping`. To me it's unexpected, but this is not a blocker for me.

codeboten:
Hey @srebhan 

Thanks for all the reviews! With regards to using the OTLP exporter instead of the current client code, in order to utilize it, it would require creating various implementing various components specified in the SDK (CheckpointSet, Aggregator) which are likely to change in the next iteration of the opentelemetry metrics signal specification. There's no exposed interface that allows me to use the protobuf more directly and pass in the slice of `ResourceMetrics`. 

If you're ok w/ the current implementation, I would be happy to refactor the client code in the future, once the metrics API/SDK has stabilized. I just don't want to implement something that's not trivial (I spent some hours trying today, and i'm not sure the code is any simpler) and have to throw it away in the near future.

I've updated the code to remove the separate ping function, and instead incorporated the check into the connect function.

codeboten:
Looks like there's another PR implementing this as well https://github.com/influxdata/telegraf/pull/9228

codeboten:
@srebhan not sure how to move this PR forward here, any help is much appreciated.

sjwang90:
@srebhan and some other Telegraf engineers will discuss the two OpenTelemetry output plugin PRs next week. We should have an answer for you @codeboten by Weds - I believe there's nothing we need from your right now.

sjwang90:
@codeboten The team would like to proceed with #9228 and that PR is close to being merged. If you would like to take a review at the PR and address any concerns.

Otherwise, thanks so much for your contribution. Either way we should be getting this OpenTelemetry Output plugin into master very soon!

codeboten:
Thanks @sjwang90, I tested the other PR and couldn't see any data being output to the OpenTelemetry collector. See my comment here:
https://github.com/influxdata/telegraf/pull/9228#issuecomment-840101808

Running the same test I did with the other PR with the code in this PR I see the following output:

```
otel-collector              | Metric #0
otel-collector              | Descriptor:
otel-collector              |      -> Name: processes.unknown
otel-collector              |      -> Description:
otel-collector              |      -> Unit:
otel-collector              |      -> DataType: IntGauge
otel-collector              | IntDataPoints #0
otel-collector              | Data point labels:
otel-collector              |      -> host: Alexs-MacBook-Pro.local
otel-collector              | StartTime: 0
otel-collector              | Timestamp: 1620854350000000000
otel-collector              | Value: 0
otel-collector              | ResourceMetrics #203
otel-collector              | InstrumentationLibraryMetrics #0
otel-collector              | InstrumentationLibrary telegraf unknown
otel-collector              | Metric #0
otel-collector              | Descriptor:
otel-collector              |      -> Name: processes.idle
otel-collector              |      -> Description:
otel-collector              |      -> Unit:
otel-collector              |      -> DataType: IntGauge
otel-collector              | IntDataPoints #0
otel-collector              | Data point labels:
otel-collector              |      -> host: Alexs-MacBook-Pro.local
otel-collector              | StartTime: 0
otel-collector              | Timestamp: 1620854350000000000
otel-collector              | Value: 0
otel-collector              | ResourceMetrics #204
otel-collector              | InstrumentationLibraryMetrics #0
otel-collector              | InstrumentationLibrary telegraf unknown
otel-collector              | Metric #0
otel-collector              | Descriptor:
otel-collector              |      -> Name: processes.blocked
otel-collector              |      -> Description:
otel-collector              |      -> Unit:
otel-collector              |      -> DataType: IntGauge
otel-collector              | IntDataPoints #0
otel-collector              | Data point labels:
otel-collector              |      -> host: Alexs-MacBook-Pro.local
otel-collector              | StartTime: 0
otel-collector              | Timestamp: 1620854350000000000
otel-collector              | Value: 0
```

sjwang90:
@codeboten Thanks for your feedback. @jacobmarble is going to take a look at your comments. Let's provide any updates on #9228 to keep it all the discussion in the same place.

sjwang90:
Closing since #9228

PR #9116, opened 2021-04-12
Remove error return type from metric.New method

Jagularr:
the `metric.New `method never returns an error, which led to a lot of unnecessary checks in the code. This PR removes the error return type from it and refactors the 94 files that referenced it.

### DIFF
--- a/plugins/parsers/dropwizard/parser.go
+++ b/plugins/parsers/dropwizard/parser.go
@@ -227,11 +227,7 @@ func (p *parser) readDWMetrics(metricType string, dwms interface{}, metrics []te
 			parsed, err := p.seriesParser.Parse([]byte(measurementName))
 			var m telegraf.Metric
 			if err != nil || len(parsed) != 1 {
-				m, err = metric.New(measurementName, map[string]string{}, map[string]interface{}{}, tm)
-				if err != nil {
-					log.Printf("W! failed to create metric of type '%s': %s\n", metricType, err)
-					continue
-				}
+				m = metric.New(measurementName, map[string]string{}, map[string]interface{}{}, tm)
 			} else {
 				m = parsed[0]
 				m.SetTime(tm)
--- a/plugins/parsers/dropwizard/parser_test.go
+++ b/plugins/parsers/dropwizard/parser_test.go
@@ -497,10 +497,7 @@ func containsAll(t1 map[string]string, t2 map[string]string) bool {
 	return true
 }
 
-func Metric(v telegraf.Metric, err error) telegraf.Metric {
-	if err != nil {
-		panic(err)
-	}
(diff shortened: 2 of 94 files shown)


plugins/parsers/dropwizard/parser.go
```diff
@@ -227,11 +227,7 @@ func (p *parser) readDWMetrics(metricType string, dwms interface{}, metrics []te
 			parsed, err := p.seriesParser.Parse([]byte(measurementName))
```
helenosheaa, line 227: I'm wondering if this error needs to be returned instead?

Jagularr: It looks like it's handled in the if statement two lines below, I'm unsure if that's what you meant though

helenosheaa: Yeah mainly just thought why are we nil checking it but then not either returning the error or logging it. I'm aware this isn't related to what you've changed so probably fine to leave it as is.

Jagularr: I agree, although as it's out of scope for this PR I'll probably leave it. and thanks for reviewing!

helenosheaa (approved):

plugins/parsers/dropwizard/parser.go
ssoroka: yeah this logic is definitely bad. there's a panic hiding in here too. You don't have to fix it in this PR, but please circle back to it and/or open an issue

plugins/parsers/dropwizard/parser_test.go
```diff
@@ -497,10 +497,7 @@ func containsAll(t1 map[string]string, t2 map[string]string) bool {
 	return true
 }
 
-func Metric(v telegraf.Metric, err error) telegraf.Metric {
-	if err != nil {
-		panic(err)
-	}
+func Metric(v telegraf.Metric) telegraf.Metric {
 	return v
 }
```
ssoroka, line 502: this function probably is not necessary any more and can be removed.

plugins/parsers/influx/parser_test.go
```diff
@@ -15,10 +15,7 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func Metric(v telegraf.Metric, err error) telegraf.Metric {
-	if err != nil {
-		panic(err)
-	}
+func Metric(v telegraf.Metric) telegraf.Metric {
 	return v
 }
```
ssoroka, line 21: can be removed

ssoroka (approved):

plugins/parsers/dropwizard/parser_test.go
Jagularr: yup!

Jagularr:
!retry-checks

PR #9123, opened 2021-04-13
Config Option to Enable Logging with Local Time

### LINKED ISSUE #3462: Allow logging in local timezone
For some reason the timestamp in the  "/var/log/telegraf.log" is UTC instead of my local timezone

2017-11-11T09:32:04Z E! Error writing to output [influxdb]: Could not write to any InfluxDB server in cluster
2017-11-11T09:33:00Z E! InfluxDB Output Error: Post http://192.168.1.200:8086/write?consistency=any&db=telegraf: dial tcp 192.168.1.200:8086: getsockopt: connection ref
used

### Relevant telegraf.conf:
 
  debug = false
  quiet = false
  logfile = "/var/log/telegraf.log"

My /etc/localtime points to the right timezone file ... 

/etc/localtime -> /usr/share/zoneinfo/Europe/Stockholm

...so everytihng looks correct but something must be missing?

### System info:

uname -a
Linux kl_backside_2 4.4.50 #1 Sun Mar 26 05:10:49 EEST 2017 armv6l GNU/Linux

telegraf -version
Telegraf v1.4.3 (git: release-1.4 b473b6a6596078f142a623ca38ff7030b7acec8f)

date command output
Sat Nov 11 12:52:25 CET 2017

Jagularr:
This hopefully resolves #3462.

You'd be able to enable `log_with_local_time` so that all logs will use the time of the machine it's being ran on. This will apply for all types of logs (stdout, file).

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -359,14 +359,14 @@ var agentConfig = `
   ## If set to -1, no archives are removed.
   # logfile_rotation_max_archives = 5
 
-  ## Inidicates whether or not to use local time as the logging prefix.
-  # log_with_local_time = false
+  ## Pick a timezone to use when logging or type 'local' for local time.
+  ## Example: America/Chicago
+  # log_with_timezone = ""
 
   ## Override default hostname, if empty use os.Hostname()
   hostname = ""
   ## If set to true, do no set the "host" tag in the telegraf agent.
   omit_hostname = false
-
 `
 
 var outputHeader = `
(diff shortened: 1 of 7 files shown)


config/config.go
```diff
@@ -188,6 +188,9 @@ type AgentConfig struct {
 	// If set to -1, no archives are removed.
 	LogfileRotationMaxArchives int `toml:"logfile_rotation_max_archives"`
 
+	// Inidicates whether or not to use local time as the logging prefix.
+	LogWithLocalTime bool `toml:"log_with_local_time"`
```
ssoroka, line 192: something like `log_with_timezone` where timezone is a string and can be `local` or any other timezone would be quite useful and a bit more flexible.

Jagularr: Good idea, done!

sspaink (approved):

PR #9124, opened 2021-04-13
Add custom docker image info to nvidia plugin

### LINKED ISSUE #8531: Add a new Telegraf:cuda Docker image to run the Nvidia-smi input plugin
### Proposal:

Running the Nvidia-smi plugin from the Docker Telegraf image.
I propose to add a new build to the list of the officially available Telegraf images based on nvidia/cuda, by specifiying clearly in the README that the Telegraf:cuda Docker image is required to run the Nvidia-smi plugin.

### Current behavior:

Nvidia-smi cannot be executed inside the Docker container as it does not include cuda.

### Desired behavior:

Make it possible to run the Nvidia-smi plugin in Docker.

helenosheaa:
resolves #8531
Add context on how to use nvidia plugin with a custom nvidia/cuda docker image

### DIFF
--- a/plugins/inputs/nvidia_smi/README.md
+++ b/plugins/inputs/nvidia_smi/README.md
@@ -90,4 +90,4 @@ nvidia_smi,compute_mode=Default,host=8218cf,index=2,name=GeForce\ GTX\ 1080,psta
 Note that there seems to be an issue with getting current memory clock values when the memory is overclocked.
 This may or may not apply to everyone but it's confirmed to be an issue on an EVGA 2080 Ti.
 
-**NOTE:** For use with docker generate your own custom docker image based on nvidia/cuda which also installs a telegraf package.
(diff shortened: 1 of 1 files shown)


ssoroka:
I'm not sure that this is the only method, it's probably possible to use volume mount binding to inject the required binary without rebuilding the docker container.

helenosheaa:
@ssoroka Ok I've added about mount binding as an alternative option in there also with a link to the relevant docker page.

Jagularr (approved):

ssoroka (approved):
this is probably correct. bwahaha

BlueskyFR:
May I make my own PR instead of this one with more detailed information?
I would like to add a Dockerfile example instead of a simple note at the end.

helenosheaa:
@BlueskyFR if you have a docker example you can add to the plugin folder we'd be happy to review a PR for it. I think it's also worth keeping this change to the README.

PR #9130, opened 2021-04-14
SQL Server - sqlServerRingBufferCPU - removed whitespaces

Trovalo:
### Required for all PRs:

Removed a pair of whitespace chars from the **sqlServerRingBufferCPU** SQL statement and added some formatting.
This query exists only for the on-prem version of SQL Server (`database_type = "SQLServer"`)

If you were unlucky enough to have some SQL 2008 the query wouldn't work as the whitespace char is not allowed inside the statement.


Trovalo:
!signed-cla

plugins/inputs/sqlserver/sqlserverqueries.go
```diff
@@ -1170,25 +1170,24 @@ WITH utilization_cte AS
 ),
 processor_Info_cte AS
 (
-	SELECT (cpu_count / hyperthread_ratio) as number_of_physical_cpus
-	  FROM sys.dm_os_sys_info
+	SELECT ([cpu_count] / [hyperthread_ratio]) as [number_of_physical_cpus]
```
Jagularr, line 1173: Out of curiosity, what's the purpose for the square brackets?

Jagularr (approved):
Seems fair enough

plugins/inputs/sqlserver/sqlserverqueries.go
Trovalo: The square bracket does not change anything is just a way to encapsulate the column name, in SQL Server you can use `"col"` or `[col]`. it's mandatory and automatically used when the column name contains special chars or spaces, otherwise, it's not strictly needed.

I like to always use it, but it was also used in the whole query, except for that line

Trovalo:
@sspaink can you have a look at this and eventually merge it?

sspaink (approved):

PR #9133, opened 2021-04-14
Add time.now() starlark processor example test

Jagularr:
This adds the ability to mock the time.now() function in starlark and thus be able to create an example in which time.now() is used to set on metrics.

### DIFF
--- a/plugins/processors/starlark/starlark.go
+++ b/plugins/processors/starlark/starlark.go
@@ -46,10 +46,11 @@ type Starlark struct {
 
 	Log telegraf.Logger `toml:"-"`
 
-	thread    *starlark.Thread
-	applyFunc *starlark.Function
-	args      starlark.Tuple
-	results   []telegraf.Metric
+	thread           *starlark.Thread
+	applyFunc        *starlark.Function
+	args             starlark.Tuple
+	results          []telegraf.Metric
+	starlarkLoadFunc func(module string, logger telegraf.Logger) (starlark.StringDict, error)
 }
 
 func (s *Starlark) Init() error {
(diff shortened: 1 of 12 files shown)


plugins/processors/starlark/starlark.go
```diff
@@ -63,7 +64,7 @@ func (s *Starlark) Init() error {
 	s.thread = &starlark.Thread{
 		Print: func(_ *starlark.Thread, msg string) { s.Log.Debug(msg) },
 		Load: func(thread *starlark.Thread, module string) (starlark.StringDict, error) {
-			return loadFunc(module, s.Log)
+			return s.starlarkLoadFunc(module, s.Log)
```
srebhan, line 67: What is the reason behind this?

plugins/processors/starlark/starlark_test.go
```diff
@@ -3256,3 +3279,22 @@ func parseErrorMessage(t *testing.T, lines []string, header string) string {
...
+
+	if module == "time.star" {
+		customModule := result["time"].(*starlarkstruct.Module)
+		customModule.Members["now"] = starlark.NewBuiltin("now", testNow)
+		result["time"] = customModule
+	}
+
+	return result, nil
+}
+
```
srebhan, line 3297: IIRC you can change the `NowFunc` of the time module at runtime after initialization. Have you tried this?

srebhan (commented):
The code looks good, but I'd like to prevent messing with the load function if possible. Can you please check if you can set time`s `NowFunc` in the tests instead of overriding the load!?

ssoroka (approved):
looks good. I think there's issues with replacing the load function because the module is loaded at runtime. Looks fine the way it is.

plugins/processors/starlark/starlark_test.go
Jagularr: That possibly could work (would have to ensure its set before loadFunc is called) but IMO overriding the method for tests allows other tests to re-use the same 'test' load where we can override functions that otherwise couldn't be tested, like time.now.

srebhan: Well it means that you add new modules in two places now, but it seems you made your choice already anyway...

PR #9138, opened 2021-04-16
Add OAuth2 to HTTP input

### LINKED ISSUE #5309: Add OAUTH2 Support for HTTP Input Plugin
## Feature Request
Add OAUTH2 Support for HTTP Input Plugin

### Proposal:

When using OAUTH2, you need to get token first via a HTTP POST Request , then you will able to send a succeeding GET Request using the previously retrieved token.

Ref.: https://aaronparecki.com/oauth-2-simplified/

Example: 

1. POST https://api.authorization-server.com/token
grant_type=client_credentials&client_id=CLIENT_ID&client_secret=CLIENT_SECRET

2. GET curl -H "Authorization: Bearer RsT5OjbzRn430zqMLgV3Ia" 
https://api.authorization-server.com/1/me

 
### Current behavior:
n/a

### Desired behavior:
OAUTH2 support

modern authentication mechanisms like OAUTH2 are becoming more and more common

Jagularr:
This should add support for oauth2 authentication ability for the http input client. Be advised that this is mostly directly taken from the HTTP output plugin. Should resolve #5309

### DIFF
--- a/plugins/common/http/config.go
+++ b/plugins/common/http/config.go
@@ -2,10 +2,10 @@ package httpconfig
 
 import (
 	"context"
-	"log"
 	"net/http"
 	"time"
 
+	"github.com/influxdata/telegraf"
 	"github.com/influxdata/telegraf/config"
 	"github.com/influxdata/telegraf/plugins/common/proxy"
 	"github.com/influxdata/telegraf/plugins/common/tls"
@@ -20,6 +20,7 @@ type HTTPClientConfig struct {
 	ClientSecret string   `toml:"client_secret"`
 	TokenURL     string   `toml:"token_url"`
 	Scopes       []string `toml:"scopes"`
+	Log          telegraf.Logger
 
 	Timeout config.Duration `toml:"timeout"`
 
@@ -63,7 +64,7 @@ func (h *HTTPClientConfig) CreateClient(ctx context.Context) (*http.Client, erro
 		ctx = context.WithValue(ctx, oauth2.HTTPClient, client)
 		client = oauthConfig.Client(ctx)
 	} else if h.ClientID != "" || h.ClientSecret != "" || h.TokenURL != "" {
-		log.Printf("One of the following fields is empty: Client ID, Client Secret or Token URL. Skipping OAuth.")
+		h.Log.Warnf("One of the following fields is empty: Client ID, Client Secret or Token URL. Skipping OAuth.")
 	}
 
 	return client, nil
--- a/plugins/common/http/config.go
+++ b/plugins/common/http/config.go
@@ -2,6 +2,7 @@ package httpconfig
 
 import (
 	"context"
+	"log"
 	"net/http"
 	"time"
 
@@ -61,6 +62,8 @@ func (h *HTTPClientConfig) CreateClient(ctx context.Context) (*http.Client, erro
 		}
 		ctx = context.WithValue(ctx, oauth2.HTTPClient, client)
 		client = oauthConfig.Client(ctx)
+	} else if h.ClientID != "" || h.ClientSecret != "" || h.TokenURL != "" {
+		log.Printf("One of the following fields is empty: Client ID, Client Secret or Token URL. Skipping OAuth.")
 	}
 
 	return client, nil
--- a/plugins/common/http/config.go
+++ b/plugins/common/http/config.go
@@ -13,7 +13,7 @@ import (
 )
 
 // Common HTTP client struct.
-type HttpClientConfig struct {
+type HTTPClientConfig struct {
 	// OAuth2 Credentials
 	ClientID     string   `toml:"client_id"`
 	ClientSecret string   `toml:"client_secret"`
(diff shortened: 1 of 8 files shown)


plugins/inputs/http/http.go
```diff
@@ -108,31 +123,54 @@ func (*HTTP) Description() string {
...
-		h.SuccessStatusCodes = []int{200}
+	if h.ClientID != "" && h.ClientSecret != "" && h.TokenURL != "" {
+		oauthConfig := clientcredentials.Config{
+			ClientID:     h.ClientID,
+			ClientSecret: h.ClientSecret,
+			TokenURL:     h.TokenURL,
+			Scopes:       h.Scopes,
+		}
+		ctx = context.WithValue(ctx, oauth2.HTTPClient, client)
+		client = oauthConfig.Client(ctx)
```
srebhan, line 170: I might make sense to log (maybe as debug) if at least one of those fields is empty but at least one of the other fields isn't, i.e. if the user forgot to specify one of the fields.

srebhan (commented):
Looks good. I have one comment regarding logging if the user _forgot_ to specify one field but nothing big.

However, I wonder if it makes sense to move this to `plugins/common` similar to `proxy` as I expect other plugins using the `http` transport will also add this in the future... What do you think about making this a common infrastructure?

plugins/inputs/http/http.go
Jagularr: Agreed it would help, I'm not sure if it would be worth adding more complexity to the code though?

Jagularr:
> Looks good. I have one comment regarding logging if the user _forgot_ to specify one field but nothing big.
> 
> However, I wonder if it makes sense to move this to `plugins/common` similar to `proxy` as I expect other plugins using the `http` transport will also add this in the future... What do you think about making this a common infrastructure?

Makes a lot of sense, just did a commit to do that.

plugins/inputs/http/http.go
srebhan: From a user's perspective you might wonder why the connection fails if you forgot one of those and you will have no means to find out why. It might be just me, but spending extra lines to help the user finding problems in his/her config is worth the two lines
```golang
else if  h.ClientID != "" || h.ClientSecret != "" || h.TokenURL != "" {
    h.Log.Warnf("One of the fields client ID=%q, client secret=%q or token URL=%q is empty. Skipping OAuth.", h.ClientID, h.ClientSecret, h.TokenURL)
}
```

srebhan (commented):
I thought more about doing it _exactly_ the same way as `proxy` or `TLS` but I'm also fine with this approach. This only leaves the user config warning issue... I would prefer it but if you have a strong opinion I don't want to be in your way. :-)

plugins/inputs/http/http.go
Jagularr: Sure, I'll add that in. thanks!

reimda (changes requested):
Could you factor out the common code from plugins/http/http.go and make it use plugins/common/http/*.go?  It would be nice to have all the oauth2 stuff in one place.

Looks great otherwise.  Thanks for adding tests!

Jagularr:
!retry-checks

plugins/common/http/config.go
```diff
@@ -0,0 +1,73 @@
...
+	"golang.org/x/oauth2"
+	"golang.org/x/oauth2/clientcredentials"
+)
+
+// Common HTTP client struct.
+type HTTPClientConfig struct {
+	// OAuth2 Credentials
+	ClientID     string   `toml:"client_id"`
+	ClientSecret string   `toml:"client_secret"`
+	TokenURL     string   `toml:"token_url"`
```
srebhan, line 21: Can we please also seperate this into something like `plugins/common/oauth/oauth.go`. It might be also useful for other transports or in existing plugins.

srebhan (commented):
I still think it might be worth to have a standalone `oauth` package without being bound to `http`...

plugins/common/http/config.go
Jagularr: Yup.

Jagularr:
@srebhan yup. just added it as it's own package. Let me know if that's what you were expecting, or if it's not exactly right.

Jagularr:
!retry-checks

srebhan (commented):
Looks mostly good, my only concern is about the logger (see my comment).

plugins/common/oauth/config.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+type OAuth2Config struct {
+	// OAuth2 Credentials
+	ClientID     string   `toml:"client_id"`
+	ClientSecret string   `toml:"client_secret"`
+	TokenURL     string   `toml:"token_url"`
+	Scopes       []string `toml:"scopes"`
+
+	Log telegraf.Logger
```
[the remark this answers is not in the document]
Jagularr: Great catch, this would cause a nil pointer exception. Thanks. Removed.

srebhan (approved):
Looks good to me.

davihar:
@Jagularr I see that you included part of the config functionality that https://cs.opensource.google/go/x/oauth2/+/a41e5a78:clientcredentials/clientcredentials.go;l=44;drc=a41e5a7819143202e8d21fb9ac9c659fbd501fb3;bpv=1;bpt=1 exposes but did not include EndpointParams.  How can this be added so that I can do what this unit test shows for adding audience value:  https://cs.opensource.google/go/x/oauth2/+/a41e5a78:clientcredentials/clientcredentials_test.go;l=25;drc=a41e5a7819143202e8d21fb9ac9c659fbd501fb3;bpv=1;bpt=1

debu99:
Is this code merged? i didn't see it in master branch

PR #9140, opened 2021-04-16
Fix CI config to check if branch is master before skipping

sspaink:
Related to: https://github.com/influxdata/telegraf/pull/9076

In order to support skipping a job depending on file changes, only works on non-master branches. This pull requests updates the config to check the current branch.


Jagularr (approved):
makes sense!

PR #9141, opened 2021-04-16
Modbus rework

srebhan:
### Required for all PRs:

Restructuring and cleaning the modbus input plugin. This is preparation work to incorporate features from PR #7941 and #8013 as well as various feature-requests around modbus into the current plugin's code-base.
As a side effect the structure of the plugin as well as unit-tests are simplified and IMO easier to follow. The used modbus-library is migrated from the unmaintained goburrow version to the grid-x fork. Additionally, all existing linter-issues are nuked by this PR.


plugins/inputs/modbus/modbus_test.go
```diff
@@ -614,144 +625,244 @@ func TestHoldingRegisters(t *testing.T) {
...
+	for i := 0; i < 4000; i++ {
+		fc := fieldDefinition{}
 		fc.Name = fmt.Sprintf("coil-%v", i)
 		fc.Address = []uint16{uint16(i)}
 		fcs = append(fcs, fc)
 
-		t.Run(fc.Name, func(t *testing.T) {
-			_, err = client.WriteSingleCoil(fc.Address[0], writeValue)
-			assert.NoError(t, err)
-		})
```
ssoroka, line 684: removing the t.Run here changes the functionality, instead I think the tests will stop on the first failure instead of reporting all of them.

plugins/inputs/modbus/modbus.go
```diff
@@ -18,51 +17,46 @@ import (
...
+type fieldConverterFunc func(bytes []byte) interface{}
+
+type requestSet struct {
+	coil     []request
+	discrete []request
+	holding  []request
+	input    []request
+}
+
+type field struct {
```
ssoroka, line 50: is this still used?

plugins/inputs/modbus/modbus.go
```diff
@@ -173,552 +167,253 @@ func (m *Modbus) Init() error {
...
-	addrs := []uint16{}
-	for _, field := range fields {
-		addrs = append(addrs, field.Address...)
+	timestamp := time.Now()
+	for retry := 0; retry <= m.Retries; retry++ {
+		timestamp = time.Now()
+		if err := m.gatherFields(); err != nil {
+			if mberr, ok := err.(*mb.Error); ok && mberr.ExceptionCode == mb.ExceptionCodeServerDeviceBusy && retry < m.Retries {
+				m.Log.Infof("Device busy! Retrying %d more time(s)...", m.Retries-retry)
+				time.Sleep(time.Duration(m.RetriesWaitTime))
```
ssoroka, line 203: probably inherited code, but we might want to consider putting non-responsive devices into a list and processing that list over until it's empty, rather than sleeping mid-list.

plugins/inputs/modbus/modbus.go
```diff
@@ -173,552 +167,253 @@ func (m *Modbus) Init() error {
...
-}
 
-func disconnect(m *Modbus) error {
-	u, err := url.Parse(m.Controller)
-	if err != nil {
-		return err
-	}
+	m.handler.SetSlave(m.SlaveID)
+	m.client = mb.NewClient(m.handler)
+	m.isConnected = false
```
ssoroka, line 279: didn't the previous version try to connect here?

plugins/inputs/modbus/modbus.go
```diff
@@ -173,552 +167,253 @@ func (m *Modbus) Init() error {
...
-	case "FLOAT32", "UFIXED":
-		if len(bytes) == 2 {
-			e16 := convertEndianness16(t.ByteOrder, bytes)
-			return scale16UtoFloat(t.Scale, e16)
-		} else if len(bytes) == 4 {
-			e32 := convertEndianness32(t.ByteOrder, bytes)
-			return scale32UtoFloat(t.Scale, e32)
-		} else {
-			e64 := convertEndianness64(t.ByteOrder, bytes)
-			return scale64UtoFloat(t.Scale, e64)
```
ssoroka, line 549: looks like the behavior of this might have changed? Might not be covered by tests.

ssoroka (commented):
left some comments. These are some pretty massive changes. Would be good to get someone with a larger modbus installation to test out this build before we merge it.

plugins/inputs/modbus/modbus.go
srebhan: It's a but unfortunate, but yes, the `requestSet` has a list of `request` instances and the `request` (see _request.go_) has a list of `field`s associated with / filled by that request. I thought moving out the request would be good, but if you like it more to have it in here it's only the type-definition and one function.

plugins/inputs/modbus/modbus.go
srebhan: Well the plugin can handle only one device and if that device is busy we cannot do anything in the meantime... I thought about putting off the sleep but it adds complexity for no gain IMO.

plugins/inputs/modbus/modbus.go
srebhan: True, however modbus devices are usually quite restricted in the number of open connections, so I removed the connect here as we always connect in the gather cycle if not already connected. The idea was to connect as late as possible.

One feature that is requested (and will be committed on top of this PR) is to connect/disconnect in each cycle to allow other collection agents to access the device. This is especially useful for serial connections...

plugins/inputs/modbus/modbus.go
srebhan: Why do you think it has changed? The only difference is that this uses the number of bytes for determining the conversion and the new code uses the number of registers (=16bit). Furthermore, `TestHoldingRegisters` will check all those conversions.

One thing that changed though (for integers only) is that we only produce 64-bit output instead of 16, 32 or 64-bit variants for now in order to minimize the number of type conversion functions...

plugins/inputs/modbus/modbus_test.go
srebhan: Well the idea here is to write the test values into the mobus-server registers. If this fails it will fail for all values, i.e. the server doesn't work... But I can change this back if you want me to.

srebhan:
@ssoroka I fully agree that it should see some testing. I will give it a first test-run next week with about 20 devices, but would be happy if we get some more coverage with different devices...

srebhan:
Added some debugging output to show the raw-values on debug to verify type-conversions. Furthermore, extend the README with a trouble-shooting section. Thanks @binueda for his review of the Trouble-shooting section on slack. :-)

srebhan:
!retry-failed

plugins/inputs/modbus/modbus.go
ssoroka: did we drop support for `FLOAT64-IEEE` or any other types?

srebhan: No we didn't (take a look at the tests). The type conversions moved to separate files (`type_conversions*.go`).

In preparation of new configuration schemes I tried to unified the names because e.g. `FLOAT32`, `FLOAT32-IEEE`, `FIXED` and `UFIXED` caused a LOT of confusion. So the idea is to have a clear regime for future configuration schemes and still support the "old" configuration style. Therefore, the types are now normalized as `FLOAT32` & `UFIXED` to `UINTx` and `FIXED` to `INTx` (depending on the bit-size) in the configuration setup (see `normalizeInputDatatype` in `configuration_original.go`).

srebhan:
!retry-failed

ssoroka (approved):
approved; waiting for user testing

wz2b:
If the intent here is to eliminate 8013 and merge in its features there are a few things it needs to be able to do:

 * Specify a unit address per-request, rather than once for the entire driver instance
 * Allow you to 'skip' registers that come back as in a "read multiple" (holding, input, or discrete) request that you don't care about i.e. they shouldn't end up as fields, but you also shouldn't be forced to make them individual requests
 * Per the modbus organization's statement we really should replace the word "slave" with the word "unit" (continuing to silently support "slave_id" is not a problem)

In particularly, the "gateway" functionality provided by #8013 is needed, and I don't think it's addressed here.  The issue is allowing you to talk to multiple devices attached to a single gateway without having to have separate TCP connections for each unit.  This is a common scenario that occurs even when talking to devices that are not traditionally "gateways."  As an example, one of the more popular solar inverters out there is the SMA SB series (e.g. SB3000).  It has multiple register sets located inside it, and there are occasions where you want to be able to use both at the same time.  To select one vs the other you use a different unit address.  Essentially, this inverter acts as two "devices" in one.  The other common scenario is strings of power meters e.g. the Square D smart breakers where each breaker is its own device (with its own unit address).

If this PR addresses the above then it's fine; if it's not, then I don't think this replaces or merges 8013.  If I'm wrong please let's discuss.

srebhan:
@wz2b as discussed on Slack, this is only a first stage. I hope submitted #9279 clarifies things. Thanks for offering your testing of #9279 on slack. Looking forward to your comments!

srebhan:
@mirkocomparetti-synesis thanks for testing!
Looking at the code it might be possible to add this feature (bitwise registers as BOOL) by parsing the datatype field. If this is ok with you, I will postpone adding it to a later point in time. Please add an issue so I won't forget. Thanks in advance!

srebhan:
@ssoroka testing was successful. :-)

mirkocomparetti-synesis:
Agreed @srebhan it can be postponed! Here it is the ticket https://github.com/influxdata/telegraf/issues/9284 for your reference.

plugins/inputs/modbus/modbus.go
```diff
@@ -173,552 +167,265 @@ func (m *Modbus) Init() error {
...
+	m.isConnected = false
+	return err
+}
 
-			// check scale
-			if item.Scale == 0.0 {
-				return fmt.Errorf("invalid scale '%f' in '%s' - '%s'", item.Scale, n, item.Name)
-			}
+func (m *Modbus) gatherFields() error {
+	for slaveID, requests := range m.requests {
```
garciaolais, line 298: remove slaveID
`for _, requests := range m.requests {`

srebhan: Done.

PR #9150, opened 2021-04-19
Adding queries for collecting QDS data for SQL DB

masree:
### Required for all PRs:

resolves #

Summary of PR:

Igor started the [PR ](https://github.com/influxdata/telegraf/pull/8465/files)to collect data from query store runtime and wait statistics for SQL DB and SQL MI. Continuing his PR to completion as he does not own this feature anymore. There are a few review comments that needed to be addressed from old PR. Steps to address them:

1. Split the PR into 2 - SQL DB and SQL MI
2. Address the review comments related to SQL DB
3. Get it reviewed from community
4. Merge SQL DB PR
5. Create a new PR for SQL MI as this needs more work (we need to iterate the individual DBs in a managed instance)
6. Repeat 2-4


andrewzayats:
@denzilribeiro @dimitri-furman  could you please review?
End-to-end tests were done using telegraf standalone configuration and Azure SQL Insights solution.

plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -679,6 +679,138 @@ SELECT
...
+	MAX(rs.max_query_max_used_memory) AS max_query_max_used_memory,
+	SUM(rs.avg_query_max_used_memory * rs.count_executions) / SUM(rs.count_executions) AS avg_query_max_used_memory,
+	MAX(rs.max_duration) as [max_duration],
+	SUM(rs.avg_duration * rs.count_executions) / SUM(rs.count_executions) as avg_duration,
+	MAX(rs.max_num_physical_io_reads) as max_num_physical_io_reads,
+	SUM(rs.avg_num_physical_io_reads * rs.count_executions) / SUM(rs.count_executions) as avg_num_physical_io_reads,
+	MAX(rs.max_page_server_io_reads) as max_page_server_io_reads,
+	SUM(rs.avg_page_server_io_reads * rs.count_executions) / SUM(rs.count_executions) as avg_page_server_io_reads,
+	MAX(rs.max_log_bytes_used) as [max_log_bytes_used],
+	SUM(rs.avg_log_bytes_used * rs.count_executions) / SUM(rs.count_executions) as avg_log_bytes_used
```
dimitri-furman, line 765: This is missing a couple of important measures (memory grants and tempdb usage):
1. max_used_memory
2. tempdb_space_used

We should add these.

plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -679,6 +679,138 @@ SELECT
...
+	'sqlserver_azuredb_querystore_wait_stats' AS [measurement],
+	REPLACE(@@SERVERNAME, '\', ':') AS [sql_instance],
+	DB_NAME() AS [database_name],
+	CONVERT(nvarchar(35), MAX(rsi.start_time), 126) AS interval_start_time,
+	CONVERT(nvarchar(35), MAX(rsi.end_time), 126) AS interval_end_time,
+	CAST(MAX(q.query_id) AS nvarchar(20)) AS query_id,
+	CONVERT(nvarchar(20), MAX(q.query_hash), 1) as query_hash,
+	CAST(ws.plan_id AS nvarchar(20)) AS plan_id,
+	CONVERT(nvarchar(20), MAX(p.query_plan_hash), 1) as query_plan_hash,
+	ws.wait_category_desc AS wait_category_s,
```
dimitri-furman, line 796: Looks like a minor inconsistency in column name suffixes: `wait_category_s` but `exec_type_desc`.

plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -679,6 +679,138 @@ SELECT
...
+	ws.execution_type AS exec_type_d,
+	MAX(ws.execution_type_desc) AS exec_type_desc,
+	SUM(ws.total_query_wait_time_ms) AS total_query_wait_time_ms_d,
+	CAST(SUM(IIF(avg_query_wait_time_ms = 0, 0, ws.total_query_wait_time_ms / avg_query_wait_time_ms)) AS BIGINT) AS count_executions,
+	MAX(ws.max_query_wait_time_ms) AS max_query_wait_time_ms_d,
+	MAX(q.query_parameterization_type) AS query_param_type_d
```
dimitri-furman, line 802: Curious, why do we need parameterization type with waits?

plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -679,6 +679,138 @@ SELECT
...
+	CONVERT(nvarchar(35), MAX(rsi.end_time), 126) AS interval_end_time,
+	CAST(MAX(q.query_id) AS nvarchar(20)) AS query_id,
+	CONVERT(nvarchar(20), MAX(q.query_hash), 1) as query_hash,
+	CAST(ws.plan_id AS nvarchar(20)) AS plan_id,
+	CONVERT(nvarchar(20), MAX(p.query_plan_hash), 1) as query_plan_hash,
+	ws.wait_category_desc AS wait_category_s,
+	ws.execution_type AS exec_type_d,
+	MAX(ws.execution_type_desc) AS exec_type_desc,
+	SUM(ws.total_query_wait_time_ms) AS total_query_wait_time_ms_d,
+	CAST(SUM(IIF(avg_query_wait_time_ms = 0, 0, ws.total_query_wait_time_ms / avg_query_wait_time_ms)) AS BIGINT) AS count_executions,
```
dimitri-furman, line 800: This is a very imprecise way to calculate the number of executions, each execution can wait for drastically different durations. Why do we need count of executions here?

plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -679,6 +679,138 @@ SELECT
...
+	MAX(rs.max_page_server_io_reads) as max_page_server_io_reads,
+	SUM(rs.avg_page_server_io_reads * rs.count_executions) / SUM(rs.count_executions) as avg_page_server_io_reads,
+	MAX(rs.max_log_bytes_used) as [max_log_bytes_used],
+	SUM(rs.avg_log_bytes_used * rs.count_executions) / SUM(rs.count_executions) as avg_log_bytes_used
+FROM sys.query_store_query q WITH (NOLOCK)
+	JOIN sys.query_store_plan p WITH (NOLOCK) ON q.query_id = p.query_id
+	JOIN sys.query_store_runtime_stats rs WITH (NOLOCK) ON p.plan_id = rs.plan_id
+	JOIN sys.query_store_runtime_stats_interval rsi WITH (NOLOCK) ON rs.runtime_stats_interval_id = rsi.runtime_stats_interval_id
+WHERE rsi.start_time >= @currIntervalStartDate AND rsi.end_time <= @currIntervalEndDate
+GROUP BY rs.plan_id, rs.execution_type, rs.runtime_stats_interval_id;
```
dimitri-furman, line 771: I would add OPTION (RECOMPILE) to this query and the waits query below. As Query Store data grows and is purged, cached plans may become suboptimal. With the frequency of execution we have for these, the overhead is minimal.

Speaking of query execution time, do we have timeout set on the client? Sometimes, queries against QDS will end up with poor plans and take a very long time. We should protect against that with a query execution timeout.

dimitri-furman (changes requested):

plugins/inputs/sqlserver/azuresqlqueries.go
andrewzayats: @dimitri-furman, I added option (recompile) to the queries. As for command timeout it hasn't been implemented in the plugin yet. And I don't think it's a good idea to impement command timeout support in this pull request. There is already issue #2824 and closed pull request #6677 related to adding timeout support.

dimitri-furman: Agreed, that PR addresses this.

plugins/inputs/sqlserver/azuresqlqueries.go
andrewzayats: @dimitri-furman if this calculation is imprecise, what can you suggest instead? This field is necessary for Azure SQL Insights solution.

dimitri-furman: To get the precise count, you would need to aggregate runtime stats in the same way as you are aggregating waits, and then join the two result sets. But before we do this, I'd like to understand how this column is used in SQL Insights. Let's discuss offline.

plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -679,6 +679,140 @@ SELECT
...
+
+const sqlAzureDBQueryStoreWaitStatistics = sqlAzureDBPartQueryPeriod + `
+
+/*QDS is not enabled on read-only replicas: https://docs.microsoft.com/en-us/sql/t-sql/functions/databasepropertyex-transact-sql?view=sql-server-ver15*/
+IF DATABASEPROPERTYEX(DB_Name(), 'Updateability') = 'READ_ONLY'  
+	RETURN
+
+/* Query the data from query store. Group  by plan_id, runtime_stats_interval_id, execution_type and wait_category, and specified here:
+https://docs.microsoft.com/en-us/sql/relational-databases/system-catalog-views/sys-query-store-wait-stats-transact-sql?view=sql-server-ver15
+*/
```
dimitri-furman, line 786: We can add a perf optimization here and skip this query if Query Store is configured not to capture waits. Check `wait_stats_capture_mode` in `sys.database_query_store_options`.

plugins/inputs/sqlserver/azuresqlqueries.go
andrewzayats: @dimitri-furman, the query already contains field **max_query_max_used_memory**. It looks like what you mentioned. 2 fields related to tempdb were added: **max_tempdb_space_used** and **avg_tempdb_space_used**

plugins/inputs/sqlserver/azuresqlqueries.go
andrewzayats: @dimitri-furman, fixed. Field names with suffix **_d** also fixed.

plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -679,6 +679,138 @@ SELECT
...
+	SET @currIntervalStartTimestamp = @lastIntervalEndTimestamp;
+
+/* Convert start and end times to datetime */
+DECLARE @currIntervalEndDate datetimeoffset = DATEADD(second, @currIntervalEndTimestamp % @secondsInDay, DATEADD(day, @currIntervalEndTimestamp / @secondsInDay, 0));
+DECLARE @currIntervalStartDate datetimeoffset = DATEADD(second, @currIntervalStartTimestamp % @secondsInDay, DATEADD(day, @currIntervalStartTimestamp / @secondsInDay, 0));
+`
+const sqlAzureDBQueryStoreRuntimeStatistics = sqlAzureDBPartQueryPeriod + `
+
+/*QDS is not enabled on read-only replicas: https://docs.microsoft.com/en-us/sql/t-sql/functions/databasepropertyex-transact-sql?view=sql-server-ver15*/
+IF DATABASEPROPERTYEX(DB_Name(), 'Updatability') = 'READ_ONLY'
```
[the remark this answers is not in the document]
andrewzayats: @dimitri-furman, Updateability fixed. As for **is_read_only** I need a bit more details from you. You said that we needed to collect recent Query Store data right after the db had been made read-only. But we cannot get exact date when the db was made read-only. This means that we should always collect Qury Store data related to the DB and ignore **is_read_only** flag. Could you please clarify this?

plugins/inputs/sqlserver/README.md
```diff
@@ -62,11 +62,11 @@ GO
 
   # database_type = "AzureSQLDB"
 
-  ## A list of queries to include. If not specified, all the above listed queries are used.
+  ## A list of queries to include. 
   # include_query = []
 
-  ## A list of queries to explicitly ignore.
-  # exclude_query = []
+  ## A list of queries to explicitly ignore. QDS queries are excluded by default.
```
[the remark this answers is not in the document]
andrewzayats: Done

plugins/inputs/sqlserver/README.md
```diff
@@ -107,6 +108,10 @@ GO
   ## This setting/metric is optional and is disabled by default.
   # health_metric = false
 
+  ## Toggling this to true enables telegraf to start collecting data from query store.
+  ## This setting is optional and is disabled by default as it the queries are expensive and collection interval should be >=15m
```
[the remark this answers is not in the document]
andrewzayats: @dimitri-furman the text above is about telegraf interval. I agree that the text should be fixed.

plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -679,6 +679,138 @@ SELECT
...
+	RETURN
+
+/* Query the data from query store. Group by plan_id, execution_type, runtime_stats_interval_id, as specified here:
+https://docs.microsoft.com/en-us/sql/relational-databases/system-catalog-views/sys-query-store-runtime-stats-transact-sql?view=sql-server-ver15
+*/
+SELECT
+	'sqlserver_azuredb_querystore_runtime_stats' AS [measurement],
+	REPLACE(@@SERVERNAME, '\', ':') AS [sql_instance],
+	DB_NAME() AS [database_name],
+	CONVERT(nvarchar(35), MAX(rsi.start_time), 126) AS interval_start_time,
```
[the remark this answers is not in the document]
andrewzayats: Done.

plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -679,6 +679,138 @@ SELECT
...
+DECLARE @lastIntervalEndTimestamp BIGINT = 0;
+
+/* Get the time when the query was previously called */
+IF @QueryData != '' AND @QueryData IS NOT NULL
+	SET @lastIntervalEndTimestamp = CAST(@QueryData AS BIGINT);
+
+DECLARE @secondsInDay INT = 86400;
+
+/* Get query aggregation interval from query store */
+DECLARE @timeGrain BIGINT = (SELECT interval_length_minutes * 60 FROM sys.database_query_store_options WITH (NOLOCK));
```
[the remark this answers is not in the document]
andrewzayats: NOLOCK was removed from QDS queries

plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -679,6 +679,138 @@ SELECT
...
+If the query was never called before, start time is end time minus @timeGrain */
+DECLARE @currIntervalStartTimestamp BIGINT = @currIntervalEndTimestamp - @timeGrain;
+
+/* If the query was called before, start time is when the query was previously called, so that we return all the records since the previous call */
+IF @lastIntervalEndTimestamp != 0
+	SET @currIntervalStartTimestamp = @lastIntervalEndTimestamp;
+
+/* Convert start and end times to datetime */
+DECLARE @currIntervalEndDate datetimeoffset = DATEADD(second, @currIntervalEndTimestamp % @secondsInDay, DATEADD(day, @currIntervalEndTimestamp / @secondsInDay, 0));
+DECLARE @currIntervalStartDate datetimeoffset = DATEADD(second, @currIntervalStartTimestamp % @secondsInDay, DATEADD(day, @currIntervalStartTimestamp / @secondsInDay, 0));
```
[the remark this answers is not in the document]
andrewzayats: @dimitri-furman, what about using **AT TIME ZONE 'UTC'** everywhere we need to compare or output datetimeoffsets? It may help to avoid issues related to DST. And it looks like DST is not applicable to Azure SQL DB, it always uses UTC.

plugins/inputs/sqlserver/README.md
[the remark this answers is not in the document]
dimitri-furman: Looks like there are more instances of QDS in the PR, please search for all.

plugins/inputs/sqlserver/azuresqlqueries.go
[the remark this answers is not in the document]
dimitri-furman: It may work functionally for time zone issues on MI and SQL VM. But there could be perf impact in Query Store collection queries because we would need to use AT TIME ZONE on the column side of the predicate, and that will prevent index seeks.

The change would do nothing to solve potential problems when Query Store interval duration is changed.

More generally, I don't see the value in back and forth conversions between bigint and datetimeoffset, they seem to just add unnecessary complexity. Am I overlooking the goal there? It just seems overall easier to cache datetimeoffset if that's what we actually use in the queries to filter.

plugins/inputs/sqlserver/azuresqlqueries.go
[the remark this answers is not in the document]
dimitri-furman: Yes, we should always collect from primary even if it's read-only. Something like this:
```
IF DATABASEPROPERTYEX(DB_Name(), 'Updateability') = 'READ_ONLY'
   AND
   EXISTS (
          SELECT *
          FROM sys.databases
          WHERE is_read_only = 0
          )
    RETURN -- skip replicas, but don't skip read-only primary
```

plugins/inputs/sqlserver/azuresqlqueries.go
dimitri-furman: Yes, I missed `max_query_max_used_memory` being there already, thanks for adding the other one.

plugins/inputs/sqlserver/azuresqlqueries.go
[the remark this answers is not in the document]
andrewzayats: @dimitri-furman, If to use the code above, in the case of readonly db the queries will execute on all replicas, Primary and Secondaries. At the same time we can use the following views to distinguish between primary an secondary replicas: **dm_database_replica_states** for _read-scale replicas_ and **dm_geo_replication_link_status** for _geo-replicas_. I have no info about their performance impact. Could you please share your thoughts?

plugins/inputs/sqlserver/azuresqlqueries.go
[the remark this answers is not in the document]
andrewzayats: @dimitri-furman, the interval calculation logic was changed and now it doesn't use bigint.

plugins/inputs/sqlserver/azuresqlqueries.go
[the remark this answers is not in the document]
dimitri-furman: Good catch, thanks. 

dm_database_replica_states does not work on Hyperscale, so can't use that. There will be a new DMV in coming months that will work for all service tiers including geo-replicas, we can address this at that time. 

For now, can leave this code as is.

plugins/inputs/sqlserver/azuresqlqueries.go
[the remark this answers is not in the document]
dimitri-furman: Thanks for making this change, looks good.

I haven't looked at Managed Instance and SQL Server queries, but for all recent changes made for SQL DB Query Store collection, similar changes should be made for MI and SQL Server too.

plugins/inputs/sqlserver/azuresqlqueries.go
andrewzayats: @dimitri-furman, in order to not confuse with **count_executions** from QDS runtime stats, the column was renamed to **count_waits**.

dimitri-furman: This is still problematic I am afraid. For one thing, "count" implies that the value is an integer, but this is fractional. For another, "wait" has a very specific meaning in the context of MSSQL query execution, i.e. see the definition of `waiting_tasks_count` in [sys.dm_os_wait_stats](https://docs.microsoft.com/en-us/sql/relational-databases/system-dynamic-management-views/sys-dm-os-wait-stats-transact-sql). This new name makes it look like this is the same thing, but it's actually something quite different and artificial.

If this is only needed for SQL Insights, then I don't think we should have this column in this query at all. The data we need is already collected in the runtime stats query for each query hash (`count_executions`), so a calculation involving waits for a query hash should be done in KQL, not as part of collection.

andrewzayats: @dimitri-furman, I'am afraid that using of count_executions (taken from sys.query_store_runtime_stats) leads to wrong results in most cases, because the same query may face with different wait types when executed multiple times, i.e. a query may be executed 10 times but it may face with memory waits only 6 times and this value then used to calculate wait statistics in QDS. This field cannot be calculated in KQL after all aggregations in the telegraf's query. Without this field we won't be able to calculate correct average values for wait statistics in KQL. And this count is not fractional, when you divide total on avg (both taken from **sys.query_store_wait_stats**) you will almost always receive integer (or smth. very close to integer in rare cases) because this integer value is used to calculate the avg. And I don't know why **sys.query_store_wait_stats** doen't contain any field containing count, it looks like an issue. Azure SQL DB QDS wait stats metrics (can be enabled by turning on **QueryStoreWaitStatistics** in Diagnostic settings) contain count. And often it differs from **count_executions**.
I don't understand, why you mentioned **sys.dm_os_wait_stats**. The field you mentioned has different name, it is located in another view and therefore has different meaning. These fields can have similar but not the same meaning.

dimitri-furman: I checked with a QDS SME and confirmed that we account for every query execution when calculating `avg_query_wait_time_ms`, not just the ones where the query actually waited, which was my assumption. So the current calculation in the query does work, and produces an accurate number of query executions per interval *and wait category* (so it won't usually match execution count per interval from runtime stats). 

This means that the original column name, `count_executions`, was actually accurate. `count_waits` is a bit misleading because it implies that we are counting the number of waits, which can be more or less than the number of executions (a query could wait multiple times during execution, and some executions may not wait at all as you pointed out). I think we should revert this name change.

Apologies for the confusion and runaround.

Minor style nit: it should be `ws.avg_query_wait_time_ms`, not `avg_query_wait_time_ms`, in two places.

andrewzayats:
@dimitri-furman @denzilribeiro, are we good to merge this pull request?

dimitri-furman:
@andrewzayats looks good to me. I believe @denzilribeiro wanted to review as well, but he is away for a couple of days. If your timeline allows, please hold off on merging until EOD Thursday June 10.

denzilribeiro (changes requested):
Looks great  overall  and very close
I still have 1 concern, mainly collecting too much data from query store under these circumstances below
- Plugin cannot to SQL for 2 days, telegraf is up so has cache, next iteration will collect 2 days worth of query store data and that that query is super expensive.
Ideally we limit collection to max of 1 hour or 1 query store interval( if that interval is configured for more than 1 hour), we can discuss offline on right threshold.

andrewzayats:
> Looks great overall and very close
> I still have 1 concern, mainly collecting too much data from query store under these circumstances below
> 
> * Plugin cannot to SQL for 2 days, telegraf is up so has cache, next iteration will collect 2 days worth of query store data and that that query is super expensive.
>   Ideally we limit collection to max of 1 hour or 1 query store interval( if that interval is configured for more than 1 hour), we can discuss offline on right threshold.

@denzilribeiro, I added a limit on the max QDS collection period. I think 3h is good enough for this.

dimitri-furman:
> > Looks great overall and very close
> > I still have 1 concern, mainly collecting too much data from query store under these circumstances below
> > 
> > * Plugin cannot to SQL for 2 days, telegraf is up so has cache, next iteration will collect 2 days worth of query store data and that that query is super expensive.
> >   Ideally we limit collection to max of 1 hour or 1 query store interval( if that interval is configured for more than 1 hour), we can discuss offline on right threshold.
> 
> @denzilribeiro, I added a limit on the max QDS collection period. I think 3h is good enough for this.

With this change, will we collect anything if QDS is configured with a daily interval?

andrewzayats:
> > > Looks great overall and very close
> > > I still have 1 concern, mainly collecting too much data from query store under these circumstances below
> > > 
> > > * Plugin cannot to SQL for 2 days, telegraf is up so has cache, next iteration will collect 2 days worth of query store data and that that query is super expensive.
> > >   Ideally we limit collection to max of 1 hour or 1 query store interval( if that interval is configured for more than 1 hour), we can discuss offline on right threshold.
> > 
> > 
> > @denzilribeiro, I added a limit on the max QDS collection period. I think 3h is good enough for this.
> 
> With this change, will we collect anything if QDS is configured with a daily interval?

@dimitri-furman, after adding a limitation on the telegraf interval daily interval is supported. Without this limit QDS data loss is possible.

denzilribeiro (changes requested):

plugins/inputs/sqlserver/README.md
```diff
@@ -107,6 +108,12 @@ GO
   ## This setting/metric is optional and is disabled by default.
   # health_metric = false
 
+  ## Toggling this to true enables telegraf to start collecting data from Query Store.
+  ## This setting is optional and is disabled by default.
+  ## Collection interval should comply with the following restriction: 15m =< collection interval <= 2h. 
+  ## Intervals shorter than 15m may cause higher performance impact on source. Intervals longer than 2h may cause loss of data.
```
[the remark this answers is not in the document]
dimitri-furman: Minor suggestion: let's not talk about unqualified "loss of data" :), mention Query Store data specifically, or "loss of monitoring data".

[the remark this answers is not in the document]
andrewzayats: >I am still not understanding this - what is the data loss it is only if telegraf cannot connect right?

@denzilribeiro, No. Data loss is possible, if you have QDS interval of 1h, time limit 3h and a telegraf interval equal to 4h.

>Also this makes it sound like you cannot set to > 2 hours or should not?

Should not. A customer can set an interval larger than recommended, but he will face with issues. The logic controlling intervals is outside of the plugin so we cannot set a restriction on the interval. The only thing we can do is to recommend.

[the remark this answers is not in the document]
ssoroka: I would say "cause some query store data not to be collected" rather than loss of data. No data is being lost.

plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -93,8 +96,12 @@ servers = [
...
 # include_query = []
 
-## A list of queries to explicitly ignore.
-# exclude_query = []
+## A list of queries to explicitly ignore. Query Store queries are excluded by default.
+# exclude_query = [AzureSQLDBQueryStoreRuntimeStatistics, AzureSQLDBQueryStoreWaitStatistics]
+
+## Toggling this to true enables telegraf to start collecting data from query store.
+## This setting is optional and is disabled by default as it the queries are expensive and collection interval should be >=15m
+# query_store_collection = false
```
ssoroka, line 104: is this something that can be covered by the exclude queries without adding a new boolean flag?

plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -104,6 +111,7 @@ servers = [
 
 # include_query = []
 
+## A list of queries to explicitly ignore.
 # exclude_query = []
```
ssoroka, line 115: should match readme

plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -165,6 +173,12 @@ func initQueries(s *SQLServer) error {
 		queries["AzureSQLDBPerformanceCounters"] = Query{ScriptName: "AzureSQLDBPerformanceCounters", Script: sqlAzureDBPerformanceCounters, ResultByRow: false}
 		queries["AzureSQLDBRequests"] = Query{ScriptName: "AzureSQLDBRequests", Script: sqlAzureDBRequests, ResultByRow: false}
 		queries["AzureSQLDBSchedulers"] = Query{ScriptName: "AzureSQLDBSchedulers", Script: sqlAzureDBSchedulers, ResultByRow: false}
+		// These two Query Store queries are not run by default as they are expensive and collection interval should be >=15m.
+		// If user needs data from Query Store, they will need to toggle the setting to true. By default it is turned off
+		if s.QueryStoreCollection == true {
```
ssoroka, line 178:
```suggestion
		if s.QueryStoreCollection {
```

ssoroka: shouldn't this be based on the include/exclude queries instead?

plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -337,9 +352,18 @@ func (s *SQLServer) Stop() {
...
-func (s *SQLServer) gatherServer(pool *sql.DB, query Query, acc telegraf.Accumulator) error {
+func (s *SQLServer) gatherServer(pool *sql.DB, server string, query Query, acc telegraf.Accumulator) error {
+	var rows *sql.Rows
+	var err error
+
 	// execute query
-	rows, err := pool.Query(query.Script)
+	if query.HasCachedData {
+		cachedData, _ := query.DataCache.Get(server)
+		rows, err = pool.Query(query.Script, sql.Named("p1", cachedData))
```
ssoroka, line 362: what's the purpose of the cache here?

ssoroka (changes requested):
Not sure I understand how the cache works. There's also no new tests for it.

plugins/inputs/sqlserver/sqlserver.go
andrewzayats: @ssoroka, justification from @denzilribeiro, why do we need this flag, available at the following link: [https://github.com/influxdata/telegraf/pull/8465#issuecomment-823711117](https://github.com/influxdata/telegraf/pull/8465#issuecomment-823711117)

plugins/inputs/sqlserver/sqlserver.go
andrewzayats: @ssoroka, this question is covered in the comment about config flag **query_store_collection**:
>is this something that can be covered by the exclude queries without adding a new boolean flag?

[https://github.com/influxdata/telegraf/pull/9150#discussion_r652166136](https://github.com/influxdata/telegraf/pull/9150#discussion_r652166136)

plugins/inputs/sqlserver/sqlserver.go
andrewzayats: @ssoroka, in the current plugin implementation the cache is used to save state of Query Store queries. More info about saving state in Query Store queries you can get from PR #8465 and from comment [https://github.com/influxdata/telegraf/pull/8465#issuecomment-751884278](https://github.com/influxdata/telegraf/pull/8465#issuecomment-751884278)

andrewzayats:
@ssoroka, could you please review the PR again?

plugins/inputs/sqlserver/README.md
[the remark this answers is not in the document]
andrewzayats: The text is fixed

plugins/inputs/sqlserver/README.md
```diff
@@ -270,9 +275,10 @@ These are metrics for Azure SQL Database (single database) and are very similar
...
 - *AzureSQLOsWaitstats*: Wait time in ms from `sys.dm_os_wait_stats`, number of waiting tasks, resource wait time, signal wait time, max wait time in ms, wait type, and wait category. The waits are categorized using the same categories used in Query Store. These waits are collected as they occur and instance wide
-- *AzureSQLDBRequests: Requests which are blocked or have a wait type from `sys.dm_exec_sessions` and `sys.dm_exec_requests`
-- *AzureSQLDBSchedulers* - This captures `sys.dm_os_schedulers` snapshots.
-
+- *AzureSQLDBRequests*: Requests which are blocked or have a wait type from `sys.dm_exec_sessions` and `sys.dm_exec_requests`
+- *AzureSQLDBSchedulers*: This captures `sys.dm_os_schedulers` snapshots
+- AzureSQLDBQueryStoreRuntimeStatistics: Collects queries runtime statistics from `sys.query_store_runtime_stats` 
+- AzureSQLDBQueryStoreWaitStatistics: Collects queries wait statistics from `sys.query_store_wait_stats`
```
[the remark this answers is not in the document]
andrewzayats: We left mentions of QDS views that contain all metrics and added "and other views related to Query Store". I believe its enough. Otherwise we have to fix all other queries not related to Query Store.

plugins/inputs/sqlserver/azuresqlqueries.go
andrewzayats: Parameterization was removed

plugins/inputs/sqlserver/azuresqlqueries.go
andrewzayats: Fixed

plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -679,6 +679,137 @@ SELECT
...
+SELECT TOP 1 @queryStartTime = start_time,  @currIntervalEndTime = end_time
+FROM sys.query_store_runtime_stats_interval
+WHERE end_time < @currTime
+ORDER BY runtime_stats_interval_id DESC;
+
+/* Query Store is disabled OR interval is already collected */
+if @currIntervalEndTime IS NULL OR @currIntervalEndTime < @currTimeLimit OR @currIntervalEndTime = @lastIntervalEndTime
+	RETURN;
+
+SET @currIntervalStartTime = IIF(@lastIntervalEndTime IS NULL OR @lastIntervalEndTime < @currTimeLimit, @queryStartTime, @lastIntervalEndTime);
```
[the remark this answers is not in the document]
andrewzayats: The limitation logic was changed to the proposed one.

plugins/inputs/sqlserver/sqlserver.go
andrewzayats: Fixed

plugins/inputs/sqlserver/README.md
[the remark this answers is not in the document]
andrewzayats: Done.

dimitri-furman (approved):

denzilribeiro (approved):

plugins/inputs/sqlserver/azuresqlqueries.go
andrewzayats: Done.

andrewzayats:
@sjwang90, @ssoroka seems to be busy. Can you please help us get this PR reviewed and merged?

sjwang90:
@andrewzayats Sure we can get someone to take a look. Can you make sure all tests are passing?

andrewzayats:
> @andrewzayats Sure we can get someone to take a look. Can you make sure all tests are passing?

@sjwang90, sqlserver tests are OK ([link](https://app.circleci.com/pipelines/github/influxdata/telegraf/5410/workflows/b13c9c13-f0d7-4bf1-97e3-d42b90c379b0/jobs/98103?invite=true#step-106-501)). The failing test refers to input plugin **tail** ([link](https://app.circleci.com/pipelines/github/influxdata/telegraf/5410/workflows/b13c9c13-f0d7-4bf1-97e3-d42b90c379b0/jobs/98103?invite=true#step-106-602)). We are not responsible for this plugin and probably branch **master** already contains a fix.

plugins/inputs/sqlserver/sqlserver.go
srebhan: I think the question is more can we have something like
```toml
enabled_features = ["query_store_collection",...]
```
to not add a list of booleans as new features are added.

plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -556,3 +592,25 @@ func init() {
...
+type QueryDataCache struct {
+	mx sync.RWMutex
+	m  map[string]string
+}
+
+func (c *QueryDataCache) Get(key string) (string, bool) {
+	c.mx.RLock()
+	defer c.mx.RUnlock()
+	val, ok := c.m[key]
+	return val, ok
```
srebhan, line 605:
```suggestion
	return c.m[key]
```

srebhan (commented):
Hey @masree, thanks for the PR. I have only a minor comment but I'm, to be honest, not the right person to review the SQL statements.
However, can you please comment on Stevens question regarding a "feature-enable list" instead of using booleans to switch on/off features? I think it is much more extensible...

plugins/inputs/sqlserver/sqlserver.go
andrewzayats: The proposed code change leads to compile error. The Get method should return 2 values: the value itself and success flag. The suggested change leads to the case when the method returns only one value.

srebhan: Sorry for the noise!

plugins/inputs/sqlserver/sqlserver.go
masree: @srebhan : On second thought to what @ssoroka called out, we can just use exclude queries without adding a new boolean flag. This means that for old config + new telegraf bits, QDS data would start to collect. Is that an acceptable scenario and can we add extra documentation for users to take latest config to avoid such scenarios? I agree that adding too many optional parameters is not scalable and would it be possible to just use the existing exclude property instead of another property enabled_features as well. let us know your thoughts

denzilribeiro:
> However, can you please comment on Stevens question regarding a "feature-enable list" instead of using booleans to switch on/off features? I think it is much more extensible...

@ssoroka , @srebhan  -- This  is no different than turning queries on/off right aka feature-enable  list or include_queries list is the same ? Unless the feature-enable groups a set of queries into a "feature" that you turn on/off.  The difference with query store is that it is generally different in terms of granularity of collection and the query is a bit more heavy duty than say the other light weight queries and isn't really a live monitoring metric but lets you analyze queries and perf that contributed to resource usage in time. 

The crux of the problem is we wanted to  disable by default and have someone explicitly enable, so we wanted a disable by default, explicitly opt in behavior.

Take for example someone who had telegraf prior to this PR, doesn't change their config files to exclude this explicitly, they get this collector turned on by default and will have to explicitly add to exclude_query list -- if that is fine then we can just use the exclude_list . That isn't just for query store related collector only but should agree upon an approach for any of the more "advanced" collectors or bit more heavy weight collectors. If exclude_query is the way to go that is fine just we would introduce new things for existing customers that don't change their config file but update telegraf. Query store being a bit of a different type of collector thought of using explicit flag but I get the point that a flag for a feature in time can explode the conf section.
 
We do understand that the same could be achieved with this ( and just use exclude list) but it won't be disable by default, rather onus is on customers.
exclude_query = ["SQLServerAvailabilityReplicaStates", "SQLServerDatabaseReplicaStates","AzureSQLDBQueryStoreRuntimeStatistics","AzureSQLDBQueryStoreWaitStatistics"]

srebhan:
@denzilribeiro the difference is in the extensibility of both schemes. Imagine adding another "thing to collect".

With the boolean approach you will need to extend the config (sampleConfig, README, struct) with another option that is say disabled by default.

With the feature list approach, you just implement the feature and add the "feature enable string" to the sampleconfig and README **documentation**. No need to change the actual configuration. Furthermore, if you have more than a handful of those feature enable/disable toggles, a include list is much more readable that a bulky list of true/false options.
You can still keep the backward compatibility by carefully crafting the `feature_include`/`feature_exclude` lists.

So yes, both approaches are equivalent feature-wise, but the latter is much more extensible/readable in the long-run.

plugins/inputs/sqlserver/sqlserver.go
srebhan: @masree I think you should better use `include_query` instead of `exclude_query` to guarantee an _explicit_ opt-in. Otherwise, people that have an existing `exclude_query` list will *automagically* get a new query which is surely not intended (given that they use the option in the first place). But yeah reusing that mechanism was exactly the point of @ssoroka...

denzilribeiro:
> You can still keep the backward compatibility by carefully crafting the `feature_include`/`feature_exclude` lists.
> So yes, both approaches are equivalent feature-wise, but the latter is much more extensible/readable in the long-run.

Right now feature_include/feature_exclude is  the same as exclude_query /include_query which is already implemented. So if the decision is still to revert to that, then it is fine 

At some point look at this a bit more holistically. The issue I have with include/exclude you don't have a notion of "default collectors" for those that don't specify the include/exclude lists and so that moves the onus to customers.

Say we have queries  a,b,c,d in existing version and now we introduce a query e that we want enabled by default, and a query f that we want disabled by default
a. Customers that have an include list in existing conf file have to add query e
b. Customers with an exclude list in their conf file will have to add query f to exclude it
c. Customers that do not have an include/exclude list will get both e and f.

Sure we can change the default conf to reflect it, but a customer that already has a conf and doesn't update it won't get it, maybe I am missing something :)

srebhan:
> Right now feature_include/feature_exclude is the same as exclude_query /include_query which is already implemented. So if the decision is still to revert to that, then it is fine

I'm fine with using the existing lists.

> At some point look at this a bit more holistically. The issue I have with include/exclude you don't have a notion of "default collectors" for those that don't specify the include/exclude lists and so that moves the onus to customers.

I'm not sure I understand your comment correctly. You can still define a default `include` or `exclude` set in your plugin and put that into the sample config. If the user chooses to modify it, he/she will start with the default and adapt to the needs.

> Say we have queries a,b,c,d in existing version and now we introduce a query e that we want enabled by default, and a query f that we want disabled by default

Then I would specify:
```toml
# query_include = ["a", "b", "c", "d", "e"]
# query_exclude = []
```
in your sample config and set these options when setting up the plugin in `init()` (lowercase `i`):
```golang
func init() {
	inputs.Add("sqlserver", func() telegraf.Input {
		return &SQLServer{
			Servers: []string{defaultServer},
			IncludeQuery: []string{"a", "b", "c", "d", "e"},
		}
	})
}
```

> a. Customers that have an include list in existing conf file have to add query e

Right, but that is the _semantics_ behind an `include` list. You explicitly _whitelist_ a set of features. So you don't want to get new stuff *automagically*. :-)

> b. Customers with an exclude list in their conf file will have to add query f to exclude it

No because you did not include it in the first place, so you don't need to exclude it. Note you are free to craft the `include` list such that only ever backward-compatible features are enabled. The `exclude` list will then superseed the include list.

> c. Customers that do not have an include/exclude list will get both e and f.

See above. Just do not specify `e` and `f` in your default include list.

> Sure we can change the default conf to reflect it, but a customer that already has a conf and doesn't update it won't get it, maybe I am missing something :)

Right, but again, if somebody sets an `include` list, he/she will get _exactly_ the stuff he/she specified. That's the intention, no surprises. ;-)

denzilribeiro:
Will change to include/exclude lists for now and revisit any other changes later.

Just to clarify concern wasn't ever of someone using include lists as they get exactly what they want, no more, no less. Concern was folks that weren't using include lists.. or using exclude lists in getting a potentially heavier collector since the default for all collectors as premise is by default all collectors enabled.

My experience has been in my limited experience with telegraf that  folks don't necesarily update their conf files often yet update telegraf major versions but then I maybe wrong

plugins/inputs/sqlserver/sqlserver.go
masree: Yea, I think with the existing include/exclude parameter, we should be able to achieve this behavior rather than introducing another property to control the same

srebhan:
> [...] Concern was folks that weren't using include lists.. or using exclude lists in getting a potentially heavier collector since the default for all collectors as premise is by default all collectors enabled.

Well almost. Premise is be backward-compatible. That is, if you add a new feature, you now have to explicitly define the include list (instead of "use all") to exclude the new feature. This way, people not using the include-list will still get what they got before.

> My experience has been in my limited experience with telegraf that folks don't necesarily update their conf files often yet update telegraf major versions but then I maybe wrong

No you are right and that's exactly the point. Please craft the `include` list in a way that you get the same results with the old and new version of telegraf.

andrewzayats:
@srebhan, the flag was removed. Are we good to merge this pull request?

plugins/inputs/sqlserver/README.md
```diff
@@ -66,20 +66,25 @@ GO
...
   ## AzureSQLMIResourceStats, AzureSQLMIResourceGovernance, AzureSQLMIDatabaseIO, AzureSQLMIServerProperties, AzureSQLMIOsWaitstats, 
   ## AzureSQLMIMemoryClerks, AzureSQLMIPerformanceCounters, AzureSQLMIRequests, AzureSQLMISchedulers
 
   # database_type = "AzureSQLManagedInstance"
 
+  ## A list of queries to include.
   # include_query = []
 
+  ## A list of queries to explicitly ignore.
   # exclude_query = []
```
srebhan, line 88: Somehow `database_type`, `include_query` and `exclude_query` entries are duplicated. Can you please cleanup those?

plugins/inputs/sqlserver/README.md
```diff
@@ -66,20 +66,25 @@ GO
...
-  ## A list of queries to include. If not specified, all the above listed queries are used.
+  ## A list of queries to include. 
   # include_query = []
 
   ## A list of queries to explicitly ignore.
-  # exclude_query = []
+  ## Query Store queries (AzureSQLDBQueryStoreRuntimeStatistics and AzureSQLDBQueryStoreWaitStatistics) are excluded by default.
+  ## If you enable Query Store queries, the collection interval should comply with the following restriction: 15m =< collection interval <= 2h. 
+  ## Intervals shorter than 15m may cause higher performance impact on source. Intervals longer than 2h may cause some Query Store data not to be collected.
+  # exclude_query = ["AzureSQLDBQueryStoreRuntimeStatistics", "AzureSQLDBQueryStoreWaitStatistics"]
```
srebhan, line 76: This will not work. You need to set
```toml
# include_query = ["AzureSQLMIResourceStats", "AzureSQLMIResourceGovernance", "AzureSQLMIDatabaseIO", "AzureSQLMIServerProperties", "AzureSQLMIOsWaitstats", "AzureSQLMIMemoryClerks", "AzureSQLMIPerformanceCounters", "AzureSQLMIRequests", "AzureSQLMISchedulers"]
# exclude_query = []
```
to guarantee a consistent behavior if the user uncomments `include_query`. You can still have `exclude_query = ["AzureSQLDBQueryStoreRuntimeStatistics", "AzureSQLDBQueryStoreWaitStatistics"]`, even though it requires the user to modify two settings if he/she wants to enable one of the excluded queries.

plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -38,6 +38,8 @@ type Query struct {
 	Script         string
 	ResultByRow    bool
 	OrderedColumns []string
+	HasCachedData  bool
+	DataCache      *QueryDataCache
```
srebhan, line 42: Btw. do we really need those exported?

plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -99,16 +101,21 @@ servers = [
...
 ## AzureSQLMIResourceStats, AzureSQLMIResourceGovernance, AzureSQLMIDatabaseIO, AzureSQLMIServerProperties, AzureSQLMIOsWaitstats, 
 ## AzureSQLMIMemoryClerks, AzureSQLMIPerformanceCounters, AzureSQLMIRequests, AzureSQLMISchedulers
 
 # database_type = "AzureSQLManagedInstance"
 
+## A list of queries to include.
 # include_query = []
 
+## A list of queries to explicitly ignore.
 # exclude_query = []
```
srebhan, line 119: See my comment in `README.md`.

plugins/inputs/sqlserver/sqlserver.go
srebhan: @ssoroka can this be marked resolved?

srebhan (commented):
@andrewzayats, @masree I have some more minor comments.

The most important one is to change the documentation (`README.md` and `sampleConfig`) such that the `include_query` list is specified explicitly to make sure the user can uncomment the option without a change in behavior.

plugins/inputs/sqlserver/README.md
andrewzayats: @srebhan, they are duplicated intentionally. **README** and **sampleConfig** contain different **include_query** and **exclude_query** per each **database_type**. Do you think this is wrong?

plugins/inputs/sqlserver/sqlserver.go
andrewzayats: What do you mean under **exported**? If you need more details about the purpose of DataCache you can get them from the following comment [https://github.com/influxdata/telegraf/pull/9150#discussion_r652240508](https://github.com/influxdata/telegraf/pull/9150#discussion_r652240508)

plugins/inputs/sqlserver/sqlserver.go
andrewzayats: @srebhan, please look at my reply on your comment under README.md.

plugins/inputs/sqlserver/README.md
andrewzayats: @srebhan, done.

plugins/inputs/sqlserver/README.md
srebhan: Ok fine with me.

plugins/inputs/sqlserver/sqlserver.go
srebhan: I mean that those struct members are visible and accessible outside of the plugin if starting with an uppercase letter (see https://golang.org/doc/effective_go#names). To prevent (accidental) messing with those members, you should make them start with a lowercase-letter.

plugins/inputs/sqlserver/sqlserver.go
```diff
@@ -96,19 +98,24 @@ servers = [
 # database_type = "AzureSQLDB"
 
 ## A list of queries to include. If not specified, all the above listed queries are used.
-# include_query = []
+# include_query = ["AzureSQLDBResourceStats", "AzureSQLDBResourceGovernance", "AzureSQLDBWaitStats", "AzureSQLDBDatabaseIO", "AzureSQLDBServerProperties", "AzureSQLDBOsWaitstats", "AzureSQLDBMemoryClerks", "AzureSQLDBPerformanceCounters", "AzureSQLDBRequests", "AzureSQLDBSchedulers"]
```
srebhan, line 101: Hey @andrewzayats, you changed this in the documentation, but not in the code. With the current code, your new query will automagically appear with old configs. Please initialize `IncludeQuery` as described here.

dimitri-furman:
@srebhan unfortunately we will not be able to continue work on this PR in foreseeable future. My impression is that it is close to meeting the quality bar for merging once your comment is addressed. Perhaps this can be finished by other contributors at some point in the future. Apologies, and thanks for understanding. 

CC: @masree

sjwang90:
Closing this issue. If anyone is interested in picking this up please feel free to re-open.

PR #9153, opened 2021-04-20
added is_active parameter to snmp input plugin

denizzengin:
### Required for all PRs:

resolves #

If we need to multiple config file (telegraf.d) and just active/passive (not any other config) then there is no other way but  delete and create. is_active make this for snmp config and support  other input implementations.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -827,6 +827,13 @@ func (c *Config) LoadConfigData(data []byte) error {
 					}
 				case []*ast.Table:
 					for _, t := range pluginSubTable {
+						isActive  := false						
+						if _, ok := t.Fields["is_active"]; ok {
+							c.getFieldBool(t, "is_active", &isActive)
+							if !isActive {
+								continue
+							}	
+						}
 						if err = c.addInput(pluginName, t); err != nil {
 							return fmt.Errorf("error parsing %s, %w", pluginName, err)
 						}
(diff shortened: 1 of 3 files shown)


Jagularr (changes requested):
Thanks for the PR! It's not very clear what `is_active` means (does it simply enable or disable an input plugin?), if so, it should be better documented in the config, and not linked with SNMP only in any way, as it doesn't seem unique to SNMP.

denizzengin:
> Thanks for the PR! It's not very clear what `is_active` means (does it simply enable or disable an input plugin?), if so, it should be better documented in the config, and not linked with SNMP only in any way, as it doesn't seem unique to SNMP.

Yes, it is only enable or disable config (specially multiple config). I implemented only snmp side so linked with SNM plugin but as i said it can be support others with simple implementation 
Also, i think if we use this parameter below, it's fairly simple that means load or not a config file.
[[inputs.xxx]] 
is_active = true/false

reimda:
Hi @denizzengin. I don't think I understand the problem you're trying to solve. What benefit does this new setting have over commenting out the plugin section in the config?

denizzengin:
> Hi @denizzengin. I don't think I understand the problem you're trying to solve. What benefit does this new setting have over commenting out the plugin section in the config?

Hi @reimda. Lets talk our needs,
 We have multiple separated device config in telegraf.d config directory. We also have a outer service and it updates or deletes config file and reload config. So  we collect metrics but at the same time, we need to make inactive  device config file for a while. We don't want to delete config and recreate because it can be limited time and just want to update above is_active parameter.

Actually, I added extra comment because this implementation only snmp input. If someone wants to add as same as mine in the future, I think it can be helpful. 
thanks for review.

PR #9156, opened 2021-04-20
Fix for #9135

### LINKED ISSUE #9135: Proper Configuration for inputs.aliyuncms
I am trying the aliyuncms plugin for getting metrics from Ali Cloud.

### Relevant telegraf.conf:
I tried to assess the aliyuncms plugin 
Following configurations used
```
[agent]
  [[inputs.aliyuncms]]
    discovery_regions = ["eu-central-1"]
    alicloud_access_key = "asdadasdadasd"
    alicloud_secret_key = "jabsdjasjdanjkdnjasd"
    [[inputs.aliyuncms.metrics]]
      names = ["InstanceActiveConnection", "InstanceNewConnection"]

  [[outputs.file]]
    files = ["stdout"]
```

### Expected behavior:

Metrics from Ali

### Actual behavior:

Error:

```
/telegraf-1.19.0/usr/bin# ./telegraf --config telegraf.conf
2021-04-15T13:21:31Z I! Starting Telegraf
2021-04-15T13:21:31Z E! [telegraf] Error running agent: Error loading config file telegraf.conf: plugin inputs.aliyuncms: line 2: configuration specified the fields ["alicloud_secret_key" "alicloud_access_key"], but they weren't used
la-la-laa-vm:~/telegraf-1.19.0/usr/bin#
```
### Additional info:
Release: telegraf-nightly (April 15-2021)https://dl.influxdata.com/telegraf/nightlies/telegraf-nightly_linux_amd64.tar.gz

What is the right configuration?

i-prudnikov:
### Required for all PRs:

resolves #9135 

According to symptoms, it looks like Alicloud changes the API behaviour, this fix is to address more strict metrics querying specifying region.
So we need to specify region for metric gathering. As we already have discovery region, I've decided to obsolete `discovery_region` attribute in favour of `regions` attribute that specify both: metrics and discovery scope.

Old configs are still supported and deprecation warning is printed if discovery_region is used. 
In case only `discovery_regions` is specified then `regions` is populated from it. If both specified then `discovery_regions` is omitted.


i-prudnikov:
CLA signed: !signed-cla

Jagularr (commented):
Provided that this solves the issue that other users were having, which it sounds like it does, this makes sense to me.

vipinvkmenon:
What a relief :D.... I had opened to create this PR myself....if there wasn't movement :)

plugins/inputs/aliyuncms/aliyuncms.go
```diff
@@ -23,84 +23,92 @@ import (
...
+	  
+	  ## Deprecated, use regions instead
+	  #discovery_regions = ["cn-hongkong"]
+	  
+	  ## how often the discovery API call executed (default 1m)
+	  #discovery_interval = "1m"
+	  
+	  ## Metrics to Pull (Required)
+	  [[inputs.aliyuncms.metrics]]
+	  ## Metrics names to be requested,
```
srebhan, line 96: The `sampleConfig` content should start with two-spaces to generate a consistent config file. Can you please fix this formatting issue!?

plugins/inputs/aliyuncms/aliyuncms.go
```diff
@@ -131,7 +139,8 @@ type (
 		PublicKeyID       string `toml:"public_key_id"`
 		RoleName          string `toml:"role_name"`
 
-		DiscoveryRegions  []string        `toml:"discovery_regions"`
+		Regions           []string        `toml:"regions"`
+		DiscoveryRegions  []string        `toml:"discovery_regions"` //this is obsolete
```
srebhan, line 143: It would be helpful to state the version making this obsolete. I'd say something like
```suggestion
		DiscoveryRegions  []string        `toml:"discovery_regions"` // obsolete since 1.19, use "regions" instead
```

plugins/inputs/aliyuncms/aliyuncms.go
```diff
@@ -326,71 +380,73 @@ func (s *AliyunCMS) updateWindow(relativeTo time.Time) {
...
+		req.RegionId = region
+
+		for more := true; more; {
+			resp, err := s.client.DescribeMetricList(req)
+			if err != nil {
+				return errors.Errorf("failed to query metricName list: %v", err)
+			} else if resp.Code != "200" {
+				s.Log.Errorf("failed to query metricName list: %v", resp.Message)
+				break
+			}
```
srebhan, line 401: As you return in the `if` you can simplify to
```suggestion
			if err != nil {
				return errors.Errorf("failed to query metricName list: %v", err)
			} 
			if resp.Code != "200" {
				s.Log.Errorf("failed to query metricName list: %v", resp.Message)
				break
			}
```

plugins/inputs/aliyuncms/aliyuncms.go
```diff
@@ -326,71 +380,73 @@ func (s *AliyunCMS) updateWindow(relativeTo time.Time) {
...
 
-		if len(datapoints) == 0 {
-			s.Log.Debugf("No metrics returned from CMS, response msg: %s", resp.Message)
-			break
-		}
+			var datapoints []map[string]interface{}
+			if err = json.Unmarshal([]byte(resp.Datapoints), &datapoints); err != nil {
```
srebhan, line 404: Use a local error
```suggestion
			if err := json.Unmarshal([]byte(resp.Datapoints), &datapoints); err != nil {
```

plugins/inputs/aliyuncms/aliyuncms.go
```diff
@@ -326,71 +380,73 @@ func (s *AliyunCMS) updateWindow(relativeTo time.Time) {
...
+				for key, value := range datapoint {
+					switch key {
+					case "instanceId", "BucketName":
+						tags[key] = value.(string)
+						if metric.discoveryTags != nil { //discovery can be not activated
+							//Skipping data point if discovery data not exist
+							if _, ok := metric.discoveryTags[value.(string)]; !ok &&
+								!metric.AllowDataPointWODiscoveryData {
+								s.Log.Warnf("Instance %q is not found in discovery, skipping monitoring datapoint...", value.(string))
+								continue NextDataPoint
```
srebhan, line 427: Wouldn't a `break` suffice here?

plugins/inputs/aliyuncms/discovery.go
```diff
@@ -21,39 +22,15 @@ import (
...
 type discoveryRequest interface {
 }
 
 type aliyunSdkClient interface {
 	ProcessCommonRequest(req *requests.CommonRequest) (response *responses.CommonResponse, err error)
 }
 
-type discoveryTool struct {
+// DiscoveryTool is a object that provides discovery feature
+type DiscoveryTool struct {
```
srebhan, line 33: Why do you need to export it?

plugins/inputs/aliyuncms/aliyuncms.go
```diff
@@ -237,17 +272,34 @@ func (s *AliyunCMS) Init() error {
...
+	//Check regions
+	if len(s.DiscoveryRegions) != 0 {
+		s.Log.Warn("'discovery_regions' is deprecated. Use 'regions' instead.")
+	}
+
+	if len(s.Regions) == 0 {
+		s.Regions = s.DiscoveryRegions
+	} else {
+		s.Log.Warn("'discovery_regions' is deprecated and ignored in favour of 'regions'.")
+	}
```
srebhan, line 284: I don't get this. Why do you issue this warning in case the user specified `regions`?

plugins/inputs/aliyuncms/aliyuncms.go
```diff
@@ -237,17 +272,34 @@ func (s *AliyunCMS) Init() error {
...
+
+	if len(s.Regions) == 0 {
+		s.Regions = aliyunRegionList
+		s.Log.Warnf("'regions' is not set. Metrics will be queried across %d regions:\n%s",
```
srebhan, line 288: I think this is an `info` rather than a warning as it is a valid choice of the user isn't it?

plugins/inputs/aliyuncms/aliyuncms.go
```diff
@@ -326,71 +380,73 @@ func (s *AliyunCMS) updateWindow(relativeTo time.Time) {
...
+				datapointTime := int64(0)
+				tags := map[string]string{}
+				for key, value := range datapoint {
+					switch key {
+					case "instanceId", "BucketName":
+						tags[key] = value.(string)
+						if metric.discoveryTags != nil { //discovery can be not activated
+							//Skipping data point if discovery data not exist
+							if _, ok := metric.discoveryTags[value.(string)]; !ok &&
+								!metric.AllowDataPointWODiscoveryData {
```
srebhan, line 425: I think
```suggestion
							_, ok := metric.discoveryTags[value.(string)]
							if !ok && !metric.AllowDataPointWODiscoveryData {
```
will increase readability. :-)

plugins/inputs/aliyuncms/aliyuncms.go
```diff
@@ -438,7 +494,7 @@ func (s *AliyunCMS) prepareTagsAndDimensions(metric *Metric) {
 L:
 	for {
 		select {
-		case s.discoveryData = <-s.dt.dataChan:
+		case s.discoveryData = <-s.dt.DataChan:
```
srebhan, line 497: Can you please explain why you do need to export the channel!?

plugins/inputs/aliyuncms/discovery.go
```diff
@@ -65,10 +42,17 @@ type discoveryTool struct {
 	wg       sync.WaitGroup              //WG for primary discovery goroutine
 	interval time.Duration               //Discovery interval
 	done     chan bool                   //Done channel to stop primary discovery goroutine
-	dataChan chan map[string]interface{} //Discovery data
+	DataChan chan map[string]interface{} //Discovery data
```
srebhan, line 45: Why do you need to export it?

srebhan (commented):
Thanks for tackling this so quickly! I have some comments in the code. Exporting of the discovery part and the channel seems unnecessary, so please revert if you don't have any good reason of doing it. :-)
@vipinvkmenon can you please test if this fixes your problem and add a comment!?

vipinvkmenon:
Tried another configuration:
```
[agent]
  [[inputs.aliyuncms]]
   regions = ["eu-central-1"]
   project = "acs_nat_gateway"
   access_key_id  = "xxx"
   access_key_secret = "xxx"
   period = "10m"
   [[inputs.aliyuncms.metrics]]
     names = ["SessionNewConnection"]
   allow_dps_without_discovery = false

  [[outputs.file]]
    files = ["stdout"]
```

got the output:
```
 ./telegraf --config confx
2021-04-21T17:24:41Z I! Starting Telegraf
2021-04-21T17:24:41Z I! Loaded inputs: aliyuncms
2021-04-21T17:24:41Z I! Loaded aggregators:
2021-04-21T17:24:41Z I! Loaded processors:
2021-04-21T17:24:41Z I! Loaded outputs: file
2021-04-21T17:24:41Z I! Tags enabled: host=xxx
2021-04-21T17:24:41Z I! [agent] Config: Interval:10s, Quiet:false, Hostname:"xxx", Flush Interval:10s
2021-04-21T17:24:41Z W! [inputs.aliyuncms] 'discovery_regions' is deprecated and ignored in favour of 'regions'.
2021-04-21T17:24:41Z E! [inputs.aliyuncms] Discovery tool is not activated: no discovery support for project "acs_nat_gateway"
aliyuncms_acs_nat_gateway,host=xxx,instanceId=nxw-gwxxxxxxxx,userId=11xxx session_new_connection_value=10 1619025600000000000
```

Indicating the plugin works for `acs_nat_gateway` and also the property `allow_dps_without_discovery = false` works

plugins/inputs/aliyuncms/aliyuncms.go
i-prudnikov: Fixed

plugins/inputs/aliyuncms/aliyuncms.go
i-prudnikov: agree

plugins/inputs/aliyuncms/aliyuncms.go
i-prudnikov: Agree, fixed

plugins/inputs/aliyuncms/aliyuncms.go
i-prudnikov: Agree, fixed

plugins/inputs/aliyuncms/discovery.go
i-prudnikov: Well, actually the discovery tool can be used as a separate object that provides discovery feature,  and hypothetically can  be pushed to a separate (from the plugin) package. 
If this smth. that is a kind of way too early, then we can revert  back to non exported type and channel.
However please not that there is also some methods of `DiscoveryTool` object that are exported (again keeping in mind this possibility).
Please do let me know your opinion, and I can push updated to this PR.

plugins/inputs/aliyuncms/aliyuncms.go
i-prudnikov: Actually I miss additional clause. What I was going to do is this:
```go
	if len(s.Regions) == 0 {
		s.Regions = s.DiscoveryRegions
	} else if len(s.DiscoveryRegions) != 0 {
		s.Log.Warn("'discovery_regions' is deprecated and ignored in favour of 'regions'.")
	}
```
Just to point out explicitly if both `regions` and `discovery_regions` specified, then `discovery_regions` is ignored in favour of `regions`

plugins/inputs/aliyuncms/aliyuncms.go
i-prudnikov: Agree, fixed

plugins/inputs/aliyuncms/aliyuncms.go
i-prudnikov: Please have a look on my comment regarding the `discoveryTool` object

plugins/inputs/aliyuncms/discovery.go
i-prudnikov: Please have a look on my comment regarding the `discoveryTool` object

plugins/inputs/aliyuncms/aliyuncms.go
vipinvkmenon: This message is a bit confusing. Considering that this plugin is yet to make it to the stable release and since `discovery_regions` is anyway deprecated from API side...could it be removed out altogether? Or it still serves a purpose?

i-prudnikov: @vipinvkmenon 

> Maybe the message could be added only if discoveryRegions & regions is configured?

This is exactly what I propose, have a look on full block that represents this logic:
```go
	//Check regions
	if len(s.DiscoveryRegions) != 0 {
		s.Log.Warn("'discovery_regions' is deprecated. Use 'regions' instead.")
	}

	if len(s.Regions) == 0 {
		s.Regions = s.DiscoveryRegions
	} else if len(s.DiscoveryRegions) != 0 {
		s.Log.Warn("'discovery_regions' is deprecated and ignored in favour of 'regions'.")
	}

	if len(s.Regions) == 0 {
		s.Regions = aliyunRegionList
		s.Log.Infof("'regions' is not set. Metrics will be queried across %d regions:\n%s",
			len(aliyunRegionList), strings.Join(aliyunRegionList, ","))
	}
```
So this line 
```go
s.Log.Warn("'discovery_regions' is deprecated and ignored in favour of 'regions'.")
```
Would be executed only in case both `regions` and `discovery_regions` set.

plugins/inputs/aliyuncms/aliyuncms.go
srebhan: Oh I think my comment was confusing. You really have to have two spaces from the line start as the initial indentation is also used for the string. You can see the difference when generating a config file with `/telegraf --input-filter aliyuncms --aggregator-filter " " --processor-filter " "  --output-filter file config > aliyun.conf`.

plugins/inputs/aliyuncms/aliyuncms.go
srebhan: Agreed but not changed. :-)

plugins/inputs/aliyuncms/discovery.go
srebhan: I prefer to not change it in this PR as it's kind of a scope creeping. Furthermore, if it will become a common thing to be used by other packages we should move it to `plugins/common`.

plugins/inputs/aliyuncms/aliyuncms.go
srebhan: @reimda, @ssoroka  are you ok with removing the `discovery_regions` option here as the plugin did not make it into any release yet IIRC?

plugins/inputs/aliyuncms/aliyuncms.go
srebhan: @i-prudnikov either I'm seeing some cached version of the file or you did not push your changes. For me it still says `Warnf`!?

plugins/inputs/aliyuncms/aliyuncms.go
srebhan: Still unchanged. I checked out via git and also this code is unchanged.

srebhan (commented):
It seems like the code is not yet reflecting the `resolved` state in the comment. Did you maybe forget to push your latest changes?

i-prudnikov:
> It seems like the code is not yet reflecting the `resolved` state in the comment. Did you maybe forget to push your latest changes?

Sorry, looks like I forgot to push changes. Pushed just now.

plugins/inputs/aliyuncms/discovery.go
i-prudnikov: Ok, noted, I will switch it to non-exported.

plugins/inputs/aliyuncms/aliyuncms.go
i-prudnikov: I will switch it to non-exported

plugins/inputs/aliyuncms/discovery.go
i-prudnikov: I will switch it to non-exported

plugins/inputs/aliyuncms/aliyuncms.go
i-prudnikov: Forgot to push changes, now should be fixed

plugins/inputs/aliyuncms/aliyuncms.go
i-prudnikov: Forgot to push changes, now should be fixed

plugins/inputs/aliyuncms/aliyuncms.go
i-prudnikov: Forgot to push changes, now should be fixed

plugins/inputs/aliyuncms/aliyuncms.go
i-prudnikov: Forgot to push changes, now should be fixed

plugins/inputs/aliyuncms/discovery.go
```diff
@@ -292,7 +277,8 @@ func NewDiscoveryTool(regions []string, project string, lg telegraf.Logger, cred
 	}, nil
 }
 
-func (dt *discoveryTool) parseDiscoveryResponse(resp *responses.CommonResponse) (discData []interface{}, totalCount int, pageSize int, pageNumber int, err error) {
+//func (dt *discoveryTool) parseDiscoveryResponse(resp *responses.CommonResponse) (data []interface{}, totalCount int, pageSize int, pageNumber int, err error) {
```
srebhan, line 280: Can you please remove the commented line.

plugins/inputs/aliyuncms/discovery.go
```diff
@@ -379,21 +369,21 @@ func (dt *discoveryTool) getDiscoveryData(cli aliyunSdkClient, req *requests.Com
...
 			}
-
 			return preparedData, nil
 		}
 	}
 }
 
-func (dt *discoveryTool) getDiscoveryDataAllRegions(limiter chan bool) (map[string]interface{}, error) {
+// GetDiscoveryDataAcrossRegions returns discovery data across defined (when creating) list of regions
+func (dt *discoveryTool) GetDiscoveryDataAcrossRegions(lmtr chan bool) (map[string]interface{}, error) {
```
srebhan, line 386: Please remove the commented line and unexport `GetDiscoveryDataAcrossRegions()`.

plugins/inputs/aliyuncms/discovery.go
```diff
@@ -482,14 +474,16 @@ func (dt *discoveryTool) Start() {
...
+// all data is read from 'DataChan'
 func (dt *discoveryTool) Stop() {
 	close(dt.done)
 
 	//Shutdown timer
 	timer := time.NewTimer(time.Second * 3)
 	defer timer.Stop()
 L:
-	for { //Unblock go routine by reading from dt.dataChan
+	for { //Unblock go routine by reading from dt.DataChan
```
srebhan, line 486: Please just keep the former. :-)

srebhan (commented):
Almost there. Just a few minor things, see comments in the code...

plugins/inputs/aliyuncms/discovery.go
i-prudnikov: For the sake of consistency I will switch all functions of `discoveryTool` object + `NewDiscoveryTool` function to be un-exported, as discussed.

i-prudnikov: Done

plugins/inputs/aliyuncms/discovery.go
i-prudnikov: done

plugins/inputs/aliyuncms/aliyuncms.go
reimda: It's ok to remove discovery_regions.  I just double checked that aliyuncms didn't make it into 1.18.  Thanks @i-prudnikov for keeping backward compatibility in mind.  It's fortunate that we don't need it this time.

plugins/inputs/aliyuncms/discovery.go
i-prudnikov: Done

plugins/inputs/aliyuncms/aliyuncms.go
i-prudnikov: Ok, `discovery_regions` removed.

srebhan: Thanks @i-prudnikov and @reimda!

srebhan (approved):
Code looks good. @i-prudnikov the bot thinks you are missing the CLA, if you already signed it, please write a comment with ` !signed-cla`. Let me know if there are problems...

i-prudnikov:
> Code looks good. @i-prudnikov the bot thinks you are missing the CLA, if you already signed it, please write a comment with ` !signed-cla`. Let me know if there are problems...

Hello, I've already put the comment with `!signed-cla`, here it is once again:  !signed-cla

i-prudnikov:
> Thanks so much for the pull request!
> 🤝 ✒️ Just a reminder that the CLA has not yet been signed, and we'll need it before merging. Please [sign the CLA](https://www.influxdata.com/legal/cla/) when you get a chance, then post a comment here saying `!signed-cla`

` !signed-cla`

sspaink:
!signed-cla

sspaink:
@i-prudnikov sorry about the CLA bot confusion, there was an issue that should be resolved now.

i-prudnikov:
> @i-prudnikov sorry about the CLA bot confusion, there was an issue that should be resolved now.

Sure, NP!
@sspaink,  @reimda, @srebhan, sorry for  the off-topic, but could somebody review this:  https://github.com/influxdata/telegraf/pull/7749

reimda:
> @sspaink, @reimda, @srebhan, sorry for the off-topic, but could somebody review this: #7749

Sure, we'll take a look.  thanks!

vipinvkmenon:
@i-prudnikov in aliyuncms `acs_ecs_dashboard` etc are called Namespace but in the plugin, you've called it Project. Any reason for the same? Maybe some documentation to understand this.

i-prudnikov:
> @i-prudnikov in aliyuncms `acs_ecs_dashboard` etc are called Namespace but in the plugin, you've called it Project. Any reason for the same? Maybe some documentation to understand this.
@vipinvkmenon , I think that the source of confusion is this page: https://help.aliyun.com/document_detail/28619.html?spm=a2c4g.11186623.6.690.1938ad41wg8QSq  - here they use term `project`, but in the API spec they are using term `namespace`.

srebhan:
@vipinvkmenon can we please stay on-topic here! ;-P There is a slack channel to discuss this kind of things... :-)

vipinvkmenon:
From the test perspective the PR works :)

sspaink (approved):
lgtm! @i-prudnikov once you've resolved the conflicts, I will merge this pr. Thanks!

PR #9159, opened 2021-04-20
Add support for gzip compression to cloud_pubsub input and output

### LINKED ISSUE #9157: Google Cloud Pubsub: allows messages to be gzipped for sending and receiving
## Feature Request

When using cloud_pubsub input & output plugins it would be great if these plugin could compress/decompress the message body.

Current behavior:
All messages to Pubsub are sent without any compression.

Desired behavior:
Add a content encoding option for message payloads, can be set to "gzip" or  "identity" to apply no encoding.

### Use case:
Currently we send quite a lot of data over pubsub and would love to be able to gzip the message data to help reduce cost. Also currently there is no option in pubsub client to gzip and the suggested approach is to gzip before sending the messages. https://github.com/googleapis/google-cloud-go/issues/1546#issuecomment-804352269

nadeem-syed:
### Required for all PRs:
- [ X] Updated associated README.md.
- [ X] Wrote appropriate unit tests.

resolves #9157

### DIFF
--- a/plugins/inputs/cloud_pubsub/pubsub.go
+++ b/plugins/inputs/cloud_pubsub/pubsub.go
@@ -44,7 +44,8 @@ type PubSub struct {
 
 	Base64Data bool `toml:"base64_data"`
 
-	Log telegraf.Logger
+	Log             telegraf.Logger
+	ContentEncoding string `toml:"content_encoding"`
 
 	sub     subscription
 	stubSub func() subscription
@@ -57,6 +58,7 @@ type PubSub struct {
 
 	undelivered map[telegraf.TrackingID]message
 	sem         semaphore
+	decoder     internal.ContentDecoder
 }
 
 func (ps *PubSub) Description() string {
@@ -88,7 +90,11 @@ func (ps *PubSub) Start(ac telegraf.Accumulator) error {
 	if ps.Project == "" {
 		return fmt.Errorf(`"project" is required`)
 	}
-
+	var err error
+	ps.decoder, err = internal.NewContentDecoder(ps.ContentEncoding)
+	if err != nil {
+		return err
+	}
 	ps.sem = make(semaphore, ps.MaxUndeliveredMessages)
 	ps.acc = ac.WithTracking(ps.MaxUndeliveredMessages)
 
@@ -175,15 +181,18 @@ func (ps *PubSub) onMessage(ctx context.Context, msg message) error {
 		return fmt.Errorf("message longer than max_message_len (%d > %d)", len(msg.Data()), ps.MaxMessageLen)
 	}
 
-	var data []byte
+	var data = msg.Data()
+	data, err := ps.decoder.Decode(data)
+	if err != nil {
+		return fmt.Errorf("unable to decode message: %v", err)
+	}
+
 	if ps.Base64Data {
-		strData, err := base64.StdEncoding.DecodeString(string(msg.Data()))
+		strData, err := base64.StdEncoding.DecodeString(string(data))
 		if err != nil {
 			return fmt.Errorf("unable to base64 decode message: %v", err)
 		}
 		data = strData
-	} else {
-		data = msg.Data()
 	}
 
 	metrics, err := ps.parser.Parse(data)
@@ -366,4 +375,8 @@ const sampleConfig = `
   ## Optional. If true, Telegraf will attempt to base64 decode the
   ## PubSub message data before parsing
   # base64_data = false
+
+  ## Content encoding for message payloads, can be set to "gzip" or
+  ## "identity" to apply no encoding.
+  # content_encoding = "identity"
 `
--- a/plugins/outputs/cloud_pubsub/pubsub.go
+++ b/plugins/outputs/cloud_pubsub/pubsub.go
@@ -62,6 +62,14 @@ const sampleConfig = `
   ## Optional. If true, published PubSub message data will be base64-encoded.
   # base64_data = false
 
+  ## Content encoding for message payloads, can be set to "gzip" or
+  ## "identity" to apply no encoding.
+
+  ## Please note that when send_batched = false each pubsub message contains only
+  ## a single metric, it is recommended to use compression with batch format
+  ## for best results.
+  # content_encoding = "identity"
+
   ## Optional. PubSub attributes to add to metrics.
   # [outputs.cloud_pubsub.attributes]
   #   my_attr = "tag_value"
@@ -79,6 +87,7 @@ type PubSub struct {
 	PublishNumGoroutines  int             `toml:"publish_num_go_routines"`
 	PublishTimeout        config.Duration `toml:"publish_timeout"`
 	Base64Data            bool            `toml:"base64_data"`
+	ContentEncoding       string          `toml:"content_encoding"`
 
 	Log telegraf.Logger `toml:"-"`
 
@@ -89,6 +98,7 @@ type PubSub struct {
 
 	serializer     serializers.Serializer
 	publishResults []publishResult
+	encoder        internal.ContentEncoder
 }
 
 func (ps *PubSub) Description() string {
@@ -114,6 +124,12 @@ func (ps *PubSub) Connect() error {
 
 	if ps.stubTopic == nil {
 		return ps.initPubSubClient()
+	}
+
+	var err error
+	ps.encoder, err = internal.NewContentEncoder(ps.ContentEncoding)
+	if err != nil {
+		return err
 	}
 	return nil
 }
@@ -220,6 +236,10 @@ func (ps *PubSub) toMessages(metrics []telegraf.Metric) ([]*pubsub.Message, erro
 			encoded := base64.StdEncoding.EncodeToString(b)
 			b = []byte(encoded)
 		}
+		b, err = ps.encoder.Encode(b)
+		if err != nil {
+			return nil, err
+		}
 
 		msg := &pubsub.Message{Data: b}
 		if ps.Attributes != nil {
@@ -241,8 +261,15 @@ func (ps *PubSub) toMessages(metrics []telegraf.Metric) ([]*pubsub.Message, erro
 			b = []byte(encoded)
 		}
 
+		b, err = ps.encoder.Encode(b)
+		if err != nil {
+			ps.Log.Debugf("could not encode metric: %v", err)
+			continue
+		}
+		data := make([]byte, len(b))
+		copy(data, b)
 		msgs[i] = &pubsub.Message{
-			Data: b,
+			Data: data,
 		}
 		if ps.Attributes != nil {
 			msgs[i].Attributes = ps.Attributes
--- a/plugins/outputs/cloud_pubsub/topic_stubbed.go
+++ b/plugins/outputs/cloud_pubsub/topic_stubbed.go
@@ -14,6 +14,7 @@ import (
 	"cloud.google.com/go/pubsub"
 	"github.com/influxdata/telegraf"
 	"github.com/influxdata/telegraf/config"
+	"github.com/influxdata/telegraf/internal"
 	"github.com/influxdata/telegraf/plugins/parsers"
 	"github.com/influxdata/telegraf/plugins/serializers"
 	"google.golang.org/api/support/bundler"
@@ -53,9 +54,11 @@ type (
 
 		published map[string]*pubsub.Message
 
-		bundler     *bundler.Bundler
-		bLock       sync.Mutex
-		bundleCount int
+		bundler         *bundler.Bundler
+		bLock           sync.Mutex
+		bundleCount     int
+		Base64Data      bool
+		ContentEncoding string
 	}
 )
 
@@ -64,9 +67,10 @@ func getTestResources(tT *testing.T, settings pubsub.PublishSettings, testM []te
 
 	metrics := make([]telegraf.Metric, len(testM))
 	t := &stubTopic{
-		T:         tT,
-		ReturnErr: make(map[string]bool),
-		published: make(map[string]*pubsub.Message),
+		T:               tT,
+		ReturnErr:       make(map[string]bool),
+		published:       make(map[string]*pubsub.Message),
+		ContentEncoding: "identity",
 	}
 
 	for i, tm := range testM {
@@ -85,7 +89,9 @@ func getTestResources(tT *testing.T, settings pubsub.PublishSettings, testM []te
 		PublishByteThreshold:  settings.ByteThreshold,
 		PublishNumGoroutines:  settings.NumGoroutines,
 		PublishTimeout:        config.Duration(settings.Timeout),
+		ContentEncoding:       "identity",
 	}
+	ps.encoder, _ = internal.NewContentEncoder(ps.ContentEncoding)
 	ps.SetSerializer(s)
 
 	return ps, t, metrics
@@ -179,19 +185,23 @@ func (t *stubTopic) sendBundle() func(items interface{}) {
 
 func (t *stubTopic) parseIDs(msg *pubsub.Message) []string {
 	p, _ := parsers.NewInfluxParser()
-	metrics, err := p.Parse(msg.Data)
+	d := msg.Data
+	decoder, _ := internal.NewContentDecoder(t.ContentEncoding)
+	d, err := decoder.Decode(d)
 	if err != nil {
-		// Just attempt to base64-decode first before returning error.
-		d, err := base64.StdEncoding.DecodeString(string(msg.Data))
-		if err != nil {
-			t.Errorf("unable to base64-decode potential test message: %v", err)
-		}
-		metrics, err = p.Parse(d)
-		if err != nil {
-			t.Fatalf("unexpected parsing error: %v", err)
-		}
+		t.Errorf("unable to decode message: %v", err)
 	}
-
+	if t.Base64Data {
+		strData, err := base64.StdEncoding.DecodeString(string(d))
+		if err != nil {
+			t.Errorf("unable to base64 decode message: %v", err)
+		}
+		d = strData
+	}
+	metrics, err := p.Parse(d)
+	if err != nil {
+		t.Fatalf("unexpected parsing error: %v", err)
+	}
 	ids := make([]string, len(metrics))
 	for i, met := range metrics {
 		id, _ := met.GetField("value")
--- a/plugins/inputs/cloud_pubsub/README.md
+++ b/plugins/inputs/cloud_pubsub/README.md
@@ -80,6 +80,10 @@ and creates metrics using one of the supported [input data formats][].
   ## PubSub message data before parsing. Many GCP services that
   ## output JSON to Google PubSub base64-encode the JSON payload.
   # base64_data = false
+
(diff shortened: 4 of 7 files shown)


plugins/inputs/cloud_pubsub/pubsub.go
```diff
@@ -44,7 +44,8 @@ type PubSub struct {
 
 	Base64Data bool `toml:"base64_data"`
 
-	Log telegraf.Logger
+	Log             telegraf.Logger
+	ContentEncoding string `toml:"content_encoding"`
```
Jagularr, line 49: Will this have / does it need a default value?

nadeem-syed: I think empty string is the default. https://github.com/influxdata/telegraf/blob/master/internal/content_coding.go#L17

Jagularr: Cool. gotcha.

Jagularr (approved):
Makes sense to me.

plugins/inputs/cloud_pubsub/pubsub.go
```diff
@@ -44,7 +44,8 @@ type PubSub struct {
...
```
srebhan, line 48: Only cosmetics, but if you don't mind to swap the two lines and add a toml tag to `Log` such as
```suggestion
	ContentEncoding string `toml:"content_encoding"`
	Log             telegraf.Logger `toml:"-"`
```
to avoid `Log` being set by the toml parser by accident.

srebhan (approved):
Looks good to me. Only one cosmetic comment, but it's not a must...

nadeem-syed:
I am getting a panic while testing under heavy production load :( 
```
panic: runtime error: slice bounds out of range [31556:0]

goroutine 111 [running]:
compress/flate.(*decompressor).Read(0xc0001bb900, 0xc006d10000, 0x4fe00, 0x4fe00, 0x51bb6, 0xe7b3c2a2, 0x0)
	/usr/local/go/src/compress/flate/inflate.go:338 +0x24f
compress/gzip.(*Reader).Read(0xc000156b00, 0xc006d10000, 0x4fe00, 0x4fe00, 0x1db6, 0x0, 0x0)
	/usr/local/go/src/compress/gzip/gunzip.go:251 +0x87
bytes.(*Buffer).ReadFrom(0xc0003f55f0, 0x423dde0, 0xc000156b00, 0x0, 0x0, 0x0)
	/usr/local/go/src/bytes/buffer.go:204 +0xbe
github.com/influxdata/telegraf/internal.(*GzipDecoder).Decode(0xc00022a950, 0xc0077a8000, 0x94f2, 0xa000, 0x0, 0x0, 0x0, 0x0, 0x0)
	/go/src/github.com/influxdata/telegraf/internal/content_coding.go:162 +0xbd
github.com/influxdata/telegraf/plugins/inputs/cloud_pubsub.(*PubSub).onMessage(0xc0002fc100, 0x42b8890, 0xc0005acdc0, 0x42cab28, 0xc000143380, 0x0, 0x0)
	/go/src/github.com/influxdata/telegraf/plugins/inputs/cloud_pubsub/pubsub.go:185 +0xaa
github.com/influxdata/telegraf/plugins/inputs/cloud_pubsub.(*PubSub).startReceiver.func1(0x42b8890, 0xc0005acdc0, 0x42cab28, 0xc000143380)
	/go/src/github.com/influxdata/telegraf/plugins/inputs/cloud_pubsub/pubsub.go:164 +0x67
github.com/influxdata/telegraf/plugins/inputs/cloud_pubsub.(*gcpSubscription).Receive.func1(0x42b8890, 0xc0005acdc0, 0xc000196f30)
	/go/src/github.com/influxdata/telegraf/plugins/inputs/cloud_pubsub/subscription_gcp.go:42 +0x7e
cloud.google.com/go/pubsub.(*Subscription).receive.func3(0xc0006a0930, 0xc00022aeb0, 0x42b8890, 0xc0005acdc0, 0xc000196f30)
	/go/pkg/mod/cloud.google.com/go/pubsub@v1.1.0/subscription.go:729 +0x70
created by cloud.google.com/go/pubsub.(*Subscription).receive
	/go/pkg/mod/cloud.google.com/go/pubsub@v1.1.0/subscription.go:727 +0x1fe
	```
	So I would like to hold this off, till I figure out what is causing this panic.

**Comment by srebhan** [comment]
Ok seems that 
[with this patch](https://github.com/influxdata/telegraf/files/6364113/fix_cloud_pubsub_decode_race.patch.zip) we can at least fix the panic... The problem is that the `onMessage()` callback can be called concurrently while the decoder cannot. It furthermore fixes the "some messages are neither ack'ed nor nack'ed" problem when returning early.

**Comment by nadeem-syed** [comment]
We are currently running this patch in our prod and saw approximately 90% reduction in our pubsub cost. 
It dropped from 25.11 MiB/s to 2.83 MiB/s, hence 💰 💸 

image

File: internal/content_coding.go [hunk 3]
```diff
@@ -166,7 +168,9 @@ func (d *GzipDecoder) Decode(data []byte) ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
-	return d.buf.Bytes(), nil
+	out := make([]byte, d.buf.Len())
+	copy(out, d.buf.Bytes())
```
srebhan, line 172: TBH I don't like this to be in the encoder/decoder as it now adds a copy operation for everyone even though we might not use the encoder/decoder in parallel... Can you please move the copy operation to your plugin's code!

srebhan: Let me take this to the core-devs to discuss this. Will report back here...

srebhan:
@nadeem-syed thanks for your patience. I asked the core-devs and they will give your PR another review regarding the buffer copy topic. It would be nice if you can fix the formatting in the meantime to make CI happy. :-)

plugins/inputs/cloud_pubsub/pubsub.go
```diff
@@ -170,30 +177,32 @@ func (ps *PubSub) startReceiver(parentCtx context.Context) error {
...
 	if ps.MaxMessageLen > 0 && len(msg.Data()) > ps.MaxMessageLen {
-		msg.Ack()
 		return fmt.Errorf("message longer than max_message_len (%d > %d)", len(msg.Data()), ps.MaxMessageLen)
 	}
 
-	var data []byte
+	// This function is called concurrently, but the decoder cannot.
+	ps.decoderMutex.Lock()
+	data, err := ps.decoder.Decode(msg.Data())
+	ps.decoderMutex.Unlock()
```
reimda, line 188: I took a look at this and I agree with @srebhan that it would be better not to modify the shared code in content_coding.go. Could you copy the buffer here instead of in the decoder function itself?

reimda (changes requested):

plugins/inputs/cloud_pubsub/pubsub.go
nadeem-syed: Sounds good. I have moved the copy buffer code to just the pubsub plugins.

internal/content_coding.go
nadeem-syed: Moved the copy operation to pubsub plugin's code https://github.com/influxdata/telegraf/pull/9159/commits/809f650f099438987792b495e0c8c3bd4426130d

plugins/inputs/cloud_pubsub/pubsub.go
```diff
@@ -170,30 +177,39 @@ func (ps *PubSub) startReceiver(parentCtx context.Context) error {
...
+	if err != nil {
+		return fmt.Errorf("unable to decode message: %v", err)
+	}
 	var data []byte
+	if ps.ContentEncoding == "gzip" {
+		data = make([]byte, len(b))
+		copy(data, b)
+	} else {
+		data = b
+	}
```
srebhan, line 198: Uhh I think you have to do this while holding the `ps.decoderMutex` lock, as otherwise writing to `b` might be racy.

plugins/outputs/cloud_pubsub/pubsub.go
```diff
@@ -220,8 +237,18 @@ func (ps *PubSub) toMessages(metrics []telegraf.Metric) ([]*pubsub.Message, erro
...
+		if err != nil {
+			return nil, err
+		}
+		var data []byte
+		if ps.ContentEncoding == "gzip" {
+			data = make([]byte, len(b))
+			copy(data, b)
+		} else {
+			data = b
+		}
```
srebhan, line 250: IIRC `Write()` is not called concurrently, so `toMessages()` isn't either. Therefore, I think you do not need to copy the encoded buffer here.

plugins/outputs/cloud_pubsub/pubsub.go
```diff
@@ -241,8 +268,20 @@ func (ps *PubSub) toMessages(metrics []telegraf.Metric) ([]*pubsub.Message, erro
...
+			ps.Log.Debugf("could not encode metric: %v", err)
+			continue
+		}
+		var data []byte
+		if ps.ContentEncoding == "gzip" {
+			data = make([]byte, len(b))
+			copy(data, b)
+		} else {
+			data = b
+		}
```
srebhan, line 282: See above, I think you don't need to copy here.

srebhan (changes requested):
@nadeem-syed thanks for your quick reaction. Seems like the copy logic is not yet right.

1. For the input plugin, you need to copy the buffer (as you do) but you need to hold the decoder mutex while doing so. Otherwise, the copy operation and the decoding might race for accessing the buffer.
2. For the output plugin, you don't need to copy the encoded data at all IMO, as `Write()` is not run concurrently. This way, the encoder is also not called concurrently making the copy superfluous.

This is really a tough one... :-)

plugins/inputs/cloud_pubsub/pubsub.go
```diff
@@ -170,30 +177,39 @@ func (ps *PubSub) startReceiver(parentCtx context.Context) error {
 
 // onMessage handles parsing and adding a received message to the accumulator.
 func (ps *PubSub) onMessage(ctx context.Context, msg message) error {
+	defer msg.Ack()
```
ssoroka, line 180: `defer msg.Ack()` is not correct. This will run the ack no matter what exit path is taken from the function, but you instead want to be selective, only `ack`ing when the message has been properly handled, or is incorrectly formatted and you will not be able to ever process it correctly. Specifically when the context is cancelled you don't want the message to be `ack`ed

plugins/inputs/cloud_pubsub/pubsub.go
ssoroka: this is not an issue

plugins/inputs/cloud_pubsub/pubsub.go
```diff
@@ -170,30 +177,39 @@ func (ps *PubSub) startReceiver(parentCtx context.Context) error {
...
+	ps.decoderMutex.Lock()
+	b, err := ps.decoder.Decode(msg.Data())
+	ps.decoderMutex.Unlock()
+	if err != nil {
+		return fmt.Errorf("unable to decode message: %v", err)
+	}
 	var data []byte
+	if ps.ContentEncoding == "gzip" {
+		data = make([]byte, len(b))
+		copy(data, b)
```
ssoroka, line 195: why does gzip need a copy of this?
seems like this whole conditional could just be `data = b`

ssoroka (changes requested):

sjwang90:
@nadeem-syed could you address the comments from above? We would love to get this reviewed and merged

nadeem-syed:
> @nadeem-syed could you address the comments from above? We would love to get this reviewed and merged

Sure. I will take a look soon.

plugins/outputs/cloud_pubsub/pubsub.go
nadeem-syed: IMHO the copy is needed because GzipDecoder [Decode ](https://github.com/influxdata/telegraf/blob/master/internal/content_coding.go#L169) is returning [buffer.Bytes() ](https://pkg.go.dev/bytes#Buffer.Bytes)which is a [slice](https://cs.opensource.google/go/go/+/refs/tags/go1.16.6:src/bytes/buffer.go;l=54) and we are appending [here](https://github.com/influxdata/telegraf/blob/master/plugins/outputs/cloud_pubsub/pubsub.go#L244-L245) and this causes erroneous result.  I have added [couple of tests](https://github.com/influxdata/telegraf/pull/9159/files#diff-c333930dc7e05efd47fb16f1fc53057bc5f9407e36851f5692accd06729e5e98R168-R223) to validate that. If we remove the copy those tests starts failing

plugins/inputs/cloud_pubsub/pubsub.go
nadeem-syed: Sounds good. I have reverted all the msg.Ack() changes.

plugins/inputs/cloud_pubsub/pubsub.go
nadeem-syed: IMHO the copy is needed because GzipEncoder [Encode ](https://github.com/influxdata/telegraf/blob/master/internal/content_coding.go#L125) is returning [buffer.Bytes() ](https://pkg.go.dev/bytes#Buffer.Bytes)which is a [slice](https://cs.opensource.google/go/go/+/refs/tags/go1.16.6:src/bytes/buffer.go;l=54) and if we multiple goroutines are modifying the underlying buffer, it starts acting weirdly due to race condition. 
We were initially not doing the copy and started getting bad data in our influxDB and also these errors in telegraf 
```
2021-07-28T10:14:18Z E! [inputs.cloud_pubsub] Error in plugin: unable to add message from subscription metrics-platform: unable to parse decoded message: metric parse error: expected field at 309:327: "luster_id=140046,container_image=us.gcr.io/wp-engine-development/memcached,container_name=memcached,container_version=24,cpu=cpu1,datacenter=gce-us-east1,host=host1,innodb_buffer_reduce=0,offering=shared,platform=classic,region=us-east1,server_version=20.10.7,ws_external_hats=False usage_total=3510062015i 1627467241000000000"
2021-07-28T10:14:18Z E! [inputs.cloud_pubsub] Error in plugin: unable to add message from subscription metrics-platform: unable to parse decoded message: metric parse error: expected tag at 507:177: ""
```
and this resulted in the following
error
 BTW I also saw we handle the buffer in similar way in [Influx serialization code](https://github.com/influxdata/telegraf/blob/1ea5c20aadc55863ce1acd49051e6d6db03b4d60/plugins/serializers/influx/influx.go#L105-L106). 
 By using similar code like above solved our problem in production.

plugins/outputs/cloud_pubsub/topic_stubbed.go
```diff
@@ -179,19 +185,23 @@ func (t *stubTopic) sendBundle() func(items interface{}) {
...
 	if err != nil {
-		// Just attempt to base64-decode first before returning error.
-		d, err := base64.StdEncoding.DecodeString(string(msg.Data))
-		if err != nil {
-			t.Errorf("unable to base64-decode potential test message: %v", err)
-		}
-		metrics, err = p.Parse(d)
+		t.Errorf("unable to decode message: %v", err)
+	}
+	if t.Base64Data {
```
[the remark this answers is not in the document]
nadeem-syed: I think base64 encoding decoding is done outside of decoder https://github.com/influxdata/telegraf/blob/master/plugins/inputs/cloud_pubsub/pubsub.go#L180

reimda (approved):

plugins/inputs/cloud_pubsub/pubsub.go
```diff
@@ -175,21 +182,33 @@ func (ps *PubSub) onMessage(ctx context.Context, msg message) error {
...
 			return fmt.Errorf("unable to base64 decode message: %v", err)
 		}
 		data = strData
-	} else {
-		data = msg.Data()
 	}
 
 	metrics, err := ps.parser.Parse(data)
 	if err != nil {
 		msg.Ack()
```
srebhan, line 210: I think we cannot ACK messages that fail. You should use `msg.Nack()`.

plugins/outputs/cloud_pubsub/pubsub.go
srebhan: Well but you are `Encode` here, don't you? Anyway, if you are sure you need the copy, please feel free to close this conversation as resolved.

srebhan (commented):
Hey @nadeem-syed thanks for working on this PR! The only comment I have is the `ACK` of the messages. Please check again to only `ACK` messages that were handled **successfully** and `NAK` messages with errors.

plugins/inputs/cloud_pubsub/pubsub.go
nadeem-syed: I think this an existing behavior. https://github.com/influxdata/telegraf/blob/master/plugins/inputs/cloud_pubsub/pubsub.go#L191

I think the reason for this behavior is that is if by mistake there is some bad unParsable data enter the pub-sub topic, then we don't want to keep on retrying. Hence we ACK the message and return an error

plugins/outputs/cloud_pubsub/pubsub.go
nadeem-syed: Oops I do mean to say GzipEncoder [Encode ](https://github.com/influxdata/telegraf/blob/master/internal/content_coding.go#L125) is returning [buffer.Bytes() ](https://pkg.go.dev/bytes#Buffer.Bytes)which is a [slice](https://cs.opensource.google/go/go/+/refs/tags/go1.16.6:src/bytes/buffer.go;l=54) and we are appending [here](https://github.com/influxdata/telegraf/blob/master/plugins/outputs/cloud_pubsub/pubsub.go#L244-L245) and this causes erroneous result.

plugins/inputs/cloud_pubsub/pubsub.go
ssoroka: I think we're moving away from copies in that code as well. If you can find a way to make it goroutine-safe without the copy, it will be significantly faster.

ssoroka: I'm concerned there's a more fundamental issue here. I think what I'd prefer to see is the whole oNMessage function with a mutex, then removing the decode mutex and removing the data copy. You should see all concurrency issues go away.

ssoroka (changes requested):

plugins/outputs/cloud_pubsub/pubsub.go
```diff
@@ -62,6 +62,14 @@ const sampleConfig = `
   ## Optional. If true, published PubSub message data will be base64-encoded.
   # base64_data = false
 
+  ## Content encoding for message payloads, can be set to "gzip" or
+  ## "identity" to apply no encoding.
+
+  ## Please note that when send_batched = false each pubsub message contains only
+  ## a single metric, it is recommended to use compression with batch format
+  ## for best results.
+  # content_encoding = "identity"
```
srebhan, line 71: Please sync this with the `plugins/outputs/cloud_pubsub/README.md`!

srebhan (commented):
Hey @nadeem-syed, thanks for the work here and sorry for my late response. I only have one more comment about syncing the config-option description in the output plugin with the README. Can you please take a look?

PR #9160, opened 2021-04-20
Converter processor: add support for large hexadecimal strings

### LINKED ISSUE #9147: [[processors.converter]] bug with hex-encoded strings
Hi there!
I'm facing a strange issue with `[[processors.converter]]` plugin. In my config I have several http inputs that return some hex-encoded strings. The issue is that for some reasons converter refuses to convert one of these strings to integer. In the same time other strings are converted properly

### Relevant telegraf.conf:
```toml
[[inputs.http]]
  urls = ["http://localhost:9650/ext/bc/C/rpc"]
  method = "POST"
  body = '{"jsonrpc":"2.0","method":"eth_blockNumber","params":[],"id":2}'
  headers = {"Content-Type" = "application/json"}
  data_format = "json"
  json_string_fields = ["result"]
  [inputs.http.tags]
    api_method = "eth_blocknumber"
[[inputs.http]]
  urls = ["http://localhost:9650/ext/bc/C/rpc"]
  method = "POST"
  body = '{"jsonrpc":"2.0","method":"eth_getBalance","params":["***************************", "latest"],"id":3}'
  headers = {"Content-Type" = "application/json"}
  data_format = "json"
  json_string_fields = ["result"]
  [inputs.http.tags]
    api_method = "eth_getbalance"
[[processors.converter]]
  [processors.converter.fields]
    integer = ["result"] 
```

### System info:
System: Ubuntu 18.04
Telegraf 1.18.1 (git: HEAD d3ac9a3f)
### Expected behavior:
```
sudo telegraf --config /etc/telegraf/telegraf.conf --test
2021-04-19T12:36:29Z I! Starting Telegraf 1.18.1
> http,api_method=eth_getbalance,host=avalanche-prod-node,url=http://localhost:9650/ext/bc/C/rpc id=3,result=612908836750534700000i 1618835790000000000
> http,api_method=eth_blocknumber,host=avalanche-prod-node,url=http://localhost:9650/ext/bc/C/rpc id=2,result=1147307i 1618835790000000000
```
### Actual behavior:
```
sudo telegraf --config /etc/telegraf/telegraf.conf --test
2021-04-19T12:36:29Z I! Starting Telegraf 1.18.1
> http,api_method=eth_getbalance,host=avalanche-prod-node,url=http://localhost:9650/ext/bc/C/rpc id=3 1618835790000000000
> http,api_method=eth_blocknumber,host=avalanche-prod-node,url=http://localhost:9650/ext/bc/C/rpc id=2,result=1147307i 1618835790000000000
```

and the following error in logs:
```
2021-04-19T12:34:30Z E! [processors.converter] error converting to integer [string]: 0x2139d19bb1c580ebe0
```

### Additional information
The same http requests performed with curl:
```
>curl -X POST http://localhost:9650/ext/bc/C/rpc -H "Content-type: application/json" -d '{"jsonrpc":"2.0","method":"eth_blockNumber","params":[],"id":3}'
>{"jsonrpc":"2.0","id":3,"result":"0x11826c"}
```
```
>curl -X POST http://localhost:9650/ext/bc/C/rpc -H "Content-type: application/json" -d '{"jsonrpc":"2.0","method":"eth_getBalance","params":["***************************", "latest"],"id":3}'
>{"jsonrpc":"2.0","id":3,"result":"0x2139d19bb1c580ebe0"}
```

At the same time on my another instance telegraf successfully converts `0x6e7f41d403fd0609` string to integer with the same config

Jagularr:
Large hexadecimal numbers don't seem to work right with the current `ParseFloat` method. I'm adding this functionality so that any size hexadecimal string can be converted to a float64.

Keep in mind that if the string is large, you should use the 'float' conversion and not an 'integer' conversion, as it will likely overflow and return the max int64 value.

resolves #9147


plugins/processors/converter/converter.go
```diff
@@ -452,6 +478,24 @@ func toString(v interface{}) (string, bool) {
...
+func parseHexadecimal(value string) (float64, error) {
+	i := new(big.Int)
+
+	_, success := i.SetString(value, 0)
+	if !success {
+		return 0, errors.New("unable to parse string to big int")
+	}
+
+	f := new(big.Float).SetInt(i)
+	result, _ := f.Float64()
```
reimda, line 490: I'm not sure returning the big hex number as a float is a good idea.  It will work for hex numbers a little larger than int64, but for bigger hex humbers it will hit the limit of a float64 and start losing precision.  Hex is used when data needs to be bit precise.  Any lossy format will cause trouble for someone.

reimda (changes requested):

plugins/processors/converter/converter.go
Jagularr: Thanks for reviewing. As I don't believe we can have a metric with a larger number than float64, wouldn't we encounter this problem regardless?

reimda: I don't think I understood what this PR is trying to do when I commented earlier.  If the user has a very big hex number and sets up a processor to convert it to a float, they should know that floats by nature have precision limitations.  They should expect the conversion to lose precision.

Maybe all we need to do is add a warning to the documentation so all users will be aware that this functionality exists and that it can be a lossy conversion.

reimda (approved):
Looks good

PR #9168, opened 2021-04-21
Private linter for Telegraf

### LINKED ISSUE #9019: Custom linter for golangci-lint to enforce review criteria
Create a private linter that can be run by golangci-lint that will enforce the rules defined in this wiki: https://github.com/influxdata/telegraf/wiki/Review

The scope of this issue will cover adding the structure of the private linter and a single rule as an example, future issues will be created to expand the features of the linter to cover more rules in the review document. The CI will need to be updated to build this linter and cache it.

Documentation how to add private linter: https://golangci-lint.run/contributing/new-linters/#how-to-add-a-private-linter-to-golangci-lint

sspaink:
Closes #9019

This pull request introduces a private linter for Telegraf that checks if a plugin is using the "log" package. I am hoping with this pull request to lay the foundation of the private linter so it can be extended with future pull requests.

In order to use private linters with golanci-lint, CGO is required to be enabled. There is no official binary at the moment with it enabled and the github action linter we are currently using doesn't support it either. Therefore to use this linter you need to follow some steps to run it locally, I've written out these steps in the README.md I added in this pull request.  

My plan is to address adding this linter to the CI in a separate PR, there are two approaches I can think of:
* Update the github action to allow new download URLS (host our own binary): https://github.com/golangci/golangci-lint-action/issues/201
*  Move away from the official github action and make our own solution

Here is a screenshot of part of the output you can expect when you run this linter:
image


Jagularr (approved):
Makes a lot of sense

PR #9178, opened 2021-04-23
Fixes #9177 - Specify wanted ipmi metrics

### LINKED ISSUE #9177: Make it possible to gather specific IPMI sensor data
### Proposal:
I'd like to see it be possible to gather specific IPMI sensor data (by using https://github.com/influxdata/telegraf/tree/master/plugins/inputs/ipmi_sensor). At the moment it collects all sensor data. In collectd this option is possible

### Current behavior:
The ipmi_sensor input plugin collects all sensor data.

### Desired behavior:
The ipmi_sensor input plugin is able to collect specific sensor data if specified.

### Use case: 
For our use case we collect just the inlet and exhaust temperatures of our servers and CPUs. We do not require other data.

Rubueno:
This change makes it possible to store specific ipmi metrics to the
ouptut.

### Required for all PRs:

resolves #9177

Added an option to specify which ipmi metrics one wishes to collect and store. Similar to collectd's ipmi plugin: https://collectd.org/wiki/index.php/Plugin:IPMI

### DIFF
--- a/plugins/inputs/ipmi_sensor/README.md
+++ b/plugins/inputs/ipmi_sensor/README.md
@@ -70,6 +70,10 @@ Any of the following parameters will be added to the aformentioned query if they
   ## Path to the ipmitools cache file (defaults to OS temp dir)
   ## The provided path must exist and must be writable
   # cache_path = ""
+
+  ## Store specific IPMI sensor data
+  ## If no value is specified all sensor data will be stored
+  # wantedmetrics = [ "inlet_temp" ]
 ```
 
 ### Measurements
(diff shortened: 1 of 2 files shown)


Rubueno:
!signed-cla

Rubueno:
This is my first ever written bit of Golang. I'm not sure how to properly manipulate the tests.

Jagularr (commented):
This looks like sound logic, but I believe you can already accomplish this (matching on specific tag values) with `tagpass`. With the following config line under the input you could filter only on the tags that you specify:

```
tagpass = { name = ["tag1", "tag2"] }
```

Rubueno:
@Jagularr Thank you for your comment. I will look into this on Monday when I get back to the office. I'll report back!

Rubueno:
@Jagularr You're absolutely correct. `tagpass` is the way to go.

Issue #3253, opened 2017-09-20
postgresql_extensible.query cannot connect to pgbouncer


mced:
## Bug report
With last version 1.4, many of us thought that telegraf was compatible with pgbouncer through postgresql_extensible.query input plugin.

It is not the case.

### Relevant telegraf.conf:
```
[[inputs.postgresql_extensible]]
  address = "host=localhost user=telegraf dbname=pgbouncer"

[[inputs.postgresql_extensible.query]]
  sqlquery="SHOW STATS"
  measurement="pgbouncer"
  withdbname=false
  version=0
```

### System info:
Telegraf v1.4.0 (git: release-1.4 34b7a4c3611d1ede908ef275401544c34a4a3ba3)
Ubuntu 14.04
pgbouncer 1.5.4

### Steps to reproduce:
Add telegraf user in pgbouncer config file as `admin_users`.
Then run telegraf.

### Expected behavior:

```
$ telegraf --config /etc/telegraf/telegraf.conf --debug --test
pgbouncer,name=pgbouncer,host=hostname,...
```

### Actual behavior:
```
$ telegraf --config /etc/telegraf/telegraf.conf --debug --test
* Plugin: inputs.postgresql_extensible, Collection 1
2017-09-20T10:29:17Z E! Error in plugin [inputs.postgresql_extensible]: ERROR: not allowed (SQLSTATE 08P01)
```

danielnelson:
Is this from the Open call here?  https://github.com/influxdata/telegraf/blob/master/plugins/inputs/postgresql_extensible/postgresql_extensible.go#L134

james-lawrence:
did we actually confirm it worked? the pgbouncer PR had mixed statements from others on it working? I don't remember the failures being understood and addressed.

the error mentioned here is a [protocol violation](https://www.postgresql.org/docs/9.4/static/errcodes-appendix.html).

Also looking at your configuration: its going to default to port 5432.... is pgbouncer running on that port? Also what is your pgbouncer configuration?

mced:
@danielnelson I can't say

@james-lawrence nope nobody confirmed, and you're right according to https://github.com/influxdata/telegraf/pull/2573#discussion_r134276525, it was not working.

pgbouncer is listening on 5432 port and pooling connections from localhost application to a remote database

#### pgbouncer is listening on 5432 port
```
$ psql -h localhost -d pgbouncer -U telegraf -p 5432
Password for user telegraf: 
[...]

PR #9182, opened 2021-04-23
Fix #9134 - Update pgx to v4

### LINKED ISSUE #9134: [BUG] Inputs PostgreSQL plugin will not work with CA cert validation
### Relevant telegraf.conf:

```toml
[[inputs.postgresql]]
 interval = "1m"
 address = "host=192.168.0.10 port=5432 user=user1 password=xxx database=x1 sslmode=verify-ca sslrootcert=/srv/telegraf_tls/ca.crt"
 databases = ["x1"]
[[inputs.postgresql_extensible]]
 interval = "1m"
 address = "host=192.168.0.10 port=5432 user=user1 password=xxx database=x1 sslmode=verify-ca sslrootcert=/srv/telegraf_tls/ca.crt"
 databases = ["x1"]
 [[inputs.postgresql_extensible.query]]
sqlquery='''
 SELECT t2.rolname as rolname, t3.datname as datname, queryid, calls, ...
'''
```

### System info:

- Docker container
- OS: Ubuntu 20.04.2 LTS
- Telegraf version: 1.16.3

### Docker

[Dockerfile](https://github.com/starkandwayne/pg-multi-exporter/blob/main/Dockerfile)

### Steps to reproduce:

1. Start Telegraf
2. Show the Telegraf logs

### Expected behavior:

I expect the establishment of an encrypted connection to the database and smooth functionality of the PostgreSQL plugin.

### Actual behavior:

We have configured an encrypted connection to the PostgreSQL database including CA-Cert validation in our Telegraf configuration. After starting Telegraf, the following error is displayed in the error log and no connection to the database can be established.
`[inputs.postgresql] Error in plugin: x509: cannot validate certificate for 192.168.0.10 because it doesn't contain any IP SANs
`
### Additional info:

Of course, we have already checked the certificate and established a connection to the DB using the psql client. There were no problems and the encrypted connection could be established without problems in `verify-ca` mode. Our CA certificate does not have a SAN entry, but in my opinion this entry should be checked in verify-full mode. 
I have the feeling that the database driver interprets the rules wrong and uses verify-full instead of verify-ca.

**Workaround:**
As a workaround we changed the ssl mode to `require` and were able to connect to the database.

ZPascal:
- Update pgx to v4
- Update go.mod
- Add license information
- Update PgBouncer test

### Required for all PRs:

This change fixes a bug inside the pgx library. Until now, it was unfortunately not possible to validate a CA certificate in verify-ca mode. The update of the library to V4 solved the corresponding problem. More details can be found, in the bug linked below.

resolves #9134

I updated pgx to version v4.6.0, fixed a typo, adapted the corresponding documentation and the PgBouncer functionality inside the `service.go` file to the new library version.

@helenosheaa Could you please, review this change?


sspaink:
!signed-cla

Jagularr (approved):
Makes sense to me

plugins/inputs/postgresql/service.go
```diff
@@ -111,33 +110,14 @@ func (p *Service) Start(telegraf.Accumulator) (err error) {
 	// Specific support to make it work with PgBouncer too
 	// See https://github.com/influxdata/telegraf/issues/3253#issuecomment-357505343
 	if p.IsPgBouncer {
-		d := &stdlib.DriverConfig{
-			ConnConfig: pgx.ConnConfig{
-				PreferSimpleProtocol: true,
-				RuntimeParams: map[string]string{
-					"client_encoding": "UTF8",
-				},
-				CustomConnInfo: func(c *pgx.Conn) (*pgtype.ConnInfo, error) {
```
helenosheaa, line 120: Do we no longer need any driver config? 

The [comment on 111 ](https://github.com/influxdata/telegraf/issues/3253#issuecomment-357505343) points to a comment around this.

Also [this thread](https://github.com/jackc/pgx/issues/649#issuecomment-585547940) seems to suggest we need/needed to register data types?

scripts/check-deps.sh
```diff
@@ -28,9 +28,6 @@ for dep in $(LC_ALL=C sort -u "${tmpdir}/golist"); do
 		# ignore ourselves
 		github.com/influxdata/telegraf) continue;;
 
-		# dependency is replaced in go.mod
```
helenosheaa, line 31: Related to https://github.com/satori/go.uuid/issues/73, is this resolved?

helenosheaa (commented):
Thanks for the PR! Overall looks good just a few questions in the code. 

Were you able to run the integration tests? As we don't currently run them as part of our circleci checks.

scripts/check-deps.sh
ZPascal: > Were you able to run the integration tests? As we don't currently run them as part of our circleci checks.

I ran the appropriate integration tests locally and looking especially at the tests where I made changes to the actual plugins.

ZPascal:
@helenosheaa 
> Were you able to run the integration tests? As we don't currently run them as part of our circleci checks.

I executed the appropriate integration tests locally and looking especially at the tests where I made changes to the actual plugins.

scripts/check-deps.sh
ZPascal: The problem should have been fixed [here](https://github.com/satori/go.uuid/pull/75). However when I looked at the repository a second time, I noticed that it's not really maintained. For this reason, I think it is better to implement the replace function.

go.mod
```diff
@@ -148,6 +146,3 @@ require (
 	k8s.io/client-go v0.20.4
 	modernc.org/sqlite v1.7.4
 )
-
-// replaced due to https://github.com/satori/go.uuid/issues/73
```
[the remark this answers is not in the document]
ZPascal: The problem should have been fixed, but I think it's the better solution to use the replacement functionality.

plugins/inputs/postgresql/service.go
```diff
@@ -111,33 +110,14 @@ func (p *Service) Start(telegraf.Accumulator) (err error) {
...
-					// https://github.com/jackc/pgx/issues/649
-					info.RegisterDataType(pgtype.DataType{
-						Value: &pgtype.OIDValue{},
-						Name:  "numericOID",
-						OID:   pgtype.NumericOID,
-					})
-
-					return info, nil
-				},
-			},
```
reimda, line 137: It does look like this block is important to pgbouncer support.  Does pgx v4 handle this in another way or does this PR remove pgbouncer support?

reimda (changes requested):

plugins/inputs/postgresql/service.go
ZPascal: In my opinion, the configuration of the driver is no longer needed inside v4. However, I'm not 100% sure and I don't know enough about PGBounder. For this reason, I will first try to create an integration test and clear the corresponding doubts based on this test.

plugins/inputs/postgresql/service.go
ZPascal: In my view, the configuration of the driver should be automatic and the manual setup is no longer needed in v4. Furthermore, the PR should only fix a bug and not remove a feature. To eliminate this risk I will first try to create an integration test.

plugins/inputs/postgresql/service.go
helenosheaa: That's great thanks

ZPascal:
Hi @helenosheaa,

I have looked at the PGBouncer integration test and run it in the main and feature branches. Here I could see the same error message in both branches. I get the following error message and the test is normally skipped:

**Container:**
`failed to connect to `host=localhost user=pgbouncer database=pgbouncer`: server error (ERROR: not allowed (SQLSTATE 08P01)).`

**Console:**
```
2021-05-03 14:41:11.165 UTC [1] LOG C-0x7f028b739100: pgbouncer/pgbouncer@172.21.0.1:61820 login attempt: db=pgbouncer user=pgbouncer tls=no
2021-05-03 14:41:11.165 UTC [1] LOG C-0x7f028b739100: pgbouncer/pgbouncer@172.21.0.1:61820 closing because: not allowed (age=0s)
2021-05-03 14:41:11.165 UTC [1] WARNING C-0x7f028b739100: pgbouncer/pgbouncer@172.21.0.1:61820 pooler error: not allowed
```

Furthermore, I can also see a corresponding error message in both scenarios in the container log. I am not a PGBouncer expert and would like to know if there is anything I need to consider when setting up the containers (PostgreSQL + PGBouncer) and running the tests? When was the last time the tests were run?

Anyway, I then looked at the PgBouncer integration test and the configuration of the connection within service.go and decided to use the PgBouncer integration test. I subsequently modified the container used accordingly and published it to Dockerhub. Furthermore, I modified the service.go according to the issue linked [here](https://github.com/jackc/pgx/issues/650). PgBouncer does not know the configuration parameter ```statement_cache_mode=describe```.

Afterwards I modified the integration test of PgBouncer, because there seems to be an update in PgBouncer and now other metrics are returned in a modified structure.

**Summary**
The adaptation of the integration test, the update of the container and the configuration of the connection allow the further use of PgBouncer and the execution of the PgBouncer integration test without having to skip it.

scripts/check-deps.sh
```diff
@@ -28,8 +28,8 @@ for dep in $(LC_ALL=C sort -u "${tmpdir}/golist"); do
...
-		github.com/satori/go.uuid) continue;;
+    # dependency is replaced in go.mod
+    github.com/satori/go.uuid) continue;;
```
ssoroka, line 32: can you restore the original spacing here? Thanks!

ZPascal: Thanks for the hint. It's done.

helenosheaa:
@ZPascal thanks for working on the integration test! I don't think it was in a working state previously so I really appreciate it.

What was the reason for changing to your own docker image?

I saw that you added `d.PreferSimpleProtocol = true` which was in the previous config. It may turn out we need to add in more custom config but if you address the merge conflicts, I'm happy to approve it as you've got the integration test working.

ZPascal:
@helenosheaa Gladly, no problem

> What was the reason for changing to your own docker image?

I had seen that the existing image was no longer regularly maintained and updated. Furthermore, the stats_users parameter must be set in the corresponding [configuration](https://github.com/ZPascal/pgbouncer-container/blob/main/run.sh#L30). For this reason, I decided to create an updated image.

> I saw that you added d.PreferSimpleProtocol = true which was in the previous config.
Right. It's necessary to set the parameter `PreferSimpleProtocol = true` at least in the case of PgBouncer. It's documented inside the following [issue](https://github.com/jackc/pgx/issues/650).

helenosheaa (approved):
Ok, thanks for clarifying! 

@reimda are you also happy for this to be merged now there's a working integration test?

ZPascal:
@helenosheaa Any news ?

plugins/inputs/postgresql/service.go
```diff
@@ -111,33 +110,16 @@ func (p *Service) Start(telegraf.Accumulator) (err error) {
...
-						Name:  "int8OID",
-						OID:   pgtype.Int8OID,
-					})
-					// Newer versions of pgbouncer need this defined. See the discussion here:
-					// https://github.com/jackc/pgx/issues/649
-					info.RegisterDataType(pgtype.DataType{
-						Value: &pgtype.OIDValue{},
-						Name:  "numericOID",
-						OID:   pgtype.NumericOID,
-					})
```
ssoroka, line 133: it looks like we're still losing support for these data types? Do you know that they're both covered by the new v4 library?

ZPascal: The mapping of the [data types](https://github.com/jackc/pgtype/blob/b26cd223783bf410ec9cd789ae872cef56610d93/pgtype.go#L65) is done inside the driver. Within pgx v4 the mapping has been outsourced and there is a separate repository for it. The version of the driver I am using, uses 1.3.0 of the outsourced types repositories. Furthermore, I have also made the corresponding integration test functional and could already test the corresponding extraction of the metrics there.

ssoroka: got it. just trying to make sure we didn't inadvertently remove support for these two types. Otherwise we're good to go

ssoroka (approved):

plugins/inputs/postgresql/service.go
reimda: Thanks for the link to the pgtype mapping. I didn't know that's how it works but it makes sense now.

reimda (approved):

Issue #8922, opened 2021-03-02
FreeBSD ARM support - Where did you go?


SomKen:
After reviewing my older post from https://github.com/influxdata/telegraf/issues/5282, it seemed as if telegraf would start building freebsd arm packages. After having then installed the 1.15 version of telegraf onto my pfsense, it has survived every version of pfsense Ive upgraded it to since. So, at least in this one case, it still works, but I don't see any new builds.

rsletten:
Would like this as well.

FreeWillyB:
Sign me up as well. I'm working on a pilot network right now. If all goes well, it would involve deploying about 13 or so of the Netgate SG-3100 across the state. I had planned on using Telegraf for metric reporting, but didn't realize the agent wasn't supported.

strman:
@SomKen any chance you could share the steps you took to install telegraf 1.15?

SomKen:
https://forum.netgate.com/topic/142836/telegraf-for-arm-systems-e-g-netgate-sg-3100/65

I'm tracking my progress here. There was a guide I found for building, funny enough, ARM packages on the SG-3100 using qemu / freebsd 12.2. Testing that now.

SomKen:
I just added https://forum.netgate.com/topic/142836/telegraf-for-arm-systems-e-g-netgate-sg-3100/69. I've included the then built version of telegraf. Please note this version is over 2 years old, but is still running on my SG-3100 after all the updates over the two years. I haven't had to do ANYTHING to get it working between updates. This must be the difference between PFSense and FreeBSD.... >_>

I attempted to build a new version last night, but as expected (i believe) GO failed to build after a few hours. Looking at logs, it seems to be because of a run away build, but I haven't had time to really look. I'll dig more tonight. 

Please Influx, please help us here. I will send you an SG-3100 if needed. Netgate is being less than helpful here. After seeing the recent drama related to Wiregurad and their response, it makes sense. 

https://www.reddit.com/r/networking/comments/m6zjie/wireguard_netgate_pfsense_drama/ and https://docs.opnsense.org/history/thefork.html if you want to cry.

SomKen:
Figured out how to build the armv7 packages for Freebsd without the need of an SG-3100, code changes, or emulation. I'll report back later, but GO is amazing.

SomKen:
`GOARM=7 GOOS=freebsd GOARCH=arm go build ./cmd/telegraf`

[...]

PR #9200, opened 2021-04-26
Add FreeBSD GOARCH=arm, GOARM=7 to package step

sspaink:
Close: #8922

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -263,7 +263,7 @@ endif
 
 ifdef armhf
 tars += telegraf-$(tar_version)_linux_armhf.tar.gz
-tars += telegraf-$(tar_version)_freebsd_armhf.tar.gz
+tars += telegraf-$(tar_version)_freebsd_armv7.tar.gz
 debs += telegraf_$(deb_version)_armhf.deb
 rpms += telegraf-$(rpm_version).armv6hl.rpm
 endif 
@@ -436,9 +436,9 @@ upload-nightly:
 %freebsd_i386.tar.gz: export GOOS := freebsd
 %freebsd_i386.tar.gz: export GOARCH := 386
 
-%freebsd_armhf.tar.gz: export GOOS := freebsd
-%freebsd_armhf.tar.gz: export GOARCH := arm
-%freebsd_armhf.tar.gz: export GOARM := 7
+%freebsd_armv7.tar.gz: export GOOS := freebsd
+%freebsd_armv7.tar.gz: export GOARCH := arm
+%freebsd_armv7.tar.gz: export GOARM := 7
 
 %windows_amd64.zip: export GOOS := windows
 %windows_amd64.zip: export GOARCH := amd64
--- a/Makefile
+++ b/Makefile
@@ -257,6 +257,7 @@ endif
 
 ifdef armel
 tars += telegraf-$(tar_version)_linux_armel.tar.gz
+tars += telegraf-$(tar_version)_freebsd_armhf.tar.gz
 rpms += telegraf-$(rpm_version).armel.rpm
 debs += telegraf_$(deb_version)_armel.deb
 endif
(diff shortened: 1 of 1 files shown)


Jagularr (approved):
Nice.

Makefile
```diff
@@ -263,6 +263,7 @@ endif
 
 ifdef armhf
 tars += telegraf-$(tar_version)_linux_armhf.tar.gz
+tars += telegraf-$(tar_version)_freebsd_armhf.tar.gz
```
reimda, line 266: "armhf" is a debian linux term. Freebsd calls it something different.

https://www.freebsd.org/platforms/

Since we have GOARM = 7 later on, I think we want to call this armv7
```suggestion
tars += telegraf-$(tar_version)_freebsd_armv7.tar.gz
```

It's kind of a pain that each OS and distribution seems to have a different name for the different arm versions.

reimda (changes requested):

reimda:
I'm not sure why CI ran twice for the last commit.  All the artifacts are also in telegraf-tiger's comment twice.  The duplicate artifacts are not bit identical because they have different go build ids (see https://golang.org/cmd/buildid/), but should be the same otherwise.

SomKen:
syn ack, i'll test this tonight. Thank you so much! one step closer...

reimda (approved):

SomKen:
Just moved over to this build of telegraf. Will report back on sunday with more. It doesn't seem any of the pf data is being recorded, but maybe my config is missing something.

reimda:
> It doesn't seem any of the pf data is being recorded, but maybe my config is missing something. I wasn't able to fully get this working when I compiled this myself, but I also didn't put a lot of time into it.

Thanks for testing.  Let me know what you find.  The package built with this PR should be the same as the build you made with `GOARM=7 GOOS=freebsd GOARCH=arm go build ./cmd/telegraf` but just put into a tar.gz.  I would expect them to work the same as your build.

SomKen:
> 
> 
> > It doesn't seem any of the pf data is being recorded, but maybe my config is missing something. I wasn't able to fully get this working when I compiled this myself, but I also didn't put a lot of time into it.
> 
> Thanks for testing. Let me know what you find. The package built with this PR should be the same as the build you made with `GOARM=7 GOOS=freebsd GOARCH=arm go build ./cmd/telegraf` but just put into a tar.gz. I would expect them to work the same as your build.

Found the problem with pf. It's totally unrelated to the build. You need to add the "telegraf" user created to the "proxy" group as /dev/pf is owned by root:proxy.

reimda:
@SomKen It sounds like the arm7 build passed your smoke test so we'll merge this.

Which input plugin are you using? I think it makes sense to document the necessary group membership in the plugin's README.md.  Would you be willing to make a PR to add it?

SomKen:
> 
> 
> @SomKen It sounds like the arm7 build passed your smoke test so we'll merge this.
> 
> Which input plugin are you using? I think it makes sense to document the necessary group membership in the plugin's README.md. Would you be willing to make a PR to add it?


Totally! Will do.

Plugins I have enabled. Do note swap doesn't really work correctly on freebsd so it's safe to ignore. 
```
[[inputs.cpu]]
[[inputs.disk]]
[[inputs.diskio]]
[[inputs.kernel]]
[[inputs.mem]]
[[inputs.processes]]
[[inputs.swap]]
[[inputs.system]]
[[inputs.net]]
[[inputs.pf]]

```

SomKen:
> 
> 
> @SomKen It sounds like the arm7 build passed your smoke test so we'll merge this.
> 
> Which input plugin are you using? I think it makes sense to document the necessary group membership in the plugin's README.md. Would you be willing to make a PR to add it?

https://github.com/influxdata/telegraf/pull/9232

I'm open to changes or suggestions.

PR #9223, opened 2021-04-29
plugins/inputs/dovecot: Add support for unix domain sockets

### LINKED ISSUE #9215: input.dovecot: Add support for unix domain sockets
## Feature Request

Add support for protocol 'unix' in input-plugin dovecot

### Current behavior:
only tcp is allowed

### Desired behavior:
allow unix as well

### Use case: 
cPanel uses unix domain sockets for statistics by default in their dovecot configuration.

Plus it's more safe!

I have forked and patched the code already in https://github.com/isodude/telegraf/branches/all?query=patch

isodude:
It's safer for dovecot to export metrics via a UDS instead of tcp port,
this will add support for that option.

### Required for all PRs:

resolves #9215 

dovecot: Add support for unix domain sockets as well

### DIFF
--- a/plugins/inputs/dovecot/README.md
+++ b/plugins/inputs/dovecot/README.md
@@ -14,6 +14,9 @@ the [upgrading steps][upgrading].
   ## specify dovecot servers via an address:port list
   ##  e.g.
   ##    localhost:24242
+  ## or as an UDS socket
+  ##  e.g.
+  ##    /var/run/dovecot/old-stats
   ##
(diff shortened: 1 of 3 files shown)


helenosheaa:
Thanks for opening the pr, overall I think it looks good. Do you think you could add a test for this extension of functionality?

isodude:
During testing I discovered that Server was not populated in the returned metric.
Wrote the unix socket address instead of leaving it empty.

sspaink (approved):

PR #2013, opened 2016-11-08
OpenStack Input Plugin

spjmurray:
Gathers global metrics like cluster resource utilisation, per-hypervisor resource
utilisation and per-project metrics like number of servers, volume usage etc.

Addresses issue #243

PR #9236, opened 2021-05-04
feat: Openstack input plugin

### LINKED ISSUE #243: Add OpenStack plugin
A plugin which queries the various OpenStack APIs in order to obtain statistics about a given component, i.e number of running instances, vCPUs deployed, memory in use, and so on.  By way of comparison, there's a fairly comprehensive collectd plugin that's been written [here](https://github.com/rochaporto/collectd-openstack).

singamSrikar:
added openstack input plugin to gather metrics from openstack services

### Required for all PRs:

- [ ] ~~Wrote appropriate unit tests.~~

resolves #243


singamSrikar:
Hi @sjwang90 and @srebhan ,
This is the new PR for openstack input plugin.

sjwang90:
Thanks @singamSrikar. We'll get a review on this PR.

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,1059 @@
...
+	"github.com/gophercloud/gophercloud/openstack/compute/v2/servers"
+	"github.com/gophercloud/gophercloud/openstack/identity/v3/projects"
+	"github.com/gophercloud/gophercloud/openstack/identity/v3/services"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/agents"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/networks"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/ports"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/subnets"
+	"github.com/gophercloud/gophercloud/openstack/orchestration/v1/stacks"
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/plugins/inputs"
```
srebhan, line 35: It's a matter of taste, but other plugins group the import into "builtin" packages, external github packages and telegraf internal packages. Not a must, but easier if you need to read multiple plugins...

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,1059 @@
...
+)
+
+// serviceType is an OpenStack service type
+type serviceType string
+
+const (
+	volumeV2Service serviceType = "volumev2"
+	Orchestration   serviceType = "orchestration"
+)
+
```
srebhan, line 55: Please fuse the `const` declaration to one block.

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,1059 @@
...
+	"github.com/gophercloud/gophercloud/openstack/identity/v3/projects"
+	"github.com/gophercloud/gophercloud/openstack/identity/v3/services"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/agents"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/networks"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/ports"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/subnets"
+	"github.com/gophercloud/gophercloud/openstack/orchestration/v1/stacks"
+	"github.com/influxdata/telegraf"
+	"github.com/influxdata/telegraf/plugins/inputs"
+	"log"
```
srebhan, line 36: You should never use `log` in a plugin. Please use the `telegraf.Logger` facility by declaring
```golang
```
in your plugin struct. The logger will then be set by telegraf and you can simply use it as `o.Log.Errorf` or similar.

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,1059 @@
...
+type serverDiag map[string]interface{}
+
+// serviceMap maps service id to a Service struct.
+type serviceMap map[string]services.Service
+
+// stackMap maps stack id to a Stack struct.
+type stackMap map[string]stacks.ListedStack
+
+// nova_serviceMap maps nova_service id to a nova Service struct.
+type nova_serviceMap map[string]nova_services.Service
```
srebhan, line 72: Please use those directly in the code instead of declaring them here. This way the type will be obvious in the code.

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,1059 @@
...
+	volumes.Volume
+	volumetenants.VolumeTenantExt
+}
+
+// volumeMap maps a volume id to a volume struct.
+type volumeMap map[string]volume
+
+// storagePoolMap maps a storage pool name to a StoragePool struct.
+type storagePoolMap map[string]schedulerstats.StoragePool
+
```
srebhan, line 122: See above.

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,1059 @@
...
+// the Input interface.
+func (o *OpenStack) Description() string {
+	return "Collects performance metrics from OpenStack services"
+}
+
+// sampleConfig is a sample configuration file entry.
+var sampleConfig = `
+  ## This is the recommended interval to poll.
+  interval = '30m'
+
```
srebhan, line 170: This is misleading as it is a Telegraf option. Please do not include general options here.

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,1059 @@
...
+  ## The identity endpoint to authenticate against and get the
+  ## service catalog from
+  identity_endpoint = "https://my.openstack.cloud:5000"
```
srebhan, line 173: How about `authentication_endpoint`?

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,1059 @@
...
+
+  ## The domain to authenticate against when using a V3
+  ## identity endpoint.  Defaults to 'default'
+  domain = "default"
+
+  ## The project to authenticate as
+  project = "admin"
```
srebhan, line 180: Please comment default values and set the respective default values in an `Init()` function.

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,1059 @@
...
+
+  ## The user to authenticate as, must have admin rights
+  username = "admin"
+
+  ## The user's password to authenticate with
+  password = "Passw0rd"
```
srebhan, line 186: How about shorten this to:
```suggestion
  ## User authentication credentials. Must have admin rights!
  # username = "admin"
  # password = "Passw0rd"
```

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,1059 @@
...
+
+  ## Services to be enabled
+  #enabled_services = ["stacks","services", "projects", "hypervisors", "flavors", "servers", "volumes", "storage" , "subnets", "ports", "networks", "aggregates", "nova_services", "agents"]
```
srebhan, line 189: Please be nice and spend some test instead of making this look like a commented option...
```suggestion
  ## Available services are: 
  ##  "agents", "aggregates", "flavors", "hypervisors", "networks", "nova_services",
  ##  "ports", "projects", "servers", "services", "stacks", "storage", "subnets", "volumes"
```

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,1059 @@
...
+  enabled_services = ["services", "projects", "hypervisors", "flavors", "networks", "volumes"]
```
srebhan, line 190: If this is a sensible default please use
```suggestion
  # enabled_services = ["services", "projects", "hypervisors", "flavors", "networks", "volumes"]
```
and set this list in `Init()`.

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,1059 @@
...
+
+  #Dependencies
+  # | Service | Depends on |
+  # | servers | projects, hypervisors, flavors |
+  # | volumes | projects |
```
srebhan, line 195: What does this mean? What should the user set here and how? Documentation starts with double-hash...

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,1059 @@
...
+type OpenStack struct {
+	// Configuration variables
+	IdentityEndpoint   string
+	Domain             string
+	Project            string
+	Username           string
+	Password           string
+	EnabledServices    []string
+	ServerDiagnotics   bool
+	InsecureSkipVerify bool
```
srebhan, line 133: You should probably include `tls.ClientConfig`. Take a look at e.g. the prometheus input plugin on how to use it.

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,1059 @@
...
+// OpenStack is the main structure associated with a collection instance.
+type OpenStack struct {
+	// Configuration variables
+	IdentityEndpoint   string
+	Domain             string
+	Project            string
+	Username           string
+	Password           string
+	EnabledServices    []string
+	ServerDiagnotics   bool
```
srebhan, line 132: Please add a `toml` declaration to all of the user-configurable options.

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,1059 @@
...
+
+  #Dependencies
+  # | Service | Depends on |
+  # | servers | projects, hypervisors, flavors |
+  # | volumes | projects |
+
+  ## Collect Server Diagnostics
+  server_diagnotics = false
+
+  InsecureSkipVerify = false
```
srebhan, line 200: Do not use camel-case style for config options. Those are snakes. ;-)

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,1059 @@
...
+	return sampleConfig
+}
+
+// Gather gathers resources from the OpenStack API and accumulates metrics.  This
+// implements the Input interface.
+func (o *OpenStack) Gather(acc telegraf.Accumulator) error {
+	// Perform any required set up
+	if err := o.initialize(); err != nil {
+		return err
+	}
```
srebhan, line 215: You don't need this with the change of `initialize()` to `Init()` below.

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,1059 @@
...
+		"aggregates":    o.gatherAggregates,
+		"nova_services": o.gatherNovaServices,
+		"agents":        o.gatherAgents,
+		"stacks":        o.gatherStacks,
+	}
+	for resources, gatherer := range gatherers {
+		for _, i := range o.EnabledServices {
+			if resources == i {
+				if err := gatherer(); err != nil {
+					log.Println("W!", plugin, "failed to get", resources, err)
```
srebhan, line 238: Use the telegraf logger facility with
```suggestion
					o.Log.Warnf("failed to get resource %q: %v", resources, err)
```

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,1059 @@
...
+		"ports":         o.gatherPorts,
+		"networks":      o.gatherNetworks,
+		"aggregates":    o.gatherAggregates,
+		"nova_services": o.gatherNovaServices,
+		"agents":        o.gatherAgents,
+		"stacks":        o.gatherStacks,
+	}
+	for resources, gatherer := range gatherers {
+		for _, i := range o.EnabledServices {
+			if resources == i {
```
srebhan, line 236: Wouldn't
```suggestion
	for _, service := range o.EnabledServices {
		gatherer := gatherers[service]
```
be much easier?

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,1059 @@
...
+		"agents":        o.accumulateAgents,
+		"stacks":        o.accumulateStacks,
+	}
+	for resources, accumulator := range accumulators {
+		for _, i := range o.EnabledServices {
+			if resources == i {
+				accumulator(acc)
+			}
+		}
+	}
```
srebhan, line 267: See above.

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,1059 @@
...
+	return nil
+}
+
+// initialize performs any necessary initialization functions
+func (o *OpenStack) initialize() error {
```
srebhan, line 272: Just rename this to
```suggestion
func (o *OpenStack) Init() error {
```
and telegraf will automatically call this function after all user-config options are set.

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,1059 @@
...
+	authOption := gophercloud.AuthOptions{
+		IdentityEndpoint: o.IdentityEndpoint,
+		DomainName:       o.Domain,
+		TenantName:       o.Project,
+		Username:         o.Username,
+		Password:         o.Password,
+	}
+	provider, err := openstack.NewClient(authOption.IdentityEndpoint)
+	if err != nil {
+		return fmt.Errorf("Unable to create Newclient for OpenStack endpoint: %v", err)
```
srebhan, line 283: Error messages should start lowercase and not end with punctuation.

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,1059 @@
...
+	}
+
+	tlsconfig := &tls.Config{}
+	tlsconfig.InsecureSkipVerify = o.InsecureSkipVerify
```
srebhan, line 287: You might want to use `tls.ClientConfig` to support custom certificates etc.

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,1059 @@
...
+	transport := &http.Transport{TLSClientConfig: tlsconfig}
+	provider.HTTPClient = http.Client{
+		Transport: transport,
+	}
+
+	err = openstack.Authenticate(provider, authOption)
+
+	if err != nil {
+		return fmt.Errorf("Unable to authenticate OpenStack user: %v", err)
+	}
```
srebhan, line 297: How about
```suggestion
	if err := openstack.Authenticate(provider, authOption); err != nil {
		return fmt.Errorf("unable to authenticate user: %v", err)
	}
```

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,1059 @@
...
+
+	// Create required clients and attach to the OpenStack struct
+	if o.identity, err = openstack.NewIdentityV3(provider, gophercloud.EndpointOpts{}); err != nil {
+		return fmt.Errorf("unable to create V3 identity client: %v", err)
+	}
+
+	o.services = serviceMap{}
+	o.gatherServices()
```
srebhan, line 305: Please check the returned error!

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,1059 @@
...
+	o.volumes = volumeMap{}
+	o.storagePools = storagePoolMap{}
+	o.subnets = subnetMap{}
+	o.networks = networkMap{}
+	o.ports = portMap{}
+	o.aggregates = aggregateMap{}
+	o.nova_services = nova_serviceMap{}
+	o.agents = agentMap{}
+	o.diag = serverDiag{}
+	return nil
```
srebhan, line 344: You might want to add a newline before `return`, but not a must.

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,1059 @@
...
+	nova_services, err := nova_services.ExtractServices(page)
+	if err != nil {
+		return fmt.Errorf("unable to extract nova_services")
+	}
+	for _, nova_service := range nova_services {
+		o.nova_services[nova_service.ID] = nova_service
+	}
+	return nil
+}
+
```
srebhan, line 394: All those function look very similar, you first call `List` with empty options, then `AllPages()`, `ExtractServices` etc. The only difference is the used OpenStack member variables. The fields `services`, `nova_services`, etc. are even of the same interface type.

So how about unifying these functions and implement them only once. Return a map and set the respective members outside of this function (e.g. in `Gather()`) if you need to do this at all...

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,1059 @@
...
+	for server_id, diagnostic := range o.diag {
+		tags := tagMap{
+			"server_id": server_id,
+		}
+		fields := fieldMap{
+			"server_id": server_id,
+		}
+		switch reflect.TypeOf(diagnostic).Kind() {
+		case reflect.Map:
+			s := reflect.ValueOf(diagnostic).MapRange()
```
srebhan, line 593: Couldn't you do the same with
```suggestion
		switch s:= diagnostic.(type) {
		case map[string]interface{}:
```
or even better
```suggestion
		s, ok := diagnostic.(map[string]interface{})
		if !ok {
			o.Log.Debugf("unknown type for diagnostics: %T", diagnostic)
			continue
		}
```

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,1059 @@
...
+
+// accumulateServerDiagnostics accumulates statistics from the compute(nova) service.
+func (o *OpenStack) accumulateServerDiagnostics(acc telegraf.Accumulator) {
+	for server_id, diagnostic := range o.diag {
+		tags := tagMap{
+			"server_id": server_id,
+		}
+		fields := fieldMap{
+			"server_id": server_id,
+		}
```
srebhan, line 590: Please decide for the one or the other. It's bad habit to have the same value as tag and field. I would opt for using a `tag` if cardinality permits...

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,1059 @@
...
+						fields["port_"+port_map[tap_list[0]]+"_"+strings.Join(tap_list[1:], "_")] = s.Value().Interface().(float64)
+						tags["port_"+port_map[tap_list[0]]] = tap_list[0]
+					}
+				} else {
+					fields[s.Key().Interface().(string)] = s.Value().Interface().(float64)
+				}
+			}
+			fields["no_of_ports"] = port_count
+		}
+		acc.AddFields("openstack_server_diagnostics", fields, tags)
```
srebhan, line 614: Do we really want different measurements here or would using `openstack` as measurement name with a `tag` `type="server diagnostics"` be better?

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,1059 @@
...
+// accumulateStacks accumulates statistics from the stack service.
+func (o *OpenStack) accumulateStacks(acc telegraf.Accumulator) {
+	for _, stack := range o.stacks {
+		tags := tagMap{
+			"id": stack.ID,
+		}
+		fields := fieldMap{
+			"creation_time":       stack.CreationTime.Format("2006-01-02T15:04:05.999999999Z07:00"),
+			"description":         stack.Description,
+			"id":                  stack.ID,
```
srebhan, line 627: Please use only a tag. You can still convert the tag to a field using processors.

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,1059 @@
...
+			"stack_name":          stack.Name,
+			"stack_status":        stack.Status,
+			"stack_status_reason": stack.StatusReason,
+			"updated_time":        stack.UpdatedTime.Format("2006-01-02T15:04:05.999999999Z07:00"),
```
srebhan, line 631: Please think about which information can be a `tag`. I guess `stack.Name` or `stack.Status` are good candidates without growing the cardinality too much.

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,1059 @@
...
+  ## This is the recommended interval to poll.
+  interval = '30m'
+
+  ## The identity endpoint to authenticate against and get the
+  ## service catalog from
+  identity_endpoint = "https://my.openstack.cloud:5000"
+
+  ## The domain to authenticate against when using a V3
+  ## identity endpoint.  Defaults to 'default'
+  domain = "default"
```
srebhan, line 177: This value (`"default"`) does not match the actually set default (`"Default"`) in `init()`... "When documentation does not match the code probably both are wrong..." ;-P

srebhan (changes requested):
Hey @singamSrikar! Thanks for pushing this forward!
I have some comments in the code. The biggest request I have is to unify the `gather...()` functions to reduce the amount of code. They all share a common structure so I guess there is potential for optimization. :-)

Take a look at the comments in the code please and let me know when you are ready for another round or if you have some questions/comments.

plugins/inputs/openstack/openstack.go
singamSrikar: That is a really good idea but might not be possible in this case as every type(Openstack services) has different keys(in terms of maps/json) or columns(in terms of DB table) it would not be possible to check whether if a particular column is applicable to a service or not (e.g. in case if the value is null/empty). Please let me know if you have a different approach towards this.

srebhan: Well the line-protocol supports these different "series" in the same measurement... But it might be difficult for other setups to deal with it. I rely on you in this point to do a sensible analysis given your use case. :-) Just asking to be sure you thought about it... In case you think it's better to have multiple measurements, please mark this conversation as "resolved".

helenosheaa:
Hi @singamSrikar, thanks for the PR, we are really interested in moving forward with it! 

Would you be able to address @srebhan's comments? Also the linter has picked up on a few things. 

Finally it would be great to see some unit tests so we can ensure the plugin is working as expected.

singamSrikar:
Hi @helenosheaa, I have addressed almost all the comments except for one major change. I am working on that, will push the code once it is done.

srebhan:
@singamSrikar thanks for working on this! Please let me know once you pushed your changes so I can take a look.

plugins/inputs/openstack/openstack.go
singamSrikar: I don't think this would work because diagnostic type is map[string]interface{} @srebhan

srebhan: Well then use
```suggestion
		s, ok := diagnostic.(map[string]interface{}) {
		if !ok {
			o.Log.Debugf("unknown type for diagnostics: %T", diagnostic)
			continue
		}
```
My point is that you don't need to use the `reflect` package but you can use type-assertion directly.

singamSrikar: Yeah that's true, thank you.

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,1059 @@
...
+  ## The recommended interval to poll is '30m'
+
+  ## The identity endpoint to authenticate against and get the service catalog from.
+  authentication_endpoint = "https://my.openstack.cloud:5000"
+
+  ## The domain to authenticate against when using a V3 identity endpoint. Defaults to 'default'.
+  # domain = "default"
+
+  ## The project to authenticate as. Defaults to 'admin'.
+  # project = "admin"
```
srebhan, line 130: Please remove the `Defaults to` comments as it is redundant with the declaration below.

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,1059 @@
...
+	authOption := gophercloud.AuthOptions{
+		IdentityEndpoint: o.IdentityEndpoint,
+		DomainName:       o.Domain,
+		TenantName:       o.Project,
+		Username:         o.Username,
+		Password:         o.Password,
+	}
+	provider, err := openstack.NewClient(authOption.IdentityEndpoint)
+	if err != nil {
+		return fmt.Errorf("unable to create Newclient for OpenStack endpoint %v", err)
```
srebhan, line 174:
```suggestion
		return fmt.Errorf("unable to create client for OpenStack endpoint %v", err)
```

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,1059 @@
...
+		return err
+	}
+
+	provider.HTTPClient = http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: tlsCfg,
+			Proxy:           http.ProxyFromEnvironment,
+		},
+		Timeout: time.Duration(o.Timeout),
+	}
```
srebhan, line 188: You might want to use `common/http`'s `CreateClient()` to also cover proxying.

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,1059 @@
...
+
+	o.gather_servers = false
+
+	return nil
+}
+
+// Gather gathers resources from the OpenStack API and accumulates metrics.  This
+// implements the Input interface.
+func (o *OpenStack) Gather(acc telegraf.Accumulator) error {
+
```
srebhan, line 234: Please remove those newlines at the start of the function.

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,1059 @@
...
+		"stacks":        o.gatherStacks,
+	}
+
+	for _, service := range o.EnabledServices {
+		if service != "services" {
+			gatherer := gatherers[service]
+			if err := gatherer(); err != nil {
+				o.Log.Warnf("failed to get resource %q %v", service, err)
+			}
+		}
```
srebhan, line 259: Please invert the condition to save an indention level. Furthermore, I guess it's an error if you cannot get a service so use `acc.AddError()`:
```suggestion
		if service == "services" {
			continue
		}
		
		gatherer := gatherers[service]
		if err := gatherer(); err != nil {
			acc.AddError(fmt.Errorf("failed to get resource %q %v", service, err))
		}
```

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,1059 @@
...
+	}
+
+	for _, service := range o.EnabledServices {
+		accumulator := accumulators[service]
+		accumulator(acc)
+	}
+
+	if o.ServerDiagnotics && !o.gather_servers {
+		if err := o.gatherServers(); err != nil {
+			o.Log.Warnf("failed to get resource servers %v", err)
```
srebhan, line 287: Maybe also use `acc.AddError()`?

plugins/inputs/openstack/openstack.go
srebhan: I now see that you cannot unify the gather functions as those are referencing the packages directly instead of "class" instances.

srebhan (changes requested):
Hey @singamSrikar, thanks for the changes. The code definitively made a big step forward. However, there are still some issues left:

1. Please join the `gather*()` and `accumulate*()` functions to save two lookups and duplicate looping over the query results. It furthermore will dramatically reduce the number of member variables for the `OpenStack` struct.
2. Check the cardinality of your created output. The code has multiple instances of using a time as `tag` which asks for trouble with respect to cardinality. Please avoid using "often changing" or unbound values for `tags` and rather output them as `fields`.
3. Please check the "secret leaking" in your output. You e.g. output `admin_pass` which will store credential information into a database possibly accessible by an audience not intended to know this information.

There are also some more minor things, please check the comments in the code.

I'm looking forward to the next round. Thanks for working on this PR!

plugins/inputs/openstack/openstack.go
singamSrikar: I am assuming that acc.AddError will not terminate the process/flow and collect metrics for rest of the configured services. Is my assumption right?

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,1059 @@
...
+
+// accumulateStacks accumulates statistics from the stack service.
+func (o *OpenStack) accumulateStacks(acc telegraf.Accumulator) {
+	for _, stack := range o.stacks {
+		tags := map[string]string{
+			"creation_time": stack.CreationTime.Format("2006-01-02T15:04:05.999999999Z07:00"),
+			"description":   stack.Description,
+			"id":            stack.ID,
+			"name":          stack.Name,
+			"stack_tags":    strings.Join(stack.Tags[:], ","),
```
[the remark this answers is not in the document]
singamSrikar: Actually we wanted to use them directly but the tags in Openstack are of array format ([ ]string) where as in Telegraf it is map.
So, is there any way we can use actual Openstack tags directly? 

We thought of two ways,
1. to use static names as keys in map with tags as values(eg: "tag_1": "prod"), this will also be of not much use because of uncertainty of where the tag would be placed (tag_n).
2. to use tag values as keys and set value to "true" (eg: "prod": "true"), with this approach the cardinality might increase drastically if there are large number of unique tags in Openstack for a particular service type.

Still think option 2 is worth incorporating but wasn't sure if the Influx community will agree. What do you think?

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,1059 @@
...
+// accumulateServers accumulates statistics about servers.
+func (o *OpenStack) accumulateServers(acc telegraf.Accumulator) {
+	var compute_hosts = map[string]map[string]string{}
+	// establishing relation between server host_id and hypervisor host_name.
+	for _, project := range o.projects {
+		compute_hosts[project.ID] = map[string]string{}
+		for _, hypervisor := range o.hypervisors {
+			h := sha256.New224()
+			h.Write([]byte(string(project.ID) + string(hypervisor.HypervisorHostname)))
+			compute_hosts[project.ID][hex.EncodeToString(h.Sum(nil))] = hypervisor.HypervisorHostname
```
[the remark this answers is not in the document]
singamSrikar: The host id of a server in Openstack is a combination of project/tenant id and hypervisor id, so to correlate I am generating the host id( hex.EncodeToString(h.Sum(nil)) ), and adding the hypervisor details(host_name) in server metrics which is masked in host id of a server.

plugins/inputs/openstack/openstack.go
srebhan: Yes your assumption is correct. That's the intended use-case for `acc.AddError()`.

plugins/inputs/openstack/openstack.go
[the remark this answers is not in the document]
srebhan: Oh I see. Didn't notice that when making the comment. I agree that option 2 is worth a thought. My preferred way to handle cardinality issues would be to use the telegraf `taginclude`/`tagexclude` mechanisms. Let me discuss this with the core-developers and I'll get back to you on this issue. Ok?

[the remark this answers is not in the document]
singamSrikar: That's sounds good.

plugins/inputs/openstack/openstack.go
singamSrikar: Thanks for confirming.

plugins/inputs/openstack/openstack.go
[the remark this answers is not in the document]
srebhan: Yeah but you are _already_ in a map that is _unique_ to the `project_id`, so you do not need to incorporate this into your hash. So the only one left in defining your hash (the only other part written to `h`) is `hypervisor.HypervisorHostname`, so why don't you use this in the map directly?
The only reason that makes sense to me is that the hash you compute is what is done to derive `server.HostID` which you are using later during lookup. But this is a fragile thing as whenever Openstack changes the way `server.HostID` is computed this plugin will break. So I think it's better to use fields that are consistent and do not rely on knowing the internals of Openstack.

Am I missing something?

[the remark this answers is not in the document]
singamSrikar: To know on which compute/hypervisor node the server is launched on, we will need an id/name of hypervisor to be present in server struct, but we have `server.HostID` which is a combination/hash of `project_id` and `hypervisor.HypervisorHostname` ( i.e. `h.Write([]byte(string(project.ID) + string(hypervisor.HypervisorHostname)))`). So, without the hash there is no correlation between server and the hypervisor upon which it is launched.

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,1059 @@
...
+		tags["fault_message"] = server.Fault.Message
+		tags["host_name"] = host_name
+		tags["project"] = project
+
+		fields := map[string]interface{}{
+			"progress":         server.Progress,
+			"accessIPv4":       server.AccessIPv4,
+			"accessIPv6":       server.AccessIPv6,
+			"addresses":        len(server.Addresses),
+			"adminPass":        server.AdminPass,
```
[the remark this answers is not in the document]
singamSrikar: Thinking of mentioning 'Measurement filtering' option in README for this.

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,1059 @@
...
+			"status":               volume.Status,
+			"availability_zone":    volume.AvailabilityZone,
+			"created_at":           volume.CreatedAt.Format("2006-01-02T15:04:05.999999999Z07:00"),
+			"updated_at":           volume.UpdatedAt.Format("2006-01-02T15:04:05.999999999Z07:00"),
+			"name":                 volume.Name,
+			"description":          volume.Description,
+			"volume_type":          volume.VolumeType,
+			"snapshot_id":          volume.SnapshotID,
+			"source_volid":         volume.SourceVolID,
+			"user_id":              volume.UserID,
```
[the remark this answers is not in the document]
singamSrikar: Thinking of mentioning 'Measurement filtering' option in README for this.

plugins/inputs/openstack/openstack.go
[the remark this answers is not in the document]
srebhan: Please don't. Instead provide a user-configurable option (e.g. `output_secrets`) that is set to **not** outputting these information by default. Same for the other instances.

plugins/inputs/openstack/openstack.go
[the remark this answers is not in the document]
srebhan: Maybe I'm missing something, but you have a hash consisting of `h = hash(project.ID+hypervisor.HypervisorHostname)` that you use in a `m = map[project.ID][h]`. As the hash `h` contains the project-ID it is unique there, so is the map entry (as the first map level also contains the project-ID).
So if you just use `map[project.ID][hypervisor.HypervisorHostname]` you have the same "uniqueness" as the project-ID in the former example is redundant. Alternatively you could use `map[h]` and use the hash, but this would complicate the lookup I guess.

So the question is: Why do you need to use the project-ID twice?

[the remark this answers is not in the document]
singamSrikar: Let me put it this way, if H1 and H2 are hypervisors, and P1 and P2 be projects then `server.HostID` for a server launched on H1 in P1 is hash(P1+H1) and for the a server deployed on H1 in P2 is hash(P2+H1), and similarly for server on H2 in P1 is hash(P1+H2) and for server on H2 in P2 is hash(P2+H2). The server struct consists `project.ID` and `HostID`, so to know on which Hypervisor the server is deployed on, I am creating a `map[project.ID]{HostID : hypervisor.HypervisorHostname}` to obtain the `hypervisor.HypervisorHostname` using the `server.HostID`.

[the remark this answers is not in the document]
srebhan: Ok, but who does set `server.HostID = hash(P1+H1)`? Openstack? If so, how do find out if they later change to `server.HostID = hash(P1+H1+somesecretsource)` or `server.HostID = some_different_hash_function(P1+H1)`?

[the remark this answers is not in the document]
singamSrikar: I have changed the flow which will add host name without need of hash.

plugins/inputs/openstack/openstack.go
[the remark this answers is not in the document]
singamSrikar: Added configurable option 'output_secrets'.

plugins/inputs/openstack/openstack.go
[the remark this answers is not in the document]
singamSrikar: Any update on this topic?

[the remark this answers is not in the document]
srebhan: Hey @singamSrikar, sorry for this taking longer than expected. I talked to @reimda and we think that adding the tags as separate metric tags is the best way. The idea is to allow the user to specify a `tag_prefix` e.g. `tag_prefix = "openstack_tag_"` (which could be the default). Furthermore, let the user specify a `tag_value`, as Openstack tags do not have a value. The default could be empty `tag_value = ""`. 

We then convert the OpenStack tags e.g. `red,blue,green` into metric tags. With the default values for `tag_prefix` and `tag_value` this will result in `openstack_tag_red=""`, `openstack_tag_green=""`, `openstack_tag_blue=""` on the metric.
This gives the user maximum flexibility with sensible defaults. We could even disable those tags by setting empty `tag_prefix=""` .

What do you think?

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,908 @@
...
+	// Locally cached resources
+	flavors     map[string]flavors.Flavor
+	hypervisors []hypervisors.Hypervisor
+	diag        map[string]interface{}
+	projects    map[string]projects.Project
+	services    map[string]services.Service
+}
+
+// ContainsService indicates whether a particular service is enabled
+func (o *OpenStack) ContainsService(t string) bool {
```
srebhan, line 89: Please do not export this function.

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,908 @@
...
+		if service.Type == t {
+			return true
+		}
+	}
+
+	return false
+}
+
+// InEnabledServices indicates whether a particular service is included in EnabledServices
+func (o *OpenStack) InEnabledServices(t string) bool {
```
srebhan, line 100: Same here, please do only export functions that need to be called from outside of the plugin. This usually boils down to the interface-functions required by Telegraf.

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,908 @@
...
+	}
+
+	// Create required clients and attach to the OpenStack struct
+	if o.identity, err = openstack.NewIdentityV3(provider, gophercloud.EndpointOpts{}); err != nil {
+		return fmt.Errorf("unable to create V3 identity client %v", err)
+	}
+
+	if err := o.gatherServices(); err != nil {
+		o.Log.Warnf("failed to get resource openstack services %v", err)
+	}
```
srebhan, line 208: Does it make sense to continue if this fails?

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,908 @@
...
+	for _, service := range o.EnabledServices {
+		if service == "services" {
+			o.accumulateServices(acc)
+			continue
+		}
+		gatherer := gatherers[service]
+		if err := gatherer(acc); err != nil {
+			acc.AddError(fmt.Errorf("failed to get resource %q %v", service, err))
+		}
+	}
```
srebhan, line 266: The comment says
> Note service harvesting must come first as the other gatherers are dependant on this information.

However, the loop does not guarantee that `services` are accumulated first. I also assume that `services` always has to be "enabled" i.e. in `EnabledServices`, so you can do the following
```suggestion
	// Gather resources.  Note service harvesting must come first as the other
	// gatherers are dependant on this information.
	gatherers := map[string]func(telegraf.Accumulator) error{
		"projects":      o.gatherProjects,
		"hypervisors":   o.gatherHypervisors,
		"flavors":       o.gatherFlavors,
		"servers":       o.gatherServers,
		"volumes":       o.gatherVolumes,
		"storage_pools": o.gatherStoragePools,
		"subnets":       o.gatherSubnets,
		"ports":         o.gatherPorts,
		"networks":      o.gatherNetworks,
		"aggregates":    o.gatherAggregates,
		"nova_services": o.gatherNovaServices,
		"agents":        o.gatherAgents,
		"stacks":        o.gatherStacks,
	}

	o.accumulateServices(acc)
	for _, service := range o.EnabledServices {
		if service == "services" {
			continue
		}
		gatherer := gatherers[service]
		if err := gatherer(acc); err != nil {
			acc.AddError(fmt.Errorf("failed to get resource %q %v", service, err))
		}
	}
```
Or filter out `services` from the list of `EnabledServices` in `Init()`.

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,908 @@
...
+	stacks, err := stacks.ExtractStacks(page)
+	if err != nil {
+		return fmt.Errorf("unable to extract stacks %v", err)
+	}
+	for _, stack := range stacks {
+		tags := map[string]string{
+			"description": stack.Description,
+			"name":        stack.Name,
+			"stack_tags":  strings.Join(stack.Tags[:], ","),
```
srebhan, line 293: Just putting a comment here as a marker regarding our `tag` conversion discussion. :-)

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,908 @@
...
+		}
+		fields := map[string]interface{}{
+			"status":        stack.Status,
+			"id":            stack.ID,
+			"status_reason": stack.StatusReason,
+			"creation_time": stack.CreationTime.Format("2006-01-02T15:04:05.999999999Z07:00"),
+			"updated_time":  stack.UpdatedTime.Format("2006-01-02T15:04:05.999999999Z07:00"),
```
srebhan, line 300: It's a matter of taste if you want to store the times as `string` (as you do) or as a nanosecond `int64` timestamp. If you want to do computations, the `int64` is easier on downstream queries, however the `string` is easier to read for humans. How about adding a user-configurable flag that let's you decide the format e.g. `timestamp_format = "unix_ns"`. Your format string could be the default...

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,908 @@
...
+	services, err := services.ExtractServices(page)
+	if err != nil {
+		return fmt.Errorf("unable to extract services %v", err)
+	}
+	for _, service := range services {
+		o.services[service.ID] = service
+	}
+
+	return nil
+}
```
srebhan, line 323: Can you please move this function up as the first of the `gatherXYZ()` group as it seems special in terms of structure and importance.

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,908 @@
...
+			allocation_pools = append(allocation_pools, pool.Start+"-"+pool.End)
+		}
+		tags := map[string]string{
+			"network_id":        subnet.NetworkID,
+			"name":              subnet.Name,
+			"description":       subnet.Description,
+			"ip_version":        strconv.Itoa(subnet.IPVersion),
+			"cidr":              subnet.CIDR,
+			"gateway_ip":        subnet.GatewayIP,
+			"dns_nameservers":   strings.Join(subnet.DNSNameservers[:], ","),
```
srebhan, line 377: These information are potentially volatile leading to large cardinality. How about making them fields, as Usually people won't filter the metrics using this information.

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,908 @@
...
+	for _, port := range ports {
+		tags := map[string]string{
+			"network_id":      port.NetworkID,
+			"name":            port.Name,
+			"description":     port.Description,
+			"status":          port.Status,
+			"tenant_id":       port.TenantID,
+			"project_id":      port.ProjectID,
+			"device_owner":    port.DeviceOwner,
+			"security_groups": strings.Join(port.SecurityGroups[:], ","),
```
srebhan, line 414: This will make it impossible to e.g. group-by `security_groups`. If so, you can also move the information to a field.

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,908 @@
...
+				}
+				if len(server.AttachedVolumes) == 0 {
+					acc.AddFields("openstack_server", fields, tags)
+				} else {
+					for _, AttachedVolume := range server.AttachedVolumes {
+						fields["volume_id"] = AttachedVolume.ID
+						acc.AddFields("openstack_server", fields, tags)
+					}
+				}
+			}
```
srebhan, line 735: How about moving the inner of the `for` loop to a separate function (e.g. `gatherServerDetails()`) and then do
```golang
		if server_gather {
			for _, server := range servers {
				o.gatherServerDetails(acc, server)
			}
		}
```

srebhan (changes requested):
Hey @singamSrikar, this looks nice already. However, some more comments in the code. Most prominent is the `tag` conversion we discussed and (additionally) my request for let the user specify the `timestamp_format` to allow users to save the times as timestamps.

In general, I want to ask you to think about your categorization into `tags` and `fields`. In e.g. InfluxDB, tags are used for efficiently filter/group metrics, while fields usually contain the information or values to compute on. Looking at your resulting metrics, I sometimes have a hard time imagining how you want to use those `tags`. Especially if those are potentially changing arbitrarily *and*/*or* if you concatenate them.

Let's take the list of DNS servers (in this example `dns1`, `dns2` and `dns3`). You concatenate those (without sorting), so for 5 different servers you will (potentially) get the following tags
```
server1,dns_servers="dns1, dns2, dns3",...
server2,dns_servers="dns2",...
server3,dns_servers="dns3, dns1",...
server4,dns_servers="dns3, dns2, dns1",...
server5,dns_servers="dns1, dns2",...
```
What do you do with these `tags`? 
It might be interesting to find out how many `servers` query which `dns_server`, but you cannot get this information (at least not with non-exponential effort).
If you do not care about this kind of query/grouping/filtering, please move the `dns_servers` to `fields` as they are purely informative.
If you however want to do this kind of query/grouping/filtering, your only choice is to keep `dns_servers` as tags and duplicate the metrics
```
server1,dns_servers="dns1",...
server1,dns_servers="dns2",...
server1,dns_servers="dns3",...
server2,dns_servers="dns2",...
server3,dns_servers="dns3",...
server3,dns_servers="dns1",...
server4,dns_servers="dns3",...
server4,dns_servers="dns2",...
server4,dns_servers="dns1",...
server5,dns_servers="dns1",...
server5,dns_servers="dns2",...
```

plugins/inputs/openstack/openstack.go
singamSrikar: There are two methods for 'services' one to gather resources(gatherServices function) and store in a map, other to output the data to accumulator(accumulateServices function). 'gatherServices' is called in Init() and depending on whether 'services' is in EnabledServices, accumulateServices method is called.

plugins/inputs/openstack/openstack.go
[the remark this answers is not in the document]
singamSrikar: Hey @srebhan, I do not think that would a good idea as 1)  tags `red,blue,green` would create 3 new columns in the measurement and similarly there could be n different tags leading to n different columns increasing cardinality drastically (for eg:. hypervisor has about 30 tags), and 2) with tag value as empty string we would not be able to determine if a particular entry in the measurement has this tag or not.

plugins/inputs/openstack/openstack.go
singamSrikar: I agree, but wouldn't that create a problem in InfluxDB as `unix_ns` is `int64` format and human readable is in `string` (to different formats for same field/tag)?
And storing `unix_ns` as `string` wouldn't add any value, so I think there is only one possibility if `unix_ns` is needed, that is to store time in both formats, in two different fields.
Thoughts?

plugins/inputs/openstack/openstack.go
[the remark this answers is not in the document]
srebhan: @singamSrikar I share your concerns about cardinality, but let me clarify things

> 1) tags red,blue,green would create 3 new columns in the measurement and similarly there could be n different tags leading to n different columns increasing cardinality drastically (for eg:. hypervisor has about 30 tags)

For influxdb the additional tags are **not** a big issue as they are only stored once per series. The _space_ increase is negligible. However, the combinatorics is not, i.e. if you always have the same 30 or 1000 tags you only have one series and we only store this tag-set _once_ per series **not** once per datapoint.

> with tag value as empty string we would not be able to determine if a particular entry in the measurement has this tag or not.

You might be right here. We can circumvent it by setting a defined (maybe even user-defined) value.

Let me additionally comment on cardinality. In your current scheme, you will produce at least the same cardinality compared to separating out the tags. The cardinality is defined by the measurement _name_, the field _name_ and the tag **set** (consisting of tag _name_ **and** _value_). By joining the Openstack tags you create a certain metric tag _value_ resulting in a specific series. Even worse, if you do not guarantee the same tag order for each datapoint when joining you might produce __different__ series for the same tag-set! In the end the cardinality when joining data the way you do is the same as if you have separate tags. And worse...

If you join the Openinput tags, you basically render them useless. Users cannot query them (that's what tags are for in the first place), but the are degraded to simple display values. Imagine you have multiple stacks/hypervisors a first one with the tags `production, region-eu, customer_x` and another one with `region-us, customer_y, ha, production, priority_high` in you schema you will get two series, the same for my schema. However, I can then craft a query returning all `production` entries by `SELECT ... WHERE "openstack_tag_production" = "true"` or similar. You cannot do this as you would need to specify the complete sets **in the correct order** for filtering.
Furthermore, my scheme allows the user to _drop_ certain tags using the telegraf machinery...

plugins/inputs/openstack/openstack.go
srebhan: Ok, can you please then add a comment to the `if service == "services" {` part explaining why you are doing it the way you do it?

plugins/inputs/openstack/openstack.go
srebhan: What I meant was that the user can decide if he/she wants a human readable format or a machine readable format in influxdb. Currently, you get a "human" readable format and, if you want to do computations with it, you need to convert it in your application.

plugins/inputs/openstack/openstack.go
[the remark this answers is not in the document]
singamSrikar: Thank you @srebhan for clarifying.
I think we can query tags using regex in present config but yeah if additional tags are not a big issue then I'll implement tags as key-value pairs tags as keys and default value as 'true' and will also add tag_prefix and tag_value options.

plugins/inputs/openstack/openstack.go
singamSrikar: can we implement this as similar to openstack tags by using `security_groups` as prefix?

plugins/inputs/openstack/openstack.go
srebhan: Sorry, but I still do not see an issue here. Can you please help me, why you see a problem in creating e.g. a _field_ called `creation_time` and let the user decide if that field contains either a _string_ (human readable time as above) or a _number_ (machine readable time)? If you _know_ what "format" `creation_time` is in (e.g. `unix_ns`), then the option could be `human_readable_timestamps = true/false`. Where do you see an issue?

plugins/inputs/openstack/openstack.go
srebhan: Sure. Sounds good to me.

plugins/inputs/openstack/openstack.go
singamSrikar: The issue is if a user initially configured say `unix_ns` (default) and later chose to switch to human readable format then I think the user would not be able to store data in same DB as `creation_time` field would be of type `int` but the new data has `creation_time` field as type `string` (human readable).

srebhan: Well then the user can still create a new series by adding a tag or by changing the metric name. I think we should not work around these kind of issues in telegraf. You can still explicitly change the field name using a processor if you need to.

srebhan (commented):
Hey @singamSrikar, thanks for the updates! We are very close and I only have some minor cleanup requests. Please take a look at the comments.

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,997 @@
...
+// ContainsService indicates whether a particular service is enabled
+func ContainsService(t string, services map[string]services.Service) bool {
+	for _, service := range services {
+		if service.Type == t {
+			return true
+		}
+	}
+
+	return false
+}
```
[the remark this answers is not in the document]
singamSrikar: Yeah it used to be like that, but in previous comment you have said `please do only export functions that need to be called from outside of the plugin. This usually boils down to the interface-functions required by Telegraf` so I changed it to this.

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,997 @@
...
+			"id":            stack.ID,
+			"status_reason": stack.StatusReason,
+		}
+		if !o.HumanReadableTS {
+			fields["creation_time"] = stack.CreationTime.UnixNano()
+			fields["updated_time"] = stack.UpdatedTime.UnixNano()
+		} else {
+			fields["creation_time"] = stack.CreationTime.Format("2006-01-02T15:04:05.999999999Z07:00")
+			fields["updated_time"] = stack.UpdatedTime.Format("2006-01-02T15:04:05.999999999Z07:00")
+		}
```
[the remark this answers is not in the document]
singamSrikar: Yeah I thought of that but we need return type for the func and  `humanReadable` is of `string` and `unix` is of `int` datatypes, right?

plugins/inputs/openstack/openstack.go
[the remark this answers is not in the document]
srebhan: Yeah but the function is still [exported](https://tour.golang.org/basics/3) as it starts with an uppercase letter. I proposed to use a lowercase letter `containsService()` (see my suggestion above) but keep it as a struct member.

plugins/inputs/openstack/openstack.go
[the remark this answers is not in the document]
srebhan: You are correct, that's why you need to return an `interface{}` type (I updated my example) which is similar to a `(void *)` in C. As the fields in the metrics are _always_ of type `interface{}` you won't run into trouble.

srebhan:
@singamSrikar after resolving the `username`/`password` thing, please think about some testing. Can we for example mockup some server responses and see if we can gather the correct metrics?

singamSrikar:
> Looks good to me. Thanks @singamSrikar for the contribution. The only issue I see is the missing tests, but given the code I guess it is not easily possible to mock some stack responses, is it?

yeah, that is right. There are few applications which can mock the Openstack APIs but that would be an additional dependency.

plugins/inputs/openstack/openstack.go
```diff
@@ -0,0 +1,958 @@
...
+	if p, ok := o.openstackProjects[server.TenantID]; ok {
+		project = p.Name
+	}
+	tags["tenant_id"] = server.TenantID
+	tags["name"] = server.Name
+	tags["host_id"] = server.HostID
+	tags["status"] = strings.ToLower(server.Status)
+	tags["key_name"] = server.KeyName
+	tags["host_name"] = hostName
+	tags["project"] = project
```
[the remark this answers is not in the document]
singamSrikar: @powersj  @srebhan  I have updated the README with the recommendation for `modifiers` (`tagexclude`) and also added a hyperlink to modifiers documentation.

plugins/inputs/openstack/README.md
```diff
@@ -0,0 +1,358 @@
...
+* networking  v2
+* orchestration  v1
+
+## Configuration and Recommendations
+### Recommendations
+
+#### The recommended polling interval is 30 minutes.  This can be reduced on smaller deployments with a handful of VMs, but will need to be increased on hubs with hundreds or thousands of VMs as it can have a performance impact. This can be done using `interval` option.
+`interval`: Overrides the interval setting of the agent for the plugin. How often to gather this metric. Normal plugins use a single global interval, but if one particular input should be run less or more often, you can configure that here. 
+#### Highly recommend to use [modifiers](https://github.com/influxdata/telegraf/blob/master/docs/CONFIGURATION.md#modifiers) like tagexclude to discard unwanted tags to keep tag cardinality in check.
+#### Also recommend to poll different Openstack services at different interval as per your need. For example: If you want to poll "nova_services" more frequently than the rest then use seperate configuration like shown below
```
powersj, line 27: No need to put it under a header, you can replace the above four lines with this:

```
Due to the large number of unique tags that this plugin generates, in order to keep the cardinality down it is **highly recommended** to use [modifiers](https://github.com/influxdata/telegraf/blob/master/docs/CONFIGURATION.md#modifiers) like `tagexclude` to discard unwanted tags. 

For deployments with only a small number of VMs and hosts, a small polling interval (e.g. seconds-minutes) is acceptable. For larger deployments, polling a large number of systems will impact performance. Use the `interval` option to change how often the plugin is run:

`interval`: How often a metric is gathered. Setting this value at the plugin level overrides the global agent interval setting.

Also, consider polling OpenStack services at different intervals depending on your requirements. This will help with load and cardinality as well.
```

powersj (changes requested):
Almost there :)

singamSrikar:
> Almost there :)

Hey @powersj, I have added the changes as suggested to README doc, please take a look and let me know if any changes are required.

srebhan (approved):
Happy when you are... ;-)

MyaLongmire (approved):

yankcrime:
I can't help but feel like a nod to the original author of a large part of this plugin, @spjmurray, would've been the right thing to do here - especially since he unsuccessfully attempted to shepherd the [original PR](https://github.com/influxdata/telegraf/pull/2013) for several years before eventually giving up due to lack of traction on the InfluxData side.

srebhan:
Hey @yankcrime, I'm sorry for this lurking around so long (though I'm not with InfluxData). As you might have noticed I tried to reactivate the "old" PR, but unfortunately the original author did not react on any of my attempts. This PR was linked to the original one and I really appreciate the work @spjmurray put into this, but I have no means to reach him beside Github comments...
So sorry if this feels disrespectful, it never was meant like this!

Issue #7433, opened 2020-04-29
Actual real-world examples of SNMP


kiler129:
## Feature Request
SNMP is hard... and annoying. I looked at telegraf to add it to my stack hopefully improving the experience. However, so far all the "ancient" tools seems way easier to configure and use, probably because of the lack of documentation for telegraf snmp input plugin.

- There's some documentation at https://www.influxdata.com/integration/snmp/ which references "Cookbook Directory" and other seemingly useful things which are nowhere to be found
- There's another part of the docs in the repo itself: https://github.com/influxdata/telegraf/tree/master/plugins/inputs/snmp
- A lot of information is scattered around GH issues

I think the documentation lacks a one fundamental thing: a real-world example. The second documentation starts talking about that topic mentioning `input_juniper_srx.<device_hostname>.toml.conf. For example: input_juniper_srx.core-fw-1.toml.conf` part, but no further information is available.

### Proposal:
A single example configuration file (files?) should be provided with the SNMP plugin. The usage of TOML doesn't exactly make things easier. The 30+ years protocol is hard on its own ;)

In a real world there isn't gonna be one host with one set of fields. 

### Current behavior:
After literally 3h of googling and trying, as a person with well over ten years of experience, I gave up. This is not exactly encouraging.

What I wanted to achieve is seemingly simple:
 - Grab data from 5 devices
   - 1 of them is unique (i.e. all snmp parameters are different)
   - 2 of them are exactly the same (i.e. all field/table related stuff shouldn't be shared while stuff related to the connection like `agents` or `retries` must be different)
   - 3 of them are from the same vendor but with differences (i.e. there's a base set of field/table entries with additional ones per specific device type or even that specific device, or in simpler words how extension of "templates" should be handles)
 - Push all data to InfluxDB

From what I see now fields can be just dumped into one bucket. This is literally impossible to handle since all devices will have different MIBs. 

### Desired behavior:
Provide examples of real-world configs.

### Use case:
Making the barrier of entry lower - currently the bar is set so high that I'm having a hard time convincing myself to the stack, not to even mention the company :)

danielnelson:
First of all, I definitely agree with you that it can be quite difficult to setup this plugin and that there is plenty of room for improvement.  I'd love to see hear more about configuring with other tools, maybe we can adopt some of their ideas.

[...]

PR #9241, opened 2021-05-05
feat: telegraf to merge tables with different indexes

### LINKED ISSUE #3076: SNMP Plugin: Merging/joining two tables with seperate indexes
## Feature Request

Pulling data from one table with index into another tables values.

### Proposal:

Sometimes one table in SNMP has a field with contains an index value for another SNMP table and it's useful to have the first tables field/values but grab a field or more than one field from another table where an index exists in the first.

### Current behaviour:

Two tables can only be merged when the share the same index.

### Desired behaviour:

Merge or extract fields from two tables where the indexes are not the same but in one table a field points to the index in the second table.

### Use case: [Why is this important (helps with prioritising requests)]

Here is an example use case of collecting CPU statistics of Cisco networking devices. In this case from the first table (1.3.6.1.4.1.9.9.109.1.1.1.1) in that table .1.3.6.1.4.1.9.9.109.1.1.1.1.2 points to an index in the entity table 1.3.6.1.2.1.47.1.1.1.1 and I'm interested in pulling in the value of 1.3.6.1.2.1.47.1.1.1.1.7 which is the text description of the hardware (i.e. the CPU name)

So to pull out the CPU utilisation of 1 and 5 minutes plus the CPU description/name (preferably as a tag) from another table.

Here is an example I tried to use but this does not work as each table is not indexed the same.

```
  [[inputs.snmp.table]]
    name = "cpuTable"
    inherit_tags = [ "hostname" ]
    index_as_tag = true

    [[inputs.snmp.table.field]]
      name = " cpmCPUTotalPhysicalIndex"
      oid = ".1.3.6.1.4.1.9.9.109.1.1.1.1.2"
      is_tag = true

    [[inputs.snmp.table.field]]
      name = "cpmCPUTotal1minRev"
      oid = ".1.3.6.1.4.1.9.9.109.1.1.1.1.7"

    [[inputs.snmp.table.field]]
      name = "cpmCPUTotal5minRev"
      oid = ".1.3.6.1.4.1.9.9.109.1.1.1.1.8"

    [[inputs.snmp.table.field]]
      name = "entPhysicalName"
      oid = "1.3.6.1.2.1.47.1.1.1.1.7"
```

This is the index into the entity table where 16733769 is an index there.

```
snmpwalk -v 2c -c XXXXXXXX router1.example.com ".1.3.6.1.4.1.9.9.109.1.1.1.1.2"
iso.3.6.1.4.1.9.9.109.1.1.1.1.2.34 = INTEGER: 16733769
iso.3.6.1.4.1.9.9.109.1.1.1.1.2.82 = INTEGER: 39967263
iso.3.6.1.4.1.9.9.109.1.1.1.1.2.98 = INTEGER: 3480210
iso.3.6.1.4.1.9.9.109.1.1.1.1.2.514 = INTEGER: 24635790
iso.3.6.1.4.1.9.9.109.1.1.1.1.2.530 = INTEGER: 38114433

```
one of the CPU utilisation values I want to grab.

```
snmpwalk -v 2c -c XXXXXXXX router1.example.com ".1.3.6.1.4.1.9.9.109.1.1.1.1.7"
iso.3.6.1.4.1.9.9.109.1.1.1.1.7.34 = Gauge32: 11
iso.3.6.1.4.1.9.9.109.1.1.1.1.7.82 = Gauge32: 2
iso.3.6.1.4.1.9.9.109.1.1.1.1.7.98 = Gauge32: 2
iso.3.6.1.4.1.9.9.109.1.1.1.1.7.514 = Gauge32: 82
iso.3.6.1.4.1.9.9.109.1.1.1.1.7.530 = Gauge32: 7
```

Looking at the entity table the following index need to be grabbed and merged into the first table and the rest of the table ignored.

```
snmpwalk -v 2c -c XXXXXXXX router1.example.com "1.3.6.1.2.1.47.1.1.1.1.7" | egrep "16733769|39967263|3480210|24635790|38114433"
iso.3.6.1.2.1.47.1.1.1.1.7.3480210 = STRING: "0/6/* - cpu"
iso.3.6.1.2.1.47.1.1.1.1.7.16733769 = STRING: "0/2/* - host"
iso.3.6.1.2.1.47.1.1.1.1.7.24635790 = STRING: "0/RP0/* - host"
iso.3.6.1.2.1.47.1.1.1.1.7.38114433 = STRING: "0/RP1/* - host"
iso.3.6.1.2.1.47.1.1.1.1.7.39967263 = STRING: "0/5/* - cpu"
```

helotpl:
### Required for all PRs:

resolves #3076


helotpl:
!signed-cla

helotpl:
Solves https://github.com/influxdata/telegraf/issues/3076

.gitignore
```diff
@@ -3,3 +3,4 @@
 /telegraf.exe
 /telegraf.gz
 /vendor
+.vscode/launch.json
```
Hipska, line 6: Please don't add these editor specific in our project, but on your global `.gitignore` file.

plugins/inputs/snmp/snmp.go
```diff
@@ -3,6 +3,7 @@ package snmp
 import (
 	"bufio"
 	"bytes"
+	"errors"
```
Hipska, line 6: This import already exists, please do a rebase to current master and add your changes on that, this is creating merge conflicts.

Hipska (changes requested):
General idea looks good, but it seems you can only merge 1 table with another. What if I also want a third table to be merged with a fourth?

Hipska:
Please also update README file and default configuration. And add some tests for this specific case. You might want to set your PR as draft while working on it.

.gitignore
```diff
@@ -4,5 +4,3 @@
 /telegraf.exe
 /telegraf.gz
 /vendor
-.DS_Store
-process.yml
```
Hipska, line 8: Please don't touch the `.gitignore` file at all.

While I do agree `.DS_Store` should not be in there, please create a separate PR for that.

helotpl: First time doing git rebase. Well pull request is marked as draft so you should not be bothered.

Hipska: Okay, great! Let me know when you are ready.

helotpl: Ok managed to restore it. Thanks for noticing this.

helotpl:
> Please also update README file and default configuration. And add some tests for this specific case. You might want to set your PR as draft while working on it.

helotpl:
> General idea looks good, but it seems you can only merge 1 table with another. What if I also want a third table to be merged with a fourth?

I think now PR is ready for another review. Merging more tables could be possible, but it requires total rework of this patch, and configuration will be not friendly for users that want just one merge. I'm not up for this task, prepping unit tests, and other extra tasks makes this not worth my time. In my production environment I have two telegrafs running. Second one is only for tasks that require joins. Before this patch I was doing merges with kapacitor, that required using "pre-" tables with short retention policy, and final joined table. The problem is that kapacitor join buffers unjoined points indefinitely, and this consumes memory constantly. So before this PR I had to restart kapacitor every week.

plugins/inputs/snmp/README.md
```diff
@@ -203,6 +203,23 @@ One [metric][] is created for each row of the SNMP table.
       # translate = true
 ```
 
+### Two Table Join
+Snmp plugin can join two snmp tables that have different indexes. For this to work one table
+should have translation field that return index of second table as value. Examples
+of such fields are:
+ * Cisco portTable with translation field: ```1.3.6.1.4.1.9.5.1.4.1.1.11``` (portIfIndex),
```
Hipska, line 210:
```suggestion
 * Cisco portTable with translation field: `1.3.6.1.4.1.9.5.1.4.1.1.11` (portIfIndex),
```
Similar for the other occurrences in your PR

Hipska:
No, that is indeed not very easy to understand, but I think that is mainly because it evolved from the idea of only merging 2 tables for the complete agent.

While typing that, I'm just wondering if we just misunderstood each other? Lets rephrase:
If I create a second table definition `[[inputs.snmp.table]]`, will the `secondary_index` fields in there interfere with the ones from the first table? There is still confusion about this since you still didn't update the example config in the readme.

helotpl:
Ok you are right here. Looks like I was lucky that several tables that I joined did not have any conflict. Also I wonder if this will work when there are many hosts and translation index return different values.

helotpl:
Ok looking at code again, the `func (t Table) Build(gs snmpConnection, walk bool) (*RTable, error)` function works per table, so between tables conflict should not happen, because `secIdxTab` is local there. But this might not work if different hosts return different translation values.

helotpl:
On second thought `func (t Table) Build(gs snmpConnection, walk bool) (*RTable, error)` works per `smnpConnection` so different `secIdxTab` between hosts. Any idea how to test this in unit tests? They run `tb, err := tbl.Build(tsc, true)` so `tsc` is a connection here. I think this code is safe to work between different joined tables and between different hosts.

helotpl:
I don't know why these ci/circleci tests fail. Looks like it is no caused by my changes. Can you rerun these tests?

helotpl:
@Hipska I think this is still PR is still ready for review.

Hipska:
!retry-failed

helotpl:
> !retry-failed

All checks have passed
21 successful and 1 skipped checks

Hipska (changes requested):
Still no adapted default config provided for `inputs.snmp.table.field` in the section just above your new section called "Two Table Join". Your new section could as well be a paragraph in the "Table" section.

For the rest, all seems fine for me.

Hipska (approved):

plugins/inputs/snmp/README.md
```diff
@@ -201,8 +201,42 @@ One [metric][] is created for each row of the SNMP table.
...
+Snmp plugin can join two snmp tables that have different indexes. For this to work one table
+should have translation field that return index of second table as value. Examples
+of such fields are:
+ * Cisco portTable with translation field: `CISCO-STACK-MIB::portIfIndex`,
+which value is IfIndex from ifTable
+ * Adva entityFacilityTable with translation field: `ADVA-FSPR7-MIB::entityFacilityOneIndex`,
+which value is IfIndex from ifTable
+ * Cisco cpeExtPsePortTable with translation field: `CISCO-POWER-ETHERNET-EXT-MIB::cpeExtPsePortEntPhyIndex`,
+which value is index from entPhysicalTable
+
```
reimda, line 233: Could you add a configuration example to the docs that uses the new join functionality? I think an example would help beginners and provide a starting point for people to copy-paste and modify. Since you have already mentioned these three examples, maybe base the example on them too?

Is there anything users need to keep in mind when doing two joins? I mean a config that for example joins one table to ifTable and also joins another table to entPhysicalTable.

reimda (dismissed):
Looking good

plugins/inputs/snmp/README.md
Hipska: Maybe that is something that could be covered by adding an example config in the testdata folder as explained in #7433

helotpl: Added one example, but it presents both tables before join, and after join. Adding all threes examples would take too much space.

helotpl:
@reimda is there anything you would like to change?

plugins/inputs/snmp/README.md
```diff
@@ -201,6 +201,113 @@ One [metric][] is created for each row of the SNMP table.
...
+[[inputs.snmp.table.field]]
+name = "EntPhyIndex"
+oid = "CISCO-POWER-ETHERNET-EXT-MIB::cpeExtPsePortEntPhyIndex"
+secondary_index_table = true    # enables joining
+
+[[inputs.snmp.table.field]]
+name = "EntPhysicalName"
+oid = "ENTITY-MIB::entPhysicalName"
+secondary_index_use = true      # this tag is indexed from secondary table
+is_tag = true
```
ssoroka, line 303: does it make sense to reference the table by name instead of just enabling this via booleans?

Hipska: Oh yeah, that also solves the issue if you have multiple of that kind of tables on a host.

`secondary_index_table = "entityTable"`

I love that idea. What do you think @helotpl?

helotpl: lines 294:303 are `[[inputs.snmp.table.fields]]` that are inside single named `[[inputs.snmp.table]]`, there is no way to mix two joins between different `[[inputs.snmp.table]]`, this gives no advantage. We could try to talk about two joins inside single `[[inputs.snmp.table]]`, but this requires rework for this patch not just altering configuration. If you really want this, then perhaps someone else can work on this.

Hipska: I still think you're not understanding what I mean, ignore it.

helotpl: In that case could you present full configuration for quoted example with your suggestions applied?

PR #9244, opened 2021-05-05
Cleanup metric type docs

Hipska:
It's not deprecated.


reimda (approved):
Thanks @Hipska for clarifying this. The type is necessary for prometheus so it isn't going away.

ssoroka (changes requested):
this is good, but we should also mention that this is only used by certain plugins, primarily prometheus, and statsd.

Hipska:
!retry-failed

ssoroka (approved):

Issue #8877, opened 2021-02-18
Graphite serializer doesn't respect tag guidelines


Hipska:
The filter is too restrictive compared to what is allowed.

### Steps to reproduce:

```
metric,tag=foo bar,foo bar=baz field=123 1234567890000000
```
### Expected behavior:

```
metric.field;tag=foo bar;foo bar=baz 123 1234567890
```

### Actual behavior:

```
metric.field;tag=foo_bar;foo_bar=baz 123 1234567890
```

### Additional info:

https://graphite.readthedocs.io/en/latest/tags.html
> Tag names must have a length >= 1 and may contain any ascii characters except `;!^=`. Tag values must also have a length >= 1, they may contain any ascii characters except `;` and the first character must not be `~`.

PR #9249, opened 2021-05-07
Allow more characters in graphite tags

### LINKED ISSUE #9248: Graphite Tag support replaces characters that should be allowed
## Feature Request

### Proposal:
I'd like to see Telegraf support the full power of tags when using the Graphite data format, with Tag Support turned on, by passing them almost unchanged from the inputs, only stripping out charaters that are strictly disallowed.

### Current behavior:
When using any output that supports Graphite with Tag Support enabled, Telegraf was stripping out characters that should actually be allowed. THis is caused by the fact that the serializer was using the same sanitizing code for tag names/values *and* the fields, and measurements. In reality, the rules are different for each aspect.

Metrics like the following:
 ```
vsphere_host_datastore,cluster=PROD\ -\ SAP,datacenter=NA\ -\ CAN1,datastoreID=5f87613a-f628220e-04b2-1402ec8723bc,esxhost=can1vm32.can.na.int.corp.com,vcenter=vc.int.corp.com datastoreIops_average=0i,datastoreMaxQueueDepth_latest=64i  1620420360000000000
> vsphere_host_disk,cluster=PROD\ -\ SAP,datacenter=NA\ -\ CAN1,datastoreID=5f87613a-f628220e-04b2-1402ec8723bc,esxhost=can1vm32.can.na.int.corp.com,lunID=naa.60060e80129f6d0050409f6d00000101,vcenter=vc.int.corp.com busResets_summation=0i,commandsAborted_summation=0i 1620420360000000000
```

Become
```
vsphere_host_datastore.datastoreIops_average;cluster=PROD_-_SAP,datacenter=NA_-_CAN1,datastoreID=5f87613a-f628220e-04b2-1402ec8723bc,esxhost=can1vm32.can.na.int.corp.com,vcenter=vc.int.corp.com 0 1620420360000000
vsphere_host_datastore.datastoreMaxQueueDepth_latest;cluster=PROD_-_SAP,datacenter=NA_-_CAN1,datastoreID=5f87613a-f628220e-04b2-1402ec8723bc,esxhost=can1vm32.can.na.int.corp.com,vcenter=vc.int.corp.com 64 1620420360000000
vsphere_host_disk.busResets_summation;cluster=PROD_-_SAP,datacenter=NA_-_CAN1,datastoreID=5f87613a-f628220e-04b2-1402ec8723bc,esxhost=can1vm32.can.na.int.corp.com,lunID=naa.60060e80129f6d0050409f6d00000101,vcenter=vc.int.corp.com 0 1620420360000000
vsphere_host_disk.commandsAborted_summation;cluster=PROD_-_SAP,datacenter=NA_-_CAN1,datastoreID=5f87613a-f628220e-04b2-1402ec8723bc,esxhost=can1vm32.can.na.int.corp.com,lunID=naa.60060e80129f6d0050409f6d00000101,vcenter=vc.int.corp.com 0 1620420360000000
```
### Desired behavior:

In reality, the [Graphite spec](https://graphite.readthedocs.io/en/latest/tags.html#carbon) says that the disallowed characters for tags are much simpler than currently setup for (`;!^=` for tag names, `;` for tag values).

So the above metrics would become:

```
vsphere_host_datastore.datastoreIops_average;cluster=PROD - SAP,datacenter=NA - CAN1,datastoreID=5f87613a-f628220e-04b2-1402ec8723bc,esxhost=can1vm32.can.na.int.corp.com,vcenter=vc.int.corp.com 0 1620420360000000
vsphere_host_datastore.datastoreMaxQueueDepth_latest;cluster=PROD - SAP,datacenter=NA - CAN1,datastoreID=5f87613a-f628220e-04b2-1402ec8723bc,esxhost=can1vm32.can.na.int.corp.com,vcenter=vc.int.corp.com 64 1620420360000000
vsphere_host_disk.busResets_summation;cluster=PROD - SAP,datacenter=NA - CAN1,datastoreID=5f87613a-f628220e-04b2-1402ec8723bc,esxhost=can1vm32.can.na.int.corp.com,lunID=naa.60060e80129f6d0050409f6d00000101,vcenter=vc.int.corp.com 0 1620420360000000
vsphere_host_disk.commandsAborted_summation;cluster=PROD - SAP,datacenter=NA - CAN1,datastoreID=5f87613a-f628220e-04b2-1402ec8723bc,esxhost=can1vm32.can.na.int.corp.com,lunID=naa.60060e80129f6d0050409f6d00000101,vcenter=vc.int.corp.com 0 1620420360000000
```

The difference is subtle, but the `cluster` and `datacenter` tags no longer have had their spaces replaced with underscores. This is a very simplistic example, but there are many other characters that are being replaced needlessly.

### Use case: 
This is important because it makes dashboards and front-ends have cleaner looking data in the fields and drop-downs.

### PR DESCRIPTION
**Pull request by G-regL**
### Required for all PRs:

resolves #9248

Added a new function to the Graphite Serializer that will use a specifc sanitzer for tags, along with supporting changes elsewhere.

As it's a relatively breaking change, it's hidden behind a (currently poorly named) feature flag called "graphite_tag_new_sanitize". That name needs to be changed, but I just can't for the life of me think of a better one.

### CODE REVIEW

File: plugins/outputs/graphite/graphite.go [hunk 1]
```diff
@@ -15,8 +15,9 @@ import (
 )
 
 type Graphite struct {
-	GraphiteTagSupport bool   `toml:"graphite_tag_support"`
-	GraphiteSeparator  string `toml:"graphite_separator"`
+	GraphiteTagSupport     bool   `toml:"graphite_tag_support"`
+	GraphiteTagNewSanitize bool   `toml:"graphite_tag_new_sanitize"`
```
helenosheaa, line 19: Instead of a bool option, how about a string option of sanitize which defaults to "strict" (the previous behavior) just allows for more future extension

plugins/serializers/graphite/README.md
```diff
@@ -64,4 +66,7 @@ cpu_usage_user;cpu=cpu-total;dc=us-east-1;host=tars 0.89 1455320690
 cpu_usage_idle;cpu=cpu-total;dc=us-east-1;host=tars 98.09 1455320690
 ```
 
+When the `graphite_tag_new_sanitize` option is enabled, the tags support the
+full range of characters defined by Graphite.
```
helenosheaa, line 70: could be worth expanding what the two sanitize settings are here

plugins/serializers/graphite/graphite.go
```diff
@@ -28,6 +28,9 @@ var (
 	)
 
 	fieldDeleter = strings.NewReplacer(".FIELDNAME", "", "FIELDNAME.", "")
+
+	allowedCharsTagName  = regexp.MustCompile(`[^ "-:\<>-\]_a-~\p{L}]`)
```
helenosheaa, line 32: Should we rename the other allowedChars on line 19 to strictAllowedChars to distinguish the two

plugins/serializers/graphite/graphite.go
```diff
@@ -248,14 +252,20 @@ func SerializeBucketNameWithTags(
...
 		if k == "name" {
 			k = "_name"
 		}
-		tagsCopy = append(tagsCopy, sanitize(k+"="+v))
+		if tagNewSanitize {
+			tagsCopy = append(tagsCopy, sanitizeTag(k, v))
+		} else {
+			tagsCopy = append(tagsCopy, sanitize(k+"="+v))
+		}
+
```
helenosheaa, line 268: remove extra empty line here to appease the linter

plugins/serializers/graphite/graphite.go
```diff
@@ -248,14 +252,20 @@ func SerializeBucketNameWithTags(
...
+	tagNewSanitize bool,
 ) string {
 	var out string
 	var tagsCopy []string
 	for k, v := range tags {
 		if k == "name" {
 			k = "_name"
 		}
-		tagsCopy = append(tagsCopy, sanitize(k+"="+v))
+		if tagNewSanitize {
```
helenosheaa, line 263: Change this to check the string value instead

plugins/serializers/graphite/graphite.go
```diff
@@ -316,3 +326,12 @@ func sanitize(value string) string {
 	// Replace any remaining illegal chars
 	return allowedChars.ReplaceAllLiteralString(value, "_")
 }
+
+func sanitizeTag(name string, value string) string {
```
helenosheaa, line 330: Again wondering about clarity of naming thats more descriptive of the difference strictSanitize for the one above and sanitize for the new one perhaps?

helenosheaa (changes requested):
Overall looks good. Just a few comments around different naming for clarity, linter errors and not adding a bool option to allow for future extension.

plugins/outputs/graphite/graphite.go
G-regL: Yeah, that makes sense. Any thoughts on the name for the mode at the heart of this PR? Maybe "compatible", or "full"?

helenosheaa: I think "compatible" makes sense as it's in line with the graphite spec

helenosheaa:
I think you could link this PR to #8877 also

plugins/serializers/graphite/README.md
G-regL: Added a bunch of text that describes what each "mode" does.

plugins/serializers/graphite/graphite.go
G-regL: Named each variable for what mode they apply to.

plugins/serializers/graphite/graphite.go
G-regL: Named each Sanitize function to match what it does (`strict` versus `compatible`).

helenosheaa (approved):
Looks good, thanks for making those changes so quickly!

G-regL:
No problem. I really want this feature, as I'm about to go through a massive new deployment and would love tag values to look right. Any idea what release it will apply to?

helenosheaa:
I think it will make it into the next feature release v1.19.0 which will be in June

plugins/serializers/graphite/graphite.go
```diff
@@ -308,11 +316,18 @@ func buildTags(tags map[string]string) string {
...
 	value = dropChars.Replace(value)
 	// Replace any remaining illegal chars
-	return allowedChars.ReplaceAllLiteralString(value, "_")
+	return strictAllowedChars.ReplaceAllLiteralString(value, "_")
+}
+
+func compatibleSanitize(name string, value string) string {
+	name = compatibleAllowedCharsName.ReplaceAllLiteralString(name, "_")
+	value = compatibleAllowedCharsValue.ReplaceAllLiteralString(value, "_")
+	value = regexp.MustCompile(`^[~]*(.*)`).FindStringSubmatch(value)[1]
```
ssoroka, line 331: please move this out to a variable so it gets checked earlier, is named, and is reused. I think it's also possible for this function to not return the match you're expecting, so maybe we can't assume the `[1]` match exists?

ssoroka (changes requested):
Looks good. the regexes are a bit dense but the test coverage looks decent. One comment from me.

plugins/serializers/graphite/graphite.go
G-regL: Yeah, the regexes are a bit dense, but it's the only way I can think to effieciently exclude the disallowed characters.
You'd like the last regex (checking for leading tilde) to be broken out to a variable? 
I've tested the function with inputs where the value has, and doesn't have, tildes and it seems to always return what I expect. I'll try some more to see if it works properly.

G-regL: I [tested](https://play.golang.org/p/E-loRxX2Zeo) a whole pile of possible values that might break the existance of the `[1]`, and I wasn't able to do so in any case. Even providing an empty value doesn't do it, probably because the matching group is so open, that it accepts anything.

Hipska (approved):

ssoroka (approved):

PR #9250, opened 2021-05-07
update HTTP v2 listener readme

sjwang90:
default config should _not_ be set to `data_format = "influx"` since if users are reading in influx LP they should be using the `inputs.influxdb_listener`. 
- Add note about using `inputs.influxdb_listener
- Change data format in default config to `data_format.= "json"`


sspaink (approved):

reimda:
This PR was merged directly to the release-1.18 branch. Usually we merge PRs to master and only cherry pick bugfixes to release branches.  I've cherry picked this to master and I'm going to remove it from release-1.18.

sspaink:
@reimda thanks for fixing this! I didn't notice it didn't go into master

PR #9256, opened 2021-05-11
feat: Modbus connection settings (serial)

### LINKED ISSUE #8225: input.modbus TCP connection handling
It seems that after successfully gathering data over modbus TCP, the connection is not released.
This makes the telegraf plugin have unique access to the modbus device, and other connections fail even if telegraf is not actively gathering data but it's idle waiting for the next interval.
Can this be fixed?

### LINKED ISSUE #9253: Modbus Serial port timeout for discontinued address poll
### Relevant telegraf.conf:

``` [[inputs.modbus]]
#   ## Connection Configuration
#   ##
#   ## The plugin supports connections to PLCs via MODBUS/TCP or
#   ## via serial line communication in binary (RTU) or readable (ASCII) encoding
#   ##
#   ## Device name
trace_connection = true
close_connection = false
name = "BM"
slave_id = 10
timeout = "3s"
busy_retries =2
busy_retries_wait = "2s"

#   ## Serial (RS485; RS232)
   controller = "file:///dev/ttyAMA0"
   baud_rate = 9600
   data_bits = 8
   parity = "E"
   stop_bits = 1
   transmission_mode = "RTU"

holding_registers = [
        { name = "Watts Total-CH1",     byte_order = "CDAB",   data_type = "FLOAT32-IEEE", scale=1.0,address = [4100,4101]},
        { name = "Watts Total-CH2",     byte_order = "CDAB",   data_type = "FLOAT32-IEEE", scale=1.0,address = [4202,4203]},

   ]


```

### System info:

Ubuntu, linux , arm64

### Steps to reproduce:
1. Telegraf modbus module configured to read a serial device /dev/tty*
2. Create holding register address value is not continuous
```
{ name = "Watts Total-CH1",     byte_order = "CDAB",   data_type = "FLOAT32-IEEE", scale=1.0,address = [4100,4101]},
{ name = "Watts Total-CH2",     byte_order = "CDAB",   data_type = "FLOAT32-IEEE", scale=1.0,address = [4202,4203]},
```
3. Run the telegraf in debug mode
`--debug log`
`2021-05-10T19:35:45Z D! [agent] Starting service inputs`
`2021-05-10T19:35:50Z D! [inputs.modbus] modbus: send 0a 03 10 04 00 02 80 71` **<-- First command at 9:35:50**
`2021-05-10T19:35:50Z D! [inputs.modbus] modbus: recv 0a 03 04 00 00 00 00 40 f3`
`2021-05-10T19:35:50Z D! [inputs.modbus] got holding@4100[2]: [0 0 0 0]`
`2021-05-10T19:35:50Z D! [inputs.modbus] ~CH1 offset 0 with len 4: [0 0 0 0] --> 0` **<-- First data received at 9:35:50**
`2021-05-10T19:35:50Z D! [inputs.modbus] modbus: send 0a 03 10 6a 00 02 e1 ac` **<-- Next Command send at 9:35:50**
`2021-05-10T19:35:54Z D! [inputs.modbus] modbus: recv` 
`2021-05-10T19:35:54Z E! [inputs.modbus] Error in plugin: serial: timeout `**<-- timeout received after 3 sec**

### Expected behavior:

Serial port subsequent command will need a delay from the time data is read. 

`2021-05-10T19:35:45Z D! [agent] Starting service inputs`
`2021-05-10T19:35:50Z D! [inputs.modbus] modbus: send 0a 03 10 04 00 02 80 71` **<-- First command at 9:35:50**
`2021-05-10T19:35:50Z D! [inputs.modbus] modbus: recv 0a 03 04 00 00 00 00 40 f3`
`2021-05-10T19:35:50Z D! [inputs.modbus] got holding@4100[2]: [0 0 0 0]`
`2021-05-10T19:35:50Z D! [inputs.modbus] ~CH1 offset 0 with len 4: [0 0 0 0] --> 0` **<-- First data received at 9:35:50**
`2021-05-10T19:35:51Z D! [inputs.modbus] modbus: send 0a 03 10 6a 00 02 e1 ac` **<-- Next Command send at 9:35:51**
`2021-05-10T19:35:54Z D! [inputs.modbus] modbus: recv 0a 03 04 00 00 00 00 40 f3`
`2021-05-10T19:35:54Z E! [inputs.modbus] ~CH2 offset 0 with len 4: [0 0 0 0] --> 0 `**<-- second data received at 9:35:54**

### Additional info:
Serial poll  rate should be greater than 100ms , found in mbpoll 
`-l #          Poll rate in ms, ( > 100, 1000 is default)`

Using mbpoll verified command send immediately after the poll completion, generated connection timeout subsequent command respond with data. device may not be ready to receive the next command 
```
root@ubuntu:/etc/telegraf# mbpoll -b 9600 -p even -m rtu -a 10 -r 4201,4203,4205  -t 4:float /dev/ttyAMA0 -1 -v
debug enabled
Set device=/dev/ttyAMA0
mbpoll 1.4-25 - FieldTalk(tm) Modbus(R) Master Simulator
Copyright © 2015-2019 Pascal JEAN, https://github.com/epsilonrt/mbpoll
This program comes with ABSOLUTELY NO WARRANTY.
This is free software, and you are welcome to redistribute it
under certain conditions; type 'mbpoll -w' for details.

Opening /dev/ttyAMA0 at 9600 bauds (E, 8, 1)
Set response timeout to 1 sec, 0 us
Protocol configuration: Modbus RTU
Slave configuration...: address = [10]
                        start reference = [4201,4203,4205]
Communication.........: /dev/ttyAMA0,       9600-8E1
                        t/o 1.00 s, poll rate 1000 ms
Data type.............: 32-bit float (little endian), output (holding) register table

-- Polling slave 10...
[0A][03][10][68][00][02][40][6C]
Waiting for a confirmation...
<0A><03><04><00><00><00><00><40><F3>
[4201]:         0
[0A][03][10][6A][00][02][E1][AC]
Waiting for a confirmation...
ERROR Connection timed out: select
Read output (holding) register failed: Connection timed out
[0A][03][10][6C][00][02][01][AD]
Waiting for a confirmation...
<0A><03><04><00><00><00><00><40><F3>
[4205]:         0


```

srebhan:
### Required for all PRs:

resolves #9253
resolves #8225

With this PR we add options to enable debugging of modbus connections and to configure the connection handling more fine-grained.
With `close_connection_after_gather` you can force the plugin to close and reopen the connection in each gather cycle. This is especially important for long intervals and/or modbus devices with a restricted number of simultaneous connections.
The second option `pause_between_requests` allows to add a pause between two requests (e.g. when querying non-consecutive registers). This is required by some serial devices (see Issue #9253).

### DIFF
--- a/plugins/inputs/modbus/README.md
+++ b/plugins/inputs/modbus/README.md
@@ -44,7 +44,7 @@ Registers via Modbus TCP or Modbus RTU/ASCII.
 
   ## Trace the connection to the modbus device as debug messages
   ## Note: You have to enable telegraf's debug mode to see those messages!
-  # trace_connection = false
+  # debug_connection = false
 
   ## Measurements
   ##
@@ -102,6 +102,11 @@ Registers via Modbus TCP or Modbus RTU/ASCII.
     ## from multiple instances you might want to only stay connected during gather and disconnect afterwards.
     # close_connection_after_gather = false
 ```
+
+### Notes
+You can debug Modbus connection issues by enabling `debug_connection`. To see those debug messages Telegraf has to be started with debugging enabled (i.e. with `--debug` option). Please be aware that connection tracing will produce a lot of messages and should **NOT** be used in production environments.
+
+Please use `pause_between_requests` with care. Especially make sure that the total gather time, including the pause(s), does not exceed the configured collection interval. Note, that pauses add up if multiple requests are sent!
 
 ### Metrics
 


rwalli:
workarounds = { close_connection_after_gather = true }

is working for me.

srebhan:
@rwalli btw using `workarounds.close_connection_after_gather = true` should also work according to the TOML syntax or
```toml
[inputs.modbus.workarounds]
    close_connection_after_gather = true
```
:-)

Thanks for testsing!

plugins/inputs/modbus/README.md
```diff
@@ -38,6 +38,18 @@ Registers via Modbus TCP or Modbus RTU/ASCII.
...
+
+  ## Enable workarounds required by some devices to work correctly
+  # workarounds = {
+      ## Pause between read requests sent to the device. This might be necessary for (slow) serial devices.
+      # pause_between_requests = "0ms"
+      ## Close the connection after every gather cycle. Usually the plugin closes the connection after a certain
+      ## idle-timeout, however, if you query a device with limited simultaneous connectivity (e.g. serial devices)
+      ## from multiple instances you might want to only stay connected during gather and disconnect afterwards.
+      # close_connection_after_gather = false
+	# }
```
Hipska, line 52:
```suggestion
  # }
```

plugins/inputs/modbus/modbus.go
```diff
@@ -97,6 +104,24 @@ const sampleConfig = `
...
+
+  ## Enable workarounds required by some devices to work correctly
+  # workarounds = {
+      ## Pause between read requests sent to the device. This might be necessary for (slow) serial devices.
+      # pause_between_requests = "0ms"
+			## Close the connection after every gather cycle. Usually the plugin closes the connection after a certain
+			## idle-timeout, however, if you query a device with limited simultaneous connectivity (e.g. serial devices)
+      ## from multiple instances you might want to only stay connected during gather and disconnect afterwards.
+			# close_connection_after_gather = false
+	# }
```
Hipska, line 123:
```suggestion
  # workarounds = {
      ## Pause between read requests sent to the device. This might be necessary for (slow) serial devices.
      # pause_between_requests = "0ms"
      ## Close the connection after every gather cycle. Usually the plugin closes the connection after a certain
      ## idle-timeout, however, if you query a device with limited simultaneous connectivity (e.g. serial devices)
      ## from multiple instances you might want to only stay connected during gather and disconnect afterwards.
      # close_connection_after_gather = false
    # }
```

Hipska (approved):
Any reason to go for the `workarounds = { }` notation instead of `[inputs.modbus.workarounds]`?

plugins/inputs/modbus/README.md
rwalli: if i remember correctly "[inputs.modbus.workarounds]" only worked at the end of a "[[inputs.modbus]]" section.

srebhan:
> Any reason to go for the `workarounds = { }` notation instead of `[inputs.modbus.workarounds]`?

No special intention behind it. It just matches the register definition a bit better, but if you insist on the `[inputs.modbus.workarounds]` notation for the sample it's also fine with me. However, @rwalli is right, you cannot configure other options afterwards as everything is then assigned to the `workarounds` table. That's the downside of having no grouping in this part of the TOML syntax.

Hipska:
It seems like the `[]` notation seems more preferred by telegraf/InfluxDB?

srebhan:
@Hipska I agree, however using this notation will probably lead to problems for less-experienced users as this
```toml
[[inputs.modbus]]
    ...
    holding_registers = [
        { name = "power_factor", byte_order = "AB",   data_type = "FIXED", scale=0.01,  address = [8]},
        { name = "voltage",      byte_order = "AB",   data_type = "FIXED", scale=0.1,   address = [0]},
        { name = "energy",       byte_order = "ABCD", data_type = "FIXED", scale=0.001, address = [5,6]}
    ]

    [inputs.modbus.workarounds]
        pause_between_requests = "100ms"
```
will work just find but

[[inputs.modbus]]
    ...
    [inputs.modbus.workarounds]
        pause_between_requests = "100ms"

    holding_registers = [
        { name = "power_factor", byte_order = "AB",   data_type = "FIXED", scale=0.01,  address = [8]},
        { name = "voltage",      byte_order = "AB",   data_type = "FIXED", scale=0.1,   address = [0]},
        { name = "energy",       byte_order = "ABCD", data_type = "FIXED", scale=0.001, address = [5,6]}
    ]
```
will fail as now `holding_registers` actually becomes `inputs.modbus.workarounds.holding_registers` instead of `inputs.modbus.holding_registers`. In contrast

[[inputs.modbus]]
    ...
    workarounds = {
        pause_between_requests = "100ms"
    }

    holding_registers = [
        { name = "power_factor", byte_order = "AB",   data_type = "FIXED", scale=0.01,  address = [8]},
        { name = "voltage",      byte_order = "AB",   data_type = "FIXED", scale=0.1,   address = [0]},
        { name = "energy",       byte_order = "ABCD", data_type = "FIXED", scale=0.001, address = [5,6]}
    ]
```
will work in both positions due to the explicit start `{` and end `}` markers.

Should we really put the fragile formatting in the example?

Hipska:
I agree, but just saying it has always been like this. See for example tag/field filtering documentation…

plugins/inputs/modbus/README.md
```diff
@@ -142,7 +156,15 @@ In case you get an `exception '2' (illegal data address)` error you might try to
...
+If your data still looks corrupted, please post your configuration, error message and/or the output of `byte_order="ABCD" data_type="UINT32"` to one of the telegraf support channels (forum, slack or as issue).
+
+#### Workarounds
+Some Modbus devices need special read characteristics when reading data and will fail otherwise. For example, there are certain serial devices that need a certain pause between register read requests. Others might only offer a limited number of simultaneously connected devices, like serial devices or some ModbusTCP devices. In case you need to access those devices in parallel you might want to disconnect immediately after the plugin finished reading.
+
+To allow this plugin to also handle those "special" devices there is the `workarounds` configuration options. In case your documentation states certain read requirements or you get read timeouts or other read errors you might want to try one or more workaround options.
+If you find that workarounds are required for your device, please let us know.
```
Hipska, line 165: I would put something like: "If you find that other/more workarounds ..."

Hipska (approved):

plugins/inputs/modbus/modbus.go
```diff
@@ -88,19 +95,23 @@ const sampleConfig = `
...
+
   ## Serial (RS485; RS232)
   # controller = "file:///dev/ttyUSB0"
   # baud_rate = 9600
   # data_bits = 8
   # parity = "N"
   # stop_bits = 1
+  # transmission_mode = "RTU"
+
+  ## Trace the connection to the modbus device as debug messages
```
reimda, line 107: How is this different than running all of telegraf with debug logging? Is it the same output, but just for this plugin?

plugins/inputs/modbus/modbus.go
```diff
@@ -441,6 +497,11 @@ func (m *Modbus) collectFields(acc telegraf.Accumulator, timestamp time.Time, ta
 	}
 }
 
+// Implement the logger interface of the modbus client
+func (m *Modbus) Printf(format string, v ...interface{}) {
+	m.Log.Debugf(format, v...)
```
reimda, line 502: These won't show up unless the user also has the --debug command line option.

plugins/inputs/modbus/modbus.go
```diff
@@ -345,6 +386,9 @@ func (m *Modbus) gatherRequestsCoil(requests []request) error {
 			request.fields[i].value = uint16((bytes[idx] >> bit) & 0x01)
 			m.Log.Debugf("  field %s with bit %d @ byte %d: %v --> %v", field.name, bit, idx, (bytes[idx]>>bit)&0x01, request.fields[i].value)
 		}
+
+		// Some (serial) devices require a pause between requests...
+		time.Sleep(time.Until(nextRequest))
```
reimda, line 391: There's an opportunity here for the user to configure more sleep time than the total collection interval. Should we warn them if for example they have 10 requests with 1 second sleep between and that won't fit into a 5 second collection interval?

(same goes for discrete, holding, input, etc. in addition to coil)

plugins/inputs/modbus/modbus.go
srebhan: My understanding is that the usual `debug` mode enables all debug output concerning the **telegraf** plugin. However, with this option you can enable the (connection) debug output of the **underlying library**. This underlying library might (and actually does) output a huge amount of messages which buries the "usual" debugging output of telegraf under tons of raw modbus connection messages rendering it useless. Nevertheless, it is sometimes useful to also see those underlying raw data to debug (connection) problems.
Therefore the idea of this option is to enable another debug level beyond the usual telegraf debugging.

plugins/inputs/modbus/modbus.go
srebhan: Right, therefore the text for the option enabling this output states

>  ## Trace the connection to the modbus device as debug messages

suggesting that you need both, `--debug` and the `debug_connection` option enabled.

Do you have a suggestion on how to improve either the text or the handling? As discussed elsewhere, we might also implement a telegraf `trace` level, but IIRC the decision was to stick with debug and individual options...

plugins/inputs/modbus/modbus.go
srebhan: The correct solution would be to allow the plugin to read the configured interval of the plugin. This would also ease the life of plugins that need to specify the collection interval for reading.

The problem with your suggestion is that there might be long intervals (e.g. 15 minutes) with long requests adding up to e.g. 1 minute rendering this a valid use-case. As long as we cannot know the collection interval we cannot decide on the validity of the sleep interval, potentially either classifying valid use-cases as errors or vice versa.

plugins/inputs/modbus/modbus.go
reimda: Let's add something to the readme to remind users that when they enable debug_connnection, they need to enable debug. Maybe a link to https://github.com/influxdata/telegraf/blob/master/docs/CONFIGURATION.md#agent. That's good enough for me

plugins/inputs/modbus/modbus.go
reimda: It would be nice for the plugin to have some way to learn what its collection interval is, but since that doesn't exist now, let's also document this so the user is aware of it. Something like, "make sure not to add pauses that are too long or the plugin will miss the next collection interval"

Hipska: That's also a thing I believe is missing from the snmp input docs (or any other that has timeouts)

srebhan: @reimda I agree and will add some words there.

PR #9269, opened 2021-05-13
Prometheus remote write parser readme fix path

helenosheaa:


sjwang90 (approved):
i before e except after c!

PR #9271, opened 2021-05-13
feat: set user agent when scraping prom metrics

russorat:
### Required for all PRs:

this change updates the user agent used when scraping a prometheus endpoint from the generic Go-Http-Client to Telegraf/1.18 used in other plugins (the influxdb_v2 output)

### DIFF
--- a/plugins/inputs/prometheus/prometheus.go
+++ b/plugins/inputs/prometheus/prometheus.go
@@ -435,7 +435,7 @@ func (p *Prometheus) gatherURL(u URLAndAddress, acc telegraf.Accumulator) error 
 
 func (p *Prometheus) addHeaders(req *http.Request) {
 	for header, value := range p.headers {
-		req.Header.Set(header, value)
+		req.Header.Add(header, value)
 	}
 }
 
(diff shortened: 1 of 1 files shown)


plugins/inputs/prometheus/prometheus.go
```diff
@@ -273,6 +275,11 @@ func (p *Prometheus) Gather(acc telegraf.Accumulator) error {
 			return err
 		}
 		p.client = client
+		p.headers = map[string]string{
+			"User-Agent": internal.ProductToken(),
+			"Accept":     acceptHeader,
+		}
+
```
Hipska, line 282:
```suggestion
```

Hipska (approved):

Hipska (approved):

plugins/inputs/prometheus/prometheus.go
```diff
@@ -427,6 +433,12 @@ func (p *Prometheus) gatherURL(u URLAndAddress, acc telegraf.Accumulator) error
 	return nil
 }
 
+func (p *Prometheus) addHeaders(req *http.Request) {
+	for header, value := range p.headers {
+		req.Header.Set(header, value)
```
reimda, line 438: The code called req.Header.Add() before but now it calls req.Header.Set(). Any reason for the change? If not, let's switch back to Add to be compatible with the previous code.

reimda (changes requested):

plugins/inputs/prometheus/prometheus.go
russorat, line 438: I was just copying some other addHeaders code. all good.

```suggestion
		req.Header.Add(header, value)
```

reimda (approved):

PR #9274, opened 2021-05-17
inputs.ping: Add an option to specify packet size

### LINKED ISSUE #9222: Add an option for packet size to the ping plugin
### Use case: Sometimes, ICMP pings work just fine when the packet size is small. There is value in being able to change the size to help diagnose network issues.

rajiv-k:
### Required for all PRs:

resolves #9222 

This patch adds an option to specify the number of data bytes to be sent.
This corresponds to the "-s" in the ping(8) command.

It is only applicable for the "native" method.

### DIFF
--- a/plugins/inputs/ping/ping.go
+++ b/plugins/inputs/ping/ping.go
@@ -18,6 +18,10 @@ import (
 	"github.com/influxdata/telegraf/plugins/inputs"
 )
 
+const (
+	defaultPingDataBytesSize = 56
+)
+
 // HostPinger is a function that runs the "ping" function using a list of
 // passed arguments. This can be easily switched with a mocked ping function
 // for unit test purposes (see ping_test.go)
@@ -75,7 +79,7 @@ type Ping struct {
 	Percentiles []int
 
 	// Packet size
-	Size int
+	Size *int
 }
 
 func (*Ping) Description() string {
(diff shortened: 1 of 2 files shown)


rajiv-k:
!signed-cla

rajiv-k:
!signed-cla

rajiv-k:
Looks like an unrelated test failure to me. How do I re-run the tests?

plugins/inputs/ping/ping.go
```diff
@@ -172,6 +179,10 @@ func (p *Ping) nativePing(destination string) (*pingStats, error) {
 		pinger.SetNetwork("ip6")
 	}
 
+	if p.Method == "native" {
+		pinger.Size = p.Size
```
sspaink, line 183: Doesn't seem like `p.Size` is being initialized anywhere, wouldn't this be 0 if this isn't configured?

rajiv-k: `p.Size` will indeed default to 0, but that's the size of the _data bytes_. With the ICMP header included, that will form an ICMP packet of 8 bytes.

sspaink: Thank you for the explanation, that makes sense. Because the README for the size flag is `56`, I wonder if we also default to `56` instead of `0` as well?

It looks like go-ping originally defaulted to `16` so that could be an appropriate default as well, but `56` probably makes more sense.

rajiv-k: Sure! Let's default to 56.

sspaink (approved):

PR #9275, opened 2021-05-17
Fix reading config files starting with http:

### LINKED ISSUE #9218: Telegraf --config does not read files if files are named http:....
### Relevant telegraf.conf:

```toml
# contents does not matter, as it is never read:
[[inputs.http_response]]
  urls = [ "https://www.influxdata.com/" ]
  response_timeout = "5s"
  follow_redirects = false
  interval         = "120s"
```

### System info:

(most likely OS independent)
RHEL7, RHEL8, ...

### Steps to reproduce:
1.  Create the above configuration in the file  https::##www.influxdata.com.conf: 
```
cat > 'https:##www.influxdata.com'.conf << EOF
[[inputs.http_response]]
  urls = [ "https://www.influxdata.com/" ]
  response_timeout = "5s"
  follow_redirects = false
  interval         = "120s"
EOF
```
2. Create a "verification" link:
```
ln 'https:##www.influxdata.com'.conf www.influxdata.com.conf
```
3. Make telegraf fail on the first:
```
 telegraf --test --config 'https:##www.influxdata.com'.conf
2021-04-29T12:41:46Z I! Starting Telegraf 1.18.0
2021-04-29T12:41:46Z E! [telegraf] Error running agent: Error loading config file https:##www.influxdata.com.conf: Retry 0 of 3 failed connecting to HTTP config server Get "https:#%23www.influxdata.com.conf": http: no Host in request URL
```
4. and succeed on the other:
```
telegraf --test --config www.influxdata.com.conf
2021-04-29T12:42:34Z I! Starting Telegraf 1.18.0
> http_response,host=...,method=GET,result=success,server=https://www.influxdata.com/,status_code=200 content_length=224068i,http_response_code=200i,response_time=0.397142678,result_code=0i,result_type="success" 1619700155000000000
```

### Expected behavior:

Telegraf should read the configuration file if it exists. The filename is valid and file exists.

### Actual behavior:

Telegraf tries to fetch a non-existing URL

### Additional info:

We use the naming standard as we perform remote monitoring of thousands of servers.
We use the naming scheme in a lot of other places too (kapacitor being one of them).

Bug is verified present in 1.14.3 and 1.18.0

akrantz01:
resolves #9218 

Fixed an error where telegraf attempted to fetch files starting with `http:` or `https:` over HTTP/S rather than read them from the filesystem.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -50,6 +50,10 @@ var (
 		`\`, `\\`,
 	)
 	httpLoadConfigRetryInterval = 10 * time.Second
+
+	// fetchUrlRe is a regex to determine whether the requested file should
+	// be fetched from a remote or read from the filesystem.
+	fetchUrlRe = regexp.MustCompile(`^\w+://`)
 )
 
 // Config specifies the URL/user/password for the database that telegraf
@@ -902,8 +906,7 @@ func escapeEnv(value string) string {
 }
 
 func loadConfig(config string) ([]byte, error) {
-	content, err := ioutil.ReadFile(config)
-	if err != nil {
+	if fetchUrlRe.MatchString(config) {
 		u, err := url.Parse(config)
 		if err != nil {
 			return nil, err
@@ -913,11 +916,11 @@ func loadConfig(config string) ([]byte, error) {
 		case "https", "http":
 			return fetchConfig(u)
 		default:
-			// If it isn't a https scheme, an error occurred
+			// If it isn't a https or file scheme, try it as a file
 		}
 	}
 
-	return content, err
+	return ioutil.ReadFile(config)
 }
 
 func fetchConfig(u *url.URL) ([]byte, error) {
--- a/config/config.go
+++ b/config/config.go
@@ -915,6 +915,8 @@ func loadConfig(config string) ([]byte, error) {
 		switch u.Scheme {
 		case "https", "http":
 			return fetchConfig(u)
+		default:
+			return nil, fmt.Errorf("scheme %q not supported", u.Scheme)
 		}
 	}
 
--- a/config/config.go
+++ b/config/config.go
@@ -50,10 +50,6 @@ var (
 		`\`, `\\`,
 	)
 	httpLoadConfigRetryInterval = 10 * time.Second
-
-	// fetchURLRe is a regex to determine whether the requested file should
-	// be fetched from a remote or read from the filesystem.
-	fetchURLRe = regexp.MustCompile(`^\w+://`)
 )
 
 // Config specifies the URL/user/password for the database that telegraf
@@ -906,18 +902,20 @@ func escapeEnv(value string) string {
 }
 
 func loadConfig(config string) ([]byte, error) {
-	if fetchURLRe.MatchString(config) {
-		u, err := url.Parse(config)
-		if err != nil {
-			return nil, err
-		}
+	u, err := url.Parse(config)
+	if err != nil {
+		return nil, err
+	}
 
-		switch u.Scheme {
-		case "https", "http":
-			return fetchConfig(u)
-		default:
-			// If it isn't a https or file scheme, try it as a file
-		}
+	if u.Host == "" {
+		return ioutil.ReadFile(config)
+	}
+
+	switch u.Scheme {
+	case "https", "http":
+		return fetchConfig(u)
+	default:
+		// If it isn't a https or file scheme, try it as a file
 	}
 
 	return ioutil.ReadFile(config)
--- a/config/config.go
+++ b/config/config.go
@@ -50,6 +50,10 @@ var (
 		`\`, `\\`,
 	)
 	httpLoadConfigRetryInterval = 10 * time.Second
+
+	// fetchURLRe is a regex to determine whether the requested file should
+	// be fetched from a remote or read from the filesystem.
+	fetchURLRe = regexp.MustCompile(`^\w+://`)
 )
 
 // Config specifies the URL/user/password for the database that telegraf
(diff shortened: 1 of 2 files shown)


akrantz01:
!signed-cla

srebhan (changes requested):
No. This means files take precedence over URLs even though you specified a valid (remote) URI. You can achieve the same with specifying `file://http:whateverfileyouneed`.

config/config.go
```diff
@@ -902,18 +902,22 @@ func escapeEnv(value string) string {
...
-	switch u.Scheme {
-	case "https", "http":
-		return fetchConfig(u)
-	default:
-		// If it isn't a https scheme, try it as a file.
+		switch u.Scheme {
+		case "https", "http":
+			return fetchConfig(u)
+		default:
+			// If it isn't a https scheme, an error occurred
```
Hipska, line 916: In this case, the actual return of `loadConfig` would be `nil, nil`, which is not a good situation, please provide a meaningful error message or reuse the error from `ioutil.ReadFile`. (the `err` from that one is overridden by the `err` from `url.Parse`)

Hipska (changes requested):
I agree with @srebhan, currently the preference is load url and otherwise load file. Now you are first trying to load file and otherwise try to load url. I think the team needs to discuss what preference we want.

ssoroka:
Thanks for the PR!

I don't think this change is unreasonable. Hipska has a good point about handling the error messages, so we need to pay attention to that. Probably the best way to decide if we should parse it as a file or a url is to look for a character sequence that clearly identifies it. Parsing files as urls will still give you a url, which is what caused the original problem. This PR _does_ in fact resolve that problem, it just comes at the cost of checking the disk for matching files first.. Which honestly isn't really a big deal.

If we wanted to stick closely to the intent of the user configuring the plugin, we could match against the regex `^\w+://` and if it matches use url parsing, and if it doesn't, use file parsing. This will remove all ambiguity, and allow for clear error messages that say things like `"couldn't read the file %q"` or `"couldn't parse the url %q"`.

srebhan:
@ssoroka well my fear is that you, with the change applied, cannot specify a URL if a file with that name exists. Let's make an example. Say a) you want to access a file named `http://myfile` and b) you want to access a server at `http://myserver` and let's assume you have (for whatever reason) the files `http://myfile` and `http://myserver` on-disk...

With the current implementation you can achieve this by specifying
for a) `file://'http://myfile'`
for b) `http://myserver`
and it fill access the file for a) and the server for b).

With the change you specify
for a) `http://myfile`
for b) `http://myserver`
while correct for a) it now will access the **file** `http://myserver` for b) instead of the server.

Sorry @akrantz01 if my comment was too harsh, but I think this kind of implicit behavior is cumbersome and might lead to unintended effects.

srebhan:
Hmmm and after reading the code again I see that my example is wrong. :-) So currently specifying `file://'http://myfile` will try to read `'file://http://myfile'` instead of `'http://myfile'`... So I see your point now @akrantz01...

The team probably should discuss this as @Hipska said, but my preference would be to implement the handling of the `file://` scheme such as
```golang
func loadConfig(config string) ([]byte, error) {
	if strings.HasPrefix(config, "file://") {
		return ioutil.ReadFile(strings.TrimPrefix(config, "file://"))
	}

	u, err := url.Parse(config)
	if err != nil {
		return nil, err
	}

	switch u.Scheme {
	case "https", "http":
		return fetchConfig(u)
	}

	// If it isn't known scheme, try it as a file.
	return ioutil.ReadFile(config)
}
```

Hipska:
Isn't `fetchConfig` also able to handle `file://` urls? So case scheme is 'file' then do something ;)

akrantz01:
To the best of my knowledge, I don't think you can create files with forward slashes in the name on any platform, so I think it'd be best to go with @ssoroka's suggestion. It wouldn't require any changes from the user and would still handle actual URLs properly.

Regarding @Hipska's point, we could still add a case for `file://` URLs so users could be explicit if they wanted.

akrantz01:
Actually, if we wanted to support `file://` URLs through `fetchConfig`, it would take a bit more work since the default HTTP client doesn't support the file scheme.

Hipska (changes requested):
I would do something like this: (pseudo code)
```
if u.Scheme in ["https", "http"] and u.Host not empty { 
    return fetchConfig(u)
}
```

srebhan:
We will get problem with parameters like `file://https:test`. Check out https://goplay.space/#SqDP6ybX7Af or
[the attached code](https://github.com/influxdata/telegraf/files/6509573/config_loading_test.txt). That's why I tried to handle the `file://` part first. :-)

akrantz01:
@ssoroka and I decided that it would be simplest just to use the regex before parsing the URL, since then the implementation isn't dependent on the internals of `url.Parse` if they ever change. It was also decided to not support `file://`, since it isn't used anywhere else.

Hipska (approved):

config/config.go
```diff
@@ -902,17 +906,20 @@ func escapeEnv(value string) string {
...
-	switch u.Scheme {
-	case "https", "http":
-		return fetchConfig(u)
-	default:
-		// If it isn't a https scheme, try it as a file.
+		switch u.Scheme {
+		case "https", "http":
+			return fetchConfig(u)
+		default:
+			// If it isn't a https scheme, try it as a file
```
srebhan, line 919: You can leave this one out as it does nothing and move the comment next to the last return. Just a minor comment though.

srebhan (approved):
@akrantz01 thanks for keep going with this persistent issue.
My only (_very_ minor) comment is to remove the `default` clause, but this is not a blocker.

srebhan:
!rety-failed

srebhan:
!retry-failed

config/config.go
Hipska: I think the linter will complain then?

Hipska (approved):

srebhan (approved):
Perfect thanks!

config/config.go
```diff
@@ -902,17 +906,19 @@ func escapeEnv(value string) string {
...
 
-	switch u.Scheme {
-	case "https", "http":
-		return fetchConfig(u)
-	default:
-		// If it isn't a https scheme, try it as a file.
+		switch u.Scheme {
+		case "https", "http":
+			return fetchConfig(u)
+		}
```
ssoroka, line 918: maybe have a default case that errors instead of falling through to reading it as a file.
```suggestion
		default:
			return nil, fmt.Errorf("scheme %q not supported", u.Scheme)
		}
```

ssoroka (approved):
Looks good. left a suggestion on erroring if the requested scheme doesn't exist.

Hipska (approved):

PR #9276, opened 2021-05-18
vSphere input: need to be able to configure the historical interval duration

### LINKED ISSUE #9247: Unable to use the VSphere input for historical statistics when historical interval is not 5m (300s)
### Relevant telegraf.conf:

```toml

```

### System info:

CentOS 7
Telegraf 1.17.2 but verified problem exists in Master
vSphere 7.0.2.00100

### Steps to reproduce:

1. Configure vSphere daily statistics interval to be anything other than default 5m
2. Configure telegraf to query all statistics from the vSphere you just modified config on
3. Start telegraf

### Expected behavior:

Operate as normal pulling all statistics

### Actual behavior:

For all historical statistics this error is thrown in the logs and no statistics are recorded.
2021-05-06T17:41:50Z D! [inputs.vsphere] Find(Datastore, /*/datastore/**) returned 17 objects
2021-05-06T17:41:50Z E! [inputs.vsphere] Getting metric metadata. Discovery will be incomplete. Error: ServerFaultCode: A specified parameter was not correct: interval
2021-05-06T17:41:50Z D! [inputs.vsphere] Found 0 metrics for xxxxxxx

### Additional info:
Plan to add a config option for "HistoricalInterval" which defaults to 300 but can be overridden so that it matches the daily interval configured in vSphere.  This will replace the currently hard coded value of 300.

gandalf80:

resolves #9247

Added a configuration option "historical_interval_duration" to the vSphere configuration that defaults to 300s but can be overriden.  Updated the input plugin readme to clarify that the 5min historical interval is a default but can be modified in vCenter and that queries for historical data must match the interval set in vCenter.


plugins/inputs/vsphere/vsphere.go
```diff
@@ -50,13 +50,14 @@ type VSphere struct {
...
-	ObjectDiscoveryInterval config.Duration
-	Timeout                 config.Duration
+	MaxQueryObjects            int
+	MaxQueryMetrics            int
+	CollectConcurrency         int
+	DiscoverConcurrency        int
+	ForceDiscoverOnInit        bool
+	ObjectDiscoveryInterval    config.Duration
+	Timeout                    config.Duration
+	HistoricalIntervalDuration int
```
Hipska, line 60: Instead of `int` please also make it a `config.Duration`. That simplifies it for all users.

Hipska (changes requested):

plugins/inputs/vsphere/vsphere.go
gandalf80: Yeah I had gone back and forth on whether it should include "Duration" at the end of the name, I think it makes sense to drop it...  
Switched to using config.Duration.  New to Go and time.Duration is....different :)

Hipska (approved):

sspaink (approved):

PR #8651, opened 2021-01-05
feat: postgresql output

phemmer:
This PR provides a PostgreSQL output plugin.

This is a continuation of #3428 but massively reworked. There are a lot of breaking changes to address issues/limitations, as well as performance optimizations.  
The performance optimizations are not minor either. The original code from #3428 benchmarked at around 250 points per second (using real production data). Its been a bit since I last ran a benchmark, but last time I did I was able to get around 60,000 points per second.

closes #3408 closes #3428 

Major changes from #3428:
* Deterministic tags when using `tags_as_foreign_keys`.  
  Previous code utilized an auto-generated serial to obtain the tag ID for a set of tags. This resulted in a severe bottleneck as the tag IDs have to be looked up from the database for every single series. The new code hashes the tags & their values (within telegraf) to generate the tag ID.
* Connection pooling & concurrency.  
  The plugin allows multiple simultaneous inserts into the database. Each batch is split into a separate insert (COPY) per measurement/table, which can run in parallel. If the plugin receives batches faster than a single connection can write, each batch will also be inserted in parallel.
* Use of `numeric` data type for `uint64`.  
  Previous code used `bigint` for `uint64`. This would result in overflow errors when inserting values larger than an `int64` as `bigint` is a signed 64-bit integer. `numeric` is an arbitrary precision exact value numeric data type. It is less performant than `bigint`, but it's the only datatype that can hold the full `uint64` range.
* More powerful schema modification template statements.  
  For table creation & column addition, a more powerful mechanism was implemented. The new template statements use golang's text/template library with lots of provided variables, methods, & functions to allow for a virtually any use case. Some example configs are provided below to demonstrate. Comprehensive documentation on this functionality will be added.
* Better error handling.
  The old code didn't use transactions, and would infinitely retry the entire batch on error. This resulted in things like duplicate inserts. As mentioned earlier, each measurement is a separate sub-batch so this mitigates some of the scope of errors. Each sub-batch is inserted in a transaction, so there's no risk of duplicates. In addition, the plugin is able to discern between temporary and permanent errors. A permanent error is something like bad SQL that will always fail now matter how many times you retry it. A temporary error is something like a failed connection, where a retry may succeed. Temporary errors will infinitely retry (with incremental backoff), while permanent errors will discard the sub-batch.

Note that there are breaking changes in the db schema:
[...]

PR #9280, opened 2021-05-19
Add SQL output plugin

reimda:
Add a SQL Output plugin that supports multiple DBMSes. This PR is an update of #4205 by @illuusio, adding integration tests and the ability to store the metric timestamp.

This plugin uses a simple, hard-coded database schema. The table name is the metric name. There is a column per field and a column per tag. It writes a row for every input metric. This means multiple metrics are never merged into a single row, even if they have the same metric name, tags, and timestamp.

This plugin generates simple SQL that is likely to work on any DBMS. It doesn't use any feature that is specific to a particular DBMS. If you want to use a feature that is specific to a particular DBMS, you may be able to set it up manually outside of this plugin or you may need to use a different plugin altogether. This means this plugin doesn't intend to remove the need for other SQL output plugins that are specialized for a single DBMS, for example #8651 for Postgres.

There is basic support to automatically create tables. Automatic table creation assumes input metrics always have the same tags and fields, and that the type of each field never changes. (The plugin doesn't attempt to alter existing tables to add or change columns based on input metrics). Use cases that don't have metrics with static schemas may require the user to create tables manually or to use a different plugin.

The integration tests run each DBMS in a docker container. Since these tests are relatively slow they are skipped in go's short mode unit tests. This is to allow short mode to complete quickly so it can be run as part of in CI builds for PRs. I intend to add a new nightly CI job that runs non short mode tests so they will be exercised regularly.

Draft PR to do list:
- Currently the only integration test is for mariadb but I plan to add similar tests for sqlite and postgres.
- The metric timestamp is always stored in a column called "timestamp" but I plan to make the column name configurable. I'll also try to make timestamp writing optional to allow use of db insertion timestamp like #4205 does.
- Add nightly CI job for integration tests


plugins/outputs/sql/sql.go
```diff
@@ -0,0 +1,256 @@
...
+
+func (p *Sql) generateInsert(tablename string, columns []string) string {
+
+	var placeholder, quoted []string
+	for _, column := range columns {
+		placeholder = append(placeholder, fmt.Sprintf("?"))
+		quoted = append(quoted, quoteIdent(column))
+	}
+
+	sql := fmt.Sprintf("INSERT INTO %s(%s) VALUES(%s)", quoteIdent(tablename), strings.Join(quoted, ","), strings.Join(placeholder, ","))
```
ssoroka, line 182: column names might need to be escaped for some weirder column names.

ssoroka (commented):
looks good

plugins/outputs/sql/sql.go
reimda: Yes, there's also opportunity for a sql injection attack through the names of metrics, fields, and tags currently. I'll have to escape them to get them to work and sanitize them so they can't be misused.

illuusio:
Is there any change to get this in?

reimda:
> Is there any change to get this in?

Hi @illuusio, I'd like to complete the to-do list in the PR description and add escaping like ssoroka mentioned in review. I also want to add to the readme, probably reusing some of this PR's description and adding a couple config examples.

I hope to get these things done in the next few days and have the PR reviewed and merged in time for the 1.19.0 release candidate on June 2.

illuusio:
@reimda Should I close my PR to favor this? If everything goes like this I'll be happy camper with 1.19.

reimda:
@illuusio This PR is your PR with tests and a couple fixes added, so it's safe to close your PR. The tests and fixes will make it easier to pass code review and hopefully a little easier to use.

I'm glad to hear you're happy about the changes I've made and the plan to get it into 1.19. I should have contacted you before updating your PR, but I was in a hurry because I have a need for this too. Thanks for your work writing #4205!

plugins/outputs/sql/sql.go
reimda: I added identifier quoting and a check for invalid characters that should cover this.

PR #9289, opened 2021-05-21
Issue 8914 x509sni

### LINKED ISSUE #8914: inputs.x509_cert unexpected behavior with SNI
### Relevant telegraf.conf:

```toml
[global_tags]

[agent]
  interval = "60s"
  round_interval = false
  metric_batch_size = 500
  metric_buffer_limit = 5000
  collection_jitter = "0s"
  flush_interval = "60s"
  flush_jitter = "0s"
  precision = ""
  debug = false
  quiet = false
  logfile = ""
  hostname = ""
  omit_hostname = false

[[inputs.x509_cert]]
  sources = ["https://www.underverse.net:443", "https://www.gynpraxis-bahls.de:443"]
  [inputs.x509_cert.tagdrop]
    common_name  = ["Let's Encrypt Authority X3", "R3"]
```

### System info:

Telegraf 1.17.3 on ArchLinux

### Steps to reproduce:

I ran two little Websites on one of my Servers. DNS resolves to the same IP and i use SNI to distinguish between the sites.
After upgrading telegraf to 1.17.3 (from 1.17.0) i realized that the behavior changed of the x509 plugin.
Before i got the right certificates from the server and got 2 different measurements from the plugin.
After the update it shows for both entries the same certificate (its the default one).
So i guess PR #7897 with 545d996771cd1b8e94e4f2ef98fea60575c27ed7 (introduced in 1.17.1) breaks the original behavior of the plugin. 

1. Run the above config with telegraf 1.17.0 and inspect output (expected behavior).
2. Run the above config with telegraf 1.17.3 and inspect output (actual behavior).

### Expected behavior:

```
2021-02-26T14:07:10Z I! Starting Telegraf 1.17.0
2021-02-26T14:07:10Z D! [agent] Initializing plugins
2021-02-26T14:07:10Z D! [agent] Starting service inputs
> x509_cert,common_name=underverse.net,host=evelyn,issuer_common_name=R3,public_key_algorithm=RSA,san=*.underverse.net\,underverse.net,serial_number=4c11c31923c7824b9055dbd729f06c41ab2,signature_algorithm=SHA256-RSA,source=https://www.underverse.net:443,verification=valid age=5208391i,enddate=1616916039i,expiry=2567608i,startdate=1609140039i,verification_code=0i 1614348431000000000
2021-02-26T14:07:10Z D! [agent] Stopping service inputs
> x509_cert,common_name=gynpraxis-bahls.de,host=evelyn,issuer_common_name=R3,public_key_algorithm=RSA,san=gynpraxis-bahls.de\,www.gynpraxis-bahls.de,serial_number=32b96a849248c02179ae09479551704d315,signature_algorithm=SHA256-RSA,source=https://www.gynpraxis-bahls.de:443,verification=valid age=5208429i,enddate=1616916001i,expiry=2567570i,startdate=1609140001i,verification_code=0i 1614348431000000000
2021-02-26T14:07:10Z D! [agent] Input channel closed
2021-02-26T14:07:10Z D! [agent] Stopped Successfully
```

### Actual behavior:

```
2021-02-26T14:07:31Z I! Starting Telegraf 1.17.3
2021-02-26T14:07:31Z D! [agent] Initializing plugins
2021-02-26T14:07:31Z D! [agent] Starting service inputs
> x509_cert,common_name=underverse.net,host=evelyn,issuer_common_name=R3,public_key_algorithm=RSA,san=*.underverse.net\,underverse.net,serial_number=4c11c31923c7824b9055dbd729f06c41ab2,signature_algorithm=SHA256-RSA,source=https://www.underverse.net:443,verification=valid age=5208412i,enddate=1616916039i,expiry=2567587i,startdate=1609140039i,verification_code=0i 1614348451000000000
> x509_cert,common_name=underverse.net,host=evelyn,issuer_common_name=R3,public_key_algorithm=RSA,san=*.underverse.net\,underverse.net,serial_number=4c11c31923c7824b9055dbd729f06c41ab2,signature_algorithm=SHA256-RSA,source=https://www.gynpraxis-bahls.de:443,verification=valid age=5208412i,enddate=1616916039i,expiry=2567587i,startdate=1609140039i,verification_code=0i 1614348451000000000
2021-02-26T14:07:31Z D! [agent] Stopping service inputs
2021-02-26T14:07:31Z D! [agent] Input channel closed
2021-02-26T14:07:31Z D! [agent] Stopped Successfully
```

### LINKED ISSUE #9278: inputs.x509_cert not handling multiple sources correctly
### Relevant telegraf.conf:

```toml
[[inputs.x509_cert]]
   sources = [
      "https://web1.domain.edu:443",
      "https://web2.domain.edu:443",
      "https://web3.domain.edu:443",
   ]

  timeout = "5s"
```

### System info:

telegraf-1.17.1+ (versions before are ok), RHEL 8.3 and CentOS 7.9

### Steps to reproduce:

1. Setup a x509_cert input which has multiple servers in the sources block, where they each resolve to different hostnames (cert for server1 is not valid for server2, etc)
2. telegraf --config sslcert-check.conf --test

### Expected behavior:

As long as all certificates in the sources list are valid there should be no verification errors reported.

### Actual behavior:

Verification errors are printed like so:
```
# telegraf --config sslcert-check.conf --test | egrep -o 'verification_error=".*"'
verification_error="x509: certificate is valid for web2.domain.edu not web1.domain.edu"
verification_error="x509: certificate is valid for web3.domain.edu not web1.domain.edu"
```
From the errors it looks like it's trying to verify all certs listed against the hostname for the first server given in the list.

### Additional info:

Same config has no such errors prior to version 1.17.1.

Breaking each server out into individual checks is still fine and gives no issues:
```
[[inputs.x509_cert]]
   sources = [
      "https://web1.domain.edu:443",
   ]
[[inputs.x509_cert]]
   sources = [
      "https://web2.domain.edu:443",
   ]

#...etc...
```
But would still be nice to have the old behavior where one [[inputs.x509_cert]] block could include all the servers needed to be monitored.

### LINKED ISSUE #9384: x509 Certificate Input Plugin v1.19.0 does not work with URL
### Relevant telegraf.conf:

```
[[inputs.x509_cert]]
sources = ["https://my.server.com:443"] 

```

### System info:

Windows Server 2019
Telegraf v1.19.0

### Steps to reproduce:

1. Configure inputs.x509_cert
2. telegraf.exe --config "C:\Program Files\Telegraf\telegraf.conf" --test --input-filter x509_cert --debug

### Expected behavior:

Certificate details returned
> x509_cert,common_name=*.server.com,country=GB <snip>
> x509_cert,common_name=Root Cert <snip> 

### Actual behavior:

No certificate details are returned
2021-06-16T16:52:07Z E! [inputs.x509_cert] could not find file: &{https:\\my.server.com:443 false false  <nil>}

### Additional info:

Telegraf v1.18.1 on the same system does not show this issue.

jjh74:
### Required for all PRs:

- [x] Updated associated README.md. (no need to update README)
- [x] Wrote appropriate unit tests. (tests already exist)

resolves #8914
resolves #9278
resolves #9384

1. commit  448eac0 tries to ensure that fileglobs are not tried for remote uris (sources starting with tcp://, udp:// or https://). This is because strings.Index(source, ":\\\\") != 1 (https://github.com/influxdata/telegraf/blob/master/plugins/inputs/x509_cert/x509_cert.go#L70) matches almost all sources -> fileglobs are tried for remote sources.
2. commit 4e1e6f3 tries to ensure that c.tlsCfg.ServerName is reset between multiple certs/sources. Without this telegraf uses same SNI for multiple certs. Leading to telegraf requesting wrong certificates and failing validation because SNI/SAN doesn't match. (Longer discussion in: https://github.com/influxdata/telegraf/issues/8914)

### DIFF
--- a/plugins/inputs/x509_cert/x509_cert.go
+++ b/plugins/inputs/x509_cert/x509_cert.go
@@ -66,8 +66,7 @@ func (c *X509Cert) SampleConfig() string {
 func (c *X509Cert) sourcesToURLs() error {
 	for _, source := range c.Sources {
 		if strings.HasPrefix(source, "file://") ||
-			strings.HasPrefix(source, "/") ||
-			strings.Index(source, ":\\") != 1 {
+			strings.HasPrefix(source, "/") {
 			source = filepath.ToSlash(strings.TrimPrefix(source, "file://"))
 			g, err := globpath.Compile(source)
 			if err != nil {
@@ -82,7 +81,6 @@ func (c *X509Cert) sourcesToURLs() error {
 			if err != nil {
 				return fmt.Errorf("failed to parse cert location - %s", err.Error())
 			}
-
 			c.locations = append(c.locations, u)
 		}
 	}
--- a/plugins/inputs/x509_cert/x509_cert.go
+++ b/plugins/inputs/x509_cert/x509_cert.go
@@ -125,6 +125,9 @@ func (c *X509Cert) getCert(u *url.URL, timeout time.Duration) ([]*x509.Certifica
 		conn := tls.Client(ipConn, c.tlsCfg)
 		defer conn.Close()
 
+		// reset SNI between requests
+		defer func() { c.tlsCfg.ServerName = "" }()
+
 		hsErr := conn.Handshake()
 		if hsErr != nil {
 			return nil, hsErr
@@ -311,6 +314,14 @@ func (c *X509Cert) Init() error {
 		tlsCfg = &tls.Config{}
 	}
 
+	if tlsCfg.ServerName != "" && c.ServerName == "" {
+		// Save SNI from tlsCfg.ServerName to c.ServerName and reset tlsCfg.ServerName.
+		// We need to reset c.tlsCfg.ServerName for each certificate when there's
+		// no explicit SNI (c.tlsCfg.ServerName or c.ServerName) otherwise we'll always (re)use
+		// first uri HostName for all certs (see issue 8914)
+		c.ServerName = tlsCfg.ServerName
+		tlsCfg.ServerName = ""
+	}
 	c.tlsCfg = tlsCfg
 
 	return nil
--- a/plugins/inputs/x509_cert/x509_cert_test.go
+++ b/plugins/inputs/x509_cert/x509_cert_test.go
@@ -316,6 +316,16 @@ func TestGatherCertMustNotTimeout(t *testing.T) {
 	assert.True(t, acc.HasMeasurement("x509_cert"))
 }
 
+func TestSourcesToURLs(t *testing.T) {
+	m := &X509Cert{
+		Sources: []string{"https://www.influxdata.com:443", "tcp://influxdata.com:443", "file:///dummy_test_path_file.pem", "/tmp/dummy_test_path_glob*.pem"},
+	}
+	require.NoError(t, m.Init())
+
+	assert.Equal(t, len(m.globpaths), 2)
+	assert.Equal(t, len(m.locations), 2)
+}
+
 func TestServerName(t *testing.T) {
 	tests := []struct {
 		name     string


plugins/inputs/x509_cert/x509_cert.go
```diff
@@ -129,10 +131,12 @@ func (c *X509Cert) getCert(u *url.URL, timeout time.Duration) ([]*x509.Certifica
 
 		hsErr := conn.Handshake()
 		if hsErr != nil {
+			c.tlsCfg.ServerName = "" // reset SNI, otherwise we reuse c.tlsCfg.ServerName
```
srebhan, line 134: I think you better make this a
```golang
defer func() { c.tlsCfg.ServerName = "" }()
```
maybe before doing the handshake in order to be prepared for future "exit-on-error" additions.

plugins/inputs/x509_cert/x509_cert.go
```diff
@@ -246,6 +250,16 @@ func (c *X509Cert) collectCertURLs() ([]*url.URL, error) {
...
 	now := time.Now()
+
+	if c.tlsCfg.ServerName != "" && c.ServerName == "" {
+		// Save SNI from c.tlsCfg.ServerName to c.ServerName and reset c.tlsCfg.ServerName.
+		// We need to reset c.tlsCfg.ServerName for each certificate when there's
+		// no explicit SNI (c.tlsCfg.ServerName or c.ServerName) otherwise we'll always use
+		// first uri HostName for all certs (see issue 8914)
+		c.ServerName = c.tlsCfg.ServerName
+		c.tlsCfg.ServerName = ""
+	}
```
srebhan, line 261: I'm not sure I understand why you need to do this. Why not just reset `c.tlsCfg.ServerName`?

srebhan (commented):
@jjh74 thanks for digging into this and submitting this PR. I added a comment and a question in the code, so it would be nice if you can comment on those.

However, taking a quick look at the code I think the issue arises from a conceptual weakness in the code. We reuse `c.tlsCfg` for multiple `location`s and are thus stateful in the code. IMO we should restructure the code such that we create one `c.tlsCfg` for each `location` in an `Init()` function making it an array. This way we can setup everything correctly there and do not need to clear fields in the loop over the locations. @jjh74 what is your opinion on this suggestion?

plugins/inputs/x509_cert/x509_cert.go
jjh74: Purpose for this is that if you've a config like:
```
[[inputs.x509_cert]]
  sources = [ "https://www.example.com:443", "tcp://example.org:443", "https://some.domain.com:443" ]
  tls_server_name = "why.example.com"
  # or
  #server_name = "why.example.com"
```
Then you can "force" all certs/requests to have why.example.com SNI. I'm not sure if this is a real use case, but this way you can force SNI (and use that SNI for validation) ?
I think w/out this first request will have why.example.com SNI and other requests will use example.org and some.domain.com (because c.tlsCfg.ServerName is reset between requests).

plugins/inputs/x509_cert/x509_cert.go
jjh74: Thanks, this seems less error prone. I changed this with commit 858b677

plugins/inputs/x509_cert/x509_cert.go
srebhan: Yeah but shouldn't it be `c.ClientConfig.ServerName` then? As far as I can see, this `if` condition is only ever true the first time `Gather()` is called as afterwards `c.tlsCfg.ServerName` will _always_ be empty. If this is the intention, then please move this code to `Init()` and we are on the same page. :-)

akrantz01:
In addition to what @srebhan said, I think it'd be a good idea to add a unit test to make sure that URLs and file paths/globs are parsed correctly, so this doesn't come up again in the future. Also, please fix the issue raised by the linter.

plugins/inputs/x509_cert/x509_cert.go
```diff
@@ -64,10 +65,15 @@ func (c *X509Cert) SampleConfig() string {
...
 
 func (c *X509Cert) sourcesToURLs() error {
+	uriregexp, err := regexp.Compile("^(((udp|tcp)[46]?)|https)://")
+	if err != nil {
+		return fmt.Errorf("could not compile uri regexp %v", err)
+	}
+
 	for _, source := range c.Sources {
-		if strings.HasPrefix(source, "file://") ||
+		if !uriregexp.MatchString(source) && (strings.HasPrefix(source, "file://") ||
```
srebhan, line 74: Maybe I miss something, but shouldn't `uriregexp.MatchString(source)` always be `false` if any of the other conditions is true? I think it is redundant, isn't it?

srebhan (commented):
I have one remaining question regarding `uriregexp` in the code...

plugins/inputs/x509_cert/x509_cert.go
jjh74: Original code is: https://github.com/influxdata/telegraf/blob/master/plugins/inputs/x509_cert/x509_cert.go#L68-L70
and AFAIK `strings.Index(source, ":\\") != 1` will match pretty much everything (and current master complains for https:// and tcp:// sources):
```
[inputs.x509_cert] could not find file: &{https://example.org:443 false false }
[inputs.x509_cert] could not find file: &{tcp://2.example.org:443 false false }
```
I'm not quite sure what the purpose of the strings.Index(source, ":\\") != 1 is, but 
maybe this could be rewritten without regexp (untested):
```
if (strings.HasPrefix(source, "file://") ||
   strings.HasPrefix(source, "/")) &&
   !(strings.Index(source, ":\\") == 1) {
```

jjh74:
> @jjh74 thanks for digging into this and submitting this PR. I added a comment and a question in the code, so it would be nice if you can comment on those.
> 
> However, taking a quick look at the code I think the issue arises from a conceptual weakness in the code. We reuse `c.tlsCfg` for multiple `location`s and are thus stateful in the code. IMO we should restructure the code such that we create one `c.tlsCfg` for each `location` in an `Init()` function making it an array. This way we can setup everything correctly there and do not need to clear fields in the loop over the locations. @jjh74 what is your opinion on this suggestion?

I have no strong opinion one way or another.
Perhaps using `c.tlsCfg` for each `location` would make it easier to add other TLS configurations (like tls min version or cipher suites) ? (But writing configuration where you'll add lot's of parameters for each `sources` seems quite unreadable and it's probably easier to use multiple `[[inputs.x509_cert]]`)

I think tests are missing a test case where you have more than one https/tcp sources and the test would check that correct SNI is used for each one.

plugins/inputs/x509_cert/x509_cert.go
srebhan: OMG! I think `strings.Index(source, ":\\") != 1` this is just the inverse of what it is supposed to be. Making this `strings.Index(source, ":\\") == 1` should be correct to cover `C:\\` type of paths.

srebhan:
Currently we are using the user configuration and create **one** `tlsCfg` as a kind of template. However, this way we need to take care for resetting location-dependent properties of that template while we gather.
My suggestion is to use the user configuration and create **multiple** `tlsCfg`, one per location. We then initialize the location-dependent properties once (e.g. at `Init()`) and don't need to do anything during gather...

plugins/inputs/x509_cert/x509_cert.go
jjh74: Then
```
if strings.Index(source, ":\\") == 1 {
    source = "file://" + filepath.ToSlash(source)
}
```
(https://github.com/influxdata/telegraf/blob/master/plugins/inputs/x509_cert/x509_cert.go#L78-L80)
is redundant and will never match ?

Do you know what the intention of the original code is ? Should it try to find fileglobs for sources like `C:\something` ? Now I think it tries to find globs for sources starting with `/` or `file://` and treat `C:\something` as `file://C:\something` (without trying to glob) ?

srebhan: Ok I walked through the PR adding this and it seems the discussion point was if Windows actually is supported by `globpath`. If it isn't, the code should be
```golang
		if strings.HasPrefix(source, "file://") ||
		   strings.HasPrefix(source, "/") {
			source = filepath.ToSlash(strings.TrimPrefix(source, "file://"))
			g, err := globpath.Compile(source)
			if err != nil {
				return fmt.Errorf("could not compile glob %v: %v", source, err)
			}
			c.globpaths = append(c.globpaths, g)
		} else {
			if strings.Index(source, ":\\") == 1 {
				source = "file://" + filepath.ToSlash(source)
			}
			u, err := url.Parse(source)
			if err != nil {
				return fmt.Errorf("failed to parse cert location - %s", err.Error())
			}
			c.locations = append(c.locations, u)
		}
```
Otherwise it should be 
```golang
		if strings.HasPrefix(source, "file://") ||
		   strings.HasPrefix(source, "/") || 
		   strings.Index(source, ":\\") == 1 {
			source = filepath.ToSlash(strings.TrimPrefix(source, "file://"))
			g, err := globpath.Compile(source)
			if err != nil {
				return fmt.Errorf("could not compile glob %v: %v", source, err)
			}
			c.globpaths = append(c.globpaths, g)
		} else {
			u, err := url.Parse(source)
			if err != nil {
				return fmt.Errorf("failed to parse cert location - %s", err.Error())
			}
			c.locations = append(c.locations, u)
		}
```

jjh74: I updated the code with expectation that `globpath` works on windows. (I'll try to test this with the windows binaries from CI build and if I can't get globpath to work with something like `C:\temp\*.pem` I'll change it to version that ignores globpath for `driveletter:\`).

srebhan:
@jjh74 seems like the windows test is failing (even for non-globbing tests)... :-( Maybe conversion of path-separators might be involved...

aslgithub:
Is there any progress on this ? It seems to be the pro-offered fix for the fact that https://github.com/influxdata/telegraf/issues/6952 has fundamentally broken x509_cert for sources other than file ?
I've logged https://github.com/influxdata/telegraf/issues/9384

jjh74:
@srebhan I just pushed a commit changing `sourcesToURLs()` so globpaths are tried only for `file://` and `/` sources (ie. expect that globpath is not working on windows). (You can still use sources like `file:///c:/WINDOWS/Temp/*.pem` to try globpath on windows). Hopefully CI now builds/works on windows.

srebhan (approved):
Looks good to me. The CI error is unrelated (in `plugins/inputs/tcp_listener`). Thanks for caring @jjh74.

srebhan:
!retry-failed

srebhan:
@jjh74 one unrelated test keeps failing, can you please try to rebase on latest master to see if this fixes the issue?

jjh74:
@srebhan now should be rebased to current master

srebhan:
@jjh74 seems like something went wrong with the rebase as you now have 139 files changed. 8-/

jjh74:
@srebhan new attempt with current master. I hope it goes better this time.

srebhan (approved):

sspaink (approved):
Awesome! Thank you for fixing these issues and explaining everything so clearly.

Nkmol:
Any ETA for a new release of this? I noticed this is essential for the multi-chain solution. For example the common Traefik LetsEncrypt setup uses a multichain setup that uses the server name to correctly pick the chain (LetsEncrypt vs Default Cert).

reimda:
> Any ETA for a new release of this?

The next scheduled patch release is July 7. We've been considering making an early patch release to fix a few regressions in 1.19.0, including this one, but we haven't settled on a date yet. It might be June 30.

PR #9290, opened 2021-05-21
feat: allow pulling configuration from separate InfluxDB instance

### LINKED ISSUE #8027: Add flag to Telegraf run command to pass a token for retrieving configuration from separate Influx host
## Feature Request

Add a flag to the Telegraf run command to pass a token for the InfluxDB instance hosting the Telegraf configuration.  This would allow for Telegraf to get its token from a different instance than to which it is writing data.

Could look like `--config-host-token` or `--config-fetch-token` or `--config-token` (because it may be implicit that it's specific to fetching that config).  All together might look like:
```
telegraf -config <url/to/config> --config-fetch-token <fetch_token>
```

## Motivation

1) Could offer flexibility to use a token through another means than an environment variable or hard-coding it into the configuration file.
2) It's possible that some Telegraf instances may want to pull configuration from a local (or "on-site", in IoT and other "fog" model use cases) instance of InfluxDB but still write out to another instance, say Cloud.

akrantz01:
resolves #8027 

Allows pulling configuration from a separate InfluxDB instance than data is pushed to by adding the `INFLUX_CONFIG_TOKEN` environment variable. There is an environment variable called `INFLUX_TOKEN` that can serve the same purpose, however, the default configuration generated by InfluxDB uses that for replacement within the config for the `token` field (shown below).

<details><summary>auto-generated configuration example</summary>
<p>

```toml
[[outputs.influxdb_v2]]	
  ## The URLs of the InfluxDB cluster nodes.
  ##
  ## Multiple URLs can be specified for a single cluster, only ONE of the
  ## urls will be written to each interval.
  ## urls exp: http://127.0.0.1:8086
  urls = ["http://172.17.0.4:8086"]

  ## Token for authentication.
  token = "$INFLUX_TOKEN"

  ## Organization is the name of the organization you wish to write to; must exist.
  organization = "organization"

  ## Destination bucket to write into.
  bucket = "bucket"
```
</p>
</details>

I ended up implementing and then subsequently removing the CLI flag since it would only work for authenticating with InfluxDB when pulling configuration. I decided that it might not be a good idea to have a flag named `--config-token` that only works for pulling configuration from InfluxDB as it could cause some confusion given the name.

### DIFF
--- a/agent/agent_test.go
+++ b/agent/agent_test.go
@@ -22,35 +22,35 @@ func TestAgent_OmitHostname(t *testing.T) {
 func TestAgent_LoadPlugin(t *testing.T) {
 	c := config.NewConfig()
 	c.InputFilters = []string{"mysql"}
-	err := c.LoadConfig("../config/testdata/telegraf-agent.toml", "")
+	err := c.LoadConfig("../config/testdata/telegraf-agent.toml")
 	assert.NoError(t, err)
 	a, _ := NewAgent(c)
 	assert.Equal(t, 1, len(a.Config.Inputs))
 
 	c = config.NewConfig()
 	c.InputFilters = []string{"foo"}
-	err = c.LoadConfig("../config/testdata/telegraf-agent.toml", "")
+	err = c.LoadConfig("../config/testdata/telegraf-agent.toml")
 	assert.NoError(t, err)
 	a, _ = NewAgent(c)
 	assert.Equal(t, 0, len(a.Config.Inputs))
 
 	c = config.NewConfig()
 	c.InputFilters = []string{"mysql", "foo"}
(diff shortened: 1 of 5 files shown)


reimda:
The environment variable INFLUX_TOKEN is only used in telegraf code in these lines. You can already use one environment variable to hold the token used to get the config, and a different environment variable to hold the token for writing metrics. You just need to change the environment variable in the config file to be something other than INFLUX_TOKEN, for example, INFLUX_WRITE_TOKEN. Then telegraf will use INFLUX_TOKEN to get config and INFLUX_WRITE_TOKEN for writing metrics.

I don't think it's necessary to make telegraf try INFLUX_CONFIG_TOKEN first, and if it isn't set, fall back to INFLUX_TOKEN.

Issue #9211, opened 2021-04-28
outputs.influxdb writing to missing databases or retentions result in endless retries


passing:
### Relevant telegraf.conf:
```toml
[agent]
    flush_interval = "1s"

[[inputs.influxdb_listener]]
    database_tag = "__database__"
    retention_policy_tag = "__retention__"

[[outputs.influxdb]]
    urls = ["http://influxdb:8086"]
    skip_database_creation = true
    database_tag = "__database__"
    exclude_database_tag = true
    retention_policy_tag = "__retention__"
    exclude_retention_policy_tag = true

[[outputs.file]]
    files = ["stdout"]
```

### Reproduce the issue

with the `telegraf.conf` file, execute the following to run an influxdb and telegraf in docker:
```sh
docker run -d --rm --name influxdb -p 8086:8086 influxdb:1.8.3
curl "http://localhost:8086/query" -d "q=CREATE DATABASE telegraf WITH DURATION 1d"

docker run --rm -ti --link influxdb -v $(pwd):/etc/telegraf -p 8186:8186 telegraf:1.18.1

# to cleanup later, run: docker stop influxdb
```

now metrics can be sent like this
```sh
curl "http://localhost:8186/write?db=telegraf" -d "test value=1"
curl "http://localhost:8186/write?db=telegraf&rp=autogen" -d "test value=1"
```

#### Problem 1

writing to an invalid database:
```sh
curl "http://localhost:8186/write?db=invalid" -d "test value=1"
```

results in this error repeating forever
```
2021-04-28T12:37:11Z E! [outputs.influxdb] When writing to [http://influxdb:8086]: 404 Not Found: database not found: "invalid"
2021-04-28T12:37:11Z E! [agent] Error writing to outputs.influxdb: could not write any address
```

#### Problem 2
writing to an invalid retention:
```sh
curl "http://localhost:8186/write?db=telegraf&rp=invalid" -d "test value=1"
```

results in this error repeating forever
```
2021-04-28T12:37:32Z E! [outputs.influxdb] When writing to [http://influxdb:8086]: 500 Internal Server Error: retention policy not found: invalid
2021-04-28T12:37:32Z E! [agent] Error writing to outputs.influxdb: could not write any address
```

### Expected behavior:

it would be expected, that when these errors occur, the metrics get dropped and don't get retried forever (while the telegraf buffer is filling up more and more)

### Actual behavior:

metrics fill up the buffer of telegraf and writing to influx is retried forever

PR #9296, opened 2021-05-24
Bugfix outputs influxdb endless retires

MyaLongmire:
### Required for all PRs:

- [ x ] Updated associated README.md. - no update needed
- [ x ] Wrote appropriate unit tests. - no new unit test needed

[outputs.influxdb writing to missing databases or retentions result in endless retries #9211](https://github.com/influxdata/telegraf/issues/9211)

Fixed the repeating 404: database not found error by checking for the error and that skipping database creation is set to true in the config file and then printing an error and returning.

Fixed the repeating 500: Interal Server Error: no retention policy error by handling it as an API error.


plugins/outputs/influxdb/http.go
```diff
@@ -317,7 +318,6 @@ func (c *httpClient) Write(ctx context.Context, metrics []telegraf.Metric) error
 					c.config.URL, dbrp.Database, err)
 			}
 		}
-
```
helenosheaa, line 320: add back in the new line here

helenosheaa, line 320: add back in new line

plugins/outputs/influxdb/http.go
```diff
@@ -374,6 +374,12 @@ func (c *httpClient) writeBatch(ctx context.Context, db, rp string, metrics []te
 		desc = writeResp.Err
 	}
 
+	//checks for any 4xx code and drops metric and retrying will not make the request work
+	if len(resp.Status) > 0 && resp.Status[0] == '4' && c.config.SkipDatabaseCreation {
+		c.log.Errorf("E! [outputs.influxdb] Failed to write metric (will be dropped: %s): %s\n", resp.Status, desc)
+		return nil
+	}
```
helenosheaa, line 381: I'm wondering if we should handle this case within the database not found error handling which is influxdb.go file from line 222. As the error is the same "database not found", but we weren't handling the case where SkipDatabaseCreation is true.

helenosheaa (commented):
Thanks for the PR! Just a question around the location of one part of the logic.

sspaink (approved):

Issue #5720, opened 2019-04-12
Improve speed of snmp plugin loading


danielnelson:
### Current behavior:

For each plugin, snmptranslate is called on each field, information is not shared between plugin instances.

### Desired behavior:

Add a shared cache for loading SNMP MIB data to improve the startup speed when using large tables, many fields, or many plugins.  Use more efficient lookup strategy, perhaps `snmptranslate -Tso -m <mib-file>` to avoid loading the cache file multiple times.

When Telegraf is reloaded, the cache should be cleared and rebuilt.

### Use case:

Currently Telegraf can be very slow loading SNMP inputs, in some cases taking up to 10 minutes.  This is made worse when you have large MIB files.

danielnelson:
This is also an issue for the snmp_trap input: https://github.com/influxdata/telegraf/issues/4377#issuecomment-570390135

danielnelson:
The gosmi project has native parsing of MIBs now: https://github.com/sleepinggenius2/gosmi, this could potentially be a way to address.

Hipska:
Indeed, I also mentioned this library for #6330

loganmc10:
I did a real quick and dirty implementation of gosmi for the snmp_trap plugin (I needed it for my case, snmptranslate was too slow and Telegraf was pinned at 100% CPU).

https://github.com/loganmc10/telegraf/commit/b019c8891ca8afdc98b1251f4a4cddd2ee01b6ce

I'm not really a Go programmer so I don't feel too comfortable bringing this all the way to a PR, but hopefully that commit can be helpful to someone that wants to submit a PR.

The performance difference is quite dramatic, I was pinned at 100% CPU before, now Telegraf is at about 1% CPU usage...

reimda:
Hi @loganmc10, thanks for sharing your work with gosmi and Telegraf. If you're not really a Go programmer you're doing a good job of faking it.

I understand that you don't feel comfortable working through the PR process to get your work merged into telegraf. If you are ok with someone starting from your branch and improving it and getting it reviewed and merged into Telegraf, would you be able to submit it as a PR? Even though it looks like you've already signed the Telegraf CLA, I don't think we're allowed to take code from your branch unless you submit it as a PR. You could mention in the PR description that you're not able to make more changes and that someone else should take it over. Thanks!

loganmc10:
Sure: https://github.com/influxdata/telegraf/pull/9307 anyone has my permission to edit/modify/improve as needed

PR #9307, opened 2021-05-26
Use gosmi for SNMP traps

loganmc10:
As requested in: https://github.com/influxdata/telegraf/issues/5720#issuecomment-849107895

I'm submitting a very early implementation of using gosmi to translate SNMP OIDs for the snmp_trap plugin.

Using snmptranslate does not work well. It's slow to call it, and it requires a mutex lock to deal with the cache, which pretty much destroys any parallelism.

We ingest about 30 traps per second. Because of the varying OID index values coming in, the snmptranslate cache didn't really work as expected for us. Telegraf was pinned at 100% CPU usage.

I'm not a professional Go programmer and I'm not so great at unit tests. I'm hoping someone can take it from here and finish it off. We're using this implementation daily and it appears to work well for us. (less than 1% CPU usage from Telegraf to process those 30 traps/second).

The biggest shortfall of what I've upload here is that it uses a hardcoded path of ```/usr/share/snmp/mibs``` for the MIBs, it should probably be something that is configured as a plugin option. Also, the folder recursion is not very good, it doesn't go multiple levels into the folder structure.

There is also a ```gosmi.Exit()``` function that should be called to clean things up that I never implemented here.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,7 +18,7 @@ require (
 	github.com/Shopify/sarama v1.27.2
 	github.com/StackExchange/wmi v0.0.0-20210224194228-fe8f1750fd46 // indirect
 	github.com/aerospike/aerospike-client-go v1.27.0
-	github.com/alecthomas/units v0.0.0-20190924025748-f65c72e2690d
+	github.com/alecthomas/units v0.0.0-20210208195552-ff826a37aa15
 	github.com/aliyun/alibaba-cloud-sdk-go v1.61.1004
 	github.com/amir/raidman v0.0.0-20170415203553-1ccc43bfb9c9
 	github.com/antchfx/xmlquery v1.3.5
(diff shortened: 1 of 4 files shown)


reimda:
Thanks @loganmc10 for the PR. This is a good start

To address the hardcoded path, on platforms that use net-snmp maybe we should use net-snmp's mib paths? We could have telegraf query net-snmp on startup with `net-snmp-config --default-mibdirs` or `snmptranslate -Dinit_mib .1.3 2>&1 |grep MIBDIR` (see https://net-snmp.sourceforge.io/wiki/index.php/TUT:Using_and_loading_MIBS)

loganmc10:
One of the benefits of this change would be to remove the dependency on net-snmp, so it would probably be better to either read the env variable MIBDIRS (which is what net-snmp uses), or add an option to the plugin that defaults to ```/usr/share/snmp/mibs```

loganmc10:
Closing in favour of https://github.com/influxdata/telegraf/pull/9343

PR #9317, opened 2021-05-28
feat: add test input plugin

### LINKED ISSUE #9316: inputs.test plugin
## Feature Request

support an input plugin of the following type:
```toml
[[inputs.test]]
  metrics = [
    'weather,state=ny temperature=81.3'
  ]
```

### Proposal:

support the above

### Current behavior:

possible through creating a file and using inputs.file or inputs.tail

### Desired behavior:

above

### Use case: 

Very useful for debugging and creating demos of processor/aggregator behavior. Allows self-contained config files that demo a specific configuration behavior

In the future this could potentially be paired with a special output that expects that a specific metric shows up; could be an interesting way to write documentation-focused tests.

akrantz01:
### Required for all PRs:

resolves #9316

Added an input plugin to gather constant metrics as defined by the plugin's configuration. Each interval the defined metrics are parsed from the desired format and outputted.

### DIFF
--- a/plugins/inputs/all/all.go
+++ b/plugins/inputs/all/all.go
@@ -182,6 +182,7 @@ import (
 	_ "github.com/influxdata/telegraf/plugins/inputs/teamspeak"
 	_ "github.com/influxdata/telegraf/plugins/inputs/temp"
 	_ "github.com/influxdata/telegraf/plugins/inputs/tengine"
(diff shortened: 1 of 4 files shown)


MyaLongmire:
As this has been sitting for a while and the same thing can be accomplished with tail or file input I am going to close this.

PR #9319, opened 2021-05-28
feat(x509_cert): add proxy support

### LINKED ISSUE #9174: Add proxy support for inputs plugin x509_cert
## Feature Request

Add proxy support for inputs plugin x509_cert 

### Proposal:
add option **http_proxy**  as in http_response to support access a target via proxy

### Current behavior:
no proxy support

### Desired behavior:
proxy support is added

### Use case: 
Can't access a target site directly and need to configure a proxy

akrantz01:
### Required for all PRs:

resolves #9174

Adds proxy support for the `x509_cert` plugin. The supported proxy methods are HTTP CONNECT (TCP only) and SOCKS5.

~~**Note:** #9289 should be merged before this PR since all URLs are currently parsed as globs causing telegraf to fail to pull the certificate. Similarly, the `TestGatherTCPCert` test will fail because of this.~~ Rebased to master.

### DIFF
--- a/plugins/common/proxy/proxy.go
+++ b/plugins/common/proxy/proxy.go
@@ -2,6 +2,7 @@ package proxy
 
 import (
 	"fmt"
+	"golang.org/x/net/proxy"
 	"net/http"
 	"net/url"
 )
@@ -21,4 +22,20 @@ func (p *HTTPProxy) Proxy() (proxyFunc, error) {
 		return http.ProxyURL(address), nil
 	}
 	return http.ProxyFromEnvironment, nil
+}
+
+type TCPProxy struct {
+	ProxyURL string `toml:"proxy_url"`
+}
+
+func (p *TCPProxy) Proxy() (proxy.Dialer, error) {
+	if len(p.ProxyURL) > 0 {
+		parsed, err := url.Parse(p.ProxyURL)
+		if err != nil {
+			return nil, fmt.Errorf("error parsing proxy url %q: %w", p.ProxyURL, err)
+		}
+
+		return proxy.FromURL(parsed, proxy.Direct)
+	}
+	return proxy.FromEnvironment(), nil
 }
--- a/plugins/common/proxy/proxy.go
+++ b/plugins/common/proxy/proxy.go
@@ -9,39 +9,50 @@ import (
 )
 
 type HTTPProxy struct {
+	UseProxy     bool   `toml:"use_proxy"`
 	HTTPProxyURL string `toml:"http_proxy_url"`
 }
 
 type proxyFunc func(req *http.Request) (*url.URL, error)
 
 func (p *HTTPProxy) Proxy() (proxyFunc, error) {
-	if len(p.HTTPProxyURL) > 0 {
-		address, err := url.Parse(p.HTTPProxyURL)
-		if err != nil {
-			return nil, fmt.Errorf("error parsing proxy url %q: %w", p.HTTPProxyURL, err)
+	if p.UseProxy {
+		if len(p.HTTPProxyURL) > 0 {
+			address, err := url.Parse(p.HTTPProxyURL)
+			if err != nil {
+				return nil, fmt.Errorf("error parsing proxy url %q: %w", p.HTTPProxyURL, err)
+			}
+			return http.ProxyURL(address), nil
 		}
-		return http.ProxyURL(address), nil
+
+		return http.ProxyFromEnvironment, nil
 	}
-	return http.ProxyFromEnvironment, nil
+
+	return nil, nil
 }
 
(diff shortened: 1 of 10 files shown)


mariuszjedrzejewski:
Is there any chances to resolve conficts and merge proxy support changes? Can't wait for this feature...

sspaink:
@akrantz01 do you have time rebase this pr? Thanks!

akrantz01:
Done!

powersj:
I pushed a commit to resolve some conflicts post-README-revamp that was done. It required removing the description/sample-config consts and a test. Will wait on CI and then merge.

powersj:
@akrantz01 I tried, but seem to have failed :) would you mind rebasing once more? sorry!

akrantz01:
I'm a bit confused why `TestGatherTCPCert` needed to be removed. It seems like it would be helpful to keep that test.

powersj:
> I'm a bit confused why `TestGatherTCPCert` needed to be removed. It seems like it would be helpful to keep that test.

You are right, I meant to only remove `TestStrings`

plugins/common/proxy/proxy.go
```diff
@@ -22,3 +24,25 @@ func (p *HTTPProxy) Proxy() (proxyFunc, error) {
...
+		if err != nil {
+			return nil, fmt.Errorf("error parsing proxy url %q: %w", p.ProxyURL, err)
+		}
+
+		if dialer, err = proxy.FromURL(parsed, proxy.Direct); err != nil {
+			return nil, err
+		}
+	} else {
+		dialer = proxy.FromEnvironment()
+	}
```
reimda, line 45: If the user doesn't set the proxy url, I would expect not to use a proxy at all, not to use the system proxy settings.

Does proxy.FromEnvironment() not use a proxy in this case or does it use the system proxy settings?

Maybe we need two settings: proxy_url for the url, and something like use_proxy (bool). If use_proxy is true but proxy_url isn't set, then use the system proxy.

reimda (changes requested):

plugins/common/proxy/proxy.go
powersj: Having worked on a corporate network that required a proxy for some things, but not all, I prefer that applications only use the proxy when explicitly told to.

plugins/common/proxy/proxy.go
```diff
@@ -4,21 +4,56 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+
+	"golang.org/x/net/proxy"
 )
 
 type HTTPProxy struct {
+	UseProxy     bool   `toml:"use_proxy"`
```
powersj, line 12: Can we rename `UseProxy` to `UseSystemProxy`? and `use_proxy` to `use_system_proxy`

plugins/common/proxy/proxy.go
```diff
@@ -4,21 +4,56 @@ import (
...
 }
 
 type proxyFunc func(req *http.Request) (*url.URL, error)
 
 func (p *HTTPProxy) Proxy() (proxyFunc, error) {
-	if len(p.HTTPProxyURL) > 0 {
-		address, err := url.Parse(p.HTTPProxyURL)
-		if err != nil {
-			return nil, fmt.Errorf("error parsing proxy url %q: %w", p.HTTPProxyURL, err)
+	if p.UseProxy {
```
powersj, line 19: @reimda, @sspaink and I were chatting and we believe this would break existing users who only have the proxy URL defined. In order to add this in a backwards-compatible way we need to implement a priority as such:

1. system proxy: if the `UseSystemProxy` is enabled, use the system proxy
2. HTTP proxy URL: if the `HTTPProxyURL` is set use this
3. no proxy

We may have users with configs today that only have the URL defined and not the boolean.

powersj (changes requested):
@akrantz01 thank you for making the update so quickly! We have one more set of changes, but we know you are probably busy with schools work. If you would rather one of us pick this up please let us know! We really appreciate all the effort you always put into these PRs.

Thanks!

plugins/common/proxy/proxy.go
akrantz01: Sounds good. Should I make `TCPProxy` act the same way for consistency?

MyaLongmire: yes if you have the time :)

reimda (approved):
Thanks Alex!

reimda:
@powersj I think Alex addressed all the concerns. Yours was the last review that requested changes. Could you take another look? If we all like this, we can resolve the conflicts one last time before merging.

powersj:
Rebase + CI and ready to merge

PR #9322, opened 2021-06-01
Suricata alerts

### LINKED ISSUE #9321: Add alert metrics for suricata input
## Feature Request

### Proposal:

Suricata IDS is mostly used to trigger alerts based on network traffic, the proposal is to also parse those alerts in the telegraf suricata input.

### Current behavior:

Currently Suricata Telegraf input is only able to parse metrics of event type "stats": https://suricata.readthedocs.io/en/latest/performance/statistics.html

### Desired behavior:

Optional flag to enable alerts measurements in telegraf.

Input configuration:

```toml
[[inputs.suricata]]
  ...
  # Enable disable alerts measurement
  alerts = true
```

Instead of having one only field `suricata`, I would add multiple for every suricata event type: `suricta_stats`, `suricata_alerts`,  etc for future event types.

Possible output:

```text
suricata_stats,host=myhost,thread=FM#01 flow_mgr_rows_empty=0,flow_mgr_rows_checked=65536,flow_mgr_closed_pruned=0,flow_emerg_mode_over=0,flow_mgr_flows_timeout_inuse=0,flow_mgr_rows_skipped=65535,flow_mgr_bypassed_pruned=0,flow_mgr_flows_removed=0,flow_mgr_est_pruned=0,flow_mgr_flows_notimeout=1,flow_mgr_flows_checked=1,flow_mgr_rows_busy=0,flow_spare=10000,flow_mgr_rows_maxlen=1,flow_mgr_new_pruned=0,flow_emerg_mode_entered=0,flow_tcp_reuse=0,flow_mgr_flows_timeout=0 1568368562545197545
suricata_alerts,host=myhost action=allowed,category=Misc activity,gid=1,rev=0,severity=3,signature=LOCAL_TCP_DOS,signature_id=6 1568368562545174807
```

### Use case: 

Track any alert generated by Suricata IDS

rogercoll:
### Required for all PRs:

- [ x ] Updated associated README.md.
- [ x ] Wrote appropriate unit tests.

resolves #9321

The code add the ability to parse logs of "alert" event type. Currently, there is only support for "stats" type. The PR separates the parse in different functions depending on the event type, it can be very helpful for future types integration.


plugins/inputs/suricata/suricata.go
```diff
@@ -36,18 +37,21 @@ type Suricata struct {
...
   # This is expected to be a filename of a
   # unix socket to be created for listening.
   source = "/var/run/suricata-stats.sock"
 
   # Delimiter for flattening field keys, e.g. subitem "alert" of "detect"
   # becomes "detect_alert" when delimiter is "_".
   delimiter = "_"
+  
+  # Detect alert logs 
+  alerts = false
```
srebhan, line 54: You should leave default field values commented and use a double `#` for comments:
```suggestion
  
  ## Detect alert logs 
  # alerts = false 
```
I see it's also "wrong" in the lines above. So I'd be happy if you can correct those too. :-)

plugins/inputs/suricata/suricata.go
```diff
@@ -150,27 +154,35 @@ func flexFlatten(outmap map[string]interface{}, field string, v interface{}, del
...
 
-func (s *Suricata) parse(acc telegraf.Accumulator, sjson []byte) {
-	// initial parsing
-	var result map[string]interface{}
-	err := json.Unmarshal(sjson, &result)
-	if err != nil {
-		acc.AddError(err)
+func (s *Suricata) parseAlert(acc telegraf.Accumulator, result map[string]interface{}) {
+	if _, ok := result["alert"].(map[string]interface{}); !ok {
+		s.Log.Debug("The 'alert' sub-object does not have required structure")
```
srebhan, line 167: It's only cosmetics but you should change to
```suggestion
		s.Log.Debug("'alert' sub-object does not have required structure")
```

plugins/inputs/suricata/suricata.go
```diff
@@ -150,27 +154,35 @@ func flexFlatten(outmap map[string]interface{}, field string, v interface{}, del
...
-	// check for presence of relevant stats
-	if _, ok := result["stats"]; !ok {
-		s.Log.Debug("Input does not contain necessary 'stats' sub-object")
-		return
+	totalmap := make(map[string]interface{})
+	for k, v := range result["alert"].(map[string]interface{}) {
+		//source and target fields are maps
+		err := flexFlatten(totalmap, k, v, s.Delimiter)
+		if err != nil {
+			s.Log.Debug(err.Error())
```
srebhan, line 176: Can you maybe state in the message where the problem occurred, like
```suggestion
			s.Log.Debugf("Flattening alert failed: %v", err)
```

plugins/inputs/suricata/suricata.go
```diff
@@ -150,27 +154,35 @@ func flexFlatten(outmap map[string]interface{}, field string, v interface{}, del
...
+			// we skip this subitem as something did not parse correctly
```
srebhan, line 177: You might want to make this more "future proof" by explicitly `continue` here. Not a must though.

plugins/inputs/suricata/suricata.go
```diff
@@ -215,6 +227,28 @@ func (s *Suricata) parse(acc telegraf.Accumulator, sjson []byte) {
...
 }
 
+func (s *Suricata) parse(acc telegraf.Accumulator, sjson []byte) {
+	// initial parsing
+	var result map[string]interface{}
+	err := json.Unmarshal(sjson, &result)
+	if err != nil {
+		acc.AddError(err)
+		return
+	}
```
srebhan, line 237: How about actually propagating the error upward?

plugins/inputs/suricata/suricata.go
```diff
@@ -215,6 +227,28 @@ func (s *Suricata) parse(acc telegraf.Accumulator, sjson []byte) {
...
+	_, ok2 := result["alert"]
+	if !ok && !ok2 {
+		s.Log.Debug("Input does not contain necessary 'stats' or 'alert' sub-object")
+		return
+	}
+	if ok {
+		s.parseStats(acc, result)
+	} else if ok2 && s.Alerts {
+		s.parseAlert(acc, result)
+	}
```
srebhan, line 249: I think you should only check for the presence of the `alerts` field if you want to parse them...
```suggestion
	if _, ok := result["stats"]; !ok {
		s.Log.Debugf("Invalid input without 'stats': %v", result)
		return fmt.Errorf("input does not contain 'stats' object")
	}
	if _, ok2 := result["alert"]; s.Alerts && !ok2 {
		s.Log.Debugf("Invalid input without 'alert' when trying to parse them: %v", result)
		return fmt.Errorf("input does not contain 'alert' object")
	}
	
	s.parseStats(acc, result)
	if s.Alerts {
		s.parseAlert(acc, result)
	}
```

plugins/inputs/suricata/suricata.go
```diff
@@ -226,6 +260,7 @@ func init() {
 		return &Suricata{
 			Source:    "/var/run/suricata-stats.sock",
 			Delimiter: "_",
+			Alerts:    false,
```
srebhan, line 263: You don't need to set this as it is the default in golang.

plugins/inputs/suricata/README.md
```diff
@@ -94,6 +98,19 @@ All fields are numeric.
...
     - ...
 
+Some fields of the Suricata alerts are strings, for example the signatures. See https://suricata.readthedocs.io/en/suricata-6.0.0/output/eve/eve-json-format.html?highlight=priority#event-type-alert for more information.
+
+- suricata_alert
+  - fields:
+    - action
+    - gid
+    - severity
+    - signature
```
srebhan, line 108: Does it make sense to convert `severity` and maybe also `action` to a tag? You could add a
```toml
as_tag = ["action", "severity"]
```
list option for this to allow the user to make this choice...

srebhan (commented):
Thank you very much @rogercoll for this PR. It's very nicely written. I do have only some minor comments in the code. Can you please check those.

plugins/inputs/suricata/suricata.go
rogercoll: Agree

rogercoll:
@srebhan Thanks, sure! Very good comments. Just reviewing them and running tests again.

plugins/inputs/suricata/suricata.go
rogercoll: @srebhan With your solution we force that "stats" event type must be always present. I have seen in some light environments that only "alert" types are configured as outputs in suricata. What do you think on making result["stats"] mandatory?

plugins/inputs/suricata/suricata.go
rogercoll: Great!

plugins/inputs/suricata/suricata.go
rogercoll: Make sense

plugins/inputs/suricata/suricata.go
rogercoll: Indeed, the Suricata event_type can only be of one type.

plugins/inputs/suricata/README.md
rogercoll: That would be very useful, working on it. Thanks

rogercoll: In some cases severity is not defined, I would not add this (like the current stats approach) to prevent errors

plugins/inputs/suricata/suricata.go
srebhan: Understood, so you either see `stats` **OR** `alerts`. I this correct? If so, please mark this as resolved even though I don't like `ok2`... :-)

plugins/inputs/suricata/README.md
srebhan: Can be done by processors later on.

plugins/inputs/suricata/suricata.go
```diff
@@ -101,7 +105,10 @@ func (s *Suricata) readInput(ctx context.Context, acc telegraf.Accumulator, conn
 			if rerr != nil {
 				return rerr
 			} else if len(line) > 0 {
```
srebhan, line 107: I think our linter will complain here. You should just have
```suggestion
			if rerr != nil {
				return rerr
			} 
			if len(line) > 0 {
```
as you exit on error anyway making the `else` superfluous.

plugins/inputs/suricata/suricata.go
```diff
@@ -150,27 +157,36 @@ func flexFlatten(outmap map[string]interface{}, field string, v interface{}, del
 		}
 	case float64:
 		outmap[field] = v.(float64)
+	case string:
+		outmap[field] = v.(string)
```
srebhan, line 161: Probably also a leftover from old code, but you can simplify to
```suggestion
	case float64:
		outmap[field] = t
	case string:
		outmap[field] = t
```

srebhan (commented):
@rogercoll thanks for the quick response. Found two more cosmetics :-) and would be happy if you can check them. I'll also trigger the CI tests and if those are ok I guess we are good to go.

plugins/inputs/suricata/suricata.go
rogercoll: Exactly. Neither do I. I we add more fields we will have to be sure to properly handle them. Thanks.

rogercoll:
@srebhan All comments closed. I see that the testing pipeline for mac is failing. I thing this is not related to this PR, am I right?

Thanks

rogercoll:
@srebhan Is there any update on this? Can we rerun the pipelines?

Thanks

srebhan:
!retry-failed

srebhan (approved):
Looks good to me. Thanks for submitting this!

reimda (changes requested):
Would you add a test case that covers the output metric format? In suricata_test.go, the func TestSuricata covers the existing "suricata" metric but we also need to test the new "suricata_alert" metric.

rogercoll:
@reimda sure, I will get back to you.

Thanks

rogercoll:
@reimda "suricata_alert" metric test added in new commits.

Thanks!

srebhan (approved):
Still looks good to me. :-)

sspaink (approved):
lgtm

wmandra:
@rogercoll this works well, however I'd like to make one suggestion. The current implementation will only output source/destination ip/port fields for rules where the "target" metadata keyword is used. Unfortunately none of the rules signatures provided by et/open, oisf, tgreen/hunting, or sslbl use this keyword, and neither do any of the default rules. 

However, src_ip, src_port, dest_ip, dest_port are included in the json data heading along with other useful fields like in_iface and proto.

The test data in the PR doesn't accurately reflect 90+% of what will be generated by actual rules in production. I would suggest adding test4.json as:
`{"timestamp":"2021-05-30T20:07:13.208777+0200","flow_id":1696236471136137,"in_iface":"s1-suricata","event_type":"alert","src_ip":"10.0.0.5","src_port":18715,"dest_ip":"179.60.192.3","dest_port":80,"proto":"TCP","alert":{"action":"allowed","gid":1,"signature_id":6,"rev":0,"signature":"Corrupted HTTP body","severity": 3,"category":"Misc activity","severity":3},"flow":{"pkts_toserver":1,"pkts_toclient":0,"bytes_toserver":174,"bytes_toclient":0,"start":"2021-05-30T20:07:13.208777+0200"}}`

This is more accurate and will then cause the unit test to fail as the plugin will not parse the src/dest fields in the event.

wmandra:
As far as I can see on a live Suricata installation src_ip, src_port, dest_ip, dest_port, and proto fields are always included in the outer json of the alert event. I haven't seen a single alert event where source and target are included within the inner "alert" json.

PR #9323, opened 2021-06-01
Prevent x509_cert from hanging on UDP connection

### LINKED ISSUE #9310: x509_cert hangs when using UDP connection
Using the `x509_cert` input plugin, if you try to get metrics from a certificate served over UDP, the plugin will seemingly hang forever. I believe this is because the `crypto/tls` package does not support TLS over UDP (DTLS).

### Relevant telegraf.conf:

```toml
[[inputs.x509_cert]]
  sources = ["udp://127.0.0.1:4444"]

[[outputs.file]]
```

### System info:

Telegraf: 1.18.3
OS: Ubuntu 20.04.2
OpenSSL: 1.1.1f

### Steps to reproduce:

1. Generate a certificate keypair using OpenSSL
```shell
openssl ecparam -out key.pem -name prime256v1 -genkey
openssl req -new -sha256 -key key.pem -out server.csr
openssl x509 -req -sha256 -days 365 -in server.csr -signkey key.pem -out cert.pem
```
2. Start a DTLS server using OpenSSL
```shell
openssl s_server -dtls -cert cert.pem -key key.pem -accept 4444
```
3. Start telegraf

### Expected behavior:

Telegraf outputs the certificate information. [Logs](https://gist.github.com/akrantz01/a595e8e75907b788c26230b0a73a46ad#file-expected-log)

### Actual behavior:

Telegraf hangs indefinitely and cannot be stopped with ctrl+c. [Logs](https://gist.github.com/akrantz01/a595e8e75907b788c26230b0a73a46ad#file-actual-log)

akrantz01:
### Required for all PRs:

resolves #9310 

Fixes the bug where the `x509_cert` plugin would hang indefinitely when you try to gather certificate metrics from a UDP address. This was done by adding UDP support to the plugin using the [pion/dtls](https://github.com/pion/dtls) library.

~~**Note**: #9289 should be merged before this PR since all URLs are currently parsed as globs causing telegraf to fail to pull the certificate. Similarly, the `TestGatherUDPCert` test will fail because of this.~~ Rebased to master.


helenosheaa (commented):
Looks good! 

One note: if you look at the ci/circleci: test-go-1_15  check you'll see CircleCI is complaining about additional licenses needing to be added for the following:

```
+github.com/pion/logging
+github.com/pion/transport
+github.com/pion/udp
```

jjh74:
@akrantz01 Does udp:// mean DTLS(rfc6347) after this PR ? (udp:// only works with DTLS servers and not with some homegrown tls on top of udp socket). Maybe `README.md` could mention that udp:// means DTLS ?

akrantz01:
That'd probably be good to mention. However, there shouldn't be any additional problems with homegrown implementations over UDP since previously, it fails to send a packet if you configure it to pull from a UDP address as the standard library isn't equipped to handle it.

plugins/inputs/x509_cert/x509_cert_test.go
```diff
@@ -260,6 +262,36 @@ func TestGatherChain(t *testing.T) {
...
+	defer listener.Close()
+
+	go func() {
+		_, err := listener.Accept()
+		require.NoError(t, err)
+	}()
+
+	m := &X509Cert{
+		Sources: []string{"udp://" + listener.Addr().String()},
+	}
```
srebhan, line 285: You need to set the logging facility here or you will get segfaults (that are the failing CI tests you see):
```suggestion
	m := &X509Cert{
		Sources: []string{"udp://" + listener.Addr().String()},
		Log:     testutil.Logger{},
	}
```

srebhan (commented):
@akrantz01 thanks for looking into this. The code looks quite good and I added a comment to fix the failing CI tests. Let me know if I can be of any further help.

plugins/inputs/x509_cert/x509_cert_test.go
```diff
@@ -260,6 +262,37 @@ func TestGatherChain(t *testing.T) {
...
+
+	addr := &net.UDPAddr{IP: net.ParseIP("127.0.0.1"), Port: 0}
+	listener, err := dtls.Listen("udp", addr, cfg)
+	require.NoError(t, err)
+	defer listener.Close()
+
+	go func() {
+		_, err := listener.Accept()
+		require.NoError(t, err)
```
srebhan, line 280: You shall not use any `require` methods that terminate (i.e. call `require.Fail` internally) in goroutines as this will only end the goroutine and not the test. See [the go release notes](https://tip.golang.org/doc/go1.16#vet) for reference also describing a potential workaround.

In this specific instance I think just ignoring the error and quitting the goroutine might be sufficient as the test relying on the accept will probably fail anyway...

srebhan (approved):
Looks good to me. Thanks for the nice PR!

MyaLongmire (approved):

PR #9339, opened 2021-06-08
feat: Auto-create bucket when non-existent in influxdb_v2

### LINKED ISSUE #9301: [influx 2.0 output plugin] auto-create bucket if not already present in influx 2.0
## Feature Request

We are using telegraf to write data to influx. The behavior of influxdb output plugin for influx v1.x creates the database specified it not already present. The same is not true of buckets in influx v2.0 

### Proposal: 
The buckets mentioned in the output plugin should be created if not created already 

### Current behavior:
The influx v2.0 bucket to which data should be written should already be present 

### Desired behavior:
If the bucket specified is not present, the bucket should be created and data to should be written to that bucket

### Use case: 
1. When moving from influx v1.x where telegraf creates databases if not present, telegraf taking care of creating buckets if not present provides parity 
2. When multiple database have to moved from influx 1.x to 2.x, the buckets would have to be created manually

akrantz01:
### Required for all PRs:

resolves #9301

Adds automatically creating a bucket if it doesn't exist in the `influxdb_v2` output plugin.

### DIFF
--- a/plugins/outputs/influxdb_v2/README.md
+++ b/plugins/outputs/influxdb_v2/README.md
@@ -33,7 +33,7 @@ The InfluxDB output plugin writes metrics to the [InfluxDB v2.x] HTTP service.
   ## If true, no create bucket requests will be sent. Set to true when using
   ## Telegraf with a user without permissions to create buckets or when the
   ## the bucket already exists
-  # skip_database_creation = false
+  # skip_bucket_creation = false
 
   ## Timeout for HTTP messages.
   # timeout = "5s"
--- a/plugins/outputs/influxdb_v2/README.md
+++ b/plugins/outputs/influxdb_v2/README.md
@@ -30,6 +30,11 @@ The InfluxDB output plugin writes metrics to the [InfluxDB v2.x] HTTP service.
   ## If true, the bucket tag will not be added to the metric.
   # exclude_bucket_tag = false
 
+  ## If true, no create bucket requests will be sent. Set to true when using
+  ## Telegraf with a user without permissions to create buckets or when the
+  ## the bucket already exists
+  # skip_database_creation = false
+
   ## Timeout for HTTP messages.
   # timeout = "5s"
 
--- a/plugins/outputs/influxdb_v2/README.md
+++ b/plugins/outputs/influxdb_v2/README.md
@@ -30,13 +30,11 @@ The InfluxDB output plugin writes metrics to the [InfluxDB v2.x] HTTP service.
   ## If true, the bucket tag will not be added to the metric.
   # exclude_bucket_tag = false
 
-  ## If true, no create bucket requests will be sent. Set to true when using
-  ## Telegraf with a user without permissions to create buckets or when the
-  ## the bucket already exists
-  # skip_bucket_creation = false
+  ## Creates any buckets if they do not already exist
+  # create_buckets = false
 
   ## Default retention rule for any newly created buckets. If unset, the default
-  ## retention rule is forever. Ignored if skip_bucket_creation is false.
+  ## retention rule is forever. Ignored if create_buckets is false.
   # default_bucket_retention = "0s"
 
   ## Timeout for HTTP messages.
--- a/plugins/outputs/influxdb_v2/README.md
+++ b/plugins/outputs/influxdb_v2/README.md
@@ -35,6 +35,10 @@ The InfluxDB output plugin writes metrics to the [InfluxDB v2.x] HTTP service.
   ## the bucket already exists
   # skip_bucket_creation = false
 
+  ## Default retention rule for any newly created buckets. If unset, the default
+  ## retention rule is forever. Ignored if skip_bucket_creation is false.
+  # default_bucket_retention = "0s"
+
   ## Timeout for HTTP messages.
   # timeout = "5s"
 
--- a/plugins/outputs/influxdb_v2/http.go
+++ b/plugins/outputs/influxdb_v2/http.go
@@ -59,38 +59,45 @@ type orgInfo struct {
 
 // createBucketRequest is the payload used for creating a bucket
 type createBucketRequest struct {
-	Name  string `json:"name"`
-	OrgID string `json:"orgID"`
-	// TODO: support custom retention rule
+	Name           string          `json:"name"`
+	OrgID          string          `json:"orgID"`
+	RetentionRules []retentionRule `json:"retentionRules"`
+}
+
+type retentionRule struct {
+	EverySeconds int64  `json:"everySeconds"`
+	Type         string `json:"type"`
 }
 
 type HTTPConfig struct {
-	URL                *url.URL
-	Token              string
-	Organization       string
-	Bucket             string
-	BucketTag          string
-	ExcludeBucketTag   bool
-	SkipBucketCreation bool
-	Timeout            time.Duration
-	Headers            map[string]string
-	Proxy              *url.URL
-	UserAgent          string
-	ContentEncoding    string
-	TLSConfig          *tls.Config
+	URL                    *url.URL
+	Token                  string
+	Organization           string
+	Bucket                 string
+	BucketTag              string
+	ExcludeBucketTag       bool
+	SkipBucketCreation     bool
+	DefaultBucketRetention int64
+	Timeout                time.Duration
+	Headers                map[string]string
+	Proxy                  *url.URL
+	UserAgent              string
+	ContentEncoding        string
+	TLSConfig              *tls.Config
 
 	Serializer *influx.Serializer
 }
 
 type httpClient struct {
-	ContentEncoding    string
-	Timeout            time.Duration
-	Headers            map[string]string
-	Organization       string
-	Bucket             string
-	BucketTag          string
-	ExcludeBucketTag   bool
-	SkipBucketCreation bool
+	ContentEncoding        string
+	Timeout                time.Duration
+	Headers                map[string]string
+	Organization           string
+	Bucket                 string
+	BucketTag              string
+	ExcludeBucketTag       bool
+	SkipBucketCreation     bool
+	DefaultBucketRetention int64
 
 	client               *http.Client
 	createBucketExecuted map[string]bool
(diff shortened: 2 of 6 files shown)


plugins/outputs/influxdb_v2/http.go
```diff
@@ -201,12 +238,19 @@ func (c *httpClient) Write(ctx context.Context, metrics []telegraf.Metric) error
 		}
 
 		for bucket, batch := range batches {
+			if !c.SkipBucketCreation && !c.createBucketExecuted[bucket] {
+				if err := c.CreateBucket(ctx, bucket); err != nil {
+					log.Printf("W! [outputs.influxdb_v2] When writing to [%s]: bucket %q creation failed: %v\n", c.URL(), bucket, err)
```
reimda, line 243: Doesn't it need to fail early in this case so it doesn't try to write the batch later to a nonexistent bucket? Or are there times when the code tries to create a bucket that already exists?

Do we need to approach bucket creation differently? Maybe we should try to write the batch first, watching for "bucket not found" errors. If there's no bucket then create the bucket and retry.

plugins/outputs/influxdb_v2/http.go
```diff
@@ -303,6 +358,124 @@ func (c *httpClient) writeBatch(ctx context.Context, bucket string, metrics []te
...
+	if resp.StatusCode == 200 && len(orgIDResp.Orgs) == 1 {
+		return orgIDResp.Orgs[0].ID, nil
+	}
+
+	return "", fmt.Errorf("failed to get ID for org %q (do you have org-level read permissions?)", c.Organization)
+}
+
+// CreateBucket creates a new bucket in the configured organization if it doesn't already exist
+func (c *httpClient) CreateBucket(ctx context.Context, bucket string) error {
+	orgID, err := c.getOrgID(ctx)
```
reimda, line 403: The org ID never changes, right? Should we get it only once and reuse it in subsequent calls to CreateBucket?

plugins/outputs/influxdb_v2/README.md
```diff
@@ -30,6 +30,15 @@ The InfluxDB output plugin writes metrics to the [InfluxDB v2.x] HTTP service.
   ## If true, the bucket tag will not be added to the metric.
   # exclude_bucket_tag = false
 
+  ## If true, no create bucket requests will be sent. Set to true when using
+  ## Telegraf with a user without permissions to create buckets or when the
+  ## the bucket already exists
+  # skip_bucket_creation = false
```
reimda, line 36: To make this backward compatible, we need to default to the plugin's previous behavior and let users opt in to get the new behavior. Also, using a negative setting (`skip_action = false`) is not as clear as a positive setting (`act = true`) for users writing their toml config and for devs reading the code.

I understand the v1 plugin has `# skip_database_creation = false` and you're trying to make it match. A negative setting makes a little more sense in v1 because db creation was the default from the start. I don't think we need to mimic the v1 setting so closely since the v2 plugin has been around for years and never defaulted to creating buckets. 

 Could we change this to `# create_buckets = false`?

plugins/outputs/influxdb_v2/README.md
```diff
@@ -30,6 +30,15 @@ The InfluxDB output plugin writes metrics to the [InfluxDB v2.x] HTTP service.
...
+
+  ## Default retention rule for any newly created buckets. If unset, the default
+  ## retention rule is forever. Ignored if skip_bucket_creation is false.
+  # default_bucket_retention = "0s"
```
reimda, line 40: The v1 plugin also lets you specify the retention policy in a metric tag, not just a toml setting (see `retention_policy_tag` and `exclude_retention_policy_tag`). I wonder if we need them here. I don't know how often they're used and how necessary they are.

Do you think we need to add similar functionality to get the retention policy from the metric?

reimda (changes requested):

plugins/outputs/influxdb_v2/README.md
akrantz01: The retention policy is only needed when the bucket is created. The tag would only be necessary if the user was using `bucket_tag` and wanted different retention policies for buckets that did not already exist.

It would be trivial to implement. However, I'm not sure it's worth having a tag on every metric for the retention policy when it may never be used. Though I guess that decision could be left up to the user.

reimda: I think the users who use v1's retention_policy_tag do have them on every metric.

Let's not add the retention policy tags functionality in this PR. If there are users who need it with v2, they will file an issue and we can add it in the future.

plugins/outputs/influxdb_v2/http.go
akrantz01: Honestly, I'm not super sure why it's there. It's from the InfluxDB v1 plugin, which I was using as a reference.

My best guess is that it was to prevent unnecessary calls to create a database. However, I think it may have been accidentally circumvented since an attempt to create a database is made if a `DatabaseNotFoundError` was encountered. This would happen if the first `CreateDatabase` call failed with a 403 error since the database wouldn't have been created, but the `createDatabaseExecuted` entry would be `true`.

Moving all the `createBucketExecuted` stuff into `CreateBucket` and having it only create the bucket on when a `BucketNotFound` error is encountered would be a viable option. So then, as you suggested, we would always try to write the batch first, creating the bucket on any failure, and then retrying on the next flush interval..`

sspaink (approved):

plugins/outputs/influxdb_v2/influxdb.go
```diff
@@ -164,37 +176,53 @@ func (i *InfluxDB) Write(metrics []telegraf.Metric) error {
 			return nil
 		}
 
+		if apiError, ok := err.(*BucketNotFoundError); ok {
+			if i.CreateBuckets {
+				if err := client.CreateBucket(ctx, apiError.Bucket); err != nil {
+					i.Log.Errorf("When writing to [%s]: bucket %q not found and failed to recreate", client.URL(), apiError.Bucket)
+				}
```
reimda, line 183: Do we need to retry the write of the current batch of metrics once we create the bucket? If we don't, the batch is lost for the client.

plugins/outputs/influxdb_v2/influxdb.go
```diff
@@ -164,37 +176,53 @@ func (i *InfluxDB) Write(metrics []telegraf.Metric) error {
...
 	}
 
 	c, err := NewHTTPClient(config)
 	if err != nil {
-		return nil, fmt.Errorf("error creating HTTP client [%s]: %v", url, err)
+		return nil, fmt.Errorf("error creating HTTP client [%s]: %v", u, err)
+	}
+
+	if i.CreateBuckets {
+		if err := c.CreateBucket(ctx, c.Bucket); err != nil {
```
reimda, line 223: Do we need to call createbucket here? If we handle the error in Write, this seems unneeded.

reimda (changes requested):

plugins/outputs/influxdb_v2/influxdb.go
akrantz01: I'm a bit confused by this. It automatically retries the write since an error is returned when the bucket does not exist. Are you suggesting we call `client.Write` again once the bucket is created successfully instead of waiting for the next flush interval?

reimda: Yes, I think we should call client.Write again right after the bucket is created.

The current code tries to write to each node in the cluster in random sequence. If it fails to write to one node because the bucket doesn't exist, it creates the bucket on that node and then moves on to the next node in the sequence to retry the write.

It makes sense to me to write the metrics to the same node we just created the bucket on. Since that node just succeeded in creating a bucket, it's likely healthy (not overloaded) and able to accept the metrics successfully. Why switch to a different node to retry writing?

The InfluxDB.Write func only returns an error if there's an error from all the nodes, so it's not likely that the metrics will have to wait until the next flush interval. I'd prefer not to wait until the next flush interval anyway. Who knows if the metric buffer will run out of space and cause old metrics to be lost. It's safer to handle the error close to where it happened and also not difficult.

It also makes the code more straightforward not to have to deal with multiple cluster nodes while also dealing with missing buckets. There's no reason to have to think about the two concepts at the same time.

ssoroka:
We may want to hold off on this. this is analogous to creating databases when a new record comes in, and it potentially affects permissions and other things, so I'm not sure it's a pattern we want to perpetuate.

reimda (approved):

dgnorton:
Creating a bucket in InfluxDB OSS requires a token with elevated privileges, which wouldn't be ideal to have that in most production environments.

Other than parity with v1, is this feature mostly desirable when initially prototyping a solution with InfluxDB? Or, is there a normal production situation where telegraf clients need to create buckets?

If this is mostly to facilitate easier prototyping with InfluxDB and we have this feature turned off by default, I don't think we should merge this change because both creating a bucket or turning on this feature require the user to figure out the cause of telegraf writes failing and take some action. It's safer for the user to create the bucket manually and then run telegraf with lower permissions.

Thoughts?

sjwang90:
Comments from phil: 

- we’re in the business of getting data into influxdb so we should make that as easy as possible. It’d kinda a weird use case to auto create buckets but it seems like a nice option versus rejecting data so I’m in favor of providing that option. I think it should be off by default (as in the pr) as free tier customers only get 2 buckets.
- I definitely think the default retention policy for these auto generated buckets should NOT be infinite. It should be 30 days, which would work with the free tier. Infinite is kinda like a foot gun for us and customers - if the customer doesn’t “really” know about which buckets are auto-created they won’t know they are getting charged for infinite data retention.
- Lastly, I worry that telegraf becomes a ddos source by creating thousands of buckets per second or something through a mis-configuration or something. I believe we are susceptible to this sort of thing now if i just wrote a bash script to do it, but we shouldn’t make our own ecosystem possible to spam us - I’d like to see a default rate limit for how often telegraf will auto create new buckets (probs 1/min?). Really if a customer is auto-creating buckets more frequently, they are “doing it wrong”. We on the cloud2 side will likely have an api limit here soon so telegraf should be prepared for that and process 429 error codes.

reimda:
@dgnorton Other than feature parity with the 1.x plugin, the only use case I'm aware of is the one mentioned in the issue this PR resolves, which is to simplify migration from 1.x to 2.x. If the 1.x instance has many databases it could become difficult to create buckets manually. In this case I would expect the user to use a token with elevated privileges for the migration, then switch to a token with lower privileges for ordinary data gathering. @RagHarnoor do you have other reasons for requesting this feature?

It sounds like we should hold off on merging this due to the concerns David and Phil brought up. If nothing else we need to default to a non-infinite retention policy and limit accidental creation of buckets.

reimda:
I'm going to close this due to the concerns raised in previous comments. If there is demand for this feature and someone would like to address those concerns, we can reopen this. Before reopening, we will need to clarify the use cases in an issue.

PR #9343, opened 2021-06-08
Use gosmi for SNMP traps

MyaLongmire:
### Required for all PRs:

resolves #9307 

config file now takes paths to where mib files are located
plugin now uses gosmi instead of snmptranslate
new unit test 
added fake mibs for the tests


MyaLongmire:
@loganmc10 Would mind reviewing this?

plugins/inputs/snmp_trap/snmp_trap.go
```diff
@@ -104,27 +102,46 @@ func init() {
...
-	cmd := exec.Command(arg0, args...)
-	var out bytes.Buffer
-	cmd.Stdout = &out
-	err := internal.RunTimeout(cmd, time.Duration(timeout))
-	if err != nil {
-		return nil, err
-	}
-	return out.Bytes(), nil
+func (s *SnmpTrap) Init() error {
+	// must init, append path for each directory, laod module for every file
```
loganmc10, line 114: > laod module for every file

typo here

plugins/inputs/snmp_trap/snmp_trap.go
```diff
@@ -366,48 +383,17 @@ func makeTrapHandler(s *SnmpTrap) gosnmp.TrapHandlerFunc {
...
 
-	scanner := bufio.NewScanner(bytes.NewBuffer(out))
-	ok := scanner.Scan()
-	if err = scanner.Err(); !ok && err != nil {
-		return e, err
-	}
-
-	e.oidText = scanner.Text()
+	e.oidText = node.RenderQualified()
+	fmt.Printf("node %v\n", node)
```
loganmc10, line 396: leftover debugging

plugins/inputs/snmp_trap/snmp_trap.go
```diff
@@ -104,27 +102,46 @@ func init() {
...
 
-func (s *SnmpTrap) Init() error {
-	s.cache = map[string]mibEntry{}
-	s.execCmd = realExecCmd
+func (s *SnmpTrap) getMibsPath() error {
+	var folders []string
+	for _, mibPath := range s.Path {
+		gosmi.AppendPath(mibPath)
+		folders = append(folders, mibPath)
+		filepath.Walk(mibPath, func(path string, info os.FileInfo, err error) error {
```
loganmc10, line 126: I believe this folder walking code could be improved. I don't think it's truly recursive. For instance, if path is "/usr/share/snmp/mibs" and the user has mibs in "/usr/share/snmp/mibs/1/2/file.mib" it won't get picked up

plugins/inputs/snmp_trap/snmp_trap_test.go
```diff
@@ -150,6 +127,15 @@ func sendTrap(t *testing.T, goSNMP gosnmp.GoSNMP, trap gosnmp.SnmpTrap) {
...
 
+// func sendTrapGoSmi(t *testing.T, goSMI gosmi.SmiNode, trap gosnmp.SnmpTrap) {
+// 	goSMI, err := gosmi.GetNodeByOID(types.OidMustFromString(goSMI.Oid.String()))
+// 	if err != nil {
+// 		t.Errorf("GetNodeByOid err: %v", err)
+// 	}
+// 	// entry.e.oidText = goSMI.RenderQualified()
+// 	goSMI.RenderQualified()
+// }
+
```
reimda, line 138: Looks like this is left over and should be deleted.
```suggestion
```

plugins/inputs/snmp_trap/snmp_trap_test.go
```diff
@@ -1293,19 +1288,13 @@ func TestReceiveTrap(t *testing.T) {
 				PrivProtocol: tt.privProto,
 				PrivPassword: tt.privPass,
 			}
-			require.Nil(t, s.Init())
-			// Don't look up oid with snmptranslate.
-			s.execCmd = fakeExecCmd
+
+			require.NoError(t, s.Init())
+
```
reimda, line 1293: Oops, it looks like s.Init() was there originally. We must have removed it before we found it was definitely necessary and put it back :)

loganmc10:
I think it looks good, I don't know about the unit tests, but I assume once they are passing that is good.

I know that gosmi has a ```gosmi.Exit()```, I haven't looked at what it does, but I assume it should be called somewhere? I don't know if Telegraf plugins have some kind of an exit function.

plugins/inputs/snmp_trap/snmp_trap_test.go
```diff
@@ -1331,4 +1320,103 @@ func TestReceiveTrap(t *testing.T) {
...
+		},
+		lookupFunc: lookup,
+		Log:        testutil.Logger{},
+		Version:    "2c",
+		Path:       []string{testDataPath},
+		// Path: []string{
+		// 	"/usr/share/snmp/mibs",
+		// 	// "/usr/share/snmp/mibs/ietf",
+		// 	// "/usr/share/snmp/mibs/iana",
+		// },
```
reimda, line 1398: Remove this leftover debugging code
```suggestion
```

reimda (commented):
Just a couple things I noticed in a very quick review

Hipska (commented):
Looks good in general, but I cannot test since I'm not using this plugin. I'm using the `snmp` plugin.

Anyway, I think it would be better to add the MIB path to the general snmp config (internal/snmp/config.go) and add the generic translation/lookup functions in that section, so both plugins can benefit from it.

Hipska:
@MyaLongmire this PR will not solve #5720 in this current state. As it is about `snmp` plug-in and the PR only covers `snmp_trap`.

Hipska:
The readme still says it uses `snmptranslate`, that's not correct I guess?

MyaLongmire:
It is not correct and I will fix this in a new pr, good catch!

PR #9345, opened 2021-06-09
Fix Mac script to install go

sspaink:
This pull request caused a bug in the shell script to install Go for mac: https://github.com/influxdata/telegraf/pull/9335 causing the Mac build to fail currently: https://app.circleci.com/pipelines/github/influxdata/telegraf/5226/workflows/6e336a4d-749b-42fb-83d0-3cf0ba309879/jobs/95226. This wasn't caught in the pull request due to the script `check-file-changes.sh` which didn't run the CI if a shell script changed. Also updated the `check-file-changes.sh` script to take into consideration for release branches which was another issue with the script.

@pierwill can you help verify that this change looks good?

### DIFF
--- a/scripts/check-file-changes.sh
+++ b/scripts/check-file-changes.sh
@@ -5,7 +5,7 @@ BRANCH="$(git rev-parse --abbrev-ref HEAD)"
 echo $BRANCH
 if [[ "$BRANCH" != "master" ]] && [[ "$BRANCH" != release* ]]; then # This should never skip for master and release branches
     # Ask git for all the differences between this branch and master
-    # Then use grep to look for changes in the .circleci/ directory, anything named *.go* or *.mod* or *.sum* or *.sh* or *Makefile
+    # Then use grep to look for changes in the .circleci/ directory, anything named *.go or *.mod or *.sum or *.sh or Makefile
     # If no match is found, then circleci step halt will stop the CI job but mark it successful
     git diff master --name-only --no-color | egrep -e "^(\.circleci\/.*)$|^(.*\.(go|mod|sum|sh))$|^Makefile$" || circleci step halt;
 fi
--- a/scripts/check-file-changes.sh
+++ b/scripts/check-file-changes.sh
@@ -1,7 +1,11 @@
 #!/bin/bash
+# To prevent the tests/builds to run for only a doc change, this script checks what files have changed in a pull request.
 
 BRANCH="$(git rev-parse --abbrev-ref HEAD)"
 echo $BRANCH
-if [[ "$BRANCH" != "master" ]] && [[ "$BRANCH" != release* ]]; then
-    git diff master --name-only --no-color | egrep -e "^(\.circleci\/.*)|(.*\.(go|mod|sum|sh))|Makefile$" || circleci step halt;
+if [[ "$BRANCH" != "master" ]] && [[ "$BRANCH" != release* ]]; then # This should never skip for master and release branches
+    # Ask git for all the differences between this branch and master
+    # Then use grep to look for changes in the .circleci/ directory, anything named *.go* or *.mod* or *.sum* or *.sh* or *Makefile
+    # If no match is found, then circleci step halt will stop the CI job but mark it successful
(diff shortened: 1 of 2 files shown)


pierwill:
Looks good to me! 👍 👍

scripts/mac_installgo.sh
```diff
@@ -12,9 +12,12 @@ path="/usr/local/Cellar"
 # Download Go and verify Go tarball. (Note: we aren't using brew because
 # it is slow to update and we can't pull specific minor versions.)
 setup_go () {
+
+    brew install coreutils # Install sha256sum util
```
reimda, line 16: Macos doesn't come with sha256sum, but it has a similar utility `shasum`. I think `shasum -a 256 --check` does the same thing as `sha256sum --check`

https://ss64.com/osx/shasum.html

Using brew to install sha256sum is an option, but using a native utility might be worth it to avoid having circleci persist the filesystem changes. Looking at the CI output for this PR, in this case it doesn't look like layer persistence is taking much time.

scripts/check-file-changes.sh
```diff
@@ -2,6 +2,6 @@
 
 BRANCH="$(git rev-parse --abbrev-ref HEAD)"
 echo $BRANCH
-if [[ "$BRANCH" != "master" ]]; then
-    git diff master --name-only --no-color | egrep -e "^(\.circleci\/.*)|(.*\.(go|mod|sum))|Makefile$" || circleci step halt;
+if [[ "$BRANCH" != "master" ]] && [[ "$BRANCH" != release* ]]; then
+    git diff master --name-only --no-color | egrep -e "^(\.circleci\/.*)|(.*\.(go|mod|sum|sh))|Makefile$" || circleci step halt;
```
reimda, line 6: Thanks for adding the check for release branches!

We ask git for all the differences between this branch and master, then use grep to look for changes in the .circleci/ directory, anything named `*.go*` or `*.mod*` or `*.sum*` or `*.sh*` or `*Makefile`. I think there's an operator precedence error here with how ^$ and | work. You probably need ^$ in each alternate section. The first ^ only applies to .cirleci/.* and the last $ only applies to Makefile.

Why do we filter the differences with grep? Is it to avoid builds for doc-only changes? Could you add a comment so it will be more obvious?

scripts/mac_installgo.sh
```diff
@@ -12,9 +12,12 @@ path="/usr/local/Cellar"
...
+
     echo "installing go"
     curl -OL https://golang.org/dl/go${GO_VERSION}.${GO_ARCH}.tar.gz --output go${GO_VERSION}.${GO_ARCH}.tar.gz
```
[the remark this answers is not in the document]
sspaink: Good catch! I removed the unnecessary -O.

scripts/mac_installgo.sh
sspaink: I think using a native tool is a good idea, I've updated the script to use it instead. Evidently you need two spaces between the SHA and the filename and it just throws a cryptic message otherwise, but its working now! Also for some reason it is still using 1.16.2, went ahead and updated that.

scripts/check-file-changes.sh
```diff
@@ -1,7 +1,11 @@
 #!/bin/bash
+# To prevent the tests/builds to run for only a doc change, this script checks what files have changed in a pull request.
 
 BRANCH="$(git rev-parse --abbrev-ref HEAD)"
 echo $BRANCH
-if [[ "$BRANCH" != "master" ]]; then
-    git diff master --name-only --no-color | egrep -e "^(\.circleci\/.*)|(.*\.(go|mod|sum))|Makefile$" || circleci step halt;
+if [[ "$BRANCH" != "master" ]] && [[ "$BRANCH" != release* ]]; then # This should never skip for master and release branches
+    # Ask git for all the differences between this branch and master
+    # Then use grep to look for changes in the .circleci/ directory, anything named *.go* or *.mod* or *.sum* or *.sh* or *Makefile
```
reimda, line 8: Nitpick to change the comments to match the regexp now that it has ^$ in each alternate section:
```suggestion
    # Then use grep to look for changes in the .circleci/ directory, anything named *.go or *.mod or *.sum or *.sh or Makefile
```

reimda (changes requested):

PR #9350, opened 2021-06-10
Fix wireguard in mod file

sspaink:
The wireguard depedency in the mod file is causing an issue with dependabot, updating should resolve it.
The error from dependabot:
```
Dependabot failed to update your dependencies because there was an error resolving your Go dependency files.

Dependabot encountered the following error:

go: golang.zx2c4.com/wireguard/wgctrl@v0.0.0-20200205215550-e35592f146e4 requires
	golang.zx2c4.com/wireguard@v0.0.20200121: reading golang.zx2c4.com/wireguard/go.mod at revisio
```


srebhan (commented):
Hey @sspaink, the Windows build fails with `FAIL	github.com/influxdata/telegraf/plugins/inputs/wireguard [build failed]` which looks not totally unrelated. :-) Can you please take a look there?

PR #9355, opened 2021-06-11
Add queries for collecting Query Store data in Azure SQL MI

andrewzayats:
### Required for all PRs:

Summary of PR:

Add queries for collecting Query Store runtime and wait statistics in Azure SQL MI.
This PR is based on the following PRs: [9150](https://github.com/influxdata/telegraf/pull/9150) and [8465](https://github.com/influxdata/telegraf/pull/8465).


andrewzayats:
@denzilribeiro @dimitri-furman could you please review?

plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -1217,3 +1217,275 @@ SELECT
...
+
+const sqlAzureMIPartQueryPeriod = `
+DECLARE @QueryData NVARCHAR(MAX) = ?1; --input parameter
+
+IF SERVERPROPERTY('EngineEdition') <> 8 BEGIN /*not Azure Managed Instance*/
+	DECLARE @ErrorMessage AS nvarchar(500) = 'Telegraf - Connection string Server:'+ @@SERVERNAME + ',Database:' + DB_NAME() +' is not an Azure Managed Instance. Check the database_type parameter in the telegraf configuration.';
+	RAISERROR (@ErrorMessage,11,1)
+	RETURN
+END;
+
```
dimitri-furman, line 1229: One general comment: do we have support for only collecting QDS data from some databases on an MI instance? For instances with many (dozens) of databases, it is likely that customers may want to collect only from a few. We may also find that collecting from all is too heavy. Can we add configurations along the lines of IncludeQueryStoreForDatabases and/or ExcludeQueryStoreForDatabases? Default would be collect from all.

plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -1217,3 +1217,275 @@ SELECT
...
+	s.value('./collectionTime[1]', 'datetimeoffset') AS [collectionTime]
+FROM @XmlQueryData.nodes('/collectionTimes/ts') t(s);
+
+/* Delete records older than <Max INTERVAL_LENGTH_MINUTES> * 2 */
+DELETE FROM #CachedDbCollectionTimes WHERE collectionTime <= DATEADD(dd, -2, SYSDATETIMEOFFSET());
+
+DECLARE @SqlText NVARCHAR(MAX) = N'
+
+DECLARE @lastIntervalEndTime DATETIMEOFFSET;
+SELECT TOP 1 @lastIntervalEndTime = [collectionTime] FROM #CachedDbCollectionTimes WHERE [id] = DB_ID();
```
dimitri-furman, line 1287: Don't you need an ORDER BY here, considering you use TOP 1? There is no PK/unique constraint on the `id` column (should it be added?), so duplicates are possible. If you do add a constraint, can remove TOP 1.

dimitri-furman (changes requested):

plugins/inputs/sqlserver/azuresqlqueries.go
andrewzayats: @dimitri-furman, currently telegraf doesn't support any DB filtering. If we add simple list IncludeQueryStoreForDatabases to a sqlserver plugin config section it will affect all servers in this section and in many cases this behavior may be undesirable. In order to allow per-server DB filtering the sqlserver plugin configuration schema and the plugin's config processing may require significant changes. Also DB filteringng may be also applied to other queries. That is why I think that DB filtering should be implemented as a separate PR.

plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -1248,7 +1248,8 @@ BEGIN
 	FETCH NEXT FROM dbCursor INTO @dbName;
 
 	WHILE @@Fetch_Status=0 BEGIN
-		SET @CurrSqlText = N''USE ''+ QUOTENAME(@dbName ,''"'') + @queryText
+		SET @CurrSqlText = N''USE ''+ QUOTENAME(@dbName ,''['') 
+				+ @queryText
```
dimitri-furman, line 1252: You don't need the second argument for QUOTENAME at all, default is []. 

I meant `QUOTENAME(@dbName) + CONCAT(CHAR(13), CHAR(10)) + @queryText`, sorry if I wasn't clear.

plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -1217,3 +1217,275 @@ SELECT
...
+	RETURN
+END;
+
+DROP PROCEDURE IF EXISTS #ExecForeachDb
+/* Create a temporary procedure that iterates through all DBs that the user has access to in this managed instance, except system DBs, and execute the query specified with @queryText */
+EXEC('CREATE PROCEDURE #ExecForeachDb @queryText NVARCHAR(MAX)
+AS
+BEGIN
+	SET NOCOUNT ON;
+
```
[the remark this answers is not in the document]
andrewzayats: @dimitri-furman, as I understand we need SP-level TRY/CATCH for cursor deallocation only. But why should we do this if the cursor is deallocated implicitly after leaving the scope of the stored proc? Could you please clarify this?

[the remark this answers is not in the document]
dimitri-furman: Yes, you are correct. I overlooked that the cursor is declared as LOCAL.

plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -1217,3 +1217,275 @@ SELECT
...
+	DECLARE dbCursor CURSOR FAST_FORWARD LOCAL FOR 
+	SELECT 
+		[name] 
+	FROM sys.databases
+	WHERE HAS_DBACCESS([name]) = 1 
+		AND [name] NOT IN (''master'', ''tempdb'', ''model'', ''msdb'');
```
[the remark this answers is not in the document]
andrewzayats: @dimitri-furman, it looks like HAS_DBACESS covers the cases you provided. But DB state may change after the check with DBACCESS. For such cases I added more error codes to ignore in the catch block.

plugins/inputs/sqlserver/azuresqlqueries.go
dimitri-furman: But that would disable all collection, right? The intent is to disable voluminous and heavy QDS collection for databases where it's not needed.

andrewzayats: @dimitri-furman, why would it disable all collection? A custemer can simply give access to DBs he wants to monitor. Other sql queries (not QDS) don't require per-DB access, they use instance-level views and won't be affected. Also a customer can simply create a separate account for QDS monitoring.
In any case limiting access now is the only way to select DBs for QDS monitoring. In one of my previous comments I wrote why adding additional configuration parameter in this PR is a bad idea.

plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -1276,24 +1276,31 @@ END');
...
+	s.value('./ct[1]', 'datetimeoffset') AS [collectionTime]
 FROM @XmlQueryData.nodes('/collectionTimes/ts') t(s);
 
 /* Delete records older than <Max INTERVAL_LENGTH_MINUTES> * 2 */
 DELETE FROM #TelegrafCachedDbCollectionTimes WHERE collectionTime <= DATEADD(dd, -2, SYSDATETIMEOFFSET());
 
 DECLARE @SqlText NVARCHAR(MAX) = N'
 
+DECLARE @currDbLogicalGuid UNIQUEIDENTIFIER;
+SELECT @currDbLogicalGuid = logical_database_guid FROM sys.dm_user_db_resource_governance WHERE [database_name] = DB_NAME()
```
dimitri-furman, line 1296: Any reason to use `database_name = DB_NAME()` vs. `database_id = DB_ID()`? It is a remote chance, but a database could be renamed between the time this query runs and the time we collect.

Please terminate query with a semicolon for consistency.

andrewzayats: @dimitri-furman, the main reason of doing so is that for some DBs DB_ID() returns  values that differ from the values in the DMV. You can try to execute the following code:

`use master`
`select database_id FROM sys.dm_user_db_resource_governance WHERE [database_name] = 'master'`
`select DB_ID()`

In my environment I received the following values: 32763, 1.
I faced with this issue on a system DB, but I am not sure that such case is impossible with a user DB.
Could you please clarify why the ids are different? Is this a Sql Server issue?

dimitri-furman: I see. This will only happen for system databases, and only on MI. There are local and replicated versions of system databases on MI with different IDs, and unfortunately metadata is not always consistent. It does not impact us because we are not collecting from system databases, so you can match on `database_id`. For user databases, `database_id` in this view will always match DB_ID() for the current database.

plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -1217,3 +1217,275 @@ SELECT
...
+
+IF SERVERPROPERTY('EngineEdition') <> 8 BEGIN /*not Azure Managed Instance*/
+	DECLARE @ErrorMessage AS nvarchar(500) = 'Telegraf - Connection string Server:'+ @@SERVERNAME + ',Database:' + DB_NAME() +' is not an Azure Managed Instance. Check the database_type parameter in the telegraf configuration.';
+	RAISERROR (@ErrorMessage,11,1)
+	RETURN
+END;
+
+DROP PROCEDURE IF EXISTS #ExecForeachDb
+/* Create a temporary procedure that iterates through all DBs that the user has access to in this managed instance, except system DBs, and execute the query specified with @queryText */
+EXEC('CREATE PROCEDURE #ExecForeachDb @queryText NVARCHAR(MAX)
```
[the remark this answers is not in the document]
andrewzayats: Done.

plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -1217,3 +1217,275 @@ SELECT
...
+		AND [name] NOT IN (''master'', ''tempdb'', ''model'', ''msdb'');
+
+	DECLARE @CurrSqlText nvarchar(max);
+	DECLARE @dbName sysname;
+
+	OPEN dbCursor;
+	FETCH NEXT FROM dbCursor INTO @dbName;
+
+	WHILE @@Fetch_Status=0 BEGIN
+		SET @CurrSqlText = N''USE ''+ QUOTENAME(@dbName ,''"'') + @queryText
```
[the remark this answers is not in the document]
andrewzayats: Done.

plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -1217,3 +1217,275 @@ SELECT
...
+		END CATCH;
+		FETCH NEXT FROM dbCursor INTO @dbName;
+	END
+
+	CLOSE dbCursor;
+	DEALLOCATE dbCursor;
+END');
+
+/* Obtain the previous times the query was called -- they are an array because we are working with a list of DBs in the Managed Instance -- and store them in a temporary table */
+IF OBJECT_ID ('tempdb.dbo.#CachedDbCollectionTimes') IS NOT NULL DROP TABLE #CachedDbCollectionTimes;
```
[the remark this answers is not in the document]
andrewzayats: Done.

plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -1217,3 +1217,275 @@ SELECT
...
+CREATE TABLE #CachedDbCollectionTimes (
+	[id] INT,
```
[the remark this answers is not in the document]
andrewzayats: Done.

plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -1217,3 +1217,275 @@ SELECT
...
+	[collectionTime] DATETIMEOFFSET
+);
+
```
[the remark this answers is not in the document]
andrewzayats: Done.

plugins/inputs/sqlserver/azuresqlqueries.go
```diff
@@ -1217,3 +1217,275 @@ SELECT
...
+	avg_duration float,
+	max_num_physical_io_reads bigint,
+	avg_num_physical_io_reads float,
+	max_page_server_io_reads bigint,
+	avg_page_server_io_reads float,
+	max_log_bytes_used bigint,
+	avg_log_bytes_used float,
+	max_tempdb_space_used bigint,
+	avg_tempdb_space_used float
+);
```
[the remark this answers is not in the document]
andrewzayats: Done.

plugins/inputs/sqlserver/azuresqlqueries.go
andrewzayats: PK was added.

plugins/inputs/sqlserver/azuresqlqueries.go
andrewzayats: Done.

dimitri-furman (approved):

sjwang90:
@andrewzayats Can you please resolve the conflicts on this PR?

dimitri-furman:
@sjwang90 my last comment on #9150 applies to this PR as well, so this can be closed, at least until there are any takers interested in completing this.

sjwang90:
Thanks @dimitri-furman! I close it but have added the `help wanted` label if anyone chooses to pick this up in the future.

PR #9359, opened 2021-06-13
Fix import of sqlite and ignore it on all platforms that require CGO.

srebhan:
### Required for all PRs:

Fix unintended pulling in of CGO by importing the sqlite SQL driver. CGO is required for some platforms when using sqlite. We simply do not support sqlite for those platforms and get rid of the CGO dependency. Fixes CI errors like
```
[...]
package github.com/influxdata/telegraf/cmd/telegraf
	imports github.com/influxdata/telegraf/plugins/outputs/all
	imports github.com/influxdata/telegraf/plugins/outputs/sql
	imports modernc.org/sqlite
	imports modernc.org/libc
	imports modernc.org/libc/unistd: build constraints exclude all Go files in /go/pkg/mod/modernc.org/libc@v1.9.5/unistd
[...]
```

### DIFF
--- a/plugins/inputs/sql/drivers_sqlite.go
+++ b/plugins/inputs/sql/drivers_sqlite.go
@@ -1,4 +1,4 @@
-// +build linux,freebsd
+// +build linux,freebsd,darwin
 // +build !mips !mips64
 
 package sql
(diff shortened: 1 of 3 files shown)


reimda (approved):
looks good to me

reimda:
Thanks @srebhan!

PR #9366, opened 2021-06-14
added a check for oid and name to prevent empty metrics

### LINKED ISSUE #8840: SNMP input can produce metrics with empty metric name
### Steps to reproduce:
The snmp input can have a table with no oid or name but with fields.  (see the plugin readme https://github.com/influxdata/telegraf/blob/ea4feb1a07edfea3243cc328f1994f63852e9f96/plugins/inputs/snmp/README.md#table)  If the table name and the table oid are not set, the plugin produces metrics with no name.

### Relevant telegraf.conf:
Here's an example snmp input that exhibits the problem.
```toml
[[inputs.snmp]]
  agents = ["udp://127.0.0.1:161"]
  community = "public"
  version = 2
  [[inputs.snmp.table]]
    #oid = "IF-MIB::ifXTable"
    #name = "ifXTable"
    index_as_tag = true
    [[inputs.snmp.table.field]]
      oid = "IF-MIB::ifHCInOctets"
      is_tag = false
    [[inputs.snmp.table.field]]
      oid = "IF-MIB::ifHCOutOctets"
      is_tag = false
[[outputs.file]]
  files = ["stdout"]
  data_format = "influx"

```
For this example you need a snmp agent running on localhost that allows access to IF-MIB::ifXTable with the public community string.  I used net-snmp.

If you uncomment either the name or the oid settings, the metric is named so metrics are output.

### System info:
I tested with telegraf master on linux but this is not platform specific.

### Expected behavior:
I didn't expect telegraf to try to produce a metric with no name.  Once I understood this was happening I would have expected an error from the snmp input or the agent itself or the file output.

The snmp input should provide a specific error, like "error: SNMP table at telegraf.conf line 234 is not named.  One or both of the oid and name settings must be set".

### Actual behavior:

Telegraf doesn't produce any metrics

MyaLongmire:
### Required for all PRs:

resolves #8840 SNMP input can produce metrics with empty metric name

Added a check for empty oid and empty name
fixed an existing unit test to give the table a name
and added a test for a table with empty oid and empty name

### DIFF
--- a/plugins/inputs/snmp/snmp.go
+++ b/plugins/inputs/snmp/snmp.go
@@ -173,6 +173,12 @@ type Table struct {
 
 // Init() builds & initializes the nested fields.
 func (t *Table) Init() error {
+	//makes sure oid or name is set in config file
+	//otherwise snmp will produce metrics with an empty name
+	if t.Oid == "" && t.Name == "" {
+		return fmt.Errorf("SNMP table in config file is not named. One or both of the oid and name settings must be set")
+	}
+
 	if t.initialized {
 		return nil
 	}
--- a/plugins/inputs/snmp/snmp_test.go
+++ b/plugins/inputs/snmp/snmp_test.go
@@ -199,11 +199,12 @@ func TestSnmpInit_noTranslate(t *testing.T) {
 			{Oid: ".1.1.1.3"},
 		},
 		Tables: []Table{
-			{Fields: []Field{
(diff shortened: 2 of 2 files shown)


sspaink (approved):

Hipska:
Did this pass all tests? Cause it makes telegraf to crash when you use ifName processor!!

Hipska:
It did pass the tests, more disturbing is that the specific ifname test has been disabled in #8725

Issue #9352, opened 2021-06-10
json_v2 parser dealing with null data


russorat:
### Relevant telegraf.conf:
```toml
[[inputs.http]]
  #interval = "1m"

  urls = [
    "https://earthquake.usgs.gov/earthquakes/feed/v1.0/summary/all_hour.geojson"
  ]

  name_override = "usgs"

  tagexclude = ["url"]

  method = "GET"

  timeout = "10s"

  data_format = "json_v2"
  [[inputs.http.json_v2]]
        [[inputs.http.json_v2.object]]
            path = "features"
            tags = ["type", "id"]
            disable_prepend_keys = true

```

### System info:

➜  telegraf git:(master) ✗ ./telegraf --version
Telegraf unknown (git: master da7f2c7a)

### Steps to reproduce:

```sh
➜  telegraf git:(master) ✗ ./telegraf --config ~/telegraf.d/inputs.http.json_v2.conf --test
2021-06-10T22:33:58Z I! Starting Telegraf 
2021-06-10T22:33:59Z E! [inputs.http] Error in plugin: [url=https://earthquake.usgs.gov/earthquakes/feed/v1.0/summary/all_hour.geojson]: unknown format '<nil>' for field  'properties_tz'
2021-06-10T22:33:59Z E! [telegraf] Error running agent: input plugins recorded 1 errors
```

### Expected behavior:

Many JSON files have null values. I expected to be able to configure the json_v2 parser to ignore null values (or just automatically ignore them). 

### Actual behavior:

Telegraf complains about an unknown format for the field.

PR #9368, opened 2021-06-14
Update json_v2 parser to handle null types

sspaink:
resolves: #9352

Instead of throwing an error when encountering a null type in the input JSON, I've updated the parser to ignore it because they are valid types but there isn't anything we can do with it. This follows the same behavior as the original json parser.


reimda (approved):
Looks good, thanks for adding so many good tests!

PR #9388, opened 2021-06-16
Sqlserver input: require authentication method to be specified

### LINKED ISSUE #9362: SQL Server input no longer functions correctly with Windows Integrated authentication
### Relevant telegraf.conf:

```toml
# Read metrics from Microsoft SQL Server
[[inputs.sqlserver]]
  ## Specify instances to monitor with a list of connection strings.
  ## All connection parameters are optional.
  ## By default, the host is localhost, listening on default port, TCP 1433.
  ##   for Windows, the user is the currently running AD user (SSO).
  ##   See https://github.com/denisenkom/go-mssqldb for detailed connection
  ##   parameters.

  servers = [
    "Server=localhost;app name=telegraf;log=1;"
  ]

  database_type = "SQLServer"
```

### System info:
telegraf 1.19.0-rc0 running on Windows OS

### Steps to reproduce:

1. Add SQL Server input configuration intended to use Windows Integrated authentication (i.e. no user id/password fields) and database_type of `SQLServer`
2. Run telegraf binary with above configuration

### Expected behavior:
Telegraf agent should connect to SQL Server instance using Windows Integrated authentication using the user identity of the running process.

### Actual behavior:
Telegraf incorrectly determines that it should connection using Azure Active Directory (AAD) authentication but `SQLServer` database type was specified and therefore throws error
`[inputs.sqlserver] Error in plugin: database connection failed : AAD auth is not supported for SQL VM i.e. DatabaseType=SQLServer`

reimda:
The sqlserver input plugin uses windows authentication (aka integrated security) when there is no username and password in the connection string.

#8822 added support for AAD authentication but only when there is no password in the connection string. Since the new check for no password happens before the previous connection logic, this causes windows authentication to be unavailable, which is a regression.

This PR adds a setting to enable AAD authentication. This allows use of windows authentication as before. It's backward compatible because the default of the new setting is the old behavior. New users who want to use AAD will to opt in to the new auth behavior by changing the setting to "AAD".

resolves #9362

### DIFF
--- a/plugins/inputs/sqlserver/README.md
+++ b/plugins/inputs/sqlserver/README.md
@@ -52,6 +52,10 @@ GO
     "Server=192.168.1.10;Port=1433;User Id=<user>;Password=<pw>;app name=telegraf;log=1;",
   ]
 
+  ## Authentication method
+  ## valid methods: "default", "AAD"
+  # auth_method = "default"
+
   ## "database_type" enables a specific set of queries depending on the database type. If specified, it replaces azuredb = true/false and query_version = 2
   ## In the config file, the sql server plugin section should be repeated each with a set of servers for a specific database_type.
   ## Possible values for database_type are - "AzureSQLDB" or "AzureSQLManagedInstance" or "SQLServer"
@@ -197,11 +201,12 @@ EXECUTE ('GRANT VIEW DATABASE STATE TO [<Monitoring_VM_Name>]')
 - On the SQL Server resource of the database(s) being monitored, go to "Firewalls and Virtual Networks" tab and allowlist the monitoring VM IP address.
 - On the Monitoring VM, update the telegraf config file with the database connection string in the following format. Please note AAD based auth is currently only supported for Azure SQL Database and Azure SQL Managed Instance (but not for SQL Server), as described [here](https://docs.microsoft.com/en-us/azure/azure-sql/database/security-overview#authentication).
 - On the Monitoring VM, update the telegraf config file with the database connection string in the following format.
-- On the Monitoring VM, update the telegraf config file with the database connection string in the following format. The connection string only provides the server and database name, but no password (since the VM's system-assigned managed identity would be used for authentication).
+- On the Monitoring VM, update the telegraf config file with the database connection string in the following format. The connection string only provides the server and database name, but no password (since the VM's system-assigned managed identity would be used for authentication). The auth method must be set to "AAD"
 ```toml
   servers = [
     "Server=<Azure_SQL_Server_Name>.database.windows.net;Port=1433;Database=<Azure_SQL_Database_Name>;app name=telegraf;log=1;",
   ]
+  auth_method = "AAD"
 ```
 - Please note AAD based auth is currently only supported for Azure SQL Database and Azure SQL Managed Instance (but not for SQL Server), as described [here](https://docs.microsoft.com/en-us/azure/azure-sql/database/security-overview#authentication).
 
--- a/plugins/inputs/sqlserver/README.md
+++ b/plugins/inputs/sqlserver/README.md
@@ -53,8 +53,8 @@ GO
   ]
 
   ## Authentication method
-  ## valid methods: "default", "AAD"
-  # auth_method = "default"
+  ## valid methods: "connection_string", "AAD"
+  # auth_method = "connection_string"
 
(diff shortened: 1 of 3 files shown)


reimda:
@cerilewis could you test the PR build once CI passes?

@avinash-nigam what do you think of this solution?

aslgithub:
@reimda I've found this via https://github.com/influxdata/telegraf/issues/9362
If its any help I've tested the telgraf-tiger artifact build and can confirm this build fixes the issue.

I've a thought if it helps in any way, normally in my experience whilst no user or password is provided in the connection string for Windows Integrated Authentication, normally the setting Integrated Security=true would be set to indicate the intention e.g.
"Server=localhost;Port=1433;Integrated Security=true;log=1;",

cerilewis:
@reimda I can confirm that using the config that failed previously with 1.19.0rc0 now works with the tiger build above.
Thank you for addressing this.

aslgithub:
@srebhan Tagging you in on this to see if you can help merge this please?

plugins/inputs/sqlserver/README.md
```diff
@@ -52,6 +52,10 @@ GO
     "Server=192.168.1.10;Port=1433;User Id=<user>;Password=<pw>;app name=telegraf;log=1;",
   ]
 
+  ## Authentication method
+  ## valid methods: "default", "AAD"
+  # auth_method = "default"
```
srebhan, line 57: Should we name it `"credentials"` or `"connection param"`? `"default"` is not very descriptive. :-)

srebhan (commented):
The code looks good to me, however, I think we should give the "default" auth method a more descriptive name e.g. "credentials" or "connection params".

plugins/inputs/sqlserver/README.md
sspaink: @akrantz01  and I agree, we think "credentials" would be more descriptive.

reimda: I switched to "connection_string". I considered using that initially but I thought it was too long. Also I thought it would be confusing for #9362 because in that case, the username/password is intentionally left out of the connection string to tell the client to use Windows Integrated auth. So to use WIA you would say auth_method = "connection_string", then not put auth in the connection string.

I don't know what's best here. "default" is not descriptive but not incorrect. "connection_string" may be the next best option.

aslgithub: Sorry, but I don't like this; missing credentials should not be assumed...
The convention as far as I understand it for Windows Integrated Auth is to specify 'Integrated Security=true' in the connection string e.g. 
"Server=localhost;Port=1433;Integrated Security=true;log=1;"

reimda: Hi @aslgithub, the issue here isn't personal preference or convention. If I could have my preference I would require an explicit setting to enable windows integrated security like you suggest, either in the connection string or as an option in a telegraf setting like "auth_method".

The issue is backward compatibility. Some existing users of this plugin (pre 1.19.0) use windows integrated auth without the "Integrated Security=true" connection parameter and without credentials. If sql server and the golang driver allow you to use windows integrated auth by omitting credentials, and telegraf allowed you to do it that way (and there are telegraf users that did it that way), we need to support it. We can't break the config for these users and force them to debug why their connection no longer works after they upgrade.

The design decision to try integrated security if there are no credentials in the connection string was made in the past, I assume by the SQL Server team, possibly many years ago. It never was our choice to do it this way.

aslgithub: @reimda Aren't you now going to break this anyway by forcing all users to set an auth method or have I misunderstood the intention ?

sspaink (approved):
LGTM!

plugins/inputs/sqlserver/README.md
reimda: The default setting for auth_method will be to act like pre-1.19.0 telegraf ("connection_string"). Users who want to use Azure Active Directory will need to opt in by setting auth_method to "AAD". This way people upgrading to 1.19.1 from any version before 1.19.0 will not be affected at all.

srebhan (approved):
Looks good to me.

ssoroka (approved):
This looks good, but you may want to consider adding a test to ensure future changes don't break support for this, but that's optional here and not required to merge. just let us know if you want to wait on merging.

PR #9395, opened 2021-06-18
Http plugin add cookie auth

### LINKED ISSUE #5932: input http_response cookies and sessions
## Feature Request

In http_response input, can you add the ability to retrieve cookie headers and reuse them ?
1) option to auth on a specific url (different than url to query for the test)
2) option to choose cookies mandatory variables to validate the auth state

idea found on https://community.influxdata.com/t/can-you-use-cookies-with-the-http-response-plugin-in-telegraf/3825 

In my case, there are php sessions cookies and others cookies crafted with javascript so I will have to write my own curl script or ask my developpers to give me a real URL for a checktest using telegraf http_response input.

jh125486:
### Required for all PRs:

resolves #5932

1. Added a few configuration options for `inputs.http` cookie auth: 
   auth url 
   body 
   method
   refresh interval
2. HTTP Client is updated to use a CookieJar
3. During plugin `Init()`, a initial auth is attempted and is successful, a Go routine is fired off for cookie auth renewal

### DIFF
--- a/plugins/common/cookie/cookie.go
+++ b/plugins/common/cookie/cookie.go
@@ -1,0 +1,91 @@
+package cookie
+
+import (
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"net/http/cookiejar"
+	"strings"
+	"time"
+
+	"github.com/influxdata/telegraf/config"
+)
+
+type CookieAuthConfig struct {
+	URL    string `toml:"cookie_auth_url"`
+	Method string `toml:"cookie_auth_method"`
+
+	// HTTP Basic Auth Credentials
+	Username string `toml:"cookie_auth_username"`
+	Password string `toml:"cookie_auth_password"`
+
+	Body    string          `toml:"cookie_auth_body"`
+	Renewal config.Duration `toml:"cookie_auth_renewal"`
+
+	client *http.Client
+}
+
+func (c *CookieAuthConfig) Start(client *http.Client) (err error) {
+	c.client = client
+
+	c.setDefaults()
+
+	// add cookie jar to HTTP client
+	if c.client.Jar, err = cookiejar.New(nil); err != nil {
+		return err
+	}
+
+	// continual auth renewal
+	ticker := time.NewTicker(time.Duration(c.Renewal))
+	go func() {
+		for range ticker.C {
+			_ = c.auth()
+		}
+	}()
+
+	// initial auth will immediately error out the Init() if auth fails
+	return c.auth()
+}
+
+func (c *CookieAuthConfig) setDefaults() {
+	if c.Renewal == 0 {
+		c.Renewal = config.Duration(5 * time.Minute)
+	}
+	if c.Method == "" {
+		c.Method = http.MethodPost
+	}
+}
+
+func (c *CookieAuthConfig) auth() error {
+	var body io.ReadCloser
+	if c.Body != "" {
+		body = ioutil.NopCloser(strings.NewReader(c.Body))
+		defer body.Close()
+	}
+
+	req, err := http.NewRequest(c.Method, c.URL, body)
+	if err != nil {
+		return err
+	}
+
+	if c.Username != "" || c.Password != "" {
+		req.SetBasicAuth(c.Username, c.Password)
+	}
+
+	resp, err := c.client.Do(req)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+
+	if _, err = io.Copy(ioutil.Discard, resp.Body); err != nil {
+		return err
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("bad response code: %v", resp.StatusCode)
+	}
+
+	return nil
+}
--- a/plugins/common/cookie/cookie.go
+++ b/plugins/common/cookie/cookie.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/influxdata/telegraf/config"
+	"github.com/sirupsen/logrus"
 )
 
 type CookieAuthConfig struct {
@@ -29,32 +30,29 @@ type CookieAuthConfig struct {
 func (c *CookieAuthConfig) Start(client *http.Client) (err error) {
 	c.client = client
 
-	c.setDefaults()
+	if c.Method == "" {
+		c.Method = http.MethodPost
+	}
 
 	// add cookie jar to HTTP client
 	if c.client.Jar, err = cookiejar.New(nil); err != nil {
 		return err
 	}
 
-	// continual auth renewal
-	ticker := time.NewTicker(time.Duration(c.Renewal))
-	go func() {
-		for range ticker.C {
-			_ = c.auth()
-		}
-	}()
+	// continual auth renewal if set
+	if c.Renewal > 0 {
+		ticker := time.NewTicker(time.Duration(c.Renewal))
+		go func() {
+			for range ticker.C {
+				if err := c.auth(); err != nil {
+					logrus.WithError(err).Error("cookie auth renewal failure")
+				}
+			}
+		}()
+	}
 
 	// initial auth will immediately error out the Init() if auth fails
 	return c.auth()
-}
-
-func (c *CookieAuthConfig) setDefaults() {
-	if c.Renewal == 0 {
-		c.Renewal = config.Duration(5 * time.Minute)
-	}
-	if c.Method == "" {
-		c.Method = http.MethodPost
-	}
 }
 
 func (c *CookieAuthConfig) auth() error {
@@ -69,7 +67,7 @@ func (c *CookieAuthConfig) auth() error {
 		return err
 	}
 
-	if c.Username != "" || c.Password != "" {
+	if c.Username != "" {
 		req.SetBasicAuth(c.Username, c.Password)
 	}
 
--- a/plugins/common/cookie/cookie.go
+++ b/plugins/common/cookie/cookie.go
@@ -9,8 +9,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/influxdata/telegraf"
 	"github.com/influxdata/telegraf/config"
-	"github.com/sirupsen/logrus"
 )
 
 type CookieAuthConfig struct {
@@ -27,7 +27,7 @@ type CookieAuthConfig struct {
 	client *http.Client
 }
 
-func (c *CookieAuthConfig) Start(client *http.Client) (err error) {
+func (c *CookieAuthConfig) Start(client *http.Client, log telegraf.Logger) (err error) {
 	c.client = client
 
 	if c.Method == "" {
@@ -39,20 +39,23 @@ func (c *CookieAuthConfig) Start(client *http.Client) (err error) {
 		return err
 	}
 
+	if err = c.auth(); err != nil {
+		return err
+	}
+
 	// continual auth renewal if set
 	if c.Renewal > 0 {
 		ticker := time.NewTicker(time.Duration(c.Renewal))
 		go func() {
 			for range ticker.C {
-				if err := c.auth(); err != nil {
-					logrus.WithError(err).Error("cookie auth renewal failure")
+				if err = c.auth(); err != nil {
+					log.Errorf("Error in plugin: [url=%v]: %v", c.URL, err)
 				}
 			}
 		}()
 	}
 
-	// initial auth will immediately error out the Init() if auth fails
-	return c.auth()
+	return nil
 }
 
 func (c *CookieAuthConfig) auth() error {
@@ -82,7 +85,10 @@ func (c *CookieAuthConfig) auth() error {
 	}
 
 	if resp.StatusCode != http.StatusOK {
-		return fmt.Errorf("bad response code: %v", resp.StatusCode)
+		return fmt.Errorf("cookie auth renewal received status code: %v (%v)",
+			resp.StatusCode,
+			http.StatusText(resp.StatusCode),
+		)
 	}
 
 	return nil
--- a/plugins/common/cookie/cookie.go
+++ b/plugins/common/cookie/cookie.go
@@ -48,8 +48,8 @@ func (c *CookieAuthConfig) Start(client *http.Client, log telegraf.Logger) (err 
 		ticker := time.NewTicker(time.Duration(c.Renewal))
 		go func() {
 			for range ticker.C {
-				if err = c.auth(); err != nil {
-					log.Errorf("Error in plugin: [url=%v]: %v", c.URL, err)
+				if err := c.auth(); err != nil && log != nil {
+					log.Errorf("renewal failed for %q: %v", c.URL, err)
 				}
 			}
 		}()
--- a/plugins/common/cookie/cookie_test.go
+++ b/plugins/common/cookie/cookie_test.go
@@ -85,15 +85,15 @@ func (s fakeServer) checkResp(t *testing.T, expCode int) {
 	}
 }
 
-func (s fakeServer) checkAuthCount(t *testing.T, expCount int) {
+func (s fakeServer) checkAuthCount(t *testing.T, atLeast int32) {
 	t.Helper()
-	require.EqualValues(t, expCount, atomic.LoadInt32(s.int32))
+	require.GreaterOrEqual(t, atomic.LoadInt32(s.int32), atLeast)
 }
 
 func TestAuthConfig_Start(t *testing.T) {
 	const (
-		renewal      = 20 * time.Millisecond
-		renewalCheck = 55 * time.Millisecond
+		renewal      = 10 * time.Millisecond
+		renewalCheck = 5 * renewal
 	)
 	type fields struct {
 		Method   string
--- a/plugins/common/cookie/cookie_test.go
+++ b/plugins/common/cookie/cookie_test.go
@@ -12,6 +12,7 @@ import (
 	"github.com/google/go-cmp/cmp"
 	"github.com/influxdata/telegraf/config"
 	"github.com/influxdata/telegraf/plugins/common/cookie"
+	"github.com/influxdata/telegraf/testutil"
 	"github.com/stretchr/testify/require"
 )
 
@@ -188,7 +189,7 @@ func TestAuthConfig_Start(t *testing.T) {
 				renewal:  renewal,
 				endpoint: authEndpointWithBasicAuth,
 			},
-			wantErr: fmt.Errorf("bad response code: 401"),
+			wantErr: fmt.Errorf("cookie auth renewal received status code: 401 (Unauthorized)"),
 			assert: func(t *testing.T, c *cookie.CookieAuthConfig, srv fakeServer) {
 				// should have never Cookie Authed
 				srv.checkAuthCount(t, 0)
@@ -229,7 +230,7 @@ func TestAuthConfig_Start(t *testing.T) {
 				renewal:  renewal,
 				endpoint: authEndpointWithBody,
 			},
-			wantErr: fmt.Errorf("bad response code: 401"),
+			wantErr: fmt.Errorf("cookie auth renewal received status code: 401 (Unauthorized)"),
 			assert: func(t *testing.T, c *cookie.CookieAuthConfig, srv fakeServer) {
 				// should have never Cookie Authed
 				srv.checkAuthCount(t, 0)
@@ -254,7 +255,7 @@ func TestAuthConfig_Start(t *testing.T) {
 				Renewal:  config.Duration(tt.args.renewal),
 			}
 
-			if err := c.Start(srv.Client()); tt.wantErr != nil {
+			if err := c.Start(srv.Client(), testutil.Logger{Name: "cookie_auth"}); tt.wantErr != nil {
 				require.EqualError(t, err, tt.wantErr.Error())
 			} else {
 				require.NoError(t, err)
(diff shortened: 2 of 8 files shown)


plugins/inputs/http/http.go
```diff
@@ -64,6 +78,12 @@ var sampleConfig = `
   ## HTTP entity-body to send with POST/PUT requests.
   # body = ""
 
+  ## Cookie-based authentication
+  # cookie_auth_url = "https://localhost/authMe"
+  # cookie_auth_method = "POST" # default: POST
+  # cookie_auth_body = '{"username": "user", "password": "pa$$word", "authenticate": "me"}'
+  # cookie_auth_renewal = "8h" # default: "5m"
```
srebhan, line 85: Please skip the extra `default` comment as it currently breaks toml IIRC. The standard is to specify the default in the commented option like
```suggestion
  # cookie_auth_method = "POST"
  # cookie_auth_body = '{"username": "user", "password": "pa$$word", "authenticate": "me"}'
  # cookie_auth_renewal = "5m"
```

plugins/inputs/http/http.go
```diff
@@ -6,16 +6,24 @@ import (
...
+	"github.com/influxdata/telegraf/config"
 	"github.com/influxdata/telegraf/internal"
 	httpconfig "github.com/influxdata/telegraf/plugins/common/http"
 	"github.com/influxdata/telegraf/plugins/inputs"
 	"github.com/influxdata/telegraf/plugins/parsers"
 )
 
+var (
+	defaultCookieAuthRenewal = 5 * time.Minute
+	defaultCookieAuthMethod  = http.MethodPost
```
srebhan, line 24: It's a matter of habits, but I would move this to the actual usage in `Init()`. Alternatively make this a constant.

plugins/inputs/http/http.go
```diff
@@ -116,13 +136,69 @@ func (h *HTTP) Init() error {
...
+			h.CookieAuthMethod = defaultCookieAuthMethod
+		}
+		// add cookie jar to HTTP client
+		if h.client.Jar, err = cookiejar.New(nil); err != nil {
+			return err
+		}
+		// start auth ticker
+		if authErr := h.startCookieAuth(renewalInterval); authErr != nil {
+			return authErr
+		}
```
srebhan, line 155: Please stick to `err` for errors wherever possible. This should also work
```suggestion
		if err := h.startCookieAuth(renewalInterval); err != nil {
			return err
		}
```

plugins/inputs/http/http.go
```diff
@@ -116,13 +136,69 @@ func (h *HTTP) Init() error {
...
 }
 
+func (h *HTTP) startCookieAuth(interval time.Duration) error {
+	// continual auth ticker
+	ticker := time.NewTicker(interval)
+	go func() {
+		for range ticker.C {
+			_ = h.doCookieAuth()
+		}
+	}()
```
srebhan, line 172: How about
```suggestion
	ticker := time.AfterFunc(interval, func() { _ = h.doCookieAuth() })
```

plugins/inputs/http/http.go
```diff
@@ -116,13 +136,69 @@ func (h *HTTP) Init() error {
...
+}
+
+func (h *HTTP) doCookieAuth() error {
+	request, err := h.newRequest(h.CookieAuthURL, h.CookieAuthMethod, h.CookieAuthBody)
+	if err != nil {
+		return err
+	}
+	defer request.Body.Close()
+
+	errBadRespCode := fmt.Errorf("bad response code")
```
srebhan, line 185: Please move this to the only usage as it does not add any value but makes the code harder to debug.

plugins/inputs/http/http.go
```diff
@@ -116,13 +136,69 @@ func (h *HTTP) Init() error {
...
+		return respErr
+	}
+	if _, err = io.Copy(ioutil.Discard, resp.Body); err != nil {
+		return err
+	}
+	if err = resp.Body.Close(); err != nil {
+		return err
+	}
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("%w: %v", errBadRespCode, resp.StatusCode)
```
srebhan, line 197: Just use
```suggestion
		return fmt.Errorf("bad response code: %v", resp.StatusCode)
```
here.

plugins/inputs/http/http.go
```diff
@@ -116,13 +136,69 @@ func (h *HTTP) Init() error {
...
+	resp, respErr := h.client.Do(request)
+	if respErr != nil {
+		return respErr
+	}
+	if _, err = io.Copy(ioutil.Discard, resp.Body); err != nil {
+		return err
+	}
+	if err = resp.Body.Close(); err != nil {
+		return err
+	}
```
srebhan, line 195: I think you want this to be
```suggestion
	resp, respErr := h.client.Do(request)
	if respErr != nil {
		return respErr
	}
	defer resp.Body.Close()

	if _, err := io.Copy(ioutil.Discard, resp.Body); err != nil {
		return err
	}
```
to avoid skipping the body-close if the copy fails (which should never be the case I guess).

plugins/inputs/http/http.go
```diff
@@ -235,8 +281,46 @@ func (h *HTTP) gatherURL(
 	return nil
 }
 
+func (h *HTTP) newRequest(url, method, body string) (*http.Request, error) {
+	bodyRC, err := makeRequestBodyReader(h.ContentEncoding, body)
+	if err != nil {
+		return nil, err
+	}
+
```
srebhan, line 289: Are you missing the
```suggestion
	defer bodyRC.Close()

```
here?

plugins/inputs/http/http.go
```diff
@@ -116,13 +136,69 @@ func (h *HTTP) Init() error {
...
+			_ = h.doCookieAuth()
+		}
+	}()
+
+	// initial auth will immediately error out the Init() if auth fails
+	return h.doCookieAuth()
+}
+
+func (h *HTTP) doCookieAuth() error {
+	request, err := h.newRequest(h.CookieAuthURL, h.CookieAuthMethod, h.CookieAuthBody)
```
srebhan, line 179: I don't think reusing the existing http-client will work here. Let me explain why: in `newRequest()` you are setting the parameters for the request including (shortened)
```golang
	bearer := "Bearer " + strings.Trim(string(token), "\n")
	request.Header.Set("Authorization", bearer)

	request.Header.Set("Content-Encoding", "gzip")

	for k, v := range h.Headers {
		request.Host = v
		request.Header.Add(k, v)
	}

	request.SetBasicAuth(h.Username, h.Password)
```
Now you use a completely different URL to issue the request, reusing the data originally specified for one URL on another URL. Not only, these setting might potentially settings between the URL but you also **leak sensible data** to a (probably) different server than initially intended by the user.

This being said, I think you need to setup a second http-client to perform the cookie-authentication, potentially also allowing the user to specify encoding, credentials, headers,...

plugins/inputs/http/README.md
```diff
@@ -22,6 +22,12 @@ The HTTP input plugin collects metrics from one or more HTTP(S) endpoints.  The
   ## HTTP entity-body to send with POST/PUT requests.
   # body = ""
 
+  ## Cookie-based authentication
+  # cookie_auth_url = "https://localhost/authMe"
+  # cookie_auth_method = "POST" # default: POST
+  # cookie_auth_body = '{"username": "user", "password": "pa$$word", "authenticate": "me"}'
+  # cookie_auth_renewal = "8h" # default: "5m"
+
```
srebhan, line 30: Please sync to sampleConfig after apply changes for my comments there.

srebhan (changes requested):
Hey @jh125486, thanks for tackling this issue!

The code, with one exeception, looks good with only minor issues (see my comments in the code). However, I have a big concern reusing the http client for the cookie authentication as you are also reusing e.g. username and password with a URL potentially different to the settings in `urls`. **This might leak sensible data.**

Furthermore, I ask myself if this new feature wouldn't be a good candidate to add to `plugins/common/http` so it can be reused by other plugins as well. Do you think this makes sense?

plugins/inputs/http/http.go
jh125486: Having them as `vars` was a force of habit for changing them during tests.  I'll move them into `Init()`.

plugins/inputs/http/http.go
jh125486: I was under the impression that `time.AfterFunc()`  only fires once?

plugins/inputs/http/http.go
jh125486: Will do.  Another golangci-lint complaint.

plugins/inputs/http/http.go
jh125486: Sure.  Force of habit to check errors.

plugins/inputs/http/http.go
jh125486: The request isn't sent by function exit, so I thought that closing the body would cause a read error in http client.

plugins/inputs/http/http.go
jh125486: Yes, I can do that.
The original use case was for a single host, so that's the direction I took since the URL slice is already "bound" to a single HTTP Basic config (user/pass) in the same config.

jh125486:
> Hey @jh125486, thanks for tackling this issue!
> 
> The code, with one exeception, looks good with only minor issues (see my comments in the code). However, I have a big concern reusing the http client for the cookie authentication as you are also reusing e.g. username and password with a URL potentially different to the settings in `urls`. **This might leak sensible data.**
> 
> Furthermore, I ask myself if this new feature wouldn't be a good candidate to add to `plugins/common/http` so it can be reused by other plugins as well. Do you think this makes sense?

I can move it to `plugins/common/http` ... I wasn't even aware that was there really :)

plugins/inputs/http/http.go
srebhan: Oh I think you are right. Sorry for the noise!

plugins/inputs/http/http.go
srebhan: Well but not closing the body will leak memory I think... :-/

plugins/inputs/http/http.go
srebhan: TBH, I also don't like the ability to specify multiple URLs as it calls for errors for exactly the reason you mentioned. However, we should not make things worse by adding another issue. :-)

plugins/common/cookie/cookie.go
```diff
@@ -0,0 +1,91 @@
...
+		return err
+	}
+
+	// continual auth renewal
+	ticker := time.NewTicker(time.Duration(c.Renewal))
+	go func() {
+		for range ticker.C {
+			_ = c.auth()
+		}
+	}()
```
srebhan, line 45: Can we somehow also model the case of non-repetitive retrieval. I.e. we only get the cookie once without the need to specify an incredibly large renewal period. We could for example check for a "zero" renewal time and only request the cookie once...

plugins/common/cookie/cookie.go
```diff
@@ -0,0 +1,91 @@
...
+	Username string `toml:"cookie_auth_username"`
+	Password string `toml:"cookie_auth_password"`
+
+	Body    string          `toml:"cookie_auth_body"`
+	Renewal config.Duration `toml:"cookie_auth_renewal"`
+
+	client *http.Client
+}
+
+func (c *CookieAuthConfig) Start(client *http.Client) (err error) {
```
srebhan, line 29: Hmm, is there a reason for passing in a client? You could also setup the client here...

plugins/common/cookie/cookie.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+func (c *CookieAuthConfig) setDefaults() {
+	if c.Renewal == 0 {
+		c.Renewal = config.Duration(5 * time.Minute)
+	}
+	if c.Method == "" {
+		c.Method = http.MethodPost
+	}
+}
```
srebhan, line 58: Can you please move this to the `Start()` function so we have the whole flow in one place.

plugins/common/cookie/cookie.go
```diff
@@ -0,0 +1,91 @@
...
+	// add cookie jar to HTTP client
+	if c.client.Jar, err = cookiejar.New(nil); err != nil {
+		return err
+	}
+
+	// continual auth renewal
+	ticker := time.NewTicker(time.Duration(c.Renewal))
+	go func() {
+		for range ticker.C {
+			_ = c.auth()
```
srebhan, line 43: The only issue I see here is that we cannot see errors produced in `auth`. Do you have a good idea on how to pass the error on to the caller? Alternatively you might at least want to log the errors...

plugins/common/cookie/cookie.go
```diff
@@ -0,0 +1,91 @@
...
+		body = ioutil.NopCloser(strings.NewReader(c.Body))
+		defer body.Close()
+	}
+
+	req, err := http.NewRequest(c.Method, c.URL, body)
+	if err != nil {
+		return err
+	}
+
+	if c.Username != "" || c.Password != "" {
```
srebhan, line 72: Maybe just do 
```suggestion
	if c.Username != "" {
```
to avoid cases where a password is set without username?

srebhan (commented):
Wow that's really nice. I have some more comments and questions, but I think we are very close.  Thanks for working on this!

plugins/common/cookie/cookie.go
jh125486: I honestly don't have a good solution.... The telegraf logger instance isn't passed into any of the plugins, so I guess it's up to each plugin to do their own logging?

I can just do a raw log with `log`.

plugins/common/cookie/cookie.go
jh125486: The `client` has to have the same `cookiejar`, so might as well use the same `client`.

srebhan: Understood. Maybe you can add your comment above as a small comment to the function...

plugins/common/cookie/cookie.go
srebhan: Well the correct solution would be to add a 
```golang
```
line to the struct and pass the parents logger. However, with this solution we need to care for the case where the parent does not set this `Log` (e.g. due to the fact that it has no logger itself).

Alternatively, you could create an `error` channel that the caller can subscribe to. Would that make sense?

plugins/common/cookie/cookie.go
```diff
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/influxdata/telegraf/config"
+	"github.com/sirupsen/logrus"
```
srebhan, line 13: Please don't do this. We blacklisted `log` as we want a common logging infrastructure to be used. Circumventing the blacklist by using `logrus` is not very nice. ;-)

plugins/common/cookie/cookie.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	// continual auth renewal if set
+	if c.Renewal > 0 {
+		ticker := time.NewTicker(time.Duration(c.Renewal))
+		go func() {
+			for range ticker.C {
+				if err := c.auth(); err != nil {
+					logrus.WithError(err).Error("cookie auth renewal failure")
+				}
```
srebhan, line 49: How about using a channel like (untested):
```suggestion
				if err := c.auth(); err != nil {
					select {
						case c.ErrorChan <- err:
						default:
					}
				}
```
You need to init `c.ErrorChan = make(chan error)` in `Start()` then and the caller can the read from the channel to log the errors. We could even provide a convenience function for that passing in a `telegraf.Logger` instance...

plugins/common/cookie/cookie_test.go
```diff
@@ -0,0 +1,268 @@
...
+			name: "zero renewal does not renew",
+			args: args{
+				renewal:  0,
+				endpoint: authEndpointNoCreds,
+			},
+			assert: func(t *testing.T, c *cookie.CookieAuthConfig, srv fakeServer) {
+				// should have Cookie Authed once
+				srv.checkAuthCount(t, 1)
+				srv.checkResp(t, http.StatusOK)
+				time.Sleep(renewalCheck)
```
srebhan, line 135: This is critical as we often had problems with flakey tests in the past due to timing issues. Please avoid this kind of "timing" as the time will depend on the load of the CI infrastructure for example.
TBH I have no good idea on how replace this, but the past experience with failing CI tests for unrelated PRs due to timing scares me when seeing this. :-)

srebhan (commented):
Almost. I had one idea about passing the `auth()` errors to the caller using a non-blocking channel. Maybe you can check if this is doable.
Furthermore, I see the timing issues in your tests. While I do not have a good idea on how to do it correctly in your case, I fear flakey tests in later (unrelated) PRs (we have been there...).

plugins/common/cookie/cookie_test.go
jh125486: Yeah, unfortunately the only way to not have this be flakey is to extend out the timing, at the expense of slow tests.

plugins/common/cookie/cookie.go
jh125486: `error` channel or embedded `Log` is good... only issue is since this package is in `common` and used in another `common` pkg (`httpconfig`), I'm not sure how the original caller would use either solution.

I can refactor `httpconfig` to embed a `Log` field (and pass it into the CookieAuth), but I'm not sure how `Log` would even get set on input/output plugins e.g. in `plugins/inputs/http/http.go`.

> Almost. I had one idea about passing the `auth()` errors to the caller using a non-blocking channel. Maybe you can check if this is doable.

My intention with returning the `auth()` error in a blocking fashion was to raise an immediate failure in the case of bad creds in a config. 
I figured if the initial auth() worked (with good creds), then future auth failures would be intermittent (and non blocking.)

srebhan: Well you could do the first call to `auth()` separately in the function before entering the timer and return the error code in a blocking fashion.
However, for the further calls to auth I'd like to have a way for the caller to know if everything is alright still. You could also provide a callback function. No matter what you decide to do, the caller should have a way to react to the failing auth. 
Maybe the best way would be to just provide an exported `Auth()` function and let the caller do the timing... :-)

plugins/common/cookie/cookie_test.go
srebhan: True. For metrics we have the `acc.Wait()` call in the testutil package that uses conditional variables. But it won't work here as you want to test if the call happens only once. However, for the other tests that might be a possible solution.

plugins/common/cookie/cookie.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	// continual auth renewal if set
+	if c.Renewal > 0 {
+		ticker := time.NewTicker(time.Duration(c.Renewal))
+		go func() {
+			for range ticker.C {
+				if err = c.auth(); err != nil {
+					log.Errorf("Error in plugin: [url=%v]: %v", c.URL, err)
+				}
```
srebhan, line 53: I think we should allow plugins to omit the logging if it e.g. does not have a logger itself.
```suggestion
				if err := c.auth(); err != nil && log != nil {
					log.Errorf("renewal failed for %q: %v", c.URL, err)
				}
```

srebhan (commented):
One more suggestion for allowing to omit the logging. There might be some plugins that do not have a logger.
Otherwise this looks good.

plugins/common/cookie/cookie.go
jh125486: Done.

srebhan (approved):
Looks good to me. Thanks @jh125486 for working hard to get this PR to the current state!

helenosheaa (commented):
Thanks for the PR, looks good. I think it would be useful if you could provide more information in the readme as to the use-case of this. So new users can understand why they might want to use verses when they don't need to use this setting.

jh125486:
>  provide more information in the readme as to the use-case of this. So new users can understand why they might want to use verses when they don't need to use this setting.

The `README`s currently do not reference any of the options within the toml, so what section name should I create for this?

srebhan:
@jh125486 I guess @helenosheaa wants a small section of text in `plugins/inputs/http/README.md` (and `plugins/outputs/http/README.md` accordingly) describing under which circumstances a user should fill this. Is there e.g. a certain popular service that uses this type of authentication that you could use as an example?
So something like
```
Cookie authentication allows to derive an authentication cookie from a server providing your credentials as body. A popular example is XYZ which work with the following configuration...
```

jh125486:
> @jh125486 I guess @helenosheaa wants a small section of text in `plugins/inputs/http/README.md` (and `plugins/outputs/http/README.md` accordingly) describing under which circumstances a user should fill this. Is there e.g. a certain popular service that uses this type of authentication that you could use as an example?
> So something like
> 
> ```
> Cookie authentication allows to derive an authentication cookie from a server providing your credentials as body. A popular example is XYZ which work with the following configuration...
> ```

Correct, so what should the name of that section be?

helenosheaa:
@jh125486 perhaps - Optional Cookie Authentication Settings: with the explanation and an example config.

MyaLongmire:
Can you please add a sample config? Also, I am a little confused by your description of the optional cookie authentication. It seems a little wordy.

jh125486:
> This is looking great so far! Can you please add a sample config? Also, is there any way to make your optional cookie authentication section a little shorter? It seems a touch wordy.

1. An example config in addition to the example config in the README? What should that file be called?
2. That section is three sentences long. What is the max number of sentences I should have?

srebhan:
@MyaLongmire there is the usual sample-config part in the `http` input/output sections (options starting with `cookie_auth_`). Do you have a suggestion on how to shorten the text?

MyaLongmire (approved):

Issue #7421, opened 2020-04-28
Wrong type for MySQL enum variables with OFF value


fxedel:
### Relevant telegraf.conf:

```toml
[[inputs.mysql]]
  servers = ["..."]
  metric_version = 2
  interval_slow = "0s"
  # ...
```

### System info:

Telegraf 1.14.0 (git: HEAD fefd7ff1)  
Debian GNU/Linux 9 (stretch)

### Description

Some MySQL variables, like [transaction_write_set_extraction](https://dev.mysql.com/doc/refman/5.7/en/replication-options-binary-log.html#sysvar_transaction_write_set_extraction), allow for multiple string values including `OFF`, e.g.: `OFF`, `MURMUR32` or `XXHASH64`.

I expected these variables to be gathered as string, however, the Off value is converted to the integer 0 in [ParseValue](https://github.com/influxdata/telegraf/blob/7ee776d67a6a5bef7205f04bad90f6cdc79b9884/plugins/inputs/mysql/v2/convert.go#L49). This conversion is good for simple On/Off variables, but doesn't fit for enumerations like transaction_write_set_extraction as it produces errors like these:

> 2020-04-28T08:35:30Z E! [outputs.influxdb] When writing to [https://...:8086]: received error partial write: field type conflict: input field "transaction_write_set_extraction" on measurement "mysql_variables" is type string, already exists as type integer dropped=3; discarding points

### Proposed fix

I don't know what the best solution would be; maybe a hardcoded list of variables that should parse On/Off as string?

Issue #6624, opened 2019-11-06
MySQL Metrics Collision


Esity:
When using one of the most recent versions of Telegraf(1.12.4) and using the MySQL input plugin against multiple MySQL based systems there is some sort of collision going on to cause write drops.

We purged all MySQL measurements in InfluxDB, waited for compactions to finish. We then re enabled the mysql input plugin and immediatly started seeing drops in InfluxDB. Here is the config we used

```
[[inputs.mysql]]
  servers = ["root:password@tcp(127.0.0.1:3306)/"]
  metric_version = 2
  perf_events_statements_digest_text_limit  = 120
  perf_events_statements_limit              = 250
  perf_events_statements_time_limit         = 86400
  table_schema_databases                    = []
  gather_table_schema                       = false
  gather_process_list                       = false
  gather_user_statistics                    = false
  gather_info_schema_auto_inc               = false
  gather_innodb_metrics                     = false
  gather_slave_status                       = false
  gather_binary_logs                        = false
  gather_table_io_waits                     = false
  gather_table_lock_waits                   = false
  gather_index_io_waits                     = false
  gather_event_waits                        = false
  gather_file_events_stats                  = false
  gather_perf_events_statements             = false
  interval_slow                             = "30m"
```

We run Percona, MySQL and MariaDB across many servers. What we noticed is that something is causing the collision and the dropped writes but I can't seem to figure out the exact metric.

All configs on all MySQL based servers are identical except the username/password

danielnelson:
Could you add the log message during write?

Esity:
We are currently sending between 20k and 25k to each InfluxDB env. We have debug enabled for our writers to see what is going on so I will have to do some digging

Esity:
@danielnelson got it

`2019-11-06T17:09:43Z E! [outputs.influxdb]: when writing to [http://hostname]: received error partial write: field type conflict: input field "ssl_ctx_verify_depth" on measurement "mysql" is type float, already exists as type integer dropped=1; discarding points`

danielnelson:
This is #5529, I'll bump up the priority.   Workaround is `fielddrop`:
```toml
[[inputs.mysql]]
  fielddrop = ["ssl_ctx_verify_depth"] # Workaround #5529
```

Closing this one and we can continue on the other issue.

Issue #5711, opened 2019-04-11
MySQL Plugin: inconsistent measurements


bluebob:
### Relevant telegraf.conf:
[[inputs.mysql]]
  servers = ["randompasswordremoved@tcp(127.0.0.1:3306)/"]
  metric_version = 2
  perf_events_statements_digest_text_limit  = 120
  perf_events_statements_limit              = 250
  perf_events_statements_time_limit         = 86400
  table_schema_databases                    = []
  gather_table_schema                       = false
  gather_process_list                       = false
  gather_user_statistics                    = false
  gather_info_schema_auto_inc               = false
  gather_innodb_metrics                     = false
  gather_slave_status                       = false
  gather_binary_logs                        = false
  gather_table_io_waits                     = false
  gather_table_lock_waits                   = false
  gather_index_io_waits                     = false
  gather_event_waits                        = false
  gather_file_events_stats                  = false
  gather_perf_events_statements             = false
  interval_slow                   = "30m"
  fielddrop = ["caching_sha2_password_rsa_public_key","rsa_public_key"]

### System info:
Telegraf 1.10.2 running on RHEL7.5 64 bit

### Steps to reproduce:

Telegraf was configured as per documentation with mysql block as given above, but outputting to a Kafka topic. A separate Telegraf (1.10.0) instance then consumes this topic and writes to an InfluxDB database that has had all mysql* measurements removed.

### Expected behavior:
No errors seen

### Actual behavior:
The Telegraf that is ingesting the data into Influx reports the following repeatedly:

2019-04-11T13:30:06Z E! [outputs.influxdb]: when writing to [http://hostnameremoved:8086]: received error partial write: field type conflict: input field "innodb_data_pending_fsyncs" on measurement "mysql" is type float, already exists as type integer dropped=1; discarding points

Given that innodb_metrics appears to be switched off in the configuration I'm a bit confused as to why this field is being created, and doubly so that it's being written with different value types.

### Additional info:

MySQL instance being polled is community version 8.0.13 running on RHEL7-64bit

danielnelson:
related #5529

bluebob:
Changing the fielddrop line to:

`  fielddrop = ["caching_sha2_password_rsa_public_key", "rsa_public_key", "innodb_data_pending_fsyncs", "innodb_row_lock_current_waits"]`

seems to stop all error messages from the InfluxDB ingest Telegraf.

bglinkerman:
I have run into similar issues with this plugin:

[...]

Issue #5055, opened 2018-11-28
MySQL plugin: received error partial write


bolek2000:
### Relevant telegraf.conf:
```
# # Read metrics from one or many mysql servers
[[inputs.mysql]]
#   ## specify servers via a url matching:
#   ##  [username[:password]@][protocol[(address)]]/[?tls=[true|false|skip-verify|custom]]
#   ##  see https://github.com/go-sql-driver/mysql#dsn-data-source-name
#   ##  e.g.
#   ##    servers = ["user:passwd@tcp(127.0.0.1:3306)/?tls=false"]
#   ##    servers = ["user@tcp(127.0.0.1:3306)/?tls=false"]
#   #
#   ## If no servers are specified, then localhost is used as the host.
#   servers = ["tcp(127.0.0.1:3306)/"]
#    servers = telegraf@unix
servers = ["telegraf@unix(/var/run/mysqld/mysqld.sock)/information_schema"]
#   
#   ## Selects the metric output format.
#   ##
#   ## This option exists to maintain backwards compatibility, if you have
#   ## existing metrics do not set or change this value until you are ready to
#   ## migrate to the new format.
#   ##
#   ## If you do not have existing metrics from this plugin set to the latest
#   ## version.
#   ##
#   ## Telegraf >=1.6: metric_version = 2
#   ##           <1.6: metric_version = 1 (or unset)
   metric_version = 2
#
#   ## the limits for metrics form perf_events_statements
   perf_events_statements_digest_text_limit  = 120
   perf_events_statements_limit              = 250
   perf_events_statements_time_limit         = 86400
#   #
#   ## if the list is empty, then metrics are gathered from all databasee tables
#   table_schema_databases                    = []
#   #
#   ## gather metrics from INFORMATION_SCHEMA.TABLES for databases provided above list
  gather_table_schema                       = false
#   #
#   ## gather thread state counts from INFORMATION_SCHEMA.PROCESSLIST
  gather_process_list                       = true
#   #
#   ## gather user statistics from INFORMATION_SCHEMA.USER_STATISTICS
  gather_user_statistics                    = true
#   #
#   ## gather auto_increment columns and max values from information schema
  gather_info_schema_auto_inc               = true
#   #
#   ## gather metrics from INFORMATION_SCHEMA.INNODB_METRICS
  gather_innodb_metrics                     = true
#   #
#   ## gather metrics from SHOW SLAVE STATUS command output
  gather_slave_status                       = false
#   #
[...]

PR #9403, opened 2021-06-21
Fix inconsistent metric types in mysql

### LINKED ISSUE #6671: Mysql input: Slave_IO_Running can have values other than Yes and No
### Relevant telegraf.conf:

```
[[inputs.mysql]]
   servers = [ "monitoring:XXX@tcp(127.0.0.1:3306)/"]
   gather_process_list                       = true
   gather_user_statistics                    = true
   gather_info_schema_auto_inc               = true
   gather_slave_status                       = true
```

### System info:
Ubuntu 16.04
telegraf 1.12.5-1
influxdb 1.7.9-1
mariadb-server-10.0 10.0.38-0ubuntu0.16.04.1

### Steps to reproduce:

1. Have a replicated mysql server with its plugin mysql configured and monitoring slave side.
    Since everything is right, the value of slave_slave_io_running is 0 (In mysql SHOW SLAVE STATUS\G: `Slave_IO_Running: Yes`
    If slave is stopped (STOP SLAVE;), the value is slave_slave_io_running is 1. (In mysql SHOW SLAVE STATUS\G: `Slave_IO_Running: No`)
2. Stop master myqsl (service mysql stop)
3. The value of slave_slave_io_running is now "Connecting"  (In mysql SHOW SLAVE STATUS\G: `Slave_IO_Running: Connecting`

Since in normal operation, slave_slave_io_running is an integer, when master crashes, it value become a string and influxdb throw it 
`received error partial write: field type conflict: input field "slave_slave_io_running" on measurement "mysql" is type string, already exists as type integer dropped=1; discarding points`

### Expected behavior:

slave_slave_io_running should return an integer value (1 or more in case of connecting)

### Actual behavior:

slave_slave_io_running may return a string or an integer, depending on the content.

### LINKED ISSUE #8319: plugin: input.mysql: 'innodb_available_undo_logs' different datatype
### System info:

Telegraf 1.15.2 (git: HEAD cd037b49)
plugin: inputs.mysql (one for master and one for slave)

### Config

```
##########
# MASTER #
##########

[[inputs.mysql]]
  servers = ["probe:${CORE_DB_PASSWORD}@tcp(xxxx:3306)/"]
  metric_version = 2

  gather_table_schema = false
  gather_process_list = true
  gather_user_statistics = false
  gather_info_schema_auto_inc = false
  gather_innodb_metrics = false
  gather_slave_status = false
  gather_binary_logs = false
  gather_global_variables = true
  gather_table_io_waits = false
  gather_table_lock_waits = false
  gather_index_io_waits = false
  gather_event_waits = false
  gather_file_events_stats = false
  gather_perf_events_statements = false

[[inputs.mysql]]
  servers = ["probe:${CORE_DB_PASSWORD}@tcp(xxx:3307)/"]
  metric_version = 2

  gather_table_schema = false
  gather_process_list = true
  gather_user_statistics = false
  gather_info_schema_auto_inc = false
  gather_innodb_metrics = false
  gather_slave_status = false
  gather_binary_logs = false
  gather_global_variables = true
  gather_table_io_waits = false
  gather_table_lock_waits = false
  gather_index_io_waits = false
  gather_event_waits = false
  gather_file_events_stats = false
  gather_perf_events_statements = false

[[outputs.influxdb_v2]]
  urls = ["${INFLUXDB_URL}"]

  token = "${INFLUXDB_TOKEN}"
  organization = "xxx"
  bucket = "telegraf"
  bucket_tag = "bucket"
  exclude_bucket_tag = true
```

### Expected behavior:

Field `innodb_available_undo_logs` should have the same data type

### Actual behavior:

Field `innodb_available_undo_logs` has different data type (int for master, float for slave)

### Additional info:

master: `SHOW GLOBAL STATUS`: 
```
innodb_available_undo_logs: 128
```

slave: `SHOW GLOBAL STATUS`:
```
innodb_available_undo_logs: 18446744073709551615
```

`telegraf --config input_mysql.conf --test`:
```
> mysql,host=xxxx,server=xxxx:3306 innodb_available_undo_logs=128i,innodb_num_open_files=278i,innodb_os_log_pending_fsyncs=0i,innodb_os_log_pending_writes=0i,innodb_os_log_written=512i,innodb_page_size=16384i,innodb_pages_created=420253i,innodb_pages_read=2678435i,innodb_pages_written=0i,innodb_row_lock_current_waits=0i,innodb_row_lock_time=106780i,innodb_row_lock_time_avg=17i,innodb_row_lock_time_max=20866i,innodb_row_lock_waits=5987i,innodb_rows_deleted=339344i,innodb_rows_inserted=3017851i,innodb_rows_read=10347509999129i,innodb_rows_updated=31345161i,innodb_truncated_status_writes=0i,key_blocks_not_flushed=0i 1603728599000000000
> mysql,host=xxxx,server=xxxx:3307 innodb_available_undo_logs=18446744073709552000,innodb_num_open_files=4442i,innodb_os_log_pending_fsyncs=0i,innodb_os_log_pending_writes=0i,innodb_os_log_written=0i,innodb_page_size=16384i,innodb_pages_created=0i,innodb_pages_read=830425i,innodb_pages_written=0i,innodb_row_lock_current_waits=0i,innodb_row_lock_time=0i,innodb_row_lock_time_avg=0i,innodb_row_lock_time_max=0i,innodb_row_lock_waits=0i,innodb_rows_deleted=0i,innodb_rows_inserted=0i,innodb_rows_read=10081762371564i,innodb_rows_updated=0i,innodb_truncated_status_writes=0i,key_blocks_not_flushed=0i 1603728599000000000
```

error
```
2020-10-26T15:42:00Z E! [agent] Error writing to outputs.influxdb_v2: 500 Internal Server Error: internal error: unexpected error writing points to database: partial write: field type conflict: input field "innodb_available_undo_logs" on measurement "mysql" is type float, already exists as type integer dropped=8
```

Under `outputs.influxdb_v2`, I see a parameter `influx_uint_support`, does this solve the issue?

fxedel:
Fixed two issues with inconsistent metric types in mysql:

* mysql variables that are stored as uint64 in mysql and are too big for int64 (currently parsed as float by telegraf)
* mysql variables with "ON", "OFF" and other string values (currently parsed as integer for "ON"/"OFF" and string otherwise)

The fix included uint parsing (if int fails) and a mapping for some MySQL statuses/variables, pinning them to a specific type.

Note that InfluxDB 1.x doesn't support uint, so outputted values will be capped to max int, but that's a problem with the InfluxDB plugin and not MySQL. (Possible fix: add influxdb option to output uints as floats).

Fixes #8319, #5711, #5055, #7421

See #5529, #6624

This could be easily extended to fix #6671 as well.

### DIFF
--- a/plugins/inputs/mysql/v2/convert.go
+++ b/plugins/inputs/mysql/v2/convert.go
@@ -21,12 +21,20 @@ func ParseInt(value sql.RawBytes) (interface{}, error) {
 	return v, err
 }
 
+func ParseUint(value sql.RawBytes) (interface{}, error) {
+	return strconv.ParseUint(string(value), 10, 64)
+}
+
 func ParseBoolAsInteger(value sql.RawBytes) (interface{}, error) {
 	if bytes.EqualFold(value, []byte("YES")) || bytes.EqualFold(value, []byte("ON")) {
 		return int64(1), nil
 	}
 
 	return int64(0), nil
+}
+
+func ParseString(value sql.RawBytes) (interface{}, error) {
+	return string(value), nil
 }
 
 func ParseGTIDMode(value sql.RawBytes) (interface{}, error) {
@@ -70,12 +78,27 @@ func ParseValue(value sql.RawBytes) (interface{}, error) {
 }
 
 var GlobalStatusConversions = map[string]ConversionFunc{
-	"ssl_ctx_verify_depth": ParseInt,
-	"ssl_verify_depth":     ParseInt,
+	"innodb_available_undo_logs":    ParseUint,
+	"innodb_buffer_pool_pages_misc": ParseUint,
+	"innodb_data_pending_fsyncs":    ParseUint,
+	"ssl_ctx_verify_depth":          ParseUint,
+	"ssl_verify_depth":              ParseUint,
 }
 
 var GlobalVariableConversions = map[string]ConversionFunc{
-	"gtid_mode": ParseGTIDMode,
+	"delay_key_write":                  ParseString,        // ON, OFF, ALL
+	"enforce_gtid_consistency":         ParseString,        // ON, OFF, WARN
+	"event_scheduler":                  ParseBoolAsInteger, // YES, NO, DISABLED
+	"gtid_mode":                        ParseGTIDMode,
+	"have_openssl":                     ParseBoolAsInteger, // alias for have_ssl
+	"have_ssl":                         ParseBoolAsInteger, // YES, DISABLED
+	"have_symlink":                     ParseBoolAsInteger, // YES, NO, DISABLED
+	"session_track_gtids":              ParseString,
+	"session_track_transaction_info":   ParseString,
+	"slave_skip_errors":                ParseString,
+	"ssl_fips_mode":                    ParseString,
+	"transaction_write_set_extraction": ParseString,
+	"use_secondary_engine":             ParseString,
 }
 
 func ConvertGlobalStatus(key string, value sql.RawBytes) (interface{}, error) {


fxedel:
!signed-cla

srebhan (approved):
Looks good codewise, but I have no way of testing this...

MyaLongmire:
@Eraac @bluebob @bolek2000 This pull request might resolve the issues you opened, if you are still having the issues would you mind testing this fix? https://github.com/influxdata/telegraf/pull/9403#issuecomment-865172715

reimda:
Hi @fxedel I took a quick look at the changes here. I'm a little concerned about backward compatibility. If someone was using "transaction_write_set_extraction" or any of the other ~20 variables which change type because of this PR, they may run into issues.  Do you think these fixes would require a new metric_version=3?

fxedel:
Hi @reimda, the issues that could be introduced by this PR are actually the same issues that we're facing today. For example `transaction_write_set_extraction` is currently stored either as int (when `OFF`) or string (when `MURMUR32`) – see #7421. With this PR, it's always stored as string. If the variable has always been set to `OFF`, it will produce new issues, but, as least for InfluxDB, only until the next shard is created. If the variable has always been changing, the issues were always present and will disappear in future. If the variable has always been set to `MURMUR32`, there will be no issues.

As this is merely a bug fix, I wouldn't introduce a new metric version, as all users of v2 would benefit in the long-term. A typical shard duration is one week, so InfluxDB users would see no issues after at most seven days.

Of course, a new metric version v3 would also be acceptable.

plugins/inputs/mysql/v2/convert.go
```diff
@@ -70,12 +81,27 @@ func ParseValue(value sql.RawBytes) (interface{}, error) {
...
 
 var GlobalVariableConversions = map[string]ConversionFunc{
-	"gtid_mode": ParseGTIDMode,
+	"delay_key_write":                  ParseString,        // ON, OFF, ALL
+	"enforce_gtid_consistency":         ParseString,        // ON, OFF, WARN
+	"event_scheduler":                  ParseBoolAsInteger, // YES, NO, DISABLED
+	"gtid_mode":                        ParseGTIDMode,
+	"have_openssl":                     ParseBoolAsInteger, // alias for have_ssl
+	"have_ssl":                         ParseBoolAsInteger, // YES, DISABLED
+	"have_symlink":                     ParseBoolAsInteger, // YES, NO, DISABLED
```
reimda, line 98: event_scheduler and have_symlink have three states: YES, NO, DISABLED. If we use ParseBoolAsInteger, YES will return 1, but both NO and DISABLED return 0. This makes it impossible to know which of those two mysql returned. Is there a reason you didn't convert with ParseString in this case?

Similarly, have_ssl and have_openssl have YES and DISABLED. Using ParseBoolAsInteger isn't ambiguous here because there are only two options, but if we return 0 for NO everywhere else, maybe DISABLED should be some other value?

plugins/inputs/mysql/v2/convert.go
```diff
@@ -70,12 +81,27 @@ func ParseValue(value sql.RawBytes) (interface{}, error) {
...
+	"session_track_gtids":              ParseString,
+	"session_track_transaction_info":   ParseString,
+	"slave_skip_errors":                ParseString,
+	"ssl_fips_mode":                    ParseString,
+	"transaction_write_set_extraction": ParseString,
+	"use_secondary_engine":             ParseString,
```
reimda, line 104: Are there other variables that may need to be parsed as string? Did you add only the variables here that we know are causing trouble or were you able to find a list of vars and expected values in mysql docs or code and make an exhaustive list? The current list of fixes is great, I just wonder if there are more out there that need to be fixed.

reimda (changes requested):
Thanks for the explanation, I see what you mean that issues that could be introduced by this PR are actually already happening. I think this change will fix a lot of the existing mysql conversion problems.

I have a couple more questions after looking a little more closely at the code.

plugins/inputs/mysql/v2/convert.go
fxedel: Hey @reimda, I intended to make this PR as backwards-compatible as possible. `have_symlink` is probably almost always `YES` or `NO`, which used to be parsed as integer. `DISABLED` is only set if the MySQL server was started with a specific option. Changing this variable to string would probably cause a type conflict for most users, so I decided to keep the integer type and make `DISABLED` also 0.

Similar considerations apply to `have_ssl` as well.

However, I've got no strong opinion on this, so I could also parse all `DISABLED` fields as string.

plugins/inputs/mysql/v2/convert.go
fxedel: Hey @reimda, I took variables from https://dev.mysql.com/doc/refman/5.7/en/server-system-variables.html. I guess there is no simple list of string variables.

We can extend our mapping with more variables at any time.

PR #9411, opened 2021-06-22
feat: Support AWS Web Identity Provider

### LINKED ISSUE #9073: Support AWS Web Identity Provider
The AWS SDKs have support for authenticating to the AWS API by using a Web Identity Provider. This is very convenient when running services inside of EKS ([Elastic Kubernetes Service](https://aws.amazon.com/eks/)) as it allows you to grant specific services inside of Kubernetes access to a specific AWS IAM role, which controls which permission that service has.

[This page](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts-technical-overview.html) talks about how to configure it in Kubernetes, while [this page](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts-minimum-sdk.html) lists which SDK versions support this feature. It mentions you need at least aws-sdk-go version 1.23.13, this codebase is currently on version 1.34.34, which is plenty:
https://github.com/influxdata/telegraf/blob/ffe92a7658d31e99750bcf81e87b94e849ed9364/go.mod#L29

Usually, it would then be enough to run the service inside EKS with the correct configuration set up for the service account, etc., and the service would then use the provided role when interacting with AWS, but that doesn't work for Telegraf. I'm guessing it's missing the necessary configuration inside https://github.com/influxdata/telegraf/blob/master/config/aws/credentials.go.

Would that be possible to add?

The Python SDK documentation also explains it pretty well:
https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html#assume-role-with-web-identity-provider

akrantz01:
### Required for all PRs:

resolves #9073 

Adds support for authenticating with the AWS API using a [Web Identity Provider](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_providers_oidc.html) using the fields `role_arn`, `web_identity_token_file` and `role_session_name`. However, only `role_arn` and `web_identity_token_file` are required to use this authentication method as `role_session_name` will default to the current timestamp.

### DIFF
--- a/config/aws/credentials.go
+++ b/config/aws/credentials.go
@@ -9,16 +9,16 @@ import (
 )
 
 type CredentialConfig struct {
-	Region               string
-	AccessKey            string
-	SecretKey            string
-	RoleARN              string
-	Profile              string
-	Filename             string
-	Token                string
-	EndpointURL          string
-	RoleSessionName      string
-	WebIdentityTokenFile string
+	Region               string `toml:"region"`
+	AccessKey            string `toml:"access_key"`
+	SecretKey            string `toml:"secret_key"`
+	RoleARN              string `toml:"role_arn"`
+	Profile              string `toml:"profile"`
+	Filename             string `toml:"shared_credential_file"`
+	Token                string `toml:"token"`
+	EndpointURL          string `toml:"endpoint_url"`
+	RoleSessionName      string `toml:"role_session_name"`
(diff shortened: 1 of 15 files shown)


sspaink (approved):
Great job, I like the way you've reorganized the credential config looks a lot nicer.

ssoroka (approved):

PR #9419, opened 2021-06-23
feat: Add support of aggregator as Starlark script

### LINKED ISSUE #8869: Officially support aggregator Starlark scripts
## Feature Request

support calling push() in Starlark scripts so that Starlark processor scripts can fully implement all the features that aggregators would need.

### Proposal:

allow running starlark as an aggregator. this would probably mean supporting something like this:
```toml
[[aggregators.starlark]]
  period = "30s"
  delay = "5m"
  grace = "5m" 
  drop_original = true
  source = '''
state = {}

def apply(metric): # or should this be named add() like the aggregators?
    last = state["last"]
    if last != None:
        metric.fields["value"] = last.fields["value'] + metric.fields["value"]
    state["last"] = deepcopy(metric) # do I need to copy it if I'm not returning it?
    return None

def push(): # called when we push this group of metrics down the line
    last = state["last"]
    state["last"] = None
    return last
'''
```

(see [Aggregator Plugins](https://github.com/influxdata/telegraf/blob/master/docs/CONFIGURATION.md#aggregator-plugins) for details on aggregator settings)

### Current behavior:

There is no call to push, and as such it's not possible to guarantee the aggregate record will be written out in the right aggregation window. This can be problematic for infrequently collected metrics.

### Desired behavior:

support writing aggregators in Starlark.

### Use case: 

People have asked for this, and will continue to build it in processors if we don't add it, though not ideal.

essobedo:
### Required for all PRs:

resolves #8869


essobedo:
@ssoroka It is not yet ready since the doc is missing but I would like to be sure that it is the right direction so could you please tell me WDYT of this PR?

plugins/aggregators/starlark/README.md
```diff
@@ -0,0 +1,252 @@
+# Starlark Processor
```
essobedo, line 1: Ignore me as long as it is a draft

plugins/aggregators/starlark/starlark.go
```diff
@@ -0,0 +1,137 @@
+package starlark //nolint
```
essobedo, line 1: Needed to avoid getting `import-shadowing: The name 'starlark' shadows an import name (revive)`

plugins/aggregators/starlark/starlark.go
```diff
@@ -0,0 +1,137 @@
...
+  ## Source of the Starlark script.
+  source = '''
+def add(cache, metric):
+  cache["last"] = metric
+
+def push(cache, accumulator):
+  last = cache.get("last")
+  if last != None:
+	  accumulator.add_fields(last.name, last.fields, last.tags)
+'''
```
ssoroka, line 28: should the push automatically clear the cache or should that be in the example?

plugins/aggregators/starlark/testdata/merge.star
```diff
@@ -0,0 +1,25 @@
...
+
+load('time.star', 'time')
+
+def add(cache, metric):
+    metrics = cache.get("metrics")
+    if metrics == None:
+        metrics = {}
+        cache["metrics"] = metrics
+        cache["ordered"] = []
+    m = metrics.get(metric)
```
ssoroka, line 11: I don't understand using metric as a key name here. is it automatically converted to a string or something?

plugins/aggregators/starlark/testdata/merge.star
```diff
@@ -0,0 +1,25 @@
...
+    if m == None:
+        m = deepcopy(metric)
+        metrics[metric] = m 
+        cache["ordered"].append(m)
+    else:
+        for k, v in metric.fields.items():
+            m.fields[k] = v
```
ssoroka, line 18: I'm surprised you don't need to write this back to the cache after changing it.

plugins/aggregators/starlark/testdata/merge.star
```diff
@@ -0,0 +1,25 @@
...
+
+def push(cache, accumulator):
+    accumulator.set_precision(time.nanosecond)
+    metrics = cache.get("ordered")
+    if metrics != None:
+        for m in metrics:
+             accumulator.add_metric(m)
```
ssoroka, line 25: is ordered literally the same objects as the one in the metrics cache?

plugins/aggregators/starlark/testdata/min_max.star
```diff
@@ -0,0 +1,40 @@
...
+                elif v > min_max["max"]:
+                    aggregate["fields"][k]["max"] = v
+        
+def push(cache, accumulator):
+    for a in cache:
+        fields = {}
+        for k in cache[a]["fields"]:
+            fields[k + "_min"] = cache[a]["fields"][k]["min"]
+            fields[k + "_max"] = cache[a]["fields"][k]["max"]
+        accumulator.add_fields(cache[a]["name"], fields, cache[a]["tags"])
```
ssoroka, line 40: any thoughts on why we should go with an accumulator as opposed to just returning an array of metrics?

plugins/processors/starlark/starlark.go
```diff
@@ -144,9 +89,9 @@ func (s *Starlark) Start(_ telegraf.Accumulator) error {
 }
 
 func (s *Starlark) Add(metric telegraf.Metric, acc telegraf.Accumulator) error {
-	s.args[0].(*Metric).Wrap(metric)
+	s.args[0].(*common.Metric).Wrap(metric)
```
ssoroka, line 92: should this be a type check instead of a type assertion? if it fails it will panic. is that controlled by the user script?

ssoroka (commented):
looking pretty good. left a few questions

plugins/processors/starlark/starlark.go
essobedo: I simply did what we currently do in the [Starlark processor](https://github.com/influxdata/telegraf/blob/master/plugins/processors/starlark/starlark.go#L147). I can use a type check instead but if so I will modify the processor too for consistency. WDYT?

plugins/aggregators/starlark/testdata/min_max.star
essobedo: If you check well, to test the implementation of the aggregator based on a Starlark script, I simply rewrote 2 existing aggregators which are minmax and merge with the corresponding Starlak script. And I tested them with the exact same tests. In this case, it is the minmax aggregator, and this is actually [the equivalent of the Push function](https://github.com/influxdata/telegraf/blob/master/plugins/aggregators/minmax/minmax.go#L89-L98). 

Regarding your initial question, to be honest, I don't know 😄

plugins/aggregators/starlark/testdata/merge.star
essobedo: It is the equivalent of the [Push function of the merge aggregator](https://github.com/influxdata/telegraf/blob/master/plugins/aggregators/merge/merge.go#L46-L48), it is only a simple rewrite, unless I miss something. Keep in mind that they are only here for testing purpose as POC.

plugins/aggregators/starlark/testdata/merge.star
essobedo: It is possible because `metrics` is a dictionary which is actually [an hashtable](https://github.com/google/starlark-go/blob/master/starlark/hashtable.go) and I added [an implementation of the function `Hash()`](https://github.com/influxdata/telegraf/pull/9419/files#diff-68dab7d4d08265bd0c597d5e5106b8b9fc5d15ea3ebe48743c3a8f7ef9b6f172R64) to be able to store a metric as a key

plugins/aggregators/starlark/starlark.go
essobedo: The cache is automatically cleared [on Reset](https://github.com/influxdata/telegraf/pull/9419/files#diff-e85b90e00ee23e36285bc26a1d203912e20d681bb6435d13b3a408d9d7a6e781R116)

plugins/aggregators/starlark/testdata/merge.star
essobedo: It is actually what we do in [scale.star](https://github.com/influxdata/telegraf/blob/master/plugins/processors/starlark/testdata/scale.star#L12).Maybe it works because behind the scenes, it actually manipulates the pointer of the metric?

sjwang90:
@essobedo - would you like someone to continue reviewing this PR? If so, feel free to take it out of its draft state.

essobedo:
@sjwang90  
> @essobedo - would you like someone to continue reviewing this PR? If so, feel free to take it out of its draft state.
Yes please

essobedo:
> @essobedo - would you like someone to continue reviewing this PR? If so, feel free to take it out of its draft state.

@sjwang90 yes please 🙏

plugins/aggregators/starlark/README.md
```diff
@@ -0,0 +1,91 @@
+# Starlark Aggregator
+
+The `starlark` aggregator calls a Starlark function to add the metrics to the aggregator,
+then calls another Starlark function to push the aggregates to the accumulator. Both Starlark functions
+have an argument called `cache` which is a [dict-like][dict] object allowing to keep temporary the metrics to
+aggregate. The `cache` is automatically cleared by the the aggregator.
```
reimda, line 6: The starlark processor already has a global dict called 'state'. This serves a very similar purpose to the cache argument here. It seems like the cache argument could be removed and the aggregator starlark code could use the state dict.

https://github.com/influxdata/telegraf/blob/56398237c4a931dba438db553a1ee1487926c072/plugins/processors/starlark/README.md#L170

If we have both dicts, I think users who write both aggregators and processors may be confused by the similarity. If we can cover both use cases with the same dict, it will be more user friendly.

plugins/aggregators/starlark/README.md
```diff
@@ -0,0 +1,91 @@
...
+```
+
+### Usage
+
+The Starlark code should contain a function called `add` that takes a cache as first argument and a metric as
+second argument. The function will be called with each metric to add them to the cache, and doesn't return anything.
+
+```python
+def add(cache, accumulator):
+	cache["last"] = metric
```
reimda, line 54:
```suggestion
def add(cache, metric):
	cache["last"] = metric
```

plugins/aggregators/starlark/README.md
```diff
@@ -0,0 +1,91 @@
...
+```
+
+The Starlark code should also contain a function called `push` that takes a cache as first argument and an accumulator as second argument. The function will be called to compute the aggregation from the metrics available in the cache, and doesn't return anything.
+
+```python
+def push(cache, metric):
+  last = cache.get("last")
+  if last != None:
+	  accumulator.add_fields(last.name, last.fields, last.tags)
```
reimda, line 63:
```suggestion
def push(cache, accumulator):
  last = cache.get("last")
  if last != None:
	  accumulator.add_fields(last.name, last.fields, last.tags)
```

plugins/aggregators/starlark/testdata/min_max.star
reimda: In the starlark processor you just return the new metric or list of metrics from the apply function. I think it would make sense to follow the same pattern here and return the new metric from the push function instead of adding it to the accumulator. If we return the new metrics, there will be no need for the function to have an accumulator at all.

It might also make sense to rename push() to apply() to make starlark aggregators more like starlark processors.

reimda (changes requested):

plugins/aggregators/starlark/README.md
essobedo: I understand your point but if we do so it means that the state has to be added explicitly in all the Starlark script like [in this example] (https://github.com/influxdata/telegraf/blob/master/plugins/processors/starlark/testdata/compare_metrics.star#L10-L12). In the processor, it has been done this way as it is seen as a corner case while in case of an aggregator it is required to have a shared state/cache that's why I added it explicitly as an argument of the functions. WDYT?

essobedo:
@srebhan Hi, could you please tell me what you think of this PR?

essobedo:
@sjwang90 Yes it is what I did except for remarks for which I gave an answer, as I did not get any approval, I considered them as still open

plugins/aggregators/starlark/testdata/min_max.star
srebhan: I second that.

essobedo: @srebhan my bad, I did it already, I just forgot to resolve it

plugins/common/starlark/starlark.go
```diff
@@ -0,0 +1,138 @@
...
+// Call calls the function fn with the specified positional and keyword arguments.
+func (s *StarlarkCommon) Call(fn starlark.Value, args starlark.Tuple) (starlark.Value, error) {
+	return starlark.Call(s.thread, fn, args, nil)
+}
+
+func NewStarlarkCommon(fn func(module string, logger telegraf.Logger) (starlark.StringDict, error)) StarlarkCommon {
+	return StarlarkCommon{
+		starlarkLoadFunc: fn,
+	}
+}
```
srebhan, line 105: I would rather avoid this...

plugins/processors/starlark/starlark_test.go
```diff
@@ -3387,3 +3308,29 @@ func testLoadFunc(module string, logger telegraf.Logger) (starlark.StringDict, e
...
 }
+
+func newStarlarkFromSrouce(source string) *Starlark {
+	inner := common.NewStarlarkCommon(testLoadFunc)
+	inner.Log = testutil.Logger{}
+	inner.Source = source
+	return &Starlark{
+		StarlarkCommon: inner,
+	}
+}
```
srebhan, line 3319: Typo here, furthermore I would initialize the inner struct directly using
```suggestion
func newStarlarkFromSource(source string) *Starlark {
	plugin := Starlark {
		StarlarkCommon: common.StarlarkCommon{
			starlarkLoadFunc: testLoadFunc,
			Log: testutil.Logger{},
			Source: source,
		},
	}
	return &plugin
}
```
You could then avoid setting the source and this way merge `newStarlarkFromSrouce`, `newStarlarkFromScript` and `newStarlarkNoScript` by just setting the corresponding field in the caller. Or you move this code to the caller directly.

srebhan (changes requested):
Thank you @essobedo  for the cool PR. I think it makes sense to provide a scriptable aggregator. We need to work a bit on the structure though. I second the view of @reimda to use the `state` as a shared thing to persist values and I think there is nothing speaking against such an approach. Furthermore, we should follow the Aggregator interface also in starlark, i.e. having a `add`, `push` and `reset` function as people are used to those and there is documentation on what they are intended to do.

What do you think?

plugins/aggregators/starlark/README.md
```diff
@@ -0,0 +1,87 @@
...
+```python
+def add(cache, metric):
+  cache["last"] = metric
+```
+
+The Starlark code should also contain a function called `apply` that takes a cache as argument. The function will be called to compute the aggregation from the metrics available in the cache, and returns the metrics to push to the accumulator.
+
+```python
+def apply(cache):
+  return cache.get("last")
```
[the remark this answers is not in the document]
essobedo: It was initially called `push` but I have been asked to rename it to `apply` like in the processor plugin

plugins/aggregators/starlark/README.md
```diff
@@ -0,0 +1,87 @@
...
+```
+
+## Usage
+
+The Starlark code should contain a function called `add` that takes a cache as first argument and a metric as
+second argument. The function will be called with each metric to add them to the cache, and doesn't return anything.
+
+```python
+def add(cache, metric):
+  cache["last"] = metric
```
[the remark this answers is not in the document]
essobedo: As I explained in a previous comment, if I do so we will need to define the state in all the Starlark scripts ([like in this example](https://github.com/influxdata/telegraf/blob/master/plugins/processors/starlark/testdata/compare_metrics.star#L10-L12)) which I wanted to avoid

plugins/aggregators/starlark/starlark.go
```diff
@@ -0,0 +1,163 @@
...
+				s.Log.Errorf("Invalid type returned in list: %s", v.Type())
+			}
+		}
+	case *common.Metric:
+		m := rv.Unwrap()
+		acc.AddMetric(m)
+	case starlark.NoneType:
+	default:
+		s.Log.Errorf("Invalid type returned: %T", rv)
+	}
```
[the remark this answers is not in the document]
essobedo: I did not move it to common, because there are some differences with the processor. For example, I'm not sure that it makes to drop a metric from an aggregator, does-it?

plugins/processors/starlark/starlark_test.go
essobedo: If so `starlarkLoadFunc` must be public

essobedo:
@srebhan thank you for your feedbacks, I addressed your remarks, please check again

plugins/aggregators/starlark/starlark.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+func (s *Starlark) Reset() {
+	s.Call(s.resetFunc, s.resetArgs) //nolint - error already checked within the Call function
```
srebhan, line 136: Can you please log that error here similar to `Push()`? The user might see unexpected things due to an incomplete reset if that call fails...

essobedo: Unless I missed your point, the error is already logged in the `Call` method itself as you can see [here](https://github.com/influxdata/telegraf/pull/9419/files#diff-2be540a22df916b660a365845298b907db028df3946588562400b20f747196b3R113-R119)

plugins/common/starlark/starlark.go
```diff
@@ -0,0 +1,154 @@
...
+	var fn *starlark.Function
+	if fn, ok = globalFn.(*starlark.Function); !ok {
+		return nil, fmt.Errorf("%s is not a function", fnName)
+	}
+
+	if fn.NumParams() != expectedParams {
+		return nil, fmt.Errorf("%s function must take %d parameter(s)", fnName, expectedParams)
+	}
+	return fn, nil
+}
```
srebhan, line 88: How about calling this `AddFunction` and extend the scope. You could do
```suggestion
func (s *StarlarkCommon) AddFunction(name string, params ...starlark.Tuple) (*starlark.Function, starlark.Tuple, error) {
	if _, found := s.globals[name]; !found {
		return nil, nil, fmt.Errorf("%s is not defined", name)
	}

	fn, ok := globalFn.(*starlark.Function)
	if !ok {
		return nil, nil, fmt.Errorf("%s is not a function", fnName)
	}

	if fn.NumParams() != len(params) {
		return nil, nil, fmt.Errorf("%s function must take %d parameter(s)", name, len(params))
	}
	
	p := make(starlark.Tuple, len(params))
	for i, param := range params {
		p[i] = param
	}
	return fn, p, nil
}
```

srebhan: Or even better, make a `functions` and a `parameters` map in the struct and do
How about calling this `AddFunction` and extend the scope. You could do
```suggestion
func (s *StarlarkCommon) AddFunction(name string, params ...starlark.Tuple) error {
	if _, found := s.globals[name]; !found {
		return fmt.Errorf("%s is not defined", name)
	}

	fn, ok := globalFn.(*starlark.Function)
	if !ok {
		return fmt.Errorf("%s is not a function", name)
	}

	if fn.NumParams() != len(params) {
		return nil, nil, fmt.Errorf("%s function must take %d parameter(s)", name, len(params))
	}
	
	s.functions[name] = fn
	s.parameters[name] = params

	return nil
}
```

plugins/common/starlark/starlark.go
```diff
@@ -0,0 +1,154 @@
...
+	rv, err := starlark.Call(s.thread, fn, args, nil)
+	if err != nil {
+		if err, ok := err.(*starlark.EvalError); ok {
+			for _, line := range strings.Split(err.Backtrace(), "\n") {
+				s.Log.Error(line)
+			}
+		}
+	}
+	return rv, err
+}
```
srebhan, line 121: If you change the internal function and parameter representation to a map you could do
```suggestion
func (s *StarlarkCommon) Call(name string) (starlark.Value, error) {
	fn, ok := s.functions[name]
	if !ok {
		return nil, fmt.Errorf("function %q does not exist", name)
	}
	args, ok := s.parameters[name]
	if !ok {
		return nil, fmt.Errorf("params for function %q do not exist", name)
	}
	
	rv, err := starlark.Call(s.thread, fn, args, nil)
	if err != nil {
		if err, ok := err.(*starlark.EvalError); ok {
			for _, line := range strings.Split(err.Backtrace(), "\n") {
				s.Log.Error(line)
			}
		}
	}
	return rv, err
}
```

plugins/aggregators/starlark/starlark.go
srebhan: You are correct, but only for starlark errors (type starlark.EvalError). All other issues will go unnoticed...

srebhan (commented):
Very nice update @essobedo! Some more ideas to move more code into common... :-)

plugins/common/starlark/starlark.go
essobedo: Hum, I'm not really a big fan of this map approach, using a map data structure just to store one element for the processor and three for the aggregator sounds like overkill, don't you agree?

essobedo:
@srebhan I improved the function `InitFunction` based on your remark and used the same code to manage the resulting value.

plugins/common/starlark/starlark.go
srebhan: Yeah it's kind of overkill, but it moves more code into common, making it easier for people to reuse it. There is e.g. a discussion on scripting http requests/responses which could also use Starlark and maybe would need to add a bunch of functions.
And last but not least, look at the beauty of this `AddFunction` above! ;-P

plugins/common/starlark/starlark.go
```diff
@@ -0,0 +1,199 @@
...
+				s.Log.Error(line)
+			}
+		} else {
+			s.Log.Error(err.Msg)
+		}
+	}
+	return rv, err
+}
+
+func (s *StarlarkCommon) AddResultingValue(rv starlark.Value, acc telegraf.Accumulator, onListMetricItem func(m telegraf.Metric) bool,
```
srebhan, line 135: `acc` has to be the first parameter.

srebhan (commented):
Thanks @essobedo for the update. It seems like I underestimated the complexity of the addition logic (in `AddResultingValue`)... Can you please revert this change as I think it is now worse in terms of readability? Sorry again for asking for the change in the first place. :-(

plugins/common/starlark/starlark.go
essobedo: ok then, I'll add the maps

essobedo:
@srebhan function `AddResultingValue` and related changes have been reverted. A map has been used as requested to store the functions and the arguments, please check again

plugins/common/starlark/starlark.go
```diff
@@ -0,0 +1,180 @@
...
+	rv, err := starlark.Call(s.thread, fn, args, nil)
+	if err != nil {
+		if err, ok := err.(*starlark.EvalError); ok {
+			for _, line := range strings.Split(err.Backtrace(), "\n") {
+				s.Log.Error(line)
+			}
+		} else {
+			s.Log.Error(err.Msg)
+		}
+	}
```
srebhan, line 145: How about checking this on the caller site? This would avoid the necessity of the `//nolint` statements. You can still keep the stacktrack logic in a function to avoid this duplication...

srebhan (commented):
Almost there. I still want to avoid the `//nolint` "magic" when calling `Call` and have a suggestion. Please let me know what you think.

essobedo:
> Almost there. I still want to avoid the `//nolint` "magic" when calling `Call` and have a suggestion. Please let me know what you think.

fixed, please check again

plugins/common/starlark/starlark.go
```diff
@@ -133,7 +133,10 @@ func (s *StarlarkCommon) Call(name string) (starlark.Value, error) {
 	if !ok {
 		return nil, fmt.Errorf("params for function %q do not exist", name)
 	}
-	rv, err := starlark.Call(s.thread, fn, args, nil)
+	return starlark.Call(s.thread, fn, args, nil)
+}
+
+func (s *StarlarkCommon) LogError(err error) {
 	if err != nil {
```
srebhan, line 140: I think this can be removed as you can assume that there is an error.

essobedo: @srebhan fixed

essobedo:
Please note that the build failure is not related to this change

plugins/aggregators/starlark/starlark.go
```diff
@@ -0,0 +1,143 @@
...
+	_, err := s.Call("add")
+	if err != nil {
+		s.LogError(err)
+	}
+}
+
+func (s *Starlark) Push(acc telegraf.Accumulator) {
+	rv, err := s.Call("push")
+	if err != nil {
+		s.LogError(err)
```
srebhan, line 100: Maybe we should use `acc.AddError` here to allow for statistics? Same for the instances below.

plugins/common/starlark/starlark.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+
+	builtins := starlark.StringDict{}
+	builtins["Metric"] = starlark.NewBuiltin("Metric", newMetric)
+	builtins["deepcopy"] = starlark.NewBuiltin("deepcopy", deepcopy)
+	builtins["catch"] = starlark.NewBuiltin("catch", catch)
+	s.addConstants(&builtins)
+
+	program, err := s.sourceProgram(builtins, "")
+
```
srebhan, line 52: Please remove this newline.

plugins/common/starlark/starlark.go
```diff
@@ -0,0 +1,180 @@
...
+}
+
+func (s *StarlarkCommon) GetParameters(name string) (starlark.Tuple, bool) {
+	parameters, found := s.parameters[name]
+	return parameters, found
+}
+
+func (s *StarlarkCommon) AddFunction(name string, params ...starlark.Value) error {
+	globalFn, found := s.globals[name]
+
```
srebhan, line 84: Please remove this newline.

plugins/common/starlark/starlark.go
```diff
@@ -0,0 +1,180 @@
...
+// Add all the constants defined in the plugin as constants of the script
+func (s *StarlarkCommon) addConstants(builtins *starlark.StringDict) {
+	for key, val := range s.Constants {
+		sVal, err := asStarlarkValue(val)
+		if err != nil {
+			s.Log.Errorf("Unsupported type: %T", val)
+		}
+		(*builtins)[key] = sVal
+	}
+}
```
srebhan, line 115: Please return the error here and pass it on to `Init()` which can fail early instead of running with a missing constant
```suggestion
func (s *StarlarkCommon) addConstants(builtins *starlark.StringDict) error {
	for key, val := range s.Constants {
		sVal, err := asStarlarkValue(val)
		if err != nil {
			return fmt.Errorf("converting type %T failed: %v", val, err)
		}
		(*builtins)[key] = sVal
		return nil
	}
}
```

plugins/common/starlark/starlark.go
```diff
@@ -0,0 +1,180 @@
...
+
+func (s *StarlarkCommon) sourceProgram(builtins starlark.StringDict, filename string) (*starlark.Program, error) {
+	if s.Source != "" {
+		_, program, err := starlark.SourceProgram(filename, s.Source, builtins.Has)
+		return program, err
+	}
+	_, program, err := starlark.SourceProgram(s.Script, nil, builtins.Has)
+	return program, err
```
srebhan, line 123:
```suggestion
	var src interface{}
	if s.Source != "" {
		src = s.Source
	}
	_, program, err := starlark.SourceProgram(s.Script, src, builtins.Has)
	return program, err
```

srebhan (commented):
This is beautiful @essobedo! Some more minor comments... :-/

plugins/aggregators/starlark/starlark.go
essobedo: The accumulator is not available in case of the other instances

srebhan (approved):
Perfect @essobedo. Thank you very much for your work!

plugins/processors/starlark/starlark_test.go
```diff
@@ -3387,3 +3319,32 @@ func testLoadFunc(module string, logger telegraf.Logger) (starlark.StringDict, e
 func testNow(thread *starlark.Thread, _ *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
 	return starlarktime.Time(time.Date(2021, 4, 15, 12, 0, 0, 999, time.UTC)), nil
 }
+
+func newStarlarkFromSrouce(source string) *Starlark {
```
reimda, line 3323:
```suggestion
func newStarlarkFromSource(source string) *Starlark {
```

plugins/common/starlark/metric.go
```diff
@@ -61,7 +61,7 @@ func (m *Metric) Truth() starlark.Bool {
 }
 
 func (m *Metric) Hash() (uint32, error) {
-	return 0, errors.New("not hashable")
+	return uint32(m.Unwrap().HashID()), nil
```
reimda, line 64: I don't think this implementation of Hash makes sense. It is based on the metric name and tags, but not the fields. When you hash something you need to include all variables in the struct. Since it doesn't include the fields, it makes the merge.star example a little hard to follow.

We either need to include the fields or leave it as not hashable like before. This will require some changes in merge.star, but those changes will make it make more sense.

reimda (changes requested):
Thanks @essobedo! This PR is getting very close. Could you take a look at how metrics are hashed? If we can resolve that (and a typo I found), I think it's ready to merge.

plugins/processors/starlark/starlark_test.go
essobedo: Good catch!

essobedo:
@reminda Fixed, please check again

srebhan (approved):

reimda (approved):

reimda:
Thanks @essobedo all your work on this!

MyaLongmire:
@essobedo Thank you so much for this pr! There is a leftover `//nolint` comment that is keeping the file from being linted. I removed it in this [pr](https://github.com/influxdata/telegraf/pull/10138) but didn't touch anything else, just wanted to let you know.

PR #9426, opened 2021-06-24
Add new output plugin for Azure Data Explorer(ADX)

### LINKED ISSUE #9230: Telegraf output plugin for Azure Data Explorer
## Feature Request

### Use case:  Lot of large enterprise customers requested for an ADX output plugin so they can ingest data collected using various input plugins directly to ADX. 
Azure Data Explorer is a big data analytics platform, purpose built for logs and telemetry analytics scenarios.

minwal:
### Required for all PRs:

It resolves #9230 

It is a new output plugin for Azure Data Explorer, its been tested with SQL Server input plugin but it can work with any input plugin of Telegraf.


srebhan (changes requested):
@minwal thanks for this marvelous work! 
However, there are some points that need some work. The main items are

1. Remove the ability to specify a serialzer as you only support JSON as far as I can see.
2. Consider my comments on `ingestor` handling.
3. Implementation of `context` for `createAzureDataExplorerTableForNamespace()` and maybe addition of a `timeout` setting.
4. Check if user settings in `Init()`.
5. Rework the test to use `require` and get rid of global variables.
6. Add more tests to test corner and error cases (e.g. invalid user settings, server not reachable,...) or multiple metrics.

The list looks long, but this is a very nice first contribution and I'm pretty sure we get this flying. Thanks again for your contribution, much appreciated.

srebhan:
Oh and please change the PR title and maybe explain a bit the available features in the description. Remember, the PR title is the later commit message in the log. ;-)
Please also take a look at the CI errors. We cannot merge the PR if those fail.
Furthermore, you need to sign the CLA and, if submitting on behalf of a company, the CCLA.

eashi:
Thanks a lot @srebhan for the detailed review.

We have changed most of the review comments. I just need to understand the effect of implementing the context:
>     3. Implementation of `context` for `createAzureDataExplorerTableForNamespace()` and maybe addition of a `timeout` setting.

It seems that the code will still run to the end even if the flush interval is elapsed (I put `time.Sleep` to test it). It's true that we get a warning:
```
W! [agent] ["outputs.azure_data_explorer"] did not complete within its flush interval
```

...but the code still executes without blocking or terminating early. So what is the effect of not implementing the context and the timeout? 

**Note**: the underlying code does have a timeout, but it defaults to 10 minutes.

srebhan (changes requested):
Hey @emadashi-purple and @minwal,

thanks for your quick response. The code looks much better this round, however, there are still some comments left from the last review round. I copy-pasted them over here. It would be nice if you can tackle them.
I especially want to hear your opinion regarding my comment on not tracking the `ingestors` but rather use trash ones. Not sure if this is possible or not.
Furthermore, please get rid of the global variables in the unit test as this might produce side-effects between tests which should be avoided in all circumstances.
You also need to update `docs/LICENSE_OF_DEPENDENCIES.md` with your newly added libs. Please check the CI errors...

srebhan:
First of all thank you for submitting this code.

Regarding the context implementation / timeout handling:

> It seems that the code will still run to the end even if the flush interval is elapsed (I put `time.Sleep` to test it). It's true that we get a warning:
> ```
> W! [agent] ["outputs.azure_data_explorer"] did not complete within its flush interval
> ```
> ...but the code still executes without blocking or terminating early. So what is the effect of not implementing the context and the timeout?

The issue here is that the buffer will fill between the runs. Imaging that you collect say 2,000 metrics per minute and you use the default buffer limit of 10,000. Let's further assume that for the next 10 minutes (your internal default timeout) you cannot flush the data. Then after 5 minutes your data will overflow.
Usually, plugins allow the user to specify a timeout (with a sensible default e.g. 10s) and produce an error. The result might be the same in this case, but it gives the user a certain amount of control over the processing.

> **Note**: the underlying code does have a timeout, but it defaults to 10 minutes.
IMO this is much to large for telegraf. E.g. I have a setup gathering ~2,000 metrics per minute and I do expect a data-sink to ingest that data withing a minute (TBH much less :-)) or so considering an error otherwise.

So my suggestion is to add a 
```golang
Timeout config.Duration `toml:"timeout"`
```
option to your struct and use that timeout in a context for `Write()`.

plugins/outputs/azure_data_explorer/azure_data_explorer.go
```diff
@@ -0,0 +1,207 @@
...
+			if err != nil {
+				return err
+			}
+
+			//create a new ingestor client for the namespace
+			adx.ingesters[namespace], err = createIngestor(adx.client, adx.Database, namespace)
+			if err != nil {
+				return err
+			}
+		}
```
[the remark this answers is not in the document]
emadashi-purple: We need to group the metrics per unique metric name and then send them in batches to Azure Data Explorer.
If I understood you correct, if we put them in one loop this means that we have to send each metric individually which is prohibitive due to performance and Kusto SDK limitation.

Is my understanding of your suggestion correct? Is there a better way?

plugins/outputs/azure_data_explorer/azure_data_explorer_test.go
```diff
@@ -0,0 +1,152 @@
...
+				client:         tC.client,
+				ingesters:      map[string]localIngestor{},
+				createIngestor: tC.createIngestor,
+				serializer:     serializer,
+			}
+
+			errorInWrite := plugin.Write(testutil.MockMetrics())
+
+			if tC.expected["error"] != "" {
+				require.EqualError(t, errorInWrite, tC.expected["error"].(string))
```
srebhan, line 89: You might want to move `error` to an own `string` field in your test-table to avoid the type assertion and simplify code.

plugins/outputs/azure_data_explorer/azure_data_explorer_test.go
```diff
@@ -0,0 +1,152 @@
...
+				require.Equal(t, createTableString, tC.client.queries[0])
+
+				createTableMappingString := fmt.Sprintf(createTableMappingCommandExpected, expectedNameOfMetric, expectedNameOfMetric)
+				require.Equal(t, createTableMappingString, tC.client.queries[1])
+			}
+		})
+	}
+}
+
+func TestInirBlankEndpoint(t *testing.T) {
```
srebhan, line 109:
```suggestion
func TestInitBlankEndpoint(t *testing.T) {
```

plugins/outputs/azure_data_explorer/azure_data_explorer_test.go
```diff
@@ -0,0 +1,152 @@
...
+	plugin := AzureDataExplorer{
+		Endpoint:       "",
+		Database:       "",
+		ClientID:       "",
+		ClientSecret:   "",
+		TenantID:       "",
```
srebhan, line 115: Remove those lines as this is the default value anyway.

srebhan (commented):
Nice progress here.
I still want to here your opinion about moving the `ingestor` creation one for-loop down. Is it really necessary to track those ingestors over multiple writes? If not there is a huge potential to simplify the code further.
Other than that only very minor comments remain. One testing left-over and removal of dead code for example. Please walk through the comments.

Thanks again for working on this!

plugins/outputs/azure_data_explorer/azure_data_explorer.go
```diff
@@ -0,0 +1,205 @@
+package azure_data_explorer
+
+// simpleoutput.go
```
srebhan, line 3: Please also remove.

plugins/outputs/azure_data_explorer/azure_data_explorer.go
```diff
@@ -0,0 +1,205 @@
...
+		reader := bytes.NewReader(mPerNamespace)
+
+		time.Sleep(20 * time.Second)
+		_, errorIngesting := adx.ingesters[key].FromReader(context.TODO(), reader, ingest.FileFormat(ingest.JSON), ingest.IngestionMappingRef(fmt.Sprintf("%s_mapping", key), ingest.JSON))
+		if errorIngesting != nil {
+			adx.Log.Errorf("sending ingestion request to Azure Data Explorer for metric %q failed: %v", key, errorIngesting)
+		}
+	}
+	return nil
+}
```
srebhan, line 143: Here my suggestion regarding the ingestion handling:
```suggestion
func (adx *AzureDataExplorer) Write(metrics []telegraf.Metric) error {
	metricsPerNamespace := make(map[string][]byte)

	// Group metrics by name and serialize them
	for _, m := range metrics {
		namespace := m.Name()
		metricInBytes, err := adx.serializer.Serialize(m)
		if err != nil {
			return err
		}

		if existingBytes, ok := metricsPerNamespace[namespace]; ok {
			metricsPerNamespace[namespace] = append(existingBytes, metricInBytes...)
		} else {
			metricsPerNamespace[namespace] = metricInBytes
		}
	}

	ctx := context.Background()
	ctx, cancel := context.WithTimeout(ctx, time.Duration(adx.Timeout))
	defer cancel()

	// Push the metrics namespace-wise
	format := ingest.FileFormat(ingest.JSON)
	for namespace, mPerNamespace := range metricsPerNamespace {
		reader := bytes.NewReader(mPerNamespace)

		//create a table for the namespace
		if err := createAzureDataExplorerTableForNamespace(ctx, adx.client, adx.Database, namespace); err != nil {
			return fmt.Errorf("creating table for %q failed: %v", namespace, err)
		}

		//create a new ingestor client for the namespace
		ingestor, err := adx.createIngestor(adx.client, adx.Database, namespace)
		if err != nil {
			return fmt.Errorf("creating ingestor for %q failed: %v", namespace, err)
		}

		mapping := ingest.IngestionMappingRef(fmt.Sprintf("%s_mapping", namespace), ingest.JSON)
		if _, err := ingestor.FromReader(ctx, reader, format, mapping); err != nil {
			adx.Log.Errorf("sending ingestion request to Azure Data Explorer for metric %q failed: %v", namespace, err)
		}
	}

	return nil
}
```

plugins/outputs/azure_data_explorer/azure_data_explorer.go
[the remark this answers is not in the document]
srebhan: I added my suggestion of the write method to the code to illustrate my idea. Please take a look. :-)

emadashi-purple:
@srebhan This was an amazing review! thanks a lot for all the comments and suggestions.

I have changed the loop so it's closer to what you have suggested, but kept the ingestors life span :D.
The reason is performance: when I asked the Kusto SDK team they said that reusing the ingestors is much better for performance and they suggested to keep them alive and reuse them as long as possible.

plugins/outputs/azure_data_explorer/azure_data_explorer.go
```diff
@@ -0,0 +1,196 @@
...
+  client_id = ""
+
+  ## The client secret of the Service Principal in Azure that has ingestion rights to the Azure Data Exploer Cluster
+  client_secret = ""
+
+  ## The tenant ID of the Azure Subsciption in which the Service Principal belongs to
+  tenant_id = ""
+
+  ## Timeout for Azure Data Explorer operations
+  # timeout = "15s"
```
srebhan, line 72: Please also update `README.md`.

plugins/outputs/azure_data_explorer/azure_data_explorer.go
```diff
@@ -0,0 +1,196 @@
...
+
+		mapping := ingest.IngestionMappingRef(fmt.Sprintf("%s_mapping", namespace), ingest.JSON)
+		if _, err := adx.ingesters[namespace].FromReader(ctx, reader, format, mapping); err != nil {
+			adx.Log.Errorf("sending ingestion request to Azure Data Explorer for metric %q failed: %v", namespace, err)
+		}
+	}
+	return nil
+}
+
+func createAzureDataExplorerTableForNamespace(ctx context.Context, client localClient, database string, tableName string) error {
```
srebhan, line 144: Would it make sense to make this a member of `(adx *AzureDataExplorer)`? This way you don't need to pass `client` and `database` as parameters.

plugins/outputs/azure_data_explorer/azure_data_explorer_test.go
```diff
@@ -0,0 +1,147 @@
...
+				"createTableCommand":        "",
+				"createTableMappingCommand": "",
+			},
+			expectedWriteError: "creating table for \"test1\" failed: Something went wrong",
+		},
+	}
+
+	for _, tC := range testCases {
+		t.Run(tC.name, func(t *testing.T) {
+			serializer, _ := telegrafJson.NewSerializer(time.Second)
```
srebhan, line 71:
```suggestion
			serializer, err := telegrafJson.NewSerializer(time.Second)
			require.NoError(t, err)
```

plugins/outputs/azure_data_explorer/azure_data_explorer_test.go
```diff
@@ -0,0 +1,147 @@
...
+
+			errorInWrite := plugin.Write(testutil.MockMetrics())
+
+			if tC.expectedWriteError != "" {
+				require.EqualError(t, errorInWrite, tC.expectedWriteError)
+			} else {
+				require.NoError(t, errorInWrite)
+
+				expectedNameOfMetric := tC.expected["metricName"]
+				createdIngestor := plugin.ingesters[tC.expected["metricName"].(string)]
```
srebhan, line 94: How about
```suggestion
				expectedNameOfMetric := tC.expected["metricName"].(string)
				createdIngestor := plugin.ingesters[expectedNameOfMetric]
```

srebhan (commented):
Nice job! There are only small cosmetic things left. Please make sure to update `README.md` with the `timeout` option.
Furthermore, it would be nice if you could add
>The reason [for keeping the ingestors over writes] is performance: when I asked the Kusto SDK team they said that reusing the ingestors is much better for performance and they suggested to keep them alive and reuse them as long as possible.

as a comment to the code for future readers ensuring we keep the performance. ;-)

plugins/outputs/azure_data_explorer/azure_data_explorer.go
emadashi-purple: I usually try to pass thin parameters to my functions to give flexibility for testing and make the dependencies clear. This is part of this habit.
Nonetheless, I have changed the code according to the suggestion :).

minwal:
@srebhan - Thanks for your awesome review comments. It helped us in refining our code. Are there any further comments? FYI - there are lot of folks waiting for this PR to be approved:) Also, may I know when will be the next Telegraf release?

srebhan (approved):
Looks good to me. Thanks @minwal and @emadashi-purple for the productive collaboration and for the positive review feedback! ;-)

Regarding release I cannot say. **Usually** if your plugin gets merged it will be in v1.20. Telegraf **usually** has a 3 month release cycle which would make it Mid of September.

@sjwang90 can you maybe comment on the next planned release date?

sjwang90:
@minwal We target releases on a quarterly basis. Our next Telegraf 1.20 should occur around mid-September.

docs/LICENSE_OF_DEPENDENCIES.md
```diff
@@ -138,6 +139,7 @@ following works:
 - github.com/karrick/godirwalk [BSD 2-Clause "Simplified" License](https://github.com/karrick/godirwalk/blob/master/LICENSE)
 - github.com/kballard/go-shellquote [MIT License](https://github.com/kballard/go-shellquote/blob/master/LICENSE)
 - github.com/klauspost/compress [BSD 3-Clause Clear License](https://github.com/klauspost/compress/blob/master/LICENSE)
+- github.com/kylelemons/godebug [Apache License](https://github.com/kylelemons/godebug/blob/master/LICENSE)
```
ssoroka, line 142: Looks like we can remove this as it's not part of the project.

plugins/outputs/azure_data_explorer/azure_data_explorer.go
```diff
@@ -0,0 +1,200 @@
...
+			return err
+		}
+		if existingBytes, ok := metricsPerNamespace[namespace]; ok {
+			metricsPerNamespace[namespace] = append(existingBytes, metricInBytes...)
+		} else {
+			metricsPerNamespace[namespace] = metricInBytes
+		}
+	}
+	ctx := context.Background()
+	ctx, cancel := context.WithTimeout(ctx, time.Duration(adx.Timeout))
```
ssoroka, line 120: should this timeout be longer if there's a lot of namespaces? seems likely it could not submit in time if there's a lot of unique measurements

plugins/outputs/azure_data_explorer/README.md
```diff
@@ -0,0 +1,63 @@
...
+```
+.create-merge table ['table-name']  (['fields']:dynamic, ['name']:string, ['tags']:dynamic, ['timestamp']:datetime)
+```
+
+The corresponding table mapping would be like the following:
+```
+.create-or-alter table ['table-name'] ingestion json mapping 'table-name_mapping' '[{"column":"fields", "Properties":{"Path":"$[\'fields\']"}},{"column":"name", "Properties":{"Path":"$[\'name\']"}},{"column":"tags", "Properties":{"Path":"$[\'tags\']"}},{"column":"timestamp", "Properties":{"Path":"$[\'timestamp\']"}}]'
+```
+
+**Note**: This plugin will automatically create Azure Data Explorer tables and corresponding table mapping as per the above mentioned commands. Since the `Metric` object is a complex type, the only output format supported is JSON, so make sure to set the `data_format` configuration in `telegraf.conf` to `json`.
```
ssoroka, line 57: Since you're creating the serializer yourself manually, you don't need to tell users to set the serializer to json.

plugins/outputs/azure_data_explorer/azure_data_explorer_test.go
```diff
@@ -0,0 +1,148 @@
...
+	testCases := []struct {
+		name               string
+		inputMetric        []telegraf.Metric
+		client             *fakeClient
+		createIngestor     ingestorFactory
+		expected           map[string]interface{}
+		expectedWriteError string
+	}{
+		{
+			name:        "Valid metric",
```
ssoroka, line 34: looks like you could add a test that verifies that fields and tags and timestamps are serialized properly.

ssoroka (changes requested):

plugins/outputs/azure_data_explorer/azure_data_explorer.go
yjlim214: Just increased the default timeout to 20 seconds according to your comment. In all cases, this is a config value and the user should be able to update it if necessary.

docs/LICENSE_OF_DEPENDENCIES.md
emadashi-purple: But we use it as an eventual dependency, if I run the command `go mod why github.com/kylelemons/godebug/pretty` I will have the result:
```
# github.com/kylelemons/godebug/pretty
github.com/influxdata/telegraf/plugins/outputs/azure_data_explorer
github.com/Azure/azure-kusto-go/kusto/ingest
github.com/kylelemons/godebug/pretty
```

And if I run the command `make check-deps` when I remove the line 142, I get the result:
```
./scripts/check-deps.sh
--- /var/folders/km/r01j6n215ld0_ps3k_x9h08w0000gn/T/tmp.4MvOyb57/LICENSE_OF_DEPENDENCIES.md    2021-07-07 12:59:28.000000000 +1000
+++ /var/folders/km/r01j6n215ld0_ps3k_x9h08w0000gn/T/tmp.4MvOyb57/HEAD  2021-07-07 12:59:28.000000000 +1000
@@ -139,0 +140 @@
+github.com/kylelemons/godebug


The docs/LICENSE_OF_DEPENDENCIES.md file does not contain the expected entries.

Lines prefixed with '+' should be added to LICENSE_OF_DEPENDENCIES.md and '-'
lines should be removed.

Include a link to the appropriate licenses for any additions.
make: *** [check-deps] Error 1
```

minwal:
As suggested by @sjwang90, adding new feature in plugin in the same PR. This feature allows to add all the metrics into a single Azure Data Explorer(ADX) table rather than creating separate ADX table for each metric.

srebhan (approved):
Still fine with me.

ssoroka (approved):

PR #9433, opened 2021-06-24
Remove experimental snowflake support

### LINKED ISSUE #9407: Startup failure with caching on 1.19
### Relevant telegraf.conf:

```toml
  
# Configuration for influxdb server to send metrics to
[[outputs.influxdb]]
  ## The full HTTP or UDP endpoint URL for your InfluxDB instance.
  ## Multiple urls can be specified as part of the same cluster,
  ## this means that only ONE of the urls will be written to each interval.
  # urls = ["udp://localhost:8089"] # UDP endpoint example
  urls = ["http://apm-internal.ravenslingshot.com:8086"] # required
  ## The target database for metrics (telegraf will create it if not exists).
  database = "metrics" # required
  
  ## Retention policy to write to. Empty string writes to the default rp.
  retention_policy = ""
  ## Write consistency (clusters only), can be: "any", "one", "quorum", "all"
  write_consistency = "any"
  
  ## Write timeout (for the InfluxDB client), formatted as a string.
  ## If not provided, will default to 5s. 0s means no timeout (not recommended).
  timeout = "5s"
  username=""
  password=""

  ## Set the user agent for HTTP POSTs (can be useful for log differentiation)
  # user_agent = "telegraf"
  ## Set UDP payload size, defaults to InfluxDB UDP Client default (512 bytes)
  # udp_payload = 512
  
  ## Optional SSL Config
  # ssl_ca = "/etc/telegraf/ca.pem"
  # ssl_cert = "/etc/telegraf/cert.pem"
  # ssl_key = "/etc/telegraf/key.pem"
  ## Use SSL but skip chain & host verification
  
  
#                            PROCESSOR PLUGINS                                #


# # Print all metrics that pass through this filter.
# [[processors.printer]]



#                            AGGREGATOR PLUGINS                               #

# # Keep the aggregate min/max of each metric passing through.
# [[aggregators.minmax]]
#   ## General Aggregator Arguments:
#   ## The period on which to flush & clear the aggregator.
#   period = "30s"
#   ## If true, the original metric will be dropped by the
#   ## aggregator and will not get sent to the output plugins.
#   drop_original = false



#                            INPUT PLUGINS                                    #

[[inputs.cpu]]
  percpu = true
  totalcpu = true
  collect_cpu_time = false
  
[[inputs.system]]
  # no configuration
  
# Read metrics about swap memory usage
[[inputs.swap]]
  # no configuration

# Read metrics about memory usage
[[inputs.mem]]
  # no configuration

[[inputs.net]]
  interfaces = ["ens5"]

# Statsd Server
[[inputs.statsd]]
  ## Address and port to host UDP listener on
  service_address = ":8125"
  
  ## The following configuration options control when telegraf clears it's cache
  ## of previous values. If set to false, then telegraf will only clear it's
  ## cache when the daemon is restarted.
  ## Reset gauges every interval (default=true)
  delete_gauges = true
  ## Reset counters every interval (default=true)
  delete_counters = true
  ## Reset sets every interval (default=true)
  delete_sets = true
  ## Reset timings & histograms every interval (default=true)
  delete_timings = true
  
  ## Percentiles to calculate for timing & histogram stats
  percentiles = [90]
  
  ## separator to use between elements of a statsd metric
  metric_separator = "."
  
  ## Parses datadog extensions to the statsd format
  datadog_extensions = true

  ## Statsd data translation templates, more info can be read here:
  ## https://github.com/influxdata/telegraf/blob/master/docs/DATA_FORMATS_INPUT.md#graphite
  # templates = [
  #     "cpu.* measurement*"
  # ]

  ## Number of UDP messages allowed to queue up, once filled,
  ## the statsd server will start dropping packets
  
  ## Number of timing/histogram values to track per-measurement in the
  ## calculation of percentiles. Raising this limit increases the accuracy
  ## of percentiles but also increases the memory usage and cpu time.
  percentile_limit = 1000
  
# Generic socket listener capable of handling multiple socket types.
[[inputs.socket_listener]]
  ## URL to listen on
  service_address = "udp://:8126"
  
  ## Maximum socket buffer size (in bytes when no unit specified).
  ## For stream sockets, once the buffer fills up, the sender will start backing up.
  ## For datagram sockets, once the buffer fills up, metrics will start dropping.
  ## Defaults to the OS default.
  read_buffer_size = "16MiB"
    
  ## Data format to consume.
  ## more about them here:
  data_format = "influx"
  
  tag_keys = [
    "server",
    "file_transfer_state",
    "file_transfer_error"
  ] 
```

### System info:

Telegraf 1.19, Ubuntu 18.04

### Steps to reproduce:

1. ... Install telegraf 1.18 with above config using influx apt repository
2. ... Things work
3. Upgrade to telegraf 1.9
4. Look for error in /var/log/syslog

### Expected behavior:

Telegraf should continue to run as it did

### Actual behavior:

failed to open. Ignored. open /etc/telegraf/.cache/snowflake/ocsp_response_cache.json: no such file or directory\n" func="gosnowflake.(*defaultLogger).Errorf" file="log.go:120"

Reverting to 1.18 fixes the problem.

reimda:
resolves #9407

Since the Snowflake db driver requires a writable user home directory and the linux packages use a telegraf user with a read-only home directory, snowflake support breaks telegraf for linux users using the packages.

This PR removes the experimental Snowflake support to quickly fix the problem and reduce the issue impact. We expect to add snowflake support back in the future.


go.mod
```diff
@@ -9,6 +9,7 @@ require (
 	code.cloudfoundry.org/clock v1.0.0 // indirect
 	collectd.org v0.5.0
 	github.com/Azure/azure-event-hubs-go/v3 v3.2.0
+	github.com/Azure/azure-pipeline-go v0.2.3 // indirect
```
srebhan, line 12: Where does this come from?

srebhan (approved):
The change looks good. I'm only curious on why a (indirect) dependency to `azure-pipeline-go` is added in `go.mod`. :-/

go.mod
reimda: go mod tidy added it. I'm not sure why

srebhan: Strange. I can live with this change if you can. ;-)

srebhan (approved):

PR #9447, opened 2021-06-29
fixed percentiles not being able to be ints

### LINKED ISSUE #9441: Statsd percentile setting fails on integer values
Before 1.19.0, the statsd percentile setting would accept integer or floating point values. After 1.19.0, it doesn't accept integer values.

This config snippet was valid before 1.19.0:
```toml
[[inputs.statsd]]
  percentiles = [90]
```

On 1.19.0 it will trigger an error like this:
`Error loading config file /etc/telegraf/telegraf.conf: error parsing statsd, line 186: (statsd.Statsd.Percentiles) cannot unmarshal TOML integer into float64`

Since users of the statsd with percentiles before 1.19.0 may have entered percentiles as integers, it's likely this will cause a failure when they upgrade to 1.19.0. (see #9407)

The setting parsing changed in #8969 as part of a cleanup refactor. 

Expected fix: make the setting work like before and accept integer values as well as floating point values.

As a workaround, users can change their telegraf.conf to have floating point values for this setting. The failing config above works with ".0" added to the setting:
```toml
[[inputs.statsd]]
  percentiles = [90.0]
```

MyaLongmire:
### Required for all PRs:

resolves #9441 

added a Number struct and an unmarshal that will allow for ints or floats

### DIFF
--- a/config/types.go
+++ b/config/types.go
@@ -14,6 +14,9 @@ type Duration time.Duration
 // Size is an int64
 type Size int64
 
+// Number will get parsed as an int or float depending on what is passed
+type Number float64
+
 // UnmarshalTOML parses the duration from the TOML config file
 func (d *Duration) UnmarshalTOML(b []byte) error {
 	var err error
@@ -85,4 +88,14 @@ func (s *Size) UnmarshalTOML(b []byte) error {
 
 func (s *Size) UnmarshalText(text []byte) error {
 	return s.UnmarshalTOML(text)
+}
+
+func (n *Number) UnmarshalTOML(b []byte) error {
+	value, err := strconv.ParseFloat(string(b), 64)
+	if err != nil {
+		return err
+	}
+
+	*n = Number(value)
+	return nil
 }
--- a/config/types.go
+++ b/config/types.go
@@ -14,9 +14,6 @@ type Duration time.Duration
 // Size is an int64
 type Size int64
 
-// Number will get parsed as an int or float depending on what is passed
-type Number float64
-
 // UnmarshalTOML parses the duration from the TOML config file
 func (d *Duration) UnmarshalTOML(b []byte) error {
 	var err error
@@ -88,14 +85,4 @@ func (s *Size) UnmarshalTOML(b []byte) error {
 
 func (s *Size) UnmarshalText(text []byte) error {
 	return s.UnmarshalTOML(text)
-}
-
-func (n *Number) UnmarshalTOML(b []byte) error {
-	value, err := strconv.ParseFloat(string(b), 64)
-	if err != nil {
-		return err
-	}
-
-	*n = Number(value)
-	return nil
 }
--- a/plugins/inputs/statsd/statsd.go
+++ b/plugins/inputs/statsd/statsd.go
@@ -37,6 +37,20 @@ const (
 
 var errParsing = errors.New("error parsing statsd line")
 
+type Number struct {
+	Value float64
+}
+
+func (n *Number) UnmarshalTOML(b []byte) error {
+	value, err := strconv.ParseFloat(string(b), 64)
+	if err != nil {
+		return err
+	}
+
+	n.Value = value
+	return nil
+}
+
 // Statsd allows the importing of statsd and dogstatsd data.
 type Statsd struct {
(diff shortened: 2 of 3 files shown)


sspaink (dismissed):
Good work, overall it looks good but can you add another unit test that reads from a toml config with a int instead of a float to verify this change works? Something like this:
```
func TestConfig(t *testing.T) {
        // Load a sample toml config
	c := config.NewConfig()
	require.NoError(t, c.LoadConfig("./testdata/statsd_int.toml"))
        // Create a statsd input plugin struct and load it with expected values
	s := inputs.Inputs["statsd"]().(*Statsd)
	s.Percentiles = 90.0
	require.Equal(t, s, c.Inputs[0].Input)
}
```

srebhan:
Hey @MyaLongmire, thanks for looking into this. IMO the problem here is an issue in the TOML parser as `90` is a perfect `float64` value and should be parsed as such by the parser. So TBH I rather prefer to fix [this function](https://github.com/influxdata/toml/blob/2a2e3012f7cfbef64091cc79776311e65dfa211b/decode.go#L273) instead of circumventing the problem by doing the parsing manually.

MyaLongmire:
> Good work, overall it looks good but can you add another unit test that reads from a toml config with a int instead of a float to verify this change works? Something like this:
> 
> ```
> func TestConfig(t *testing.T) {
>         // Load a sample toml config
> 	c := config.NewConfig()
> 	require.NoError(t, c.LoadConfig("./testdata/statsd_int.toml"))
>         // Create a statsd input plugin struct and load it with expected values
> 	s := inputs.Inputs["statsd"]().(*Statsd)
> 	s.Percentiles = 90.0
> 	require.Equal(t, s, c.Inputs[0].Input)
> }
> ```

I added a test that ensures ints can be added in the config, although I changed your idea so I wouldn't have to add a test data folder

MyaLongmire:
> Hey @MyaLongmire, thanks for looking into this. IMO the problem here is an issue in the TOML parser as `90` is a perfect `float64` value and should be parsed as such by the parser. So TBH I rather prefer to fix [this function](https://github.com/influxdata/toml/blob/2a2e3012f7cfbef64091cc79776311e65dfa211b/decode.go#L273) instead of circumventing the problem by doing the parsing manually.
> What do you think?

How would we know that 90 is supposed to be a float while parsing it? I feel like putting it in config copies how it used to be in internal.

srebhan:
> How would we know that 90 is supposed to be a float while parsing it?
The TOML parser has the struct it unmarshalls to. If you look at the function I linked, `fv reflect.Value` is the reference to the struct field the parser should unmarshal to. In the big `switch fv.Kind() {` statement below, the parser converts the _known_ TOML Integer value to the _known_ field type. It's only that it does not take "converting an int to a float" into account. We just would need to add
```golang
	case reflect.Float32, reflect.Float64:
		f := float64(i)
		if fv.OverflowFloat(f) {
			return &errorOutOfRange{fv.Kind(), f}
		}
		fv.SetFloat(f)
```

> I feel like putting it in config copies how it used to be in internal.
Right, and you can do this as it restores the functionality. However, it is a workaround and the problem occurs for _any_ `float64` field exposed as a configuration option. So you would need to convert _all_ of those fields to `config.Number`... :-S

reimda:
@srebhan @MyaLongmire I don't think we want to make this change apply to all float64 fields in the config. TOML's design intentionally has a separate float and int type so making a toml int field initialize a telegraf float setting feels wrong. I think this change should be a very isolated workaround only for the statsd plugin. Statsd is special only for backward compatibility.

I'd prefer not to put Number in config/types.go because I don't want other plugins to do this in the future. What do you think about moving Number from config/ into plugins/inputs/statsd/ and putting in a comment saying it is for backward compatibility and asking people not to do the same thing elsewhere?

srebhan:
@reimda: Fine with me.

In general I think using a TOML int as float is a valid use-case. Maybe the parser could have a `strict_typing` setting, but using `90` as a float is perfectly valid. Using `90.0` as int is not. :-)

srebhan (approved):
Perfect. Thanks @MyaLongmire for fixing this one!

sspaink (approved):

